Issue #4605, opened 2021-12-21
Decouple Service from components


bogdandrutu:
Right now the `Service` (configuration, logic) is very connected with the rest of components. There are use-cases where we need to allow users to consume the otel components in a different "Service" or a different way to define the service configuration. To achieve this we should move the `config.Config` and `config.Service` under the `service` package and allow components to not depend on anything service related.

alolita:
Hi @bogdandrutu what is the reasoning behind these changes? Are you building a new service package? Is there a design document you can share with the rest of us? It would be good for maintainers to also do design docs for changes which other maintainers can also have a chance to review and provide feedback on.

alolita:
Also, I thought we had a project policy of contributors from one org having reviews for their changes from approvers from other orgs (not all from the same org) for all components on OTEL - Collector included. Has this policy changed? Just wondering.

cc: @open-telemetry/governance-committee

dyladan:
> Also, I thought we had a project policy of contributors from one org having reviews for their changes from approvers from other orgs (not all from the same org) for all components on OTEL - Collector included. Has this policy changed? Just wondering.
> 
> cc: @open-telemetry/governance-committee

I think that policy only applied to oteps/specification. For many SIGs it's not viable to enforce such a rule with the limited number of active approvers.

bogdandrutu:
@alolita here are the answers:

> what is the reasoning behind these changes?

The reason as explained in the issue is to decouple the service definition from the components (e.g. otlp receiver). As it is logical, the service uses components not the other way around.

> Are you building a new service package?

No. But we want to not limit ourself in the future. 

> Is there a design document you can share with the rest of us?

As the issue says this needs only one struct move, so it is a pretty simple change.

jpkrohling:
> There are use-cases where we need to allow users to consume the otel components in a different "Service" or a different way to define the service configuration.

I'm not opposed to this change, but could you give a concrete example of the above? It would certainly help making the case for this change.

bogdandrutu:
@jpkrohling internally @tigrannajaryan was looking to build an internal service based on these components, he can tell you more about the status of that.

Also, it does not cost us too much (almost nothing) to keep this door open.

Aneurysm9:
I would like to know more about this change as well.  Without an understanding of what the goal is and the use cases for it, it is hard to evaluate whether the change is good or correct.

> internally @tigrannajaryan was looking to build an internal service based on these components

[...]

PR #4608, opened 2021-12-21
Deprecate `config.Config` and `config.Service`, use `service.Config*`

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4605

Followup PRs:
* Move service/internal/configunmarshaler as private struct into service/ to avoid circular dependency (no public API change)
* Change service/internal/builders to private structs in service/ or don't pass config, instead only pass the components and pipeline from the config.

While doing this, I realize that the "service" section in our config is a bit useless, we can think of promoting everything under "service::" one level up.


service/serviceconfig/service.go
```diff
@@ -114,32 +120,16 @@ type ServiceTelemetryMetrics struct {
 	Address string `mapstructure:"address"`
 }
 
-func (s ServiceTelemetryMetrics) validate() error {
+func (s ServiceTelemetryMetrics) Validate() error {
```
Aneurysm9, line 123: Do these need to be exported?  I don't see them used anywhere else.

bogdandrutu: Done.

Aneurysm9:
I don't think my questions and concerns from #4605 have been addressed.  I still don't know what the plan is here.

codeboten (commented):
After reviewing the code in the PR I think the only question i have is the same as @Aneurysm9 mentioned before around what the plan to decouple `configunmarshaler` is

bogdandrutu:
Converting this to draft, to block myself on merging, since we should do the backwards compatible change.

config/moved_service.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 // Service defines the configurable components of the service.
+// Deprecated: Use serviceconfig.Service
```
codeboten, line 24: The deprecation message needs a version number

codeboten (commented):
>The reason is that in the "config" package after this change, there is no concept of "Service" so that packages offers utils for what it provides. The rest of the ConfigUnmarshaler will leave somewhere in the service package.

Right, in that case any reason not to move it as part of this change? This way we at least avoid introducing a dependency between on `serviceconfig` from a `config` package.

bogdandrutu:
> This way we at least avoid introducing a dependency between on serviceconfig from a config package.

You will still have that since we need to deprecate things, unless we are willing to copy the code.

config/moved_service.go
bogdandrutu: Done.

bogdandrutu:
> Right, in that case any reason not to move it as part of this change? This way we at least avoid introducing a dependency between on serviceconfig from a config package.

After some time, I got to the state you asked for. We don't have any circular dependency with `config` since the unmarshaler is now internal in the service. There is still a followup as documented in the description, but that is easy and non breaking to solve.

config/moved_service.go
```diff
@@ -110,28 +114,13 @@ type ServiceTelemetryMetrics struct {
...
+// Deprecated: [v0.52.0] Use service.ConfigServicePipeline
 type Pipeline struct {
 	Receivers  []ComponentID `mapstructure:"receivers"`
 	Processors []ComponentID `mapstructure:"processors"`
 	Exporters  []ComponentID `mapstructure:"exporters"`
 }
 
-// Pipelines is a map of names to Pipelines.
-type Pipelines map[ComponentID]*Pipeline
+// Deprecated: will be removed soon.
```
codeboten, line 125: Missing version

codeboten (approved):
Just one comment, otherwise the change looks good.

config/moved_service.go
bogdandrutu: Done.

Issue #4623, opened 2022-01-03
configgrpc needs to be refactored in accordance to the change in confighttp


hyunuk:
**Is your feature request related to a problem? Please describe.**

https://github.com/open-telemetry/opentelemetry-collector/pull/4441#discussion_r761532499
> It can be shareable. I removed usages of these values in confighttp and made connection to middleware.CompressionType.
To connect these values with configgrpc, I discussed with @Aneurysm9 and since the changes in confighttp are quite a lot, he will create an issue about refactoring configgrpc after this PR is being merged.

Since the PR for `confighttp` introduced `CompressionType` that can be shareable with grpc, `configgrpc` part needs to be refactored.

**Describe the solution you'd like**
use `middleware.CompressionType` instead of it's own constants.

PR #4624, opened 2022-01-03
Refactor configgrpc for compression methods

hyunuk:
**Description:**
- use `middleware.CompressionType` instead of grpc's own constants.
- Not to use a map to validate compression keys.
- added tests for each compression methods.
- throw an error for zlib and deflate, which only be supported in confighttp

**Link to tracking Issue:**
#4623

### DIFF
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -44,24 +44,7 @@ import (
 	"go.opentelemetry.io/collector/internal/middleware"
 )
 
-// Compression gRPC keys for supported compression types within collector.
-const (
-	CompressionUnsupported = ""
-	CompressionGzip        = "gzip"
-	CompressionSnappy      = "snappy"
-	CompressionZstd        = "zstd"
-)
-
-var (
-	// Map of opentelemetry compression types to grpc registered compression types.
-	gRPCCompressionKeyMap = map[string]string{
-		CompressionGzip:   gzip.Name,
-		CompressionSnappy: snappy.Name,
-		CompressionZstd:   zstd.Name,
-	}
-
-	errMetadataNotFound = errors.New("no request metadata found")
-)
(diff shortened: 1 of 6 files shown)


config/configgrpc/configgrpc.go
```diff
@@ -373,14 +356,18 @@ func (gss *GRPCServerSettings) ToServerOption(host component.Host, settings comp
...
 }
 
-// GetGRPCCompressionKey returns the grpc registered compression key if the
-// passed in compression key is supported, and CompressionUnsupported otherwise.
-func GetGRPCCompressionKey(compressionType string) string {
-	compressionKey := strings.ToLower(compressionType)
-	if encodingKey, ok := gRPCCompressionKeyMap[compressionKey]; ok {
-		return encodingKey
+// GetGRPCCompressionName returns compression name registered in grpc.
+func GetGRPCCompressionName(compressionType middleware.CompressionType) (string, error) {
```
bogdandrutu, line 360: Why is this public? Can we make this private?

Aneurysm9 (approved):

config/configgrpc/configgrpc.go
```diff
@@ -83,7 +66,7 @@ type GRPCClientSettings struct {
 	Endpoint string `mapstructure:"endpoint"`
 
 	// The compression key for supported compression types within collector.
-	Compression string `mapstructure:"compression"`
+	Compression middleware.CompressionType `mapstructure:"compression"`
```
bogdandrutu, line 69: There is a problem with this, similar in http case. We are exposing publicly an "internal" type (under internal/** path). We need to move `middleware.CompressionType` in a public package. We can do this in a separate PR, but I want you to have a tracking issue.

hyunuk: Posted a new issue #4627

bogdandrutu (approved):

Issue #4565, opened 2021-12-17
Have a way to disable the legacy HTTP endpoint


rgl:
**Describe the bug**

There is no way to disable the legacy HTTP endpoint.

This behavior [seems to be hardcoded without a way to disable](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/otlp.go#L146-L157).

**Steps to reproduce**

When using this library from Tempo, I've explicitly set the HTTP endpoint as:

```yaml
distributor:
  receivers:
    # see https://github.com/grafana/tempo/blob/v1.2.1/modules/distributor/config.go#L33
    # see https://github.com/open-telemetry/opentelemetry-collector/tree/main/receiver/otlpreceiver
    otlp:
      protocols:
        http:
          endpoint: 0.0.0.0:4318
```

**What did you expect to see?**

Only port 4318 listening.

**What did you see instead?**

It also opens the legacy HTTP listener at port 55681.

**What version did you use?**

I'm using it tru Tempo 1.2.1. I think its using v0.21.0:

* https://github.com/grafana/tempo/blob/v1.2.1/go.mod#L57
* https://github.com/grafana/tempo/blob/v1.2.1/go.sum#L2065-L2066

Also note that the latest collector version (v0.41.0) also uses the same code, so I guess it also has the same behavior.

**What config did you use?**

See the description.

**Environment**

OS: Ubuntu 20.04
Compiler(if manually compiled): N/A.

**Additional context**

Originally asked for help at https://github.com/grafana/tempo/discussions/1181

jpkrohling:
Interesting, I thought we enabled it only when no values were explicitly provided. I think it might be time to remove this anyway... Thoughts from other approvers and maintainers?

cc @open-telemetry/collector-maintainers , @open-telemetry/collector-approvers

tigrannajaryan:
Let's remove legacy ports completely. I think we kept it in grace period for long enough.

PR #4631, opened 2022-01-04
otlp receiver: Fix HTTP server config source and legacy port mutation

rmfitzpatrick:
**Description:**
Fixing a bug - When https://github.com/open-telemetry/opentelemetry-collector/pull/4519 was added it revealed that default http endpoint usage results in the otlp receiver modifying its config and that the http server starting logic disregarded the config parameter.

Reported via https://github.com/signalfx/splunk-otel-collector/issues/1064

**Testing:**
Added unit test

**Documentation:**
Updated changelog, but none needed otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 
 - Fix merge config map provider to close the watchers (#4570)
 - Fix expand map provider to call close on the base provider (#4571)
+- `otlp` receiver: Fix legacy port cfg value override and HTTP server starting bug (#4631)
 
 ## v0.41.0 Beta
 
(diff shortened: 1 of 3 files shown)


bogdandrutu:
@rmfitzpatrick why not fixing https://github.com/open-telemetry/opentelemetry-collector/issues/4565?

rmfitzpatrick:
> @rmfitzpatrick why not fixing #4565?

Because it sounds like they may be removed in total in the future and I opted to fix a bug in the meantime: https://github.com/open-telemetry/opentelemetry-collector/issues/4565#issuecomment-996907080

bogdandrutu:
I think the decision on that issue is that "the future" is now :)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- `otlp` receiver: Fix legacy port cfg value override and HTTP server starting bug (#4631)
```
dmitryax, line 5: This line should go to the Bug fixes section

dmitryax (commented):
Thanks for fixing it!

It looks good to me. 

@bogdandrutu I would merge this one before cutting the new release this week, and we can disable the the legacy endpoint after that

dmitryax (approved):

Issue #4630, opened 2022-01-04
Remove requirement that new components be presented across multiple PRs


Aneurysm9:
The [current process for adding a new component](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#when-adding-a-new-component) is not working for many reviewers and approvers.

[Quoting @anuraaga](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/6926#issuecomment-1004696786):

> * First PR ends up basically being a rubber stamp PR. The second PR tends to be massive anyways, and splitting out rubber-stamp type content doesn't help me that much, I can skim them and focus more on the code even within a single PR without the second PR feeling any more overwhelming
> * A PR with only boilerplate often stubs out empty methods for interface implementations and it can be hard to follow what is meant to be reviewed as final code and what is in-progress waiting for the second PR
> * GitHub will often assign a different reviewer for first and second PRs, making any context lost

Alternatives should be proposed and an update to the policy adopted.

bogdandrutu:
@jpkrohling @Aneurysm9 @anuraaga the logic of this for new "Components" (Processor/Receiver/etc.) is not as "rubber stamps" PRs. Let's look at the suggestion to better understand, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#when-adding-a-new-component:

**1:**
```
* First PR should include the overall structure of the new component:
  * Readme, configuration, and factory implementation usually using the helper
    factory structs.
  * This PR is usually trivial to review, so the size limit does not apply to
    it.
```

This PR as you can read should add a "readme" and "configuration" which are the most important user interfaces. They should definitely not be "rubber stamp PR." and if anything I think this is the most important PR out of all. This PR is a good for explaining what is the purpose of the new component, and how users can use it and. Also this PR is short enough, so if we cannot find sponsors, or the other maintainers have other ideas we don't feel bad throwing this away.

**2:**
```
* Second PR should include the concrete implementation of the component. If the
  size of this PR is larger than the recommended size consider splitting it in
  multiple PRs.
```

This is all about internal implementation, and things that can easily be fixed in the future without breaking changes. Also nothing stops the reviewer to split this if massive. I lots of times ask for this to be split into more simpler/easier to review PRs that are focus on one "logic".

**3:**
```
* Last PR should enable the new component and add it to the `otelcontribcol`
  binary by updating the `components.go` file. The component must be enabled
[...]

PR #4600, opened 2021-12-21
Move all utils configmapproviders to internal to prepare interface change

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #4636, opened 2022-01-05
Transform expand provider to converter

bogdandrutu:

### DIFF
--- a/service/internal/configprovider/expand.go
+++ b/service/internal/configprovider/expand.go
@@ -15,45 +15,21 @@
 package configprovider // import "go.opentelemetry.io/collector/service/internal/configprovider"
 
 import (
-	"context"
-	"fmt"
 	"os"
 
 	"go.opentelemetry.io/collector/config"
-	"go.opentelemetry.io/collector/config/configmapprovider"
 )
 
-type expandMapProvider struct {
-	base configmapprovider.Provider
-}
-
-// NewExpand returns a Provider, that expands all environment variables for a
-// config.Map provided by the given Provider.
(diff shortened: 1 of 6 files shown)


service/internal/configprovider/expand.go
```diff
@@ -15,45 +15,20 @@
...
-	"go.opentelemetry.io/collector/config/configmapprovider"
 )
 
-type expandMapProvider struct {
-	base configmapprovider.Provider
-}
-
-// NewExpand returns a Provider, that expands all environment variables for a
+// NewExpandConverter returns a Provider, that expands all environment variables for a
 // config.Map provided by the given Provider.
```
dmitryax, line 24: Please update the comment

dmitryax (approved):

Aneurysm9 (dismissed):
#4600 is going to break things for downstream distributions and we need to pause on digging a deeper hole.

bogdandrutu:
@Aneurysm9 if still unclear, see the next PR #4637. Also as commented in the #4600 it is not as bad as you think, and with #4637 you will not miss anything.

Aneurysm9:
Perhaps we should add this to the discussion on https://github.com/open-telemetry/opentelemetry-collector/issues/4630.  With all of the fragmented PRs happening in this space with zero communication of plan or desired end-state, I have no idea what works and what doesn't.  I will review this in the morning when I return to work.

bogdandrutu:
@Aneurysm9 this is not a component FYI, so your argument does not stand for that issue.

bogdandrutu:
@Aneurysm9 anything else blocking this or can be merged?

PR #4637, opened 2022-01-05
Embed merge support into the default ConfigProvider.

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/4636


service/config_provider.go
```diff
@@ -64,34 +64,41 @@ type ConfigProvider interface {
...
 
-// newConfigProvider returns a new ConfigProvider that provides the configuration using the given
-// `configMapProvider` and the given `configUnmarshaler`.
-func newConfigProvider(configMapProvider configmapprovider.Provider, configUnmarshaler configunmarshaler.ConfigUnmarshaler, cfgMapConverters ...ConfigMapConverterFunc) ConfigProvider {
+// NewConfigProvider returns a new ConfigProvider that provides the configuration:
+// * Retrieve the config.Map by merging all retrieved maps from all the configmapprovider.Provider in order.
+// * Then applies all the ConfigMapConverterFunc in the given order.
+// * Then unmarshalls the final config.Config using the given configunmarshaler.ConfigUnmarshaler.
+//
+// This is a temporary API.
```
dmitryax, line 81: Can we add few words what to expect instead going forward?

dmitryax (commented):
I believe this change needs a Changelog line

rmfitzpatrick:
Wanted to mention that this signature update will allow our downstream distribution to adopt the most recent core version after https://github.com/open-telemetry/opentelemetry-collector/pull/4600 hid a number of our required helpers, leaving us in the position to have to port and shim them all. Would greatly appreciate it and https://github.com/open-telemetry/opentelemetry-collector/pull/4636 landing.

service/config_provider.go
Aneurysm9: Seconded.  I think I would prefer to see this take functional options as it would allow for more flexibility in how the set of providers and converters are constructed.

service/config_provider.go
```diff
@@ -154,5 +160,44 @@ func (cm *configProvider) closeIfNeeded(ctx context.Context) error {
...
+
+func mergeRetrieve(ctx context.Context, onChange func(*configmapprovider.ChangeEvent), providers []configmapprovider.Provider) (configmapprovider.Retrieved, error) {
+	var retrs []configmapprovider.Retrieved
+	retCfgMap := config.NewMap()
+	for _, p := range providers {
+		retr, err := p.Retrieve(ctx, onChange)
+		if err != nil {
+			return nil, err
+		}
+		cfgMap, err := retr.Get(ctx)
```
Aneurysm9, line 174: Is there ever a reason to call `Retrieve()` independent of `Get()`, or vice versa?  Does this need to be a two-stage process?

bogdandrutu: Consider to read the API:
https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configmapprovider/provider.go#L38
https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configmapprovider/retrieved.go#L26

In summary the decision to retrieve an object with two function was preferred via returning the map + a close function. If you think the other option is better please file a short draft PR with how you would envision that, or an issue.

bogdandrutu: Please "unresolve" this comment if you feel the answer is not good enough.

service/config_provider.go
bogdandrutu: Even using "options" would not help me, since I will have to do the split of the `ConfigUnmarshaler` which will cause some breaking changes. Until that happens I prefer to mark this experimental to ensure the right expectations.

dmitryax (approved):
The code looks good to me. 

I would still ask to put some info in Changelog, maybe mention these functions as replacement for removed `configmapprovider.NewDefault`, `configmapprovider.NewExpand`, `configmapprovider.NewMerge` in the same Breaking changes line. 

Or at least mention this PR in https://github.com/open-telemetry/opentelemetry-collector/pull/4600

bogdandrutu:
@dmitryax done.

dmitryax (approved):

PR #4640, opened 2022-01-05
Add OTLP grpc compression benchmarks

jack-berg:
Per request in PR #4632, splitting out the addition of compression benchmarks from the change in the default OTLP grpc compression algo. 

Related to Issue #4587.


bogdandrutu (approved):
Very nice

PR #4642, opened 2022-01-05
Add otlptext support for exponential histogram

### LINKED ISSUE #4641: otlptext support for ExponentialHistogram
**Describe the bug**
The `otlptext` package does not recognize ExponentialHistogram data types.

**Steps to reproduce**
Demonstrated in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6666 using a logging exporter and a statsd receiver configured to produce exponential histograms.

The data type should print its buckets in a human readable way.

**What did you see instead?**
Nothing prints.

**What version did you use?**
60823e2

**What config did you use?**

```
receivers:
  statsd:
    endpoint: :8127
    aggregation_interval: 10s
    is_monotonic_counter: true
    timer_histogram_mapping:
    - statsd_type: histogram
      observer_type: histogram
      histogram:
        max_size: 16
    - statsd_type: timing
      observer_type: histogram
      histogram:
        max_size: 128

exporters:
  logging:
    loglevel: debug
   
service:
  pipelines:
    metrics:
      receivers: [statsd]
      exporters: [logging]
```

jmacd:
**Description:** 
Add support for printing OTLP exponential histograms

**Link to tracking Issue:** 
Fixes #4641 

**Testing:**

I ran `go test -v` with a modified test to verify that the printed format (which the test does not actually validate for any data point) is correct. 

The math equations used here are copied from https://github.com/open-telemetry/opentelemetry-go/pull/2393 where they have been extensively tested.

### DIFF
--- a/internal/otlptext/databuffer.go
+++ b/internal/otlptext/databuffer.go
@@ -146,6 +146,14 @@ func (b *dataBuffer) logExponentialHistogramDataPoints(ps pdata.ExponentialHisto
 
 		scale := int(p.Scale())
 		factor := math.Ldexp(math.Ln2, -scale)
+		// Note: the equation used here, which is
+		//   math.Exp(index * factor)
+		// reports +Inf as the _lower_ boundary of the bucket nearest
+		// infinity, which is incorrect and can be addressed in various
+		// ways.  The OTel-Go implementation of this histogram pending
+		// in https://github.com/open-telemetry/opentelemetry-go/pull/2393
+		// uses a lookup table for the last finite boundary, which can be
+		// easily computed using `math/big` (for scales up to 20).
 
 		negB := p.Negative().BucketCounts()
 		posB := p.Positive().BucketCounts()
--- a/internal/otlptext/databuffer.go
+++ b/internal/otlptext/databuffer.go
@@ -17,6 +17,7 @@ package otlptext // import "go.opentelemetry.io/collector/internal/otlptext"
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/otlptext/databuffer.go
```diff
@@ -128,6 +133,54 @@ func (b *dataBuffer) logDoubleHistogramDataPoints(ps pdata.HistogramDataPointSli
...
+
+		negB := p.Negative().BucketCounts()
+		posB := p.Positive().BucketCounts()
+
+		for i := 0; i < len(negB); i++ {
+			pos := len(negB) - i - 1
+			index := p.Negative().Offset() + int32(pos)
+			count := p.Negative().BucketCounts()[pos]
+			lower := math.Exp(float64(index) * factor)
+			upper := math.Exp(float64(index+1) * factor)
```
tigrannajaryan, line 166: These calculations are inherent property of how we define the histogram, right? Ideally we would want the calculation logic to be encapsulated in the `ExponentialHistogramDataPoint` struct and be callable as methods here. I am not sure if it is easily doable since the struct I think is auto-generated. As an alternate perhaps extract this logic as common helper functions somewhere so that any code that needs to read the histogram data can reuse it.

jmacd: I was a bit conflicted about this -- the code in open-telemetry/opentelemetry-go#2393 has these equations written in an internal package, and if those packages were exposed we wouldn't need to implement them here.

There is hand-written code in the `model/pdata` package, so we can add methods outside of the generated code.

OTOH, I had been thinking of this as a low-level dependency, and a non-critical use of the exponential histogram where it does not warrant bringing in a larger library. I documented how this equation is not perfect near infinity, which I think is a detail that does not matter for a logging exporter. I'd also be happy to create helper methods in  the `model/pdata.ExponentialHistogramDataPoint` using these equations for now, leave a TODO stating to replace the hard-coded equation after open-telemetry/opentelemetry-go#2393 or an equivalent library merges.

bogdandrutu (approved):

internal/otlptext/databuffer.go
jmacd: I will follow up.

PR #4645, opened 2022-01-05
Use config to setup skip-compilation instead of the flag

bogdandrutu:


cmd/otelcorecol/builder-config.yaml
```diff
@@ -1,3 +1,5 @@
+SkipCompilation: true
```
bogdandrutu, line 1: @jpkrohling this is possible right now, but I think would be good to actually define a `mapstructure` tag with a value like `skip_compilation`? What do you think?

Aneurysm9: What advantage does this have over a CLI flag?  It seems that this would prevent use of this config to directly build a collector binary.  It may not be used that way by `make genotelcorecol`, but why foreclose the option?

bogdandrutu: I am saying that right now, the way how viper works, makes this available. Is this what we want? Is this the name we want?

bogdandrutu: Question: did you know that this is possible? 

You do have a good point on keeping the flag, but do we need this config then?

jpkrohling: > Is this what we want? Is this the name we want?

I wasn't aware and am surprised that it works, although it does make sense now that you pointed it out. I see two different config types: runtime (CLI flags) and distribution-related config. They are internally stored in the same struct, but are logically different.

bogdandrutu: @jpkrohling can you take this and make a proposal to fix this one way or the other?

jpkrohling: I would just not document this flag and not support it being used this way. Not sure we need a way to actively prevent people from using it.

bogdandrutu: If you want to avoid the config support you can have a "mapstructure:-"

jpkrohling: True, that's an easy fix.

bogdandrutu: @jpkrohling you are not moving away from owning this :))

jpkrohling: lol , that wasn't my intention :-) I'll change this PR tomorrow morning.

bogdandrutu: @jpkrohling it looks like someone is keep trying to escape from this 🗡️

jpkrohling: I do have this in my queue but didn't find the time yet to do the coding... Too many code reviews to perform :-)

bogdandrutu: Maybe you can split the PRs to reduce your review time 😜

jpkrohling: Touche!

jpkrohling: Done!

jpkrohling:
@bogdandrutu, I think this is now ready for review.

bogdandrutu:
@jpkrohling it is funny to review my own PR :)))

bogdandrutu (commented):
I cannot Approve but LGTM :)))

jpkrohling (approved):
I'm approving @bogdandrutu's PR by proxy, as he approved it himself based on the code I added to it. Not sure how I feel about it, but I guess it's OK in the end :-)

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Remove deprecated `configtest.LoadConfig` and `configtest.LoadConfigAndValidate` (#4659)
 - Move service.ConfigMapConverterFunc to config.MapConverterFunc (#4673)
   - Add context to config.MapConverterFunc (#4678)
+- Builder: the skip compilation should only be supplied as a CLI flag. Previously, it was possible to specify that in the YAML file, contrary to the original intention (#4645)
```
tigrannajaryan, line 17: This would be nice to include in the PR description/commit message, otherwise it is not clear why the change is made.

tigrannajaryan (approved):

Issue #4627, opened 2022-01-04
CompressionType shouldn't be in internal/** path.


hyunuk:
**Is your feature request related to a problem? Please describe.**

There is a problem with this, similar in http case. We are exposing publicly an "internal" type (under internal/** path). We need to move `middleware.CompressionType` in a public package. We can do this in a separate PR, but I want you to have a tracking issue.

_Originally posted by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/pull/4624#discussion_r777778018_

**Describe the solution you'd like**

Since every code in `internal/middleware/compression.go` is for `config/confighttp.go`, except `CompressionType`, move the code from `internal/middleware/compression.go` to `config/confighttp.go` and delete this file. `config/configgrpc.go` will refer `confighttp` for `CompressionType` after this change is applied.

**Describe alternatives you've considered**

`config` package also has an internal folder, so move this to the `config/internal`.

bogdandrutu:
> Since every code in internal/middleware/compression.go is for config/confighttp.go, except CompressionType, move the code from internal/middleware/compression.go to config/confighttp.go and delete this file. config/configgrpc.go will refer confighttp for CompressionType after this change is applied.

Only that type needs to be public, the rest of the code MUST NOT be public unless in an internal.

Alternatively we can have a `configcompression` small package for this.

hyunuk:
Completed.

PR #4441, opened 2021-11-16
Support Snappy and Zstd for confighttp.go

hyunuk:
**Description:**
This PR provides [snappy](github.com/golang/snappy) and [zstd](https://github.com/klauspost/compress/tree/master/zstd) as additional compression methods for `confighttp`, which are the same libraries as `configgrpc` uses.

The `compression` field in `exporter/otlphttpexporter/config.go` has been removed and integrated into `confighttp.HTTPClientSettings`.

**Link to tracking Issue:**
#2766

PR #4651, opened 2022-01-07
fix: move compression.go into confighttp.go

hyunuk:
**Description:**

This PR moved every code in `internal/middleware/compression.go` to `config/confighttp.go` and tests as well.
It also changed functions unexported except `CompressionType`.

_Note: Functions in `compression.go` was iniatially created for `otlphttpexporter`, but it was changed to the `confighttp` in this PR(see #4441 and [the code](https://github.com/open-telemetry/opentelemetry-collector/pull/4441/files#diff-a13e181e395f2cb03704cecf9c08f081decf273f1042b7244ad732b01637ce24L93-L99))._

**Link to tracking Issue:**
#4627


bogdandrutu (commented):
Not sure if it is ok to make `configgrpc` depend on `confighttp`. Can we maybe consider a `configcompression`? Idea being other protocols (e.g. plain UDP) can use these constants as well.

Another option is `confignet`? Personally I prefer `configcompression`, any thoughts @tigrannajaryan @open-telemetry/collector-approvers ?

config/configcompression/compressionType.go
```diff
@@ -0,0 +1,49 @@
...
+type CompressionType string
+
+const (
+	CompressionGzip    CompressionType = "gzip"
+	CompressionZlib    CompressionType = "zlib"
+	CompressionDeflate CompressionType = "deflate"
+	CompressionSnappy  CompressionType = "snappy"
+	CompressionZstd    CompressionType = "zstd"
+	CompressionNone    CompressionType = "none"
+	CompressionEmpty   CompressionType = ""
```
bogdandrutu, line 28: Interested about the "empty" value... Should it be a public value for it? What is the use-case for it?

hyunuk: `CompressionEmpty` and `CompressionNone` was used in a different package before but now they are only used in here. I'll make them unexported.

config/configgrpc/configgrpc.go
```diff
@@ -39,6 +39,7 @@ import (
 	"go.opentelemetry.io/collector/client"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configauth"
+	"go.opentelemetry.io/collector/config/configcompression"
```
jpkrohling, line 42: No strong feelings, but I think I would prefer confignet, if we believe we might need another shared configs than compression.

config/configgrpc/configgrpc.go
```diff
@@ -357,13 +358,13 @@ func (gss *GRPCServerSettings) ToServerOption(host component.Host, settings comp
 }
 
 // getGRPCCompressionName returns compression name registered in grpc.
-func getGRPCCompressionName(compressionType middleware.CompressionType) (string, error) {
+func getGRPCCompressionName(compressionType configcompression.CompressionType) (string, error) {
 	switch compressionType {
-	case middleware.CompressionGzip:
+	case configcompression.CompressionGzip:
```
jpkrohling, line 363:
```suggestion
	case configcompression.Gzip:
```

If we stick with configcompression, remove "compression" from the property name.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
   - The merge functionality is now embedded into `service.NewConfigProvider` (#4637).
 - Move `configtest.LoadConfig` and `configtest.LoadConfigAndValidate` to `servicetest` (#4606)
 - Builder: Remove deprecated `include-core` flag (#4616)
+- Move `compression.go` into `confighttp.go` (#4651)
```
dmitryax, line 18: It doesn't seem like a breaking change. It's rather an additional public API / enhancement. Also please move it to "Unreleased" section and reword it to indicate what public API is exposed in which package.

config/configgrpc/configgrpc.go
hyunuk: done. Thank you

CHANGELOG.md
hyunuk: done. Thank you

config/configgrpc/configgrpc.go
bogdandrutu: I don't mind using the `confignet`, @tigrannajaryan @open-telemetry/collector-approvers please give at least one opinion for `configcompression` vs `confignet`

codeboten: I find `configcompression` more meaningful as to what the contents of the package are today. `confignet` doesn't make me think of compression necessarily, and it would be possible for these options to be used outside of network transport (ie for compressing data before saving it to a datastore) in which case it would be confusing for the option to be in `confignet`. It's not a strong preference but just what comes to mind off top my head.

codeboten (approved):

config/confighttp/confighttp.go
```diff
@@ -325,3 +333,158 @@ func authInterceptor(next http.Handler, authenticate configauth.AuthenticateFunc
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
+
+type compressRoundTripper struct {
```
bogdandrutu, line 337: Can we have this in a separate file in the same dir to avoid super large files?

bogdandrutu (approved):

config/confighttp/confighttp.go
bogdandrutu: Thank you @hyunuk

bogdandrutu:
please fix lint

config/configcompression/compressionType.go
```diff
@@ -0,0 +1,49 @@
...
+
+type CompressionType string
+
+const (
+	Gzip    CompressionType = "gzip"
+	Zlib    CompressionType = "zlib"
+	Deflate CompressionType = "deflate"
+	Snappy  CompressionType = "snappy"
+	Zstd    CompressionType = "zstd"
+	None    CompressionType = "none"
```
bogdandrutu, line 27: One PR ago you told me that this does not need to be public

PR #4652, opened 2022-01-07
Add external process execution guidelines

tigrannajaryan:
The guidelines are necessary to explain how we want to generally approach external process execution. This was recently brought up in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6512

If we accept these guidelines the following should happen:
- The prometheusexecreceiver should be modified to allow only a hard-coded list of exporters.
- The fluentbitextension should be either removed or significantly
  limited in terms of what locations and what executable file names it can allow.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6512
  will be rejected, possibly substituted by a plugin system that @zenmoto
  referred to in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6512#issuecomment-996922645
  if we find useful to have such plugin system.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -260,6 +260,24 @@ the event happens.
 Make log message human readable and also include data that is needed for easier
 understanding of what happened and in what context.
 
+### Executing External Processes
+
+The components should avoid executing arbitrary external processes with arbitrary command
+line arguments based on user input, including input received from the network or input
+read from the configuration file. Failure to follow this rule can result in arbitrary
+remote code execution, compelled by malicious actors that can craft the input.
+
+The following limitations are recommended:
+- If an external process needs to be executed limit and hard-code the location where
+  the executable file may be located, instead of allowing the input to dictate the
+  full path to the executable.
+- If possible limit the name of the executable file to be one from a hard-coded
+  list defined at compile time.
+- If command line arguments need to be passed to the process do not take the arguments
+  from the user input directly. Instead, compose the command line arguments indirectly,
+  if necessary, deriving the value from the user input. Limit as much as possible the
+  possible space of values for command line arguments.
+
 ### Observability
 
 Out of the box, your users should be able to observe the state of your component. 


CONTRIBUTING.md
```diff
@@ -260,6 +260,24 @@ the event happens.
...
+### Executing External Processes
+
+The components should avoid executing arbitrary external processes with arbitrary command
+line arguments based on user input, including input received from the network or input
+read from the configuration file. Failure to follow this rule can result in arbitrary
+remote code execution, compelled by malicious actors that can craft the input.
+
+The following limitation are recommended:
+- If an external process needs to be executed limit and hard-code the location where
+  the executable file may be located, instead of allowing the input to dictate the
```
Aneurysm9, line 272:
```suggestion
- If an external process needs to be executed, limit and hard-code the absolute path
  to the executable file instead of allowing the input to dictate the
```

We should be clear that an absolute path should be specified and no reliance on `$PATH` should be allowed.

tigrannajaryan: I do not think we need to limit this to the absolute path only. As long as it is a strictly predictable path I think it should be allowed. For example one good approach could that we we put `otelcol` executable in some directory, e.g. `/usr/bin/otelcol` and allow execution from a subdirectory with a hard-coded name, e.g. anything located in `/usr/bin/otelcolscripts/` is allowed. In the code this can be expressed as a directory relative to the directory where `otelcol` executable is located, yet I believe it is sufficiently safe.
I do agree that reliance on `$PATH` should not be allowed.
What do you think?

Aneurysm9: I think that allowing relative paths are dangerous.  If the `otelcol` executable is in `/usr/bin` then specifying `.` or `..` or `../sbin` seem undesirable.  Having the behavior change if the executable is moved also seems potentially surprising.  That said, I think what you've proposed would be better than doing nothing and shouldn't prevent us from further tightening the restrictions in the future.

tigrannajaryan: > If the otelcol executable is in /usr/bin then specifying . or .. or ../sbin seem undesirable. 

Absolutely. I was unclear in my description. I think the end user should not be allowed to specify arbitrary relative paths (nor they should be allowed to specify arbitrary absolute paths for that matter, see below). However there may be an implementation logic that uses a hard-coded relative path to a well known place which cannot be exploited.

For example imagine we want to have a receiver that allows scripts to be executed and their output captured as log records. It would be absolutely unacceptable to allow the path to the scripts to be a config option and allow this path to be either a relative path or an absolute path because that would mean a compromised config file allows malicious actors to execute arbitrary commands. Limiting this config option to absolute paths doesn't prevent this. The problem is in allowing arbitrary directories to be specified in the config file.

As opposed to that imagine that we have a receiver that allows scripts to be executed, but the limitation is that the scripts must be located in a specific directory. To achieve this we can hard-code the directory to be for example `/usr/bin/otelcolscripts` and this absolute path can literally be a hard-coded string in the receiver implementation. Alternatively the receiver implementation can calculate this directory to be an `otelcolscripts` subdirectory relative to the location where `otelcol` executable is stored. This relative approach allows more flexibility in the placement of the `otelcol` executable and the scripts that we want to be executable, yet is limited enough that it is impossible to run arbitrary commands from `/usr/bin` or any other directory that is not explicitly blessed by virtue of having the right name.

tigrannajaryan:
@open-telemetry/collector-approvers please review.

jpkrohling (approved):

dmitryax (approved):
LGTM

CONTRIBUTING.md
```diff
@@ -260,6 +260,24 @@ the event happens.
...
 understanding of what happened and in what context.
 
+### Executing External Processes
+
+The components should avoid executing arbitrary external processes with arbitrary command
+line arguments based on user input, including input received from the network or input
+read from the configuration file. Failure to follow this rule can result in arbitrary
+remote code execution, compelled by malicious actors that can craft the input.
+
+The following limitation are recommended:
```
mx-psi, line 270:
```suggestion
The following limitations are recommended:
```

mx-psi (approved):

bogdandrutu (approved):
For some of these statements may be better to have some concrete examples of "do" and "don't do"

tigrannajaryan:
> For some of these statements may be better to have some concrete examples of "do" and "don't do"

Good point. We have open issues for prometheusexecreceiver and fluentbitextension, I think once we decide how exactly these need to work we can link to their implementations as examples.

PR #4656, opened 2022-01-07
Define a type for onChange func instead of func pointer

bogdandrutu:

### DIFF
--- a/config/configmapprovider/provider.go
+++ b/config/configmapprovider/provider.go
@@ -40,15 +40,15 @@ type Provider interface {
 	// contains the value to be injected in the configuration and the corresponding watcher that
 	// will be used to monitor for updates of the retrieved value.
 	//
-	// onChange callback is called when the config changes. onChange may be called from
-	// a different go routine. After onChange is called Retrieved.Get should be called
+	// watcher callback is called when the config changes. watcher may be called from
+	// a different go routine. After watcher is called Retrieved.Get should be called
 	// to get the new config. See description of Retrieved for more details.
-	// onChange may be nil, which indicates that the caller is not interested in
+	// watcher may be nil, which indicates that the caller is not interested in
 	// knowing about the changes.
 	//
 	// If ctx is cancelled should return immediately with an error.
 	// Should never be called concurrently with itself or with Shutdown.
-	Retrieve(ctx context.Context, onChange func(*ChangeEvent)) (Retrieved, error)
+	Retrieve(ctx context.Context, watcher WatcherFunc) (Retrieved, error)
 
 	// Shutdown signals that the configuration for which this Provider was used to
 	// retrieve values is no longer in use and the Provider should close and release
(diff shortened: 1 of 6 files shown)


config/configmapprovider/provider.go
```diff
@@ -62,6 +62,8 @@ type Provider interface {
 	Shutdown(ctx context.Context) error
 }
 
+type Watcher func(*ChangeEvent)
```
bogdandrutu, line 65: Other name suggestions are also appreciate.

tigrannajaryan: `WatcherFunc` perhaps?

tigrannajaryan (approved):

config/configmapprovider/provider.go
bogdandrutu: Good suggestion, will do.

config/configmapprovider/provider.go
```diff
@@ -40,15 +40,15 @@ type Provider interface {
 	// contains the value to be injected in the configuration and the corresponding watcher that
 	// will be used to monitor for updates of the retrieved value.
 	//
-	// onChange callback is called when the config changes. onChange may be called from
+	// watcher callback is called when the config changes. onChange may be called from
 	// a different go routine. After onChange is called Retrieved.Get should be called
```
codeboten, line 44:
```suggestion
	// a different go routine. After watcher is called Retrieved.Get should be called
```

config/configmapprovider/provider.go
```diff
@@ -40,15 +40,15 @@ type Provider interface {
...
 	// to get the new config. See description of Retrieved for more details.
 	// onChange may be nil, which indicates that the caller is not interested in
```
codeboten, line 46:
```suggestion
	// watcher may be nil, which indicates that the caller is not interested in
```

service/config_provider.go
```diff
@@ -145,7 +145,7 @@ func (cm *configProvider) Watch() <-chan error {
 }
 
 func (cm *configProvider) onChange(event *configmapprovider.ChangeEvent) {
-	// TODO: Remove check for configsource.ErrSessionClosed when providers updated to not call onChange when closed.
+	// TODO: Remove check for configsource.ErrSessionClosed when providers updated to not call watcher when closed.
```
codeboten, line 148: should this have been changed? the method name here is still onChange

codeboten (commented):
Just confirming this is a change to improve the usability of the interface? I didn't see an issue associated or a description for the change.

bogdandrutu:
@codeboten yes, this PR is just to offer a more intuitive API, essentially instead of relying on the name of the variable to suggest what the function does, we define a type with a good name.

service/config_provider.go
bogdandrutu: GoLand decided to do this.

config/configmapprovider/provider.go
bogdandrutu: Done.

config/configmapprovider/provider.go
bogdandrutu: Done.

config/configmapprovider/provider.go
bogdandrutu: Done.

Aneurysm9 (approved):

Issue #4567, opened 2021-12-17
Support more config sources


pmm-sumo:
**Background**

A collector must be provided with a configuration to run (there's no default one). For this purpose, CollectorSettings includes [ConfigMapProvider](https://github.com/open-telemetry/opentelemetry-collector/blob/6888ffc9d7e3a3b0f669572025a64dd4730ac37f/service/settings.go#L62-L67) which allows to access te settings via [Provider](https://github.com/open-telemetry/opentelemetry-collector/blob/6888ffc9d7e3a3b0f669572025a64dd4730ac37f/config/configmapprovider/provider.go#L23-L51) interface.

Typically, the collector is run with a [default config map provider](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/command.go#L33-L35). It reads the config from the file (provided as a required `--config <filename>` argument) and optional `--set ...` which be also used to set arbitrary config params.

AWS OTel Collector is providing a custom provider, which extends it with a capability to read config from a preset environment variable (in its case, `ADT_CONFIG_CONTENT`), which, if found, is expanded into a full config map.

Currently, it's not possible to provide a custom config map provider via OpenTelemetry Collector Builder (which is mentioned in #4548)

**Describe the solution you'd like**

The default config map provider could be extended with an ability to specify other sources of config than the file. E.g.:

* `--config env:SOME_ENV_VARIABLE` - the config would be taken from `SOME_ENV_VARIABLE`, similarly as in case of AWS OTel Collector
* `--config https://example.com/some_config.yaml` - the config could be retrieved via a GET request (possible additional capabilities would be required here, like header with a token, etc.)
* ... 

Separate providers for each config source type could be created, which could be routed basing on the string identifying the source. 

Possibly, the providers could be registered in a way that would make it easier to interact with when using OpenTelemetry Collector Builder (e.g. could be provided by additional argument to the builder)

pmm-sumo:
@tigrannajaryan @jpkrohling I recall that there were some more ideas mentioned during the SIG. Perhaps just adding an additional set of Provider implementations (for env variable, http source, etc.) and some way to route to them accordingly might be sufficient though?

tigrannajaryan:
> @tigrannajaryan @jpkrohling I recall that there were some more ideas mentioned during the SIG. Perhaps just adding an additional set of Provider implementations (for env variable, http source, etc.) and some way to route to them accordingly might be sufficient though?

[...]

PR #4657, opened 2022-01-07
Change configmapprovider.Provider to accept an location for retrieve

bogdandrutu:
This is a preliminary change in order to support single/multiple local/remote config sources.
This PR changes the Provider to accept an URI as the resource identifier.

1. The current way is still fully supported and will continue to be supported, but there are more ways to specify the file:

Command line(currently supported): `./otelcol --config=local.yaml`
Command line(currently supported): `./otelcol --config=/path/to/local.yaml`
Command line(new support): `./otelcol --config=file:/path/to/local.yaml`

### DIFF
--- a/config/configmapprovider/env.go
+++ b/config/configmapprovider/env.go
@@ -16,33 +16,33 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 
 	"go.opentelemetry.io/collector/config"
 )
 
-type envMapProvider struct {
-	envName string
-}
+const envSchemeName = "env"
+
+type envMapProvider struct{}
 
 // NewEnv returns a new Provider that reads the configuration from the given environment variable.
-func NewEnv(envName string) Provider {
-	return &envMapProvider{
-		envName: envName,
-	}
+//
+// This Provider supports "env" scheme, and can be called with a selector:
+// `env:NAME_OF_ENVIRONMENT_VARIABLE`
+func NewEnv() Provider {
+	return &envMapProvider{}
 }
 
-func (emp *envMapProvider) Retrieve(context.Context, WatcherFunc) (Retrieved, error) {
-	if emp.envName == "" {
-		return nil, errors.New("config environment not specified")
+func (emp *envMapProvider) Retrieve(_ context.Context, location string, _ WatcherFunc) (Retrieved, error) {
+	if !strings.HasPrefix(location, envSchemeName+":") {
+		return nil, fmt.Errorf("%v location is not supported by %v provider", location, envSchemeName)
 	}
 
-	content := os.Getenv(emp.envName)
-
+	content := os.Getenv(location[len(envSchemeName)+1:])
 	var data map[string]interface{}
 	if err := yaml.Unmarshal([]byte(content), &data); err != nil {
 		return nil, fmt.Errorf("unable to parse yaml: %w", err)
--- a/config/configmapprovider/file.go
+++ b/config/configmapprovider/file.go
@@ -16,34 +16,46 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io/ioutil"
+	"path"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 
 	"go.opentelemetry.io/collector/config"
 )
 
-type fileMapProvider struct {
-	fileName string
+const fileSchemeName = "file"
+
+type fileMapProvider struct{}
+
+// NewFile returns a new Provider that reads the configuration from a file.
+//
+// This Provider supports "file" scheme, and can be called with a "location" that follows:
+//   file-location = "file:" local-path
+//   local-path    = [ drive-letter ] file-path
+//   drive-letter  = ALPHA ":"
+// The "file-path" can be relative or absolute, and it can be any OS supported format.
+//
+// Examples:
+// `file:path/to/file` - relative path (unix, windows)
+// `file:/path/to/file` - absolute path (unix, windows)
+// `file:c:/path/to/file` - absolute path including drive-letter (windows)
+// `file:c:\path\to\file` - absolute path including drive-letter (windows)
+func NewFile() Provider {
+	return &fileMapProvider{}
 }
 
-// NewFile returns a new Provider that reads the configuration from the given file.
-func NewFile(fileName string) Provider {
-	return &fileMapProvider{
-		fileName: fileName,
-	}
-}
-
-func (fmp *fileMapProvider) Retrieve(context.Context, WatcherFunc) (Retrieved, error) {
-	if fmp.fileName == "" {
-		return nil, errors.New("config file not specified")
+func (fmp *fileMapProvider) Retrieve(_ context.Context, location string, _ WatcherFunc) (Retrieved, error) {
+	if !strings.HasPrefix(location, fileSchemeName+":") {
+		return nil, fmt.Errorf("%v location is not supported by %v provider", location, fileSchemeName)
 	}
 
-	content, err := ioutil.ReadFile(fmp.fileName)
+	// Clean the path before using it.
+	content, err := ioutil.ReadFile(path.Clean(location[len(fileSchemeName)+1:]))
 	if err != nil {
-		return nil, fmt.Errorf("unable to read the file %v: %w", fmp.fileName, err)
+		return nil, fmt.Errorf("unable to read the file %v: %w", location, err)
 	}
 
 	var data map[string]interface{}
--- a/config/configmapprovider/provider.go
+++ b/config/configmapprovider/provider.go
@@ -24,11 +24,11 @@ import (
 //
 // The typical usage is the following:
 //
-//		r := mapProvider.Retrieve()
+//		r := mapProvider.Retrieve("file:/path/to/config")
 //		r.Get()
 //		// wait for onChange() to be called.
 //		r.Close()
-//		r = mapProvider.Retrieve()
+//		r = mapProvider.Retrieve("file:/path/to/config")
 //		r.Get()
 //		// wait for onChange() to be called.
 //		r.Close()
@@ -40,7 +40,11 @@ type Provider interface {
 	// contains the value to be injected in the configuration and the corresponding watcher that
 	// will be used to monitor for updates of the retrieved value.
 	//
-	// watcher callback is called when the config changes. watcher may be called from
+	// `location` must follow the "<scheme>:<opaque_data>" format. This format is compatible
+	// with the URI definition (see https://datatracker.ietf.org/doc/html/rfc3986). The "<scheme>"
+	// must be always included in the `location`.
+	//
+	// `watcher` callback is called when the config changes. watcher may be called from
 	// a different go routine. After watcher is called Retrieved.Get should be called
 	// to get the new config. See description of Retrieved for more details.
 	// watcher may be nil, which indicates that the caller is not interested in
@@ -48,7 +52,7 @@ type Provider interface {
 	//
 	// If ctx is cancelled should return immediately with an error.
 	// Should never be called concurrently with itself or with Shutdown.
-	Retrieve(ctx context.Context, watcher WatcherFunc) (Retrieved, error)
+	Retrieve(ctx context.Context, location string, watcher WatcherFunc) (Retrieved, error)
 
 	// Shutdown signals that the configuration for which this Provider was used to
 	// retrieve values is no longer in use and the Provider should close and release
--- a/service/config_provider.go
+++ b/service/config_provider.go
@@ -17,6 +17,7 @@ package service // import "go.opentelemetry.io/collector/service"
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 
 	"go.uber.org/multierr"
@@ -64,7 +65,8 @@ type ConfigProvider interface {
 }
 
 type configProvider struct {
-	configMapProviders []configmapprovider.Provider
+	locations          []string
+	configMapProviders map[string]configmapprovider.Provider
 	cfgMapConverters   []ConfigMapConverterFunc
 	configUnmarshaler  configunmarshaler.ConfigUnmarshaler
 
@@ -78,9 +80,16 @@ type configProvider struct {
 // * Then applies all the ConfigMapConverterFunc in the given order.
 // * Then unmarshalls the final config.Config using the given configunmarshaler.ConfigUnmarshaler.
 //
+// The `configMapProviders` is a map of pairs <scheme,Provider>.
+//
 // Notice: This API is experimental.
-func NewConfigProvider(configMapProviders []configmapprovider.Provider, cfgMapConverters []ConfigMapConverterFunc, configUnmarshaler configunmarshaler.ConfigUnmarshaler) ConfigProvider {
+func NewConfigProvider(
+	locations []string,
+	configMapProviders map[string]configmapprovider.Provider,
+	cfgMapConverters []ConfigMapConverterFunc,
+	configUnmarshaler configunmarshaler.ConfigUnmarshaler) ConfigProvider {
 	return &configProvider{
+		locations:          locations,
 		configMapProviders: configMapProviders,
 		cfgMapConverters:   cfgMapConverters,
 		configUnmarshaler:  configUnmarshaler,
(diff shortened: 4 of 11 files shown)


config/configmapprovider/file.go
```diff
@@ -16,34 +16,48 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
...
+const fileSchemeName = "file"
+
+type fileMapProvider struct{}
+
+// NewFile returns a new Provider that reads the configuration from a file.
+//
+// This Provider supports "file" scheme (see https://datatracker.ietf.org/doc/html/rfc8089), and can be called with an
+// uri with or without the scheme name:
+// `/path/to/file` - no scheme, no hostname
+// `file:/path/to/file` - no hostname
```
Aneurysm9, line 38: Would this mean that relative paths are no longer supported?

config/configmapprovider/env.go
```diff
@@ -18,31 +18,39 @@ import (
...
+	if url.Scheme != "" && url.Scheme != envSchemeName {
+		return nil, ErrSchemeIsNotSupported
 	}
 
-	content := os.Getenv(emp.envName)
+	if url.Host == "" {
+		return nil, errors.New("config environment not specified")
+	}
 
+	content := os.Getenv(url.Host)
```
Aneurysm9, line 53: This feels wrong as the value is not used as a host identifier. I wonder whether it might make more sense to specify that the --config flag can take values that specify a source by using <source_id>:<opaque_data>. This would allow sources where specifying them as a URI makes sense to do so but doesn't force sources like this one into that mold.

service/config_provider.go
```diff
@@ -79,8 +81,13 @@ type configProvider struct {
 // * Then unmarshalls the final config.Config using the given configunmarshaler.ConfigUnmarshaler.
 //
 // Notice: This API is experimental.
-func NewConfigProvider(configMapProviders []configmapprovider.Provider, cfgMapConverters []ConfigMapConverterFunc, configUnmarshaler configunmarshaler.ConfigUnmarshaler) ConfigProvider {
+func NewConfigProvider(
+	urls []*url.URL,
+	configMapProviders map[string]configmapprovider.Provider,
```
Aneurysm9, line 86: Is there a reason to require available providers and their scheme mappings to be included here explicitly rather than having a provider registry that providers hook into with an `init()` function?  That would seem to avoid issues that could be caused here if providers aren't included or have their scheme mapping changed around.

config/configmapprovider/provider.go
```diff
@@ -48,7 +58,14 @@ type Provider interface {
...
 	// Should never be called concurrently with itself or with Shutdown.
-	Retrieve(ctx context.Context, watcher WatcherFunc) (Retrieved, error)
+	Retrieve(ctx context.Context, url *url.URL, watcher WatcherFunc) (Retrieved, error)
+
+	// Scheme returns the implemented scheme name (see https://datatracker.ietf.org/doc/html/rfc3986#section-3.1)
+	// by this Provider.
+	//
+	// Implementation must make sure to not conflict with well known scheme names,
+	// see https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
+	Scheme() string
```
Aneurysm9, line 68: Is there any provision for a provider supporting multiple schemes, or multiple providers supporting the same scheme?  For instance, the prototype S3 provider I created could use either `s3://` or `http://` URLs.

bogdandrutu: Hmm.. There is a problem with that "http" cannot be restricted to only "s3", and if you do register the "http" for s3 an alternative that loads from any HTTP endpoint cannot exists.

config/configmapprovider/file.go
bogdandrutu: They are, I even have a test:

```
func TestFile(t *testing.T) {
	fp := NewFile()
	ret, err := fp.Retrieve(context.Background(), mustParseURL(filepath.Join("testdata", "default-config.yaml")), nil)
	require.NoError(t, err)
	cfg, err := ret.Get(context.Background())
	assert.NoError(t, err)
	expectedMap := config.NewMapFromStringMap(map[string]interface{}{
		"processors::batch":         nil,
		"exporters::otlp::endpoint": "localhost:4317",
	})
	assert.Equal(t, expectedMap, cfg)
	assert.NoError(t, ret.Close(context.Background()))
	assert.NoError(t, fp.Shutdown(context.Background()))
}
```

will update comment to reflect that it is still possible.

service/config_provider.go
bogdandrutu: That is about "initialization". I am trying at this point to focus more on the "Provider" changes to ensure I can achieve everything.

Not sure if we need the global + init() or have the "Builder" do the right thing, and have in the builder "cfgmapproviders" section. Will think more about this in a followup after I am done with `Provider` interface.

Aneurysm9: I think it should generally be the case that the builder is a way to build a working collector, and not the only way.  To the extent that we pile responsibility for needed functionality on the builder it seems like it would become more mandatory and less optional.

Will this PR stay as a draft until the followup is completed, or at least a spec agreed?  I'd prefer to avoid churn here if there is uncertainty about the way forward.

config/configmapprovider/env.go
bogdandrutu: Indeed, was thinking about this option as well. I liked the URL because it has RFC and clear requirements, but your suggestion also works if we make "source_id" required to ensure that if ":" in the opaque_data we parse correctly. Or we put a requirement that ":" needs to be escaped, but I think requiring the "source_id:" in the Provider API is fine, and for the "config" flag I think it is fine to expect that no files have ":"?

config/configmapprovider/provider.go
Aneurysm9: Yup, not sure there's a great solution to that other than not supporting HTTP URIs for S3 configs.  Maybe something could be done with patterns instead of schemes, or a chain of responsibility ending in a generic HTTP loader that could give other providers a chance to determine if a given URI should be handled.

bogdandrutu: But a generic HTTP/GET provider should work for the s3 case as well correct? Does not need to be something special.

Aneurysm9: For objects that are public, it would.  The prototype I built instead decomposed the URI to extract the information needed to use the S3 client and thus could be used for any object, even with the HTTP syntax.  This would be necessary if the object were in a different region as the [`s3://` syntax](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#accessing-a-bucket-using-S3-format) does not support specifying a region.

service/config_provider.go
bogdandrutu: I understand the churn problem, but these 2 are kind of different problems: finalizing the Provider interface vs how the Providers are passed to the Collector.

I would say that manually passing the list of available providers should be the way to go:
1. Avoids using global state;
2. Consistent with other components, I would ask why are these special?
3. Global + Init() is sometimes unexpected for the distribution owners, e.g. one component that they include may by mistake or intentionally depend on a provider (call init()) and make it available in the distribution
4. The builder is not the only way. The distribution that have custom builds will have to write this map instead of just "includes", which is a bit of extra work.

Based on the first 3 reasons I would go with an API where distributions have to manually pass the available providers that they want for that distribution.

config/configmapprovider/provider.go
bogdandrutu: Not sure what you are arguing for though:
1. I think one provider is responsible for only one "scheme" ("source_id");
2. If a well known URL scheme is defined for that name, then the provider MUST follow the RFCs that defines it.

config/configmapprovider/env.go
bogdandrutu: Changed, let me know about the `Provider` and `File` (which follows the URL definition, which means it is impossible to have "file:relative/path", only "relative/path" is accepted without the scheme "file:".

pmm-sumo:
Related issue: https://github.com/open-telemetry/opentelemetry-collector/issues/4567

config/configmapprovider/provider.go
```diff
@@ -16,19 +16,24 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
 
 import (
 	"context"
+	"errors"
 )
 
+// ErrSchemeIsNotSupported can be returned by providers that receive a `Retrieve` call
+// with unsupported scheme name.
+var ErrSchemeIsNotSupported = errors.New("telemetry type is not supported")
```
tigrannajaryan, line 24:
```suggestion
var ErrSchemeIsNotSupported = errors.New("scheme is not supported")
```

config/configmapprovider/provider.go
```diff
@@ -40,15 +45,27 @@ type Provider interface {
 	// contains the value to be injected in the configuration and the corresponding watcher that
 	// will be used to monitor for updates of the retrieved value.
 	//
-	// watcher callback is called when the config changes. watcher may be called from
+	// `selector` must follow the "<scheme>:<opaque_data>" format. This format is compatible with the URI definition
```
tigrannajaryan, line 48: Is `<scheme>:` part optional?

service/config_provider.go
tigrannajaryan: >Consistent with other components, I would ask why are these special?

Passing a `Factories` of config providers would be consistent with what we do with other component types.

service/config_provider.go
```diff
@@ -79,8 +81,13 @@ type configProvider struct {
 // * Then unmarshalls the final config.Config using the given configunmarshaler.ConfigUnmarshaler.
 //
 // Notice: This API is experimental.
-func NewConfigProvider(configMapProviders []configmapprovider.Provider, cfgMapConverters []ConfigMapConverterFunc, configUnmarshaler configunmarshaler.ConfigUnmarshaler) ConfigProvider {
+func NewConfigProvider(
+	selectors []string,
+	configMapProviders map[string]configmapprovider.Provider,
```
tigrannajaryan, line 86: >selectors []string,

Please document this parameter.

>configMapProviders map[string]configmapprovider.Provider,

What is the key of this map?

service/config_provider.go
```diff
@@ -96,8 +103,25 @@ type ConfigMapConverterFunc func(*config.Map) error
...
 	return NewConfigProvider(
-		[]configmapprovider.Provider{configmapprovider.NewFile(configFileName), configmapprovider.NewProperties(properties)},
-		[]ConfigMapConverterFunc{configprovider.NewExpandConverter()},
+		[]string{configFileName},
+		map[string]configmapprovider.Provider{
+			"file": configmapprovider.NewFile(),
+			"env":  configmapprovider.NewEnv(),
+		},
+		[]ConfigMapConverterFunc{
+			func(c *config.Map) error {
```
tigrannajaryan, line 112: It is not clear what this function does. It would be useful to extract it as a regular function, give it a name and document.

service/config_provider.go
```diff
@@ -163,11 +187,20 @@ func (cm *configProvider) Shutdown(ctx context.Context) error {
 	return multierr.Combine(cm.closeIfNeeded(ctx), mergeShutdown(ctx, cm.configMapProviders))
 }
 
-func mergeRetrieve(ctx context.Context, onChange func(*configmapprovider.ChangeEvent), providers []configmapprovider.Provider) (configmapprovider.Retrieved, error) {
+func mergeRetrieve(ctx context.Context, onChange func(*configmapprovider.ChangeEvent), selectors []string, providers map[string]configmapprovider.Provider) (configmapprovider.Retrieved, error) {
```
tigrannajaryan, line 190: We have `map[string]configmapprovider.Provider` in many places but it is not clear what `string` means here. Can we define a type alias (either for the key or for the map) to make the type of the key clear?

service/config_provider.go
```diff
@@ -163,11 +187,20 @@ func (cm *configProvider) Shutdown(ctx context.Context) error {
...
-		retr, err := p.Retrieve(ctx, onChange)
+	for _, selector := range selectors {
+		// For backwards compatibility, empty url scheme means "file".
+		scheme := "file"
+		if idx := strings.Index(selector, ":"); idx != -1 {
+			scheme = selector[:idx]
+		}
+		p, ok := providers[scheme]
+		if !ok {
+			return nil, configmapprovider.ErrSchemeIsNotSupported
```
tigrannajaryan, line 201: It would be useful to include the name of the unsupported scheme in the error message.

config/configmapprovider/provider.go
tigrannajaryan, line 48: It is not clear why this is called `selector`. Would `location` or `url` be a better name?

config/configmapprovider/file.go
bogdandrutu: Done.

service/config_provider.go
bogdandrutu: Here we are talking about passing the available components vs register to a global registry. The name Factory vs Provider we can debate :)

config/configmapprovider/provider.go
bogdandrutu: It is not really an URL, I had that version but it doesn't fit well for env. @Aneurysm9 had a comment that in order to have env:NAME you had to look for host in the URL and he did not like it.

Location may be better, will update

config/configmapprovider/provider.go
bogdandrutu: Done.

service/config_provider.go
bogdandrutu: Done.

config/configmapprovider/provider.go
bogdandrutu: Done.

bogdandrutu:
@tigrannajaryan indeed, I removed that since already too large.

service/config_provider.go
bogdandrutu: The "scheme"

config/configmapprovider/provider.go
bogdandrutu: Currently is not specified, should it be?

config/configmapprovider/env.go
Aneurysm9: > and for the "config" flag I think it is fine to expect that no files have ":"?

Unless we're supporting VMS, that's probably fine :)

service/config_provider.go
tigrannajaryan: I understand. I was just supporting the idea that it does not need to be a global registry. We don't do global registries for component factories, and I don't think we need a global registry for config providers.

service/config_provider.go
bogdandrutu: Reduced the usage to just one place, added comment about the meaning. Please reopen if anything is missing.

config/configmapprovider/provider.go
```diff
@@ -40,15 +45,27 @@ type Provider interface {
...
 	// Should never be called concurrently with itself or with Shutdown.
-	Retrieve(ctx context.Context, watcher WatcherFunc) (Retrieved, error)
+	Retrieve(ctx context.Context, location string, watcher WatcherFunc) (Retrieved, error)
+
+	// Scheme returns the implemented scheme name (see https://datatracker.ietf.org/doc/html/rfc3986#section-3.1)
+	// by this Provider.
+	//
+	// Implementation must make sure to not conflict with well known scheme names, or follow the standard,
+	// see https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
+	Scheme() string
```
Aneurysm9, line 68: This doesn't seem to be used anywhere and it's not clear to me where it would be useful.  Are there plans for it or can it be removed?

Aneurysm9:
> > Multiple Local Config Files:
> > Command line: ./otelcol --config=config.yaml --config=config2.yaml
> 
> Is this already implemented in this PR or planned for the future? I could not find the code that does this.

I don't believe so.  We'll need to convert `--config` to use [`Func()`](https://pkg.go.dev/flag#FlagSet.Func) or [`Var()`](https://pkg.go.dev/flag#FlagSet.Var) with an implementation that can build up a `[]string`.  `pflag` has [`StringArray()`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringArray), which would do what we need.

bogdandrutu:
@Aneurysm9 Indeed, I tried to respond https://github.com/open-telemetry/opentelemetry-collector/pull/4657#issuecomment-1010152653

config/configmapprovider/provider.go
bogdandrutu: Good call, can be added later. I will remove it for the moment.

Aneurysm9:
Github is annoying with its "some comments can be threaded, some cannot" nonsense...

config/configmapprovider/provider.go
bogdandrutu: Done.

service/config_provider.go
bogdandrutu: Done.

tigrannajaryan (approved):
LGTM

Aneurysm9 (approved):

PR #4666, opened 2022-01-11
Change Properties Provider to be a converter, does not match the Provider interface

bogdandrutu:
Resolves TODOs left in https://github.com/open-telemetry/opentelemetry-collector/pull/4657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## 🛑 Breaking changes 🛑
 
 - Change configmapprovider.Provider to accept a location for retrieve (#4657)
+- Change Properties Provider to be a Converter (#4666)
 - Define a type `WatcherFunc` for onChange func instead of func pointer (#4656)
 - Remove deprecated `configtest.LoadConfig` and `configtest.LoadConfigAndValidate` (#4659)
 
--- a/config/configmapprovider/properties.go
+++ b/config/configmapprovider/properties.go
@@ -16,7 +16,6 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
 
 import (
 	"bytes"
-	"context"
 	"strings"
 
 	"github.com/knadh/koanf/maps"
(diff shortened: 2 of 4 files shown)


config/configmapprovider/properties.go
```diff
@@ -25,31 +24,26 @@ import (
...
 //
 // Properties must follow the Java properties format, key-value list separated by equal sign with a "."
 // as key delimiter.
 //  ["processors.batch.timeout=2s", "processors.batch/foo.timeout=3s"]
-// TODO: Change this to be a service.ConfigMapConverterFunc.
-func NewProperties(properties []string) Provider {
-	return &propertiesMapProvider{
-		properties: properties,
+// TODO: Find the right place for service.ConfigMapConverterFunc and return that. Currently this will cause circular dependency.
+func NewPropertiesConverter(properties []string) func(*config.Map) error {
```
dmitryax, line 34: Why do we want this to be a converter? I'm not saying it cannot be, but it still can be considered as a source of config map... 

Maybe `MergePropertiesConverter` in that case?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## 🛑 Breaking changes 🛑
 
 - Change configmapprovider.Provider to accept a location for retrieve (#4657)
+- Change Properties Provider to be a Converter (#4666)
```
dmitryax, line 8: Should we have snippet on how to replace to configmap.NewProperties usage to the new converter in the PR description or here?

config/configmapprovider/properties.go
bogdandrutu: Override? since it is clear. The "Provider" now accepts a "location" (see https://github.com/open-telemetry/opentelemetry-collector/pull/4657) a.k.a an "url" or something similar from where to retrieve the config, here it is just overriding some values, it does not fetch any source to retrieve a config.

dmitryax: `OverridePropertiesConverter` sounds good!

CHANGELOG.md
bogdandrutu: Not sure what needs to be documented, since this is not directly used, usually used via the `service.NewDefaultConfigProvider` which does the right thing.

config/configmapprovider/properties.go
bogdandrutu: I think correctly is "overwrite" since we replace that:

To **"overwrite"** something is to put something else in its place, destroying the thing overwritten. To **"override"** something is to cause something else to operate instead of it without harming or changing the thing overridden.

dmitryax (approved):

PR #4667, opened 2022-01-11
Remove unnecessary import comment from test file

bogdandrutu:

### DIFF
--- a/config/configmapprovider/retrieved_test.go
+++ b/config/configmapprovider/retrieved_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package configmapprovider // import "go.opentelemetry.io/collector/config/configmapprovider"
+package configmapprovider
 
 import (
 	"context"


config/configmapprovider/retrieved_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package configmapprovider // import "go.opentelemetry.io/collector/config/configmapprovider"
```
dmitryax, line 15: We add these annotations for usability automatically in all packages. Why removing here?

dmitryax: https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile#L74-L76

bogdandrutu: The tool adds only for non-test, check that almost all test don't have this, all other files have it.

dmitryax: oh, I see. tests are skipped... nvm then

dmitryax (approved):

PR #4668, opened 2022-01-11
Replace service.ConfigMapConverterFunc with config.MapConverter interface

bogdandrutu:
This PR tries to solve 2 "todos" from the code, where the current `service.ConfigMapConverterFunc` cannot be used as return type because of circular deps.

### DIFF
--- a/config/configmapconverter.go
+++ b/config/configmapconverter.go
@@ -1,0 +1,52 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config // import "go.opentelemetry.io/collector/config"
+
+import (
+	"context"
+)
+
+// MapConverter interface to convert a config.Map. This allows distributions
+// (or components) to build backwards compatible config converters.
+type MapConverter interface {
+	Convert(ctx context.Context, cfgMap *Map) error
+
+	// privateMapConverter is an unexported func to disallow direct implementation.
+	privateMapConverter()
+}
+
+// ConvertFunc specifies the function invoked when the MapConverter.Convert is being called.
+type ConvertFunc func(context.Context, *Map) error
+
+// Convert implements the MapConverter.Convert.
+func (f ConvertFunc) Convert(ctx context.Context, cfgMap *Map) error {
+	if f == nil {
+		return nil
+	}
+	return f(ctx, cfgMap)
+}
+
+type converter struct {
+	ConvertFunc
+}
+
+func (converter) privateMapConverter() {}
+
+// NewMapConverter returns a MapConverter configured with the provided options.
+func NewMapConverter(convertFunc ConvertFunc) MapConverter {
+	return &converter{
+		ConvertFunc: convertFunc,
+	}
+}
--- a/config/configmapconverter_test.go
+++ b/config/configmapconverter_test.go
@@ -1,0 +1,38 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config // import "go.opentelemetry.io/collector/config"
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewRetrieved_NilConvertFunc(t *testing.T) {
+	mc := NewMapConverter(nil)
+	cfgMap := NewMapFromStringMap(map[string]interface{}{"foo": "bar"})
+	assert.NoError(t, mc.Convert(context.Background(), cfgMap))
+	assert.Equal(t, map[string]interface{}{"foo": "bar"}, cfgMap.ToStringMap())
+}
+
+func TestNewConverter_ReturnError(t *testing.T) {
+	expectedErr := errors.New("my error")
+	mc := NewMapConverter(func(ctx context.Context, m *Map) error {
+		return expectedErr
+	})
+	assert.ErrorIs(t, mc.Convert(context.Background(), NewMap()), expectedErr)
+}
--- a/service/internal/configprovider/expand.go
+++ b/service/internal/configprovider/expand.go
@@ -15,21 +15,20 @@
 package configprovider // import "go.opentelemetry.io/collector/service/internal/configprovider"
 
 import (
+	"context"
 	"os"
 
 	"go.opentelemetry.io/collector/config"
 )
 
-// NewExpandConverter returns a service.ConfigMapConverterFunc, that expands all environment variables for a given config.Map.
-//
-// This does not directly return service.ConfigMapConverterFunc to avoid circular dependencies, not a problem since it is internal.
-func NewExpandConverter() func(*config.Map) error {
-	return func(cfgMap *config.Map) error {
+// NewExpandConverter returns a config.MapConverter, that expands all environment variables for a given config.Map.
+func NewExpandConverter() config.MapConverter {
+	return config.NewMapConverter(func(_ context.Context, cfgMap *config.Map) error {
 		for _, k := range cfgMap.AllKeys() {
 			cfgMap.Set(k, expandStringValues(cfgMap.Get(k)))
 		}
 		return nil
-	}
+	})
 }
 
 func expandStringValues(value interface{}) interface{} {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Change Properties Provider to be a Converter (#4666)
 - Define a type `WatcherFunc` for onChange func instead of func pointer (#4656)
 - Remove deprecated `configtest.LoadConfig` and `configtest.LoadConfigAndValidate` (#4659)
+- Replace service.ConfigMapConverterFunc with config.MapConverter interface (#4668)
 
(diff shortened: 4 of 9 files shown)


config/configmapconverter.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+)
+
+// MapConverter interface to convert a config.Map. This allows distributions
+// (or components) to build backwards compatible config converters.
+type MapConverter interface {
+	Convert(ctx context.Context, cfgMap *Map) error
+
+	// privateMapConverter is an unexported func to disallow direct implementation.
+	privateMapConverter()
```
Aneurysm9, line 27: This seems overly-complex.  Why is implementation of this interface disallowed?  If that is necessary, why have `converter` as an intermediary rather than just adding the private method to `ConvertFunc`?

service/internal/configprovider/expand.go
```diff
@@ -15,21 +15,20 @@
...
 	"go.opentelemetry.io/collector/config"
 )
 
-// NewExpandConverter returns a service.ConfigMapConverterFunc, that expands all environment variables for a given config.Map.
-//
-// This does not directly return service.ConfigMapConverterFunc to avoid circular dependencies, not a problem since it is internal.
-func NewExpandConverter() func(*config.Map) error {
-	return func(cfgMap *config.Map) error {
+// NewExpandConverter returns a config.MapConverter, that expands all environment variables for a given config.Map.
+func NewExpandConverter() config.MapConverter {
```
Aneurysm9, line 25: Please move this back to `configmapprovider`.  I don't see any reason for it to be internal and it is needed by downstream distributions.  Doing it here can save me the trouble of creating another PR to do it.

config/configmapconverter_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewRetrieved_NilConvertFunc(t *testing.T) {
+	mc := NewMapConverter(nil)
+	cfgMap := NewMap()
+	assert.NoError(t, mc.Convert(context.Background(), cfgMap))
+	assert.Len(t, cfgMap.AllKeys(), 0)
```
Aneurysm9, line 29: This should probably test with a `Map` that has some random values in it.  The implementation could replace the provided map with a new empty map and still pass this test.

config/configmapconverter.go
bogdandrutu: Consistency with other places.

service/internal/configprovider/expand.go
bogdandrutu: This is a separate request, can you please open an issue. Trying to keep discussion focus on the change.

config/configmapconverter.go
bogdandrutu: > If that is necessary, why have converter as an intermediary rather than just adding the private method to ConvertFunc?

Because if I have `ConverterFunc` with the private method users can do `var cmc MapConverter = ConvertFunc(myFunc)` which defeats the purpose because if want to add new func on the interface I have to add it in the `ConvertFunc` as well which is not very clear.

config/configmapconverter_test.go
bogdandrutu: Comment should be on the "dependent" PR.

config/configmapconverter.go
Aneurysm9: If you add new functions onto the interface you need to implement them on `converter` as well.  How is that any different?

In the Go SIG we made the decision to only use the private method pattern of preventing implementation where we do not want outside implementations.  In other words, interfaces that were for the SDK to implement and to be consumed by others.  This seems the opposite, something that is likely to be implemented by others and only consumed be the collector framework.

config/configmapconverter_test.go
Aneurysm9: What?

service/internal/configprovider/expand.go
Aneurysm9: So I can spend weeks chasing changes you're making while waiting for approval?  Or, you can `git mv` and insert `map` back into the package name while you're in the midst of making any semblance of a rationale for making this internal obsolete.

config/configmapconverter.go
bogdandrutu: ...........

Here is simple, if we add a new function (indeed all new funcs have to have a "default" implementation), we will add them to the `converter` for sure, but all the current implementation that use `NewConverter` will continue to work.

Just give a try if you want to add a `CloseFunc`:
```golang
type MapConverter interface {
	Convert(ctx context.Context, cfgMap *Map) error

	Close(ctx context.Context) error

	privateMapConverter()
}

// ConvertFunc specifies the function invoked when the MapConverter.Convert is being called.
type ConvertFunc func(context.Context, *Map) error

// Convert implements the MapConverter.Convert.
func (f ConvertFunc) Convert(ctx context.Context, cfgMap *Map) error {
	if f == nil {
		return nil
	}
	return f(ctx, cfgMap)
}

// CloseFunc specifies the function invoked when the Retrieved.Close is being called.
type CloseFunc func(context.Context) error

// Close implements the Retrieved.Close.
func (f CloseFunc) Close(ctx context.Context) error {
	if f == nil {
		return nil
	}
	return f(ctx)
}

// MapConverterOption represents the possible options for NewMapConverter.
type MapConverterOption func(*retrieved)

// WithClose overrides the default `Close` function for a Retrieved.
// The default always returns nil.
func WithClose(closeFunc CloseFunc) MapConverterOption {
	return func(o *converter) {
		o.CloseFunc = closeFunc
	}
}

type converter struct {
	ConvertFunc
	CloseFunc
}

func (converter) privateMapConverter() {}

// NewMapConverter returns a MapConverter configured with the provided options.
func NewMapConverter(convertFunc ConvertFunc, options ...MapConverterOption) MapConverter {
	ret := &converter{
		ConvertFunc: convertFunc,
	}

	for _, op := range options {
		op(ret)
	}

	return ret
}
```

bogdandrutu: > In the Go SIG we made the decision to only use the private method pattern of preventing implementation where we do not want outside implementations. In other words, interfaces that were for the SDK to implement and to be consumed by others. This seems the opposite, something that is likely to be implemented by others and only consumed be the collector framework.

This is a different SIG, and we went with different approaches, so we will be consistent with this SIG previous work.

config/configmapconverter_test.go
bogdandrutu: My bad, that is true, there is another "Empty" test in the dependent PR (which I updated :)) based on your comment). Will update this as well.

config/configmapconverter.go
Aneurysm9: > < giant wall of code >
...

This isn't how interfaces in Go work.  This is excessively complicated for no apparent reason.  If you have a `Converter` interface:

```go
type Converter interface {
	Convert(ctx context.Context, cfgMap *Map) error
}
```

That can be implemented directly and used by the code that needs it.  No need for a private function or an unexported wrapper type or an initializer.  A `ConverterFunc` type could be added for convenience.

If you want to add a `Closer` interface, then do so:

```go
type Closer interface {
	Close(ctx context.Context) error
}
```

If a `Converter` needs to also be a `Closer`, it implements that interface.  The consumer can check if it is given a `Closer` and invoke `Close()` as necessary.  If the consumer, for some reason, really needs a `Converter` to also be guaranteed to be a `Closer`, it can define a `ConvertCloser` interface:

```go
type ConvertCloser interface {
	Converter
	Closer
}
```

That's it.  Nothing more.  No need to be more complicated than that.

service/internal/configprovider/expand.go
bogdandrutu: Not going to argue more, but I can tell you that I will not mix that in the same PR since is a different change, I can do it after this PR.

I've learned a new English word "semblance":
> the outward appearance or apparent form of something, especially when the reality is different.

config/configmapconverter.go
Aneurysm9: > This is a different SIG, and we went with different approaches, so we will be consistent with this SIG previous work.

As all of the changes you are making to the configuration system show, nothing is yet set in stone.  We can change prior decisions when we find that they were mistaken.  We did that in the Go SIG on this very topic.  It's okay.

config/configmapconverter_test.go
bogdandrutu: Done, sorry again for the confusion.

config/configmapconverter.go
bogdandrutu: Please open an issue with the proposal, or a draft PR. Happy to review proposals, for the moment as I said I will keep things consistent, otherwise if I am the one who applies a new pattern I will have to change the pattern everywhere, and I will not do that, so happy to review things :).

FYI: I do know how the optional interfaces work in `go` :), but we decided not to go with that path since it complicates the usage especially when you have to check lots of optional implementations. This decision

Aneurysm9 (changes requested):
There is disagreement regarding the approach to take in defining this new interface and no indication of consensus within the SIG.  This should be discussed at the next SIG meeting.

codeboten (commented):
@bogdandrutu @Aneurysm9 looking through the code it does look like a pattern that's been used (using private methods to prevent users from implementing the interface), although it's unclear to me whether the `MapConverter` interface is or isn't an interface we'd want users to implement. Based on this comment

```
// MapConverter interface to convert a config.Map. This allows distributions
// (or components) to build backwards compatible config converters.
```

I would expect distros to want to be able to implement this interface, but maybe it's not the intended use. Maybe my question here is how are users expected to use `MapConverter` in their distros (or components)?

Added this to the agenda for tomorrow's SIG. @open-telemetry/collector-approvers please take a look.

bogdandrutu:
@codeboten 

> although it's unclear to me whether the MapConverter interface is or isn't an interface we'd want users to implement.

`MapConverter` I believe that in general will be mostly implemented by us and distro owners, not sure if it matters who is implemented but I think this is the answer to your question.

>I would expect distros to want to be able to implement this interface, but maybe it's not the intended use. Maybe my question here is how are users expected to use MapConverter in their distros (or components)?

I see distros wanted to offer different capabilities like "hardcoded" config for components, or automatically add a component, etc. It is a very powerful capability that we give to the distos, so I expect that it will be used quiet a lot.

tigrannajaryan:
> This PR tries to solve 2 "todos" from the code, where the current `service.ConfigMapConverterFunc` cannot be used as return type because of circular deps.

Isn't it easier and cleaner to move `ConfigMapConverterFunc` to `config` package where `config.Map` lives currently? That eliminates the circular dependency and the type definition nicely fits in the `config` package. We will then have `config.Map` and `config.ConfigMapConverterFunc` that we can use everywhere and no circular dependencies.

Is the goal the elimination of todos and circular dependency or we really want to replace the function by an interface?

bogdandrutu:
@tigrannajaryan here is the PR https://github.com/open-telemetry/opentelemetry-collector/pull/4673

bogdandrutu:
PR was partially merged, and left an issue to discuss between type func vs interface.

PR #4458, opened 2021-11-18
Support env substitution in configuration file if variable not set

neelayu:
**Description:**
FEAT: Support for expanding environment variable in case it is not set. 
The basic syntax is:
`${ENV_VAR:-DEFAULT_VALUE}`

**Link to tracking Issue:** #4384 

**Testing:**  Unit Tests
This PR is similar to #4387, instead this doesn't use extra dependencies. 

I think this is a relatively common ask especially when configuration is allowed via environment variables. A fallback is expected in case the variable is not set.

PR #4672, opened 2022-01-12
Move expand config map converter back out of internal package

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

Moving the environment variable expansion to an internal package broke a use case we have in the ADOT collector downstream distribution.  We have [had to duplicate it](https://github.com/aws-observability/aws-otel-collector/pull/868) in order to ensure that we can ship a release.

The converter is now sitting alone in that internal package and other converter implementations exist outside of that package.  It is useful to downstream distributions and should not be locked away.


bogdandrutu (changes requested):
We need to discuss this in the SIG meeting.

Aneurysm9:
> We need to discuss this in the SIG meeting to determine if this is needed or not.

You seemed to agree it was worthwhile [yesterday](https://github.com/open-telemetry/opentelemetry-collector/pull/4668#discussion_r782600217):

> I will not mix that in the same PR since is a different change, I can do it after this PR.

What changed?

neelayu:
+1 to the need of this PR. We also need this in our custom distribution. 
Also #4458 would have to be merged with the latest changes.

bogdandrutu:
@neelayu how do you use it? Is the Default provider that already uses it not enough?

Aneurysm9:
There is capability in the default provider that cannot be replicated with any other provider if a user needs to change the default provider.  That alone should be reason enough to unwind this change.

neelayu:
> @neelayu how do you use it? Is the Default provider that already uses it not enough?

As mentioned by @Aneurysm9, we are using a custom config provider. And the only "converter" is made private now, so cant use it anymore.

bogdandrutu:
> As mentioned by @Aneurysm9, we are using a custom config provider. And the only "converter" is made private now, so cant use it anymore.

Can you elaborate a bit more? I am curious on what are the features that you need, if you can share.

Aneurysm9:
> Can you elaborate a bit more? I am curious on what are the features that you need, if you can share.

Without this being in an externally usable package it is impossible to use it with any other converter or provider that is not in the default provider.  That renders the existence of the ability to implement custom providers and converters illusory.  It doesn't matter what the specific use case is, it matters that a whole category of use cases are not possible while this converter remains locked in an internal package.

bogdandrutu:
@Aneurysm9 I know how amazon is using it, and actually you don't need that now with the support for `--config=env:FOO` so not sure why you are pushing on this, see the code that you have https://github.com/aws-observability/aws-otel-collector/blob/main/pkg/config/config_factory.go, can be translated into:

```golang
func GetConfigProvider() service.ConfigProvider {
	// aws-otel-collector supports loading yaml config from Env Var
	// including SSM parameter store for ECS use case
	if configContent, ok := os.LookupEnv(envKey); ok {
		log.Printf("Reading AOT config from environment: %v\n", configContent)
		return service.NewDefaultConfigProvider("env:"+envKey, nil /* was not supported before */)
	}

	return service.NewDefaultConfigProvider(getConfigFlag(), getSetFlag())
}
```

Aneurysm9:
That is the current usage.  If we implement a custom `ConfigMapProvider` or `MapConverterFunc`, how do we use that with the expand converter?

jcchavezs:
Our use case is: We retrieve the collector config from a bigger config file which includes this collector config internally. Before parse/unmarshal this config we need to expand the environment variables and then return that as retrieved. Currently we don't have means to do that as the only options are to expand env vars when the config comes from a file.

tigrannajaryan:
Based on the comments it seems to me we need to put together a short design doc that clarifies how custom distributions are supposed to load their configuration. Just by looking at the source code I do not clearly see that picture.
@bogdandrutu would you be able to put together a doc that shows the use cases and code snippets that shows how the particular use case is handled? I think it will help to either confirm that the current API is good enough or show the gaps that need to be addressed and demonstrate the need for the changes (including this particular change and perhaps any other changes needed). It should be faster to have this discussion on a design doc than on a code change PR.

bogdandrutu:
@tigrannajaryan I can do that, but only in first week of Feb, I have some pto planned.

alolita:
Hi @bogdandrutu - I have two parts in this request for you -

1/ Restore NewExpandConverter: We request you to restore the functionality provided by the NewExpandConverter public function (which is not public any more). This will enable collector-core users to maintain the flexibility to build custom config providers downstream. This is something we need now (before you go on PTO).

2/ Communicate design changes before making breaking changes: As @tigrannajaryan also requested, please provide your design thinking on how various use cases for developers using collector-core for custom config support can better use these changes. How do we use the new API vs the old API and how do we make the best use of it? It would be great to have a doc for this. PRs and issues are tough to have design discussions on if there are breaking changes. This can happen after you’re back from PTO :-)

Thanks for considering this request.

jcchavezs:
+1 on @alolita's request `1`. I also wonder if the code changes can be done without the need of @bogdandrutu to be actively present. This is supposed to be an OSS project and this kind of efforts are supposed to be driven by a community, not by a single person. Do you agree on that?

As per `2` I think specially with breaking changes we need to be clear with motivations and defintively point to a discussion rather than just opening a PR and someone else approving it without any sense of transparency and definitively missing a huge piece of context for such decision. This project has great automations for many things, I wonder if an automation for enforcing a link to an issue or a doc can be bring in (I am up for working on that).

dashpole:
Discussed with @bogdandrutu at the collector sig meeting today.  The general consensus was that this can be added back, but with the expectation that it will significantly change soon.  The open question from the sig meeting was:

Given our intent to change this (in major ways) soon, should we add a deprecation notice?

bogdandrutu:
SIG Discussion: The idea was to embed this (similar with merge functionality) into our implementation of the `service.ConfigProvider` (https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/config_provider.go#L67). This way we can extend the functionality that we currently have to "expand" env vars, to also expand any other source. This is what the current experimental configsource package tried to do.

Bogdan's comment: Instead of exposing this, maybe change the [constructor](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/config_provider.go#L86) to always append this in the slice of converters (probably at the end).

bogdandrutu:
@dashpole I think my current proposal (came to my mind after the meeting) to embed this in the current implementation is easier... what do you think?

Aneurysm9:
> Bogdan's comment: Instead of exposing this, maybe change the [constructor](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/config_provider.go#L86) to always append this in the slice of converters (probably at the end).

I dislike that approach.  Much as we shouldn't prevent users from extending the default provider, I don't think that we should enforce that they include particular functionality.  This part of the collector is a framework and should provide options.  Mandatory behavior should be limited as much as possible.

jcchavezs:
> I dislike that approach. Much as we shouldn't prevent users from extending the default provider, I don't think that we should enforce that they include particular functionality. This part of the collector is a framework and should provide options. Mandatory behavior should be limited as much as possible.

Agree but also the "probably at the end" is very unnexplicit and that will lead to missbehaviour and confusion. If something has to happen in specific order, be explicit or give it an own behaviour.

bogdandrutu:
@jcchavezs not sure about your comment, since you quoted something that disagrees with adding that by default, so that "probably at the end" is not applying.


 In my proposal I said "probably" because I want to let the author of the PR to think/decide about that behavior. I think, I should have used a "question mark" to signal that the author has to think about that, I was not trying to propose to leave "probably at the end" :)

jcchavezs:
I might be unclear. My point is, we shouldn't be plugin anything magically because it isn't explicit. User could decide to include env vars and do redactions after the expansion and forcing the expand at the end is both not explicit and leads to misbehaviours. If we wanna add something at the end in all cases then we can make it explicit by adding a something like a "postConverter" or whatever but in general I pretty much agree with @Aneurysm9 we must not do assumptions on what users will do, instead provide the alternatives to make it up theirselves.

bogdandrutu:
@jcchavezs As mentioned in the SIG meeting, if this is exposed will be removed soon, so all the ideas are trying to remove that extra breaking. Alternative proposal was to put this public but add a deprecation warning.

dmitryax:
> Alternative proposal was to put this public but add a deprecation warning.

I like this option more than others. I don't think we should add it implicitly to `NewConfigProvider`. We probably should give users an option to use it with a warning so they are ready that this can be removed going forward.

bogdandrutu:
@dmitryax I agree, maybe you can create an alternative PR since the author of this PR does not want to consider alternative approaches.

Aneurysm9:
> I agree, maybe you can create an alternative PR since the author of this PR does not want to consider alternative approaches.

I have said no such thing and you know that.

I do not think that a deprecation warning is appropriate as we don't know that it will be deprecated since we don't have a design for the future state, nor do we know what it will be replaced with.  What would we say, "this may change in the future"?  That's implied by the v0.x version.

Since this is not a breaking change we should make it now and deal with any future breaking changes in the area as separate.

dashpole:
As another data-point, Google also uses the Expand converter for its build of the collector. We're just a little (lot) behind head right now, so we haven't run into the change yet.  We've implemented our own WatchForUpdate source for the File config provider.

dashpole (approved):
While we are still on v0.x versions, I'd also prefer not adding deprecation warnings, although I'm not sure debating it is a good use of time.  We likely plan to change a large number of public functions in the near future, and I wouldn't advocate adding warnings everywhere.  Adding a warning just for this seems strange, and doing it everywhere seems like unnecessary work.  We aren't committing to keeping the package by merging this either.

bogdandrutu:
@dashpole with the new https://github.com/open-telemetry/opentelemetry-collector/pull/4712 I disagree with you. Since we are expecting from this moment to do this.

Aneurysm9:
This is not a breaking change.

dashpole:
> with the new #4712 I disagree with you. Since we are expecting from this moment to do this.

I left [a comment](https://github.com/open-telemetry/opentelemetry-collector/pull/4712/files#r789039147) there.  It isn't clear to me if something should be deprecated _before_ it is replaced (if it is being replaced), or during/after replacement.  If before, then I think adding a deprecation notice here is correct.  If during/after, we should omit it until we replace the package.

codeboten:
@tigrannajaryan @open-telemetry/collector-approvers please review this PR, based on the feedback here, it looks like it would help out users to move this back.

codeboten (approved):

jpkrohling:
@bogdandrutu, could you please take another look and reset your review if appropriate?

bogdandrutu:
@jpkrohling @codeboten I think I made it clear what my request was to unblock this...

jpkrohling:
@bogdandrutu there are so many comments here that I lost track of it.

codeboten:
Can you confirm that adding a deprecation notice here would move the PR forward? I'm just trying to unblock users until the full design moving forward is communicated as suggested by @tigrannajaryan.

Aneurysm9:
> @bogdandrutu Can you confirm that adding a deprecation notice here would move the PR forward? I'm just trying to unblock users until the full design moving forward is communicated as suggested by @tigrannajaryan.

We don't need a deprecation notice until there is something to replace it with.  As it stands we don't even know that this will be deprecatid.

jcchavezs:
> We don't need a deprecation notice until there is something to replace it with. As it stands we don't even know that this will be deprecatid.

How would that deprecation notice would look like if there is no replacement yet? Is there at least any link for a doc on how this will change in the near future? If any maybe a `experimental` tag would be helpful.

Aneurysm9:
> > We don't need a deprecation notice until there is something to replace it with. As it stands we don't even know that this will be deprecatid.
> 
> How would that deprecation notice would look like if there is no replacement yet? Is there at least any link for a doc on how this will change in the near future? If any maybe a `experimental` tag would be helpful.
> 
> Would that work for you @Aneurysm9 and @bogdandrutu?

An `experimental` notice is unnecessary as the entire collector framework is experimental at this point, as evidenced by its `v0.x` version.  

We should be generally moving in the direction of stability.  This change undoes a destabilizing change that has been impacting users for weeks.  We can argue all we want about where we want to go after that and how to get there, but can we for now just unbreak the things we have broken?

@bogdandrutu you know what the right thing to do here is.  Please do it.

jcchavezs:
> An experimental notice is unnecessary as the entire collector framework
is experimental at this point, as evidenced by its v0.x version.

Agree on that, tho some APIs are less likely to change than others I think.

José Carlos Chávez


man. 24. jan. 2022 kl. 22:05 skrev Anthony Mirabella <
***@***.***>:

bogdandrutu:
@jcchavezs @codeboten any note that this is very likely to change soon (deprecated, experimental, etc.) is ok. I just want to make sure distributions know that support for this in the collector is very likely to change or remove as a standalone thing.

bogdandrutu (approved):

PR #4673, opened 2022-01-12
Move service.ConfigMapConverterFunc to config.MapConverterFunc

bogdandrutu:
The drop of the Config prefix from the func name was forced by go lint, which does not like to repeat the package name as the prefix for types.

### DIFF
--- a/config/configmap.go
+++ b/config/configmap.go
@@ -30,6 +30,11 @@ const (
 	KeyDelimiter = "::"
 )
 
+// MapConverterFunc is a converter function for the config.Map that allows distributions
+// (in the future components as well) to build backwards compatible config converters.
+// TODO: Consider to add a context as the first argument.
+type MapConverterFunc func(*Map) error
+
 // NewMap creates a new empty config.Map instance.
 func NewMap() *Map {
 	return &Map{k: koanf.New(KeyDelimiter)}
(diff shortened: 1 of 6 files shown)


mx-psi (approved):

tigrannajaryan (approved):
LGTM. 
@Aneurysm9 this is a breaking change, but should be an easy fix if you depend on it in your distro, so I assume not a problem, right?

config/configmap.go
```diff
@@ -30,6 +30,11 @@ const (
 	KeyDelimiter = "::"
 )
 
+// MapConverterFunc is a converter function for the config.Map that allows distributions
+// (in the future components as well) to build backwards compatible config converters.
+// TODO: Consider to add a context as the first argument.
+type MapConverterFunc func(*Map) error
```
Aneurysm9, line 36: I believe that this is a valuable addition and should be done now.

Aneurysm9 (approved):
> this is a breaking change, but should be an easy fix if you depend on it in your distro, so I assume not a problem, right?

Yes, I believe we can work with this.  #4672 is more important to address breakage that we could not fix without code duplication.

I think the addition of a `context.Context` to this function signature is a valuable addition and was done in #4668 so could be done here, but as long as it is done before the next release there wouldn't need to be two breaking changes here to deal with.

Issue #4669, opened 2022-01-12
restrict message size for otlpreceiver's http endpoint


Frefreak:
**Is your feature request related to a problem? Please describe.**
Currently there seems to be no way to restrict the message size for otlpreceiver's http request. Grpc endpoint have a `max_recv_msg_size_mib` to achieve this so it might be useful to have a configuration for http endpoint also.

**Describe the solution you'd like**
1. add a new config for this.
or
2. provides a way to enable user to pass extra middlewares when creating otlpreceiver's factory.

**Additional context**
We use kafkaexporter with retry queue enabled. However the kafka have a message size restriction (a `recordbatch` can not exceeds 1MB by default). We'd like a way to make the otlpreceiver in sync with exporter so that a large message got rejected in receiver in the first place. In this case the client can be aware of this instead of thinking the request is success but could not find the span in the final UI due to exporting failure.

Frefreak:
closing as it is fixed by 4677

PR #4677, opened 2022-01-13
Add `max_request_body_size` to confighttp.HTTPServerSettings

Frefreak:
**Description:**
Add a max_recv_size option to otlpreceiver's http endpoint.
When not specified it is the default zero value 0, in which case will not introduce the size limit.

**Link to tracking Issue:** #4669

**Testing:**
- a new `TestHTTPMaxRecvSize` test is added in **otlp_test.go** (this uses the existing global json data to test)
- also manually tested

screenshot of a packet capture session for **x-protobuf** content-type
setting `max_recv_size` to 74 for a size 75 message:
image
image

**Documentation:**
update README for otlpreceiver and CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ## 💡 Enhancements 💡
 
 - Remove expand cases that cannot happen with config.Map (#4649)
+- Add `max_recv_size` to otlpreceiver's http settings (#4677)
 
 ## v0.42.0 Beta
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ## 💡 Enhancements 💡
 
 - Remove expand cases that cannot happen with config.Map (#4649)
-- Add `max_recv_size` to otlpreceiver's http settings (#4677)
+- Add `max_request_body_size` to otlpreceiver's http settings (#4677)
 
 ## v0.42.0 Beta
 
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -232,7 +232,6 @@ func (hss *HTTPServerSettings) ToListener() (net.Listener, error) {
 // returned by HTTPServerSettings.ToServer().
 type toServerOptions struct {
 	errorHandler
-	maxRequestBodySize int64
 }
 
 // ToServerOption is an option to change the behavior of the HTTP server
@@ -244,15 +243,6 @@ type ToServerOption func(opts *toServerOptions)
 func WithErrorHandler(e errorHandler) ToServerOption {
 	return func(opts *toServerOptions) {
 		opts.errorHandler = e
-	}
-}
-
-// WithMaxRequestBodySize introduce a request body size upper bound for the HTTP handler.
-// This uses http.MaxBytesReader under the hood and when the client sends a request body
-// that exceeds this limit, 400 will be returned.
-func WithMaxRequestBodySize(maxRequestBodySize int64) ToServerOption {
-	return func(opts *toServerOptions) {
-		opts.maxRequestBodySize = maxRequestBodySize
 	}
 }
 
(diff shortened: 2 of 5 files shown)


receiver/otlpreceiver/config.md
```diff
@@ -71,11 +71,12 @@ Config defines configuration for OTLP receiver.
...
+| Name          | Type                                                      | Default      | Docs                                                                               |
+| ----          | ----                                                      | -------      | ----                                                                               |
+| endpoint      | string                                                    | 0.0.0.0:4318 | Endpoint configures the listening address for the server.                          |
+| tls           | [configtls-TLSServerSetting](#configtls-TLSServerSetting) | <no value>   | TLSSetting struct exposes TLS client configuration.                                |
+| cors          | [confighttp-CORSSettings](#confighttp-CORSSettings)       | <no value>   | CORSSettings configures a receiver for HTTP cross-origin resource sharing (CORS).  |
+| max_recv_size | int                                                       | 0            | MaxRecvSize configures the maximum allowed body size for a single request.         |
```
mx-psi, line 79: Is there any way to make the name more explicit? Maybe `max_request_size`? Or `max_payload_size`?

Frefreak: `max_request_size` looks better. To be more explicit, how about `max_request_body_size`?

mx-psi, line 79: We should also specify the units here (bytes), and say that 0 means "any size".

Frefreak: updated

Frefreak: doc updated

mx-psi (approved):

receiver/otlpreceiver/config.md
Frefreak: Forgot to mention the default "0" case. added

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 ## 💡 Enhancements 💡
 
 - Remove expand cases that cannot happen with config.Map (#4649)
+- Add `max_request_body_size` to otlpreceiver's http settings (#4677)
```
bogdandrutu, line 15: This is not correct, this adds the option to all the receivers that use the HttpClientSettings and not just otlp, consider to update everywhere

config/confighttp/confighttp.go
```diff
@@ -237,6 +241,15 @@ func WithErrorHandler(e middleware.ErrorHandler) ToServerOption {
 	}
 }
 
+// WithMaxRecvSize introduce a request body size upper bound for the HTTP handler.
+// This uses http.MaxBytesReader under the hood and when the client sends a request body
+// that exceeds this limit, 400 will be returned.
+func WithMaxRecvSize(maxRecvSize int64) ToServerOption {
```
bogdandrutu, line 247: Why public?

CHANGELOG.md
Frefreak: CHANGELOG updated

config/confighttp/confighttp.go
bogdandrutu: `WithErrorHandler` is a func pointer, cannot be set in Configuration via yaml, this is not the case here. We don't have similar things for any other part of the config that can be configured in the yaml file.

Frefreak: Indeed. I removed the `maxRequestBodySize` field in toServerOptions and the *With* function. Just a `MaxRequestBodySize` field in `HTTPServerSettings`  is already enough.

bogdandrutu:
@codeboten @jpkrohling please review

bogdandrutu (approved):

receiver/otlpreceiver/otlp_test.go
```diff
@@ -668,6 +668,63 @@ func TestHTTPUseLegacyPortWhenUsingDefaultEndpoint(t *testing.T) {
...
+	require.NoError(t, err)
+	_, err = ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Equal(t, 200, resp.StatusCode)
+
+	err = r.Shutdown(context.Background())
+	require.NoError(t, err)
+
+	cfg.Protocols.HTTP.MaxRequestBodySize--
+	r, err = NewFactory().CreateTracesReceiver(
```
jpkrohling, line 708: I would have preferred this to be a separate test, but this can be done in a follow-up PR if you agree.

jpkrohling (approved):

codeboten (approved):
👍

receiver/otlpreceiver/otlp_test.go
Frefreak: > I would have preferred this to be a separate test, but this can be done in a follow-up PR if you agree.

I will prepare a follow-up PR later today.

PR #4679, opened 2022-01-13
Transform component.ExtensionFactory into a struct. 

bogdandrutu:
Other benefits/things discovered
* No need to have the helper packages, we can remove all "helper" packages to component (independent of this PR).
* Remove the internalinterface, and dependency between packages (component, internal, helpers). May help in the future with splitting into different packages.

### DIFF
--- a/component/extension.go
+++ b/component/extension.go
@@ -18,6 +18,7 @@ import (
 	"context"
 
 	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/internal/internalinterface"
 )
 
 // Extension is the interface for objects hosted by the OpenTelemetry Collector that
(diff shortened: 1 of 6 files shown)


bogdandrutu:
@tigrannajaryan @codeboten @mx-psi @jpkrohling give me ideas.

dmitryax:
Wrapping it all in a struct maybe seems not as clean as an interface, but straighten the dependency chain is much valuable I believe. Right now it's very convoluted.

mx-psi:
+1 to @dmitryax's comment: interface is cleaner, but this simplifies things while being as flexible as before

jpkrohling:
Sorry for my radio silence on this, I came back to this issue a few times but I do not have a strong opinion either way. You all seem to agree that a struct is preferable over the interface. Is this a pattern we want to adopt elsewhere as well, like in the configauth, by removing the ServerAuthenticator interface, for instance?

component/extension.go
```diff
@@ -52,22 +53,39 @@ type ExtensionCreateSettings struct {
...
+
+// CreateExtension creates a service extension based on the given config.
+func (f CreateExtensionFunc) CreateExtension(ctx context.Context, set ExtensionCreateSettings, cfg config.Extension) (Extension, error) {
+	return f(ctx, set, cfg)
+}
+
+// ExtensionFactory is a factory for extensions to the service.
+type ExtensionFactory struct {
+	internalinterface.BaseInternal
+	CfgType config.Type
```
bogdandrutu, line 81: not the best name, since conflicts with "Type()"

component/extension.go
```diff
@@ -52,22 +53,39 @@ type ExtensionCreateSettings struct {
...
+type CreateExtensionFunc func(context.Context, ExtensionCreateSettings, config.Extension) (Extension, error)
+
+// CreateExtension creates a service extension based on the given config.
+func (f CreateExtensionFunc) CreateExtension(ctx context.Context, set ExtensionCreateSettings, cfg config.Extension) (Extension, error) {
+	return f(ctx, set, cfg)
+}
+
+// ExtensionFactory is a factory for extensions to the service.
+type ExtensionFactory struct {
+	internalinterface.BaseInternal
```
bogdandrutu, line 80: Soon we will remove when all factories will be struct and this can be something internal in that package (or removed completely).

extension/extensionhelper/factory.go
```diff
@@ -15,60 +15,29 @@
...
-	internalinterface.BaseInternal
-	cfgType                config.Type
-	createDefaultConfig    CreateDefaultConfig
-	createServiceExtension CreateServiceExtension
-}
+// Deprecated: use component.CreateExtensionFunc.
+type CreateServiceExtension = component.CreateExtensionFunc
 
-// NewFactory returns a component.ExtensionFactory.
+// Deprecated: use component.ExtensionFactory directly.
```
bogdandrutu, line 31: Naturally the return type of this should be `*component.ExtensionFactory` but that will be a breaking change, also the component.Factories also uses the non pointers. Maybe we should keep the interface and have the same trick with not allowing extension?

jpkrohling: If we have an interface, why wouldn't we allow people to extend it?

mx-psi: > If we have an interface, why wouldn't we allow people to extend it?

If we add a new signal type (say, profiles), how would you add a new method to create a profiles exporter/receiver/processor?

jpkrohling: This reasoning should definitely be documented.

dmitryax:
@bogdandrutu this PR can be closed now, right?

PR #4682, opened 2022-01-14
Change to comment for some description in pull request template

khanhntd:
**Description:** 
Instead of manually deleting the description paragraph, it would be better for developers to see those as comments and only needs to review as less as possible before creating the PR. Therefore, it will decrease the time for them to open a PR.

**Link to tracking Issue:** 
N/A

**Testing:** 
Capture a picture with the final pull request template that is attached below.

**Documentation:** 
N/A

**Atteched Image:**
image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,14 +5,12 @@ and are not aimed at stabilizing and preparing the Collector for the release wil
 
 _Delete this paragraph before submitting._
 
-**Description:** <Describe what has changed. 
-Ex. Fixing a bug - Describe the bug and how this fixes the issue.
-Ex. Adding a feature - Explain what this achieves.>
+**Description:** <Describe what has changed.>
+<!--Ex. Fixing a bug - Describe the bug and how this fixes the issue.
+Ex. Adding a feature - Explain what this achieves.-->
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -5,14 +5,12 @@ and are not aimed at stabilizing and preparing the Collector for the release wil
 
 _Delete this paragraph before submitting._
 
-**Description:** <Describe what has changed. 
-Ex. Fixing a bug - Describe the bug and how this fixes the issue.
-Ex. Adding a feature - Explain what this achieves.>
+**Description:** <Describe what has changed.>
+<!--Ex. Fixing a bug - Describe the bug and how this fixes the issue.
+Ex. Adding a feature - Explain what this achieves.-->
```
bogdandrutu, line 10: Not sure about this, since users can leave it in the commit comment, the UI will not show us but will be merged in the final PR comment. Is that a problem?

khanhntd: Interesting. I thought the UI will show us if we did not delete it since I have experienced it once when creating a PR with aws-otel-collector. IMO, if we asked the developers to delete it, then why should we keep it in the commit comment though?

PR #4711, opened 2022-01-19
Update README.md

anthonysalvato3:
max_elapsed_time is set to a default of 5 minutes in queued_retry.go, but was listed as 120s in this readme.

**Testing:** I ran a collector instance and exported to an empty IP address with the default retry settings. The retries continued for more than 120s and less than 300s. So I concluded that of these two numbers, 300s must be correct.


mx-psi (approved):
LGTM, see https://github.com/open-telemetry/opentelemetry-collector/blob/04baaae533dc2fea3bfddce1df1c72c3d0694042/exporter/exporterhelper/queued_retry.go#L57

codeboten (approved):
Thanks for fixing this!

anthonysalvato3:
Do I need to create another equivalent PR with the "Skip Changelog" label?

bogdandrutu:
@anthonysalvato3 we have the "power" :)) I added the label for you.

Issue #4674, opened 2022-01-12
Interfaces to be implemented by users should not have unexported methods


Aneurysm9:
In #4668 I expressed concern that a new interface was created that unnecessarily included an unexported method with the goal of preventing direct implementation of the interface.  Instead, providing a "partial" implementation to an initializer was required, with the initializer wrapping the partial implementation in an unexported struct that had an implementation of the unexported interface method with an empty body.  However, since the interface had only one real method and that is the method provided to the initializer, it seemed to serve no purpose and only added complexity.

The response to this concern was that adding the unexported interface method was in conformance with practice throughout this repo and must remain for consistency.  There appears to be no formal statement of this practice or of a recommendation that it be used.  It is also not used consistently throughout this repository and attempting to do so would likely have significant downstream impact.

Of the 69 exported interfaces I found in this repository only 13 include an unexported method, including the `internalinterface.InternalInterface` that was created for this purpose but only used directly in one exported interface (and indirectly in five more that embed that exported interface).  One of those interfaces (`configmapprovider.Retrieved`) was only recently changed to include an unexported method in #4577, with no explanation offered for why.

```
component.Factory (through embedding internalinterface.InternalInterface)
component.ExporterFactory (through embedding component.Factory)
component.ExtensionFactory (through embedding component.Factory)
component.ProcessorFactory (through embedding component.Factory)
component.ReceiverFactory (through embedding component.Factory)
experimental/component.ConfigSourceFactory (through embedding component.Factory)
config.Exporter
config.Extension
config.Processor
config.Receiver
configmapprovider.Retrieved
consumertest.Consumer
internalinterface.InternalInterface
```

I think that we should generally discourage the inclusion of unexported methods in exported interfaces and have a high bar for justifying the use of the pattern.

bogdandrutu:
First of all, please stop the personal accusations (because I know that is the case in things like `One of those interfaces (configmapprovider.Retrieved) was only recently changed to include an unexported method in #4577, with no explanation offered for why.`).

> One of those interfaces (configmapprovider.Retrieved) was only recently changed to include an unexported method in #4577, with no explanation offered for why. 

The explanation is very simple, once a pattern was adopted by the project we can use that without having to explain every time we use that patter "why".

Aneurysm9:
> The explanation is very simple, once a pattern was adopted by the project we can use that without having to explain every time we use that patter "why".

[...]

PR #4712, opened 2022-01-19
Add guidelines for breaking changes

jpkrohling:
As a result of the discussion involving #4672, we agreed during the SIG Collector meeting that we'd start recommending and enforcing breaking changes to be performed in two steps.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -359,6 +359,8 @@ version (`vM.N+1`).
 - when renaming types, functions, or attributes, we MUST create the new name and MUST deprecate the old one in one
   version (step 1), and MAY remove it in a subsequent version (step 2). For simple renames, the old name SHALL call the
   new one.
+- when a feature is being replaced in favor of an existing one, we MUST mark a feature as deprecated in one version, and
+  MAY remove it in the next one
 
 When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it
 behind a [feature
@@ -383,9 +385,19 @@ that each of the following steps is done in a separate version:
 1. Current version `v0.N` has `func GetFoo() Foo`
 1. We now need to also return an error. We do it by creating a new function that will be equivalent to the existing one,
    so that current users can easily migrate to that: `func MustGetFoo() Foo`, which panics on errors. We release this in
-   `v0.N+1`, deprecating the existing `func GetFoo() Foo` with it, perhaps adding a log entry with a warning and an
-   entry to the changelog.
+   `v0.N+1`, deprecating the existing `func GetFoo() Foo` with it, adding an entry to the changelog and perhaps a log
+   entry with a warning.
 1. On `v0.N+2`, we change `func GetFoo() Foo` to `func GetFoo() (Foo, error)`.
+
+#### Example #3 - Changing the arguments of a function
+
+1. Current version `v0.N` has `func GetFoo() Foo`
+1. We now decided to do something that might be blocking as part of `func GetFoo() Foo`, so, we start accepting a
+   context: `func GetFooWithContext(context.Context) Foo`. We release this in `v0.N+1`, deprecating the existing `func
+   GetFoo() Foo` with it, adding an entry to the changelog and perhaps a log entry with a warning. The existing `func
+   GetFoo() Foo` is changed to call `func GetFooWithContext(context.Background()) Foo`.
+1. On `v0.N+2`, we change `func GetFoo() Foo` to `func GetFoo(context.Context) Foo` if desired or remove it entirely if
+   needed.
 
 #### Exceptions
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -348,22 +348,39 @@ or report a warning back to the collector with a clear error saying CGO is requi
 ### Breaking changes
 
 Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
-without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
-changelog, possibly with a line instructing users how to move forward.
+without a good reason. Hence, when a change is known to cause a breaking change, it MUST be clearly marked in the
+changelog, and SHOULD include a line instructing users how to move forward.
 
 We also strive to perform breaking changes in two stages: 
 
-- when we need to remove something, we mark a feature as deprecated in one version, and remove it in the next one
-- when renaming types, functions, or attributes, we create the new name first, deprecating the old one in one
-version, removing it in the next version
-- when creating new interfaces, provide helpers so that consumers can be forward compatible without much effort. One
+- when we need to remove something, we MUST mark a feature as deprecated in one version, and MAY remove it in the next one
+- when renaming types, functions, or attributes, we MUST create the new name first and MUST deprecate the old one in one
+version, and MAY remove it in the next version. For simple renames, the old name SHALL call the new one.
+- when creating new interfaces, helpers MUST be provided so that consumers can be forward compatible without much effort. One
 example can be found in the `configauth` package, where a [`ServerAuthenticator`](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/serverauth.go#L28-L40)
 interface exists, we as well as a [helper](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/default_serverauthenticator.go#L59-L71)
 that most implementations should use.
 
-When deprecating a feature affecting end-users, consider first deprecating the feature, then hiding it behind
-a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/),
-and eventually removing it. 
+When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind
+a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/) in the next version,
+and eventually removing it after yet another version.
+
+#### Example #1 - Renaming a function
+
+1. Current version `v0.N` has `func GetFoo() Bar`
+1. We now decided that `GetBar` is a better name. As such, on `v0.N+1` we add a new `func GetBar() Bar` function,
+changing the existing `func GetFoo() Bar` to be an alias of the new function. Additionally, a log entry with a warning
+is added to the old function, along with an entry to the changelog.
+1. On `v0.N+2`, we MAY remove `func GetFoo() Bar`.
+
+#### Example #2 - Changing the return values of a function
+
+1. Current version `v0.N` has `func GetFoo() Foo`
+1. We now need to also return an error. We do it by creating a new function that will be equivalent to the existing
+one, so that current users can easily migrate to that: `func MustGetFoo() Foo`, which panics on errors. We release
+this in `v0.N+1`, deprecating the existing `func GetFoo() Foo` with it, adding a log entry with a warning and an
+entry to the changelog.
+1. On `v0.N+2`, we change `func GetFoo() Foo` to `func GetFoo() (Foo, error)`.
 
 ## Updating Changelog
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -403,7 +403,9 @@ that each of the following steps is done in a separate version:
 
 While the above is what we strive to follow, we acknowledge that some changes might be unfeasible to achieve in a
 non-breaking manner. Exceptions to the outlined rules are acceptable if consensus can be obtained from approvers in the
-pull request they are proposed. A reason for requesting the exception MUST be given in the pull request.
+pull request they are proposed. A reason for requesting the exception MUST be given in the pull request. Until unanimity
+is obtained, approvers and maintainers are encouraged to discuss the issue at hand. If a consensus (unanimity) cannot be
+obtained, the maintainers are then tasked in getting a decision using its regular means (voting, TC help, ...).
 
 ## Updating Changelog
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -358,8 +358,8 @@ We also strive to perform breaking changes in two stages:
 version, and MAY remove it in the next version. For simple renames, the old name SHALL call the new one.
 
 When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind
-a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/) in the next version,
-and eventually removing it after yet another version.
+a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
+in a subsequent version, and eventually removing it after yet another version.
 
 #### Example #1 - Renaming a function
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -351,12 +351,19 @@ Whenever possible, we adhere to semver as our minimum standards. Even before v1,
 without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
 changelog, possibly with a line instructing users how to move forward.
 
-We also strive to perform breaking changes in two stages, or using feature flags: 
+We also strive to perform breaking changes in two stages: 
 
 - when we need to remove something, we mark a feature as deprecated in one version, and remove it in the next one
 - when renaming types, functions, or attributes, we create the new name first, deprecating the old one in one
 version, removing it in the next version
-- when creating new interfaces, provide helpers so that consumers can be forward compatible without much effort
+- when creating new interfaces, provide helpers so that consumers can be forward compatible without much effort. One
+example can be found in the `configauth` package, where a [`ServerAuthenticator`](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/serverauth.go#L28-L40)
+interface exists, we as well as a [helper](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/default_serverauthenticator.go#L59-L71)
+that most implementations should use.
+
+When deprecating a feature affecting end-users, consider first deprecating the feature, then hiding it behind
+a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/),
+and eventually removing it. 
 
 ## Updating Changelog
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -356,9 +356,9 @@ version (`vM.N+1`).
 
 - when we need to remove something, we MUST mark a feature as deprecated in one version, and MAY remove it in the next
   one
-- when renaming types, functions, or attributes, we MUST create the new name and MUST deprecate the old one in one
-  version (step 1), and MAY remove it in a subsequent version (step 2). For simple renames, the old name SHALL call the
-  new one.
+- when renaming or refactoring types, functions, or attributes, we MUST create the new name and MUST deprecate the old
+  one in one version (step 1), and MAY remove it in a subsequent version (step 2). For simple renames, the old name
+  SHALL call the new one.
 - when a feature is being replaced in favor of an existing one, we MUST mark a feature as deprecated in one version, and
   MAY remove it in the next one
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -353,29 +353,32 @@ changelog, and SHOULD include a line instructing users how to move forward.
 
 We also strive to perform breaking changes in two stages: 
 
-- when we need to remove something, we MUST mark a feature as deprecated in one version, and MAY remove it in the next one
-- when renaming types, functions, or attributes, we MUST create the new name first and MUST deprecate the old one in one
-version, and MAY remove it in the next version. For simple renames, the old name SHALL call the new one.
+- when we need to remove something, we MUST mark a feature as deprecated in one version, and MAY remove it in the next
+  one
+- when renaming types, functions, or attributes, we MUST create the new name and MUST deprecate the old one in one
+  version (step 1), and MAY remove it in a subsequent version (step 2). For simple renames, the old name SHALL call the
+  new one.
 
-When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind
-a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
+When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it
+behind a [feature
+flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
 in a subsequent version, and eventually removing it after yet another version.
 
 #### Example #1 - Renaming a function
 
 1. Current version `v0.N` has `func GetFoo() Bar`
 1. We now decided that `GetBar` is a better name. As such, on `v0.N+1` we add a new `func GetBar() Bar` function,
-changing the existing `func GetFoo() Bar` to be an alias of the new function. Additionally, a log entry with a warning
-is added to the old function, along with an entry to the changelog.
+   changing the existing `func GetFoo() Bar` to be an alias of the new function. Additionally, a log entry with a
+   warning is added to the old function, along with an entry to the changelog.
 1. On `v0.N+2`, we MAY remove `func GetFoo() Bar`.
 
 #### Example #2 - Changing the return values of a function
 
 1. Current version `v0.N` has `func GetFoo() Foo`
-1. We now need to also return an error. We do it by creating a new function that will be equivalent to the existing
-one, so that current users can easily migrate to that: `func MustGetFoo() Foo`, which panics on errors. We release
-this in `v0.N+1`, deprecating the existing `func GetFoo() Foo` with it, adding a log entry with a warning and an
-entry to the changelog.
+1. We now need to also return an error. We do it by creating a new function that will be equivalent to the existing one,
+   so that current users can easily migrate to that: `func MustGetFoo() Foo`, which panics on errors. We release this in
+   `v0.N+1`, deprecating the existing `func GetFoo() Foo` with it, perhaps adding a log entry with a warning and an
+   entry to the changelog.
 1. On `v0.N+2`, we change `func GetFoo() Foo` to `func GetFoo() (Foo, error)`.
 
 ## Updating Changelog
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -351,7 +351,8 @@ Whenever possible, we adhere to semver as our minimum standards. Even before v1,
 without a good reason. Hence, when a change is known to cause a breaking change, it MUST be clearly marked in the
 changelog, and SHOULD include a line instructing users how to move forward.
 
-We also strive to perform breaking changes in two stages: 
+We also strive to perform breaking changes in two stages, deprecating it first (`vM.N`) and breaking it in a subsequent
+version (`vM.N+1`).
 
 - when we need to remove something, we MUST mark a feature as deprecated in one version, and MAY remove it in the next
   one
@@ -362,7 +363,12 @@ We also strive to perform breaking changes in two stages:
 When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it
 behind a [feature
 flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
-in a subsequent version, and eventually removing it after yet another version.
+in a subsequent version, and eventually removing it after yet another version. This is how it would look like, considering
+that each of the following steps is done in a separate version:
+
+1. Mark the feature as deprecated, add a short lived feature flag with the feature enabled by default
+1. Change the feature flag to disable the feature by default, deprecating the flag at the same time
+1. Remove the feature and the flag
 
 #### Example #1 - Renaming a function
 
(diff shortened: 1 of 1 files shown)


jpkrohling:
@dashpole, do you have a good resource about how to do feature flags in the collector? If so, I'll add a link.

CONTRIBUTING.md
```diff
@@ -345,6 +345,18 @@ with clear instructions on how to install the required libraries.
 Furthermore, if your package requires CGO, it MUST be able to compile and operate in a no op mode
 or report a warning back to the collector with a clear error saying CGO is required to work.
 
+### Doing breaking changes
```
codeboten, line 348:
```suggestion
### Breaking changes
```

codeboten (approved):

CONTRIBUTING.md
```diff
@@ -345,6 +345,18 @@ with clear instructions on how to install the required libraries.
 Furthermore, if your package requires CGO, it MUST be able to compile and operate in a no op mode
 or report a warning back to the collector with a clear error saying CGO is required to work.
 
+### Breaking changes
+
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
+changelog, possibly with a line instructing users how to move forward.
+
+We also strive to perform breaking changes in two stages, or using feature flags:
```
bogdandrutu, line 354: Feature flags do not really help with API changes (only with functionality) since they are evaluated at runtime. Alternative can be "build tags"

CONTRIBUTING.md
```diff
@@ -345,6 +345,18 @@ with clear instructions on how to install the required libraries.
...
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
+changelog, possibly with a line instructing users how to move forward.
+
+We also strive to perform breaking changes in two stages, or using feature flags: 
+
+- when we need to remove something, we mark a feature as deprecated in one version, and remove it in the next one
+- when renaming types, functions, or attributes, we create the new name first, deprecating the old one in one
+version, removing it in the next version
+- when creating new interfaces, provide helpers so that consumers can be forward compatible without much effort
```
bogdandrutu, line 359: Not sure if we want to commit to this. Maybe as an "optional"? The helper will be remove anyway next release, so the dependency has to change.

bogdandrutu:
> do you have a good resource about how to do feature flags in the collector? If so, I'll add a link.

https://github.com/open-telemetry/opentelemetry-collector/tree/main/service/featuregate

But as I explained, these are not helping with API changes.

CONTRIBUTING.md
dashpole: Agreed.  I meant it WRT functionality, or configuration.

CONTRIBUTING.md
```diff
@@ -345,6 +345,18 @@ with clear instructions on how to install the required libraries.
...
+### Breaking changes
+
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
+changelog, possibly with a line instructing users how to move forward.
+
+We also strive to perform breaking changes in two stages, or using feature flags: 
+
+- when we need to remove something, we mark a feature as deprecated in one version, and remove it in the next one
+- when renaming types, functions, or attributes, we create the new name first, deprecating the old one in one
```
dmitryax, line 357: Maybe we should we explicitly allow moving things across packages and package renames? Otherwise it'll be very difficult to work towards Collector 1.0 given amount of changes we might need to do based on the today's SIG discussions.

dmitryax (commented):
Should we explicitly allow package

bogdandrutu:
Question: What is the expected way to change a func signature (breaking change, but assuming it just returns 2 values instead of 1) in the new "world":
1. Create a new `func` and deprecate the old one - downside probably we will not use the best name for the func since that is the old func. Can also be solved by doing this in multiple steps, add a temporary (not ideal name), remove the deprecated one, deprecate the temporary and reuse the good name, remove temporary.
2. Accept some "breaking changes" without deprecation.

CONTRIBUTING.md
```diff
@@ -345,6 +345,18 @@ with clear instructions on how to install the required libraries.
 Furthermore, if your package requires CGO, it MUST be able to compile and operate in a no op mode
 or report a warning back to the collector with a clear error saying CGO is required to work.
 
+### Breaking changes
```
bogdandrutu, line 348: Question: What is the expected way to change a func signature (breaking change, but assuming it just returns 2 values instead of 1) in the new "world":
1. Create a new `func` and deprecate the old one - downside probably we will not use the best name for the func since that is the old func. Can also be solved by doing this in multiple steps, add a temporary (not ideal name), remove the deprecated one, deprecate the temporary and reuse the good name, remove temporary.
2. Accept some "breaking changes" without deprecation.

deleted the initial comment, and moved it here to allow threads :D

CONTRIBUTING.md
Aneurysm9: Do we have examples of such helpers?  It's not clear to me what this would look like.

CONTRIBUTING.md
Aneurysm9: If we move types or functions to a new package we should be able to leave behind a forwarding type or function.  If we can't do that because it would introduce circular dependencies then we do it the other way around and have the new type/function reference the prior location until it is removed and the definition moved to the new location.

dmitryax: Yes, but it still adds a lot of work for us to maintain this approach while this kind of changes is a simple package rename on the client side. I would vote for allowing these changes, let's see what other folks think

Aneurysm9: pdata is mostly generated code.  We can generate it twice in different packages.  

I think that at this point in the lifecycle of the project it is worthwhile to consider that adding burdens on those proposing making breaking changes is a good thing.  It should perhaps no longer be the case that we make breaking changes in an offhand manner because it is easy to do.

CONTRIBUTING.md
mx-psi: Maybe we should have an "API breaking changes" section and a "End-user facing breaking changes" section?

CONTRIBUTING.md
jpkrohling: I had the configauth example in mind, where we have a ServerAuthenticator interface, as well as a `NewServerAuthenticator` function that returns an implementation based on the provided options.

https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/default_serverauthenticator.go#L59-L71

CONTRIBUTING.md
jpkrohling: How would we do it after v1? We'd likely have a new function name that is composed by the ideal name + differences in behavior. Like:

- step 1 (current API): `func GetFoo() Foo`
- step 2 (current+1): needs to return error, then:
- `func MustGetFoo() Foo` (panics on error, is the replacement for `GetFoo() Foo`, which gets deprecated in favor of the new MustGetFoo
- step 3 (current+2): `func GetFoo() (Foo, error)` is added, `func MustGetFoo() Foo` is not touched

CONTRIBUTING.md
jpkrohling: I added something about it to the doc.

CONTRIBUTING.md
jpkrohling: I added a paragraph about user-facing changes and feature flags, removing it from this paragraph here. Let me know if you think this should be on its own section.

CONTRIBUTING.md
jpkrohling: If we have agreement, I'll add this example to the doc.

CONTRIBUTING.md
mx-psi: Looks good to me like this!

CONTRIBUTING.md
bogdandrutu: Fine with this. After v1 I will have to survive with the bad name :))

CONTRIBUTING.md
```diff
@@ -345,6 +345,25 @@ with clear instructions on how to install the required libraries.
...
```
bogdandrutu, line 348: What about changing an interface. Adding a func? Do we want to go with optional interface? I don't think there is a way if the interface is used in multiple funcs as argument, because you need to deprecate every one.

jpkrohling: I don't have the perfect answer, but this question is relevant also for post-v1 and the answer should be the same.

bogdandrutu: Post v1 I think the willingness to pay any cost is a "given" :))

CONTRIBUTING.md
```diff
@@ -345,6 +345,25 @@ with clear instructions on how to install the required libraries.
...
+
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
+changelog, possibly with a line instructing users how to move forward.
+
+We also strive to perform breaking changes in two stages: 
+
+- when we need to remove something, we mark a feature as deprecated in one version, and remove it in the next one
```
dashpole, line 356: Like the point below ("we create the new name first"), I think we should introduce the replacement code (if it is being replaced) before we deprecate existing functionality?

CONTRIBUTING.md
```diff
@@ -345,6 +345,25 @@ with clear instructions on how to install the required libraries.
 Furthermore, if your package requires CGO, it MUST be able to compile and operate in a no op mode
 or report a warning back to the collector with a clear error saying CGO is required to work.
 
+### Breaking changes
+
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
+changelog, possibly with a line instructing users how to move forward.
```
Aneurysm9, line 352:
```suggestion
changelog, preferably with a line instructing users how to move forward.
```

or

```suggestion
changelog, and SHOULD include a line instructing users how to move forward.
```

CONTRIBUTING.md
```diff
@@ -345,6 +345,25 @@ with clear instructions on how to install the required libraries.
 Furthermore, if your package requires CGO, it MUST be able to compile and operate in a no op mode
 or report a warning back to the collector with a clear error saying CGO is required to work.
 
+### Breaking changes
+
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it should be clearly marked in the
```
Aneurysm9, line 351:
```suggestion
without a good reason. Hence, when a change is known to cause a breaking change, it MUST be clearly marked in the
```

CONTRIBUTING.md
Aneurysm9: I very much agree with this, though it is attached to a line about removing indentifiers.  If we're purely removing something without replacement (and with no intent to ever offer a replacement) then there is nothing to have in place before the deprecation notice.

If, however, we're removing something to be replaced with something else, even if it is not a renaming but a change of design, we should definitely have the replacement in place before removing the existing code.  We should also, where possible, change the existing implementation to forward to the new implementation, or re-implement in terms of the new model, before removal.

CONTRIBUTING.md
jpkrohling: Should I add something to the PR to address this?

CONTRIBUTING.md
```diff
@@ -345,6 +345,25 @@ with clear instructions on how to install the required libraries.
...
+changelog, possibly with a line instructing users how to move forward.
+
+We also strive to perform breaking changes in two stages: 
+
+- when we need to remove something, we mark a feature as deprecated in one version, and remove it in the next one
+- when renaming types, functions, or attributes, we create the new name first, deprecating the old one in one
+version, removing it in the next version
+- when creating new interfaces, provide helpers so that consumers can be forward compatible without much effort. One
```
Aneurysm9, line 359: Interfaces (generally) [cannot be added to](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20220114162006-9d54fb35363c/apidiff#readme-interfaces) in a backward compatible manner and I think we should avoid contortions to try to give the appearance that they can.  I also think that we should be [very careful about adding unexported methods](https://github.com/open-telemetry/opentelemetry-collector/issues/4674) to interfaces in an attempt to reserve the ability to add to them later.

The `ServerAuthenticator` helper seems reasonable in that it provides sane default behavior while not being the only way to implement the interface, but some of it is now unnecessary since [nil `StartFunc` and `ShutdownFunc`](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/component/componenthelper/component.go#L24) have that sane default behavior and an authenticator with default AuthN functionality is useless.  If we were creating this interface today, would we still want this helper or would the existing ways of implementing an `Extension` be sufficient?

I don't think this is the place to discuss the merits of that particular interface, but I do think it might be the case that guidance for how to handle interface compatibility is more complicated and should be discussed and added separately from the rest of this guidance which is more straightforward.

CONTRIBUTING.md
```diff
@@ -345,6 +345,25 @@ with clear instructions on how to install the required libraries.
...
+example can be found in the `configauth` package, where a [`ServerAuthenticator`](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/serverauth.go#L28-L40)
+interface exists, we as well as a [helper](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/default_serverauthenticator.go#L59-L71)
+that most implementations should use.
+
+When deprecating a feature affecting end-users, consider first deprecating the feature, then hiding it behind
```
Aneurysm9, line 364: A feature flag should probably be added with the deprecation notice, but not yet disabling the feature.  Then the flag default can be set to disable the feature by default, and finally the flag and the feature removed.

CONTRIBUTING.md
jpkrohling: I see where it was ambiguous and have addressed that.

CONTRIBUTING.md
jpkrohling: This is in line with what I had in mind. I changed the text to be clearer.

CONTRIBUTING.md
jpkrohling: I agree with you, but in practical terms, when developing an authenticator, I might not care whether my authenticator is a `ServerAuthenticator` or a `ValidatableServerAuthenticator`. I'd prefer to only provide an `Authenticate` function.

Packages like `confighttp` and `configgrpc` might care about the validation aspect of the component, and `configauth` might provide a default `Validate` that complies with that new interface. IMO, this is the value of providing those helpers: new interfaces can be composed and used in consuming places without having to adapt all the current implementations of the simpler interface.

Aneurysm9: In the case of validation I assume the default behavior if no validator is provided is to pass validation.  That could also be achieved by testing for an implementation of a `Validater` interface and only performing validation if it is implemented.  This puts the burden on the consumer, who is in the best position to know whether the new capability is optional (thus taking the interface testing approach) or mandatory (and thus taking an interface composition approach and requiring a `ValidatableServerAuthenticator`).

Aneurysm9: I am strongly opposed to this being a `MUST`, as it is in the current revision.  I think if the discussion of interfaces were removed from this PR and handled separately that I could approve this PR, but as it stands I cannot.

CONTRIBUTING.md
dashpole: I'm still a bit unclear... Which category (removal vs renaming) would general refactoring fall into? For example, if I split `ListAndWatch()` into `List()` and `Watch()`, would that be covered by the removal or renaming category here?  Does "renaming" cover all refactoring for which there is a similar replacement, or are refactors considered removals + new additions?  What would this imply for https://github.com/open-telemetry/opentelemetry-collector/pull/4672#issuecomment-1017804080?

CONTRIBUTING.md
jpkrohling: Alright, I just removed it from here. We can discuss about it and add a better version in the future.

CONTRIBUTING.md
jpkrohling: Do you think the new version is satisfactory?

CONTRIBUTING.md
```diff
@@ -356,10 +356,6 @@ We also strive to perform breaking changes in two stages:
...
 - when renaming types, functions, or attributes, we MUST create the new name first and MUST deprecate the old one in one
 version, and MAY remove it in the next version. For simple renames, the old name SHALL call the new one.
-- when creating new interfaces, helpers MUST be provided so that consumers can be forward compatible without much effort. One
-example can be found in the `configauth` package, where a [`ServerAuthenticator`](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/serverauth.go#L28-L40)
-interface exists, we as well as a [helper](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/config/configauth/default_serverauthenticator.go#L59-L71)
-that most implementations should use.
 
 When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind
 a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/) in the next version,
```
bogdandrutu, line 361: Is this step necessary? Waiting 2 versions is not enough?

bogdandrutu:
@codeboten lots of changes since your review, I asked you to do another round please.

CONTRIBUTING.md
Aneurysm9: Deprecating features where we impact end users should be where we're most conservative.  I think only two releases between announcement and removal may not be enough time.  The [project guidelines on long-term support](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#long-term-support) are measured in years, not weeks.  We're not there yet, I don't think, but I do think we're beyond the point where a very short notice period is appropriate.  Distributions like ADOT that release once per month could entirely miss the notice period if it is a single release.

CONTRIBUTING.md
Aneurysm9: I think it's still a little unclear that the feature gate should be added at the same time as the deprecation notice.  That would give users the option to toggle the gate to see how it functions with the feature removed before that becomes the default.  This somewhat extends the time available to detect issues with removing the feature and gives users an option beyond simply filing an issue saying that removing the deprecated behavior might impact them; they can actually test it and know.

CONTRIBUTING.md
bogdandrutu: The current text anyway limits to 2 releases, is just that in the second release you have a gate to enable that back... So what I am asking is if the gate is necessary since we remove it in 2 releases anyway.

Aneurysm9: We should then `s/the next/a subsequent/` as I don't think a single release window with notice is sufficient.

CONTRIBUTING.md
jpkrohling: > give users the option to toggle the gate to see how it functions with the feature removed before that becomes the default

True, I haven't thought about that. How about this, each step being a version?

1. mark the feature as deprecated, add feature flag with the feature enabled by default
2. change the feature flag to disable it by default
3. remove the flag

CONTRIBUTING.md
jpkrohling: The rules are the same: MUST mark as deprecated in one version, MAY remove in the next. The difference is that renaming causes the deprecated code to call the new one. So, `ListAndWatch` gets deprecated and calls `List()` and `Watch()`, until it eventually gets removed.

CONTRIBUTING.md
```diff
@@ -345,6 +345,38 @@ with clear instructions on how to install the required libraries.
...
 
+### Breaking changes
+
+Whenever possible, we adhere to semver as our minimum standards. Even before v1, we strive to not break compatibility
+without a good reason. Hence, when a change is known to cause a breaking change, it MUST be clearly marked in the
+changelog, and SHOULD include a line instructing users how to move forward.
+
+We also strive to perform breaking changes in two stages: 
+
+- when we need to remove something, we MUST mark a feature as deprecated in one version, and MAY remove it in the next one
```
tigrannajaryan, line 356: Should we also indicate deprecations in the changelog?

CONTRIBUTING.md
```diff
@@ -345,6 +345,38 @@ with clear instructions on how to install the required libraries.
...
+- when renaming types, functions, or attributes, we MUST create the new name first and MUST deprecate the old one in one
```
tigrannajaryan, line 357: Is this 3 different steps in 3 versions or 2 steps in 2 versions? Wording is not clear.

Also in some rare case it may not possible, e.g. we have A and B and need to swap the names (I can't think of a more realistic use case, but I suspect there may be). I think SHOULD is more appropriate than MUST here.

CONTRIBUTING.md
```diff
@@ -345,6 +345,38 @@ with clear instructions on how to install the required libraries.
...
+
+When deprecating a feature affecting end-users, consider first deprecating the feature in one version, then hiding it behind
+a [feature flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
+in a subsequent version, and eventually removing it after yet another version.
+
+#### Example #1 - Renaming a function
+
+1. Current version `v0.N` has `func GetFoo() Bar`
+1. We now decided that `GetBar` is a better name. As such, on `v0.N+1` we add a new `func GetBar() Bar` function,
+changing the existing `func GetFoo() Bar` to be an alias of the new function. Additionally, a log entry with a warning
```
tigrannajaryan, line 368: Log entry may be excessive in some case (if it is in the hot path) and can be against existing guidelines about logging. Let's not overdo it, maybe recommend it as a possibility if it is not in the hot path.

CONTRIBUTING.md
```diff
@@ -345,6 +345,38 @@ with clear instructions on how to install the required libraries.
...
+is added to the old function, along with an entry to the changelog.
+1. On `v0.N+2`, we MAY remove `func GetFoo() Bar`.
+
+#### Example #2 - Changing the return values of a function
```
tigrannajaryan, line 372: This probably can be generalized for any function signature change, not just return value change or perhaps add one more example to show parameter type/list changes.

CONTRIBUTING.md
jpkrohling: Yes, and it's mentioned earlier in this section:

> Hence, when a change is known to cause a breaking change, it MUST be clearly marked in the
> changelog, and SHOULD include a line instructing users how to move forward

CONTRIBUTING.md
jpkrohling: Two steps in two versions. I changed the text a bit to try to clarify this.

> Also in some rare case it may not possible, e.g. we have A and B and need to swap the names (I can't think of a more realistic use case, but I suspect there may be). I think SHOULD is more appropriate than MUST here.

How about adding a paragraph about handling exceptions to the rules in general?

CONTRIBUTING.md
jpkrohling: Agree. I added the word "perhaps" there. Is this sufficient?

CONTRIBUTING.md
Aneurysm9: I think Tigran is asking about something slightly different.  That language would apply when a feature is removed.  But before that it should be deprecated.  In many cases there will be a replacement feature added concurrent with the deprecation, and the two should be announced together in the changelog:

> - Added feature X.  This will replace feature Y, which is now deprecated.

If something is being removed with no replacement, though, we should be sure to note its deprecation and pending removal in the changelog.

CONTRIBUTING.md
Aneurysm9: 3. remove the flag and feature

but, yeah, that's in alignment with what I think will work best.

CONTRIBUTING.md
jpkrohling: I'm sorry, I think I'm still missing something. Can you provide a suggestion to address this? I still see the current text as addressing both situations, requiring changelog entries.

CONTRIBUTING.md
Aneurysm9: I think it is sufficiently covered.

CONTRIBUTING.md
jpkrohling: Done. I changed the wording a bit, so, I'd appreciate another review of this part.

CONTRIBUTING.md
```diff
@@ -345,6 +345,41 @@ with clear instructions on how to install the required libraries.
 Furthermore, if your package requires CGO, it MUST be able to compile and operate in a no op mode
 or report a warning back to the collector with a clear error saying CGO is required to work.
 
+### Breaking changes
```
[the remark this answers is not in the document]
jpkrohling: Tigran had a comment going the other direction (https://github.com/open-telemetry/opentelemetry-collector/pull/4712#discussion_r789689021): what to do when it's impossible or unfeasible to do it as we described here? I'll add a section about exceptions.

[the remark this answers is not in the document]
jpkrohling: I added a new paragraph about the exceptions.

CONTRIBUTING.md
jpkrohling: I added a new paragraph about the exceptions.

CONTRIBUTING.md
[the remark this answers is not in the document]
Aneurysm9: +1

Rules are meant to be broken, when you know why they exist and why they're the wrong thing to do in a given situation.  I think the callout about consensus among approvers being necessary to make an exception is a good one.  We shouldn't make exceptions with a single approval. 

Perhaps not worth mentioning in this doc, but something we should all keep in mind is that we should take care to ensure that exceptions do not become the norm.  If that is happening then the policy is wrong or we're not being faithful to it and our users.

CONTRIBUTING.md
tigrannajaryan: Consider scenario:
1. Feature A exists.
1. Feature B is added.
1. After a while Feature B is found to be sufficient to cover use cases where Feature A is used.
1. Feature A is marked deprecated.
1. Feature A is removed.

I think we need to have a changelog about deprecation in step 4. Deprecation _alone_ is noteworthy enough to warrant a changelog entry even if it is not happening simultaneously with some other change.

tigrannajaryan (approved):
LGTM. This reflects our intent. Any further clarifications can be added via additional PRs.

tigrannajaryan:
@jpkrohling I can merge this once the comments are resolved.

CONTRIBUTING.md
jpkrohling: So, this?

> - when a feature is being replaced in favor of an existing one, we MUST mark a feature as deprecated in one version, and
>  MAY remove it in the next one

CONTRIBUTING.md
jpkrohling: Can this be marked as resolved, or is there still something I need to clarify in the doc?

CONTRIBUTING.md
jpkrohling: Done

jpkrohling:
PR updated. I think there's only one open comment about features being replaced by existing ones. Not sure there's anything to do in the other open comment.

CONTRIBUTING.md
dashpole: Can you expand renaming to be clear it covers refactoring as well?
`when renaming or refactoring types, functions, or attributes`...

jpkrohling: Got it! Done.

jpkrohling:
> Not sure there's anything to do in the other open comment.

That one was resolved, only one comment is still open but I think it has been addressed already.

CONTRIBUTING.md
Aneurysm9: `s/the next one/a subsequent version/`

While we are presently trying to move as fast as possible and have a reasonable ability to expect users to adapt to change, these guidelines should outlast this phase, particularly when it comes to features or behavior for end users that may not be compilation breaking.  We should be prepared to have deprecation periods that last not days or weeks, but many months spanning several releases.

jpkrohling: My bad, I used "a subsequent one" elsewhere and should have used it here as well. Fixed now.

Aneurysm9 (approved):

dashpole (approved):

CONTRIBUTING.md
```diff
@@ -345,6 +345,65 @@ with clear instructions on how to install the required libraries.
...
+   context: `func GetFooWithContext(context.Context) Foo`. We release this in `v0.N+1`, deprecating the existing `func
+   GetFoo() Foo` with it, adding an entry to the changelog and perhaps a log entry with a warning. The existing `func
+   GetFoo() Foo` is changed to call `func GetFooWithContext(context.Background()) Foo`.
+1. On `v0.N+2`, we change `func GetFoo() Foo` to `func GetFoo(context.Context) Foo` if desired or remove it entirely if
+   needed.
+
+#### Exceptions
+
+While the above is what we strive to follow, we acknowledge that some changes might be unfeasible to achieve in a
+non-breaking manner. Exceptions to the outlined rules are acceptable if consensus can be obtained from approvers in the
```
bogdandrutu, line 405: should we define what consensus means? half + 1? You are defining something here that is not well explained.

jpkrohling: I understand "consensus" as unanimity, but I'm not a native speaker. My intention was that we, as engineers, should be able to find a solution that satisfies everyone, even if it means making compromises along the way.

Aneurysm9: I don't think consensus needs to mean unanimity, but generally at least that there are not strong/formal objections.  I'm a fan of the [w3c consensus model](https://www.w3.org/2021/Process-20211102/#Consensus).

I think that we will not always be able to satisfy everyone, but we must ensure that we have exhausted reasonable attempts to do so and have policies and procedures for dealing with formal objections.  We should generally strive to minimize opposition and look for consensus options that have the most support and least resistance.

jpkrohling: Do we need to specify in this doc what consensus means to us? Do we know what it means to us?

Aneurysm9: I would prefer that consensus be defined at a higher layer than contributing docs.  It is a concept that should apply to most things we do and applies not only to this SIG.  Perhaps we need to define for ourselves how to apply the general concept in particular situations, but that should probably be done in reference to a globally used definition.

bogdandrutu: Without defining what it means, how can we apply this?

jpkrohling: We might be overthinking this, but here's a proposal: 

>  A reason for requesting the exception MUST be given in the pull request. Until unanimity
> is obtained, approvers and maintainers are encouraged to discuss the issue at hand. If a consensus (unanimity) cannot be
> obtained, the maintainers are then tasked in getting a decision using its regular means (voting, TC help, ...).

jpkrohling: @bogdandrutu, is this proposal acceptable?

bogdandrutu: Ok, sounds reasonable as long as the approvers are reasonable :)

bogdandrutu (approved):

Issue #4681, opened 2022-01-13
[Proposal] Try to move helpers to create [*]Factory types in `component` directly.


bogdandrutu:
There is a draft PR https://github.com/open-telemetry/opentelemetry-collector/pull/4679, which shows that it can improve things, also it can definitely help with package dependencies graph (component -> internal, helpers -> {component, internal}), which does not allow us for the moment to split in separate modules the `component` package. This proposal does not suggest to split in separate modules (clarification for people that like to find different interpretation), just lists this impediment for future possibilities.

bogdandrutu:
From @dmitryax, see comment https://github.com/open-telemetry/opentelemetry-collector/pull/4679#issuecomment-1012585449

> Wrapping it all in a struct maybe seems not as clean as an interface, but straighten the dependency chain is much valuable I believe. Right now it's very convoluted.

PR #4719, opened 2022-01-20
Use directly the consumer helper funcs in the other helpers packages.

bogdandrutu:
Try to understand if the Component helper is necessary, or only having the helper funcs is enough.

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/4681


Aneurysm9:
I think this looks like an improvement.  Does it even break anything or is it just removing a layer of indirection?

jpkrohling (approved):

bogdandrutu:
>  Does it even break anything or is it just removing a layer of indirection?

@Aneurysm9 not all my PRs are breaking things :))

Issue #4665, opened 2022-01-11
Replace "configmapprovider.Retrieved" with returning config.Map and closeFunc?


bogdandrutu:
This idea was considered initially, but because the "Watcher" was an optional interface (not a func passed as argument) it couldn't be implemented in a nice way. Now that we changed the way the "watcher" is passed we can simplify the Provider interface.

**Current:**
```
type Provider interface {
	Retrieve(ctx context.Context, location string, watcher WatcherFunc) (Retrieved, error)
	Shutdown(ctx context.Context) error
}

type Retrieved interface {
	Get(ctx context.Context) (*config.Map, error)

	Close(ctx context.Context) error
}
```

**After:**
```
type Provider interface {
	// CloseFunc can be replaced with Done/DoneFunc/StopWatchFunc.
	Retrieve(ctx context.Context, location string, watcher WatcherFunc) (*config.Map, CloseFunc, error)  
	Shutdown(ctx context.Context) error
}
```

I believe this is what @Aneurysm9 confusion/idea was, but I cannot speak for them.

Aneurysm9:
This is somewhat more like what I'd expect, but it's also not clear to me why the `Retrieved` needed to be closed and then obtained again.  Or, in this new model, why invoking `watcher` should cause `CloseFunc` to be invoked and then `Retrieve`.  If the provider knows that the configuration has changed, can it provide the new configuration directly to `watcher`?

bogdandrutu:
@Aneurysm9 not really, because we may have multiple Providers, and if one of them discover (watches) a change we need to stop watching for changes in the other ones and retrieve the whole config again. The reason to retrieve the entire config and not just the one that trigger the watcher (a simple solution would be to keep the result of all the retrieves and redo the operations, but cannot happen this way, see next) is because if we want to support the "embedded configs" (allow users to have in their config files "receiver::awsxray::token::${voltdb:path/to/my/token}" the changed config may not need some of the retrieved values, so keeping a watcher will be useless.

bogdandrutu:
Resolved. `Retrieved` is a struct now.

PR #4721, opened 2022-01-21
Replace configmapprovider.Retrieved with returning config.Map and StopWatchFunc

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4665

Alternative considered: was to replace configmapprovider.Retrieved with a struct.


config/configmapprovider/provider.go
```diff
@@ -66,6 +68,15 @@ type Provider interface {
 	Shutdown(ctx context.Context) error
 }
 
+// CloseFunc signals that the configuration for which it was used to retrieve values is
+// no longer in use and should close and release any watchers that it may have created.
+//
+// Should block until all resources are closed, and guarantee that `onChange` is not
+// going to be called after it returns except when `ctx` is cancelled.
+//
+// TODO: Find a good name for this func.
```
bogdandrutu, line 77: Any ideas?

bogdandrutu (commented):
Will fix the changelog when I get an approver :)

config/configmapprovider/provider.go
Aneurysm9: This feels like a reasonable name.  Externally, it'll be referenced as `configmapprovider.CloseFunc`, which is a bit of a mouthful but has all the relevant info.

If this doesn't feel right, what about a `Close` method on the `Provider` interface?

Aneurysm9 (commented):
Generally looks good.  I think it could maybe be simplified further by making the `CloseFunc` a `Close` method on the `Provider` interface.  Would be interested in hearing if you had reasons for not doing that.

bogdandrutu:
@Aneurysm9 close right now is designed to stop/close (maybe stop watching func is a better name because of this) watching for the specified "location" in the retrieve, not for all the locations. So we can move it to provider if we change the behavior as well, but that has less flexibility for the future.

Aneurysm9:
> @Aneurysm9 close right now is designed to stop/close (maybe stop watching func is a better name because of this) watching for the specified "location" in the retrieve, not for all the locations. So we can move it to provider if we change the behavior as well, but that has less flexibility for the future.

Ok, I think I've got it now.  I think for some reason I was assuming earlier binding of the location.  This does seem like the right API given that the stop/close is scoped to the retrieved location and the provider is not.

> Please read the API description (what is now merged, and what is here) and let me know if that is not clear about the behavior I mentioned, because we need to improve it otherwise.

The description of "typical usage" of the `Provider` interface does need to be updated as it no longer reflects the expected usage pattern.  But I don't think that my misunderstanding is a fault of the description.

config/configmapprovider/provider.go
bogdandrutu: I tried `StopWatchFunc`, if not better I can revert.

bogdandrutu:
> The description of "typical usage" of the Provider interface does need to be updated as it no longer reflects the expected usage pattern. 

Done.

config/configmapprovider/provider.go
Aneurysm9: That makes me think of a timer/stopwatch.  Maybe `EndWatchFunc`?

config/configmapprovider/provider.go
```diff
@@ -24,15 +26,13 @@ import (
 //
 // The typical usage is the following:
 //
-//		r := mapProvider.Retrieve("file:/path/to/config")
-//		r.Get()
+//		cfg, stopWatch, err := mapProvider.Retrieve("file:/path/to/config")
```
Aneurysm9, line 29:
```suggestion
//		cfg, stopWatch, err := mapProvider.Retrieve(ctx, "file:/path/to/config", onChange)
```

config/configmapprovider/provider.go
```diff
@@ -24,15 +26,13 @@ import (
...
 //		// wait for onChange() to be called.
-//		r.Close()
-//		r = mapProvider.Retrieve("file:/path/to/config")
-//		r.Get()
+//		if stopWatch != nil { stopWatch() }
+//		cfg, stopWatch, err = mapProvider.Retrieve("file:/path/to/config")
```
Aneurysm9, line 32:
```suggestion
//		cfg, stopWatch, err = mapProvider.Retrieve(ctx, "file:/path/to/config", onChange)
```

config/configmapprovider/provider.go
bogdandrutu: Happy to try that, between `EndWatchFunc` or `CloseWatchFunc` or `DoneFunc` (does not seem to end, but saw this pattern)?

Aneurysm9: `DoneFunc` also seems good.  I think I would be happy with either that or `EndWatchFunc`.  Would like to know what others think.

bogdandrutu: FYI: Usually (for unknown reasons) you don't get feedback unless you tag people :) @tigrannajaryan @codeboten @jpkrohling @dmitryax

config/configmapprovider/provider.go
bogdandrutu: Done.

config/configmapprovider/provider.go
bogdandrutu: Done.

config/configmapprovider/provider.go
bogdandrutu: Tried `EndWatchFunc` looks ok...

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Deprecate `configtelemetry.Level.Set()` (#4700)
 - Remove support to some arches and platforms from `ocb` (opentelemetry-collector-builder) (#4710)
 - Remove deprecated legacy path ("v1/trace") support for otlp http receiver (#4720)
+- Replace configmapprovider.Retrieved with returning config.Map and StopWatchFunc (#4721)
```
Aneurysm9, line 22:
```suggestion
- Replace configmapprovider.Retrieved with returning config.Map and EndWatchFunc (#4721)
```

config/configmapprovider/provider.go
```diff
@@ -24,15 +26,13 @@ import (
 //
 // The typical usage is the following:
 //
-//		r := mapProvider.Retrieve("file:/path/to/config")
-//		r.Get()
+//		cfg, stopWatch, err := mapProvider.Retrieve(ctx, "file:/path/to/config", onChange)
```
Aneurysm9, line 29:
```suggestion
//		cfg, endWatch, err := mapProvider.Retrieve(ctx, "file:/path/to/config", onChange)
```

GitHub won't let me make a single suggestion here, but apply to the rest of this docblock.

Aneurysm9 (approved):

config/configmapprovider/file.go
```diff
@@ -47,25 +47,23 @@ func NewFile() Provider {
 	return &fileMapProvider{}
 }
 
-func (fmp *fileMapProvider) Retrieve(_ context.Context, location string, _ WatcherFunc) (Retrieved, error) {
+func (fmp *fileMapProvider) Retrieve(_ context.Context, location string, _ WatcherFunc) (*config.Map, EndWatchFunc, error) {
```
jpkrohling, line 50: Given this is an API, would it be better to return a specific type containing the map and the func? A func that returns 3 values leaves a bad aftertaste, especially for a pre-v1 API, in my opinion.

bogdandrutu: So you think the alternative consideration is better. I think having a simple struct with 2 public members is as good as this and definitely better than an interface (as it was before).

jpkrohling: My reasoning is that it's easy to add a new field to the return object type if needed, without breaking current callers, while if you add an extra value to the return values, you break them. For some reason, I think it's easier to start needing 4 values when you have 3 than needing 3 when you have 2 (one of them being `error`).

bogdandrutu: Sounds reasonable to me, will either change this or maybe do another one to look at that option as well.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/pull/4789 here is the alternative, let me know.

bogdandrutu:
Based on the comments in the alternative PR https://github.com/open-telemetry/opentelemetry-collector/pull/4789, closing this.

Issue #2701, opened 2021-03-14
Support Multiple Config Files/An import statement


PettitWesley:
### Feature Request: Configuration can be split into multiple files

It'd be ideal if the collector could work off of a set of configuration files rather than a single one. There are probably two basic ways to accomplish this:
1. Continue to support a single config file but allow it to import other others. This is the same as how your main script might import other code. You have an import statement somewhere in the config file and it references a path to find other config.
2. Support a list of config files in the entrypoint for the collector. 

The way this should work is that the collector should evaluate each config file and merge them.

### Why is this useful?

As the collector grows, the configurations that users create will get more and more complex. For the same reasons that in school your prof taught you not to put all your C code in one file, its generally useful to organize configuration logically into multiple files.

This is also useful when you have multiple people/entities working on building the final configuration. Consider the following scenario: A few years from now when OTel has truly eaten the world of instrumentation, Ellen and Akshay are building the config for their application. The config for their unified collector must container the following:
- Ellen's logging config, which includes multiple destinations, custom decorations, and parsing
- Akshay's trace config, with its own complex set of processor definitions
- Configuration for [AWS Container Insights metrics, which is very lengthy as well](https://github.com/aws-observability/aws-otel-collector/blob/main/config/ecs/container-insights/otel-task-metrics-config.yaml).
- Some configuration for collecting security logs from their platform, which they have received from their company's centralized security team. (In this future world, the collector can also be used to collect security events). 

Could they concatenate all of these configs together into one file? Definitely. It'd be inconvenient though. Note that each piece of the config has a different owner. This means that updates might come independently. Thus it's useful if they can remain in separate files. Regardless of how the team stores and deploys their config, it'll be easier to manage if they can keep it in multiple files.

### Examples in other projects

Fluentd has `@include` and Fluent Bit has `@INCLUDE` for importing configuration from other files. https://docs.fluentd.org/configuration/config-file#6-reuse-your-config-the-include-directive

This is very commonly used. 

### Merging configs

This bit which will need some discussion. 

Ideally, the configurations should be merge-able. This means I can define a processor in config file A and then reference it in a pipeline in config file B. Basically the final configuration is the concatenation of all the configs.

bogdandrutu:
/cc @pjanotti 

I think the proposed "config source" will address this.

PettitWesley:
@pjanotti Can you give details on this config source thing?

tigrannajaryan:
[...]

PR #4727, opened 2022-01-21
Enable end users to provide multiple files for config location

bogdandrutu:
Restriction (compatible with `--set`): Per flag entry can be defined only one location, to define multiple locations use multiple flag entries (which will be merged in the given order):

```bash
./otelcol --cofig=file:/path/to/first/file --config=file:/path/to/second/file
```

Fixes: https://github.com/open-telemetry/opentelemetry-collector/issues/2701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Deprecate `configtelemetry.Level.Set()` (#4700)
 - Remove support to some arches and platforms from `ocb` (opentelemetry-collector-builder) (#4710)
 - Remove deprecated legacy path ("v1/trace") support for otlp http receiver (#4720)
+- Change the `service.NewDefaultConfigProvider` to accept a slice of strings (#4727).
 
 ## 💡 Enhancements 💡
 
@@ -29,6 +30,7 @@
 - Move `compression.go` into `confighttp.go` to internalize functions in `compression.go` file. (#4651)
   - create `configcompression` package to manage compression methods in `confighttp` and `configgrpc`
 - Add support for cgroupv2 memory limit (#4654)
+- Enable end users to provide multiple files for config location (#4727)
 
 ## 🧰 Bug fixes 🧰
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - Deprecate `configtelemetry.Level.Set()` (#4700)
 - Remove support to some arches and platforms from `ocb` (opentelemetry-collector-builder) (#4710)
 - Remove deprecated legacy path ("v1/trace") support for otlp http receiver (#4720)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Updated confighttp `ToClient` to support passing telemetry settings for instrumenting otlphttp exporter(#4449)
 - Deprecate `configtelemetry.Level.Set()` (#4700)
 - Remove support to some arches and platforms from `ocb` (opentelemetry-collector-builder) (#4710)
+- Change the `service.NewDefaultConfigProvider` to accept a slice of strings (#4727).
```
Aneurysm9, line 22:
```suggestion
- Change the `service.NewDefaultConfigProvider` to accept a slice of location strings (#4727).
```

service/flags.go
```diff
@@ -39,14 +37,15 @@ func (s *stringArrayValue) Set(val string) error {
...
+	return "[" + strings.Join(s.values, ", ") + "]"
 }
 
 func flags() *flag.FlagSet {
 	flagSet := new(flag.FlagSet)
 	featuregate.Flags(flagSet)
 
-	configFlag = flagSet.String("config", defaultConfig, "Path to the config file")
+	flagSet.Var(configFlag, "config", "Locations to the config file(s), note that only a"+
+		" single (first) location can be set per flag entry e.g. `-config=file:/path/to/first --config=file:path/to/second`.")
```
Aneurysm9, line 48:
```suggestion
		" single location can be set per flag entry e.g. `-config=file:/path/to/first --config=file:path/to/second`.")
```

I think the `(first)` bit was in relation to changing arrays in configs through `--set`.

bogdandrutu: Yeah, copy paste ftw :)))

Aneurysm9 (approved):

jpkrohling (approved):
This is probably our last breaking change before the new procedure :-) LGTM, especially as @Aneurysm9 has approved, signaling that one of the known downstream distributions is OK with this change.

Aneurysm9:
> This is probably our last breaking change before the new procedure :-) LGTM, especially as @Aneurysm9 has approved, signaling that one of the known downstream distributions is OK with this change.

I don't think it breaks behavior or semantics and the API change is easy to account for and there are already breaks in this API in this release, so let's do this! :)

bogdandrutu:
@codeboten @tigrannajaryan I want at least one more +1

codeboten (approved):
👍

bogdandrutu:
> This is probably our last breaking change before the new procedure :-) LGTM, especially as @Aneurysm9 has approved, signaling that one of the known downstream distributions is OK with this change.

You don't want to know how much code I can break in one day :))))))

Issue #4730, opened 2022-01-21
In `service.NewConfigProvider` return an error if len(configLocations) is 0.


bogdandrutu:

noisersup:
I would like to take this task

bogdandrutu:
This is done, the `confmap.Resolver` does that.

PR #4734, opened 2022-01-23
Return an error in service.NewConfigProvider if there is no configLocations

noisersup:
**Description:** 
Add length checking and error returning of locations in `service.NewConfigProvider` and made the whole code compatible with this change.

**Link to tracking Issue:** #4730


noisersup:
I signed it

bogdandrutu (commented):
@jpkrohling based on your example and considering we don't want to have the "Must" version, how do you see this? Here is what I would do to strictly follow your proposal:

version 1:
```golang
MustNewConfigProvider
// Deprecated: use MustNewConfigProvider
NewConfigProvider
```

version 2:
```golang
// Deprecated: we don't want to support this, handle the error don't be lazy
MustNewConfigProvider
NewConfigProvider -> return error
```

version 3:
```golang
NewConfigProvider -> return error
```

bogdandrutu:
@jpkrohling my rule of thumb "do not expose more than needed, and must version is not needed" :)

noisersup:
As I can understand I should add `MustNewConfigProvider` function which will be version of `NewConfigProvider` without len(locations) checking and without error returning.

Any other changes in code I made (Error handling in all places where call to `NewConfigProvider` occures) should stay in pull request?

jpkrohling:
The Must version will just call the NewConfigProvider for now. You'll deprecate NewConfigProvider on this PR and this will be released as part of 0.44.0

On a subsequent PR, to be merged on a version coming after 0.45, you'll change the NewConfigProvider to return an error. Your MustNewConfigProvider will then check for errors and panic in case it sees one. On this version, you deprecate the MustNewConfigProvider.

For 0.46, you can remove the MustNewConfigProvider func.

I know this is convoluted, but this has two main reasons:

1. people currently using the NewConfigProvider deserve having some time to adapt to this change
1. we need to learn and get used to not break things

noisersup:
Ok, that's reasonable.
Should I make another 2 PRs now or it would be better to wait for version updates?

jpkrohling:
A draft PR would be welcome. I'll create two milestones, perhaps we'll remember about those PRs when time comes :-)

bogdandrutu (approved):
This looks good now, let's just make sure this branch is up to date and you resolve merge conflicts. Thanks

bogdandrutu:
Merging this, and will followup with a changelog change.

Issue #4722, opened 2022-01-21
Access to Host in authInterceptor


orloffv:
I use auth interceptor https://github.com/open-telemetry/opentelemetry-collector/blob/04baaae533dc2fea3bfddce1df1c72c3d0694042/config/confighttp/confighttp.go#L333
and i want to get Host inside, but in go(https://pkg.go.dev/net/http#Request) Host property stay separately, not inside header.
Do you have some idea, how to get Host inside auth interceptor?
My idea is add some proxy before collector(like nginx) and add new header like that:
```
X-SMTH-Host: Host;
```

jpkrohling:
We recently had a PR where regular headers were added as attributes to the client.Info:

https://github.com/open-telemetry/opentelemetry-collector/blob/459f1566475f572a9655a1c1c16185aa62e9c3c7/config/confighttp/clientinfohandler.go#L42-L58

Perhaps this code here could store the headers in a var, and enhance it with the host information before calling `client.FromMetadata`? Would you be open to trying it out?

orloffv:
@jpkrohling I understand correctly, you are suggesting that I can try add the code to save the Host in this place?
or i can check, maybe Host already save?

jpkrohling:
> you are suggesting that I can try add the code to save the Host in this place?

Yes :-)

orloffv:
@jpkrohling
```client.FromMetadata``` it's mistake? you mean ```client.FromContext```

orloffv:
@jpkrohling Can you check my pr? https://github.com/open-telemetry/opentelemetry-collector/pull/4736

jpkrohling:
Thank you for the PR!

orloffv:
merged

orloffv:
@jpkrohling oh LOL!

```contextWithClient``` call inside ```ServeHTTP```

https://github.com/open-telemetry/opentelemetry-collector/blob/da91c56fe91999e93621756b87b082e156ef00da/config/confighttp/clientinfohandler.go#L44-L63

https://github.com/open-telemetry/opentelemetry-collector/blob/da91c56fe91999e93621756b87b082e156ef00da/config/confighttp/clientinfohandler.go#L37-L40


```ServeHTTP``` calls after authenticate method :( 

https://github.com/open-telemetry/opentelemetry-collector/blob/04baaae533dc2fea3bfddce1df1c72c3d0694042/config/confighttp/confighttp.go#L331-L341

[...]

PR #4736, opened 2022-01-24
Add Host property for Metadata

orloffv:
**Description:** Add Host property for Metadata

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4722

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 - Move `compression.go` into `confighttp.go` to internalize functions in `compression.go` file. (#4651)
   - create `configcompression` package to manage compression methods in `confighttp` and `configgrpc`
 - Add support for cgroupv2 memory limit (#4654)
+- Add Host property for Metadata (#Add Host property for Metadata)
 
 ## 🧰 Bug fixes 🧰
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   - create `configcompression` package to manage compression methods in `confighttp` and `configgrpc`
 - Add support for cgroupv2 memory limit (#4654)
 - Enable end users to provide multiple files for config location (#4727)
-- Add Host property for Metadata (#Add Host property for Metadata)
+- Add Host property for Metadata (#4736)
 
 ## 🧰 Bug fixes 🧰
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
   - create `configcompression` package to manage compression methods in `confighttp` and `configgrpc`
 - Add support for cgroupv2 memory limit (#4654)
 - Enable end users to provide multiple files for config location (#4727)
-- Add Host property for Metadata (#4736)
+- `client.Info`: Add Host property for Metadata (#4736)
 
 ## 🧰 Bug fixes 🧰
 
--- a/client/client.go
+++ b/client/client.go
@@ -131,6 +131,8 @@ type AuthData interface {
 	GetAttributeNames() []string
 }
 
+const MetadataHostName = "Host"
+
 // NewContext takes an existing context and derives a new context with the
 // client.Info value stored on it.
 func NewContext(ctx context.Context, c Info) context.Context {
--- a/client/client.go
+++ b/client/client.go
@@ -107,6 +107,9 @@ type Info struct {
 	// this connection.
 	Auth AuthData
 
+	// Host fro the incomping request (http.Request.Host)
+	Host string
+
 	// Metadata is the request metadata from the client connecting to this connector.
 	// Experimental: *NOTE* this structure is subject to change or removal in the future.
 	Metadata Metadata
--- a/client/client.go
+++ b/client/client.go
@@ -107,9 +107,6 @@ type Info struct {
 	// this connection.
 	Auth AuthData
 
-	// Host from the incoming request (http.Request.Host)
-	Host string
-
 	// Metadata is the request metadata from the client connecting to this connector.
 	// Experimental: *NOTE* this structure is subject to change or removal in the future.
 	Metadata Metadata
--- a/client/client.go
+++ b/client/client.go
@@ -107,7 +107,7 @@ type Info struct {
 	// this connection.
 	Auth AuthData
 
-	// Host fro the incomping request (http.Request.Host)
+	// Host from the incoming request (http.Request.Host)
 	Host string
 
 	// Metadata is the request metadata from the client connecting to this connector.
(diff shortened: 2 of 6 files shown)


client/client.go
```diff
@@ -107,6 +107,9 @@ type Info struct {
 	// this connection.
 	Auth AuthData
 
+	// Host fro the incomping request (http.Request.Host)
```
neelayu, line 110:
```suggestion
	// Host from the incoming request (http.Request.Host)
```

client/client.go
```diff
@@ -107,6 +107,9 @@ type Info struct {
 	// this connection.
 	Auth AuthData
 
+	// Host from the incoming request (http.Request.Host)
+	Host string
```
jpkrohling, line 111: This should be part of the metadata, not its own top-level attribute.

config/confighttp/clientinfohandler.go
```diff
@@ -53,6 +53,8 @@ func contextWithClient(req *http.Request, includeMetadata bool) context.Context
 		cl.Metadata = client.NewMetadata(req.Header.Clone())
 	}
 
+	cl.Host = req.Host
```
jpkrohling, line 56: Something like:

```go
cl.Metadata["host"] = req.Host
```

Make sure to do something similar for gRPC.

orloffv: I think it's not good idea, because now Metadata equals req.Header
why not to use special property without mutate Metadata(eq headers)?

orloffv: And one more problem
```
    IncludeMetadata bool `mapstructure:"include_metadata,omitempty"`
```

jpkrohling: >  I think it's not good idea, because now Metadata equals req.Header
> why not to use special property without mutate Metadata(eq headers)?

Because it does not scale and assumes that host is somewhat more relevant than other properties that are part of the HTTP envelope.

tigrannajaryan: @jpkrohling I think this shows that we need to define the `Metadata` field semantics better. It is currently not clear whether it is incoming connection transport-specific (which means readers of this field become transport-specific too) or it is transport-agnostic (which requires well-defined, standardized set of keys that all transports should map their metadata too). In the later case "Host" would be one of the standard keys that all transports should use to store the value of the host (whatever the meaning of that host is for that particular transport).

I am not sure what is the right way. Ideally readers of the auth Info should not care how (via what transport) the request was received, so transport-agnosticity is preferable for creating e.g. processors that use auth info. However, I am not sure standardization of keys is really possible across transports.

orloffv: sure, on http2 we already can access to Host header with another key - ":authority"

jpkrohling (approved):

config/confighttp/clientinfohandler.go
jpkrohling: Indeed, we need better documentation for this feature as a whole.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
   - create `configcompression` package to manage compression methods in `confighttp` and `configgrpc`
 - Add support for cgroupv2 memory limit (#4654)
 - Enable end users to provide multiple files for config location (#4727)
+- Add Host property for Metadata (#4736)
```
jpkrohling, line 38:
```suggestion
- `client.Info`: Add Host property for Metadata (#4736)
```

jpkrohling (approved):

config/confighttp/clientinfohandler.go
```diff
@@ -50,7 +50,12 @@ func contextWithClient(req *http.Request, includeMetadata bool) context.Context
 	}
 
 	if includeMetadata {
-		cl.Metadata = client.NewMetadata(req.Header.Clone())
+		md := req.Header.Clone()
+		if len(md.Get("Host")) == 0 && req.Host != "" {
+			md.Add("Host", req.Host)
```
bogdandrutu, line 55: Should "Host" be a const in the "client" package? Seems to be an important convention that we have.

config/configgrpc/configgrpc.go
```diff
@@ -400,7 +400,11 @@ func contextWithClient(ctx context.Context, includeMetadata bool) context.Contex
 	}
 	if includeMetadata {
 		if md, ok := metadata.FromIncomingContext(ctx); ok {
-			cl.Metadata = client.NewMetadata(md.Copy())
+			copiedMD := md.Copy()
+			if len(md["Host"]) == 0 && len(md[":authority"]) > 0 {
+				copiedMD["Host"] = md[":authority"]
+			}
```
bogdandrutu, line 406: should we remove the ":authority" in this case? (Not sure, but want to make sure somebody thinks about this).

orloffv: I think no, because it's real header

config/confighttp/clientinfohandler.go
orloffv: ok

config/configgrpc/configgrpc.go
jpkrohling: Agree, at least as part of this PR here. We might want to start a discussion for something like "semantic conventions" for the headers.

config/configgrpc/configgrpc.go
```diff
@@ -400,7 +400,11 @@ func contextWithClient(ctx context.Context, includeMetadata bool) context.Contex
 	}
 	if includeMetadata {
 		if md, ok := metadata.FromIncomingContext(ctx); ok {
-			cl.Metadata = client.NewMetadata(md.Copy())
+			copiedMD := md.Copy()
+			if len(md[client.MetadataHostName]) == 0 && len(md[":authority"]) > 0 {
```
jpkrohling, line 404: I see now what @bogdandrutu meant earlier. I agree with the approach taken here, even though we end up with the host information duplicated in the client.Info. I don't think this is a problem.

jpkrohling (approved):

config/configgrpc/configgrpc.go
jpkrohling: I was agreeing in keeping `:authority` here, as it's the key used by gRPC, and would expect to see it there. The second part of my comment was about having a single value where downstream components would expect the "host" value to exist, but this is being addressed in this PR already. Sorry about the noise.

bogdandrutu: > "semantic conventions" for the headers.

This also may include us having the constant values for things like "Host", that's why I got confused. So you are ok with having that constant?

jpkrohling: Yes, at least for the moment. If we keep finding ourselves doing it often, we might want to split that out.

orloffv:
ready to merge or this PR has some problems?

jpkrohling:
@bogdandrutu would you please do the honours?

Issue #1591, opened 2020-08-19
Reload TLS certificates


pavolloffay:
**Is your feature request related to a problem? Please describe.**

I would like to renew certificates without shutting down OTEL service.

**Describe the solution you'd like**

Watch for filesystem changes and load certificates if they change.

**Describe alternatives you've considered**

Restart the service if certs have to be changed.

**Additional context**

A similar feature is being done in https://github.com/jaegertracing/jaeger/pull/2389

tigrannajaryan:
This perhaps can be done as part of config hot-reload feature.

pavolloffay:
Do you have any pointer to issue/PR? 

Note that the certs names might not change only the content of the files.

fujin:
Would very much love this feature! I am using cert-manager to provision certificates (via Kubernetes Secret) to my OpenTelemetry Collectors for the oltp receiver. We fortunately have an HPA on the deployment so pods are replaced frequently such that a certificate renewal is likely to be picked up.

bogdandrutu:
This is done by #4737

PR #4737, opened 2022-01-24
feat(configtls): Add support for reloading TLS certificates

wdullaer:
**Description:**
This PR adds support for reloading TLS certificates.
The go TLS config already has good support for reloading certificates: rather than storing them on the config, a function that produces a certificate on demand can be provided.

Rather than rely on file system notifications, which would require a dedicated go routine and quite some refactoring to handle gracefully, I have chosen to save the currently loaded certificate in a closure. Every `reload-interval` the `GetCertificate` function will attempt to reload the certificate from disk.
If the reload fails, the old certificate will be returned.

This approach has a number of benefits:
* The certificate is eagerly loaded, so config errors will still be caught at the same time
* When `reload-interval` is 0 the approach reduces to the current behaviour, functionally and performance-wise
* No additional go routines are spawned
* No explicit polling: certificates are only reloaded when they are requested

The main downside is the additional configuration surface, but I think it's worth the tradeoff.

Since even fast moving certificates live for a few hours to a few days, the overhead of this approach should be minimal. Most people will be able to get away with `reload-interval` set to `24h`.

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector/issues/1591

**Testing:**
Tests were added which simulate and validate the reloading behaviour.
We are also using a similar approach in production for both HTTP and gRPC services in our own go codebases.

**Documentation:**
I added the `ReloadInterval` config parameter with godoc. If this needs additional documentation, please let me know.


config/configtls/configtls_test.go
```diff
@@ -201,3 +206,166 @@ func TestLoadTLSServerConfig(t *testing.T) {
...
+		assert.NoError(t, fdc2.Close())
+
+		fdk2, err := os.Open(path.Join("testdata", "test-key2.pem"))
+		assert.NoError(t, err)
+		_, err = io.Copy(keyFile, fdk2)
+		assert.NoError(t, err)
+		assert.NoError(t, fdk2.Close())
+
+		// Wait ReloadInterval to ensure a reload will happen
+		time.Sleep(options.ReloadInterval)
```
jpkrohling, line 288: This might make this test unreliable, as it might take longer to read/reload than the actual reload interval. How about allowing consumers to be called back whenever the cert is reloaded? You can then register a callback here in the test, and use it with a wait group instead of relying on unreliable timings.

config/configtls/configtls_test.go
```diff
@@ -201,3 +206,166 @@ func TestLoadTLSServerConfig(t *testing.T) {
...
+		assert.NoError(t, err)
+		_, err = io.Copy(certFile, fdc2)
+		assert.NoError(t, err)
+		assert.NoError(t, fdc2.Close())
+
+		fdk2, err := os.Open(path.Join("testdata", "test-key2.pem"))
+		assert.NoError(t, err)
+		_, err = io.Copy(keyFile, fdk2)
+		assert.NoError(t, err)
+		assert.NoError(t, fdk2.Close())
```
jpkrohling, line 357: A good chunk of this and the previous tests are the same, perhaps split it into its own func? Like, "refreshCerts" ?

config/configtls/testdata/test-cert2.pem
```diff
@@ -0,0 +1,14 @@
+-----BEGIN CERTIFICATE-----
```
jpkrohling, line 1: It would be good to have documented somewhere how you generated this so that we can generate it again at a later point using the same features you used.

config/configtls/configtls.go
```diff
@@ -104,14 +109,26 @@ func (c TLSSetting) loadTLSConfig() (*tls.Config, error) {
...
 		if err != nil {
 			return nil, fmt.Errorf("failed to load TLS cert and key: %w", err)
 		}
-		certificates = append(certificates, tlsCert)
+		getCertificate = func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			now := time.Now()
+			if c.ReloadInterval != 0 && lastReload.Add(c.ReloadInterval).Before(now) {
+				var newCert tls.Certificate
+				newCert, err = tls.LoadX509KeyPair(c.CertFile, c.KeyFile)
+				if err == nil {
```
Aneurysm9, line 125: If there was an error loading the new cert it should be logged.

config/configtls/configtls.go
```diff
@@ -104,14 +109,26 @@ func (c TLSSetting) loadTLSConfig() (*tls.Config, error) {
...
+					cert = newCert
+					lastReload = now
```
Aneurysm9, line 127: Should there be some form of locking on this function to prevent a flood of attempts to reload the cert when it expires?

config/configtls/testdata/test-cert2.pem
Aneurysm9: Rather than committing certs that inevitably come with all manner of edge cases and ways they will break and the need to occasionally regenerate them, I've found it easier to simply generate test certs on the fly, into a tempdir if they need to live on a filesystem somewhere.

bogdandrutu: Most likely https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile#L326

config/configtls/configtls_test.go
```diff
@@ -201,3 +206,166 @@ func TestLoadTLSServerConfig(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotNil(t, tlsCfg)
 }
+
+func TestCertificateReload(t *testing.T) {
+	t.Run("Should eagerly load the given certificate", func(t *testing.T) {
```
bogdandrutu, line 211: Consider to have 3 tests instead of a test with 3 `t.Run`

bogdandrutu (commented):
This PR needs the CLA signed before more detailed review.

config/configtls/configtls_test.go
wdullaer: If you look at the code this tests, you'll see that this is an entirely synchronous function.
`GetCertificates` reloads the certificates lazily, the operation is only performed if a new certificate is requested. The `ReloadInterval` config parameter is there to ensure we only reload once in a while rather than every time a cert is needed, as this will impact performance.

By sleeping for `ReloadInterval` here we ensure the time since the last reload is strictly larger than `ReloadInterval` (since we do some other work before) and the cert will always be reloaded.
This test will never fail, the sleep is not there to pave over a potential async race condition.

config/configtls/configtls_test.go
wdullaer: Sure, I can refactor this to look more like a table test.

config/configtls/testdata/test-cert2.pem
wdullaer: I'll update that build script so it produces both the certs.

config/configtls/configtls.go
wdullaer: I would agree, but I can use some guidance on how you'd prefer to pass in a logger here: we need access to the logger when config is evaluated so it can be put into the closure.

An alternative is to just return the error out of `GetCertificates`: this will cause the TCP connection to fail, but will log the error we return here.
The reason I didn't is because the most likely reason for failure is that the new certs are in the process of being written, which is transient and will typically succeed afterwards.

config/configtls/configtls.go
wdullaer: We don't really consider cert expiry in the code.
Concurrent reloading is possible in a highly trafficked server: all invocations would load the same cert, and have a slightly different `now`.
So the end result of the race would be the same value for `cert` and any of the values of `now`. This is probably still fine functionally.

Maybe a bit of resources can be saved with careful usage of an RWLock, so we only ever reload once.

config/configtls/configtls_test.go
wdullaer: I'll refactor this into something that looks more like a table test.

config/configtls/configtls_test.go
jpkrohling: So, your test is waiting 100ms for nothing? That's not nice to other tests ;-) Wouldn't the same work if you set it to, say, 100 Microseconds? Or, anything higher than 0, for that matter...

wdullaer: Any value will do, it can definitely be shorter.

wdullaer:
Thank you very much for the very quick turnaround on this PR. It's very refreshing to get so much good feedback so quickly.

Regarding the CLA: we are having some issues with EasyCLA and haven't been able to approve this yet. We're still trying to figure out what's wrong, which is unfortunate.

I haven't been able to work on this today, but I will address the comments this week.

wdullaer:
* I've updated the Makefile and gen-certs script to also produce certs for the configtls tests.
* The new reload tests have been turned into a table test
* The wait times have been reduces to 100 Microseconds

jpkrohling (approved):

config/configtls/configtls.go
Aneurysm9: If a returned error gets logged that may be preferable.  Failed connections can and should be retried and, if the failure is indeed transient, then it will soon succeed.  But if the failure isn't transient things will eventually start failing with expired or otherwise invalid certs, all the while the user is expecting that the new cert is being used when it isn't.

Makefile
```diff
@@ -319,13 +319,18 @@ check-contrib:
 	make -C $(CONTRIB_PATH) for-all CMD="go mod edit -dropreplace go.opentelemetry.io/collector"
 
 # List of directories where certificates are stored for unit tests.
-CERT_DIRS := config/configgrpc/testdata \
-             config/confighttp/testdata
+CERT_DIRS := localhost|""|config/configgrpc/testdata \
+             localhost|""|config/confighttp/testdata \
+			 example1|"-1"|config/configtls/testdata \
+			 example2|"-2"|config/configtls/testdata
```
bogdandrutu, line 325: Some strange indentation?

wdullaer:
* I've fixed the indentation in the Makefile (tabs and spaces will fight forever)
* `GetCertfificate` now returns an error if reload fails (along with matching test)

Do you still want me to look at locking around the reload?

jpkrohling:
This also needs a changelog entry.

wdullaer:
* Added Changelog
* Added `GetClientCertificate` implementation: all other unit tests should pass now

bogdandrutu (changes requested):
Blocking this (please help me understand if I am wrong), since this change does not document/implement any thread-safety documentation/guards.

> 	// GetClientCertificate may be called multiple times for the same
	// connection if renegotiation occurs or if TLS 1.3 is in use.

Clearly there needs to be a protection here.

config/configtls/configtls.go
```diff
@@ -104,14 +109,30 @@ func (c TLSSetting) loadTLSConfig() (*tls.Config, error) {
...
+				if err != nil {
+					return nil, err
+				}
+				cert = newCert
+				lastReload = now
+			}
+			return &cert, nil
+		}
+		getCertificate = func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) { return getCert() }
+		getClientCertificate = func(cri *tls.CertificateRequestInfo) (*tls.Certificate, error) { return getCert() }
```
bogdandrutu, line 135: For better "isolation/encapsulation" consider to create a new type/struct that offers "getCert" func. To better understand what properties are passed to the func/what changes, etc.

wdullaer: Having a functional programming background, I like closures, but it's easy enough to refactor this way.

bogdandrutu: I may understand that, but that's how we cannot easily see that "lastReload" is changed within a multi thread code without being protected, and there may be other variables as well. So I really believe that my suggestion is the right call here, especially when dealing with multi threaded code.

wdullaer:
I don't think that comment is about concurrent use.
I think the authors here are trying to point out a difference with server certificates, where `GetCertificate` will only be called once per connection.
In TLS 1.2 and lower, the server can ask for renegotiation on a connection (and often does for on sites with both public and mTLS protected content), which would lead to `GetClientCertificate` being called more than once during the lifetime of a connection.
I don't know enough about TLS 1.3 (which no longer supports renegotiation) in which circumstances this would happen there.
I'm assuming this may be relevant for some enterprise authentication schemes where you have `GetClientCertificate` truly dynamically produce a cert that takes into account the specific connection, which is not what we're doing here.

As I believe I mentioned earlier: the lack of locking around the reload might cause some wasted cycles in servers that have a lot of concurrent connections being established. When the 'timer expires', we might reload the certificate concurrently, spending some compute we don't really need to, but this 'race' will not lead to any inconsistent data: every concurrent call is loading the same data and the resulting `lastReload` timestamp will be off by a few ms, which is irrelevant for the use case.

It does seem to be a recurring concern though, so I'll implement and R/W lock around the reload.

bogdandrutu:
> As I believe I mentioned earlier: the lack of locking around the reload might cause some wasted cycles in servers that have a lot of concurrent connections being established. When the 'timer expires', we might reload the certificate concurrently, spending some compute we don't really need to, but this 'race' will not lead to any inconsistent data: every concurrent call is loading the same data and the resulting lastReload timestamp will be off by a few ms, which is irrelevant for the use case.

I would quote this:  _if your program has a race condition, the behaviour is undefined_ read more details https://groups.google.com/g/golang-nuts/c/EHHMCdcenc8. Unfortunately cannot find that intel blog post (link does not seem to work anymore) which I read long time ago.

config/configtls/configtls.go
```diff
@@ -85,6 +91,59 @@ type TLSServerSetting struct {
...
+		defer r.lock.Unlock()
+		cert, err := tls.LoadX509KeyPair(r.CertFile, r.KeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load TLS cert and key: %w", err)
+		}
+		r.cert = &cert
+		r.lastReload = now
+		return r.cert, nil
+	}
+	defer r.lock.RUnlock()
```
bogdandrutu, line 143: If you really care about the extra cycle, you can:
1. Precalculate the "next" reload instead of always calculating `r.lastReload.Add(r.ReloadInterval)`
2. Can use an atomic variable for the "nextReloadTime" instead of a rlock, that will complicate a bit the logic, but can be done.

bogdandrutu: I would for the moment drop the 2. point, but first may worth doing.

bogdandrutu (approved):

config/configtls/configtls.go
wdullaer: precalculating sounds good, I'll add that in.

I'm not super sold on the atomic variable: I don't think this code is that performance critical to justify it.

wdullaer:
I just wanted to say thank you to all the reviewers for the quick and super valuable feedback.
This has probably been the smoothest interaction I've had so far with a large open source project.

I know how frustrating maintaining a project can be, so I wanted to take the time to thank everyone who was involved here!

PR #4739, opened 2022-01-24
fix call to NewDefaultConfigProvider

codeboten:
Still unclear how this didn't fail CI before 🤔 

The call in the test was still passing in a `string` instead of `[]string`

### DIFF
--- a/service/config_provider_test.go
+++ b/service/config_provider_test.go
@@ -354,7 +354,7 @@ func TestBackwardsCompatibilityForFilePath(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		provider := NewDefaultConfigProvider(tt.location, []string{})
+		provider := NewDefaultConfigProvider([]string{tt.location}, []string{})
 		_, err := provider.Get(context.Background(), factories)
 		assert.Contains(t, err.Error(), tt.errMessage, tt.name)
 	}


service/config_provider_test.go
```diff
@@ -354,7 +354,7 @@ func TestBackwardsCompatibilityForFilePath(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		provider := NewDefaultConfigProvider(tt.location, []string{})
+		provider := NewDefaultConfigProvider([]string{tt.location}, []string{})
```
Aneurysm9, line 357: Should the test cases be updated to contain `[]string` and have at least one that has multiple entries?

codeboten: The test cases in `TestConfigProvider_Errors` already have coverage for multiple entries: https://github.com/open-telemetry/opentelemetry-collector/blob/c3be6f7064d0276b2c6646a60a32b7c5fffb6dd7/service/config_provider_test.go#L111

The `NewDefaultConfigProvider` just passes parameters along to `NewConfigProvider`.

Aneurysm9: I'm not sure those are effective tests for `NewDefaultConfigProvider` or that they really test what should be happening when any config provider receives multiple locations, but that's not really a problem for this PR to solve.  

I think it would be marginally clearer for the test cases here to be `[]string`, even if they only had a single entry, but don't feel it is worth holding up a breakfix change for that.

Aneurysm9 (approved):

dmitryax (approved):

service/config_provider_test.go
bogdandrutu: > any config provider

The ConfigProvider is an interface, and this is one of our supported implementations, every implementation can do whatever they want... even not accepting MapProviders, or locations, or expand env var, etc....

bogdandrutu (approved):

PR #4741, opened 2022-01-24
service.NewConfigProvider

Brandnamecarr:
Description: Added a check to the length of the config locations, returns a new error if length is 0.

Link to issue: [https://github.com/open-telemetry/opentelemetry-collector/issues/4730](url)

First PR, go easy on me! Thanks.

### DIFF
--- a/service/config_provider.go
+++ b/service/config_provider.go
@@ -16,6 +16,7 @@ package service // import "go.opentelemetry.io/collector/service"
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


Aneurysm9:
> First PR, go easy on me! Thanks.

Thanks for the contribution!  Unfortunately, it is impossible to review in its current state as it appears as if the entire file has been deleted and rewritten, so we can't easily tell what actually changed.  My initial guess as to what may cause this is a line ending mismatch.  Can you ensure that your editor adapts to the existing line ending type rather than rewriting with platform-preferred line endings?  The line endings should be `\n` and not `\r\n`.

bogdandrutu (changes requested):
Something strange happens in this PR, I think you replace some characters that are not visible... Are you on windows and replaced all "new lines"?

bogdandrutu:
Closing this PR since the first PR was already merged to do this.

Issue #4231, opened 2021-10-20
Design and modify internal config to API to allow plugable full-config sources


tigrannajaryan:
The ConfigSource experimental API currently allows to inject values of the configuration settings from remote sources. The [remote configuration capabilities](https://github.com/open-telemetry/opentelemetry-collector/issues/4190) however require us to be able to inject entire configurations, not just individual values.

We may need to modify the internal config API / ConfigSource / ParserProvider and other necessary parts of our API to support these use cases.

This task requires first designing the API, then refactoring the Collector codebase to use these APIs.

bogdandrutu:
I think this needs to become "unification" of MapProvider and ConfigSource, we should not have 2

tigrannajaryan:
I would like to keep this open to make sure we are tracking it correctly. It may be that #3924 is all we need, which is fine, we will reference it from here, but maybe there is other work needed, so let's keep this open for now.

tigrannajaryan:
Note that the [Supervisor model](https://docs.google.com/document/d/1KtH5atZQUs9Achbce6LiOaJxLbksNJenvgvyKLsJrkc/edit#) may render some of the tasks listed as unnecessary, e.g. reverting of bad configs can become Supervisor responsibility. The exact functionality of the Supervisor is still to be defined.

PR #4468, opened 2021-11-23
Config sources to support full remote config and value substitution

tigrannajaryan:
Resolves: https://github.com/open-telemetry/opentelemetry-collector/issues/4190

This is a draft PR (not production quality code) that shows roughly what the end
goal is. Don't merge this PR, it is merely to show what the final destination
looks like. If we agree that this is a good end goal I will split this into
series of smaller PRs and will submit them.

I am primarily looking for the feedback on the new concept and the APIs. There
is no need to review all the implementation details, they are not production
quality yet.

## Configuration File Format Changes

Introduced 2 new optional sections in the configuration file format:
config_sources and merge_configs. config_sources defines one or more
configuration sources. Configuration sources are a new type of component that is
pluggable like all other component types by implementing its Factory.
Configuration sources must implement a configmapprovider.Provider interface.

The merge_configs section contains a sequence of config source references. The
configs from these sources is retrieved and merged in the specified order to
form the top-level configuration of the Collector. The locally specified
top-level configuration is merged last (thus has the highest precedence).

This approach allows to maintain backward compatibility. Any previously valid
configuration file remains valid. New configuration files may specify only
config_sources/merge_configs section, move the local configuration to a separate
file and use that file as one of the configuration sources.

This approach also does not require us to introduce a new --config_sources
command line option.

## Configuration Provider Interfaces

Introduced the concept of config MapProvider and config ValueProvider, both
implementing a base Provider interface.

MapProvider is equivalent to the old Provider interface and allows retrieving
configuration maps.

ValueProvider can return configuration values, which includes primitive values,
arrays, maps. ValueProvider also supports returning different values based on
the selector and params passed which allows to parameterize the returned value
at the place of usage.

## Command Line Invocation

ValueProvicer-type config sources can be invoked directly from the command line.
The --config option now accepts either a file path (which was the existing
functionality) or values in the form of `<sourcename>:<selector>?<params>`
(which is a new functionality). The new functionality is triggered by the
presence of a colon in the --config option value. This makes it impossible to
pass file names that use a colon character, but we assume that this is an
acceptable limitation since colon is often used as a file path delimiter
character in Unix systems anyway.

The new functionality allows for example for the following command lines:
`./otelcol --config=config.yaml`
`./otelcol --config=files:config.yaml`
`./otelcol --config=http://example.com/path/to/config.yaml`
`./otelcol --config=https://example.com/path/to/config.yaml?foo=bar&bar=baz`

## Example Config Sources

[...]

Issue #4001, opened 2021-09-09
Config unmarshalling doesn't work properly for slice values 


dmitryax:
**Describe the bug**
If we have a default config for a component where there is a slice value with 2 and more item, there is no way to override that value with a smaller slice.

The problem is that [mapstructure.decodeMap func](https://github.com/mitchellh/mapstructure/blob/4664f9ee512b4dfa71b501c4960d469c205b2cf1/mapstructure.go#L1046) that we use for unmarshalling compares elements of the slice one by one and does not override the length of the result. It leads to the problem that we cannot override a default slice value with a slice of smaller size. 

This is an unexpected behaviour for users. We should be able to fully override default slice values.

**Steps to reproduce**
Real example is https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/dotnetdiagnosticsreceiver configuration:

Try to override the `counters` value with a 1-item slice like this:
```
receivers:
  dotnet_diagnostics:
    counters: ["MyCounters"]
``` 

**What did you expect to see?**
A slice config value should be fully overridden. In the example above, we should get `counters: ["MyCounters"]`

**What did you see instead?**
Only subset of the default slice items are overridden, the rest of the default slice stays the same. In the example above, we get `counters: ["MyCounters", "Microsoft.AspNetCore.Hosting"]`

PR #4742, opened 2022-01-25
Add support in the confmap.Resolver to expand embedded config URIs inside configuration

bogdandrutu:
Proposal on how we can extend our current config provider to support expanding (including) config "parts". This will fix https://github.com/open-telemetry/opentelemetry-collector/issues/4231 and will fully implement same capabilities as proposed in https://github.com/open-telemetry/opentelemetry-collector/pull/4468.

This design is based on the https://docs.google.com/document/d/1XlDpS8-a4Ds6uoe6C6WgO8mCpDaQNufcHSEmBpiCdI0/edit format which we already support for the "--config" flag.

## `confmap.Provider` changes

Extend the capability of a `confmap.Provider` to return a `Provided` value that can encapsulate a `confmap.Conf` or any partial config value (float64/int64/bool/string/map[string]interface{}/[]interface{} where map values or slice values have the same restrictions). The current design allows `recursive` expansions, so values returned by expanding one embedded config URI will be searched for expanding again.

## Known caveats

1. The proposed mechanism does not allow expanding values within other values, for example this case `${http://example.com/get_config?os=${env:OSNAME}}` will NOT expand first the OSNAME env var.

## Examples of usages

**Retrieving token from zookeeper (scheme not yet defined)**

```yaml
exporters:
  signalfx:
    # this will read the access_token value from "zk" config provider.
    access_token: ${zk:zk1,zk2,zk2/path/to/data}
```

**Retrieving an entire exporter config from a file**

```yaml
exporters:
  # this will read the entire config for the otlp/file exporter from the given file
  otlp/file_config: ${file:path/to/config/file/for/the/exporter.yaml}
```

**Expanding env vars**

```yaml
exporters:
  otlp:
    endpoint: ${env:MY_ENV}

  otlp/backwards_compatible:
    endpoint: ${MY_ENV}
```

## Expand `Converter` vs `embedded` config

The current proposal to support `embedding` config URIs, does not support the following cases that are supported by the expand envvar `Converter`:

1. Using `$` without brackets:

```yaml
exporters:
  otlp/another_way_to_use_env_vars:
    endpoint: $MY_ENV
```

2. Expanding values within a string value and concatenate:

```yaml
exporters:
  otlp/localhost_enpoint:
    endpoint: "localhost:${OTLP_PORT_ENV}"
```

Because of these missing features the Expand `Converter` will be maintain for a period of time until a final decision between supporting or not the 2 missing features is made. If we decide to not support any of these features we can start at least by logging errors and disable the functionality behind a featuregate until we know if this is needed or not.

### DIFF
--- a/confmap/provider.go
+++ b/confmap/provider.go
@@ -16,6 +16,7 @@ package confmap // import "go.opentelemetry.io/collector/confmap"
 
 import (
 	"context"
+	"fmt"
 )
 
 // Provider is an interface that helps to retrieve a config map and watch for any
@@ -81,7 +82,7 @@ type ChangeEvent struct {
 
 // Retrieved holds the result of a call to the Retrieve method of a Provider object.
 type Retrieved struct {
-	conf      *Conf
+	rawConf   interface{}
 	closeFunc CloseFunc
 }
 
@@ -101,17 +102,39 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
 }
 
 // NewRetrieved returns a new Retrieved instance that contains the data from the raw deserialized config.
-func NewRetrieved(rawConf map[string]interface{}, opts ...RetrievedOption) (Retrieved, error) {
+// The rawConf can be one of the following types:
+//   - Primitives: int, int32, int64, float32, float64, bool, string;
+//   - []interface{};
+//   - map[string]interface{};
+func NewRetrieved(rawConf interface{}, opts ...RetrievedOption) (Retrieved, error) {
+	if err := checkRawConfType(rawConf); err != nil {
+		return Retrieved{}, err
+	}
 	set := retrievedSettings{}
 	for _, opt := range opts {
 		opt(&set)
 	}
-	return Retrieved{conf: NewFromStringMap(rawConf), closeFunc: set.closeFunc}, nil
+	return Retrieved{rawConf: rawConf, closeFunc: set.closeFunc}, nil
 }
 
 // AsConf returns the retrieved configuration parsed as a Conf.
-func (r Retrieved) AsConf() (*Conf, error) {
-	return r.conf, nil
+func (r *Retrieved) AsConf() (*Conf, error) {
+	if r.rawConf == nil {
+		return New(), nil
+	}
+	val, ok := r.rawConf.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("retrieved value (type=%T) cannot be used as a Conf", r.rawConf)
+	}
+	return NewFromStringMap(val), nil
+}
+
+// AsRaw returns the retrieved configuration parsed as an interface{} which can be one of the following types:
+//   - Primitives: int, int32, int64, float32, float64, bool, string;
+//   - []interface{} - every member follows the same rules as the given interface{};
+//   - map[string]interface{} - every value follows the same rules as the given interface{};
+func (r *Retrieved) AsRaw() (interface{}, error) {
+	return r.rawConf, nil
 }
 
 // Close and release any watchers that Provider.Retrieve may have created.
@@ -129,3 +152,15 @@ func (r Retrieved) Close(ctx context.Context) error {
 
 // CloseFunc a function equivalent to Retrieved.Close.
 type CloseFunc func(context.Context) error
+
+func checkRawConfType(rawConf interface{}) error {
+	if rawConf == nil {
+		return nil
+	}
+	switch rawConf.(type) {
+	case int, int32, int64, float32, float64, bool, string, []interface{}, map[string]interface{}:
+		return nil
+	default:
+		return fmt.Errorf("unsupported type=%T for retrieved config", rawConf)
+	}
+}
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -38,6 +38,8 @@ type Resolver struct {
 	sync.Mutex
 	closers []CloseFunc
 	watcher chan error
+
+	enableExpand bool
 }
 
 // ResolverSettings are the settings to configure the behavior of the Resolver.
@@ -115,20 +117,14 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 		// For backwards compatibility:
 		// - empty url scheme means "file".
 		// - "^[A-z]:" also means "file"
-		scheme := "file"
-		if idx := strings.Index(uri, ":"); idx != -1 && !driverLetterRegexp.MatchString(uri) {
-			scheme = uri[:idx]
-		} else {
-			uri = scheme + ":" + uri
+		if driverLetterRegexp.MatchString(uri) {
+			uri = "file:" + uri
 		}
-		p, ok := mr.providers[scheme]
-		if !ok {
-			return nil, fmt.Errorf("scheme %q is not supported for uri %q", scheme, uri)
-		}
-		ret, err := p.Retrieve(ctx, uri, mr.onChange)
+		ret, err := mr.retrieveValue(ctx, location{uri: uri, defaultScheme: "file"})
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("cannot retrieve the configuration: %w", err)
 		}
+		mr.closers = append(mr.closers, ret.Close)
 		retCfgMap, err := ret.AsConf()
 		if err != nil {
 			return nil, err
@@ -136,7 +132,18 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 		if err = retMap.Merge(retCfgMap); err != nil {
 			return nil, err
 		}
-		mr.closers = append(mr.closers, ret.Close)
+	}
+
+	if mr.enableExpand {
+		cfgMap := make(map[string]interface{})
+		for _, k := range retMap.AllKeys() {
+			val, err := mr.expandValueRecursively(ctx, retMap.Get(k))
+			if err != nil {
+				return nil, err
+			}
+			cfgMap[k] = val
+		}
+		retMap = NewFromStringMap(cfgMap)
 	}
 
 	// Apply the converters in the given order.
@@ -186,4 +193,83 @@ func (mr *Resolver) closeIfNeeded(ctx context.Context) error {
 		err = multierr.Append(err, ret(ctx))
 	}
 	return err
+}
+
+func (mr *Resolver) expandValueRecursively(ctx context.Context, value interface{}) (interface{}, error) {
+	for i := 0; i < 100; i++ {
+		val, changed, err := mr.expandValue(ctx, value)
+		if err != nil {
+			return nil, err
+		}
+		if !changed {
+			return val, nil
+		}
+		value = val
+	}
+	return nil, errors.New("too many recursive expansions")
+}
+
+func (mr *Resolver) expandValue(ctx context.Context, value interface{}) (interface{}, bool, error) {
+	switch v := value.(type) {
+	case string:
+		// If it doesn't have the format "${scheme:opaque}" no need to expand.
+		if !strings.HasPrefix(v, "${") || !strings.HasSuffix(v, "}") {
+			return value, false, nil
+		}
+		uri := v[2 : len(v)-1]
+		// For backwards compatibility:
+		// - empty scheme means "env".
+		ret, err := mr.retrieveValue(ctx, location{uri: uri, defaultScheme: "env"})
+		if err != nil {
+			return nil, false, err
+		}
+		mr.closers = append(mr.closers, ret.Close)
+		val, err := ret.AsRaw()
+		return val, true, err
+	case []interface{}:
+		nslice := make([]interface{}, 0, len(v))
+		nchanged := false
+		for _, vint := range v {
+			val, changed, err := mr.expandValue(ctx, vint)
+			if err != nil {
+				return nil, false, err
+			}
+			nslice = append(nslice, val)
+			nchanged = nchanged || changed
+		}
+		return nslice, nchanged, nil
+	case map[string]interface{}:
+		nmap := map[string]interface{}{}
+		nchanged := false
+		for mk, mv := range v {
+			val, changed, err := mr.expandValue(ctx, mv)
+			if err != nil {
+				return nil, false, err
+			}
+			nmap[mk] = val
+			nchanged = nchanged || changed
+		}
+		return nmap, nchanged, nil
+	}
+	return value, false, nil
+}
+
+type location struct {
+	uri           string
+	defaultScheme string
+}
+
+func (mr *Resolver) retrieveValue(ctx context.Context, l location) (Retrieved, error) {
+	uri := l.uri
+	scheme := l.defaultScheme
+	if idx := strings.Index(uri, ":"); idx != -1 {
+		scheme = uri[:idx]
+	} else {
+		uri = scheme + ":" + uri
+	}
+	p, ok := mr.providers[scheme]
+	if !ok {
+		return Retrieved{}, fmt.Errorf("scheme %q is not supported for uri %q", scheme, uri)
+	}
+	return p.Retrieve(ctx, uri, mr.onChange)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -192,6 +192,7 @@ There isn't a valid core binary for this release. Use v0.57.2 instead.
 - Use OpenCensus `metric` package for process metrics instead of `stats` package (#5486)
 - Update OTLP to v0.18.0 (#5530)
 - Log histogram min/max fields with `logging` exporter (#5520)
+- Add support in the `confmap.Resolver` to expand embedded config URIs inside configuration (#4742)
 
 ### 🧰 Bug fixes 🧰
 
--- a/confmap/README.md
+++ b/confmap/README.md
@@ -33,23 +33,35 @@ The `Resolver` receives as input a set of `Providers`, a list of `Converters`, a
 `configURI` that will be used to generate the resulting, or effective, configuration in the form of a `Conf`,
 that can be used by code that is oblivious to the usage of `Providers` and `Converters`.
 
+`Providers` are used to provide an entire configuration when the `configURI` is given directly to the `Resolver`,
+or an individual value (partial configuration) when the `configURI` is embedded into the `Conf` as a values using
+the syntax `${configURI}`.
+
 ```terminal
               Resolver                   Provider
-                 │                          │
    Resolve       │                          │
 ────────────────►│                          │
                  │                          │
               ┌─ │        Retrieve          │
               │  ├─────────────────────────►│
-              │  │                          │
+              │  │          Conf            │
               │  │◄─────────────────────────┤
-   foreach    │  │                          │
+  foreach     │  │                          │
   configURI   │  ├───┐                      │
               │  │   │Merge                 │
               │  │◄──┘                      │
               └─ │                          │
+              ┌─ │        Retrieve          │
+              │  ├─────────────────────────►│
+              │  │    Partial Conf Value    │
+              │  │◄─────────────────────────┤
+  foreach     │  │                          │
+  embedded    │  │                          │
+  configURI   │  ├───┐                      │
+              │  │   │Replace               │
+              │  │◄──┘                      │
+              └─ │                          │
                  │            Converter     │
-                 │                │         │
               ┌─ │     Convert    │         │
               │  ├───────────────►│         │
     foreach   │  │                │         │
@@ -57,15 +69,15 @@ that can be used by code that is oblivious to the usage of `Providers` and `Conv
               └─ │                          │
                  │                          │
 ◄────────────────┤                          │
-                 │                          │
 ```
 
 The `Resolve` method proceeds in the following steps:
 
 1. Start with an empty "result" of `Conf` type.
 2. For each config URI retrieves individual configurations, and merges it into the "result".
-2. For each "Converter", call "Convert" for the "result".
-4. Return the "result", aka effective, configuration.
+3. For each embedded config URI retrieves individual value, and replaces it into the "result".
+4. For each "Converter", call "Convert" for the "result".
+5. Return the "result", aka effective, configuration.
 
 ### Watching for Updates
 After the configuration was processed, the `Resolver` can be used as a single point to watch for updates in the
--- a/confmap/provider/internal/provider.go
+++ b/confmap/provider/internal/provider.go
@@ -24,7 +24,7 @@ import (
 // * yamlBytes the yaml bytes that will be deserialized.
 // * opts specifies options associated with this Retrieved value, such as CloseFunc.
 func NewRetrievedFromYAML(yamlBytes []byte, opts ...confmap.RetrievedOption) (confmap.Retrieved, error) {
-	var rawConf map[string]interface{}
+	var rawConf interface{}
 	if err := yaml.Unmarshal(yamlBytes, &rawConf); err != nil {
 		return confmap.Retrieved{}, err
 	}
--- a/confmap/provider/internal/provider_test.go
+++ b/confmap/provider/internal/provider_test.go
@@ -46,10 +46,13 @@ func TestNewRetrievedFromYAMLWithOptions(t *testing.T) {
 
 func TestNewRetrievedFromYAMLInvalidYAMLBytes(t *testing.T) {
 	_, err := NewRetrievedFromYAML([]byte("[invalid:,"))
-	require.Error(t, err)
+	assert.Error(t, err)
 }
 
 func TestNewRetrievedFromYAMLInvalidAsMap(t *testing.T) {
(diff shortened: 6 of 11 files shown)


bogdandrutu:
Bogdan's ideas:

> 1. This PR requires a change in the current Provider interface to support returning a config.Map (maybe a map[string]interface{}) or a string (maybe accept any interface{} and set that)

I think we should change the return type to be "interface" and when used in the "--config" flag the return value MUST be "map[string]interface{}" and when used in the embedding/including case value MUST be one of "string", "[]interface{}" or "map[string]interface{}" as the "primitive" values for yaml configs.

> 2. It is a bit unclear how to support watching for config updates, there are 2 options: a) extend the Converter to also accept a WatcherFunc; b) remove expand converter and embed this capability into our implementation of the ConfigProvider (allow both constructors NewConfigProvider and NewDefaultConfigProvider to allow this to be configured somehow).

Here is where the most contingent point may be. My preference is to embed this into our implementation of the ConfigProvider, but if this is not ok because we cannot find a way to expose a configuration to enable/disable this, we may want to go with the a) option.

> 3. The current "os.Expand" does not allow to return an error or to handle an error. It may require us to copy that or look for an alternative API.

I need to investigate more this.

bogdandrutu:
@codeboten @jcchavezs here is why blindly exposing the current Expand will not help. There are still things to be changed/fully decide on, but we should be able to support this in the future I think, since there are couple of issues related to this, including the current experimental configsource package. If we finalize this, we can then remove all the experimental change and we have a very powerful mechanism of "merging" configs or "including" configs parts using same format "uri" like.

jpkrohling:
```
access_token: ${env:MY_ENV?default=foo}
```

Is this open to discussion? Shouldn't we make it like other systems, like bash? For instance:

```
${env:MY_ENV:-foo}
```

jpkrohling:
On point 2, I'm not quite sure I understand the option b:

> b) remove expand converter and embed this capability into our implementation of the ConfigProvider (allow both constructors NewConfigProvider and NewDefaultConfigProvider to allow this to be configured somehow).

Did you mean a downstream implementation, on Splunk's distribution?

jpkrohling:
> The current "os.Expand" does not allow to return an error or to handle an error. It may require us to copy that or look for an alternative API.

Can we brainstorm a bit on whether we *need* this error? I also think we do, but what are the implication in case we decide to make this a "can't fail" function? What happens when Zookeeper is offline after the first parsing (ie, can't update the config)?

bogdandrutu:
@jpkrohling related to the alternative for the env: The only requirement that we have right now in the "Provider" is that it starts with "env:" (the schema), so we can do anything we want when we will change the ["env" provider](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configmapprovider/env.go) to support default values. Personally I like keeping things more or less like "uri" and I am not sure that syntax you proposed is uri compatible. But to make sure we understand each other, that is a different decision that we have to make when we extend the "env" to support default values. This PR is about supporting same syntax as in the "--config" flag to include partial configs in the config.

config/configmapprovider/expand.go
```diff
@@ -17,52 +17,79 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
...
+func NewExpandConverter(configMapProviders map[string]Provider) config.MapConverterFunc {
+	// TODO: Figure out how to support watching for changes in a Converter, or if expand should be a converter.
+	return func(ctx context.Context, cfgMap *config.Map) error {
 		for _, k := range cfgMap.AllKeys() {
-			cfgMap.Set(k, expandStringValues(cfgMap.Get(k)))
+			cfgMap.Set(k, expandStringValues(cfgMap.Get(k), func(location string) string {
+				// This allows escaping environment variable substitution via $$, e.g.
+				// - $FOO will be substituted with env var FOO
+				// - $$FOO will be replaced with $FOO
+				// - $$$FOO will be replaced with $ + substituted env var FOO
```
codeboten, line 38: Is there a case there this functionality would not be desirable? i'm wondering if any providers would prefer not having the functionality between lines 39-48.

I'm thinking of a case where someone maybe provides an alternate env provider.... i don't know if this is a strong case though.

config/configmapprovider/expand.go
```diff
@@ -17,52 +17,79 @@ package configmapprovider // import "go.opentelemetry.io/collector/config/config
...
+				} else {
+					// MapProviders require the location to always have the scheme.
+					location = scheme + ":" + location
+				}
+				p, ok := configMapProviders[scheme]
+				// TODO: Figure out how to return an error when not registered.
+				if !ok {
+					return ""
+				}
+				// TODO: Figure out how to return an error when retrieve returns an error.
```
codeboten, line 54: probably would need to register an error callback somewhere along the way here.

codeboten (commented):
I like where this functionality is going, I would love to hear from folks that are implementing downstream providers whether they have any concerns with this approach. To be clear this mean `NewExpandConverter` remains as the public API to use moving forward?

tigrannajaryan:
Two comments:

1. I like the usage syntax of `${scheme:locator}`. I think it works for simple cases, such as `${env:VARNAME}` and also for more complicated cases e.g. `${http://example.com/get_config?os=windows}`. One thing that is not clear is if the expansion is recursive, i.e. can I do this for example: `${http://example.com/get_config?os=${env:OSNAME}}`.
I agree with Bogdan that it is preferable to make things look like an URI if possible. Staying consistent helps makes the end user's life easier. I do not think we should mix bash-like syntax with URI-like syntax unless we are forced to.

2. I have a problem understanding how this all works just by looking at the source code. 

It would be great to have a design document that explains the _entire_ picture of how the configuration is supposed to work, what are the individual concepts we deal with (what is Converter, what is Provider, etc), how the concepts interact with each other, how the Service uses the concepts, which of the concepts are user-extensible or user-definable and how, which ones are internal only. Ideally have a document with some diagrams and flows, including the fetching flow (pull config), the watching flow (push/notify from sources).

Once the concepts are explained at the high-level I think it will be easier to decide what it should look in the code.

Aneurysm9:
I would second what Tigran has said about needing a design document detailing the complete system.  It's not clear to me how this is supposed to fit together.  It looks like this would expect the expand converter to have a list of providers that it would attempt to fetch replacements from, but I would expect replacements to be simple scalar values for the most part while the data fetched from providers would typically be a config map with many values.  How would an individual value be selected from a location that provided many? How would this system avoid fetching the same configuration multiple times to perform multiple replacements from the same provider?

I'm really not comfortable changing the return type of retrieved configuration to `interface{}`.  Why is that necessary and what alternatives are available?

mx-psi:
I also agree that having a design document here would be very helpful. I like the direction this goes in, but I feel like I would be able to contribute to the discussion better if I had a design doc. 

> I'm really not comfortable changing the return type of retrieved configuration to interface{}. Why is that necessary and what alternatives are available?

The crux of the problem here seems to be supporting both simple values and complex objects; e.g. you want to have both a `${source:API_TOKEN}`, which would return a `string` and `${source:RECEIVER_CONFIGURATION}`, which would return a `config.Map`. Maybe the solution is for providers to unmarshal into a config object rather than returning a value? The problem with an approach like that is that it allows providers to do too much.

bogdandrutu:
@mx-psi very well pointed the problem.

mx-psi:
To avoid the `interface{}` type, if we are willing to accept some verbosity, we can emulate a sum type in Go with an internal interface:

```go

type ConfigObject interface {
   internal.InternalInterface
}

type ScalarValue struct{
  internal.BaseInternal
  s string
}

func (s *ScalarValue) String() string {
  return s.s
}

func NewScalarConfigObject(s string) ConfigObject {
  return &ScalarValue{s}
}

type ConfigMap struct{
   internal.BaseInternal
   c *config.Map
}

func NewConfigMapConfigObject(c *config.Map) ConfigObject {
  return &ConfigMap{c}
}

func (c *ConfigMap) ConfigMap() *config.Map {
  return c.c
}

func example() {

       object, _ := something.Get(ctx)
        switch object.(type) {
        case *ScalarValue:
          // do something with object.String()
        case *ConfigMap:
          // do something with object.ConfigMap()
        }
}
```

There is even [some tooling support](https://github.com/BurntSushi/go-sumtype) to ensure exhaustive case-checking.

bogdandrutu:
@mx-psi here is an alternative, will offer a strRetrieved and a NewRetrievedFromString:

```golang
// Retrieved holds the result of a call to the Retrieve method of a Provider object.
type Retrieved interface {
	MergeTo(dest *Map, destPath string) error
	Close(ctx context.Context) error
}

// mapRetrieved holds the result of a call to the Retrieve method of a Provider object.
type mapRetrieved struct {
	cfgMap *Map
	CloseFunc
}

// NewRetrievedFromMap returns a new Retrieved instance that contains a Map data.
// * cfgMap the Map that will be merged to the given map in the MergeTo.
// * CloseFunc specifies a function to be invoked when the configuration for which it was
//   used to retrieve values is no longer in use and should close and release any watchers
//	 that it may have created.
func NewRetrievedFromMap(cfgMap *Map, close CloseFunc) Retrieved {
	return &mapRetrieved{cfgMap: cfgMap, CloseFunc: close}
}

func (r *mapRetrieved) MergeTo(dest *Map, destPath string) error {
	return dest.k.MergeAt(r.cfgMap.k, destPath)
}

// CloseFunc a function to close and release any watchers that it may have created.
//
// Should block until all resources are closed, and guarantee that `onChange` is not
// going to be called after it returns except when `ctx` is cancelled.
//
// Should never be called concurrently with itself.
type CloseFunc func(context.Context) error

func (f CloseFunc) Close(ctx context.Context) error {
	if f == nil {
		return nil
	}
	return f(ctx)
}

```

bogdandrutu:
@codeboten @mx-psi at this point, the PR is fully functional, but relies on https://github.com/open-telemetry/opentelemetry-collector/pull/5198 to be fully implemented and add support for `MergeToAt` that was discussed there.

As it is now implemented it has some limitations:
* It does not expand values in an array; If this is important to support we should not offer `MergeToAt` but instead, we should offer a `AsRaw() interface{}`. Maybe this also influences and we can have a "AsMap() *config.Map" instead of the current proposal of `MergeTo` in #5198

bogdandrutu:
@tigrannajaryan 

> I like the usage syntax of ${scheme:locator}. I think it works for simple cases, such as ${env:VARNAME} and also for more complicated cases e.g. ${http://example.com/get_config?os=windows}. One thing that is not clear is if the expansion is recursive, i.e. can I do this for example: ${http://example.com/get_config?os=${env:OSNAME}}.

I did not intend for the first version to make it work for that "recursive" example. The reason is that you can use for the moment something like `${env:LINK_WITH_OSNAME}`. We can extend later the functionality but for the moment I would reject to support more than minimum required, any extra fancy feature costs us maintenance time (and trust me a lot of time we spend on this).

> I have a problem understanding how this all works just by looking at the source code.

It would be great to have a design document that explains the entire picture of how the configuration is supposed to work, what are the individual concepts we deal with (what is Converter, what is Provider, etc), how the concepts interact with each other, how the Service uses the concepts, which of the concepts are user-extensible or user-definable and how, which ones are internal only. Ideally have a document with some diagrams and flows, including the fetching flow (pull config), the watching flow (push/notify from sources).

`Provider`: as the name states in provides configuration in any supported forms (as a map, or as a value for the embedding case), and allows the "service" that uses them to watch/monitor for changes.

`Converter`: this is mostly to support implementing backwards compatible changes (a.k.a converts a provided map to another map) OR any other conversion that doesn't "provide" new configuration since we don't have a mechanism to watch/monitor for changes. A nice usage of converters can be seen in Splunk distro (https://github.com/signalfx/splunk-otel-collector/blob/main/internal/configconverter/move_hec_tls.go).

I think the current way we use converter is wrong, but works, since we use for the "--set" flag and for expanding `envvar`. The `envvar` converter as you can see in this PR (latest snapshot) will be soon removed since we implement the "expanding" in the `service.ConfigProvider`. The "--set" flag usage can be also removed from being a converter, and transform into a "yaml:..." location, this is a functionality that we should implement (not suggesting to remove the --set flag, just how we consume it in the service.ConfigProvider).

Now, how the default (called default because so far the `service.ConfigProvider` is an interface that we provide a "default" implementation, others may do other things) `service.ConfigProvider` works, you can also read the [comment](https://pkg.go.dev/go.opentelemetry.io/collector@v0.49.0/service#NewConfigProvider):
1. Retrieves individual configurations (as a Map) from all given `locations`.
2. In the retrieved order all retrieved config maps will be merged in one map.
3. After that we will expand the "locations" (yaml values that follow the format `${scheme:opaque}`, for the moment we do not allow this to be part of a longer string, or recursive how you called it).
4. After that we call the converters in the given order
5. After that we call the unmarshaler.
6. Then we return the given result, while watching for all watchers (from point 1 and 3).

mx-psi:
> It does not expand values in an array

I think this is important to eventually support, what is the problem? Is this related to #4001? If this is potentially solvable in the future, could we return a 'not supported' error when people try to use this feature in an array so that we don't have to solve this problem now.

> I did not intend for the first version to make it work for that "recursive" example. The reason is that you can use for the moment something like ${env:LINK_WITH_OSNAME}. We can extend later the functionality but for the moment I would reject to support more than minimum required, any extra fancy feature costs us maintenance time (and trust me a lot of time we spend on this).

How would it be extended? To add `${url:${env:URL_WITH_NAME}}` in the future, the `${..}` part needs to be handled in some way now (via an error?), otherwise these characters could be valid in a URI from a custom provider and there would be ambiguity between "load this from the url stored in the environment variable `URL_WITH_NAME`" and "load this from the url `"${env:URL_WITH_NAME}"`"

tigrannajaryan:
> `Provider`: as the name states in provides configuration in any supported forms (as a map, or as a value for the embedding case), and allows the "service" that uses them to watch/monitor for changes.
> 
> `Converter`: this is mostly to support implementing backwards compatible changes (a.k.a converts a provided map to another map) OR any other conversion that doesn't "provide" new configuration since we don't have a mechanism to watch/monitor for changes. A nice usage of converters can be seen in Splunk distro (https://github.com/signalfx/splunk-otel-collector/blob/main/internal/configconverter/move_hec_tls.go).

@bogdandrutu It would be nice to add relevant parts of this explanation to the code as comments (and anything else you think can be useful for understanding the mechanics of config providers).

I have not been following this PR closely for a while now and do not have a specific feedback on the PR. Let me know if you want me to review it (and having a design described in more details somewhere would help with a review tremendously, since I am still struggling a bit with understanding how it all works).

bogdandrutu:
> I have not been following this PR closely for a while now and do not have a specific feedback on the PR. Let me know if you want me to review it (and having a design described in more details somewhere would help with a review tremendously, since I am still struggling a bit with understanding how it all works).

It is in my personal opinion overkill to ask for a "design doc" where to explain how the code works. I think that this will very fast get outdated, and code should be nicely written and documented that you don't need to read a document that explains how the code works. If we I failed to do that in this PR, it means that I need to restructure/redesign the code a bit more so that it is clear what it does.

config/configmapprovider/expand.go
bogdandrutu: No longer in the PR.

config/configmapprovider/expand.go
bogdandrutu: No longer needed.

confmap/provider.go
```diff
@@ -101,17 +102,39 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
...
+	}
+	val, ok := r.rawConf.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("retrieved value (type=%T) cannot be used as a Conf", r.rawConf)
+	}
+	return NewFromStringMap(val), nil
+}
+
+// AsRaw returns the retrieved configuration parsed as an interface{} which can be one of the following types:
+//   * Primitive float/int/bool/string;
```
tigrannajaryan, line 133: Line 106 says `Primitives: int, int32, int64, float32, float64, bool, string;`. Is the difference intentional?

confmap/resolver.go
```diff
@@ -187,3 +188,59 @@ func (mr *Resolver) closeIfNeeded(ctx context.Context) error {
...
+			nmap[mk] = val
+		}
+		return nmap, nil
+	}
+	return value, nil
+}
+
+func (mr *Resolver) retrieveValue(ctx context.Context, uri string, defaultScheme string) (Retrieved, error) {
+	// For backwards compatibility:
+	// - empty scheme means "env".
```
tigrannajaryan, line 234: Comment is incorrect.

confmap/resolver.go
```diff
@@ -115,28 +115,21 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
...
-			uri = scheme + ":" + uri
+		if driverLetterRegexp.MatchString(uri) {
+			uri = "file:" + uri
 		}
-		p, ok := mr.providers[scheme]
-		if !ok {
-			return nil, fmt.Errorf("scheme %q is not supported for uri %q", scheme, uri)
-		}
-		ret, err := p.Retrieve(ctx, uri, mr.onChange)
+		ret, err := mr.retrieveValue(ctx, uri, "file")
```
tigrannajaryan, line 121: Nit: it is a bit unclear what "file" means here. Also, two string parameters are easy to support in the wrong order. s an alternate pass a struct with 2 fields, which will also make clear what the meaning of the parameters is because fields have names.

confmap/resolver.go
```diff
@@ -187,3 +188,59 @@ func (mr *Resolver) closeIfNeeded(ctx context.Context) error {
...
+		}
+		location := v[2 : len(v)-1]
+		// For backwards compatibility:
+		// - empty scheme means "env".
+		ret, err := mr.retrieveValue(ctx, location, "env")
+		if err != nil {
+			return nil, err
+		}
+		mr.closers = append(mr.closers, ret.Close)
+		return ret.AsRaw()
```
tigrannajaryan, line 207: To confirm: we don't want to apply `expandValue` recursively to the returned result here?

tigrannajaryan (approved):

confmap/provider.go
bogdandrutu: Mistake, fixed.

confmap/resolver.go
bogdandrutu: How many times would we do this? Until no changes (can get into infinite loop, maybe try for 10 times if still resolving then return error?)?

I thought it is not useful, but I can see now point in having it (after you asked for it). Let's discuss the behavior.

confmap/resolver.go
bogdandrutu: Commented on the param. Let me know if you think the struct will be very useful and I can do it.

bogdandrutu: Done.

confmap/resolver.go
tigrannajaryan: I can imagine 2 use cases:
- If you have a local config file where env variables are used (quite common, e.g. in Splunk distro) you may want to specify that the value must be retrieved from a remote source instead of providing it directly.
- The opposite: a config fetched from a remote location, but which refers to a local env variables (e.g. refer to $PATH or $TEMP).

>How many times would we do this? Until no changes (can get into infinite loop, maybe try for 10 times if still resolving then return error?)?

Yeah, probably repeat until no more changes with some limit to avoid infinite looping. Any reasonable use case I can imagine should be 2 rounds, so limiting to something like 5 or 19 attempts should be fine.

mx-psi: I would suggest not introducing recursive expanding on this PR. We can forbid usage of `$`, `{` and `}` in the URI on this PR and deal with this problem later

confmap/provider.go
```diff
@@ -101,17 +102,39 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
 }
 
 // NewRetrieved returns a new Retrieved instance that contains the data from the raw deserialized config.
-func NewRetrieved(rawConf map[string]interface{}, opts ...RetrievedOption) (Retrieved, error) {
+// The rawConf can be one of the following types:
+//   * Primitives: int, int32, int64, float32, float64, bool, string;
+//   * []interface{};
+//   * map[string]interface{};
+func NewRetrieved(rawConf interface{}, opts ...RetrievedOption) (Retrieved, error) {
+	if err := checkRawConfType(rawConf); err != nil {
```
mx-psi, line 110: I think it would be good to have a unit test passing a `rawConf` of an unsupported type

confmap/provider.go
```diff
@@ -101,17 +102,39 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
 }
 
 // NewRetrieved returns a new Retrieved instance that contains the data from the raw deserialized config.
-func NewRetrieved(rawConf map[string]interface{}, opts ...RetrievedOption) (Retrieved, error) {
+// The rawConf can be one of the following types:
```
mx-psi, line 105: Some of the fields we have on the Collector configuration don't have these types, e.g. `zapcore.Level` for the log level, or `configcompression.CompressionType` for the compression type. It's fine not to support them on the first version, but do we have a plan to support them eventually?

confmap/resolver.go
bogdandrutu: There are two problems, and probably we should address both somehow:
1. Embedding `${http://my.example.com?os=${env:OS_ENV}}`
2. Recursive `${file:/path/to/my/file}` which returns a section of the config as a `map[string]interface{}` for example like `{"key": ${env:VALUE}}`. Then should we expand `${env:VALUE}`?

confmap/provider.go
bogdandrutu: `zapcore.Level` is the "unmarshaled" type, the config should be "string" correct? The provider should not "unmarshal" into objects, only into primitives. The next part is other component job.

mx-psi: Right, that makes sense, thanks for clarifying

confmap/resolver.go
mx-psi: @jeromeinsf I think everyone agrees on leaving the implementation of (1) for later and restricting the allowed characters now. For (2), I believe @bogdandrutu wants to add support now with some recursion limit

confmap/provider.go
bogdandrutu: Done.

confmap/resolver.go
```diff
@@ -187,3 +188,82 @@ func (mr *Resolver) closeIfNeeded(ctx context.Context) error {
...
+		}
+		value = val
+	}
+	return nil, errors.New("too many recursive expansions")
+}
+
+func (mr *Resolver) expandValue(ctx context.Context, value interface{}) (interface{}, bool, error) {
+	switch v := value.(type) {
+	case string:
+		// If it doesn't have the format "${scheme:opaque}" no need to expand.
```
mx-psi, line 209: Something like `${HOST_IP}:4317` does not get expanded then? Why don't we use `os.Expand` like we are using on the expand provider?

mx-psi (commented):
I only have one remaining question

mx-psi:
I created #5706 for the character set allowed in URIs. I think we can deal with this independently, since we potentially already have this problem with the environment variables (in principle [any character except for `=` and NUL is allowed](https://stackoverflow.com/a/2821183))

confmap/resolver.go
bogdandrutu: Because `os.Expand` only supports returning strings (not a partial sub-map).

mx-psi: Right, but the problem of dealing with `${HOST_IP}:4317` still remains then? That would only be supported by environment variables?

bogdandrutu: Do we want to support this? What is a good use-case?

mx-psi: Here are some examples on contrib using this with environment variables:
- [sapm exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8782bd5d6b6bf44ae3d214b5c8c0d56e8293e744/exporter/sapmexporter/examples/signalfx-k8s.yaml#L71 )
- [kubeletstats receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/kubeletstatsreceiver#service-account-authentication-example) (see `endpoint`)
- [basicauth extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/basicauthextension#configuration) (see `inline`)

Why should environment variables work differently from other providers?

mx-psi: The `${THING_A}${THING_B}` pattern is also interesting. Right now, IIUC, we interpret this as "get the thing with URI `THING_A}${THING_B`", which also seems counter to what I would expect to happen (I guess this relates to #5706 too).

bogdandrutu: To make progress, I disabled this feature from the Resolver. For the moment no way to enable this, but will send another PR that addresses the remaining issues and use a featuregate to enable this maybe.

mx-psi: Makes sense. Sorry to bring this up so late, I hadn't realized until now

PR #4649, opened 2022-01-06
Remove expand cases that cannot happen with config.Map

bogdandrutu:
Using Koanf fixed the issue with AllKeys that did not correctly flatten the input map.

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #4748, opened 2022-01-25
expandStringValues function support to map[string]interface{}

kilingzhang:
**Description:** 

The use and expansion of environment variables is supported in the Collector configuration. For example:

```
processors:
  attributes/example:
    actions:
      - key: "${DB_KEY}"
        action: "${OPERATION}"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 🧰 Bug fixes 🧰
+
+- ExpandStringValues function support to map[string]interface{} (#4748) 
+
 ## v0.43.0 Beta
 
 ## 🛑 Breaking changes 🛑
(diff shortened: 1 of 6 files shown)


bogdandrutu:
@kilingzhang consider to add a test to prove that this is needed. See https://github.com/open-telemetry/opentelemetry-collector/pull/4649 where I explicitly added tests to prove that is not needed, maybe I missed something, but would be good to have a unit tests for that.

bogdandrutu:
Pressed by mistake closed instead of comment sorry.

kilingzhang:
> @kilingzhang consider to add a test to prove that this is needed. See #4649 where I explicitly added tests to prove that is not needed, maybe I missed something, but would be good to have a unit tests for that.

ok . I added some tests to prove it was necessary .

bogdandrutu (approved):
Good fix. Thanks

bogdandrutu:
Can you please rebase (resolve merge conflicts)?

codeboten (approved):

Issue #4258, opened 2021-10-25
Support optional fields in proto


codeboten:
A [change proposed](https://github.com/open-telemetry/opentelemetry-proto/pull/336) in the opentelemetry-proto uses `optional` fields, which are not supported by the gogo proto library. A few options have been proposed on how to move forward:
1. fork the gogo library and add support for optional fields
2. replace the library with manual serialization

MovieStoreGuy:
@codeboten,

Gogo has been looking for a new owner for some time now, would be better just to use the official protobuf lib for go?

https://github.com/golang/protobuf

codeboten:
There have now been 3 prototypes to attempt at adding support for optional fields:

1. Fork gogoproto and add optional feature to it: https://github.com/open-telemetry/opentelemetry-collector/pull/4259 
2. Use custom type field in gogo: https://github.com/open-telemetry/opentelemetry-collector/pull/4495
3. Use google protobuf directly in combination with vtproto for faster serialization/deserialization: https://github.com/open-telemetry/opentelemetry-collector/pull/4658 

With options 1 & 2, the burden of maintenance moving forward continues to be high, in addition to continuing a dependency on a library which is no longer maintained. The performance of option 3 essentially means we'll have to spend significant amounts of effort in another library (vtproto).

As mentioned in the SIG call on Jan 12, as a result of the previous 3 prototypes, the current thinking is to move ahead with prototyping serialization/deserialization OTLP from/to pdata directly. This has been done by the Java SIG as per the discussion [here](https://github.com/open-telemetry/opentelemetry-specification/discussions/1996)

tigrannajaryan:
> As mentioned in the SIG call on Jan 12, as a result of the previous 3 prototypes, the current thinking is to move ahead with prototyping serialization/deserialization OTLP from/to pdata directly. This has been done by the Java SIG as per the discussion [here](https://github.com/open-telemetry/opentelemetry-specification/discussions/1996)

Can we use the generated code as a starting point, which we then tweak to get rid of gogoproto dependency? This should significantly lower the implementation effort. I think Bogdan's idea of implementing a limited code generator just for our use-case is also worth exploring.

[...]

PR #4750, opened 2022-01-25
add optional fields manually

codeboten:
**Description:** 
The following PR manually adds support for Min/Max optional fields to ExponentialHistogramDataPoint/HistogramDataPoint. Some things to note:
* Once this PR is merged, we will no longer be able to use genproto to update the proto files as this is editing them manually.
* Ideally in the future we would move all the fields to serialize/deserialize directly from pdata, that being said I didn't want to start with such a major refactor here.
* Cleanup to remove proto generation code could also be done here, if we want to move forward with this change, I will open a separate PR to address this.

I believe once there is a pattern for a given field type, this approach should be easy enough to follow.

**Link to tracking Issue:** #4258 

**Testing:** ~~Added tests for serializing/deserializing the data. I plan on doing some manual tests with a mock server that uses the proto as generated by the google proto library.~~ This has been done: https://github.com/open-telemetry/opentelemetry-collector/pull/4750#issuecomment-1030212547


model/pdata/optional.go
```diff
@@ -0,0 +1,38 @@
...
+
+// NewOptionalDouble returns a new OptionalDouble from the given byte array.
+func NewOptionalDouble(value float64) OptionalDouble {
+	return OptionalDouble{orig: data.NewOptionalDouble(value)}
+}
+
+// IsEmpty returns true if id doesn't contain at least one non-zero byte.
+func (t OptionalDouble) IsEmpty() bool {
+	return t.orig.IsEmpty()
+}
```
bogdandrutu, line 34: Would be good to be kind of in sync with generated protobuf code, is this how they call the func?

model/pdata/generated_metrics.go
```diff
@@ -1366,6 +1366,26 @@ func (ms HistogramDataPoint) SetFlags(v MetricDataPointFlags) {
...
 
+// Min returns the min associated with this HistogramDataPoint.
+func (ms HistogramDataPoint) Min() OptionalDouble {
+	return OptionalDouble{orig: ((*ms.orig).Min)}
+}
+
+// SetMin replaces the min associated with this HistogramDataPoint.
+func (ms HistogramDataPoint) SetMin(v OptionalDouble) {
+	(*ms.orig).Min = v.orig
+}
```
bogdandrutu, line 1377: Another approach:

```golang
// Min returns the min associated with this HistogramDataPoint.
func (ms HistogramDataPoint) Min() float64 {
	return (*ms.orig).Min.DoubleValue
}

func (ms HistogramDataPoint) HasMin() bool {
	!(*ms.orig).Min.IsEmpty()
}


func (ms HistogramDataPoint) SetMin(v float64) {
	(*ms.orig).Min.DoubleValue = v
}
```

model/pdata/optional.go
codeboten: The generated protos checks for `nil` values. You can see the generated protos here

https://github.com/codeboten/opentelemetry-collector/pull/10/commits/7811abd5b1cf0e69a148a875a1c56a32cf2eda15#diff-7276c207e936f474133fad60cb485d044f4a8cc3931300c5b82b34fa2a238abdR810-R821

Here's an example of GetMin:
```
func (x *HistogramDataPoint) GetMin() float64 {
	if x != nil && x.Min != nil {
		return *x.Min
	}
	return 0
}
```

It uses the following to marshal:

```
if m.Max != nil {
		i -= 8
		binary.LittleEndian.PutUint64(dAtA[i:], uint64(math.Float64bits(float64(*m.Max))))
		i--
		dAtA[i] = 0x61
	}
	if m.Min != nil {
		i -= 8
		binary.LittleEndian.PutUint64(dAtA[i:], uint64(math.Float64bits(float64(*m.Min))))
		i--
		dAtA[i] = 0x59
	}
```

Following the `IsEmpty` pattern was done to remain consistent with the trace/span ID custom types already present in pdata.

model/pdata/generated_metrics.go
codeboten: Are you suggesting we do away with the `OptionalDouble` type in pdata? I don't have a strong preference to keeping it and this approach would give us essentially the same functionality without an additional type.

bogdandrutu: That is my idea.. What do you think?

codeboten: Just tried it and I agree it's better. I've added the `HasMin` `HasMax` methods to metrics.go instead of creating a new optional field generator. We can always add one later if needed.

codeboten: Decided to add a template for optional fields after all, since this will be re-used for min/max and sum in the near future. PTAL @bogdandrutu

codeboten:
To validate the proto works as expected, I tested this PR against a Python gRPC server which used the generated code from https://github.com/open-telemetry/opentelemetry-proto/pull/279. Here's the export request with min/max fields:

```
      histogram {
        data_points {
          count: 10
          bucket_counts: 21
          bucket_counts: 32
          min: 0.1
          max: 100.0
        }
      }
```

And here's another with only one of the fields:

```
      histogram {
        data_points {
          count: 10
          bucket_counts: 21
          bucket_counts: 32
          max: 100.0
        }
      }
```

codeboten:
I believe to move this PR forward, I can split it into two PRs: 
1. all the internals to support optionals (this is everything in pdata except adding the min/max fields). This could be merged before the protos to make adding the fields easy once the proto PRs merge.
2. the code that adds min/max + otlp serialization. This PR could either exist in draft mode or I can wait to open until the proto is merged.

Thoughts?

model/internal/data/optional.go
```diff
@@ -0,0 +1,71 @@
...
+// limitations under the License.
+
+package data // import "go.opentelemetry.io/collector/model/internal/data"
+
+import (
+	encoding_binary "encoding/binary"
+	"math"
+)
+
+type OptionalDouble struct {
```
bogdandrutu, line 22: This is "OptionalFixed64" correct (is not the int64 from proto is the fixed64)?

model/internal/data/protogen/metrics/v1/metrics.pb.go
```diff
@@ -2484,6 +2492,20 @@ func (m *HistogramDataPoint) MarshalToSizedBuffer(dAtA []byte) (int, error) {
...
+		}
+		dAtA[i] = 0x61
+	}
+	if !m.Min.IsEmpty() {
+		i -= m.Min.Size()
+		if _, err := m.Min.MarshalTo(dAtA[i:]); err != nil {
+			return 0, err
+		}
+		dAtA[i] = 0x59
+	}
```
bogdandrutu, line 2508: Is this manually written? How can we maintain this?

model/internal/data/optional.go
codeboten: Correct, this is a `fixed64`  field.

model/internal/data/protogen/metrics/v1/metrics.pb.go
codeboten: Correct, the process here would be that it would have to be added back to the proto after `make genproto` is run. Any thoughts on how to best address this? I suppose we could add this marshal/unmarshal methods outside of the generated code, and remove the function as part of `genproto`?

I'm open to alternatives here.

tigrannajaryan: Can we automate the patching as part of `make genproto` step? Also extract the manually written as much as possible as a func and move to a regular (non-generated) file and call it from here (minimize the line count of manual code present in this auto-generated file).

tigrannajaryan: Also add a comment clearly saying where does the manual code in this file start and where it ends. This should help to locate the right place to manually patch after re-generation and if auto-patching fails.

codeboten:
Closing in favour of #4929

PR #4762, opened 2022-01-28
Return an error in `service.NewConfigProvider` if there is no configLocations (v0.45)

noisersup:
**Description:** 
As was said in #4734, this commit:  

- changes `service.NewConfigProvider` to return an error when provided `configLocations` length is equals to 0,
- deprecates `service.MustNewConfigProvider` and makes it `panic` under the same condition as above,
- changes back all references in code from "Must" version to normal and handles error.

When the version 0.44 will come out I will undraft this PR.
Also I'm not sure is it possible to apply only third commit (the previous 2 are from my last PR) but I will remove them when my last PR will be merged.

**Link to tracking Issue:** #4730 

**Testing:** `make all`

### DIFF
--- a/service/config_provider.go
+++ b/service/config_provider.go
@@ -75,32 +75,24 @@ type configProvider struct {
 	watcher chan error
 }
 
-// ConfigProviderOptions has options that change the behavior of ConfigProvider
-// returned by NewConfigProvider()
-type ConfigProviderOptions struct {
-	configMapProviders map[string]configmapprovider.Provider
-	cfgMapConverters   []config.MapConverterFunc
-	configUnmarshaler  configunmarshaler.ConfigUnmarshaler
-}
-
 // ConfigProviderOption is an option to change the behavior of ConfigProvider
 // returned by NewConfigProvider()
-type ConfigProviderOption func(opts *ConfigProviderOptions)
+type ConfigProviderOption func(opts *configProvider)
 
 func WithConfigMapProviders(c map[string]configmapprovider.Provider) ConfigProviderOption {
-	return func(opts *ConfigProviderOptions) {
+	return func(opts *configProvider) {
 		opts.configMapProviders = c
 	}
 }
 
 func WithCfgMapConverters(c []config.MapConverterFunc) ConfigProviderOption {
-	return func(opts *ConfigProviderOptions) {
+	return func(opts *configProvider) {
 		opts.cfgMapConverters = c
 	}
 }
 
 func WithConfigUnmarshaler(c configunmarshaler.ConfigUnmarshaler) ConfigProviderOption {
-	return func(opts *ConfigProviderOptions) {
+	return func(opts *configProvider) {
 		opts.configUnmarshaler = c
 	}
 }
@@ -134,13 +126,14 @@ func MustNewConfigProvider(
 // The `configMapProviders` is a map of pairs <scheme,Provider>.
 //
 // Notice: This API is experimental.
-func NewConfigProvider(
-	locations []string,
-	opts ...ConfigProviderOption) (ConfigProvider, error) {
-
-	configOpts := &ConfigProviderOptions{}
-	for _, o := range opts {
-		o(configOpts)
+func NewConfigProvider(locations []string, opts ...ConfigProviderOption) (ConfigProvider, error) {
+	// Set default ConfigProvider values
+	configMapProvider := map[string]configmapprovider.Provider{
+		"file": configmapprovider.NewFile(),
+		"env":  configmapprovider.NewEnv(),
+	}
+	cfgMapConverter := []config.MapConverterFunc{
+		configmapprovider.NewExpandConverter(),
 	}
 
 	if len(locations) == 0 {
@@ -149,43 +142,37 @@ func NewConfigProvider(
 	// Safe copy, ensures the slice cannot be changed from the caller.
 	locationsCopy := make([]string, len(locations))
 	copy(locationsCopy, locations)
-	return &configProvider{
+
+	provider := configProvider{
 		locations:          locationsCopy,
-		configMapProviders: configOpts.configMapProviders,
-		cfgMapConverters:   configOpts.cfgMapConverters,
-		configUnmarshaler:  configOpts.configUnmarshaler,
+		configMapProviders: configMapProvider,
+		cfgMapConverters:   cfgMapConverter,
+		configUnmarshaler:  configunmarshaler.NewDefault(),
 		watcher:            make(chan error, 1),
-	}, nil
+	}
+
+	// Override default values with user options
+	for _, o := range opts {
+		o(&provider)
+	}
+
+	return &provider, nil
 }
 
-// Deprecated: use NewDefaultConfigProvider instead
+// Deprecated: use NewConfigProvider instead
 // MustNewDefaultConfigProvider returns the default ConfigProvider, and it creates configuration from a file
 // defined by the given configFile and overwrites fields using properties.
 func MustNewDefaultConfigProvider(configLocations []string, properties []string) ConfigProvider {
-	configProvider, err := NewDefaultConfigProvider(configLocations, properties)
-	if err != nil {
-		panic(err)
-	}
-	return configProvider
-}
-
-// NewDefaultConfigProvider returns the default ConfigProvider, and it creates configuration from a file
-// defined by the given configFile and overwrites fields using properties.
-func NewDefaultConfigProvider(configLocations []string, properties []string) (ConfigProvider, error) {
-	configMapProvider := map[string]configmapprovider.Provider{
-		"file": configmapprovider.NewFile(),
-		"env":  configmapprovider.NewEnv(),
-	}
 	cfgMapConverter := []config.MapConverterFunc{
 		configmapprovider.NewOverwritePropertiesConverter(properties),
 		configmapprovider.NewExpandConverter(),
 	}
-	return NewConfigProvider(
-		configLocations,
-		WithCfgMapConverters(cfgMapConverter),
-		WithConfigUnmarshaler(configunmarshaler.NewDefault()),
-		WithConfigMapProviders(configMapProvider),
-	)
+
+	configProvider, err := NewConfigProvider(configLocations, WithCfgMapConverters(cfgMapConverter))
+	if err != nil {
+		panic(err)
+	}
+	return configProvider
 }
 
 func (cm *configProvider) Get(ctx context.Context, factories component.Factories) (*config.Config, error) {
--- a/service/config_provider.go
+++ b/service/config_provider.go
@@ -75,6 +75,7 @@ type configProvider struct {
 	watcher chan error
 }
 
+// Deprecated: use NewConfigProvider instead
 // MustNewConfigProvider returns a new ConfigProvider that provides the configuration:
 // * Retrieve the config.Map by merging all retrieved maps from all the configmapprovider.Provider in order.
 // * Then applies all the ConfigMapConverterFunc in the given order.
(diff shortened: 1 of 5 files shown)


service/config_provider.go
```diff
@@ -76,6 +76,27 @@ type configProvider struct {
...
+// Deprecated: use NewConfigProvider instead
+// MustNewConfigProvider returns a new ConfigProvider that provides the configuration:
+// * Retrieve the config.Map by merging all retrieved maps from all the configmapprovider.Provider in order.
+// * Then applies all the ConfigMapConverterFunc in the given order.
+// * Then unmarshalls the final config.Config using the given configunmarshaler.ConfigUnmarshaler.
+//
+// The `configMapProviders` is a map of pairs <scheme,Provider>.
+//
+// Notice: This API is experimental.
+func MustNewConfigProvider(
```
bogdandrutu, line 88: If you want to go this path, you need to "deprecate" the current `NewConfigProvider` and in 2 weeks we change that to return error.

Same for the "NewDefault", need a must version and "deprecate"

noisersup: I deprecated it [here](https://github.com/open-telemetry/opentelemetry-collector/pull/4734/commits/22eb74f37ccd026ae0d8090575658a7e0bb57844)

The current PR was supposed to be the second step of the three given in this [comment](https://github.com/open-telemetry/opentelemetry-collector/pull/4734/commits/22eb74f37ccd026ae0d8090575658a7e0bb57844). It will be undraft after releasing version with the previous PR merged.

Also, sorry for an oversight at `NewDefaultConfigProvider` I'm going to push the commit with the must version to the [previous PR](https://github.com/open-telemetry/opentelemetry-collector/pull/4734).

bogdandrutu:
Plese rebase.

noisersup:
Should I add entry to changelog?

In my last PR `Skip Changelog` flag was added.

bogdandrutu:
@noisersup yes add a changelog entry, and rebase so we can merge this.

noisersup:
done

bogdandrutu:
@noisersup I am sorry, I missed this release, but I think that is good, because I think we can do even better now. What about we start using an Option pattern here, and keep only the locations (as required) field?

cc @codeboten @tigrannajaryan @jpkrohling 

```
func NewConfigProvider(locations []string, opts ...ConfigProviderOption) (ConfigProvider, error) {
}
```

bogdandrutu:
@noisersup looks like we have an agreement here, do you mind implementing that change?

noisersup:
@bogdandrutu Sure, I'll implement that.

noisersup:
I hope I understood the requirements

service/config_provider.go
```diff
@@ -75,6 +75,37 @@ type configProvider struct {
...
+
+// ConfigProviderOption is an option to change the behavior of ConfigProvider
+// returned by NewConfigProvider()
+type ConfigProviderOption func(opts *ConfigProviderOptions)
+
+func WithConfigMapProviders(c map[string]configmapprovider.Provider) ConfigProviderOption {
+	return func(opts *ConfigProviderOptions) {
+		opts.configMapProviders = c
+	}
+}
```
bogdandrutu, line 94: Not sure if we want a map as argument or one of the next two options:

1.
```golang
func WithConfigMapProvider(scheme string, p configmapprovider.Provider) ConfigProviderOption {
	return func(opts *ConfigProviderOptions) {
		opts.configMapProviders[scheme] = p
	}
}
```

2.
```golang

// Change configmapprovider.Provider to have a `Scheme()` func on it.

func WithConfigMapProvider(p configmapprovider.Provider) ConfigProviderOption {
	return func(opts *ConfigProviderOptions) {
		opts.configMapProviders[p.Scheme()] = p
	}
}
```

service/config_provider.go
```diff
@@ -75,6 +75,37 @@ type configProvider struct {
...
+
+func WithCfgMapConverters(c []config.MapConverterFunc) ConfigProviderOption {
+	return func(opts *ConfigProviderOptions) {
+		opts.cfgMapConverters = c
+	}
+}
```
bogdandrutu, line 100: The order of converters matters, the question I have is: do we ask the user to give a slice of them or one by one:

```golang

func WithCfgMapConverters(c config.MapConverterFunc) ConfigProviderOption {
	return func(opts *ConfigProviderOptions) {
		opts.cfgMapConverters = append(opts.cfgMapConverters, c)
	}
}

return NewConfigProvider(
	configLocations,
	WithCfgMapConverter(configmapprovider.NewOverwritePropertiesConverter(properties)),
	WithCfgMapConverter(configmapprovider.NewExpandConverter())
	)
```

bogdandrutu (commented):
I documented couple of options. Probably we can extract issues out of these and move forward with the current PR which does improve lots of the things.

service/config_provider.go
bogdandrutu: I think this is not a good idea, since if we define a default "converter" user cannot remove/change that.

service/config_provider.go
```diff
@@ -75,6 +75,37 @@ type configProvider struct {
 	watcher chan error
 }
 
+// ConfigProviderOptions has options that change the behavior of ConfigProvider
+// returned by NewConfigProvider()
+type ConfigProviderOptions struct {
+	configMapProviders map[string]configmapprovider.Provider
+	cfgMapConverters   []config.MapConverterFunc
+	configUnmarshaler  configunmarshaler.ConfigUnmarshaler
+}
```
bogdandrutu, line 84: this type is not needed. You can change the option to be: `type ConfigProviderOption func(opts *configProvider)`. External user cannot and should not be able to extend the `ConfigProviderOption` anyway.

service/config_provider.go
```diff
@@ -102,42 +136,56 @@ func MustNewConfigProvider(
...
+		panic(err)
+	}
+	return configProvider
 }
 
-// Deprecated: [v.0.44.0] use MustNewDefaultConfigProvider instead
 // NewDefaultConfigProvider returns the default ConfigProvider, and it creates configuration from a file
 // defined by the given configFile and overwrites fields using properties.
-func NewDefaultConfigProvider(configLocations []string, properties []string) ConfigProvider {
+func NewDefaultConfigProvider(configLocations []string, properties []string) (ConfigProvider, error) {
```
bogdandrutu, line 174: This should become the default values for `NewConfigProvider`. So I would remove this and change NewConfigProvider to have these map providers (user will be able to override this with the `WithCfgMapProviders`) and these map converters (user will be able to override this with the `WithCfgMapConverters`).

bogdandrutu (approved):

service/config_provider.go
noisersup: I wanted to implement both approaches in seperate commits but I have an issue with the first.

Due to the fact that 2 of the functions (`service.MustNewConfigProvider` and `service.TestConfigProvider_Errors`) that use `WithConfigMapProviders` are using `map[string]configmapprovider.Provider` approach I need to translate these maps to `[]ConfigProviderOption` and after that append this array to other Options and pass them to `NewConfigProvider` function.

example:
```go
// Convert map[string]configmapprovider.Provider to []ConfigProviderOption

configMapProvidersOpts := []ConfigProviderOption{}
for scheme, provider := range configMapProviders {
configMapProvidersOpts = append(configMapProvidersOpts, WithConfigMapProvider(scheme, provider))
}

configProvider, err := NewConfigProvider(locations, append(configMapProvidersOpts, WithCfgMapConverters(cfgMapConverters), WithConfigUnmarshaler(configUnmarshaler))...)
```

should I create a function that converts config maps to `[]ConfigProviderOption` and appends them to provided options 
to avoid reused code or it's not that big deal?

```go
func AppendConfigMapProvider(
	opts []ConfigProviderOption,
	configMapProviders map[string]configmapprovider.Provider) []ConfigProviderOption {
	// Convert map[string]configmapprovider.Provider to []ConfigProviderOption
	for scheme, provider := range configMapProviders {
		opts = append(opts, WithConfigMapProvider(scheme, provider))
	}
	return opts
}

```

service/config_provider.go
```diff
@@ -75,6 +75,29 @@ type configProvider struct {
...
+	return func(opts *configProvider) {
+		opts.configMapProviders = c
+	}
+}
+
+func WithCfgMapConverters(c []config.MapConverterFunc) ConfigProviderOption {
+	return func(opts *configProvider) {
+		opts.cfgMapConverters = c
+	}
+}
```
Aneurysm9, line 92:
```suggestion
func WithConfigMapConverters(c ...config.MapConverterFunc) ConfigProviderOption {
	return func(opts *configProvider) {
		opts.cfgMapConverters = append(opts.cfgMapConverters, c...)
	}
}
```

I think I would prefer to see these append to the config, allowing for multiple invocations of the options, possibly provided by separate packages.  The invocation of `NewConfigProvider()` ultimately controls the order in which they're applied.  A small change would be required there to set the default values after applying options if the list is empty, rather than setting it first and allowing it to be overwritten.

The same applies to `WithConfigMapProviders()`.  There order is not important, so multiple maps can be merged.

Aneurysm9:
The title of this PR is no longer adequately descriptive, can you please update it?

@open-telemetry/collector-approvers please review this PR as it has changed in scope and may have more significant impact than initially proposed.

service/config_provider.go
Aneurysm9: > // Change configmapprovider.Provider to have a `Scheme()` func on it.

I think I prefer this approach.  We require that providers have a scheme anyways, so why not a way to determine what they think their scheme is?

> should I create a function that converts config maps to []ConfigProviderOption and appends them to provided options to avoid reused code or it's not that big deal?

An unexported helper function to do this might be useful, but at least `service.MustNewConfigProvider()` is expected to go away in the near future, so I'm not sure how big a deal it is.

service/config_provider.go
```diff
@@ -75,6 +75,29 @@ type configProvider struct {
...
+// returned by NewConfigProvider()
+type ConfigProviderOption func(opts *configProvider)
+
+func WithConfigMapProviders(c map[string]configmapprovider.Provider) ConfigProviderOption {
+	return func(opts *configProvider) {
+		opts.configMapProviders = c
+	}
+}
+
+func WithCfgMapConverters(c []config.MapConverterFunc) ConfigProviderOption {
```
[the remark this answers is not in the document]
bogdandrutu: +1

service/config_provider.go
```diff
@@ -75,6 +75,29 @@ type configProvider struct {
 	watcher chan error
 }
 
+// ConfigProviderOption is an option to change the behavior of ConfigProvider
+// returned by NewConfigProvider()
+type ConfigProviderOption func(opts *configProvider)
```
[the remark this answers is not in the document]
bogdandrutu: Same effect. Interface with a private func or a func working on a private type. No real difference. We can discuss this style in a separate issue/PR

service/config_provider.go
bogdandrutu: I am confused about your comment: `I think I would prefer to see these append to the config, allowing for multiple invocations of the options, possibly provided by separate packages.` are you suggesting a global?

Even without the "default" problem, your suggestion does not trivially allow "no converters" case.

Also these append thing is unclear to me, I would rather provide only one "Converter" and a "MultiConverter" implementation if the list is a problem.

service/config_provider.go
bogdandrutu: > I think I prefer this approach. We require that providers have a scheme anyways, so why not a way to determine what they think their scheme is?

After more thoughts on this, I think is the right thing to do, since this will allow us to use the same pattern in the builder to support more providers - builder will look for NewProvider in the given import (similar to NewFactory for components.

Per @Aneurysm9 comment the `service.MustNewConfigProvider()` is going away, and the other one is a test. I would rather look into an example how a distribution with extra providers will use this. Or an example written for this case.

service/config_provider.go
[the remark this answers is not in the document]
bogdandrutu: Context: We already have this pattern, which offers the same flexibility/protection. So let's discuss separately.

PR #4770, opened 2022-01-31
Add Host to authInterceptor

orloffv:
**Description:** Add Host to authInterceptor

**Link to tracking Issue:** Closes https://github.com/open-telemetry/opentelemetry-collector/issues/4722

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -25,6 +25,7 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel"
 
+	"go.opentelemetry.io/collector/client"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
 	"go.opentelemetry.io/collector/config/configauth"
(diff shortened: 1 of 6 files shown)


orloffv:
conflicts, because wait for revert https://github.com/open-telemetry/opentelemetry-collector/pull/4736

config/confighttp/confighttp.go
```diff
@@ -339,7 +340,11 @@ type CORSSettings struct {
 
 func authInterceptor(next http.Handler, authenticate configauth.AuthenticateFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx, err := authenticate(r.Context(), r.Header)
+		headers := r.Header.Clone()
```
jpkrohling, line 343: Should we remove the same logic that was added as part of the previous PR? Should we do something similar for gRPC here as well?

orloffv: 1) I reverted all pervious PR - https://github.com/open-telemetry/opentelemetry-collector/pull/4736
2) For me - no, because this problem only for http. If you want, i can add

jpkrohling: Instead of reverting that commit, use this PR here alone to show the differences. I believe it's about just moving the logic to a different place. And yes, please fix it for gRPC as well.

bogdandrutu:
Consider to rebase since changes are merged with previous PRs and hard to review.

orloffv:
@bogdandrutu @jpkrohling ok
first commit - revert,
two commits add logic for http&grpc handler

orloffv:
friendly ping

jpkrohling (approved):
Looks good, but have you checked your downstream authenticator will benefit from this change? Or perhaps are you still missing something?

orloffv:
yes i already check and use it in my production

jpkrohling:
We have a similar request, #4793, which caused discussions to handle this in a generic way. I'm not sure we should merge this as is and break consumers soon, or hold this change to wait for that, preventing people from having this at all.

Given that you seem to have a custom build for production, I'm right now more keen on waiting for the final solution.

Issue #4633, opened 2022-01-04
Replace path.Join method with filepath.Join method


codeboten:
As per @bogdandrutu's [comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/6480#issuecomment-1005016589), we should follow the same pattern in the core repo.

shree007:
@jpkrohling  @codeboten I may work on it.

jpkrohling:
Go ahead.

PR #4780, opened 2022-01-31
Replace usage of path.Join with filepath.Join, better windows support

bogdandrutu:

### DIFF
--- a/config/configmap_test.go
+++ b/config/configmap_test.go
@@ -17,7 +17,7 @@ package config
 import (
 	"fmt"
 	"io/ioutil"
-	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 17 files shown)


dmitryax (approved):

codeboten (approved):
Were these leftover from https://github.com/open-telemetry/opentelemetry-collector/issues/4633?

codeboten (commented):
Appears this made windows tests mad :D

bogdandrutu:
@codeboten maybe, that issue was all about `path.Join(".",` but it is hard to do only in a specific case. For consistency and easy to follow rule, I want to use "filepath" everywhere.

bogdandrutu:
> Appears this made windows tests mad :D

This is because in 2 cases we had to use "path" package, when we are constructing URL path :))

PR #4787, opened 2022-02-01
[Builder]: Remove test that depends on the contrib

bogdandrutu:
Remove unnecessary test for replace since the `cmd/otelcorecol` tests the latest binary with replace statements already.


cmd/builder/test/core.builder.yaml
```diff
@@ -13,3 +13,7 @@ receivers:
 exporters:
   - import: go.opentelemetry.io/collector/exporter/loggingexporter
     gomod: go.opentelemetry.io/collector v0.43.1
+
+replaces:
```
jpkrohling, line 17: I still see value in having a test that is stable, independent from the current core code. The replaces test could have a manifest similar to the one you have here in this PR.

bogdandrutu: What is the "backwards" compatibility we guarantee here? Because having this "stable" release means that is the version we support to build, do we support 10 versions behind? Last version? What is the guarantee we have?

This changes to say we support only the "current" version, and not behind.

jpkrohling: No, we test the latest released version here.

bogdandrutu: Done, clarify and change to test with latest release and latest merged code. These are the things we guarantee at this moment correct?

jpkrohling (approved):

cmd/builder/test/core.builder.yaml
jpkrohling: Correct

PR #4789, opened 2022-02-01
Transform configmapprovider.Retrieved to a struct

bogdandrutu:
The non deprecated core is moved to provider.go for readability improvement.

Deprecates all the old interface funcs, and helpers to construct the struct.


bogdandrutu:
This is an alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/4721, @jpkrohling what do you think?

jpkrohling (commented):
I like this better than the current version or the original PR.

config/configmapprovider/provider.go
```diff
@@ -77,3 +79,23 @@ type ChangeEvent struct {
 	// Any non-nil error indicates that there was a problem with watching the config changes.
 	Error error
 }
+
+// Retrieved holds the result of a call to the Retrieve method of a Provider object.
+type Retrieved struct {
+	Cfg *config.Map
```
Aneurysm9, line 85:
```suggestion
	Map *config.Map
```

Not fond of the `Cfg` field name.

service/config_provider.go
```diff
@@ -213,8 +205,8 @@ func (cm *configProvider) Shutdown(ctx context.Context) error {
 // https://tools.ietf.org/id/draft-kerwin-file-scheme-07.html#syntax
 var driverLetterRegexp = regexp.MustCompile("^[A-z]:")
 
-func (cm *configProvider) mergeRetrieve(ctx context.Context) (configmapprovider.Retrieved, error) {
-	var retrs []configmapprovider.Retrieved
+func (cm *configProvider) mergeRetrieve(ctx context.Context) (*configmapprovider.Retrieved, error) {
+	var retrs []configmapprovider.CloseFunc
```
Aneurysm9, line 209:
```suggestion
	var closers []configmapprovider.CloseFunc
```

This is now a different type of list and should be named accordingly.

Aneurysm9 (commented):
I think I'm good with this approach if this is the consensus.

config/configmapprovider/retrieved.go
```diff
@@ -21,38 +21,42 @@ import (
...
-	//
-	// Should block until all resources are closed, and guarantee that `onChange` is not
-	// going to be called after it returns except when `ctx` is cancelled.
-	//
-	// Should never be called concurrently with itself or Get.
-	Close(ctx context.Context) error
+// Get returns the config Map. Should never be called after Close.
+// Should never be called concurrently with itself or Close.
+//
+// Deprecated: Not needed, will be removed soon.
```
dmitryax, line 27: Should we update the workflow now https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configmapprovider/provider.go#L27-L37 ?

config/configmapprovider/retrieved.go
```diff
@@ -21,38 +21,42 @@ import (
...
+func (r Retrieved) Close(ctx context.Context) error {
+	if r.CloseFunc == nil {
+		return nil
+	}
+	return r.CloseFunc(ctx)
 }
 
 // GetFunc specifies the function invoked when the Retrieved.Get is being called.
+// Deprecated: Not needed, will be removed soon.
```
codeboten, line 49: Can you add a note for users to know why this is "Not needed" 😬

codeboten (commented):
I agree with @jpkrohling, I think this approach is cleaner than #4721

service/config_provider.go
bogdandrutu: Done.

config/configmapprovider/provider.go
bogdandrutu: Done.

config/configmapprovider/retrieved.go
```diff
@@ -21,38 +21,42 @@ import (
...
+// Get returns the config Map. Should never be called after Close.
+// Should never be called concurrently with itself or Close.
+//
+// Deprecated: Not needed, will be removed soon.
+func (r Retrieved) Get(ctx context.Context) (*config.Map, error) {
+	return r.Cfg, nil
+}
 
-	// privateRetrieved is an unexported func to disallow direct implementation.
-	privateRetrieved()
```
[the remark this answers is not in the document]
bogdandrutu: Indeed.

config/configmapprovider/retrieved.go
bogdandrutu: done.

config/configmapprovider/retrieved.go
bogdandrutu: :)) I don't know how to explain :)) more than you have access to the member...

bogdandrutu:
@codeboten @jpkrohling @dmitryax @Aneurysm9 can I get an approve as well?

dmitryax:
Please add a changelog line

config/configmapprovider/retrieved.go
```diff
@@ -21,38 +21,42 @@ import (
...
+// Close signals that the configuration for which it was used to retrieve values is
+// no longer in use and should close and release any watchers that it may have created.
+//
+// Should block until all resources are closed, and guarantee that `onChange` is not
+// going to be called after it returns except when `ctx` is cancelled.
+//
+// Should never be called concurrently with itself or Get.
+//
+// Deprecated: Not needed, will be removed soon. You have access to the struct members.
```
dmitryax, line 40: Don't you think this one can still be useful?

bogdandrutu: Why? May be... but I want to limit the public api "when in doubt, leave it out" :)

dmitryax: I mean r.Close() seems cleaner than r.CloseFunc() with a condition :) 

If we are considering adding it later maybe makes sense to keep it now

bogdandrutu: That is the whole idea, I am in doubt :) sooo.. I prefer to leave it out.

bogdandrutu: Are you 100% convince that we need that "helper" func?

bogdandrutu: I tried one more option, but still not convinced that this helper is really needed.

dmitryax (approved):

PR #4793, opened 2022-02-02
Include Query Params in request map for configauth

neelayu:
**Description:**
Feature to include query params along with header in confighttp to pass to authenticator function.

**Link to tracking Issue:** NA

**Testing:** Unit

It would be good to have query params added to the "headers" map when considering http auth. The reason is for custom auth extensions could need data from query string(ex api_key, token) for authentication. This PR includes the query params in the headers itself which is renamed to a more generic `requestMap`. This allows non-breaking change for existing gRPC auth as well.

### DIFF
--- a/config/configauth/default_serverauthenticator.go
+++ b/config/configauth/default_serverauthenticator.go
@@ -58,7 +58,7 @@ func WithShutdown(shutdownFunc componenthelper.ShutdownFunc) Option {
 // NewServerAuthenticator returns a ServerAuthenticator configured with the provided options.
 func NewServerAuthenticator(options ...Option) ServerAuthenticator {
 	bc := &defaultServerAuthenticator{
-		AuthenticateFunc: func(ctx context.Context, headers map[string][]string) (context.Context, error) { return ctx, nil },
+		AuthenticateFunc: func(ctx context.Context, requestMap map[string][]string) (context.Context, error) { return ctx, nil },
 		StartFunc:        func(ctx context.Context, host component.Host) error { return nil },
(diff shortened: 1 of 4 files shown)


bogdandrutu (changes requested):
Soon we will copy everything 10 times... I think we need to stop limiting. If you want have "headers" + "other stuff" instead of creating maps after maps.

@jpkrohling please provide a design doc of where you want to get with this.

neelayu:
Sure. The func signature defines headers which works for both gRPC and HTTP. However the actually request instances are different for both. So we can make use of this approach, or pass the value in the context, which I feel might obfuscate a lot of things.

jpkrohling:
I'm converting this to a draft, as it's clear that we want to have a better generic way to handle this. @neelayu, could you please create an issue to track this? I'll share my thoughts there once the issue is opened.

neelayu:
Created #4806

bogdanov1609:
> I'm converting this to a draft, as it's clear that we want to have a better generic way to handle this. @neelayu, could you please create an issue to track this? I'll share my thoughts there once the issue is opened.

@neelayu @jpkrohling could you please provide relevant issue for this pr? We are interested in this feature and maybe can help.

neelayu:
> > I'm converting this to a draft, as it's clear that we want to have a better generic way to handle this. @neelayu, could you please create an issue to track this? I'll share my thoughts there once the issue is opened.
> 
> @neelayu @jpkrohling could you please provide relevant issue for this pr? We are interested in this feature and maybe can help.

@bogdanov1609 

#4806 is the issue where the discussions were taking place. Thanks!

PR #4804, opened 2022-02-03
Remove usage of LogRecord.Name field

tigrannajaryan:
The LogRecord Name field [is removed](https://github.com/open-telemetry/opentelemetry-specification/pull/2271) from the specification and is [going to be removed](https://github.com/open-telemetry/opentelemetry-proto/pull/357) from the OTLP.


dmitryax (approved):

model/otlp/json_test.go
```diff
@@ -55,11 +55,11 @@ var logsOTLP = func() pdata.Logs {
 	il := rl.InstrumentationLibraryLogs().AppendEmpty()
 	il.InstrumentationLibrary().SetName("name")
 	il.InstrumentationLibrary().SetVersion("version")
-	il.LogRecords().AppendEmpty().SetName("testMetric")
+	il.LogRecords().AppendEmpty().SetSeverityText("testMetric")
```
codeboten, line 58: The usage here is a bit confusing to me as you're setting what used to be the name as a severity text, I would expect this field to be something like "INFO" or "WARN" rather than testMetric. I understand this is an arbitrary string field but it may cause confusion for others.

tigrannajaryan: Fixed.

model/otlp/json_test.go
```diff
@@ -55,11 +55,11 @@ var logsOTLP = func() pdata.Logs {
 	il := rl.InstrumentationLibraryLogs().AppendEmpty()
 	il.InstrumentationLibrary().SetName("name")
 	il.InstrumentationLibrary().SetVersion("version")
-	il.LogRecords().AppendEmpty().SetName("testMetric")
+	il.LogRecords().AppendEmpty().SetSeverityText("Error")
```
bogdandrutu, line 58: Extremely confused by the choice of replacing this with Severity. Should severity not be an enum?

tigrannajaryan: SeverityText is a string (SeverityNumber is an enum). 

In this particular test it does not matter what field we use. This is merely to populate a log record that we later check exists after serialization. The checks are not comprehensive, they only verified a single field (Name) and I did not aim to improve the tests in any way, merely replaced Name by SeverityText to keep the test as is.

pmm-sumo (approved):

PR #4809, opened 2022-02-05
Remove deprecated funcs in pdata

bogdandrutu:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
   - Deprecated `receiverhelper.WithLogs` in favour of `component.WithLogsReceiver`
   - Deprecated `receiverhelper.NewFactory` in favour of `component.NewReceiverFactory`
 - Remove support for legacy otlp/http port. (#4916)
+- Remove deprecated funcs in pdata (#4809)
 
 ### 💡 Enhancements 💡
 
--- a/model/pdata/logs.go
+++ b/model/pdata/logs.go
@@ -92,19 +92,6 @@ func (ld Logs) ResourceLogs() ResourceLogsSlice {
 	return newResourceLogsSlice(&ld.orig.ResourceLogs)
 }
 
-// Deprecated: [v0.44.0] use LogRecordSlice
-type LogSlice = LogRecordSlice
-
-// Deprecated: [v0.44.0] use `LogRecords`.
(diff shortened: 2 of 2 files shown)


model/pdata/logs.go
```diff
@@ -96,19 +96,6 @@ func (ld Logs) ResourceLogs() ResourceLogsSlice {
 	return newResourceLogsSlice(&ld.orig.ResourceLogs)
 }
```
codeboten, line 98: Note for future maintainers: it would be helpful to identify the version in which a Deprecated message was added. This would help us ensure a version has been released that includes the deprecation message without having to dig through the git history. Any thoughts on how this would be best captured?

codeboten (approved):

model/pdata/logs.go
dmitryax: Maybe right in the comment?

```
// Deprecated as of 0.42.0: LogRecordSlice
```

dmitryax (approved):

model/pdata/logs.go
bogdandrutu: @dmitryax please update contributions guide with this. Thanks

dmitryax: What I suggested doesn't work because prefix `// Deprecated:` is required. Prepared a PR with a similar syntax that fits the requirement https://github.com/open-telemetry/opentelemetry-collector/pull/4822

CHANGELOG.md
```diff
@@ -6,12 +6,18 @@
...
+
+- Remove deprecated funcs in pdata (#4809)
+
 ## v0.45.0 Beta
 
 ### 🛑 Breaking changes 🛑
 
 - Remove deprecated funcs in configtelemetry (#4808)
 - Deprecate `service/defaultcomponents` go package (#4622)
+- Remove deprecated funcs in pdata (#4809)
```
jpkrohling, line 20: This should be removed.

jpkrohling (changes requested):
For some reason, I'm getting conflicts that are not apparent via the web UI when rebasing this to remove the changelog entry. Perhaps someone could do it instead?

bogdandrutu:
@jpkrohling PTAL :)

bogdandrutu:
@jpkrohling ping :)

PR #4810, opened 2022-02-05
Use go mod compat, drops support for reproducibility with go 1.16

bogdandrutu:


Makefile
```diff
@@ -90,8 +90,7 @@ gofmt:
 .PHONY: gotidy
 gotidy:
 	$(MAKE) for-all CMD="rm -fr go.sum"
-	$(MAKE) for-all CMD="go mod tidy -go=1.16"
-	$(MAKE) for-all CMD="go mod tidy -go=1.17"
```
jpkrohling, line 94: What are the practical effects of having 1.16 and then 1.17 vs. just 1.17?

Aneurysm9: Aren't we already there by virtue of having a `go 1.17` directive?  I thought we were also using some 1.17 features/functions in places.

Aneurysm9: > What are the practical effects of having 1.16 and then 1.17 vs. just 1.17?

My understanding is that `-go=1.16` will rewrite the mod/sum files as 1.16 would, with complete transitive dependency info and then `-go=1.17` will rewrite from there to the condensed form with inlined indirect dependencies, but using the minimum version calculated from the previous 1.16 files.  This ensures that the same versions are selected as would be by 1.16.

Using 1.17 directly does not calculate the versions that 1.16 would select and may result in different versions being selected, accounting for the differences that [@codeboten mentioned](https://github.com/open-telemetry/opentelemetry-collector/pull/4810#pullrequestreview-875048500).

Aneurysm9 (approved):

Issue #4782, opened 2022-02-01
[pdatagen] Generate types for Metric oneof fields


dmitryax:
pdatagen should generate types and type accessors for Metric oneof fields: NumberDataPoint/Exemplar.Type and  Metric.DataType

Resolves TODO from https://github.com/open-telemetry/opentelemetry-collector/pull/4773#discussion_r796192215.

PR #4811, opened 2022-02-05
[pdatagen] Generate Metric type accessors

dmitryax:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/4782

### DIFF
--- a/model/pdata/generated_metrics.go
+++ b/model/pdata/generated_metrics.go
@@ -609,6 +609,24 @@ func (ms Metric) SetUnit(v string) {
 	(*ms.orig).Unit = v
 }
 
+// DataType returns the type of the data for this Metric.
+// Calling this function on zero-initialized Metric will cause a panic.
+func (ms Metric) DataType() MetricDataType {
+	switch ms.orig.Data.(type) {
+	case *otlpmetrics.Metric_Gauge:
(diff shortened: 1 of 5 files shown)


model/pdata/generated_metrics.go
```diff
@@ -1106,6 +1124,18 @@ func (ms NumberDataPoint) SetTimestamp(v Timestamp) {
 	(*ms.orig).TimeUnixNano = uint64(v)
 }
 
+// Type returns the type of the value for this NumberDataPoint.
+// Calling this function on zero-initialized NumberDataPoint will cause a panic.
+func (ms NumberDataPoint) Type() MetricValueType {
```
bogdandrutu, line 1129: On the list of "may fix": MetricValueType does not sound as the best name for the enum. Maybe NumberValueType?
PointValueType? Not sure, just open an issue, this name does not sound very clear to me, because I would ask why Histogram is not a value type.

bogdandrutu (approved):

model/pdata/generated_metrics.go
dmitryax: Makes sense. Filed an issue https://github.com/open-telemetry/opentelemetry-collector/issues/4819

Issue #4815, opened 2022-02-07
Confusing mention to `--new-metrics` flag in the documentation


acm19:
**Describe the bug**
`--new-metrics` flag is mentioned here: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md, this flag has been removed: https://github.com/open-telemetry/opentelemetry-collector/pull/2105. Mention to it is confusing.

**Steps to reproduce**
Not applicable.

**What did you expect to see?**
No mention to a flag that doesn't exist for this version of the software. 

**What did you see instead?**
Not applicable.

**What version did you use?**
Latest main branch.

**What config did you use?**
Not applicable.

**Environment**
Not applicable.

**Additional context**
If somebody is using an older version of the software they should check out the old documentation but current documentation shouldn't mention inexistent flags. I'm happy to create a PR.

PR #2105, opened 2020-11-10
Remove legacy metrics, they were marked as legacy for ~12 months

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/1082

This PR removes:
- Legacy receiver/exporter metrics:
  - otelcol/receiver/received_spans
  - otelcol/receiver/dropped_spans
  - otelcol/receiver/received_timeseries
  - otelcol/receiver/dropped_timeseries
  - otelcol/exporter/received_spans
  - otelcol/exporter/dropped_spans
  - otelcol/exporter/received_timeseries
  - otelcol/exporter/dropped_timeseries
  - otelcol/exporter/received_logs
  - otelcol/exporter/dropped_logs
- For processors remove the legacy metrics (new metrics have the same data, different names):
  - e.g. "spans_dropped" -> "processor/spans_dropped"
  - e.g. "batch_send_size_bytes" -> "processor/batch/batch_send_size_bytes"

All the new metrics were enabled when using the --new-metrics flag. The PR also removes two flags:
- "--new-metrics"
- "--legacy-metrics"

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #4825, opened 2022-02-09
Remove reference to `--new-metrics` in the docs

acm19:
Removes reference to inexistent flag `--new-metrics` as it's default and
this flag doesn't exist anymore. It was removed here:
https://github.com/open-telemetry/opentelemetry-collector/pull/2105

Resolves: https://github.com/open-telemetry/opentelemetry-collector/issues/4815


jpkrohling:
Thanks for your effort, but we seem to have a PR for this already: #4823

Issue #4826, opened 2022-02-09
Collector Lifecycle Inconsistencies


cpheps:
**Describe the bug**
There are a few lifecycle inconsistencies in the collector lifecycle management. 

1. In `Collector.Run` the collector state is not updated if the collector [errors on startup](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L221-L227). Leaving the collector in a `Starting` state even though `Run` has excited with and error.
2. There's a race condition/panic in `Collector.Shutdown` if `Shutdown` is called before the [logger is initialized](https://github.com/open-telemetry/opentelemetry-collector/blob/18fb01178b794a0b8e2f3b0e48ab6f81056ae2b5/service/collector.go#L175) it will panic in the `Shutdown` [recover](https://github.com/open-telemetry/opentelemetry-collector/blob/18fb01178b794a0b8e2f3b0e48ab6f81056ae2b5/service/collector.go#L116).

**Steps to reproduce**

1. Call `Collector.Run` with a bad config and then call `Collector.GetState` to observe state is still `Starting`.
2. It's harder to reliably trigger the race condition but if you call `Collector.Run` with a bad config then call `Collector.Shutdown` it will have the same panic.

**What did you expect to see?**

1. State updates correctly changed to `Closed` or maybe a new `Failed` or `Error` state to indicate a failed startup.
2. `Shutdown` never panics no matter when it's called.

**What did you see instead?**
1. State is `Started`
2. `Shutdown` panics

**What version did you use?**
v0.43.0

**What config did you use?**
Any invalid configuration will work.

**Environment**
OS: macOS Monterey (12.2)
Compiler: go 1.17.6

**Additional context**
I'd be happy to submit a PR to fix the above I think the only open question is if the collector errors on startup should the state be `Closed` or should we introduce a new state to represent a failure?

It's a bit unclear how `Collector.Run` and `Collector.GetState` are supposed to be used by the caller. It seems like `Run` should be called an expected to block until the collector finishes (with or without an error). It's not clear if `GetState` is meant to be constantly monitored to help with managing a collector or if it's purely informational on what the collector is doing.

cpheps:
Created a PR to fix the Shutdown race condition as that was straight forward and didn't require discussion. Keeping issue open as there still needs to be a decision on the collectors state when it fails to startup.

PR #4827, opened 2022-02-09
Set Collector Log in New to avoid Shutdown panic

cpheps:
Fixing a bug - Fixes Shutdown panic/race condition described in #4826. This would occur by calling `Shutdown` before the logger was initialized. Added a `Nop` logger when creating a new logger so that a logger will always exist when the collector exists.

**Link to tracking Issue:** #4826

**Testing:** Added a unit test to verify `Shutdown` could be called before a collector is started and a logger is configured.


bogdandrutu:
I think the access to the log pointer must be atomic

cpheps:
> I think the access to the log pointer must be atomic

@bogdandrutu I agree looks like a non-atomic issue would have existed before this change. I think making the log pointer atomic should be it's own issue/pr. I can write it up if you agree.

cpheps:
@dmitryax Do you mind reviewing looks like Github assigned you as reviewer.

Issue #4832, opened 2022-02-10
Split pdata into multiple packages


dmitryax:
## Problem

As a result of [discussion](https://github.com/open-telemetry/opentelemetry-collector/issues/4705) on how to handle experimental data types, a decision was made to split pdata package into multiple modules per signal and publish them as separate modules. `pdata` package incapsulates internal proto data structures into private fields. Given that currently all signal data structures (logs, metrics, traces) use common structures like `Resource`, `AttributeMap`, `AttributeValue`, they need direct access to the private proto fields of the common structures. Splitting them as is cannot be done since it removes the access to the private fields of common data types from other signal packages. There are several ways how this can be solved, all of them have some drawbacks.

### 1. Wrap common pdata structs

Common pdata structures like `Resource` and others can be wrapped into another struct that is accessible in a public module while wrapped part can be placed into an internal package that all other signals have access to.

Cons:
- Complicates common data structures
- Additional computation overhead to wrap/unwrap

### 2. Add new exported functions to common data structures to be used only by other pdata signals

Private functions of common data types that currently used to pass private proto fields can be exported. In that case other signal modules can have access to it. 

For example:

```
func newResource(orig *otlpresource.Resource) Resource
```

can be changed to 

```
func NewResourceFromProto(orig *otlpresource.Resource) Resource {
```

Pros:
- Less additional complexiy 

Cons:
- Introducing exported functions that are needed for internal use only.

### 3. Put `pdata` into an internal package as is and use aliases to it in exported modules

The whole pdata package can be put as is into an internal package. And exported packages will contain only aliases to the internal pdata package.

Pros:
- Easy to generate and maintain

Cons:
- Slightly complicates the module structure and godoc.

## Proposal

The third solution seems to be the most appropriate. It keeps the overhead minimal while introducing no changes to existing pdata structures. It just adds another layer of reference in the code and generated documentation.

bogdandrutu:
My first question/reaction: Considering all these problems... Do we still want to split the packages?

dmitryax:
Also the only downside of the 3 approach can be mitigated to some extent. Signal specific parts of the `pdata` code that is put under internal in https://github.com/open-telemetry/opentelemetry-collector/pull/4833 potentially can be moved to exported p<signal> modules and only `pcommon` will stay with the aliases.

Aneurysm9:
[...]

PR #4833, opened 2022-02-10
Move pdata into an internal package and use aliases in exported package

dmitryax:
This change is the first step towards splitting the pdata package into multiple modules. It implements the approach 3 outlined [here](https://github.com/open-telemetry/opentelemetry-collector/issues/4832).

In this PR, all files from `model/pdata` are moved to `model/internal/pdata` as is, new files with aliases to the internal pdata package are added to the existing `model/pdata` package.

If this is accepted, next it can be easily used to generate other p\<signal\> packages in _addition_ to existing exported pdata package until we find out what to do with components relying on it:
- If we are going to split other components, they can be changed along with switching to the new p\<signal\> packages.
- If we don't split them, we will need to run all the components relying on pdata through the 3-step deprecation process.


bogdandrutu (approved):
This change is necessary to move forward with the plan of splitting the pdata package.

Aneurysm9 (changes requested):
I don't believe that there is yet consensus on https://github.com/open-telemetry/opentelemetry-collector/issues/4832 or agreement that this PR is necessary to effect a split of pdata into multiple modules.  We need to hold off on merging this until a plan for #4832 is agreed.

bogdandrutu:
Timeline of events:
1. Decision to split in multiple packages was made long time, and everyone agreed on.
2. During the implementation of that decision @dmitryax and @bogdandrutu, wrongly made an assumption that the internal package cannot be shared between modules, and will have to expose publicly things that we don't want. That's when the https://github.com/open-telemetry/opentelemetry-collector/issues/4832 was created.
3. During last SIG @Aneurysm9 pointed that the sharing of internal packages is possible, based on go example with the internal/global package. This was documented in the issue.

Now: because of 3) the solution for splitting is trivial (not sure if more documentation is needed). But essentially will have a module with all the "internal not exported" that will be shared by all the other packages that need access to that. This module does not necessary need to be separate module or be inside the `common` module that anyway every other module depends on.

dmitryax:
Looks like we have a consensus [here](https://github.com/open-telemetry/opentelemetry-collector/issues/4832) and [here](https://github.com/open-telemetry/opentelemetry-collector/pull/4918).

@codeboten @Aneurysm9 @tigrannajaryan PTAL at this PR

bogdandrutu:
@open-telemetry/collector-maintainers are we ok to move forward with this intermediate step, since for the final PR there are still "subtle" things to discuss as the final name of the package "trace" vs "ptrace" the final name of the public structs "TracesUnmarshaler" vs "Unmarshaler" which we can focus on a smaller PR that only exposes these.

tigrannajaryan:
> @open-telemetry/collector-maintainers are we ok to move forward with this intermediate step, since for the final PR there are still "subtle" things to discuss as the final name of the package "traces" vs "ptraces" the final name of the public structs "TracesUnmarshaler" vs "Unmarshaler" which we can focus on a smaller PR that only exposes these.

I am OK.

bogdandrutu (approved):

PR #4869, opened 2022-02-16
[confighttp] Allow CORS requests with configured auth

svrakitin:
**Description:** 

If authenticator is configured it will deny CORS requests which do not carry auth. This change reorders auth and CORS middlewares so CORS takes precedence.

**Testing:**

There is a test which ensures that auth will be skipped for CORS requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 - Transform configmapprovider.Retrieved interface to a struct (#4789)
 - Added feature gate summary to zpages extension (#4834)
 
+### 🧰 Bug fixes 🧰
+
+- `confighttp`: Allow CORS requests with configured auth (#4869)
+
 ### 🚩 Deprecations 🚩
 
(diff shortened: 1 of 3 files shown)


jpkrohling (approved):

jpkrohling:
Could you please add an entry to the changelog, under bug fixes?

svrakitin:
@jpkrohling Yeah, on it.

bogdandrutu:
@jpkrohling I am not an expert at all in CORS, but I trust you :) if you know about CORS there are couple of related issues that you may want to also look into :)

jpkrohling:
> I am not an expert at all in CORS

Nobody is :-) This PR does look right though, as CORS handling should indeed happen before auth, as those are pre-flight (OPTIONS method) requests.

> if you know about CORS there are couple of related issues that you may want to also look into :)

Ping me there, I can take a look and assign to myself if I think I can help there.

PR #4879, opened 2022-02-17
Check for Unmarshallable internal arguments

bogdandrutu:
This has a limitation that do not check for the top level (the result pass to Unmarshal call). The reason is because cannot easily avoid a circular call (can be done by keeping a state in the Map).

Before doing more things, would like to know if this is something we want to support.

/CC @open-telemetry/collector-approvers

### DIFF
--- a/config/configmap.go
+++ b/config/configmap.go
@@ -62,23 +62,13 @@ func (l *Map) AllKeys() []string {
 
 // Unmarshal unmarshalls the config into a struct.
 // Tags on the fields of the structure must be properly set.
-func (l *Map) Unmarshal(rawVal interface{}) error {
-	decoder, err := mapstructure.NewDecoder(decoderConfig(rawVal))
-	if err != nil {
-		return err
-	}
-	return decoder.Decode(l.ToStringMap())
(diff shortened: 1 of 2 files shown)


dmitryax:
What is the goal for this change? Do we want this to automatically detect any configuration parts that are `Unmarshallable` (not only component configs)? Or is there another reason?

bogdandrutu:
> Do we want this to automatically detect any configuration parts (not only component configs) that are Unmarshallable? 

@dmitryax YES :)

djaglowski:
I believe I have found need of this change for migrating `pkg/stanza` unmarshaling to mapstructure/confmap. It appears the PR timed out without any specific objections. Can this be reopened?

djaglowski (commented):
This requires a changelog entry and a couple lint fixes, but looks good to me otherwise.

djaglowski:
@bogdandrutu, this just needs a couple minor things. Do you want to update it? Alternatively, mind if I open a new PR with the same + fixes?

bogdandrutu:
@djaglowski go for it, happy to review :)

PR #4249, opened 2021-10-22
Fix expand nil struct pointers, add tests to confirm the fix

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #4880, opened 2022-02-17
Add tests for expandNilStructPointersFunc, left TODO to investigate

bogdandrutu:
Follow-up after https://github.com/open-telemetry/opentelemetry-collector/pull/4249, I was not able to finish that, and want to see if somebody else can continue. Added tests (which are always good to validate the current behavior), but left few TODOs about things to investigate.


codeboten (commented):
I'll try to pick the investigation next week. For now it's good to add these tests. 👍 

The missing space caused lint check to fail.

codeboten (approved):

PR #4886, opened 2022-02-18
[processor/memorylimiter] Only drops traces, not logs or metrics

dloucasfx:
Signed-off-by: Dani Louca <dlouca@splunk.com>

**Description:** 

This change https://github.com/open-telemetry/opentelemetry-collector/commit/062e64f1d7466400cbc5b0bd727ec94175a0e0de caused the memory limiter to "only" start the `checkMemLimits` routine for the ml instance used by the traces processor . 
In other words, metrics and logs processor will NOT [drop/refuse](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/memorylimiterprocessor/memorylimiter.go#L206) data and will pass them down to the next consumer regardless of the current memory pressure as their instance of ml->forcingDrop will not be set.

The simplest solution, is to call start for each processor (metrics, logs, traces) , but this will not be efficient as we'll be running 3 instances of `checkMemLimits`, ie: multiple GC .
But at the same we need to allow multiple instances, with different configs, example: `memory_limiter/another` and `memory_limiter`

````
extensions:
  memory_ballast:
    size_mib: 4

receivers:
  otlp:
    protocols:
      grpc:
      http:
processors:
  memory_limiter:
    check_interval: 2s
    limit_mib: 10

  memory_limiter/another:
    check_interval: 1s
    limit_mib: 100
exporters:
  logging:
    logLevel: info

service:
  telemetry:
    logs:
      level: "info"
  pipelines:
    metrics:
      receivers: [otlp]
      processors: [memory_limiter]
      exporters: [logging]
    metrics/default:
      receivers: [otlp]
      processors: [memory_limiter]
      exporters: [logging]
    traces:
      receivers: [otlp]
      processors: [memory_limiter/another]
      exporters: [logging]
  extensions: [memory_ballast]
````

The fix adds a global map to keep track of the different instance and add ~~sync once~~ mutex for the start and shutdown call, so only the first processor can launch the `checkMemLimits` routine and the last one to call `shutdown` to take it down.
If shutdown was called and no `checkMemLimits` has started, then we'll return an error message; unit tests were updated to handle this.

**Testing:** 
Tested with above config and using splunk otel instance with valid data.
Made sure only a single `checkMemLimits` is running when there is a single config for memory-limiter and more than one when we have multiple.
I also verified that under memory pressure, when we pass the soft limit, all data types, traces, logs and metrics are getting dropped.

One we agree on this solution, I will look into adding more unit test to validate the change

### DIFF
--- a/processor/memorylimiterprocessor/factory.go
+++ b/processor/memorylimiterprocessor/factory.go
@@ -16,6 +16,7 @@ package memorylimiterprocessor // import "go.opentelemetry.io/collector/processo
 
 import (
 	"context"
+	"sync"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
@@ -30,14 +31,24 @@ const (
 
 var processorCapabilities = consumer.Capabilities{MutatesData: false}
 
+type factory struct {
+	// memoryLimiters stores memoryLimiter instances with unique configs that multiple processors can reuse.
+	// This avoids running multiple memory checks (ie: GC) for every processor using the same processor config.
+	memoryLimiters map[config.ComponentID]*memoryLimiter
+	lock           sync.Mutex
+}
+
 // NewFactory returns a new factory for the Memory Limiter processor.
 func NewFactory() component.ProcessorFactory {
+	f := &factory{
+		memoryLimiters: map[config.ComponentID]*memoryLimiter{},
+	}
 	return component.NewProcessorFactory(
 		typeStr,
 		createDefaultConfig,
-		component.WithTracesProcessor(createTracesProcessor),
-		component.WithMetricsProcessor(createMetricsProcessor),
-		component.WithLogsProcessor(createLogsProcessor))
+		component.WithTracesProcessor(f.createTracesProcessor),
+		component.WithMetricsProcessor(f.createMetricsProcessor),
+		component.WithLogsProcessor(f.createLogsProcessor))
 }
 
 // CreateDefaultConfig creates the default configuration for processor. Notice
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: processor/memorylimiterprocessor/memorylimiter.go [hunk 1]
```diff
@@ -149,12 +153,16 @@ func (ml *memoryLimiter) start(_ context.Context, host component.Host) error {
...
+	ml.startOnce.Do(func() {
+		ml.startMonitoring()
+	})
 	return nil
 }
 
 func (ml *memoryLimiter) shutdown(context.Context) error {
-	ml.ticker.Stop()
+	ml.shutDownOnce.Do(func() {
+		ml.ticker.Stop()
```
rmfitzpatrick, line 164: If one processor is used in multiple pipelines isn't there a chance this could be preemptive and halt functionality for remainders that have yet to shutdown? We probably want to keep an atomic or waitgroup for the last instance to call.

dloucasfx: I was not aware that we can reload the config dynamically to result of multiple start/shutdown; other than a unit test, how can I trigger this as a user?

@bogdandrutu thanks for pointing out the `SharedComponent` it did the trick to get rid of the sync.once for the start/shutdown, however, if I tried to the get rid of the `getMemoryLimiter` factory logic and fully count on `GetOrAdd` then it won't work well as the `SharedComponent` does not handle errors of object initialization; 
In other words, if my func looks like this... 

````
func createTracesProcessor(
	_ context.Context,
	set component.ProcessorCreateSettings,
	cfg config.Processor,
	nextConsumer consumer.Traces,
	var err error
	var ml *memoryLimiter
	mlInstance := memLimiters.GetOrAdd(cfg, func() component.Component {
		ml, err = newMemoryLimiter(set, cfg.(*Config))
		return ml
	})
	if err != nil {
		return nil, err
	}
	return processorhelper.NewTracesProcessor(
		cfg,
		nextConsumer,
		mlInstance.Unwrap().(*memoryLimiter).processTraces,
		processorhelper.WithCapabilities(processorCapabilities),
		processorhelper.WithStart(mlInstance.Start),
		processorhelper.WithShutdown(mlInstance.Shutdown))

}
```
....and `newMemoryLimiter` was supplied wrong config to cause an error, this component will still be cached by `SharedComponent`  and yes, the error check outside the `SharedComponent` scope will result of returning `nil, err` but subsequent `create*Processor`, with the same config, will pick up the cached value component; in my case, causing unit tests to fail.

The only workaround I can think of is to unwrap the object and check if it's nil, but I would loose the error message.
```
	if mlInstance.Unwrap().(*memoryLimiter) == nil {
		return nil, fmt.Errorf("memory limiter object failed to get initialized")
	}
```
Also note, that this can cause some type of leaks as those objects are cached and we'll not call shutdown on them to clear them up from . This issue exists at the moment `SharedComponent` map.

I might be missing something that does not require changing the behavior of `SharedComponent` create.

dloucasfx: @rmfitzpatrick 
I tried a combination of atomic and waitgroup, but that would not be 100% thread safe as I need to sync multiple steps, so I reverted to basic mutex which works fine for this usage and performance should not take a hit as the start/shutdown should not be called frequently

@bogdandrutu 
I looked at `SharedComponent` implementation and it uses sync once per instance, so this and in addition to lack of error checking, I don't think `SharedComponent` is a good fit here. 
And as I mentioned in my previous comment, in case of an error, the component will leak, that should also be addressed

@tigrannajaryan @bogdandrutu
Is it possible to call shutdown without calling start? I am checking for that and it using it in the test, but was not sure if there are extra precautions needed

rmfitzpatrick (approved):

processor/memorylimiterprocessor/factory.go
```diff
@@ -48,57 +57,76 @@ func createDefaultConfig() config.Processor {
...
-		ml.processLogs,
+		memLimiter.processLogs,
 		processorhelper.WithCapabilities(processorCapabilities),
-		processorhelper.WithShutdown(ml.shutdown))
+		processorhelper.WithStart(memLimiter.start),
+		processorhelper.WithShutdown(memLimiter.shutdown))
+}
+
+func (f *factory) getMemoryLimiter(set component.ProcessorCreateSettings, cfg config.Processor) (*memoryLimiter, error) {
+	f.lock.Lock()
```
tigrannajaryan, line 118: What's the purpose of the lock? I don't think create*Processor functions are called concurrently anywhere. Are they?

processor/memorylimiterprocessor/memorylimiter.go
```diff
@@ -81,6 +82,9 @@ type memoryLimiter struct {
 	configMismatchedLogged bool
 
 	obsrep *obsreport.Processor
+
+	lock       sync.Mutex
```
tigrannajaryan, line 86: To help clarify what the lock is for it can be useful to do this:
```suggestion
	refCounterLock       sync.Mutex
```

processor/memorylimiterprocessor/factory.go
```diff
@@ -30,14 +31,22 @@ const (
 
 var processorCapabilities = consumer.Capabilities{MutatesData: false}
 
+type factory struct {
+	memoryLimiters map[config.ComponentID]*memoryLimiter
```
tigrannajaryan, line 35: Add comments to explain what the purpose of the map is.

processor/memorylimiterprocessor/factory.go
```diff
@@ -48,57 +57,76 @@ func createDefaultConfig() config.Processor {
...
 		nextConsumer,
-		ml.processLogs,
+		memLimiter.processLogs,
 		processorhelper.WithCapabilities(processorCapabilities),
-		processorhelper.WithShutdown(ml.shutdown))
+		processorhelper.WithStart(memLimiter.start),
+		processorhelper.WithShutdown(memLimiter.shutdown))
+}
+
+func (f *factory) getMemoryLimiter(set component.ProcessorCreateSettings, cfg config.Processor) (*memoryLimiter, error) {
```
tigrannajaryan, line 117: Add some comments to explain the purpose of this function.

tigrannajaryan (commented):
Logic looks good to me. Just some minor comments.

processor/memorylimiterprocessor/factory.go
dloucasfx: Not very familiar with the caller, I was under the impression that pipelines run in their own routine and based on another processor in contrib https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/factory.go#L50-L57 my assumption was that we need to keep the map thread safe.

You also mentioned that future work is in place to add dynamic config loading and wondered if concurrency will also be added if it's not there.
I don't think it will hurt to keep this lock in place, but if you prefer I take it out, I can do that.

processor/memorylimiterprocessor/memorylimiter.go
dloucasfx: 👍

processor/memorylimiterprocessor/factory.go
dloucasfx: 👍

processor/memorylimiterprocessor/factory.go
dloucasfx: 👍

dloucasfx:
@tigrannajaryan 
I updated the PR, the only thing left:
1) question whether a lock is needed. See my reply in the review comment
2) I just want to make sure you have seen my comment here https://github.com/open-telemetry/opentelemetry-collector/pull/4886#discussion_r812995156
mainly around 
```
Is it possible to call shutdown without calling start? I am checking for that and it using it in the test, but was not sure if there are extra precautions needed
```

Thanks!

tigrannajaryan:
>Is it possible to call shutdown without calling start? I am checking for that and it using it in the test, but was not sure if there are extra precautions needed

No, this would be a violation of the contract.

tigrannajaryan (approved):
LGTM.
I assume you are going to re-test this for the use case you had. Please let me know how it goes.

Issue #4775, opened 2022-01-31
[pdata] Inconsistent behavior of `oneof` proto field accessors


dmitryax:
`NumberDataPoint/Exemplar` accessors to `oneof value` proto field (`IntVal` and `DoubleVal`) return zero value in case of mismatch with underlying proto data type.

`Metric` accessors to `oneof data` proto field (`Gauge`, `Sum` etc) cause a panic in case of mismatch with underlying proto data type.

Should we consider making it consistent and changing the `NumberDataPoint/Exemplar.Value` accessors to panic as well (in a backward compatible manner)? It also saves us a cycle currently spent on the type comparison.

cc @bogdandrutu @codeboten

codeboten:
`IntVal`/`DoubleVal` is consistent with the implementation for Attributes, are you suggesting that should be changed as well?

> panic as well (in a backward compatible manner)?

Can you elaborate on how this would work?

dmitryax:
> IntVal/DoubleVal is consistent with the implementation for Attributes, are you suggesting that should be changed as well?

Yes, I think we should do the same for `AttributeValue`. Especially because `AttributeValue` getters return complex data types initialized with zero values and detached from the receiver argument. For example let's look at the following code which is currently valid:

```
initialMap := pdata.NewAttributeMap()
initialMap.InsertInt("key", 9)
intValue, _ := initialMap.Get("key")
mapValue := intValue.MapVal()  // I believe this should cause panic instead on returning a new `AttributeMap` detached from `initialMap`
```

And `Metric` `oneof` accessors do cause panics:
```
m := pdata.NewMetric()
m.SetDataType(pdata.MetricDataTypeGauge)
m.Sum() // Panic!
```

We could also return `NewSum()` from `m.Sum()` in case of mismatch to bring it to consistency with `AttributeValue` and `NumberDataPoint/Exemplar`, but I don't think it's a good idea. I think current `Metric` behavior is what we need adapt in other `oneof` field accessors.

> > panic as well (in a backward compatible manner)?

> Can you elaborate on how this would work?

Do you mean backward compatibility? We would need to introduce replacement methods and deprecate current ones. For example `IntVal` -> `IntValue`, `MapVal` -> `MapValue`. Maybe we can keep the later going forward or switch back during the `pdata` -> `pmetric` package restructure.

dmitryax:
I might be missing some context justifying the current behavior. @open-telemetry/collector-approvers please let me know

[...]

PR #4909, opened 2022-02-23
Change behavior of AttributeValue oneof value field accessors

dmitryax:
Make it consistent with Metric oneof data field: raise panic in case of type mismatch instead of returning a zero value.

The following methods are deprecated in favor of the new accessors:
- StringVal -> StringValue
- IntVal -> IntValue
- DoubleVal -> DoubleValue
- BoolVal -> BoolValue
- MapVal -> MapValue
- SliceVal -> SliceValue
- BytesVal -> BytesValue

Corresponding setter methods are also renamed for consistency without changing its logic:
- SetStringVal -> SetStringValue
- SetIntVal -> SetIntValue
- SetDoubleVal -> SetDoubleValue
- SetBoolVal -> SetBoolValue
- SetBytesVal -> SetBytesValue

Updates: https://github.com/open-telemetry/opentelemetry-collector/issues/4775

## Benchmarks

goos: darwin
goarch: amd64
pkg: go.opentelemetry.io/collector/model/internal/pdata
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz

### Before

```
BenchmarkAttributeValueStringDeprecatedAccessor-16               	1000000000	         0.6580 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueIntDeprecatedAccessor-16                  	1000000000	         0.6568 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueDoubleDeprecatedAccessor-16               	1000000000	         0.6514 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBoolDeprecatedAccessor-16                 	1000000000	         0.7169 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBytesDeprecatedAccessor-16                	1000000000	         0.6539 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueMapDeprecatedAccessor-16                  	487263888	         2.425 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueArrayDeprecatedAccessor-16                	508004030	         2.372 ns/op	       0 B/op	       0 allocs/op
```

### After

```
BenchmarkAttributeValueStringAccessor-16                         	1000000000	         0.6713 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueIntAccessor-16                            	1000000000	         0.6626 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueDoubleAccessor-16                         	1000000000	         0.6548 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBoolAccessor-16                           	1000000000	         0.7929 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBytesAccessor-16                          	1000000000	         0.6972 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueMapAccessor-16                            	1000000000	         0.9166 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueArrayAccessor-16                          	1000000000	         0.9209 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/internal/otlptext/databuffer_test.go
+++ b/internal/otlptext/databuffer_test.go
@@ -24,6 +24,22 @@ import (
 
 func TestNestedArraySerializesCorrectly(t *testing.T) {
 	ava := pdata.NewAttributeValueArray()
+	ava.SliceValue().AppendEmpty().SetStringValue("foo")
+	ava.SliceValue().AppendEmpty().SetIntValue(42)
+
+	ava2 := pdata.NewAttributeValueArray()
+	ava2.SliceValue().AppendEmpty().SetStringValue("bar")
+	ava2.CopyTo(ava.SliceValue().AppendEmpty())
+
+	ava.SliceValue().AppendEmpty().SetBoolValue(true)
+	ava.SliceValue().AppendEmpty().SetDoubleValue(5.5)
+
+	assert.Equal(t, 5, ava.SliceValue().Len())
+	assert.Equal(t, "[foo, 42, [bar], true, 5.5]", attributeValueToString(ava))
+}
+
+func TestNestedArraySerializesCorrectlyDeprecated(t *testing.T) {
+	ava := pdata.NewAttributeValueArray()
 	ava.SliceVal().AppendEmpty().SetStringVal("foo")
 	ava.SliceVal().AppendEmpty().SetIntVal(42)
 
@@ -40,11 +56,11 @@ func TestNestedArraySerializesCorrectly(t *testing.T) {
 
 func TestNestedMapSerializesCorrectly(t *testing.T) {
 	ava := pdata.NewAttributeValueMap()
-	av := ava.MapVal()
+	av := ava.MapValue()
 	av.Insert("foo", pdata.NewAttributeValueString("test"))
 
 	ava2 := pdata.NewAttributeValueMap()
-	av2 := ava2.MapVal()
+	av2 := ava2.MapValue()
 	av2.InsertInt("bar", 13)
 	av.Insert("zoo", ava2)
 
@@ -53,6 +69,6 @@ func TestNestedMapSerializesCorrectly(t *testing.T) {
      -> zoo: MAP({"bar":13})
 }`
 
-	assert.Equal(t, 2, ava.MapVal().Len())
+	assert.Equal(t, 2, ava.MapValue().Len())
 	assert.Equal(t, expected, attributeValueToString(ava))
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@
 - Rename `pdata.AttributeMap.Delete` to `pdata.AttributeMap.Remove` (#4914)
 - pdata: deprecate funcs working with InternalRep (#4957)
 
+### 🚩 Deprecations 🚩
+
+- Change behavior of pdata.AttributeValue oneof value field accessors (#4909)
+  - Deprecated `AttributeValue.StringVal` in favor of `AttributeValue.StringValue`
+  - Deprecated `AttributeValue.IntVal` in favor of `AttributeValue.IntValue`
+  - Deprecated `AttributeValue.DoubleVal` in favor of `AttributeValue.DoubleValue`
+  - Deprecated `AttributeValue.BoolVal` in favor of `AttributeValue.BoolValue`
+  - Deprecated `AttributeValue.MapVal` in favor of `AttributeValue.MapValue`
+  - Deprecated `AttributeValue.SliceVal` in favor of `AttributeValue.SliceValue`
+  - Deprecated `AttributeValue.BytesVal` in favor of `AttributeValue.BytesValue`
+  - Deprecated `AttributeValue.SetStringVal` in favor of `AttributeValue.SetStringValue`
+  - Deprecated `AttributeValue.SetIntVal` in favor of `AttributeValue.SetIntValue`
+  - Deprecated `AttributeValue.SetDoubleVal` in favor of `AttributeValue.SetDoubleValue`
+  - Deprecated `AttributeValue.SetBoolVal` in favor of `AttributeValue.SetBoolValue`
+  - Deprecated `AttributeValue.SetBytesVal` in favor of `AttributeValue.SetBytesValue`
+
 ### 💡 Enhancements 💡
 
 - Add `pdata.AttributeMap.RemoveIf`, which is a more performant way to remove multiple keys (#4914)
--- a/internal/otlptext/databuffer.go
+++ b/internal/otlptext/databuffer.go
@@ -256,17 +256,17 @@ func (b *dataBuffer) logLinks(description string, sl pdata.SpanLinkSlice) {
 func attributeValueToString(av pdata.AttributeValue) string {
 	switch av.Type() {
 	case pdata.AttributeValueTypeString:
-		return av.StringVal()
+		return av.StringValue()
 	case pdata.AttributeValueTypeBool:
-		return strconv.FormatBool(av.BoolVal())
+		return strconv.FormatBool(av.BoolValue())
 	case pdata.AttributeValueTypeDouble:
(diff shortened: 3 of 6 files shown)


bogdandrutu:
@dmitryax if we break this class anyway. Should we do the rename where the funcs have the new behavior, and the deprecated (old object) keeps the old behavior and old name for the funcs?

dmitryax:
> The idea being the new struct AnyValue (based on the suggestion) should have only the new getters.

Makes sense. I'll combine other renames related to AttributeValue within this change

internal/otlptext/databuffer_test.go
```diff
@@ -24,27 +24,27 @@ import (
 
 func TestNestedArraySerializesCorrectly(t *testing.T) {
```
jpkrohling, line 25: I'm OK with the changes from this PR, but I feel like it could also have a "proof" that this is backwards compatible. Like, by keeping this test using the old functions, changing it only when the old ones are removed. Or duplicating this test, with the one using the old functions being removed when the functions are removed.

jpkrohling (approved):

internal/otlptext/databuffer_test.go
dmitryax: Copied the test using old funcs with `Deprecated` suffix. I did the same for this test already https://github.com/open-telemetry/opentelemetry-collector/pull/4909/files#diff-5e341afb2ee55aae2edd3b7b3b3d714dc5d2233ec814131a720f74b02de8754fR28

dmitryax:
> @dmitryax if we break this class anyway. Should we do the rename where the funcs have the new behavior, and the deprecated (old object) keeps the old behavior and old name for the funcs?

@bogdandrutu after giving it a second thought I would rather keep this PR as is and will change `AttributeValue` to `AnyValue` in the next PR, because mixing method renames with the struct name change makes it more complicated to follow IMO

codeboten (approved):

Aneurysm9:
Should we return an error instead of `panic()`ing?  I'm not sure it is a good idea to cause an application to crash due to a mismatched expectation that can be handled gracefully.

dmitryax:
> Should we return an error instead of panic()ing? I'm not sure it is a good idea to cause an application to crash due to a mismatched expectation that can be handled gracefully.

We do allow panic in other places of pdata already. It's always expected that the type is checked before accessing the data. This is how it's being used almost everywhere in contrib (I'm updating couple of places where it's missed).

If we return error, 
1. It'll complicate the clients. As I mentioned, we check the type everywhere before accessing the data, so it'll leave all the clients omitting errors like `val, _ := ...Value()` making the code look unpleasant. Or we would have to add redundant error checks which will be never be executed given the type is validated before the call.
2. It'll bring back the logic that we are replacing in this PR with all the proto generated conditions + returning an additional error. I think it'll make it even less performant than before.

Aneurysm9:
> We do allow panic in other places of pdata already. It's always expected that the type is checked before accessing the data. This is how it's being used almost everywhere in contrib (I'm updating couple of places where it's missed).

I'm not sure that "we do <undesirable thing> elsewhere" is a solid argument for doing it in new places.  Indeed, I'm not sure I've seen any reason at all offered for why we want to do this potentially very disruptive change.

> If we return error,
>
>    It'll complicate the clients. As I mentioned, we check the type everywhere before accessing the data, so it'll leave all the clients omitting errors like val, _ := ...Value() making the code look unpleasant. Or we would have to add redundant error checks which will be never be executed given the type is validated before the call.

We apparently don't type check everywhere since you've opened at least one issue to add missing type checking.  Indeed, I'm not sure that we type check in most places or even anywhere.  A cursory search for `StringVal()` in the contrib returns dozens of hits where attribute string values are accessed unchecked, or only checked by looking for an empty string after access.  Each of these places would become a landmine waiting to explode a running process with the attendant potential for losing data.

As for whether it will complicate clients, I don't agree.  Clients either have to check the attribute type first or they have  to check for an error or zero value after access.  In any scenario there is checking that needs to happen, but only in one scenario does failing to check every single time result in a dead process and lost data.

>    It'll bring back all the proto generated conditions logic that we are replacing in this PR + an additional error. I think it'll make it even less performant than before.

Do you have benchmarks for this?  This is a significant change that should be made after testing and proving the benefits, not on instinct.

dmitryax:
> Do you have benchmarks for this? This is a significant change that should be made after testing and proving the benefits, not on instinct.

Makes sense. I will provide benchmarks for the 3 cases so we can discuss the API based on that:
1. current implementation
2. this PR
3. value + error

dmitryax:
@Aneurysm9 I tried the approach that you suggested here https://github.com/open-telemetry/opentelemetry-collector/pull/4987

The benchmarks are the following:

**Before**

```
BenchmarkAttributeValueStringDeprecatedAccessor-16               	1000000000	         0.6580 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueIntDeprecatedAccessor-16                  	1000000000	         0.6568 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueDoubleDeprecatedAccessor-16               	1000000000	         0.6514 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBoolDeprecatedAccessor-16                 	1000000000	         0.7169 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBytesDeprecatedAccessor-16                	1000000000	         0.6539 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueMapDeprecatedAccessor-16                  	487263888	         2.425 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueArrayDeprecatedAccessor-16                	508004030	         2.372 ns/op	       0 B/op	       0 allocs/op
```

**With panic (This PR)**

```
BenchmarkAttributeValueStringAccessor-16                         	1000000000	         0.6713 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueIntAccessor-16                            	1000000000	         0.6626 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueDoubleAccessor-16                         	1000000000	         0.6548 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBoolAccessor-16                           	1000000000	         0.7929 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueBytesAccessor-16                          	1000000000	         0.6972 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueMapAccessor-16                            	1000000000	         0.9166 ns/op	       0 B/op	       0 allocs/op
BenchmarkAttributeValueArrayAccessor-16                          	1000000000	         0.9209 ns/op	       0 B/op	       0 allocs/op
```

**With error (https://github.com/open-telemetry/opentelemetry-collector/pull/4987)**

```
BenchmarkAttributeValueStringAccessor-16                            353806957            3.464 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueIntAccessor-16                               346630108            3.479 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueDoubleAccessor-16                            356147574            3.358 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueBoolAccessor-16                              350702415            3.472 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueBytesAccessor-16                             331781389            3.677 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueMapAccessor-16                               341309820            3.559 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueArrayAccessor-16                             346380688            3.493 ns/op           0 B/op          0 allocs/op
```

bogdandrutu:
@dmitryax I think this needs to be closed based on the decision in https://github.com/open-telemetry/opentelemetry-collector/issues/4775?

dmitryax:
Yes, closing this one

Issue #4694, opened 2022-01-18
Replace global configtelemetry.UseOpenTelemetryForInternalMetrics with a featuregate


bogdandrutu:
The definition of the variable is [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtelemetry/configtelemetry.go#L39).

This was intended to be one of the first usage of the `featuregate` package, but we did not finish that.

shree007:
@bogdandrutu I may work on it

bogdandrutu:
you "may" or you "will" :) ?

shree007:
> you "may" or you "will" :) ?

I mean please assign to me, I will work on it.

splunkericl:
Hey @shree007 , have you made any progress on this? If not we will be willing to take over the work as we run into the same issue.

shree007:
Unfortunetly I could not look into it. You may proceed.

PR #4912, opened 2022-02-23
Add feature flag to allow enabling otel for internal metrics

splunkericl:
**Description:**
Previously, the variable to gate using open telemetry metrics was a constant and always set to false. This prevents any otel collector users to enable the feature.

This change will now use feature gate flag instead. The user can register the feature flag in their code base and enable the feature.

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector/issues/4694

**Testing:**
Added unit test to run otel collector with the flag enabled. Verify and assert the metrics endpoint.

**Documentation:**
feature gate ID variable documentation was added.


service/telemetry.go
```diff
@@ -50,6 +51,10 @@ const AddCollectorVersionTag = true
 const (
 	zapKeyTelemetryAddress = "address"
 	zapKeyTelemetryLevel   = "level"
+
+	// UseOtelForInternalMetricsfeatureGateID is the feature gate ID that controls whether the collector uses open
+	// telemetry for internal metrics.
+	UseOtelForInternalMetricsfeatureGateID = "telemetry.useOtelForInternalMetrics"
```
bogdandrutu, line 57: Does this need to be public?

bogdandrutu (commented):
Please add a changelog entry

config/configtelemetry/configtelemetry.go
```diff
@@ -36,8 +36,6 @@ const (
 	levelDetailedStr = "detailed"
 )
 
-const UseOpenTelemetryForInternalMetrics = false
```
bogdandrutu, line 39: We should deprecate this for the moment and make it one of the option for one release to enable this.

service/telemetry.go
splunkericl: Ideally yes. If the user wants to enable the feature they can use the variable directly instead of having constant of their own.

config/configtelemetry/configtelemetry.go
splunkericl: As in keep the variable so it is not a breaking change for the release?

service/telemetry.go
bogdandrutu: The intention is that it is enabled by the flag not code. that is how we envision the use of "featureflags",

config/configtelemetry/configtelemetry.go
bogdandrutu: yes, deprecate and keep if for 1 release, and next release we remove it.

service/telemetry.go
splunkericl: ok will keep it as private

service/telemetry.go
```diff
@@ -50,6 +51,10 @@ const AddCollectorVersionTag = true
 const (
 	zapKeyTelemetryAddress = "address"
 	zapKeyTelemetryLevel   = "level"
+
+	// useOtelForInternalMetricsfeatureGateID is the feature gate ID that controls whether the collector uses open
+	// telemetry for internal metrics.
+	useOtelForInternalMetricsfeatureGateID = "telemetry.useOtelForInternalMetrics"
```
bogdandrutu, line 57: not sure if camel case is the right choice here. @dashpole what is the k8s recommendation here?

service/telemetry.go
```diff
@@ -98,7 +103,7 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
 	instanceID := instanceUUID.String()
 
 	var pe http.Handler
-	if configtelemetry.UseOpenTelemetryForInternalMetrics {
+	if featuregate.IsEnabled(useOtelForInternalMetricsfeatureGateID) {
```
Aneurysm9, line 106: Where is this gate registered?  I don't see it anywhere.  If there is a pre-existing gate in this package should it's `ID ` field be used directly here to ensure consistency?  See, [this example of registration](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/metrics_receiver_test.go#L38-L39) and [use](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/factory.go#L89).

service/telemetry.go
splunkericl: Looking at the [k8s feature gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) it seems like they follow the camel case convention. Seems like this is also the [suggestion](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/featuregate/README.md) of the original author for the feature gate in the repo also.

config/configtelemetry/configtelemetry.go
Aneurysm9: Should this be used as the default value of `enabled` in the feature gate registration to ensure that changes to its value are still honored in the absence of an explicit feature gate CLI flag while it remains?

service/telemetry.go
splunkericl: If I am understanding correctly of how feature gate should be used, the user should register the feature gate in their `init` function and set `Enabled` to `true` for this feature. 

I don't think I am understanding the example of registration you linked. How are those translated to feature flag?

config/configtelemetry/configtelemetry.go
splunkericl: yes we can.

service/telemetry.go
dashpole: camel-case is what other feature flags use, so we should go with that.  Kubernetes doesn't use `.` to categorize, but we already do that elsewhere (e.g. `receiver.otlp.newFeatureToAdd`).  This looks correct to me.

service/telemetry.go
Aneurysm9: My apologies, I appear to have pasted the wrong link to the registration example.  It should be [this instead](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/config.go#L69-L77).

The code with the feature to be controlled by a feature gate is responsible for creating and registering the gate with the feature gate registry in an `init()` function.  The registry can then be manipulated by the user providing CLI flags with instructions to enable or disable a gated feature.  Finally, at runtime, the code with the feature to be controlled inquires whether the feature is enabled using the `featuregate.IsEnabled()` function.

In the present case, there should be something like this in `service/telemetry.go`:

```go
var useOtelForInternalMetrics = featuregate.Gate{
  ID: "telemetry.useOtelForInternalMetrics",
  Enabled: configtelemetry.UseOpenTelemetryForInternalMetrics,
  Description: "Controls whether to use the OTel SDK for internal metrics",
}

func init() {
  featuregate.Register(useOtelForInternalMetrics)
}
```

service/collector_test.go
```diff
@@ -135,6 +136,19 @@ func TestCollector_Start(t *testing.T) {
 	}, time.Second*2, time.Millisecond*200)
 }
 
+func TestCollector_Start(t *testing.T) {
+	testCollectorStartHelper(t)
+}
+
+func TestCollector_StartWithOtelInternalMetrics(t *testing.T) {
+	featuregate.Register(featuregate.Gate{
```
dashpole, line 144: note that setting a feature-gate during a unit test affects all other unit tests as well.  If you want to set it during a test only, you need to set it back at the end.

@Aneurysm9 I've added a test util for my feature gate here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7177/files#diff-a4da287cd470d6d0ff4cdcd03a3d5cf4a5ccb76c9dea73d4cc961bde6819999fR34 for setting/resetting it during tests.  I could make that more generic and add it to the featuregate package if you think that would be helpful for others.

service/telemetry.go
splunkericl: ok I see. I had a different understanding of where registration should happen but i will add it

service/collector_test.go
Aneurysm9: That could be useful.  To this point I believe that all of the feature gates we've added have been well localized and included in the component configuration as unexported fields with the `featuregate.IsEnabled()` call in `createDefaultConfig()`.  This meant that testing could happen without involving the feature gate registry by simply providing different configuration.  Your use is a bit more invasive, though, so I see where that approach makes more sense.  I'm not sure which style will end up being more common in the end.

service/telemetry.go
bogdandrutu: @splunkericl @dashpole thanks, looks good.

service/telemetry.go
```diff
@@ -50,8 +51,20 @@ const AddCollectorVersionTag = true
...
+	// useOtelForInternalMetricsfeatureGateID is the feature gate ID that controls whether the collector uses open
+	// telemetry for internal metrics.
+	useOtelForInternalMetricsfeatureGateID = "telemetry.useOtelForInternalMetrics"
 )
 
+func init() {
+	featuregate.Register(featuregate.Gate{
+		ID:          useOtelForInternalMetricsfeatureGateID,
+		Description: "controls whether the collector to uses open telemetry for internal metrics",
+		Enabled:     configtelemetry.UseOpenTelemetryForInternalMetrics,
```
dashpole, line 64: We shouldn't reverence `UseOpenTelemetryForInternalMetrics` if it is deprecated.

Aneurysm9: I think this is appropriate in this case.  The feature flag is replacing compile-time variable setting of this variable as the control mechanism for this feature.  Using it here as the default value allows for users who are still setting it at compile time to not yet have a functional break.  It is similar to deprecating a function and implementing a replacement in terms of the deprecated function.

config/configtelemetry/configtelemetry.go
```diff
@@ -36,6 +36,8 @@ const (
 	levelDetailedStr = "detailed"
 )
 
+// Deprecated: UseOpenTelemetryForInternalMetrics has been deprecated. Uses feature flag in
+// service.useOtelForInternalMetricsfeatureGateID to handle this feature
```
bogdandrutu, line 40: the "feature flag" is `telemetry.useOtelForInternalMetrics` not the internal variable.

service/collector_test.go
bogdandrutu: @dashpole here is what I was pushing to do https://github.com/open-telemetry/opentelemetry-collector/pull/4108#discussion_r726260681 to enable easier testing. You can pass then a "testing registry" to not have to deal with globals. Global state also does not allow parallel testing.

service/telemetry.go
dashpole: Am I understanding correctly that "setting it at compile time" means "carrying a patch to the collector", or is there a way to override constants I'm not aware of? I don't think we should consider constants to be part of our config surface for stability purposes. If we explicitly told users to enable this functionality via patches, then I guess its alright for this case, but i'd prefer if we didn't do this going forward.  In general, deprecated code should reference the new code, not the other way around.

bogdandrutu: it is with `-ldflags`, something like:

`-ldflags go.opentelemetry.io/collector/service.UseOpenTelemetryForInternalMetrics=true`

splunkericl: The constant can't be override. But I was more thinking this is to keep the same `Enabled` state between the constant and the new feature flag. What if another person changes the `Enabled` state to `true` before we can remove the constant. Then there will be inconsistency.

It's not great to back reference deprecated variable but this will be removed after a release.

config/configtelemetry/configtelemetry.go
splunkericl: updated

service/collector_test.go
splunkericl: I have set and unset the flag for this specific test for now. If a more generic approach is needed to apply for testing with gate lets do it in a separate issue.

bogdandrutu (approved):

service/telemetry.go
```diff
@@ -50,8 +51,20 @@ const AddCollectorVersionTag = true
...
+
+	// useOtelForInternalMetricsfeatureGateID is the feature gate ID that controls whether the collector uses open
+	// telemetry for internal metrics.
+	useOtelForInternalMetricsfeatureGateID = "telemetry.useOtelForInternalMetrics"
 )
 
+func init() {
+	featuregate.Register(featuregate.Gate{
+		ID:          useOtelForInternalMetricsfeatureGateID,
+		Description: "controls whether the collector to uses open telemetry for internal metrics",
```
Aneurysm9, line 63:
```suggestion
		Description: "controls whether the collector uses OpenTelemetry for internal metrics",
```

Aneurysm9 (approved):

service/telemetry.go
dashpole: TIL about `ldflags` :).  Feel free to resolve this thread.

dashpole (approved):

codeboten (approved):

codeboten (commented):
@splunkericl please address build failure issues

splunkericl:
@codeboten I have made  a couple changes to fix lint, import order and unit test. can you run the approval workflow again?

splunkericl:
@ dmitryax can you please take a look on the change?

Issue #4756, opened 2022-01-27
Efficient way of removing multiple keys from AttributeMap


anuraaga:
Currently `AttributeMap` only has `Delete`, which will cause an array truncation every time it's called (I think). It would be good if we can only have such an operation happen once when deleting multiple keys. In Java, we added `removeIf` which accepts a function executed against all keys, perhaps something like that could be implemented here.

bogdandrutu:
We also have "RemoveIf" for slices (see one example https://pkg.go.dev/go.opentelemetry.io/collector/model@v0.43.1/pdata#AttributeValueSlice.RemoveIf) with same signature so we can consider that.

bogdandrutu:
@dashpole you can take this one if you want.

PR #4914, opened 2022-02-23
Add RemoveIf to attribute map, and rename Delete to Remove

dashpole:
**Description:**

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4756 by adding RemoveIf for AttributeMap.  The implementation (and TODO) are based on the RemoveIf for AttributeValueSlice: https://github.com/open-telemetry/opentelemetry-collector/blob/model/v0.43.1/model/pdata/generated_common.go#L181

**Testing:**
Unit tests added.

Benchmark added to compare performance of Remove to RemoveIf when deleting a large number of elements.

```
$ go test -bench=.
goos: linux
goarch: amd64
pkg: go.opentelemetry.io/collector/model/pdata
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
...
BenchmarkAttributeMap_Remove-16                                  	   94230	     12897 ns/op
BenchmarkAttributeMap_RemoveIf-16                                	  374329	      3345 ns/op
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,13 @@
 ### 🛑 Breaking changes 🛑
 
 - Remove `Type` funcs in pdata (#4933)
+- Rename `pdata.AttributeMap.Delete` to `pdata.AttributeMap.Remove` (#4914)
 
-## 🧰 Bug fixes 🧰
+### 💡 Enhancements 💡
+
+- Add `pdata.AttributeMap.RemoveIf`, which is a more performant way to remove multiple keys (#4914)
+
+### 🧰 Bug fixes 🧰
 
 - Collector `Run` will now exit when a context cancels (#4954)
 
--- a/model/internal/pdata/common.go
+++ b/model/internal/pdata/common.go
@@ -513,7 +513,7 @@ func (am AttributeMap) Get(key string) (AttributeValue, bool) {
 
 // Delete deletes the entry associated with the key and returns true if the key
 // was present in the map, otherwise returns false.
-// Deprecated: Use Remove instead.
+// Deprecated: [v0.46.0] Use Remove instead.
 func (am AttributeMap) Delete(key string) bool {
 	return am.Remove(key)
 }
--- a/model/internal/pdata/common.go
+++ b/model/internal/pdata/common.go
@@ -525,6 +525,26 @@ func (am AttributeMap) Delete(key string) bool {
 	return false
 }
 
+// DeleteIf deletes the entries for which the function in question returns true
+func (am AttributeMap) DeleteIf(f func(string, AttributeValue) bool) {
+	newLen := 0
+	for i := 0; i < len(*am.orig); i++ {
+		akv := &(*am.orig)[i]
+		if f(akv.Key, AttributeValue{&akv.Value}) {
+			continue
+		}
+		if newLen == i {
+			// Nothing to move, element is at the right place.
+			newLen++
+			continue
+		}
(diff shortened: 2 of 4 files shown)


model/pdata/common.go
```diff
@@ -525,6 +525,26 @@ func (am AttributeMap) Delete(key string) bool {
 	return false
 }
 
+// Delete deletes the entry for which the function in question
+func (am AttributeMap) DeleteIf(f func(string, AttributeValue) bool) {
```
bogdandrutu, line 529: For consistency, rename this to "RemoveIf" since that is what we use in slices? It is just a suggestion, opposite argument can be made since we call "Delete" not "Remove" the other API. Maybe change of fixing that as well.

dashpole: Changed DeleteIf -> RemoveIf and Delete -> Remove.  I added a deprecation warning to Delete, and made it call Remove

model/pdata/common.go
```diff
@@ -513,7 +513,14 @@ func (am AttributeMap) Get(key string) (AttributeValue, bool) {
 
 // Delete deletes the entry associated with the key and returns true if the key
 // was present in the map, otherwise returns false.
+// Deprecated: Use Remove instead.
```
bogdandrutu, line 516: Add the version to the deprecated statement

dashpole: done

bogdandrutu (approved):
LGTM, @codeboten @tigrannajaryan since this changes a bit more specifically `Delete` -> `Remove`, I want your opinion.

model/pdata/common.go
```diff
@@ -525,6 +532,26 @@ func (am AttributeMap) Delete(key string) bool {
...
+		}
+		if newLen == i {
+			// Nothing to move, element is at the right place.
+			newLen++
+			continue
+		}
+		(*am.orig)[newLen] = (*am.orig)[i]
+		newLen++
+	}
+	// TODO: Prevent memory leak by erasing truncated values.
```
tigrannajaryan, line 551: Just curious: is it necessary to erase the individual elements? Is it not sufficient that the slice length is reduced? Doesn't GC understand that?

tigrannajaryan (approved):
No strong opinion on Delete vs Remove name. Works either way for me.

tigrannajaryan:
> No strong opinion on Delete vs Remove name. Works either way for me.

Correction: Remove/RemoveIf is preferable to be consistent with slices that have RemoveIf.

dashpole:
Leaving the comment here, since it will be easier to read than in a thread...
I think GC might actually be handling this correctly, and not leaking memory.  My experimental code: https://github.com/dashpole/dashpole_demos/commit/3a2c7686711e625a512b57d8cd9f3063e9cc4638#diff-f48531acb0ca1af513103682c6d325f99a4397db9023fa8a3bad4f4700e867a2R9

Memory usage over time:
Screen Shot 2022-02-28 at 3 15 09 PM

model/pdata/common.go
dashpole: I think you might be Right.  See my comment below

tigrannajaryan:
> Leaving the comment here, since it will be easier to read than in a thread... I think GC might actually be handling this correctly, and not leaking memory. 

Thanks for confirming!

model/pdata/common.go
dashpole: removed the TODO

jpkrohling (approved):

PR #4918, opened 2022-02-25
Split pdata package by telemetry signal type

dmitryax:
This is the next change after https://github.com/open-telemetry/opentelemetry-collector/pull/4833 following the approach outlined as Option 3 in [this issue](https://github.com/open-telemetry/opentelemetry-collector/issues/4832). It adds new aliases for pdata split into different packages by signal type. The following package structure is proposed:

```
/model
  /pdata (aliases to /internal/pdata)
    /traces (additional aliases to traces parts of internal/pdata)
    /logs (additional aliases to logs parts of internal/pdata)
    /metrics (additional aliases to metrics parts of internal/pdata)
```

Next steps:

1. Remove traces, logs and metrics parts from /model/pdata.
2. Move traces, logs and metrics implementation from /model/internal/pdata to /model/pdata/\<signals\> to replace the aliases - not a breaking change.

Once all the steps are done the following package structure will be achieved:

```
/model
  /pdata (aliases to /internal/pdata where only common parts are left)
    /traces (actual implementation of traces related pdata parts)
    /logs (actual implementation of logs related pdata parts)
    /metrics (actual implementation of metrics related pdata parts)
```


bogdandrutu (commented):
LGTM this is how I would envisioned to continue that PR.

tigrannajaryan:
LGTM to me as well. Only small question: do we want common aliases to be in `model/pdata` or `model/pdata/common`?

dmitryax:
> Only small question: do we want common aliases to be in model/pdata or model/pdata/common?

This is a good question.

`model/pdata` pros:

- I think names look more descriptive, e.g. pdata.InstrumentationLibrary vs common.InstrumentationLibrary.
- No need to introduce another breaking change.

`model/pdata/common` pros:
- Maybe more cleaner packages/modules structure when all the signals+common on the same level.

I'd prefer `model/pdata`, but not firm on this.

tigrannajaryan:
> I'd prefer `model/pdata`, but not firm on this.

Works for me.

model/pdata/traces/alias_trace.go
```diff
@@ -0,0 +1,62 @@
...
+
+// This file contains aliases for traces data structures.
+
+import "go.opentelemetry.io/collector/model/internal/pdata"
+
+// Marshaler is an alias for pdata.TracesMarshaler interface.
+type Marshaler = pdata.TracesMarshaler
+
+// Unmarshaler is an alias for pdata.TracesUnmarshaler interface.
+type Unmarshaler = pdata.TracesUnmarshaler
```
bogdandrutu, line 25: If you do this kind of changes, you also need to change the funcs in the interface to not include "Traces". I think for that reason there was another proposal to use `ptraces` as package and keep the names exactly how they were.

dmitryax: Do you mean to change `UnmarshalLogs` func in pdata.<strike>Traces</strike>Unmarshaler to `Unmarshal`? In that case how would unmarshalers implement all the functions to unmarshal logs, traces and metrics if they the same name?

bogdandrutu: Do we need that?

dmitryax: > Do we need that?

What do you mean?

bogdandrutu: >  In that case how would unmarshalers implement all the functions to unmarshal logs, traces and metrics if they the same name?

Do we need this capability?

Aneurysm9: I don't think it would be necessary to change `UnmarshalTraces` to `Unmarshal` as there would generally not be references to such implementing functions that start with `traces` and, if even if there were, they would not stutter due to the interposition of `Unmarshal` before `Traces`.  Indeed, the stdlib follows this practice with, e.g., [`json.Marshaler`](https://pkg.go.dev/encoding/json#Marshaler) having an `UnmarshalJSON` method.  As @dmitryax points out, this is the only way to support unmarshaling multiple types of values from a single implementation.

dmitryax: > Do we need this capability?

Looking through the code, I don't see where we actually need the combined (un)marshalers. Probably we can just split  them. For example [here](https://github.com/open-telemetry/opentelemetry-collector/blob/9f2ae4ab8187d81475bcbbaa4d175c27d9bce9cd/model/otlp/pb_unmarshaler.go#L25) `pbUnmarshaler` becomes `pbLogsUnmarshaler` etc. @Aneurysm9 @bogdandrutu WDYT?

Aneurysm9:
> If we are going to split other components, they can be changed along with switching to the new \<signals\> packages.
> If we don't split them, we will need to run all the components relying on pdata through the 3-step deprecation process.

Can you elaborate on this?  I'm not sure I follow.

Aneurysm9:
If we choose this route, should we jump straight to this PR and skip #4833?  I'm not seeing the benefit of it as an intermediate step.  Is there something I'm missing there?

model/pdata/traces/alias_trace.go
Aneurysm9: I think that works for our uses, but I would still want to see the methods in the interfaces be `Unmarshal<Signal>`.  This provides flexibility for components that may support multiple signals and use these interfaces.  The [`awskinesisexporter` encoding system](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/awskinesisexporter/internal/batch/encode.go#L48-L50) has an example of such a use.

dmitryax: Makes sense. Then we probably don't need to split it

dmitryax:
>> If we are going to split other components, they can be changed along with switching to the new <signals> packages.
>> If we don't split them, we will need to run all the components relying on pdata through the 3-step deprecation process.

> Can you elaborate on this? I'm not sure I follow.

I tried to outline the next steps towards moving from unified pdata to the new separate packages. To do that we need to update all the clients of pdata. There is no yet decision whether we will be splitting all other packages by the signal type. So there are two possible ways:
1. Split all the client packages and update them to use new `pdata/signal` packages,
2. Otherwise we need to run the clients through the 2-step (not 2) deprecation process as outlined [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes)

dmitryax:
> If we choose this route, should we jump straight to this PR and skip https://github.com/open-telemetry/opentelemetry-collector/pull/4833? I'm not seeing the benefit of it as an intermediate step. Is there something I'm missing there?

I would prefer granular changes as they are pretty much independent steps

alolita:
For the sake of user clarity, we should support intuitive conventions 

Example -
/pdata/traces
/pdata/metrics
/pdata/logs

which is what this PR proposed and is implementing. So +1 on this PR.

bogdandrutu:
@alolita thanks for the feedback. I think the current PR I think we use different package names.

dmitryax:
@bogdandrutu 

I added a small usage example of the new pdata packages as further split of consumer package: https://github.com/open-telemetry/opentelemetry-collector/pull/5087.

I also added a PR with the alternative package structure and the same usage example: https://github.com/open-telemetry/opentelemetry-collector/pull/5087

I think I like structure from this PR more because:
- The package names are probably better aligned with names of other packages that we will be splitting going forward even there is a possible chance of import conflict.
- At least some API names can be shorten like plog.LogsMarshaler -> logs.Marshaler, plog.NewLogs = logs.New.

Let me know WDYT. I will also update more packages to show usage of both version more extensively.

bogdandrutu:
> The package names are probably better aligned with names of other packages that we will be splitting going forward even there is a possible chance of import conflict.

As mentioned in the other PRs, I think we agreed to not do this. Also if from the same repo we export conflicting packages I think we fail a bit the language pattern :)

> At least some API names can be shorten like plog.LogsMarshaler -> logs.Marshaler, plog.NewLogs = logs.New.

This is nice indeed.

dmitryax:
> Also if from the same repo we export conflicting packages I think we fail a bit the language pattern :)

I would agree on this one. Is this the only reason why you prefer https://github.com/open-telemetry/opentelemetry-collector/pull/5087 over this PR?

Aneurysm9:
> > Also if from the same repo we export conflicting packages I think we fail a bit the language pattern :)
> 
> I would agree on this one. Is this the only reason why you prefer #5087 over this PR?

I disagree.  I think it is less about whether there are multiple packages with the same package name within the repository and more about whether they are likely to be imported at the same time.  What other packages would conflict with these names?

dmitryax:
> What other packages would conflict with these names?

It depends on whether we will be splitting other packages. If we split them, it'll be at least `component` and `consumer` packages. Currently we don't have anything conflicting.

dmitryax:
As I described [here](https://github.com/open-telemetry/opentelemetry-collector/pull/5086#issuecomment-1081235460), it's not required to change any clients to split the pdata package. We can split it right away. 

The only thing we need to decide is which package structure we are going to adopt. I updated all the core's code in this PR and https://github.com/open-telemetry/opentelemetry-collector/pull/5087 to use new packages and deprecated the old one. 

There are few conflicts with existing `metrics`, `traces` and `logs` local variables in this PR. Now I'm on a fence and maybe prefer https://github.com/open-telemetry/opentelemetry-collector/pull/5087 slightly more because `plog`, `ptrace`, `pmetric` and `pcommon` are pretty unique names and won't cause any conflicts. I'm going to suggest a poll so we can vote and decide on tomorrow's SIG meeting.

cc @Aneurysm9 @bogdandrutu @tigrannajaryan

dmitryax:
Results of the poll from today's SIG meeting https://docs.google.com/document/d/1D_jK39GYkp9wAPgRoCYntnzq4LBs_YNnu64vXJGc9nI/edit#heading=h.oeev18el9qk1:
- 2 votes for this structure.
- 7 votes for the alternative structure.

So I'm closing this PR in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/5087

PR #4920, opened 2022-02-25
Remove use of unnecessary internal interface in factory

bogdandrutu:
This PR does not change the design (which is to not allow external implementation),
but simplifies the way how this is achieved, now that the helpers are in the same package.


codeboten (approved):

bogdandrutu:
I doubled checked with @codeboten that they saw latest version :)

PR #4925, opened 2022-02-28
Reword exceptions section to be more general

mx-psi:
**Description:**

Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/4903#issuecomment-1049671154.

Moves 'Exceptions' subsection to its own section and rewords to apply more generally to all rules and procedures that may be set in this repository.

The intent is to formalize how to deal with exceptions to existing rules.

The specific motivation was to address how should we determine exceptions to the Go version support policy if e.g. a new version has a feature that is deemed particularly useful or necessary, but it can apply to any other rule we come up with.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -457,14 +457,6 @@ that each of the following steps is done in a separate version:
 1. On `v0.N+2`, we change `func GetFoo() Foo` to `func GetFoo(context.Context) Foo` if desired or remove it entirely if
    needed.
 
-#### Exceptions
-
-While the above is what we strive to follow, we acknowledge that some changes might be unfeasible to achieve in a
-non-breaking manner. Exceptions to the outlined rules are acceptable if consensus can be obtained from approvers in the
-pull request they are proposed. A reason for requesting the exception MUST be given in the pull request. Until unanimity
-is obtained, approvers and maintainers are encouraged to discuss the issue at hand. If a consensus (unanimity) cannot be
-obtained, the maintainers are then tasked in getting a decision using its regular means (voting, TC help, ...).
-
 ## Updating Changelog
 
 An entry into the [Changelog](./CHANGELOG.md) is required for the following reasons:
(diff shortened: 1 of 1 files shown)


mx-psi:
cc @bogdandrutu @Aneurysm9 from previous PR, also cc @jpkrohling since you wrote the original text and may have feedback.

CONTRIBUTING.md
```diff
@@ -502,3 +494,12 @@ go: github.com/golangci/golangci-lint@v1.31.0 requires
 `go env GOPROXY` should return `https://proxy.golang.org,direct`. If it does not, set it as an environment variable:
 
 `export GOPROXY=https://proxy.golang.org,direct`
+
+## Exceptions and changes in the rules
+
+While the ruling on this and other documents in this repository is what we strive to follow, we acknowledge that rules may be 
+unfeasible to be applied in some situations or that procedures may need to change over time. Exceptions or changes to the rules
```
Aneurysm9, line 501:
```suggestion
unfeasible to be applied in some situations. Exceptions to the rules 
```

We should separate changes from exceptions.  I'd like to avoid a situation where there is a disconnect between expectations where some may expect that they are discussing an exception and others think they are discussing a permanent change.

bogdandrutu: Good feedback.

CONTRIBUTING.md
```diff
@@ -502,3 +494,12 @@ go: github.com/golangci/golangci-lint@v1.31.0 requires
 `go env GOPROXY` should return `https://proxy.golang.org,direct`. If it does not, set it as an environment variable:
 
 `export GOPROXY=https://proxy.golang.org,direct`
+
+## Exceptions and changes in the rules
```
mx-psi, line 498: Suggestion to be in line with Anthony's comment

```suggestion
## Exceptions
```

CONTRIBUTING.md
mx-psi: That's good feedback, thanks, I also changed the title to be in line with this.

CONTRIBUTING.md
```diff
@@ -502,3 +494,12 @@ go: github.com/golangci/golangci-lint@v1.31.0 requires
 `go env GOPROXY` should return `https://proxy.golang.org,direct`. If it does not, set it as an environment variable:
 
 `export GOPROXY=https://proxy.golang.org,direct`
+
+## Exceptions
+
+While the ruling on this and other documents in this repository is what we strive to follow, we acknowledge that rules may be
```
Aneurysm9, line 500:
```suggestion
While the rules in this and other documents in this repository is what we strive to follow, we acknowledge that rules may be 
```

Aneurysm9 (approved):

bogdandrutu (approved):

codeboten (approved):

PR #4933, opened 2022-02-28
Removing deprecated Type functions

codeboten:
These have been replaced with ValueType.


bogdandrutu (commented):
For the `pdata` changes, based on my experience is better to keep deprecated funcs for 2 releases.

codeboten:
> For the `pdata` changes, based on my experience is better to keep deprecated funcs for 2 releases.

Happy to keep this one opened until the release this week.

bogdandrutu (approved):

PR #4934, opened 2022-03-01
Use json for writing floats as strings

dashpole:
**Description:**

This is the change that would be required by https://github.com/open-telemetry/opentelemetry-specification/pull/2343.  It changes how small floats are represented by using the json marshaller.  The test would fail with the current implementation, as it returns ".000000009" instead of "9e-9".


model/pdata/common.go
```diff
@@ -379,7 +379,8 @@ func (a AttributeValue) AsString() string {
 		return strconv.FormatBool(a.BoolVal())
 
 	case AttributeValueTypeDouble:
-		return strconv.FormatFloat(a.DoubleVal(), 'f', -1, 64)
+		jsonStr, _ := json.Marshal(a.DoubleVal())
+		return string(jsonStr)
```
bogdandrutu, line 383: Why? Should we use what json.Marshal uses instead? The reason is I believe json.Marshal is extremely slow because of reflection etc.

dashpole: We can do that.  It would involve copying https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/encoding/json/encode.go;l=585.

dashpole: Done. Benchmark improvement:

With `json.Marshal`:

`BenchmarkAttributeValueFloat_AsString-16    	 3270664	       364.8 ns/op`

With the new implementation:

`BenchmarkAttributeValueFloat_AsString-16    	 6703143	       174.0 ns/op`

I'll let you decide if the change is worth it :).

dashpole:
This is ready for review

model/pdata/common.go
bogdandrutu: As most likely a future maintainer of the codebase what would be your preference 😁?

dashpole: I think its probably worth it.  We risk drifting from the upstream implementation, but it hasn't changed in years, so we are probably safe.  The benchmark differences are fairly large.

bogdandrutu (approved):

bogdandrutu:
@dashpole maybe rebase to re-trigger checks?

dashpole:
pushed again

bogdandrutu:
@dashpole not sure what happens,.... maybe open a new PR from a cloned branch?

dashpole:
it woke up!

dashpole:
rebased

PR #4936, opened 2022-03-01
Add a NewConfigProvider that uses options for settings, and does input validation

bogdandrutu:
This is a fork of https://github.com/open-telemetry/opentelemetry-collector/pull/4762

Also, deprecates `service.MustNewConfigProvider` and `service.MustNewDefaultConfigProvider`in favor of `service.NewConfigProvider`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,6 @@
 
 - Deprecate configmapprovider package, replace with mapconverter (#5167)
 - Deprecate `service.MustNewConfigProvider` and `service.MustNewDefaultConfigProvider`in favor of `service.NewConfigProvider` (#4762)
-  - There is no equivalent to provide custom ConfigUnmarshaler, if you need to apply changes to the parsed config wrap the default ConfigProvider.
->>>>>>> 2daa6ab9 (Change deprecated NewConfigProvider to use options and do input validation)
 
 ### 💡 Enhancements 💡
 
@@ -50,7 +48,6 @@
 
 - Move MapProvider to config, split providers in their own package (#5030)
 - API related to `pdata.AttributeValue` is deprecated in favor of `pdata.Value` (#4978)
-- API related to `pdata.AttributeValue` is deprecated in favor of `pdata.Value` (#4975)
   - `pdata.AttributeValue` struct is deprecated in favor of `pdata.Value`
   - `pdata.AttributeValueType` type is deprecated in favor of `pdata.ValueType`
   - `pdata.AttributeValueType...` constants are deprecated in favor of `pdata.ValueType...`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 ### 🚩 Deprecations 🚩
 
 - Deprecate configmapprovider package, replace with mapconverter (#5167)
+- Deprecate `service.MustNewConfigProvider` and `service.MustNewDefaultConfigProvider`in favor of `service.NewConfigProvider` (#4762)
+  - There is no equivalent to provide custom ConfigUnmarshaler, if you need to apply changes to the parsed config wrap the default ConfigProvider.
+>>>>>>> 2daa6ab9 (Change deprecated NewConfigProvider to use options and do input validation)
 
 ### 💡 Enhancements 💡
 
@@ -47,6 +50,7 @@
 
 - Move MapProvider to config, split providers in their own package (#5030)
 - API related to `pdata.AttributeValue` is deprecated in favor of `pdata.Value` (#4978)
+- API related to `pdata.AttributeValue` is deprecated in favor of `pdata.Value` (#4975)
   - `pdata.AttributeValue` struct is deprecated in favor of `pdata.Value`
   - `pdata.AttributeValueType` type is deprecated in favor of `pdata.ValueType`
   - `pdata.AttributeValueType...` constants are deprecated in favor of `pdata.ValueType...`
--- a/service/collector_test.go
+++ b/service/collector_test.go
@@ -37,7 +37,7 @@ import (
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
-	"go.opentelemetry.io/collector/config/configmapprovider"
+	"go.opentelemetry.io/collector/config/mapconverter/overwritepropertiesmapconverter"
 	"go.opentelemetry.io/collector/internal/testcomponents"
 	"go.opentelemetry.io/collector/internal/testutil"
 	"go.opentelemetry.io/collector/service/featuregate"
@@ -62,11 +62,12 @@ func TestCollector_StartAsGoRoutine(t *testing.T) {
 	factories, err := testcomponents.NewDefaultFactories()
 	require.NoError(t, err)
 
-	cfgProvider, err := NewConfigProvider(
-		[]string{
-			filepath.Join("testdata", "otelcol-config.yaml"),
-			"yaml:service::telemetry::metrics::address: localhost:" + strconv.FormatUint(uint64(testutil.GetAvailablePort(t)), 10),
-		})
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
+		filepath.Join("testdata", "otelcol-config.yaml"),
+		"yaml:service::telemetry::metrics::address: localhost:" + portAsString(testutil.GetAvailablePort(t)),
+	}
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	set := CollectorSettings{
@@ -106,13 +107,17 @@ func testCollectorStartHelper(t *testing.T) {
 	}
 
 	metricsPort := testutil.GetAvailablePort(t)
-	cfgProvider, err := NewConfigProvider(
-		[]string{filepath.Join("testdata", "otelcol-config.yaml")},
-		WithConfigMapConverters([]config.MapConverterFunc{
-			configmapprovider.NewOverwritePropertiesConverter(
-				[]string{"service.telemetry.metrics.address=localhost:" + strconv.FormatUint(uint64(metricsPort), 10)},
-			),
-		}))
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
+		filepath.Join("testdata", "otelcol-config.yaml"),
+	}
+	cfgSet.MapConverters = append([]config.MapConverterFunc{
+		overwritepropertiesmapconverter.New(
+			[]string{"service.telemetry.metrics.address=localhost:" + portAsString(metricsPort)},
+		)},
+		cfgSet.MapConverters...,
+	)
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	col, err := New(CollectorSettings{
@@ -184,7 +189,11 @@ func TestCollector_ShutdownNoop(t *testing.T) {
 	factories, err := testcomponents.NewDefaultFactories()
 	require.NoError(t, err)
 
-	cfgProvider, err := NewConfigProvider([]string{filepath.Join("testdata", "otelcol-config.yaml")})
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
+		filepath.Join("testdata", "otelcol-config.yaml"),
+	}
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	set := CollectorSettings{
@@ -208,7 +217,11 @@ func TestCollector_ShutdownBeforeRun(t *testing.T) {
 	factories, err := testcomponents.NewDefaultFactories()
 	require.NoError(t, err)
 
-	cfgProvider, err := NewConfigProvider([]string{filepath.Join("testdata", "otelcol-config.yaml")})
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
+		filepath.Join("testdata", "otelcol-config.yaml"),
+	}
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	set := CollectorSettings{
@@ -242,7 +255,11 @@ func TestCollector_ClosedStateOnStartUpError(t *testing.T) {
 	factories, err := testcomponents.NewDefaultFactories()
 	require.NoError(t, err)
 
-	cfgProvider, err := NewConfigProvider([]string{filepath.Join("testdata", "otelcol-invalid.yaml")})
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
+		filepath.Join("testdata", "otelcol-invalid.yaml"),
+	}
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	// Load a bad config causing startup to fail
@@ -280,7 +297,11 @@ func TestCollector_ReportError(t *testing.T) {
 	factories, err := testcomponents.NewDefaultFactories()
 	require.NoError(t, err)
 
-	cfgProvider, err := NewConfigProvider([]string{filepath.Join("testdata", "otelcol-config.yaml")})
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
+		filepath.Join("testdata", "otelcol-config.yaml"),
+	}
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	col, err := New(CollectorSettings{
@@ -315,10 +336,12 @@ func TestCollector_ContextCancel(t *testing.T) {
 	factories, err := testcomponents.NewDefaultFactories()
 	require.NoError(t, err)
 
-	cfgProvider, err := NewConfigProvider([]string{
+	cfgSet := newDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
 		filepath.Join("testdata", "otelcol-config.yaml"),
-		"yaml:service::telemetry::metrics::address: localhost:" + strconv.FormatUint(uint64(testutil.GetAvailablePort(t)), 10),
-	})
+		"yaml:service::telemetry::metrics::address: localhost:" + portAsString(testutil.GetAvailablePort(t)),
+	}
+	cfgProvider, err := NewConfigProvider(cfgSet)
 	require.NoError(t, err)
 
 	set := CollectorSettings{
@@ -408,4 +431,8 @@ func assertZPages(t *testing.T) {
 	for _, path := range paths {
 		testZPagePathFn(t, path)
 	}
+}
+
+func portAsString(p uint16) string {
+	return strconv.FormatUint(uint64(p), 10)
 }
--- a/service/collector_test.go
+++ b/service/collector_test.go
@@ -36,6 +36,8 @@ import (
 	"go.uber.org/zap/zapcore"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/config/configmapprovider"
 	"go.opentelemetry.io/collector/internal/testcomponents"
 	"go.opentelemetry.io/collector/internal/testutil"
 	"go.opentelemetry.io/collector/service/featuregate"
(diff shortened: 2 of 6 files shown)


service/collector.go
```diff
@@ -50,6 +50,20 @@ const (
 	Closed
 )
 
+func (s State) String() string {
```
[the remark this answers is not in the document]
bogdandrutu: Yeah, I made this PR, then I realized that I wanted to cleanup some other things :)) so I created the cleanup PRs.

service/config_provider_test.go
```diff
@@ -165,7 +165,9 @@ func TestConfigProvider_Errors(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			cfgW := MustNewConfigProvider(tt.locations, tt.parserProvider, tt.cfgMapConverters, tt.configUnmarshaler)
+			cfgW, err := NewConfigProvider(tt.locations, WithConfigMapProviders(tt.parserProvider), WithConfigMapConverters(tt.cfgMapConverters), WithConfigUnmarshaler(tt.configUnmarshaler))
```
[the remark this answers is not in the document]
bogdandrutu: We usually don't do that...

service/config_provider.go
```diff
@@ -100,44 +126,49 @@ func MustNewConfigProvider(
...
+
+	// Override default values with user options
+	for _, o := range opts {
+		o(&provider)
+	}
+
+	return &provider, nil
 }
 
+// Deprecated: use NewConfigProvider instead
```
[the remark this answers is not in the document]
bogdandrutu: done.

service/config_provider_test.go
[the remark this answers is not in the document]
mx-psi: That's fair, let's not do it then

CHANGELOG.md
```diff
@@ -121,6 +122,11 @@
 
 - `confighttp`: Allow CORS requests with configured auth (#4869)
 
+### 🚩 Deprecations 🚩
```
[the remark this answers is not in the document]
bogdandrutu: Thanks for catching this.

service/config_provider_test.go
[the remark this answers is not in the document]
jpkrohling: To be honest, I've been requesting those on the PRs I'm reviewing. I think it's important to assert that the behavior didn't change.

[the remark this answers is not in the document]
bogdandrutu: Done.

service/config_provider.go
```diff
@@ -100,44 +126,49 @@ func MustNewConfigProvider(
 // The `configMapProviders` is a map of pairs <scheme,Provider>.
```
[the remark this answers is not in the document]
bogdandrutu: Done...

bogdandrutu:
@open-telemetry/collector-approvers PTAL

service/config_provider.go
```diff
@@ -68,50 +68,112 @@ type ConfigProvider interface {
...
+// withConfigUnmarshaler overwrites the default unmarshaler.
+// This func is private because providing custom ConfigUnmarshaler is not necessary since users can wrap/implement
+// ConfigProvider if needed to change the resulted config.
+func withConfigUnmarshaler(c configunmarshaler.ConfigUnmarshaler) ConfigProviderOption {
+	return func(opts *configProvider) {
+		opts.configUnmarshaler = c
+	}
+}
+
+// Deprecated: [v0.48.0] use NewConfigProvider instead.
```
Aneurysm9, line 109:
```suggestion
// Deprecated: [v0.49.0] use NewConfigProvider instead.
```

service/config_provider.go
```diff
@@ -68,50 +68,112 @@ type ConfigProvider interface {
...
+// WithConfigMapConverters overwrites the default converters.
+func WithConfigMapConverters(c []config.MapConverterFunc) ConfigProviderOption {
+	return func(opts *configProvider) {
+		opts.configMapConverters = c
+	}
+}
+
+// withConfigUnmarshaler overwrites the default unmarshaler.
+// This func is private because providing custom ConfigUnmarshaler is not necessary since users can wrap/implement
+// ConfigProvider if needed to change the resulted config.
```
Aneurysm9, line 102: I'm not sure I understand this, or the comment on the same topic in the changelog.  Can you provide an example here of how it would be done?  It seems to me that the `config.Map` has already been turned into a `config.Config` and so there is no ability to influence that translation, only the `config.Map` or the `config.Config`.

This is a capability that exists now.  Are we sure that it is unused and will not present issues if we remove it?  We have made assumptions before about how people use the config processing system and got it very wrong.  I'd like to avoid that happening again.

service/config_provider.go
```diff
@@ -68,50 +68,112 @@ type ConfigProvider interface {
...
+func WithConfigMapProvider(mp config.MapProvider) ConfigProviderOption {
+	return func(opts *configProvider) {
+		opts.configMapProviders[mp.Scheme()] = mp
+	}
+}
+
+// WithConfigMapConverters overwrites the default converters.
+func WithConfigMapConverters(c []config.MapConverterFunc) ConfigProviderOption {
+	return func(opts *configProvider) {
+		opts.configMapConverters = c
```
Aneurysm9, line 96:
```suggestion
		opts.configMapConverters = append(opts.configMapConverters, c)
```

This should append rather than replace so that multiple `WithConfigMapConverters` options can be composed.

service/config_provider.go
```diff
@@ -68,50 +68,112 @@ type ConfigProvider interface {
...
+	for _, o := range opts {
+		o(&provider)
 	}
+
+	return &provider, nil
 }
 
-// MustNewDefaultConfigProvider returns the default ConfigProvider from slice of location strings
-// (e.g. file:/path/to/config.yaml) and property overrides (e.g. service.telemetry.metrics.address=localhost:8888).
+// Deprecated: [v0.48.0] use NewConfigProvider instead.
```
Aneurysm9, line 165:
```suggestion
// Deprecated: [v0.49.0] use NewConfigProvider instead.
```

service/config_provider.go
```diff
@@ -68,50 +68,112 @@ type ConfigProvider interface {
...
+
+	provider := configProvider{
 		locations:          locationsCopy,
-		configMapProviders: configMapProviders,
-		cfgMapConverters:   cfgMapConverters,
-		configUnmarshaler:  configUnmarshaler,
-		watcher:            make(chan error, 1),
+		configMapProviders: make(map[string]config.MapProvider),
+		configMapConverters: []config.MapConverterFunc{
+			configmapprovider.NewExpandConverter(),
```
Aneurysm9, line 151: This drops the `NewOverwritePropertiesConverter`.  While equivalent functionality may be available from the `yamlmapprovider` package I would expect this to surprise anyone using the `--set` flag with a suddenly non-functioning deployment.

service/config_provider.go
bogdandrutu: The order of the converters matter, so we cannot do append here because will confuse lots of users. The reason why matters is because I may implement a converter that adds a property using an "envvar" then if the env converter is called before or after produces different results.

service/config_provider.go
bogdandrutu: Done.

service/config_provider.go
bogdandrutu: Done.

service/config_provider.go
bogdandrutu: The `NewOverwritePropertiesConverter` cannot be added here, because it needs the values for `--set`. This is not a regression and not a "non-functioning" deployment, because:
* If users use our distros we take care of that in NewWindowsService and NewCommand;
* If users use a custom distro that does not use NewWindowsService and/or NewCommand and they use:
  *  `MustNewConfigProvider` - that did not have support for `--set` callers needed to specifically use `NewOverwritePropertiesConverter`; When they transition nothing to do;
  *  `MustNewDefaultConfigProvider` that did have support for `--set`; When transition callers need to specifically use `NewOverwritePropertiesConverter` for that;

service/config_provider.go
bogdandrutu: I am 99.99% sure (that the only use-case we heard so far and seen is to alter an unmarshalled `config.Config`, and not to change the unmarshalling logic), that being said, now we have a clear process in place for deprecation. If we make a mistake here, I expect users to come back and tell us that they need that functionality and stop the removal.

I am happy to actually expose this as deprecated, with the "proposed transition", and a note that this will not be removed for at least 2 versions to ensure that possible users are coming and telling us.

bogdandrutu: Reverted this, since is a different change. will submit a different PR to argue about this.

service/config_provider.go
mx-psi: >  we cannot do append here because will confuse lots of users. 

Should `NewConfigProvider` return an error if multiple `WithConfigMapConverters` functions are passed then?

bogdandrutu: Not sure... Personally I don't see a reason to do that.

bogdandrutu: But also this is complicated. We want to have a `default` for this like the expand converter (which probably should not be a converter but that is a separate discussion), if user overwrites that should be accepted correct? Only if it double overwrites is error? Also API accepts slice, so that forces users to think about this...

If you really think that this should be an error let me know, but I don't see this as an error, also the users of this API are very few people :))

CHANGELOG.md
```diff
@@ -42,6 +44,7 @@
 
 - Move MapProvider to config, split providers in their own package (#5030)
 - API related to `pdata.AttributeValue` is deprecated in favor of `pdata.Value` (#4978)
+- API related to `pdata.AttributeValue` is deprecated in favor of `pdata.Value` (#4975)
```
codeboten, line 47: Seems like a typo here
```suggestion
```

service/config_provider_test.go
```diff
@@ -34,11 +34,12 @@ import (
 )
 
 type mockProvider struct {
-	retM *config.Map
-	errR error
-	errS error
-	errW error
-	errC error
+	shceme string
```
codeboten, line 37:
```suggestion
	scheme string
```

service/config_provider_test.go
```diff
@@ -55,7 +56,10 @@ func (m *mockProvider) Retrieve(_ context.Context, _ string, watcher config.Watc
 }
 
 func (m *mockProvider) Scheme() string {
-	return "mock"
+	if m.shceme == "" {
```
codeboten, line 59:
```suggestion
	if m.scheme == "" {
```

service/config_provider_test.go
```diff
@@ -55,7 +56,10 @@ func (m *mockProvider) Retrieve(_ context.Context, _ string, watcher config.Watc
...
+		return "mock"
+	}
+	return m.shceme
```
codeboten, line 62:
```suggestion
	return m.scheme
```

service/config_provider_test.go
```diff
@@ -77,86 +81,72 @@ func TestConfigProvider_Errors(t *testing.T) {
...
 		},
 		{
 			name:      "retrieve_err",
 			locations: []string{"mock:", "err:"},
-			parserProvider: map[string]config.MapProvider{
-				"mock": &mockProvider{},
-				"err":  &mockProvider{errR: errors.New("retrieve_err")},
+			parserProvider: []config.MapProvider{
+				&mockProvider{},
+				&mockProvider{shceme: "err", errR: errors.New("retrieve_err")},
```
codeboten, line 103:
```suggestion
				&mockProvider{scheme: "err", errR: errors.New("retrieve_err")},
```

service/config_provider_test.go
```diff
@@ -77,86 +81,72 @@ func TestConfigProvider_Errors(t *testing.T) {
...
-			parserProvider: func() map[string]config.MapProvider {
+			parserProvider: func() []config.MapProvider {
 				cfgMap, err := configtest.LoadConfigMap(filepath.Join("testdata", "otelcol-nop.yaml"))
 				require.NoError(t, err)
-				return map[string]config.MapProvider{
-					"mock": &mockProvider{},
-					"err":  &mockProvider{retM: cfgMap, errC: errors.New("close_err")},
+				return []config.MapProvider{
+					&mockProvider{},
+					&mockProvider{shceme: "err", retM: cfgMap, errC: errors.New("close_err")},
```
codeboten, line 149:
```suggestion
					&mockProvider{scheme: "err", retM: cfgMap, errC: errors.New("close_err")},
```

service/config_provider_test.go
bogdandrutu: I like how IDE was smart enough so if I wrote a typo once, helped me to survive with that :)))

service/config_provider_test.go
bogdandrutu: Done.

service/config_provider_test.go
bogdandrutu: Done.

service/config_provider_test.go
bogdandrutu: Done.

service/config_provider_test.go
bogdandrutu: Done.

CHANGELOG.md
bogdandrutu: Done.

bogdandrutu:
@codeboten thanks for the review, PTAL

service/config_provider.go
codeboten: I think the argument being an array makes it clear enough that folks should set multiple with a single call here.

I can see how it's confusing for `WithConfigMapConverters` to overwrite an option each time it's called while `WithConfigMapProvider` appends to the configuration. I don't have a good idea of how to discern between the two other than with documentation.

Maybe `WithConfigMapProvider` should also overwrite the option to remain consistent?

codeboten (commented):
Thanks for addressing my suggestions. Just one comment in the discussion w/ @Aneurysm9 & @mx-psi. It would be good to agree on the way to move forward.

service/config_provider.go
Aneurysm9: `WithConfigMapProvider` modifies a map, though, where order does not matter and multiple providers may come from multiple sources.  If anything, I'd like to see `WithConfigMapConverters` move in that direction, creating a set of available converters that could then be referenced in a list of conversion steps to apply.  I don't know that there's anything that can be done one way but not the other, perhaps with the exception of having a component with a single method that returns a slice of options intended to register providers and converters.

bogdandrutu: The provider is only one so not sure... We can accept a slice there as well. Also the providers are called in the order of the location argument so their order of registration does not matter.

Another option is to have only one Converter, and add support for MultiConverter as a helper implementation, and then that will overwrite the default converter again

service/config_provider.go
Aneurysm9: We need to commit to maintaining support for `--set`, then, because removal of that flag will break existing deployments.  We cannot decide that having two ways to do the same thing is wrong and remove it.

bogdandrutu: Who we?

bogdandrutu: The flag is not removed as pointed. Not sure what you are talking, sorry.

bogdandrutu: How is this discussion relevant to this PR?

service/config_provider.go
bogdandrutu: @codeboten maybe another good option is o have a "ConfigProviderSettings" instead of using options. And would be clear that we have a map there for providers and a slice for converters. Users can see them easily then?

codeboten: I think that would work. 👍

bogdandrutu: Will give a try in a different PR or commit

jpkrohling: I think the Settings solution is the best solution to solve the ambiguity that multiple `WithConfigMapConverters` options would bring.

service/config_provider.go
```diff
@@ -68,50 +68,110 @@ type ConfigProvider interface {
...
-		makeConfigMapProviderMap(filemapprovider.New(), envmapprovider.New(), yamlmapprovider.New()),
-		[]config.MapConverterFunc{
+		WithConfigMapConverters([]config.MapConverterFunc{
 			configmapprovider.NewOverwritePropertiesConverter(properties),
 			configmapprovider.NewExpandConverter(),
-		},
-		configunmarshaler.NewDefault())
+		}))
+	if err != nil {
+		panic(err)
```
jpkrohling, line 172: I guess this is so early in the code that we don't have even a basic logging facility set, right?

jpkrohling (commented):
LGTM, the only contention point seems to be about calling WithConfigMapConverters multiple times. Once that is sorted out, either by returning an error on the second call or by using a settings object instead of options, this would be ready to be merged.

service/config_provider.go
bogdandrutu: Yes

service/collector_test.go
```diff
@@ -106,13 +107,17 @@ func testCollectorStartHelper(t *testing.T) {
...
 	metricsPort := testutil.GetAvailablePort(t)
-	cfgProvider, err := NewConfigProvider(
-		[]string{filepath.Join("testdata", "otelcol-config.yaml")},
-		WithConfigMapConverters([]config.MapConverterFunc{
-			configmapprovider.NewOverwritePropertiesConverter(
-				[]string{"service.telemetry.metrics.address=localhost:" + strconv.FormatUint(uint64(metricsPort), 10)},
-			),
-		}))
+	cfgSet := NewDefaultConfigProviderSettings()
+	cfgSet.Locations = []string{
```
jpkrohling, line 111: I didn't know this before, but I prefer this so much more than the previous iteration.

service/config_provider.go
bogdandrutu: Done, settings is available now.

jpkrohling (approved):

service/config_provider.go
```diff
@@ -68,50 +68,105 @@ type ConfigProvider interface {
...
+	Locations []string
+
+	// MapProviders is a map of pairs <scheme, config.MapProvider>.
+	// It is required to have at least one config.MapProvider.
+	MapProviders map[string]config.MapProvider
+
+	// MapConverters is a slice of config.MapConverterFunc.
+	MapConverters []config.MapConverterFunc
+
+	// The configunmarshaler.ConfigUnmarshaler to be used to unmarshal the config.Map into config.Config.
```
codeboten, line 94:
```suggestion
	// Unmarshaler is the configunmarshaler.ConfigUnmarshaler to be used to unmarshal the config.Map into config.Config.
```

service/config_provider.go
```diff
@@ -68,50 +68,105 @@ type ConfigProvider interface {
...
+	// It is required to not be nit, use configunmarshaler.NewDefault() by default.
```
codeboten, line 95:
```suggestion
	// It is required to not be nil, use configunmarshaler.NewDefault() by default.
```

service/config_provider.go
```diff
@@ -68,50 +68,105 @@ type ConfigProvider interface {
...
+	Unmarshaler configunmarshaler.ConfigUnmarshaler
+}
+
+func newDefaultConfigProviderSettings() ConfigProviderSettings {
```
codeboten, line 99: I'm probably missing this somewhere, but as a user is there a way for me to obtain the defaults from outside this package? I'm thinking of the case where someone may only want to override the locations but not the default providers/converters

codeboten (commented):
Overall I like this a lot. Just a question and a couple of nits

service/config_provider.go
bogdandrutu: I think, if we go with the same model as the Factories for components, we should not have any "defaults" every distribution will pick their "defaults" that they commit to support. All the components that we use in the "default" config are also public, so a distro can emulate the same if this is what they want.

codeboten (approved):

service/config_provider.go
mx-psi: This still needs to be applied

service/command.go
```diff
@@ -29,7 +31,17 @@ func NewCommand(set CollectorSettings) *cobra.Command {
...
+				cfgSet := newDefaultConfigProviderSettings()
+				cfgSet.Locations = getConfigFlag()
+				// Append the "overwrite properties converter" as the first converter.
+				cfgSet.MapConverters = append(
+					[]config.MapConverterFunc{overwritepropertiesmapconverter.New(getSetFlag())},
+					cfgSet.MapConverters...)
+				set.ConfigProvider, err = NewConfigProvider(cfgSet)
+				if err != nil {
+					return err
+				}
```
mx-psi, line 44: This and the `collector_windows` code are the same, should we extract it to a function?

mx-psi (approved):
Using a struct resolves the ambiguity of repeating an option, I only have two nits but otherwise LGTM

Issue #4906, opened 2022-02-22
`componenttest.NewNop[Exporter|Processor|etc.]Factory` do not respect contribution guideline


bogdandrutu:
They return a global unique instance instead of creating new instance, see related comment https://github.com/open-telemetry/opentelemetry-collector/pull/4899#discussion_r811568953

jpkrohling:
Should this be closed:

Haleygo:
@bogdandrutu  I'd like to work on it :) Is that ok?
I will remake those New.. func under /componenttest, let them return new instance instead of using global instance.
/assign

PR #4940, opened 2022-03-02
Refactor some new factory func under component/componenttest

Haleygo:
**Description:** 
Fixing a bug: Refactor some new factory func under component/componenttest.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4906


bogdandrutu (approved):

Haleygo:
> @Haleygo there were some tests written where the assumption was that the instance is the same.
@bogdandrutu Sorry I missed that test. 
Now I changed those NewFactory func, not let them use the same global Factory. And they can't be asserted equal with func in them.
Can I delete those asserts now or should I add some other test?

PR #4945, opened 2022-03-02
Release prep for v0.46.0

Aneurysm9:


cmd/otelcorecol/builder-config.yaml
```diff
@@ -2,29 +2,29 @@ dist:
   module: go.opentelemetry.io/collector/cmd/otelcorecol
   name: otelcorecol
   description: Local OpenTelemetry Collector binary, testing only.
-  version: 0.45.1-dev
-  otelcol_version: 0.45.0
+  version: 0.46.1-dev
```
codeboten, line 5:
```suggestion
  version: 0.46.0-dev
```

Aneurysm9: `v0.46.0-dev` would sort before `v0.46.0` and would imply that it is work toward `v0.46.0`, which is already released.  I think this should either stay as `v0.46.1-dev` or be `v0.47.0-dev`.

codeboten (approved):

Issue #4951, opened 2022-03-03
add the option to enable server reflection on otlp grpc receiver


tvaintrob:
**Describe the solution you'd like**
a configuration to enable server reflection on the grpc otlp receiver something like the following:
```yaml
receivers:
  otlp:
    protocols:
      grpc:
        enable_reflection: true
```

**Describe alternatives you've considered**
thought about adding a new receiver plugin but seems like the better solution is to add the functionality to the default one

id be happy to implement this one myself and submit a PR

tigrannajaryan:
Please clarify the purpose of this feature. What does server reflection do?

jpkrohling:
The biggest advantage is listing the resources a server provides. It allows a tool like grpccurl to be more user-friendly. We'd likely not have this enabled in production, but I can see how this can help debug connectivity and configuration issues.

WarriorOfWire:
I'm trying to write a tool that sends messages directly to opentelemetry collector and I receive this when sending messages via your proto service definition:
```
(logged by my client - Channel established, request made it to otel collector, this is the response from otel collector)

{ code: Unimplemented, message: "unknown service opentelemetry.proto.metrics_service.MetricsService", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }
```

The very next thing I did to understand what the collector is looking for was to reach for `grpcurl`'s `list` verb.
```
➜  ~ grpcurl -plaintext localhost:4317 list                        
Failed to list services: server does not support the reflection API
```
I still do not know what's wrong.

Here is how I'm running otel:
```
docker run -p 4317:4317 --name otel --rm otel/opentelemetry-collector
```

Not supporting reflection at all is quite hostile to those integrating with the collector. I actually thought the missing reflection was a bug!

AlexDHoffer:
Why is this not supported? Would love to be able to grpcurl this thing without it being a huge pain

brandtkilian:
If that can be of any interest as a workaround, I am using `grpcui`, then I cloned [the protobuf repo](https://github.com/open-telemetry/opentelemetry-proto) and I am using something along:

```bash
grpcui -plaintext -import-path/path/to/opentelemetry-proto/ -proto /path/to/opentelemetry-proto/opentelemetry/proto/collector/metrics/v1/metrics_service.proto 127.0.0.1:{your_otlp_grpc_port}
```

Then I can craft my export metrics test queries from a web ui.

[...]

PR #4953, opened 2022-03-03
added option to enable server reflection on the otlp receiver

tvaintrob:
**Description:** Adding a feature - added option to enable grpc reflection on the otlp receiver
**Link to tracking Issue:** #4951
**Testing:** no additional testing
**Documentation:** no additional documentation

### DIFF
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -157,6 +157,9 @@ type GRPCServerSettings struct {
 	// Include propagates the incoming connection's metadata to downstream consumers.
 	// Experimental: *NOTE* this option is subject to change or removal in the future.
 	IncludeMetadata bool `mapstructure:"include_metadata,omitempty"`
+
+	// EnableReflection will add grpc reflection apis to the receiver
+	EnableReflection bool `mapstructure:"enable_reflection,omitempty"`
 }
 
 // SanitizedEndpoint strips the prefix of either http:// or https:// from configgrpc.GRPCClientSettings.Endpoint.
--- a/receiver/otlpreceiver/otlp.go
+++ b/receiver/otlpreceiver/otlp.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/gorilla/mux"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/component/componenterror"
@@ -121,6 +122,10 @@ func (r *otlpReceiver) startProtocolServers(host component.Host) error {
 
 		if r.logReceiver != nil {
 			otlpgrpc.RegisterLogsServer(r.serverGRPC, r.logReceiver)
+		}
+
+		if r.cfg.GRPC.EnableReflection {
+			reflection.Register(r.serverGRPC)
 		}
 
 		err = r.startGRPCServer(r.cfg.GRPC, host)


config/configgrpc/configgrpc.go
```diff
@@ -157,6 +157,9 @@ type GRPCServerSettings struct {
 	// Include propagates the incoming connection's metadata to downstream consumers.
 	// Experimental: *NOTE* this option is subject to change or removal in the future.
 	IncludeMetadata bool `mapstructure:"include_metadata,omitempty"`
+
+	// EnableReflection will add grpc reflection apis to the receiver
+	EnableReflection bool `mapstructure:"enable_reflection,omitempty"`
```
mx-psi, line 162: What do you think about having a `reflection` section with an `enabled` boolean instead? In case we want to add any other reflection related settings in the future.

mx-psi (commented):
This needs a changelog entry

config/configgrpc/configgrpc.go
tvaintrob: sounds good, ill make the changes

tvaintrob:
how should i update the changelog?

mx-psi:
> how should i update the changelog? @mx-psi

You can add this to 'Unreleased' under a new enhacenments section on the `CHANGELOG.md` file

config/configgrpc/configgrpc.go
```diff
@@ -124,6 +124,11 @@ type KeepaliveEnforcementPolicy struct {
 	PermitWithoutStream bool          `mapstructure:"permit_without_stream,omitempty"`
 }
 
+// GRPCReflectionConfig defines settings for gRPC reflection configuration.
+type GRPCReflectionConfig struct {
```
mx-psi, line 128: Settings sounds more consistent with the rest of the file
```suggestion
// GRPCReflectionSettings defines settings for gRPC reflection configuration.
type GRPCReflectionSettings struct {
```

tvaintrob: done

receiver/otlpreceiver/otlp.go
```diff
@@ -123,6 +124,12 @@ func (r *otlpReceiver) startProtocolServers(host component.Host) error {
 			otlpgrpc.RegisterLogsServer(r.serverGRPC, r.logReceiver)
 		}
 
+		if r.cfg.GRPC.Reflection != nil {
+			if r.cfg.GRPC.Reflection.Enabled {
+				reflection.Register(r.serverGRPC)
```
mx-psi, line 129: The `ConfigGRPCSettings` struct is used by components other than the otlpreceiver. With this implementation, other components will have a `reflection.enabled` option available, but it won't do anything unless implemented. 

Unfortunately, gRPC does not expose this as a `ServerOption`, so the solution may have to be to introduce a `ToServer` method as it happens with the HTTPServer settings.

@jpkrohling (picking you since you were requested a review) what do you think is the best approach here?

config/configgrpc/configgrpc.go
```diff
@@ -124,6 +124,11 @@ type KeepaliveEnforcementPolicy struct {
 	PermitWithoutStream bool          `mapstructure:"permit_without_stream,omitempty"`
 }
 
+// GRPCReflectionSettings defines settings for gRPC reflection configuration.
+type GRPCReflectionSettings struct {
+	Enabled bool `mapstructure:"enabled,omitempty"`
+}
```
bogdandrutu, line 130: What is this? It is not used anywhere.

receiver/otlpreceiver/otlp.go
jpkrohling: Agree, I think the cleanest would indeed have a ToServer.

PR #4954, opened 2022-03-03
Collector Run will listen for context done

### LINKED ISSUE #4842: Collector Run does not listen for context Done
**Describe the bug**
[Collector.Run](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L207) takes in a context but it doesn't appear that it actually listens for the context to be canceled at all. 

**Steps to reproduce**

```go
// Collector setup
...
ctx, cancel := context.WithCancel(context.Background())
cancel()

// Will not exit if configured correctly will continue to run
collector.Run(ctx)
```

Since `Run` is a blocking call I would expect a canceled context to cause `Run` to return with the `ctx.Err()` as the returned error.

**What did you see instead?**
A correctly configured collector will move into a running state and continue to work

**What version did you use?**
v0.44.0

**What config did you use?**
Any valid configuration will work

**Environment**
OS: macOS Monterey 12.2

**Additional context**
Looking at `Run` there are two sections of operations:

- Startup
- Running and waiting for a shutdown/reconfigure event

At the very least [runAndWaitForShutdownEvent](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L207) should add a `ctx.Done()` case to the select statement and allow it gracefully shutdown. Ideally there would also be an exit point during startup. It looks like the context is being passed through the stack during startup and may trigger in a certain set of circumstances (I didn't follow each stack to it's bottom so unsure).

I was just curious and searched the entire repo for `ctx.Done()` and found a single use in [otlp_test.go](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L207). Based on that I'm wondering if there are other places in the collector we should be listening for a closed context.

cpheps:
Fixes #4842 

Added listening for `<-ctx.Done()` in `Run` select so if the passed in context cancels the collector will too.

There was a weird pattern though where internal `shutdown` takes a context and currently it's the passed in one. For the case of context cancelling I called the `shutdown` function with background context. Wondering if we want to put a timer on this or if we want to always use background context on `shutdown`. I could see a scenario where the collector is shutting down and the passed in context is canceled and interrupts the graceful shutdown.

**Link to tracking Issue:** #4842 

**Testing:** Added unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Remove `Type` funcs in pdata (#4933)
 
+## 🧰 Bug fixes 🧰
+
+- Collector `Run` will now exit when a context cancels ()
+
 ## v0.46.0 Beta
 
 ### 🛑 Breaking changes 🛑
--- a/service/collector.go
+++ b/service/collector.go
@@ -174,6 +174,10 @@ LOOP:
 		case <-col.shutdownChan:
 			col.logger.Info("Received shutdown request")
 			break LOOP
+		case <-ctx.Done():
+			col.logger.Error("Context done, terminating process", zap.Error(ctx.Err()))
+			// Call shutdown with background context as the passed in context has been canceled
+			return col.shutdown(context.Background())
 		}
 	}
 	return col.shutdown(ctx)
--- a/service/collector_test.go
+++ b/service/collector_test.go
@@ -253,6 +253,43 @@ func TestCollector_ReportError(t *testing.T) {
 	assert.Equal(t, Closed, col.GetState())
 }
 
+// TestCollector_ContextCancel tests that the collector gracefully exits on context cancel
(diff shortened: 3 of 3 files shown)


bogdandrutu (approved):
I am confused if we need `Shutdown` at all at this point. I think this is a good change, but would like to think about.

cpheps:
> I am confused if we need Shutdown at all at this point. I think this is a good change, but would like to think about.

I think it's still useful to have `Shutdown`. Since the collector is likely the core thing being run in a binary containing it I would think most use cases for context would be from signal intercepts. If you didn't want to intercept signals `Shutdown` provides a clear way to stop the collector. I think using context or `Shutdown` achieve the same behavior but calling `Shutdown` is more deliberate on the part of the wrapper code.

PR #4968, opened 2022-03-08
Add `pipeline` key with pipeline identifier to processor loggers

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### 💡 Enhancements 💡
 
 - Add `pdata.AttributeMap.RemoveIf`, which is a more performant way to remove multiple keys (#4914)
+- Add `pipeline` key with pipeline identifier to processor loggers (#4968)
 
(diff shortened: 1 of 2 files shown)


MovieStoreGuy (approved):

jpkrohling:
Is there an issue tracking this? Looks sane, but I would love to understand what triggered this.

Aneurysm9:
> Is there an issue tracking this? Looks sane, but I would love to understand what triggered this.

It was triggered by [a message in #otel-collector slack](https://cloud-native.slack.com/archives/C01N6P7KR6W/p1646713341467289).  It's a bit unfortunate that there doesn't seem to be a good way to tie pipelines to the exporters or receivers they use, but given that they can be part of multiple pipelines having the component name in the logger is probably the best we can do there.

jpkrohling (approved):

codeboten (approved):

Issue #4917, opened 2022-02-24
Prototype test friendly featureflags


bogdandrutu:
See the idea https://github.com/open-telemetry/opentelemetry-collector/pull/4108#discussion_r726260681

yankay:
Hi @bogdandrutu, would you please review the PR.

PR #4969, opened 2022-03-08
change feature registry as a prototype to make test friendly

yankay:
change feature registry as a prototype to make test friendly.
* change  type registry to public
* create a public func "GetRegistry" to return the prototype
* remove the golbal function: List, IsEnabled, Apply
* change the unit test

**Link to tracking Issue:**  https://github.com/open-telemetry/opentelemetry-collector/issues/4917

Signed-off-by: Kay Yan <kay.yan@daocloud.io>

### DIFF
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -58,8 +58,12 @@ const (
 )
 
 func init() {
+	initWithRegistry(featuregate.GetRegistry())
+}
+
+func initWithRegistry(registry *featuregate.Registry) {
 	//nolint:staticcheck
-	featuregate.Register(featuregate.Gate{
+	registry.Register(featuregate.Gate{
 		ID:          useOtelForInternalMetricsfeatureGateID,
 		Description: "controls whether the collector to uses open telemetry for internal metrics",
 		Enabled:     configtelemetry.UseOpenTelemetryForInternalMetrics,
@@ -112,7 +116,7 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
 	instanceID := instanceUUID.String()
 
 	var pe http.Handler
-	if featuregate.IsEnabled(useOtelForInternalMetricsfeatureGateID) {
+	if featuregate.GetRegistry().IsEnabled(useOtelForInternalMetricsfeatureGateID) {
 		otelHandler, err := tel.initOpenTelemetry()
 		if err != nil {
 			return err
--- a/service/collector_test.go
+++ b/service/collector_test.go
@@ -151,13 +151,14 @@ func TestCollector_Start(t *testing.T) {
 
 func TestCollector_StartWithOtelInternalMetrics(t *testing.T) {
 	resetCollectorTelemetry()
-	originalFlag := featuregate.IsEnabled(useOtelForInternalMetricsfeatureGateID)
+	reg := featuregate.NewRegistry()
+	originalFlag := reg.IsEnabled(useOtelForInternalMetricsfeatureGateID)
 	defer func() {
-		featuregate.Apply(map[string]bool{
+		reg.Apply(map[string]bool{
 			useOtelForInternalMetricsfeatureGateID: originalFlag,
 		})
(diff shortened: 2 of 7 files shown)


service/telemetry.go
```diff
@@ -59,7 +59,7 @@ const (
 
 func init() {
 	//nolint:staticcheck
-	featuregate.Register(featuregate.Gate{
+	featuregate.GetRegistry().Register(featuregate.Gate{
```
bogdandrutu, line 62: To see the benefit of this change, we should split init into something like:

```golang
func init() {
	initWithRegistry(featuregate.GetRegistry())
}

func initWithRegistry(registry *featuregate.Registry) {
	registry.Register(featuregate.Gate{
		ID:          useOtelForInternalMetricsfeatureGateID,
		Description: "controls whether the collector to uses open telemetry for internal metrics",
		Enabled:     configtelemetry.UseOpenTelemetryForInternalMetrics,
	})
}
```

Then change all the tests to run `initWithRegistry` and test using a custom "Registry" per test. Does it make sense?

yankay: HI @bogdandrutu 

We add a new func 'newRegistry()' to create object in test.
and add the func 'initWithRegistry' with unit test.

Would you please review it?

bogdandrutu:
Please rebase. 

@open-telemetry/collector-approvers please review

bogdandrutu:
@Aneurysm9 consider please to read the PR and the issue associated. 

This is mostly to help `testing` by having the possibility to design internally the funcs to accept a registry, and don't have to deal with globals in unit-tests.

```golang
func init() {
	initWithRegistry(featuregate.GetRegistry())
}

func initWithRegistry(registry *featuregate.Registry) {
	// ... do work
}
```

With this you can write unittests that do not change the global state, and importantly they can be run in parallel and guarantee to be isolated.

Aneurysm9:
> @Aneurysm9 consider please to read the PR and the issue associated.

I have read the PR and associated issue.  Why would you assume that I hadn't done so before commenting on the implementation in the PR?

> This is mostly to help `testing` by having the possibility to design internally the funcs to accept a registry, and don't have to deal with globals in unit-tests.
> 
> With this you can write unittests for `initWithRegistry` that do not change the global state, and importantly they can be run in parallel and guarantee to be isolated.

How does this help with testing?  At most it allows testing of the `init()` implementation, which should be calling a function with a statically defined `Gate` struct.  Testing that is not the interesting part.  Testing the use of the gate is the interesting part and, without doing the sort of dependency injection that I mentioned above, that still depends on the global instance and is in no way improved by this PR.  This comes at the cost of breaking the API for all consumers of the package.

bogdandrutu:
@Aneurysm9 

> How does this help with testing? At most it allows testing of the init() implementation, which should be calling a function with a statically defined Gate struct. 

> If we are to provide a Registry for consumers to use we need to be very careful to ensure that it is available to all consumers that may need to test flag status at all points where flag status may need to be checked. 

All these are making a wrong assumption that we need to "provide" the registry. I trust that engineers can design/refactor their code (similarly to the init example) in a way that the interesting functionality that use the "Gate"/"calls the registry" can be extracted into a separate func that accepts a registry as an argument. So then they can write unit tests. Same pattern by the way is used in the "go sdk for TracerProvider vs global", idea being you write the instrumentation plugins (code that uses the gates) to accept a "Registry" but if not provided then use the global.

bogdandrutu:
@Aneurysm9 here is actually the continuation of this PR and how tests can be improved https://github.com/open-telemetry/opentelemetry-collector/pull/5160/commits/f1b821c705bd37e111b4d9edbd3175060abbefcd (part of a cloned PR https://github.com/open-telemetry/opentelemetry-collector/pull/5160)

bogdandrutu:
Closing this, since we have the continuation https://github.com/open-telemetry/opentelemetry-collector/pull/5160

Issue #4967, opened 2022-03-07
Featuregates don't work until late in initialization


aabmass:
**Describe the bug**

The `--feature-gates` CLI flag is not applied until this line:

https://github.com/open-telemetry/opentelemetry-collector/blob/f980c9ef25b1110288ab1b0a297419ec83b541a1/cmd/otelcorecol/main.go#L34

As a result any `featuregate.IsEnabled(myFeatureGateID)` checks during component factory creation (L16 in the same file) will return the default value regardless of the CLI flag. This also effects the contrib repo, for example in the [googlecloud exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/de10b84532d20bd04d4e7088ecdfe5d034c409e8/exporter/googlecloudexporter/factory.go#L45) this check always returns false.

**Steps to reproduce**

Perform a `featuregate.IsEnabled()` check in a `NewFactory()` method, for example here https://github.com/open-telemetry/opentelemetry-collector/blob/83c6ed136695cb4eb75ec9c65544aa56eea8f0db/exporter/otlpexporter/factory.go#L35

**What did you expect to see?**
featuregate checks should work early during factory creation. They may be used to enable/disable features early on during initialization. They shouldn't have different result depending on the time they are called.

**What did you see instead?**
the `--feature-gates` CLI flag has no effect if checked during factory creation.

**What version did you use?**
main

**What config did you use?**
N/A

**Environment**
Linux/go1.18

cc @dashpole

PR #4977, opened 2022-03-10
Move initialization of component factories after flag parsing

dashpole:
**Description:**

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4967

Instead of assembling factories before creating and executing the cobra service, construct factories within the cobra run command so it happens after feature flag parsing.

**Testing:**

Manually tested by adding a feature gate to the OTLP receiver, and checking its value during the NewFactory() call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,10 @@
 - The `featuregates` were not configured from the "--feature-gates" flag on windows service (#5060)
 - Fix Semantic Convention Schema URL definition for 1.5.0 and 1.6.1 versions (#5103)
 
+### 🧰 Bug fixes 🧰
+
+- Fix the use of feature gates within factory initialization. (#4977)
+
 ## v0.47.0 Beta
 
 ### 🛑 Breaking changes 🛑
--- a/cmd/builder/internal/builder/templates/main.go.tmpl
+++ b/cmd/builder/internal/builder/templates/main.go.tmpl
@@ -8,32 +8,47 @@ package main
 import (
 	"log"
 
+	"github.com/spf13/cobra"
+
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/service"
+	"go.opentelemetry.io/collector/service/featuregate"
 )
 
 func main() {
-	factories, err := components()
-	if err != nil {
-		log.Fatalf("failed to build components: %v", err)
-	}
-
 	info := component.BuildInfo{
 		Command:     "{{ .Distribution.Name }}",
 		Description: "{{ .Distribution.Description }}",
 		Version:     "{{ .Distribution.Version }}",
 	}
 
-	if err := run(service.CollectorSettings{BuildInfo: info, Factories: factories}); err != nil {
-		log.Fatal(err)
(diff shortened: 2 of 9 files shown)


bogdandrutu (commented):
I understand the problem, and the solution is nice. I want to propose for thinking another alternative, can we only keep as our public API the Collector and let the generator deal with flags etc.? I think that way we don't have to change settings. The reason to keep only that is to separate concerns collector is a lib and builder creates the main func.

dashpole:
The issue i've run into is that flag parsing is done as part of `cmd.Execute()` on the cobra command.

Currently, `component.Factories` is an input to `service.CollectorSettings`, which is an input to `service.NewCommand()`.  We call `cmd.Execute` on the command returned by `service.NewCommand()`, which calls [command.ParseFlags](https://github.com/spf13/cobra/blob/9369465955fcc766aed3e7cf574277ee553251d4/command.go#L1625).

So as long as completed factories is an input to the command, we will not have parsed flags prior to instantiating factories.

I think options are:

1. Switch to instantiating factories within the command (this PR, or something similar).
2. Don't use cobra for the feature gate flag, and just parse it prior to instantiating factories.  We lose the flag-related benefits of cobra, and (may?) have similar issues with other flags, if the values are referenced during factory creation.
3. Stop using cobra all-together (not a real option, just for completeness)
4. Leave it as-is, and document that feature-gates should not be used from within `NewFactory()` calls.

dashpole:
Note: it looks like I would need to make the API change in one PR, then update the builder, and then switch to the new API in a second PR here.

bogdandrutu:
> Don't use cobra for the feature gate flag, and just parse it prior to instantiating factories. We lose the flag-related benefits of cobra, and (may?) have similar issues with other flags, if the values are referenced during factory creation.

I still believe that removing "NewCommand" (this is a simple func anyway) and "NewWindowsService" from our public "offering" and let people use the builder to create the main func is a valid option. I am on PTO and will look into this more when back on 16th.

dashpole:
I was able to implement your suggestion above by moving NewService to the builder.  One issue is that the builder integration tests pin a version of the collector (currently 0.38.0).  Since this changes some functions in service (GetConfigFlag(), GetSetFlag(), Flags()), we will need to make those changes, cut a release, and then merge the rest of this PR.

dashpole:
If the changes here look reasonable, i'll open a separate PR to make the changes to service flag functions

bogdandrutu:
@dashpole for me this looks right. I think as part of generating the "cmd" (via the builder) directory we should generate the `NewCommand` as well as the `WindowsHandler`. I would be curious about what others think? A different option is to move these to a different package than the current "service" and maybe consider to use different `Settings` definition where we can have a "provider" for factories as your initial example.

dashpole:
I think the approach here is simpler, and has less of an API surface than doing the latter.  I'll open the change to the public interfaces.

dashpole:
Windows probably doesn't need to Reimplement the entire NewCommand.  It can just perform flag parsing from the builder.

dashpole:
This will continue to fail until it can be rebased on https://github.com/open-telemetry/opentelemetry-collector/pull/5130

PR #4980, opened 2022-03-11
Change behavior of oneof numeric value field accessors

dmitryax:
The following methods are deprecated in favor of the new accessors that are more performant and require numeric type match:
  - `Exemplar.IntVal` -> `Exemplar.IntValue`
  - `Exemplar.DoubleVal` -> `Exemplar.DoubleValue`
  - `NumberDataPoint.IntVal` -> `NumberDataPoint.IntValue`
  - `NumberDataPoint.DoubleVal` -> `NumberDataPoint.DoubleValue`

Corresponding setter methods are also renamed for consistency without changing its logic:
  - `Exemplar.SetIntVal` -> `Exemplar.SetIntValue`
  - `Exemplar.SetDoubleVal` -> `Exemplar.SetDoubleValue`
  - `NumberDataPoint.SetIntVal` -> `NumberDataPoint.SetIntValue`
  - `NumberDataPoint.SetDoubleVal` -> `NumberDataPoint.SetDoubleValue`

Updates: https://github.com/open-telemetry/opentelemetry-collector/issues/4775

Similar to: https://github.com/open-telemetry/opentelemetry-collector/pull/4909

PR #4987, opened 2022-03-13
Make AttributeValue accessors to return error in case of type mismatch

dmitryax:
This is an alternative solution to https://github.com/open-telemetry/opentelemetry-collector/pull/4909 proposed by @Aneurysm9

The following methods are deprecated in favor of the new accessors:
- StringVal -> StringValue
- IntVal -> IntValue
- DoubleVal -> DoubleValue
- BoolVal -> BoolValue
- MapVal -> MapValue
- SliceVal -> SliceValue
- BytesVal -> BytesValue

The new accessors return additional error. The error is raised if accessor is called for a wrong type, e.g. `IntValue()` is called on `AttributeValue` of `AttributeValueTypeString` type.

Corresponding setter methods are also renamed for consistency without changing its logic:
- SetStringVal -> SetStringValue
- SetIntVal -> SetIntValue
- SetDoubleVal -> SetDoubleValue
- SetBoolVal -> SetBoolValue
- SetBytesVal -> SetBytesValue

## Benchmarks

goos: darwin
goarch: amd64
pkg: go.opentelemetry.io/collector/model/internal/pdata
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz

### Before

```
BenchmarkAttributeValueStringDeprecatedAccessor-16                  1000000000           0.6064 ns/op          0 B/op          0 allocs/op
BenchmarkAttributeValueIntDeprecatedAccessor-16                     1000000000           0.6103 ns/op          0 B/op          0 allocs/op
BenchmarkAttributeValueDoubleDeprecatedAccessor-16                  1000000000           0.6048 ns/op          0 B/op          0 allocs/op
BenchmarkAttributeValueBoolDeprecatedAccessor-16                    1000000000           0.6933 ns/op          0 B/op          0 allocs/op
BenchmarkAttributeValueBytesDeprecatedAccessor-16                   1000000000           0.6501 ns/op          0 B/op          0 allocs/op
BenchmarkAttributeValueMapDeprecatedAccessor-16                     556089897            2.182 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueArrayDeprecatedAccessor-16                   541185199            2.179 ns/op           0 B/op          0 allocs/op

```

### After

```
BenchmarkAttributeValueStringAccessor-16                            353806957            3.464 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueIntAccessor-16                               346630108            3.479 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueDoubleAccessor-16                            356147574            3.358 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueBoolAccessor-16                              350702415            3.472 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueBytesAccessor-16                             331781389            3.677 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueMapAccessor-16                               341309820            3.559 ns/op           0 B/op          0 allocs/op
BenchmarkAttributeValueArrayAccessor-16                             346380688            3.493 ns/op           0 B/op          0 allocs/op
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@
 - Rename `pdata.AttributeMap.Delete` to `pdata.AttributeMap.Remove` (#4914)
 - pdata: deprecate funcs working with InternalRep (#4957)
 
+### 🚩 Deprecations 🚩
+
+- Change behavior of pdata.AttributeValue oneof value field accessors (#4909)
+  - Deprecated `AttributeValue.StringVal` in favor of `AttributeValue.StringValue`
+  - Deprecated `AttributeValue.IntVal` in favor of `AttributeValue.IntValue`
+  - Deprecated `AttributeValue.DoubleVal` in favor of `AttributeValue.DoubleValue`
+  - Deprecated `AttributeValue.BoolVal` in favor of `AttributeValue.BoolValue`
+  - Deprecated `AttributeValue.MapVal` in favor of `AttributeValue.MapValue`
+  - Deprecated `AttributeValue.SliceVal` in favor of `AttributeValue.SliceValue`
+  - Deprecated `AttributeValue.BytesVal` in favor of `AttributeValue.BytesValue`
+  - Deprecated `AttributeValue.SetStringVal` in favor of `AttributeValue.SetStringValue`
+  - Deprecated `AttributeValue.SetIntVal` in favor of `AttributeValue.SetIntValue`
+  - Deprecated `AttributeValue.SetDoubleVal` in favor of `AttributeValue.SetDoubleValue`
+  - Deprecated `AttributeValue.SetBoolVal` in favor of `AttributeValue.SetBoolValue`
+  - Deprecated `AttributeValue.SetBytesVal` in favor of `AttributeValue.SetBytesValue`
+
 ### 💡 Enhancements 💡
 
 - Add `pdata.AttributeMap.RemoveIf`, which is a more performant way to remove multiple keys (#4914)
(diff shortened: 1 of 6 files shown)


dmitryax:
I see some pros and cons of this approach comparing to https://github.com/open-telemetry/opentelemetry-collector/pull/4909

Pros:

- Safe usage, no chance to catch a runtime error.

Cons:

- Clunky interface raising only one error which can be omitted by user if the type is checked in advance. We have to explicitly communicate this to assure client that there is no other error can be returned.
- Degraded performance.

If we decide to go with this approach we have to update : 
- [numeric value field accessors](https://github.com/open-telemetry/opentelemetry-collector/pull/4980) and 
- [Metric data field accessors](https://github.com/open-telemetry/opentelemetry-collector/blob/main/model/internal/pdata/generated_metrics.go#L630-L663)

Both of them will likely introduce the performance hit as well.

bogdandrutu:
I don't like this especially because of the extra check when I already know the type. we should go with some standard way of doing this. 

1. For example let's have the same behavior as Google protobuf https://github.com/open-telemetry/opentelemetry-proto-go/blob/ee3f9617238186538f35760934fc1727e807b5c3/otlp/common/v1/common.pb.go#L131 unfortunately no easy way to check for nil on the struct case.
2. Panic, this very similar to a type cast, and if you don't check the conversion status you get panic. So not sure why doing this. I read the comments in the other PR but don't prove to me why panic is not natural especially how type casting works in golang

dmitryax:
Closing based on decision made in https://github.com/open-telemetry/opentelemetry-collector/issues/4775

PR #4995, opened 2022-03-14
Remove all deprecated funcs/structs from v0.46.0

bogdandrutu:


model/internal/pdata/common.go
```diff
@@ -511,9 +511,7 @@ func (am AttributeMap) Get(key string) (AttributeValue, bool) {
 	return AttributeValue{nil}, false
 }
 
-// Delete deletes the entry associated with the key and returns true if the key
-// was present in the map, otherwise returns false.
-// Deprecated: [v0.46.0] Use Remove instead.
+// Deprecated: [v0.47.0] Use Remove instead.
```
codeboten, line 514: Is the intent here to keep this for 2 versions? if so i'd leave the number to 0.46.0 to remind us that it's been deprecated for 2 versions :)

bogdandrutu: As you can see, this is just fixing the fact that it was deprecated in v0.47.0 not in v0.46.0 :) This is not removed.

codeboten: Ah yes. Thank you for clarifying

codeboten (approved):

Issue #4988, opened 2022-03-13
Avoid using `Attribute` in common pdata components related to AnyValue proto field


dmitryax:
## Problems

All of the pdata wrappers related to [AnyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L27) fields include `Attribute` part in its name because the field used to be applicable to the `attributes` fields only, like [here](https://github.com/open-telemetry/opentelemetry-proto/blob/17c68a9d2b67d702153d91fa985ecf2b87f74b7c/opentelemetry/proto/logs/v1/logs.proto#L206). But it's not the case anymore since `pdata.LogRecord.Body` uses [AnyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L135) wrapped by `pdata.AttributeValue`. This is the main goal that intended to be addressed in https://github.com/open-telemetry/opentelemetry-collector/issues/4818.

But the are couple of other naming issues related to this problem that need to be addressed:

### 1. AttributeMap naming

`pdata.AttributeMap` struct is a wrapper around [repeated KeyValue proto field](https://github.com/open-telemetry/opentelemetry-proto/blob/17c68a9d2b67d702153d91fa985ecf2b87f74b7c/opentelemetry/proto/common/v1/common.proto#L58). The name seems reasonable when it's used for attributes like [here](https://github.com/open-telemetry/opentelemetry-proto/blob/17c68a9d2b67d702153d91fa985ecf2b87f74b7c/opentelemetry/proto/trace/v1/trace.proto#L237). But it also can be used as an instance of [AnyValue](https://github.com/open-telemetry/opentelemetry-proto/blob/17c68a9d2b67d702153d91fa985ecf2b87f74b7c/opentelemetry/proto/common/v1/common.proto#L36). For example the following line currently returns a value of `AttributeMap` type while there is nothing related to _attribute_ in the call:

```go
pdata.NewLogs().ResourceLogs().AppendEmpty().LogRecords().AppendEmpty().Body().MapVal()
```

### 2. AttributeValueSlice naming

There are couple of problem with this name:

1. It can be easily confused with all the other `AttributeValue...` types like `AttributeValueString`, `AttributeValueInt`, `AttributeValueMap`, `AttributeValueArray`, etc. They both have the same naming schema but different meaning:

[...]

PR #5001, opened 2022-03-15
Remove "Attribute" part from common pdata collections names

dmitryax:
All of the pdata wrappers for collections include Attribute part in its name because the fields used to be part of the attributes fields only. But it's not the case anymore since pdata.LogRecord.Body uses AnyValue that can contain any common collection. This change renames AttributeMap and AttributeValueSlice collections by removing Attribute part from their names and making them consistent.

Renames:
  - Deprecate `pdata.AttributeMap` struct in favor of `pdata.Map`
  - Deprecate `pdata.NewAttributeMap` func in favor of `pdata.NewMap`
  - Deprecate `pdata.NewAttributeMapFromMap` func in favor of `pdata.NewMapFromRaw`
  - Deprecate `pdata.AttributeValueSlice` struct in favor of `pdata.Slice`
  - Deprecate `pdata.NewAttributeValueSlice` func in favor of `pdata.NewSlice`

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4988


model/internal/pdata/common.go
```diff
@@ -446,23 +446,23 @@ func newAttributeKeyValueBytes(k string, v []byte) otlpcommon.KeyValue {
...
 	orig := []otlpcommon.KeyValue(nil)
-	return AttributeMap{&orig}
+	return Map{&orig}
 }
 
-// NewAttributeMapFromMap creates a AttributeMap with values
+// NewMapFromRaw creates a Map with values
 // from the given map[string]AttributeValue.
-func NewAttributeMapFromMap(attrMap map[string]AttributeValue) AttributeMap {
+func NewMapFromRaw(attrMap map[string]AttributeValue) Map {
```
dmitryax, line 462: This name is chosen for consistency with existing `Map.AsRaw` method

dmitryax: The only difference between them is that `AsRaw` returns `map[string]interface{}` not `map[string]AttributeValue` but I think it's ok. The types shouldn't be changed given their use cases.

model/internal/pdata/common.go
```diff
@@ -471,21 +471,21 @@ func NewAttributeMapFromMap(attrMap map[string]AttributeValue) AttributeMap {
...
-func newAttributeMap(orig *[]otlpcommon.KeyValue) AttributeMap {
-	return AttributeMap{orig}
+func newMap(orig *[]otlpcommon.KeyValue) Map {
+	return Map{orig}
 }
 
-// Clear erases any existing entries in this AttributeMap instance.
-func (am AttributeMap) Clear() {
+// Clear erases any existing entries in this Map instance.
+func (am Map) Clear() {
```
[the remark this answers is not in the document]
dmitryax: Thanks. Updated

model/internal/pdata/common.go
bogdandrutu: Hmm... I don't like this inconsistency... let's think about this because it is a subtle difference that may confuse users.

model/pdata/common_alias.go
```diff
@@ -48,11 +48,26 @@ var (
...
-// Aliases for functions to create pdata.AttributeMap.
+// Deprecated: [v0.47.0] Use Map instead.
+type AttributeMap = pdata.Map
+
+// Aliases for functions to create pdata.Map.
 var (
-	NewAttributeMap        = pdata.NewAttributeMap
-	NewAttributeMapFromMap = pdata.NewAttributeMapFromMap
+	NewMap        = pdata.NewMap
+	NewMapFromRaw = pdata.NewMapFromRaw
```
bogdandrutu, line 60: nit: separate PR, I would prefer to rename `internal/pdata` to maybe just `internal`, I am confused about the 2 packages named the same, when code refers to one vs the other.

model/internal/pdata/common.go
dmitryax: Actually I think we can change NewMapFromRaw to take `map[string]interface{}` as well. I checked it's being used in tests only, so performance overhead (if there is any) doesn't matterr. Let me know what do you think. It'll make client code even more concise. Let me know what do you think

bogdandrutu: Would like to see that code..

dmitryax: Made this change in the second commit to this PR. Please let me know what do you think

model/internal/pdata/common.go
```diff
@@ -131,6 +131,35 @@ func NewValueBytes(v []byte) Value {
 	return Value{orig: &otlpcommon.AnyValue{Value: &otlpcommon.AnyValue_BytesValue{BytesValue: v}}}
 }
 
+func newValueFromRaw(iv interface{}) Value {
+	switch tv := iv.(type) {
+	case nil:
+		return NewValueEmpty()
+	case string:
+		return NewValueString(tv)
+	case int64:
```
bogdandrutu, line 140: Here I think we should accept all type of `ints` (int8, int32, int, int64) and `uints` (uint8, uint32, uint, uint64)

model/internal/pdata/common.go
```diff
@@ -131,6 +131,35 @@ func NewValueBytes(v []byte) Value {
...
+		return NewValueInt(tv)
+	case float64:
```
bogdandrutu, line 142: we should also accept `float32` here.

bogdandrutu (approved):

model/internal/pdata/common.go
dmitryax: I assume you skipped `int16` and `uint16` unintentionally

model/internal/pdata/common.go
dmitryax: Done

model/internal/pdata/common.go
dmitryax: Added all variants of int also `pdata.Map` and `pdata.Slice` as another representation of map and slice

model/internal/pdata/common.go
```diff
@@ -132,6 +132,65 @@ func NewValueBytes(v []byte) Value {
...
+		return NewValueDouble(tv)
+	case bool:
+		return NewValueBool(tv)
+	case []byte:
+		return NewValueBytes(tv)
+	case Map:
+		mv := NewValueMap()
+		mv.MapVal().EnsureCapacity(tv.Len())
+		tv.CopyTo(mv.MapVal())
+		return mv
```
bogdandrutu, line 173: I would not add this for the moment. If we do this we need `Value` itself as well, and that also needs to copy :)

model/internal/pdata/common.go
```diff
@@ -132,6 +132,65 @@ func NewValueBytes(v []byte) Value {
...
+	case map[string]interface{}:
+		mv := NewValueMap()
+		mv.MapVal().EnsureCapacity(len(tv))
```
bogdandrutu, line 176: CopyTo should do the right thing in terms of optimizing the allocation, if that is not the case we should fix that, but we should not have to call `EnsureCapacity`.

bogdandrutu (approved):

model/internal/pdata/common.go
dmitryax: Not sure I understand which `Value` do you mean here :) Removed for now

bogdandrutu (approved):

PR #4978, opened 2022-03-10
Rename pdata.AttributeValue to pdata.AnyValue

dmitryax:
`pdata.AttributeValue` is a wrapper over `opentelemetry.proto.common.v1.AnyValue` type. It's not exclusively used with `pdata.AttributeMap`, `pdata.LogRecord.Body` also uses which makes the name misleading. This change renames all API related to `pdata.AttributeValue` to `pdata.AnyValue` keeping the old API as deprecated

Updates: https://github.com/open-telemetry/opentelemetry-collector/issues/4818

PR #5004, opened 2022-03-15
[CHANGELOG] Move all deprecation entries to "Deprecations" section

dmitryax:
Move all the deprecation entries from "Breaking Changes" to its own section as discussed https://github.com/open-telemetry/opentelemetry-collector/pull/4978#discussion_r825459643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,12 @@
 ### 🛑 Breaking changes 🛑
 
 - Remove `Type` funcs in pdata (#4933)
-- Rename `pdata.AttributeMap.Delete` to `pdata.AttributeMap.Remove` (#4914)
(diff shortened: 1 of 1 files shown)


codeboten (approved):
Thanks for fixing this!

codeboten (commented):
Not sure if it's worth doing, but there are a few other deprecated notes in the changelog, see version 0.14.0. Those may be too old to be worth addressing though. 🤷

dmitryax:
> Not sure if it's worth doing, but there are a few other deprecated notes in the changelog, see version 0.14.0. Those may be too old to be worth addressing though. 🤷

It makes sense to update all of them. Updated the PR

bogdandrutu (approved):

PR #5034, opened 2022-03-17
[pdata] Change outcome of misuse of oneof collection accessors.

dmitryax:
Change misuse outcome of `pdata.AttributeMap.MapVal()` and `pdata.AttributeValueSlice.SliceVal()` function calls. In case of type mismatch, they now return an invalid zero-initialized instance instead of returning a valid detached collection.

This PR implements the approach that was agreed upon in https://github.com/open-telemetry/opentelemetry-collector/issues/4775.


Aneurysm9 (approved):

model/internal/pdata/common.go
```diff
@@ -186,27 +186,27 @@ func (v Value) BoolVal() bool {
...
 //
 // Calling this function on zero-initialized Value will cause a panic.
 func (v Value) MapVal() AttributeMap {
 	kvlist := v.orig.GetKvlistValue()
 	if kvlist == nil {
-		return NewAttributeMap()
+		return AttributeMap{}
 	}
 	return newAttributeMap(&kvlist.Values)
 }
```
bogdandrutu, line 199:
```golang
func (v Value) MapVal() AttributeMap {
	kvlist, ok := v.orig.GetValue().(*otlpcommon.AnyValue_KvlistValue)
	if !ok {
		return AttributeMap{}
	}
	return newAttributeMap(&kvlist.KvlistValue.Values)
}
```

model/internal/pdata/common.go
```diff
@@ -186,27 +186,27 @@ func (v Value) BoolVal() bool {
...
 
 // SliceVal returns the slice value associated with this Value.
-// If the Type() is not ValueTypeArray then returns an empty slice. Note that modifying
-// such empty slice has no effect on this Value.
+// If the Type() is not ValueTypeArray then returns an invalid slice. Note that using
+// such slice can cause panic.
 //
 // Calling this function on zero-initialized Value will cause a panic.
 func (v Value) SliceVal() AttributeValueSlice {
```
bogdandrutu, line 206:
```golang
func (v Value) SliceVal() AttributeValueSlice {
	arr, ok := v.orig.GetValue().(*otlpcommon.AnyValue_ArrayValue)
	if !ok {
		return AttributeValueSlice{}
	}
	return newAttributeValueSlice(&arr.ArrayValue.Values)
}
```

model/internal/pdata/common.go
dmitryax: Hm... `v.orig.Get<other>Value()` essentially implement the same instructions. So I'll just update these two

dmitryax: done

bogdandrutu (approved):

bogdandrutu:
Failing tests :)

dmitryax:
@bogdandrutu the failing test is checking the following input which unlikely to be caught in reality but I think we still need to handle it:

```
	orig := &otlpcommon.AnyValue{Value: &otlpcommon.AnyValue_KvlistValue{KvlistValue: nil}}
	m1 = Value{orig: orig}
	assert.EqualValues(t, AttributeMap{}, m1.MapVal())
```

To handle I would have to add another if check to the construction that you suggested, It'll look like this (Updated):

```
func (v Value) MapVal() AttributeMap {
	kvlist, ok := v.orig.GetValue().(*otlpcommon.AnyValue_KvlistValue)
	if !ok || kvlist.KvlistValue == nil {
		return AttributeMap{}
	}
	return newAttributeMap(&kvlist.KvlistValue.Values)
}
```

I run the benchmarks showing that original code of this PR appeared to be more performant:

Original:
```
BenchmarkAttributeValueMapAccessor-16    	1000000000	         0.7589 ns/op
BenchmarkAttributeValueArrayAccessor-16    	1000000000	         1.124 ns/op
```

Updated:
```
BenchmarkAttributeValueMapAccessor-16    	1000000000	         1.102 ns/op
BenchmarkAttributeValueArrayAccessor-16    	838671705	         1.350 ns/op
```

dmitryax:
I've brought back the original code for map and slice value getters for now

bogdandrutu (approved):

PR #5035, opened 2022-03-18
[pdata] Change outcome of Metric data accessors misuse

dmitryax:
Change outcome of `pdata.Metric.<Gauge|Sum|Histogram|ExponentialHistogram>()` functions misuse. In case of type mismatch, they don't panic right away but return an invalid zero-initialized instance for consistency with other OneOf field accessors

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4775


Aneurysm9 (approved):

model/internal/cmd/pdatagen/internal/base_fields.go
```diff
@@ -64,10 +64,17 @@ const oneOfTypeAccessorHeaderTestTemplate = `func Test${structName}${originField
...
+// zero-initiazed instance of ${returnType}. Note that using such ${returnType} instance can cause panic.
+//
 // Calling this function on zero-initialized ${structName} will cause a panic.
 func (ms ${structName}) ${fieldName}() ${returnType} {
-	return new${returnType}((*ms.orig).${originOneOfFieldName}.(*${originStructType}).${originFieldName})
+	v := ms.orig.Get${originFieldName}()
+	if v == nil {
+		return ${returnType}{}
+	}
+	return new${returnType}(v)
```
bogdandrutu, line 77: nit: This has an extra "if check" in the get. Try to do the type conversion here and use ok directly.

bogdandrutu (approved):

model/internal/cmd/pdatagen/internal/base_fields.go
```diff
@@ -64,10 +64,17 @@ const oneOfTypeAccessorHeaderTestTemplate = `func Test${structName}${originField
 	assert.Equal(t, "", ${typeName}(1000).String())`
 
 const accessorsOneOfMessageTemplate = `// ${fieldName} returns the ${lowerFieldName} associated with this ${structName}.
-// Calling this function when ${originOneOfFieldName}Type() != ${typeName} will cause a panic.
+//
+// Calling this function when ${originOneOfFieldName}Type() != ${typeName} returns an invalid 
+// zero-initiazed instance of ${returnType}. Note that using such ${returnType} instance can cause panic.
```
mx-psi, line 69:
```suggestion
// zero-initialized instance of ${returnType}. Note that using such ${returnType} instance can cause panic.
```

model/internal/cmd/pdatagen/internal/base_fields.go
dmitryax: done

model/internal/cmd/pdatagen/internal/base_fields.go
dmitryax: thanks for catching this. fixed

mx-psi (approved):

PR #5036, opened 2022-03-18
Run `make gotidy`

dashpole:
Some of the go.sum files seem to be out-of-date.  Run `make gotidy` to update them.


dashpole:
This is strange, the behavior of `make gotidy` and `make genotelcorecol` seem to be different, even though both are running `go mod tidy -compat=1.17` in the `cmd/otelcorecol` directory...

PR #5053, opened 2022-03-21
Deprecate featureflags.FlagValue.SetSlice, unnecessary public

bogdandrutu:


service/featuregate/flags_test.go
```diff
@@ -96,33 +109,3 @@ func TestFlagValue_withPlus(t *testing.T) {
...
-		},
-		{
-			name:     "repeated items",
-			input:    []string{"foo", "-bar", "-foo"},
-			expected: FlagValue{"foo": true, "bar": false},
-		},
-	} {
-		t.Run(tc.name, func(t *testing.T) {
-			v := FlagValue{}
-			assert.NoError(t, v.SetSlice(tc.input))
```
jpkrohling, line 124: Could you please keep this, removing only when the public function is removed?

bogdandrutu: Why? We test that via the "comma separated" cases, so the only untested code is the call from `SetSlice` to `setSlice`.

jpkrohling: Which is exactly what is being deprecated :-)

bogdandrutu: I don't think I understand:

```golang
// fully tested
func Set() {
  // calls setSlice()
}

func SetSlice() {
  // Not tested 1 line
  return setSlice()
}

// fully tested
func setSlice() {}
}
```

bogdandrutu: The "setSlice" is very well tested, it is only that deprecated func that is not tested which does very little (to nothing).

jpkrohling (approved):

PR #5054, opened 2022-03-21
Deprecate LogRecord.Name(), it was deprecated in the data model

bogdandrutu:

### DIFF
--- a/model/internal/pdata/generated_log.go
+++ b/model/internal/pdata/generated_log.go
@@ -640,6 +640,8 @@ func (ms LogRecord) SetSeverityNumber(v SeverityNumber) {
 }
 
 // Name returns the name associated with this LogRecord.
+//
+// Deprecated: [v0.48.0] it was removed from the data model.
 func (ms LogRecord) Name() string {
 	return (*ms.orig).Name
 }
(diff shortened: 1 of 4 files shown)


model/internal/pdata/generated_log.go
```diff
@@ -640,6 +640,8 @@ func (ms LogRecord) SetSeverityNumber(v SeverityNumber) {
 }
 
 // Name returns the name associated with this LogRecord.
+//
+// Deprecated: [v0.48.0] it was removed from the data model.
```
jpkrohling, line 644: Is there a recommendation for current users of this? Would just removing this call result in the same data quality, or should this `name` be added as an attribute?

jpkrohling: Is the same information supposed to be conveyed in a different way?

tigrannajaryan: No. The field is deprecated because there is no clear understanding on what the field is supposed to carry. Hence, no replacement either. It is unconditionally removed without attempting to support it in any graceful way. There is an ongoing discussion about whether we can have a proper definition of what "Name" should be with precisely defined semantics. If that happens we may add the field, but it will be considered a completely new field, without any graceful transition required.

After keeping this deprecation notice in the codebase we should just delete the field soon.

jpkrohling: Great, thanks for the clarification.

jpkrohling (approved):

PR #5060, opened 2022-03-22
Deprecate global flag in featuregates

bogdandrutu:
Distributions that build without the builder will need to declare it as they do for "config" and "set".


service/collector_windows.go
```diff
@@ -88,6 +90,7 @@ func (s *WindowsService) start(elog *eventlog.Log, colErrorChannel chan error) e
 	if err := flags().Parse(os.Args[1:]); err != nil {
 		return err
 	}
+	featuregate.Apply(gatesList)
```
bogdandrutu, line 93: This PR also fixes the use of featuregates on windows, where the gate flag was not applied.

dmitryax (approved):

Issue #5063, opened 2022-03-22
Support OTLP v0.15.0


codeboten:
Update the proto to the latest release: https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v0.15.0. There are a few tasks to do here:
- [x] #5071
- [x] https://github.com/open-telemetry/opentelemetry-collector/issues/4258
- [x] #5072
- [x] #5073
- [x] #5074

dmitryax:
Will this include renaming of `pdata.InstrumentationLibrary` or we need another ticket for that?

PR #5064, opened 2022-03-22
update proto to 0.15.0

### LINKED ISSUE #4258: Support optional fields in proto

codeboten:
This includes changes to support optional types:
- update sed script to update optional type to oneof
- bump proto to 0.15.0
- make genproto
- update sum to OptionalType

Fixes #4258 #5071 

Part of #5063


model/internal/pdata/generated_metrics.go
```diff
@@ -1428,14 +1428,26 @@ func (ms HistogramDataPoint) SetCount(v uint64) {
 	(*ms.orig).Count = v
 }
 
+// Sum_Type returns the type of the sum_ for this HistogramDataPoint.
+// Calling this function on zero-initialized HistogramDataPoint will cause a panic.
+func (ms HistogramDataPoint) Sum_Type() OptionalType {
```
bogdandrutu, line 1433: I think we should stick with previous public API that has a 'HasSum() bool'

codeboten: Fixed, please take a look.

model/internal/pdata/generated_metrics.go
```diff
@@ -1430,12 +1430,18 @@ func (ms HistogramDataPoint) SetCount(v uint64) {
 
 // Sum returns the sum associated with this HistogramDataPoint.
 func (ms HistogramDataPoint) Sum() float64 {
-	return (*ms.orig).Sum
+	return (*ms.orig).GetSum()
```
bogdandrutu, line 1433: nit: separate PR we should update this comment that a "zero-initialized" value is returned if the value is not present (HasSum() == false)

bogdandrutu (approved):

Issue #5065, opened 2022-03-23
[pdata] Rename `ValueTypeArray` to `ValueTypeSlice` for consistency with pdata.Slice


dmitryax:
Currently we have the following wrappers for proto AnyValue:

```
const (
	ValueTypeEmpty ValueType = iota
	ValueTypeString
	ValueTypeInt
	ValueTypeDouble
	ValueTypeBool
	ValueTypeMap
	ValueTypeArray
	ValueTypeBytes
)
```

And the following wrappers for collection of `pdata.Value` :

```
pdata.Slice
pdata.Map
```

So we have pdata.NewValueMap().MapVal() returning a pdata.Map value.

But pdata.NewValueArray().SliceVal() returning a pdata.Slice value.

We need to rename `ValueTypeArray` to `ValueTypeSlice` for consistency.

More details in the thread started here: https://github.com/open-telemetry/opentelemetry-collector/issues/4988#issuecomment-1073419719

dmitryax:
@Aneurysm9 I think we don't even need to deprecate `ValueTypeArray` because it wasn't released yet. Let me know WDYT

PR #5066, opened 2022-03-23
[pdata] Rename ValueTypeArray to ValueTypeSlice

dmitryax:
`pdata.ValueTypeArray` is removed without deprecation because it was introduced few commits before and not released yet.

Closes: https://github.com/open-telemetry/opentelemetry-collector/issues/5065


bogdandrutu (approved):

model/internal/pdata/common_test.go
```diff
@@ -976,17 +976,17 @@ func generateTestBytesMap() Map {
 	})
 }
 
-func TestAttributeValueArray(t *testing.T) {
-	a1 := NewValueArray()
-	assert.EqualValues(t, ValueTypeArray, a1.Type())
+func TestAttributeValueSlice(t *testing.T) {
```
bogdandrutu, line 979: Separate PR consider to remove `Attribute` from test names.

bogdandrutu (approved):

model/internal/pdata/common_test.go
dmitryax: Sounds good. I'll do

bogdandrutu:
Please rebase

dmitryax:
done

dmitryax:
_contrib-tests / contrib_tests_ is failing because of removed `ValueTypeArray` which is expected. `ValueTypeArray` isn't part of to the latest release, it was added few after 0.47.0, so I believe it can be removed right away. Let me know if there are any objections.

cc @bogdandrutu @Aneurysm9

Aneurysm9 (approved):

Aneurysm9:
> `ValueTypeArray` isn't part of to the latest release, it was added after 0.47.0, so I believe it can be removed right away.

I think that's correct.  We don't have any releases with that identifier so there is no need to perpetuate it for compatibility.

PR #3890, opened 2021-08-25
Enable h2c traffic for http servers.

a-feld:
**Description:** Added support for h2c traffic on http servers generated by confighttp.

This is in preparation for combining gRPC / HTTP ports for OTLP (see #3765 for the complete use case)

**Testing:** The HTTP receive tests have been updated to check for both http/1.1 and h2c traffic when "insecure" mode / non-TLS is used.

PR #3765, opened 2021-08-03
[PROTOTYPE] Combine gRPC and HTTP servers

a-feld:
# gRPC / HTTP server consolidation 🚧  [PROTOTYPE] 🚧 

This PR consolidates the gRPC server and HTTP servers into 1 centralized server. The architecture of this solution enables:
* A consolidated 4317 port for both gRPC and HTTP traffic
* More configurable transports given that the official h2 / h2c golang servers are used rather than the gRPC implementation
* A smaller footprint for the binary given that the gRPC library is no longer compiled into the receiver

## Notes

* We should probably discuss direction on how to transport headers into the processors / exporters. There are valid reasons for wanting to access the headers - one example being the implementation of a multi-tenant solution where API keys may change.

Issue #1256, opened 2020-07-02
Unsplit the OTLP receiver protocols, use a shared port


jmacd:
**Is your feature request related to a problem? Please describe.**
This is a request to revert https://github.com/open-telemetry/opentelemetry-collector/pull/1223

**Describe the solution you'd like**
I would like both protocols to share a port, as this reduces configuration cost for users and system operators.

**Describe alternatives you've considered**
The alternative is to use a single port. The PR https://github.com/open-telemetry/opentelemetry-collector/pull/1223 did not provide a good justification for this change.

**Additional context**
At LightStep we once had a multi-port satellite, and it was a huge problem for our users. We eventually were able to support a single-port configuration with both gRPC and HTTP, secure and insecure settings.

austinlparker:
Mostly just echoing Josh's comment here. Multi-port satellites caused us no end of headaches for users.

bogdandrutu:
@jmacd happy to revert if you show me how to make TLS work with port multiplexing.

jmacd:
I'm asking around Lightstep for someone who can contribute here. Thanks.

jpkrohling:
For context, #1227 might have been the reason for the change.

jmacd:
Just want to repeat that "no end of headaches for users" is not an exaggeration. It was an awful time. I'm looking for someone to help on this.

MatthewDolan:
@bogdandrutu Can you expand more on what about TLS needs to work to make this port viable?

MatthewDolan:
I think there recommendation from grpc-go for supporting both http and grpc on the same port is to utilize the `ServerHTTP(...)` method on the grpc server.

https://godoc.org/google.golang.org/grpc#Server.ServeHTTP

Something similar to the example code in the docs:

```
if r.ProtoMajor == 2 && strings.HasPrefix(r.Header.Get("Content-Type"), "application/grpc") {
	grpcServer.ServeHTTP(w, r)
} else {
	httpMux.ServeHTTP(w, r)
}
```

The TLS support needs to be handled at the http layer because the request has to be decrypted before it can be routed to either the grpc or http mux.

@bogdandrutu, is that the information you were looking for? Apologies if I am missing context from another thread.

bogdandrutu:
@jmacd can you please tell me if all protocols are enabled or you still want user to configure which protocols to enable?

tcolgate:
[...]

PR #5067, opened 2022-03-23
Implement gRPC deframing directly on HTTP receiver port.

anuraaga:
This is basically the opposite of the Java exporter work detailed in https://github.com/open-telemetry/opentelemetry-specification/discussions/1996. Notably, implementing unary gRPC is fairly simple with off-the-shelf HTTP frameworks. 

This isn't complete but I can finish it up if there is interest in this approach. IMO it's the best way to achieve gRPC and HTTP receivers on the same port.

Verified it basically works with the Java integration tests in https://github.com/open-telemetry/opentelemetry-java/compare/main...anuraaga:otlp-unify?expand=1

Notable functionality improvements (?) with this approach

- Only have to implement server auth extension once for HTTP and gRPC as this would just use the HTTP extensions
- Can use HTTP 1.1 with gRPC if someone wants to (only client I know of that supports this is Armeria Java but maybe there are more, but no official ones)

### DIFF
--- a/receiver/otlpreceiver/otlp.go
+++ b/receiver/otlpreceiver/otlp.go
@@ -15,10 +15,16 @@
 package otlpreceiver // import "go.opentelemetry.io/collector/receiver/otlpreceiver"
 
 import (
+	"compress/gzip"
 	"context"
+	"encoding/binary"
 	"fmt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"io"
 	"net"
 	"net/http"
+	"strconv"
 	"sync"
 
 	"github.com/gorilla/mux"
@@ -185,8 +191,55 @@ func (r *otlpReceiver) registerTraceConsumer(tc consumer.Traces) error {
 		r.httpMux.HandleFunc("/v1/traces", func(resp http.ResponseWriter, req *http.Request) {
 			handleUnmatchedRequests(resp, req)
 		})
+
+		r.httpMux.HandleFunc(
+			"/opentelemetry.proto.collector.trace.v1.TraceService/Export",
+			func(resp http.ResponseWriter, req *http.Request) {
+				prepareGrpcResponse(resp)
+				body, err := deframeGrpc(req)
+				if err != nil {
+					writeGrpcError(resp, err)
+					return
+				}
+				otlpReq, err := pbEncoder.unmarshalTracesRequest(body)
+				if err != nil {
+					writeGrpcError(resp, err)
+					return
+				}
+				otlpResp, err := r.traceReceiver.Export(req.Context(), otlpReq)
+				if err != nil {
+					writeGrpcError(resp, err)
+					return
+				}
+				msg, err := pbEncoder.marshalTracesResponse(otlpResp)
+				if err != nil {
+					writeGrpcError(resp, err)
+					return
+				}
+				writeGrpcMessage(resp, msg)
+			}).Methods(http.MethodPost)
 	}
 	return nil
+}
+
+func writeGrpcMessage(resp http.ResponseWriter, msg []byte) {
+	_, err := resp.Write([]byte{0})
+	if err != nil {
+		writeGrpcError(resp, err)
+	}
+	length := make([]byte, 4)
+	binary.BigEndian.PutUint32(length, uint32(len(msg)))
+	_, err = resp.Write(length)
+	if err != nil {
+		writeGrpcError(resp, err)
+		return
+	}
+	_, err = resp.Write(msg)
+	if err != nil {
+		writeGrpcError(resp, err)
+		return
+	}
+	writeGrpcStatus(resp, status.New(codes.OK, ""))
 }
 
 func (r *otlpReceiver) registerMetricsConsumer(mc consumer.Metrics) error {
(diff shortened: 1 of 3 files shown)


anuraaga:
@tigrannajaryan @open-telemetry/collector-maintainers

receiver/otlpreceiver/otlp.go
```diff
@@ -185,10 +191,57 @@ func (r *otlpReceiver) registerTraceConsumer(tc consumer.Traces) error {
...
+					return
+				}
+				writeGrpcMessage(resp, msg)
+			}).Methods(http.MethodPost)
 	}
 	return nil
 }
 
+func writeGrpcMessage(resp http.ResponseWriter, msg []byte) {
+	_, err := resp.Write([]byte{0})
```
anuraaga, line 226: Wouldn't be hard to implement compressed responses, didn't bother as we know for now the responses are always empty

jpkrohling:
Is this related to the previous work done by @a-feld as part of #3765? I need to review the notes myself as well, but I remember him mentioning some difficulties and caveats.

anuraaga:
Thanks @jpkrohling I didn't see that PR but the implementation is quite similar indeed. We implement the gRPC wire protocol here, and in particular the definition of the OTLP spec, so I think there are no strict differences. There are gRPC features like retry and load balancing that the OTel spec does not specify that we probably shouldn't support in the collector in general. But luckily these are client concerns and I'm not personally aware of gRPC-specific server side features that we might miss so maybe not such a big deal. But I'm not 100% sure that we miss any gRPC feature, I do know we don't miss any OTel spec feature though.

anuraaga:
I found https://github.com/open-telemetry/opentelemetry-collector/pull/3890 too to split the changes up. I don't see any reason to not merge that other than a lack of timely review. If there is interest in trying this again, I can resubmit the PRs. Don't want to submit if they'll just stalebot though :)

jpkrohling:
> Don't want to submit if they'll just stalebot though :)

There were a couple of concerns from what I remember:

- backwards compatibility 
- supportability (do we have people who'd have the time and ability to support this in the long run?)
- performance
- TLS, per-RPC auth, ... 

If none of them are problems, we should certainly move forward. It would still be good to present a proposal and/or make a demo during the SIG call.

anuraaga:
> do we have people who'd have the time and ability to support this in the long run

This is an important question that I can't answer. If maintainers aren't at all invested in this approach we should not pursue it I think. 

But I think it's good for users and should be a priority for maintainers. Maybe unfairly strict, but I would say that because the TC chose to support gRPC (unnecessarily), we would generally expect members to ensure a good experience and do everything it take to make it happen. Well I guess @bhs's values don't actually prioritize this so feel free to ignore :-)

tigrannajaryan:
> > do we have people who'd have the time and ability to support this in the long run
> 
> This is an important question that I can't answer. If maintainers aren't at all invested in this approach we should not pursue it I think.

It does not look like a huge maintenance burden to me if this is all of the code that we need. What do other @open-telemetry/collector-maintainers think?

@anuraaga If I remember correctly in the last attempt (I think Bogdan tried it) we couldn't make mTLS work. Can you confirm that it works with this implementation?

anuraaga:
@tigrannajaryan Thanks I will verify mTLS. Also let me know anything else to verify will be happy to confirm

anuraaga:
@tigrannajaryan I was able to verify mTLS works with grpc-java after a small fix in https://github.com/open-telemetry/opentelemetry-collector/pull/5067/commits/7c1823925366b96f41b493dbee89b3c799726752. Interestingly it seems this fix is needed regardless of this PR - currently it isn't possible to use HTTP/2 with the http receiver which is probably not intended. With it, it'll use ALPN to negotiate, which means HTTP/2 on newer HTTP implementations (and all gRPC clients) and 1.1 on older ones

tigrannajaryan:
> Interestingly it seems this fix is needed regardless of this PR - currently it isn't possible to use HTTP/2 with the http receiver which is probably not intended.

Yeah, OTLP/HTTP receiver must support HTTP/2. Thanks for catching this.

Let's see some feedback from other maintainers and if everyone is happy we can move forward. Thanks for undertaking this @anuraaga

Aneurysm9:
My recollection of the last time we discussed this was that @jmacd relayed issues with supportability and end-user confusion that, in addition to some technical difficulties, led to us abandoning the effort.  I'm good with giving this another go, but would feel more comfortable if we reviewed the recordings of the prior discussions and/or had Josh restate his concerns and we were sure they were all addressed.

tigrannajaryan:
> My recollection of the last time we discussed this was that @jmacd relayed issues with supportability and end-user confusion that, in addition to some technical difficulties, led to us abandoning the effort. I'm good with giving this another go, but would feel more comfortable if we reviewed the recordings of the prior discussions and/or had Josh restate his concerns and we were sure they were all addressed.

@jmacd do you remember what was the end-user confusion caused by? I would expect that allowing to receive both gRCP and plain HTTP on one port should help to reduce confusion since the user does not need to choose 2 different ports on the sending side.

jpkrohling:
> I would expect that allowing to receive both gRCP and plain HTTP on one port should help to reduce confusion since the user does not need to choose 2 different ports on the sending side

About that, we had discussions here as well: https://github.com/open-telemetry/opentelemetry-collector/issues/1256

tigrannajaryan:
Once we are are happy with this implementation we also need to update the OTLP spec to reflect the port merging and it needs to be done in a way that is not a breaking change. So, we need to perform some sort of graceful transition.

jpkrohling:
We discussed this during the SIG call today, and there are a few things to consider:

- @jmacd mentioned in the linked issue that Lightstep reverted their decision to go with a single port due to the number of problems they had. This reflects the fact that this is a critical part of the collection path, and errors or performance issues here might be hard to solve. @bogdandrutu mentioned that he'd not be comfortable in having this code here without the backing of another community, more focused on this layer, like gRPC
- If we do decide to keep this as an experiment, we should only do it after GA: it's easy to converge the ports in the future, but it's hard to split them after GA. By then, if we have enough users providing feedback and once we feel like this is rock solid, we might consider converging the ports for v2.

anuraaga:
Alright will go ahead and close this PR

Issue #4759, opened 2022-01-27
Allow builder to accept multiple config providers


jpkrohling:
This is the builder part of #4567.

jpkrohling:
Question to @pavankrish123, @pmm-sumo, and @bogdandrutu: how would the main.go look like to support this? Once I know that, we can find a way to support it in the builder.

Without looking into many details about the config provider feature, I think it could work in a similar way as the components: it could accept a list of factories which would be loaded in order. The builder could then have a new config option, "configproviders", accepting go modules for those providers like we do for the other components.

Aneurysm9:
I think for this to work we'd need either a conventional way to instantiate a provider given just a package name, or we'd need to include an option to specify a function to invoke to instantiate the provider.  

We should probably do the same for map converters, as well.

jpkrohling:
> or we'd need to include an option to specify a function to invoke to instantiate the provider.

We assume "NewFactory" for the components:

https://github.com/open-telemetry/opentelemetry-collector/blob/32301d929b413918efeefc1d9a1165b03ac71988/cmd/builder/internal/builder/templates/components.go.tmpl#L27

pmm-sumo:
I like the idea with "configproviders" (or "configmapproviders" actually?). 

As @Aneurysm9 noted, doing it at the level of configmapproviders also brings a question how to handle map converters and perhaps unmarshaller (separate set of config entries I presume?). I think refactoring it to make it consistent with other components ("NewFactory()") would be the way to go

Aneurysm9:
One limitation of the conventional approach is that it can only support one (provider|converter|unmarshaller) per package.  The collector currently has several of each in common packages and those would need to be restructured.

atoulme:
This seems like something we should address in a short order. Any ideas?

evan-bradley:
I think the providers are in a good place for this to move forward. Per the discussion in https://github.com/open-telemetry/opentelemetry-collector/issues/6956, we want to configure providers using URI fragments, which are passed when they resolve a config URI, so the `New()` function in each provider package should be sufficient to instantiate the provider.

[...]

PR #5068, opened 2022-03-23
Add Scheme to MapProvider interface

bogdandrutu:
This is another required step in order to support configuring map providers via the Builder.

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/4759

This is a breaking change, since adds a new func to an interface. It is hard to make this backwards compatible, and would like to move forward with this exception.


mx-psi (approved):
LGTM, I find this to be an acceptable breakage of the rules (but would like to hear from at least one other approver)

dmitryax (approved):

dmitryax:
Seems reasonable to me as well

tigrannajaryan:
Does `Scheme()` need to be part of `MapProvider` interface? I see an analogy with components where we made `Type` part of the component factory, not of the component instance itself. Using this same approach we can introduce factories that would then allow creating `MapProvider`s in MustNewDefaultConfigProvider.

bogdandrutu:
> I see an analogy with components where we made Type part of the component factory, not of the component instance itself.

I have a different view of this, kind of the `MapProvider` is the equivalent of the `Factory` and the `Retrieved` value is the component.

tigrannajaryan:
> > I see an analogy with components where we made Type part of the component factory, not of the component instance itself.
> 
> I have a different view of this, kind of the `MapProvider` is the equivalent of the `Factory` and the `Retrieved` value is the component.

Yeah, I guess that works too. The factory-based approach would probably be a bit more uniform with what we do with components from naming perspective, but I think your approach is fine too.

codeboten (approved):

PR #5037, opened 2022-03-18
Run unit tests on go 1.18 in addition to 1.17

dashpole:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5032.

In order to merge this PR, an admin for the repo will need to change the branch protection rules for `main` to require the `build-and-test / unittest (1.18)` and `build-and-test / unittest (1.17)`  checks instead of the `unittest` check.

I don't think this needs a changelog entry.

PR #5075, opened 2022-03-23
Upgrade to Go 1.18

emaderer:
Upgrade to Go 1.18


Makefile.Common
```diff
@@ -34,7 +34,7 @@ fmt:
 .PHONY: tidy
 tidy:
 	rm -fr go.sum
-	$(GOCMD) mod tidy -compat=1.17
+	$(GOCMD) mod tidy -compat=1.18
```
Aneurysm9, line 37: We cannot do this at this time.  We still need to support building with 1.17.  See our [version support policy](https://github.com/open-telemetry/opentelemetry-collector/blob/main/README.md#compatibility).

Aneurysm9 (changes requested):
We cannot drop support for 1.17 at this time.  Support for [testing against 1.18](https://github.com/open-telemetry/opentelemetry-collector/pull/5037) has already been merged.

Makefile.Common
emaderer: I see, thanks for pointing that out!

PR #5076, opened 2022-03-23
Deprecation of pdata methods following OTLP v0.15.0

### LINKED ISSUE #5073: deprecate InstrumentationLibrary* in pdata in favour of Scope*

codeboten:
Follow up to #5064 

Adding deprecation methods for the following pdata methods:
  - InstrumentationLibrary is now Scope
  - InstrumentationLibrarySpans is now ScopeSpans
  - InstrumentationLibraryLogs is now ScopeLogs
  - InstrumentationLibraryMetrics is now ScopeMetrics

Fixes #5073 #5072


tigrannajaryan:
Does this implement the [graceful transition rules](https://github.com/open-telemetry/opentelemetry-proto/blob/3c2915c01a9fb37abfc0415ec71247c4978386b0/opentelemetry/proto/trace/v1/trace.proto#L58) or it will be in a separate PR?

codeboten:
@tigrannajaryan it does not implement them yet. I will do so before marking it ready for review 👍

codeboten:
Sorry @tigrannajaryan I misinterpreted the question in your comment and thought you were talking about the whether this PR would include the deprecated methods to allow users of pdata to transition off of the old names. Those are not included in this PR.

The update to support the transition rules will be done as a separate PR to address #5074

tigrannajaryan:
> Sorry @tigrannajaryan I misinterpreted the question in your comment and thought you were talking about the whether this PR would include the deprecated methods to allow users of pdata to transition off of the old names. Those are now included in this PR.
> 
> The update to support the transition rules will be done as a separate PR to address #5074

OK. Let's plan to do this work before the next release. We want the new release to handle the transition as it is described in the proto spec.

codeboten:
💯

@tigrannajaryan i started here https://github.com/open-telemetry/opentelemetry-collector/pull/5085

tigrannajaryan (approved):

docs/troubleshooting.md
```diff
@@ -176,7 +176,7 @@ With the modified configuration if you re-run the test above the log output shou
 2020-11-11T04:08:17.344Z	DEBUG	loggingexporter/logging_exporter.go:353	ResourceSpans #0
```
bogdandrutu, line 176: Why this change?

bogdandrutu (approved):

docs/troubleshooting.md
cbandy: The coloring here is not a diff; only L179 is changing.

It looks to me like GitHub is highlighting that this code block is identified as `json` but is not valid JSON.

codeboten: The only change is `InstrumentationLIbrarySpans` -> `ScopeSpans`

bogdandrutu (approved):

Issue #5061, opened 2022-03-22
`HTTPClientSettings.ToClient` is inconsistent with other `To*` config methods


mx-psi:
`confighttp.HTTPClientSettings.ToClient` takes an id-extension map while `configgrpc.HTTPServerSettings.ToServer` and all the `configrpc.*.To*` methods take a `component.Host`, in all cases to support authentication. 

Ideally all would be consistent by passing `component.Host` to all of them, although I don't want to assume this is the best solution just because it's the most used one.

cc @jpkrohling since this is auth related.

bogdandrutu:
I would vote to make it consistent and use `component.Host`.

shree007:
@jpkrohling  I may work on it.

fatsheep9146:
Is this issue still open?

mx-psi:
> Is this issue still open?

It is, #5083 was closed as stale

fatsheep9146:
> > Is this issue still open?
> 
> It is, #5083 was closed as stale

Could I work on this?

mx-psi:
Assigned :)

mx-psi:
Reopening since we still have to:
1. remove `ToClient` in a future version
2. add it back with the 'correct' argument

fatsheep9146:
> Reopening since we still have to:
> 
> 1. remove `ToClient` in a future version
> 2. add it back with the 'correct' argument

These two works should both be done until at least version v0.56.0?

mx-psi:
@fatsheep9146 Yes, for now we have to wait!

mx-psi:
@fatsheep9146 I think you can continue doing this now!

PR #5083, opened 2022-03-24
opentelemetry-collector-5061 replacing ext map with host

shree007:
**Description:** 
Adding `component.Host` into `confighttp.HTTPClientSettings.ToClient `to follow similar norm in other places

**Link to tracking Issue:** #5061 

**Testing:**
Paasing `componenttest.NewNopHost()` into testing method


bogdandrutu:
@shree007 can you update the PR to pass the tests in this repo (not the contrib)

codeboten (commented):
@shree007 any update on moving this work forward?

shree007:
> 

Should I remove that check ?

config/confighttp/confighttp_test.go
```diff
@@ -255,6 +249,7 @@ func TestHTTPClientSettingWithAuthConfig(t *testing.T) {
 		shouldErr    bool
 		settings     HTTPClientSettings
 		extensionMap map[config.ComponentID]component.Extension
+		host         component.Host
```
jpkrohling, line 252: No cases are setting these. Should it really be added here?

shree007: https://github.com/open-telemetry/opentelemetry-collector/pull/5083/files#diff-ae54372cbf36c6ca3ce262dc39d7db67e1d71bba20c8838a9906bc123133c8b6R312
Here its being used.

jpkrohling: It's consumed there, but no tests are creating a custom value for it. Each test is defining a map of components, but that's not being used anymore. Shouldn't you create a host based on that map?

shree007: @jpkrohling  let me walk through it again to end to end flow. I will create host based on that map accordingly.

bogdandrutu: It is not used anywhere.

bogdandrutu (commented):
Please make PR pass tests

shree007:
Ok. I will do today

PR #5086, opened 2022-03-25
Introduce additional consumer packages split by telemetry type

dmitryax:
This change adds another set of packages representing "consumer" split by telemetry signal type. The same approach is taken as for the pdata split. Common parts are moved to internal package and then they are used from both existing combined and split consumer packages.

This is just one step towards splitting the pdata usage to illustrate one option of new pdata structure as suggested in this PR: https://github.com/open-telemetry/opentelemetry-collector/pull/4918


bogdandrutu:
I think during the discussion about the split of `pdata` we agreed that we are not going to split all the packages, only `pdata`. I want to make sure we do not go on a different direction.

mx-psi:
> I think during the discussion about the split of `pdata` we agreed that we are not going to split all the packages, only `pdata`. I want to make sure we do not go on a different direction.

My interpretation of what happened is "we agreed for now on pdata split and nothing else", not "we agreed not to split other packages". I agree if we were to split other stuff by telemetry type, it would need further discussion

dmitryax:
> we agreed for now on pdata split and nothing else", not "we agreed not to split other packages

This was my understanding as well. And this PR is just to illustrate how further split can be done to support the further discussion. We need to do "something" with other packages using pdata in order to provide the backward compatible transition to the new pdata packages. This one is a simple example of what can be done for that, others can be more problematic to split, but we have to somehow rename/move them anyway.

Issue #5027, opened 2022-03-17
Split otlp/otlpgrpc


bogdandrutu:
Decision was made to split the pdata package into multiple per signal packages, see https://github.com/open-telemetry/opentelemetry-collector/issues/4832.

As a followup we need to split the `otlp` and `otlpgrpc` as well since they are part of the same current module `model`. Here are some options:
1. Extract the current packages into separate module without no other changes.
2. Move everything from `otlp` and `otlpgrpc` into the equivalent signal package decided in #4832.

I personally believe that the 2nd option is cleaner and gives us the "encapsulation" we need.

dmitryax:
Looks like we need to make a decision here to proceed with https://github.com/open-telemetry/opentelemetry-collector/pull/5087.

Currently we have `otlp` and `otlpgrpc` packages under `/model`.  IMO having them under `model` is not much cleaner than merging them into `pdata`. If we go with the 2 approach, we can have the following structure:

```
/pdata
  /ptrace (including trace related API from `pdata/otlp` and `pdata/grpcotlp`
  /pmetric (including metrics related API from `pdata/otlp` and `pdata/grpcotlp`
  /plog (including logs related API from `pdata/otlp` and `pdata/grpcotlp`
/semconv
```

If we go with 1, we need to figure out another structure. I'm thinking of something like this:

```
/model
  /pdata
    /ptrace (including trace related API from `pdata/otlp` and `pdata/grpcotlp`
    /pmetric (including metrics related API from `pdata/otlp` and `pdata/grpcotlp`
    /plog (including logs related API from `pdata/otlp` and `pdata/grpcotlp`
  /semvconv
  /otlp
    /otlptrace
    /otlpmetric
    /otlplog
  /otlpgrpc
    /grpctrace
    /grpcmetric
    /grpclog
```

We can also remove `/model` for this structure and have `pdata`, `otlp`, `otlpgrpc` and `semvconv` at the root level, but this will  require moving `/model/internal` to `/internal`

cc @Aneurysm9 @bogdandrutu @tigrannajaryan , any other ideas?

tigrannajaryan:
From what I understand `otlpgrpc` is currently used by otlp receiver/exporter only. Is this true? If so then why does it need to be a public package? Can we move it to `internal/otlp` or something like that?

tigrannajaryan:
[...]

PR #5087, opened 2022-03-25
Split pdata package by telemetry signal type

dmitryax:
This is an alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/4918 with the following package structure:

```
/model
  /pcommon (aliases to common API in /internal/pdata: Resource, Map, Value etc)
  /ptrace (aliases to traces API of internal/pdata)
  /plog (aliases to logs API of internal/pdata)
  /pmetric (aliases to metrics API of internal/pdata)
```


bogdandrutu (changes requested):
I like this, except the split of consumer which is out of scope for the pdata split.

dmitryax:
Removed the split of consumer

dmitryax:
Based on the results of [the poll](https://docs.google.com/document/d/1D_jK39GYkp9wAPgRoCYntnzq4LBs_YNnu64vXJGc9nI/edit#heading=h.oeev18el9qk1) https://github.com/open-telemetry/opentelemetry-collector/pull/4918 is closed in favor of this PR

bogdandrutu (approved):

bogdandrutu:
@open-telemetry/collector-maintainer please review

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@
   instance for consistency with other OneOf field accessors (#5035)
 - Update OTLP to v0.15.0 (#5064)
 - Adding support for transition from older versions of OTLP to OTLP v0.15.0 (#5085)
+- Split pdata package separated by telemetry signal type (#5087)
```
codeboten, line 80: Please add a note about the deprecation of `pdata.*` aliases as well

exporter/exporterhelper/internal/persistent_queue_test.go
```diff
@@ -167,20 +168,20 @@ func TestPersistentQueue_ConsumersProducers(t *testing.T) {
...
+func newTraces(numTraces int, numSpans int) ptrace.Traces {
+	traces := ptrace.NewTraces()
 	batch := traces.ResourceSpans().AppendEmpty()
 	batch.Resource().Attributes().InsertString("resource-attr", "some-resource")
 	batch.Resource().Attributes().InsertInt("num-traces", int64(numTraces))
 	batch.Resource().Attributes().InsertInt("num-spans", int64(numSpans))
 
 	for i := 0; i < numTraces; i++ {
-		traceID := pdata.NewTraceID([16]byte{1, 2, 3, byte(i)})
+		traceID := pcommon.NewTraceID([16]byte{1, 2, 3, byte(i)})
```
codeboten, line 179: I can see why NewTraceID and NewSpanID are in `pcommon` as they're used by multiple signals, but since they're always tied to the tracing signal, would it not make more sense for them to live in `ptrace`?

codeboten (commented):
Overall this looks good, just a couple of comments

exporter/exporterhelper/internal/persistent_queue_test.go
dmitryax: I'd put them in pcommon just to maintain a clean dependancy chain and avoid any dependancies across the signal packages.

CHANGELOG.md
dmitryax: Replaced by an entry in "Deprecations" section

codeboten (approved):

exporter/exporterhelper/internal/persistent_queue_test.go
bogdandrutu: Crazy consideration/idea. Can we remove them and replace with a `[8]byte` and `[16]byte` for the APIs in span/exemplar/etc.?

bogdandrutu: This can be a followup issue to discuss.

dmitryax: Sure, submitted https://github.com/open-telemetry/opentelemetry-collector/issues/5137

tigrannajaryan (approved):
LGTM.

I think in our discussions we can continue referring to the internal data collectively as `pdata` even when it is totally eliminated from the codebase. It is a convenient and catchy name.

dmitryax:
> I think in our discussions we can continue referring to the internal data collectively as pdata even when it is totally eliminated from the codebase. It is a convenient and catchy name.

I agree on using `pdata` name as reference to the all pipeline data structures. We agreed on using `p<signal>` as module names, but we still can to move them under an extra `pdata` directory and have the following structure if we want to:

```
/model
  /pdata
    /pcommon (aliases to common API in /internal/pdata: Resource, Map, Value etc)
    /ptrace (aliases to traces API of internal/pdata)
    /plog (aliases to logs API of internal/pdata)
    /pmetric (aliases to metrics API of internal/pdata)
  /semconv
  ... 
```

@open-telemetry/collector-approvers please share you feedback if you think it'll be better.

tigrannajaryan:
> We agreed on using `p<signal>` as module names, but we still can to move them under an extra `pdata` directory

This approach works for me too. No strong opinion one way or another.

bogdandrutu:
@dmitryax @tigrannajaryan what about a `model` -> `pdata` than `semconv` is it's own thing outside pdata module.

dmitryax:
> what about a model -> pdata than semconv is it's own thing outside pdata module.

I was also thinking about it. It means that we need to move `model/otlp` and `model/otlpgrpc` to `pdata/p<signal>`, right? I can submit a followup PR to see how it looks.

tigrannajaryan:
> @dmitryax @tigrannajaryan what about a `model` -> `pdata` than `semconv` is it's own thing outside pdata module.

We have other packages in `model` which strictly speaking are not `pdata` unless we stretch the term (which I would prefer not to). `pdata` in my mind is in-memory representation of Collector pipeline data.

bogdandrutu:
> I was also thinking about it. It means that we need to move model/otlp and model/otlpgrpc to pdata/p<signal>, right? I can submit a followup PR to see how it looks.

See https://github.com/open-telemetry/opentelemetry-collector/issues/5027

> We have other packages in model which strictly speaking are not pdata unless we stretch the term (which I would prefer not to). pdata in my mind is in-memory representation of Collector pipeline data.

Besides semconv nothing after we fix https://github.com/open-telemetry/opentelemetry-collector/issues/5027

Aneurysm9:
> We have other packages in `model` which strictly speaking are not `pdata` unless we stretch the term (which I would prefer not to). `pdata` in my mind is in-memory representation of Collector pipeline data.

This is how I think of it as well.  It seems sufficiently different from OTLP, even if it is presently implemented as an API for managing an OTLP representation, that I think we should be careful about intermixing the concepts too much.

dmitryax:
Looks like we have an agreement on how to merge `otlp` and `otlpgrpc` packages into `pdata` in https://github.com/open-telemetry/opentelemetry-collector/issues/5027. According to that we will have the following two modules instead if `/model`:

```
/pdata
  /ptrace
    /ptracegrpc
  /pmetric
    /pmetricgrpc
  /plog
    /ploggrpc

/semconv
```

If everyone is ok with this structure, I'm going to proceed with the following action items:

1. Move `/model/internal` to `/internal`
2. Update this PR to have `/pdata/p<signal>` instead of `/model/p<signal>`
3. Address https://github.com/open-telemetry/opentelemetry-collector/issues/5027 by introducing `/model/p<signal>/p<signal>oltp` packages

cc @bogdandrutu @tigrannajaryan @Aneurysm9

dmitryax:
I tried a RP for (1) item https://github.com/open-telemetry/opentelemetry-collector/pull/5161. It works but introduces interdependencies between go.opentelemetry.io/collector and go.opentelemetry.io/collector/model modules. So I'd rather update this PR in a different way to avoid interdependencies but use more aliasing for the deprecated APIs.

dmitryax:
It appeared to be pretty difficult to do what we agreed on by the steps that I outlined without introducing interdependencies between packages. To make it in a clean way split of all pdata related parts into a new module has to be done in one shot. I submitted another PR for that instead of reusing this one: https://github.com/open-telemetry/opentelemetry-collector/pull/5168

PR #5090, opened 2022-03-26
[model/semconv] Add definitions for semantic conventions v1.9.0

dmitryax:


model/semconv/v1.9.0/schema.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package semconv // import "go.opentelemetry.io/collector/model/semconv/v1.9.0"
+
+// SchemaURL is the schema URL that matches the version of the semantic conventions
+// that this package defines. Conventions packages starting from v1.4.0 must declare
+// non-empty schema URL in the form https://opentelemetry.io/schemas/<version>
+const SchemaURL = "https://opentelemetry.io/schemas/v1.9.0"
```
cbandy, line 20: Should this be `https://opentelemetry.io/schemas/1.9.0`? xref: https://github.com/open-telemetry/opentelemetry-go/issues/2743

```console
$ curl -fS https://opentelemetry.io/schemas/v1.9.0
curl: (22) The requested URL returned error: 404 

$ curl -fS https://opentelemetry.io/schemas/1.9.0
file_format: 1.0.0
schema_url: https://opentelemetry.io/schemas/1.9.0
versions:
  1.9.0:
  1.8.0:
    spans:
      changes:
        - rename_attributes:
            attribute_map:
              db.cassandra.keyspace: db.name
              db.hbase.namespace: db.name
  1.7.0:
  1.6.1:
  1.5.0:
  1.4.0:
```

dmitryax: @cbandy Good catch, thanks! Looks like this field was defined incorrectly for previous versions as well

dmitryax: Updated

codeboten (approved):

dmitryax:
I would wait for confirmation from @Aneurysm9 as original author of the SemConv definitions that the change of SchemaURL is needed, or if there is any reason to keep the `https://opentelemetry.io/schemas/v` prefix.

tigrannajaryan:
> I would wait for confirmation from @Aneurysm9 as original author of the SemConv definitions that the change of SchemaURL is needed, or if there is any reason to keep the `https://opentelemetry.io/schemas/v` prefix.

@dmitryax I believe the "v" prefix should not be present. It is incorrectly added.

dmitryax:
@tigrannajaryan thanks for confirming

tigrannajaryan (approved):

Aneurysm9 (approved):

PR #5091, opened 2022-03-26
[model/semconv] Add missing files for semconv v1.7.0 and v1.8.0

dmitryax:
Some files that are supposed to be created manually were missed in semconv v1.7.0 and v1.8.0

### DIFF
--- a/model/semconv/v1.7.0/nonstandard.go
+++ b/model/semconv/v1.7.0/nonstandard.go
@@ -1,0 +1,22 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package semconv // import "go.opentelemetry.io/collector/model/semconv/v1.7.0"
+
+const (
+	OtelLibraryName       = "otel.library.name"
+	OtelLibraryVersion    = "otel.library.version"
+	OtelStatusCode        = "otel.status_code"
+	OtelStatusDescription = "otel.status_description"
+)
--- a/model/semconv/v1.7.0/schema.go
+++ b/model/semconv/v1.7.0/schema.go
@@ -1,0 +1,20 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


model/semconv/v1.7.0/schema.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package semconv // import "go.opentelemetry.io/collector/model/semconv/v1.7.0"
+
+// SchemaURL is the schema URL that matches the version of the semantic conventions
+// that this package defines. Conventions packages starting from v1.4.0 must declare
+// non-empty schema URL in the form https://opentelemetry.io/schemas/<version>
+const SchemaURL = "https://opentelemetry.io/schemas/v1.7.0"
```
cbandy, line 20: I noticed that these URLs cannot be downloaded, so I opened https://github.com/open-telemetry/opentelemetry-go/issues/2743.

dmitryax: This file is created manually in this repo. I updated it

dmitryax:
@Aneurysm9 PTAL if the change of SchemaURL value is needed, or if there is any reason to keep the `https://opentelemetry.io/schemas/v` prefix.

model/semconv/v1.7.0/nonstandard.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package semconv // import "go.opentelemetry.io/collector/model/semconv/v1.7.0"
+
+const (
+	OtelLibraryName       = "otel.library.name"
+	OtelLibraryVersion    = "otel.library.version"
+	OtelStatusCode        = "otel.status_code"
+	OtelStatusDescription = "otel.status_description"
+)
```
tigrannajaryan, line 22: I assume this is an exact copy from 1.6.1 directory, right?

tigrannajaryan (approved):

Aneurysm9 (approved):

model/semconv/v1.7.0/nonstandard.go
dmitryax: Yes, it is. I didn't find any changes on these definitions up to 1.10.0

bogdandrutu: If we managed to not have these (especially because they are manually written), is this a sign that we don't need them?

bogdandrutu: If we managed to not have these (especially because they are manually written), is this a sign that we don't need them?

bogdandrutu: If we managed to not have these (especially because they are manually written), is this a sign that we don't need them?

dmitryax: I can add a test verifying that the files are created

dmitryax: added the test https://github.com/open-telemetry/opentelemetry-collector/pull/5106

codeboten (approved):

PR #5103, opened 2022-03-28
Fix Semantic Convention Schema URL definition for 1.5.0 and 1.6.1

dmitryax:
This change fixes the sem conv SchemaURL values, as reported in https://github.com/open-telemetry/opentelemetry-collector/pull/5090#discussion_r835926493

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
 ### 🧰 Bug fixes 🧰
 
 - The `featuregates` were not configured from the "--feature-gates" flag on windows service (#5060)
+- Fix Semantic Convention Schema URL definition for 1.5.0 and 1.6.1 versions (#5103)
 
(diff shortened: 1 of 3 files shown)


dmitryax:
This is technically a breaking change, so we need to make sure that this is needed indeed. I'll keep it in draft for now

cc @Aneurysm9 @bogdandrutu

bogdandrutu (approved):
I think this is a good fix.

/cc @tigrannajaryan

tigrannajaryan (approved):
Yes, this is definitely a bug. The correct URLs don't have "v" prefix, see the Schema Files directory: https://github.com/open-telemetry/opentelemetry-specification/tree/main/schemas

Since Schemas at OpenTelemetry are still Experimental we are free to make this breaking change (really a bug fix).

Aneurysm9 (approved):

Issue #4819, opened 2022-02-08
[pdata] Consider renaming of MetricValueType


dmitryax:
`MetricValueType` type sounds confusing. It's not a type of any metric value, it's applicable to numeric values only.

Originally raised by @bogdandrutu https://github.com/open-telemetry/opentelemetry-collector/pull/4811#discussion_r801186353:

> On the list of "may fix": MetricValueType does not sound as the best name for the enum. Maybe NumberValueType? PointValueType? Not sure, just open an issue, this name does not sound very clear to me, because I would ask why Histogram is not a value type.

Aneurysm9:
I think it is going to get confusing if there is any more than this and https://github.com/open-telemetry/opentelemetry-collector/issues/4818.  If we're going to consider renaming a number of types or methods or fields in the pdata package we should have a single overarching issue and a design doc for it so that we can consider the change holistically.

dmitryax:
These issues are created as soon they discovered. At this point I don't plan to create any new. I believe discussing each of these renames under dedicated issues is more convenient than putting them together as they are pretty much independent. If other @open-telemetry/collector-approvers think it's worth an overarching issue and a doc, I can do it.

PR #5104, opened 2022-03-28
[pdata] Rename MetricValueType to NumberType

dmitryax:
`MetricValueType` name is confusing. It's not a type of any metric value, it's applicable to numeric values only.

This change proposes renaming of `pdata.MetricValueType` to `pdata.NumberType`.

### DIFF
--- a/model/internal/pdata/generated_metrics.go
+++ b/model/internal/pdata/generated_metrics.go
@@ -1161,14 +1161,14 @@ func (ms NumberDataPoint) SetTimestamp(v Timestamp) {
 
 // ValueType returns the type of the value for this NumberDataPoint.
 // Calling this function on zero-initialized NumberDataPoint will cause a panic.
-func (ms NumberDataPoint) ValueType() MetricValueType {
+func (ms NumberDataPoint) ValueType() NumberType {
 	switch ms.orig.Value.(type) {
 	case *otlpmetrics.NumberDataPoint_AsDouble:
-		return MetricValueTypeDouble
+		return NumberTypeDouble
 	case *otlpmetrics.NumberDataPoint_AsInt:
-		return MetricValueTypeInt
+		return NumberTypeInt
 	}
-	return MetricValueTypeNone
+	return NumberTypeNone
 }
 
 // DoubleVal returns the doubleval associated with this NumberDataPoint.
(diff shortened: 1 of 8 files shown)


model/internal/pdata/generated_metrics.go
```diff
@@ -1161,14 +1161,14 @@ func (ms NumberDataPoint) SetTimestamp(v Timestamp) {
 
 // ValueType returns the type of the value for this NumberDataPoint.
 // Calling this function on zero-initialized NumberDataPoint will cause a panic.
-func (ms NumberDataPoint) ValueType() MetricValueType {
+func (ms NumberDataPoint) ValueType() NumberType {
```
bogdandrutu, line 1164: What is the convention here? I want to make sure we follow a convention between `pdata.Value` and `pdata.Metric` for func and enum type names.

dmitryax: Type accessor functions for `pdata.Metric`, `pdata.NumberDataPoint` and `pdata.Examplar` (except for `pdata.Value`) are called based on the oneof proto field name, but there is no convention for the type name.

Type accessor func naming:
- `pdata.Metric` has `DataType()`
- `pdata.NumberDataPoint` and `pdata.Exemplar` have `ValueType()`
- But `pdata.Value` has `Type()`

I don't think `pdata.Value` should follow the same convention for the type accessor. It would be `pdata.Value.ValueType()`, I think `Value` in the func name is redundant given that `pdata.Value` itself doesn't have any other fields as `pdata.Metric`, `pdata.NumberDataPoint` and `pdata.Exemplar` have.

Enum type naming:
There is no convention for that. We just use names that makes sense.
- `pdata.Metric.DataType()` returns `pdata.MetricDataType`
- `pdata.NumberDataPoint.ValueType()` and `pdata.Exemplar.ValueType()` return `pdata.NumberType`
- `pdata.Value.Type()` returns `ValueType`
Probably in order to adopt any convention here, we would need to split `pdata.NumberType` and use different types for `pdata.NumberDataPoint.ValueType()` and `pdata.Exemplar.ValueType()`.

Please let me know WDYT.

dmitryax: Also I think the `NumberType` definitions should be moved to pdata/common

dmitryax: @bogdandrutu please let me know WDYT ^

dmitryax:
I accidentally pushed it to a branch in core which I cannot rebase anymore. Submitted another PR instead https://github.com/open-telemetry/opentelemetry-collector/pull/5227

bogdandrutu:
Should we close this and delete the branch then?

dmitryax:
Yes this can be closed, need help with removing the branch. I kept it to not lose the discussion above

Issue #4971, opened 2022-03-08
Document and provide a way to disable global state modification by Collector


mx-psi:
**Is your feature request related to a problem? Please describe.**

On initialization, the Collector modifies global state from external dependencies in a number of ways; for example (probably not an exhaustive list!):

- Modify the global gRPC logger https://github.com/open-telemetry/opentelemetry-collector/blob/52f866991b4bf4c07b371c043da03cb765b76bc2/service/internal/telemetrylogs/logger.go#L76
- Modify the global meter provider when using the OTel metrics feature flag https://github.com/open-telemetry/opentelemetry-collector/blob/f980c9ef25b1110288ab1b0a297419ec83b541a1/service/telemetry.go#L213 
- Modify the global trace provider https://github.com/open-telemetry/opentelemetry-collector/blob/52f866991b4bf4c07b371c043da03cb765b76bc2/service/collector.go#L215

All of these provide useful features, but it makes it difficult to use the Collector as a library: in the grpc logger example in particular, it can lead to data races since the logger access is not atomic, in the global providers it may interact in unexpected ways with other parts of a codebase.

 **Describe the solution you'd like**

The Collector's documentation will explicitly state any global state modifications of externally accessible structs. It also provides an option to disable these modifications. In the cases listed above, something like #4970 would work, since it could be the telemetry provider's responsibility to modify global telemetry state.

dmitryax:
@mx-psi will you be able to work on this one?

mx-psi:
Sure :)

bogdandrutu:
Removed https://github.com/open-telemetry/opentelemetry-collector/pull/5138

mx-psi:
👍 will focus just on the gRPC logger then

Issue #4970, opened 2022-03-08
Allow overriding telemetry providers


mx-psi:
**Is your feature request related to a problem? Please describe.**

When embedding the OpenTelemetry Collector on an existing application, said application may have an existing telemetry system for logging, metrics and/or tracing. With the current API, while some features of the telemetry may be configurable through the configuration provider (e.g. one can enable/disable the metrics telemetry), the telemetry can't be easily integrated into existing systems.

Currently, the logging features can be overriden since one can pass `LoggingOptions`, https://github.com/open-telemetry/opentelemetry-collector/blob/f980c9ef25b1110288ab1b0a297419ec83b541a1/service/settings.go#L64-L65 which effectively allows one to change the `zap.Core` and have arbitrary logging, but the metrics and traces providers cannot be modified apart from the config exposed in the YAML.

**Describe the solution you'd like**

A new `TelemetryFactory` can be passed to `CollectorSettings`. This provider would have a type similar to:

```go
type TelemetryFactory func(*config.ServiceTelemetry) telemetry.TelemetrySettings
```

A default `TelemetryFactory` is provided for distributions that want to keep the current behavior or build upon the current behavior. This factory can take options such as `zap.Logger` options.

The `LoggingOptions` field is eventually removed, since it's superseded by the default factory.

**Describe alternatives you've considered**

`CollectorSettings` could instead have new `MetricsOptions` and `TracingOptions` of some kind, although this seems more restrictive.

**Additional context**

I think this will allow creating the telemetry providers at the same point in the initialization as today, but care should be taken to ensure that this can be done as soon as possible.

The factory for telemetry settings should be extendable in the future for new types of telemetry that OpenTelemetry may add. This is effectively the same problem as the one we have with component factories, and we can apply the same solution here.

pmm-sumo:
I think having more flexible options here could be also helpful when implementing [own telemetry reporting](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#own-telemetry-reporting) capability of OpAMP (or any other use-case where collector's own telemetry is sent out)

bogdandrutu:
@pmm-sumo if otel-go library is not capable of supporting that (via an extra exporter for example) then we failed to implement the library sdk. I don't think for your use-case this issue is relevant.

bogdandrutu:
@mx-psi I am failing to see what the proposed solution looks like. Let's start by first define what "capabilities"/"mean" to run the collector as library.

[...]

PR #5107, opened 2022-03-28
Telemetry provider PoC

mx-psi:
**Description:**

This is a proof of concept in support of #4970.

A `TelemetryProvider` is defined to support overriding the telemetry on the Collector. A default implementation is provided, but alternatives can be provided to override telemetry providers and also to deal with #4971.

The `TelemetryProvider` interface has two methods: the `SetupTelemetry` and `ZPages`. The latter is needed since the span processor needs to be passed as part of the zpages server registration. This could be a separate, optional interface, similar to what happens now for ZPages on `component.Host`.

**Link to tracking Issue:** #4970


mx-psi:
I refactored this a bit to show how to skip GRPC logger as part of a 'test telemetry provider' and to adapt to main changes

jmacd:
FWIW https://github.com/open-telemetry/opentelemetry-collector/pull/8111#issuecomment-1654148294
I recommend we lean on the OTel global providers instead of supporting a different way to inject those dependencies.

PR #5114, opened 2022-03-29
Upgrade to latest build tools go.opentelemetry.io/build-tools

bogdandrutu:


internal/tools/go.mod
```diff
@@ -4,16 +4,16 @@ go 1.17
 
 require (
 	github.com/client9/misspell v0.3.4
-	github.com/golangci/golangci-lint v1.45.2
+	github.com/golangci/golangci-lint v1.45.0
```
Aneurysm9, line 7: Is this downgrade to deal with the modified tag in one of its dependencies?  Can that dependency version be excluded instead?

Aneurysm9 (approved):

internal/tools/go.mod
bogdandrutu: Not sure what happened, was definitely not intentional.

PR #5115, opened 2022-03-30
fix bug with nil Sum

codeboten:
A bug was caught in the prometheus tests in the contrib repository where a histogram datapoint that contained a nil sum was copied into a new datapoint, resulting in the sum being initialized to 0.

Added a test to validate the issue is fixed. I will add another issue to add this test to the generated tests for future use.


codeboten:
Skipping changelog as this bug has not been released yet.

Aneurysm9 (approved):

dmitryax (approved):

Issue #5113, opened 2022-03-29
Builder vs make gotidy output missmatch


bogdandrutu:
Investigate (and fix after) why the result of `make genotelcorecol` produces a different `cmd/otelcorecol/go.sum` than `make gotidy` on the head and no changes in `cmd/otelcorecol/go.mod`.

/cc @jpkrohling

PR #5116, opened 2022-03-30
Fix mismatch go.sum between `make gotidy` and `make genotelcorecol`

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>

**Description:** Just as the title says.

**Link to tracking Issue:** #5113 

**Testing:** Manual test by running `make gotidy` and `make genotelcorecol`


bogdandrutu (approved):

jpkrohling:
I'm pretty sure the "all" was added later to fix another problem. Could you please check git blame for this line?

rleungx:
> I'm pretty sure the "all" was added later to fix another problem. Could you please check git blame for this line?

Yes. The original PR is https://github.com/open-telemetry/opentelemetry-collector-builder/pull/50, but it seems to be unnecessary. 
The problem it's going to solve is that we need to update the modules before building which https://github.com/open-telemetry/opentelemetry-collector-builder/pull/20 solved. And https://github.com/open-telemetry/opentelemetry-collector-builder/pull/66 add `tidy` command before `download` in `GetModules`. So I think we even don't need a `download` here.

PR #5117, opened 2022-03-30
Remove Tigran since he is no longer approver/maintainer on contrib

bogdandrutu:


jpkrohling (changes requested):
He should still be there:

> In order to have more people comfortable with the release process, and in order to decrease the burden on a small number of volunteers, all core approvers are release managers from time to time

https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md

bogdandrutu:
@jpkrohling officially (because of TC role, he has that access) @tigrannajaryan should not be able to push the tags in contrib. who should do that in this case?

jpkrohling:
> It is possible that a core approver isn't a contrib approver. In that case, the release manager should coordinate with a contrib approver for the steps requiring such role, like the publishing of tags.

https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md

tigrannajaryan:
I think I should be able to do the contrib release because I have the access, but we need to figure out in the future if this rule remains since other core approvers who are not contrib approvers will not be able to. No rush for now since there is noone else like that at the moment.

jpkrohling:
I stand by the original text of the release docs: 

> In that case, the release manager should coordinate with a contrib approver for the steps requiring such role, like the publishing of tags.

tigrannajaryan:
> I stand by the original text of the release docs:
> 
> > In that case, the release manager should coordinate with a contrib approver for the steps requiring such role, like the publishing of tags.

Right, I missed that part. We are good.

PR #5130, opened 2022-03-30
Make collector flag functions public to allow collector builders to use them

dashpole:
This is the set of public changes required for https://github.com/open-telemetry/opentelemetry-collector/pull/4977.  After that, we need to wait for a release before making changes in that PR.

Part of https://github.com/open-telemetry/opentelemetry-collector/issues/4967

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 💡 Enhancements 💡
 
+- Make flag functions public so they can be used by collector builders (#5130)
+
 ### 🧰 Bug fixes 🧰
 
 ## v0.48.0 Beta
--- a/service/collector_windows.go
+++ b/service/collector_windows.go
@@ -94,7 +94,7 @@ func (s *WindowsService) Execute(args []string, requests <-chan svc.ChangeReques
 
 func (s *WindowsService) start(elog *eventlog.Log, colErrorChannel chan error) error {
 	// Parse all the flags manually.
-	if err := flags().Parse(os.Args[1:]); err != nil {
+	if err := Flags().Parse(os.Args[1:]); err != nil {
 		return err
 	}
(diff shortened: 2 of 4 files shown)


bogdandrutu (commented):
I have a philosophical question: Who is the user of the "Flags" API? Do we want to support the use-case where users are writing their own "main" func, or we support only the use-case where they use the builder?

jpkrohling:
> Do we want to support the use-case where users are writing their own "main" func, or we support only the use-case where they use the builder?

I see the builder as one user that writes its own main func. Users may use the builder once and adapt the generated source code for their needs.

bogdandrutu:
@neelayu we have an issue to allow custom providers and did 80% of that work already.. Next reason :)

jpkrohling:
> So the builder should/can generate that file as well? We should not offer that in our "distributed" modules.

People can use the builder to create their own distributions. The recommendation is to just use a manifest file and let the builder generate the sources and compile a binary, but I do see how people would be interested in just generating a bootstrap code and adapt it from there. This is why we have the option "--skip-compilation". In fact, we have something similar for the contrib repository: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/otelcontribcol

bogdandrutu:
@jpkrohling correct... but do we want to support the use-case where they manually write all the files?

jpkrohling:
> but do we want to support the use-case where they manually write all the files?

We do not want to support the conflict resolution between the generated sources and the customizations, but we want to support the source code generation, especially if users find it easier to check the manifests vs. the main.go file. As a developer of a downstream distribution, I would probably isolate my changes outside of the generated files, perhaps just adding a line or two to call my custom func defined elsewhere. Otherwise, I wouldn't see the point in generating the main func.

PR #5136, opened 2022-04-01
Update OTLP to 0.16.0

tigrannajaryan:
Removed LogRecord.Name field (was deprecated earlier)


bogdandrutu:
For reasons like https://github.com/open-telemetry/opentelemetry-collector/runs/5792037755?check_suite_focus=true we decided to try to keep pdata deprecated fields for 2 versions.

tigrannajaryan:
I removed LogRecord.Name usage from influxdb otel converter: https://github.com/influxdata/influxdb-observability/pull/61

PR #5138, opened 2022-04-01
Do not set TraceProvider to global

bogdandrutu:
The reason is because soon we will have the config for the tracing/metrics providers in the main config (already started).
When this happens then we can have users "reconfigure" that via the watch API, and re-setting the global is not allowed in otel (also not restart friendly). Because of that we should break this behavior now, and not install the global.


Aneurysm9 (approved):
I think this should already be a non-issue since the provider in the `TelemetrySettings` should be used by all components.  Definitely good to eliminate this, though.

codeboten (approved):

codeboten (commented):
Probably worth a changelog in case this breaks anyone using the global provider.

bogdandrutu:
@codeboten thanks for reminding me... I forgot to push the changes :))

PR #5144, opened 2022-04-04
Remove deprecated funcs from otlpgrpc

bogdandrutu:
Because, most likely for the moment only otlp exporter/receiver is using this package, we can remove this after just one version, no third-party deps.


codeboten (approved):
👍

Issue #1123, opened 2020-06-15
OTLP exporter status codes to retry on are incorrect


kbrockhoff:
**Describe the bug**
OTLP exporter retries sending of data when receiving back a ResourceExhausted status code even though the call will never succeed.

**Steps to reproduce**
Remove this line from correctness_test.go and then run tests.
    send_batch_size: 1024

**What did you expect to see?**
Log message reporting the problem and dropping of the message.

**What did you see instead?**
Nothing. Have to put a breakpoint in exporter/otlpexporter/exporter.go at line 116 to identity the problem.

**What version did you use?**
Version: 6dcfa475e2c0a31d891801a0992f62e889674162

**What config did you use?**
Config: See testbed/correctness/correctness_test.go

**Environment**
OS: OS X
Compiler(if manually compiled): go1.14.3 darwin/amd64

**Additional context**

bogdandrutu:
@kbrockhoff  is this resolved?

tigrannajaryan:
@kbrockhoff I see this is assigned to you. Did you intend to work on this?

kbrockhoff:
I won't have time to work on it for at least two weeks. If somebody else does not pick it up by then, I will work on it

tigrannajaryan:
@kbrockhoff OK. From the description I think what you are saying is that we should treat ResourceExhausted as a fatal error and should not retry. Is ResourceExhausted fatal in the sense that retries will never succeed?

kbrockhoff:
In my experience if the message rate is high enough to trigger throttling, you will never be able to catch up. Dropping data is the only way to restore service.

bogdandrutu:
I think this is fixed.

PR #5147, opened 2022-04-04
[exporter/otlp] Retry RESOURCE_EXHAUSTED only if the server returns RetryInfo

svrakitin:
**Description:**

This makes us retry on`ResourceExhausted` only if retry info is provided. It also makes code a bit more explicit.

In my case it caused an issue in production where upstream denied requests above `max_recv_msg_size` and we kept retrying.

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector/issues/1123


Aneurysm9:
Can we instead change that case to determine whether the retry info has been provided in the response?  The spec [requires retry of transient errors](https://opentelemetry.io/docs/reference/specification/protocol/exporter/#retry) and says that `RESOURCE_EXHAUSTED` [`SHOULD` be treated as transient](https://opentelemetry.io/docs/reference/specification/protocol/otlp/#otlpgrpc-response).  I think it makes sense to treat it as transient if retry info is provided and permanent otherwise.

svrakitin:
@Aneurysm9 Makes sense, gonna work on this.

svrakitin:
@Aneurysm9 I updated the PR to separate "always-retryable" and "never-retryable" error codes. Does it makes sense? Meanwhile I am going to write some tests for that as those seem to be missing.

Aneurysm9:
> I updated the PR to separate "always-retryable" and "never-retryable" error codes. 

This seems like a good approach.

tigrannajaryan (dismissed):
I do not understand what this change does. A couple questions:
- Is the [OTLP specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-response) correct or something needs to be changed there?
- If the specification is correct which part of the specification is currently violated by the implementation in this repo that this PR fixes?

(Blocking temporarily to avoid accidental merging until all is clarified).

svrakitin:
@tigrannajaryan This implementation effectively treats `ResourceExhausted` as transient (retryable) only if upstream signals backpressure. Otherwise, it will be treated as permanent.

It doesn't violate the spec as this only modifies the client behavior to drop data immediately if `ResourceExhausted` can't be retried. I would suggest to update the spec to recommend to retry on `ResourceExhausted` only if upstream signals backpressure.

tigrannajaryan:
> It doesn't violate the spec as this only modifies the client behavior to drop data immediately if `ResourceExhausted` can't be retried.

This **does** appear to contradict what the spec says:
```
The client SHOULD interpret gRPC status codes as retryable or not-retryable according to the following table:

|gRPC Code|Retryable?|
|---------|----------|
|RESOURCE_EXHAUSTED|Yes|
```

So, to make it clear, you are saying that RESOURCE_EXHAUSTED should not be an unconditional "Yes" in that table. It should be conditional:
- If the response includes a `RetryInfo` then RESOURCE_EXHAUSTED is retryable.
- If the response does not include a `RetryInfo` then RESOURCE_EXHAUSTED is not retryable.

Is this correct?

tigrannajaryan:
So, I think we need to update spec to reflect this. I believe this can be considered a bug in the specification, so it is an allowed change.

svrakitin:
@tigrannajaryan Should I prepare the change to the spec before proceeding with this?

exporter/otlpexporter/otlp.go
```diff
@@ -141,38 +141,49 @@ func processError(err error) error {
 		return nil
 	}
 
-	// Now, this is this a real error.
+	// Now, this is a real error.
 
-	if !shouldRetry(st.Code()) {
-		// It is not a retryable error, we should not retry.
+	// Check if error is permanent, in which case we shouldn't ever retry.
```
tigrannajaryan, line 146: I have a hard-time reading the logic in lines 147-163.

Can we refactor this to avoid using a pair of shouldNeverRetry/shouldAlwaysRetry functions and instead show the entirety of the logic in one place?

Presumably all we need to do is modify the shouldRetry function and modify its behavior for the ResourceExhausted case. No other changes should be necessary if [this is correct](https://github.com/open-telemetry/opentelemetry-collector/pull/5147#issuecomment-1091952940).

tigrannajaryan:
> @tigrannajaryan Should I prepare the change to the spec before proceeding with this?

Yes, that would be great.

exporter/otlpexporter/otlp.go
svrakitin: Alright, gonna make it simpler

svrakitin:
I made the code follow the new addition to the spec and added the test to check `RetryInfo`/no-`RetryInfo` cases.

tigrannajaryan (commented):
Much better now, thanks.
Let's wait for https://github.com/open-telemetry/opentelemetry-specification/pull/2480 to be merged before merging this.

tigrannajaryan (approved):

bogdandrutu:
@svrakitin please resolve conflicts, and I will merge

bogdandrutu:
@svrakitin unit tests are failing

svrakitin:
@bogdandrutu Yeah, I didn't notice that the tests now use `go.uber.org/atomic`. Updated the test.

PR #5148, opened 2022-04-04
[model/semconv] Add `generated_` prefix to generated files

dmitryax:
For consistency with other generated filenames across the collector repos. This also helps to distinguish files that has to be manually created for each semconv release.


model/semconv/v1.6.1/generated_trace.go
```diff
@@ -14,7 +14,7 @@
 
 // Code generated from semantic convention specification. DO NOT EDIT.
 
-package semconv // import "go.opentelemetry.io/collector/model/semconv/v1.6.1"
```
bogdandrutu, line 17: Should this be kept?

dmitryax: Looks like the old version of the porto tool used to keep this line in generated files and now it's removed. I'll make sure this PR doesn't change it

dmitryax: Updated

Aneurysm9 (approved):

PR #5151, opened 2022-04-05
Deprecate configunmarshaler package, move it to internal

bogdandrutu:
**Folllowup:**

* When the deprecated funcs/types are removed the internal package can be moved to service/internal.

**Motivation:**

* This package is removed because with the latest addition of the `service.ConfigProvider` the usecase to change the unmarshaled config.Config can be achieved by wrapping/implementing that interface, so no clear use-case for this. In the future we can expose it again if we have good reasons.
* During the review of another PR, this was mentioned as something some approvers/maintainers were concerned about what to do with this package. See https://github.com/open-telemetry/opentelemetry-collector/pull/4608#pullrequestreview-890442136

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/4605


config/configunmarshaler/unmarshaler.go
```diff
@@ -15,12 +15,14 @@
...
-	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/internal/configunmarshaler"
 )
 
-// ConfigUnmarshaler is the interface that unmarshalls the collector configuration from the config.Map.
-type ConfigUnmarshaler interface {
-	// Unmarshal the configuration from the given parser and factories.
-	Unmarshal(v *config.Map, factories component.Factories) (*config.Config, error)
-}
+// Deprecated: if you need to update the config.Config implement custom (or wrap) service.ConfigProvider.
```
codeboten, line 21: Please add version

internal/configunmarshaler/defaultunmarshaler.go
```diff
@@ -75,17 +81,17 @@ type configSettings struct {
 	Service    map[string]interface{}                        `mapstructure:"service"`
 }
 
-type defaultUnmarshaler struct{}
+type DefaultUnmarshaler struct{}
```
codeboten, line 84: Is this used outside of this package?

config/configunmarshaler/unmarshaler.go
bogdandrutu: Done.

internal/configunmarshaler/defaultunmarshaler.go
bogdandrutu: Nope, initially I did not define the interface in the internal and I had to return it. But I figured that I have to define the interface since it is expose as argument.

codeboten (approved):

bogdandrutu:
Holding this until https://github.com/open-telemetry/opentelemetry-collector/pull/4936 is merged and I will merge the changes in https://github.com/open-telemetry/opentelemetry-collector/pull/5155 into this.

bogdandrutu:
@codeboten PTAL.

service/config_provider.go
```diff
@@ -92,8 +92,9 @@ type ConfigProviderSettings struct {
 	// MapConverters is a slice of config.MapConverterFunc.
 	MapConverters []config.MapConverterFunc
 
-	// The configunmarshaler.ConfigUnmarshaler to be used to unmarshal the config.Map into config.Config.
-	// It is required to not be nil, use configunmarshaler.NewDefault() by default.
+	// Deprecated: [v0.49.0] because providing custom ConfigUnmarshaler is not necessary since users can wrap/implement
+	// ConfigProvider if needed to change the resulted config. This functionality will be kept for at least 2 minor versions,
+	// and if nobody express a need for it will be removed.
```
codeboten, line 97:
```suggestion
	// Deprecated: [v0.49.0] providing custom ConfigUnmarshaler is not necessary since users can wrap/implement
	// ConfigProvider if needed to change the resulted config. This functionality will be kept for at least 2 minor versions,
	// and if no one expresses a need for it will be removed.
```

codeboten (approved):

PR #5162, opened 2022-04-06
[pdata] Extract model/internal/pdata package into model/internal

dmitryax:
To avoid confusion with `model/pdata` package and `pdata` module which we will likely have instead of `model` based on the latest discussion in https://github.com/open-telemetry/opentelemetry-collector/pull/5087


bogdandrutu (approved):
this is an internal only change +1

bogdandrutu:
I think we may have an issue, because I remember I asked you to do this to remove the confusion between the current public vs internal package.

dmitryax:
> I think we may have an issue, because I remember I asked you to do this to remove the confusion between the current public vs internal package.

What issue do you mean? Yes I remember that you asked this.

PR #5168, opened 2022-04-08
Split `model` by signal type and move it to the new `pdata` module

dmitryax:
Split all `pdata` related code by type and move it from `model` to the new module `pdata`.

- `model/pdata` and `model/otlp` are moved to `pdata/plog`, `pdata/pmetric` and `pdata/ptrace`.
- `model/otlpgrpc` is moved to `pdata/plogotlp`, `pdata/pmetricotlp` and `pdata/ptraceotlp`.

Now all the APIs in `model` except for `model/semconv` are deprecated.

This is another PR following the latest decisions made in https://github.com/open-telemetry/opentelemetry-collector/pull/5087 and https://github.com/open-telemetry/opentelemetry-collector/issues/5027.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/4832 and https://github.com/open-telemetry/opentelemetry-collector/issues/5027


tigrannajaryan:
2 questions:
- pdata is now one Go module. I thought we planned to split into 4 modules, one common and 3 for signals types. Was this decision changed or it is coming in a future PR?
- What is the plan for model/semconv? Is model going to remain a Go module with just semconv package?

dmitryax:
@tigrannajaryan 

> pdata is now one Go module. I thought we planned to split into 4 modules, one common and 3 for signals types. Was this decision changed or it is coming in a future PR?

The plan is the same. New modules for each signal + `pdata/pcommon` will be added later. They depend on `pdata/internal`, so we need to make `pdata` or `pdata/internal` a module anyway. I think we can keep additional `pdata` module for that.

> What is the plan for model/semconv? Is model going to remain a Go module with just semconv package?

I think we agreed on moving it from `/model/semconv` to another separate `/semconv` module. This will be done separately.

pdata/internal/semconv/template.j2
```diff
@@ -0,0 +1,87 @@
+{%- macro to_go_attr_type(type, val) -%}
```
tigrannajaryan, line 1: Is this a copy of template.j2 we have elsewhere?

tigrannajaryan (approved):

pdata/internal/semconv/template.j2
dmitryax: Oh, this was copied by accident. Thanks for catching it

dmitryax:
Look like something wrong with coverage calculation. I verified that the coverage is the same between the moved parts

dmitryax:
> @dmitryax look into updating this file https://github.com/open-telemetry/opentelemetry-collector/blob/main/.codecov.yml

Oh that's why! Thanks for pointing on this

pdata/ptrace/alias.go
```diff
@@ -0,0 +1,62 @@
...
+// This file contains aliases for trace data structures.
+
+// TracesMarshaler is an alias for internal.TracesMarshaler interface.
+type TracesMarshaler = internal.TracesMarshaler
+
+// TracesUnmarshaler is an alias for internal.TracesUnmarshaler interface.
+type TracesUnmarshaler = internal.TracesUnmarshaler
+
+// TracesSizer is an alias for internal.TracesSizer interface.
+type TracesSizer = internal.TracesSizer
```
bogdandrutu, line 28: Question: Should we remove for these the prefix "Traces"?

pdata/plog/json.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/gogo/protobuf/jsonpb"
+
+	"go.opentelemetry.io/collector/pdata/internal"
+	otlplogs "go.opentelemetry.io/collector/pdata/internal/data/protogen/logs/v1"
+	"go.opentelemetry.io/collector/pdata/internal/otlp"
+)
+
+// NewJSONLogsMarshaler returns a LogsMarshaler. Marshals to OTLP json bytes.
+func NewJSONLogsMarshaler() LogsMarshaler {
```
bogdandrutu, line 28: I would consider to remove the "Logs" from the `NewJSONLogsMarshaler`? What do you think? Same for proto.

pdata/plog/pb.go
```diff
@@ -0,0 +1,59 @@
...
+
+package plog // import "go.opentelemetry.io/collector/pdata/plog"
+
+import (
+	"go.opentelemetry.io/collector/pdata/internal"
+	otlplogs "go.opentelemetry.io/collector/pdata/internal/data/protogen/logs/v1"
+)
+
+// NewProtobufLogsMarshaler returns a LogsMarshaler. Marshals to OTLP binary protobuf bytes.
+func NewProtobufLogsMarshaler() LogsMarshaler {
```
bogdandrutu, line 23: I think we should shortcut to just `Proto` instead of `Protobuf` or update the `plogotlp.LogsRequest` to use `MarshalProtobuf`.

pdata/ptrace/alias.go
dmitryax: Yes, I think this prefix isn't needed given that we still have `TracesSize` method in the interface

pdata/plog/json.go
dmitryax: Makes sense. In that case we also rename interfaces `plog.LogsMarshaler` and `plog.LogsUnmarshaler` to `plog.Marshaler` and `plog.Unmarshaler`, but keep names of all the methods like `MarshalLogs` and `UnmarshalLogs`. Same goes to `plog.Sizer` as pointed in the discussion above.

How does it sound?

pdata/plog/pb.go
dmitryax: Yeah, good catch. With all the discussions above this signature would become:

```golang
func NewProtoMarshaler() Marshaler {
```

pdata/plog/json.go
bogdandrutu: I like it. In this PR was referring to the `NewJSONLogsMarshaler` -> `NewJSONMarshaler`

bogdandrutu: But yes, you mentioned all the changes I was thinking about in your comments :)

pdata/plog/plogotlp/logs.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (lr LogsRequest) Logs() plog.Logs {
+	return internal.LogsFromOtlp(&otlplogs.LogsData{ResourceLogs: lr.orig.ResourceLogs})
+}
+
+// LogsClient is the client API for OTLP-GRPC Logs service.
+//
+// For semantics around ctx use and closing/ending streaming RPCs, please refer to https://godoc.org/google.golang.org/grpc#ClientConn.NewStream.
+type LogsClient interface {
```
dashpole, line 121: should this just be `plogotlp.Client` (and `plogotlp.Server`), similar to other changes suggested?
Same question for other signals.

dmitryax: This sounds reasonable to me. I think we can rename `plogotlp.Client` and `plogotlp.Server` but we should keep `plogotlp.LogsResponse` and `plogotlp.LogsRequest` in this package as is. Let me know if this is what you meant.

pdata/plog/json.go
dmitryax: Applied this one and other suggestions.

pdata/plog/plogotlp/logs.go
dashpole: I would also lean towards `plogotlp.Request` and `plogotlp.Response`, but I don't feel quite as strongly about that.

bogdandrutu: I like dropping the "Logs" prefix for Request/Response.

dmitryax: I was thinking that removing `Logs` prefix from `Request` and `Response` would leave it without enough connection to `plog.Logs`, but probably `Request.SetLogs` and `Request.Logs` is enough. So I'm ok with removing prefix as well.

bogdandrutu: They will always be "referenced" as "plogotlp.Request" when using the type (based on golang rule).

dashpole (approved):

bogdandrutu (approved):

dmitryax:
🎉

PR #5181, opened 2022-04-11
Small nits in bounded memory, unnecessary atomics and pointers

bogdandrutu:
* Capacity is immutable, so access no longer atomic.
* The items channel is immutable, so no reason to use pointer.

### DIFF
--- a/exporter/exporterhelper/internal/bounded_memory_queue.go
+++ b/exporter/exporterhelper/internal/bounded_memory_queue.go
@@ -28,51 +28,47 @@ import (
 // channels, with a special Reaper goroutine that wakes up when the queue is full and consumers
 // the items from the top of the queue until its size drops back to maxSize
 type boundedMemoryQueue struct {
-	workers       int
 	stopWG        sync.WaitGroup
 	size          *uatomic.Uint32
-	capacity      *uatomic.Uint32
 	stopped       *uatomic.Uint32
-	items         *chan interface{}
+	items         chan interface{}
 	onDroppedItem func(item interface{})
 	factory       func() consumer
 	stopCh        chan struct{}
+	capacity      uint32
 }
 
 // NewBoundedMemoryQueue constructs the new queue of specified capacity, and with an optional
 // callback for dropped items (e.g. useful to emit metrics).
 func NewBoundedMemoryQueue(capacity int, onDroppedItem func(item interface{})) ProducerConsumerQueue {
-	queue := make(chan interface{}, capacity)
 	return &boundedMemoryQueue{
 		onDroppedItem: onDroppedItem,
-		items:         &queue,
+		items:         make(chan interface{}, capacity),
 		stopCh:        make(chan struct{}),
-		capacity:      uatomic.NewUint32(uint32(capacity)),
 		stopped:       uatomic.NewUint32(0),
 		size:          uatomic.NewUint32(0),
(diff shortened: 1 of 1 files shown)


dmitryax (approved):

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -42,14 +41,13 @@ type boundedMemoryQueue struct {
...
-	queue := make(chan interface{}, capacity)
 	return &boundedMemoryQueue{
 		onDroppedItem: onDroppedItem,
-		items:         &queue,
+		items:         make(chan interface{}, capacity),
 		stopCh:        make(chan struct{}),
-		capacity:      uatomic.NewUint32(uint32(capacity)),
-		stopped:       uatomic.NewUint32(0),
-		size:          uatomic.NewUint32(0),
+		size:          int32(0),
```
mx-psi, line 48: I understand why we remove the use of atomics for `capacity`, but why are we removing the use of `go.uber.org/atomic` since we still need atomic access? Atomics in Go have [a number of limitations](https://pkg.go.dev/sync/atomic#pkg-note-BUG) and it's very easy to hit some of them without using a safe wrapper like the one from Uber. 

In particular, it is very easy to cause panics on ARM 32 bit if a struct field is accessed atomically and it is not the first field in the struct (I don't think it happens here because it needs to be a 64 bit word, but it's very easy to make that mistake).

bogdandrutu: Sorry, I don't understand how the uber wrapper helps with that. The guarantee is that "first field of an allocated struct" not when embedding the struct. So if you use the uber wrapper wrong (not as a pointer) you have the same problem.

Though you are correct that because of the "NewX" func in uber is less likely that users will not use pointers. I do understand that using a simple pointer like `*int64` has the same guarantees, so we should make a decision about what do we recommend in this repo.

I am ok to recommend "uber/atomic" but we should be consistent and use it everywhere, one of my motivation was also because this is the only place where we use that.

mx-psi: > So if you use the uber wrapper wrong (not as a pointer) you have the same problem.

I think the `NewX` func nudges you to always use it as a pointer (or at least that's what it does for me 😄), but you are right, uber is a safer wrapper, not safe. I prefer using uber/atomic because you don't have that nudge otherwise, and running linters to check for aligned access is hard (you need to run on arm 32 bits, which we can't do today on CI).

bogdandrutu: I think then we should document that we recommend using uber atomics in this repo (if others feel the same, I will open an issue to discuss, I am fine with that). I will revert this PR to non controversial changes (capacity -> non-atomic; channel does not need to be pointer).

hanjm: I think a lint is useful to keep consistent, like prometheus does.https://github.com/prometheus/prometheus/blob/main/.golangci.yml#L29

bogdandrutu: @hanjm please comment on the issue https://github.com/open-telemetry/opentelemetry-collector/issues/5200

mx-psi (approved):

PR #5197, opened 2022-04-12
Fix translation from otlp.Request to pdata representation

bogdandrutu:
The problem that this PR tries to fix, is that changes in the traces object return by the otlp.Request are not reflected in the Request.

The solution is to wrap the request proto instead of the data proto, since for the marshalers we don't have an equivalent object that shares the same problem as Request.


dmitryax (approved):
LGTM

PR #5198, opened 2022-04-12
Change config.Retrieved to allow retrieving different types, map, string, etc.

bogdandrutu:


mx-psi:
I will have a look tomorrow morning :) I saw this and it looks reasonable at a first glance, but I haven't had time to give it a proper look

mx-psi (commented):
This makes sense to me, I think my only objection is that it can be a bit cryptic to figure out how it works by just looking at the code, so we probably want more docs/explicitness.

The detail about the `""` path in particular should be documented; I wonder if we should have a `Path` type alias with an `RootPath = ""` constant and make it `MergeTo(dest *Map, destPath Path)`

bogdandrutu:
@mx-psi thanks for the feedback. I did revert the change to make `Retrieved` interface, since it was not backwards compatible (first deprecate all the fields and expose the funcs, then will remove the fields and can transform in interface). I also change the `MoveTo` to not accept a destination (since in lots of places we need to call this), and will add later a `MoveToAt` that accepts a destination as well and will document at that moment the behavior of the path (if we need).

config/mapconverter/expandmapconverter/expand_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"go.opentelemetry.io/collector/config"
-	"go.opentelemetry.io/collector/config/mapprovider/filemapprovider"
+	"go.opentelemetry.io/collector/config/configtest"
```
bogdandrutu, line 26: Changes in this file are extracted as https://github.com/open-telemetry/opentelemetry-collector/pull/5229 which will be merged first.

service/config_provider.go
```diff
@@ -144,21 +144,22 @@ func (cm *configProvider) Get(ctx context.Context, factories component.Factories
...
 	}
 
-	ret, err := cm.mergeRetrieve(ctx)
+	retMap, closer, err := cm.mergeRetrieve(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("cannot retrieve the configuration: %w", err)
 	}
-	cm.closer = ret.CloseFunc
+
+	cm.closer = closer
```
mx-psi, line 152: Why not `ret.Close` instead? (so that you don't have to change the return type of `mergeRetrieve`)

config/mapprovider.go
```diff
@@ -81,20 +81,39 @@ type ChangeEvent struct {
...
+// * CloseFunc specifies a function to be invoked when the configuration for which it was
+//   used to retrieve values is no longer in use and should close and release any watchers
+//	 that it may have created.
+func NewRetrievedFromMap(cfgMap *Map, close CloseFunc) Retrieved {
+	return Retrieved{Map: cfgMap, CloseFunc: close}
+}
+
+// MergeTo merges the retrieved configuration into the given Map.
+// Note that the retrieved configuration must not be modified.
+func (r Retrieved) MergeTo(dest *Map) error {
```
mx-psi, line 101: Should we handle the case where `dest` is nil explicitly and return an error?

mx-psi, line 101: Sanity check: suppose we have

```
exporters:
   otlp:
     endpoint: ${provider:uri}
```

How would this be supported? IIUC the intention is to have the caller do something like:

```go
m := SomeMap()
r := SomeRetrieved()
subsection, _ := m.Sub("exporters::otlp") // Parent of section?
r.MergeTo(subsection)
```

Is that right? I am not sure I see how this work now that the path has been removed from the interface

bogdandrutu: What I am proposing is to have in the end both "MergeTo" and "MergeToAt". This is to limit the use of the "special case" of empty string path.

mx-psi: Ah, I see! Okay, then let's go with this first, yes :)

mx-psi: > We don't do that in lots of other places... 

That's fair, I was thinking about the close function case, but it seems like we mostly only do this for function pointers.

> Let's have an issue about validating the input arguments, and make a decision for the entire repo?

Sounds good, I am now doubtful about what the best approach is but it seems like a good idea to have the discussion in any case.

config/mapprovider.go
```diff
@@ -81,20 +81,39 @@ type ChangeEvent struct {
...
 
-// CloseFunc a function to close and release any watchers that it may have created.
+// NewRetrievedFromMap returns a new Retrieved instance that contains a Map data.
+// * cfgMap the Map that will be merged to the given map in the MergeTo.
+// * CloseFunc specifies a function to be invoked when the configuration for which it was
+//   used to retrieve values is no longer in use and should close and release any watchers
+//	 that it may have created.
+func NewRetrievedFromMap(cfgMap *Map, close CloseFunc) Retrieved {
+	return Retrieved{Map: cfgMap, CloseFunc: close}
+}
```
bogdandrutu, line 97: @mx-psi last question here, should we go with "close" as "RetrievedOption" because literally is optional here? so we will define a "Option" and we will have `WithClose`, instead of asking users to pass nil when no close needed?

bogdandrutu:
@mx-psi per the feedback from the expand PR (https://github.com/open-telemetry/opentelemetry-collector/pull/4742), I changed this to "AsMap" instead of the merge. The problem with the "Merge" like patter, is because it is impossible to merge (parse into a location) if the expanded value is part of an array, so in order to not have to suffer and limit the functionality to not work on arrays, I decided to go with the pattern where we return the value.

config/mapprovider.go
mx-psi: Sounds better, yes, it's also how we pass the function pointers on the component helpers, so we already have the precedent there.

mx-psi:
> @mx-psi per the feedback from the expand PR (#4742), I changed this to "AsMap" instead of the merge. The problem with the "Merge" like patter, is that is impossible to merge (parse into a location) if the expanded value is part of an array, so in order to not have to suffer and limit the functionality to not work on arrays, I decided to go with the pattern where we return the value.

Retrieved would have a `AsMap` and... `MergeToAt` methods? Also, I don't think I understand now the arrays issue and it would help if you can share an example (here or in the other PR) about what the issue is.

mx-psi:
Ah, I think I understand now, let me spell it out loud to confirm. Say we have:

```yaml
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  nop

processors:
  transform:
    traces:
      queries:
        -  ${env:MY_QUERY}
        - keep_keys(resource.attributes, "service.name", "service.namespace", "cloud.region")
        - set(name, attributes["http.route"])
service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [transform]
      exporters: [nop]
```

We then don't have a way to specify the path to the environment variable, we can't do something like `processors::transform::queries::[0]`. Is that what you are pointing at?

bogdandrutu:
Indeed, unless we invent a path format like you proposed, we cannot reference that env variable in the array. We can also go that path, but I feel is too hacky

mx-psi:
@bogdandrutu How does something like this (naming issues aside) sound?

```go

type Segment {
  key string
  index int
}

func Key(key string) Segment {return Segment{key: key, index: -1}}
func Index(idx int) Segment {return Segment{key: "", index: idx}}

type Path struct{
  segments []string
}

func NewPath(segments ...Segment) Path {return Path{segments}}
```

and then have

```go
func (r Retrieved) MergeToAt(dest *Map, path Path) { 
  // ...
}
```

We don't need to support `Index` now and just have the `Key` function for now. Something like this 
1. would allow us to support it in the future and
2. would avoid having to return `interface{}`.

----

There is still the problem of something like:

```
processors:
  transform:
    traces:
      queries:
        - keep_keys(resource.attributes, ${env:SOME_ENV_VAR}, "service.namespace", "cloud.region")
```

which I am not sure how it will be supported, even if solving the array problem.

bogdandrutu:
The last example is a substring in a string from what I understand. With the return of interface... We may be able to solve that, since we can check if it is a value that we can stringify and concat it. With the merge we need to have a way to say merge in this string between this index and that index

mx-psi:
> The last example is a substring in a string from what I understand.

Yes, that's it.

> I am curious about this requirement would avoid having to return interface{}. we already return and accept interface{} in the [Map](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configmap.go#L84) (and all the similar concepts do the same). Why is that so important?

So this was mentioned by @Aneurysm9 on https://github.com/open-telemetry/opentelemetry-collector/pull/4742#issuecomment-1024455907. I agree that whenever possible avoiding `interface{}` is a good idea: having explicit types allows us (and potential future users) to catch some kind of errors at compile time instead of at runtime. I don't know if Anthony had some other benefit in mind.

I think the alternatives have been explored here and at #4742 at length and in my mind, unless the "substring in a string" case is handled in some special way, it seems like the only one that does not seem prohibitively complex or limited is returning an `interface{}` indeed.

mx-psi:
> > having explicit types allows us (and potential future users) to catch some kind of errors at compile time instead of at runtime.
> 
> To be honest, because we need to unmarshal these values into the config, there is no way you will catch errors at compilation, unless I am missing something. Also the return value depends on external sources like what you have in zookeeper, voltdb, etc....

That is a fair point. I still feel like it's generally better to have explicit types but because of our discussion above, I am happy with the `interface{}` solution since the alternatives seem too complex. 

The only missing thing in my mind to merge this PR is to deal with the Close function as an option from https://github.com/open-telemetry/opentelemetry-collector/pull/5198#discussion_r853368828

bogdandrutu:
@mx-psi done, PTAL one more time if you have time before your day ends :)

PR #5203, opened 2022-04-13
Update BoundedMemoryQueue to consume remaining items on close

### LINKED ISSUE #5110: Question about shutdown the queuedRetrySender
Hi! We have a question about shutdown the `queuedRetrySender`.

In https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry_inmemory.go#L147 comments, it seems will drain the queue before shutdown the `queuedRetrySender`. 
```
func (qrs *queuedRetrySender) shutdown() {
	// Cleanup queue metrics reporting
	if qrs.cfg.Enabled {
		_ = globalInstruments.queueSize.UpsertEntry(func() int64 {
			return int64(0)
		}, metricdata.NewLabelValue(qrs.fullName))
	}

	// First Stop the retry goroutines, so that unblocks the queue numWorkers.
	close(qrs.retryStopCh)

	// Stop the queued sender, this will drain the queue and will call the retry (which is stopped) that will only
	// try once every request.
	if qrs.queue != nil {
		qrs.queue.Stop()
	}
}
```
But in https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/bounded_memory_queue.go#L83, while the `select` run into the `case <-q.stopCh`, it will return and then the items in queue will not be consumed by `itemConsumer`. Are we understand correctly or missing some key points？
```
func (q *boundedMemoryQueue) StartConsumers(num int, callback func(item interface{})) {
	factory := func() consumer {
		return consumerFunc(callback)
	}
	q.workers = num
	q.factory = factory
	var startWG sync.WaitGroup
	for i := 0; i < q.workers; i++ {
		q.stopWG.Add(1)
		startWG.Add(1)
		go func() {
			startWG.Done()
			defer q.stopWG.Done()
			itemConsumer := q.factory()
			queue := *q.items
			for {
				select {
				case item, ok := <-queue:
					if ok {
						q.size.Sub(1)
						itemConsumer.consume(item)
					} else {
						// channel closed, finish worker
						return
					}
				case <-q.stopCh:
					// the whole queue is closing, finish worker
					return
				}
			}
		}()
	}
	startWG.Wait()
}
```

_Originally posted by @ralphgj in https://github.com/open-telemetry/opentelemetry-collector/discussions/5088_

TylerHelmuth:
**Description:** 
This PR updated the BoundedMemoryQueue to consume any remaining items on shutdown.  The timing of the colsure of the items queue during `Stop()` has been moved up so that items in the queue can be consumed in a non-blocking way.

**Link to tracking Issue:**
Fixes #5110 

**Testing:**
Unit Tests


exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -129,9 +133,9 @@ func (q *boundedMemoryQueue) Produce(item interface{}) bool {
 // and releases the items channel. It blocks until all consumers have stopped.
 func (q *boundedMemoryQueue) Stop() {
 	q.stopped.Store(1) // disable producer
+	close(q.items)
 	close(q.stopCh)
```
bogdandrutu, line 137: I think we should just simply remove entirely the `stopCh`, and rely on the line 77 to exit for the worker. That is probably the cleanest.

bogdandrutu, line 137: I think we should just simply remove entirely the `stopCh`, and rely on the line 77 to exit for the worker. That is probably the cleanest.

TylerHelmuth: Ok, I can make that change.

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -95,6 +95,52 @@ func TestBoundedQueue(t *testing.T) {
...
+// only after Stop will mean the consumers are still locked while
+// trying to perform the final consumptions.
+func TestShutdownWhileNotEmpty(t *testing.T) {
+	q := NewBoundedMemoryQueue(10, func(item interface{}) {})
+
+	consumerState := newConsumerState(t)
+
+	q.StartConsumers(1, func(item interface{}) {
+		consumerState.record(item.(string))
+		//
```
bogdandrutu, line 111: Empty comment?

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -66,21 +64,11 @@ func (q *boundedMemoryQueue) StartConsumers(numWorkers int, callback func(item i
...
-					}
-				case <-q.stopCh:
-					// the whole queue is closing, finish worker
-					return
-				}
+			for item := range q.items {
+				q.size.Sub(1)
+				itemConsumer.consume(item)
 			}
+			return
```
bogdandrutu, line 71: unnecessary return statement at the end of the func

bogdandrutu (approved):

bogdandrutu:
@TylerHelmuth please add a changelog entry.

PR #5226, opened 2022-04-14
Update pdata godoc strings

dmitryax:
- Use copy of the godoc comments from pdata/internal instead of mentioning aliases in pdata/p<signal> public API docs.
- Update godoc strings in both pdata/p<signal> and pdata/internal to keep them consistent.

### DIFF
--- a/pdata/internal/cmd/pdatagen/internal/base_slices.go
+++ b/pdata/internal/cmd/pdatagen/internal/base_slices.go
@@ -111,11 +111,18 @@ func fillTest${structName}(tv ${structName}) {
 	}
 }`
 
-const commonSliceAliasTemplate = `// ${structName} is an alias for ${refPackage}.${structName} struct.
-${extraStructComment}type ${structName} = ${refPackage}.${structName}
+const commonSliceAliasTemplate = `// ${structName} logically represents a slice of ${elementName}.
+//
+// This is a reference type. If passed by value and callee modifies it, the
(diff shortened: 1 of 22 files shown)


model/pdata/generated_common_alias.go
```diff
@@ -19,18 +19,31 @@ package pdata
...
 import "go.opentelemetry.io/collector/pdata/pcommon"
 
-// InstrumentationScope is an alias for pcommon.InstrumentationScope struct.
+// InstrumentationScope is a message representing the instrumentation scope information.
+//
+// This is a reference type, if passed by value and callee modifies it the
+// caller will see the modification.
+//
+// Must use NewInstrumentationScope function to create new instances.
+// Important: zero-initialized instance is not valid for use.
```
bogdandrutu, line 28: Don't need to update the deprecated once.

dmitryax: This is generated code. It would require some effort to split the generator to produce different godoc strings between the old and new aliases.

bogdandrutu: I think you can make the "generator" no longer generate the `model/pdata` directory. It is gone :)

dmitryax: You're right. This is easy to do :)

dmitryax: done

bogdandrutu (approved):

PR #5227, opened 2022-04-15
[pdata] Split `pmetric.MetricValueType` into two more specific types

dmitryax:
Split `pmetric.MetricValueType` into `NumberDataPointValueType` and `ExemplarValueType`

- `NumberDataPoint.ValueType()` now returns a `NumberDataPointValueType` that can be `NumberDataPointValueTypeInt` or `NumberDataPointValueTypeDouble`.

- `ExemplarValueType` now returns a `ExemplarValueType` that can be `ExemplarValueInt` or `ExemplarValueDouble`.

This change doesn't provide backward compatibility. All usage of the old types has to be replaced with the upgrade.

Updated PR from https://github.com/open-telemetry/opentelemetry-collector/pull/5104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Remove pdata deprecated funcs from 2 versions (v0.48.0) ago. (#5219)
 - Remove non pdata deprecated funcs/structs (#5220)
+- Split `pmetric.MetricValueType` into `NumberDataPointValueType` and `ExemplarValueType` (#5004)
 
 ### 🚩 Deprecations 🚩
 
--- a/internal/otlptext/databuffer.go
+++ b/internal/otlptext/databuffer.go
@@ -100,9 +100,9 @@ func (b *dataBuffer) logNumberDataPoints(ps pmetric.NumberDataPointSlice) {
 		b.logEntry("StartTimestamp: %s", p.StartTimestamp())
 		b.logEntry("Timestamp: %s", p.Timestamp())
 		switch p.ValueType() {
-		case pmetric.MetricValueTypeInt:
+		case pmetric.NumberDataPointValueTypeInt:
 			b.logEntry("Value: %d", p.IntVal())
-		case pmetric.MetricValueTypeDouble:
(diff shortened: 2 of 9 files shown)


bogdandrutu:
Curios about the package change... what is the motivation?

dmitryax:
It seems to me like a more generic type that potentially can be applied to any numeric field in other (new) signals. But I'm open to other opinions

bogdandrutu:
For "Exemplar" is the type of the "Measurement" correct? Can be int64/float64, maybe in the future `enum` or even `string`. For the "NumberDataPoint" that is kind of guarantee to be a number since used for Sum aggregation.

1. Should we have 2 different types?
2. It feels a bit overkill to see this as generic.

dmitryax:
@bogdandrutu makes sense. If we can anticipate a change in exemplars, it's better to have two different types instead of trying to come up with a generic name for both of them.

If we want to be consistent, the naming should be the following:

- `func (ms NumberDataPoint) ValueType() NumberDataPointValueType` where `NumberDataPointValueType` can be `NumberDataPointValueTypeInt` or `NumberDataPointValueTypeDouble`.

- `func (ms Exemplar) ValueType() ExemplarValueType` where `ExemplarValueType` `ExemplarValueTypeInt` or `ExemplarValueTypeDouble`.

`NumberDataPointValueTypeInt` seems a bit too verbose but any other shorter name breaks the consistency and makes the name unnecessary more generic. So I think we can stick with these name.

WDYT?

bogdandrutu:
I am fine with longer names for consistency reasons.

dmitryax:
Updated the PR. 

I believe that changing all the functions using these types through a 2-step backward compatible way would be an overkill and will be much more complicated for users than updating the types themselves once. So I think we can afford this as another exception to the backward compatibility rule.

dmitryax:
After giving it a second though, I think it still maybe better to provide a backward compatible solution even if it can be misused. Submitted another PR: https://github.com/open-telemetry/opentelemetry-collector/pull/5227

PR #5233, opened 2022-04-19
[pdata] Split `pmetric.MetricValueType` into more specific types

dmitryax:
Split `pmetric.MetricValueType` into `NumberDataPointValueType` and `ExemplarValueType`

1. `pmetric.NumberDataPoint.ValueType()` now returns a `NumberDataPointValueType` that can be `NumberDataPointValueTypeInt` or `NumberDataPointValueTypeDouble`.

2. `pmetric.Exemplar.ValueType()` now returns a `ExemplarValueType` that can be `ExemplarValueInt` or `ExemplarValueDouble`.

(1) use case is most commonly being used. Thats why we keep backward compatibility for it. We don't keep backward compatibility for (2) it to avoid mixing the split types.

Replaces https://github.com/open-telemetry/opentelemetry-collector/pull/5227

### DIFF
--- a/model/pdata/metrics_alias.go
+++ b/model/pdata/metrics_alias.go
@@ -103,16 +103,16 @@ const (
 
 // MetricValueType is an alias for pmetric.MetricValueType type.
 // Deprecated: [v0.49.0] Use pmetric.MetricValueType instead.
-type MetricValueType = pmetric.MetricValueType
+type MetricValueType = pmetric.MetricValueType //nolint:staticcheck
 
 const (
 
 	// Deprecated: [v0.49.0] Use pmetric.MetricValueTypeNone instead.
-	MetricValueTypeNone = pmetric.MetricValueTypeNone
+	MetricValueTypeNone = pmetric.MetricValueTypeNone //nolint:staticcheck
 
 	// Deprecated: [v0.49.0] Use pmetric.MetricValueTypeInt instead.
-	MetricValueTypeInt = pmetric.MetricValueTypeInt
+	MetricValueTypeInt = pmetric.MetricValueTypeInt //nolint:staticcheck
 
 	// Deprecated: [v0.49.0] Use pmetric.MetricValueTypeDouble instead.
-	MetricValueTypeDouble = pmetric.MetricValueTypeDouble
+	MetricValueTypeDouble = pmetric.MetricValueTypeDouble //nolint:staticcheck
 )
--- a/pdata/pmetric/alias.go
+++ b/pdata/pmetric/alias.go
@@ -74,10 +74,34 @@ const (
 )
 
 // MetricValueType specifies the type of NumberDataPoint.
-type MetricValueType = internal.MetricValueType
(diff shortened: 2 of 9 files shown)


model/pdata/metrics_alias.go
```diff
@@ -103,16 +103,16 @@ const (
...
+	MetricValueTypeNone = pmetric.MetricValueTypeNone //nolint:staticcheck
 
 	// Deprecated: [v0.49.0] Use pmetric.MetricValueTypeInt instead.
-	MetricValueTypeInt = pmetric.MetricValueTypeInt
+	MetricValueTypeInt = pmetric.MetricValueTypeInt //nolint:staticcheck
 
 	// Deprecated: [v0.49.0] Use pmetric.MetricValueTypeDouble instead.
-	MetricValueTypeDouble = pmetric.MetricValueTypeDouble
+	MetricValueTypeDouble = pmetric.MetricValueTypeDouble //nolint:staticcheck
 )
```
bogdandrutu, line 118: Already deprecated, we don't need to make changes to this anymore.

pdata/pmetric/alias.go
```diff
@@ -74,10 +74,42 @@ const (
...
+)
+
+// ExemplarValueType specifies the type of Exemplar measurement value.
+type ExemplarValueType = internal.MetricValueType
+
+const (
+	ExemplarValueTypeNone   = internal.MetricValueTypeNone
+	ExemplarValueTypeInt    = internal.MetricValueTypeInt
+	ExemplarValueTypeDouble = internal.MetricValueTypeDouble
+)
```
bogdandrutu, line 115: I don't think this is that used, so I would propose an exception here and do the correct thing, as long as we can build successfully the contrib. Give a try :)

bogdandrutu: My idea is to see if we do the right thing for exemplar (no backwards compatible change) is ok.

dmitryax: Makes sense. We can probably break `MetricValueType` -> `ExemplarValueType`, but keep `MetricValueType` <-> `NumberDataPOintValueType` backward compatible 👍

model/pdata/metrics_alias.go
dmitryax: This change is just to avoid linter complains about deprecated API usage.

pdata/pmetric/alias.go
dmitryax: Updated PR. Looks like Exemplar types are used at in `pkg/translator/opencensus/metrics_to_oc.go`  I don't see any other errors

bogdandrutu (approved):
I am ok with the exception for the `ExemplarValueType`.

cc @codeboten @tigrannajaryan

tigrannajaryan:
> I am ok with the exception for the `ExemplarValueType`.
> 
> cc @codeboten @tigrannajaryan

SGMT, please check the contrib codebase for usage too and if it is very rarely used then I am fine with this approach.

tigrannajaryan (approved):

codeboten (approved):

PR #5236, opened 2022-04-20
[receiver/otlp] Add tests for YAML edge cases

mx-psi:
**Description:** 

Add two test cases for edge cases in writing the OTLP receiver YAML:

- A protocol key is explicitly set as `null`.
- A protocol key is explicitly set to an empty object, `{}`.

In both cases the protocol will be enabled.

These test cases would usually not be set by end-users but they can happen if there is some processing of YAML (e.g. by something like Helm). I think it's important to have them here to avoid regressions if we ever change the underlying code for configuration loading


mx-psi:
I don't think this requires a changelog since it only adds unit tests to avoid future regressions.

codeboten (approved):

PR #5244, opened 2022-04-21
Extend unmarshal hook to check map key string to any TextUnmarshaler not only ComponentID

bogdandrutu:


config/configmap.go
```diff
@@ -170,51 +163,62 @@ var (
...
+
+		m := reflect.MakeMap(reflect.MapOf(t.Key(), reflect.TypeOf(true)))
+		for k := range data.(map[string]interface{}) {
+			tKey := reflect.New(t.Key())
+			if err := tKey.Interface().(encoding.TextUnmarshaler).UnmarshalText([]byte(k)); err != nil {
+				return nil, err
+			}
+
+			if m.MapIndex(reflect.Indirect(tKey)).IsValid() {
+				return nil, fmt.Errorf("duplicate name %q after trimming spaces %v", k, tKey)
```
mx-psi, line 218: This error may be true in one particular case, but it's not with a general `UnmarshalText` implementation
```suggestion
				return nil, fmt.Errorf("duplicate name %q after unmarshaling %v", k, tKey)
```

mx-psi (approved):
LGTM, only a nit about the error message

PR #5245, opened 2022-04-21
Fix nil pointer exception in service.NewConfigProvider

dmitryax:
Deprecated `ConfigProviderSettings.Unmarshaler` is still being used in `service.NewConfigProvider` causing nil pointer exception it's not provided.


dmitryax:
Changelog entry is skipped because deprecation of `ConfigProviderSettings.Unmarshaler` is not released yet

bogdandrutu (approved):

PR #5249, opened 2022-04-23
Otlp http receiver transport

jonathonfletcher:
**Description:**

Enable protocols.http.transport config, comparable to protocols.grpc.transport config.

Useful in mixed ipv4 / ipv6 environments where the `"tcp"` transport does not bind as desired. enables use of transport `"tcp4"` and transport `"tcp6"` in http receiver equivalent to same in gprc receiver.

**Testing:**

Tested with:
- `transport: tcp4`
- `transport: tcp6`
- `transport: tcp`

to confirm same bind as grpc receiver with same config setting.Are there gprc receiver transport configure tests in-tree - the existing tests all seem to wire to "tcp".

**Documentation:** 

Added line in `receiver/otlpreceiver/config.md` under `confighttp-HTTPServerSettings` with `transport` equivalent to same under `configgrpc-GRPCServerSettings`.


config/confighttp/confighttp.go
```diff
@@ -203,6 +203,9 @@ type HTTPServerSettings struct {
 	// Endpoint configures the listening address for the server.
 	Endpoint string `mapstructure:"endpoint"`
```
bogdandrutu, line 204: If needed to extend with transport, we should "embed" (using mapstructure `squash`) https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confignet/confignet.go#L22 into this config.

config/confighttp/confighttp.go
```diff
@@ -203,6 +203,9 @@ type HTTPServerSettings struct {
...
 
+	// Server Transport config. For transport only "tcp" / "tcp4" / "tcp6" are valid options.
+	Transport string `mapstructure:"transport"`
```
bogdandrutu, line 207: We need unit tests for this.

bogdandrutu: Is this comment correctly place in this thread? I don't feel that a unittest will break anything.

jpkrohling: This is not a unit test O_O

jpkrohling: Sorry, just realized I hijacked your comment. My comments are a different concern than yours.

PR #5256, opened 2022-04-25
Remove usage of unnecessary omitempty

bogdandrutu:
`omitempty` has effect only if the config is used as a source to encode/decode from, but the config is always the destination in our case, so it has no effect.


jpkrohling (approved):
Looks good, a grep didn't return any other instance of `omitempty` other than the ones brought by generated code (inside pdata/internal/data/protogen)

PR #5258, opened 2022-04-26
Reduced logging level for collector shutdown due to context cancel

### LINKED ISSUE #5242: Collector Stopping Due to Context Cancel is Noisy
**Describe the bug**
In the collector when context passed into `Run` is cancelled the collector logs and error message [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L182). When the context is from a `signal.NotifyContext` call it gives the user an error log and stack trace. This can cause the user to think there is something wrong when the collector correctly shut down.

Example of error log:

```
2022-04-20T14:25:06.545-0400	error	service/collector.go:182	Context done, terminating process	{"error": "context canceled"}
go.opentelemetry.io/collector/service.(*Collector).runAndWaitForShutdownEvent
	/home/otel/go/pkg/mod/go.opentelemetry.io/collector@v0.49.0/service/collector.go:182
go.opentelemetry.io/collector/service.(*Collector).Run
	/home/otel/go/pkg/mod/go.opentelemetry.io/collector@v0.49.0/service/collector.go:255
main.runInteractive
	/home/otel/git/observiq-otel-collector/cmd/collector/main.go:51
main.run
	/home/otel/git/observiq-otel-collector/cmd/collector/main_others.go:27
main.main
	/home/otel/git/observiq-otel-collector/cmd/collector/main.go:40
runtime.main
	/usr/local/go/src/runtime/proc.go:250
```

**Steps to reproduce**

1. Pass a context from `signal.NotifyContext` into the collector.
2. Let the program run for a bit then SIGTERM it and observe the error log.

A more graceful shutdown message that doesn't print a stack trace. Likely just changing the log level to a debug log. Maybe print and error 

**What did you see instead?**
Error log with a stack trace

**What version did you use?**
Version: `v0.49.0`

**Config**
Any configuration works. `DisableGracefulShutdown` must be set to `true` so the wrapping program can intercept the signals.

**Additional context**

I believe we should reduce the log level of exiting due to a context. If we want to error maybe only do so on a `context.DeadlineExceeded` error.

cpheps:
Resolves #5242

Reduced the logging level when the collector exits due to a context cancel. When the collector is exited with a canceled context it would print an error with a stack trace. This lead to confusing logs when the collector correctly shut down due to a context that was closed by signal intercepts.

~~Kept the context error log if the error is `context.DeadlineExceeded` as it's much more likely if the collector exceeded some deadline the user should be notified of such.~~

**Link to tracking Issue:** #5242 

**Testing:** Tested locally with signal intercepts. Current unit tests are valid.

### DIFF
--- a/service/collector.go
+++ b/service/collector.go
@@ -183,15 +183,9 @@ LOOP:
 			col.logger.Info("Received shutdown request")
 			break LOOP
 		case <-ctx.Done():
-			col.logger.Info("Context done, terminating process")
+			// Log the error for the context as a string or else zap.Error prints a stack trace.
+			col.logger.Info("Context done, terminating process", zap.String("Context Error", ctx.Err().Error()))
 
-			// Log error for when a context has timed out.
-			// A context that is canceled could be do to intercepted signals or intentional
-			// and we don't want to log an error with a stack trace to the user.
-			err := ctx.Err()
-			if errors.Is(err, context.DeadlineExceeded) {
-				col.logger.Error("Context deadline exceeded", zap.Error(err))
-			}
 			// Call shutdown with background context as the passed in context has been canceled
 			return col.shutdown(context.Background())
 		}
--- a/service/collector.go
+++ b/service/collector.go
@@ -183,8 +183,7 @@ LOOP:
 			col.logger.Info("Received shutdown request")
 			break LOOP
 		case <-ctx.Done():
-			// Log the error for the context as a string or else zap.Error prints a stack trace.
-			col.logger.Info("Context done, terminating process", zap.String("Context Error", ctx.Err().Error()))
+			col.logger.Info("Context done, terminating process", zap.Error(ctx.Err()))
 
 			// Call shutdown with background context as the passed in context has been canceled
 			return col.shutdown(context.Background())
--- a/service/collector.go
+++ b/service/collector.go
@@ -183,7 +183,15 @@ LOOP:
 			col.logger.Info("Received shutdown request")
 			break LOOP
 		case <-ctx.Done():
-			col.logger.Error("Context done, terminating process", zap.Error(ctx.Err()))
+			col.logger.Info("Context done, terminating process")
+
+			// Log error for when a context has timed out.
+			// A context that is canceled could be do to intercepted signals or intentional
+			// and we don't want to log an error with a stack trace to the user.
+			err := ctx.Err()
+			if errors.Is(err, context.DeadlineExceeded) {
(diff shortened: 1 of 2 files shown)


service/collector.go
```diff
@@ -183,7 +183,15 @@ LOOP:
 			col.logger.Info("Received shutdown request")
 			break LOOP
 		case <-ctx.Done():
-			col.logger.Error("Context done, terminating process", zap.Error(ctx.Err()))
+			col.logger.Info("Context done, terminating process")
```
bogdandrutu, line 186: I would include the error anyway in this message, or is that the one that prints the stacktrace?

service/collector.go
```diff
@@ -183,7 +183,15 @@ LOOP:
...
+
+			// Log error for when a context has timed out.
+			// A context that is canceled could be do to intercepted signals or intentional
+			// and we don't want to log an error with a stack trace to the user.
+			err := ctx.Err()
+			if errors.Is(err, context.DeadlineExceeded) {
```
bogdandrutu, line 192: Why is this error special?

service/collector.go
cpheps: Yeah the `zap.Error(ctx.Err())` prints the stack trace. I think if you don't wrap it in `zap.Error` it won't print the stack but rather just the error string. I would be ok doing that I think.

service/collector.go
cpheps: I figured you still wanted to know if the collector was stopped due to a DeadlineExceed error as that would be less expected than a cancel. I think though the above comment of maybe just logging the error with no stack trace would allow us to remove this.

codeboten (approved):

service/collector.go
```diff
@@ -183,7 +183,9 @@ LOOP:
 			col.logger.Info("Received shutdown request")
 			break LOOP
 		case <-ctx.Done():
-			col.logger.Error("Context done, terminating process", zap.Error(ctx.Err()))
+			// Log the error for the context as a string or else zap.Error prints a stack trace.
+			col.logger.Info("Context done, terminating process", zap.String("Context Error", ctx.Err().Error()))
```
bogdandrutu, line 187: Wrote a test to prove this. Is the "error" log level that prints the stacktrace, not the zap.Error field. So this PR can be as simple as changing `col.logger.Error` to `col.logger.Info`

cpheps: Good catch. Not sure how I got confused on it but appreciate the double check.

bogdandrutu (approved):

PR #5268, opened 2022-04-27
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/builder-release.yaml
+++ b/.github/workflows/builder-release.yaml
@@ -5,8 +5,13 @@ on:
     tags:
       - 'cmd/builder/v*'
 
+permissions:
+  contents: read
+
 jobs:
   goreleaser:
+    permissions:
+      contents: write  # for goreleaser/goreleaser-action to create a GitHub release
     runs-on: ubuntu-20.04
     steps:
       -
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -4,8 +4,15 @@ on:
     branches: [ main ]
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   CodeQL-Build:
+    permissions:
+      actions: read  # for github/codeql-action/init to get workflow details
+      contents: read  # for actions/checkout to fetch code
+      security-events: write  # for github/codeql-action/autobuild to send a status report
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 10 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -4,8 +4,15 @@ on:
     branches: [ main ]
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   CodeQL-Build:
+    permissions:
+      actions: read  # for github/codeql-action/init to get workflow details
```
bogdandrutu, line 13: Is this not by default?

naveensrinivasan: No, it isn't. But that is required for Private repos. So I think for this one it isn't required.

.github/workflows/stale-pr.yaml
```diff
@@ -3,8 +3,14 @@ on:
...
     - cron: "12 3 * * *" # arbitrary time not to DDOS GitHub
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
```
mx-psi, line 13: This looks sensible but I think we don't have a good way to test it, so I am afraid this PR could break this and we wouldn't notice in a long time

.github/workflows/builder-release.yaml
```diff
@@ -5,8 +5,13 @@ on:
     tags:
       - 'cmd/builder/v*'
 
+permissions:
+  contents: read
+
 jobs:
   goreleaser:
+    permissions:
+      contents: write  # for goreleaser/goreleaser-action to create a GitHub release
```
mx-psi, line 14: ditto here (well, here it would not take a long time to notice but it would affect the release, so I would prefer to have this tested first)

mx-psi (commented):
I think it's fine to add these, at least on the ones where it's clear that we only need read (for the rest I think we want some proof that it works before merging). I also would like to understand what the default permissions are

Issue #5270, opened 2022-04-27
[pdata] Fix copying of metric fields of primitive items slice type


dmitryax:
Same issues as https://github.com/open-telemetry/opentelemetry-collector/issues/5266 but for metric fields of primitive item slices type:

- BucketCounts
- ExplicitBounds

PR #5271, opened 2022-04-27
[pdata] Fix copying of metric fields of primitive items slice type

dmitryax:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5270

### DIFF
--- a/pdata/internal/generated_pmetric.go
+++ b/pdata/internal/generated_pmetric.go
@@ -1489,8 +1489,20 @@ func (ms HistogramDataPoint) CopyTo(dest HistogramDataPoint) {
 		dest.SetSum(ms.Sum())
 	}
 
-	dest.SetBucketCounts(ms.BucketCounts())
-	dest.SetExplicitBounds(ms.ExplicitBounds())
+	if len(ms.orig.BucketCounts) == 0 {
+		dest.orig.BucketCounts = nil
+	} else {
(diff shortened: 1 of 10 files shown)


pdata/internal/generated_pmetric.go
```diff
@@ -1489,8 +1489,20 @@ func (ms HistogramDataPoint) CopyTo(dest HistogramDataPoint) {
 		dest.SetSum(ms.Sum())
 	}
 
-	dest.SetBucketCounts(ms.BucketCounts())
-	dest.SetExplicitBounds(ms.ExplicitBounds())
+	if ms.orig.BucketCounts != nil {
```
bogdandrutu, line 1492:
```golang
if len(ms.orig.BucketCounts) == 0 {
	dest.orig.BucketCounts = nil
} else {
	dest.orig.BucketCounts = make([]uint64, len(ms.orig.BucketCounts))
	copy(dest.orig.BucketCounts, ms.orig.BucketCounts)
}
```

dmitryax: But this would be not exact copy... if `ms.orig.BucketCounts` has `[]uint64{}`, `dest.orig.BucketCounts` will be `nil`

bogdandrutu: and? do we need to distinguished between these?

bogdandrutu (approved):

PR #5272, opened 2022-04-27
[bugfix] temporary fix around grpclogger

codeboten:
Calling SetLoggerV2 isn't thread safe and should only be called `before any grpc functions`. See more details: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/logging/settable#pkg-overview

This hacky workaround checks if the collector's logger has been set, if so, don't bother updating the grpc logger. Looking for alternative implementations here.

### DIFF
--- a/service/collector.go
+++ b/service/collector.go
@@ -202,18 +202,11 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 		return fmt.Errorf("failed to get config: %w", err)
 	}
 
-	// TODO: hacky workaround to unblock v0.50.0
-	configureGRPCLogging := false
-	if col.logger == nil {
-		configureGRPCLogging = true
-	}
 	if col.logger, err = telemetrylogs.NewLogger(cfg.Service.Telemetry.Logs, col.set.LoggingOptions); err != nil {
 		return fmt.Errorf("failed to get logger: %w", err)
 	}
 
-	// setting the grpc logger multiple times causes
-	// a race condition
-	if configureGRPCLogging {
+	if !col.set.SkipSettingGRPCLogger {
 		telemetrylogs.SetColGRPCLogger(col.logger, cfg.Service.Telemetry.Logs.Level)
 	}
 
--- a/service/collector.go
+++ b/service/collector.go
@@ -201,11 +201,21 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to get config: %w", err)
 	}
+
+	// TODO: hacky workaround to unblock v0.50.0
+	configureGRPCLogging := false
+	if col.logger == nil {
+		configureGRPCLogging = true
+	}
 	if col.logger, err = telemetrylogs.NewLogger(cfg.Service.Telemetry.Logs, col.set.LoggingOptions); err != nil {
 		return fmt.Errorf("failed to get logger: %w", err)
 	}
 
-	telemetrylogs.SetColGRPCLogger(col.logger, cfg.Service.Telemetry.Logs.Level)
+	// setting the grpc logger multiple times causes
+	// a race condition
+	if configureGRPCLogging {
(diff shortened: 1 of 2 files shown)


service/collector.go
```diff
@@ -201,11 +201,21 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
...
 	if col.logger, err = telemetrylogs.NewLogger(cfg.Service.Telemetry.Logs, col.set.LoggingOptions); err != nil {
 		return fmt.Errorf("failed to get logger: %w", err)
 	}
 
-	telemetrylogs.SetColGRPCLogger(col.logger, cfg.Service.Telemetry.Logs.Level)
+	// setting the grpc logger multiple times causes
+	// a race condition
+	if configureGRPCLogging {
+		telemetrylogs.SetColGRPCLogger(col.logger, cfg.Service.Telemetry.Logs.Level)
+	}
```
bogdandrutu, line 218: clean this logic and use a "sync.Once" for this behavior? Also this should be "global var sync.Once" since multiple collector instances should not race on this.

tigrannajaryan: Is it worth documenting that `cfg.Service.Telemetry.Logs.Level` is not reconfigurable when a ConfigProvider watches and returns a new config?

bogdandrutu: this is only for grpc logs.

tigrannajaryan: Yes, I mean changing `cfg.Service.Telemetry.Logs.Level` will have no effect on grpc logs. Probably a minor detail.

codeboten: After reviewing this further, the problem is that the testbed is configuring the receivers before the service has a chance to configure the logger, using once won't help here as the grpc receivers are already starting before the grpc logger is set.

I was planning on adding an option to the collector to skip setting the grpc logger, which can then be configured in the InProcessCollector for the testbed. Thoughts?

tigrannajaryan: >I was planning on adding an option to the collector to skip setting the grpc logger, which can then be configured in the InProcessCollector for the testbed. Thoughts?

I think this is fine for the tests.

service/settings.go
```diff
@@ -64,6 +64,9 @@ type CollectorSettings struct {
 	// LoggingOptions provides a way to change behavior of zap logging.
 	LoggingOptions []zap.Option
 
+	// SkipSettingGRPCLogger avoids setting the grpc logger
+	SkipSettingGRPCLogger bool
```
jpkrohling, line 68: I might still be in need of a coffee, but how does this work? This is a new value, and I see this value only being read, never written, so, it will always be false. The conditional on service/collector.go would then always yield true, which will call the same code as before this PR.

tigrannajaryan: I think it will be set in correctness tests, which are in the contrib repo, not in this repo.

BTW, I am not sure why we moved the all of the testbed from core to contrib. I consider the testbed to be a core part of the project. We run the testbed for otlp receiver/exporter in this repo, now we don't anymore and any problem discovery is delayed until the contrib consumes the new core.

bogdandrutu: I do agree that maybe the framework of testbed can live here, the reason we moved it was because we don't want to give API guarantees to the framework, and core is closer and closer to offer also API stability.

bogdandrutu (approved):

service/settings.go
tigrannajaryan: >the reason we moved it was because we don't want to give API guarantees to the framework

That's a good point. Maybe we can keep testbed go module on 0.x forever to indicate it is unstable and also have explicit warning that no-one should use this API except contrib (yes, I know people don't pay attention to warnings :-) ).

mx-psi:
This fixed #4971; It's a good temporary fix but I don't love this because I feel like it's yet another individual flag for customizing telemetry, and I would like to instead have something more general like the telemetry providers I proposed on #4970. 

I haven't had time recently to push for that but I intend to continue the work there in the next few weeks, and would appreciate comments/thoughts/feedback from @codeboten and others on that issue.

codeboten:
I'll take a look and provide feedback after today's release @mx-psi

PR #5160, opened 2022-04-06
Change featureflags package to make it test friendly

bogdandrutu:
Fixes:  https://github.com/open-telemetry/opentelemetry-collector/issues/4917

PR #5293, opened 2022-04-29
[service] nit: move all telemetry fields into TelemetrySettings.

mx-psi:
**Description:** 

Just a small nit to make `Collector` and `service` more consistent by moving every telemetry-related struct into a `component.TelemetrySettings`.

The only thing I don't like is that the field name conflicts with the one from #5160, better name ideas are accepted :)

### DIFF
--- a/service/collector.go
+++ b/service/collector.go
@@ -26,7 +26,6 @@ import (
 	"syscall"
 
 	"go.opentelemetry.io/contrib/zpages"
-	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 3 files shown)


bogdandrutu (approved):

bogdandrutu:
> The only thing I don't like is that the field name conflicts with the one from https://github.com/open-telemetry/opentelemetry-collector/pull/5160, better name ideas are accepted :)

That is temporary, when changed to otel metrics (hopefully soon) we will not have that problem anymore.

PR #5294, opened 2022-04-29
[extension/zpages] Register span processor within zpages extension code

mx-psi:
**Description:** 

Use [`RegisterSpanProcessor`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TracerProvider.RegisterSpanProcessor) to handle zpages registration instead of registering it at tracer provider creation time.

This allows removing al zpages references from `service.Collector` and all `zpagesSpanProcessor` references on `service.service`.

This removes special code from the service moving it to its rightful place, and reduces service's API. It will also allow me to simplify the telemetry providers PoC.


service/collector.go
```diff
@@ -244,10 +241,9 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 // Run starts the collector according to the given configuration given, and waits for it to complete.
 // Consecutive calls to Run are not allowed, Run shouldn't be called once a collector is shut down.
 func (col *Collector) Run(ctx context.Context) error {
-	col.zPagesSpanProcessor = zpages.NewSpanProcessor()
 	col.tracerProvider = sdktrace.NewTracerProvider(
 		sdktrace.WithSampler(internal.AlwaysRecord()),
```
bogdandrutu, line 245: This is needed only for the `zPagesSpanProcessor`. If not registered, we don't need to locally record all the spans.

bogdandrutu (approved):

service/collector.go
bogdandrutu: Somehow... there is something I don't like here ... (sorry, most likely is not this code, but the overall structure has some problems).

An idea that I had in mind for long time, is to actually embed the "zpages" into Host:
* Host will have to start an "http.Server" that can be shared between "health check", "zpages", "status", "etc.".
* If we do this, the zpages/health extensions (maybe) becomes almost a no-op.

How I would design this is probably on the Host interface allow any component to "Register" a path/handler similar with `http.ServerMux`, something like:
```golang
RegisterHandler(pattern string, handler http.Handler)
```

Then we should extend the "service" configuration to accept an http server configuration for this, and maybe some options to say "register trace zpages", "register metrics zpages", etc.

The `service.Host` can accept this instance of the `http.Server` (similar to how you construct things like telemetry).

mx-psi: Let's focus on the zpages case first, since I am having a bit of trouble visualizing how this would work. There are two parts for zpages:

1. `RegisterZPages`, which needs to live in the service/host since it needs to know about the available pipelines
2. the `/debug/tracez` handler, which can be dealt within the `zpages` extension code or within service/host code

In this case, what you would want is to: for (1), change it so that you don't need to pass the `http.ServeMux` but rather have this be done within the service code (you would still need a special method on the component.Host to activate this, right?) and for (2) call the new `RegisterHandler` method from `component.Host` to register the `tracez` handler.

I guess the part that confuses me the most is 
> maybe some options to say "register trace zpages", "register metrics zpages", etc.

since I don't understand where these options live (do they live in `service.svcSettings`? are these methods part of `service.Host` but not part of the `component.Host` interface?)

bogdandrutu: TL;DR: Who owns the "http.Server" instance. right now that is the main reason why we have the `zpages` extension, if we change the `serviceHost` to have a  default (can be shared between components) `http.Server` the need of zpages extension goes away. If we do this, then the `serviceHost` can own the entire `zpages` logic.

bogdandrutu: > since I don't understand where these options live (do they live in service.svcSettings? are these methods part of service.Host but not part of the component.Host interface?)

Somewhere here https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/service.go#L24

bogdandrutu:
@mx-psi can you please rebase this, I feel that I've seen some of the changes in other already merged PRs

extension/zpagesextension/zpagesextension.go
```diff
@@ -17,30 +17,52 @@ package zpagesextension // import "go.opentelemetry.io/collector/extension/zpage
...
+	// Use an anonymous interface to support non-SDK trace providers.
+	sdktracer, ok := zpe.telemetry.TracerProvider.(interface {
+		RegisterSpanProcessor(SpanProcessor trace.SpanProcessor)
+	})
+	if ok {
+		sdktracer.RegisterSpanProcessor(zpe.zpagesSpanProcessor)
+		zPagesMux.Handle(path.Join("/debug", tracezPath), zpages.NewTracezHandler(zpe.zpagesSpanProcessor))
+		zpe.telemetry.Logger.Info("Registered zPages span processor on tracer provider")
+	} else {
+		zpe.telemetry.Logger.Info("zPages span processor registration is not available")
```
bogdandrutu, line 55: Maybe warning?

extension/zpagesextension/zpagesextension.go
```diff
@@ -17,30 +17,52 @@ package zpagesextension // import "go.opentelemetry.io/collector/extension/zpage
...
 	hostZPages, ok := host.(interface {
 		RegisterZPages(mux *http.ServeMux, pathPrefix string)
 	})
 	if ok {
-		zpe.logger.Info("Register Host's zPages")
 		hostZPages.RegisterZPages(zPagesMux, "/debug")
+		zpe.telemetry.Logger.Info("Registered Host's zPages")
 	} else {
-		zpe.logger.Info("Host's zPages not available")
+		zpe.telemetry.Logger.Info("Host's zPages not available")
```
bogdandrutu, line 65: Warn?

extension/zpagesextension/zpagesextension.go
```diff
@@ -69,12 +91,27 @@ func (zpe *zpagesExtension) Shutdown(context.Context) error {
 	if zpe.stopCh != nil {
 		<-zpe.stopCh
 	}
+
+	// Try to unregister the zpages span processor with the UnregisterSpanProcessor method.
+	// https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TracerProvider.UnregisterSpanProcessor.
+	// Use an anonymous interface to support non-SDK trace providers.
+	sdktracer, ok := zpe.telemetry.TracerProvider.(interface {
+		UnregisterSpanProcessor(SpanProcessor trace.SpanProcessor)
+	})
```
bogdandrutu, line 100: nit: I would actually define a private interfaces instead of anonymous? Maybe also check for both funcs from the beginning?

extension/zpagesextension/zpagesextension.go
```diff
@@ -69,12 +91,27 @@ func (zpe *zpagesExtension) Shutdown(context.Context) error {
...
+	if ok {
+		sdktracer.UnregisterSpanProcessor(zpe.zpagesSpanProcessor)
+		zpe.telemetry.Logger.Info("Unregistered zPages span processor on tracer provider")
+	} else {
+		zpe.telemetry.Logger.Info("zPages span processor unregistration is not available")
```
bogdandrutu, line 105: same warning?

extension/zpagesextension/zpagesextension.go
```diff
@@ -69,12 +91,27 @@ func (zpe *zpagesExtension) Shutdown(context.Context) error {
 	if zpe.stopCh != nil {
 		<-zpe.stopCh
 	}
+
+	// Try to unregister the zpages span processor with the UnregisterSpanProcessor method.
+	// https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#TracerProvider.UnregisterSpanProcessor.
+	// Use an anonymous interface to support non-SDK trace providers.
```
bogdandrutu, line 97: nit: I think is a good reason, very unlikely since the processor is a SDK concept :)

extension/zpagesextension/zpagesextension.go
mx-psi: done

extension/zpagesextension/zpagesextension.go
mx-psi: done

extension/zpagesextension/zpagesextension.go
mx-psi: I did a single private interface https://github.com/open-telemetry/opentelemetry-collector/blob/9ac886315100151251823a89705ca4c43f3a2f4c/extension/zpagesextension/zpagesextension.go#L41-L54

extension/zpagesextension/zpagesextension.go
mx-psi: done

extension/zpagesextension/zpagesextension.go
mx-psi: If people want to wrap the official SDK with something of their own we can support them in this way, while we wouldn't be able to otherwise, I think that's a good enough reason to keep it with an interface

service/collector.go
mx-psi: Okay, I think I get it now, and it sounds like a good idea. I created #5353 to track this. I still think this PR makes sense to merge, since it reduces the settings one needs to pass to the `service` struct , which can help us expose it.

bogdandrutu: Based on that issue, should we actually move that into the service itself?

bogdandrutu:
merging this for the moment, but I think we can do even better :)

Issue #5297, opened 2022-04-29
Prepend windowsEventLogCore instead of appending


braydonk:
**Is your feature request related to a problem? Please describe.**
When creating a new Windows service, a `zap` core for Windows Event Log output is created and appended to the `CollectorSettings.LoggingOptions` that were passed in by the user. Unfortunately, this means the `windowsEventLogCore` is the first in line of zap cores to be executed. This leads to the LoggingOptions passed in not being applied to the Windows Event Log output.

**Describe the solution you'd like**
It would be preferable if the `windowsEventLogCore` was instead prepended to the `LoggingOptions` in [newWithWindowsEventLogCore](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector_windows.go#L154) so that the options the user passed in will wrap the event log core, instead of the event log core wrapping the user's options.

braydonk:
I have opened a draft PR that shows the change I would like to make. I will be running a manual test first thing Monday.

braydonk:
I finally successfully ran a manual test that confirmed my suspicion and confirmed that the fix in the PR mentioned above works.

braydonk:
Fix is merged, thanks!

braydonk:
This regressed in #9726.

braydonk:
With #11051 merged, this can be closed again.

PR #5298, opened 2022-04-29
service: prepend the windowsEventLogCore

braydonk:
When the `windowsEventLogCore` is *appended* to the logging options, it makes the event log core run before any user specified options. This is undesirable because it means the logs are sent to the event log before any of the user's options are applied. This PR changes `newWithWindowsEventLogCore` to *prepending* the `windowsEventLogCore` option to the array of LoggingOptions, so it will always run after the user specified options.

**Link to tracking Issue:** #5297 

**Testing:** 
* `make` and `make fmt`
* **MANUAL TEST PENDING**


jsuereth (approved):

dashpole (approved):
Please add a changelog entry

dashpole (approved):

codeboten (approved):

Issue #5263, opened 2022-04-26
[pdata] Defensive setters/getters for primitive value slices


dmitryax:
This is generalized issue made from https://github.com/open-telemetry/opentelemetry-collector/issues/4181.

We have the following setters accepting slices of primitive types (with similar getters) that take the value as is without using defensive copying:

```
Value.SetBytesVal(bv []byte)
HistogramDataPoint.SetBucketCounts(v []uint64)
HistogramDataPoint.SetExplicitBounds(v []float64)
```

We need to make at least the setters to use defensive copies of the slices passed as arguments, so any change to the value slice that used as the argument to the setter doesn't affect stored value.

An open question if we should do the same for getters and return copies instead of the stored values.

UPD:

The decision was made to wrap primitive slices into another structure like `ImmutableUInt64Slice`. In order to update all the methods working directly with primitive slices in a backward compatible way, we are going to make several changes across multiple OTel Collector releases:

**0.51.0**: Prefix all the methods working with primitive slices with `M` by introducing new methods and deprecating old ones.
**0.52.0**: Remove methods that are not prefixed with `M`.
**0.53.0**: Introduce methods that are not prefixed with `M` back, but now working with new immutable slice structs
**0.54.0**: Remove methods prefixed with `M`.

dmitryax:
> An open question if we should do the same for getters and return copies instead of the stored values.

If we don't do the same for getters, we introduce another way to mutate the slices. And doing it that way may cause some subtle bugs. E.g. if exporter uses one of these getters and then changes the returned value, it makes the exporter mutating which is hard to track in this scenario, and this can also happen accidentally.

I vote for using defensive copying for getters as well setters. It makes it easier to track pdata mutations.

dmitryax:
@open-telemetry/collector-approvers please share your opinions

tigrannajaryan:
We can try "moving" setters to avoid the copying performance impact (bytes can be potentially large):
```go
Value.SetBytesVal(bv *[]byte) {
  target = *bv
  *bv = nil
}
```
If the caller needs to hold a copy for some reason they can make copy before the call.

Not very idiomatic, I know.

Probably unnecessary for histograms since the bucket is cannot be very large (I think?)

dmitryax:
@tigrannajaryan I think that's ok to use for setter, but what about getters?

tigrannajaryan:
> @tigrannajaryan I think that's ok to use for setter, but what about getters?

[...]

PR #5299, opened 2022-04-29
[pdata] Use immutable slices for primitive type slices

dmitryax:
Use immutable slices for primitive types slices to restrict mutations:
  - `Value.SetBytesVal` func is deprecated in favor of `Value.SetIBytesVal` func that takes 
    `pcommon.ImmutableByteSlice` instead of []byte.
  - `Value.BytesVal` func is deprecated in favor of `Value.IBytesVal` func that returns `pcommon.ImmutableByteSlice` 
    instead of []byte.
  - `<HistogramDataPoint|Buckets>.SetBucketCounts` funcs are deprecated in favor of 
    `<HistogramDataPoint|Buckets>.SetIBucketCounts` funcs that take `pcommon.ImmutableUInt64Slice` instead of []uint64.
  - `<HistogramDataPoint|Buckets>.BucketCounts` funcs are deprecated in favor of
    `<HistogramDataPoint|Buckets>.IBucketCounts` funcs that return `pcommon.ImmutableUInt64Slice` instead of []uint64.
  - `HistogramDataPoint.SetExplicitBounds` func is deprecated in favor of `HistogramDataPoint.SetIExplicitBounds` func 
    that takes `pcommon.ImmutableFloat64Slice` instead of []float64.
  - `HistogramDataPoint.ExplicitBounds` func func is deprecated in favor of `HistogramDataPoint.IExplicitBounds` 
    returns `pcommon.ImmutableFloat64Slice` instead of []float64.

Use `pcommon.New...ImmutebleSliceFromValue` to convert primitive slice into `pcommon....ImmutebleSlice`, original slice will be reused and not available anymore.

Use `pcommon.New...ImmutebleSliceFromCopy` to create `pcommon....ImmutebleSlice` from a copy of the original slice that will not be affected.

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/5263


dmitryax:
Given that slices of primitive types are not widely used and doing this change in a backward compatible way would require a several iterations, I believe we can allow this change as another exception.

dmitryax:
@tigrannajaryan I applied the approach that you recommended in https://github.com/open-telemetry/opentelemetry-collector/issues/5263 PTAL once you have a chance

I like how it ended up looking

pdata/internal/common.go
```diff
@@ -500,7 +487,7 @@ func (v Value) asRaw() interface{} {
 	case ValueTypeInt:
 		return v.IntVal()
 	case ValueTypeBytes:
-		return v.BytesVal()
+		return v.BytesVal().Value()
```
tigrannajaryan, line 490: For the future, it appears one of the common places we call asRaw() from is when we need to convert it to to JSON string in `AsString()`. We can have a non-copying version of asRaw() to be used in `AsString()`. Can be a future PR (we can postpone it until we really see the need because of bad performance).

pdata/internal/generated_immutable_slice.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ImmutableByteSlice represents a []byte slice that cannot be mutated.
+type ImmutableByteSlice struct {
+	value []byte
+}
+
+// NewImmutableByteSlice creates a new ImmutableByteSlice by moving an underlying array of the provided []byte slice.
+// Note that the argument value will be cleared after the call,
+// use NewImmutableByteSliceFromCopy if it needs to be preserved.
+func NewImmutableByteSlice(val *[]byte) ImmutableByteSlice {
```
tigrannajaryan, line 28: My slight concern is if this may be a bit dangerous if the caller is not careful and tries to work with the nil-ed pointer after the call. However I don't think this is going to be called with a pointer variable, it will be most likely called by taking an address of any existing slice, so it should be safe, there is no way to cause a nil pointer access and panic.
So, I think this is fine. A bit non-idiomatic for Go, but I think it should be ok.

tigrannajaryan (approved):

pdata/internal/common.go
dmitryax: Sounds good. Another option can be to introduce `Value.AsJSON() method

pdata/internal/generated_immutable_slice.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ImmutableUIntSlice represents a []uint64 slice that cannot be mutated.
+type ImmutableUIntSlice struct {
+	value []uint64
+}
+
+// NewImmutableUIntSlice creates a new ImmutableUIntSlice by moving an underlying array of the provided []uint64 slice.
+// Note that the argument value will be cleared after the call,
+// use NewImmutableUIntSliceFromCopy if it needs to be preserved.
+func NewImmutableUIntSlice(val *[]uint64) ImmutableUIntSlice {
```
bogdandrutu, line 104: Uint64 in the name in case we ever need the real uint?

pdata/internal/generated_immutable_slice.go
```diff
@@ -0,0 +1,132 @@
...
+	if val != nil {
+		is.value = *val
+		*val = nil
+	}
+	return is
+}
+
+// NewImmutableByteSliceFromCopy creates a new ImmutableByteSlice by copying the provided []byte slice.
+// Use NewImmutableByteSlice if the provided slice is not needed anymore.
+func NewImmutableByteSliceFromCopy(val []byte) ImmutableByteSlice {
```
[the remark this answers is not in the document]
dmitryax: Makes sense, signatures of `CopyFrom` and `MoveFrom` will be different from `CopyTo` and `MoveTo` that we have but I think that's ok. `CopyFromRaw` or `CopyFromBytes` probably to verbose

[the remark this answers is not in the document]
dmitryax: Maybe `CopyRaw` and `MoveRaw` as another option?

pdata/internal/generated_immutable_slice.go
dmitryax: Yes, makes sense. But we still keep `NewImmutableDoubleSlice(val *[]float64)`, right?

pdata/internal/generated_immutable_slice.go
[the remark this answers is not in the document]
dmitryax: We also have getter method called `Value`. So setters can use the term. For example:
- `Value() []byte`, `MoveValue(val *[]byte)`, `CopyValue(val []byte)`
- `Raw() []byte`, `MoveRaw(val *[]byte)`, `CopyRaw(val []byte)`

pdata/internal/generated_immutable_slice.go
bogdandrutu: For the "primitives" slices I would go with the go terminology.

dmitryax:
> Found a case where I need the number of buckets for histograms and number of counts. Consider to expose this since it is immutable.

They are all exposed already through the aliases

pdata/internal/generated_immutable_slice.go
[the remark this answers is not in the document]
dmitryax: I tried this approach at it seems like the interface becomes clunky. In order to use a new `ImmutableByteSlice` user would have to write 3 lines:
```golang
ibs := ImmutableByteSlice()
ibs.MoveFrom(&[]{1,2,3})
use(ibs)
```

instead of

```golang
use(NewImmutableByteSlice(&[]{1,2,3}))
```

Unless we `MoveValue` and `CopyValue` returns the `ImmutableByteSlice` instance itself which doesn't seem ideal.

Also use case when the same instance can be used for several calls of `MoveFrom` or `CopyFrom` seems unrealistic.

What about `NewImmutableByteSliceFromValue` and `NewImmutableByteSliceFromCopy`?

dmitryax:
@bogdandrutu PTAL

pdata/internal/generated_immutable_slice.go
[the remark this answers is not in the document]
bogdandrutu: > Also use case when the same instance can be used for several calls of MoveFrom or CopyFrom seems unrealistic.

Not really, for histogram bucket boundaries will lots of times be the case correct?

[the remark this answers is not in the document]
dmitryax: I'm going to apply these changes in contrib and will see how it looks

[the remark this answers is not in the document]
dmitryax: @bogdandrutu I applied the changed in config and didn't run into use cases where the `Copy/Move` functionality would be preferable. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9940

dmitryax:
Look like this functionality is being used for histogram metrics in all the distributions. So it may be better to run this change through 2-step deprecation process:
1. Introduce new functions: `Value.SetIBytesVal`, `Value.IBytesVal`, `<HistogramDataPoint|Buckets>.SetIBucketCounts`, `<HistogramDataPoint|Buckets>.IBucketCounts`, `HistogramDataPoint.IExplicitBounds` and deprecate existing ones.
2. Deprecate functions prefixed with `I` and replace `Value.SetBytesVal`, `Value.BytesVal`, `<HistogramDataPoint|Buckets>.SetBucketCounts` with them.
3. Remove functions prefixed with `I`.

Need more opinions if it makes sense or we can go through a one time breaking change as currently done in this PR.

cc @Aneurysm9 @dashpole @pmm-sumo

dmitryax:
I see there are two projects that will be affected by this change, and some contrib modules depend on them:
- https://github.com/influxdata/influxdb-observability 
- https://github.com/GoogleCloudPlatform/opentelemetry-operations-go

So it'll be pretty difficult to make as one breaking change :(

bogdandrutu:
I agree. We should not break these projects...

dmitryax:
@bogdandrutu I updated the PR to be not a breaking change

dmitryax:
To kickstart the process of upgrading the methods in backward compatible in 0.51.0, decision was made to make a simple rename of the methods first https://github.com/open-telemetry/opentelemetry-collector/pull/5344

dmitryax:
@bogdandrutu I updated this PR. PTAL whenever you have a chance

pdata/internal/cmd/pdatagen/internal/immutable_slice_structs.go
```diff
@@ -0,0 +1,225 @@
...
+// limitations under the License.
+
+package internal // import "go.opentelemetry.io/collector/pdata/internal/cmd/pdatagen/internal"
+
+import (
+	"os"
+	"strings"
+)
+
+const immutableSliceTemplate = `// ${structName} represents a []${itemType} slice that cannot be mutated.
```
bogdandrutu, line 22: Not sure if it is worth to generate these slices. I will give some attention to this generator while you are away.

pdata/internal/generated_immutable_slice.go
```diff
@@ -0,0 +1,162 @@
...
+	is := ImmutableFloat64Slice{}
+	if len(val) != 0 {
+		is.value = make([]float64, len(val))
+		copy(is.value, val)
+	}
+	return is
+}
+
+// RawSlice returns a copy of the []float64 slice.
+func (is ImmutableFloat64Slice) RawSlice() []float64 {
```
bogdandrutu, line 97: I think we use the "AsRaw" pattern? (see the `pcommon.Value` as an example)

pdata/internal/generated_immutable_slice.go
```diff
@@ -0,0 +1,162 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Code generated by "model/internal/cmd/pdatagen/main.go". DO NOT EDIT.
+// To regenerate this file run "go run model/internal/cmd/pdatagen/main.go".
+
+package internal
+
+// ImmutableByteSlice represents a []byte slice that cannot be mutated.
+type ImmutableByteSlice struct {
```
bogdandrutu, line 21: We should comment that the same "instance" can be used with multiple "set" calls, since it is immutable.

pdata/internal/generated_immutable_slice.go
```diff
@@ -0,0 +1,162 @@
...
+	value []byte
+}
+
+// NewImmutableByteSliceMoveFrom creates a new ImmutableByteSlice by replacing an underlying value of the []byte slice.
+// Note that the argument value will be cleared after the call,
+// use NewImmutableByteSliceCopyFrom if it needs to be preserved.
+func NewImmutableByteSliceMoveFrom(val *[]byte) ImmutableByteSlice {
```
bogdandrutu, line 28: Don't feel the best about these names for constructors, if anyone has an idea, will be appreciated. An option is to keep `func NewImmutableByteSlice(val []byte) ImmutableByteSlice` or maybe `func NewImmutableByteSlice(values ...byte) ImmutableByteSlice` since I believe that will force always to create a new "slice" so it will be easier to use than the current "CopyFrom" current form?

bogdandrutu (commented):
overall looks good.

pdata/internal/generated_immutable_slice.go
dmitryax: I like `func NewImmutableByteSlice(values ...byte) ImmutableByteSlice` form

dmitryax: I think we can keep `func NewImmutableByteSlice(values []byte) ImmutableByteSlice` form. I tried the `...byte` construction and looks like it doesn't force copying.

pdata/internal/generated_immutable_slice.go
dmitryax: Added

pdata/internal/generated_immutable_slice.go
dmitryax: Sounds good. updated

pdata/internal/generated_immutable_slice.go
bogdandrutu: Let's remove the move constructor for the moment? What do you think?

> I tried the ...byte construction and looks like it doesn't force copying.

Can you share the experiment or what you did?

pdata/internal/generated_pmetric.go
```diff
@@ -1444,24 +1444,24 @@ func (ms HistogramDataPoint) SetSum(v float64) {
...
+func (ms HistogramDataPoint) SetBucketCounts(v ImmutableUInt64Slice) {
+	(*ms.orig).BucketCounts = v.value
 }
 
-// MExplicitBounds returns the mexplicitbounds associated with this HistogramDataPoint.
-func (ms HistogramDataPoint) MExplicitBounds() []float64 {
-	return (*ms.orig).ExplicitBounds
+// ExplicitBounds returns the explicitbounds associated with this HistogramDataPoint.
+func (ms HistogramDataPoint) ExplicitBounds() ImmutableFloat64Slice {
+	return ImmutableFloat64Slice{(*ms.orig).ExplicitBounds}
```
bogdandrutu, line 1459: nit: use named arguments for initialization?

dmitryax: Updated

pdata/internal/generated_immutable_slice.go
dmitryax: I updated the constructor to the following form:

```
func NewImmutableByteSlice(val ...byte) ImmutableByteSlice {
	is := ImmutableByteSlice{}
	is.value = val
	return is
}
```

Then tried this:

```
	b := []byte{1,2,3}
	is := pcommon.NewImmutableByteSliceM(b...)
	fmt.Println(is.At(0))
	b[0] = 10
	fmt.Println(is.At(0))
```

The output was:
```
1
10

```

dmitryax:
As discussed with @bogdandrutu, since we don't have a need for the optimized version of the constructor where the primitive is reused in the immutable slice, I'm removing `NewImmutable...SliceMoveFrom` constructors until there is a need for them.

I confirmed that we don't need this optimization in contrib at this point https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/11058

pdata/internal/common.go
```diff
@@ -262,10 +268,18 @@ func (v Value) SliceVal() Slice {
 // If the Type() is not ValueTypeBytes then returns false.
 // Calling this function on zero-initialized Value will cause a panic.
 // Modifying the returned []byte in-place is forbidden.
+// Deprecated: [0.54.0] Use BytesVal() instead.
 func (v Value) MBytesVal() []byte {
 	return v.orig.GetBytesValue()
 }
 
+// BytesVal returns the ImmutableByteSlice value associated with this Value.
+// If the Type() is not ValueTypeBytes then returns false.
```
tigrannajaryan, line 277: The comment is not accurate. This does not return bool.

dmitryax: Thanks. fixed

bogdandrutu (approved):

PR #5305, opened 2022-05-03
Change config.Retrieved to be an interface

bogdandrutu:
This change will allow to implement support for "retrieving" any type of values, since we can offer implementations of the retrieved for string/etc.

### DIFF
--- a/config/mapprovider.go
+++ b/config/mapprovider.go
@@ -16,6 +16,8 @@ package config // import "go.opentelemetry.io/collector/config"
 
 import (
 	"context"
+	"fmt"
+	"reflect"
 )
 
 // MapProvider is an interface that helps to retrieve a config map and watch for any
@@ -80,9 +82,25 @@ type ChangeEvent struct {
 }
 
 // Retrieved holds the result of a call to the Retrieve method of a Provider object.
-type Retrieved struct {
-	cfgMap    *Map
-	closeFunc CloseFunc
+type Retrieved interface {
+	// AsMap returns the retrieved configuration parsed as a Map.
+	AsMap() (*Map, error)
+
+	// AsValue returns the retrieved configuration parsed as an interface{} which can be one of the following types:
+	//   * Primitive float/int/bool/string;
+	//   * []interface{} - every member follows the same rules as the given interface{};
+	//   * map[string]interface{} - every value follows the same rules as the given interface{};
+	AsValue() (interface{}, error)
+
+	// Close and release any watchers that MapProvider.Retrieve may have created.
+	//
+	// Should block until all resources are closed, and guarantee that `onChange` is not
+	// going to be called after it returns except when `ctx` is cancelled.
+	//
+	// Should never be called concurrently with itself.
+	Close(ctx context.Context) error
+
+	unexportedFunc()
 }
 
 type retrievedSettings struct {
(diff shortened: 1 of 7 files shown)


bogdandrutu:
cc @mx-psi since you review the initial proposal https://github.com/open-telemetry/opentelemetry-collector/pull/5198

config/mapprovider.go
```diff
@@ -80,11 +80,19 @@ type ChangeEvent struct {
 }
 
 // Retrieved holds the result of a call to the Retrieve method of a Provider object.
-type Retrieved struct {
-	// Deprecated: [v0.50.0] Use NewRetrievedFromMap to initialize, and Retrieved.AsMap to access.
-	*Map
-	// Deprecated: [v0.50.0] Use NewRetrievedFromMap to initialize, and Retrieved.Close to access.
-	CloseFunc
+type Retrieved interface {
```
bogdandrutu, line 83: Alternative, can make this a struct with only internal members for the moment.

mx-psi: I don't have strong feelings either way... it feels like we should have some criteria on when unimplementable interface is preferable  vs. when do we prefer a struct with private fields but, lacking that, this is fine to me

bogdandrutu: In general I think unless 2 completely different implementation of the same thing you should not use an interface. Let me try the other version to compare :)

bogdandrutu: First trivial change is here https://github.com/open-telemetry/opentelemetry-collector/pull/5363, this should be non controversial. After that is merged I will make another PR to show how we can have with a struct the 2 possible values vs interface.

bogdandrutu: I kind of implement the interface with only one struct, see this updated PR, I think we don't necessary need the "interface" change anymore. Waiting for feedback.

config/mapprovider.go
```diff
@@ -110,26 +128,62 @@ func NewRetrievedFromMap(cfgMap *Map, opts ...RetrievedOption) Retrieved {
...
-//
-// Should never be called concurrently with itself.
-func (r Retrieved) Close(ctx context.Context) error {
-	if r.closeFunc == nil {
-		return nil
+// NewRetrievedFromString returns a new Retrieved instance that contains a string value.
+// * cfgMap the Map that will be merged to the given map in the MergeTo.
+// * CloseFunc specifies a function to be invoked when the configuration for which it was
+//   used to retrieve values is no longer in use and should close and release any watchers
+//	 that it may have created.
```
mx-psi, line 151: docs are out of date here and above

mx-psi, line 151: docs are out of date here and above

config/mapprovider.go
```diff
@@ -110,26 +128,62 @@ func NewRetrievedFromMap(cfgMap *Map, opts ...RetrievedOption) Retrieved {
...
+
+type valueRetrieved struct {
+	value interface{}
+	CloseFunc
+}
+
+func (r *valueRetrieved) AsMap() (*Map, error) {
+	val, ok := r.value.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("retrieved value (type=%v) cannot be used as a *Map", reflect.TypeOf(r.value))
```
mx-psi, line 168: You can [use `%T`](https://pkg.go.dev/fmt#hdr-Printing) in a format string to get the type
```suggestion
		return nil, fmt.Errorf("retrieved value (type=%T) cannot be used as a *Map", r.value)
```

bogdandrutu:
This is outdated at this point.

Issue #5310, opened 2022-05-04
[pdata] Inconsistently exported AsRaw methods


dmitryax:
Currently only `Map` has exported `AsRaw` method and `NewMapFromRaw` function. Very similar `Slice` structure has the same but not exported functions: `asRaw` and `newSliceFromRaw`. Another structure that has the same unexported functions is `Value`.

I propose to expose at least `Slice.AsRaw()` and `NewSliceFromRaw` for consistency with `Map`. Not sure about `Value`, probably we can keep it can stay unexported until there is a need.

bogdandrutu:
do you also need a fromRaw for the slice?

dmitryax:
Right. I missed that. Updated the issue description and the PR

PR #5311, opened 2022-05-04
[pdata] Make Slice.AsRaw() method exported

dmitryax:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/5310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### 💡 Enhancements 💡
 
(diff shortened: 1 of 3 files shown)


bogdandrutu (approved):
What is the need for this method? Just curious.

dmitryax:
It's needed in contrib in one place https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9c214be44477ef84daa0770a3f0464fc25eae7cf/exporter/lokiexporter/encode_json.go#L98

Also it seems inconsistent that Map has it while Slice doesn't

PR #5347, opened 2022-05-11
[pdata] Remove deprecated primitive slice getters/setters

dmitryax:
This is the second step in a sequence of backward changes to replace primitive slices with immutable structs https://github.com/open-telemetry/opentelemetry-collector/issues/5263


bogdandrutu:
We probably need to coordinate with @dashpole since most likely Google exporter uses these fields.

dmitryax:
Yes, a few other distros are using it as well. They should migrate to the methods prefixed with `M` by the next release according to the schedule defined https://github.com/open-telemetry/opentelemetry-collector/issues/5263 . Let me know if that schedule makes sense

dmitryax:
@bogdandrutu the contrib test is passing now

bogdandrutu (approved):

PR #5357, opened 2022-05-12
Move telemetry initialization to the service, because config is part of the service config.

bogdandrutu:


bogdandrutu:
cc @mx-psi

mx-psi (commented):
The deprecation of `GetLogger` makes total sense to me (maybe move it to a different PR so that this doesn't block it?).

For the other part, I have thought about this in the past weeks and I was considering it the other way around (i.e. we should move the `Telemetry` section out of `ServiceConfig`, at least in the Go API) but this could be an alternative approach too. The open questions to me for this approach are:

1. How do we support customizing telemetry for Go API users? With the API before this PR, we could expose the service, but with this approach we can't do that (we could still do something like telemetry providers or something like it I guess?)
2. We force the service initialization to happen first, since it's the one that creates the telemetry settings, that you probably want to use all over your application. This may not fit well on existing applications; for `service.Collector` in particular it means that adding new functionality to the Collector needs to happen on the service/extensions if we want to use `TelemetrySettings` on it. Is that okay?

bogdandrutu:
> The deprecation of GetLogger makes total sense to me (maybe move it to a different PR so that this doesn't block it?).

Will do.

> How do we support customizing telemetry for Go API users? With the API before this PR, we could expose the service, but with this approach we can't do that (we could still do something like telemetry providers or something like it I guess?)

I want to have separation of concerns, and here is where I got with my thoughts:
1. `Collector` struct (not sure the right package/name of the struct) is responsible to resolve a "Config", watch for  config updates, watch for kernel signals, and owns the "Service" (means start/stop/etc.).
2. `service.Service` struct (we can discuss about the name later, because this is somehow confusing with "service" part in config), is responsible for all the "parts" that constitute the "otelcol service". Essentially given a configuration for everything (components, pipelines, telemetry, and the right factories (the current `component.Factories` probably should be moved to `service.Factories` and probably extended to provide support for `PipelineFactory` and `TelemetryLogsFactory` etc.)) creates all the components, builds the pipelines.
3. `service.Host` struct is responsible to "run" the components that needs to be up and running during the lifetime of a service (extensions + pipelines(receivers, processors, exporters)) and owns their lifetime. It also allows these components to collaborate with the "service" via things like reporting status, etc.

> We force the service initialization to happen first, since it's the one that creates the telemetry settings, that you probably want to use all over your application. This may not fit well on existing applications; for service.Collector in particular it means that adding new functionality to the Collector needs to happen on the service/extensions if we want to use TelemetrySettings on it. Is that okay?

I personally thing there are few places where "functionality" should live:
1. If it is a pipeline component it has to be a receiver/processor/exporter.
2. It it is a component that needs to run for the entire lifetime of the service, OR may need to interact with other pipeline components, OR the Host (access telemetry, report status, access other components), it will be an "extension".
3. For things that are extremely generic and useful for all the previous defined components (extension/receiver/processor/exporter) we add this as part of the "Host".

mx-psi (approved):
The separation of concerns looks reasonable in principle, and this is all internal, so I am approving so as not to block your work, and we can continue the discussion when I am back from PTO and I think more about it :)

service/telemetry.go
```diff
@@ -57,7 +57,7 @@ const (
 )
 
 type collectorTelemetryExporter interface {
-	init(col *Collector) error
+	init(srv *service) error
```
codeboten, line 60: nit: `svc` makes more sense to me here, as `srv` is associated with server, and not service. not a huge deal though.

codeboten (commented):
Please resolve conflicts, otherwise the change looks good.

PR #5378, opened 2022-05-17
Improve map providers tests, env use constant, file add missing invalid

bogdandrutu:


dmitryax (commented):
Should invalid-yaml.yaml be moved or can be removed?

bogdandrutu:
Do you mean copy? It is correct now, the file provider has a test that reads an "invalid" config, and was passing because file did not exists (not because of invalid yaml, hence the diff in coverage). For env provider we provide a "const" invalid yaml string.

dmitryax:
I just don't understand why `testdata/invalid-yaml.yaml` file was moved from `envmapprovider` to `filemapprovider` in this PR. Shouldn't the file be removed instead?

bogdandrutu:
> I just don't understand why testdata/invalid-yaml.yaml file was moved from envmapprovider to filemapprovider in this PR. Shouldn't the file be removed instead?

TL;DR -> NO 

The `envmapprovider` tests were updated to no longer read from files the input, and instead define the input yaml string as const (visible in the PR).

The `filemapprovider`, had a test for an [TestInvalidYAML](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/mapprovider/filemapprovider/mapprovider_test.go#L54), which as mentioned in the previous comment passed in the past, but not because of the invalid yaml, but because of the missing file. By "moving" (actually you can see this as fixing that test and adding the invalid file, which github sees it as a "move").

dmitryax:
Ok, thanks for explaing

dmitryax (approved):

PR #5382, opened 2022-05-17
BreakingChange: Change to use interface instead of func for MapConverter

bogdandrutu:
The main reason to do this, is because if in the future we want to extend the capabilities of the "MapConverter" we don't have any way to do that. By using an interface we can define "optional" interfaces that some converters may implement to enable future improvements.

### DIFF
--- a/config/mapconverter.go
+++ b/config/mapconverter.go
@@ -1,0 +1,34 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config // import "go.opentelemetry.io/collector/config"
+
+import (
+	"context"
+)
+
+// MapConverter is a converter interface for the config.Map that allows distributions
+// (in the future components as well) to build backwards compatible config converters.
+type MapConverter interface {
+	// Convert applies the conversion logic to the given "cfgMap".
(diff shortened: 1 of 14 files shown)


config/mapconverter.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// MapConverter is a converter interface for the config.Map that allows distributions
+// (in the future components as well) to build backwards compatible config converters.
+type MapConverter interface {
+	// Convert applies the conversion logic to the given "cfgMap".
+	Convert(ctx context.Context, cfgMap *Map) error
+}
+
+// Deprecated: Implement MapConverter interface.
```
codeboten, line 28: Missing version

config/mapconverter/expandmapconverter/expand.go
```diff
@@ -21,16 +21,20 @@ import (
...
+// New returns a config.MapConverter, that expands all environment variables for a given config.Map.
 //
 // Notice: This API is experimental.
-func New() config.MapConverterFunc {
-	return func(_ context.Context, cfgMap *config.Map) error {
-		for _, k := range cfgMap.AllKeys() {
-			cfgMap.Set(k, expandStringValues(cfgMap.Get(k)))
-		}
-		return nil
+func New() config.MapConverter {
```
codeboten, line 29: This should be deprecated first

codeboten (commented):
I prefer the use of an interface as per the change. Is there a way to deprecate the existing converter's `New` functions before changing them? Otherwise this breaks the "breaking change" policy

bogdandrutu:
> Otherwise this breaks the "breaking change" policy

Indeed that's why it is called "braking change" and, yes we need @open-telemetry/collector-approvers to say they are ok with this.

bogdandrutu:
> New functions before changing them

The result of the `New` function is only (I haven't seen any other way) used in the "converters" slice of `service.ConfigProvider`. So this will not be that disruptive for users, the most annoying will be if they implement converters by themselves...

dmitryax:
I'm Ok with accepting this exception. This API doesn't seem to be widely adopted at this point

tigrannajaryan (approved):
I am OK with the change.

bogdandrutu:
@codeboten ptal

codeboten (approved):

Issue #3262, opened 2021-05-23
[processors/batch] Size calculations inconsistent, causing unbounded batch size in bytes


timcosta:
**Describe the bug**
The `batch` processor is not limiting batch size by bytes, but rather by item count. I believe this is due to confusing nomenclature, let me try to explain...

In [`config.go`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/config.go#L32) we define `SendBatchSize` and `SendBatchMaxSize`, neither of which specify the unit being counted, i.e whether the size is counting bytes or items.

In [`batch_processor.go`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L66) `size()` explicitly refers to bytes, and `itemCount()` is explicitly the number of items based on the documentation.

The problem begins in [`processItem()`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L154) where the line `for bp.batch.itemCount() >= bp.sendBatchSize` begins to mix definitions of count and size and compares the number of items to the configured batch size.

When we later [`sendItems()`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L177) we have two metrics `statBatchSendSize` and `statBatchSendSizeBytes` which are properly counted using `itemCount()` and `size()` respectively. 

In the [`export()`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L319) functions we double down on the mix up with `if sendBatchMaxSize > 0 && bl.logCount > sendBatchMaxSize` again mixing count and size terms when the config and function documentation shows that they are distinct values.

In [`split_logs.go`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/splitlogs.go#L23) the comparison conflates count and size once again.

The tests for this appear to be misleading, because they are explicitly [counting bytes](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor_test.go#L580) for the size, but then use a [`logsPerRequest`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor_test.go#L588) variable when calculating the expected number of batches.

[...]

PR #5385, opened 2022-05-18
fix: Send correct batch stats when SendBatchMaxSize is set

njvrzm:
**Description:** 
This fixes a bug with the batch processor's `batch_send_size` and `batch_send_size_bytes` metrics. Their values were being calculated before `SendBatchMaxSize` was applied.

We observed this issue during performance testing. With `SendBatchMaxSize` set to a small enough value that it almost always took effect, graphs of `batch_send_size` showed an odd sawtooth pattern and the total values were far in excess of the number of items actually sent. Looking at some individual metrics the issue was clear - with a `SendBatchMaxSize` of 100, for instance, we'd see `batch_send_size` metrics looking like 1000, 900, 800, 700... as the full size of the batch queue was recorded but not sent each time.

With this change the various `export` methods report the actual count of items sent, and byte size sent if requested, and the `sendItems` method records those values.

**Testing:** 
I added a test called `TestBatchProcessorSentBySize_withMaxSize`, based on `TestBatchProcessorSentBySize` but with `SendBatchMaxSize` set and with all spans delivered in a single request so that the batch size is predictable. It does not attempt to validate the `batch_send_size_bytes` metric - the splitting of batches makes the method used in the original test fail due to different amounts of overhead.

(Incidentally, `TestBatchProcessorSentBySize` itself is rather brittle. Changing either `sendBatchSize` or `spansPerRequest` so that the former is not a multiple of the latter makes the test fail in several ways.)


processor/batchprocessor/batch_processor.go
```diff
@@ -244,17 +245,24 @@ func (bt *batchTraces) add(item interface{}) {
 	td.ResourceSpans().MoveAndAppendTo(bt.traceData.ResourceSpans())
 }
 
-func (bt *batchTraces) export(ctx context.Context, sendBatchMaxSize int) error {
+func (bt *batchTraces) export(ctx context.Context, sendBatchMaxSize int, returnBytes bool) (int, int, error) {
```
jaronoff97, line 248: question: should we do the `stats.Record` call within export rather than returning it? That way we don't need to make any function signature changes?

jaronoff97: i can see why that would be annoying because then instead of a single stats record call, you have to make one in each batch processor. What do you think?

jaronoff97 (approved):

processor/batchprocessor/batch_processor.go
njvrzm: Yeah, I'd rather change the signature, especially since it's only used in one place, than duplicate the stats recording code.

processor/batchprocessor/batch_processor.go
```diff
@@ -175,15 +175,16 @@ func (bp *batchProcessor) resetTimer() {
...
 func (bp *batchProcessor) sendItems(triggerMeasure *stats.Int64Measure) {
-	// Add that it came form the trace pipeline?
-	stats.Record(bp.exportCtx, triggerMeasure.M(1), statBatchSendSize.M(int64(bp.batch.itemCount())))
-
-	if bp.telemetryLevel == configtelemetry.LevelDetailed {
-		stats.Record(bp.exportCtx, statBatchSendSizeBytes.M(int64(bp.batch.size())))
-	}
-
-	if err := bp.batch.export(bp.exportCtx, bp.sendBatchMaxSize); err != nil {
+	detailed := bp.telemetryLevel == configtelemetry.LevelDetailed
```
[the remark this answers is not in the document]
njvrzm: It looks like the [default](https://github.com/open-telemetry/opentelemetry-collector/blob/e6b5cd54d1fccb1d30aebfd76dd0714c7ed99842/service/internal/configunmarshaler/defaultunmarshaler.go#L198-L203) for this setting is `LevelBasic`, so eliminating the check here would be a behavior change.

codeboten (commented):
Thanks for catching and fixing this. Please add a changelog entry. Also can you confirm whether or not this addresses https://github.com/open-telemetry/opentelemetry-collector/issues/3262

njvrzm:
Thanks for having a look, @codeboten.
* This does not address #3262.
* I've added a changelog entry.
* Should I try to address the codecov check failure as well?

PR #5395, opened 2022-05-18
[Telemetry Query Language] Updating doc to match naming standards

TylerHelmuth:
**Description:**
Updates function names/parameters in doc to match function naming standards.

### DIFF
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -147,49 +147,54 @@ the design phase. The Python-like language currently assumes statements can only
 
 Remove a forbidden attribute such as `http.request.header.authorization` from spans only
 
-`delete(attributes["http.request.header.authorization"]) from traces`
-`delete(attributes["http.request.header.authorization"])`
+`delete(attributes["http.request.header.authorization"]) from traces`  
+`delete(attributes["http.request.header.authorization"])`  
 
 Remove all attributes except for some
 
-`keep_keys(attributes, "http.method", "http.status_code") from metrics`
-`keep_keys(attributes, "http.method", "http.status_code")`
+`keep_keys(attributes, "http.method", "http.status_code") from metrics`  
+`keep_keys(attributes, "http.method", "http.status_code")`  
 
 Reduce cardinality of an attribute
 
-`replace_wildcards("/user/*/list/*", "/user/{userId}/list/{listId}", attributes["http.target"]) from traces`
-`replace_wildcards("/user/*/list/*", "/user/{userId}/list/{listId}", attributes["http.target"])`
+`replace_match(attributes["http.target"], "/user/*/list/*", "/user/{userId}/list/{listId}") from traces`  
+`replace_match(attributes["http.target"], "/user/*/list/*", "/user/{userId}/list/{listId}")`  
 
 Reduce cardinality of a span name
 
-`replace_wildcards("GET /user/*/list/*", "GET /user/{userId}/list/{listId}", name) from traces`
-`replace_wildcards("GET /user/*/list/*", "GET /user/{userId}/list/{listId}", name)`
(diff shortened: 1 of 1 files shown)


docs/processing.md
```diff
@@ -147,49 +147,54 @@ the design phase. The Python-like language currently assumes statements can only
 
 Remove a forbidden attribute such as `http.request.header.authorization` from spans only
 
-`delete(attributes["http.request.header.authorization"]) from traces`
-`delete(attributes["http.request.header.authorization"])`
+`delete(attributes["http.request.header.authorization"]) from traces`  
+`delete(attributes["http.request.header.authorization"])`
```
bogdandrutu, line 151: Why do we double all the lines w/o a "from" statement?

TylerHelmuth: According to the [example description](https://github.com/open-telemetry/opentelemetry-collector/blob/ab796818282d61ba0bec0e372adb126f73c169a2/docs/processing.md?plain=1#L145) its showing 2 different types of languages

bogdandrutu: I don't understand sorry. Do you want to say that the only diff is "from signal" everywhere?

TylerHelmuth: Some of the example show the difference better

```
Filtering out signals such as by removing all metrics with a `http.target` of `/health`

`drop() where attributes["http.target"] = "/health" from metrics`
`if attributes["http.target"] = "/health": drop()`

Attach information from resource into telemetry, for example adding certain resource fields as metric attributes

`set(attributes["k8s_pod"], resource.attributes["k8s.pod.name"]) from metrics`
`attributes["k8s_pod"] = resource.attributes["k8s.pod.name"]`
```

But ya, for the scenarios where the work is all done in a function its basically the same

docs/processing.md
```diff
@@ -147,49 +147,54 @@ the design phase. The Python-like language currently assumes statements can only
...
-`delete(resource.attributes["process.command_line"])`
+`delete(resource.attributes["process.command_line"]) from traces`  
+`delete(resource.attributes["process.command_line"])`  
 
 Filtering out signals such as by removing all metrics with a `http.target` of `/health`
 
-`drop() where attributes["http.target"] = "/health" from metrics`
-`if attributes["http.target"] = "/health": drop()`
+`drop() where attributes["http.target"] = "/health" from metrics`  
+`if attributes["http.target"] = "/health": drop()`
```
bogdandrutu, line 181: How do we say the same thing here that only drop attributes from "metrics" points not spans (second line does not have anything to signal that).

docs/processing.md
bogdandrutu: Ya, but the second line does not have a "signal" part, which means that the "python" code applies to all not only to metrics.

docs/processing.md
TylerHelmuth: I think some of this is holdover from the original issue/pr discussion where it was being debated how the query language would interact with different signals.  The doc currently says "The Python-like language currently assumes statements can only be applied to one signal.", which at the same time implies that the SQL-like language could be applied across multiple.

In reality the collector architecture (and as a result the transform processor) won't let a function work on multiple signals at a time.  I think we could remove all the `from ...` statements (simplifying the comparison examples where possible) and update the doc to state a query can only work on 1 signal at a time.

TylerHelmuth:
Pinging @anuraaga again in case you missed it

docs/processing.md
anuraaga: Yeah we can probably remove them now and even the python-like syntax, don't think we have real plans to experiment with it at this point

TylerHelmuth: @bogdandrutu opened up #5408 to address this issue

bogdandrutu (approved):

PR #5400, opened 2022-05-19
[pdata] Rename `Value.NewValueBytes ` to `Value.NewValueBytes`

dmitryax:
Deprecate `pcommon.Value.NewValueBytes` in favor of `Value.NewValueMBytes` in preparation of migration to the new immutable slices.

This constructor was missed in https://github.com/open-telemetry/opentelemetry-collector/pull/5344. We will still stick to the schedule defined in https://github.com/open-telemetry/opentelemetry-collector/issues/5263. The deprecated function will be just replaced without removal in 0.53.0.


dmitryax:
@bogdandrutu PTAL

bogdandrutu (approved):

Issue #5398, opened 2022-05-19
Allow specifying Collector's own Resource in the config


tigrannajaryan:
In order to allow external Supervisor to control the Collector we want to the Supervisor to be able to tell the Collector to use a specific `service.instance.id` instead of auto-generating one.

One of the ways to allow this is to add a sub-section in the `service.telemetry` config section where any Resource attributes can be specified. This allow to specify not only `service.instance.id`, but also `service.namespace` (to .g. specify the team this Collector belongs too) or `deployment.environment` or even the `service.name` (defaults to current "io.opentelemetry.collector").

For example we can have the following in the config:
```yaml
service:
  telemetry:
    resource:
      service.instance.id: 01G3EN4NW306AFVGQT5ZYC0GEK
      service.namespace: onlineshop
      deployment.environment: production
```

PR #5402, opened 2022-05-19
Allow specifying Collector's own Resource in the config

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/5398

This adds a sub-section in the `service.telemetry` config section where any Resource attributes can be specified. All specified attributes will included as own metrics labels.

If `service.instance.id` is not specified it will be auto-generated. Previously `service.instance.id` was always auto-generated, so the default of the new behavior matches the old behavior.

For example we can have the following in the config:

```yaml
service:
  telemetry:
    resource:
      service.instance.id: 01G3EN4NW306AFVGQT5ZYC0GEK
      service.namespace: onlineshop
      deployment.environment: production
```

### DIFF
--- a/config/moved_service.go
+++ b/config/moved_service.go
@@ -38,6 +38,13 @@ type Service struct {
 type ServiceTelemetry struct {
 	Logs    ServiceTelemetryLogs    `mapstructure:"logs"`
 	Metrics ServiceTelemetryMetrics `mapstructure:"metrics"`
+
+	// Resource specifies user-defined attributes to include with all emitted telemetry.
+	// For metrics the attributes become Prometheus labels.
+	// Note that some attributes are added automatically (e.g. service.version) even
+	// if they are not specified here. In order to suppress such attributes the
+	// attribute must be specified in this map with null YAML value (nil string pointer).
+	Resource map[string]*string `mapstructure:"resource"`
 }
 
 // ServiceTelemetryLogs defines the configurable settings for service telemetry logs.
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -94,6 +94,7 @@ func (tel *colTelemetry) init(col *Collector) error {
 func (tel *colTelemetry) initOnce(col *Collector) error {
 	logger := col.telemetry.Logger
 	cfg := col.service.config.Telemetry
+	resource := cfg.Resource
 
 	level := cfg.Metrics.Level
 	metricsAddr := cfg.Metrics.Address
@@ -109,8 +110,29 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
 
 	logger.Info("Setting up own telemetry...")
 
-	instanceUUID, _ := uuid.NewRandom()
-	instanceID := instanceUUID.String()
+	// Construct telemetry attributes from resource attributes.
+	telAttrs := map[string]string{}
+	for k, v := range resource {
+		// nil value indicates that the attribute should not be included in the telemetry.
+		if v != nil {
+			telAttrs[k] = *v
+		}
+	}
+
+	if _, ok := resource[semconv.AttributeServiceInstanceID]; !ok {
+		// AttributeServiceInstanceID is not specified in the config. Auto-generate one.
+		instanceUUID, _ := uuid.NewRandom()
+		instanceID := instanceUUID.String()
+		telAttrs[semconv.AttributeServiceInstanceID] = instanceID
+	}
+
+	if AddCollectorVersionTag {
+		if _, ok := resource[semconv.AttributeServiceVersion]; !ok {
+			// AttributeServiceVersion is not specified in the config. Use the actual
+			// build version.
+			telAttrs[semconv.AttributeServiceVersion] = version.Version
+		}
+	}
 
 	var pe http.Handler
 	if tel.registry.IsEnabled(useOtelForInternalMetricsfeatureGateID) {
@@ -120,7 +142,7 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
 		}
 		pe = otelHandler
 	} else {
-		ocHandler, err := tel.initOpenCensus(col, instanceID)
+		ocHandler, err := tel.initOpenCensus(col, telAttrs)
 		if err != nil {
 			return err
 		}
@@ -131,8 +153,7 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
 		"Serving Prometheus metrics",
 		zap.String(zapKeyTelemetryAddress, metricsAddr),
 		zap.String(zapKeyTelemetryLevel, level.String()),
-		zap.String(semconv.AttributeServiceInstanceID, instanceID),
-		zap.String(semconv.AttributeServiceVersion, version.Version),
+		zap.Any("Resource", resource),
 	)
 
 	mux := http.NewServeMux()
@@ -152,7 +173,7 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
 	return nil
 }
 
-func (tel *colTelemetry) initOpenCensus(col *Collector, instanceID string) (http.Handler, error) {
+func (tel *colTelemetry) initOpenCensus(col *Collector, telAttrs map[string]string) (http.Handler, error) {
 	processMetricsViews, err := telemetry2.NewProcessMetricsViews(getBallastSize(col.service.host))
 	if err != nil {
 		return nil, err
@@ -178,10 +199,8 @@ func (tel *colTelemetry) initOpenCensus(col *Collector, instanceID string) (http
 
 	opts.ConstLabels = make(map[string]string)
 
-	opts.ConstLabels[sanitizePrometheusKey(semconv.AttributeServiceInstanceID)] = instanceID
-
-	if AddCollectorVersionTag {
-		opts.ConstLabels[sanitizePrometheusKey(semconv.AttributeServiceVersion)] = version.Version
+	for k, v := range telAttrs {
+		opts.ConstLabels[sanitizePrometheusKey(k)] = v
 	}
 
 	pe, err := prometheus.NewExporter(opts)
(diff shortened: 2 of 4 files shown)


config/moved_service.go
```diff
@@ -36,8 +36,9 @@ type Service struct {
 // ServiceTelemetry defines the configurable settings for service telemetry.
 // Deprecated: [v0.52.0] Use service.ConfigServiceTelemetry
 type ServiceTelemetry struct {
-	Logs    ServiceTelemetryLogs    `mapstructure:"logs"`
-	Metrics ServiceTelemetryMetrics `mapstructure:"metrics"`
+	Logs     ServiceTelemetryLogs    `mapstructure:"logs"`
+	Metrics  ServiceTelemetryMetrics `mapstructure:"metrics"`
+	Resource map[string]string       `mapstructure:"resource"`
```
bogdandrutu, line 41: can this be otel/attributes.Set? Do we want to allow non string values?

service/telemetry.go
```diff
@@ -109,8 +110,22 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
...
 
-	instanceUUID, _ := uuid.NewRandom()
-	instanceID := instanceUUID.String()
+	if resource == nil {
+		resource = map[string]string{}
+	}
+
+	var instanceID string
+	instanceID = resource[semconv.AttributeServiceInstanceID]
+	if instanceID == "" {
```
bogdandrutu, line 119: Do we want to enforce this? I always felt that we should not have this by default.

service/telemetry.go
```diff
@@ -109,8 +110,22 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
...
 	logger.Info("Setting up own telemetry...")
 
-	instanceUUID, _ := uuid.NewRandom()
-	instanceID := instanceUUID.String()
+	if resource == nil {
+		resource = map[string]string{}
+	}
+
+	var instanceID string
+	instanceID = resource[semconv.AttributeServiceInstanceID]
```
bogdandrutu, line 118: Allow users to decide to not have this... or have it empty if they want?
```golang
instanceID, ok := resource[semconv.AttributeServiceInstanceID]
if !ok {
}
```

service/telemetry.go
```diff
@@ -109,8 +110,22 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
...
+	if instanceID == "" {
+		// AttributeServiceInstanceID is not specified in the config. Auto-generate one.
+		instanceUUID, _ := uuid.NewRandom()
+		instanceID = instanceUUID.String()
+		resource[semconv.AttributeServiceInstanceID] = instanceID
+	}
+
+	if AddCollectorVersionTag {
+		resource[semconv.AttributeServiceVersion] = version.Version
+	}
```
bogdandrutu, line 128: Shouldn't you check if already there?

service/telemetry.go
```diff
@@ -109,8 +110,22 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
...
+	var instanceID string
+	instanceID = resource[semconv.AttributeServiceInstanceID]
+	if instanceID == "" {
+		// AttributeServiceInstanceID is not specified in the config. Auto-generate one.
+		instanceUUID, _ := uuid.NewRandom()
+		instanceID = instanceUUID.String()
+		resource[semconv.AttributeServiceInstanceID] = instanceID
+	}
+
+	if AddCollectorVersionTag {
```
bogdandrutu, line 126: I think we need an issue to make this feature gate or remove it, since I believe nobody sets this.

service/telemetry.go
tigrannajaryan: I think we need the default. Otherwise there is no way to differentiate metrics coming from different instances.

service/telemetry.go
tigrannajaryan: Yes, good catch.

service/telemetry.go
tigrannajaryan: Done. You can now provide empty string to suppress built-in attributes.

service/telemetry.go
tigrannajaryan: Fixed.

service/telemetry.go
tigrannajaryan: @open-telemetry/collector-approvers I would like opinions on this. Is this a) valuable and b) not-surprising behaviour?

service/telemetry.go
bogdandrutu: Yes, that's why I asked for an "issue" :)

tigrannajaryan: Issue created: https://github.com/open-telemetry/opentelemetry-collector/issues/5404

service/telemetry.go
mx-psi: Empty strings are valid on resource attributes so it does feel a bit like surprising behavior to me. 

I don't have a good replacement for this though, only thing I can think of is maybe just `service.instance.id: null`? since null attributes are [not valid](https://github.com/open-telemetry/opentelemetry-specification/blob/0c793b450098e8d7a739e575897707f7db08d8ce/specification/common/README.md#attribute).

TylerHelmuth:
This is a great addition, can the new feature be documented specifically on opentelemetry.io?  [There is a section of telemetry in the collector config doc](https://opentelemetry.io/docs/collector/configuration/#service).  It doesn't list every config option, but I think adding this feature to the explicitly mentioned options would be good.

tigrannajaryan:
> This is a great addition, can the new feature be documented specifically on opentelemetry.io? [There is a section of telemetry in the collector config doc](https://opentelemetry.io/docs/collector/configuration/#service). It doesn't list every config option, but I think adding this feature to the explicitly mentioned options would be good.

I think we need to look into using https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/configschema to generate the documentation and see if we can push it to opentelemetry.io (or pull from there). I don't think it is sustainable to document things here in the code and also on the website and keep the docs in sync manually. We can look into it in a separate PR, that's out of the scope for this PR.

service/telemetry.go
tigrannajaryan: Yeah, `null` is probably a better choice. Let me try that.

service/telemetry.go
```diff
@@ -109,8 +110,32 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
...
+		}
+	}
+
+	// Empty Resource attributes are an indication by the user to suppress and stop
+	// emitting the particular attribute.
+	for k, v := range resource {
+		if v == nil {
+			delete(resource, k)
+		}
+	}
```
dmitryax, line 138: I think we should use new `map[string]string` instead of mutating original collector configuration object (including this place https://github.com/open-telemetry/opentelemetry-collector/pull/5402/files#diff-52da5566fa270d6a253f7a301b66f4a1dbda7c2c42f3781f8357efe7f08dff10R121). Then we can also remove this redundant check https://github.com/open-telemetry/opentelemetry-collector/pull/5402/files#diff-52da5566fa270d6a253f7a301b66f4a1dbda7c2c42f3781f8357efe7f08dff10R206.

WDYT?

tigrannajaryan: Done.

dmitryax (approved):

tigrannajaryan:
@bogdandrutu @mx-psi ok to merge?

config/moved_service.go
```diff
@@ -38,6 +38,13 @@ type Service struct {
 type ServiceTelemetry struct {
 	Logs    ServiceTelemetryLogs    `mapstructure:"logs"`
 	Metrics ServiceTelemetryMetrics `mapstructure:"metrics"`
+
+	// Resource specifies user-defined attributes to include with all emitted telemetry.
+	// For metrics the attributes become Prometheus labels.
```
bogdandrutu, line 43: This is not 100% true, only if the exporter is prometheus.

service/telemetry.go
```diff
@@ -109,8 +110,29 @@ func (tel *colTelemetry) initOnce(col *Collector) error {
...
+		instanceUUID, _ := uuid.NewRandom()
+		instanceID := instanceUUID.String()
+		telAttrs[semconv.AttributeServiceInstanceID] = instanceID
+	}
+
+	if AddCollectorVersionTag {
+		if _, ok := resource[semconv.AttributeServiceVersion]; !ok {
+			// AttributeServiceVersion is not specified in the config. Use the actual
+			// build version.
+			telAttrs[semconv.AttributeServiceVersion] = version.Version
```
mx-psi, line 133: On non-core distros, this will show `latest`, right? We probably want to explicitly pass a version on the Collector settings to avoid this

tigrannajaryan: I haven't verified, but I think you are right. This PR didn't change the behavior.

config/moved_service.go
tigrannajaryan: Maybe I can just delete this line to void misleading?

service/telemetry.go
mx-psi: You are right, I missed that this was not introduced by this PR. I will make a note to reproduce and open an issue then. We probably want to get rid of `version.Version` entirely and leave passing the version to the builder instead.

tigrannajaryan: There is another semi-related issue: for Agent Management the Supervisor [needs a way](https://docs.google.com/document/d/1KtH5atZQUs9Achbce6LiOaJxLbksNJenvgvyKLsJrkc/edit#heading=h.ioikt02qpy5f) to fetch the Collector's version.

If we are going to change how versions are embedded in the Collector I want to use the opportunity to see if we can make it fetchable by the Supervisor. A few options:
- Have a manifest file with version number recorded in it and include the manifest in Collector's installation packages.
- Try to fetch version number from executable directly (as a symbol?). Not very desirable since it requires platform specific code.
- Read from CLI by using `--version` flag (not very reliable since the output format is not precisely defined).
- Something else?

tigrannajaryan: I submitted a separate issue for this https://github.com/open-telemetry/opentelemetry-collector/issues/5438

PR #5408, opened 2022-05-23
[Telemetry Query Language] Updated function signal expectations

TylerHelmuth:
**Description:** 
Removes the python-like syntax examples.  Changes language to be clear functions only work one one signal at a time.

Related to #5395

### DIFF
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -92,7 +92,7 @@ The processors implementing this use case are `k8sattributesprocessor`, `resourc
 
 When looking at the use cases, there are certain common features for telemetry mutation and metric generation.
 
-- Identify the type of signal (span, metric, log, resource), or apply to all signals
+- Identify the type of signal (`span`, `metric`, `log`).
 - Navigate to a path within the telemetry to operate on it
 - Define an operation, and possibly operation arguments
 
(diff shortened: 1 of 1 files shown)


docs/processing.md
```diff
@@ -208,15 +239,24 @@ exporters:
...
     # Assuming group_by is defined in a contrib extension module, not baked into the "query" processor
     extensions: [group_by]
-    expressions:
-      - drop() where attributes["http.target"] = "/health"
-      - delete(attributes["http.request.header.authorization"])
-      - replace_wildcards("/user/*/list/*", "/user/{userId}/list/{listId}", attributes["http.target"])
-      - set(attributes["k8s_pod"], resource.attributes["k8s.pod.name"]) from metric
-      - group_by(trace_id, 2m) from span
+    traces:
+      queries:
```
bogdandrutu, line 246: so only for "traces" you have "queries"?

bogdandrutu (approved):

docs/processing.md
TylerHelmuth: Good catch, they should all have "queries"

docs/processing.md
```diff
@@ -313,5 +355,3 @@ There are some known issues and limitations that we hope to address while iterat
 - Handling array-typed attributes
 - Working on a array of points, rather than a single point
 - Metric alignment - for example defining an expression on two metrics, that may not be at the same timestamp
-- The collector has separate pipelines per signal - while the query language could apply cross-signal, we will need
```
anuraaga, line 316: It's probably still worth keeping this line. It would be nice if eventually it could happen cross signal some day as some of our examples show :)

anuraaga (approved):

docs/processing.md
TylerHelmuth: Haha I'll add it back to keep the hope alive

bogdandrutu (approved):

Issue #5431, opened 2022-05-26
Add Move() function in pdata slices


sokoide:
**Is your feature request related to a problem? Please describe.**
We are making a spanevent filter processor which filters out unnecessary filters per span based on rules we define.
Currently, we do it as below. However, RemoveIf and CopyTo are not efficient because it iterates all span events and causes a memory copy.

```go
filteredEvents, err := EvaluateStatementWithSpans(stmt, span.Events()) // EvaluateStatementWithSpans is our function

span.Events().RemoveIf(func(_ ptrace.SpanEvent) bool {
	return true
})
filteredEvents.CopyTo(span.Events())
```

**Describe the solution you'd like**
We'd like to add `Move` function in pdata slices so that we can do like this.

```go
filteredEvents, err := EvaluateStatementWithSpans(stmt, span.Events())

filteredEvents.Move(span.Events())
```
**Describe alternatives you've considered**
The one in the first section is the current implementation which is inefficient.

**Additional context**
We are sending a lot of OTEL events to a satellite OTEL collector which filters out unnecessary events and forwards it to a central OTEL collector.

bogdandrutu:
I am very confused, because your proposed solution is way slower than `RemoveIf` which is exactly defined for this use-case.

```golang
span.Events().RemoveIf(func(se ptrace.SpanEvent) bool {
	return !keepEvent(se)
})
```

If anything, you need a `Clear` func. For `MoveTo` that you propose I don't see a good reason to have when we have `MoveAndAppendTo`.

sokoide:
@bogdandrutu, thank you for the example. Here is what we'd like to achieve.

We'll filter out events based on our query with Goyacc and get the filtered events as `filteredEvents` in our span event filter processor.
Then the processor sets it back to the original span's `span.Events()` like this.

```go
filteredEvents, err := EvaluateStatementWithSpans(stmt, span.Events())
filteredEvents.Move(span.Events())
```

If we use `RemoveIf` (to clear the orig) and `MoveAndAppendTo` to replace it, it'll iterates N times in the former and copies N times in the latter.

If we add `Clear` as you suggested, it'll be better.
```go
func (es SpanEventSlice) Clear() {
    *es.orig = nil
}
```

If we have `Move`, I think it'd be even better.

```go
package internal

import (
        "testing"
)

// RemoveIf -> MoveAndAppendTo is slow because it iterates all events in RemoveIf
[...]

PR #5432, opened 2022-05-26
[pdata] add Move function (#5431)

sokoide:
**Description:**
Add `Move` function for slices in pdata so that we can replace span events in a span. This is useful to make a filter processor.
Please refer to #5431 for more detailed usage example.

**Link to tracking Issue:**
#5431 

**Testing:**
Unit tests added in pdata/internal/cmd/pdatagen/internal/base_slices.go -> Test${structtName}_Move.
Tested in all slices generated by `go run pdata/internal/cmd/pdagtagen/main.go` -> `make test`

### DIFF
--- a/pdata/internal/cmd/pdatagen/internal/base_slices.go
+++ b/pdata/internal/cmd/pdatagen/internal/base_slices.go
@@ -20,6 +20,14 @@ import (
 )
 
 const commonSliceTemplate = `
+// Move moves all elements from the current slice and replaces the dest.
+// The current slice will be cleared.
+func (es ${structName}) Move(dest ${structName}) {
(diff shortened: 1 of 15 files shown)


bogdandrutu:
@sokoide we have `MoveAndAppendTo`, is that not enough for you?

sokoide:
> @sokoide we have `MoveAndAppendTo`, is that not enough for you?

Hello @bogdandrutu, Thank you for taking a look!
Currently, we use `MoveAndAppendTo` like #5431 which needs N loops for deleteing old ones and anotehr N loops with copy to append filtered events. We'd like to improve it by the `Move` (1 pointer copy) for the span event filter processor scenario. How does that sound?

PR #5441, opened 2022-05-30
Bump to go.pkg.in/yaml.v3

mx-psi:
**Description:** Bump YAML dependency to v3, which was recently released. Changes are listed here https://github.com/go-yaml/yaml#compatibility.

Open question: Does this need a changelog? It does have some user-facing changes (e.g. `on/off` are now valid boolean values)


codeboten (approved):
I'm ok without a changelog for this, since `on/off` is only there to be backwards compatible with yaml 1.1, and we were already compatible with 1.2

PR #5453, opened 2022-06-01
Avoid duplicate code, add NewRetrievedFromYAML, replace NewRetrievedFromConf with NewRetrieved

bogdandrutu:
* It is ok to change NewRetrievedFromMap to NewRetrieved since it was just moved, not yet released.

This is a step towards supporting providing any value type.

### DIFF
--- a/confmap/provider.go
+++ b/confmap/provider.go
@@ -100,21 +100,22 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
 	}
 }
 
-// NewRetrievedFromMap returns a new Retrieved instance that contains a Map data.
-// * conf the Map that will be merged to the given map in the MergeTo.
-// * CloseFunc specifies a function to be invoked when the configuration for which it was
-//   used to retrieve values is no longer in use and should close and release any watchers
-//	 that it may have created.
-func NewRetrievedFromMap(conf *Conf, opts ...RetrievedOption) Retrieved {
+// NewRetrieved returns a new Retrieved instance that contains the data from the raw deserialized config.
+func NewRetrieved(rawConf map[string]interface{}, opts ...RetrievedOption) (Retrieved, error) {
 	set := retrievedSettings{}
 	for _, opt := range opts {
 		opt(&set)
 	}
-	return Retrieved{conf: conf, closeFunc: set.closeFunc}
+	return Retrieved{conf: NewFromStringMap(rawConf), closeFunc: set.closeFunc}, nil
 }
 
-// AsMap returns the retrieved configuration parsed as a Map.
-func (r Retrieved) AsMap() (*Conf, error) {
+// Deprecated: [v0.53.0] Use AsConf.
+func (r *Retrieved) AsMap() (*Conf, error) {
+	return r.AsConf()
+}
+
+// AsConf returns the retrieved configuration parsed as a Conf.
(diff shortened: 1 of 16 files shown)


confmap/provider.go
```diff
@@ -100,21 +102,38 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
...
 
-// NewRetrievedFromMap returns a new Retrieved instance that contains a Map data.
-// * conf the Map that will be merged to the given map in the MergeTo.
-// * CloseFunc specifies a function to be invoked when the configuration for which it was
-//   used to retrieve values is no longer in use and should close and release any watchers
-//	 that it may have created.
-func NewRetrievedFromMap(conf *Conf, opts ...RetrievedOption) Retrieved {
+// Deprecated: [v0.53.0] Use NewRetrievedFromYAML, please inform the otel collector maintainers if
+// NewRetrievedFromYAML does not cover all your needs.
+func NewRetrievedFromConf(conf *Conf, opts ...RetrievedOption) Retrieved {
```
mx-psi, line 107: I think removing this will make our (Datadog's) use case more difficult. In our case, there is no 1:1 mapping between the Datadog Agent's YAML and the underlying Collector configuration and even if there were, we don't see the YAML directly, just a Viper object.

We could do something like build a `confmap.Conf`, marshal it to YAML and pass it to `NewRetrievedFromYAML`, but it would be more cumbersome

bogdandrutu: Was looking for feedback like this. Will not deprecate this, or provide an API that accepts 'map[string] interface {}' is that ok?

mx-psi: `map[string]interface{}` would be okay, yes!

bogdandrutu: Done.

confmap/provider.go
```diff
@@ -100,21 +102,33 @@ func WithRetrievedClose(closeFunc CloseFunc) RetrievedOption {
...
 		opt(&set)
 	}
-	return Retrieved{conf: conf, closeFunc: set.closeFunc}
+	return Retrieved{conf: NewFromStringMap(rawConf), closeFunc: set.closeFunc}, nil
+}
+
+// NewRetrievedFromYAML returns a new Retrieved instance that contains the deserialized data from the yaml bytes.
+// * yamlBytes the yaml bytes that will be deserialized.
+// * opts specifies options associated with this Retrieved value, such as CloseFunc.
+func NewRetrievedFromYAML(yamlBytes []byte, opts ...RetrievedOption) (Retrieved, error) {
```
mx-psi, line 117: Could this be an internal function on `confmap/provider/internal`? The only use outside of the provider folder is on `service/mapresolver_test.go`, would it be feasible to remove it from there? I think `NewRetrieved` is general enough for fulfilling any use case, so it would be nice to have just that.

bogdandrutu: I can do that for the moment until outside world will file an issue that they need it as well :)

bogdandrutu: Actually I like it, done.

bogdandrutu:
@mx-psi PTAL

bogdandrutu:
Split into #5463 #5465

mx-psi:
All of them LGTM! :)

bogdandrutu:
All changes are merged from the split PRs.

PR #4117, opened 2021-09-26
Including metrics on number of batches handled through persistent queue

pmm-sumo:
Testing: Unit tests updated, manual tests to follow

PR #5457, opened 2022-06-02
[exporterhelper] Proposal: enable Persistent Queue feature by default

pmm-sumo:
**Description:** 

Persistent Queue is a feature that currently requires "enable_unstable" tag provided to the  the build to enable it. We have been using it in [our distro](https://github.com/SumoLogic/sumologic-otel-collector/) for several months already and it looks stable. It seems that others are using this capability as well ([example](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9327#issuecomment-1135494065))

I would like to propose enabling it by default, which also simplifies the build process a little bit.


bogdandrutu (changes requested):
Definitely we should not drop the in-memory version.

bogdandrutu (changes requested):
I read more about the code, and my previous concern is NOT valid. But `queuedRetrySender` implementation is different, metrics exposed are different, etc. This has more implications, so please make the change by deleting the "_experimental" file and bring to "_inmemory" file changes needed. So we can see exactly what changes.

bogdandrutu:
> BTW, we had https://github.com/open-telemetry/opentelemetry-collector/pull/4117#pullrequestreview-769158884 few months ago on the additional metrics that I proposed to include with persistent queue. Do you think it's the right time to get back to it or we should rather give more time the collector metrics to stabilize?

I think we should definitely do this in a separate PR, second I think (need to confirm, but this is my first thought) we should include data type as a separate label instead of adding it as part of name.

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -71,49 +71,136 @@ func (qCfg *QueueSettings) Validate() error {
...
+	requestUnmarshaler internal.RequestUnmarshaler
 }
 
-func newQueuedRetrySender(id config.ComponentID, _ config.DataType, qCfg QueueSettings, rCfg RetrySettings, _ internal.RequestUnmarshaler, nextSender requestSender, logger *zap.Logger) *queuedRetrySender {
+func (qrs *queuedRetrySender) fullName() string {
+	if qrs.signal == "" {
+		return qrs.id.String()
+	}
+	return fmt.Sprintf("%s-%s", qrs.id.String(), qrs.signal)
+}
```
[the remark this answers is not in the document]
pmm-sumo: Each signal must have a different name for the persistent queue as otherwise we would be writing multiple signal types to the same underlying storage. In any case, I changed the code a bit so for in-memory queue the name is not depending on the signal type (and stays as it was before)

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -71,49 +71,140 @@ func (qCfg *QueueSettings) Validate() error {
...
 
-func newQueuedRetrySender(id config.ComponentID, _ config.DataType, qCfg QueueSettings, rCfg RetrySettings, _ internal.RequestUnmarshaler, nextSender requestSender, logger *zap.Logger) *queuedRetrySender {
+func buildFullName(persistentQueueEnabled bool, name string, signal config.DataType) string {
+	if !persistentQueueEnabled || signal == "" {
+		return name
+	}
+	// For persistent queue, we must use signal type, since the full name is used to identify storage
+	// (which must be separate for each signal type)
+	return fmt.Sprintf("%s-%s", name, signal)
+}
```
bogdandrutu, line 100: I don't believe this is the right thing to do... I think we should have 2 different labels one for signal one for name instead of concatenating. As I mentioned, we should have this in a separate PR and debate there.

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -54,7 +53,8 @@ func NewDefaultQueueSettings() QueueSettings {
 		// This is a pretty decent value for production.
 		// User should calculate this from the perspective of how many seconds to buffer in case of a backend outage,
 		// multiply that by the number of requests per seconds.
-		QueueSize: 5000,
+		QueueSize:                5000,
+		PersistentStorageEnabled: false,
```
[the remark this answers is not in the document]
bogdandrutu: But if someone has a file_extension and a DB extension (storage) both, still better to force users to specify exactly what extension to use.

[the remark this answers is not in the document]
bogdandrutu: But if someone has a file_extension and a DB extension (storage) both, still better to force users to specify exactly what extension to use.

exporter/exporterhelper/queued_retry_inmemory.go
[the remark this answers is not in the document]
bogdandrutu: The name is also used for the "queue_size" label, that's why I don't like this change because you are changing the metrics that we emit.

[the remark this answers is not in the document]
bogdandrutu: So, you need to rename the func to "buildStorageName" and call it only when calling the constructor of the `NewPersistenQueue` the one in the struct `fullName` should be kept the same since we use that in the metrics.

exporter/exporterhelper/queued_retry_inmemory.go
pmm-sumo: Yeah, I see your point. The current design is unexpectedly changing the label used for the metric. I will fix this

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -71,49 +71,131 @@ func (qCfg *QueueSettings) Validate() error {
...
+		stopCh:         retryStopCh,
+		logger:         sampledLogger,
+		// Following three functions actually depend on queuedRetrySender
+		onTemporaryFailure: qrs.onTemporaryFailure,
+	}
+
+	if !qCfg.PersistentStorageEnabled {
+		qrs.queue = internal.NewBoundedMemoryQueue(qrs.cfg.QueueSize, func(item interface{}) {})
+	}
+	// The Persistent Queue is initialized separately as it needs extra information about the component
```
bogdandrutu, line 122: Let's initialize both in the same place?

pmm-sumo:
@swiatekm-sumo FYI

tigrannajaryan:
@pmm-sumo do you plan to continue working on this?

swiatekm-sumo:
@tigrannajaryan I'm planning to pick this up from @pmm-sumo sometime this week.

tigrannajaryan:
@swiatekm-sumo sounds good, thank you.

exporter/exporterhelper/README.md
```diff
@@ -28,24 +28,19 @@ The following configuration options can be modified:
...
+> :warning: The capability is under development. To use it, an [filestorage] extension needs to be setup, which requires OpenTelemetry Collector Contrib
 
-With this build tag set, additional configuration option can be enabled:
+To use persistent queue, following setting needs to be enabled:
 
 - `sending_queue`
-  - `persistent_storage_enabled` (default = false): When set, enables persistence via a file storage extension
-    (note, `enable_unstable` build tag needs to be enabled first, see below for more details)
+  - `storage` (default = none): When set, enables persistence and uses the component specified as a storage extension for persistent queue
+  - `persistent_storage_enabled` (default = false): When set, enables persistence using the only available storage extension (fails when no storage extension is enabled or more than one storage extension is available)
```
tigrannajaryan, line 39: Is this setting necessary? If the `storage` is unspecified the persistent queue is disabled. Isn't that enough for control? Or do we want to have this so that we can enable the storage without knowing the name of the extension?

Alternate suggestion: can we delete `persistent_storage_enabled` setting and make `*` (or some other character sequence) a special value for `storage` setting which finds the only available storage extension?

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !enable_unstable
-// +build !enable_unstable
-
 package exporterhelper // import "go.opentelemetry.io/collector/exporter/exporterhelper"
 
+// TODO: rename the file to queued_retry_sender.go after merging the PR
```
tigrannajaryan, line 17: Why this TODO? Why can't we rename before merging the PR?

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -43,6 +40,12 @@ type QueueSettings struct {
 	NumConsumers int `mapstructure:"num_consumers"`
 	// QueueSize is the maximum number of batches allowed in queue at a given time.
 	QueueSize int `mapstructure:"queue_size"`
+	// StorageID if not empty, enables the persistent storage and uses the component specified
+	// as a storage extension for the persistent queue
+	StorageID *config.ComponentID `mapstructure:"storage"`
+	// StorageEnabled describes whether persistence via a file storage extension is enabled using the single
+	// default storage extension.
+	StorageEnabled bool `mapstructure:"persistent_storage_enabled"`
```
tigrannajaryan, line 48: What happens is `storage` is specified and `persistent_storage_enabled` is also true? Which setting wins?

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -71,49 +75,142 @@ func (qCfg *QueueSettings) Validate() error {
...
+		stopCh:         retryStopCh,
+		logger:         sampledLogger,
+		// Following three functions actually depend on queuedRetrySender
+		onTemporaryFailure: qrs.onTemporaryFailure,
+	}
+
+	if qCfg.StorageEnabled || qCfg.StorageID != nil {
+		qrs.queue, qrs.queueStartFunc = internal.NewPersistentQueue(qrs.fullName, qrs.signal, qrs.cfg.QueueSize, qrs.logger, qrs.requestUnmarshaler)
+		// TODO: following can be further exposed as a config param rather than relying on a type of queue
+		qrs.requeuingEnabled = true
```
tigrannajaryan, line 171: Looks like this is a new functionality that didn't exist before. Can we remove this from this PR and add in a subsequent PR? This PR should be only about enabling Persistent Queue by default.

exporter/exporterhelper/queued_retry_test.go
```diff
@@ -56,7 +56,7 @@ func TestQueuedRetry_DropOnPermanentError(t *testing.T) {
 	})
 
 	ocs.run(func() {
-		// This is asynchronous so it should just enqueue, no errors expected.
+		// This is asynchronous so it should just enqueue, no errors expectedError.
```
tigrannajaryan, line 59: Is this change and other similar changes below intentional?

tigrannajaryan:
> Allright, from what I can see @pmm-sumo addressed all the review comments from @bogdandrutu and we're mostly missing some tests to make codecov happy. I'll add those in my own branch, then open a new PR referencing this one. Is that ok @bogdandrutu @tigrannajaryan ?

@swiatekm-sumo sounds good. I also left some comments above.

exporter/exporterhelper/queued_retry_inmemory.go
pmm-sumo: @tigrannajaryan this is due to [this suggestion](https://github.com/open-telemetry/opentelemetry-collector/pull/5457#pullrequestreview-994187835)

exporter/exporterhelper/README.md
pmm-sumo: `storage` is a new setting which allows to explicitly select which storage extension to use. However, when only only storage extension is specified in the config, we want to be able to automatically attach it and having `storage: ...` specified is not needed. It would be still helpful though to have control if file-backed or memory-backed queue is used. Hence `persistent_storage_enabled`. I think any suggestions on how to make it more elegant are more than welcome but I think we want to retain such kind of functionality

exporter/exporterhelper/queued_retry_inmemory.go
pmm-sumo: I think the more trickier example is `persistent_storage_enabled: false` and `storage` referring some existing storage extension. Basing on the name, I would assume `persistent_storage_enabled` should have precedence

exporter/exporterhelper/queued_retry_inmemory.go
swiatekm-sumo: It used to be enabled here: https://github.com/open-telemetry/opentelemetry-collector/blob/e8dd9536f5e53228606c325f12d4bfc4b4293acc/exporter/exporterhelper/queued_retry_experimental.go#L170 so this actually preserves existing behaviour.

exporter/exporterhelper/queued_retry_test.go
swiatekm-sumo: Looks like an overzealous replace, I'll fix it.

tigrannajaryan:
Is this entirely superseded by https://github.com/open-telemetry/opentelemetry-collector/pull/5711
If yes, then close this PR.

PR #5485, opened 2022-06-06
Remove confmap.Conf.Set, not used anywhere for the moment

bogdandrutu:


bogdandrutu:
cc @mx-psi can you tell me about your usage in DD?

I will "deprecate" it, but initially I am trying to see if contrib breaks.

mx-psi:
> cc @mx-psi can you tell me about your usage in DD?
> 
> I will "deprecate" it, but initially I am trying to see if contrib breaks. **Update:** contrib passed, so change this to deprecated.

We use it to build up a `confmap.Conf` manually from Viper (see [here](https://github.com/DataDog/datadog-agent/blob/c4a81894ddbc0be09b1f6683ffbeb372d8d8426f/pkg/otlp/config.go#L60)). It's feasible to replace it with a `Merge` or handle it with something else, I just trusted this code more than using Viper 😄

mx-psi (approved):

PR #5487, opened 2022-06-07
Add per-metric stability guidelines

djaglowski:
Resolves [(contrib)#9963](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/9963).

Possibly these should be defined in contrib or in a separate file. I'd appreciate feedback on the guidelines themselves as well as appropriate placement in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Objectives:
 - Extensible: Customizable without touching the core code.
 - Unified: Single codebase, deployable as an agent or collector with support for traces, metrics and logs.
 
-## Stability levels
+### Component stability levels
 
 The collector components and implementation are in different stages of stability, and usually split between
 functionality and configuration. The status for each component is available in the README file for the component. While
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,34 +63,67 @@ Objectives:
...
 The component is ready for general availability. Bugs and performance problems should be reported and there's an expectation that the component owners will work on them. Breaking changes, including configuration options and the component's output are not expected to happen without prior notice, unless under special circumstances.
 
-### Deprecated
+#### Deprecated
 
 The component is planned to be removed in a future version and no further support will be provided. Note that new issues will likely not be worked on. When a component enters "deprecated" mode, it is expected to exist for at least two minor releases. See the component's readme file for more details on when a component will cease to exist.
 
+### Metric stability levels
+
+Receivers that produce metrics should declare a stability level for each metric. A for each metric may not exceed the The status for each metric should be listed in the receiver's `documentation.md` file according to the following definitions:
```
bogdandrutu, line 96: `A for each metric` something sounds off here.

README.md
```diff
@@ -63,34 +63,67 @@ Objectives:
...
+The component is ready for general availability. Bugs and performance problems should be reported and there's an expectation that the component owners will work on them. Breaking changes, including configuration options and the component's output are not expected to happen without prior notice, unless under special circumstances.
+
+#### Deprecated
+
+The component is planned to be removed in a future version and no further support will be provided. Note that new issues will likely not be worked on. When a component enters "deprecated" mode, it is expected to exist for at least two minor releases. See the component's readme file for more details on when a component will cease to exist.
+
+### Stability level progression
+
+Comonents and metric stability levels should progress according to the following guidelines:
+- A component that produces metrics should produce at least one metric of the same stability level. For example, a Stable component should produce at least one Stable metric.
```
bogdandrutu, line 121: Why? If we did not have metrics in a stable component, and start adding them now? How do I get to stable directly?

README.md
```diff
@@ -63,34 +63,67 @@ Objectives:
...
+
+#### Stable
+
+The component is ready for general availability. Bugs and performance problems should be reported and there's an expectation that the component owners will work on them. Breaking changes, including configuration options and the component's output are not expected to happen without prior notice, unless under special circumstances.
+
+#### Deprecated
+
+The component is planned to be removed in a future version and no further support will be provided. Note that new issues will likely not be worked on. When a component enters "deprecated" mode, it is expected to exist for at least two minor releases. See the component's readme file for more details on when a component will cease to exist.
+
+### Stability level progression
```
bogdandrutu, line 118: Are you allowed to remove a deprecated metric, or disable to save cost (backend cost as well as runtime cost)?

README.md
djaglowski: My intention here is to address primarily how it "should" be. Then we would take steps to get there.

As to the reasoning here, this just seems necessary in order to give meaning to the component status. I would question whether a scraper can be stable if it produces no stable metrics.

README.md
djaglowski: Fixed

README.md
djaglowski: Thanks for the reminder to incorporate default enabled state as well. I will add considerations for this.

djaglowski: I've proposed some guidelines for enabled status within each stability level.

README.md
djaglowski: I was mostly trying to address scrapers, but didn't use the term because I'm not sure it's a user-facing term. (I said receivers that produce metrics instead, but maybe this isn't specific enough.)

That said, I think all of this would apply to own telemetry as well, with the exception of this line.

bogdandrutu: How is this different than "receiver that produce logs" or "spans" should we actually call this "telemetry"?

I would probably have a different term for "own observability telemetry" vs "telemetry on behalf of others" :))

djaglowski: > I would probably have a different term for "own observability telemetry" vs "telemetry on behalf of others" :))

Makes sense. I think "own telemetry" would be clear and is already used elsewhere (eg OpAMP). Perhaps "scraped telemetry" then for the other term. 

> How is this different than "receiver that produce logs" or "spans" should we actually call this "telemetry"?

The way I see it is that metrics have named instances, each of which has its own detailed data model. Maybe I'm overlooking something, but I don't think logs and spans have an equivalent really.

In any case, I can try to organize this PR to leave room for future stability guidelines for other signals, but I only have detailed thoughts for metrics at this point. What do you think of organizing it like this:

```
## Metric stability levels

### "Own metrics" vs "Scraped metrics"

// TODO describe the difference and give a couple examples

The following levels and guidelines apply to "own metrics" and "scraped metrics", except where otherwise noted.

...

### Scraped metrics

Receivers whose sole purpose is to scrape metrics (ie. "scrapers") should produce at least one metric of the same stability level. For example, a Stable component should produce at least one Stable metric.
```

Then if someone can reuse some of these guidelines for logs and/or spans, they can extract them into a "telemetry" header.

README.md
```diff
@@ -63,34 +63,67 @@ Objectives:
...
+#### Stable
 
 The component is ready for general availability. Bugs and performance problems should be reported and there's an expectation that the component owners will work on them. Breaking changes, including configuration options and the component's output are not expected to happen without prior notice, unless under special circumstances.
 
-### Deprecated
+#### Deprecated
 
 The component is planned to be removed in a future version and no further support will be provided. Note that new issues will likely not be worked on. When a component enters "deprecated" mode, it is expected to exist for at least two minor releases. See the component's readme file for more details on when a component will cease to exist.
 
+### Metric stability levels
```
jpkrohling, line 94: How about adding a "Since" property, stating the version of the collector the metric started being reported?

djaglowski:
[A closely related issue was opened on the spec repo](https://github.com/open-telemetry/opentelemetry-specification/issues/2610) at roughly the same time as this proposal was drafted. I think that issue should be resolved before this moves forward.

PR #5522, opened 2022-06-13
Cleanup internal/testcomponents

bogdandrutu:
* Expose example processor to allow start/shutdown tests.

### DIFF
--- a/internal/testcomponents/example_exporter.go
+++ b/internal/testcomponents/example_exporter.go
@@ -25,13 +25,12 @@ import (
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
-// ExampleExporter is for testing purposes. We are defining an example config and factory
-// for "exampleexporter" exporter type.
-type ExampleExporter struct {
+const expType = "exampleexporter"
+
+// ExampleExporterConfig config for ExampleExporter.
+type ExampleExporterConfig struct {
 	config.ExporterSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct
 }
-
-const expType = "exampleexporter"
 
 // ExampleExporterFactory is factory for ExampleExporter.
 var ExampleExporterFactory = component.NewExporterFactory(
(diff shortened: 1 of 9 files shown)


internal/testcomponents/example_exporter.go
```diff
@@ -25,81 +25,76 @@ import (
...
 	component.WithTracesExporter(createTracesExporter),
 	component.WithMetricsExporter(createMetricsExporter),
 	component.WithLogsExporter(createLogsExporter))
 
-// CreateDefaultConfig creates the default configuration for the Exporter.
 func createExporterDefaultConfig() config.Exporter {
-	return &ExampleExporter{
+	return &struct {
+		config.ExporterSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct
+	}{
```
tigrannajaryan, line 41: This looks different from what we do with real components. Perhaps for the example to be closer to the reality keep the ExampleExporterConfig struct declared at the top lexical level.

internal/testcomponents/example_exporter.go
```diff
@@ -25,81 +25,76 @@ import (
...
-// ExampleExporter is for testing purposes. We are defining an example config and factory
-// for "exampleexporter" exporter type.
-type ExampleExporter struct {
-	config.ExporterSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct
-}
-
 const expType = "exampleexporter"
 
-// ExampleExporterFactory is factory for ExampleExporter.
+// ExampleExporterFactory is factory for ExampleExporterConfig.
```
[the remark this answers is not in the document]
bogdandrutu: Mistake, thanks for catching this.

internal/testcomponents/example_exporter.go
bogdandrutu: This is not an example despite the name, it is used for testing. But will update to follow the standard.

internal/testcomponents/example_processor_test.go
```diff
@@ -0,0 +1,37 @@
...
+	rcv := &ExampleProcessor{}
+	host := componenttest.NewNopHost()
+	assert.False(t, rcv.Started)
+	err := rcv.Start(context.Background(), host)
+	assert.NoError(t, err)
+	assert.True(t, rcv.Started)
+
+	err = rcv.Shutdown(context.Background())
+	assert.NoError(t, err)
+	assert.True(t, rcv.Started)
```
fatsheep9146, line 36: I think these lines should be same as other tests

```
	assert.False(t, rcv.Stopped)
	err = rcv.Shutdown(context.Background())
	assert.NoError(t, err)
	assert.True(t, rcv.Stopped)
```

bogdandrutu: Thanks, done.

mx-psi (approved):

PR #5542, opened 2022-06-15
Extract builder for pipelines components into one struct

bogdandrutu:

### DIFF
--- a/service/internal/pipelines/pipelines.go
+++ b/service/internal/pipelines/pipelines.go
@@ -1,0 +1,169 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pipelines // import "go.opentelemetry.io/collector/service/internal/pipelines"
+
+import (
+	"context"
+	"net/http"
+	"sort"
+
+	"go.uber.org/multierr"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/service/internal/builder"
+	"go.opentelemetry.io/collector/service/internal/zpages"
+)
+
+const (
+	zPipelineName  = "zpipelinename"
+	zComponentName = "zcomponentname"
+	zComponentKind = "zcomponentkind"
+)
+
+// Pipelines is set of all pipelines created from exporter configs.
+type Pipelines struct {
+	telemetry component.TelemetrySettings
+
+	exporters *builder.BuiltExporters
+	pipelines builder.BuiltPipelines
+	receivers builder.Receivers
+}
+
+// Build builds all pipelines from config.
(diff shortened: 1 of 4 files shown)


service/internal/pipelines/pipelines.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+const (
+	zPipelineName  = "zpipelinename"
+	zComponentName = "zcomponentname"
+	zComponentKind = "zcomponentkind"
+)
+
+// BuiltPipelines is set of all pipelines created from exporter configs.
+type BuiltPipelines struct {
```
mx-psi, line 37: What about simply `Pipelines`?

service/internal/pipelines/pipelines.go
```diff
@@ -0,0 +1,169 @@
...
+	telemetry component.TelemetrySettings
+
+	exporters *builder.BuiltExporters
+	pipelines builder.BuiltPipelines
+	receivers builder.Receivers
+}
+
+// Build builds all pipelines from config.
+func Build(ctx context.Context, telemetry component.TelemetrySettings, buildInfo component.BuildInfo, cfg *config.Config, factories component.Factories) (*BuiltPipelines, error) {
```
mx-psi, line 46: Should we have a `PipelinesFactory` struct instead? Something like:

```
type PipelinesFactory struct{
  // ... 
}

// NewFactory creates a new Pipelines factory.
func NewFactory(telemetry component.TelemetrySettings, buildInfo component.BuildInfo, factories component.Factories) (*PipelinesFactory, error)

// Build a Pipeline for a given configuration.
func (p *PipelinesFactory) Build(ctx context.Context, cfg *config.Config) (*BuiltPipelines, error)
```

service/internal/pipelines/pipelines.go
```diff
@@ -0,0 +1,169 @@
...
+		exporters: exporters,
+		pipelines: pipelines,
+	}, nil
+}
+
+// StartAll starts all pipelines.
+//
+// Start with exporters, processors (in revers configured order), then receivers.
+// This is important so that components that are earlier in the pipeline and reference components that are
+// later in the pipeline do not start sending data to later pipelines which are not yet started.
```
mx-psi, line 74:
```suggestion
// later in the pipeline do not start sending data to later components which are not yet started.
```

service/internal/pipelines/pipelines.go
mx-psi, line 46: It feels weird to pass both `component.TelemetrySettings` and `service.Config` as arguments, given that the config struct has `service.ConfigServiceTelemetry` within it. I definitely want to be able to customize the `component.TelemetrySettings` for my use case, and I think several other people have expressed their need for this, so not passing the telemetry structs in some way does not feel like an option. Some possibilities are:

1. Move `service.ConfigServiceTelemetry` out of its current location in the configuration. I think this is not a realistic option if it affects the YAML, since it will break too many things, but we could do it with a custom unmarshaling function if we want to go this way.
2. Do a 'telemetry provider' kind of thing. It makes this somewhat more clunky to use for e.g. my use case, but maybe it's the way to go?
3. Do nothing and just document it; accept that we can't change it and that we have this weirdness in the API

bogdandrutu: I will track this in a separate issue https://github.com/open-telemetry/opentelemetry-collector/issues/5548

service/internal/pipelines/pipelines.go
bogdandrutu: Done.

service/internal/pipelines/pipelines.go
bogdandrutu: Done.

service/internal/pipelines/pipelines.go
bogdandrutu: Will address this issue separately, I am planning to actually send only what is needed (no telemetry settings).

mx-psi: 👍 , yeah, thinking about this later, the `Extensions` field is also a problem, so probably we can just send exactly what we need

mx-psi (approved):

PR #5544, opened 2022-06-15
Add tests for the new pipelines builder

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/5542


service/internal/pipelines/pipelines_test.go
```diff
@@ -0,0 +1,292 @@
...
+		{configFile: "unknown_processor_factory.yaml"},
+		{configFile: "unknown_receiver_config.yaml"},
+		{configFile: "unknown_receiver_factory.yaml"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.configFile, func(t *testing.T) {
+			if test.configFile == "unknown_receiver_config.yaml" {
+				t.Skip("This is a small issue with current implementation which will be fixed in #5512." +
+					"In real binary this will not be hit, since the validation of the config will catch this issue in advance.")
```
mx-psi, line 229: What's the issue here?

mx-psi (approved):

service/internal/pipelines/pipelines_test.go
bogdandrutu: The problem is that we build the receivers by looking at configs, and if used in one pipeline we create/initialize it. With this logic if there is a receiver in a pipeline without a config we don't create/initialize it, but this is not a "Valid" config and fails `config.Validate()`.

Issue #5550, opened 2022-06-16
Make consumererror a separate module, and include exporterhelper.NewThrottleRetry


dashpole:
**Is your feature request related to a problem? Please describe.**

Context: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/433

I'd like to use consumererror outside of the upstream repositories in our library.

Right now consumererror isn't its own module, so it pulls in more dependencies than needed (which might make breaking changes to core harder).

Second, I think it might make more sense for `exporterhelper.NewThrottleRetry` to live in `consumererror` (maybe NewThrottled?), so that users of `consumererror` don't need to depend on exporterhelper as well.

Third, the google cloud APIs all use GRPC clients, similar to the otlp exporter, and would end up duplicating the retry-handling code: https://github.com/open-telemetry/opentelemetry-collector/blob/772a3755f4e8e132e4a4068d2884ecb3ddbaa48f/exporter/otlpexporter/otlp.go#L131.  It would be nice to make that logic usable as a library (possible also in consumererror) so it can be used by other exporters without duplicating it.

bogdandrutu:
The consumererror as is right now is part of a small package `consumer`.

dashpole:
Do you think it is OK us to take an external dependency on `consumer`?

bogdandrutu:
> Do you think it is OK us to take an external dependency on consumer?

We are working on stabilizing this package as we speak, see https://github.com/open-telemetry/opentelemetry-collector/milestone/31. When that is done for sure, before that I would only recommend if you are a "trusted" client and will not annoy if we still break some of the APIs.

bogdandrutu:
Closing this issue for the "separate module part".

PR #5551, opened 2022-06-17
Make internal/testdata and consumer their own go modules

dashpole:
**Description:**

Part of https://github.com/open-telemetry/opentelemetry-collector/issues/5550.

This makes it possible to use consumer without pulling in the entire base module.  The consumer module depends on internal/testdata


dashpole:
I chose to add a go module in consumer, rather than consumer/consumererror because they have roughly the same set of dependencies (both depend only on pdata).  I can change this if requested.

.github/dependabot.yml
```diff
@@ -18,6 +18,14 @@ updates:
...
     schedule:
       interval: "weekly"
+  - package-ecosystem: "gomod"
+    directory: "/consumer"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "gomod"
+    directory: "/internal/testdata"
+    schedule:
+      interval: "weekly"
```
bogdandrutu, line 28: This is a package that we don't want to expose.

consumer/consumererror/throttle.go
```diff
@@ -0,0 +1,53 @@
...
+	return "Throttle (" + t.delay.String() + "), error: " + t.err.Error()
+}
+
+func (t throttled) Unwrap() error {
+	return t.err
+}
+
+// ThrottleRetryDelay returns the duration of time we should wait to retry, or
+// zero if the error is not a throttled retry error
+func ThrottleRetryDelay(err error) time.Duration {
```
bogdandrutu, line 40: Why this instead of exposing "Throttle" type and add a func on it `RetryDelay`?

bogdandrutu (commented):
I would for sure separate the 2 independent changes:
1. The move of the error.
2. The exposing of the package.

.github/dependabot.yml
dashpole: I don't believe this exposes any additional modules.  They are still internal, IIUC.  It does reduce the number of dependencies you pull in when you reference this package.

consumer/consumererror/throttle.go
dashpole: Moved to #5569

.github/dependabot.yml
dashpole: I started working on moving required functionality into the consumer package, but it involves copying basically the entire testdata package into consumer, and didn't seem worth it.  The cleanest way would be just to duplicate the package inside consumer: https://github.com/open-telemetry/opentelemetry-collector/pull/5578

PR #5569, opened 2022-06-21
Move NewThrottleRetry from exporterhelper to consumererror

dashpole:
**Description:**

Part of https://github.com/open-telemetry/opentelemetry-collector/issues/5550.

NewThrottleRetry is an error that is similar in kind to a permanent error: it informs callers how to deal with retries. I think it makes the most sense for the two to live in the same package, since users of NewPermanent will likely also need NewThrottleRetry.

This also makes getting the duration of the throttle retry error cleaner.

It deprecates the previous NewThrottleRetry function.

**Testing:**

Added unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -378,6 +378,7 @@ There isn't a valid core binary for this release. Use v0.57.2 instead.
     that takes `pcommon.ImmutableFloat64Slice` instead of []float64.
   - `HistogramDataPoint.MExplicitBounds` func func is deprecated in favor of `HistogramDataPoint.ExplicitBounds`
     returns `pcommon.ImmutableFloat64Slice` instead of []float64.
+- Deprecate `exporterhelper.NewThrottleRetry` in favor of `consumererror.NewThrottleRetry` (#5551)
 
 ### 💡 Enhancements 💡
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -378,7 +378,7 @@ There isn't a valid core binary for this release. Use v0.57.2 instead.
     that takes `pcommon.ImmutableFloat64Slice` instead of []float64.
   - `HistogramDataPoint.MExplicitBounds` func func is deprecated in favor of `HistogramDataPoint.ExplicitBounds`
     returns `pcommon.ImmutableFloat64Slice` instead of []float64.
-- Deprecate `exporterhelper.NewThrottleRetry` in favor of `consumererror.NewThrottleRetry` (#5551)
+- Deprecate `exporterhelper.NewThrottleRetry` in favor of `consumererror.NewThrottleRetry` (#5569)
 
 ### 💡 Enhancements 💡
 
--- a/consumer/consumererror/throttle.go
+++ b/consumer/consumererror/throttle.go
@@ -1,0 +1,38 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package consumererror // import "go.opentelemetry.io/collector/consumer/consumererror"
+
+import (
+	"time"
+)
+
+type Throttle struct {
+	delay time.Duration
+}
+
+func (t Throttle) Error() string {
+	return "Throttle (" + t.delay.String() + ")"
+}
+
+func (t Throttle) RetryDelay() time.Duration {
+	return t.delay
+}
+
+// NewThrottleRetry creates a new throttle retry error.
+func NewThrottleRetry(delay time.Duration) error {
+	return Throttle{
+		delay: delay,
+	}
+}
--- a/exporter/exporterhelper/queued_retry.go
+++ b/exporter/exporterhelper/queued_retry.go
@@ -321,25 +321,11 @@ func (qrs *queuedRetrySender) send(req internal.Request) error {
 }
 
 // TODO: Clean this by forcing all exporters to return an internal error type that always include the information about retries.
-type throttleRetry struct {
-	err   error
-	delay time.Duration
-}
-
-func (t throttleRetry) Error() string {
-	return "Throttle (" + t.delay.String() + "), error: " + t.err.Error()
-}
-
-func (t throttleRetry) Unwrap() error {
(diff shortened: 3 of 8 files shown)


consumer/consumererror/throttle.go
```diff
@@ -0,0 +1,53 @@
...
+	return "Throttle (" + t.delay.String() + "), error: " + t.err.Error()
+}
+
+func (t throttled) Unwrap() error {
+	return t.err
+}
+
+// ThrottleRetryDelay returns the duration of time we should wait to retry, or
+// zero if the error is not a throttled retry error
+func ThrottleRetryDelay(err error) time.Duration {
```
dashpole, line 40: From @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/pull/5551:

> Why this instead of exposing "Throttle" type and add a func on it RetryDelay?

bogdandrutu: Why do you need to change the constructor?

dashpole: I was assuming you wanted to add additional data to the error, but I can't think of any examples where that would be useful :).

bogdandrutu: I am waiting for you to do the change, unless I misunderstand something here.

dashpole: Ah, OK.  I'll make the change.  I thought you were just interested in why I had deviated from the previous implementation.  I can go back to that.

dashpole: Sorry for the delay.  Done.

dashpole:
Sorry for the wait @bogdandrutu.  This should be ready for review now

consumer/consumererror/throttle.go
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package consumererror // import "go.opentelemetry.io/collector/consumer/consumererror"
+
+import (
+	"time"
+)
+
+// TODO: Clean this by forcing all exporters to return an internal error type that always include the information about retries.
```
bogdandrutu, line 21: No longer the case?

exporter/exporterhelper/queued_retry.go
```diff
@@ -320,26 +320,10 @@ func (qrs *queuedRetrySender) send(req internal.Request) error {
...
-func (t throttleRetry) Error() string {
-	return "Throttle (" + t.delay.String() + "), error: " + t.err.Error()
-}
-
-func (t throttleRetry) Unwrap() error {
-	return t.err
-}
-
 // NewThrottleRetry creates a new throttle retry error.
+// Deprecated: [v0.54.0] Use consumererror.NewThrottleRetry instead
```
bogdandrutu, line 324: Update to `0.60.0`

consumer/consumererror/throttle.go
```diff
@@ -0,0 +1,46 @@
...
+func (t Throttle) Unwrap() error {
+	return t.err
+}
+
+func (t Throttle) RetryDelay() time.Duration {
+	return t.delay
+}
+
+// NewThrottleRetry creates a new throttle retry error.
+func NewThrottleRetry(err error, delay time.Duration) error {
```
bogdandrutu, line 41: Do we do wrapping here or ask to use something like `multierr.Append`?

consumer/consumererror/throttle.go
dashpole: I don't think this forces exporters to use this.  The TODO is copied from the old implementation.

consumer/consumererror/throttle.go
dashpole: We are wrapping.  Throttle has an Unwrap() function.  Users don't need to use multierr.

exporter/exporterhelper/queued_retry.go
dashpole: done

consumer/consumererror/throttle.go
bogdandrutu: I understand what we are doing, sorry for the wrong question. Was trying to ask if we should do automatic wrapping or ask users to use something like `multierr.Append`?

dashpole: I think we should automatically wrap here.  The expectation is that the error passed is based on a single request, rather than a multierr error, so I think wrapping is correct here.  If there are multiple things that need to be retried, those could be combined via multierr.Append and then potentially un-combined when we need to find the longest retry time, but we should still be wrapping in this case.

bogdandrutu: > The expectation is that the error passed is based on a single request, rather than a multierr error, so I think wrapping is correct here.

It does not matter if it is one or more requests, the fact that matters is there needs to be an existing error.

Still not understand why enforce wrapping, sorry. You are saying that we cannot have a Throttle error without having another error, which I am not sure I understand.

My suggestion is that `Throttle` is an independent error, and if users want to say that `Throttle` is caused by another error they will do `multierr.Append(NewThrottle(...), myErr)` or `multierr.Combine(NewThrottle(...), myErr)`. 

I am trying to understand if we need to force users to have a previous error or not.

dashpole: I would expect the error to give some reason as to "why" we need to retry or are throttled.

If we use multierr.Append/Combine, then it isn't clear if those errors were for two different events (e.g. in the GCP exporter, maybe I split a batch of data into two, made two requests, and got two errors), or for the same event (e.g. one request with an error which I am retrying).

bogdandrutu: OK, what is the order of arguments in other cases like this (if you seen any error like this one)? For me seems more natural to pass the delay first. Also in all `fmt.Errorf` statements I see the "cause" as last thing.

consumer/consumererror/throttle.go
dashpole: left the comment where it was in exporterhelper.

consumer/consumererror/throttle.go
dashpole: mmm.  I looked at a few other error classes (e.g. gRPC status), and it seems better to not include other errors.  Putting the error into the throttle error doesn't actually make the error more readable, since they get printed out as a list anyways.

I simplified this by removing the error as an argument

CHANGELOG.md
```diff
@@ -378,6 +378,7 @@ There isn't a valid core binary for this release. Use v0.57.2 instead.
     that takes `pcommon.ImmutableFloat64Slice` instead of []float64.
   - `HistogramDataPoint.MExplicitBounds` func func is deprecated in favor of `HistogramDataPoint.ExplicitBounds`
     returns `pcommon.ImmutableFloat64Slice` instead of []float64.
+- Deprecate `exporterhelper.NewThrottleRetry` in favor of `consumererror.NewThrottleRetry` (#5551)
```
dashpole, line 381:
```suggestion
- Deprecate `exporterhelper.NewThrottleRetry` in favor of `consumererror.NewThrottleRetry` (#5569)
```

PR #5571, opened 2022-06-21
Fix initialization of the MetricProvider

bogdandrutu:
The problem was that the MetricProvider is initialized into the "service.telemetry.MetricProvider" after components were created, so all components will get a no-op implementation even if the feature gate is enabled. This change was not a trivial change because the process telemetry initialization requires the ballast size, which is available after the extensions are initialized, because of that I split the initialization of the MetricProvider/oc.Registry from the initialization of the process telemetry.

### DIFF
--- a/service/collector_test.go
+++ b/service/collector_test.go
@@ -133,6 +133,7 @@ func TestCollectorReportError(t *testing.T) {
 }
 
 func TestCollectorFailedShutdown(t *testing.T) {
+	t.Skip("This test was using telemetry shutdown failure, switch to use a component that errors on shutdown.")
 	factories, err := componenttest.NopFactories()
 	require.NoError(t, err)
 
@@ -143,7 +144,7 @@ func TestCollectorFailedShutdown(t *testing.T) {
 		BuildInfo:      component.NewDefaultBuildInfo(),
 		Factories:      factories,
 		ConfigProvider: cfgProvider,
-		telemetry:      &mockColTelemetry{},
+		telemetry:      newColTelemetry(featuregate.NewRegistry()),
 	})
 	require.NoError(t, err)
 
@@ -265,7 +266,7 @@ func ownMetricsTestCases(version string) []ownMetricsTestCase {
 		}}
 }
 
-func testCollectorStartHelper(t *testing.T, telemetry collectorTelemetryExporter, tc ownMetricsTestCase) {
+func testCollectorStartHelper(t *testing.T, telemetry *telemetryInitializer, tc ownMetricsTestCase) {
 	factories, err := componenttest.NopFactories()
 	zpagesExt := zpagesextension.NewFactory()
 	factories.Extensions[zpagesExt.Type()] = zpagesExt
@@ -288,7 +289,7 @@ func testCollectorStartHelper(t *testing.T, telemetry collectorTelemetryExporter
 		// Set the metrics address to expose own metrics on.
 		"service::telemetry::metrics::address": metricsAddr,
 	}
-	// Also include resource attributes under the service.telemetry.resource key.
+	// Also include resource attributes under the service::telemetry::resource key.
 	for k, v := range tc.userDefinedResource {
 		extraCfgAsProps["service::telemetry::resource::"+k] = v
 	}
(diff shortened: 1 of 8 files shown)


service/collector_test.go
```diff
@@ -133,6 +133,7 @@ func TestCollectorReportError(t *testing.T) {
 }
 
 func TestCollectorFailedShutdown(t *testing.T) {
+	t.Skip("This test was using telemetry shutdown failure, switch to use a component that errors on shutdown.")
```
tigrannajaryan, line 136: Is this unnecessary now? Maybe delete it?

service/service.go
```diff
@@ -22,27 +22,31 @@ import (
...
+	colTelemetry *colTelemetry
 }
 
 func newService(set *settings) (*service, error) {
 	srv := &service{
 		buildInfo: set.BuildInfo,
 		config:    set.Config,
 		telemetry: component.TelemetrySettings{
+			Logger:         zap.NewNop(),
 			TracerProvider: trace.NewNoopTracerProvider(),
```
tigrannajaryan, line 50: Why do we initialize with Nop instances here, and later on lines 62-70 immediately reassign the values? Looks like  this Nop initialization is not needed.

service/service.go
```diff
@@ -22,27 +22,31 @@ import (
...
 type service struct {
-	buildInfo component.BuildInfo
-	config    *Config
-	telemetry component.TelemetrySettings
-	host      *serviceHost
+	buildInfo    component.BuildInfo
+	config       *Config
+	telemetry    component.TelemetrySettings
+	host         *serviceHost
+	colTelemetry *colTelemetry
```
tigrannajaryan, line 41: It is a bit confusing that we have `telemetry` and `colTelemetry` fields. Later in the code when they are used it is not clear what the difference is. 

Some renaming suggestions:
- `service.colTelemetry` -> `service.telemetryInitializer`
- `service.telemetry` -> `service.telemetrySettings` or (`TelemetrySettings` -> `TelemetryProviders` and `service.telemetry` -> `service.telemetryProviders`)

service/service.go
```diff
@@ -72,6 +82,10 @@ func newService(set *settings) (*service, error) {
 		return nil, fmt.Errorf("cannot build pipelines: %w", err)
 	}
 
+	if err = telemetry.RegisterProcessMetrics(srv.colTelemetry.ocRegistry, getBallastSize(srv.host)); err != nil {
```
tigrannajaryan, line 85: Add a comment to explain why this is after `Build`.

service/service.go
```diff
@@ -64,6 +69,11 @@ func newService(set *settings) (*service, error) {
 		return nil, fmt.Errorf("failed to get logger: %w", err)
 	}
 
+	if err = srv.colTelemetry.init(set.BuildInfo, srv.telemetry.Logger, set.Config.Service.Telemetry, set.AsyncErrorChannel); err != nil {
+		return nil, fmt.Errorf("failed to initialize telemetry: %w", err)
+	}
+	srv.telemetry.MeterProvider = srv.colTelemetry.mp
```
tigrannajaryan, line 75: Can we return this from `colTelemetry.init()`? Generally, I think we should avoid working with the private fields of `colTelemetry` here. Perhaps to enforce it move `colTelemetry` to a separate package.

tigrannajaryan (commented):
I feel like telemetry initialization could use some refactoring. It is a bit difficult to understand and `service` also works with the internal fields of `colTelemetry`, violating its encapsulation and separation of concerns.
Can be done in a separate PR.

service/service.go
tigrannajaryan: Yes, https://github.com/open-telemetry/opentelemetry-collector/pull/5570 fixed it for TraceProvider. We have the same issue with the Logger here.

service/collector_test.go
bogdandrutu: As you can see it is a TODO, to use a different design, and re-enable. Not sure how to do this without blowing the size of this PR.

service/service.go
bogdandrutu: Done.

service/service.go
bogdandrutu: Will do that, when moving to `service/telemetry` package, in a soon followup PR.

service/service.go
bogdandrutu: Done, longer names (Java ftw 👍🏽  )

tigrannajaryan (approved):

PR #5580, opened 2022-06-22
add stability level to component factory

codeboten:
The following allows the setting of a stability level on the factory. This will let us consolidate warnings across all components to notify users of deprecated, unmaintained components.

### DIFF
--- a/component/component.go
+++ b/component/component.go
@@ -105,6 +105,9 @@ const (
 )
 
 // StabilityLevel represents the stability level of the component created by the factory.
+// The stability level is used to determine if the component should be used in production
+// or not. For more details see:
+// https://github.com/open-telemetry/opentelemetry-collector#stability-levels
 type StabilityLevel int
 
 const (
--- a/component/component.go
+++ b/component/component.go
@@ -104,6 +104,37 @@ const (
 	KindExtension
 )
 
+// StabilityLevel represents the stability level of the component created by the factory.
+type StabilityLevel int
+
+const (
+	_ StabilityLevel = iota // skip 0, start types from 1.
+	StabilityLevelUnmaintained
+	StabilityLevelDeprecated
+	StabilityLevelInDevelopment
+	StabilityLevelAlpha
+	StabilityLevelBeta
+	StabilityLevelStable
+)
+
+func (sl StabilityLevel) String() string {
+	switch sl {
+	case StabilityLevelUnmaintained:
+		return "unmaintained"
+	case StabilityLevelDeprecated:
+		return "deprecated"
+	case StabilityLevelInDevelopment:
+		return "in development"
+	case StabilityLevelAlpha:
+		return "alpha"
+	case StabilityLevelBeta:
+		return "beta"
+	case StabilityLevelStable:
+		return "stable"
+	}
+	return "undefined"
+}
+
 // Factory is implemented by all component factories.
 //
 // This interface cannot be directly implemented. Implementations must
@@ -116,11 +147,16 @@ type Factory interface {
 }
 
 type baseFactory struct {
-	cfgType config.Type
+	cfgType   config.Type
+	stability StabilityLevel
 }
 
 func (baseFactory) unexportedFactoryFunc() {}
 
 func (bf baseFactory) Type() config.Type {
 	return bf.cfgType
+}
+
+func (bf baseFactory) StabilityLevel() StabilityLevel {
+	return bf.stability
 }
--- a/component/component.go
+++ b/component/component.go
@@ -147,7 +147,7 @@ type Factory interface {
 	Type() config.Type
 
 	// StabilityLevel gets the stability level of the component.
-	StabilityLevel(config.Type) StabilityLevel
+	StabilityLevel(config.DataType) StabilityLevel
 
 	unexportedFactoryFunc()
 }
@@ -163,8 +163,8 @@ func (bf baseFactory) Type() config.Type {
 	return bf.cfgType
 }
 
-func (bf baseFactory) StabilityLevel(cfgType config.Type) StabilityLevel {
-	if val, ok := bf.stability[cfgType]; ok {
+func (bf baseFactory) StabilityLevel(dt config.DataType) StabilityLevel {
+	if val, ok := bf.stability[dt]; ok {
 		return val
 	}
 	return StabilityLevelUndefined
--- a/component/component.go
+++ b/component/component.go
@@ -111,7 +111,7 @@ const (
 type StabilityLevel int
 
 const (
-	_ StabilityLevel = iota // skip 0, start types from 1.
+	StabilityLevelUndefined = iota // skip 0, start types from 1.
 	StabilityLevelUnmaintained
 	StabilityLevelDeprecated
 	StabilityLevelInDevelopment
@@ -146,12 +146,15 @@ type Factory interface {
 	// Type gets the type of the component created by this factory.
 	Type() config.Type
 
+	// StabilityLevel gets the stability level of the component.
+	StabilityLevel(config.Type) StabilityLevel
+
 	unexportedFactoryFunc()
 }
 
 type baseFactory struct {
 	cfgType   config.Type
-	stability StabilityLevel
+	stability map[config.Type]StabilityLevel
 }
 
 func (baseFactory) unexportedFactoryFunc() {}
@@ -160,6 +163,9 @@ func (bf baseFactory) Type() config.Type {
 	return bf.cfgType
 }
 
-func (bf baseFactory) StabilityLevel() StabilityLevel {
-	return bf.stability
+func (bf baseFactory) StabilityLevel(cfgType config.Type) StabilityLevel {
+	if val, ok := bf.stability[cfgType]; ok {
+		return val
+	}
+	return StabilityLevelUndefined
 }
--- a/component/componenttest/nop_processor.go
+++ b/component/componenttest/nop_processor.go
@@ -44,9 +44,9 @@ func NewNopProcessorFactory() component.ProcessorFactory {
 				ProcessorSettings: config.NewProcessorSettings(config.NewComponentID("nop")),
 			}
 		},
-		component.WithTracesProcessor(createTracesProcessor),
-		component.WithMetricsProcessor(createMetricsProcessor),
-		component.WithLogsProcessor(createLogsProcessor),
+		component.WithTracesProcessorAndStabilityLevel(createTracesProcessor, component.StabilityLevelInDevelopment),
+		component.WithMetricsProcessorAndStabilityLevel(createMetricsProcessor, component.StabilityLevelInDevelopment),
+		component.WithLogsProcessorAndStabilityLevel(createLogsProcessor, component.StabilityLevelInDevelopment),
 	)
 }
 
--- a/component/componenttest/nop_processor.go
+++ b/component/componenttest/nop_processor.go
@@ -44,9 +44,9 @@ func NewNopProcessorFactory() component.ProcessorFactory {
 				ProcessorSettings: config.NewProcessorSettings(config.NewComponentID("nop")),
 			}
 		},
-		component.WithTracesProcessorAndStabilityLevel(createTracesProcessor, component.StabilityLevelInDevelopment),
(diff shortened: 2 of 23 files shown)


component/component.go
```diff
@@ -104,6 +104,37 @@ const (
 	KindExtension
 )
 
+// StabilityLevel represents the stability level of the component created by the factory.
```
mx-psi, line 107: Should we link the README docs here?

component/exporter.go
```diff
@@ -157,6 +157,13 @@ func WithLogsExporter(createLogsExporter CreateLogsExporterFunc) ExporterFactory
 	}
 }
 
+// WithExporterStabiityLevel overrides the default "unmaintained" stability level.
+func WithExporterStabilityLevel(sl StabilityLevel) ExporterFactoryOption {
```
mx-psi, line 161: Stability level can potentially be defined independently for each signal type (so e.g. you can be beta for traces, stable for metrics and alpha for logs on a given exporter). How would one set this? (not sure if this is what you are alluding to in the README)

codeboten: @mx-psi right, this is what I meant with:

> the stability level is set per component factory, rather than per component. 

I'm open to suggestions on how to better solve this. I understand that in its current state, this would not work for the OTLP exporter for example, which is marked stable for traces/metrics but beta for logs.

mx-psi: Okay, I wasn't sure if this is what you were talking about. 

Here is an alternative I can think of:

1. Add a new 'unsupported' stability level, which would be the zero value.
2. Replace `WithExporterStabilityLevel` with a new `WithExporterMetadata` passing the metadata:

```go
// ExporterMetadata defines human-readable metadata about the exporter
type ExporterMetadata struct {
  // StabilityMetrics defines the stability level for the metrics exporter.
  StabilityMetrics StabilityLevel
  // StabilityTraces defines the stability level for the traces exporter. 
  StabilityTraces StabilityLevel
  // StabilityLogs defines the stability level for the logs exporter.
  StabilityLogs StabilityLevel

  // Add further metadata in the future if we come up with anything worth putting here (e.g. warnings?)
}

func WithExporterMetadata(metadata ExporterMetadata) ExporterFactoryOption { 
  // ...
}
```

3. When applying a `With[Signal]Exporter` options, if `Stability[Signal]` is 'unsupported', replace it with 'unmaintained' instead.
4. After having applied all options, check that the metadata is consistent with the available create functions (i.e. that we don't mark as 'unsupported' anything that is actually supported).

This is complicated because we have to account for (i) in what order you pass the options and (ii) the possibility of mismatch between what one says in `ExporterMetadata` and what one passes in other options, but the public API is... satisfying? to me.

codeboten: I've added a `StabilityLevelUndefined` level, which is the default for any component that doesn't set the level for a particular signal. I've opted to go w/ a map as a parameter, this means in the future we'll be able to add additional signals without having to modify the struct.

PTAL at the updated implementation, it addresses the concern i had around not being able to set different stability levels for different components of the same factory.

component/exporter.go
```diff
@@ -157,6 +157,13 @@ func WithLogsExporter(createLogsExporter CreateLogsExporterFunc) ExporterFactory
...
 
+// WithExporterStabiityLevel overrides the default "unmaintained" stability level.
+func WithExporterStabilityLevel(sl map[config.Type]StabilityLevel) ExporterFactoryOption {
+	return func(o *exporterFactory) {
+		o.stability = sl
+	}
+}
+
 // NewExporterFactory returns a ExporterFactory.
 func NewExporterFactory(cfgType config.Type, createDefaultConfig ExporterCreateDefaultConfigFunc, options ...ExporterFactoryOption) ExporterFactory {
```
mx-psi, line 168: Using a map makes sense to me, but someone can still do

```go
component.NewExporterFactory(
	typeStr,
	createDefaultConfig,
	component.WithTracesExporter(createTracesExporter),
	component.WithExporterStabilityLevel(map[config.Type]component.StabilityLevel{
		config.TracesDataType:  component.StabilityLevelStable,
		config.MetricsDataType: component.StabilityLevelStable, // WRONG: no support for metrics
}))
```

I think we should detect and error out on this case

mx-psi (commented):
LGTM except for a case where I think we should error out

component/component.go
```diff
@@ -112,15 +146,26 @@ type Factory interface {
 	// Type gets the type of the component created by this factory.
 	Type() config.Type
 
+	// StabilityLevel gets the stability level of the component.
+	StabilityLevel(config.Type) StabilityLevel
```
bogdandrutu, line 150: Is this "config.DataType"? I know is the same, but this is what I think makes sense.

component/exporter.go
```diff
@@ -157,6 +157,13 @@ func WithLogsExporter(createLogsExporter CreateLogsExporterFunc) ExporterFactory
 	}
 }
 
+// WithExporterStabiityLevel overrides the default "unmaintained" stability level.
+func WithExporterStabilityLevel(sl map[config.Type]StabilityLevel) ExporterFactoryOption {
```
bogdandrutu, line 161: Can this be just `WithStabilityLevel`? Another idea is to change `WithTracesExporter` to accept a `StabilityLevel` as well?

mx-psi: > Can this be just WithStabilityLevel?

The problem with this is that e.g. the receiver factory options also live in this package, so when we want to add this for receivers the name would clash.

bogdandrutu: We can do a trick, the `WithStabilityLevel` can return a type that is all option types Receiver/Exporter/Etc.

mx-psi: Also,

> Another idea is to change WithTracesExporter to accept a StabilityLevel as well?

sounds better to me, but I am not sure how we do this without breaking changes

service/internal/pipelines/pipelines.go
```diff
@@ -304,6 +304,23 @@ func Build(ctx context.Context, settings component.TelemetrySettings, buildInfo
...
+	switch sl {
+	case component.StabilityLevelDeprecated:
+		logger.Info("Component has been deprecated and will be removed in future releases.", zap.String(components.ZapStabilityKey, sl.String()))
+	case component.StabilityLevelUnmaintained:
+		logger.Info("Component is unmaintained and actively looking for contributors.", zap.String(components.ZapStabilityKey, sl.String()))
+	case component.StabilityLevelInDevelopment:
+		logger.Info("Component is under development.", zap.String(components.ZapStabilityKey, sl.String()))
+	case component.StabilityLevelAlpha:
+	case component.StabilityLevelBeta:
+	case component.StabilityLevelStable:
```
fatsheep9146, line 317: I think L315 to L318 should be changed to

```
case component.StabilityLevelAlpha, component.StabilityLevelBeta, component.StabilityLevelStable:
   logger.Debug("Stability level", zap.String(components.ZapStabilityKey, sl.String()))
```
or

```
	case component.StabilityLevelAlpha:
             fallthrough
	case component.StabilityLevelBeta:
             fallthrough
	case component.StabilityLevelStable:
             logger.Debug("Stability level", zap.String(components.ZapStabilityKey, sl.String()))
```

component/exporter.go
bogdandrutu: 2 step, `WithTracesExporterWithStability`?

codeboten: So it sounds like we can go either with:
1. adding the stability level to the traces exporter func
2. adding a WithStabilityLevel and keep WithTracesExporter (and all the other functions like it) as is

@mx-psi @bogdandrutu  I think it makes sense to have this be a parameter that all components will use, which would make option 2 make more sense.

mx-psi: The reason I like (1) more than (2) is that cases like https://github.com/open-telemetry/opentelemetry-collector/pull/5580#discussion_r907060692 are detected at compile time (missing/extra argument on a function) instead of at factory build time, but I don't feel strongly about it.

service/internal/pipelines/pipelines.go
```diff
@@ -304,6 +304,23 @@ func Build(ctx context.Context, settings component.TelemetrySettings, buildInfo
 	return exps, nil
 }
 
+func logStabilityMessage(logger *zap.Logger, sl component.StabilityLevel) {
+	switch sl {
+	case component.StabilityLevelDeprecated:
+		logger.Info("Component has been deprecated and will be removed in future releases.", zap.String(components.ZapStabilityKey, sl.String()))
+	case component.StabilityLevelUnmaintained:
+		logger.Info("Component is unmaintained and actively looking for contributors.", zap.String(components.ZapStabilityKey, sl.String()))
```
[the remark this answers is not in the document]
codeboten: Added, PTAL

component/exporter.go
```diff
@@ -157,6 +157,13 @@ func WithLogsExporter(createLogsExporter CreateLogsExporterFunc) ExporterFactory
 	}
 }
 
+// WithExporterStabiityLevel overrides the default "unmaintained" stability level.
```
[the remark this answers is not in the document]
codeboten: Fixed the comment to "undefined"

service/internal/pipelines/pipelines.go
codeboten: Fixed

component/exporter.go
codeboten: @mx-psi @bogdandrutu updated the PR with the suggestion of going w/ `WithTracesExporterAndStabilityLevel`. I think this is the right way to move forward. If folks agree, I will continue w/ the TODO items in the description

service/internal/pipelines/pipelines.go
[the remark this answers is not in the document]
TylerHelmuth: LGTM

component/component.go
codeboten: Yup 👍🏻

component/exporter.go
mx-psi: I agree that this is the right way to move forward 👍

codeboten: Done

component/componenttest/nop_processor.go
```diff
@@ -44,9 +44,9 @@ func NewNopProcessorFactory() component.ProcessorFactory {
 				ProcessorSettings: config.NewProcessorSettings(config.NewComponentID("nop")),
 			}
 		},
-		component.WithTracesProcessor(createTracesProcessor),
-		component.WithMetricsProcessor(createMetricsProcessor),
-		component.WithLogsProcessor(createLogsProcessor),
+		component.WithTracesProcessorAndStabilityLevel(createTracesProcessor, component.StabilityLevelInDevelopment),
```
bogdandrutu, line 47: :)) should the nop be stable?

component/exporter.go
```diff
@@ -149,30 +149,57 @@ type exporterFactory struct {
 }
 
 // WithTracesExporter overrides the default "error not supported" implementation for CreateTracesExporter.
+// Deprecated: [v0.55.0] Use WithTracesExporterAndStabilityLevel instead.
```
bogdandrutu, line 152: Make this call into `WithTracesExporterAndStabilityLevel` with `Undefined` level?

service/internal/pipelines/pipelines.go
```diff
@@ -304,6 +304,21 @@ func Build(ctx context.Context, settings component.TelemetrySettings, buildInfo
 	return exps, nil
 }
 
+func logStabilityMessage(logger *zap.Logger, sl component.StabilityLevel) {
```
bogdandrutu, line 307: Should we have a "InfoMessage" (or find a better name) on `component.StabilityLevel` (can even accept zap.Log or return the string) to help with this? 

Reason: you forget to do the same in `extensions` package, where you will need to duplicate this.

bogdandrutu (approved):
Overall looks good to me. I would split the logging / zpages part.

service/internal/pipelines/pipelines.go
codeboten: Sounds good, I thought about it and ended up choosing to go this route, but you're right we'll need the same thing for extensions.

I'll follow up this PR with a fix for this that can be used in extensions

PR #5602, opened 2022-06-27
Update flags API

### LINKED ISSUE #5443: [Proposal]: Enhance [*]Flags to not need [*]Flag, and to simplify usage
**Current API(s):**

Right now users can check for a specific flag by using `HasFlag` API, which hides some complicated logic of using bit operations, but requires multiple structs/types.

```golang
// MetricDataPointFlags defines how a metric aggregator reports aggregated values.
// It describes how those values relate to the time interval over which they are aggregated.
type MetricDataPointFlags uint32

const (
	// MetricDataPointFlagsNone is the default MetricDataPointFlags.
	MetricDataPointFlagsNone = MetricDataPointFlags(otlpmetrics.DataPointFlags_FLAG_NONE)
)

// NewMetricDataPointFlags returns a new MetricDataPointFlags combining the flags passed
// in as parameters.
func NewMetricDataPointFlags(flags ...MetricDataPointFlag) MetricDataPointFlags {
	var flag MetricDataPointFlags
	for _, f := range flags {
		flag |= MetricDataPointFlags(f)
	}
	return flag
}

// HasFlag returns true if the MetricDataPointFlags contains the specified flag
func (d MetricDataPointFlags) HasFlag(flag MetricDataPointFlag) bool {
	return d&MetricDataPointFlags(flag) != 0
}

// String returns the string representation of the MetricDataPointFlags.
func (d MetricDataPointFlags) String() string {
	return otlpmetrics.DataPointFlags(d).String()
}

// MetricDataPointFlag allow users to configure DataPointFlags. This is achieved via NewMetricDataPointFlags.
// The separation between MetricDataPointFlags and MetricDataPointFlag exists to prevent users accidentally
// comparing the value of individual flags with MetricDataPointFlags. Instead, users must use the HasFlag method.
type MetricDataPointFlag uint32

const (
	// MetricDataPointFlagNoRecordedValue is flag for a metric aggregator which reports changes since last report time.
	MetricDataPointFlagNoRecordedValue = MetricDataPointFlag(otlpmetrics.DataPointFlags_FLAG_NO_RECORDED_VALUE)
)
```

**Proposal API:**

Essentially hide the bit operation from the users and provide helpers to access different flags from the bitset

```golang
// MetricDataPointFlags defines how a metric aggregator reports aggregated values.
// It describes how those values relate to the time interval over which they are aggregated.
type MetricDataPointFlags uint32

// NewMetricDataPointFlags returns a new MetricDataPointFlags combining the flags passed
// in as parameters.
func NewMetricDataPointFlags() MetricDataPointFlags {
	return MetricDataPointFlags(otlpmetrics.DataPointFlags_FLAG_NONE)
}

// NoRecordedValue returns true if the MetricDataPointFlags contains the specified flag
func (d MetricDataPointFlags) NoRecordedValue() bool {
	return d&MetricDataPointFlags(otlpmetrics.DataPointFlags_FLAG_NO_RECORDED_VALUE) != 0
}

// HasFlag returns true if the MetricDataPointFlags contains the specified flag
func (d MetricDataPointFlags) SetNoRecordedValue(val bool) {
	....
}


// String returns the string representation of the MetricDataPointFlags.
func (d MetricDataPointFlags) String() string {
	return otlpmetrics.DataPointFlags(d).String()
}
```

TylerHelmuth:
**Description:** 
This PR updates the Flags API to hide all references to the Flags themselves.  Instead, setting flags and checking if a flag is set is handled via functions that know about the flag being set/checked.

**Link to tracking Issue:**
Fixes #5443

**Testing:**
Updated unit tests


TylerHelmuth:
@bogdandrutu @dmitryax is this what you had in mind with #5443?  This implementation is currently a breaking change to show final state, but I think we'd need to handle deprecating HasFlag and the original constructor.

pdata/internal/metrics.go
```diff
@@ -192,19 +192,18 @@ const (
...
+
+// NoRecordedValue returns true if the MetricDataPointFlags contains the NO_RECORDED_VALUE flag.
+func (d MetricDataPointFlags) NoRecordedValue() bool {
+	return d&MetricDataPointFlags(otlpmetrics.DataPointFlags_FLAG_NO_RECORDED_VALUE) != 0
 }
 
-// HasFlag returns true if the MetricDataPointFlags contains the specified flag
-func (d MetricDataPointFlags) HasFlag(flag MetricDataPointFlag) bool {
-	return d&MetricDataPointFlags(flag) != 0
+func (d MetricDataPointFlags) SetNoRecordedValue() MetricDataPointFlags {
```
bogdandrutu, line 205: Should accept bool to allow set/unset?

bogdandrutu (commented):
I like it.

pdata/internal/metrics.go
TylerHelmuth: Do we use &^ for this?

TylerHelmuth: And if that's the case, I'd prefer a Set function and a Remove function. Passing False or True to a Set function makes it seem like it's a bool being set. I'd rather have 2 different functions that describe what they do.

bogdandrutu: > Do we use &^ for this?

That is an implementation detail

> I thought once a flag was bitwise ORed it couldn't be undone.

That may be the case for some flags, but definitely not for all.

> And if that's the case, I'd prefer a Set function and a Remove function. Passing False or True to a Set function makes it seem like it is a bool being set. I'd rather have 2 different functions that describe what they do.

For me the reason to have a "Set" with bool, is because matches the getter. In the end the flag represents a "bool" correct? which can be true of false.

@mx-psi @codeboten @tigrannajaryan any comments on this? What is your preference?

mx-psi: I prefer `Set` with a bool, for consistency with other `Set` functions on pdata.

TylerHelmuth: Ok, I updated the function to set/remove based on a bool parameter.

TylerHelmuth:
What should the strategy be to make this backwards compatible?  The `NewMetricDataPointFlags` func change is probably the most breaking of the changes.

pdata/internal/metrics.go
```diff
@@ -192,19 +192,21 @@ const (
...
-// NewMetricDataPointFlags returns a new MetricDataPointFlags combining the flags passed
-// in as parameters.
-func NewMetricDataPointFlags(flags ...MetricDataPointFlag) MetricDataPointFlags {
-	var flag MetricDataPointFlags
-	for _, f := range flags {
-		flag |= MetricDataPointFlags(f)
-	}
-	return flag
+// NewMetricDataPointFlags returns a new MetricDataPointFlags with no flags set.
+func NewMetricDataPointFlags() MetricDataPointFlags {
```
bogdandrutu, line 196: You need one more change, since this type is no longer immutable, we need to remove the setter from the parent struct

bogdandrutu: The reason is because that is the pattern we use

bogdandrutu:
> What should the strategy be to make this backwards compatible? The NewMetricDataPointFlags func change is probably the most breaking of the changes.

If too hard we can do an exception, otherwise a 2 step process is preferred, add a `NewEmptyMetricDataPointFlags` deprecate the current one, then remove the current add the new signature and deprecate the empty.

pdata/internal/metrics.go
TylerHelmuth: I am not sure what you mean by "parent struct"

TylerHelmuth:
> If too hard we can do an exception, otherwise a 2 step process is preferred, add a `NewEmptyMetricDataPointFlags` deprecate the current one, then remove the current add the new signature and deprecate the empty.

Done

pdata/internal/metrics_test.go
```diff
@@ -691,15 +691,20 @@ func TestMetricsClone(t *testing.T) {
 func TestMetricsDataPointFlags(t *testing.T) {
 	gauge := generateTestGauge()
 
-	gauge.DataPoints().At(0).SetFlags(NewMetricDataPointFlags())
+	gauge.DataPoints().At(0).SetFlags(NewEmptyMetricDataPointFlags())
```
bogdandrutu, line 694: No need to set this correct?

pdata/internal/metrics_test.go
```diff
@@ -691,15 +691,20 @@ func TestMetricsClone(t *testing.T) {
...
 	assert.True(t, gauge.DataPoints().At(0).Flags() == MetricDataPointFlagsNone)
-	assert.False(t, gauge.DataPoints().At(0).Flags().HasFlag(MetricDataPointFlagNoRecordedValue))
+	assert.False(t, gauge.DataPoints().At(0).Flags().NoRecordedValue())
 	assert.Equal(t, "FLAG_NONE", gauge.DataPoints().At(0).Flags().String())
 
-	gauge.DataPoints().At(0).SetFlags(NewMetricDataPointFlags(MetricDataPointFlagNoRecordedValue))
+	gauge.DataPoints().At(0).SetFlags(NewEmptyMetricDataPointFlags().SetNoRecordedValue(true))
```
bogdandrutu, line 699: gauge.DataPoints().At(0).Flags().SetNoRecordedValue(true)

pdata/internal/metrics.go
bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/blob/c368d497a2046011da4a39ee977ceed5f841c266/pdata/internal/generated_pmetric.go#L1209

pdata/internal/metrics_test.go
TylerHelmuth: Technically it works with the original function, but Goland got all upset bc it was deprecated.  If you want to keep it so that the deprecation/transition process has less modifications thats find with me.

pdata/internal/metrics_test.go
TylerHelmuth: If you're meaning 

```Go
gauge.DataPoints().At(0).SetFlags(gauge.DataPoints().At(0).Flags().SetNoRecordedValue(true))
```

That would work.  But

```Go
gauge.DataPoints().At(0).Flags().SetNoRecordedValue(true)
```

wouldn't actually change the flag returned by Flags() since .SetNoRecordedValue(true) currently returns a new MetricDataPointFlags

bogdandrutu: I think we should make `SetNoRecordedValue` to mutate and not have the `SetFlags`, similar with other "mutable" structs.

TylerHelmuth: We might need to keep `SetFlags` so that `CopyTo` works as expected.  Or we need a new `CopyTo` func on `MetricDataPointFlags` so that we can do `ms.Flags().CopyTo(dest.Flags())`

TylerHelmuth: Also, how should `SetNoRecordedValue` mutate?  Should we change `.Flags()` to return a pointer (I think it would be the only pdata field to do this) or should we actually make a new `SetFlagNoRecordedValue(bool)` function to replace `SetFlags`.  Then we could do 

```Go
gauge.DataPoints().At(0).SetFlagNoRecordedValue(true)
assert.False(t, gauge.DataPoints().At(0).Flags() == MetricDataPointFlagsNone)
assert.True(t, gauge.DataPoints().At(0).Flags().NoRecordedValue())
```

bogdandrutu: See https://github.com/open-telemetry/opentelemetry-collector/pull/5609

bogdandrutu: That PR, is only to show how I think it makes more sense, and is consistent with all other "mutable" values.

TylerHelmuth: @bogdandrutu I updated MetricDataPointFlags to be a struct like your example PR.  I think the result is pretty clean and friendly to work with.  In addition to `NoRecordedValue` and `SetNoRecordedValue` I added a `None` and `Reset` so that users can interacted with the default flag.   `None` replaces the need for `gauge.DataPoints().At(0).Flags() == MetricDataPointFlagsNone`.

pdata/internal/metrics.go
```diff
@@ -185,42 +185,69 @@ func (at MetricAggregationTemporality) String() string {
...
 }
 
-// MetricDataPointFlag allow users to configure DataPointFlags. This is achieved via NewMetricDataPointFlags.
-// The separation between MetricDataPointFlags and MetricDataPointFlag exists to prevent users accidentally
-// comparing the value of individual flags with MetricDataPointFlags. Instead, users must use the HasFlag method.
-type MetricDataPointFlag uint32
+// Reset resets the flags to be FLAG_NONE.
+func (ms MetricDataPointFlags) Reset() {
+	*ms.orig = uint32(otlpmetrics.DataPointFlags_FLAG_NONE)
+}
```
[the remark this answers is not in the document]
TylerHelmuth: My thought for adding them was to try to keep some backwards compatible capabilities and to future-proof the API.

The API today exposes `MetricDataPointFlagsNone` which can be used to compared against `Flags()` to see if no flags are set.  Without a `None()` function we'll only be able to check whether or not `NoRecordedValue` is set or not.  I couldn't find any use of that exposed field in Contrib so maybe it doesn't matter but it is technically possible today.

Since that is the only flag right now a return of false from `NoRecordedValue` means that no flags are set, but that wouldn't necessarily be true if more flags are added in the future. It feels like being able to interact with the default state of the enum via dedicated functions future proofs the API.

TylerHelmuth:
Not stale, waiting for @bogdandrutu to return from PTO.

pdata/internal/metrics.go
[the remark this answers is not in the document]
TylerHelmuth: Although I do think `None()` is good ergonomics for checking if there are no flags set, I do recognize that is not being done anywhere in Core or Contrib.  If a third flag is ever added to the protos and pdata is updated to handle that flag then we should add `None()` and `Rest()`.

TylerHelmuth:
@bogdandrutu ready for another review.

pdata/pmetric/alias.go
```diff
@@ -54,25 +54,9 @@ const (
...
-// MetricDataPointFlag allow users to configure DataPointFlags. This is achieved via NewMetricDataPointFlags.
-// The separation between MetricDataPointFlags and MetricDataPointFlag exists to prevent users accidentally
-// comparing the value of individual flags with MetricDataPointFlags. Instead, users must use the HasFlag method.
-type MetricDataPointFlag = internal.MetricDataPointFlag
-
-const (
-	// MetricDataPointFlagNoRecordedValue is flag for a metric aggregator which reports changes since last report time.
-	MetricDataPointFlagNoRecordedValue = internal.MetricDataPointFlagNoRecordedValue
-)
-
```
bogdandrutu, line 75: We need to go via deprecation, cannot directly delete the code.

bogdandrutu (commented):
I think this is the final state that I envisioned, the problem is that we need to do this in multiple steps and deprecate things first.

pdata/pmetric/alias.go
TylerHelmuth: Agreed.  I will get started on that now that the end state is finalized.

pdata/internal/cmd/pdatagen/internal/metrics_structs.go
```diff
@@ -489,6 +493,15 @@ var exemplar = &messageValueStruct{
 	},
 }
 
+var dataPointFlagsFieldStruct = &messageValueField{
+	fieldName:       "FlagsStruct",
```
TylerHelmuth, line 497: Open to suggestions on this name.

pdata/pmetric/alias.go
TylerHelmuth: Re-added the original solution and added a new field to pdata called `FlagsStruct` that exposes the new capabilities.

pdata/internal/metrics.go
```diff
@@ -183,6 +183,62 @@ func (at MetricAggregationTemporality) String() string {
...
 
+// EmptyMetricDataPointFlags defines how a metric aggregator reports aggregated values.
+// It describes how those values relate to the time interval over which they are aggregated.
+//
+// This is a reference type, if passed by value and callee modifies it the
+// caller will see the modification.
+//
+// Must use NewEmptyMetricDataPointFlags function to create new instances.
+// Important: zero-initialized instance is not valid for use.
+type EmptyMetricDataPointFlags struct {
```
bogdandrutu, line 194: s/EmptyMetricDataPointFlags/MetricDataPointFlagsStruct?

If I understand, this is a temporary name because of the name conflict, correct?

pdata/internal/cmd/pdatagen/internal/metrics_structs.go
bogdandrutu: Sounds reasonable.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
 - Remove deprecated `config.ServiceTelemetry` (#5565)
 - Remove deprecated `config.ServiceTelemetryLogs` (#5565)
 - Remove deprecated `config.ServiceTelemetryMetrics` (#5565)
+- Removed `HasFlag`, `MetricDataPointFlagsNone`, and `MetricDataPointFlagNoRecordedValue`. Use new Flags API to check for Flags and to set and remove Flags (#5602)
```
bogdandrutu, line 40: These are deprecated not deleted, correct?

Also, please move them to unreleased section.

pdata/internal/metrics.go
TylerHelmuth: Ya temporary. In a future PR we'd switch everything back to the names being used today.

CHANGELOG.md
TylerHelmuth: Good catch, need to update this again.

pdata/internal/cmd/pdatagen/internal/metrics_structs.go
TylerHelmuth: The existing `Flags()` api doesn't actually show as deprecated when used bc it auto-generated code. Is that ok or should we flag it and point people to `FlagsStruct`?

CHANGELOG.md
TylerHelmuth: Done.

pdata/internal/metrics.go
bogdandrutu: Please use "MetricDataPointFlagsStruct" name.

TylerHelmuth: Done

pdata/internal/cmd/pdatagen/internal/metrics_structs.go
bogdandrutu: I think we should mark that as deprecated, otherwise cannot expect people to change. We can do that in a separate PR.

bogdandrutu (approved):

pdata/internal/cmd/pdatagen/internal/metrics_structs.go
TylerHelmuth: Ok.

Issue #5564, opened 2022-06-21
Restructure packages from current "service" package


bogdandrutu:
This is a proposal to restructure current large "service" package with multiple dependencies, and too many features to match the configuration "service" part:

- [x] Will create a package under "service" directory for every config service subcomponent (telemetry, pipelines, extensions). These packages will follow the format:
    * `Config` struct for configuration, similar to all other components.
    * Will not use a "Factory" since I don't think we want necessary to allow this to be pluginable, but will offer an API like `New(context.Context, Settings)` or `Build(context.Context, Settings)`.
- [x] Telemetry initialization logic will be moved into this package `service/telemetry.go` file.
- [x] The current `service` package will become the package that implements the current service part of the config. Which means also will implement the "Host" interface. This package will have a `Config` which will represent the current "service::" configuration.
- [x] Create a new "collector" package and move the current "Collector", "windows.SvcHandler", "Command", "ConfigProvider" structs and related constructors/helpers.
    * Things related to `flags` will also be moved there.
- [x] `service/featuregate` will be moved  as a top directory `featuregate`.

Advantages:
1. Better match of the features/structure with the configuration structure.
2. Allow external parties to use a "sub-component" of the collector (telemetry, extensions, etc.). This will fix https://github.com/open-telemetry/opentelemetry-collector/issues/4970

PR #5606, opened 2022-06-29
Move extensions for internal to public package

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-collector/issues/5564

### DIFF
--- a/service/extensions/config.go
+++ b/service/extensions/config.go
@@ -1,0 +1,22 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package extensions // import "go.opentelemetry.io/collector/service/extensions"
+
+import (
+	"go.opentelemetry.io/collector/config"
+)
+
+// Config represents the ordered list of extensions configured for the service.
+type Config []config.ComponentID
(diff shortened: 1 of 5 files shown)


service/extensions/extensions.go
```diff
@@ -128,18 +128,15 @@ type Settings struct {
...
 	Factories map[config.Type]component.ExtensionFactory
-
-	// ServiceExtensions are the ordered list of extensions configured for the service.
-	ServiceExtensions []config.ComponentID
 }
 
-// Build builds Extensions from config.
-func Build(ctx context.Context, set Settings) (*Extensions, error) {
+// New creates a new Extensions from Config.
+func New(ctx context.Context, set Settings, cfg Config) (*Extensions, error) {
```
TylerHelmuth, line 134: I like the idea of using `Config` like other components do, but what gets put in `Config` vs `Settings`?

TylerHelmuth: That distinction makes sense, thanks for clarifying.

service/extensions/config.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package extensions // import "go.opentelemetry.io/collector/service/extensions"
+
+import (
+	"go.opentelemetry.io/collector/config"
+)
+
+// Config represents the ordered list of extensions configured for the service.
+type Config []config.ComponentID
```
mx-psi, line 22: Should we have a struct in case we ever want to add more fields?

```suggestion
// Config defines the Extensions configuration
type Config struct {
  // Extensions configured for the service
  Extensions []config.ComponentID
}
```

Also, does the extension order really matter? We do start them in that order, but maybe we should not mention that in the docs, so that we are free to change that order (e.g. start extensions in parallel) if we ever want to.

mx-psi (commented):
Just one comment, also needs changelog entry

service/extensions/config.go
tigrannajaryan: >Also, does the extension order really matter? We do start them in that order, but maybe we should not mention that in the docs, so that we are free to change that order (e.g. start extensions in parallel) if we ever want to.

Extensions may depend on other extensions (you can use GetExtensions() to access another extension), so the start order can be important (I don't we have an actual use case where it matters, but we may have in the future).

mx-psi: That's a good point, we should keep the order mention then.

bogdandrutu: Thanks for the discussion, seems to be resolved.

mx-psi (approved):

PR #5626, opened 2022-07-04
Restore "Shutdown complete" log

### LINKED ISSUE #5625: Restore the "Shutdown complete" log from collector

The collector v0.52.0 and before used to write an INFO log on shutdown that said "Shutdown complete".
This was useful e.g. in testing, to automatically verify that the collector has shut down completely.

```console
2022-07-04T12:05:28.513+0200	info	service/collector.go:278	Shutdown complete.
```

In v0.53.0 and later, this log has been [removed](https://github.com/open-telemetry/opentelemetry-collector/commit/1aa0e5ab3667721f282b557346779cb4ebec4903#diff-780b10647e2c9aaf7a1a80898c979f5018106a37a92b0d2ce4728017b28be2f6L278).

Now I don't seem to have a clear way of verifying that the collector has shut down.

I propose to restore this single log. Hopefully this is a simple and painless change. 🤞

astencel-sumo:
Fixes #5625 

**Testing:** 

I have run:

```console
make install-tools
make
make otelcorecol
```

After the otelcorecol binary was built, I ran it with a simple config:

```yaml
exporters:
  logging:

receivers:
  otlp:
    protocols:
      http:

service:
  pipelines:
    traces:
      receivers:
      - otlp
      exporters:
      - logging
```

and got the following log output, including the "Shutdown complete" log:

```console
2022-07-04T12:58:21.506+0200    info    service/telemetry.go:103        Setting up own telemetry...
2022-07-04T12:58:21.507+0200    info    service/telemetry.go:138        Serving Prometheus metrics      {"address": ":8888", "level": "basic"}
2022-07-04T12:58:21.507+0200    info    extensions/extensions.go:42     Starting extensions...
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:74       Starting exporters...
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:78       Exporter is starting... {"kind": "exporter", "data_type": "traces", "name": "logging"}
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:82       Exporter started.       {"kind": "exporter", "data_type": "traces", "name": "logging"}
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:86       Starting processors...
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:98       Starting receivers...
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:102      Receiver is starting... {"kind": "receiver", "name": "otlp", "pipeline": "traces"}
2022-07-04T12:58:21.507+0200    info    otlpreceiver/otlp.go:88 Starting HTTP server on endpoint 0.0.0.0:4318   {"kind": "receiver", "name": "otlp", "pipeline": "traces"}
2022-07-04T12:58:21.507+0200    info    pipelines/pipelines.go:106      Receiver started.       {"kind": "receiver", "name": "otlp", "pipeline": "traces"}
2022-07-04T12:58:21.507+0200    info    service/collector.go:215        Starting otelcorecol... {"Version": "0.54.0-dev", "NumCPU": 16}
2022-07-04T12:58:21.507+0200    info    service/collector.go:128        Everything is ready. Begin running and processing data.

^C2022-07-04T12:58:24.650+0200  info    service/collector.go:159        Received signal from OS {"signal": "interrupt"}
2022-07-04T12:58:24.650+0200    info    service/collector.go:231        Starting shutdown...
2022-07-04T12:58:24.650+0200    info    pipelines/pipelines.go:118      Stopping receivers...
2022-07-04T12:58:24.651+0200    info    pipelines/pipelines.go:125      Stopping processors...
2022-07-04T12:58:24.651+0200    info    pipelines/pipelines.go:132      Stopping exporters...
2022-07-04T12:58:24.651+0200    info    extensions/extensions.go:56     Stopping extensions...
2022-07-04T12:58:24.651+0200    info    service/collector.go:246        Shutdown complete.
```

### DIFF
--- a/service/collector.go
+++ b/service/collector.go
@@ -243,6 +243,7 @@ func (col *Collector) shutdown(ctx context.Context) error {
 		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown collector telemetry: %w", err))
 	}
 
+	col.service.telemetrySettings.Logger.Info("Shutdown complete.")
 	col.setCollectorState(Closed)
 
 	return errs


tigrannajaryan (approved):

astencel-sumo:
Should this PR be marked with the `Skip-Changelog` label?
Or is it worth adding this change to the Changelog?

service/collector.go
```diff
@@ -243,6 +243,7 @@ func (col *Collector) shutdown(ctx context.Context) error {
 		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown collector telemetry: %w", err))
 	}
 
+	col.service.telemetrySettings.Logger.Info("Shutdown complete.")
```
bogdandrutu, line 246: This happens after the logger is closed.

bogdandrutu (changes requested):
Cannot do this unfortunately

service/collector.go
tigrannajaryan: The log output in the PR description shows the "Shutdown complete.". How does this work if the logger is closed?

bogdandrutu: It may be that the logger is not closed is just flushed, so the output is not deterministic?

jedr: Oh I think I know why it works ;)
https://github.com/open-telemetry/opentelemetry-collector/blob/v0.55.0/service/service.go#L136

Perhaps I should first propose fixing that TODO and then put a "Logging finished" or similar message right before `Logger.Sync()` call.

astencel-sumo: Oh I think I know why it works ;)
https://github.com/open-telemetry/opentelemetry-collector/blob/v0.55.0/service/service.go#L136

Perhaps I should first propose fixing that TODO and then put a "Logging finished" or similar message right before `Logger.Sync()` call.

tigrannajaryan: What is the reason we need to shutdown the Logger at all? Why shouldn't we keep it alive until the process is exited?

tigrannajaryan: Make sense. So, I think it is best to output "Service is shutdown" log message, followed by `Logger.Sync()` at the end of `service.Shutdown()`.

bogdandrutu: >What is the reason we need to shutdown the Logger at all? Why shouldn't we keep it alive until the process is exited? 

This is not ok, because user can change the logger config, so we need to "close" it every time a config has changed (a.k.a service shutdown) not until the process ends.

tigrannajaryan: >This is not ok, because user can change the logger config, so we need to "close" it every time a config has changed (a.k.a service shutdown) not until the process ends.

I sill don't understand why we need to "close" it. I assume you are thinking about the service restart scenario, right? When restarting a new logger is created with potentially new config. Why is it a problem to keep the old logger around? It will be garbage collected when the new logger is created and assigned to the service's logger field. At least that's what I expect, maybe I am missing something.

bogdandrutu: On the same note: "I also don't understand why is so important to have this log message?"

We definitely need to flush it to be deterministic, otherwise if after the restart we "end" too soon the previous log may miss things and may not be flushed. So why do we care to have this log and this debate? I think we should make "service" (instance) own the logger and its lifetime.

bogdandrutu:
Replaced by https://github.com/open-telemetry/opentelemetry-collector/pull/5913

PR #5651, opened 2022-07-08
Update the builder to use a consistent binary name.

### LINKED ISSUE #5581: Builder tool name is inconsistent
**Describe the bug**
The builder tool is alternately named `ocb`, `builder` or `opentelemetry-collector-builder`, depending on where you look.

**Steps to reproduce**
Download a release binary, e.g. [ocb_0.54.0_darwin_amd64](https://github.com/open-telemetry/opentelemetry-collector/releases/download/v0.54.0/ocb_0.54.0_darwin_amd64)

```
$ ./ocb_0.54.0_darwin_arm64 --help
OpenTelemetry Collector distribution builder (0.54.0)

Usage:
  builder [flags]
  builder [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  version     Version of opentelemetry-collector-builder

Flags:
      --config string            config file (default is $HOME/.otelcol-builder.yaml)
      --description string       A descriptive name for the OpenTelemetry Collector distribution (default "Custom OpenTelemetry Collector distribution")
      --go string                The Go binary to use during the compilation phase. Default: go from the PATH
  -h, --help                     help for builder
      --module string            The Go module for the new distribution (default "go.opentelemetry.io/collector/cmd/builder")
      --name string              The executable name for the OpenTelemetry Collector distribution (default "otelcol-custom")
      --otelcol-version string   Which version of OpenTelemetry Collector to use as base (default "0.54.0")
      --output-path string       Where to write the resulting files (default "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution2878501842")
      --skip-compilation         Whether builder should only generate go code with no compile of the collector (default false)
      --version string           The version for the OpenTelemetry Collector distribution (default "1.0.0")

Use "builder [command] --help" for more information about a command.
```

A consistent name for the builder tool.

**What did you see instead?**
The release process names the tool `ocb`, the subcommand name is `builder` and the version help calls it `opentelemetry-collector-builder`. And I guess the config file implies the tool would be `otelcol-builder`.

**What version did you use?**
Version: 0.54.0

**What config did you use?**
Config: N/A

**Environment**
OS: Darwin
Compiler(if manually compiled): N/A

jpeach:
**Description:** 

Update the builder to consistently use "ocb" in the name of the command,
version and configuration file.

**Link to tracking Issue:**

This fixes #5581.

**Testing:** 

Manually looked at version output.

**Documentation:** 

None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,6 @@ ALL_MODULES := $(shell find . -type f -name "go.mod" -exec dirname {} \; | sort 
 CMD?=
 TOOLS_MOD_DIR := ./internal/tools
 
-GOOS=$(shell $(GOCMD) env GOOS)
-GOARCH=$(shell $(GOCMD) env GOARCH)
-GH := $(shell which gh)
-
 # TODO: Find a way to configure this in the generated code, currently no effect.
 BUILD_INFO_IMPORT_PATH=go.opentelemetry.io/collector/internal/version
 VERSION=$(shell git describe --always --match "v[0-9]*" HEAD)
@@ -206,6 +202,10 @@ otelcorecol:
 .PHONY: genotelcorecol
 genotelcorecol:
 	pushd cmd/builder/ && $(GOCMD) run ./ --skip-compilation --config ../otelcorecol/builder-config.yaml --output-path ../otelcorecol && popd
+
+.PHONY: ocb
+ocb:
+	$(MAKE) -C cmd/builder ocb
 
 DEPENDABOT_PATH=".github/dependabot.yml"
 .PHONY: internal-gendependabot
@@ -411,5 +411,8 @@ endif
 		echo "'gh' command not found, can't submit the PR on your behalf."; \
 		exit 1; \
 	fi
-	gh pr create --title "[chore] prepare release $(RELEASE_CANDIDATE)"
+	$(GH) pr create --title "[chore] prepare release $(RELEASE_CANDIDATE)"
 
+.PHONY: clean
+clean:
+	test -d bin && $(RM) bin/*
--- a/cmd/builder/internal/command.go
+++ b/cmd/builder/internal/command.go
@@ -38,8 +38,13 @@ var (
 // Command is the main entrypoint for this application
 func Command() (*cobra.Command, error) {
 	cmd := &cobra.Command{
-		Use:  "builder",
-		Long: fmt.Sprintf("OpenTelemetry Collector distribution builder (%s)", version),
+		Use: "ocb",
+		Long: fmt.Sprintf("OpenTelemetry Collector Builder (%s)", version) + `
+
+ocb generates a custom OpenTelemetry Collector binary using the
+build configuration given by the "--config" argument.
+`,
+		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := initConfig(); err != nil {
 				return err
@@ -58,8 +63,14 @@ func Command() (*cobra.Command, error) {
 		},
 	}
 
-	// the external config file
-	cmd.Flags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.otelcol-builder.yaml)")
+	cmd.Flags().StringVar(&cfgFile, "config", "", "build configuration file")
+
+	// A build configuration file is always required, and there's no
+	// default. We can relax this in future by embedding the default
+	// config that is used to build otelcorecol.
+	if err := cmd.MarkFlagRequired("config"); err != nil {
+		panic(err) // Only fails if the usage message is empty, which is a programmer error.
+	}
 
 	// the distribution parameters, which we accept as CLI flags as well
 	cmd.Flags().BoolVar(&cfg.SkipCompilation, "skip-compilation", false, "Whether builder should only generate go code with no compile of the collector (default false)")
@@ -82,11 +93,7 @@ func Command() (*cobra.Command, error) {
 }
 
 func initConfig() error {
-	cfg.Logger.Info("OpenTelemetry Collector distribution builder", zap.String("version", version), zap.String("date", date))
-	// use the default path if there is no config file being specified
-	if cfgFile == "" {
-		cfgFile = "$HOME/.otelcol-builder.yaml"
-	}
+	cfg.Logger.Info("OpenTelemetry Collector Builder", zap.String("version", version), zap.String("date", date))
 
 	// load the config file
 	if err := k.Load(file.Provider(cfgFile), yaml.Parser()); err != nil {
--- a/Makefile.Common
+++ b/Makefile.Common
@@ -10,6 +10,10 @@ GO_ACC=go-acc
 LINT=golangci-lint
 IMPI=impi
 
+GOOS := $(shell $(GOCMD) env GOOS)
+GOARCH := $(shell $(GOCMD) env GOARCH)
+GH := $(shell which gh)
+
 .PHONY: test
 test:
 	$(GOTEST) $(GOTEST_OPT) ./...
--- a/cmd/builder/Makefile
+++ b/cmd/builder/Makefile
@@ -1,1 +1,5 @@
 include ../../Makefile.Common
+
+.PHONY: ocb
+ocb:
+	GO111MODULE=on CGO_ENABLED=0 $(GOCMD) build -trimpath -o ../../bin/ocb_$(GOOS)_$(GOARCH) .
--- a/cmd/builder/README.md
+++ b/cmd/builder/README.md
@@ -1,4 +1,4 @@
-# OpenTelemetry Collector builder
+# OpenTelemetry Collector Builder (ocb)
 [![CI](https://github.com/open-telemetry/opentelemetry-collector-builder/actions/workflows/go.yaml/badge.svg)](https://github.com/open-telemetry/opentelemetry-collector-builder/actions/workflows/go.yaml?query=branch%3Amain)
 
 This program generates a custom OpenTelemetry Collector binary based on a given configuration.
@@ -7,7 +7,7 @@ This program generates a custom OpenTelemetry Collector binary based on a given 
 
 ```console
 $ GO111MODULE=on go install go.opentelemetry.io/collector/cmd/builder@latest
-$ cat > ~/.otelcol-builder.yaml <<EOF
+$ cat > otelcol-builder.yaml <<EOF
 exporters:
   - gomod: "github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.40.0"
   - import: go.opentelemetry.io/collector/exporter/loggingexporter
@@ -21,7 +21,7 @@ processors:
   - import: go.opentelemetry.io/collector/processor/batchprocessor
     gomod: go.opentelemetry.io/collector v0.40.0
 EOF
-$ builder --output-path=/tmp/dist
+$ builder --config=otelcol-builder.yaml --output-path=/tmp/dist
 $ cat > /tmp/otelcol.yaml <<EOF
 receivers:
   otlp:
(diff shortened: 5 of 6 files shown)


Makefile
```diff
@@ -206,6 +203,10 @@ otelcorecol:
 genotelcorecol:
 	pushd cmd/builder/ && $(GOCMD) run ./ --skip-compilation --config ../otelcorecol/builder-config.yaml --output-path ../otelcorecol && popd
 
+.PHONY: ocb
+ocb:
+	$(MAKE) -C cmd/builder ocb
```
jpkrohling, line 208: Where is the binary being generated? At the root of the repository, or at ./bin ?

cmd/builder/internal/command.go
```diff
@@ -83,10 +83,8 @@ func Command() (*cobra.Command, error) {
 
 func initConfig() error {
 	cfg.Logger.Info("OpenTelemetry Collector distribution builder", zap.String("version", version), zap.String("date", date))
-	// use the default path if there is no config file being specified
-	if cfgFile == "" {
-		cfgFile = "$HOME/.otelcol-builder.yaml"
```
jpkrohling, line 88: Interesting, I wonder why we had this here... But if it works, I'm glad to have this code removed.

Makefile
TylerHelmuth: Based on the builder's make file it gets output in `./bin` right?

cmd/builder/internal/command.go
```diff
@@ -59,7 +59,7 @@ func Command() (*cobra.Command, error) {
 	}
 
 	// the external config file
-	cmd.Flags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.otelcol-builder.yaml)")
+	cmd.Flags().StringVar(&cfgFile, "config", "$HOME/.ocb.yaml", "config file (default is $HOME/.ocb.yaml)")
```
TylerHelmuth, line 62: This is a breaking change right?  Is it ok to break this in 1 release or do we need to take a deprecated approach?  I'm thinking of situations where users of the ocb are already taking advantage of the default `.otelcol-builder.yaml` file and that would stop working when this is released.  Could break CICD pipelines.

cmd/builder/internal/command.go
```diff
@@ -83,10 +83,8 @@ func Command() (*cobra.Command, error) {
 
 func initConfig() error {
 	cfg.Logger.Info("OpenTelemetry Collector distribution builder", zap.String("version", version), zap.String("date", date))
-	// use the default path if there is no config file being specified
-	if cfgFile == "" {
-		cfgFile = "$HOME/.otelcol-builder.yaml"
-	}
+
+	cfg.Logger.Info("Using config file", zap.String("path", cfgFile))
```
TylerHelmuth, line 87: I think leaving this log at the end makes sense.  If we move it to the top we could have the situation where we print "Using config file <filename>" and then immediately print "failed to load config file <filename>", which seems confusing.

Makefile
jpeach: Yup, it does the same as `otelcorecol`, just as a developer convenience.

cmd/builder/internal/command.go
jpeach: Yeh, it's a breaking change. We could revert (at the cost of some inconsistency), or we could fall back to `.otelcol-builder.yaml` if `.ocb.yaml` isn't present.

Medium term, I'm wondering whether it makes sense to take the config file only as an argument, not a flag, but I need to do some more research and learning beforehand so that I can make a good proposal :)

cmd/builder/internal/command.go
jpeach: Reverted this change.

cmd/builder/internal/command.go
codeboten: Falling back to `.otelcol-builder.yaml` and adding a warning about it being deprecated in the future seems like a safe way to go.

cmd/builder/internal/command.go
```diff
@@ -38,7 +38,7 @@ var (
 // Command is the main entrypoint for this application
 func Command() (*cobra.Command, error) {
 	cmd := &cobra.Command{
-		Use:  "builder",
+		Use:  "ocb",
 		Long: fmt.Sprintf("OpenTelemetry Collector distribution builder (%s)", version),
```
codeboten, line 42:
```suggestion
		Long: fmt.Sprintf("OpenTelemetry Collector Builder (%s)", version),
```

cmd/builder/internal/command.go
```diff
@@ -83,10 +83,6 @@ func Command() (*cobra.Command, error) {
 
 func initConfig() error {
 	cfg.Logger.Info("OpenTelemetry Collector distribution builder", zap.String("version", version), zap.String("date", date))
```
codeboten, line 85:
```suggestion
	cfg.Logger.Info("OpenTelemetry Collector Builder", zap.String("version", version), zap.String("date", date))
```

codeboten (commented):
Thanks for fixing the inconsistencies! Just a few suggestions, please add an entry to the changelog as well

cmd/builder/internal/command.go
jpeach: Since we need to add a fallback, what I propose is
* deprecate the required`--config` flag
* retain the current `--config` flag implementation for backward compatibility
* default to requiring a single argument that is the path to the config file to build

From my perspective, the default $HOME location for config isn't that useful, since the config ought to be under source control. Compulsory options are better expressed as arguments than flags.

I'll push an update later today.

TylerHelmuth: I am good with this approach, but it does deviate from the collector's ergonomics of passing in configuration via a --config flag.

jpeach: > does deviate from the collector's ergonomics of passing in configuration via a --config flag

Yep, IMHO distinction is that ocb is a compiler that takes a config input and generates a corresponding output, whereas the collector is a long-running service, where it's config is part of the system.

jpeach:
Using the flag:
```
$ ./builder --config=./b.conf
Flag --config has been deprecated, pass the build configuration as the first argument instead
2022-07-19T15:11:53.755+1000	INFO	internal/command.go:119	OpenTelemetry Collector Builder	{"version": "dev", "date": "unknown"}
2022-07-19T15:11:53.756+1000	INFO	internal/command.go:138	Using config file	{"path": "./b.conf"}
2022-07-19T15:11:53.756+1000	INFO	builder/config.go:99	Using go	{"go-executable": "/opt/homebrew/bin/go"}
2022-07-19T15:11:53.756+1000	INFO	builder/main.go:76	Sources created	{"path": "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1100039365"}
2022-07-19T15:11:53.984+1000	INFO	builder/main.go:108	Getting go modules
2022-07-19T15:11:54.017+1000	INFO	builder/main.go:87	Compiling
2022-07-19T15:11:55.675+1000	INFO	builder/main.go:94	Compiled	{"binary": "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1100039365/otelcol-custom"}
```
Using the arg:
```
$ ./builder b.conf
2022-07-19T15:12:24.721+1000	INFO	internal/command.go:119	OpenTelemetry Collector Builder	{"version": "dev", "date": "unknown"}
2022-07-19T15:12:24.721+1000	INFO	internal/command.go:138	Using config file	{"path": "b.conf"}
2022-07-19T15:12:24.721+1000	INFO	builder/config.go:99	Using go	{"go-executable": "/opt/homebrew/bin/go"}
2022-07-19T15:12:24.722+1000	INFO	builder/main.go:76	Sources created	{"path": "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1049441039"}
2022-07-19T15:12:24.918+1000	INFO	builder/main.go:108	Getting go modules
2022-07-19T15:12:24.945+1000	INFO	builder/main.go:87	Compiling
2022-07-19T15:12:25.740+1000	INFO	builder/main.go:94	Compiled	{"binary": "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1049441039/otelcol-custom"}
```
Using both:
```
$ ./builder --config=b.conf b.conf
Flag --config has been deprecated, pass the build configuration as the first argument instead
Error: configuration flag and argument must not both be provided
Usage:
  ocb [flags] FILE
  ocb [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  version     Version of ocb

Flags:
      --description string       A descriptive name for the OpenTelemetry Collector distribution (default "Custom OpenTelemetry Collector distribution")
      --go string                The Go binary to use during the compilation phase. Default: go from the PATH
  -h, --help                     help for ocb
      --module string            The Go module for the new distribution (default "go.opentelemetry.io/collector/cmd/builder")
      --name string              The executable name for the OpenTelemetry Collector distribution (default "otelcol-custom")
      --otelcol-version string   Which version of OpenTelemetry Collector to use as base (default "0.55.0")
      --output-path string       Where to write the resulting files (default "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1945253626")
      --skip-compilation         Whether builder should only generate go code with no compile of the collector (default false)
      --version string           The version for the OpenTelemetry Collector distribution (default "1.0.0")

Use "ocb [command] --help" for more information about a command.

Error: configuration flag and argument must not both be provided
```

Using none:
```
$ ./builder
2022-07-19T15:13:17.637+1000	INFO	internal/command.go:119	OpenTelemetry Collector Builder	{"version": "dev", "date": "unknown"}
2022-07-19T15:13:17.637+1000	ERROR	internal/command.go:123	failed to load config file	{"config-file": "$HOME/.otelcol-builder.yaml", "error": "open $HOME/.otelcol-builder.yaml: no such file or directory"}
go.opentelemetry.io/collector/cmd/builder/internal.initConfig
	/Users/jpeach/upstream/opentelemetry/opentelemetry-collector/cmd/builder/internal/command.go:123
go.opentelemetry.io/collector/cmd/builder/internal.Command.func1
	/Users/jpeach/upstream/opentelemetry/opentelemetry-collector/cmd/builder/internal/command.go:72
github.com/spf13/cobra.(*Command).execute
	/Users/jpeach/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:872
github.com/spf13/cobra.(*Command).ExecuteC
	/Users/jpeach/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:990
github.com/spf13/cobra.(*Command).Execute
	/Users/jpeach/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:918
main.main
	/Users/jpeach/upstream/opentelemetry/opentelemetry-collector/cmd/builder/main.go:26
runtime.main
	/opt/homebrew/Cellar/go/1.18.4/libexec/src/runtime/proc.go:250
2022-07-19T15:13:17.637+1000	INFO	internal/command.go:138	Using config file	{"path": "$HOME/.otelcol-builder.yaml"}
2022-07-19T15:13:17.637+1000	INFO	builder/config.go:99	Using go	{"go-executable": "/opt/homebrew/bin/go"}
2022-07-19T15:13:17.638+1000	INFO	builder/main.go:76	Sources created	{"path": "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1733631529"}
2022-07-19T15:13:17.807+1000	INFO	builder/main.go:108	Getting go modules
2022-07-19T15:13:17.835+1000	INFO	builder/main.go:87	Compiling
2022-07-19T15:13:18.649+1000	INFO	builder/main.go:94	Compiled	{"binary": "/var/folders/hf/8frlwk6s40j0wj3jpgbq94z00000gp/T/otelcol-distribution1733631529/otelcol-custom"}
```

jpeach:
> This should not happen. An info or warn-level message should be logged that the default config is being used, not an error with a stack trace.

This is the standard ocb behavior when the config file isn't found. There's a couple of things I can do here:

1. log deprecation warning when we do the fallback (I didn't do that because the logging hasn't started at the point where we fall back, but can add it if we want)
1.  only fall back if the file exists
1. log the error, but no stack

jpeach:
> This should not happen. An info or warn-level message should be logged that the default config is being used, not an error with a stack trace.

I verified with release build from github, and there is no code to expand the default ` $HOME/.otelcol-builder.yaml`, so the default has never worked AFAIK.

jpkrohling:
> I verified with release build from github, and there is no code to expand the default $HOME/.otelcol-builder.yaml, so the default has never worked AFAIK.

I was sure it used to work, so I picked an old version (0.45.0) and ran `git bisect`, which pointed me to the culprit:

```
78e9640760fc6b0d04cb454b6f94fcb4fc22f647 is the first bad commit
commit 78e9640760fc6b0d04cb454b6f94fcb4fc22f647
Author: Ryan Leung <rleungx@gmail.com>
Date:   Thu Feb 24 03:26:15 2022 +0800

    Replace viper with koanf in builder (#4757)
    
    * replace viper with koanf in builder
    
    Signed-off-by: Ryan Leung <rleungx@gmail.com>
    
    * resolve the conflicts
    
    Signed-off-by: Ryan Leung <rleungx@gmail.com>
    
    * Update go.sum
    
    Co-authored-by: Bogdan Drutu <bogdandrutu@gmail.com>
    Co-authored-by: Bogdan Drutu <lazy@splunk.com>

 cmd/builder/go.mod              | 13 ++------
 cmd/builder/go.sum              | 67 +++++++++++++++++++++++++++++++++++------
 cmd/builder/internal/command.go | 48 ++++++++++++++---------------
 3 files changed, 85 insertions(+), 43 deletions(-)
```

jpeach:
> This means this has not been working since February and we just found out about it. I agree there's no backwards compatibility problem here then. I would be OK with not having a default location for the file

@jpkrohling OK, so I propose to revert all the args changes, and just update the `--config` flag to be required and have no default. Given the history here, that is effectively backwards compatible.

> but we should still have a reasonable default configuration, so that the naked command still produces a reasonable binary

Since the regression you found, anyone depending on the default would get silently incorrect results. Because the binary that is produced without any configuration doesn't have any components, I think it's better to always require the `--config` flag for now.

I can post a follow-up PR that relaxes that and uses the `cmd/otelcorecol/builder-config.yaml` as the default if no config is specified. WDYT?

jpkrohling:
> I can post a follow-up PR that relaxes that and uses the cmd/otelcorecol/builder-config.yaml as the default if no config is specified. WDYT?

Sounds good.

jpeach:
> > I can post a follow-up PR that relaxes that and uses the cmd/otelcorecol/builder-config.yaml as the default if no config is specified. WDYT?
> 
> Sounds good.

Done, and pushed the change.

jpkrohling (approved):

PR #5659, opened 2022-07-09
[builder] add build tag flag

### LINKED ISSUE #5612: Allow builder to use Go build tags/use go command flags
It would be useful to have an option in the builder to pass Go build tags, so that we can test the functionality currently behind the `enable_unstable` build tag. We could either have an option to pass the build tags or have an option to pass arbitrary flags to the `go` command.

This would also help in open-telemetry/opentelemetry-collector-contrib/issues/11867 if we want to keep the `make otelcontribcol-unstable` command after switching to use the builder.

fatsheep9146:
Signed-off-by: Ziqi Zhao <zhaoziqi9146@gmail.com>

**Description:** 
Fix #5612

Add a build tag parameter for builder


cmd/builder/internal/builder/main.go
```diff
@@ -86,7 +86,11 @@ func Compile(cfg Config) error {
 
 	cfg.Logger.Info("Compiling")
 	// #nosec G204
-	cmd := exec.Command(cfg.Distribution.Go, "build", "-ldflags=-s -w", "-trimpath", "-o", cfg.Distribution.Name)
+	args := []string{"build", "-ldflags=-s -w", "-trimpath", "-o", cfg.Distribution.Name}
+	if cfg.Distribution.BuildTags != "" {
+		args = append(args, "-tags", cfg.Distribution.BuildTags)
+	}
+	cmd := exec.Command(cfg.Distribution.Go, args...)
```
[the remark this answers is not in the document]
jpkrohling: Previously, the nosec line was right before the exec.Command, and it's not anymore. Move it here and it will skip the linter. 

That said, I'd be more comfortable if we were to sanitize both the argument that was originally causing G204 (cfg.Distribution.Name) and the new one (cfg.Distribution.BuildTags).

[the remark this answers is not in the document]
fatsheep9146: I move the #nosec to the right place, lint is passed.
@TylerHelmuth @jpkrohling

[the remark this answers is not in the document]
jpkrohling: Are you able to sanitize it in any way, removing the need to skip the check?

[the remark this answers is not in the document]
TylerHelmuth: I agree with @jpkrohling.  It;s not the focus of the PR, but if we are gonna make changes to the Command we might as well try to go all the way and sanitize.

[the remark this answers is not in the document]
fatsheep9146: I will fix this.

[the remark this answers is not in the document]
fatsheep9146: After investigating all existing issues about the G204 error through [securego/gosec](https://github.com/securego/gosec) 

- https://github.com/securego/gosec/issues/204
- https://github.com/securego/gosec/issues/292
- https://github.com/securego/gosec/pull/339
- https://github.com/securego/gosec/issues/343
- https://github.com/securego/gosec/pull/677

even fix a bugs for G204 https://github.com/securego/gosec/pull/835 

I believe for the case of ocb, we should only escape the error by comments. This opinion is also approved by the author of gosec repo

https://github.com/securego/gosec/issues/831#issuecomment-1196292035

What do you think? @TylerHelmuth @jpkrohling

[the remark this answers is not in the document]
fatsheep9146: I think the gosec line is already in the right place =D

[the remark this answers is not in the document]
fatsheep9146: Tracking issue is already created #5781 @jpkrohling

cmd/builder/internal/command.go
```diff
@@ -93,6 +94,7 @@ build configuration given by the "--config" argument.
 	cmd.Flags().StringVar(&cfg.Distribution.OutputPath, distributionOutputPathFlag, cfg.Distribution.OutputPath, "Where to write the resulting files")
 	cmd.Flags().StringVar(&cfg.Distribution.Go, distributionGoFlag, "", "The Go binary to use during the compilation phase. Default: go from the PATH")
 	cmd.Flags().StringVar(&cfg.Distribution.Module, distributionModuleFlag, "go.opentelemetry.io/collector/cmd/builder", "The Go module for the new distribution")
+	cmd.Flags().StringVar(&cfg.Distribution.BuildTags, "build-tags", distributionBuildTags, "A comma-separated list of build tags to consider satisfied during the build")
```
TylerHelmuth, line 97: For consistency I think we should use a const instead of the string "build-tags"

jpkrohling:
The CI is failing due to https://github.com/open-telemetry/opentelemetry-collector/issues/5770

TylerHelmuth (approved):

cmd/builder/internal/command.go
```diff
@@ -39,6 +39,7 @@ const (
 	distributionOutputPathFlag     = "output-path"
 	distributionGoFlag             = "go"
 	distributionModuleFlag         = "module"
+	distributionBuildTags          = "build-tags"
```
bogdandrutu, line 42: Do we need a flag when we have configuration for this? I think we should limit the amount of flags we offer and focus on using the configuration file.

bogdandrutu (changes requested):
@jpkrohling I want to limit the use of flags in general if possible to provide in the config. What is your opinion on this?

cmd/builder/internal/command.go
fatsheep9146: @bogdandrutu @TylerHelmuth @jpkrohling If everyone agreed on this, I can remove args input and only support this field in configuration file.

bogdandrutu: Let's start with that in my opinion

bogdandrutu: Can you do the change since others don't respond, we can always add the flag later since it is backwards compatible to add more things.

fatsheep9146: Sorry for delay, I will change this as you said. @bogdandrutu

fatsheep9146: @bogdandrutu I already removed flag definition of buildtags, leaving it only in the config file.

bogdandrutu (approved):

cmd/builder/internal/command.go
jpkrohling: Just saw this here. I'm not particularly in favor of flags, but I'm in favor of consistency. Either we clearly mark that we are moving away from flags and won't add this, or we add this to be in line with the other options.

jpkrohling (dismissed):
Please, document why we don't have this new option as a flag.

cmd/builder/internal/command.go
bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/pull/5946

PR #5661, opened 2022-07-09
add validation for empty telemetry metric address when level not none

### LINKED ISSUE #5655: Setting an empty metrics address should fail fast
I would expect an empty address [`telemetry::metrics::address`] to return a validation error, instead of silently disabling telemetry. Do you think this is worth an issue?

_Originally posted by @codeboten in https://github.com/open-telemetry/opentelemetry-collector/pull/5642#discussion_r916833143_

fatsheep9146:
Signed-off-by: Ziqi Zhao <zhaoziqi9146@gmail.com>

**Description:** 
Fix #5655

### DIFF
--- a/.chloggen/valiation-for-empty-metric-address.yaml
+++ b/.chloggen/valiation-for-empty-metric-address.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add validation for empty address [telemetry::metrics::address]
+
+# One or more tracking issues or pull requests related to the change
+issues: [5661]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/service/config.go
+++ b/service/config.go
@@ -148,6 +148,10 @@ func (cfg *Config) validateService() error {
 				return fmt.Errorf("pipeline %q references exporter %q which does not exist", pipelineID, ref)
 			}
 		}
+
+		if err := cfg.Service.Telemetry.Validate(); err != nil {
+			fmt.Printf("telemetry config validation failed, %v\n", err)
+		}
 	}
 	return nil
 }
--- a/service/config_test.go
+++ b/service/config_test.go
@@ -85,7 +85,7 @@ func TestConfigValidate(t *testing.T) {
 			name: "custom-service-telemetrySettings-encoding",
 			cfgFn: func() *Config {
 				cfg := generateConfig()
-				cfg.Service.Telemetry.Logs.Encoding = "test_encoding"
+				cfg.Service.Telemetry.Logs.Encoding = "json"
 				return cfg
 			},
 			expected: nil,
@@ -236,6 +236,16 @@ func TestConfigValidate(t *testing.T) {
 				return cfg
 			},
 			expected: errors.New(`unknown pipeline datatype "wrongtype" for wrongtype`),
+		},
+		{
+			name: "invalid-telemetry-metric-config",
+			cfgFn: func() *Config {
+				cfg := generateConfig()
+				cfg.Service.Telemetry.Metrics.Level = configtelemetry.LevelBasic
+				cfg.Service.Telemetry.Metrics.Address = ""
+				return cfg
+			},
+			expected: nil,
 		},
 	}
 
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -15,6 +15,8 @@
 package telemetry // import "go.opentelemetry.io/collector/service/telemetry"
 
 import (
+	"fmt"
+
 	"go.uber.org/zap/zapcore"
 
 	"go.opentelemetry.io/collector/config/configtelemetry"
@@ -112,4 +114,15 @@ type TracesConfig struct {
 	// tracecontext and  b3 are supported. By default, the value is set to empty list and
 	// context propagation is disabled.
 	Propagators []string `mapstructure:"propagators"`
+}
+
+// Validate checks whether the current configuration is valid
+func (c *Config) Validate() error {
+
+	// Check when service telemetry metric level is not none, the metrics address should not be empty
+	if c.Metrics.Level != configtelemetry.LevelNone && c.Metrics.Address == "" {
+		return fmt.Errorf("collector telemetry metric address should exist when metric level is not none")
+	}
+
+	return nil
 }
--- a/service/telemetry/config_test.go
+++ b/service/telemetry/config_test.go
@@ -1,0 +1,63 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package telemetry
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.opentelemetry.io/collector/config/configtelemetry"
+)
+
+func TestLoadConfig(t *testing.T) {
+	tests := []struct {
+		name    string
+		cfg     *Config
+		success bool
+	}{
+		{
+			name: "basic metric telemetry",
+			cfg: &Config{
+				Metrics: MetricsConfig{
+					Level:   configtelemetry.LevelBasic,
+					Address: "127.0.0.1:3333",
+				},
+			},
+			success: true,
+		},
+		{
+			name: "invalid metric telemetry",
+			cfg: &Config{
+				Metrics: MetricsConfig{
+					Level:   configtelemetry.LevelBasic,
+					Address: "",
+				},
+			},
+			success: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.cfg.Validate()
+			if tt.success {
+				assert.NoError(t, err)
+			} else {
+				assert.Error(t, err)
+			}
+		})
+	}
+}


TylerHelmuth (approved):
LGTM

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### 💡 Enhancements 💡
 
 - Add `linux-ppc64le` architecture to cross build tests in CI
+- Add validation for empty address [telemetry::metrics::address] (#5661)
```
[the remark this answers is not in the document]
jpkrohling: +1. Not only that, but we should do this change in two stages: one deprecating the old behavior, another one changing it on a subsequent version.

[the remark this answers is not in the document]
jpkrohling: Can you check the deprecation guidelines from the contribution doc? We should have the process outlined there, along with a few examples.

config/moved_config.go
```diff
@@ -147,6 +148,12 @@ func (cfg *Config) validateService() error {
 			}
 		}
 	}
+
+	// Check when service telemetry metric level is not none, the metrics address should not be empty
+	if cfg.Service.Telemetry.Metrics.Level != configtelemetry.LevelNone && cfg.Service.Telemetry.Metrics.Address == "" {
+		return fmt.Errorf("collector telemetry metric address should exist when metric level is not none")
+	}
+
```
bogdandrutu, line 156: [Idea]: On `cfg.Service.Telemetry` add a `Validation` func and call that one.

What do you think?

CHANGELOG.md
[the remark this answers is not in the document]
fatsheep9146: I've checked the part [Breaking changes](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes) of contributing guideline. 
The listed examples: `Example #1 - Renaming a function`, `Example #2 - Changing the return values of a function` and `Example #3 - Changing the arguments of a function` all do not fit this case.

But I think this case maybe more likely to 
> when renaming or refactoring types, functions, or attributes, we MUST create the new name and MUST deprecate the old one in one version (step 1), and MAY remove it in a subsequent version (step 2). For simple renames, the old name SHALL call the new one.

So I think we should do 2 steps
- in version N+1, keep `validateService` almost unchanged but add a warning log of deprecation, and add `mustValidateService` function to add the new validation logic of telemetry metric
- in version N+2, change `mustValidateService` to become the new version of `validateService`

@jpkrohling @TylerHelmuth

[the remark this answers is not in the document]
TylerHelmuth: I believe that follows the cadence correctly.

config/moved_config.go
fatsheep9146: I saw other config package use `Validate`, not `Validation`, I think I better keep up?

@bogdandrutu

CHANGELOG.md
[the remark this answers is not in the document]
fatsheep9146: I already changed code, please help review this again =D @TylerHelmuth @jpkrohling

config/moved_config.go
fatsheep9146: @bogdandrutu I already changed this with `Validate`

bogdandrutu: yes, `Validate` sorry.

bogdandrutu (commented):
Please fix lint issues

fatsheep9146:
@bogdandrutu @jpkrohling @TylerHelmuth All checks are passed, please help review this again, thanks, =D

config/moved_config.go
```diff
@@ -96,6 +96,7 @@ func (cfg *Config) Validate() error {
 	return cfg.validateService()
 }
 
+// Deprecated: [v0.55.0] Use mustValidateService instead.
```
bogdandrutu, line 99: This is private, why do you need deprecation?

fatsheep9146: I didn't notice that. how to indicate it is deprecated for private function? could your give me a hint?

bogdandrutu: just break it/change it... nobody except us is using that func.

fatsheep9146: I've already removed the comment.

fatsheep9146:
@bogdandrutu I think the failed check is not related to this pr.

jpkrohling:
PR rebased.

fatsheep9146:
> PR rebased.

Done

jpkrohling:
The CI is failing due to https://github.com/open-telemetry/opentelemetry-collector/issues/5770

fatsheep9146:
> The CI is failing due to #5770

Done

fatsheep9146:
ping @jpkrohling, could you review this again?

jpkrohling:
`unittest (1.19)` didn't complete yet, PR rebased to see if this check is triggered.

fatsheep9146:
@jpkrohling unittest for 1.19 is triggered and passed =D

fatsheep9146:
ping @jpkrohling for review this again, thx, =D

jpkrohling:
Sorry, I was out last week. Could you please rebase this? There seems to be a conflict that needs solving.

fatsheep9146:
> Sorry, I was out last week. Could you please rebase this? There seems to be a conflict that needs solving.

I already rebased this.

fatsheep9146:
Could you review this again? @jpkrohling

jpkrohling:
@bogdandrutu, you had comments on this before as well, do you want to review it again?

fatsheep9146:
ping @bogdandrutu @jpkrohling

config/moved_config.go
```diff
@@ -147,9 +147,20 @@ func (cfg *Config) validateService() error {
 			}
 		}
 	}
+
 	return nil
 }
 
+// since mustValidateService will be used for next version, so use nolint to escape the `unused` lint error
+// nolint
+func (cfg *Config) mustValidateService() error {
```
jpkrohling, line 156: In Go, "Must*" functions typically mean that the call never fails (no errors), resulting in panic in case of an error. I'm not saying you should panic here, just that you should set the expectation straight by not using "must" as part of the name.

config/moved_config.go
```diff
@@ -147,9 +147,20 @@ func (cfg *Config) validateService() error {
...
+	if err := cfg.validateService(); err != nil {
+		return err
+	}
+
+	return cfg.Telemetry.Validate()
```
jpkrohling, line 161: Shouldn't this be part of the main Validate function? Why do you need this in a specific extra function (mustValidateService)?

config/moved_config.go
fatsheep9146: That make senses. I originally tried to follow [example 2](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#example-2---changing-the-return-values-of-a-function), so I name the new function to mustValidateService.

What name do you think is better, is `newValidateService` or `newValidateService` ok?  @jpkrohling

config/moved_config.go
fatsheep9146: This also follows [example 2](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#example-2---changing-the-return-values-of-a-function), since I introduced a new validating rule for config file, which may result that the collector with wrong config file crashes when upgraded to new version if the pr is merged. 
So I first mark old validation function as deprecated and use the new validation for next release.

jpkrohling: Oh, I see the confusion now. The example shows how one would proceed in changing the signature of a function to add a new return value (an error).

In your case, you should just add the new condition to the function. But instead of returning an error, you should just log a warning for a couple of releases and then return an error. In this case, given the bug is because the validation does not exist for this condition, adding the condition wouldn't be considered a breaking change (even though it would indeed break already faulty instances).

cc @open-telemetry/collector-contrib-approvers, for more opinions on this.

fatsheep9146: > Oh, I see the confusion now. The example shows how one would proceed in changing the signature of a function to add a new return value (an error).
> 
> In your case, you should just add the new condition to the function. But instead of returning an error, you should just log a warning for a couple of releases and then return an error. In this case, given the bug is because the validation does not exist for this condition, adding the condition wouldn't be considered a breaking change (even though it would indeed break already faulty instances).
> 
> cc @open-telemetry/collector-contrib-approvers, for more opinions on this.

Ok, I got it. So I need to do following things
- remove `mustValidateService` function, just add new validation to original `validateService` function
- for next release(N+1), just log a warning without return error, so the instance with faulty config can still start.
- for N+2 release, return error for this validation. 

@jpkrohling Do I understand this correctly?

jpkrohling: Yes, sounds good to me.

fatsheep9146: @jpkrohling  I already remove `mustValidateService` and move the validation of Telemetry config directly into `validateService ` function, and in validateService function I initialize a zap.logger to log for the validation fail. Please help review this, thank you =D

service/collector_test.go
```diff
@@ -518,27 +518,36 @@ func TestCollectorShutdownBeforeRun(t *testing.T) {
...
-	require.Error(t, col.Run(context.Background()))
+			// Expect run to error
+			require.Error(t, col.Run(context.Background()))
 
-	// Expect state to be closed
-	assert.Equal(t, Closed, col.GetState())
+			// Expect state to be closed
+			assert.Equal(t, Closed, col.GetState())
+		})
+	}
```
bogdandrutu, line 550: Remove unnecessary/unrelated changes.

config/moved_config.go
```diff
@@ -97,6 +100,31 @@ func (cfg *Config) Validate() error {
...
+		InitialFields:     cfg.Telemetry.Logs.InitialFields,
+	}
+	if zapCfg.Encoding == "console" {
+		// Human-readable timestamps for console format of logs.
+		zapCfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	}
+	logger, err := zapCfg.Build()
+	if err != nil {
+		return err
+	}
```
bogdandrutu, line 126: Why this change?

bogdandrutu (changes requested):
Lots of unrelated changes, please keep only changes in service/telemetry

config/moved_config.go
fatsheep9146: > > Oh, I see the confusion now. The example shows how one would proceed in changing the signature of a function to add a new return value (an error).
> > In your case, you should just add the new condition to the function. But instead of returning an error, you should just log a warning for a couple of releases and then return an error. In this case, given the bug is because the validation does not exist for this condition, adding the condition wouldn't be considered a breaking change (even though it would indeed break already faulty instances).
> > cc @open-telemetry/collector-contrib-approvers, for more opinions on this.
> 
> Ok, I got it. So I need to do following things
> 
> * remove `mustValidateService` function, just add new validation to original `validateService` function
> * for next release(N+1), just log a warning without return error, so the instance with faulty config can still start.
> * for N+2 release, return error for this validation.
> 


Becase I talked about the with @jpkrohling , we should first log with warning for next release(N+1), and return error in N+2 release. So I have to output a warning log in validateService function of config/move_config.go. 

But for now, the config.Config does not have a logger field in it. In order to keep the same logging format as collector, so I initialize a logger object same as https://github.com/open-telemetry/opentelemetry-collector/blob/e98116922cf6c079c246fa78eef59dfd9c2e4867/service/internal/telemetrylogs/logger.go#L26

I also think this is not the right way to output warning log in config/move_config.go, do you have any suggestion for this? @bogdandrutu

jpkrohling: It does; would that be a bit too much though? Not sure I see users wanting to enable this feature gate themselves.

fatsheep9146: Honestly, I prefer the logging way, since the validation of config may be enhanced again in the future, not only in this pull request, warning log may be more friendly to both developers and users.

@bogdandrutu @jpkrohling

bogdandrutu: Users will start to ask to configure that, etc. If you want to log something then simply return an error with a good message. Also why configuring a `zap.Logger` vs `fmt.Printf`?

fatsheep9146: Honestly, how to log the warning message also puzzled me for a while. Finally I choose `zap.Logger` because I want to keep the same logging experience with other logs for collector, as following (the first line):

```
2022-09-29T09:17:18.442+0800	warn	config/moved_config.go:180	Telemetry config validation failed	{"reason": "collector telemetry metric address should exist when metric level is not none"}

2022-09-29T09:17:18.442+0800	info	service/telemetry.go:107	Skipping telemetry setup.	{"address": "", "level": "basic"}
2022-09-29T09:17:18.442+0800	info	components/components.go:30	In development component. May change in the future.	{"kind": "exporter", "data_type": "traces", "name": "logging", "stability": "in development"}
```

If use `fmt.Printf` instead, the experience will be
```
Telemetry config validation failed: collector telemetry metric address should exist when metric level is not none

2022-09-29T09:22:57.747+0800	info	service/telemetry.go:107	Skipping telemetry setup.	{"address": "", "level": "basic"}
2022-09-29T09:22:57.747+0800	info	components/components.go:30	In development component. May change in the future.	{"kind": "exporter", "data_type": "traces", "name": "logging", "stability": "in development"}
```

I'm not saying keep the uniform logging experience is necessary, I think we can talk more about this and agree on a right way.

@bogdandrutu

fatsheep9146: Do you have more suggestions? @bogdandrutu

fatsheep9146: ping @bogdandrutu @jpkrohling 

Do you have any suggestion for push forwarding this pr?

jpkrohling: I still think that we should log a warning message in case we have an empty telemetry metrics address and level is not none, returning an error in a future version. I don't think this deserves to be behind a feature gate, though. But I'm good with whatever @bogdandrutu decides.

fatsheep9146: > Honestly, how to log the warning message also puzzled me for a while. Finally I choose `zap.Logger` because I want to keep the same logging experience with other logs for collector, as following (the first line):
> 
> ```
> 2022-09-29T09:17:18.442+0800	warn	config/moved_config.go:180	Telemetry config validation failed	{"reason": "collector telemetry metric address should exist when metric level is not none"}
> 
> 2022-09-29T09:17:18.442+0800	info	service/telemetry.go:107	Skipping telemetry setup.	{"address": "", "level": "basic"}
> 2022-09-29T09:17:18.442+0800	info	components/components.go:30	In development component. May change in the future.	{"kind": "exporter", "data_type": "traces", "name": "logging", "stability": "in development"}
> ```
> 
> If use `fmt.Printf` instead, the experience will be
> 
> ```
> Telemetry config validation failed: collector telemetry metric address should exist when metric level is not none
> 
> 2022-09-29T09:22:57.747+0800	info	service/telemetry.go:107	Skipping telemetry setup.	{"address": "", "level": "basic"}
> 2022-09-29T09:22:57.747+0800	info	components/components.go:30	In development component. May change in the future.	{"kind": "exporter", "data_type": "traces", "name": "logging", "stability": "in development"}
> ```
> 
> I'm not saying keep the uniform logging experience is necessary, I think we can talk more about this and agree on a right way.
> 
> @bogdandrutu

@bogdandrutu Do you have more suggestions? 
I think we can all agree with the opinion that we should use warning logs, not feature-gates.
Since this package does not have import `zap.Logger` , should we just use fmt.Printf?

bogdandrutu: Yes, let's use `fmt.Printf` then

fatsheep9146: Could you help review this again? @bogdandrutu

fatsheep9146: Ping @bogdandrutu

bogdandrutu (approved):

Issue #5572, opened 2022-06-22
Ability to propogate collectors self trace context beyong collector's boundary


pavankrish123:
**Is your feature request related to a problem? Please describe.**

We would like to observe the request flow from collector through various processing services. Currently collector provides  starts a span when exporting a ExportRequest and ends it after completion of the request but it does not propagate the context (default setting). The spans could be viewed through zpages extension. 

While zpages extension actually doesn't need context propagation, cases where we want to stitch collector request (as a client) to processing services down stream need  context propagation to get the single view and it will be invaluable in tracking and correlating request and  processing workflow. 

**Describe the solution you'd like**
1. Perhaps we can enable propagation by setting [autopropogator](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/propagators/autoprop) which provides to set propagators via [env variable ](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/propagators/autoprop/propagator.go#L29) `OTEL_PROPAGATORS` 

2. Be more explicit and allow setting propagators via config in 
```yaml 
service:
   telemetry: 
     traces:
        propagators: tracecontext
```

**Additional context**
I did a little experiment and added the following code 

```go
--- a/service/service.go
+++ b/service/service.go
@@ -17,6 +17,8 @@ package service // import "go.opentelemetry.io/collector/service"
 import (
        "context"
        "fmt"
+       "go.opentelemetry.io/otel"
+       "go.opentelemetry.io/otel/propagation"
 
        "go.opentelemetry.io/otel/metric/nonrecording"
        sdktrace "go.opentelemetry.io/otel/sdk/trace"
@@ -58,6 +60,12 @@ func newService(set *settings) (*service, error) {
                // needed for supporting the zpages extension
                sdktrace.WithSampler(internal.AlwaysRecord()),
        )
+       otel.SetTextMapPropagator(
+               propagation.NewCompositeTextMapPropagator(
+                       propagation.TraceContext{},
+                       propagation.Baggage{},
+               ),
+       )
 

```

The context is propagated to the downstream services

```bash 

[...]

PR #5670, opened 2022-07-11
Enable propagation of the trace context from collector

pavankrish123:
**Description:** 

Added support to propagate traceparent etc headers from collector to the downstream entities that process the export requests from collector. 

**Link to tracking Issue:** [5572](https://github.com/open-telemetry/opentelemetry-collector/issues/5572) 

**Testing:** 

running 

```bash
./otelcorecol_darwin_amd64 --config test.yaml 

```
gives 

```bash 
2022/07/27 21:59:24 request headers: map[Accept-Encoding:[gzip] B3:[e027c66374cdfd8c54704a92cd7759eb-7ce684e76639b14e-0] Content-Encoding:[gzip] Content-Length:[860] Content-Type:[application/x-protobuf] Traceparent:[00-e027c66374cdfd8c54704a92cd7759eb-7ce684e76639b14e-00] User-Agent:[Local OpenTelemetry Collector binary, testing only./0.55.0-dev (darwin/amd64)]]
```
for config 

``` yaml 
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:3000
    traces_endpoint: http://0.0.0.0:3000/v1/trace
    tls:
      insecure: true

  logging:


processors:
  batch:

service:
  telemetry:
    traces:
      propagators:
        - tracecontext
        - b3
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
```

(and)

```bash 
2022/07/28 00:14:57 request headers: map[Accept-Encoding:[gzip] Content-Encoding:[gzip] Content-Length:[859] Content-Type:[application/x-protobuf] User-Agent:[Local OpenTelemetry Collector binary, testing only./0.55.0-dev (darwin/amd64)]]
```
for config 
```
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:3000
    traces_endpoint: http://0.0.0.0:3000/v1/trace
    tls:
      insecure: true

  logging:


processors:
  batch:

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@
 - Deprecate `component.NewExtensionFactoryWithStabilityLevel` in favor of `component.NewExtensionFactory` (#5917)
 
 ### 💡 Enhancements 💡
-- Add support to unmarshalls bytes into pmetric.Metrics with `jsoniter` in jsonUnmarshaler(#5433)
 
+- Add support to unmarshalls bytes into pmetric.Metrics with `jsoniter` in jsonUnmarshaler(#5433)
 - Add httpprovider to allow loading config files stored in HTTP (#5810)
 - Added `service.telemetry.traces.propagators` configuration to set propagators for collector's internal spans. (#5572)
 


service/telemetry/config.go
```diff
@@ -103,3 +104,12 @@ type MetricsConfig struct {
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
 }
+
+// TracesConfig defines settings to configure context propagation of the collector traces. Collector uses
+// go.opentelemetry.io/contrib/propagators/autoprop package to propagate the context and OTEL_PROPAGATORS
+// environment variable can be set to configure the propagators.
```
mx-psi, line 110: We should have a more general description here
```suggestion
// TracesConfig defines the configurable settings for service telemetry traces.
```

The other details are implementation details that I don't think we should document, since it may get out of sync if we decide to use a different propagator.

service/telemetry/config.go
```diff
@@ -103,3 +104,12 @@ type MetricsConfig struct {
...
+type TracesConfig struct {
+	// DisablePropagation disables propagation of the trace context to external backend beyond exporters.
+	// The context is propagated by default.
+	DisablePropagation bool `mapstructure:"disable_propagation"`
```
mx-psi, line 114: I think we should have a propagation section with an enabled field, since (i) this is the pattern we follow elsewhere and (ii) it's possible that we will get other configuration settings related to propagation. So, something like:

```
traces:
  propagator:
    enabled: true
```

I am not familiar with propagation, so maybe the approach should be different (should you have to choose what propagator to use?)

service/telemetry/config.go
```diff
@@ -103,3 +104,12 @@ type MetricsConfig struct {
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
 }
+
+// TracesConfig defines settings to configure context propagation of the collector traces. Collector uses
+// go.opentelemetry.io/contrib/propagators/autoprop package to propagate the context and OTEL_PROPAGATORS
+// environment variable can be set to configure the propagators.
+type TracesConfig struct {
+	// DisablePropagation disables propagation of the trace context to external backend beyond exporters.
+	// The context is propagated by default.
```
mx-psi, line 113: We should have it off by default for now, and think about having it on by default later, after it has had some testing.

cmd/otelcorecol/go.mod
```diff
@@ -59,13 +59,18 @@ require (
 	go.opentelemetry.io/collector/semconv v0.54.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.32.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.32.0 // indirect
+	go.opentelemetry.io/contrib/propagators/autoprop v0.33.0 // indirect
```
mx-psi, line 62: IIUC this is a direct dependency, right? So it should be on the other require block

mx-psi (commented):
👋 I think you should open an issue to discuss the use case for this before we decide on this PR.

I also left some suggestions to make the configuration more consistent with existing practices on the Collector.

service/telemetry/config.go
pavankrish123: @jpkrohling let me know how to proceed. Per your comment in https://github.com/open-telemetry/opentelemetry-collector/issues/5572 I have enabled it by default, shall I heed @mx-psi advice here?

service/telemetry/config.go
pavankrish123: I like this structure propagation does have other possible configurations - like (type of propagator etc) - Let me restructure the propagation into its own nested struct and reissue a PR.

cmd/otelcorecol/go.mod
pavankrish123: Think this is In the context of cmd/otelcorecol module - the files in the module doesn't refer to autoprop directly (just as they don't refer to other indirect dependencies, even though they are direct dependencies for exporters, receivers, service and extensions etc. 

The top level go.mod present in the root folder has go.opentelemetry.io/contrib/propagators/autoprop as direct dependency though.

pavankrish123:
> 👋 I think you should open an issue to discuss the use case for this before we decide on this PR.
@mx-psi the linked issue is in the PR description.  https://github.com/open-telemetry/opentelemetry-collector/issues/5572

cmd/otelcorecol/go.mod
mx-psi: Right :facepalm: looking at the wrong file, thanks!

service/telemetry/config.go
jpkrohling: In the long term, it should be enabled by default, so that we won't break the chain. Given that this is a change in the current behavior, this should be a two-step change: implement the new feature and make it disabled by default, but leave a note to the changelog stating that this will be the enabled by default by N+2.

pavankrish123: Thanks @jpkrohling - I have set the propagation to false by default for now - @mx-psi please review and let me know.

mx-psi (approved):
The config struct looks sensible, and it sounds like a feature we can support at this level. 

I don't feel very comfortable with comitting to autoprop which has had a single version so far and is not 1.0, but I guess since we can still make breaking changes we can change our mind later (specially if we have this off-by-default for now)

service/telemetry/config.go
Aneurysm9: Should this even have an enable/disable flag in the config, or should it be governed by a feature gate until we're comfortable enabling it by default?  The user can always use the `none` "propagator" to disable context propagation.

service/telemetry/config.go
```diff
@@ -103,3 +104,16 @@ type MetricsConfig struct {
...
+type TracesConfig struct {
+	// Configuration related to propagation of collector traces.
+	Propagation PropagationConfig `mapstructure:"propagation"`
+}
+
+// PropagationConfig defines the configuration related to propagation of collectors traces
+type PropagationConfig struct {
+	// Enabled sets context propagation of collector traces to downstream
+	// services consuming collectors export requests.
+	Enabled bool `mapstructure:"enabled"`
```
Aneurysm9, line 118: See [my comment](https://github.com/open-telemetry/opentelemetry-collector/pull/5670#discussion_r918731880) elsewhere.  I don't think `enabled` is the right option here.  I'm not sure any option is required given that propagation can be disabled by using the `none` "propagator".

service/service.go
```diff
@@ -62,6 +64,10 @@ func newService(set *settings) (*service, error) {
 		telemetryInitializer: set.telemetry,
 	}
 
+	if set.Config.Service.Telemetry.Traces.Propagation.Enabled {
```
Aneurysm9, line 67: If this is a feature gate instead we can eliminate this once we're comfortable having propagation enabled by default.  The `none` "propagator" still provides a way for users to prevent propagation.

pavankrish123: @Aneurysm9 just for the sake of my understanding - some clarifications (sorry never dealt with feature gate module before) 
1. By feature gate are you referring to https://github.com/open-telemetry/opentelemetry-collector/tree/main/service/featuregate? 
2. If that is what you are referring to, are you suggesting that we enable the context propagation by default (assuming this is what we want to do eventually) behind a feature gate flag and users can choose to set the OTEL_PROPAGATOR to `none` should they choose to disable the propagation. Thus eliminating need to have any configuration defined completely? 
3. Is there a good example on some features behind the feature gates?

mx-psi: The feature gate sounds good to me and allows us to commit less strongly to this, but I don't think `OTEL_PROPAGATOR` should be the only way to configure this when this is stable (to be honest, I think it's bad that it can be configured at all through an environment variable that we don't control, but I guess we can't avoid that unless the OTel Go SDK changes). If `enabled` is not appropriate, we should have a configuration option equivalent to `OTEL_PROPAGATOR` on the YAML.

pavankrish123:
@Aneurysm9 adjusted the PR per your suggestions. Please let me know WDYT.

service/telemetry/config.go
pavankrish123: per suggestion from @Aneurysm9 - the propagation is enabled by default behind a feature flag :)

pavankrish123:
@jpkrohling please review the PR and let me know.

service/service.go
Aneurysm9: > we should have a configuration option equivalent to `OTEL_PROPAGATOR` on the YAML.

Agreed.  I think this could be accomplished currently by using `os.SetEnv` to set the value of `OTEL_PROPAGATOR` if it isn't already set.  I'll propose that an option be added to the autoprop interface to allow setting a default propagator by ID rather than just by providing an implementation.

> (to be honest, I think it's bad that it can be configured at all through an environment variable that we don't control, but I guess we can't avoid that unless the OTel Go library changes)

This is not going to change.  OTel in general has made the decision that SDKs should be configurable through environment variables and this is in line with that policy.  I think there is significant benefit to operators that may not be able to alter configuration in other ways but do have the ability to control the execution environment.  Ultimately, it would be little different from this configuration:

```yaml
service:
  telemetry:
    trace:
      propagator: ${OTEL_PROPAGATOR}
```

service/telemetry.go
```diff
@@ -123,6 +136,10 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger
 		telAttrs[semconv.AttributeServiceVersion] = buildInfo.Version
 	}
 
+	if tel.registry.IsEnabled(allowTraceContextPropagationFeatureGateID) {
+		otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
```
Aneurysm9, line 140:
```suggestion
		if _, exists := os.LookupEnv("OTEL_PROPAGATOR); !exists && cfg.Traces.Propagator != "" {
			// handle the error, though
			_  = os.SetEnv("OTEL_PROPAGATOR", cfg.Traces.Propagator)
		}
		otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
```

Something like this could potentially be used to enable control of the selector propagators via config if it isn't already set via environment.

pavankrish123: @Aneurysm9 can we provide a factory method to construct the right propagators from comma separated strings in the autoprop module -(which I think is the right place). Setting environment variables from within the process doesn't seem clean and side effect-y

IMHO env variables is something that operators should drive. I suggest we document OTEL_PROPAGATOR and introduce configuration `cfg.tracer.Propgators` only when auto prop adds the factory method. Besides having the factory method take in the same input as what gets passed to OTEL_PROPAGATOR makes the module more consistent. 

cc: @mx-psi  @jpkrohling

service/service.go
pavankrish123: >>>  I'll propose that an option be added to the autoprop interface to allow setting a default propagator by ID rather than just by providing an implementation.

+1

service/telemetry.go
pavankrish123: Had a chat with @Aneurysm9 - We have decided to leave the code as is with autoprop and will let users configure it via OTEL_PROPAGATORS. Once the id based factory api is available, will intoduce collector telemetry.traces.propagators configuration.

mx-psi (approved):

service/telemetry.go
```diff
@@ -123,6 +136,13 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger
 		telAttrs[semconv.AttributeServiceVersion] = buildInfo.Version
 	}
 
+	if tel.registry.IsEnabled(allowTraceContextPropagationFeatureGateID) {
```
jpkrohling, line 139: This absolutely needs to be documented. Can it be done as part of this PR?

pavankrish123: Thanks @jpkrohling   I will add the section to telemetry doc in another PR after this is merged.

bogdandrutu (changes requested):
As a rule of thumb in the collector we are moving away from env variables in general (allow them only as part of the yaml config via embedding). So I think we should offer an yaml alternative first for this feature.

jpkrohling:
> As a rule of thumb in the collector we are moving away from env variables in general

While I agree with the general feeling, this goes into a bigger standard: OpenTelemetry SDK should generally be configurable via env vars.

bogdandrutu:
> OpenTelemetry SDK should generally be configurable via env vars.

Unfortunately this is true because the TC members were lazy (starting with the person who writes this) and have not put the effort into review/propose a configuration (yaml/json) and lazily accepted env vars as the only way. /cc @jsuereth

pavankrish123:
@bogdandrutu, I have updated PR per your suggestion using the new API added to `autoprop` module in `otel-go-contrib`. The API is not released yet, so I will keep the PR in draft. But please take a look into the new commit and let me know if you think I am heading in the right direction, I am adding unit tests  as well. 

Latest tests - 


running 

```bash
./otelcorecol_darwin_amd64 --config test.yaml --feature-gates=telemetry.allowTraceContextPropagation
```

gives 

```bash 
2022/07/27 21:59:24 request headers: map[Accept-Encoding:[gzip] B3:[e027c66374cdfd8c54704a92cd7759eb-7ce684e76639b14e-0] Content-Encoding:[gzip] Content-Length:[860] Content-Type:[application/x-protobuf] Traceparent:[00-e027c66374cdfd8c54704a92cd7759eb-7ce684e76639b14e-00] User-Agent:[Local OpenTelemetry Collector binary, testing only./0.55.0-dev (darwin/amd64)]]
```
for config 

``` yaml 
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:3000
    traces_endpoint: http://0.0.0.0:3000/v1/trace
    tls:
      insecure: true

  logging:


processors:
  batch:

service:
  telemetry:
    traces:
      propagators: "tracecontext,b3"
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
```

(and)

```bash 
2022/07/28 00:14:57 request headers: map[Accept-Encoding:[gzip] Content-Encoding:[gzip] Content-Length:[859] Content-Type:[application/x-protobuf] Traceparent:[00-62f74abcde1f2ca18f0bb7faa8ba385a-9cc603f55b09d95f-00] User-Agent:[Local OpenTelemetry Collector binary, testing only./0.55.0-dev (darwin/amd64)]]
```
for config 
```
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:3000
    traces_endpoint: http://0.0.0.0:3000/v1/trace
    tls:
      insecure: true

  logging:


processors:
  batch:

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
```

service/telemetry.go
```diff
@@ -123,6 +136,19 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger
...
+		props := strings.Split(cfg.Traces.Propagators, ",")
+
+		textMapPropagators, err := autoprop.TextMapPropagator(props...)
```
jpkrohling, line 142: This feature requires the use of OpenTelemetry SDK, which is part of another feature gate, right?

jpkrohling (commented):
Docs or it didn't happen 😉

service/telemetry.go
pavankrish123: Actually no, the feature is merged in master (not yet released) - no feature gates here.

pavankrish123: @jpkrohling  - I will move the PR from draft when go contrib officially releases the feature - Docs will be included - Waiting for the release to happen.

pavankrish123: @jpkrohling Actually no, that feature gate is for using OpenTelemetry SDK metrics (only) - OpenTelemetry SDK was always used for tracing without any feature flags.

pavankrish123:
Hello @bogdandrutu @jpkrohling - The autoprop api is officially released and I have updated the PR. The PR is complete. Please review and give your blessings. :) 

- The feature is available as a collector configuration now per @bogdandrutu suggestion and it is behind a feature gate.
-  I will work on getting docs added under https://opentelemetry.io/docs/collector once the PR is merged. There is no documentation as such in this context (collectors self telemetry) - So will have to write one from scratch.

cmd/builder/go.mod
```diff
@@ -39,6 +39,6 @@ require (
 )
 
 retract (
-	v0.57.0 // Release failed, use v0.57.2
 	v0.57.1 // Release failed, use v0.57.2
+	v0.57.0 // Release failed, use v0.57.2
```
jpkrohling, line 43: What happened here?

go.mod
```diff
@@ -82,7 +87,7 @@ replace go.opentelemetry.io/collector/semconv => ./semconv
 replace go.opentelemetry.io/collector/pdata => ./pdata
 
 retract (
-	v0.32.0 // Contains incomplete metrics transition to proto 0.9.0, random components are not working.
-	v0.57.0 // Release failed, use v0.57.2
 	v0.57.1 // Release failed, use v0.57.2
+	v0.57.0 // Release failed, use v0.57.2
```
jpkrohling, line 91: Was this auto-formatted?

jpkrohling (commented):
Please, revert the "retract" changes to the go.mod files. If they were intentional, please open a new PR with those isolated changes.

Other than that, LGTM, but I believe @bogdandrutu needs to sign off as well, given he had comments before.

go.mod
pavankrish123: yes, fixed now

cmd/builder/go.mod
pavankrish123: fixed now

pavankrish123:
> Please, revert the "retract" changes to the go.mod files. If they were intentional, please open a new PR with those isolated changes.
> 
> Other than that, LGTM, but I believe @bogdandrutu needs to sign off as well, given he had comments before.

done!

pavankrish123:
>>> As a rule of thumb in the collector we are moving away from env variables in general (allow them only as part of the yaml config via embedding). So I think we should offer an yaml alternative first for this feature.

(Done!)

@bogdandrutu can you please review and let me know.

service/telemetry/config.go
```diff
@@ -103,3 +104,11 @@ type MetricsConfig struct {
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
 }
+
+// TracesConfig exposes the common Telemetry configuration for collector's internal spans.
+type TracesConfig struct {
+	// Propagators is a comma separated list of TextMapPropagators from the default registered list
+	// (tracecontext, baggage, b3, b3multi, jaeger, xray, or ottrace). The default propagator used in
+	// tracecontext and setting the vale to "none" disables the context propagation altogether.
+	Propagators string `mapstructure:"propagators"`
```
bogdandrutu, line 113: Why not a proper list in yaml?

service/telemetry/config.go
```diff
@@ -24,6 +24,7 @@ import (
 type Config struct {
 	Logs    LogsConfig    `mapstructure:"logs"`
 	Metrics MetricsConfig `mapstructure:"metrics"`
+	Traces  TracesConfig  `mapstructure:"traces"`
```
bogdandrutu, line 27: Do we need the config? can we rely only on the featuregate initially?

pavankrish123: Hi @bogdandrutu, i think it is necessary unless we are ok enabling just a default propagator once the feature gate is passed. I was instructed to have a configurable option (by you :-) )   in collectors configurations in the last commit, so I added this change. 

let me know if you changed your mind, I can remove this and just support the feature gate + default propagator.

service/telemetry/config.go
bogdandrutu: Also add a comment that this is experimental, until otel defines standard yaml configuration for the library.

service/telemetry/config.go
bogdandrutu: I am just a bit worried that we are defining this before otel does, so let's just add a note that this is experimental until otel defines a yaml config.

cmd/otelcorecol/go.mod
```diff
@@ -59,6 +59,11 @@ require (
 	go.opentelemetry.io/collector/semconv v0.56.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.34.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.34.0 // indirect
+	go.opentelemetry.io/contrib/propagators/autoprop v0.34.0 // indirect
```
bogdandrutu, line 62: Worried about this dependency, since may get quite big. Can we limit to a subset of propagators for the moment?

service/telemetry/config.go
pavankrish123: Done! updated the PR

pavankrish123:
Updated the PR per @bogdandrutu suggestion, the configuration is a list now 

```yaml
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:3000
    traces_endpoint: http://0.0.0.0:3000/v1/trace
    tls:
      insecure: true

  logging:


processors:
  batch:

service:
  telemetry:
    traces:
      propagators:
        - tracecontext
        - b3
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
```

```bash
./otelcorecol_darwin_amd64 --config test.yaml --feature-gates=telemetry.allowTraceContextPropagation
```

produces the following headers in the export request. 

```bash
2022/08/12 11:13:30 request headers: map[Accept-Encoding:[gzip] B3:[d6550dfa8b9c5e0f10acd6533391f08f-4bd7dc181ad45baf-0] Content-Encoding:[gzip] Content-Length:[276] Content-Type:[application/x-protobuf] Traceparent:[00-d6550dfa8b9c5e0f10acd6533391f08f-4bd7dc181ad45baf-00] User-Agent:[Local OpenTelemetry Collector binary, testing only./0.58.0-dev (darwin/amd64)]]
```

cmd/otelcorecol/go.mod
pavankrish123: Good concern to have, I think b3 and traceparent are the most used ones. Let me see if I can compile without getting other ones.

service/telemetry.go
bogdandrutu: I think a simpler implementation is to make this featuregate just control the default of the configuration. So the list is empty if the gate is not enabled, otherwise is `["tracecontext]`? What do you think? Then be default we don't send this, also we have the gate to change the behavior to be default send, and for people that enable the config works without any gate?

cmd/otelcorecol/go.mod
bogdandrutu: Waiting on this?

pavankrish123: @bogdandrutu the way to limit the propagtors is to not use autoprop module, the module  - if we go about using go-contrib/propagators directly then we can limit - The propagators can be enabled via configuration strings but the logic will become sort of parallel/redundant  to what autoprop is doing but with limited options. Are you ok with this ?

pavankrish123:
@bogdandrutu - just summarizing the suggestions from you, to make  sure I didn;t miss any 

1.  Limiting the propagators that we allow (tracecontext, b3) - this can be done if we don't use autoprop module at all(which pulls all the propgators)
2. Changing the default behavior of the feature gate  
   1. when feature gate flag is passed and there is no configuration in collectors yaml, we enable tracecontext by default 
   2. when feature gate flag is not passed and there is no configration in collectors yaml, we don;t propagate context at all. 
   3. when `telemetry.traces` configuration is present, we ignore the feature gate flag and honor the configuration and propagate the context 

Please let me know if my understanding of these changes you suggested is correct?

bogdandrutu:
> Limiting the propagators that we allow (tracecontext, b3) - this can be done if we don't use autoprop module at all(which pulls all the propgators)

Correct

>Changing the default behavior of the feature gate
when feature gate flag is passed and there is no configuration in collectors yaml, we enable tracecontext by default
when feature gate flag is not passed and there is no configration in collectors yaml, we don;t propagate context at all.
when telemetry.traces configuration is present, we ignore the feature gate flag and honor the configuration and propagate the context

This is too complicated, let's go with a simple thing. Remove the featuregate and make default config to be "empty list" which means this is disable by default. Then in a separate PR we can discuss about adding a featuregate to change the default config.

cmd/otelcorecol/go.mod
pavankrish123: Done!

pavankrish123:
Thanks @bogdandrutu - updated the PR per your suggestions. Kindly review

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 ### 💡 Enhancements 💡
 
 - Add httpprovider to allow loading config files stored in HTTP (#5810)
+- Add support to propagate trace context of collector's internal spans if `telemetry.allowTraceContextPropagation` featuregate is enabled (#5572)
```
bogdandrutu, line 20: Please update.

bogdandrutu (approved):

CHANGELOG.md
pavankrish123: oops! Fixed it - Please review and let me know WDYT?

pavankrish123: fixed the PR description  as well.

pavankrish123: thanks @bogdandrutu,  fixed the PR description as well.

PR #5676, opened 2022-07-12
skip changelog for [chore] prs

codeboten:
Add a check in the changelog job to skip the job if the title of the PR contains [chore]. This makes it possible for a user who's change doesn't require a changelog entry and doesn't have write access to the repository to skip this check.

For now I've left the check for the "Skip Changelog" label as well, in the future, this should be removed and the label can be removed altogether.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -28,6 +28,6 @@ jobs:
             echo "A CHANGELOG was modified. Looks good!"
           else
             echo "No CHANGELOG was modified."
-            echo "Please add a CHANGELOG entry, or add \"[chore]\" to the title of the pull request if not required."
+            echo "Please add a CHANGELOG entry, or add either \"[chore]\" to the title of the pull request or add the \"Skip Changelog\" label if not required."
             false
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog.yml
```diff
@@ -28,6 +28,6 @@ jobs:
             echo "A CHANGELOG was modified. Looks good!"
           else
             echo "No CHANGELOG was modified."
-            echo "Please add a CHANGELOG entry, or add the \"Skip Changelog\" label if not required."
+            echo "Please add a CHANGELOG entry, or add \"[chore]\" to the title of the pull request if not required."
```
TylerHelmuth, line 31:
```suggestion
            echo "Please add a CHANGELOG entry, or add either \"[chore]\" to the title of the pull request or add the \"Skip Changelog\" label if not required."
```

TylerHelmuth (commented):
This is a clever solution to the issue where users can't add labels themselves.  I do like keeping the Skip Changelog label for situations where the PR is not a Chore but also might not need a changelog.

.github/workflows/changelog.yml
TylerHelmuth: I guess it depends if you want to specifically callout the PR is a chore for the sake of getting reviewers to quickly review it haha. Seeing [chore] in the title normally tells me I could review a PR relatively fast.

tigrannajaryan (approved):

PR #5678, opened 2022-07-12
[exporter/logging] Decouple `loglevel` field from level of logged messages

mx-psi:
**Description:** Make all messages from logging exporter report at info level.

**Link to tracking Issue:**  https://github.com/open-telemetry/opentelemetry-collector/pull/5677#pullrequestreview-1036046290

cc @tigrannajaryan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 💡 Enhancements 💡
 
 - Add `linux-ppc64le` architecture to cross build tests in CI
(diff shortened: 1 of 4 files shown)


TylerHelmuth (approved):
I like this solution.  I agree with @tigrannajaryan that being able to set only the logging exporter to debug and not the entire collector is very helpful.

tigrannajaryan:
@open-telemetry/collector-approvers any objections to this change?

Issue #5674, opened 2022-07-11
Add support to handle 402 and 413 http error code in Otlp exporter


mcmho:
**Is your feature request related to a problem? Please describe.**
Collector currently does retries 402 and 413 error code.   413 is the error code for PAYLOAD TOO LARGE
 and 402 is for PAYMENT REQUIRED which a user case of this would be when license token is exhausted.  
These error code should be treated as permanent error in the Collector so that they won't get retried. 

**Describe the solution you'd like**
Currently Collector only treats 400 as permanent error.  
```
	if resp.StatusCode == http.StatusBadRequest {
		// Report the failure as permanent if the server thinks the request is malformed.
		return consumererror.NewPermanent(formattedErr)
	}
```
The solution is to add 402 and 413 to the permanent error list.  

**Describe alternatives you've considered**
Tweaking the backend to return a different error code; however, that is a hack.   It is not a permanent solution. 

**Additional context**
it does not makes sense to keep retrying, for example 413, since it will just keep failing the second time and so forth.

jpkrohling:
Sounds reasonable to me.

mcmho:
@jpkrohling ,  can I work on it?

mcmho:
Sorry, for couple attempts, I didn't realize I typed my email address wrong.   Now it is fixed.

PR #5682, opened 2022-07-14
Add support to handle 402 and 413 http error code in Otlp exporter #5674

mcmho:
- added StatusRequestEntityTooLarge (413) and StatusPaymentRequired (402) as failure permanent

**Description:**  
Collector currently does retries 402 and 413 error code. 413 is the error code for PAYLOAD TOO LARGE
and 402 is for PAYMENT REQUIRED which a user case of this would be when license token is exhausted.
These error code should be treated as permanent error in the Collector so that they won't get retried.

**Link to tracking Issue:** 
https://github.com/open-telemetry/opentelemetry-collector/issues/5674

**Testing:**  
otlp_test.go

**Documentation:**
if we currently document http error code behavior, then treating 402 and 413 as permanent failure can be documented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add `linux-ppc64le` architecture to cross build tests in CI
 - `client`: perform case insensitive lookups in case the requested metadata value isn't found (#5646)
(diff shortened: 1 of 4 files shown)


mcmho:
i am so sorry.   for a moment, i didn't realize my email address was messed up.   now is fixed.

PR #5691, opened 2022-07-15
Add guidelines about commit messages and merging/squashing commits

tigrannajaryan:
Our commit messages often are quite lacking. A typical approach is writing single line commit
messages and pushing multiple such commits to a PR. When commits are squashed the commit
messages are concatenated, and we end up with useless commit messages.

Here is an example of a commit message body (I am omitting the first line and the commit hash
to avoid blaming anyone in particular - we are all guilty):

```
    * add docstring

    * fix typo

    * fix impi

    * apply review feedback
```

This is a useless commit message, in fact, it is worse than an empty one.

I tried to keep the added paragraph in the CONTRIBUTING.md short. If you think more details and
convincing is necessary I can make it more elaborate. Note that the text I added doesn't just
call for good commit messages but has some hints about the commit and PR creation, squashing and
merging processes that make it easier to end up with better commit messages.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -196,9 +196,9 @@ so it is a useful habit to write good commit messages before the PR is created.
 Also, unless you actually want to tell a story with multiple commits make sure to squash
 into a single commit before creating the PR.
 
-When merging PRs with multiple commits, they will be squashed and GitHub will
+When maintainers merge PRs with multiple commits, they will be squashed and GitHub will
 concatenate all commit messages right before you hit the "Confirm squash and merge"
-button. Make sure to edit this concatenated message to make it right before merging.
+button. Maintainers must make sure to edit this concatenated message to make it right before merging.
 In some cases, if the commit messages are lacking the easiest approach to have at
 least something useful is copy/pasting the PR description into the commit message box
 before merging (but see above paragraph about writing good commit messages in the first place).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,6 +187,22 @@ $ git commit
 $ git push fork feature
 ```
 
+### Commit Messages
+
+Use descriptive commit messages. Here are [some recommendations](https://cbea.ms/git-commit/)
+on how to write good commit messages.
+When creating PRs GitHub will automatically copy commit messages into the PR description,
+so it is a useful habit to write good commit messages before the PR is created.
+Also, unless you actually want to tell a story with multiple commits make sure to squash
+into a single commit before creating the PR.
+
+When merging PRs with multiple commits, they will be squashed and GitHub will
+concatenate all commit messages right before you hit the "Confirm squash and merge"
+button. Make sure to edit this concatenated message to make it right before merging.
+In some cases, if the commit messages are lacking the easiest approach to have at
+least something useful is copy/pasting the PR description into the commit message box
+before merging (but see above paragraph about writing good commit messages in the first place).
+
 ## General Notes
 
 This project uses Go 1.17.* and [Github Actions.](https://github.com/features/actions)


CONTRIBUTING.md
```diff
@@ -179,6 +179,22 @@ $ git commit
...
+Use descriptive commit messages. Here are [some recommendations](https://cbea.ms/git-commit/)
+on how to write good commit messages.
+When creating PRs GitHub will automatically copy commit messages into the PR description,
+so it is a useful habit to write good commit messages before the PR is created.
+Also, unless you actually want to tell a story with multiple commits make sure to squash
+into a single commit before creating the PR.
+
+When merging PRs with multiple commits, they will be squashed and GitHub will
+concatenate all commit messages right before you hit the "Confirm squash and merge"
+button. Make sure to edit this concatenated message to make it right before merging.
```
mx-psi, line 193: Only maintainers can edit this message, so maybe this should not be on CONTRIBUTING.md?

CONTRIBUTING.md
```diff
@@ -179,6 +179,22 @@ $ git commit
...
 ```
 
+### Commit Messages
+
+Use descriptive commit messages. Here are [some recommendations](https://cbea.ms/git-commit/)
+on how to write good commit messages.
+When creating PRs GitHub will automatically copy commit messages into the PR description,
+so it is a useful habit to write good commit messages before the PR is created.
+Also, unless you actually want to tell a story with multiple commits make sure to squash
+into a single commit before creating the PR.
```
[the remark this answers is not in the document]
TylerHelmuth: I completely agree.  I don't like reviewing updates to a PR when the latest commit is "all the files" instead of only the files that changed.

We could suggest using `git reset --soft HEAD~X` (where X is the number of commits made locally) before pushing to remote branches.  This allows lots of incremental commits locally while preserving a single useful commit message for the remote.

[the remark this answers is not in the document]
tigrannajaryan: That's exactly what it says:

>make sure to squash into a single commit before creating the PR.

Note _before_.

I don't suggest to squash commits after the PR is created.

CONTRIBUTING.md
tigrannajaryan: Do we have a better place for this? I don't think we have a "maintainer's only" secret contributing guide. :-)

CONTRIBUTING.md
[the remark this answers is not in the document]
mx-psi: 👍 Makes sense, misread that

CONTRIBUTING.md
mx-psi: > I don't think we have a "maintainer's only" secret contributing guide. :-)

We don't and we shouldn't :) I don't have a good idea of where it should go, it just feels weird to have something that applies to just a handful of people on the doc that everybody reads to get started contributing

dashpole: I think we should just clarify that it is a message for maintainers (e.g. Maintainers, when merging PRs, should squash commits and edit the commit message to concisely describe the PR)

codeboten, line 193:
```suggestion
When maintainers merge PRs with multiple commits, they will be squashed and GitHub will
concatenate all commit messages right before you hit the "Confirm squash and merge"
button. Maintainers must make sure to edit this concatenated message to make it right before merging.
```

codeboten (approved):
Added a suggestion around calling out maintainers specifically (around squashing commits). Please take a look if the wording makes sense. I'm ok with adding this as a guideline.

CONTRIBUTING.md
tigrannajaryan: Done.

tigrannajaryan: >I think we should just clarify that it is a message for maintainers (e.g. Maintainers, when merging PRs, should squash commits and edit the commit message to concisely describe the PR)
Done.

tigrannajaryan:
Let's see what commit message we end up with when this is merged :-)

CONTRIBUTING.md
```diff
@@ -187,6 +187,22 @@ $ git commit
...
```
bogdandrutu, line 197: [Controversial opinion] we should never need more than one commit, if you need more commits it means you actually need more PRs.

tigrannajaryan: I don't disagree, especially because we force squashing the commits when merging, so multiple commits are only useful as a tool to aid reviewing (i.e. tell a story about how the PR was created). I wanted to leave this option open, although I don't think we use it much (or ever).

djaglowski (approved):

tigrannajaryan:
@open-telemetry/collector-approvers any other comments or good to merge?

mx-psi (approved):

Issue #5699, opened 2022-07-18
Add `ReadHeaderTimeout` values


codeboten:
The following lint errors started showing up after the latest upgrade to golangci-lint:

```
golangci-lint run
config/confighttp/confighttp.go:321:10: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
        return &http.Server{
                Handler: handler,
        }, nil
config/confighttp/compression_test.go:106:12: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
                        srv := &http.Server{
                                Handler: handler,
                        }
config/confighttp/compression_test.go:204:12: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
                        srv := &http.Server{
                                Handler: httpContentDecompressor(handler),
                        }
make: *** [lint] Error 1
```

PR #5700, opened 2022-07-18
[ci] disable G112 to unblock CI

codeboten:
Disabling G112 until https://github.com/open-telemetry/opentelemetry-collector/issues/5699 has been addressed.


TylerHelmuth (commented):
Is this a temporary solution so that builds can keep running or are we not concerned about this code?

TylerHelmuth (approved):

Issue #5654, opened 2022-07-08
Initialize config from collector binary  


svrnm:
Follow up to my question at #5648

**Is your feature request related to a problem? Please describe.**
Not all releases of the otel collector come with a default configuration (linux, windows, mac tar.gz) or the default config is not easily accessible (docker image). As a new comer I need to "go searching" ([otel docs](https://opentelemetry.io/docs/collector/configuration/), examples, ...) to have something I can start from. 

**Describe the solution you'd like**
I was wondering, if there would be interest in adding something like an `--init` option/subcommand to the collector binary that writes a `config.yml` to the current working directory. In a version 1 that command could just write a very basic config file (otlp receivers for everything, otlp exporter (disabled) & logging exporter (enabled)). In a version 2 an addition could either be a set of questions that the end-user can answer to get their perfect setup (like `npm init` or `composer init` etc do it) or options to `--init` like `--init=minimal`, or `--init=full` ...

**Describe alternatives you've considered**
Of course as an alternative the packages for linux, windows, mac could just contain a sample config.yaml.

**Additional context**

This could also simplify the getting started guides for the individual languages, when the OTLP exporter is introduced, e.g.:

The [Configure and run a local collector](https://opentelemetry.io/docs/instrumentation/python/getting-started/#configure-and-run-a-local-collector) section in the python docs (which eventually should be the same for the other languages) could be reduced to 

```
$ otelcol --init
Writing configuration file './config.yaml'
$ otelcol --config=./config.yaml
2022-07-08T11:40:08.867+0200	info	service/telemetry.go:107	Setting up own telemetry...
...
```
(or the equivalent using docker)

neelayu:
@jpkrohling I have created a PR for this issue.

Issue #5223, opened 2022-04-14
Print configuration used bei OTEL Collector at runtime


Mario-Hofstaetter:
**Is your feature request related to a problem? Please describe.**

otelcol can be configured by one or multiple `yaml` configuration files, which also support environment variable expension.
Therefore, especially when using ENV vars, it is useful to be able to view the config otelcol **_is actually running with_**.

AFAIK currently there is no such feature.

Below I will suggest solutions I am using on a daily basis with other observability tools.

To cite @mx-psi [from slack](https://cloud-native.slack.com/archives/C01N6P7KR6W/p1649947052436929?thread_ts=1649945311.538769&cid=C01N6P7KR6W):

> I think there are some security concerns here that would need to be addressed (re: some components need tokens/keys in the config that we wouldn’t want to show), but it definitely sounds like something useful

**Describe the solution you'd like**

All solutions should consider secrets like api tokens, any maybe censor them ******

# Print configuration in command line

Probably the easiest way. For example [Grafana promtail has a CLI flag `-print-config-stderr`](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#printing-promtail-config-at-runtime) which just dumps the config into the commandline, including expanded ENV vars. If the seperate `-config.expand-env` flag is omitted, it is clearly visible where ENV vars are used in the configuration. This could be done for `otelcol` in exactly the same way (however ENV vars are expanded by default in otelcol)

```text
./promtail -print-config-stderr "-config.file=promtail-config.yaml" "-config.expand-env"
```

They also [support a `-dry-run` flag](https://grafana.com/docs/loki/latest/clients/promtail/troubleshooting/#dry-running), so nothing is actually pushed (`otelcol` equivalent for dry-running would be: Don't push logs & traces, don't expose metrics), but I guess that could be a feature request on its own.

# Show configuration in HTTP endpoint

Otelcol has several HTTP endpoints (none active by default?), which could enable showing the active configuration in plain text.
For example, using `extensions: [health_check , zpages]` adds to seperate http endpoints.

Maybe an optional extension "`config`" (lack of a better name) could be configured in a similar fashion to expose the configuration via http.

- When running `prometheus`, its config (html) is available in http://localhost:9090/config

<details><summary>Click to show prometheus web config <<< </summary>

![grafik](https://user-images.githubusercontent.com/33002073/163457463-4754b2ba-eb60-42cc-be08-ca41c878f6a4.png)

</details>

[...]

PR #5710, opened 2022-07-20
Add a collector flag to list available components.

### LINKED ISSUE #5709: Show which components are built into a collector

Especially hen getting started in the ecosystem, it would be helpful for the collector to be able to show which components it has available. This lets you check what features to expect in released builds and builds you make yourself.

Some flag like `--list-components`.

jpeach:
**Description:**

Add a `--list-components` flag to the collector service. This currently
show all the available components, their types and their signal stability
levels. In future, we could also show the Go module and version, but
that would require changes to the builder, so it's omitted for now.

**Link to tracking Issue:** 

This fixes #5709.

**Testing:** 

Just manual.

```
otelcorecol jpeach$ go build . && ./otelcorecol --list-components
exporters:
	logging (traces: in development, metrics: in development, logs: in development)
	otlp (traces: stable, metrics: stable, logs: beta)
	otlphttp (traces: stable, metrics: stable, logs: beta)
receivers:
	otlp (traces: stable, metrics: stable, logs: beta)
processors:
	batch (traces: stable, metrics: stable, logs: stable)
	memory_limiter (traces: beta, metrics: beta, logs: beta)
extensions:
	memory_ballast
	zpages
```

**Documentation:**

None. Visible in usage output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Add the Collector flag `--list-components` to show the available components (#5709)
 
 ## v0.60.0 Beta
 
--- a/service/command.go
+++ b/service/command.go
@@ -16,13 +16,70 @@ package service // import "go.opentelemetry.io/collector/service"
 
 import (
 	"errors"
+	"fmt"
+	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
 
+	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/confmap"
 	"go.opentelemetry.io/collector/confmap/converter/overwritepropertiesconverter"
 	"go.opentelemetry.io/collector/service/featuregate"
 )
+
+func printComponentSet(out io.Writer, factories component.Factories) {
+	stabilityString := func(
+		traces component.StabilityLevel,
+		metrics component.StabilityLevel,
+		logs component.StabilityLevel,
+	) string {
+		var stability []string
+
+		if traces != component.StabilityLevelUndefined {
+			stability = append(stability, fmt.Sprintf("traces: %s", traces))
+		}
+
+		if metrics != component.StabilityLevelUndefined {
+			stability = append(stability, fmt.Sprintf("metrics: %s", metrics))
+		}
+
+		if logs != component.StabilityLevelUndefined {
+			stability = append(stability, fmt.Sprintf("logs: %s", logs))
+		}
+
+		return strings.Join(stability, ", ")
+	}
+
+	fmt.Fprintf(out, "exporters:\n")
+	for _, e := range factories.Exporters {
+		fmt.Fprintf(out, "\t%s (%s)\n",
+			e.Type(),
+			stabilityString(e.TracesExporterStability(), e.MetricsExporterStability(), e.LogsExporterStability()),
+		)
+	}
+
+	fmt.Fprintf(out, "receivers:\n")
+	for _, r := range factories.Receivers {
+		fmt.Fprintf(out, "\t%s (%s)\n",
+			r.Type(),
+			stabilityString(r.TracesReceiverStability(), r.MetricsReceiverStability(), r.LogsReceiverStability()),
+		)
+	}
+
+	fmt.Fprintf(out, "processors:\n")
+	for _, p := range factories.Processors {
+		fmt.Fprintf(out, "\t%s (%s)\n",
+			p.Type(),
+			stabilityString(p.TracesProcessorStability(), p.MetricsProcessorStability(), p.LogsProcessorStability()),
+		)
+	}
+
+	fmt.Fprintf(out, "extensions:\n")
+	for _, e := range factories.Extensions {
+		fmt.Fprintf(out, "\t%s (%s)\n", e.Type(), e.ExtensionStability())
+	}
+}
 
 // NewCommand constructs a new cobra.Command using the given CollectorSettings.
 func NewCommand(set CollectorSettings) *cobra.Command {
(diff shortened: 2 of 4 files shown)


service/command.go
```diff
@@ -15,13 +15,57 @@
...
+		var stability []string
+
+		for _, t := range config.AllDataTypes {
+			if s := f.StabilityLevel(t); s != component.StabilityLevelUndefined {
+				stability = append(stability, fmt.Sprintf("%s: %s", t, s))
+			}
+		}
+
+		if len(stability) == 0 {
+			fmt.Printf("\t%s\n", f.Type())
```
mx-psi, line 42: Could you use `f.Fprintf` so we can unit test this?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Expose `pcommon.NewSliceFromRaw` function (#5679)
 - `loggingexporter`: create the exporter's logger from the service's logger (#5677)
 - Add `otelcol_exporter_queue_capacity` metrics show the collector's exporter queue capacity (#5475)
+- Add the Collector flag `--list-components` to show the available components. (#5709)
```
mx-psi, line 13: This does not look like a flag to me, but like a subcommand (something like `build-info`, `components` or `show-components`?)

mx-psi:
Also this overlaps somewhat with what the [`zpages` extension](https://github.com/open-telemetry/opentelemetry-collector/tree/main/extension/zpagesextension) does. It may be a better idea to add it there, rather than on a subcommand

CHANGELOG.md
jpeach: Yeh I thought about that, but it was more intrusive, and I didn't want to introduce a new thing that the collector wasn't already doing.

service/command.go
jpeach: Yup.

jpeach:
> Also this overlaps somewhat with what the [`zpages` extension](https://github.com/open-telemetry/opentelemetry-collector/tree/main/extension/zpagesextension) does. It may be a better idea to add it there, rather than on a subcommand, which is something we haven't done before.

Yeh, this probably makes sense in zpages. However, what I was looking for here is a quick way to see what I have in a collector binary. zpages makes more sense when I have a fleet of collectors.

CHANGELOG.md
mx-psi: I think it's likely we will eventually have subcommands, some other FRs also make sense as subcommands, like #5654 or #5223. But I agree that it's a significant addition, we may want to put this behind a [feature gate](https://github.com/open-telemetry/opentelemetry-collector/tree/main/service/featuregate#usage) to signal that we haven't yet fully figured out how the final design will look like.

bogdandrutu:
Since we all agree that zpages is a must, let's do that and if we find out that is annoying for this case we can revisit. I feel that we are overthinking, let's start with what we know for sure is needed.

jpeach:
> Since we all agree that zpages is a must, let's do that

I think that's a separate PR. It doesn't work for my use case, which is to check which components are in a binary. zpages might not be enabled in the build. To use the data I have to know that zpages is a thing and have read it's documentation, then configure and run the collector as a service and curl it, which is a lot more effort that I think is necessary, and not a good UX for people who are new to the ecosystem.

jpeach:
> > Since we all agree that zpages is a must, let's do that
> 
> 
> 
> I think that's a separate PR. It doesn't work for my use case, which is to check which components are in a binary. zpages might not be enabled in the build. To use the data I have to know that zpages is a thing and have read it's documentation, then configure and run the collector as a service and curl it, which is a lot more effort that I think is necessary, and not a good UX for people who are new to the ecosystem.

@bogdandrutu so if I'm new to the otel ecosystem, I probably pull down a prebuilt collector binary. I start reading about how I need to configure receivers and exporters and all these optional components. The first thing I need to know is which components I have. IMHO zpages doesn't help this at all.

jpeach:
not stale

TylerHelmuth:
I agree with @jpeach that a subcommand/flag is more useable for this use case. Zpages requires the collector to be running with a valid configuration (that includes the extension) and then you have to know how to use zpages to get to the endpoint that lists all the components. For a brand new user this is not a user-friendly. 

I believe that it is a better user experience to directly ask the collector "what components can I use" without first needing to pass it a valid config (which requires knowledge of what components can be used). Relying on Zpages also requires the binary to include the Zpages extension, which custom distributions may not include. 

 If we really can't add a subcommand or another flag could we at least add the component list to the --help flag?

jpeach:
Rebased again. @open-telemetry/collector-approvers can we get consensus here?

jpkrohling:
> I believe that it is a better user experience to directly ask the collector "what components can I use" without first needing to pass it a valid config

This has my vote.

TylerHelmuth:
@jpeach if we don't make progress on this in the next week I'll add this PR to the SIG meeting next Wednesday.

bogdandrutu:
@TylerHelmuth @jpeach @jpkrohling please provide exactly the commands/flags that you want to add:
* Do you `list` per component type?
* Do we need to support this for windows service handler? What is the separation of things supported for windows service handler vs non windows service? Needs to better understand this.
* Do you have a flag or subcommand? I prefer subcommand. If subcommand probably we should also have a "run" subcommand for current "run" behavior of the root command and deprecate the current root. Also if the decision for flag is just because we are worried about the followup work, I am not ok with that.

jpeach:
>please provide exactly the commands/flags that you want to add

Umm, that's in the patch and the description field for this PR?

> Do you list per component type?

Do you mean a flag that would let you list one type of component and not others? No, the behavior I'm proposing is what's in the PR.

> Do we need to support this for windows service handler?

What is this? If windows has collector components, then this would list them.

> Do you have a flag or subcommand?

It's a flag, because the implementation is trivial and doesn't require a lot of refactoring or backwards compatibility.

bogdandrutu:
> What is this? If windows has collector components, then this would list them.

https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector_windows.go#L45

> It's a flag, because the implementation is trivial and doesn't require a lot of refactoring or backwards compatibility.

I think I already answered to this.

> Also if the decision for flag is just because we are worried about the followup work, I am not ok with that.

jpeach:
OK, sounds like we don't have consensus for this change.

jpkrohling:
> Do you have a flag or subcommand? I prefer subcommand

Not a strong preference, but as a user, I would prefer a flag like '--help' (as I see this as a specialization of '--help'). When looking at the binary, I know it's for running a collector, but before doing that, I might just want to check what's the version and which components it has.

PR #5711, opened 2022-07-20
[exporterhelper] enable Persistent Queue feature by default

swiatekm-sumo:
**Description:**
Persistent Queue is a feature that currently requires "enable_unstable" tag provided to the the build to enable it. We have been using it in [our distro](https://github.com/SumoLogic/sumologic-otel-collector/) for several months already and it looks stable. It seems that others are using this capability as well (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9327#issuecomment-1135494065)

This change enables it by default, which also simplifies the build process a little bit.

**Testing:**
Added some tests to cover the queue creation logic and the requeue feature. They're a bit awkward, and should probably be refactored once the requeue feature is exposed in config.

Originally posted by @pmm-sumo [here](https://github.com/open-telemetry/opentelemetry-collector/pull/5457), see the review discussion therein for justification for some of the choices made.


swiatekm-sumo:
I've rebased this post 0.56.0 and fixed the remaining minor issues the CI got stuck on.

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -43,6 +40,12 @@ type QueueSettings struct {
 	NumConsumers int `mapstructure:"num_consumers"`
 	// QueueSize is the maximum number of batches allowed in queue at a given time.
 	QueueSize int `mapstructure:"queue_size"`
+	// StorageID if not empty, enables the persistent storage and uses the component specified
+	// as a storage extension for the persistent queue
+	StorageID *config.ComponentID `mapstructure:"storage"`
+	// StorageEnabled describes whether persistence via a file storage extension is enabled using the single
+	// default storage extension.
+	StorageEnabled bool `mapstructure:"persistent_storage_enabled"`
```
bogdandrutu, line 48: I don't like this config, is too complex for no good reason. I see no point in supporting `persistent_storage_enabled` instead of simply asking for the exact name.

bogdandrutu (changes requested):
Would like to agree on the config first, since that is the most important part.

exporter/exporterhelper/queued_retry_inmemory.go
swiatekm-sumo: The current state was an attempt to keep the config backwards-compatible, as this whole change was originally just about enabling the queue in the build by default. If we're going to break it, then I agree with @bogdandrutu - just make the user specify the storage explicitly and be done with it, looks like the best option in the long term.

If we do that though, I think the changelog should mention the change explicitly.

bogdandrutu: I would not call this a breaking change since this was an experimental feature.

swiatekm-sumo: It's not technically a breaking change, but it's nonetheless a feature being used in the real world, and I think it'd be polite to call it out in the changelog to make it easier for users to figure out what broke their config after an upgrade.

Either way, I'll make the change in this PR.

bogdandrutu: FYI: Waiting for the change.

swiatekm-sumo: Done, sorry for the delay. I've added a changelog entry to breaking changes, even though this technically isn't one.

CHANGELOG.md
```diff
@@ -12,6 +13,7 @@
 
 ### 💡 Enhancements 💡
 
+- Enable Persistent Queue by default (#5711)
```
tigrannajaryan, line 16: This is a bit misleading. We don't enable persistent queue by default. We just make the feature available in the default build.

Makefile
```diff
@@ -58,7 +58,7 @@ gomoddownload:
 
 .PHONY: gotest
 gotest:
-	@$(MAKE) for-all-target TARGET="test test-unstable"
+	@$(MAKE) for-all-target TARGET="test"
```
tigrannajaryan, line 61: Why is this removed? What if we have other unstable features to test?

exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -71,49 +71,128 @@ func (qCfg *QueueSettings) Validate() error {
...
+	signal             config.DataType
+	cfg                QueueSettings
+	consumerSender     requestSender
+	queue              internal.ProducerConsumerQueue
+	queueStartFunc     internal.PersistentQueueStartFunc
+	retryStopCh        chan struct{}
+	traceAttributes    []attribute.KeyValue
+	logger             *zap.Logger
+	requeuingEnabled   bool
+	requestUnmarshaler internal.RequestUnmarshaler
```
tigrannajaryan, line 121: Ideally this PR should do one thing: enable the feature. Instead I see modifications of the logic. Can we split the PR into 2 parts:
1. Change whatever logic needs to change and whatever refactoring is necessary. Keep the persistent queue feature experimental.
2. Make persistent queue feature part of stable build.
Otherwise I have a hard time understanding what changes are needed for enabling the feature, which changes are refactorings which don't change the behavior and what changes change the behavior.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🛑 Breaking changes 🛑
 
+- Require storage explicitly set for persistent queue (#5711)
```
bogdandrutu, line 7: Please comment, this happens only if you enable the experimental feature xyz...

Makefile
bogdandrutu: I think we should go with feature gates for unstable features from now on, this is an old unstable feature before the gates.

exporter/exporterhelper/README.md
```diff
@@ -28,24 +28,18 @@ The following configuration options can be modified:
...
 
-The maximum number of batches stored to disk can be controlled using `sending_queue.queue_size` parameter (which,
-similarly as for in-memory buffering, defaults to 5000 batches).
+The maximum number of batches stored to disk can be controlled using `sending_queue.queue_size` parameter (which, similarly as for in-memory buffering, defaults to 5000 batches).
 
-When `persistent_storage_enabled` is set to true, the queue is being buffered to disk using 
-[file storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage).
-If collector instance is killed while having some items in the persistent queue, on restart the items are being picked and
-the exporting is continued.
+When persistent queue is enabled, the batches are being buffered to disk using [filestorage] extension. If collector instance is killed while having some items in the persistent queue, on restart the items are being picked and the exporting is continued.
```
bogdandrutu, line 42: wrong statement, they are buffered in the storage :) the storage can be s3 or nvram (whatever the extension implements).

exporter/exporterhelper/internal/persistent_queue.go
```diff
@@ -36,12 +35,19 @@ type persistentQueue struct {
 	storage    persistentStorage
 }
 
-// NewPersistentQueue creates a new queue backed by file storage; name parameter must be a unique value that identifies the queue
-func NewPersistentQueue(ctx context.Context, name string, capacity int, logger *zap.Logger, client storage.Client, unmarshaler RequestUnmarshaler) ProducerConsumerQueue {
-	return &persistentQueue{
+// PersistentQueueStartFunc defines a function that can be used to start the persistent queue
+type PersistentQueueStartFunc func(ctx context.Context, client storage.Client)
```
bogdandrutu, line 39: why is this change necessary now and not before? if this can be "refactored" as a separate PR, please do so.

bogdandrutu (commented):
Please make sure we are not doing too many things in one PR, and keep PRs as small and focus as possible.

exporter/exporterhelper/queued_retry_inmemory.go
swiatekm-sumo: Yeah, this became too complicated along the way. I'm going to split it into three PRs:
1. Make the queue logic consistent between the in-memory queue and the persistent one. This way enabling the persistent queue is just a change to the build.
2. Enable persistent queue in the build by default.
3. Configuration changes for setting the storage for the persistent queue.

swiatekm-sumo: The first PR: https://github.com/open-telemetry/opentelemetry-collector/pull/5764. Would you prefer to close this one right now or wait until all the changes are merged individually? @tigrannajaryan @bogdandrutu ?

bogdandrutu:
This needs some rework.

swiatekm-sumo:
> This needs some rework.

I'll just close it and continue submitting the individual changes as separate PRs. After #5764, I'll do the storage configuration change next.

PR #5728, opened 2022-07-21
[chore] Update check-links workflow to match contrib logic

TylerHelmuth:
**Description:** 
Quick PR to closer align Core's checklink workflow with Contribs.  I am hoping it fixes issues like this: https://github.com/open-telemetry/opentelemetry-collector/runs/7452564621?check_suite_focus=true.

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -22,8 +22,7 @@ jobs:
       - name: Get changed files
         id: changes
         run: |
-          echo "::set-output name=md::$(git diff --name-only --diff-filter=ACMRTUXB origin/${{ github.event.pull_request.base.ref }} ${{ github.event.pull_request.head.sha }} | grep .md$ | xargs)"
-
+          echo "::set-output name=md::$(git diff --name-only --diff-filter=ACMRTUXB $(git merge-base origin/main ${{ github.event.pull_request.head.sha }}) ${{ github.event.pull_request.head.sha }} | grep .md$ | xargs)"
   check-links:
     runs-on: ubuntu-latest
     needs: changedfiles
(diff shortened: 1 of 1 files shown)


bogdandrutu:
> Description:
Quick PR to closer align Core's checklink workflow with Contribs. I am hoping it fixes issues like this: https://github.com/open-telemetry/opentelemetry-collector/runs/7452564621?check_suite_focus=true.

[just for me to understand the logic]: Can you please explain how? That to me looks like a problem in the tool.

TylerHelmuth:
> That to me looks like a problem in the tool.

I don't think so.  In the build I linked the check-links job fails because there is a bad link in `docs/release.md`.  The command being run is exactly 

```
markdown-link-check \
    --verbose \
    --config .github/workflows/check_links_config.json \
    CHANGELOG.md docs/release.md
```

So the job is correctly running against the 2 files it was told to and finding a bad link in one of them.  The problem is that by branch did not modify the `docs/release.md` file, so it shouldn't get passed to markdown-link-check.

In the "changed files" job the runs before the check-links job, the git diff command that checks what files changes is `git diff --name-only --diff-filter=ACMRTUXB origin/main f34584b43df46ad8afa134184311cd45ec07bd31 | grep .md$ | xargs`.  When I run that locally I get `CHANGELOG.md` and `docs/release.md`.  

If I run this PR's version of the git diff command for that PR instead, it only returns `CHANGELOG.md`, which is correct.

TylerHelmuth:
@djaglowski is the git wizard who [fixed this issue in Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/11443).  Comparing against the common ancestor using `git  merge-base` instead of directly against `main` is the key.

djaglowski (approved):

bogdandrutu (approved):

PR #5729, opened 2022-07-21
[chore] add checklinks to makefile

TylerHelmuth:
**Description:** 
Adds checklinks capability to make.  This allows checking links locally and aligns Core and Contrib's checklinks capabilities.


Makefile.Common
```diff
@@ -26,6 +30,12 @@ test-with-cover:
 benchmark:
 	$(GOTEST) -bench=. -run=notests ./...
 
+.PHONY: checklinks
+checklinks:
+	command -v $(MDLINKCHECK) >/dev/null 2>&1 || { echo >&2 "$(MDLINKCHECK) not installed. Run 'npm install -g markdown-link-check'"; exit 1; }
+	find . -name \*.md -print0 | xargs -0 -n1 \
+		$(MDLINKCHECK) -q -c $(SRC_ROOT)/.github/workflows/check_links_config.json || true
```
bogdandrutu, line 37: I don't think this should run per module correct? we can have only one checklink for the entire repo?

TylerHelmuth: In my opinion it is useful to have both options.  When developing in a specific component it is nice to be able to run checklinks for that components files only, as it is much faster.  This should allow Core developers to run `make checklinks` anywhere there is a makefile with that includes `Makefile.Common`.  In Contrib thats basically every component.  If thats not the situation for Core then 1 checklinks for the whole repo works as well.

bogdandrutu: I don't think that is the case in core.

TylerHelmuth: Ok, moved to Makefile instead of Common.  Can now run `make checklinks` to check all the links in Core.

codeboten (approved):

bogdandrutu (approved):

PR #5734, opened 2022-07-22
Warn on unknown environment variable; add feature gate to error out

### LINKED ISSUE #5615: Log a warning when an environment variable without a value is expanded
Related to #5614 

In https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/11846 the user was trying to use `"$ConnectionString"` as a Username. The collector automatically interprets `$` as an environment variable expansion. The environment variable was empty, so the `Username` field was interpreted as empty and failed validation. There was nothing logged in the collector to alert this user that their configuration was being mangled by the failed environment expansion.

When an environment variable expansion has no value, log a warning.

mx-psi:
**Description:**

- Add warning when expanding unknown environment variable.
- Add `confmap.expandconverter.RaiseErrorOnUnknownEnvVar` feature gate to turn this warning into an error

**Link to tracking Issue:** Fixes  #5615

**Testing:** Added unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 
 ### 💡 Enhancements 💡
 
+- Warn when expanding unknown environment variable (#5734)
+  - The `confmap.expandconverter.RaiseErrorOnUnknownEnvVar` feature gate will turn this into an error.
+
 ### 🧰 Bug fixes 🧰
 
 ## v0.56.0 Beta
--- a/confmap/converter/expandconverter/expand.go
+++ b/confmap/converter/expandconverter/expand.go
@@ -16,51 +16,123 @@ package expandconverter // import "go.opentelemetry.io/collector/confmap/convert
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"sort"
 
 	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/internal/nonfatalerror"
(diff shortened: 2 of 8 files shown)


mx-psi:
It looks like this needs a bigger refactor than what I originally thought :( We need to pass the logger to be able to log this warning, but for that we need to create the converter later in the process

mx-psi:
This is the best I could come up with, although it comes with a lot of plumbing. The logger is not available when loading the configuration, so we need to have a way to bubble up the information that we want to be logged. To do that, I created a new `nonfatalerror` package to signal that an error is not fatal.

@dyladan (or anyone else :)) can you think of a better solution for doing this?

dyladan:
Hmm I see the issue with the logger not being constructed yet. Honestly it might be better to just fail instead of gating that behavior. The only risk is that existing collector configs might start failing on new collector versions but the collector is still 0.x so I think it isn't out of bounds to do that. Might even help users catch issues they didn't know they had.

dyladan:
nonfatalerror is probably the same thing I would have come up with if you want the fail fast strategy to be gated but I might not have used the same name. The error _might_ be fatal if there is a gate set, so it's not really the same. It's more of a _sometimes_ fatal error than a nonfatal error. I'm trying to think of a word other than fatal which describes something that is normally not fatal but can be fatal if in a strict mode.

mx-psi:
Asked on #5615 for more people to participate in the discussion.

PR #5747, opened 2022-07-25
Update dependabot pr GH action to generate otelcorecol and include all files

bogdandrutu:
* No reason to "build" the collector, instead we need to make sure we generate the collector automatic files to be up to date.
* No reason not to skip any other file.


bogdandrutu:
See diff https://github.com/open-telemetry/opentelemetry-collector/pull/5744 and https://github.com/open-telemetry/opentelemetry-collector/pull/5746

codeboten (approved):

PR #5756, opened 2022-07-27
[exporter/exporterhelper] allow to specify max queue size in bytes

### LINKED ISSUE #5213: [exporter/exporterhelper] allow to specify max queue size in bytes
Currently it's possible to specify the maximum size of the queue (both memory-backed and persistent) in number of batches. It would be useful to specify max size in bytes (especially for the persistent queue)

We could add another parameter, which could be used to describe the maximum size in bytes. To avoid confusing the queue size limit in number of batches vs number of bytes, perhaps `queue_size` could be deprecated (and used as an alternative for one of the new parameters). Then we could have `queues_size_batches` and `queue_size_bytes` or so (other suggestions welcome).

When both of them are specified, hitting any would cause dropping data

aboguszewski-sumo:
**Description:** 
This PR resolves a problem described more precisely in the linked issue. 
TL;DR: The persistent queue now supports specifying an upper limit for the queue capacity in bytes. The `queue_size` parameter is now deprecated and serves as an alias for `queue_size_batches`. 

**Link to tracking Issue:** 
Resolves #5213 

**Testing:** 
Some of the old tests have been modified insignificantly, by adding a bonus parameter to some functions etc. 
The main functionality added in this PR has been tested with a test very similar to the test for the capacity specified in batches.

### DIFF
--- a/exporter/exporterhelper/README.md
+++ b/exporter/exporterhelper/README.md
@@ -17,7 +17,7 @@ The following configuration options can be modified:
 - `sending_queue`
   - `enabled` (default = true)
   - `num_consumers` (default = 10): Number of consumers that dequeue batches; ignored if `enabled` is `false`
-  - `queue_size` (default = 5000): Maximum number of batches kept in memory before dropping; ignored if `enabled` is `false`
+  - `queue_size_batches` (default = 5000): Maximum number of batches kept in memory before dropping; ignored if `enabled` is `false`
   User should calculate this as `num_seconds * requests_per_second / requests_per_batch` where:
     - `num_seconds` is the number of seconds to buffer in case of a backend outage
     - `requests_per_second` is the average number of requests per seconds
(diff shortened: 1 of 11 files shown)


bogdandrutu (changes requested):
We declared the configuration for this stable, and is probably one of the most used config, so cannot break it.

aboguszewski-sumo:
Do you mean the change from `queue_size` to `queue_size_batches`? It's fine, I can revert it since it's not a must-have change in this case. I don't think it's a breaking change though, after these changes if `queue_size` has been specified in the config, a warning is logged and that value is used instead of `queue_size_batches` so that the old builds won't break.

bogdandrutu:
> I don't think it's a breaking change though, after these changes if queue_size has been specified in the config, a warning is logged and that value is used instead of queue_size_batches so that the old builds won't break.

What is the motivation to have all this logic? Why is not the storage extension configured to have a size in bytes instead of the user?

aboguszewski-sumo:
I am not sure I understand what's the problem, so I will elaborate on what this PR is about:
- The main change (specifying the queue size in bytes) affects only the persistent queue, which is experimental at the moment (but there are plans to make it stable: #5711).
- Limiting the queue size in bytes directly in the queue config seems more intuitive. It is already possible to specify the limit in batches here, so it's logical to put other similar limits here. Moreover, the storage as the place to configure this does not sound intuitive at all - the limit of the queue is a thing that is related directly to the queue, and storage is just a tool used to implement the queue.
- The deprecation of `queue_size` in favor of `queue_size_batches` was only a suggestion from the issue and is not necessary.

cc: @swiatekm-sumo

bogdandrutu:
This PR needs to be re-created after the cleanup of enabling the persistent queue.

aboguszewski-sumo:
I've recreated this PR on another branch, but it seems that there are some issues that need to be fixed in the persistent queue before this feature can be added seamlessly. I'm closing this for now and will create another PR when it will be time for it.

Issue #5692, opened 2022-07-15
Allow user to pass in metrics namespace for prometheus exporter


splunkericl:
### Is your feature request related to a problem? Please describe.
Currently, all metrics emitted in otel collector have [namespace set ](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry.go#L176)to `otelcol`. If the otel collector user adds their own metrics they would end up having the same prefix. For our team, we are not interested to have the prefix for our metrics name.

From the otel collector user perspective, the metrics namespace should be customizable. If the user decides not to have any namespace to make the metrics more readable, it should be an option.

### Describe the solution you'd like
A new field is added in `ServiceTelemetryMetrics` in `go.opentelemetry.io/collector/config/service.go`

```
type ServiceTelemetryMetrics struct {
...
namespace string
}
```

the namespace is passed in to prometheus exporter namespace
```
opts := prometheus.Options{
	Namespace: col.service.config.Telemetry.Metrics.namespace,
}
```

### Describe alternatives you've considered
user can remove the name prefix when scraping the metric

### Additional context
N/A

dmitryax:
Hey @splunkericl. I'm curios what is your use case? 

> If the otel collector user adds their own metrics they would end up having the same prefix

What metrics do want to add that ends up with `otelcol` prefix? Can you please provide an example?

The `otelcol` namespace is applied only to a specific set of collector's metrics, not the metrics that the collector collects.

dmitryax:
Also the issue title mentions [prometheus exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter) which not part of this repo

tigrannajaryan:
@dmitryax the use case is custom builds or custom distros of the Collector which have a name that is different from "OpenTelemetry Collector". In such cases you may want to change own metric names to avoid using "otelcol" namespace to avoid confusion (in some environments people may have canonical Otel Collector and custom distros co-existing, so differentiating their metrics can be useful). The hard-coded "otelcol" namespace value is currently used as a prefix for all own metrics that the Collector exposes (by default we expose them as Prometheus metrics on http://localhost:8888/metrics).
I think the capability to customize the metric namespace fits into our general customization story where we allow for example to change the short name and full name of the Collector (via BuildInfo.Command and BuildInfo.Description).

dmitryax:
It makes sense now. Thanks for clarifying.

[...]

PR #5758, opened 2022-07-28
Allow user to pass in metrics namespace

splunkericl:
**Description:**  
Adding a configuration that allows user to configure their metrics namespace

**Link to tracking Issue:** 
https://github.com/open-telemetry/opentelemetry-collector/issues/5692

**Testing:**
Added a unit test to verify namespace can be customized

**Documentation:**
Added godoc on the new `namespace` field

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 - `ocb` now exits with an error if it fails to load the build configuration. (#5731)
 - Deprecate `HTTPClientSettings.ToClientWithHost` (#5737)
-- Add a new field `namespace` in `service.telemetry.metrics.namespace` to allow user to customize metric namespace (#5692)
+- Add a new field `service.telemetry.metrics.namespace` to allow users to customize the collector's metric namespace (#5692)
 
 ### 🧰 Bug fixes 🧰
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - `ocb` now exits with an error if it fails to load the build configuration. (#5731)
 - Deprecate `HTTPClientSettings.ToClientWithHost` (#5737)
+- Add a new field `namespace` in `service.telemetry.metrics.namespace` to allow user to customize metric namespace (#5692)
 
 ### 🧰 Bug fixes 🧰
 
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -54,6 +54,7 @@ const (
 	// useOtelForInternalMetricsfeatureGateID is the feature gate ID that controls whether the collector uses open
 	// telemetrySettings for internal metrics.
 	useOtelForInternalMetricsfeatureGateID = "telemetry.useOtelForInternalMetrics"
+	defaultNameSpace                       = "otelcol"
 )
 
 type telemetryInitializer struct {
@@ -173,8 +174,12 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
 	}
 
 	// Until we can use a generic metrics exporter, default to Prometheus.
+	namespace := defaultNameSpace
+	if cfg.Metrics.NameSpace != "" {
+		namespace = cfg.Metrics.NameSpace
+	}
 	opts := prometheus.Options{
-		Namespace: "otelcol",
+		Namespace: namespace,
 	}
 
 	opts.ConstLabels = make(map[string]string)
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -175,8 +175,8 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
 
 	// Until we can use a generic metrics exporter, default to Prometheus.
 	namespace := defaultNameSpace
-	if cfg.Metrics.NameSpace != "" {
-		namespace = cfg.Metrics.NameSpace
+	if cfg.Metrics.Namespace != nil {
+		namespace = *cfg.Metrics.Namespace
 	}
 	opts := prometheus.Options{
 		Namespace: namespace,
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -103,6 +103,10 @@ type MetricsConfig struct {
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
 
-	// Namespace indicates the telemetry namespace the service will set. If nil, a default value is assigned.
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 - `ocb` now exits with an error if it fails to load the build configuration. (#5731)
 - Deprecate `HTTPClientSettings.ToClientWithHost` (#5737)
+- Add a new field `namespace` in `service.telemetry.metrics.namespace` to allow user to customize metric namespace (#5692)
```
TylerHelmuth, line 17:
```suggestion
- Add a new field `service.telemetry.metrics.namespace` to allow users to customize the collector's metric namespace (#5692)
```

service/telemetry.go
```diff
@@ -173,8 +174,12 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
 	}
 
 	// Until we can use a generic metrics exporter, default to Prometheus.
+	namespace := defaultNameSpace
+	if cfg.Metrics.NameSpace != "" {
```
tigrannajaryan, line 178: The [issue](https://github.com/open-telemetry/opentelemetry-collector/issues/5692) description said:
>If the user decides not to have any namespace to make the metrics more readable, it should be an option.

It doesn't seem to be possible to specify empty string as a namespace with this implementation. Is it a non-goal?

tigrannajaryan (commented):
@open-telemetry/collector-approvers the issue was not triaged but the feature seems reasonable to me. Any thoughts on whether we want this accepted?

service/telemetry/config.go
```diff
@@ -102,4 +102,7 @@ type MetricsConfig struct {
 
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
+
+	// NameSpace indicates the telemetry namespace the service will set.
+	NameSpace string `mapstructure:"namespace"`
```
TylerHelmuth, line 107: I'd recommends going with `Namespace` instead, since it is a single word.

service/telemetry.go
```diff
@@ -173,8 +174,12 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
...
+		namespace = cfg.Metrics.NameSpace
+	}
 	opts := prometheus.Options{
-		Namespace: "otelcol",
+		Namespace: namespace,
```
TylerHelmuth, line 182: This will only change the namespace when using OpenCensus metrics.  For users that enable the `useOtelForInternalMetricsfeatureGateID` feature gate and use OTEL metrics, is the namespace not relevant? Or does `initOpenTelemetry` also need to be made aware of the namespace?

service/telemetry.go
splunkericl: I actually do want to support empty namespace. But I am also worried that will break a lot of existing users' metrics. As now their metrics will not have "otelcol" prefix by default. Will that kind of breaking changes be acceptable in general?

service/telemetry/config.go
splunkericl: sounds good. will update

service/telemetry.go
splunkericl: there isn't a namespace set in the opentelemetry metrics(through `useOtelForInternalMetricsfeatureGateID`). I think there are still a lot of designs around that with the beta milestone so I will keep it as it is.

service/telemetry.go
tigrannajaryan: No, we should not make breaking changes. "otelcol" must remain the default.

However, as one possible option, we can differentiate between unspecified `namespace` setting (meaning that default "otelcol" should be used) and explicitly specified empty string `namespace: ""`in the config. This can be done if `Namespace` is declared as `Namespace *string` optional setting in the `MetricsConfig`. I am not entirely sure this behavior is clear and unsurprising enough, perhaps there are other ways to allow specifying empty string, but also keep the current behavior as the default.

splunkericl: Having a string pointer as config does seem a bit confusing. Another alternative is to use an additional bool field `UseCustomNamespace`. But then now we have two fields which increases maintenance effort...

I can do either. Do you have a preference?

tigrannajaryan: >Having a string pointer as config does seem a bit confusing.

IIRC using pointers is the idiomatic way to handle/detect optional config settings in Go yaml parsers. That itself should be fine and is not a problem in the code.

What I am more concerned about is whether having the default value of a string config setting as "otelcol" is a surprising behavior for the end users or we think it is fine. It is probably fine. I am curious what other @open-telemetry/collector-approvers think.

splunkericl: ok if that is the convention. I can update to use the string pointer.

Regarding the default settings of "otelcol", I think that is ok as other otel collector also provide default configs or some sort. But if anyone from that list disagrees please comment on the thread.

tigrannajaryan: Let's see if anyone else objects and if not let's go with the pointer.

dmitryax:
Issue description is bit confusing. Can you please reply to https://github.com/open-telemetry/opentelemetry-collector/issues/5692#issuecomment-1201571189

tigrannajaryan (approved):

tigrannajaryan:
@dmitryax @TylerHelmuth any other comments or good to merge?

TylerHelmuth (approved):
LGTM.  Thanks @splunkericl!

service/telemetry/config.go
```diff
@@ -102,4 +102,7 @@ type MetricsConfig struct {
 
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
+
+	// Namespace indicates the telemetry namespace the service will set. If nil, a default value is assigned.
```
dmitryax, line 106: The description is not very clear. As an otel collector user, I don't understand it without looking into the code. Please mention that it changes namespace of exposed OTel collector's metrics and that default namespace is `otelcol`. It maybe good to also show an example of how the metrics will look like if the namespace is changed for a generic metric like `otelcol_process_uptime`

splunkericl: that is fair. Will add it to be more explicit.

splunkericl: added in the new version

bogdandrutu (changes requested):
This is only supported in opencensus (based on my knowledge), are we sure that we want to add this as another blocker to adopt opentelemetry?

PR #5762, opened 2022-07-29
Define a new API for components stability level to match extensions as well

bogdandrutu:
* Deprecates current `StabilityLevel(config.DataType)` function from the base `Factory` which is not well suited for Extension. 
* Add new specific for each component `Stability` functions.

### DIFF
--- a/component/exporter.go
+++ b/component/exporter.go
@@ -68,20 +68,29 @@ type ExporterFactory interface {
 	// tests of any implementation of the Factory interface.
 	CreateDefaultConfig() config.Exporter
 
-	// CreateTracesExporter creates a trace exporter based on this config.
+	// CreateTracesExporter creates a TracesExporter based on this config.
 	// If the exporter type does not support tracing or if the config is not valid,
 	// an error will be returned instead.
 	CreateTracesExporter(ctx context.Context, set ExporterCreateSettings, cfg config.Exporter) (TracesExporter, error)
 
-	// CreateMetricsExporter creates a metrics exporter based on this config.
+	// TracesExporterStability gets the stability level of the TracesExporter.
+	TracesExporterStability() StabilityLevel
+
+	// CreateMetricsExporter creates a MetricsExporter based on this config.
 	// If the exporter type does not support metrics or if the config is not valid,
 	// an error will be returned instead.
 	CreateMetricsExporter(ctx context.Context, set ExporterCreateSettings, cfg config.Exporter) (MetricsExporter, error)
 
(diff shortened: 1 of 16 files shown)


bogdandrutu:
Looking for feedback on this PR, before I will do the changelog and other changes.

extension/zpagesextension/factory.go
```diff
@@ -31,7 +31,7 @@ const (
 
 // NewFactory creates a factory for Z-Pages extension.
 func NewFactory() component.ExtensionFactory {
-	return component.NewExtensionFactory(typeStr, createDefaultConfig, createExtension)
+	return component.NewExtensionFactoryWithStabilityLevel(typeStr, createDefaultConfig, createExtension, component.StabilityLevelBeta)
```
jpkrohling, line 34: This is fine for this PR, but perhaps we should try to have consumers changed in a follow-up PR. This would help assert that the changes are indeed backward compatible. This did leave me wondering if a similar change wouldn't be needed in the other components from this repo, like the OTLP receiver/exporter.

component/exporter.go
```diff
@@ -68,20 +68,29 @@ type ExporterFactory interface {
...
 
-	// CreateTracesExporter creates a trace exporter based on this config.
+	// CreateTracesExporter creates a TracesExporter based on this config.
 	// If the exporter type does not support tracing or if the config is not valid,
 	// an error will be returned instead.
 	CreateTracesExporter(ctx context.Context, set ExporterCreateSettings, cfg config.Exporter) (TracesExporter, error)
 
-	// CreateMetricsExporter creates a metrics exporter based on this config.
+	// TracesExporterStability gets the stability level of the TracesExporter.
+	TracesExporterStability() StabilityLevel
```
TylerHelmuth, line 77: Is the advantage to breaking out stability level by signal and component type that we no longer need to pass in a DataType to get the stability?

bogdandrutu: also, this pattern works for Extensions which don't have a data type.

TylerHelmuth: > which don't have a data type.

makes sense.

bogdandrutu: Thanks.

codeboten (approved):

extension/zpagesextension/factory.go
codeboten: Receivers, processors, and exporters have already had stability levels added in an earlier change. Extension were the only components left out from my earlier change

PR #5764, opened 2022-07-31
[exporterhelper] Make the in-memory and persistent queues more consistent

swiatekm-sumo:
**Description:**
This change is part of an effort to enable the persistent queue by default, without requiring the `enable_unstable` tag. Right now, there are minor inconsistencies between the queue behaviours, and this PR removes them. The intent is for enabling the persistent queue in the build by default to not change any behaviour.

More specifically:
* queue names are now consistent, which required a change to a metric label for the persistent queue
* the in-memory queue has the `requeuing` feature, though it's not enabled
* added some missing tests for persistent queue configuration

This was originally part of #5711, which ended up doing too many things at once.

**Link to tracking Issue:** N/A

**Testing:**
Added some tests for parts of the persistent queue that weren't originally covered.

### DIFF
--- a/exporter/exporterhelper/queued_retry_inmemory.go
+++ b/exporter/exporterhelper/queued_retry_inmemory.go
@@ -79,36 +79,59 @@ type queuedRetrySender struct {
 	retryStopCh     chan struct{}
 	traceAttributes []attribute.KeyValue
 	logger          *zap.Logger
+	// currently this is always false for the in-memory queue
+	// it's here for consistency with the persistent queue
+	requeuingEnabled bool
 }
 
 func newQueuedRetrySender(id config.ComponentID, _ config.DataType, qCfg QueueSettings, rCfg RetrySettings, _ internal.RequestUnmarshaler, nextSender requestSender, logger *zap.Logger) *queuedRetrySender {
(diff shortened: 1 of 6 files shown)


exporter/exporterhelper/queued_retry_inmemory.go
```diff
@@ -79,36 +79,59 @@ type queuedRetrySender struct {
...
 	}
+
+	qrs.consumerSender = &retrySender{
+		traceAttribute:     traceAttr,
+		cfg:                rCfg,
+		nextSender:         nextSender,
+		stopCh:             retryStopCh,
+		logger:             sampledLogger,
+		onTemporaryFailure: qrs.onTemporaryFailure,
+	}
```
bogdandrutu, line 108: why this change?

bogdandrutu (approved):

exporter/exporterhelper/queued_retry_inmemory.go
bogdandrutu: I reverted this change.

bogdandrutu: I learned why :)), reverted my revert.

PR #5784, opened 2022-08-02
Require the storage to be explicitly set for persistent queue

swiatekm-sumo:
**Description:**

The persistent queue needs a storage client to run. Currently, this simply takes the one storage extension defined, and errors if there are multiple available. This is both restrictive and arbitrary, and mostly works because it's currently unusual for users to define multiple storage extensions.

This change forces the users to explicitly set the storage extension in queue config if they want to use the persistent queue. This setting doubles as both choosing the storage and enabling the persistent queue - when `storage: nil` (the default), we use the in-memory queue, otherwise, we use the persistent queue with the set storage extension.

This was originally discussed in #5711, and it's part of that change. As it's a breaking change, I'd like to do it before we enable the persistent queue by default in the build.

**Link to tracking Issue:** Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/10915

**Testing:**
Updated the unit tests.

**Documentation:**
Updated the queue readme.

### DIFF
--- a/exporter/exporterhelper/README.md
+++ b/exporter/exporterhelper/README.md
@@ -36,16 +36,13 @@ The following configuration options can be modified:
 With this build tag set, additional configuration option can be enabled:
 
 - `sending_queue`
-  - `persistent_storage_enabled` (default = false): When set, enables persistence via a file storage extension
+  - `storage` (default = none): When set, enables persistence and uses the component specified as a storage extension for the persistent queue
     (note, `enable_unstable` build tag needs to be enabled first, see below for more details)
 
 The maximum number of batches stored to disk can be controlled using `sending_queue.queue_size` parameter (which,
 similarly as for in-memory buffering, defaults to 5000 batches).
 
-When `persistent_storage_enabled` is set to true, the queue is being buffered to disk using 
-[file storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage).
-If collector instance is killed while having some items in the persistent queue, on restart the items are being picked and
-the exporting is continued.
+When persistent queue is enabled, the batches are being buffered using the provided storage extension - [filestorage] is a popular and safe choice. If the collector instance is killed while having some items in the persistent queue, on restart the items will be be picked and the exporting is continued.
 
 ```
                                                               ┌─Consumer #1─┐
@@ -93,9 +90,9 @@ exporters:
   otlp:
     endpoint: <ENDPOINT>
     sending_queue:
-      persistent_storage_enabled: true
+      storage: file_storage/otc
 extensions:
-  file_storage:
+  file_storage/otc:
     directory: /var/lib/storage/otc
     timeout: 10s
 service:
@@ -112,3 +109,5 @@ service:
       exporters: [otlp]
 
 ```
+
+[filestorage]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage
--- a/exporter/exporterhelper/queued_retry_experimental.go
+++ b/exporter/exporterhelper/queued_retry_experimental.go
@@ -44,8 +44,9 @@ type QueueSettings struct {
 	NumConsumers int `mapstructure:"num_consumers"`
 	// QueueSize is the maximum number of batches allowed in queue at a given time.
 	QueueSize int `mapstructure:"queue_size"`
-	// PersistentStorageEnabled describes whether persistence via a file storage extension is enabled
-	PersistentStorageEnabled bool `mapstructure:"persistent_storage_enabled"`
+	// StorageID if not empty, enables the persistent storage and uses the component specified
+	// as a storage extension for the persistent queue
+	StorageID *config.ComponentID `mapstructure:"storage"`
 }
 
 // NewDefaultQueueSettings returns the default settings for QueueSettings.
@@ -57,8 +58,7 @@ func NewDefaultQueueSettings() QueueSettings {
 		// This is a pretty decent value for production.
 		// User should calculate this from the perspective of how many seconds to buffer in case of a backend outage,
 		// multiply that by the number of requests per seconds.
-		QueueSize:                5000,
-		PersistentStorageEnabled: false,
+		QueueSize: 5000,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


exporter/exporterhelper/queued_retry_experimental.go
```diff
@@ -120,46 +120,54 @@ func newQueuedRetrySender(id config.ComponentID, signal config.DataType, qCfg Qu
...
 	}
 
-	client, err := storageExtension.GetClient(ctx, component.KindExporter, id, string(signal))
+	client, err := extension.GetClient(ctx, component.KindExporter, ownerID, string(signal))
 	if err != nil {
 		return nil, err
 	}
 
-	return &client, err
+	return client, err
```
tigrannajaryan, line 159: 👍

exporter/exporterhelper/queued_retry_experimental_test.go
```diff
@@ -80,21 +80,21 @@ func TestGetRetrySettings(t *testing.T) {
 			expectedError: errNoStorageClient,
 		},
 		{
-			desc:           "obtain default storage extension",
-			numStorages:    1,
-			storageEnabled: true,
-			expectedError:  nil,
+			desc:          "obtain storage extension by name",
+			numStorages:   2,
+			storageID:     "1",
```
tigrannajaryan, line 85: Shouldn't this be?
```suggestion
			storageID:     "file_storage/1",
```
I thought ComponentID always includes the component type name.

exporter/exporterhelper/queued_retry_experimental.go
```diff
@@ -120,46 +120,54 @@ func newQueuedRetrySender(id config.ComponentID, signal config.DataType, qCfg Qu
...
 }
 
-func getStorageClient(ctx context.Context, host component.Host, id config.ComponentID, signal config.DataType) (*storage.Client, error) {
-	var storageExtension storage.Extension
-	for _, ext := range host.GetExtensions() {
-		if se, ok := ext.(storage.Extension); ok {
-			if storageExtension != nil {
-				return nil, errMultipleStorageClients
+func (qCfg *QueueSettings) getStorageExtension(extensions map[config.ComponentID]component.Extension) (storage.Extension, error) {
+	if qCfg.StorageID != nil {
```
bogdandrutu, line 132: Few suggestions:
1. Usually prefer to check for conditions first and return than indenting the code. `if qCfg.StorageID == nil { return nil, nil } `
2. Better to pass a non pointer `ComponentID` and then no need for the func to be on the QueueSettings. You already check for nil in the caller code.

exporter/exporterhelper/README.md
```diff
@@ -36,16 +36,13 @@ The following configuration options can be modified:
...
     (note, `enable_unstable` build tag needs to be enabled first, see below for more details)
 
 The maximum number of batches stored to disk can be controlled using `sending_queue.queue_size` parameter (which,
 similarly as for in-memory buffering, defaults to 5000 batches).
 
-When `persistent_storage_enabled` is set to true, the queue is being buffered to disk using 
-[file storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage).
-If collector instance is killed while having some items in the persistent queue, on restart the items are being picked and
-the exporting is continued.
+When persistent queue is enabled, the batches are being buffered using the provided storage extension - [filestorage] is a popular and safe choice. If the collector instance is killed while having some items in the persistent queue, on restart the items will be be picked and the exporting is continued.
```
TylerHelmuth, line 45: should `[filestorage]` be a link?

exporter/exporterhelper/queued_retry_experimental.go
```diff
@@ -44,8 +44,9 @@ type QueueSettings struct {
 	NumConsumers int `mapstructure:"num_consumers"`
 	// QueueSize is the maximum number of batches allowed in queue at a given time.
 	QueueSize int `mapstructure:"queue_size"`
-	// PersistentStorageEnabled describes whether persistence via a file storage extension is enabled
-	PersistentStorageEnabled bool `mapstructure:"persistent_storage_enabled"`
+	// StorageID if not empty, enables the persistent storage and uses the component specified
+	// as a storage extension for the persistent queue
+	StorageID *config.ComponentID `mapstructure:"storage"`
```
bogdandrutu, line 49: @djaglowski I saw a PR from you related to this, can you ensure we are consistent here? Another idea is to extract this into a "configstorage" similar with "configauth" to ensure everyone uses this in the same way.

djaglowski: Yes, for reference [here is the related PR on contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12871/).

It looks like we have a difference in behavior regarding whether or not storage is automatically used if there is exactly one storage extension. This was discussed [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/10915#issuecomment-1154055913). Basically, there is a desire to have this functionality for easier configuration in a common case and for backwards compatibility.

bogdandrutu: > Basically, there is a desire to have this functionality for easier configuration in a common case and for backwards compatibility.

I don't think a behavior like that is desired for new components, I don't disagree that for backwards compatibility you may do that, but I disagree to do that for new components. The main reason is that someone comes, adds a new extension that implements the storage extension and everything breaks now.

djaglowski: Although I do not want to introduce a breaking change, I think in this case we should consider it in order to keep behavior the same for all components that use storage extensions. To my knowledge, this would only affect the stanza-based and otlpjsonfile receivers, none of which are beyond alpha status.

exporter/exporterhelper/queued_retry_experimental_test.go
swiatekm-sumo: This is just a test parameter, in the actual test code we manually make it into a config.ComponentID struct. I'll change it to `storageName` so it's less confusing.

exporter/exporterhelper/README.md
swiatekm-sumo: It should, I thought it was linked already in this file. Will fix.

exporter/exporterhelper/queued_retry_experimental.go
swiatekm-sumo: Good point, thx. I also noticed I missed a test for the wrong extension type that was in my original PR, re-added while making these changes.

bogdandrutu:
You need a rebase

bogdandrutu (approved):

PR #5787, opened 2022-08-02
Add a new label Skip Contrib Tests that skips contrib tests requirement

bogdandrutu:

### DIFF
--- a/.github/workflows/contrib-tests.yml
+++ b/.github/workflows/contrib-tests.yml
@@ -5,6 +5,8 @@ on:
     tags:
       - v[0-9]+.[0-9]+.[0-9]+.*
   pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    branches: [ main ]
 
 concurrency:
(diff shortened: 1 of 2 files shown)


TylerHelmuth (approved):

.github/workflows/contrib-tests.yml
```diff
@@ -5,6 +5,8 @@ on:
     tags:
       - v[0-9]+.[0-9]+.[0-9]+.*
   pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
codeboten, line 8: Should `ready_for_review` be included here? I'm not sure if there are other `types` that should be included maybe `edited`?

TylerHelmuth: If we add `ready_for_review` we should update the changelog workflow to match.

bogdandrutu: Done for `ready_for_review`

codeboten (approved):

PR #5791, opened 2022-08-02
add go 1.19 to tests

codeboten:
Adding the latest version of go to the tests run by CI.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5793


TylerHelmuth (approved):
Hype!

codeboten:
To pass the tests, the following changes were required:
- run make genpdata
- fix test certificates to address errors caused by the rejection of duplicate extensions in TLS handshakes

client/client.go
```diff
@@ -21,7 +21,7 @@
 // propagated down the pipeline, with the values being produced by
 // authenticators and/or receivers, and consumed by processors and exporters.
 //
-// Producers
+// # Producers
```
tigrannajaryan, line 24: Are changes in this file relevant?

.github/workflows/build-and-test.yml
```diff
@@ -108,7 +108,7 @@ jobs:
   unittest:
     strategy:
       matrix:
-        go-version: [1.18, 1.17]
+        go-version: [1.19, 1.18, 1.17]
```
tigrannajaryan, line 111: Don't we remove N-2 simultaneously we adding the new version N?

tigrannajaryan (commented):
Lot's of seemingly unrelated changes, mostly with whitespace or formatting in comments. Is this intentional?

client/client.go
codeboten: This was all done automatically by running `make fmt`, it was causing lint to fail previously.

.github/workflows/build-and-test.yml
codeboten: I figured it would be easier to add 1.19 and make necessary changes to pass. Then update the 'required' check to 1.19 and 1.18 (removing 1.17) before removing 1.17. I don't have a strong preference though.

bogdandrutu (approved):

PR #5819, opened 2022-08-04
Remove the InstrumentationLibrary to Scope translation, OTLP 0.19 no longer supports it.

bogdandrutu:
Kept a generic "Migrate[Logs|Metrics|Traces]" so that if needed in the future we will not forget to update any place.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5818


bogdandrutu:
Not sure if this should be a "breaking change". Probably to be safe we can say it is.

codeboten (commented):
I think the change itself is fine, one question i have is whether there should be a migration guide section somewhere, or at least an indicator of what OTLP protocols are supported by a collector that's easier to find than looking at the Makefile.

This would be to answer questions from users that may upgrade the collector to find that their data is no longer working because they're sending an outdated version of OTLP.

bogdandrutu:
@codeboten is that not the changelog? Why do we have the changelog then?

codeboten:
> @codeboten is that not the changelog? Why do we have the changelog then?

I guess I was thinking something that is easier to parse ie. the table here: https://github.com/open-telemetry/opentelemetry-operator/#opentelemetry-operator-vs-kubernetes-vs-cert-manager

codeboten (approved):
With the new changelog entry i think this is ok.

Issue #4714, opened 2022-01-19
"immutable" config.[Receiver|Processor|Exporter|Extension] interface


bogdandrutu:
Right now, we do have a "GetID/SetID" on the configuration interface for all our components, this causes the interface to have "mutable" funcs. It would be cleaner from a design perspective if we separate the ID from the config.

A draft PR https://github.com/open-telemetry/opentelemetry-collector/pull/3840 was done, but don't know if I have time to finish this.

rahuldimri:
@bogdandrutu If you want I can have a look into this.

PR #5822, opened 2022-08-04
Use the config pointer as the key for the memorylimiter instances instead of the ID

bogdandrutu:


processor/memorylimiterprocessor/factory.go
```diff
@@ -122,7 +122,7 @@ func (f *factory) getMemoryLimiter(set component.ProcessorCreateSettings, cfg co
 	f.lock.Lock()
 	defer f.lock.Unlock()
 
-	if memLimiter, ok := f.memoryLimiters[cfg.ID()]; ok {
+	if memLimiter, ok := f.memoryLimiters[cfg]; ok {
```
dmitryax, line 125: Why does it need to be changed? just using another wrapping struct as the key, right? cannot see any reasons for that

bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/issues/4714

dmitryax (approved):

PR #5833, opened 2022-08-05
Deprecate processorhelper.New*Processor in favor of New*ProcessorWithCreateSettings

bogdandrutu:
Main motivation is to ensure that Settings and Context are passed (to enable obsreport usage which requires settings to be passed, and to enable future logic of cancelation or other things passed in the context).

The plan is next release to deprecate the new func, and change the current func to match the new signature, then in another release to remove the new func.


processor/processorhelper/traces.go
```diff
@@ -36,14 +36,26 @@ type tracesProcessor struct {
...
-// NewTracesProcessor creates a TracesProcessor that ensure context propagation and the right tags are set.
-// TODO: Add observability metrics support
+// Deprecated: [v0.58.0] use version with NewTracesProcessorWithCreateSettings.
 func NewTracesProcessor(
 	cfg config.Processor,
 	nextConsumer consumer.Traces,
 	tracesFunc ProcessTracesFunc,
 	options ...Option,
 ) (component.TracesProcessor, error) {
+	return NewTracesProcessorWithCreateSettings(context.Background(), component.ProcessorCreateSettings{}, cfg, nextConsumer, tracesFunc, options...)
```
dmitryax, line 46: Are you planning to bring back `NewTracesProcessor` name once the old one is removed?

bogdandrutu: yes. we need intermediate to keep backwards compatibility for 2 versions.

dmitryax (approved):
LGTM. But I would like ask for another review from @open-telemetry/collector-approvers since the change is significant

bogdandrutu:
@dmitryax I don't feel is that significant, since we are making the "helper" funcs to look like the Factory funcs (which are the significant once. But more eyes are always welcome.

bogdandrutu:
Per the approvers from https://github.com/open-telemetry/opentelemetry-collector/pull/5834 I take that this is fine as well

jpkrohling:
I haven't reviewed this code in detail, but the general approach is fine.

PR #5834, opened 2022-08-05
Deprecate exporterhelper.New*Exporter in favor of New*ExporterWithContext

bogdandrutu:
Main motivation is to pass the context to the helper in case any cancelation is needed, and to match the Create[*]Exporter calls.

The plan is next release to deprecate the new func, and change the current func to match the new signature, then in another release to remove the new func.


dmitryax (approved):
Same as for https://github.com/open-telemetry/opentelemetry-collector/pull/5833. LGTM. But I would like ask for another review from @open-telemetry/collector-approvers for broader awareness

mx-psi (approved):

jpkrohling (approved):
I agree with the plan to add context to the functions, taking a few steps to achieve the final desired outcome.

PR #5839, opened 2022-08-05
Breaking Change: Change confmap.Provider to return pointer to Retrieved.

bogdandrutu:
This change makes implementations cleaner, since they can return `nil, err` in case of an error instead of a zero initialized Retrieved.

This is a breaking change, but there are very very few implementation of the Provider, so it should be safe to just break it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - ABI breaking change: `featuregate.Registry.Apply` returns error now.
 - Update minimum go version to 1.18 (#5795)
 - Remove deprecated `Flags` API from pdata (#5814)
+- Change `confmap.Provider` to return pointer to `Retrieved` (#5839)
 
 ### 🚩 Deprecations 🚩
 
--- a/confmap/provider.go
+++ b/confmap/provider.go
@@ -53,7 +53,7 @@ type Provider interface {
 	//
 	// If ctx is cancelled should return immediately with an error.
 	// Should never be called concurrently with itself or with Shutdown.
-	Retrieve(ctx context.Context, uri string, watcher WatcherFunc) (Retrieved, error)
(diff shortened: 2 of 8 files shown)


bogdandrutu:
I need to get @open-telemetry/collector-approvers consensus, since this is a breaking change without deprecation.

mx-psi (approved):
LGTM, I am fine with breaking this in one go

dmitryax (approved):

codeboten (commented):
> This is a breaking change, but there are very very few implementation of the Provider, so it should be safe to just break it.

Is there a way to know this outside of the repository? I agree with this change, however i would prefer we follow the same path we always do for breaking changes.

bogdandrutu:
> Is there a way to know this outside of the repository?

* Contrib passes (so that is a good sign).
* Based on my knowledge only AWS may have an implementation of this for http/s3. If that is the case, the change is very small for them.

> however i would prefer we follow the same path we always do for breaking changes.

* Because `Provider` is an interface it is extremely hard base on what I know to make this fully backwards compatible. We can define a new interface and provide some "Converters" then we need to change lots of other public APIs wherever we expect this Interface to expect the other one, and deprecate there, etc. Because the good names are taken we have to do everything in 3 steps. I think this is definitely overkill based on the very limited usage and the amount of changes needed.

codeboten (approved):
Right, since any changes to the interface would cause any implementor to need to update anyways. I'm ok going forward with this.

mx-psi:
S3 provider is here, I think: open-telemetry/opentelemetry-collector-contrib/pull/13113

bogdandrutu:
Looks like we have 3 approvers, and first provider coming to contrib soon. I will merge this and update contrib so we can have that one implementing the new interface.

junhaoyu-aws:
That's a great work, I will fix my s3provider PR according to this.

PR #5840, opened 2022-08-05
[pdata] Ran make genpdata

TylerHelmuth:
Regenerating pdata with `make genpdata`.  IDK if we want the comment changes, but raising the issue via this PR since the current pdata is out of sync.

### DIFF
--- a/pdata/internal/generated_ptrace.go
+++ b/pdata/internal/generated_ptrace.go
@@ -57,11 +57,10 @@ func (es ResourceSpansSlice) Len() int {
 // At returns the element at the given index.
 //
 // This function is used mostly for iterating over all the values in the slice:
-//
-//	for i := 0; i < es.Len(); i++ {
-//	    e := es.At(i)
-//	    ... // Do something with the element
-//	}
+//   for i := 0; i < es.Len(); i++ {
+//       e := es.At(i)
+//       ... // Do something with the element
+//   }
 func (es ResourceSpansSlice) At(ix int) ResourceSpans {
 	return newResourceSpans((*es.orig)[ix])
 }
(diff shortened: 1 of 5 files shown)


pdata/internal/generated_ptrace.go
```diff
@@ -1049,11 +1035,10 @@ func (es SpanLinkSlice) AppendEmpty() SpanLink {
...
 // Returns the same instance to allow nicer code like:
-//
-//	lessFunc := func(a, b SpanLink) bool {
-//	  return a.Name() < b.Name() // choose any comparison here
-//	}
-//	assert.EqualValues(t, expected.Sort(lessFunc), actual.Sort(lessFunc))
+//   lessFunc := func(a, b SpanLink) bool {
+//     return a.Name() < b.Name() // choose any comparison here
+//   }
+//   assert.EqualValues(t, expected.Sort(lessFunc), actual.Sort(lessFunc))
```
dmitryax, line 1041: Hm.. it just replaces tabs with spaces in the comments. The template uses tabs. I don't think this is an improvement. Maybe is there something changed to how `go fmt` work in your setup?

TylerHelmuth: It's the output when I run `make genpdata` on main, but let me refresh my tools install and see what happens.

bogdandrutu:
@TylerHelmuth you have a golang 1.18 the 1.19 gofmt changes this so the code is formated with 1.19

pdata/internal/generated_ptrace.go
bogdandrutu: Is old go fmt that you have @TylerHelmuth

TylerHelmuth:
gotta fix my Go version

PR #5860, opened 2022-08-08
exporterhelper: bounded memory queue, remove unused onDropItem func

bogdandrutu:


jpkrohling (approved):
Is this function used in the Jaeger implementation? I can see how this can be useful, but if we have no use right now, even after seeing how the original implementation uses it, we might as well just get rid of it.

bogdandrutu:
> Is this function used in the Jaeger implementation? I can see how this can be useful, but if we have no use right now, even after seeing how the original implementation uses it, we might as well just get rid of it.

You can see that the func is anyway called only in `Produce` func, and we do check for the `Produce` return value (returns false when called), so same functionality can be implemented outside of the queue implementation.

Issue #5706, opened 2022-07-19
Restrict characters allowed in expand provider URIs


mx-psi:
During review of #4742 I noticed that we basically allow arbitrary characters on the identifiers used for environment variable expansion (and soon expansion from other kinds of sources). This is too permissive, since it will restrict us from ever supporting recursion `${ENV_${RECURSIVE}}`, URI-like parameters for default (e.g. `${ENV?default=something}`) or other features that we may want to add in the future.

We need to validate identifiers to exclude at least these characters (`$`, `{`, `}`, `?`, `=`). Expanding the set of allowed identifiers is not a breaking change, so we should err on the side of caution and be conservative about what characters we allow.

One possibility would be to adopt the [Unicode identifier and pattern syntax](https://unicode.org/reports/tr31/#R1), which is commonly used in programming languages. We can be even more restrictive if we think any character in that set would potentially be used in the future for special syntax.

mx-psi:
Note: this should not, in principle, apply to URIs from configuration sources, otherwise we can't support things like the `yaml` provider.

mx-psi:
Since right now we only have the `env` provider, we can explore the environment variables case. Here is some info [for Windows](https://stackoverflow.com/a/20635858) (many special characters allowed) and [for Linux](https://stackoverflow.com/a/20635858) (anything goes in principle, the recommendation is to use the [Portable Character Set](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap06.html#tagtcjh_3)).

bogdandrutu:
The use-cases: `${ENV_${RECURSIVE}}` and `${ENV?default=something}` seem very different to me.

Here are some suggestions/questions:
1. I suggest that we do NOT support this `${ENV?default=something}` and only support `${env:ENV?default=something}` and the "non-schema" (not an URI like) support that we offer, is only for backwards compatibility and we offer only `${ENV}` and `$ENV`.
2. I suggest to not add any new feature to the "non-schema" expansion logic, and keep that for backwards compatibility as much as we want, maybe eventually consider to remove it.
[...]

PR #5861, opened 2022-08-08
Enforce scheme name restrictions to all confmap.Provider implementations.

bogdandrutu:
Currently was documented that the format should be compatible with the URI definition (see https://datatracker.ietf.org/doc/html/rfc3986),
but the scheme name restriction was not copied from the RFC, and no tests to enforce. This PR clearly documents the characters allowed in the scheme name and add confmaptest helper func to test for valid scheme names.

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/5706

### DIFF
--- a/confmap/confmaptest/configtest.go
+++ b/confmap/confmaptest/configtest.go
@@ -15,17 +15,46 @@
 package confmaptest // import "go.opentelemetry.io/collector/confmap/confmaptest"
 
 import (
-	"context"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+
+	"gopkg.in/yaml.v2"
 
 	"go.opentelemetry.io/collector/confmap"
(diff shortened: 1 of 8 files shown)


confmap/confmaptest/configtest.go
```diff
@@ -15,17 +15,46 @@
...
+	}
+
+	var rawConf map[string]interface{}
+	if err = yaml.Unmarshal(content, &rawConf); err != nil {
 		return nil, err
 	}
-	return ret.AsConf()
+
+	return confmap.NewFromStringMap(rawConf), nil
+}
```
bogdandrutu, line 42: I had to avoid using `fileprovider.New()` to avoid circular dependencies, since `fileprovider` uses `ValidateProviderScheme`.

mx-psi (approved):
Changes here LGTM. Should we also make `service.NewConfigProvider` fail if it we pass a provider that does not have a valid scheme?

PR #5868, opened 2022-08-08
Deprecate MetricDataPointFlags.String(), no other pdata has this

bogdandrutu:
If offering String() on pdata we should do it from the pdata generator not as ad-hoc for some of the hand written structs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
   - `component.ReceiverFactory.MetricsReceiverStability`
   - `component.ReceiverFactory.LogsReceiverStability`
 - Deprecate `obsreport.ProcessorSettings.Level` and `obsreport.ExporterSettings.Level`, use MetricsLevel from CreateSettings (#5824)
+- Deprecate MetricDataPointFlags.String(), no other pdata has this (#5868)
 - Deprecates `FlagsStruct` in favor of `Flags` (#5842)
   - `MetricDataPointFlagsStruct` -> `MetricDataPointFlags`
   - `NewMetricDataPointFlagsStruct` -> `NewMetricDataPointFlags`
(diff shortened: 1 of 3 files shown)


pdata/internal/metrics.go
```diff
@@ -258,7 +258,7 @@ func (ms MetricDataPointFlags) SetNoRecordedValue(b bool) {
 	}
 }
 
-// String returns the string representation of the MetricDataPointFlags.
+// Deprecated: [v0.58.0] will be soon removed, no String() on any other pdata.
```
dmitryax, line 261: Why need to remove this? What the alternative way to get human readable representation of the flags?

bogdandrutu: Why do we need that?

bogdandrutu: Also implementation is completely wrong, since the flags are masks not values. which means you need to mask with that and determine if present, so the human readable is a combination of all properties not one.

dmitryax: Sure, but other flags can be added separated by commas once we have more than one flag?

dmitryax: I'm ok with deprecating/removing it until there is a real use case

dmitryax (approved):

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@
   - `component.ReceiverFactory.MetricsReceiverStability`
   - `component.ReceiverFactory.LogsReceiverStability`
 - Deprecate `obsreport.ProcessorSettings.Level` and `obsreport.ExporterSettings.Level`, use MetricsLevel from CreateSettings (#5824)
+- Deprecate MetricDataPointFlags.String(), no other pdata has this (#5868)
```
dmitryax, line 40: Currently it sound like we don't `String` methods in pdata at all

```suggestion
- Deprecate MetricDataPointFlags.String(), no other pdata flags have this method (#5868)
```

PR #5877, opened 2022-08-09
Upgrade to go1.19 all workflows except testing on go1.18

bogdandrutu:

### DIFF
--- a/confmap/confmaptest/configtest.go
+++ b/confmap/confmaptest/configtest.go
@@ -16,7 +16,7 @@ package confmaptest // import "go.opentelemetry.io/collector/confmap/confmaptest
 
 import (
 	"fmt"
-	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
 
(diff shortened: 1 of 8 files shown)


confmap/confmaptest/configtest.go
```diff
@@ -28,7 +28,7 @@ import (
 // LoadConf loads a confmap.Conf from file, and does NOT validate the configuration.
 func LoadConf(fileName string) (*confmap.Conf, error) {
 	// Clean the path before using it.
-	content, err := ioutil.ReadFile(filepath.Clean(fileName))
+	content, err := os.ReadFile(filepath.Clean(fileName))
```
bogdandrutu, line 31: I had to do only a small change here.

codeboten (approved):

confmap/confmaptest/configtest.go
codeboten: i'm surprised this was the only use of ioutil that needed to be changed.

PR #5895, opened 2022-08-10
[chore] change config tests to unmarshal only the config for that component

bogdandrutu:
The main motivation for this is to allow components (including tests) to not depend on the "service" configuration and be reusable without the otelcol service.

If we agree on this we should do the same in contrib components.


mx-psi (approved):
LGTM, I think this makes sense. It will be a ton of work in contrib though 😅

PR #5896, opened 2022-08-11
Fix bug in setting the correct collector state after a config change event.

### LINKED ISSUE #5830: collector state won't be set to Running state after watch is triggered
**Describe the bug**
After configuration is updated and `watch` is triggered, the collector health state will be stuck at `Starting` even tho the collector has started completely.

**Steps to reproduce**
- Run the collector
- trigger `watch` operation in config provider
- check collector state by calling `GetState`
- state won't be in `Running` and will always be in `Starting`

collector state to be `Running`

**What did you see instead?**
Collector state is always `Starting`

**What version did you use?**
latest. https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L140

**What config did you use?**
any config

**Environment**
OS: mac catalina

kovrus:
**Description:**
After configuration is updated the collector state will be stuck at `Starting` even tho the collector has started completely.

**Link to tracking Issue:**
Fixes #5830.

**Testing:**
Unit tests.


service/collector.go
```diff
@@ -200,7 +200,7 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	if err = col.service.Start(ctx); err != nil {
 		return err
 	}
-
+	col.setCollectorState(Running)
```
bogdandrutu, line 203: I think you should remove the call from line 136 correct?

bogdandrutu (approved):

bogdandrutu:
Looks like tests are not passing https://github.com/open-telemetry/opentelemetry-collector/runs/7811347371?check_suite_focus=true#step:6:3292

kovrus:
> Looks like tests are not passing https://github.com/open-telemetry/opentelemetry-collector/runs/7811347371?check_suite_focus=true#step:6:3292

[Closing collector properly](https://github.com/open-telemetry/opentelemetry-collector/pull/5896/files#diff-39cb0a393ab609b58e27ab97198f7fdd885a5369e325d25b32879b22d6e19019R371) in tests via collector api, seems to fix the race condition.

PR #5913, opened 2022-08-15
Move log statements to service when possible and cleaner

bogdandrutu:

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -17,6 +17,7 @@ package service // import "go.opentelemetry.io/collector/service"
 import (
 	"context"
 	"fmt"
+	"runtime"
 
 	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -116,13 +122,18 @@ func (srv *service) Start(ctx context.Context) error {
 		return fmt.Errorf("cannot start pipelines: %w", err)
 	}
 
-	return srv.host.extensions.NotifyPipelineReady()
+	err := srv.host.extensions.NotifyPipelineReady()
+	srv.telemetrySettings.Logger.Info("Everything is ready. Begin running and processing data.")
```
dmitryax, line 126: Should we log it if `err != nil`?

bogdandrutu: Done.

dmitryax (approved):

Issue #5920, opened 2022-08-16
Add Size() method to pdata Structures


cpheps:
Currently all the pdata structures have a way to count the number of objects in them:

- plogs have `LogRecordCount()`
- pmetrics have `MetricCount()` and `DataPointCount()`
- ptraces have `SpanCount()`

It would be useful in some cases to have access to the size of the pdata structure. The primary use case would be so components could measure their data throughput and record it via custom telemetry metrics. I have done some benchmarks and this operation can be expensive at scale. Due to this I would not want to include the data throughput metrics in the current set of built in component metrics.

The implementation is very straight forward for all pdata types, here is an example using metrics:

```go
func (md Metrics) Size() int {
	return md.orig.Size()
}
```

This will give the protobuf size of the object, counting all the sub objects. I would want to add a warning to the function comment indicating possible performance impact at scale.

bogdandrutu:
This operation is pretty expensive, and we try to discourage the use of this in metrics, since the amount of CPU needed to calculate this is way too much compare with the usefulness. 

This may change when we change to lazy unmarshalling, but I would suggest to retain that until then. As an alternative for the moment you still have `Sizer` as an interface that the encoders may implement (the protobuf does that), see https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/ptrace/encoding.go#L33 and https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/ptrace/pb.go#L23

cpheps:
@bogdandrutu I wasn't aware of the `Sizer` interface. I think that more or less is the same thing I'm proposing. I will look at that. Closing this issue.

cpheps:
@bogdandrutu One note about this it's not very friendly to interact with. Seems like the only way to get a `Sizer` is to cast a `Marshaler` as a `Sizer`. Shown [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L230) in the batch processor. To someone importing the collector as a dependency it isn't obvious this is a safe operation without digging into the code. Seems like there should be a function analogous to `NewProtoMarshaler` and `NewProtoUnmarshaler` called `NewProtoSizer` that instantiates something that directly implements the `Sizer` interface.

bogdandrutu:
Someone closed to soon this https://github.com/open-telemetry/opentelemetry-collector/issues/3842 

[...]

PR #5929, opened 2022-08-17
Added MarshalerSizer Interface to facilitate creation of Sizer implementations

cpheps:
**Description:**
Proposing adding a new `MarshalerSizer` interface to `ptrace`, `plog`, and `pmetric` packages and alter `NewProtoMarshaler` to return this interface. This allows easy creation of a `Sizer` implementation. Currently the only way to create a sizer for the `pdata` packages is to call `NewProtoMarshaler` function then cast to a `Sizer`. This is not intuitive from an API standpoint that it should be safe to cast a `Marshaler` as a `Sizer`. 

The `MarshalerSizer` interface is a combination of the `Marshaler` and `Sizer` interfaces.

If this is merged I will create a follow up PR to update the Batch Processor to use this rather than casting from a `Marshaler`.

**Link to tracking Issue:** #5920

**Testing:** Updated tests to use the `MarshalSizer` instead of casting.

### DIFF
--- a/pdata/plog/encoding.go
+++ b/pdata/plog/encoding.go
@@ -14,6 +14,12 @@
 
 package plog // import "go.opentelemetry.io/collector/pdata/plog"
 
+// MarshalSizer is the interface that groups the basic Marshal and Size methods
+type MarshalSizer interface {
+	Marshaler
+	Sizer
+}
+
 // Marshaler marshals pdata.Logs into bytes.
 type Marshaler interface {
 	// MarshalLogs the given pdata.Logs into bytes.
--- a/pdata/plog/encoding.go
+++ b/pdata/plog/encoding.go
@@ -14,8 +14,8 @@
 
 package plog // import "go.opentelemetry.io/collector/pdata/plog"
 
-// MarshalSizer is the interface that groups the basic Marshal and Size methods
-type MarshalSizer interface {
+// MarshalerSizer is the interface that groups the basic Marshal and Size methods
+type MarshalerSizer interface {
 	Marshaler
 	Sizer
 }
--- a/pdata/plog/pb.go
+++ b/pdata/plog/pb.go
@@ -19,13 +19,9 @@ import (
 	otlplogs "go.opentelemetry.io/collector/pdata/internal/data/protogen/logs/v1"
 )
 
-// NewProtoMarshaler returns a Marshaler. Marshals to OTLP binary protobuf bytes.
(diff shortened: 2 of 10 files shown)


pdata/plog/pb.go
```diff
@@ -24,7 +24,11 @@ func NewProtoMarshaler() Marshaler {
 	return newPbMarshaler()
 }
 
-// TODO(#3842): Figure out how we want to represent/return *Sizers.
+// NewProtoSizer returns a Sizer. Calculates the size of a marshaled Logs.
+func NewProtoSizer() Sizer {
```
bogdandrutu, line 28: Not sure if this is the best option, we may consider alternative as having an interface `MarshalerSizer` similar with [`WriterSeeker`](https://pkg.go.dev/io#WriteSeeker)?

cpheps: I actually like that better. Would you expect the `NewProtoMarshaler` to return a `MarshalerSizer` or still a `Marshaler`?

bogdandrutu: NewProtoMarshaler will return the most descriptive interface that implements, so `MarshalerSizer`

cpheps: @bogdandrutu Updated to use a `MarshlerSizer` interface. Let me know if you think there needs to be additional tweaks to this implementation.

bogdandrutu (commented):
I expect to remove some of the conversions to Sizer from batch processor, do I miss something and that is not possible?

pdata/plog/encoding.go
```diff
@@ -14,6 +14,12 @@
 
 package plog // import "go.opentelemetry.io/collector/pdata/plog"
 
+// MarshalSizer is the interface that groups the basic Marshal and Size methods
+type MarshalSizer interface {
```
bogdandrutu, line 18: The convention name is `MarshalerSizer`?

cpheps: Fixing it.

cpheps:
> I expect to remove some of the conversions to Sizer from batch processor, do I miss something and that is not possible?

@bogdandrutu I wasn't sure if that should be part of this PR or not. This changes makes the current logic still valid. I was going to do a follow up PR with the batch processor changes but I can group them in here if you'd rather just have a single PR.

bogdandrutu:
@cpheps sounds good, but still a question in the issue if you can respond

bogdandrutu (approved):

Issue #5312, opened 2022-05-04
`http` receiver drops JSON requests with extra keys 


pichlermarc:
**Describe the bug**
JSON requests with extra keys are dropped by the `http` receiver. This can cause older collectors to not work with newer exporters in case new optional fields are added (as suggested in https://github.com/open-telemetry/oteps/pull/201) or renamed. 

The suggested workaround for the renamed `instrumentationLibraryMetrics` was to double-publish the same data (see [here](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L72-L76)). However, since to requests with extra keys (in this case `scopeMetrics`) are dropped, it is not possible to use newer JSON exporters that implement this spec with collector versions older than `0.48`. 

To prevent this in the future, it might make sense to allow extra keys in the request payload.

**Steps to reproduce**

1. Create a file `request-with-extra-key.json` and add the following:

```json
{
    "resourceMetrics": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "service.name",
                        "value": {
                            "stringValue": "basic-metric-service"
                        }
                    }
                ],
                "droppedAttributesCount": 0
            },
            "scopeMetrics": [
                {
                    "metrics": [
                        {
                            "description": "Example of a Counter",
                            "name": "requests",
                            "sum": {
                                "aggregationTemporality": 2,
                                "dataPoints": [
                                    {
                                        "asDouble": 1,
                                        "attributes": [],
                                        "startTimeUnixNano": 1651578664709780000,
                                        "timeUnixNano": 1651578666712823600
                                    }
                                ],
                                "isMonotonic": true
                            },
                            "unit": "1"
                        }
                    ],
                    "scope": {
                        "name": "example-exporter-collector",
                        "version": "",
                        "extraKey": "value"
                    }
                }
            ]
        }
    ]
}
```

This JSON has an key `extraKey` in the `scope`. Simulating what would happen if an optional field would be added in a future version of [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto).

2. Send the request to a collector
`curl -X POST -H "Content-Type: application/json" -d @request-with-extra-key.json -i localhost:4318/v1/metrics`

**What did you expect to see?**
A `200` response and metrics being processed and `extraKey` in scope being dropped (and possibly a log message).

**What did you see instead?**
A `400 Bad Request` response with
[...]

PR #5930, opened 2022-08-17
Fix reading scope attributes for trace JSON, remove duplicate code

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-collector/issues/5312 

### DIFF
--- a/pdata/internal/json/scope.go
+++ b/pdata/internal/json/scope.go
@@ -1,0 +1,42 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package json // import "go.opentelemetry.io/collector/pdata/internal/json"
+
+import (
+	jsoniter "github.com/json-iterator/go"
+
+	otlpcommon "go.opentelemetry.io/collector/pdata/internal/data/protogen/common/v1"
+)
+
+func ReadScope(iter *jsoniter.Iterator, scope *otlpcommon.InstrumentationScope) {
+	iter.ReadObjectCB(func(iter *jsoniter.Iterator, f string) bool {
+		switch f {
+		case "name":
+			scope.Name = iter.ReadString()
+		case "version":
+			scope.Version = iter.ReadString()
+		case "attributes":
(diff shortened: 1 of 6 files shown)


pdata/internal/json/scope.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	jsoniter "github.com/json-iterator/go"
+
+	otlpcommon "go.opentelemetry.io/collector/pdata/internal/data/protogen/common/v1"
+)
+
+func ReadScope(iter *jsoniter.Iterator, scope *otlpcommon.InstrumentationScope) {
+	iter.ReadObjectCB(func(iter *jsoniter.Iterator, f string) bool {
+		switch f {
+		case "name":
```
TylerHelmuth, line 26: Any chance there are constants from the spec/proto we can use to represent these cases?

pdata/internal/json/scope_test.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"testing"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/stretchr/testify/assert"
+
+	otlpcommon "go.opentelemetry.io/collector/pdata/internal/data/protogen/common/v1"
+)
+
+func TestReadScope(t *testing.T) {
```
TylerHelmuth, line 26: Can you add a test case that covers the situation where a field is skipped?

pdata/internal/json/scope.go
bogdandrutu: They are tags, but that needs reflection package and is expensive. Not anything else that I know, but probably we can do a proto plugin to extract these values.

TylerHelmuth: Looking through pdata seems like we use this strategy a lot, so changing,  if possible, would probably be better for a later PR.

pdata/internal/json/scope_test.go
bogdandrutu: Done.

pdata/internal/json/scope.go
bogdandrutu: Definitely will not do this in the current PR, also needs a lot of changes :)

pdata/internal/json/scope.go
```diff
@@ -0,0 +1,42 @@
...
+
+package json // import "go.opentelemetry.io/collector/pdata/internal/json"
+
+import (
+	jsoniter "github.com/json-iterator/go"
+
+	otlpcommon "go.opentelemetry.io/collector/pdata/internal/data/protogen/common/v1"
+)
+
+func ReadScope(iter *jsoniter.Iterator, scope *otlpcommon.InstrumentationScope) {
```
tigrannajaryan, line 23: Do we need to also call ReadScope for logs? I don't see it in this PR.

tigrannajaryan (approved):

pdata/internal/json/scope.go
bogdandrutu: Logs was not yet converted to the new `jsoniter`, still uses `jsonpb`.

TylerHelmuth (approved):

PR #5931, opened 2022-08-17
otlpjson/trace: skip unknown fields instead of error

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 ### 🧰 Bug fixes 🧰
 
 - Fix reading scope attributes for trace JSON, remove duplicate code. (#5930)
+- otlpjson/trace: skip unknown fields instead of error. (#5931)
 - Fix bug in setting the correct collector state after a configuration change event. (#5830)
 - Fix json trace unmarshalling for numbers (#5924):
   - Accept both string and number for int32/uint32.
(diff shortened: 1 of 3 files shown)


TylerHelmuth:
Does this change allow the collector to receiver any json payload intending to be an OTEL span without any errors?  So if a user sent `{"key": "value"}`, which matches none of our fields, would the collector skip everything and then return a 200?

bogdandrutu:
Yes, but there is no way otherwise to support adding of new fields correct?

TylerHelmuth:
Ya I guess we have to push that responsibility off to the receivers, otherwise the collector isn't forward/backwards compatible.

TylerHelmuth:
Maybe there could be an option when starting the collector to allow strict enforcement versus skipping unknown values? Some users might want the enforcement.

bogdandrutu:
@TylerHelmuth let's think about that, would you mind opening an issue for that? I do believe that by default we should not fail.

TylerHelmuth (approved):

TylerHelmuth:
> @TylerHelmuth let's think about that, would you mind opening an issue for that? I do believe that by default we should not fail.

Done: https://github.com/open-telemetry/opentelemetry-collector/issues/5935

dmitryax:
Looks like link is the description is wrong

bogdandrutu:
@dmitryax good catch, PTAL

dmitryax (approved):

PR #5933, opened 2022-08-17
Deprecate for good the FlagsStruct types

bogdandrutu:
They cannot be removed, unless we apply an exception, since the Deprecation was not correctly marked in golang/godoc because of the missing ":".

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/5443


bogdandrutu:
@open-telemetry/collector-approvers please let me know +1 (for going with removing the funcs now) or -1 (to deprecate now and remove later).

Aneurysm9 (approved):
-1 remove them later

Are there linters that could help avoid a mistake like this in the future?

bogdandrutu:
@Aneurysm9 not that I know of.

dmitryax (approved):

bogdandrutu:
Looks like people prefer mostly to deprecate now, merge this PR as is then.

PR #5936, opened 2022-08-19
Change pdata generated types to use type definition instead of aliases

bogdandrutu:
Change the way how objects are generated in `pdata`. Previously the pdata objects were generated in the `internal` package and aliases were created in the public packageas.

This PR changes this, by creating only "wrappers" object inside the internal package, and the public type is a type def of the internal type:

```golang
package internal

type LogRecord struct {
	orig *otlplogs.LogRecord
}

func GetOrigLogRecord(ms LogRecord) *otlplogs.LogRecord {
	return ms.orig
}

func NewLogRecord(orig *otlplogs.LogRecord) LogRecord {
	return LogRecord{orig: orig}
}
```

```golang
package plog

type LogRecord internal.LogRecord
```

With this approach, we still do not allow users access to the internal origin, which allows us flexibility to change to other representation (something like lazy proto, etc), but improves documentation of the pdata packages, see [current documentation](https://pkg.go.dev/go.opentelemetry.io/collector/pdata@v0.58.0/plog).


pdata/plog/generated_logs.go
```diff
@@ -30,28 +32,28 @@ import (
...
-}
+type ResourceLogsSlice internal.ResourceLogsSlice
 
 func newResourceLogsSlice(orig *[]*otlplogs.ResourceLogs) ResourceLogsSlice {
-	return ResourceLogsSlice{orig}
+	return ResourceLogsSlice(internal.NewResourceLogsSlice(orig))
+}
+
+func (ms ResourceLogsSlice) getOrig() *[]*otlplogs.ResourceLogs {
+	return internal.GetOrigResourceLogsSlice(internal.ResourceLogsSlice(ms))
```
dmitryax, line 42: Nice trick! 👍

pdata/ptrace/traces.go
```diff
@@ -12,67 +12,49 @@
...
+func (ms Traces) Clone() Traces {
 	cloneTd := NewTraces()
-	td.ResourceSpans().CopyTo(cloneTd.ResourceSpans())
+	ms.ResourceSpans().CopyTo(cloneTd.ResourceSpans())
 	return cloneTd
 }
 
 // SpanCount calculates the total number of spans.
-func (td Traces) SpanCount() int {
+func (ms Traces) SpanCount() int {
```
dmitryax, line 55: Why receiver var name is changed to `ms`?

pdata/plog/logs.go
```diff
@@ -12,66 +12,49 @@
...
-	*dest.orig = *ld.orig
-	*ld.orig = otlpcollectorlog.ExportLogsServiceRequest{}
+func (ms Logs) MoveTo(dest Logs) {
+	*dest.getOrig() = *ms.getOrig()
+	*ms.getOrig() = otlpcollectorlog.ExportLogsServiceRequest{}
 }
 
 // Clone returns a copy of Logs.
-func (ld Logs) Clone() Logs {
+func (ms Logs) Clone() Logs {
```
dmitryax, line 48: Same Q here. Is it an accidental rename?

dmitryax (commented):
Just couple of nits. Otherwise LGTM.

This is a great solution! I believe the only downside is that now we have some computational overhead spent of type conversions, right?

pdata/ptrace/traces.go
bogdandrutu: Because all generated things are "ms", so was easier to copy paste helper funcs :))

pdata/plog/logs.go
bogdandrutu: Same answer.

bogdandrutu:
> This is a great solution! I believe the only downside is that now we have some computational overhead spent of type conversions, right?

I believe the compiler should be smart enough to not have to do this :)

dmitryax (approved):
LGTM

PR #5937, opened 2022-08-19
[processor/batch] Removed casts to Sizer interface in batch processor

cpheps:
**Description:** As a follow up to #5929 removed cast to `Sizer` interface in batch processor as this is no longer required with the new `MarshalerSizer` that is returned from `NewProtoMarshaler`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@
 
 ### 💡 Enhancements 💡
 
-- Updated Batch Processor to remove unnecessary cast to `Sizer` interface (#5937)
 - Added `MarshalerSizer` interface to `ptrace`, `plog`, and `pmetric` packages. `NewProtoMarshaler` now returns a `MarshalerSizer` (#5929)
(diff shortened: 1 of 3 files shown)


TylerHelmuth (approved):
LGTM.  probably don't need a changelog entry since there isn't any user impact from that change.

cpheps:
> probably don't need a changelog entry since there isn't any user impact from that change.
@TylerHelmuth I can remove it. I think someone has to tag this PR so CI doesn't fail on not having one.

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 
 ### 💡 Enhancements 💡
 
+- Updated Batch Processor to remove unnecessary cast to `Sizer` interface (#5937)
```
bogdandrutu, line 36: Remove this, no need will skip the changelog.

bogdandrutu (approved):

CHANGELOG.md
cpheps: Done

Issue #4197, opened 2021-10-13
Support exponential histograms in metric pipelines


jmacd:
**Is your feature request related to a problem? Please describe.**

OTLP v0.11 supports an exponential histogram data point. The problem is, many exporters and processors will not initially know what to do with this data point, so some kind of configurable default behavior may be desired. The options are:

1. Reject the new data type if not handled, signal errors
2. Convert the new data type to fixed-boundary explicit histogram data points
3. Convert the new data type to adjustable-boundary explicit histogram data points

Of these, (3) is considered not useful because any destination that supports variable-boundary histograms should accept the new exponential histogram data point quickly.

**Describe the solution you'd like**
Add some kind of support for metrics pipelines to correctly perform (1) or (2) above out-of-the-box, so that users don't silently drop the new data point. For example, if exponential histograms were converted to a small number of fixed boundaries automatically, users would at least see their data. If exponential histograms become errors that the user is monitoring, that is acceptable too.

**Describe alternatives you've considered**
Option (3) was the only alternative. The code in this PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/1353 shows how we could implement option (3), but I doubt anyone wants an auto-scaling explicit boundary conversion automatically.

PR #5938, opened 2022-08-19
Use the github.com/lightstep/go-expohisto exponential histogram mapping functions

jmacd:
**Description:**
This updates the logging exporter to use the an external library of mapping functions which corrects the inclusivity for OTel specification v1.13.0 (i.e., with upper-inclusive bucket boundaries). The external library was originally pending for inclusion in the OTel-Go SDK and may be adopted in that repository eventually.

This addresses boundary conditions more correctly; with this PR the logging exporter is able to correctly print the upper boundary of the last valid bucket. Extreme values will print as "OVERFLOW" or "UNDERFLOW" in cases where this logic is unable to produce a correct string representation.

This leaves a TODO allowing more work to format subnormal values in certain corner cases. 

**Link to tracking Issue:**

Part of #4197

**Testing:** New tests introduced.


jmacd:
This is blocked, we need an otel-go release. I'll set this as a draft.

jmacd:
Note that the corresponding change of exponential histogram boundary inclusivity has been released in specification v0.13, so at this point the collector is printing boundaries incorrectly.

I have published Lightstep's production-tested exponential histogram data structure at https://github.com/lightstep/go-expohisto for use here and in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12951.

jpkrohling:
What are the chances of having the code from https://github.com/lightstep/go-expohisto being part of otel-go?

internal/testdata/metric.go
```diff
@@ -218,19 +218,25 @@ func initExponentialHistogramMetric(hm pmetric.Metric) {
 	hdp0.SetZeroCount(1)
 	hdp0.SetScale(1)
 
-	// positive index 1 and 2 are values sqrt(2), 2 at scale 1
+	// at scale 1,
+	// postitive bucket index 0 is (1, sqrt(2)] (not represented)
+	// postitive bucket index 1 is (sqrt(2), 2]
+	// postitive bucket index 2 is (2, 2*sqrt(2)]
```
codeboten, line 224:
```suggestion
	// positive bucket index 0 is (1, sqrt(2)] (not represented)
	// positive bucket index 1 is (sqrt(2), 2]
	// positive bucket index 2 is (2, 2*sqrt(2)]
```

internal/testdata/metric.go
```diff
@@ -218,19 +218,25 @@ func initExponentialHistogramMetric(hm pmetric.Metric) {
...
+	// at scale 1
+	// negative bucket index 0 is -(1, sqrt(2)]
+	// negative bucket index -1 is -(sqrt(2)/2, 1]
 	hdp0.Negative().SetOffset(-1)
 	hdp0.Negative().BucketCounts().FromRaw([]uint64{1, 1})
 
 	// The above will print:
-	// Bucket (-1.414214, -1.000000], Count: 1
-	// Bucket (-1.000000, -0.707107], Count: 1
+	// Bucket [-1.414214, -1.000000), Count: 1
```
codeboten, line 235: Should these match the changes in [metrics_with_all_types.out](https://github.com/open-telemetry/opentelemetry-collector/pull/5938/files#diff-fcf207003d9dec15c485d510f4526edfb00f5c915c7251188e73eb7927e6f14a)?

jmacd:
With respect to the question about taking a dependency on the go-expohisto reference implementation:
It's well documented that I briefly succeeded at placing this code into the otel-go metrics SDK. That code was withdrawn in the alpha release, and I feel it's not my fight at this point. I welcome the otel-go repository to use the go-expohisto implementation by reference or by copy. 

Note that the otel-collector-contrib statsd receiver has taken a go-expohisto dependency already, and that the library exposes an interface that would not strictly be needed in the context of an otel-go SDK. The statsd receiver wants to count sampled metric events, which means supporting a histogram Increment() function for counts greater than 1, which means we're asking otel-go to maintain something on behalf of the collector, essentially, if we move it out of the lightstep repo.

Note that this dependency, which is now limited to the logging exporter, is smaller than that dependency. The mapping functions are smaller than the data structure itself. 

As for the change here, and the fact that this code is complicated. The fact is, Prometheus forced OTel to change their histogram boundaries for reasons that they hold dear. We presented a ton of evidence that their request leads to more complicated implementations, but we in the end decided to do what they wanted. As a result, printing boundaries of exponential histograms **correctly** became substantially harder. If we want to print these absolutely correctly according to the specification, we need this code somewhere.

Note also that nothing about being an SDK requires printing boundaries correctly, so how to print boundaries is not covered in the SDK specification, so the code here is somewhat opinionated. I would encourage us to accept this PR and let otel-go adopt the go-expohisto dependency at its own pace.

internal/testdata/metric.go
jmacd: Yes.  Sorry, I at first changed the code to use %f like the explicit bucket histogram does, but after reading the tests for a while I decided it was worse to use %f and changed back to %.6f. Will update.

jack-berg (approved):
Just spot checking for correctness. Looks ok to me.

jmacd:
Currently, the logging exporter prints exponential histograms incorrectly according to the specification. I would like it fixed.

Issue #5882, opened 2022-08-10
Collector metrics names reported with OpenCensus and OpenTelemetry are inconsistent


dmitryax:
Currently Collector uses OpenCensus SDK to report its own metrics. Migration to OpenTelemetry metrics SDK started with a feature gate `telemetry.useOtelForInternalMetrics`. There is a significant difference in the produced metrics.

OpenCensus metrics utilize the prometheus exporter with the default metrics namespace set to `otelcol`. It means that all the reported metrics get the `otelcol` prefix, and we have metrics reported as `otelcol_exporter_enqueue_failed_log_records`, `otelcol_process_uptime`, etc.

OpenTelemetry metrics SDK doesn't have a capability to specify the namespace on all the metrics. So the metrics would be `exporter_enqueue_failed_log_records`, `process_uptime`, etc.

Before proceeding with the migration process, we need to decide what metric names would look like long term. In particular, do we want to keep `otelcol` prefix or remove it?

1. If we want to keep the prefix, we should probably stop using the namespace configuration on the OpenCensus Prometheus exporter and make sure all the metrics have the `otelcol` prefix set explicitly. This should be a no-op change for end users of the default collector's metrics.

2. If we want to remove the prefix, it will be a significant breaking change for current OTel Collector users. We should probably run it through a deprecation process using another feature gate with an extended (up to 1 year?) duration.

dashpole:
The "short_name" attribute is intended to be used for the prometheus namespace, but is currently a work-in-progress:

* https://github.com/open-telemetry/opentelemetry-specification/pull/2702
* https://github.com/open-telemetry/opentelemetry-specification/pull/2703

Also, I'm working on a migration OTel for OC -> OTel to help mitigate or prevent breaking changes when migrating:

* https://github.com/open-telemetry/oteps/pull/210

dmitryax:
@dashpole given that we currently expose Collector's metrics in Prometheus format only, do you think it makes sense to set `otelcol` as prefix on metrics instead of the namespace for now until we have an option to use Scope `short_name`? At least it'll bring the exposed metrics to a consistent state from Prometheus standpoint and will allow us to migrate to Scope `short_name` without breaking users with `telemetry.useOtelForInternalMetrics` enabled?

bogdandrutu:
Just one small clarification:

[...]

PR #5939, opened 2022-08-21
[obsreport] Replace Prometheus namespace with metrics prefix

dmitryax:
"otelcol" part of the exposed metrics that is set as Prometheus namespace is replaced the same explicit metrics prefix. This change brings names consistency between metrics that can be reported with enabled or disabled "telemetry.useOtelForInternalMetrics" feature gate.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/5882

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### 🛑 Breaking changes 🛑
 
+- "otelcol" part of the exposed metrics that is set as Prometheus namespace is replaced the same explicit metrics prefix (#5939)
+  - This is potentially a breaking change for custom distributions that expose additional OpenCensus metrics. The
+    "otelcol/" prefix must be explicitly added to those metrics to get the same names as before this change.
 - Remove deprecated fields/funcs from `service` (#5907)
   - Remove `ConfigProviderSettings.Location`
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### 🛑 Breaking changes 🛑
 
+- "otelcol" part of the exposed metrics that is set as Prometheus namespace is replaced the same explicit metrics prefix (#5939)
```
codeboten, line 7: To be clear to most users this will not be a breaking change as the prometheus metric exposed will be the same. 

```suggestion
- Update `otelcol_` part of the exposed Prometheus metrics to be set via a metric prefix instead of a namespace (#5939)
```

codeboten (commented):
Just one comment, otherwise this looks good. Hurray for consistency :)

paivagustavo:
This can also be accomplished by wrapping the `prometheus.Registry` before registering the otel-go prom exporter's Collector, which would not break any additional metric:
https://github.com/open-telemetry/opentelemetry-collector/pull/6222/files#diff-52da5566fa270d6a253f7a301b66f4a1dbda7c2c42f3781f8357efe7f08dff10R255

bogdandrutu:
#6223 closes this

PR #5943, opened 2022-08-22
Add sub command to generate sample configuration for the collector

neelayu:
**Description:** 
Adding a feature - Create a subcommand that generates a sample configuration for the collector. It has support for UI based options(like `npm init`), or via flags.

**Link to tracking Issue:** #5654

**Testing:** Unit Tests

**Documentation:** Need suggestions

Waiting for suggestions to improve the code more and opinions on how we can improve the documentation.

Sample workflow:
```console
otelcorecol git:(config-sub) ✗ go run . config --help
Print sample configuration

Usage:
  otelcorecol config [flags]

Flags:
  -f, --config-file string      Name of the config file (default "otelcol.config.yaml")
  -d, --dry-run                 Dry run
  -e, --exporter stringArray    Exporters to be configured
  -x, --extension stringArray   Extensions to be configured
  -h, --help                    help for config
  -l, --pipeline stringArray    Specify the pipelines to be configured
  -p, --processor stringArray   Processors to be configured
  -r, --receiver stringArray    Receivers to be configured
```

If no flags are provided, use prompts
It uses `github.com/AlecAivazis/survey` go library.

Screenshot 2022-08-22 at 7 35 18 PM

With flags
Screenshot 2022-08-22 at 7 37 33 PM

Output(only some parts for relevance)
```yaml
exporters:
  logging:
    # LogLevel defines log level of the logging exporter; options are debug, info, warn, error.
    loglevel: 0
    # SamplingInitial defines how many samples are initially logged during each second.
    sampling_initial: 2
    # SamplingThereafter defines the sampling rate after the initial samples are logged.
    sampling_thereafter: 500
  otlp:
     ...

 ...
service:
  ..
```
As you can see, comments are also added for better understanding. This is achieved by `github.com/open-telemetry/opentelemetry-collector-contrib/cmd/configschema`

Suggestions are welcome to improve this PR.
Thanks!


svrnm:
Hey @neelayu, thanks for creating this. That's much more than I expected when I raised #5654!

I did not review the code, leaving that to the @open-telemetry/collector-approvers, but I tried out the functionality:

I like the prompts! Two issues: 

If I skip all prompts I get the following YAML:
```yaml
exporters: {}
processors: {}
receivers: {}
extensions: {}
service:
  # Specify extensions for the collector
  extensions: []
  pipelines: {}
```
Is it possible to not have `{}` for an empty set? 

If I select all/some attributes it fails with

```
2022/08/23 12:03:16 open ../../go.mod: no such file or directory
```

This seems to be a hard-coded file path, when I copy my compiled collector binary into a folder like "./bin/bin" within the forked repository things work. You may need to verify that the final collector works independent of the existence of these files.

To mention that: I like that if I only provide the `-f <filename>` I get an empty file and not the "missing" prompts. This is really helpful if I just need an empty config that I might like to edit in a second step.

Regarding the copied comments: On the one hand I like them, on the other hand they can make a config extremely hard to read for a new-comer, e.g. `otelcol config -p batch -e otlp -d` creates a config with 106 lines. Maybe you can provide another flag that enables/disables to fill out the config with all the attributes + comments?

neelayu:
@svrnm Thanks for going through this PR and giving it a try.

> If I skip all prompts I get the following YAML:
>  yaml code
> Is it possible to not have {} for an empty set?

Yes. This is something which I want an opinion on. I'd like to add validations to check if the inputs were provided or not.

> If I select all/some attributes it fails with

```2022/08/23 12:03:16 open ../../go.mod: no such file or directory```
Correct. I have to fix this!

> on the other hand they can make a config extremely hard to read for a new-comer,
They do make the configuration file pretty large. As you mentioned, are these options better?
```
otelcol config -p batch -e otlp --no-comments
```
Will produce all fields with no comments.
```
otelcol config -p batch -e otlp --no-attrs
```
Will produce only the configuration file with key headers?
```yaml
exporters:
  otlp:
     # For configuration params, refer https://pkg.go.dev/go.opentelemetry.io/collector/exporter/otlpexporter

```
As you can see, there are no attributes, but a single line comment indicating where the info on the configuration can be found. I believe pkg.go.dev will do the job.

I think its difficult to get only "mandatory" or "minimal" configuration fields from the struct. So its either all or none.

svrnm:
> @svrnm Thanks for going through this PR and giving it a try.
> 
> > If I skip all prompts I get the following YAML:
> > yaml code
> > Is it possible to not have {} for an empty set?
> 
> Yes. This is something which I want an opinion on. I'd like to add validations to check if the inputs were provided or not.

I personally think it's ok to leave them empty.


> > on the other hand they can make a config extremely hard to read for a new-comer,
> > They do make the configuration file pretty large. As you mentioned, are these options better?
> 
> ```
> otelcol config -p batch -e otlp --no-comments
> ```
> 
> Will produce all fields with no comments.
> 
> ```
> otelcol config -p batch -e otlp --no-attrs
> ```
> 
> Will produce only the configuration file with key headers?
> ```yaml
> exporters:
>   otlp:
>      # For configuration params, refer https://pkg.go.dev/go.opentelemetry.io/collector/exporter/otlpexporter
> ```

I like that, teaches people also where they can look for details!

I am only wondering what should be the default (full details, only attributes, no comments&no attributes)?

> As you can see, there are no attributes, but a single line comment indicating where the info on the configuration can be found. I believe pkg.go.dev will do the job.
> 
> I think its difficult to get only "mandatory" or "minimal" configuration fields from the struct. So its either all or none.

That's an interesting one! What about something like:

```
otelcol config -p batch -e otlp,endpoint=https://mybackend/,entpoint.headers.apiKey=asdf,tls.insecure=true --no-attrs
```

gives you some of those attributes set (and skips the rest)?

neelayu:
Hey @svrnm I have made a few changes. I didnt really get time this week to work on it. 
Re:

> That's an interesting one! What about something like:

otelcol config -p batch -e otlp,endpoint=https://mybackend/,entpoint.headers.apiKey=asdf,tls.insecure=true --no-attrs
> gives you some of those attributes set (and skips the rest)?

I believe this is as good as knowing what params are present. So the user might as well use that and create a config file.

svrnm:
@neelayu thank you, I am looking forward to have this feature. I am eventually just a consumer of that, as an end-user but also to simplify the Getting Started docs for the collector over at opentelemetry.io :)

service/configinit_test.go
```diff
@@ -0,0 +1,275 @@
...
+	dir := t.TempDir()
+	file := fmt.Sprintf("%v/%v", dir, "otelcol_temp_config.yaml")
+	factories := components()
+	cmd := NewConfigurationInitCommand(CollectorSettings{Factories: factories})
+	cmd.SetArgs([]string{"-p", "batch", "-r", "otlp", "-m", "./..", "-f", file})
+	_ = cmd.Execute()
+
+	_, err := os.Stat(file)
+	if err != nil {
+		assert.Fail(t, "File %v was expected to be created", file)
```
jpkrohling, line 146: Why not just an `assert.NoError(t, err, "file ... was expected ...")` ?

service/configinit_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+	donec := make(chan struct{})
+	go func() {
+		defer close(donec)
+		procedure(t, c, state)
+	}()
+
+	test(c)
+
+	// Close the slave end of the pty, and read the remaining bytes from the master end.
```
jpkrohling, line 165:
```suggestion
	// Close the client end of the pty, and read the remaining bytes from the provider's end.
```

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+		Filter:  filterFunc,
+	}
+
+	promptPipelines := &survey.MultiSelect{
+		Message: "Select pipelines to enable",
+		Options: []string{"metrics", "traces", "logs"},
+	}
+
+	promptWithoutComm := &survey.Confirm{
+		Message: "Without commented configuration fields?",
```
jpkrohling, line 242: Do we need this? I think the UX can be simplified by leaving comments in the generated config file and skipping this question.

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+		Default: false,
+	}
+
+	promptDryRun := &survey.Confirm{
```
jpkrohling, line 246: How about we leave this only as a flag?

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+		Message: "Dry run(Default: N)",
+		Default: false,
+	}
+
+	promptFile := &survey.Input{
+		Message: "Name of the config file",
```
jpkrohling, line 252: Will the default config file name appear somewhere? In the previous prompt, you added "Default: N" manually, perhaps you should add the default here as well?

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+		Default: defaultConfigFileName,
+	}
+
+	_ = survey.AskOne(promptExp, &exportersSelected, survey.WithStdio(stdio.In, stdio.Out, stdio.Err))
```
jpkrohling, line 256: Why are we ignoring potential errors in this section?

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package service // import "go.opentelemetry.io/collector/service"
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"log"
```
jpkrohling, line 21: Any reasons to use stdlib here instead of zap.Logger?

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+	"github.com/spf13/cobra"
+	"golang.org/x/mod/modfile"
+	"gopkg.in/yaml.v3"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+)
+
+const (
+	commandTemplate = `Print sample configuration.
```
jpkrohling, line 39:
```suggestion
	commandTemplate = `Generate a sample configuration file.
```

service/configinit.go
```diff
@@ -0,0 +1,530 @@
...
+func NewConfigurationInitCommand(set CollectorSettings) *cobra.Command {
+	cfgCmd = &configCommand{
+		exiter: func(args ...interface{}) error {
+			log.Fatal(args...)
+			return nil
+		},
+	}
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Print sample configuration",
```
jpkrohling, line 79:
```suggestion
		Short: "Generate a sample configuration file",
```

jpkrohling (commented):
I left a few comments, and while I think this feature would be useful, I think the current implementation might need some more attention. I ran the simplest command possible, `$ go run ./ config -e otlp`, and I got a file that is full of options that are not needed while still not being very functional, given that the collector won't start due to missing values for required properties (endpoint, for instance).

Instead of having one command for all possible components, how about having an interface for components to implement? This way, component developers can decide what and how to ask users, potentially even running the validation in the process.

Concretely, OTLP Exporter would only ask for the "endpoint" property and ask users if they want to specify values for optional properties.

service/configinit_test.go
neelayu: Yes. Thats much better

service/configinit.go
neelayu: I have removed this option.

service/configinit.go
neelayu: To keep the flow simple, I have kept this as a flag and as a prompt. I am showing the prompt based on the condition
```go
if len(os.Args) <= 2 {
  showPrompt()
}
```
As you can see, keeping this as only flag would add complicated branches in my opinion.

service/configinit.go
neelayu: Yes the default file name appears in the prompt.

service/configinit.go
neelayu: I think since we only want to show error and warning logs, stdlib should be okay?

neelayu:
I have refactored the code. The basefactory essentially adds a SampleConfig() method. The authors are expected to return a configuration for the component. This is more or less what Telegraf agent does in its approach. Since I didnt want to introduce breaking changes across the contrib repo as well, I have added it as an option. Let me know if this approach works. @jpkrohling

service/configinit.go
jpkrohling: Between a more complicated code branching and better UX, I prefer better UX ;-)

service/configinit.go
jpkrohling: Consistency is better here. Unless there's a strong reason to use stdlib, use the logging library we use elsewhere in the code base.

neelayu: Got it. I am now using zap.Logger

service/configinit.go
neelayu: I was going to handle the error. Thanks for pointing it out.

service/configinit.go
neelayu: It turns out, the conditional statement is relatively less complicated. So made this change. Thanks

svrnm:
I am still a big fan of this and keen to see this happening, is there anything I can do to help this progressing? ty

mx-psi:
@jpkrohling is there anything needed here?

Aneurysm9 (commented):
Something about having all of the `*.sample.yaml` files added here doesn't feel right to me.  It seems like the structure of these files is dictated by the structure of the types they represent.  The addition here is documentary comments and sample values.  I wonder whether this could be accomplished with well-structured godoc comments that can be accessed with the reflection API.  Doing so would avoid repeating information for types that are commonly embedded in other types and also minimize the risk of changes to the meaning, interpretation, or documentation of a field not being correctly propagated to all sample configurations where it appears.

neelayu:
@Aneurysm9 I did not add `*sample.yaml` in my first few drafts. You can check the approach which I used here https://github.com/open-telemetry/opentelemetry-collector/commit/044642b3e1114ed3e32611bc89c65673778ecea7

In that I used the configschema API from contrib repo to fetch the fields and doc info. The limitation of that approach is that we can't obtain sensible defaults for fields which are mandatory.(Eg endpoint)

component/component.go
```diff
@@ -164,15 +164,24 @@ type Factory interface {
...
 	unexportedFactoryFunc()
 }
 
 type baseFactory struct {
-	cfgType config.Type
+	cfgType      config.Type
+	sampleConfig string
 }
 
 func (baseFactory) unexportedFactoryFunc() {}
```
jpkrohling, line 180: I believe this won't cause backward compatibility issues because we are forcing implementations to embed the base factory, right?

service/configinit.go
```diff
@@ -0,0 +1,546 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
+	"go.opentelemetry.io/collector/service/internal/components"
+)
+
+const (
+	sampleConfigHeader = `Opentelemetry Collector sample configuration. Double check the pipeline configuration
```
jpkrohling, line 39:
```suggestion
	sampleConfigHeader = `OpenTelemetry Collector sample configuration. Double-check the pipeline configuration
```

service/configinit.go
```diff
@@ -0,0 +1,546 @@
+// Copyright The OpenTelemetry Authors
```
jpkrohling, line 1: Not quite sure I agree this is the best package for this code.

service/configinit.go
```diff
@@ -0,0 +1,546 @@
...
+	DryRun         bool
+}
+
+var defaultConfigFileName = "otelcol.config.yaml"
+
+var cfgCmd *configCommand
+var log *zap.Logger
+var opts = new(optionsSelected)
+
+func init() {
```
jpkrohling, line 74: I don't see a strong reason to use an init function here. Perhaps name it differently and invoke it where the logger needs to be set?

service/configinit.go
```diff
@@ -0,0 +1,546 @@
...
+			if (len(os.Args) == 3 && cmd.Flags().Lookup("dry-run").Changed) || len(os.Args) <= 2 {
+				err := showPrompt(cfgCmd.available, opts, cfgCmd.stdio)
+				// err could be interrupt or error occurred due to failed write to the answer object,
+				// in either case, we can't proceed further
+				if err != nil {
+					return cfgCmd.exiter(err)
+				}
+			}
+
+			exportersNode, err := getExportersNode(set.Factories, opts.Exporters)
```
jpkrohling, line 108: Could you split the business from the plumbing? Plumbing would be setting up the logger, validating the user input, returning errors, printing the outcome, and so on. Assembling the list of components and their sample config would be the business.

jpkrohling (commented):
I like where this is going, but I believe that:

1. this needs a bit better code organization, as there's just too much happening on the configinit.go file
2. if the sample YAML files are auto-generated, there should be a couple of make targets to generate the files on demand and CI checks to ensure that those files are not changed (or are changed when the base type changes). I believe a header stating that those files are auto-generated would also be warranted

neelayu:
Thanks @jpkrohling for the comments. I just want to ensure that this approach of sample config yaml makes sense. One other approach I can think of creating a web equivalent config creation which was discussed in SIG meeting in CLI. That is for every field we ask the user to input the value, the docstring will act as help text. 
Along with it, we can have an option of `--full` which will create the configuration with no user input, the one which I created in my first commit. 
Let me know which approach we want to take. 

If we stick to sample yaml approach,  I am aware that the yaml files will have to be created via make targets. However, that would add too much code to review in this PR. So I can take it up in a separate task. Although there is one caveat here, the files will be auto-generated, but they would be bulky with many fields and large comment sections. So the onus lies on the developer(author of the component) to select(or not) which ones are relevant for a sample config.

jpkrohling:
Sorry, I need to catch-up with the collector's current state before giving any meaningful input here. Perhaps @codeboten knows what's the state around this part, especially regarding the UI that was shown some weeks ago to help with the collector's config?

Issue #273, opened 2019-08-19
Watch config file for changes


jpkrohling:
Viper has support for watching the configuration file for changes and it would be good to have a policy on what to do when that happens. Looking a bit around the code, I can see three main possibilities:

- [ ] Shutdown + init: this seems to be the easiest with the current codebase
- [ ] Add a `OnConfigChange` to all components, such as exporters and receivers, allowing them to be notified of config changes. This is the cleanest and probably best for the future, as it allows each component to hold incoming requests while in-flight requests finish.
- [ ] Nothing: just log a message stating that the config file has new content and that opentelemetry-service expects an orchestrator to kill the process to get the new config file in effect.

For the first and second options, it might also make sense to react on `SIGHUP`.

ccoqueiro:
Hello folks, i have a question from customer: "We need to filter which logs we end to you guys. Right now, we use logspout API to make hot changes on container. With that we exclude logs from being sent to LogDNA. With this new LogObserver approach, we changed logspout for fluentd. Is there a way to hot-update those exclusions in fluentd?  We thought about filtering in the OtelCollector GW that we’ll have for logs. If possible, we can centralize exclusions there, but we need a way to “hot-reload” its config." In this case, this options here would be useful? Thanks

dyladan:
I would propose the following:

1. For now we implement (1) as it is the quickest win. Make the feature opt-in with a configuration like `RestartOnConfigChange` (or opt-out if maintainers prefer?)
2. Implement stub `OnConfigChange` implementation in each component here and in contrib which logs that the component does not support live reloading
3. Handle SIGHUP to call `OnConfigChange`
4. Create a new experimental configuration `LiveReloadOnConfigChange (default false)` (name up for debate) which would prevent the collector from restarting when config changes and instead call `OnConfigChange` handlers
5. (long term) Implement `OnConfigChange` in all packages
6. (probably very long term) make `LiveReloadOnConfigChange` default true?

If this looks agreeable I would like to be assigned and start work on this.

Mario-Hofstaetter:
Another way of implementing this as done by prometheus is via an optional HTTP endpoint command:
https://prometheus.io/docs/prometheus/latest/configuration/configuration/

[...]

PR #5945, opened 2022-08-22
Watch for config file changes in fileprovider

swiatekm-sumo:
**Description:**
Make the file config provider watch for changes to the config file.

This is done via polling the file every second, with an additional grace period of two seconds for actually delivering updates to the collector. We're not using notification mechanisms provided by the OS because they're unreliable or annoyingly complex to make work for some popular use cases - most notably Kubernetes ConfigMaps. As we don't need frequent updates and our watched files are small, polling is both performant and easier to understand than, say, inotify.

This change is more complex than it technically needs to be because it tries to reasonably deal with misbehaving users. For example, it'll quietly accept a user calling `.Shutdown` before `.Close`:

```go
r, err := provider.Retrieve("file:/path/to/config")
provider.Shutdown()
r.Close()
```

and `r.Close()` is even thread-safe. If we're ok with the above resulting in a panic or a deadlock, I can make both the provider and the filewatcher simpler.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/273

**Testing:**
Added additional tests for the provider and separate unit tests for the filewatcher. I also did a couple manual e2e tests of the more interesting edge cases, like network filesystems and deep symlink chains.

### DIFF
--- a/confmap/provider/fileprovider/filewatcher.go
+++ b/confmap/provider/fileprovider/filewatcher.go
@@ -1,0 +1,175 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fileprovider // import "go.opentelemetry.io/collector/confmap/provider/fileprovider"
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"os"
+	"sync"
+	"time"
+
+	"go.uber.org/atomic"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+// pollingFileWatcher watches for changes to a file and delivers change events for each one
+// It keeps track of the state of the watched file in a simple polling loop
+//
+// The implementation intentionally doesn't use notification mechanisms provided by the OS
+// such as inotify. The reasons for this choice are:
+//
+//   - Reliability
+//     Inotify and similar mechanisms don't work in some circumstances, and require hacky workarounds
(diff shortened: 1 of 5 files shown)


TylerHelmuth (commented):
I love this idea, but my gut says we should not make it a default for fileprovide initially.  The current expectation for the file provide (and performance expectations) is that the file is not watched.  It feels appropriate to either hide this capability behind a feature gate or as a different provider/flag.  That way existing customers have to opt-in for now.

swiatekm-sumo:
> I love this idea, but my gut says we should not make it a default for fileprovide initially. The current expectation for the file provide (and performance expectations) is that the file is not watched. It feels appropriate to either hide this capability behind a feature gate or as a different provider/flag. That way existing customers have to opt-in for now.

Sounds reasonable to me. It'd be pretty straightforward to gate this, considering all the new functionality is hidden behind a single conditional right now.

swiatekm-sumo:
In all honesty, after thinking about it more, different usecases may want to have this enabled or disabled in general. In Kubernetes, for example, you typically want to treat Pods as immutable, and don't want them to quietly reload configuration just because you made a change to a ConfigMap. Arguably it should be the users' responsibility to make sure Kubernetes doesn't reload the file inside the Pod, but that's an obscure enough topic that I'm not confident in users not shooting themselves in the foot with it.

I think I'm going to open an issue about this so we can more carefully weigh the pros and cons. This probably applies more generally to configuration reloading for any provider, this one is simply the first in line to have it implemented.

confmap/provider/fileprovider/filewatcher.go
```diff
@@ -0,0 +1,175 @@
...
+
+// calculateFingerprint calculates the fingerprint used to check for file changes
+func calculateFingerprint(content []byte) []byte {
+	hash := sha256.New()
+	hash.Write(content)
+	return hash.Sum(nil)
+}
+
+func metadataEqual(first os.FileInfo, second os.FileInfo) bool {
+	return first.Size() == second.Size() && first.ModTime() == second.ModTime()
```
rapphil, line 174: nit: some file file systems might be mounted with the `norelatime` option, which can lead to funny and unlikely bugs.

swiatekm-sumo: Will it? It was my understanding that `norelatime` only affects access time, whereas I'm checking modification time here.

rapphil: My mistake, you are right. this will only affect the access time.

PR #5946, opened 2022-08-22
Mark custom distribution flags as deprecated

bogdandrutu:

### DIFF
--- a/cmd/builder/internal/command.go
+++ b/cmd/builder/internal/command.go
@@ -87,12 +87,33 @@ build configuration given by the "--config" argument.
 	// the distribution parameters, which we accept as CLI flags as well
 	cmd.Flags().BoolVar(&cfg.SkipCompilation, skipCompilationFlag, false, "Whether builder should only generate go code with no compile of the collector (default false)")
 	cmd.Flags().StringVar(&cfg.Distribution.Name, distributionNameFlag, "otelcol-custom", "The executable name for the OpenTelemetry Collector distribution")
+	if err := cmd.Flags().MarkDeprecated(distributionNameFlag, "use config distribution::name"); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


cmd/builder/internal/command.go
```diff
@@ -87,12 +87,33 @@ build configuration given by the "--config" argument.
...
 	cmd.Flags().StringVar(&cfg.Distribution.Version, distributionVersionFlag, "1.0.0", "The version for the OpenTelemetry Collector distribution")
+	if err := cmd.Flags().MarkDeprecated(distributionVersionFlag, "use config distribution::version"); err != nil {
+		return nil, err
+	}
 	cmd.Flags().StringVar(&cfg.Distribution.OtelColVersion, distributionOtelColVersionFlag, cfg.Distribution.OtelColVersion, "Which version of OpenTelemetry Collector to use as base")
+	if err := cmd.Flags().MarkDeprecated(distributionOtelColVersionFlag, "use config distribution::otelcol_version"); err != nil {
+		return nil, err
+	}
 	cmd.Flags().StringVar(&cfg.Distribution.OutputPath, distributionOutputPathFlag, cfg.Distribution.OutputPath, "Where to write the resulting files")
+	if err := cmd.Flags().MarkDeprecated(distributionOutputPathFlag, "use config distribution::output_path"); err != nil {
```
TylerHelmuth, line 106: I can see a world where `output_path` stays a flag because it is a common flag in other CLIs, but I don't feel that strongly about it.

TylerHelmuth (approved):
LGTM but let one non-critical question.

jpkrohling (approved):

cmd/builder/internal/command.go
bogdandrutu: This is usually `output` not `output_path`, or maybe I am confused about this.

TylerHelmuth: agreed, the name doesn't line up, but the concept of this flag I mean.

Issue #5859, opened 2022-08-08
While working with ImmutableSlices found them very expensive


bogdandrutu:
Majority of the use-cases to work with the new `ImmutableSlices` especially in the metrics side were related to updating some bucket counts or bucket bounds. In all these cases there was a need to update an existing slice, which is very expensive with the current approach.

I propose that we change from "Immutable[Bytes|Uint64|...]Slice" to a "[Bytes|Uint64|...]Slice" which has an api to allow set a value at an index "SetAt(i, val)". 

/cc @dmitryax can you look into this and make sure my proposal makes or does not make sense?

dmitryax:
@bogdandrutu This goes against the decision that we made in https://github.com/open-telemetry/opentelemetry-collector/issues/5263 . It means that we will need to copy the slices when the owner pdata objects are copied. 

If we go this route, I'm not sure what are we gaining by wrapping the slices at all. To fix https://github.com/open-telemetry/opentelemetry-collector/issues/5263 from the state where we were initially, we could've just make defensive copping of the slices in the setters. It seems like wrapping it in plain `[Bytes|Uint64|...]Slice` with `SetAt` just limits the API of the native go slices with no additional benefit . Or am I missing some benefits?

bogdandrutu:
What do you mean, I am a bit confused. The proposal is not to return a golang slice which has the problem that you describe, but to return a `ByteSlice` struct similar to all other `XSlice` structs that we have in the `pdata`

dmitryax:
> The proposal is not to return a golang slice which has the problem that you describe, but to return a ByteSlice struct similar to all other XSlice structs that we have in the pdata

The point of `pdata` is to forbid direct access to underlying proto structs to allow us potential changes in them, right? There are no changes that we can expect in underlying primitive types to abstract them for the same reason. To fix https://github.com/open-telemetry/opentelemetry-collector/issues/5263, we could've just added a defensive copying in the owner objects setters, e.g. `HistogramDataPoint.SetBucketCounts`. The same defensive copying must be done if we wrap the primitive slices in any other structure. Defensive copying in getters can be also considered but is not really required for https://github.com/open-telemetry/opentelemetry-collector/issues/5263.

[...]

PR #5947, opened 2022-08-22
[pdata] Migrate to mutable primitive slices

dmitryax:
Prototype PR applying an approach suggested in https://github.com/open-telemetry/opentelemetry-collector/issues/5859


internal/testdata/metric.go
```diff
@@ -197,12 +197,12 @@ func initHistogramMetric(hm pmetric.Metric) {
 	hdp1.SetSum(15)
 	hdp1.SetMin(15)
 	hdp1.SetMax(15)
-	hdp1.SetBucketCounts(pcommon.NewImmutableUInt64Slice([]uint64{0, 1}))
+	hdp1.SetBucketCounts(pcommon.NewUInt64SliceFromRaw([]uint64{0, 1}))
```
bogdandrutu, line 200: Can we do something like:

```
hdp1.BucketCounts().FromRaw([]uint64{0, 1}))
```

dmitryax: Makes sense. Let me try that

dmitryax: Hmm.. We don't have such pattern in pdata in other places. If we replace `pcommon.Value.SetBytesVal` with `pcommon.Value.BytesVal().FromRaw(...)`, we also need to change the pattern for `Map` and `Slice` types of `pcommon.Value` where we currently use another approach: `pcommon.NewMapFromRaw(...).CopyTo(dest)`.

We can probably replace all the `New...FromRaw` with New...().CopyFromRaw(...)`:

```go
hdp1.BucketCounts().CopyFromRaw([]uint64{0, 1}))
val.MapVal().CopyFromRaw(...)
```



WDYT?

bogdandrutu: I don't care,  `CopyFromRaw` sounds good, and better explain what will happen. On the minus side, FromRaw hides the "copy" so in the future is we find a solution to not copy we can optimize that. But result is the same.

bogdandrutu: But I 100% agree about Map API to be similar with this.

dmitryax: It improves the workflow for the use case with the buckets but makes it worse for Map->bytes use cases. 

E.g. instead of 
m.InsertBytes("key_bytes", pcommon.NewValueBytesFromRaw([]byte{1, 2, 3}))

user will need to have

dmitryax: To reach consistency with `pcommon.Map` and `pcommon.Slice`, we will need to deprecate/remove `Value.<Insert/Upsert>Bytes` methods. 
But it's only 6 such calls in contrib. So maybe that's ok

dmitryax: Oh and `pcommon.Value.SetBytesVal` should be removed as well. Other 6 usages in core

dmitryax: Another downside of making the slices a "complex" type, is that we will need to store a pointer to the slice in https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/generated_immutable_slice.go#L23. So another extra allocation

bogdandrutu:
@dmitryax is this still needed?

dmitryax:
Can be closed in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/5971

PR #5949, opened 2022-08-23
builder: move config module validation during Validate

bogdandrutu:

### DIFF
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
 
@@ -102,7 +103,7 @@ func (c *Config) Validate() error {
 		c.Logger.Info("Using go", zap.String("go-executable", c.Distribution.Go))
 	}
 
-	return nil
+	return multierr.Combine(validateModules(c.Extensions), validateModules(c.Receivers), validateModules(c.Exporters), validateModules(c.Processors))
 }
 
 // ParseModules will parse the Modules entries and populate the missing values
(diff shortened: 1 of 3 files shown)


cmd/builder/internal/builder/config.go
```diff
@@ -128,13 +129,18 @@ func (c *Config) ParseModules() error {
 	return nil
 }
 
-func parseModules(mods []Module) ([]Module, error) {
-	var parsedModules []Module
+func validateModules(mods []Module) error {
```
jpkrohling, line 132: As a new function, this might deserve its own tests.

jpkrohling (commented):
Looks OK and cleaner than before, but what's the motivation?

cmd/builder/internal/builder/config.go
bogdandrutu: I disagree that we should have unittests for private methods. This is an implementation detail to avoid duplicate code.

https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods

We do have unittest to check that `Validate` returns error if any of the Modules (extension, receiver, processor, exporter) are invalid.

tigrannajaryan: My 2 cents: IMO, if the private methods are fully testable via the public API calls then usually there is no need to test the private methods directly. Sometimes it is not possible or practical, and in such cases it is useful to cover the private methods by their own unit tests.

jpkrohling: I find it easier to exercise all the conditions by testing the functions in isolation, but given we have enough coverage here, it's fine to have it this way.

bogdandrutu:
@jpkrohling friendly ping

jpkrohling (approved):

PR #5951, opened 2022-08-23
Remove unnecessary posflag provider load

bogdandrutu:
This is unnecessary because:
1. It happens before flags are loaded, so will initialize things with default values, which already happens during flags initialization. See flags.StringVar for example.
2. The load using posflag provider, as defined right now (with current flags name) produce this config:
```
k := map[string]interface{} {
	"config": ""
	"description": "Custom OpenTelemetry Collector distribution"
	"go": ""
	"module": "go.opentelemetry.io/collector/cmd/builder"
	"name": "otelcol-custom otelcol-version:0.58.0"
	"output-path": "/var/folders/5c/5p_3jmvd6qx0rsvmb9j7c_s00000gn/T/otelcol-distribution1155391158"
	"skip-compilation": false
	"version": "1.0.0"
}
```
As you can see, this will whole be ignored, since all the configs (except the config flag) are sub-configs under "dist" anyway.

### DIFF
--- a/cmd/builder/internal/command.go
+++ b/cmd/builder/internal/command.go
@@ -22,7 +22,6 @@ import (
 	"github.com/knadh/koanf/parsers/yaml"
 	"github.com/knadh/koanf/providers/env"
 	"github.com/knadh/koanf/providers/file"
-	"github.com/knadh/koanf/providers/posflag"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


jpkrohling:
I'll review this one, I think we had bugs around this code which required this to be added.

jpkrohling:
I mean, this code is there to fix a previous bug. I *think* it's a bug we had when CLI options weren't being taken into consideration, but I need to manually test this PR. I'll review this later today or tomorrow.

jpkrohling (approved):

jpkrohling:
I couldn't reproduce the problem I had in mind, so, this looks good. Looking at the history of this file, I see that this line was introduced right during the migration from Viper to Koanf.

Issue #5627, opened 2022-07-04
Revert change to min TLS version


jpkrohling:
As I understood it, 1.0 is used as the default version on the server-side. This is *not* what we want here, on a newer server/library: the recommendation for a few years has been already to default to 1.2 as the minimum version, allowing users to use 1.0 or SSLvX only when absolutely needed.

_Originally posted by @jpkrohling in https://github.com/open-telemetry/opentelemetry-collector/pull/5480#discussion_r913230956_

atoulme:
I believe this is now fixed with #5956, can we close this issue?

PR #5956, opened 2022-08-24
Use TLS 1.2 by default when min_version is not defined

rapphil:
**Description:** 

Use TLS 1.2 by default when min_version is not defined. 

TLS 1.0 and 1.1 are widely deprecated. It does not make sense to offer users with an insecure default.

This PR changes the default TLS min_version for both server and client to 1.2. This is reverting part of a recent change: https://github.com/open-telemetry/opentelemetry-collector/pull/5480

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/5627

**Testing:** Unit tests were modified to account for this case.

**Documentation:** Documentation was updated describing the new default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Replace `processorhelper.New[Traces|Metrics|Logs]Exporter` with `processorhelper.New[Traces|Metrics|Logs]ProcessorWithCreateSettings` definition (#5915)
 - Replace `exporterhelper.New[Traces|Metrics|Logs]Exporter` with `exporterhelper.New[Traces|Metrics|Logs]ExporterWithContext` definition (#5914)
 - Replace ``component.NewExtensionFactory`` with `component.NewExtensionFactoryWithStabilityLevel` definition (#5917)
+- Set TLS 1.2 as default for `min_version` for TLS configuration in clients and servers in case this property is not defined (#5956)
 
 ### 🚩 Deprecations 🚩
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - Replace `processorhelper.New[Traces|Metrics|Logs]Exporter` with `processorhelper.New[Traces|Metrics|Logs]ProcessorWithCreateSettings` definition (#5915)
 - Replace `exporterhelper.New[Traces|Metrics|Logs]Exporter` with `exporterhelper.New[Traces|Metrics|Logs]ExporterWithContext` definition (#5914)
 - Replace ``component.NewExtensionFactory`` with `component.NewExtensionFactoryWithStabilityLevel` definition (#5917)
-- Set TLS 1.2 as default for `min_version` for TLS configuration in clients and servers in case this property is not defined (#5956)
+- Set TLS 1.2 as default for `min_version` for TLS configuration in case this property is not defined (affects servers). (#5956)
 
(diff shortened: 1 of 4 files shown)


jpkrohling (commented):
I'm all for this change, which has been the case in the past, but a recent change made our code relay the default to go's crypto/tls package. Would you be able to dig that issue and link it here? A git blame on the "// Defaults will ..." comment should reveal the PR/issue that changed that.

Aneurysm9 (approved):
Please add a `CHANGELOG.md` entry.

rapphil:
> I'm all for this change, which has been the case in the past, but a recent change made our code relay the default to go's crypto/tls package. Would you be able to dig that issue and link it here? A git blame on the "// Defaults will ..." comment should reveal the PR/issue that changed that.

I added the previous PR in the description.

jpkrohling:
@bogdandrutu, in the previous PR, you seemed to favor delegating the default protocol to the underlying library. I believe setting 1.2 as the min version is the right thing to do here, despite the underlying library setting a lower min version, but it would be good to have your OK on this before moving forward.

bogdandrutu:
Usually I prefer to be consistent with the standard library. Not an expert in security, and not sure what it means to use the deprecated version.

rapphil:
> Usually I prefer to be consistent with the standard library. Not an expert in security, and not sure what it means to use the deprecated version.

I agree with your approach, but I think it is not applicable in this case because promoting security best practices with secure defaults has higher priority. 

I'm not a security expert either but I think we can trust open standards to guide us here:
https://datatracker.ietf.org/doc/rfc8996/

> Removing support for older versions from implementations reduces the
   attack surface, reduces opportunity for misconfiguration, and
   streamlines library and product maintenance.

The RFC has references for possible attacks for which TLS1.0 and 1.1 are vulnerable to.

jpkrohling:
In security circles, people just assume TLS 1.0 and 1.1 are effectively broken. There might be people out there with good reasons to use those versions, so I wouldn't artificially limit its usage, but I would definitely have 1.2 as the default min version.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Replace `processorhelper.New[Traces|Metrics|Logs]Exporter` with `processorhelper.New[Traces|Metrics|Logs]ProcessorWithCreateSettings` definition (#5915)
 - Replace `exporterhelper.New[Traces|Metrics|Logs]Exporter` with `exporterhelper.New[Traces|Metrics|Logs]ExporterWithContext` definition (#5914)
 - Replace ``component.NewExtensionFactory`` with `component.NewExtensionFactoryWithStabilityLevel` definition (#5917)
+- Set TLS 1.2 as default for `min_version` for TLS configuration in clients and servers in case this property is not defined (#5956)
```
bogdandrutu, line 22: The statement is partially incorrect, since for client it was already 1.2

rapphil: It is setting explicitly, no? Before it was implicit.

bogdandrutu: But the action is not a "breaking change" since the result for the user is the same.

rapphil: Good point. I updated the changelog.

rapphil:
@bogdandrutu @jpkrohling 
Is there anything else that I can do to get your approvals?

PR #5959, opened 2022-08-25
Add FromRaw to Flags struct

bogdandrutu:
See usages in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/13617 where we had to check for the sample bit to set the value instead of just setting the entire value.

### DIFF
--- a/pdata/plog/logs.go
+++ b/pdata/plog/logs.go
@@ -248,6 +248,11 @@ func (ms LogRecordFlags) SetIsSampled(b bool) {
 	}
 }
 
+// FromRaw converts from the OTLP uint32 representation into this LogRecordFlags.
+func (ms LogRecordFlags) FromRaw(val uint32) {
+	*ms.getOrig() = val
+}
+
 // AsRaw converts LogRecordFlags to the OTLP uint32 representation.
 func (ms LogRecordFlags) AsRaw() uint32 {
 	return *ms.getOrig()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@
 - Added `service.telemetry.traces.propagators` configuration to set propagators for collector's internal spans. (#5572)
 - Remove unnecessary duplicate code and allocations for reading enums in JSON. (#5928)
 - Add "dist.build_tags" configuration option to support passing go build flags to builder. (#5659)
-- Add an AsRaw func on the flags, lots of places to encode these flags. (#5934)
+- Add an AsRaw and FromRaw for Flags, lots of places to encode these flags. (#5934) (#5959)
 - Change pdata generated types to use type definition instead of aliases. (#5936)
   - Improves documentation, and makes code easier to read/understand.
 
--- a/pdata/plog/logs_test.go
+++ b/pdata/plog/logs_test.go
@@ -157,6 +157,10 @@ func TestLogRecordFlags(t *testing.T) {
 	moveFlags.CopyTo(flags)
 	assert.True(t, flags.IsSampled())
 	assert.True(t, moveFlags.IsSampled())
+
+	flags.FromRaw(127)
+	assert.True(t, flags.IsSampled())
+	assert.Equal(t, uint32(127), flags.AsRaw())
 }
 
 func BenchmarkLogsClone(b *testing.B) {
--- a/pdata/pmetric/metrics.go
+++ b/pdata/pmetric/metrics.go
@@ -219,6 +219,11 @@ func (ms MetricDataPointFlags) SetNoRecordedValue(b bool) {
 	}
 }
 
+// FromRaw converts from the OTLP uint32 representation into this LogRecordFlags.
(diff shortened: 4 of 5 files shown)


dmitryax (approved):

TylerHelmuth (approved):

pdata/plog/logs.go
```diff
@@ -248,6 +248,11 @@ func (ms LogRecordFlags) SetIsSampled(b bool) {
 	}
 }
 
+// FromRaw converts from the OTLP uint32 representation into this LogRecordFlags.
+func (ms LogRecordFlags) FromRaw(val uint32) {
+	*ms.getOrig() = val
+}
```
tigrannajaryan, line 254: Instead of having both IsSampled and FromRaw I think a better way would be to do this:
1. Declare new type LogRecordFlags.
2. Add IsSampled/SetIsSampled methods to LogRecordFlags.
3. Have LogRecord.SetFlags/GetFlags which uses LogRecordFlags.
4. Delete IsSampled/SetIsSampled/FromRaw methods from LogRecordFlags

bogdandrutu: So to set the flag you need:

```golang
flags := logRecord.Flags()
flags.SetIsSampled(true)
logRecord.SetFlags(flags)
```

This is a bit inconsistent with current pdata similar structs. It is similar with TraceID/SpanID in that regard, but these are immutable types. The rule is that if an object is "mutable" (SetIsSampled) then we don't offer a setter to avoid mistakes when updated but not "re-set".

tigrannajaryan: The reason I believe what I suggest is the right approach is separation of concerns. OTLP is not concerned with "sampled" flag directly. OTLP is concerned that there is a byte-sized W3C flags. That's all OTLP needs to know. The inner structure of W3C byte flag is W3C concern. pdata is our abstracted representation of OTLP and must be concerned with the same concepts as OTLP, not more, not less. In the spirit of this I believe pdata API must expose W3C flags as a byte (aliased as a new type for stronger typing), but not individual bits in the W3C flags.

So, perhaps a slightly different variation of what I suggested which is more consistent with my reasoning above would be to have this:
1. Declare new type W3CTraceFlags (an alias of `byte`).
2. Add IsSampled/SetIsSampled methods to W3CTraceFlags.
3. Have LogRecord.SetTraceFlags/GetTraceFlags which uses W3CTraceFlags.
4. Delete IsSampled/SetIsSampled/FromRaw methods from LogRecordFlags.

bogdandrutu: > pdata is our abstracted representation of OTLP and must be concerned with the same concepts as OTLP

This is where I disagree, pdata is our collector data model, which is based on "OTLP" but is not "OTLP". It is the collector's data model. The structs in `ptraceotlp` are related to OTLP (protocol), the others are collector's data model, indeed based on `OTLP`.

> So, perhaps a slightly different variation of what I suggested which is more consistent with my reasoning above would be to have this:

Even this has the downside I mentioned, this does not address that, on how to change the "Sampled" flag.

bogdandrutu: An option is to make the Flags immutable and have a "WithIsSampled" which returns a new Flags struct, so then we are consistent that we have setters only for immutable types.

tigrannajaryan: >This is where I disagree, pdata is our collector data model, which is based on "OTLP" but is not "OTLP". It is the collector's data model.

Indeed. If that's they way we see it then we are free to deviate from OTLP's understanding. However, do we? What do we gain by that deviation?

On the contrary, there is value in saying "pdata is an efficient in-memory representation and API to work with OTLP data". I envision it being a separate Otel library, not even necessarily part of Collector. Any Go codebase that needs to work with OTLP data can use such a library. Think for example Schema transformation. This is now developed as a new processor in Collector, however backends also need similar functionality. If pdata becomes an independent Otel library, then Schema transformer can also be an independent Otel library that depends on pdata, but is not in the Collector codebase. Then in the Collector codebase we can have just the schematransform processor which uses both libraries.

bogdandrutu: That is fine, still I believe that we should respect the rule that we "Set" only immutable types, so we cannot have `SetIsSampled`, we need to return a new type to keep the logic that all types that we offer setters are immutable.

tigrannajaryan: I don't think it matters for non-reference types. The rule that types with setters need to be immutable is only important for reference types so that you can't use the mutability to modify the value without explicitly calling the setter. For non-reference types this is impossible, so not an issue.
Anyway, I am fine either way, this is a minor issue, works either way for me.

bogdandrutu: @tigrannajaryan please review https://github.com/open-telemetry/opentelemetry-collector/pull/5970

bogdandrutu:
Closing in favor of:
* https://github.com/open-telemetry/opentelemetry-collector/pull/5999 
* https://github.com/open-telemetry/opentelemetry-collector/pull/5996

PR #5971, opened 2022-08-25
[pdata] Migrate immutable primitive slices to mutable wrappers

dmitryax:
This PR changes API for primitive slice wrappers to allow changing items directly since it appeared to be expensive to update that data as immutable wrappers: it required making two copies to update a slice.

This is a solution alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/5947

Update https://github.com/open-telemetry/opentelemetry-collector/issues/5859

The wrappers for primitive slices doesn't represent atomic objects, they are treated the same way as other "complex" collection wrapper e.g. `pcommon.Slice` or `pcommon.Map`


bogdandrutu:
We need a rebase on this.

pdata/pcommon/generated_primitive_slice.go
```diff
@@ -0,0 +1,197 @@
...
+	if len(orig) == 0 {
+		return nil
+	}
+	copyOrig := make([]byte, len(orig))
+	copy(copyOrig, orig)
+	return copyOrig
+}
+
+// CopyFromRaw copies raw []byte into the slice ByteSlice.
+func (ms ByteSlice) CopyFromRaw(val []byte) {
```
bogdandrutu, line 60: Still not sure about the `Copy` word in this func name.

pdata/pcommon/generated_primitive_slice.go
```diff
@@ -0,0 +1,197 @@
...
+// AsRaw returns a copy of the []byte slice.
+func (ms ByteSlice) AsRaw() []byte {
+	orig := *ms.getOrig()
+	if len(orig) == 0 {
+		return nil
+	}
+	copyOrig := make([]byte, len(orig))
+	copy(copyOrig, orig)
+	return copyOrig
+}
```
bogdandrutu, line 57: AsRaw and CopyFrom both need to copy from a `[]byte`, but code is a bit different, it should not be (in both case len == 0 => nil).

one idea, have this func in internal:

```
func CopySlice[V byte | int64 | float64](arr []V) []V {
	if len(arr) == 0 {
		return nil
	}

	copyArr := make([]V, len(arr))
	for i := range arr {
		copyArr[i] = arr[i]
	}
	return copyArr
}
```

Then 

```
func (ms ByteSlice) AsRaw() []byte {
	return internal.CopySlice(*ms.getOrig())
}

func (ms ByteSlice) FromRaw(val []byte) []byte {
	*ms.getOrig() = internal.CopySlice(val)
}

```

pdata/pcommon/generated_primitive_slice.go
dmitryax: If we change `AsRaw` to copy the data as well, we should make consistent naming. Let's go without `Copy` prefix

pdata/pcommon/generated_primitive_slice.go
dmitryax: Good idea. Thanks. I applied it

pdata/pcommon/generated_primitive_slice.go
```diff
@@ -0,0 +1,209 @@
...
+func NewByteSlice() ByteSlice {
+	orig := []byte(nil)
+	return ByteSlice(internal.NewByteSlice(&orig))
+}
+
+// NewImmutableByteSlice creates a new ByteSlice by copying the provided []byte slice.
+// Deprecated: [0.60.0] Use New{structName}() and {structName}.FromRaw([]byte) instead
+func NewImmutableByteSlice(orig []byte) ByteSlice {
+	return ByteSlice(internal.NewByteSlice(&orig))
+}
```
bogdandrutu, line 46: We need to copy the orig slice correct?

```golang
ret := NewByteSlice()
ret.FromRaw(orig)
return ret
```

**OR** 

```golang
orig := copyByteSlice(orig)
return ByteSlice(internal.NewByteSlice(&orig))
```

pdata/pcommon/generated_primitive_slice.go
```diff
@@ -0,0 +1,209 @@
...
+
+// ByteSlice represents a []byte slice.
+// The instance of ByteSlice can be assigned to multiple objects since it's immutable.
+//
+// Must use NewByteSlice function to create new instances.
+// Important: zero-initialized instance is not valid for use.
+type ByteSlice internal.ByteSlice
+
+// Deprecated: [0.60.0] Use ByteSlice instead.
+type ImmutableByteSlice = ByteSlice
```
bogdandrutu, line 30: would not bother to generate the deprecated funcs/types. lets write them manually.

pdata/pcommon/generated_primitive_slice.go
dmitryax: Updated

pdata/pcommon/generated_primitive_slice.go
dmitryax: done

pdata/pcommon/generated_primitive_slice.go
```diff
@@ -0,0 +1,203 @@
...
+	return ByteSlice(internal.NewByteSlice(&orig))
+}
+
+// NewImmutableByteSlice creates a new ByteSlice by copying the provided []byte slice.
+// Deprecated: [0.60.0] Use New{structName}() and {structName}.FromRaw([]byte) instead
+func NewImmutableByteSlice(from []byte) ByteSlice {
+	orig := copyByteSlice(from)
+	return ByteSlice(internal.NewByteSlice(&orig))
+}
+
```
bogdandrutu, line 45: Same with this, we can manually write them, so next release we don't have to change the generator.

bogdandrutu (approved):
I like it.

pdata/pcommon/generated_primitive_slice.go
dmitryax: done

pdata/internal/cmd/pdatagen/internal/primitive_slice_structs.go
```diff
@@ -0,0 +1,190 @@
...
+	return ${structName}(internal.New${structName}(&orig))
+}
+
+// AsRaw returns a copy of the []${itemType} slice.
+func (ms ${structName}) AsRaw() []${itemType} {
+	return copy${structName}(*ms.getOrig())
+}
+
+// FromRaw copies raw []${itemType} into the slice ${structName}.
+func (ms ${structName}) FromRaw(val []${itemType}) {
```
dmitryax, line 45: @bogdandrutu do you think we should return ms here?

Otherwise current inline usages of `pcommon.NewImmutable...Slice()` becomes pretty clunky

dmitryax: As for me, I'm not sure about that. I don't like returning it and I don't like how it complicates the usages

dmitryax: Let me apply current state of this PR on contrib, and we'll see

dmitryax: It looks pretty good https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14016 I believe we should keep current version

bogdandrutu (approved):
Things to do in a followup PR:
* Copying can be done in-place to save allocations.
* No way of increasing the capacity or length. We can discuss about that API.
* Add CopyTo/MoveTo and remove the copy logic from the parent's CopyTo.

dmitryax:
> Things to do in a followup PR:

Sounds good. I'll follow up on those

Issue #5974, opened 2022-08-26
[pdata] Revisit Map.Insert/Update/Upsert methods


dmitryax:
`pcommon.Map.Insert/Update/Upsert` are the only methods allowing setting the mutable types on the Map. This practice  is generaly discouraged in pdata module since it potentially can lead to setting the same maps/slices across different parts of pdata.

Setting mutable types should be replaced with other methods, e.g. `Map.UpsertEmpty() Value`.

dmitryax:
@bogdandrutu I realized all the Upsert/Insert/Update methods copy the underlying data under the hood. It means that my concern about sharing the same objects in different parts of pdata is invalid. But it means that the only way to put a map or a slice as a child of another map is to generate a map and copy it, which always implies another extra allocation for the whole map.

```go
childMap := NewValueMap()
... fill childMap ...
parentMap.Upsert("key", childMap) // childMap is copied here
```

If we move to `UpsertEmptyMap`, we don't have any extra allocations

```go
childMap := parentMap.UpsertEmptyMap("key")
... fill childMap ...
```

So even if Upsert/Insert/Update do defensive copying, I think we should still deprecate them.

bogdandrutu:
This is done, since we removed or deprecate (last one is Insert) all the funcs.

Issue #5973, opened 2022-08-26
[pdata] Need a clean way to convert pcommon.Value into a Map or Slice


dmitryax:
Currently, `pcommon.Value` can be switched to a different type only between primitive types, e.g. `String` -> `Int` using `Value.SetInt`, but there is no way to convert `Int` into `Slice` for example. It's not a problem when we work with `common.Map`, but it is a problem when we need to convert a `Value` of `LogRecord.Body`. It's still possible to work around with `Value.CopyTo`, but it goes against the recommended pdata workflow when we handle pdata objects without using constructors like NewValuMap.

The suggestion is to introduce new `pcommon.Value` methods: `Value.InitMap() Map` and `Value.InitSlice() Map`. The naming is arguable.

bogdandrutu:
Here is somewhere where we do this kind of "conversion" inside metrics https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pmetric/metrics.go#L136. we need to be consistent with that as well.

dmitryax:
Completed in https://github.com/open-telemetry/opentelemetry-collector/pull/5975

dmitryax:
Completed in https://github.com/open-telemetry/opentelemetry-collector/pull/5975

PR #5975, opened 2022-08-26
[pdata] Deprecate map.Upsert/Update

dmitryax:
`pcommon.Map.Insert/Update/Upsert` methods are the only methods allowing setting the mutable types on the Map. This goes against the established practice of building pdata.

This PR add extra methods to `pcommon.Value` and `pcommon.Map` to replace `pcommon.Map.Insert/Update/Upsert` methods. Deprecation of `pcommon.Map.Insert` is delayed since they are used in many places in contrib. So we want to replaces them first.

Updates: https://github.com/open-telemetry/opentelemetry-collector/issues/5974 and https://github.com/open-telemetry/opentelemetry-collector/issues/5973


pdata/pcommon/common.go
```diff
@@ -772,6 +791,18 @@ func (m Map) Upsert(k string, v Value) {
 	}
 }
 
+// UpsertEmpty inserts or updates an empty value to the map under given key
+// and return the updated/inserted value.
+func (m Map) UpsertEmpty(k string) Value {
```
bogdandrutu, line 796: Let's do an investigation if we have usages for the map API. I kind of believe that it is a good time to maybe switch to `sync.Map` like API, call this `StoreX` and remove all the variants... what do you think? Probably separate effort, but want to investigate.

dmitryax: I was also curious if we really need all of them so I did an initial investigation. Most of the usages are Upsert/Insert that can be easily replaced with Upsert/Store . There are just few cases where Insert or Update are really needed e.g. in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/attraction/attraction.go#L319. But they can be also be replaced with Get+Upsert/Store, it's just one more iteration which is fine probably

pdata/pcommon/common.go
```diff
@@ -298,6 +298,22 @@ func (v Value) SetBytesVal(bv ImmutableByteSlice) {
 	v.getOrig().Value = &otlpcommon.AnyValue_BytesValue{BytesValue: bv.getOrig()}
 }
 
+// InitMap sets value to an empty map and returns it.
+// Calling this function on zero-initialized Value will cause a panic.
+func (v Value) InitMap() Map {
```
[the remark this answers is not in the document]
dmitryax: It's different from setters like `SetStringVal` because it returns the value. So I was thinking that it should be a different name

pdata/pcommon/common.go
bogdandrutu: also attraction will be removed and replace soon by the tql package/processor

pdata/pcommon/common.go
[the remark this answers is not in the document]
bogdandrutu: Not sure about that, I think we should ask few other people... For me I prioritize searchability, so all the "setters" are there, indeed some are a bit different. Maybe we can find a standard package in golang similar that we can get inspiration.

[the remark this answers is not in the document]
dmitryax: As agreed, I'm changing it to `SetEmptyMap`

pdata/pcommon/common.go
```diff
@@ -772,6 +791,16 @@ func (m Map) Upsert(k string, v Value) {
 	}
 }
 
+// UpsertEmpty inserts or updates an empty value to the map under given key.
+func (m Map) UpsertEmpty(k string) {
```
bogdandrutu, line 795: I found few places in contrib where returning a Value here will help. Reason is that for example you have a logic to modify a Value (let's say for LogRecord.Body()) and want to use the same logic here, it is impossible to do so.

See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/telemetryquerylanguage/contexts/internal/tqlcommon/map.go#L29 where I will simply append an `EmptyValue` then will initialize it in the `SetValue` call.

dmitryax: Makes sense. It just fells a bit inconsistent that this method returns a wrapper - Value while all others return the child value. Maybe we call it differently? Just `Upsert` or `UpsertValue`?

bogdandrutu: I prefer `UpsertEmpty` -> returns the empty value, since it is consistent with `UpsertEmtpyMap` -> returns the empty Map inside value

dmitryax: Ok, Let's keep call it `UpsertEmpty`. I'll update the PR and will add test coverage

pdata/pcommon/common.go
```diff
@@ -648,6 +664,7 @@ func (m Map) RemoveIf(f func(string, Value) bool) {
 //
 // Important: this function should not be used if the caller has access to
 // the raw value to avoid an extra allocation.
+// NOTE: The method will be deprecated in 0.60.0. Use Get and CopyTo instead.
```
bogdandrutu, line 667: Get and UpsertEmpty?

bogdandrutu (approved):

pdata/pcommon/common.go
dmitryax: I added more descriptive guidelines to the deprecation comments. PTAL

bogdandrutu (approved):

PR #5978, opened 2022-08-26
[chore] Improve startup error message

### LINKED ISSUE #5648: User experience: starting the collector without config file yields a cryptic message
When running the collector without specifying a config file, the error message that is returned doesn't immediately make it clear to users what the problem is:

```
$ ./bin/otelcorecol_linux_amd64 
Error: invalid map resolver config: no URIs
2022/07/07 16:08:11 collector server run finished with error: invalid map resolver config: no URIs
```

I would prefer to have a validation somewhere before this error, stating that a configuration file is required to run the collector. I believe we even had this a few releases ago.

TylerHelmuth:
**Description:** 
Update's the collector's execute command to return a more usable error message when no configuration is provided.

**Link to tracking Issue:** 
Fixes #5648

**Testing:**
Scenario already covered in existing unit test.


service/command.go
```diff
@@ -35,7 +37,13 @@ func NewCommand(set CollectorSettings) *cobra.Command {
 			}
 			if set.ConfigProvider == nil {
 				var err error
-				cfgSet := newDefaultConfigProviderSettings(getConfigFlag(flagSet))
+
+				configFlags := getConfigFlag(flagSet)
+				if len(configFlags) == 0 {
+					return errors.New("at least one config flag must be provided")
+				}
```
bogdandrutu, line 44: Same needs to be done for windows.

TylerHelmuth: Good catch.  I'll get that added.

TylerHelmuth: Done

bogdandrutu (approved):

PR #5987, opened 2022-08-29
docs(troubleshooting): put newer versions first

astencel-sumo:
**Documentation:**

Reordering config instructions so that newer versions are put first and older versions later.
With time, the users will be less and less interested in configuration for older versions.


tigrannajaryan (approved):
Thanks @astencel-sumo

bogdandrutu (approved):

jpkrohling (approved):

Issue #5988, opened 2022-08-29
golangci-lint reports: times.Total is deprecated


iblancasa:
**Describe the bug**
The usage of the `cpu.Total` function was deprecated in https://github.com/shirou/gopsutil/pull/1325

**Steps to reproduce**
When running `make lint`, the this error is produced:
```sh
$ make lint
golangci-lint run
service/internal/telemetry/process_telemetry.go:162:9: SA1019: times.Total is deprecated: Total returns the total number of seconds in a CPUTimesStat Please do not use this internal function. (staticcheck)
	return times.Total()
	       ^
make: *** [Makefile.Common:41: lint] Error 1
```

**What version did you use?**
Version: `master`. Commit: `1eda0c4e4f6fbd90966359e12b226db01a845223`.

bogdandrutu:
done

PR #5989, opened 2022-08-29
Fix lint error due to the usage of deprecated method

iblancasa:
Signed-off-by: Israel Blancas <iblancasa@gmail.com>

**Description:**
Fixing error reported by `golangci-lint` due to the usage of deprecated function.

**Link to tracking Issue:** #5988

**Testing:**: current tests cover the changes done.


service/internal/telemetry/process_telemetry.go
```diff
@@ -159,7 +159,11 @@ func (pm *processMetrics) updateCPUSeconds() float64 {
 		return 0
 	}
 
-	return times.Total()
+	total := times.User + times.System + times.Idle + times.Nice +
+		times.Iowait + times.Irq + times.Softirq + times.Steal + times.Guest +
+		times.GuestNice
```
dmathieu, line 164: Looking at https://github.com/shirou/gopsutil/pull/1325 and https://www.opsdash.com/blog/cpu-usage-linux.html, I wonder if we shouldn't remove `Guest` and `GuestNice`.

iblancasa: Oh! That is a really good point! Thanks for sharing that article.
Probably `gopsutil` should remove them too, right? Since:

> Guest and Guest Nice: The process (a hypervisor) is running a virtual CPU. These numbers are already included in User and Nice.

dmathieu: They do it for `getAllBusy`. But I suppose they don't want to change the legacy behavior of `Total`.

dmitryax (approved):

service/internal/telemetry/process_telemetry.go
```diff
@@ -159,7 +159,10 @@ func (pm *processMetrics) updateCPUSeconds() float64 {
 		return 0
 	}
 
-	return times.Total()
+	total := times.User + times.System + times.Idle + times.Nice +
+		times.Iowait + times.Irq + times.Softirq + times.Steal
+
+	return total
```
bogdandrutu, line 165: just 
```
return times.User + times.System + times.Idle + times.Nice +
		times.Iowait + times.Irq + times.Softirq + times.Steal
```

bogdandrutu (approved):

service/internal/telemetry/process_telemetry.go
bogdandrutu, line 165:
```suggestion
	return times.User + times.System + times.Idle + times.Nice +
		times.Iowait + times.Irq + times.Softirq + times.Steal
```

PR #5996, opened 2022-08-29
New immutable type for LogRecordFlags structs

bogdandrutu:
This is not a breaking change since the LogRecordFlags were not yet released


bogdandrutu:
@tigrannajaryan @TylerHelmuth from the discussion in https://github.com/open-telemetry/opentelemetry-collector/pull/5972 , I understand that we agreed to go with this instead of having AsRaw/FromRaw on a mutable type. Please correct me if I am wrong.

tigrannajaryan (approved):
I think this is the best approach so far. Thanks for iterating to find it.

Issue #6003, opened 2022-08-31
Persistent Queue Storage Should close client on stop


splunkericl:
**Describe the bug**
Persistent Queue(`persistentContiguousStorage`) retrieves a client from file storage and uses it to persist items. The current file storage uses a bbolt library, which locks the file it loads. This file stays locked until the fileStorageClient `Close` method is called.

However, when `persistentContiguousStorage` is [stopped](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/persistent_storage.go#L203), the client is not closed and the file underneath would stay locked. This would cause collector fail to restart when settings are updated due to file being locked.

**Steps to reproduce**
1. Enable persistent queue and use file_storage extension as the storage client.
2. Run collector
3. Update settings
4. collector attempts to restart. Persistent queue fails to initialize and timeout from file_lock.

**What did you expect to see?**
At step 4, the collector should start successfully

**What did you see instead?**
file has been locked. client is unable to open file

**What version did you use?**
Version: 0.58.0

**Environment**
OS: Mac Monterey
Compiler(if manually compiled): Go 1.17

**Additional context**
Suggestion to fix: when `persistentContiguousStorage` `stop` is [called](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/persistent_storage.go#L203). Call `pcs.client.Close(context.Background())`.

Edited the file locally and verify the collector can start successfully

PR #6005, opened 2022-08-31
Close persistent storage client on stop

splunkericl:
**Description:**
The client passed into `persistentContiguousStorage` was never closed prior. This would caused the underneath file locked forever and collector failing to restart when settings are updated.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/6003

**Testing:** Added unit test on stop

**Documentation:** None needed as fixing bug

### DIFF
--- a/exporter/exporterhelper/internal/persistent_storage_test.go
+++ b/exporter/exporterhelper/internal/persistent_storage_test.go
@@ -32,10 +32,6 @@ import (
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
-var (
-	mockClientCloseCounting = uint64(0)
-)
-
 func createStorageExtension(_ string) storage.Extension {
 	// After having storage moved to core, we could leverage storagetest.NewTestExtension(nil, path)
 	return newMockStorageExtension()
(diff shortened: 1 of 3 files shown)


exporter/exporterhelper/internal/persistent_storage_test.go
```diff
@@ -32,6 +32,10 @@ import (
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
+var (
+	mockClientCloseCounting = uint64(0)
```
bogdandrutu, line 36: Why global instead of a member in the mockStorageClient

splunkericl: If we put it inside `mockStorageClient`, we have to cast the type and extract the value. I just thought that would be uglier.

If you have strong preference to leave it in `mockStorageClient` I can make the change.

splunkericl: On second thought the counter won't work if there are multiple clients created. I will move it inside.

bogdandrutu (approved):

PR #6008, opened 2022-09-01
Change TraceID and SpanID to be type def of an array

bogdandrutu:
In the name of consistency, either change Bytes to AsRaw (see Value/Map/etc.) or remove it and use type def (see *Flags structs, etc.). Decision made by couple of approvers to go with the type def of an array.

Description written by @jpkrohling then modified by @bogdandrutu.


pdata/pcommon/spanid.go
```diff
@@ -20,37 +20,35 @@ import (
 )
 
 // EmptySpanID represents the empty (all zero bytes) SpanID.
-var EmptySpanID = NewSpanID([8]byte{})
+var EmptySpanID = SpanID([8]byte{})
```
dmitryax, line 23: I think SpanID and TraceID now can be treated the same way as byte arrays. In that case we are exposing the individual bytes for possible mutation like `EmptySpanID[1] = 123`. I believe it's applicable to any instance of SpanID and SpanID now

jpkrohling:
Wasn't it like that in the past already? Not that I'm against, but what's the reasoning for this change?

pdata/pcommon/spanid.go
bogdandrutu: >  I believe it's applicable to any instance of SpanID and TraceID now

In Go language, an array is of value type not of reference type. So when the array is assigned to a new variable, then the changes made in the new variable do not affect the original array.

So the returned SpanID/TraceID can indeed be mutated, see also the tests that proves the initial array is not mutated by changing the return value type array (not a reference to the initial array).

The problem with `EmptySpanID` is valid though, so need to change. Is `NewTraceIDEmpty` a good name?

dmitryax: Maybe `NewEmptyTraceID`?

jpkrohling:
> @dmitryax and I decided to go with this.

It would be good to have this reasoning as part of the PR's description or have it as a linked issue.

jpkrohling:
> Not good at memory, but I don't remember to see it like this.

I remember using something like `TraceID([]byte)` (or even `TraceID([8]byte)`) before and having to change it to `NewTraceID([]byte)` because it was now a struct instead of a type def of an array. My concern is going this route without understanding how we got here, only to revert in the future once we hit the same problem as we did in the past.

bogdandrutu:
> It would be good to have this reasoning as part of the PR's description or have it as a linked issue.

Here is `Consistency, either change Bytes to AsRaw (see Value/Map/etc.) or remove it and use type def (see *Flags structs, etc.). @dmitryax and I decided to go with this.`. We identify that `Bytes` is inconsistent with other pdata, and we had to fix that. Then we had the 2 options I described, and picked one.

jpkrohling:
I changed the PR description to have what you mentioned in the comment.

bogdandrutu:
> I remember using something like TraceID([]byte) (or even TraceID([8]byte)) 

That is in otel library, but not sure it was here ever. Here are all the types I remember: 

In (2020-Sep-22) https://pkg.go.dev/go.opentelemetry.io/collector@v0.11.0/consumer/pdata#TraceID it was a byte slice, which had the problem that was reference and mutable, so no good for us.

In (2020-Oct-19) https://pkg.go.dev/go.opentelemetry.io/collector@v0.13.0/consumer/pdata#TraceID it was a typedef to the internal equivalent of the current `data.TraceID`, that had the problem that exposed extra functions available on the `data.TraceID`.

In (2021-Jul-14) https://pkg.go.dev/go.opentelemetry.io/collector/model@v0.30.0/pdata#TraceID it was already a struct like today.

So we never had it like an array which has the property of not being a reference and not mutating the source. I would please ask in the future to help identify the source of you information and give me a link.

pdata/pcommon/spanid.go
bogdandrutu: I would keep it consistent with `NewValueEmpty`?

dmitryax: Ok, let's keep `NewValueEmpty`

jpkrohling:
That's perfect, looks indeed like an evolution instead of a step back.

dmitryax (approved):

PR #6009, opened 2022-09-01
Add skip-get-modules flag to support isolated envs

foadnh:
**Description:** Add `skip-get-modules` flag to the builder.

Isolated build environments like [Bazel](https://bazel.build/) or some CIs might not have access to the public internet.
These environments have different ways to inject dependencies, and there is no need to get them directly from the internet during the builder execution time.
~~Also, if we skip compilation (an existing feature) and downloading modules (this feature), there is no need for Go binary anymore.~~ Move to a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
   - `processorhelper.New[Traces|Metrics|Logs]ProcessorWithCreateSettings`
   - `component.NewExtensionFactoryWithStabilityLevel`
 
+### 💡 Enhancements 💡
+
+- Add `skip-get-modules` builder flag to support isolated environment executions (#6009)
+
 ## v0.59.0 Beta
 
 ### 🛑 Breaking changes 🛑
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -34,6 +34,7 @@ var ErrInvalidGoMod = errors.New("invalid gomod specification for module")
 type Config struct {
 	Logger          *zap.Logger
 	SkipCompilation bool `mapstructure:"-"`
+	SkipGetModules  bool `mapstructure:"-"`
 
 	Distribution Distribution `mapstructure:"dist"`
 	Exporters    []Module     `mapstructure:"exporters"`
--- a/cmd/builder/internal/builder/main.go
+++ b/cmd/builder/internal/builder/main.go
@@ -103,6 +103,11 @@ func Compile(cfg Config) error {
 
 // GetModules retrieves the go modules, updating go.mod and go.sum in the process
 func GetModules(cfg Config) error {
+	if cfg.SkipGetModules {
+		cfg.Logger.Info("Generating source codes only, will not update go.mod and retrieve Go modules.")
+		return nil
+	}
+
 	// #nosec G204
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
   - `processorhelper.New[Traces|Metrics|Logs]ProcessorWithCreateSettings`
   - `component.NewExtensionFactoryWithStabilityLevel`
 
+### 💡 Enhancements 💡
+
+- Add `skip-get-modules` builder flag to support isolated environment executions (#6009)
```
dmathieu, line 14: I think you need an extra new line after your new entry and before the previous release version.

cmd/builder/internal/builder/config.go
```diff
@@ -89,15 +90,17 @@ func NewDefaultConfig() Config {
 // Validate checks whether the current configuration is valid
 func (c *Config) Validate() error {
 	// #nosec G204
-	if _, err := exec.Command(c.Distribution.Go, "env").CombinedOutput(); err != nil {
-		path, err := exec.LookPath("go")
-		if err != nil {
-			return ErrGoNotFound
+	if !(c.SkipCompilation && c.SkipGetModules) {
```
dmathieu, line 93: This change seems unrelated to the original intent of your PR. How about splitting it into another one?

Also, a global negative condition not the most readable way to write this IMHO.
I would write this `if !c.SkipCompilation || !c.SkipGetModules`.

bogdandrutu (commented):
To be honest, I would be more aggressive into looking to remove the `go mod tidy` step and compilation. We should just generate the code from this tool. The other operations are very dependent on the environment.

jpkrohling:
> To be honest, I would be more aggressive into looking to remove the go mod tidy step and compilation

No, this goes against the spirit of the tool. Give it a manifest, and it will give you a collector distribution binary. I'm OK with having flags to fine-tune this and have the builder generate only sources if needed, but the tool should be able to generate binaries.

bogdandrutu:
> No, this goes against the spirit of the tool. Give it a manifest, and it will give you a collector distribution binary. I'm OK with having flags to fine-tune this and have the builder generate only sources if needed, but the tool should be able to generate binaries.

Then we should split the tool into 2, generator and builder :) I feel that the build part is a well covered tooling problem, only the generation is specific to the collector.

jpkrohling:
Why have separate tools when one flag can be used to skip the compilation step?

bogdandrutu:
> Why have separate tools when one flag can be used to skip the compilation step?

The reason for me is that, I believe that the tooling for go mods and go builds are there, and don't trust too much "magic" behind. The skip is fine, but I don't like how complicated logic we have to determine the go binary, etc.

bogdandrutu:
@jpkrohling anyway, having a flag for the moment is fine, I am just saying that we should probably not do this in the tool, but can have that discussion separately.

CHANGELOG.md
foadnh: Thanks, fixed it.

cmd/builder/internal/builder/config.go
foadnh: Good point. I will move it to a separate PR and create a new one once this PR is merged.

jpkrohling (approved):

PR #6026, opened 2022-09-06
Builder skip Go binary validation if not used

foadnh:
**Description:** Builder skip Go binary validation if not used

As a follow-up on (#6009), when the builder is used with `-skip-compilation` and `-skip-get-modules` flags, it only generates distribution go codes. So there is no need for Go binary.
Isolated build environments like [Bazel](https://bazel.build/) or some CIs can skip passing Go binary to the builder.

### DIFF
--- a/cmd/builder/internal/builder/config_test.go
+++ b/cmd/builder/internal/builder/config_test.go
@@ -135,3 +135,14 @@ func TestNewDefaultConfig(t *testing.T) {
 	require.NoError(t, cfg.ParseModules())
 	require.NoError(t, cfg.Validate())
 }
+
+func TestSkipGoValidation(t *testing.T) {
+	cfg := Config{
+		Distribution: Distribution{
+			Go: "invalid/go/binary/path",
+		},
+		SkipCompilation: true,
+		SkipGetModules:  true,
+	}
+	require.NoError(t, cfg.Validate())
+}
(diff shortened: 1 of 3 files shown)


jpkrohling (approved):

jpkrohling:
I believe this change deserves a changelog entry.

dmathieu:
Should a test be added?

jpkrohling:
If you can add a test, it would be awesome.

foadnh:
@jpkrohling changelog entry added.

foadnh:
@dmathieu @jpkrohling I added a test for skipping Go binary validation. There are no tests for actually Go binary validation.

cmd/builder/internal/builder/config_test.go
```diff
@@ -135,3 +135,14 @@ func TestNewDefaultConfig(t *testing.T) {
...
+
+func TestSkipGoValidation(t *testing.T) {
+	cfg := Config{
+		Distribution: Distribution{
+			Go: "invalid/go/binary/path",
+		},
+		SkipCompilation: true,
+		SkipGetModules:  true,
+	}
+	require.NoError(t, cfg.Validate())
```
dmathieu, line 147: Maybe check that `c.Distribution.Go` is nil?

PR #6029, opened 2022-09-06
[confmap] Add mapstructure hook function for unmarshalling confmap.Conf

djaglowski:
This also moves the Unmarshallable interface from `config` to `confmap` in order to avoid
an import cycle. This makes some sense anyways because the interface is concerned
specifically with unmarshalling a confmap.Conf. The config.Unmarshallable interface
is deprecated and changed to an alias of confmap.Unmarshallable.

Rehash of #4879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 
 - Add `skip-get-modules` builder flag to support isolated environment executions (#6009)
   - Skip unnecessary Go binary path validation when the builder is used with `skip-compilation` and `skip-get-modules` flags (#6026)
+- Add mapstructure hook function for confmap.Unmarshaler interface (#6029)
 
 ## v0.59.0 Beta
 
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -212,12 +212,12 @@ func mapKeyStringToMapKeyTextUnmarshalerHookFunc() mapstructure.DecodeHookFuncTy
 // by implementing the Unmarshaler interface.
 func unmarshalerHookFunc(result interface{}) mapstructure.DecodeHookFuncValue {
 	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
-		if _, ok := from.Interface().(map[string]interface{}); !ok {
+		toPtr := to.Addr().Interface()
+		if _, ok := toPtr.(Unmarshaler); !ok {
 			return from.Interface(), nil
 		}
 
-		toPtr := to.Addr().Interface()
-		if _, ok := toPtr.(Unmarshaler); !ok {
+		if _, ok := from.Interface().(map[string]interface{}); !ok {
 			return from.Interface(), nil
 		}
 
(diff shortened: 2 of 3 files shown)


config/common.go
```diff
@@ -26,13 +26,8 @@ type validatable interface {
...
 
-// Unmarshallable defines an optional interface for custom configuration unmarshalling.
-// A configuration struct can implement this interface to override the default unmarshalling.
-type Unmarshallable interface {
-	// Unmarshal is a function that unmarshalls a confmap.Conf into the unmarshable struct in a custom way.
-	// The confmap.Conf for this specific component may be nil or empty if no config available.
-	Unmarshal(component *confmap.Conf) error
-}
+// Deprecated: [v0.60.0] use confmap.Unmarshallable.
+type Unmarshallable confmap.Unmarshallable
```
bogdandrutu, line 30: `type Unmarshallable = confmap.Unmarshallable` to avoid breaking changes.

bogdandrutu (commented):
Let's start with a small PR to deprecate current interface and add it.

confmap/confmap.go
```diff
@@ -210,6 +208,33 @@ func mapKeyStringToMapKeyTextUnmarshalerHookFunc() mapstructure.DecodeHookFuncTy
...
+func unmarshalerHookFunc(result interface{}) mapstructure.DecodeHookFuncValue {
+	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
+		if _, ok := from.Interface().(map[string]interface{}); !ok {
+			return from.Interface(), nil
+		}
+
+		toPtr := to.Addr().Interface()
+		if _, ok := toPtr.(Unmarshaler); !ok {
+			return from.Interface(), nil
+		}
```
bogdandrutu, line 222: This should be the first condition correct? If this does not happen everything else does not matter.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 
 - Add `skip-get-modules` builder flag to support isolated environment executions (#6009)
   - Skip unnecessary Go binary path validation when the builder is used with `skip-compilation` and `skip-get-modules` flags (#6026)
+- Add mapstructure hook function for confmap.Unmarshaler interface (#6029)
```
bogdandrutu, line 38: Enhancements or Breaking change? :)))

I don't think if I should worry about the possible breaking change... I would prefer to not :)

bogdandrutu (approved):

CHANGELOG.md
djaglowski: I think this would only be breaking if a non-component struct already implements this interface. In that case, it would be possible that the behavior changes. However, I cannot find any examples of this in either core or contrib. The only implementation I find is otlpreceiver's Config, which will not have a behavior change because `UnmarshalReceiver` already accounts for component configs.

confmap/confmap.go
djaglowski: Sure, I moved it up

confmap/confmap_test.go
```diff
@@ -228,3 +242,68 @@ func newConfFromFile(t testing.TB, fileName string) map[string]interface{} {
...
+		"next": map[string]interface{}{
+			"string": "make sure this",
+		},
+		"another": "make sure this",
+	})
+	tc := &testConfig{}
+	assert.NoError(t, cfgMap.UnmarshalExact(tc))
+	assert.Equal(t, "make sure this", tc.Another)
+	assert.Equal(t, "make sure this is called", tc.Next.String)
+}
```
bogdandrutu, line 282: add a test where you call also tc.Unmarshal(cfgMap) to make sure that doesn't get infinite loop as well.

bogdandrutu (approved):

PR #6043, opened 2022-09-09
[pdata] Use copy instead of manual copying slices.

dmitryax:
Follow-up to item 1 in https://github.com/open-telemetry/opentelemetry-collector/pull/5971

> Copying can be done in-place to save allocations.


bogdandrutu:
@dmitryax this is great, but I also thing that if cap of dest is less than len of source we should avoid allocation. Maybe another PR.

bogdandrutu (approved):

Issue #6041, opened 2022-09-09
Replace New[Map|Slice]FromRaw with New[Map|Slice]().FromRaw


dmitryax:
After https://github.com/open-telemetry/opentelemetry-collector/pull/5971 is merged, we need to replace New[Map|Slice]FromRaw to New[Map|Slice] and FromRaw in order to have consistent interface

bogdandrutu:
Only to remove deprecate funcs left.

PR #6045, opened 2022-09-09
[pdata] Deprecate New[Slice|Map]FromRaw functions

dmitryax:
Use `New[Slice|Map]().FromRaw()` instead fo consistency with the interface provided by primitive slices.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/6041

### DIFF
--- a/pdata/pcommon/common_test.go
+++ b/pdata/pcommon/common_test.go
@@ -167,12 +167,12 @@ func TestNilOrigSetValue(t *testing.T) {
 	assert.Equal(t, []byte{1, 2, 3}, av.BytesVal().AsRaw())
 
 	av = NewValueEmpty()
-	av.SetEmptyMapVal().UpsertString("k", "v")
-	assert.Equal(t, NewMapFromRaw(map[string]interface{}{"k": "v"}), av.MapVal())
+	av.SetEmptyMapVal().FromRaw(map[string]interface{}{"k": "v"})
+	assert.Equal(t, map[string]interface{}{"k": "v"}, av.MapVal().AsRaw())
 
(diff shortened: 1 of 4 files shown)


dmitryax:
@bogdandrutu  `NewMapFromRaw` is used pretty extensively for tests in contrib to create map instances inline. This PR introduces significant inconvenience. I'm wondering if we should have some additional helper... I'll submit a draft PR in contrib

dmitryax:
Maybe returning map by `FromRaw` is ok given that we already return map by `Sort`

bogdandrutu:
Another option is to also keep the NewFromRaw as a helper for tests?

pdata/pcommon/common_test.go
```diff
@@ -168,11 +168,11 @@ func TestNilOrigSetValue(t *testing.T) {
 
 	av = NewValueEmpty()
 	av.SetEmptyMapVal().UpsertString("k", "v")
-	assert.Equal(t, NewMapFromRaw(map[string]interface{}{"k": "v"}), av.MapVal())
+	assert.Equal(t, map[string]interface{}{"k": "v"}, av.MapVal().AsRaw())
```
bogdandrutu, line 171:
```golang
	av.SetEmptyMapVal().FromRaw(map[string]interface{}{"k": "v"})
	assert.Equal(t, map[string]interface{}{"k": "v"}, av.MapVal().AsRaw())
```

bogdandrutu (approved):

pdata/pcommon/common_test.go
dmitryax: addressed

PR #6049, opened 2022-09-12
[Makefile] Clean up check-contrib target

dmitryax:
Make sure contrib is fully restored locally but ignore it in CI


jpkrohling (commented):
If the changes aren't checked in, wouldn't it just be a matter of doing a "git restore" on the specific files?

dmitryax:
> If the changes aren't checked in, wouldn't it just be a matter of doing a "git restore" on the specific files?

In that case we would need to make sure that there are no changes in contrib that are not checked in on go.mod/sum files.

jpkrohling:
You are right, I think I prefer your current approach.

jpkrohling (approved):

Issue #6022, opened 2022-09-04
Revisit current `ptrace.TraceState` API


bogdandrutu:
- [ ] Move to `pcommon`? Reason to do it is because the ids are already part of the `pcommon` and most likely this will become part of logs and metrics exemplars.
- [ ] Right now is a typedef to string. If we intend to have the type doing parsing, we may look into something like what otel-go has defined for this. We can start by just making this a proper struct and do not expose implicit conversions to/from string as we do right now because of the typedef, but have `AsRaw` and `FromRaw`.

bogdandrutu:
Will have to pick one solution between https://github.com/open-telemetry/opentelemetry-collector/pull/6053 and https://github.com/open-telemetry/opentelemetry-collector/pull/6052

dmitryax:
I think https://github.com/open-telemetry/opentelemetry-collector/pull/6052 is a better option providing extensibility that possibly can be utilized

bogdandrutu:
@jpkrohling also vote for the same. So will move forward with that proposal. Will close the alternate PR.

PR #6052, opened 2022-09-12
Replace immutable TraceState with a mutable struct

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6022

Main motivation is that we may want to allow mutating (adding an entry to the state).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,7 +100,7 @@
   - `ptrace.Span.SetTraceState` in favor of `ptrace.Span.TraceStateStruct().FromRaw`
   - `ptrace.SpanLink.TraceState` in favor of `ptrace.SpanLink.TraceStateStruct().AsRaw()`
   - `ptrace.SpanLink.SetTraceState` in favor of `ptrace.SpanLink.TraceStateStruct().FromRaw`
-  - `TraceStateStruct` is a temporary name that will be replaced back to TraceState in the next release.
+  - `TraceStateStruct` is a temporary name that will be replaced back to `TraceState` in the next release.
 
 ### 💡 Enhancements 💡
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,12 @@
   - `SummaryDataPoint.Flags` -> `SummaryDataPoint.FlagsImmutable`
   - `MetricDataPointFlags` -> `MetricDataPointFlagsImmutable`
   - `NewMetricDataPointFlags` -> `MetricDataPointFlagsImmutable`
+- Deprecate `ptrace.TraceState` in favor of `pcommon.TraceState`. (#6052)
+  - `ptrace.Span.TraceState` in favor of `ptrace.Span.TraceStateStruct().AsRaw()`
+  - `ptrace.Span.SetTraceState` in favor of `ptrace.Span.TraceStateStruct().FromRaw`
+  - `ptrace.SpanLink.TraceState` in favor of `ptrace.SpanLink.TraceStateStruct().AsRaw()`
+  - `ptrace.SpanLink.SetTraceState` in favor of `ptrace.SpanLink.TraceStateStruct().FromRaw`
+  - `TraceStateStruct` is a temporary name that will be replaced back to TraceState in the next release.
 
 ### 💡 Enhancements 💡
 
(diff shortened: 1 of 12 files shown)


pdata/ptrace/json_test.go
```diff
@@ -84,7 +84,7 @@ var tracesOTLPFull = func() Traces {
 	sp.SetDroppedLinksCount(1)
 	sp.SetEndTimestamp(pcommon.NewTimestampFromTime(time.Now()))
 	sp.SetParentSpanID(spanID)
-	sp.SetTraceState("state")
+	sp.TraceStateStruct().FromRaw("state")
```
jpkrohling, line 87: Can't we keep having the setter like it was before?

pdata/ptrace/traces.go
```diff
@@ -70,14 +70,34 @@ func (ms Traces) ResourceSpans() ResourceSpansSlice {
 	return newResourceSpansSlice(&ms.getOrig().ResourceSpans)
 }
 
-// TraceState is a string representing the tracestate in w3c-trace-context format: https://www.w3.org/TR/trace-context/#tracestate-header
+// Deprecated: [v0.60.0] use pcommon.TraceState.
 type TraceState string
 
 const (
-	// TraceStateEmpty represents the empty TraceState.
+	// Deprecated: [v0.60.0] use pcommon.TraceState.
```
jpkrohling, line 77: People using this are using a specific instance of the trace state. Perhaps the godoc tell people how to accomplish the same with the new API?

pdata/ptrace/generated_traces.go
```diff
@@ -620,14 +620,9 @@ func (ms Span) SetSpanID(v pcommon.SpanID) {
...
-// TraceState returns the tracestate associated with this Span.
-func (ms Span) TraceState() TraceState {
-	return TraceState(ms.getOrig().TraceState)
-}
-
-// SetTraceState replaces the tracestate associated with this Span.
-func (ms Span) SetTraceState(v TraceState) {
-	ms.getOrig().TraceState = string(v)
+// TraceStateStruct returns the tracestatestruct associated with this Span.
+func (ms Span) TraceStateStruct() pcommon.TraceState {
```
jpkrohling, line 624: This name is going to change in the future, right?

jpkrohling (commented):
I also prefer this over #6053

pdata/ptrace/json_test.go
bogdandrutu: The rule of pdata: If it is mutable we don't have a setter. If it is immutable then we have a setter. In this proposal the struct is mutable, so we don't offer a setter (users should change in place).

pdata/ptrace/generated_traces.go
bogdandrutu: Correct will become `TraceState`, but to do deprecation cannot reuse the same name.

pdata/ptrace/traces.go
bogdandrutu: Clarified how to use it temporary. There are only 2 usages of this in contrib, will take as an action item to provide an API or a helper to check for empty.

CHANGELOG.md
```diff
@@ -95,6 +95,11 @@
   - `SummaryDataPoint.Flags` -> `SummaryDataPoint.FlagsImmutable`
   - `MetricDataPointFlags` -> `MetricDataPointFlagsImmutable`
   - `NewMetricDataPointFlags` -> `MetricDataPointFlagsImmutable`
+- Deprecate `ptrace.TraceState` in favor of `pcommon.TraceState`. (#6052)
+  - `ptrace.Span.TraceState` in favor of `ptrace.Span.TraceStateStruct().AsRaw()`
+  - `ptrace.Span.SetTraceState` in favor of `ptrace.Span.TraceStateStruct().FromRaw`
+  - `ptrace.SpanLink.TraceState` in favor of `ptrace.SpanLink.TraceStateStruct().AsRaw()`
+  - `ptrace.SpanLink.SetTraceState` in favor of `ptrace.SpanLink.TraceStateStruct().FromRaw`
```
dmitryax, line 102: Let's maybe clarify here that `TraceStateStruct` is a temporary name that will be replaced back to `TraceState` in the next release?

dmitryax (approved):
One doc nit, otherwise LGTM

CHANGELOG.md
bogdandrutu: Done.

dmitryax (approved):

codeboten (approved):
+1 to #6052 over #6053, i couldn't come up with a useful scenario where an immutable tracestate would be preferable

PR #6060, opened 2022-09-13
[pdata] Add Append and EnsureCapacity methods to primitive slices

dmitryax:
Follow-up PR from https://github.com/open-telemetry/opentelemetry-collector/pull/5971:

> No way of increasing the capacity or length. We can discuss about that API.

Given that we have `SetAt` interface not `Append...`, I think it makes sense to have `SetLen`. But I'm open to other suggestions

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6063

cc @bogdandrutu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Delete deprecated `p[metric|log|trace].MarshalerSizer` interfaces (#6083)
 
+### 💡 Enhancements 💡
+
+- Add AppendEmpty and EnsureCapacity method to primitive pdata slices (#6060)
 
 ## v0.60.0 Beta
 
--- a/pdata/internal/cmd/pdatagen/internal/primitive_slice_structs.go
+++ b/pdata/internal/cmd/pdatagen/internal/primitive_slice_structs.go
@@ -61,6 +61,25 @@ func (ms ${structName}) SetAt(i int, val ${itemType}) {
 	(*ms.getOrig())[i] = val
 }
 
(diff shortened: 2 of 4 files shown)


bogdandrutu:
> Given that we have SetAt interface not Append..., I think it makes sense to have SetLen. But I'm open to other suggestions

Confused about why not offering an `Append(vals ...int64)` is not a valid solution as well? 

Some other question:
1. Should we also offer `EnsureCapacity`?

dmitryax:
@bogdandrutu  My reasoning for `SetLan` is coming from the usage. Usually we know length in advance and just set/update existing elements not appending anything. 

Maybe we shouldn't rely on usage and think of the slices in more abstracted way. In that case `Append` and `EnsureCapacity` probably is better, since it provides more flexibility going forward

bogdandrutu:
I think an `Append` and `EnsureCapacity` make total sense and moves this so be similar with other slices. In addition we can also have `SetLen` but want tot understand if it does allocation if not enough space or crash (for example reslicing a slice crash if not enough capacity). I think `SetLen` can wait until we see more use-cases, what do you think?

dmitryax:
> I think SetLen can wait until we see more use-cases, what do you think?

Ok sounds good. let's do that

dmitryax:
@bogdandrutu , updated

bogdandrutu (approved):

bogdandrutu:
Need a rebase

dmitryax:
Rebased

PR #6092, opened 2022-09-15
[pdata] Expose pcommon.Value.AsRaw/FromRaw methods

dmitryax:
These methods would be good to expose given that we have it for several other types. They would be pretty useful to replace custom functions in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/telemetryquerylanguage/contexts/internal/tqlcommon/value.go


bogdandrutu (approved):
Though, the link provided does not work since the from has special values for some type of primitive slices, which may actually not be supported in reality (I know there is no way to provide that now, so may be good to remove them :))

Issue #6081, opened 2022-09-14
[pdata] Revisit need of `Val` suffix of `Value` methods


dmitryax:
Currently all the setter and getter methods of `pcommon.Value` end with `Val` suffix, e.g.: `Value.BoolVal`, `Value.SetBoolVal`, `Value.MapVal` etc. The only reason for having the suffix is that otherwise we end up with a `Value.String` method unintentionally implementing `fmt.Stringer` interface.

Potentially we can remove the suffix from all the methods except for `Value.StringVal` where we can add an explicit note about the naming. I'm not sure if it's a good enough reason to introduce this exception. 

We can also keep the existing naming. I just wanted to bring this for discussion before releasing pdata 1.0. I don't have a strong preference.

@open-telemetry/collector-approvers please share your thoughts

bogdandrutu:
Another option is to call `GetString`?

dmitryax:
`Get` prefix goes against go naming recommendations, but I still like it more than other options.

If we ignore `fmt.Stringer`, `Value` of any types other than string will be printed as `""`, so for me it's the least desirable option

dmitryax:
I also like `Get` because it's getting more explicit that you can get anything from `Value` and in most cases it's a zero/invalid result.

bogdandrutu:
Yes, if we go with `GetX`, we should not panic if not correct type, and return zero initialized value.

dmitryax:
> Yes, if we go with GetX, we should not panic if not correct type, and return zero initialized value.

This is what we are doing now, it's all zero values, which also means invalid values for complex types like Map and Slice.

dmitryax:
There are 3 different parts of the related pdata API that we need to consider additionally in this issue.

1. `pcommon.Value` (https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pcommon/common.go#L70-L321) which this issue is submitted for initially.  It's a wrapper over a proto oneof message called `Value`. All getters/setters methods have `Val` suffix just to avoid implementing `fmt.Stringer` interface, but getter for type doesn't have any prefix.

[...]

PR #6099, opened 2022-09-16
[pdata] Remove Val suffix from pcommon.Value getter/setter methods

dmitryax:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/6081


bogdandrutu:
I like the new API... Want to hear more opinions. @open-telemetry/collector-approvers ?

dmitryax:
@bogdandrutu If we merge this, we will still have `(Set)[Double|Int]Val` methods on `pmetric.NumberDataPoint` and `pmetric.Examplar` structs, but I believe it's good to keep the `Val` suffix there because they have other unrelated fields and they look through another _Value_ entity, so we should keep something related to _value_ in those method names.

bogdandrutu:
Maybe that should be Value?

bogdandrutu:
@dmitryax what about the metrics data type then? We definitely are not consistent here

tigrannajaryan:
I am not sure this is an improvement. Get prefix is explicitly not recommended by Effective Go for getters. Granted, Val suffix is also not pretty.

Minor argument in favour of Val suffix: it is faster to use with autocomplete, less characters to type until you get the right hint.

However, I can live with either approach.

TylerHelmuth (approved):
I like this a lot. Get and Set feel like natural prefixes for this situation, even if it goes against Go naming standards

dmitryax:
> Maybe that should be Value?

Yes, `(Set)[Double|Int]Value` on `pmetric.NumberDataPoint` and `pmetric.Examplar` sounds better to me 👍 

> @dmitryax what about the metrics data type then? We definitely are not consistent here

We are not, yes. We don't have any prefix or suffix for the metric type getters, it's just `Sum()`, `Gauge()` etc. It goes through `Data` oneof protobuf messages called `Data`. I'm not sure what we can do there to bring consistency.  Making it `SumValue` or `SumData` doesn't feel convenient.

TylerHelmuth:
> We definitely are not consistent here 

This is pretty annoying, and if we can't use `Get|Set` everywhere then I'd _probably_ vote for consistency over anything else.  But if I was a new user to the pdata API, I think `GetString` and `SetString` are nicer than `StringVal` and `SetStringVal`.  

But I also have a bias towards languages with Getters and Setters.  An argument could be made that for Go `Get` isn't the expected name and would throw users off.

TylerHelmuth:
>  it's just Sum(), Gauge()

This makes me really wish we could just do `SetString` and `String`.  `GetSum` and `GetGauge` don't feel right.  (Which is maybe a good argument against using `Get` for `Value`.)

bogdandrutu:
@dmitryax please summarize the conversation in the issue:
* 4 places where we have this pattern right now (please enumerate them, and add links so people can find them easily), and all somehow inconsistent for naming.
* Let's play with the proposals: 1. use get/set (see how that impact other places); 2. In case of Metric/Exemplar should we have `SumData` to be consistent with `DoubleValue`?

So let's enumerate the rules that we want to apply (or exceptions if any) and apply them and see what we get. By apply just some code snippets it is fine.

jpkrohling (approved):
I preferred `BoolVal` to `GetBool` until I realized that `Get` would be a better pair to `Set` in `SetBool`.

dmitryax:
Please do not merge until a decision is made on https://github.com/open-telemetry/opentelemetry-collector/issues/6081#issuecomment-1258364386

bogdandrutu (changes requested):
To avoid merging until decision in the issue is made

exporter/loggingexporter/internal/otlptext/databuffer.go
```diff
@@ -273,17 +273,17 @@ func (b *dataBuffer) logLinks(description string, sl ptrace.SpanLinkSlice) {
 func attributeValueToString(v pcommon.Value) string {
 	switch v.Type() {
 	case pcommon.ValueTypeString:
```
bogdandrutu, line 275: `ValueTypeStr`?

CHANGELOG.md
```diff
@@ -35,6 +35,11 @@
 - Deprecate `pmetric.NumberDataPoint.[Set]?[Int|Double]Val()` in favor of `pmetric.NumberDataPoint.[Set]?[Int|Double]Value()`. (#6134)
 - Deprecate `pmetric.Exemplar.[Set]?[Int|Double]Val()` in favor of `pmetric.Exemplar.[Set]?[Int|Double]Value()`. (#6134)
 
+### 🚩 Deprecations 🚩
+
+- Rename all `pcommon.Value` getter/setter methods by removing `Val` suffix. Getters now have `Get` prefix instead.
```
bogdandrutu, line 40: Update?

exporter/loggingexporter/internal/otlptext/databuffer.go
```diff
@@ -273,17 +273,17 @@ func (b *dataBuffer) logLinks(description string, sl ptrace.SpanLinkSlice) {
...
 	case pcommon.ValueTypeInt:
-		return strconv.FormatInt(v.IntVal(), 10)
+		return strconv.FormatInt(v.Int(), 10)
 	case pcommon.ValueTypeSlice:
-		return sliceToString(v.SliceVal())
+		return sliceToString(v.Slice())
 	case pcommon.ValueTypeMap:
-		return mapToString(v.MapVal())
+		return mapToString(v.Map())
 	default:
```
bogdandrutu, line 287: This lacks the ValueTypeBytes? I will take care of this in a separate PR.

dmitryax: Yes, nice catch.

CHANGELOG.md
dmitryax: updated

exporter/loggingexporter/internal/otlptext/databuffer.go
dmitryax: renamed

pdata/pcommon/common.go
```diff
@@ -201,125 +204,196 @@ func (v Value) Type() ValueType {
 	return ValueTypeEmpty
 }
 
-// StringVal returns the string value associated with this Value.
-// If the Type() is not ValueTypeString then returns empty string.
+// Str returns the string value associated with this Value.
+// The shorter name is used instead of String to avoid implementing fmt.Stringer interface.
```
bogdandrutu, line 208: Consider to have this comment on the "SetStr" as well.

bogdandrutu (approved):

PR #6102, opened 2022-09-16
Add traceparent id in the Debug log for exporter

mcmho:
Added trace id in debug log in otlphttpexporter

**Description:** 
Enable cross correlation of log messages easier with parent trace id

if this change is agreed by the community, suggestion is to do the same thing for otlpexporter.go as well.

**Link to tracking Issue:**  
https://github.com/open-telemetry/opentelemetry-collector/issues/6368
(but should be moved back to /opentelemetry-collector instead)

**Testing:**  
tested methods with promethus self metrics


exporter/otlphttpexporter/otlp.go
```diff
@@ -124,7 +125,9 @@ func (e *exporter) pushLogs(ctx context.Context, ld plog.Logs) error {
 }
 
 func (e *exporter) export(ctx context.Context, url string, request []byte) error {
-	e.logger.Debug("Preparing to make HTTP request", zap.String("url", url))
+	var span = trace.SpanFromContext(ctx)
+	var traceId = span.SpanContext().TraceID().String()
```
TylerHelmuth, line 129: Maybe we can do this function calls only when telemetry `service.telemetry.logs.level` is set to `DEBUG`?

exporter/otlphttpexporter/otlp.go
```diff
@@ -124,7 +125,9 @@ func (e *exporter) pushLogs(ctx context.Context, ld plog.Logs) error {
...
+	e.logger.Debug("Preparing to make HTTP request", zap.String("url", url), zap.String("traceId", traceId))
```
TylerHelmuth, line 130: What happens if we are exporting a metric?  Would it print out an empty traceId?  Can we only include the traceId if we have one?

TylerHelmuth (commented):
I love this idea.  Maybe it could end up in exporterhelper instead so any trace/log exporter could take advantage of something like this?

exporter/otlphttpexporter/otlp.go
bogdandrutu: you can, use `zap.Stringer("traceId", traceId)` so delays the call to `String()`.

exporter/otlphttpexporter/otlp.go
```diff
@@ -124,7 +125,9 @@ func (e *exporter) pushLogs(ctx context.Context, ld plog.Logs) error {
 }
 
 func (e *exporter) export(ctx context.Context, url string, request []byte) error {
-	e.logger.Debug("Preparing to make HTTP request", zap.String("url", url))
+	var span = trace.SpanFromContext(ctx)
```
bogdandrutu, line 128: use `trace. SpanContextFromContext()`

exporter/otlphttpexporter/otlp.go
bogdandrutu: traceID is the traceID of the request in the collector (not from the telemetry we transport), we always have one if the otel-trace-go is configured.

exporter/otlphttpexporter/otlp.go
mcmho: all these are great feedbacks!   thank you so much!

exporter/otlphttpexporter/otlp.go
mcmho: Thank you @TylerHelmuth for being so supportive!   
Do you mean `func (ts *timeoutSender) send(req internal.Request) error` in exporthelper/common.go?
 
I see that is the common path for trace/log/metric export.

TylerHelmuth: @mcmho I don't think we have access to a logger from that function so there would need to be more work to add this feature to the helper.  @bogdandrutu what do you think; do we stick with otlphttpexporter (and otlpexporter in a future PR) for now, or try to provide this feature to all exporters from the get go?

mcmho: Thanks @TylerHelmuth!
@bogdandrutu ,  Is it not a good idea to keep a logger in 
```
// NewExporter creates a new Exporter.
func NewExporter(cfg ExporterSettings) *Exporter {
	return &Exporter{
		level:          cfg.ExporterCreateSettings.TelemetrySettings.MetricsLevel,
		spanNamePrefix: obsmetrics.ExporterPrefix + cfg.ExporterID.String(),
		mutators:       []tag.Mutator{tag.Upsert(obsmetrics.TagKeyExporter, cfg.ExporterID.String(), tag.WithTTL(tag.TTLNoPropagation))},
		tracer:         cfg.ExporterCreateSettings.TracerProvider.Tracer(cfg.ExporterID.String()),
	}
}
```

mcmho: Thanks @TylerHelmuth @bogdandrutu !

I appreciate your idea behind beauty of common code and reusability. 
For now, another need I see is that a logger would be useful for _otlpexporter_.oltp
so enabling it with a logger should be a good thing to do.

Currently both _otlphttpexporter_.exporter  and _loggingexporter_.loggingExporter have a logger,
```
type loggingExporter struct {
	logger     *zap.Logger
        ......  }
type exporter struct {
	logger     *zap.Logger
        ......  }
```
it might be good to make that symmetrical in otlpexporter.otlp.go as well:
type exporter struct {
	logger     *zap.Logger
        ......  }

And for longer term / future PR, we can attach to consolidate this feature to all exporters if we see fit.

How does that sound?   @bogdandrutu  @TylerHelmuth

mcmho: Wait... i just realized that `component.TelemetrySettings` inside `otlpexporter.exporter` has a logger.  
I can just use that with otlpexporter?

mcmho: Fixed by the latest commit.

exporter/otlphttpexporter/otlp.go
mcmho: Fixed by the latest commit.

exporter/otlpexporter/otlp.go
```diff
@@ -103,18 +105,21 @@ func (e *exporter) shutdown(context.Context) error {
 func (e *exporter) pushTraces(ctx context.Context, td ptrace.Traces) error {
 	req := ptraceotlp.NewRequestFromTraces(td)
 	_, err := e.traceExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
+	e.settings.Logger.Debug("Exporting traces for ", zap.Stringer("traceId", trace.SpanContextFromContext(ctx).TraceID()))
```
Aneurysm9, line 108:
```suggestion
	e.settings.Logger.Debug("Exporting traces", zap.Stringer("traceId", trace.SpanContextFromContext(ctx).TraceID()))
	req := ptraceotlp.NewRequestFromTraces(td)
	_, err := e.traceExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
```

I would emit these logs at the beginning of these methods so that it is obvious that exporting has started even if it takes some time.  Also, the message should be able to stand alone; the additional attributes are to add context and are not part of the log message.

exporter/loggingexporter/logging_exporter.go
```diff
@@ -29,15 +30,16 @@ import (
...
+	logLevel         zapcore.Level
 	logsMarshaler    plog.Marshaler
 	metricsMarshaler pmetric.Marshaler
 	tracesMarshaler  ptrace.Marshaler
 }
 
-func (s *loggingExporter) pushTraces(_ context.Context, td ptrace.Traces) error {
-	s.logger.Info("TracesExporter", zap.Int("#spans", td.SpanCount()))
+func (s *loggingExporter) pushTraces(ctx context.Context, td ptrace.Traces) error {
+	s.logger.Info("TracesExporter", zap.Int("#spans", td.SpanCount()),
```
Aneurysm9, line 41: These entries are emitted at `Info` while the other exporters changed in this PR are exporting at `Debug`.  Why the difference?

Aneurysm9 (commented):
It's not clear to me what value this change provides.  It's very limited in scope and expanding it to all exporters in this manner feels untenable.  What can we do to make this information automatically applied to all log entries in a consistent manner?

exporter/loggingexporter/logging_exporter.go
mcmho: @Aneurysm9  for logging_exporter.go, it has always been Info() and that is rate limited, so I leave it as is. 
as for oltp.go, I don't see any rate limiter there.   I feel like it is too frequent for Info level.   What do you think?

exporter/otlpexporter/otlp.go
mcmho: Okay, will do.  thanks @Aneurysm9

exporter/loggingexporter/logging_exporter.go
mx-psi: All logging exporter output is at info level since #5678

mx-psi:
> What can we do to make this information automatically applied to all log entries in a consistent manner?

One way of doing this is to wrap the logger's [zapcore.Core](https://pkg.go.dev/go.uber.org/zap@v1.23.0/zapcore#Core) and replace its `Write` method so that it adds this new field if the entry is at debug level. We could do that at the exporterhelper level.

It's a bit more complex, but it should be equally performant and work for all exporters

PR #6088, opened 2022-09-15
Remove deprecated funcs from pcommon.[Map|Value|Slice]

bogdandrutu:
Signed-off-by: Bogdan <bogdandrutu@gmail.com>

PR #6105, opened 2022-09-16
[pdata] Restore NewValueBytesEmpty that was removed without deprecation

dmitryax:
Appeared that we accidentally broke the deprecation rule in https://github.com/open-telemetry/opentelemetry-collector/pull/6088. This PR unblocks core dependency upgrade in contrib


bogdandrutu (approved):
Thanks for catching this.

PR #6109, opened 2022-09-16
Remove duplicate entry in CHANGELOG

bogdandrutu:


bogdandrutu:
I know that @jpkrohling will say that this is yet another reason to use the contrib tool, still I am not sold for the extra work to create a new file every time, etc. makes sense. Right now we are merging lots of PRs in core, but this will slowdown and the tool just causes way more work. Also just my feeling, we are over document things in CHANGELOG because every PR needs a CHANGELOG entry, look at the random enhancements in contrib. Anyway just me commenting about this :))

dmitryax (approved):

jpkrohling:
> I know that @jpkrohling will say that this is yet another reason to use the contrib tool

You know me too well!

Issue #5084, opened 2022-03-24
otel collector telemetry metrics server won't spawn again after restarting


splunkericl:
**Describe the bug**
When the collector is shutdown and run again(or another collector is run again), the telemetry metrics server won't spawn again. 
- collector telemetry is shut down [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L277) when shutdown is called
- when collector is run telemetry is init [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L246)
- the telemetry init implementation only [initialize this once](https://github.com/open-telemetry/opentelemetry-collector/blob/1b7618728835aca5591221b23e92d91ad43f1905/service/telemetry.go#L84)

**Steps to reproduce**
- create a otel collector
- call Run with settings
- call shutdown
- create another otel collector
- call Run with new settings
- metrics server won't spawn again

**What did you expect to see?**
When a new collector is run again. metrics server should still be spawned

**What did you see instead?**
metrics server no longer spawns due to initOnce call.

**What version did you use?**
Version: v0.46.0

**What config did you use?**
Any valid config

**Environment**
OS: macOS catalina
Compiler(if manually compiled): go 1.17.3

**Additional context**

tigrannajaryan:
>call shutdown
Are you calling [Shutdown](https://github.com/open-telemetry/opentelemetry-collector/blob/cb868e20c8ffc616c7ca84c68834bb3bdeab82c3/service/collector.go#L128)? As far as I remember that puts the Service in a terminal state. It is not expected that Run will work again after that.

More importantly, why are you doing this? If you want to reload the Service with new config you need to use the [ConfigProvider.Watch](https://github.com/open-telemetry/opentelemetry-collector/blob/cb868e20c8ffc616c7ca84c68834bb3bdeab82c3/service/config_provider.go#L46) capability.

splunkericl:
Even if we use the watch method, the internal implementation for watch is similar. See: https://github.com/open-telemetry/opentelemetry-collector/blob/cb868e20c8ffc616c7ca84c68834bb3bdeab82c3/service/collector.go#L162-L165

It would invoke the shutdown function and then Run with the new settings.

tigrannajaryan:
> It would invoke the shutdown function and then Run with the new settings.

OK, this looks like a legitimate bug in that case. 

[...]

PR #6138, opened 2022-09-22
Changed how the telemetryInitializer is created

cpheps:
**Description:** 
Changed how the `telemetryInitializer` is created. Previously only a single instance of the `telemetryInitializer` was created for a process. The caused an issue if a `Collector` instance was shutdown and a new one was created the new instance would get the global instance that was shutdown as a result of the first Collector instance being shutdown.

The new logic ensures the feature gate is only registered once but a new `telemetryInitializer` instance is created per instance of the collector.

**Link to tracking Issue:** 
Tangentially related to #5084. 

In that issue the desired design is:

> I think we are going to the direction to follow the language pattern which is after Shutdown you cannot call Run on the same instance, you have to create a new instance.

I believe this respects that design and fixes a bug where a new instance could not fully startup with the same expectations as the previous instance.

**Testing:** 
Modified unit tests. Also, tested with our OPAMP enabled [agent](https://github.com/observIQ/observiq-otel-collector) and [server](https://github.com/observiq/bindplane-op) to verify that we can create a new instance of the collector with a changed config in a single process and have telemetry still be active. Previously it would not be.

### DIFF
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -46,9 +46,6 @@ import (
 	"go.opentelemetry.io/collector/service/telemetry"
 )
 
-// internalMetricFGRegisterOnce is a Once that ensures that the internal metric feature gate is only registered once
-var internalMetricFGRegisterOnce sync.Once
-
 const (
 	zapKeyTelemetryAddress = "address"
 	zapKeyTelemetryLevel   = "level"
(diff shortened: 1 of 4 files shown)


cpheps:
For CI failures
Looks like codecov got a 404 on upload and that the contrib tests are currently broken but being fixed.

service/telemetry.go
```diff
@@ -78,10 +78,13 @@ type telemetryInitializer struct {
 }
 
 func newColTelemetry(registry *featuregate.Registry) *telemetryInitializer {
-	registry.MustRegister(featuregate.Gate{
-		ID:          useOtelForInternalMetricsfeatureGateID,
-		Description: "controls whether the collector to uses OpenTelemetry for internal metrics",
-		Enabled:     false,
+	// Ensure we only do the register once
+	internalMetricFGRegisterOnce.Do(func() {
```
djaglowski, line 82: It feels awkward to have a once-ever command occur within what is essentially a constructor. Should this just be moved elsewhere? Perhaps in an `init` func?

This might allow the test override to be removed as well.

djaglowski:
Is there a particular PR that is supposed to fix the contrib tests? If so, will you link it here so we can track it?

service/telemetry.go
cpheps: Yeah, it's not uncommon for singleton patterns to do this in getters that act as first time creations. Though the purpose of this function has changed with the new logic. I have no problem putting it into an init function. I will have to double check the tests though. If this logic needs to be tested it'll have to be broken out into it's own function that is wrapped in an init.

cpheps:
> Is there a particular PR that is supposed to fix the contrib tests? If so, will you link it here so we can track it?

I believe it was this PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14390. The contrib tests look like the fail due to a signalfx go mod error.

service/telemetry.go
cpheps: Pushed a refactor to remove the `Once` and move the register into an init function.

djaglowski (approved):
This LGTM.

djaglowski:
@open-telemetry/collector-maintainers, anyone available to take a look at this one?

Aneurysm9 (approved):

Issue #2336, opened 2021-01-06
Signal Translation Proprosal


djaglowski:
**Is your feature request related to a problem? Please describe.**
It is sometimes necessary to convert signals from one signal type to another. Signal translations may be direct one-to-one operations, or may include derivative signals such as counts, aggregations, summarizations, etc. A specific example can be seen in contrib [#403](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/403).

The OpenTelemetry Collector does not currently have a standard mechanism to support such use cases.

**Describe the solution you'd like**

Introduce a new component type, tentatively called `connectors`, which would serve the following purposes:
- Act as a formal point of translation from one signal type to another
- Avoid introducing to the pipeline any additional fan-out points
- Formalize the linking together of multiple pipelines

![image](https://user-images.githubusercontent.com/5255616/103781784-c6c7b480-5004-11eb-95ff-6659b6a58006.png)

## Rationale

First, consider that exporters may be thought of as a natural point of connection between pipelines, particularly when it comes to translation from one signal type to another. After all, exporters are fundamentally concerned with translating OTLP formatted signals to any arbitrary output format. This output format could just as well be another OTLP signal type, which could then be sent to an OTLP receiver.

Next, consider the following example scenario, simplified from the suggestion [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/403#issuecomment-747379678).
- tracing data, processed to derive metrics and exported
- tracing data then needs to be further processed before being exported

A minimal solution to this use case could depend on the addition of a new exporter, which would derive metrics from traces and emit them like the OTLP exporter. This solution would use the fanout stage of one pipeline to replicate traces, and then concurrently send traces to two additional pipelines.

```
receivers:
  some/tracingreceiver:
  otlp/tracingreceiver:
    endpoint: "0.0.0.0:1111"
  otlp/metricsreceiver:
    endpoint: "0.0.0.0:2222"
  
processors:
  some/tracingprocessor:
  some/metricsprocessor:
  
exporters:
  some/tracingexporter:
  some/metricsexporter:
  otlp/tracingexporter:
    endpoint: "0.0.0.0:1111"
  tracestometrics: # perform some derivation and then behave as otlp/metricsexporter
    endpoint: "0.0.0.0:2222"

service:
  pipelines:
    a:
      receivers: [some/tracingreceiver]
      exporters: [otlp/tracingexporter, tracestometrics]
    b:
[...]

PR #6140, opened 2022-09-22
Connectors prototype

djaglowski:
## Connectors

This is a working prototype of a new component type called "Connectors", which were proposed and described in #2336.

The general idea is that a connector acts as an exporter/receiver pair, bridging the gap between two or more pipelines. A connector can be used to merge data from multiple pipelines, or to split/replicate data to multiple pipelines. Connectors are configured declaratively, in the same way as receivers, processors, exporters, or extensions, and are integrated into the existing pipeline structure by being used in place of at least one exporter and at least one receiver.

The implementation includes two very simple connectors, `nop` and `count`, which allow for basic demonstration of functionality. (`count` is a bit contrived, but useful to illustrate a capability)

As much as possible, I've leaned on the existing notions of receivers and exporters. My goal is to allow existing functionality to do as much of the work as possible. Because of this we get a lot of useful functionality for free:
- Connectors inherit the ability (or inability) to be used as receivers in a particulate type of pipeline. Same with regard to acting as exporters. (eg. `count` cannot be used as a receiver in a logs or traces pipeline)
- When used as a receiver, a single instance can be shared by multiple pipelines. The built-in fanout capability is used automatically.

#### Count logs as they are forwarded. Report the count as a metric.

```yaml
receivers:
  otlp:

exporters:
  otlp/log_backend:
  otlp/count_of_logs:

connectors:
  count:

service:
  pipelines:
    logs:
      receivers: [otlp]
      exporters: [otlp/log_backend, count]
    metrics:
      receivers: [count]
      exporters: [otlp/count_of_logs]
```

#### Duplicate metrics. Forward one stream directly to a backend. Batch the other stream and forward it to a different backend.

```yaml
receivers:
  otlp:

processors:
  batch:

exporters:
  otlp/raw_backend:
  otlp/batched_backend:

connectors:
  nop:

service:
  pipelines:
    metrics/in:
      receivers: [otlp]
      exporters: [nop]
    metrics/raw:
      receivers: [nop]
      exporters: [otlp/raw_backend]
    metrics/batched:
      receivers: [nop]
      processors: [batch]
      exporters: [otlp/batched_backend]
```

#### Merge traces from two pipelines, then batch them.

```yaml
receivers:
  otlp/1:
  otlp/2:

processors:
  batch:

exporters:
  otlp:

connectors:
  nop:

service:
  pipelines:
    traces/1:
      receivers: [otlp/1]
      exporters: [nop]
    traces/2:
      receivers: [otlp/2]
      exporters: [nop]
    traces/out:
      receivers: [nop]
      processors: [batch]
      exporters: [otlp]
```
 

Several important aspects of the design are not yet implemented, but I wanted to share the prototype for early feedback.

TODO:
- Tests / Documentation
- Validation that a connector is actually used in two or more pipelines
  - At least once as a receiver
  - At least once as an exporter
- `ConnectorFactoryOptions` to describe and enforce valid pipeline connections
  - The `nop` connector can handle `logs -> logs`, `metrics -> metrics`, `traces -> traces`
  - The `count` connector can handle `logs -> metric`, `metrics -> metrics`, `traces -> metrics`
- Graph validation / cycle detection
  - Users should not be able to connect pipelines in such as way where data flow is cyclical.
- Start/Stop ordering. Ideally, the overall graph is drained in such a way that signals are not stranded during shutdown.

### DIFF
--- a/component/connector.go
+++ b/component/connector.go
@@ -18,21 +18,6 @@ import (
 	"go.opentelemetry.io/collector/config"
 )
 
-// Connector exports telemetry data from one pipeline to another.
-// type Connector interface {
-// 	Component
-// 	Exporter
-// 	Receiver
-// }
-
-// // ConnectorCreateSettings configures Connector creators.
-// type ConnectorCreateSettings struct {
-// 	TelemetrySettings
-
-// 	// BuildInfo can be used by components for informational purposes
-// 	BuildInfo BuildInfo
-// }
-
 // ConnectorFactory is factory interface for connectors.
 //
 // This interface cannot be directly implemented. Implementations must
@@ -60,6 +45,12 @@ type ConnectorCreateDefaultConfigFunc func() config.Connector
 func (f ConnectorCreateDefaultConfigFunc) CreateDefaultConfig() config.Connector {
 	return f()
 }
+func (f ConnectorCreateDefaultConfigFunc) createDefaultReceiverConfig() config.Receiver {
+	return f()
+}
+func (f ConnectorCreateDefaultConfigFunc) CreateDefaultExporterConfig() config.Exporter {
+	return f()
+}
 
 type connectorFactory struct {
 	baseFactory
@@ -84,9 +75,21 @@ func NewConnectorFactory(
 }
 
 func (f *connectorFactory) NewExporterFactory() ExporterFactory {
-	return NewExporterFactory(f.cfgType, nil, f.exporterFactoryOptions...)
+	return NewExporterFactory(f.cfgType, f.CreateDefaultExporterConfig, f.exporterFactoryOptions...)
 }
 
 func (f *connectorFactory) NewReceiverFactory() ReceiverFactory {
-	return NewReceiverFactory(f.cfgType, nil, f.receiverFactoryOptions...)
+	return NewReceiverFactory(f.cfgType, f.createDefaultReceiverConfig, f.receiverFactoryOptions...)
 }
+
+// TODO Implement and enforce ConnectorFactoryOptions that enumerate valid signal combos.
+//
+// Example: nopconnector
+// func AsLogsToLogsConnector() ConnectorFactoryOption
+// func AsMetricsToMetricsConnector() ConnectorFactoryOption
+// func AsTracesToTracesConnector() ConnectorFactoryOption
+//
+// Example: countconnector
+// func AsLogsToMetricsConnector() ConnectorFactoryOption
+// func AsMetricsToMetricsConnector() ConnectorFactoryOption
+// func AsTracesToMetricsConnector() ConnectorFactoryOption
--- a/component/connector.go
+++ b/component/connector.go
@@ -1,0 +1,92 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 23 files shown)


djaglowski:
CI seems to have some conflicting expectations about the builder and/or generated code, but it builds and runs locally.

connector/countconnector/count.go
```diff
@@ -0,0 +1,190 @@
...
+	return c.metricsConsumer.ConsumeMetrics(ctx, newCountMetric("metrics", md.MetricCount()))
+}
+
+func (c *countConnector) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
+	return c.metricsConsumer.ConsumeMetrics(ctx, newCountMetric("logs", ld.LogRecordCount()))
+}
+
+func newCountMetric(signalType string, count int) pmetric.Metrics {
+	ms := pmetric.NewMetrics()
+	rms := ms.ResourceMetrics().AppendEmpty()
```
Aneurysm9, line 177: Should this copy the resource and scope from the source data?

component/connector.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (f *connectorFactory) NewExporterFactory() ExporterFactory {
+	return NewExporterFactory(f.cfgType, f.CreateDefaultExporterConfig, f.exporterFactoryOptions...)
+}
+
+func (f *connectorFactory) NewReceiverFactory() ReceiverFactory {
+	return NewReceiverFactory(f.cfgType, f.createDefaultReceiverConfig, f.receiverFactoryOptions...)
+}
+
+// TODO Implement and enforce ConnectorFactoryOptions that enumerate valid signal combos.
```
djaglowski, line 85: Having thought about this some more, I believe it will be necessary to provide the full matrix of `Create<Type1>To<Type2>Func`'s, because each combination may have different behavior.

bogdandrutu:
@djaglowski thanks for doing this, I have couple of questions about the design:

* Does this fit the collector starting order? The reason I am asking is because we will have to create 2 "instances" (one receiver and one exporter) and call start twice? I want to see if you see any problems with that.
* Also this causes a bit of confusion to me that there are 2 default configs. Are we requiring to have 2 default configs?

I am struggling a bit with the design choice, I think we can add a proper new component `component.Connector` with a `config.Connector` and the factory very similar with any other factory (probably we need a way to configure next component later). The `component.Connector` can implement both interfaces `component.Exporter` and `component.Receiver`.

The late configuration for the "next consumer" can also be hacked, by passing a "lazy initialized consumer" which allows us to "swap" the next consumer there, we can check couple of designs there.

djaglowski:
Thanks for the feedback @bogdandrutu.

> Does this fit the collector starting order? The reason I am asking is because we will have to create 2 "instances" (one receiver and one exporter) and call start twice? I want to see if you see any problems with that.
> Also this causes a bit of confusion to me that there are 2 default configs. Are we requiring to have 2 default configs?
>
> I am struggling a bit with the design choice, I think we can add a proper new component component.Connector with a config.Connector and the factory very similar with any other factory (probably we need a way to configure next component later). The component.Connector can implement both interfaces component.Exporter and component.Receiver.

I agree this implementation is quite awkward for all the reasons you mentioned. Technically, I was able to work with around the dual constructor/configs issue, but authoring the individual components required some assumptions about the order of component creation, which should not be a concern of the component author. I thought perhaps there might be a way to provide a `connectorhelper` package that would abstract all that from the author, but I am still not sure quite what that would look like yet. In any case, I think you are right and I will take another pass at this where connectors are treated a proper first-class component.

> The late configuration for the "next consumer" can also be hacked, by passing a "lazy initialized consumer" which allows us to "swap" the next consumer there, we can check couple of designs there.

Good idea - I'll keep this in mind. At some point I need to take a closer look at start/stop order across the collector. I am not very familiar with this code yet but am hoping to build a topological ordering that ensures data is never blocked or stranded. If this works, I think we can pass next consumers in directly during construction, since we would build from the end of the graph backwards to the beginning in order to ensure data flow during startup.

djaglowski:
Quick update: I'm continuing to work on this and will open another PR that solidifies the new component type and also includes a graph-based implementation of `service/internal/pipelines.Pipelines`.

Issue #6136, opened 2022-09-21
Add service.name resource attribute to the collector's own telemetry


jlegoff:
**Is your feature request related to a problem? Please describe.**

The opam spec [states](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes) that an agent should have the following identifying attributes:
* service.name
* service.namespace
* service.version 
* service.instance.id 
* any other attributes that are necessary for uniquely identifying the Agent's own telemetry.

At the moment, the collector's own telemetry don't include these.

**Describe the solution you'd like**
The collector should include the above attributes - `service.namespace` and `service.instance.id` should probably be configurable. I'm not sure what the default values should be.

**Describe alternatives you've considered**
I've [suggested](https://cloud-native.slack.com/archives/C01N6P7KR6W/p1663680351659209) in slack a specific resource type might be created for agents, but the consensus seems to be that the collector is a service.

tigrannajaryan:
>At the moment, the collector's own telemetry don't include these.

This is partially correct. Collector does include `service.version` and `service.instance.id` already. `service.name` is indeed missing and based on today's SIG discussion should be added. 

Collector also allows to specify arbitrary attributes to include in its own telemetry using `resource` key under the `telemetry` key of the config. This allows the user to define `service.namespace` and any other attributes and also allows overriding `service.instance.id`.

The only thing I believe is really missing is that `service.name` should be by default included in own telemetry. I would expect it to be set to the FQDN, e.g. "io.opentelemetry.collector".

tigrannajaryan:
Some findings for posterity: I have been testing how Collector generates its own metrics using prometheusreceiver to collect them and was surprised to discover that `service.name` was included as Resource attributes on the collected metrics. I was surprised since I didn't expect this. However, after some digging I found that prometheusreceiver sets the `service.name` [equal to the configured job name](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/4d376d0d928dd82601f72abe432d763fb66610af/receiver/prometheusreceiver/internal/prom_to_otlp.go#L53). 

So, we still need to add the `service.name` as a Resource attribute, even if the prometheusreceiver overwrites it with the configured job name later (since prometheusreceiver is not the only way we will be collecting our own telemetry in the future).

jack-berg:
[...]

PR #6152, opened 2022-09-26
Add service.name resource attribute to the collector's own telemetry

tigrannajaryan:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/6136

- The default service.name is set to BuildInfo.Command (equals to "otelcol" for this repo).
- The service.name can be overridden by the end user via telemetry config setting.

### DIFF
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -117,27 +117,8 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger 
 
 	logger.Info("Setting up own telemetry...")
 
-	// Construct telemetry attributes from resource attributes.
-	telAttrs := map[string]string{}
-	for k, v := range cfg.Resource {
-		// nil value indicates that the attribute should not be included in the telemetry.
-		if v != nil {
-			telAttrs[k] = *v
-		}
-	}
-
-	if _, ok := cfg.Resource[semconv.AttributeServiceInstanceID]; !ok {
-		// AttributeServiceInstanceID is not specified in the config. Auto-generate one.
-		instanceUUID, _ := uuid.NewRandom()
-		instanceID := instanceUUID.String()
-		telAttrs[semconv.AttributeServiceInstanceID] = instanceID
-	}
-
-	if _, ok := cfg.Resource[semconv.AttributeServiceVersion]; !ok {
-		// AttributeServiceVersion is not specified in the config. Use the actual
-		// build version.
-		telAttrs[semconv.AttributeServiceVersion] = buildInfo.Version
-	}
+	// Construct telemetry attributes from build info and config's resource attributes.
+	telAttrs := buildTelAttrs(buildInfo, cfg)
 
 	if tp, err := textMapPropagatorFromConfig(cfg.Traces.Propagators); err == nil {
 		otel.SetTextMapPropagator(tp)
@@ -177,6 +158,37 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger 
 	}()
 
 	return nil
+}
+
+func buildTelAttrs(buildInfo component.BuildInfo, cfg telemetry.Config) map[string]string {
+	telAttrs := map[string]string{}
+
+	for k, v := range cfg.Resource {
+		// nil value indicates that the attribute should not be included in the telemetry.
+		if v != nil {
+			telAttrs[k] = *v
+		}
+	}
+
+	if _, ok := cfg.Resource[semconv.AttributeServiceName]; !ok {
+		// AttributeServiceName is not specified in the config. Use the default service name.
+		telAttrs[semconv.AttributeServiceName] = buildInfo.Command
+	}
+
+	if _, ok := cfg.Resource[semconv.AttributeServiceInstanceID]; !ok {
+		// AttributeServiceInstanceID is not specified in the config. Auto-generate one.
+		instanceUUID, _ := uuid.NewRandom()
+		instanceID := instanceUUID.String()
+		telAttrs[semconv.AttributeServiceInstanceID] = instanceID
+	}
+
+	if _, ok := cfg.Resource[semconv.AttributeServiceVersion]; !ok {
+		// AttributeServiceVersion is not specified in the config. Use the actual
+		// build version.
+		telAttrs[semconv.AttributeServiceVersion] = buildInfo.Version
+	}
+
+	return telAttrs
 }
 
 func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs map[string]string) (http.Handler, error) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@
 - Add AppendEmpty and EnsureCapacity method to primitive pdata slices (#6060)
 - Expose `AsRaw` and `FromRaw` `pcommon.Value` methods (#6090)
 - Convert `ValueTypeBytes` attributes in logging exporter (#6153)
+- service.name Resource attribute is added to Collector's own telemetry, defaults to the value of `BuildInfo.Command` and can be overridden in the config (#6152) 
 - Updated how `telemetryInitializer` is created so it's instanced per Collector instance rather than global to the process (#6138)
 
 ## v0.60.0 Beta
--- a/service/telemetry_test.go
+++ b/service/telemetry_test.go
@@ -1,0 +1,69 @@
+// Copyright  The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package service
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.opentelemetry.io/collector/component"
+	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
+	"go.opentelemetry.io/collector/service/telemetry"
+)
+
+func TestBuildTelAttrs(t *testing.T) {
+	buildInfo := component.NewDefaultBuildInfo()
+
+	// Check default config
+	cfg := telemetry.Config{}
+	telAttrs := buildTelAttrs(buildInfo, cfg)
+
+	assert.Len(t, telAttrs, 3)
+	assert.Equal(t, buildInfo.Command, telAttrs[semconv.AttributeServiceName])
+	assert.Equal(t, buildInfo.Version, telAttrs[semconv.AttributeServiceVersion])
+
+	_, exists := telAttrs[semconv.AttributeServiceInstanceID]
+	assert.True(t, exists)
+
+	// Check override by nil
+	cfg = telemetry.Config{
+		Resource: map[string]*string{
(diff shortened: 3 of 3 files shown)


component/build_info.go
```diff
@@ -25,6 +25,11 @@ type BuildInfo struct {
 
 	// Version string.
 	Version string
+
+	// ServiceName is the value of "service.name" attribute in the telemetry
+	// that the Collector produces about itself. This can be overridden
+	// by the end user via telemetry.Config.Resource setting.
+	ServiceName string
```
bogdandrutu, line 32: Why exposing it here? Why is not a config?

tigrannajaryan: It is also a config. I described in the PR description:

>I made the service.name a setting in BuildInfo which can be overridden when building the Collector. Other distros may choose a different service.name.
>The service.name can be also overridden by the end user via telemetry config setting.

tigrannajaryan:
@jpkrohling looks like builder is failing because it imports a specific version of the Collector instead of the latest code in local directory (and the previous version doesn't contain the newly added BuildInfo.ServiceName field). Should we change builder to import latest Collecor or there is some other thinking behind this that I am not aware?

component/build_info.go
bogdandrutu: Why do you want to have it in the BuildInfo then?

tigrannajaryan: The BuildInfo defines the default value of `service.name` if it is unspecified by the end user in the config file. This allows distributions to choose their own default value of `service.name`. Do you think it is unnecessary? Alternatively, we can hard-code the service.name to `io.opentelemtry.collector`. Is it what you are suggesting?

bogdandrutu: The BuildInfo is available to all components, is the intent to expose this to every component? If just to allow different distributions to configure we can do that in the `service.CollectorSettings`, but I don't think that is important. We should probably hardcode to `io.opentelemtry.collector` and see if any distribution wants to overwrite that and configuration is not enough.

tigrannajaryan: Sounds good. Let me change to hard-coded.

tigrannajaryan: Done.

service/telemetry.go
```diff
@@ -179,6 +165,37 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger
 	return nil
 }
 
+func buildTelAttrs(buildInfo component.BuildInfo, cfg telemetry.Config) map[string]string {
```
bogdandrutu, line 168: Separate concern, but we should look into that. When we "restart" the telemetry due to for example a config update, we generate a new instance ID which I think we should not? Let's open an issue to discuss that.

service/telemetry.go
```diff
@@ -179,6 +165,37 @@ func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger
...
+	for k, v := range cfg.Resource {
+		// nil value indicates that the attribute should not be included in the telemetry.
+		if v != nil {
+			telAttrs[k] = *v
+		}
+	}
+
+	if _, ok := cfg.Resource[semconv.AttributeServiceName]; !ok {
+		// AttributeServiceName is not specified in the config. Use the default service name.
+		telAttrs[semconv.AttributeServiceName] = defaultServiceName
```
bogdandrutu, line 180: Any idea/objection about using `BuildInfo.Command` as default instead of `io.opentelemetry.collector`? For us that will become `otelcol`.

service/telemetry.go
tigrannajaryan: Ideally, yes, the instance ID should persist. Issue created https://github.com/open-telemetry/opentelemetry-collector/issues/6158

service/telemetry.go
tigrannajaryan: Pros:
- `otelcol` is shorter and more readable.
Cons
- Different distros use the same `otelcol` BuildInfo.Command which can be confusing.
- Our own contrib repo uses `otelcontribcol` which it probably should not, it should use `otelcol` (?)

@open-telemetry/collector-approvers what do you think?

service/telemetry.go
bogdandrutu: Thanks

service/telemetry.go
```diff
@@ -57,6 +57,11 @@ const (
 	// supported trace propagators
 	traceContextPropagator = "tracecontext"
 	b3Propagator           = "b3"
+
+	// Default value of "service.name" resource attribute in the telemetry
+	// that the Collector produces about itself. This can be overridden
+	// by the end user via telemetry.Config.Resource setting.
+	defaultServiceName = "io.opentelemetry.collector"
```
jpkrohling, line 64: How can downstream distributions customize this?

service/telemetry.go
jpkrohling: I think we should differentiate between core and contrib. So, either `io.opentelemetry.collector.contrib` or `otelcol-contrib` for contrib would work for me, with a slight preference for the shorter version. On a UI that is not that wide, the shorter name would make it easier to identify whether the service relates to the core or the contrib distribution.

service/telemetry.go
bogdandrutu: Distributions can have their own "confmap.Converter` that can change the config. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/converter.go

jpkrohling: While that's a powerful tool, I believe this will be such a common thing to override that we might want to provide an easier way to customize it?

bogdandrutu: The build flag? that is very easy.

Also I don't think will be that common.

dmitryax: > Also they can have a build ld flag to change this constant correct?

Constants cannot be overridden with build flags AFIAK, only variables (even internal)

I also think it might be useful to have a cleaner way to override this in distros

bogdandrutu: We jump from 0 -> 1000 here. We did not have any car before now we want a 2M$ car. Let's see if we have this request before any more public API. We provided a way via config. Also another way is to do via config flags ("--set"), there are at least 3 different ways to achieve this, and we are asking for more.

Also we are talking about distros (2-3), for that if this is not the cleanest solution for the moment is ok, final users should have a clean solution and they have.

service/telemetry.go
bogdandrutu: The comment moved away from the initial discussion, reseting to get back to that. Can this default to `BuildInfo.Command`?

tigrannajaryan:
I think we have 3 open questions:
- Should different distributions use different service.name in the own telemetry?
- Should Core and Contrib builds use different service.name?
- Should we aim for short, more readable service.name like `otelcol` or for longer but more unambiguous, e.g. FQDN `io.opentelemetry.collector`?

Seems like we have different opinions here. Perhaps other @open-telemetry/collector-approvers and @open-telemetry/collector-contrib-approvers can chime in and suggest some arguments that help us answer these questions (i.e. provide some motivation why you think so)?

bogdandrutu:
> Should different distributions use different service.name in the own telemetry?

Who do you ask for? Every distro can do what they want. If a distro owner comes and ask me, I would suggest a different service name since is a different service (even though has lots of common things with the core service).

> Should Core and Contrib builds use different service.name?

I think so. 

> Should we aim for short, more readable service.name like otelcol or for longer but more unambiguous, e.g. FQDN io.opentelemetry.collector?

I've never seen a long name like `io.opentelemetry.collector`. I personally prefer up to 10 characters names, ideal 5-7.

service/telemetry.go
dmitryax: Given that `service.name` specification doesn't mention any uniqueness requirements, I'd vote for having a shorter name

djaglowski:
The specification says of `service.name`: `MUST be the same for all instances of horizontally scaled services.`

It also describes related fields, including `service.namespace`: `A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services.`

My interpretation of this is that `service.name` should be quite generic and `service.namespace` can optionally be used by distros and custom builds to provide additional specificity.

Applying this to Tigran's questions, I think `service.name` should be constant (eg `otelcol`), but we should support in some way that distros and custom builds can specify `service.namespace` in order to differentiate themselves.

bogdandrutu:
> Applying this to Tigran's questions, I think service.name should be constant (eg otelcol), but we should support in some way that distros and custom builds can specify service.namespace in order to differentiate themselves.

Why? I think the configuration in the end should belong to the final user and not to the distributor, since this is a service deployed/monitored by that user not by the distributor. If the distributor also deploys this for the final user they control the config so they can do whatever they want.

djaglowski:
> I think the configuration in the end should belong to the final user and not to the distributor, since this is a service deployed/monitored by that user not by the distributor. If the distributor also deploys this for the final user they control the config so they can do whatever they want.

Ok, fair point. 

I guess I'm confused about whether the `service` namespace should uniquely identify the binary, or if it is purely a concern of the final user to describe it sufficiently for their needs.

tigrannajaryan:
> we should support in some way that distros and custom builds can specify `service.namespace` in order to differentiate themselves.

We do support this already in the end user config. You can specify `service.namespace` (and any other attribute) in service.telemetry.resource config setting.

However, `service.namespace` is typically intended for grouping of team-wide services (e.g. all services that belong to Finance department), not for grouping of services by the distro. Although I can accept that such grouping may be a valid choice for some users I don't think we need to support it in any special way by encouraging distro owners to set their own `service.namespace`. I believe it is one of the attributes that truly belongs to the end users.

tigrannajaryan:
> > Should different distributions use different service.name in the own telemetry?
> 
> Who do you ask for? Every distro can do what they want. If a distro owner comes and ask me, I would suggest a different service name since is a different service (even though has lots of common things with the core service).

I am asking on behalf of the end user. What is the best from their perspective? If they happen to deploy Collectors from different distros do they want them to appear as the same service or as different services in their backend?

TylerHelmuth:
My two cents:

> Should different distributions use different service.name in the own telemetry?

Ultimately the user should have the ability to set whatever `service.name` they want and we should provide a default value.  I'd go as far as to say that users should be able to set any attributes they want on the collector's own telemetry. They are the ones that are going to use the telemetry to monitor their collectors, so we should let them do what they want with their data.

> Should Core and Contrib builds use different service.name?

Sure, but I don't have strong feelings on this at all.

> Should we aim for short, more readable service.name like otelcol or for longer but more unambiguous, e.g. FQDN io.opentelemetry.collector?

I'd vote shorter name as the default.

tigrannajaryan:
> Ultimately the user should have the ability to set whatever `service.name` they want and we should provide a default value. I'd go as far as to say that users should be able to set any attributes they want on the collector's own telemetry. They are the ones that are going to use the telemetry to monitor their collectors, so we should let them do what they want with their data.

Yes, that's exactly what we already do in this PR. We allow the user to override `service.name` and set any other attributes they want. The question is precisely about what the default value of `service.name` should be. (I updated my earlier comment to make it clear).

TylerHelmuth:
> Yes, that's exactly what we already do in this PR. We allow the user to override service.name and set any other attributes they want.

Yup, I'm agreeing with this PR's implementation.

As for actual default value, I'll put a vote in for `BuildInfo.Command` so we can gracefully achieve `otelcol` and `otelcol-contrib` without changing our build/release process.

tigrannajaryan:
I see a lot of support for using `BuildInfo.Command` as the default value for `service.name`. Anyone thinks this is a bad idea?

service/telemetry.go
tigrannajaryan: Changed to `BuildInfo.Command`.

TylerHelmuth (approved):

bogdandrutu (approved):

diego-lambrechts-wp:
Hi!. Can anyone please tell me in which section of the configuration the service.name can be set? Unfortunately I had no luck finding it in the documentation. Thank you!

PR #6163, opened 2022-09-27
Fully overrides default slice fields of config (#4001)

mcdoker18:
**Description:** 

mapstructure library doesn't override full slice during unmarshalling. Origin issue: https://github.com/mitchellh/mapstructure/issues/74#issuecomment-279886492 To address this we zeroes every slice before unmarshalling unless user provided slice is nil.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4001

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -168,6 +168,7 @@ func decodeConfig(m *Conf, result interface{}, errorUnused bool) error {
 			mapstructure.StringToTimeDurationHookFunc(),
 			mapstructure.TextUnmarshallerHookFunc(),
 			unmarshalerHookFunc(result),
+			zeroesSliceHookFunc(),
 		),
 	}
 	decoder, err := mapstructure.NewDecoder(dc)
@@ -317,6 +318,23 @@ func marshalerHookFunc(orig interface{}) mapstructure.DecodeHookFuncValue {
 			return nil, err
 		}
 		return conf.ToStringMap(), nil
+	}
+}
+
+// mapstructure library doesn't override full slice during unmarshalling.
+// Origin issue: https://github.com/mitchellh/mapstructure/issues/74#issuecomment-279886492
+// To address this we empty every slice before unmarshalling unless user provided slice is nil.
+func zeroesSliceHookFunc() mapstructure.DecodeHookFuncValue {
+	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
+		if from.Kind() == reflect.Slice && from.IsNil() {
+			return from.Interface(), nil
+		}
+
+		if to.CanSet() && to.Kind() == reflect.Slice {
+			to.Set(reflect.MakeSlice(reflect.SliceOf(to.Type().Elem()), 0, 0))
+		}
+
+		return from.Interface(), nil
 	}
 }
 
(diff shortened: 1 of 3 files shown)


confmap/confmap_test.go
```diff
@@ -334,3 +334,124 @@ func TestUnmarshalerErr(t *testing.T) {
...
+
+type (
+	innerStructWithSlices struct {
+		Ints []int `mapstructure:"ints"`
+	}
+	structWithSlices struct {
+		Strings []string              `mapstructure:"strings"`
+		Nested  innerStructWithSlices `mapstructure:"nested"`
+	}
+)
```
dmitryax, line 346: Nit: please move them inside the test function and remove the type grouping.

dmitryax (commented):
Thanks for addressing the issue! Overall LGTM. Just one nit regarding tests

confmap/confmap.go
```diff
@@ -239,6 +240,23 @@ func unmarshalerHookFunc(result interface{}) mapstructure.DecodeHookFuncValue {
 	}
 }
 
+// mapstructure library doesn't override full slice during unmarshalling.
+// Origin issue: https://github.com/mitchellh/mapstructure/issues/74#issuecomment-279886492
+// To address this we zeroes every slice before unmarshalling unless user provided slice is nil.
+func zeroesSliceHookFunc() mapstructure.DecodeHookFuncValue {
+	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
+		if from.Kind() == reflect.Slice && from.IsNil() {
```
bogdandrutu, line 248: Len() == 0? No need to re-initialize if empty?

mcdoker18: If user provides empty list in the config explicitly then we zeroes the default value of the field.

mcdoker18:
@bogdandrutu @dmitryax please can you take a look? All comments are fixed.

confmap/confmap.go
```diff
@@ -320,6 +321,23 @@ func marshalerHookFunc(orig interface{}) mapstructure.DecodeHookFuncValue {
 	}
 }
 
+// mapstructure library doesn't override full slice during unmarshalling.
+// Origin issue: https://github.com/mitchellh/mapstructure/issues/74#issuecomment-279886492
+// To address this we zeroes every slice before unmarshalling unless user provided slice is nil.
```
dmitryax, line 326:
```suggestion
// To address this we empty every slice before unmarshalling unless user provided slice is nil.
```

mcdoker18:
@dmitryax fixed

dmitryax (approved):
LGTM

confmap/confmap.go
codeboten: @bogdandrutu do you mean if both source and destination are 0 then we should just return?

codeboten (approved):
The PR looks ok to me, will merge once @bogdandrutu's comments have been addressed

confmap/confmap.go
bogdandrutu: Lets start with this:
`nil != initialized and len 0`

```yaml
foo:
```

```yaml
foo: []
```

What is the behavior we want in these cases? Please document them somewhere, in order for me to understand if the implementation is ok.

bogdandrutu: Ping on this.

Issue #6154, opened 2022-09-27
Inconsistent Clone vs CopyTo usage


bogdandrutu:
Proposal to replace the `[Metrics|Logs|Traces].Clone` methods with `[Metrics|Logs|Traces].CopyTo`.

PR #6164, opened 2022-09-27
Deprecate pdata Clone methods in favor of CopyTo

dmitryax:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/6154


bogdandrutu (approved):
There should be some usages of these funcs?

PR #6170, opened 2022-09-28
Remove usages of deprecated pdata Clone methods

dmitryax:


processor/batchprocessor/splitlogs_test.go
```diff
@@ -156,34 +156,3 @@ func TestSplitLogsMultipleILL(t *testing.T) {
 	assert.Equal(t, "test-log-int-0-0", split.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(0).SeverityText())
 	assert.Equal(t, "test-log-int-0-4", split.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(4).SeverityText())
 }
-
-func BenchmarkSplitLogs(b *testing.B) {
```
codeboten, line 160: Can you explain why the split tests have been removed? It's not super clear from the PR

dmitryax: Just to avoid reworking it. The benchmark tests were added to highlight the performance improvement of batch processor refactoring. I don't think it's needed anymore

bogdandrutu (approved):

processor/batchprocessor/splitlogs_test.go
bogdandrutu: I agree, this is a lot of the time the case for a one time improvement benchmarks that we never clean them

codeboten (approved):
Thanks for the response @dmitryax, LGTM

Issue #5476, opened 2022-06-05
[pdata] Map.Delete mutates values that referenced through Map.Get


dmitryax:
`Map.Delete` doesn't work as a clean map interface and mutates values returned by `Map.Get`. The following snippet

```go
	m := pcommon.NewMap()
	m.InsertString("k1", "v1")
	m.InsertString("k2", "v2")
	v1, _ := m.Get("k1")
	fmt.Println(v1.StringVal())
	m.Remove("k1")
	fmt.Println(v1.StringVal())
```

prints

```
v1
v2
```

instead of expected

```
v1
v1
```

dmitryax:
@bogdandrutu I believe this is a side-affect of this change https://github.com/open-telemetry/opentelemetry-collector/pull/2017 . Not sure if it warrants bringing back the following Map representation

```
type Map struct {
	orig *[]*otlpcommon.KeyValue
}
```

I would say we should. WDYT?

bogdandrutu:
This is true for other slices, and I think we specifically told (was there at one point at least if I remember) in the description that instances returned by `At` and `Get` will be invalid if Remove or even add is called.

bogdandrutu:
@dmitryax without that optimization we spend lots and lots of CPU allocating 1000s of small objects :))

tigrannajaryan:
BTW, this and other optimization won't be necessary and won't cause problems like this if we move to https://github.com/splunk/exp-lazyproto :-)

dmitryax:
Removing from 1.0 milestone since fixing it is not a breaking change for the pdata API

Issue #5588, opened 2022-06-23
[pcommon.Map] Removal then Insert results in wrong value assigned.


MovieStoreGuy:
**Describe the bug**

Removing an attribute from `pcommon.Map` that is not the end of the internal list results in overwriting keys that it shouldn't be.
It only happens with a specific operation order to trigger this.

**Steps to reproduce**
If possible, provide a recipe for reproducing the error.

```golang
import (
	"fmt"

	"go.opentelemetry.io/collector/pdata/pcommon"
)

func main() {
	attrs := pcommon.NewMapFromRaw(map[string]interface{}{
		"instrumentation.version": "v0.0.0-local",
		"vendor":                  "none",
		"hostname":                "workstation-1",
	})

	fmt.Println("The map is:", attrs.AsRaw())
	v, _ := attrs.Get("instrumentation.version")
	if attrs.Remove("instrumentation.version") {
		fmt.Println("successfully removed instrumentation.version attr")
	}
	attrs.Insert("auto-instrumentation", v)

	fmt.Println("The map is:", attrs.AsRaw())
}
```

Here I pick a value that would be in the middle of a sorted list, remove it, validate that it was removed successfully and then insert an updated value for it. 

**What did you expect to see?**

The final map should be shown as:

```
The map is: map[hostname:workstation-1 auto-instrumentation:v0.0.0-local vendor:none]
```

**What did you see instead?**
A clear and concise description of what you saw instead.

The resulting map is this:
```
The map is: map[auto-instrumentation:workstation-1 hostname:workstation-1 vendor:none]
```

**What version did you use?**
The mod file that I have is this:

```
module attrs-example

go 1.18

require go.opentelemetry.io/collector/pdata v0.54.0

require (
	github.com/gogo/protobuf v1.3.2 // indirect
	github.com/golang/protobuf v1.5.2 // indirect
	golang.org/x/net v0.0.0-20201021035429-f5854403a974 // indirect
	golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4 // indirect
	golang.org/x/text v0.3.3 // indirect
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 // indirect
	google.golang.org/grpc v1.47.0 // indirect
	google.golang.org/protobuf v1.28.0 // indirect
)
```

**What config did you use?**
N/A

**Environment**
N/A

**Additional context**
[...]

PR #2017, opened 2020-10-27
Make all KeyValue non-nullable

bogdandrutu:
Good win in the amount of allocations, and some nice win for the latency. In the current benchmarks only one attribute per resource, one attribute per span, one attribute per link, and one attribute per event.

Before:
```
goos: darwin
goarch: amd64
pkg: go.opentelemetry.io/collector/consumer/pdata
BenchmarkTracesFromOtlp
BenchmarkTracesFromOtlp-16    	     686	   1759476 ns/op	  981565 B/op	   35113 allocs/op
PASS
```

After:
```
goos: darwin
goarch: amd64
pkg: go.opentelemetry.io/collector/consumer/pdata
BenchmarkTracesFromOtlp
BenchmarkTracesFromOtlp-16    	     666	   1720189 ns/op	  940353 B/op	   29961 allocs/op
PASS
```

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #6205, opened 2022-09-29
[pdata] Make sure Map.Delete/Put doesn't affect other values

dmitryax:
This change reverts https://github.com/open-telemetry/opentelemetry-collector/pull/2017 in order to fix https://github.com/open-telemetry/opentelemetry-collector/issues/5476 and https://github.com/open-telemetry/opentelemetry-collector/issues/5588

During one of the previous SIG meetings, we discussed that we want to address the unexpected behavior and accept the performance hit until we move to custom proto serializer like https://github.com/splunk/exp-lazyproto


bogdandrutu:
Let's park this for a bit.

dmitryax:
@bogdandrutu sounds good. We also have a similar issue for slices https://github.com/open-telemetry/opentelemetry-collector/issues/6208

If we agree to merge this, I think we should do the same for slices

Issue #816, opened 2020-04-09
Migrate collector's metrics from OC to OpenTelemetry


pavolloffay:
Currently, the collector uses OpenCensus API for internal observability https://github.com/open-telemetry/opentelemetry-collector/blob/0e1b2e323d398cf6cda500953c7896aa01c173b7/observability/observability.go#L25. 

These metrics should be probably migrated to OpenTelemetry API.

Jaeger storage components use abstract metrics API and so we need to export these metrics in Jaeger OTEL collector e.g.

https://github.com/jaegertracing/jaeger/blob/bf870d9e2bffaf4758045919446be75ac9fc9783/cmd/opentelemetry-collector/app/exporter/cassandra/exporter.go#L31.

Are there plans to migrate to OpenTelemetry metrics? In Jaeger we want to avoid  first creating OC adapter and then OTEL once this repo migrates. I can submit a PR to port the metrics to OTEL API.

yurishkuro:
This would also provide the needed dogfooding for OTel SDK.

bogdandrutu:
@yurishkuro completely agree, we are waiting for 2 things:
1. The ability to use labels from the `correlation-context`.
1. The ability to configure `Views` - allow us to configure which labels from the `correlation-context` to use.

If we don't have these two things we need to manually propagate/retrieve these values every time we do a report. Which may be possible if we do have couple of helper functions. So probably the main missing feature is to support changing the default aggregations for some `instruments` like for distributions we don't want just minmaxsumcount we want a histogram.

+ @jmacd to help correct me if we can already do the histogram thing.

yurishkuro:
Why do you need context labels in the collector? Sounds a bit like nice to have, not a showstopper.

bogdandrutu:
I think I also tried to explain myself that we have the solution to use helper functions, but let me describe the design a bit:
* In the pipeline we want to record at every step metric broken down by the "receiver" so we add receiver to the context to propagate it in the pipeline instead of having it as an explicit argument. So we can have helper functions that just retrieve that from the context and use it in the metrics as labels.

And as I explained is not a showstopper, the current blocker is the ability to configure the aggregation for the distribution to be histogram vs minmaxsumcount, and probably a nice to have is the ability to select the labels for the aggregator (again not a showstopper).

bogdandrutu:
@pavolloffay to answer your question, we do want to use opentelemetry metrics. probably tracing as well :)

[...]

PR #6222, opened 2022-10-03
Instrument obsreport.receiver metrics with otel go

paivagustavo:
**Description:** Instruments existing `obsreport.Receiver` metrics  with otel-go, side-by-side with the existing opencensus metrics.

This uses the same `UseOtelForInternalMetricsfeatureGateID` feature gate, adding a `UseOtelForMetrics()` method on the `obsreportconfig` that should be used by components of the collector core to check what sdk should be used for instrumentation.

**Link to tracking Issue:** part of https://github.com/open-telemetry/opentelemetry-collector/issues/816

**Testing:** Manual tests comparing the current exported metrics and the metrics from the otel-go sdk after enabling the feature gate `--feature-gates=telemetry.useOtelForInternalMetrics`


internal/obsreportconfig/obsmetrics/obs_receiver.go
```diff
@@ -83,4 +87,40 @@ var (
 		ReceiverPrefix+RefusedLogRecordsKey,
 		"Number of log records that could not be pushed into the pipeline.",
 		stats.UnitDimensionless)
+
+	otelMeter     = global.Meter(MeterName)
+	AttrReceiver  = attribute.Key(ReceiverKey)
+	AttrTransport = attribute.Key(TransportKey)
```
bogdandrutu, line 93: Why using the global meter?

internal/obsreportconfig/obsmetrics/obs_receiver.go
```diff
@@ -83,4 +87,40 @@ var (
...
+	OtelReceiverAcceptedLogRecords, _ = otelMeter.SyncInt64().Counter(
+		ReceiverPrefix+AcceptedLogRecordsKey,
+		instrument.WithDescription("Number of log records successfully pushed into the pipeline."),
+		instrument.WithUnit(unit.Dimensionless),
+	)
+	OtelReceiverRefusedLogRecords, _ = otelMeter.SyncInt64().Counter(
+		ReceiverPrefix+RefusedLogRecordsKey,
+		instrument.WithDescription("Number of log records that could not be pushed into the pipeline."),
+		instrument.WithUnit(unit.Dimensionless),
+	)
```
bogdandrutu, line 125: these can be defined in the obsreport package now, since we don't need to share them between other internal packages.

obsreport/obsreport_receiver.go
```diff
@@ -163,23 +175,42 @@ func (rec *Receiver) endOp(
...
+			switch dataType {
+			case config.TracesDataType:
+				acceptedMeasure = obsmetrics.OtelReceiverAcceptedSpans
+				refusedMeasure = obsmetrics.OtelReceiverRefusedSpans
+			case config.MetricsDataType:
+				acceptedMeasure = obsmetrics.OtelReceiverAcceptedMetricPoints
+				refusedMeasure = obsmetrics.OtelReceiverRefusedMetricPoints
+			case config.LogsDataType:
+				acceptedMeasure = obsmetrics.OtelReceiverAcceptedLogRecords
+				refusedMeasure = obsmetrics.OtelReceiverRefusedLogRecords
```
bogdandrutu, line 189: extract separate funcs "recordWithOtel" and "recordWithOC" to help with readability.

obsreport/views.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+var storage = views{
+	v: []otelview.View{},
+}
+
+// AddViews to the OpenTelemetry Go SDK.
+// Views are only configurable at the SDK initialization, for this reason
+// components must call this function inside an init() function to be correctly initialized.
+func AddViews(v ...otelview.View) {
```
bogdandrutu, line 47: Why do you want to expose this functionality? What is the purpose of it?

bogdandrutu (commented):
We need this https://github.com/open-telemetry/opentelemetry-collector/pull/5939, which is to not break the metrics names at this time. We will at one point later revisit all names.

obsreport/views.go
bogdandrutu: This can/should be added later. There are other nicer alternatives like allowing Factories to implement additional interfaces for this, or doing it in the NewFactory helper as optionals, etc.

paivagustavo: I'll just remove this for now as this was just here as a preparation step migrations of some other components, thanks for sharing some nicer options as well!

obsreport/obsreport_receiver.go
paivagustavo: done.

internal/obsreportconfig/obsmetrics/obs_receiver.go
paivagustavo: I've moved those to be created inside of `obsreport.Receiver` using the `ReceiverCreateSettings.MeterProvider`.

obsreport/obsreport_receiver.go
```diff
@@ -63,7 +87,70 @@ func NewReceiver(cfg ReceiverSettings) *Receiver {
...
+
+func (rec *Receiver) createOtelMetrics() {
+	var err error
+	handleError := func(metricName string, err error) {
+		if err != nil {
+			rec.logger.Warn("failed to create otel instrument", zap.Error(err), zap.String("metric", metricName))
+		}
 	}
+
+	rec.acceptedSpansCounter, err = rec.meter.SyncInt64().Counter(
```
jaronoff97, line 113: thought: I wonder if these should exist in a list/map that are then just accessed as needed. i.e.

```
        telemetryMap = map[config.DataType]map[string]*syncint64.Counter
        // iterate through [DataType] x ["accepted", "refused"] to initialize
...
	telemetryMap[dataType]["accepted"].Add(receiverCtx, int64(numAccepted), rec.otelAttrs...)
	telemetryMap[dataType]["refused"].Add(receiverCtx, int64(numRefused), rec.otelAttrs...)
```

jaronoff97 (approved):
Just one thought, otherwise looks great!

paivagustavo:
> We need this https://github.com/open-telemetry/opentelemetry-collector/pull/5939, which is to not break the metrics names at this time. We will at one point later revisit all names.

@bogdandrutu  this pr actually fixes the name inconsistencies wrapping the `prometheus.Registry`, this might be easier than trying add `otelcol_` prefix to all metrics (including contrib metrics):
```
prometheus.WrapRegistererWithPrefix("otelcol_", prometheus.WrapRegistererWith(promLabels, registry))
```

If we still wants to move on with #5939, I can create a copy of `obsmetrics.ReceiverPrefix`, so this won't make `otelcol_otelcol_` metrics and fix this after that get merged.

internal/obsreportconfig/obsmetrics/obsmetrics.go
```diff
@@ -19,4 +19,10 @@ package obsmetrics // import "go.opentelemetry.io/collector/internal/obsreportco
 
 const (
 	NameSep = "/"
+
+	MeterName = "io.opentelemetry.collector"
```
bogdandrutu, line 23: Why this name? Do we have the same name for receivers/exporters? How does this affect the metrics exposed to Prometheus? /cc @dashpole 

In order to be a no-op change between OC -> Otel for the moment, I think we should have an empty name for the moment. Then when we have the final list of the metrics, we should agree on what scope name we use.

obsreport/obsreport.go
```diff
@@ -17,10 +17,17 @@ package obsreport // import "go.opentelemetry.io/collector/obsreport"
...
+	"go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 )
 
 func recordError(span trace.Span, err error) {
 	if err != nil {
 		span.SetStatus(codes.Error, err.Error())
 	}
 }
+
+func InstrumentWithOtel() bool {
```
bogdandrutu, line 31: Why public? Who is the user of this?

obsreport/obsreport_receiver.go
bogdandrutu: Accessing a map is slower than direct access. So I prefer this way.

service/telemetry.go
```diff
@@ -213,15 +221,39 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
...
 	registry := prometheus.NewRegistry()
-	if err := registry.Register(exporter.Collector); err != nil {
+
+	promLabels := make(prometheus.Labels)
+	for k, v := range attrs {
+		promLabels[sanitizePrometheusKey(k)] = v
+	}
+
+	wrappedRegisterer := prometheus.WrapRegistererWithPrefix("otelcol_", prometheus.WrapRegistererWith(promLabels, registry))
+	if err := wrappedRegisterer.Register(exporter.Collector); err != nil {
```
bogdandrutu, line 256: Let's split this change in a separate PR.

paivagustavo: Created https://github.com/open-telemetry/opentelemetry-collector/pull/6223 for this.

internal/obsreportconfig/obsmetrics/obsmetrics.go
paivagustavo: This can be removed since I'm no longer using the global meter and using the `ReceiverCreateSettings.MeterProvider` instead, there I'm using the same name as the trace name which is `cfg.ReceiverID.String()`.


> In order to be a no-op change between OC -> Otel for the moment, I think we should have an empty name for the moment.

I think this does not matter for the prometheus exporter,  as the otel scope is ignored now and will be added as additional attributes and metrics per https://github.com/open-telemetry/opentelemetry-specification/pull/2703

obsreport/obsreport.go
paivagustavo: I planned this to be public for all components that want to double instrument with otel-go, but as I've discovered the  `TelemetrySettings` struct, that seems to be a more suitable place, what do you think about that?

jpkrohling (commented):
This looks sane to me,  a review from @Aneurysm9 would be helpful though.

component/telemetry.go
```diff
@@ -33,6 +33,10 @@ type TelemetrySettings struct {
 	// MeterProvider that the factory can pass to other instrumented third-party libraries.
 	MeterProvider metric.MeterProvider
 
+	// UseOtelForMetrics controls whether the collector should use otel go sdk for
+	// internal telemetry metrics.
+	UseOtelForMetrics bool
```
bogdandrutu, line 38: Why exposing this?

bogdandrutu (changes requested):
Current PR exposes new temporary public APIs because the OC bridge is not ready and we cannot really start the proper migration to otel. 

I am not OK doing that.

paivagustavo:
@bogdandrutu  are you concerned about having this public `TelemetrySettings.UseOtelForMetrics`? Would it be better if we get rid of that and just started to record values to both sdks at the same time? Something like this:
```go
rec.recordWithOtel(receiverCtx, dataType, numAccepted, numRefused)
rec.recordWithOC(receiverCtx, dataType, numAccepted, numRefused)
```

I'm trying to understand what other paths we might have other than just waiting for the bridge to get ready.

bogdandrutu:
@paivagustavo to give you alternatives, I need to understand the goal :)

component/telemetry.go
paivagustavo: I've moved it to the internal package `obsreportconfig` so we can use it to the collector core component with otel.

internal/obsreportconfig/obsreportconfig.go
```diff
@@ -20,9 +20,21 @@ import (
...
 	"go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 )
 
+const (
+	// UseOtelForInternalMetricsfeatureGateID is the feature gate ID that controls whether the collector uses open
+	// telemetrySettings for internal metrics.
+	UseOtelForInternalMetricsfeatureGateID = "telemetry.useOtelForInternalMetrics"
+)
+
+// UseOtelForMetrics controls whether the Collector should use Otel for Collector's internal instrumentation.
```
codeboten, line 33: nit: this method doesn't control anything but reports it 

```suggestion
// UseOtelForMetrics returns whether the Collector should use Otel for Collector's internal instrumentation.
```

obsreport/obsreport_receiver.go
```diff
@@ -36,6 +42,19 @@ type Receiver struct {
 	longLivedCtx   bool
 	mutators       []tag.Mutator
 	tracer         trace.Tracer
+	meter          metric.Meter
+	logger         *zap.Logger
+
+	// Otel API
```
codeboten, line 48: Is this comment necessary here?

obsreport/obsreport_receiver.go
```diff
@@ -63,7 +83,70 @@ func NewReceiver(cfg ReceiverSettings) *Receiver {
...
+		logger: cfg.ReceiverCreateSettings.Logger,
+
+		useOtelForMetrics: obsreportconfig.UseOtelForMetrics(),
+		otelAttrs: []attribute.KeyValue{
+			attribute.String(obsmetrics.ReceiverKey, cfg.ReceiverID.String()),
+			attribute.String(obsmetrics.TransportKey, cfg.Transport),
+		},
 	}
+
+	rec.createOtelMetrics()
```
codeboten, line 96: should this be behind a check for using otel metrics?

codeboten (commented):
Just a couple of questions

service/telemetry.go
```diff
@@ -85,7 +81,7 @@ func init() {
 // registerInternalMetricFeatureGate registers the Internal Metric feature gate to the passed in registry
 func registerInternalMetricFeatureGate(registry *featuregate.Registry) {
 	registry.MustRegister(featuregate.Gate{
-		ID:          useOtelForInternalMetricsfeatureGateID,
+		ID:          obsreportconfig.UseOtelForInternalMetricsfeatureGateID,
```
Aneurysm9, line 84: This registration should move to `obsreportconfig` if the ID constant is moving there as well.

internal/obsreportconfig/obsreportconfig.go
Aneurysm9: I don't think this function is even needed since the feature gate ID constant is also exported.  It's easy enough for consumers to check the gate themselves and also hopefully avoids hiding the fact that a gate is being checked so it isn't called in a tight loop accidentally.

obsreport/obsreport_receiver.go
Aneurysm9: Or the check should be at the top of the method.  One of the two.

obsreport/obsreport_receiver.go
```diff
@@ -163,23 +246,11 @@ func (rec *Receiver) endOp(
...
-		case config.TracesDataType:
-			acceptedMeasure = obsmetrics.ReceiverAcceptedSpans
-			refusedMeasure = obsmetrics.ReceiverRefusedSpans
-		case config.MetricsDataType:
-			acceptedMeasure = obsmetrics.ReceiverAcceptedMetricPoints
-			refusedMeasure = obsmetrics.ReceiverRefusedMetricPoints
-		case config.LogsDataType:
-			acceptedMeasure = obsmetrics.ReceiverAcceptedLogRecords
-			refusedMeasure = obsmetrics.ReceiverRefusedLogRecords
+		if rec.useOtelForMetrics {
```
Aneurysm9, line 249: Maybe hide this behind a `rec.record()` method that does the switching.

obsreport/obsreport_receiver.go
paivagustavo: I've decided to add a check at the top.

obsreport/obsreport_receiver.go
paivagustavo: Created a `recordMetrics` to hide it and also to not be mistaken with the `if span.IsRecording() ` line below which is not refer a different type of "record"

service/telemetry.go
paivagustavo: done.

obsreport/obsreport_receiver.go
paivagustavo: Not really 🙃

internal/obsreportconfig/obsreportconfig.go
paivagustavo: Just removed it.

paivagustavo:
Thanks for the reviews @bogdandrutu @codeboten @Aneurysm9, I think I've addressed all the comments at this point and it should be ready to be reviewed again.

codeboten (approved):
Thanks for addressing my comments 👍🏻

Aneurysm9 (approved):

internal/obsreportconfig/obsreportconfig.go
```diff
@@ -20,9 +20,30 @@ import (
...
+func init() {
+	// register feature gate
+	RegisterInternalMetricFeatureGate(featuregate.GetRegistry())
+}
+
+// RegisterInternalMetricFeatureGate registers the Internal Metric feature gate to the passed in registry
+func RegisterInternalMetricFeatureGate(registry *featuregate.Registry) {
+	registry.MustRegister(featuregate.Gate{
+		ID:          UseOtelForInternalMetricsfeatureGateID,
+		Description: "controls whether the collector to uses OpenTelemetry for internal metrics",
```
evantorrie, line 42:
```suggestion
		Description: "controls whether the collector uses OpenTelemetry for internal metrics",
```

obsreport/obsreport_receiver.go
```diff
@@ -63,7 +83,74 @@ func NewReceiver(cfg ReceiverSettings) *Receiver {
 			tag.Upsert(obsmetrics.TagKeyTransport, cfg.Transport, tag.WithTTL(tag.TTLNoPropagation)),
 		},
 		tracer: cfg.ReceiverCreateSettings.TracerProvider.Tracer(cfg.ReceiverID.String()),
+		meter:  cfg.ReceiverCreateSettings.MeterProvider.Meter(cfg.ReceiverID.String()),
```
bogdandrutu, line 86: Not sure if this is the right name for the meter. Should we have a meter per factory type of per factory component instance?

paivagustavo: Do you mean to use `receiver/otlp` for the `otlp` receiver's meter name instead of just `otlp`, to differentiate components that have the same name? 

If that is what you meant, I think that seems better. Should tracer name be like that as well?

internal/obsreportconfig/obsreportconfig.go
```diff
@@ -103,6 +116,26 @@ func allViews() []*view.View {
 	return views
 }
 
+func receiverViews() []*view.View {
+	if featuregate.GetRegistry().IsEnabled(UseOtelForInternalMetricsfeatureGateID) {
```
paivagustavo, line 120: kinda misc, but I've added featuregate check on the OC receiver views creation.

obsreport/obsreport_receiver.go
bogdandrutu: Nope, I mean to use `cfg.ReceiverID.Type()`, so if we have multiple instances of the `otlp` receiver we have same scope for this, and the instance  name is not in the scope name (probably should by an attribute).

Aneurysm9: I don't think `otlp` is an appropriate scope name.  Shouldn't the scope here be `go.opentelemetry.io/collector/obsreport` with attributes for `componentType=receiver` and `componentID=otlp` (or `componentID=otlp/foo` if a name exists on the receiver ID)?

bogdandrutu: That is also a possibility, but you are still suggesting the same thing, that component name should not be part of the scope (which was before).

I think I like scope.name=`go.opentelemetry.io/collector/obsreport/receiver` and have a `component.id=otlp` (or `component.id=otlp/foo`) since we use `.` mostly as attribute name separator in otel (see resource attributes for example).

internal/obsreportconfig/obsreportconfig.go
bogdandrutu: why? I don't think we should code this way, the contract is to register these views only if `UseOtelForInternalMetricsfeatureGateID`, then we protect the high level function we don't have to protect any sub-function.

paivagustavo: Because if we want to use `prometheus.Registry` as the bridge, we don't want OC to emit metrics that have being migrated to otel-go if the featuregate is enabled, but we would want all the others non-migrated oc metrics to be exported as normal, hence I just removed the `receiver` metrics for now.

Once all components of the core gets instrumented with otel-go, we can move this to be broader and not register any view at all when the featuregate is enabled, but for now I would suggest to just remove it component per component.

obsreport/obsreport_receiver.go
paivagustavo: By the way, we will encounter some warns/missing metrics due to some otel-go bugs at the moment. 

Because the go sdk doesn't return a valid instrument for "duplicate" metric creation results in invalid instruments in otel-go ([tracking issue](https://github.com/open-telemetry/opentelemetry-go/issues/3229
)), only the first receiver would have valid instruments and subsequent calls to the `obsreport.NewReceiver()` will have noop metrics.

I would suggest us to just add the `component.id` (`cfg.ReceiverID.String()`) as suffix for now, adding a TODO with the tracking issue and remove that once that gets fixed. Since the prometheus exporter doesn't use the scope for nothing yet, it wouldn't change anything.

obsreport/obsreport_receiver.go
```diff
@@ -63,7 +88,74 @@ func NewReceiver(cfg ReceiverSettings) *Receiver {
...
+func (rec *Receiver) createOtelMetrics() {
+	if !rec.useOtelForMetrics {
+		return
+	}
+
+	var err error
+	handleError := func(metricName string, err error) {
+		if err != nil {
+			rec.logger.Warn("failed to create otel instrument", zap.Error(err), zap.String("metric", metricName))
+		}
```
bogdandrutu, line 115: We need to return the error to the caller instead of logging. It should not be done in this PR, but we need to file an issue for this.

bogdandrutu (approved):

obsreport/obsreport_receiver.go
bogdandrutu: I would not do that, since that will be a breaking change. But fix was release today.

paivagustavo: Right, upgrading to the new released version fixed this.

bogdandrutu (approved):
Just for my sanity, please create a separate PR for the version updates, maybe also include the 1.11 updates as well.

PR #6223, opened 2022-10-03
Add prometheus metric prefix and constant service attributes to otel-go prometheus exporter

paivagustavo:
**Description:** This wraps `prometheus.Registry` with the `otelcol_` prefix and the `service_` attributes as these are not supported directly by the OpenTelemetry Go Prometheus Exporter.

This ensures that metrics exported by otel-go is consistent with the existing opencensus metrics.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/5882

**Testing:** Manual tested by adding a otel-go metric and testing this.

### DIFF
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -249,16 +249,8 @@ func (tel *telemetryInitializer) initOpenTelemetry(attrs map[string]string) (htt
 	)
 
 	registry := prometheus.NewRegistry()
-	// OpenTelemetry specifies that resource attributes must not be attached to prometheus metrics,
-	// to be compatible with existing OpenCensus instrumentation we need to add them as constant labels
-	// to the prometheus registry.
-	// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#resource-attributes-1
-	promLabels := make(prometheus.Labels)
-	for k, v := range attrs {
-		promLabels[sanitizePrometheusKey(k)] = v
-	}
 
-	wrappedRegisterer := prometheus.WrapRegistererWithPrefix("otelcol_", prometheus.WrapRegistererWith(promLabels, registry))
+	wrappedRegisterer := prometheus.WrapRegistererWithPrefix("otelcol_", registry)
 	if err := wrappedRegisterer.Register(exporter.Collector); err != nil {
 		return nil, fmt.Errorf("failed to register prometheus collector: %w", err)
 	}
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -15,6 +15,7 @@
 package service // import "go.opentelemetry.io/collector/service"
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -31,10 +32,12 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opentelemetry.io/contrib/propagators/b3"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	otelprom "go.opentelemetry.io/otel/exporters/prometheus"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/propagation"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
 	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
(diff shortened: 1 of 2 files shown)


service/telemetry.go
```diff
@@ -225,15 +225,21 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
...
 
 	exporter := otelprom.New()
 	tel.mp = sdkmetric.NewMeterProvider(sdkmetric.WithReader(exporter))
 
 	registry := prometheus.NewRegistry()
-	if err := registry.Register(exporter.Collector); err != nil {
+	promLabels := make(prometheus.Labels)
+	for k, v := range attrs {
+		promLabels[sanitizePrometheusKey(k)] = v
+	}
```
bogdandrutu, line 239: I think the "attrs" are actually "Resource" attributes and they should be configured accordingly to the `MeterProvider`?

paivagustavo: Agreed, but right now resource attributes are not exported by the prometheus exporter. To keep the otel prometheus metrics consistent with the existing oc prometheus we need that.

paivagustavo: I've added `attrs` as `Resource` attributes to the sdk (which will be useful if we want to support other formats) but if we want to keep the parity with the existing oc metrics we still need to add them as prometheus constant labels.

If you think we should drop it, i.e., losing `service_instance_id` and `service_version` from our prom metrics, I can update the pr.

service/telemetry.go
```diff
@@ -225,15 +228,34 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
...
+func (tel *telemetryInitializer) initOpenTelemetry(attrs map[string]string) (http.Handler, error) {
 	// Initialize the ocRegistry, still used by the process metrics.
 	tel.ocRegistry = ocmetric.NewRegistry()
 
+	var resAttrs []attribute.KeyValue
+	for k, v := range attrs {
+		resAttrs = append(resAttrs, attribute.String(k, v))
+	}
+
+	res, err := resource.New(context.TODO(), resource.WithAttributes(resAttrs...))
```
bogdandrutu, line 240: `context.Background` may be better here.

service/telemetry.go
```diff
@@ -225,15 +228,34 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
...
+		sdkmetric.WithResource(res),
+		sdkmetric.WithReader(exporter),
+	)
 
 	registry := prometheus.NewRegistry()
-	if err := registry.Register(exporter.Collector); err != nil {
+	promLabels := make(prometheus.Labels)
+	for k, v := range attrs {
+		promLabels[sanitizePrometheusKey(k)] = v
+	}
```
bogdandrutu, line 255: Per your comment, please add a TODO with a link to a tracking bug, to remove this when support for using Resource attributes into Prometheus exporter is added.

bogdandrutu (commented):
LGTM overall, waiting for last nits.

service/telemetry.go
bogdandrutu: See my next comment, for the moment add them, but let's track this bug. I expect this to be fixed for prometheus soon correct.

service/telemetry.go
paivagustavo: I've added a comment about this linking to the specs, there is no bug because that is the intended behavior as per [OpenTelemetry specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#resource-attributes-1) and [OpenMetrics specs](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems).

bogdandrutu: Sorry for the delay, I was thinking about this.

Here is my understanding:
* When exporting to prometheus and we have Resource we generate a target_info metric.
* If we monitor that with collector's `prometheusreceiver` that we recommend, then the target_info metric is added back as resource. When exported to other backends most likely they get to be labels or available somehow as labels.
* I believe then we don't need anymore these values as prometheus labels.

Does it make sense to you? Maybe we protect this by another `featuregate` (do not add resource attributes as labels, enabled by default).

paivagustavo: Yes, that makes sense. Do you think we should add this `featuregate` in this pr or just remove this constant labels for now and add it back later?

bogdandrutu: lets remove them for the moment, since everything is protected by the big featuregate to enable otel

paivagustavo: Done.

bogdandrutu (approved):

PR #6228, opened 2022-10-03
[exporter/otlp] split into its own module

### LINKED ISSUE #6195: Split `otlpexporter` on its own module

codeboten:
Fixes #6195


cmd/otelcorecol/builder-config.yaml
```diff
@@ -28,5 +28,6 @@ processors:
 
 replaces:
   - go.opentelemetry.io/collector => ../../
-  - go.opentelemetry.io/collector/semconv => ../../semconv
```
bogdandrutu, line 31: Should you not have a change around the exporter module at line 16?

bogdandrutu (commented):
I expect a change in the builder

cmd/otelcorecol/builder-config.yaml
codeboten: Good point, updated.

bogdandrutu: Same thing in the default config in the builder.

codeboten: i forgot to run `make ocb`. should probably add a check for this

exporter/otlpexporter/go.mod
```diff
@@ -0,0 +1,50 @@
+module go.opentelemetry.io/collector/exporter/otlpexporter
+
+go 1.19
```
bogdandrutu, line 3: 1.18?

bogdandrutu (approved):

exporter/otlpexporter/Makefile
```diff
@@ -0,0 +1 @@
+include ../../Makefile.Common
```
bogdandrutu, line 1: empty line please.

exporter/otlpexporter/go.mod
codeboten: 👍🏻

bogdandrutu:
I think we should send also a PR to remove https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/doc.go since it belongs (will) to a module without anything in that parent directory. Also think what we want to do with https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter#general-information since it feels strange that it belongs to a different module.

Maybe we create a module "exporter" with all the exporters?

exporter/otlpexporter/Makefile
codeboten: Done

codeboten:
> I think we should send also a PR to remove https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/doc.go since it belongs (will) to a module without anything in that parent directory

I agree this makes sense once all the other exporters are split into their own modules.

> Also think what we want to do with https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter#general-information since it feels strange that it belongs to a different module.

This could likely be moved under the docs directory in something called exporter.md? Maybe there should be a components directory under docs?

> Maybe we create a module "exporter" with all the exporters?

What would the result of someone including that module be? All the exporters get pulled in? I wonder if it would be confusing for users as it may be unclear what exporters fall under that package.

bogdandrutu:
> What would the result of someone including that module be? All the exporters get pulled in? I wonder if it would be confusing for users as it may be unclear what exporters fall under that package.

"All the exporters get pulled in" pulled in where? All the exporters are available in the module, if you talk about the builder you need 2 lines one for module and one for import.

bogdandrutu:
last one about this "structure" part:
I saw somewhere in contrib that `go mod` complains if you have a module `go.opentelemetry.io/collector/exporter` (in case in the future we want to use that package and module for some shared code between exporters) and `go.opentelemetry.io/collector/exporter/otlpexporter`. Can you check that?

bogdandrutu:
It looks to be related to this issue https://github.com/golang/go/issues/30590, found starting from the code :)) https://cs.opensource.google/go/go/+/master:src/cmd/go/testdata/script/mod_dot.txt;l=30;bpv=1;bpt=0

codeboten:
Another update to this PR here:
* moved the exporter to `exporters` directory. this is more inline with opentelemetry-go: https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters
* renamed `otlpexporter` to `otlpgrpcexporter` to make this more consistent with `otlphttpexporter`

Please take a look @open-telemetry/collector-approvers

dmitryax:
Renaming modules/packages is a significant breaking change for the API consumers. Can you please add deprecated aliases under the old paths?

dmitryax:
Also I'm still not convinced that we have significant enough reasons to rename `exporter` to `exporters`. Even if I like the plural more, I think we should clearly provide the reasoning behind that in this PR or an issue. Looks like the errors that @bogdandrutu posted are not relevant anymore?

exporters/otlpgrpcexporter/config.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package otlpexporter // import "go.opentelemetry.io/collector/exporter/otlpexporter"
+package otlpgrpcexporter // import "go.opentelemetry.io/collector/exporters/otlpgrpcexporter"
```
dmitryax, line 15: Do we rename the package only but keep `otlp` as the component name?

codeboten: `otlp` in the config applies to both grpc and http exporters, my opinion is that it's fine to separate config from package name, but i don't feel strongly about it. i can revert if that's preferable.

codeboten:
> Also I'm still not convinced that we have significant enough reasons to rename `exporter` to `exporters`. Even if I like the plural more, I think we should clearly provide the reasoning behind that in this PR or an issue. Probably better to have a separate PR for that. This one is just about otlp exporter. Looks like the errors that @bogdandrutu posted are not relevant anymore?

Since this is the first time we'll publish these components as separate modules, it's the time to decide on what makes the most sense. The plural does make more sense if it's going to contain more than one exporter, and as i mentioned it does align with otel-go.

That being said, I agree that deprecating/breaking the import path for users shouldn't be done lightly.

exporters/otlpgrpcexporter/config.go
dmitryax: > otlp in the config applies to both grpc and http exporters, 

otlphttpexporter uses `otlphttp` https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/factory.go#L33

I think we should keep module names consistent with the collector component names. I would vote for keeping `otlpexporter` without renaming it. This seems like the most important one

codeboten: sorry i was thinking of the receiver 😅

codeboten:
@dmitryax I opened an issue to discuss moving components to the plural subdirectory: https://github.com/open-telemetry/opentelemetry-collector/issues/6337

bogdandrutu:
> Looks like the errors that @bogdandrutu posted are not relevant anymore?

@dmitryax the errors I mentioned will be removed if we rename the package since, "exporter" package will be free to use.

> Also I'm still not convinced that we have significant enough reasons to rename exporter to exporters.

@dmitryax can you confirm that the errors I mentioned are ok long term?

exporters/otlpgrpcexporter/config.go
mx-psi: I think we should keep it as `otlpexporter`. It would otherwise be the only exporter with a mismatch between the Go module name and the configuration identifier.

codeboten:
Closing in favour of #6343

PR #6231, opened 2022-10-04
Add options to confmap Marshal/Unmarshal, deprecate UnmarshalExact

bogdandrutu:
Officially this can be considered breaking change because of a use in a `var foo func(interface{}) error = conf.Unmarshal`, not in a normal calling usage, but this is an edgecase that we can except and move forward with this.

This will allow to add more options to Marshal/Unmarshal in the future.

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -57,19 +57,52 @@ func (l *Conf) AllKeys() []string {
 	return l.k.Keys()
 }
 
-// Unmarshal unmarshalls the config into a struct.
-// Tags on the fields of the structure must be properly set.
-func (l *Conf) Unmarshal(result interface{}) error {
-	return decodeConfig(l, result, false)
+type UnmarshalOption interface {
+	apply(*unmarshalOption)
 }
 
-// UnmarshalExact unmarshalls the config into a struct, erroring if a field is nonexistent.
+type unmarshalOption struct {
+	errorUnused bool
+}
+
+// WithErrorUnused sets an option to error when there are existing
+// keys in the original Conf that were unused in the decoding process
+// (extra keys).
+func WithErrorUnused() UnmarshalOption {
+	return unmarshalOptionFunc(func(uo *unmarshalOption) {
+		uo.errorUnused = true
+	})
+}
+
+type unmarshalOptionFunc func(*unmarshalOption)
+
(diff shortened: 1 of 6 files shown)


confmap/confmap.go
```diff
@@ -57,19 +57,55 @@ func (l *Conf) AllKeys() []string {
...
+	// If ErrorUnused is true, then it is an error for there to exist
+	// keys in the original Conf that were unused in the decoding process
+	// (extra keys).
+	errorUnused bool
+}
+
+// WithErrorUnused  then it is an error for there to exist
+// keys in the original Conf that were unused in the decoding process
+// (extra keys).
+func WithErrorUnused() UnmarshalOption {
```
bogdandrutu, line 74: Question if this should accept a bool or we allow only to set. This feels ok to me as is.

confmap/confmap.go
```diff
@@ -57,19 +57,55 @@ func (l *Conf) AllKeys() []string {
...
+type unmarshalOption struct {
+	// If ErrorUnused is true, then it is an error for there to exist
+	// keys in the original Conf that were unused in the decoding process
+	// (extra keys).
+	errorUnused bool
+}
+
+// WithErrorUnused  then it is an error for there to exist
+// keys in the original Conf that were unused in the decoding process
+// (extra keys).
```
dmitryax, line 73: The comment is a pretty unclear. Can you please update it?

confmap/confmap.go
```diff
@@ -57,19 +57,55 @@ func (l *Conf) AllKeys() []string {
...
 
-// Unmarshal unmarshalls the config into a struct.
+type UnmarshalOption interface {
+	apply(*unmarshalOption)
+}
+
+type unmarshalOption struct {
+	// If ErrorUnused is true, then it is an error for there to exist
+	// keys in the original Conf that were unused in the decoding process
+	// (extra keys).
```
dmitryax, line 67: Not sure if this one needed, seems like a redundant copy

bogdandrutu: Done.

confmap/confmap.go
bogdandrutu: Done.

dmitryax (approved):

codeboten (approved):

PR #6236, opened 2022-10-04
Tentative way to fix the Unmarshal of the root

bogdandrutu:
Downside, if the root implements Unmarshaler and the caller calls `root.Unmarshal(conf)` then the logic in `root.Unmarshal(conf)` is called twice. Is it ok to expect that users should never call directly that interface and always go via the `conf.Unmarshal(root)`?

/cc @djaglowski


djaglowski:
> Is it ok to expect that users should never call directly that interface and always go via the conf.Unmarshal(root)?

This would be tough to enforce and the implications of not following the guidance would sometimes be subtle and go unnoticed. It's probably an improvement over the current limitation, but it seems like there might be a better way. That said, I tinkered with for quite a while now and have not been able to get a better result.

PR #6237, opened 2022-10-05
Add crosslink tool

bryan-aguilar:
**Description:** Adds the [crosslink](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/crosslink) tool to simplify the function of adding replace statements to intra-repository go modules. 


bryan-aguilar:
Can this PR get a skip changelog label?

Makefile
```diff
@@ -433,3 +434,10 @@ checklinks:
 	command -v markdown-link-check >/dev/null 2>&1 || { echo >&2 "markdown-link-check not installed. Run 'npm install -g markdown-link-check'"; exit 1; }
 	find . -name \*.md -print0 | xargs -0 -n1 \
 		markdown-link-check -q -c ./.github/workflows/check_links_config.json || true
+
+# error message "failed to sync logger:  sync /dev/stderr: inappropriate ioctl for device"
+# is a known issue but does not affect function. 
+.PHONY: crosslink
+crosslink: install-tools
```
bogdandrutu, line 441: we don't require install-tools in other places (unless I forget), just try to be consistent.

bogdandrutu (commented):
Lets make this step part of github action "check" and if git diffs then fail.

Makefile
bryan-aguilar: I believe there is some inconsistency. Multimod requires `install-tools`. But I will remove it to match the majority.

bryan-aguilar:
Updated GHA and make directive.

I also added the `--prune` flag as default. Since this will be running on PR build it should continue to clean the modules clean. You can see the details of the prune command [here](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/crosslink#prune---prune---p).

Issue #6238, opened 2022-10-05
Service TelemetryInitializer doesn't clean up when newService errors


cpheps:
**Describe the bug**
When the `newService` function returns an error, as a result of or after `srv.telemetryInitializer.init` executes, it doesn't clean up any telemetry views that were registered. The means if a new collector instance is started within the same process that would attempt to register those same views will fail with a `failed to initialize telemetry` error.

**Steps to reproduce**
Have a binary that wraps the collector and starts an instance of the collector. Start the wrapped instance with an invalid config. This will cause `newService` to return an error after initializing telemetry views. Without stopping the process start another instance of the collector with any config. `newService` will return get an error similar to:

```
failed to initialize telemetry: processor/batch/batch_send_size: cannot register view \"processor/batch/batch_send_size\"; a different view with the same name is already registered\nprocessor/batch/batch_send_size_bytes: cannot register view \"processor/batch/batch_send_size_bytes\"; a different view with the same name is already registered
```

**What did you expect to see?**
A separate instance of the collector able to be started within the same process with a valid config.

**What did you see instead?**
Failure of the telemetry initialization

**What version did you use?**
v0.61.0

**What config did you use?**
Config:

```yaml
receivers:
  hostmetrics:
    collection_interval: 1h
    scrapers:
      load:
      memory:

processors:
  batch:

exporters:
  logging:
   #invalid param to cause failure
    level: info

service:
  pipelines:
    metrics:
      receivers: [hostmetrics]
      processors: [batch]
      exporters: [logging]
```

**Environment**
OS: macOS 12.5.1
Compiler(if manually compiled): go version go1.19.1 darwin/arm64

**Additional context**
This seems to be a side effect of #6138, which made a `telemetryInitializer` per instance of the collector. The problem with this is on `telemetryInitializer.init` registers views globally to the process. Before #6138 the `telemetryInitializer` was a onetime use per process so views would never attempt to be registered multiple times.

I think there's two paths to consider based on intent of use for the collector:
1. Can a single process is meant to run several instances of the Collector concurrently?
2. If only a single instance of the Collector is meant to used at a time in a process should the `telemetryInitializer` act like it is tied to the lifetime of that instance?

For number 1 I think this is not the intended use case. It seems undefined how multiple instances of the collector could coexist, especially if they had configurations for internal telemetry that contradicted one another.

[...]

PR #6239, opened 2022-10-05
Fixed issue where telemetryInitializer is not cleaned up when newService errors

cpheps:
Fixing a bug - Fixed issue where the `telemetryInitializer` was not cleaned up if `newService` errors after `telemetryInitializer.init` is called. This would cause subsequent calls to new service to error on `telemetryInitializer.init`.

Fixes: #6238 

**Testing:**
- Added a new unit test that fails without new cleanup logic
- Tested with our OPAMP enabled [agent](https://github.com/observIQ/observiq-otel-collector) and [server](https://github.com/observiq/bindplane-op). On reconfigure our agent runs a new instance of the Collector within the agent process. This was exposing the issue when a bad config is sent.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -77,7 +77,7 @@ func newService(set *settings) (*service, error) {
 
 	// If there is any error after we initialize telemetry we need to ensure we clean that up for any future service
 	defer func() {
-		if returnErrs != nil {
+		if returnErrs == nil {
 			return
 		}
 
--- a/service/service.go
+++ b/service/service.go
@@ -71,6 +71,19 @@ func newService(set *settings) (*service, error) {
 	if err = srv.telemetryInitializer.init(set.BuildInfo, srv.telemetrySettings.Logger, set.Config.Service.Telemetry, set.AsyncErrorChannel); err != nil {
 		return nil, fmt.Errorf("failed to initialize telemetry: %w", err)
 	}
+
+	// Used to capture main error and any defer errors
+	var returnErrs error
+
+	// If there is any error after we initialize telemetry we need to ensure we clean that up for any future service
+	defer func() {
+		if returnErrs != nil {
+			if shutdownErr := srv.telemetryInitializer.shutdown(); shutdownErr != nil {
+				returnErrs = multierr.Append(returnErrs, fmt.Errorf("failed to shutdown collector telemetry: %w", shutdownErr))
+			}
+		}
+	}()
+
 	srv.telemetrySettings.MeterProvider = srv.telemetryInitializer.mp
 
 	extensionsSettings := extensions.Settings{
@@ -80,7 +93,8 @@ func newService(set *settings) (*service, error) {
 		Factories: srv.host.factories.Extensions,
 	}
 	if srv.host.extensions, err = extensions.New(context.Background(), extensionsSettings, srv.config.Service.Extensions); err != nil {
-		return nil, fmt.Errorf("failed build extensions: %w", err)
+		returnErrs = multierr.Append(returnErrs, fmt.Errorf("failed build extensions: %w", err))
+		return nil, returnErrs
 	}
 
 	pipelinesSettings := pipelines.Settings{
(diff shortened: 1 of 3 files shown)


service/service.go
```diff
@@ -71,6 +71,19 @@ func newService(set *settings) (*service, error) {
 	if err = srv.telemetryInitializer.init(set.BuildInfo, srv.telemetrySettings.Logger, set.Config.Service.Telemetry, set.AsyncErrorChannel); err != nil {
 		return nil, fmt.Errorf("failed to initialize telemetry: %w", err)
 	}
+
+	// Used to capture main error and any defer errors
+	var returnErrs error
+
+	// If there is any error after we initialize telemetry we need to ensure we clean that up for any future service
+	defer func() {
+		if returnErrs != nil {
```
djaglowski, line 80: If there is no error, are we inferring that `telemetryInitializer` was properly shutdown? 

If so, it seems like there might be a way to tighten up the way we manage shutting it down. Maybe there is a way to _only_ call it here?

Otherwise, i think it might be cleaner to split `newService` into two functions, where the inner function handles all the cases that can result in a returnErr. Instead, just return from the inner function and check the error in the outer func. Then we don't need to manage this with a defer.

service/service.go
```diff
@@ -71,6 +71,19 @@ func newService(set *settings) (*service, error) {
...
+			if shutdownErr := srv.telemetryInitializer.shutdown(); shutdownErr != nil {
+				returnErrs = multierr.Append(returnErrs, fmt.Errorf("failed to shutdown collector telemetry: %w", shutdownErr))
+			}
+		}
+	}()
```
djaglowski, line 85: Business logic aside, we should flatten this by returning early in the simple case.
```suggestion
	defer func() {
		if returnErrs == nil {
			return
		}
		if shutdownErr := srv.telemetryInitializer.shutdown(); shutdownErr != nil {
			returnErrs = multierr.Append(returnErrs, fmt.Errorf("failed to shutdown collector telemetry: %w", shutdownErr))
		}
	}()
```

service/service.go
cpheps: I think I'm miss understanding. The logic here is should be if there is an error it will be stored in `returnErrs`. The defer will execute after `returnErrs` is set but before the `return` so if we return due to an error we clean up and if we cleanly return we do nothing and keep the `telemetryInitalizer` running.

Let me know if you still have concerns as I might be missing what you're intending.

service/service.go
cpheps: Yeah I agree with this, nested error checking could be cleaned up. I'll make a change.

service/service.go
cpheps: Reading back through this a few times I think I understand your point. I think there is a way to restructure this function so it doesn't need a defer.

service/service.go
```diff
@@ -71,38 +71,17 @@ func newService(set *settings) (*service, error) {
...
-
-	extensionsSettings := extensions.Settings{
-		Telemetry: srv.telemetrySettings,
-		BuildInfo: srv.buildInfo,
-		Configs:   srv.config.Extensions,
-		Factories: srv.host.factories.Extensions,
-	}
-	if srv.host.extensions, err = extensions.New(context.Background(), extensionsSettings, srv.config.Service.Extensions); err != nil {
-		return nil, fmt.Errorf("failed build extensions: %w", err)
-	}
```
djaglowski, line 74: nit: The diff shows the `srv.telemetrySettings.MeterProvider = srv.telemetryInitializer.mp` line as changed, but really there is just an extra line here. Removing it should simplify the diff.

service/service.go
```diff
@@ -153,3 +132,40 @@ func (srv *service) Shutdown(ctx context.Context) error {
 	// TODO: Shutdown TracerProvider, MeterProvider, and Sync Logger.
 	return errs
 }
+
+func (srv *service) initPipeline(set *settings) error {
+	var err error
```
djaglowski, line 137: nit: Seems kind of unusual to declare this outside of any individual condition, since each error is returned immediately.

djaglowski (commented):
LGTM, just a couple nits

service/service.go
cpheps: This was copying the pattern that was in `newService` before I refactored it. This is done because in the assignments if there is no error it assigns to an already existing field on the `srv` object. I think the alternative would be to create tmp variables to assign to the `srv` fields but I don't think that's necessarily cleaner.

djaglowski (approved):
LGTM

Aneurysm9 (approved):

codeboten (approved):
Thanks for fixing!

mx-psi (approved):

service/service.go
```diff
@@ -73,36 +73,14 @@ func newService(set *settings) (*service, error) {
...
-	if srv.host.pipelines, err = pipelines.Build(context.Background(), pipelinesSettings); err != nil {
-		return nil, fmt.Errorf("cannot build pipelines: %w", err)
-	}
-
-	if set.Config.Telemetry.Metrics.Level != configtelemetry.LevelNone && set.Config.Telemetry.Metrics.Address != "" {
-		// The process telemetry initialization requires the ballast size, which is available after the extensions are initialized.
-		if err = telemetry.RegisterProcessMetrics(srv.telemetryInitializer.ocRegistry, getBallastSize(srv.host)); err != nil {
-			return nil, fmt.Errorf("failed to register process metrics: %w", err)
+	// process the configuration and initialize the pipeline
+	if err := srv.initPipeline(set); err != nil {
```
bogdandrutu, line 77: err already defined, this will "shadow". Don't use ":"

service/service.go
```diff
@@ -153,3 +131,40 @@ func (srv *service) Shutdown(ctx context.Context) error {
 	// TODO: Shutdown TracerProvider, MeterProvider, and Sync Logger.
 	return errs
 }
+
+func (srv *service) initPipeline(set *settings) error {
```
bogdandrutu, line 135: This also initialize the extensions. Maybe have rename to `initExtensionsAndPipelines`?

cpheps:
@bogdandrutu made the suggested changes

bogdandrutu (approved):

PR #6255, opened 2022-10-07
[pdata] Add documentation defining naming conventions

dmitryax:
The documentation is intended to summarize all the naming rules followed in pdata module

### DIFF
--- a/pdata/README.md
+++ b/pdata/README.md
@@ -1,0 +1,236 @@
+# Pipeline data (pdata)
+
+Pipeline data (pdata) implements data structures that represent telemetry data in-memory. All data received 
+is converted into this format, travels through the pipeline in this format, and is converted from this format by 
+exporters when sending.
+
+Current implementation primarily uses OTLP protobuf structs as the underlying data structures for many of the 
+declared structs. We keep a pointer to OTLP protobuf in the "orig" member field. This allows efficient translation 
+to/from OTLP wire protocol. The underlying data structure is kept private so that we are free to make changes to it 
+in the future.
+
+The pdata API is designed to avoid mutable data sharing and bugs that stem from that. Each pdata instance cannot 
+contain a reference to an object that is used in another pdata instance.
+
+## API naming convention
+
+### Package names
+
+Names of pdata packages don't follow names of the protobuf packages. The pdata has a package per telemetry type 
+starting with `p`, e.g. `ptrace`, and `pcommon` package which includes pdata API for protobuf definitions from 
+`common` and `resource` protobuf packages.
+
+### Protobuf message representation in pdata
+
+Pipeline data structs SHOULD be based on the names of the underlying OTLP protobuf messages. Data types for 
+protobuf messages defined as part of another message SHOULD include the owner's name as a prefix. The following 
+examples are two pdata structs based on protobuf messages defined at the package level and as part of another 
+message: 
+
+- `pmetric.NumberDataPoint` based on a `NumberDataPoint` protobuf message defined in 
+  [metrics.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto)
+  package.
+- `pmetric.ExponentialHistogramDataPointBuckets` based on a `Buckets` protobuf message defined in 
+  `ExponentialHistogramDataPoint` message of
+  [metrics.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto)
+  package. 
+
+Exceptions to the naming rules are possible, but not encouraged. Another name can be chosen for brevity or if the 
+struct provides a different interface to manipulate the underlying data. The following exceptions are currently 
+accepted:
+
+- `plog.Logs` based on `LogsData` protobuf message.
+- `pmetric.Metrics` based on `MetricsData` protobuf message.
+- `ptrace.Traces` based on `TracesData` protobuf message.
+- `pcommon.Slice` based on `ArrayValue` protobuf message.
+- `pcommon.Map` based on `KeyValueList` protobuf message.
+- `pcommon.Value` based on `AnyValue` protobuf message.
+
+Each pdata struct MUST have an initialization function starting with `New` prefix. Usage of zero-initialized values is 
+prohibited and can cause a panic. Each pdata struct MUST provide the following methods:
+
+- `MoveTo(<type>)`: moves all the data from one struct instance to another. The destination instance is overwritten and the 
+  source instance is re-initialized as a new empty instance.
+- `CopyTo(<type>)`: deep copies all the data from one struct instance to another. The destination instance is 
+  overwritten and the source instance is not modified.
+
+Each pdata struct based on a protobuf message SHOULD have getter methods for every protobuf field. Exceptions to
+this rule are allowed if exposing a field is not desirable in pdata. For example, a deprecated protobuf field MAY
+not be exposed in pdata. Also, pdata MAY provide a different interface to manipulate protobuf fields without
+exposing them explicitly as is done with `pcommon.Map`.
+
+### Protobuf fields representation in pdata
+
+#### Singular fields
+
+Name of a pdata getter methods representing singular protobuf fields SHOULD match the name the protobuf field but 
+written in CamelCase aligned with Go naming conventions. If a protobuf field is of scalar or enum type, the 
+corresponding pdata struct MUST provide a setter method with `Set` prefix.
+
+Some fields of scalar type in a protobuf message MAY be represented by another pdata type providing an additional 
+interface, for example `pcommon.Timestamp` is another type that wraps `uint64` value and provides an additional
+interface to work with timestamps. pdata fields returning `pcommon.Timestamp` don't follow the recommended naming
+schema, `Timestamp` word is used instead of `TimeUnixNano` to represent protobuf fields that contain
+`time_unix_nano`, for example `StartTimestamp` pdata field is used for `start_time_unix_nano` protobuf fields.
+
+#### Optional fields
+
+Each optional field in a protobuf message exposed in pdata MUST have an additional method with `Has` prefix to
+determine if the field is set.
+
+#### OneOf fields
+
+A pdata struct representing a protobuf message with an exposed oneof field MUST provide getter and setter methods for 
+each option of the oneof field. Name of each setter and getter SHOULD be called the same as the protobuf oneof field 
+options. The following exception is adopted in pdata for numeric oneof protobuf fields for brevity:
+
+```protobuf
+oneof value {
+  double as_double = 4;
+  sfixed64 as_int = 6;
+}
+```
+
+is represented in pdata in a shorter form of the following methods:
+
+```golang
+DoubleValue() float64
+SetDoubleValue(float64)
+IntValue() int64
+SetIntValue(v int64)
+```
+
+If a oneof field option is another protobuf message, the setter name MUST include `Empty` in its name. Such setter 
+MUST set the oneof field to an empty initialized struct and return it. The following conditions define whether the 
+name of the oneof field must be included in the setter and getter method names:
+
+1. If a oneof field represents the whole protobuf message, name of the oneof field itself MAY be omitted in setter 
+   and getter methods. For example:
+
+```go
+func (ms Metric) Sum() Sum
+func (ms Metric) SetEmptySum() Sum
+```
+
+2. If a oneof field is relevant only to a particular field of the message, the pdata getter and setter methods MUST 
+   include name of the oneof field along with name of the option. For example:
+
+```go
+func (ms NumberDataPoint) IntValue() int64
+func (ms NumberDataPoint) SetIntValue(v int64)
+```
+
+Additionally, the pdata struct MUST provide a type getter for every exposed oneof protobuf field. Name of the getter 
+MUST be `Type` for oneof fields representing the whole protobuf message (1), or `<oneof field>Type` for oneof fields 
+relevant only to a particular field. The function MUST return an `int32` enum type called `<struct name>Type` for (1) 
+and `<struct name><oneof field>Type` for (2). For example:
+
+- `func (ms Metric) Type() MetricType` (1)
+- `func (ms NumberDataPoint) ValueType() NumberDataPointValueType` (2)
+
+The enum constants MUST be called the same as the type with a suffix named after the oneof option, e.g.: 
+
+- `MetricTypeSum` (1)
+- `NumberDataPointValueTypeString` (2)
+
+An unset oneof protobuf value MUST be represented by a pdata constant with `Empty` suffix, e.g.:
+
+- `MetricTypeEmpty` (1)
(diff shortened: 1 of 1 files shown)


dmitryax:
@open-telemetry/collector-approvers PTAL

pdata/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `plog.Logs` based on `LogsData` protobuf message.
+- `pmetric.Metrics` based on `MetricsData` protobuf message.
+- `ptrace.Traces` based on `TracesData` protobuf message.
+- `pcommon.Slice` based on `ArrayValue` protobuf message.
+- `pcommon.Map` based on `KeyValueList` protobuf message.
+- `pcommon.Value` based on `AnyValue` protobuf message.
+
+Each pdata struct MUST have an initialization function starting with `New` prefix. Usage of zero-initialized is 
+prohibited and can case panic. Each pdata struct MUST provide the following methods:
```
mx-psi, line 32:
```suggestion
prohibited and can cause a panic. Each pdata struct MUST provide the following methods:
```

pdata/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Optional fields
+
+Each optional field in protobuf message exposed in pdata MUST have an additional method with `Has` prefix to
+determine if the field is set.
+
+### OneOf fields
+
+A pdata struct representing a protobuf message with an oneof field MUST provide getter and setter methods for each 
+option of the oneof field. If an oneof field option is another protobuf message, setter name MUST include `Empty`
```
mx-psi, line 60: 'oneof' starts with a consonant sound /w/ so it should be preceded by `a`, not `an` (see [here](https://english.stackexchange.com/a/333972))
```suggestion
A pdata struct representing a protobuf message with a oneof field MUST provide getter and setter methods for each 
option of the oneof field. If a oneof field option is another protobuf message, setter name MUST include `Empty` 
```

pdata/README.md
```diff
@@ -0,0 +1,144 @@
...
+declared structs. We keep a pointer to OTLP protobuf in the "orig" member field. This allows efficient translation 
+to/from OTLP wire protocol. The underlying data structure is kept private so that we are free to make changes to it 
+in the future.
+
+API provided by pdata ensures that no underlying data can be used in more than one place. Each pdata instance cannot 
+contain a reference to an object that is already used somewhere else.
+
+## API naming convention
+
+### protobuf message representation in pdata
```
bogdandrutu, line 17: One more exception is that we have "Resource" in the common package instead of "resource" package.

pdata/README.md
```diff
@@ -0,0 +1,144 @@
...
+- `ptrace.Traces` based on `TracesData` protobuf message.
+- `pcommon.Slice` based on `ArrayValue` protobuf message.
+- `pcommon.Map` based on `KeyValueList` protobuf message.
+- `pcommon.Value` based on `AnyValue` protobuf message.
+
+Each pdata struct MUST have an initialization function starting with `New` prefix. Usage of zero-initialized is 
+prohibited and can cause a panic. Each pdata struct MUST provide the following methods:
+
+- `MoveTo(<type>)`: moves all the data from one struct instance to another. Destination instance is overridden, the 
+  source instance is re-initialized.
```
bogdandrutu, line 35: Clarify what "re-initialized" mean.

pdata/README.md
```diff
@@ -0,0 +1,144 @@
+# Pipeline data (pdata)
```
bogdandrutu, line 1: Add one section about the "flags". I think we had a good conversation with conclusions there as well.

pdata/README.md
dmitryax: We don't specify package names here. I added a small paragraph for that

pdata/README.md
dmitryax: done

pdata/README.md
dmitryax: Added sections for enums and flags

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+	AggregationTemporalityDelta = AggregationTemporality(otlpmetrics.AggregationTemporality_AGGREGATION_TEMPORALITY_DELTA)
+	AggregationTemporalityCumulative = AggregationTemporality(otlpmetrics.AggregationTemporality_AGGREGATION_TEMPORALITY_CUMULATIVE)
+)
+```
+
+### Flags
+
+Flags fields are typically defined in protobuf as `uint32` fields representing 32 distinct boolean flags. The bit 
+fields of the flags are defined as enum definitions. The exposed protobuf flags fields MUST be defined as new types 
+with `uint32` underlying type and called according to the same rules as struct names representing protobuf messages
```
bogdandrutu, line 161: Mention that is immutable.

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+)
+```
+
+### Flags
+
+Flags fields are typically defined in protobuf as `uint32` fields representing 32 distinct boolean flags. The bit 
+fields of the flags are defined as enum definitions. The exposed protobuf flags fields MUST be defined as new types 
+with `uint32` underlying type and called according to the same rules as struct names representing protobuf messages 
+and other enum types. Each pdata flags type MUST have an empty variable of the fields type with `Default` prefix. 
+Each flag MUST have a getter method returning a boolean value and a setter method with `With` prefix returning a
```
bogdandrutu, line 163: Not necessary boolean, because we may have non boolean fields in the flags (a 3 state for example).

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+	AggregationTemporalityUnspecified = AggregationTemporality(otlpmetrics.AggregationTemporality_AGGREGATION_TEMPORALITY_UNSPECIFIED)
+	AggregationTemporalityDelta = AggregationTemporality(otlpmetrics.AggregationTemporality_AGGREGATION_TEMPORALITY_DELTA)
+	AggregationTemporalityCumulative = AggregationTemporality(otlpmetrics.AggregationTemporality_AGGREGATION_TEMPORALITY_CUMULATIVE)
+)
+```
+
+### Flags
+
+Flags fields are typically defined in protobuf as `uint32` fields representing 32 distinct boolean flags. The bit 
+fields of the flags are defined as enum definitions. The exposed protobuf flags fields MUST be defined as new types
```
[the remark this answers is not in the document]
dmitryax: Ok, removed the sentence.

pdata/README.md
dmitryax: done

pdata/README.md
dmitryax: corrected

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+- `NumberDataPointValueTypeString` (2)
+
+An unset oneof protobuf value MUST be represented by a pdata constant with `Empty` suffix, e.g.:
+
+- `MetricTypeEmpty` (1)
+- `NumberDataPointValueTypeEmpty` (2)
+
+The pdata enum type SHOULD have `String()` method returning a string value of the corresponding oneof field option. 
+
+### Repeated fields
```
Aneurysm9, line 104:
```suggestion
### Repeated protobuf fields
```

pdata/README.md
```diff
@@ -0,0 +1,144 @@
...
+  source instance is re-initialized.
+- `CopyTo(<type>)`: copies all the data from one struct instance to another. Destination instance is overridden, the
+  source instance is not modified.
+
+### protobuf fields representation in pdata
+
+### Singular fields
+
+Each pdata struct based on a protobuf message SHOULD have getter methods for every protobuf field called the same as
+the protobuf field but in CamelCase. Exceptions to this rule are allowed if exposing a field is not desirable in pdata.
```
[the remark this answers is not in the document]
dmitryax: Added couple of examples. I don't think we should necessary limit reasons why fields should/can be not exposed in pdata. Just wanted to keep it open.

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+Pipeline data (pdata) implements data structures that represent telemetry data in-memory. All data received 
+is converted into this format, travels through the pipeline in this format, and is converted from this format by 
+exporters when sending.
+
+Current implementation primarily uses OTLP protobuf structs as the underlying data structures for many of the 
+declared structs. We keep a pointer to OTLP protobuf in the "orig" member field. This allows efficient translation 
+to/from OTLP wire protocol. The underlying data structure is kept private so that we are free to make changes to it 
+in the future.
+
+API provided by pdata ensures that no underlying data can be used in more than one place. Each pdata instance cannot
```
tigrannajaryan, line 12: >no underlying data can be used in more than one place

This is a bit unclear, since "used" and "place" can be interpreted in different ways.

Can we maybe say instead:
```suggestion
The pdata is designed to avoid mutable data sharing and bugs that stem from that. Each pdata instance cannot 
```

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+API provided by pdata ensures that no underlying data can be used in more than one place. Each pdata instance cannot 
+contain a reference to an object that is already used somewhere else.
+
+## API naming convention
+
+### Protobuf message representation in pdata
+
+Packages of pdata are not following names of the protobuf packages. pdata has a package per telemetry type starting 
+with `p`, e.g. `ptace`, and `pcommon` package with API for other protobuf definitions from `common` and `resource`
```
tigrannajaryan, line 20:
```suggestion
with `p`, e.g. `ptrace`, and `pcommon` package with API for other protobuf definitions from `common` and `resource` 
```

tigrannajaryan: >with API for other protobuf definitions from `common` and `resource`
>protobuf packages

I do not quite understand what this means.

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+## API naming convention
+
+### Protobuf message representation in pdata
+
+Packages of pdata are not following names of the protobuf packages. pdata has a package per telemetry type starting 
+with `p`, e.g. `ptace`, and `pcommon` package with API for other protobuf definitions from `common` and `resource` 
+protobuf packages.
+
+Pipeline data structs SHOULD be based on the names of the underlying OTLP protobuf messages. Data types for 
+protobuf messages defined as part of another message SHOULD include the owner's name as a prefix. Exceptions to these
```
tigrannajaryan, line 24: Can we add one example to make clear what this means?

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+protobuf messages defined as part of another message SHOULD include the owner's name as a prefix. Exceptions to these 
+rules are possible, but not encouraged. Another name can be chosen for brevity or if the struct provides a completely
+different interface to manipulate the underlying data. The following exceptions are currently accepted:
+
+- `plog.Logs` based on `LogsData` protobuf message.
+- `pmetric.Metrics` based on `MetricsData` protobuf message.
+- `ptrace.Traces` based on `TracesData` protobuf message.
+- `pcommon.Slice` based on `ArrayValue` protobuf message.
+- `pcommon.Map` based on `KeyValueList` protobuf message.
+- `pcommon.Value` based on `AnyValue` protobuf message.
```
tigrannajaryan, line 33: Would it be better to delete all these lines and only keep one exception example, to reduce the amount of maintenance this document requires if the source code is refactored? Let the source code be the source of truth. If you want the fact of the exception to be documented it would be better to add a comment in the source code near the name.

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+Each pdata struct MUST have an initialization function starting with `New` prefix. Usage of zero-initialized values is 
+prohibited and can cause a panic. Each pdata struct MUST provide the following methods:
+
+- `MoveTo(<type>)`: moves all the data from one struct instance to another. The destination instance is overwritten and the 
+  source instance is re-initialized as a new empty instance.
+- `CopyTo(<type>)`: copies all the data from one struct instance to another. The destination instance is overwritten and the
```
tigrannajaryan, line 40:
```suggestion
- `CopyTo(<type>)`: deep copies all the data from one struct instance to another. The destination instance is overwritten and the 
```

pdata/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+- `AsRaw() []<primitive type>`: returns copy of the underlying go slice. 
+- `FromRaw(val []<primitive type>)`: copies a go slice into the pdata wrapper instance.
+- `Len()`: returns length of the slice.
+- `At(int) <primitive type>`: returns an element from the slice by its index, can cause "index out of range" panic.
+- `EnsureCapacity(int)`: ensures that the slice have at least the requested capacity.
+- `AppendEmpty() <primitive type>`: adds another slice element with a zero value.
+- `CopyTo(<slice type>)`: copies the slice to another slice of the same type overriding the destination.
+- `MoveTo(<slice type>)`: move the slice to another slice of the same type. The destination is overridden, the 
+  source is set to an empty slice.
```
tigrannajaryan, line 130: Do we need these detailed docs given that this is all codified in the pdata generator? I am worried that we may forget to keep this up to date when we change any behavior in the generator.
The list of methods also seems incomplete, I can't find `MoveAndAppendTo` or `RemoveIf`.

tigrannajaryan (commented):
I think it is useful to document the design principles of pdata API, but I think it would be beneficial to remove from this document the parts which are already codified in the pdata generator to avoid the need to maintain the document if the generator is changed in any way.

pdata/README.md
dmitryax: Updated

pdata/README.md
dmitryax: Added

pdata/README.md
dmitryax: Updated. Thanks!

pdata/README.md
dmitryax: applied

pdata/README.md
dmitryax: Makes sense. Maybe requiring a particular set of methods is redundant here. Removed them

pdata/README.md
dmitryax: Makes sense. Removed

bogdandrutu (commented):
Please document all exceptions, see comment
https://github.com/open-telemetry/opentelemetry-collector/pull/6283#issuecomment-1275385123

dmitryax:
> Please document all exceptions, see comment https://github.com/open-telemetry/opentelemetry-collector/pull/6283#issuecomment-1275385123

We had the struct name exceptions listed in this document initially.
 
@tigrannajaryan suggested to not list them all, but use the code as source of truth https://github.com/open-telemetry/opentelemetry-collector/pull/6255#discussion_r992334428.

If we want to document field name exceptions, I believe we should do the same for struct. So the question is where we document the exceptions: source code comments or this Readme file.

Using the readme adds maintenance burden as Tigran pointed out. But I don't like putting them into the comments because the comments are intended for API users who probably don't care why particular type or field is called one way or another. So I would rather list them here, and I don't think it'll be a lot of them going forward to bother us with keeping it in sync.

dmitryax:
@bogdandrutu, I mentioned all the naming exceptions in this doc

bogdandrutu (approved):

bogdandrutu:
Not perfect, but a great start :) Thanks @dmitryax

Issue #6257, opened 2022-10-08
[pdata] Rename pdata struct that are named inconsistently with protobuf messages


dmitryax:
We want rename the following structs that are not aligned with names of underlying protobuf messages:

- [x] `pmetric.MetricAggregationTemporality` -> `pmetric.AggregationTemporality`
- [x] `ptrace.SpanStatus` -> `ptrace.Status`
- [x] `pmetric.MetricDataPointFlags` -> `pmetric.DataPointFlags`
- [x] `pmetric.ValueAtQuantile` -> `pmetric.SummaryDataPointValueAtQuantile`
- [x] `pmetric.Buckets` -> `ExponentialHistogramDataPointBuckets`

See https://github.com/open-telemetry/opentelemetry-collector/issues/6200 for the discussion

PR #6259, opened 2022-10-08
[pdata] Rename pmetric.MetricDataPointFlags to pmetric.DataPointFlags

dmitryax:
Following pdata naming rules outlined in https://github.com/open-telemetry/opentelemetry-collector/pull/6255, we want to update the struct to have the same name as defined in protobuf package.

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/6257


bogdandrutu (approved):

bogdandrutu:
where is the DataPointFlags defined?

dmitryax:
@bogdandrutu do you mean pdata definition? it's here https://github.com/open-telemetry/opentelemetry-collector/pull/6259/files#diff-39b6dfa0a0fdf8e13fb63075018e16a009a905cfd91c451af47768bac6268e96R23

bogdandrutu:
>As a new rule we want to rename all the structs for proto messages defined at the package level to not have the owner prefix in its name.

Read the description.

dmitryax:
Right, the description is a bit incorrect. This is proto enum not a message https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L321. But I believe we should apply the same rule for enums. Do you think otherwise?

dmitryax:
Updated the description

PR #6267, opened 2022-10-10
[config{grpc,http}] Add warning when using unspecified address

### LINKED ISSUE #6151: [receiver/otlp] CWE-1327 in OTLP receiver
### Introduction

Recently there was a CVE on the Go runtime (CVE-2022-27664) that could be leveraged to cause a denial of service on applications built with affected runtimes through a faulty HTTP/2 connection. We addressed this CVE on the official binaries by bumping the Go version (see open-telemetry/opentelemetry-collector-releases/pull/198).

This CVE impacted in particular the OTLP receiver. While some level of impact was unavoidable on this receiver since exposing an HTTP/2 server is part of its core functionality, the OTLP receiver default endpoints (`0.0.0.0:4317` and `0.0.0.0:4318`) made it so that this vulnerability could be leveraged more easily than it could have been. In particular, these endpoints fall prey to [CWE-1327](https://cwe.mitre.org/data/definitions/1327.html) (*"Binding to an Unrestricted IP Address"*), instead of a more restricted address.

My intention with this issue is to discuss if we can in some way mitigate the impact of future CVEs for OTLP receiver users through either code changes or improved documentation, by effectively making so that less users use `0.0.0.0` when they don't actually need to be so unrestricted.

### Possible ways to address this

We could do all, none or any other combination of the proposals below.

### Change the default endpoint to use localhost instead of `0.0.0.0`

**Description**: Change the default endpoint to be `localhost:4317` and `localhost:4318` for OTLP/gRPC and OTLP/HTTP receivers respectively.

**Advantages**: 
1. It fits with the OpenTelemetry "secure by default" philosophy.
2. It is a reasonable default for some use cases (e.g. running the bare binary).

**Downsides**: 
1. It is a breaking change for some users. We define the OTLP receiver configuration schema to be ['stable'](https://github.com/open-telemetry/opentelemetry-collector#stable). While the definition leaves some leeway for breaking changes under special circumstances, we don't have a consensus on what these circumstances should be.
2. This default will not work as nicely in some use cases (e.g. running in a containerized setup).

### Add a warning when the endpoint is `0.0.0.0`

**Description**: Add a warning message when using an `0.0.0.0` endpoint.

**Advantages**:
1. Can be done in a backwards compatible way.

**Downsides**:
1. Most people (probably) don't read the logs.
2. It's hard (impossible?) to show this warning only when it's really useful. This would be logged in cases where no action is needed (e.g. containerized setups).

### Document security recommendations for these endpoints

**Description**: We add a section to our [security recommendations](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security.md) recommending users to change the endpoint to limit exposure.

**Advantages**:
1. Can be done in a backwards compatible way.

**Downsides**:
1. Most people (probably) don't read the security documentation.

mx-psi:
**Description:** 

Add warning when using `0.0.0.0` address on an HTTP or gRPC server.

**Link to tracking Issue:** Fixes #6151


config/configgrpc/configgrpc.go
```diff
@@ -273,6 +274,22 @@ func (gss *GRPCServerSettings) ToListener() (net.Listener, error) {
...
+	switch gss.NetAddr.Transport {
+	case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
+		if host, _, err := net.SplitHostPort(gss.NetAddr.Endpoint); err != nil {
+			return nil, fmt.Errorf("failed to parse endpoint: %w", err)
+		} else if host == "0.0.0.0" || host == "::" {
+			settings.Logger.Warn(
+				"Using the 0.0.0.0 address exposes this server to every network interface, which may facilitate Denial of Service attacks",
+				zap.String(
+					"documentation",
+					"https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security.md#safeguards-against-denial-of-service-attacks",
```
codeboten, line 287:
```suggestion
					"https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks",
```

config/confighttp/confighttp.go
```diff
@@ -259,6 +261,18 @@ func WithErrorHandler(e errorHandler) ToServerOption {
...
 // ToServer creates an http.Server from settings object.
 func (hss *HTTPServerSettings) ToServer(host component.Host, settings component.TelemetrySettings, handler http.Handler, opts ...ToServerOption) (*http.Server, error) {
+	if host, _, err := net.SplitHostPort(hss.Endpoint); err != nil {
+		return nil, fmt.Errorf("failed to parse endpoint: %w", err)
+	} else if host == "0.0.0.0" || host == "::" {
+		settings.Logger.Warn(
+			"Using the 0.0.0.0 address exposes this server to every network interface, which may facilitate Denial of Service attacks",
+			zap.String(
+				"documentation",
+				"https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security.md#safeguards-against-denial-of-service-attacks",
```
codeboten, line 271:
```suggestion
				"https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks",
```

codeboten (commented):
Thanks for fixing, just a suggestion to fix the (now broken) links

config/confighttp/confighttp.go
```diff
@@ -259,6 +261,18 @@ func WithErrorHandler(e errorHandler) ToServerOption {
 
 // ToServer creates an http.Server from settings object.
 func (hss *HTTPServerSettings) ToServer(host component.Host, settings component.TelemetrySettings, handler http.Handler, opts ...ToServerOption) (*http.Server, error) {
+	if host, _, err := net.SplitHostPort(hss.Endpoint); err != nil {
```
codeboten, line 264: Any thoughts on using `IsUnspecified` here? https://pkg.go.dev/net#IP.IsUnspecified

codeboten (approved):

config/configgrpc/configgrpc.go
```diff
@@ -273,6 +274,22 @@ func (gss *GRPCServerSettings) ToListener() (net.Listener, error) {
 
 // ToServerOption maps configgrpc.GRPCServerSettings to a slice of server options for gRPC.
 func (gss *GRPCServerSettings) ToServerOption(host component.Host, settings component.TelemetrySettings) ([]grpc.ServerOption, error) {
+
+	switch gss.NetAddr.Transport {
+	case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
+		if host, _, err := net.SplitHostPort(gss.NetAddr.Endpoint); err != nil {
+			return nil, fmt.Errorf("failed to parse endpoint: %w", err)
+		} else if ip := net.ParseIP(host); ip != nil && ip.IsUnspecified() {
```
bogdandrutu, line 282: No need for else when return. Makes code easier to read.

config/confighttp/confighttp.go
```diff
@@ -259,6 +261,18 @@ func WithErrorHandler(e errorHandler) ToServerOption {
...
+		return nil, fmt.Errorf("failed to parse endpoint: %w", err)
+	} else if ip := net.ParseIP(host); ip != nil && ip.IsUnspecified() {
```
bogdandrutu, line 266: same comment here, remove else please :)

config/confighttp/confighttp.go
bogdandrutu, line 264: Can we not have http over unix sockets? (probably we don't support this).

config/configgrpc/configgrpc.go
mx-psi: `host` is local to the if block, will define it outside :+1:

config/confighttp/confighttp.go
mx-psi: Listeners created through the `HTTPServerSettings` have the protocol hardcoded to `tcp`: https://github.com/open-telemetry/opentelemetry-collector/blob/bbf6043cc764bf73305a1d9c874c688f7aa4c1ef/config/confighttp/confighttp.go#L227

I guess someone could create a custom listener ignoring the configuration but create a server with the `ToServer` method. This seems clearly against the intention of the current API (TLS configuration and Endpoint from config would be ignored if not handled explicitly). Should we care about that case?

mx-psi: If we care, I don't see a way out other than logging the error if `net.SplitHostPort` fails and skipping the unspecified check; we don't know within `ToServer` what kind of listener we will get.

config/configgrpc/configgrpc.go
```diff
@@ -273,6 +274,25 @@ func (gss *GRPCServerSettings) ToListener() (net.Listener, error) {
...
+			return nil, fmt.Errorf("failed to parse endpoint: %w", err)
+		}
+
+		if ip := net.ParseIP(host); ip != nil && ip.IsUnspecified() {
+			settings.Logger.Warn(
+				"Using the 0.0.0.0 address exposes this server to every network interface, which may facilitate Denial of Service attacks",
+				zap.String(
+					"documentation",
+					"https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks",
+				),
```
bogdandrutu, line 291: Last comment, I promise :)) 

Can we actually have a `Validate() error` on `confignet.NetAddr` and `confignet.TCPAddr`. That way you can share this code.

bogdandrutu (approved):

config/configgrpc/configgrpc.go
bogdandrutu: Hard probably to have the warning about "0.0.0.0" (since the Validate will not have access to the logger) but that can be duplicate or a standalone helper in the `confignet` package.

mx-psi: I created a `WarnOnUnspecifiedHost` helper on `config/internal`. Does that look good?

PR #6268, opened 2022-10-10
Disallow dollar sign in opaque value, allow recursive expansion to be implemented in the future

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-collector/issues/5706


bogdandrutu:
skip changelog since the feature is not yet enabled (if put in the changelog would cause confusion).

mx-psi (approved):

Issue #6256, opened 2022-10-07
[pdata] Inconsistent names of zero enum constants


dmitryax:
There are two categories of  enum types in pdata:

1. To represent enum fields in proto:
- `plog.SeverityNumber`
- `pmetrc.AggregationTemporality`
- `ptrace.SpanKind`
- `ptrace.StatusCode`

2. To represent options of oneof proto fieds:
- `pcommon.ValueType`
- `pmetric.MetricType`
- `pmetric.NumberDataPointValueType`
- `pmetric.ExemplarValueType`

Zero enums constants are called inconsistently across pdata.

For proto-defined enums (1), we mostly use the same name as defined in proto which is always `Unspecified`, e.g. `otrace.SpanKindUnspecified`. The only exception is SeverityNumber where we use `Undefined` for some reason:

```
SeverityNumberUndefined = SeverityNumber(otlplogs.SeverityNumber_SEVERITY_NUMBER_UNSPECIFIED)
```

For enums that defined in pdata (2) only, we use `Empty` and `None` suffixes for zero enum constants:

- `pcommon.ValueTypeEmpty`
- `pmetric.MetricTypeNone`
- `pmetric.NumberDataPointValueTypeNone`
- `pmetric.ExemplarValueTypeNone`

We should consider if we want to set the same suffix for all of them or allow different names. 

If we stick with `Unspecified` for (1), we need to rename at least `SeverityNumberUndefined` to `SeverityNumberUnspecified`. 

Renaming `pcommon.ValueTypeEmpty` doesn't seem desirable to me given that we have functions like `pcommon.NewValueEmpty` and `pcommon.Map.PutEmpty` that doesn't seem like should be renamed. 

Using `Empty` for `pmetric.MetricTypeNone`, `pmetric.NumberDataPointValueType` and `pmetric.ExemplarValueTypeNone` seems like acceptable option to me. Possibly `Empty` can be also applied to (1) enums as well.

@open-telemetry/collector-approvers

bogdandrutu:
> For proto-defined enums (1), we mostly use the same name as defined in proto which is always Unspecified, e.g. otrace.SpanKindUnspecified. The only exception is SeverityNumber where we use Undefined for some reason:

I think we should change everywhere with `Unspecified`

> For enums that defined in pdata (2) only, we use Empty and None suffixes for zero enum constants:

I think we should consider Empty for Value an exception since it has a specific meaning.

dmitryax:
[...]

PR #6270, opened 2022-10-10
[pdata] Rename enum zero value constants to have `Empty` suffix

dmitryax:
For consistency with `pcommon.ValueType`

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/6256

### DIFF
--- a/pdata/pmetric/metrics.go
+++ b/pdata/pmetric/metrics.go
@@ -102,7 +102,8 @@ func (ms Metrics) DataPointCount() (dataPointCount int) {
 type MetricType int32
 
 const (
-	MetricTypeNone MetricType = iota
+	// MetricTypeEmpty means that metric type is unset.
+	MetricTypeEmpty MetricType = iota
 	MetricTypeGauge
 	MetricTypeSum
 	MetricTypeHistogram
@@ -110,11 +111,14 @@ const (
 	MetricTypeSummary
 )
 
+// Deprecated: [0.62.0] Use MetricTypeEmpty instead.
+const MetricTypeNone = MetricTypeEmpty
+
 // String returns the string representation of the MetricType.
 func (mdt MetricType) String() string {
 	switch mdt {
-	case MetricTypeNone:
-		return "None"
+	case MetricTypeEmpty:
+		return "Empty"
 	case MetricTypeGauge:
 		return "Gauge"
 	case MetricTypeSum:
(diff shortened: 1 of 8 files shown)


bogdandrutu (commented):
I don't think we should do this to be honest. In case of "Value", the `Empty` has a real value (is correct), in case for `None` that is not really valid.

dmitryax:
Now I see that I misread you comment in https://github.com/open-telemetry/opentelemetry-collector/issues/6256. Do you think we should keep it as `None` or `Unspecified`?

bogdandrutu:
@dmitryax let's think a bit more about this. "empty" may not be bad option, since it expose that none of the values within the `oneof` group is present, aka the `oneof` is empty. 
* For `MetricType`, empty sounds good, though the metric may have a name (is not completely empty).
*  `NumberDataPointValueType` , empty sounds maybe that the value is `empty`?

I can be convinced to do this change, does not necessary seem the worst, since from protocol perspective we have the capability to really on the presence of the oneof (any member) or not.

dmitryax:
> "empty" may not be bad option, since it expose that none of the values within the oneof group is present, aka the oneof is empty.

That was my original idea.

> For MetricType, empty sounds good, though the metric may have a name (is not completely empty).

I believe metric with a name still can be "empty". It sounds more natural for me than `None` metric type. Maybe `Unspecified` is a good choice as well.

> NumberDataPointValueType , empty sounds maybe that the value is empty?

It still mostly applied to Value _Type_ rather value itself, but I believe it's ok to interpret it like that.

Both `Empty` and `Unspecified` sound like a good options for me, not sure about `None` tho. Maybe `Empty` is better for `MetricType` and `Unspecified` for `NumberDataPointValueType` and `ExemplarValueType`.

pdata/pmetric/metrics.go
```diff
@@ -102,19 +102,22 @@ func (ms Metrics) DataPointCount() (dataPointCount int) {
 type MetricType int32
 
 const (
-	MetricTypeNone MetricType = iota
+	MetricTypeEmpty MetricType = iota
```
bogdandrutu, line 105: Add a comment to the value that it means none of the "possible" values is set? Or something in that lines.

bogdandrutu (approved):

pdata/pmetric/metrics.go
dmitryax: Added a comment to these

bogdandrutu (approved):

PR #6276, opened 2022-10-11
Enable confmap expand feature

bogdandrutu:


confmap/resolver.go
```diff
@@ -149,17 +147,15 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
...
-		cfgMap := make(map[string]interface{})
-		for _, k := range retMap.AllKeys() {
-			val, err := mr.expandValueRecursively(ctx, retMap.Get(k))
-			if err != nil {
-				return nil, err
-			}
-			cfgMap[k] = val
+	cfgMap := make(map[string]interface{})
+	for _, k := range retMap.AllKeys() {
+		val, err := mr.expandValueRecursively(ctx, retMap.Get(k))
```
dmitryax, line 152: Shouldn't we remove `expandconverter` now? At least from `newDefaultConfigProviderSettings`?

dmitryax (commented):
Should we enable it through a feature gate at the beginning? I believe it might introduce some unexpected outcomes for distros that do it using another approach...

confmap/resolver.go
bogdandrutu: No, read this does not expand unless a scheme is provided, and that is not the case for the `expandconverter`.

bogdandrutu:
>Should we enable it through a feature gate at the beginning? I believe it might introduce some unexpected outcomes for distros that do it using another approach...

No, because that has a different syntax which does not match this one, so the mechanism will not trigger.

The reason a featureflag is not required is because is a new syntax, and if users are using this new syntax is clear that they want this feature, otherwise this feature is not triggered.

bogdandrutu:
@dmitryax convinced me that this may overlap with splunk's distros feature, so protect this by a feature gate for the moment.

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@
 - `receiver/otlp`: Make logs related to gRCPC and HTTP server startup clearer (#6174)
 - Add prometheus metric prefix and constant service attributes to Collector's own telemetry when using OpenTelemetry for internal telemetry (#6223)
 - `exporter/logging`: Apply consistent rendering of map values (#6244)
+- Add support in the confmap.Resolver to expand embedded config URIs inside configuration. (#6276)
+  - For usage details see https://github.com/open-telemetry/opentelemetry-collector/pull/4742.
```
dmitryax, line 52:
```suggestion
  - For usage details see https://github.com/open-telemetry/opentelemetry-collector/pull/6276.
```

dmitryax (approved):

CHANGELOG.md
bogdandrutu: Removed the line, no need since is already the PR listed.

Aneurysm9:
> The old way of expanding env vars (without a scheme) is still supported (for at least few versions)

This is a significant feature of the configuration system used by many users in many ways.  We need to be very careful about this transition and ensure that it is seamless as possible.  I don't think we can simply say it will go away in a couple of releases.

bogdandrutu:
@Aneurysm9 when that moment comes, happy to hear opinions, for the moment let's aim for that (be positive).

Aneurysm9:
All I'm saying is that if we want to make this change the time to plan for transition is now, not when we're ready to disable functionality that users have relied on up to this point.

PR #6279, opened 2022-10-11
Fixes #4946

### LINKED ISSUE #4946: Collector Run should exit quickly if shutdown was called prior
**Describe the bug**
Currently if `Shutdown` is called before `Run` all [this](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L227-L251) startup code runs before `runAndWaitForShutdownEvent` which will then process the shutdown. 

**Steps to reproduce**
1. Created collector
2. Call `Shutdown`
3. Call `Run`

Collector not attempt to run startup/configure code and exit early.

**What did you see instead?**
Startup code was run then exited.

**What version did you use?**
Version: v0.45.0

**What config did you use?**
Config: any valid config

**Environment**
OS: MacOS 12.2
Compiler(if manually compiled): go 1.17.6

**Additional context**
Action item from #4878

Chinwendu20:
Signed-off-by: Maureen <amaka013@gmail.com>

Heyy could you please check if this is aligned? @cpheps 

Description: Fixed collector attempting startup when `ShutDown` is called before `Run`

Link to tracking Issue: fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4946


CHANGELOG.md
```diff
@@ -48,6 +48,7 @@
 ### 🧰 Bug fixes 🧰
 
 - Fixed bug where `telemetryInitializer` is not cleaned up when `newService` errors (#6239)
+- Fixed collector attempting startup when `ShutDown` is called before `Run`
```
cpheps, line 51: Can you add the PR number in parenthesis here please? 

So just adding `(#6279)` at the end.

service/collector.go
```diff
@@ -158,7 +158,8 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 // Run starts the collector according to the given configuration, and waits for it to complete.
 // Consecutive calls to Run are not allowed, Run shouldn't be called once a collector is shut down.
 func (col *Collector) Run(ctx context.Context) error {
-	if err := col.setupConfigurationComponents(ctx); err != nil {
+	_, ok := <-col.shutdownChan
```
cpheps, line 161: If `Shutdown` is not called before `Run` this will block indefinitely causing run to never start.

I think we can add a check like this at the start:

```go
state := col.GetState()
if state == Closing || state == Closed {
  return errors.New("collector has already been shutdown and cannot be run again")
}
```

It might be worth saving that error off into a variable for easy of use and testing.

cpheps (changes requested):

CHANGELOG.md
Chinwendu20: Oh okay, thanks a lot. I do not know how I missed that

service/collector.go
Chinwendu20: I really appreciate your help and guidance towards helping me complete this issue, I guess I am still a bit new to GO and GO's concurrency is uncharted waters for me, glad for this issue that has served to help me learn it practically. So while coming up with a solution, I looked at the test case you created [here]((https://github.com/open-telemetry/opentelemetry-collector/pull/4878/files#diff-39cb0a393ab609b58e27ab97198f7fdd885a5369e325d25b32879b22d6e19019) and I created a small snippet of code to simulate that. When I ran it even without "shutdown" being called before, it worked so I thought it was a good solution:

image
I thought since there is an equivalent concurrent reading and writing of the channel in the main function and go routine, there would be no blocking. I guess I might have missed something.

It did block if shut is not called right after run, but I found that was the same case if I did not have that extra line checking if the  channel was closed or not:

image


Your solution looks really good but I guess I did not not know the collection's state would be set to closing when the `ShutDown` function was called, as I could not find it here:

image

cpheps: >Your solution looks really good but I guess I did not not know the collection's state would be set to closing when the ShutDown function was called, as I could not find it here:

It's a bit of a chain reaction. Closing the `shutdownChan` gets picked up in `Run` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L195). It then calls `col.shutdown` which sets the state to `Closing` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L209).

As for learning go concurrency if you haven't yet checking out [tour of go](https://go.dev/tour/concurrency/1) is a good starting place. Also, just playing around with it is a good way to learn. In the example above you gave this line will block until something is pushed onto the channel or the channel is closed.

```go
_, ok := <-col.shutdownChan
```

If the channel is closed `ok` will be false. If something is pushed the channel, before closing, `ok` will be true indicating the channel is still active and can be listened on again.

Chinwendu20: Thank you so much. I find that there are two shutdown functions.One with a capital 's' and the other with a small 's' The one I saw in your test and on the screenshot that I shared above is the one with a capital 's'. So when collector.ShutDown is called it does not cause any chain reaction as it is not same as collector.shutdown which is used inside in Run() function.

Thanks for the great tips on how to get started with concurrency. 

`_, ok := <-col.shutdownChan`
This line reads from the col channel  and waits till there something os pushed to the channel before it unblocks. Can we not say the same about the select statement in Run() function? As it does not have a 'default block'. I find that if you call The Run() function without the ShutDown function without the ShutDown function, before or after the Run(), we would get an error
So I guess adding the line of code that I added would not make the function behave differently

cpheps: >So when collector.ShutDown is called it does not cause any chain reaction as it is not same as collector.shutdown which is used inside in Run() function.

So here is the chain reaction that causes the collector to stop once `Shutdown` is called:
1. [Shutdown is called](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L114)
2. If the collector is running it [closes the shutdownChan](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L121)
3. In `Run` the closed channel triggers [this](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L195) case in the select and breaks out of the main loop.
4. At the end of `Run` [shutdown is called](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L205) and `Run` returns the result of `shutdown`.

> This line reads from the col channel and waits till something is pushed to the channel, causing a block. Can we not say the same about the select statement in Run() function?

This is true. The issue #4946 that you're working on is to make this happen immediately inside `Run`. Currently `Run` has some setup logic [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/collector.go#L161-L164) which calls [setupConfigurationComponents](https://github.com/open-telemetry/opentelemetry-collector/blob/f24a6ac30fa027d3e373f6280f5b4bb4b4626587/service/collector.go#L127-L156). There are a lot of things that happen in this that can all be skipped by an early detection that we've already been shut down.

Chinwendu20: Thanks a lot for your patience with me. This chain of event only happens if Run is called before ShutDown. I thought we were looking at a scenario when Shutdown is called before Run. As per your statement:

image

So the way that I am looking at it is if ShutDown is called before Run, it would not be able to receive any message, as the blocking event awaiting receipt of message from shutDownchan is not currently on (i.e. Run has not been called yet) but I guess that it is not how it works. That was my initial thought process :

image

Thanks for the clarity.

Then I guess even though the select statement causes a block, it makes that happen in the right stage of execution as opposed to where my line of code would have caused that

cpheps (commented):
These changes look good. Can we add a test for the new logic. That calls `Shutdown` then `Run` and verifies the error returned is the new one you added?

Chinwendu20:
> These changes look good. Can we add a test for the new logic. That calls `Shutdown` then `Run` and verifies the error returned is the new one you added?

Oh okay. Sounds good. On it

cpheps:
> However, it works with my initial line of code:
>
>_, ok := <-col.shutdownChan if !ok { return errors.New("collector has already been shutdown and cannot be run again") }

Your line of code works when Shutdown is called before Run. If you use the normal flow of calling just Run the collector will never fully start it'll block on that line as the `col.shutdownChan` is either waiting for something to be pushed onto the channel or a close. The channel is closed when you call `Shutdown` before `Run`.

The test you called out [here](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L53) does have a race condition now. Because `startCollector` is called [here](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L69) and kicks off a goroutine there's no guarantee that `Run` is called before `Shutdown` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L75). They'll need to be some flow control added to ensure run has started then call `Shutdown`

Chinwendu20:
> > However, it works with my initial line of code:
> > _, ok := <-col.shutdownChan if !ok { return errors.New("collector has already been shutdown and cannot be run again") }
> 
> Your line of code works when Shutdown is called before Run. If you use the normal flow of calling just Run the collector will never fully start it'll block on that line as the `col.shutdownChan` is either waiting for something to be pushed onto the channel or a close. The channel is closed when you call `Shutdown` before `Run`.
> 
> The test you called out [here](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L53) does have a race condition now. Because `startCollector` is called [here](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L69) and kicks off a goroutine there's no guarantee that `Run` is called before `Shutdown` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L75). They'll need to be some flow control added to ensure run has started then call `Shutdown`

Thanks a lot for your time. I get that that this line of code

```GO
_, ok := <-col.shutdownChan if !ok { return errors.New("collector has already been shutdown and cannot be run again") }
```
Might not be the best solution because of what you highlighted, thanks a lot for pointing that out.

However, this line of code:
```GO
state := col.GetState()
if state == Closing || state == Closed {
  return errors.New("collector has already been shutdown and cannot be run again")
}
```
 does not stop this [line] (https://github.com/open-telemetry/opentelemetry-collector/blob/4d445807f9eca3af660bec2a3fb4ca3518cd255d/service/collector.go#L161) from running because as highlighted in your previous comment and from what I have seen.

image
The collector needs to be running to trigger the chain of events that would lead to the "Closing" state. But this is the requirement:
image

So I guess that I have to figure out another solution? But then I am wondering what that would  be as the only thing that happens in the "ShutDown" function is that the "shutdown" chan is closed

Also about the flow-control that you suggested should that be in my test for this issue or the [test](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L53) that I highlighted?

cpheps:
Ohh I missed [this](https://github.com/open-telemetry/opentelemetry-collector/blob/6313054b642c1ee8bba8c097a98dfb4cf870862e/service/collector_test.go#L71-L73) part of the test that ensures it's running before we call `Shutdown`.

Do you mind pushing the code you have? I just tried your new logic on the collector and all of the tests pass. I wondering if there's something small we might be missing that's causing the issue you're seeing.

Chinwendu20:
Yeah, all the tests should pass. The issue is with the new unit test that checks the logic of the code that addresses this issue. It is failing, that is what I was pointing out and explaining the reason that I think it is so. Are you saying I should still push it? I sent a screenshot of the test in the other comment.

cpheps:
Yeah if you could push your test, even if it's broken.  That way I can test it on my end, I think it might help me to see and run it.

Chinwendu20:
Okay pushed it.

Chinwendu20:
I have updated the unit test. Realized there was a mistake in the other one but test keeps failing though

cpheps:
@Chinwendu20 Thanks. I'm a bit busy today but I'll try this out early tomorrow. I appreciate your patience so far.

cpheps:
@Chinwendu20 sorry been busy I'm looking at your test today.

cpheps:
@Chinwendu20 I see the issue with your test and it has to do with the fact that the `state` doesn't get changed to `Closed` or `Closing` until `Run` has been called. This is not ideal as that defeats the entire purpose. It does feel like the states we have are not enough to represent that a Collector is new but hasn't started yet. I feel like that is out of the scope of this PR. 

As far as your test goes I would call Run in a goroutine, then Shutdown, then Run again and it should test the intent of your change. Here's the modified test I used on your branch:

```go
func TestCollectorShutdownBeforeRunMessage(t *testing.T) {
	factories, err := componenttest.NopFactories()
	require.NoError(t, err)

	cfgProvider, err := NewConfigProvider(newDefaultConfigProviderSettings([]string{filepath.Join("testdata", "otelcol-nop.yaml")}))
	require.NoError(t, err)

	set := CollectorSettings{
		BuildInfo:      component.NewDefaultBuildInfo(),
		Factories:      factories,
		ConfigProvider: cfgProvider,
		telemetry:      newColTelemetry(featuregate.NewRegistry()),
	}
	col, err := New(set)
	require.NoError(t, err)

	// Run the collector
	wg := &sync.WaitGroup{}
	wg.Add(1)
	go func() {
		defer wg.Done()
		assert.NoError(t, col.Run(context.Background()))
	}()

	// Calling shutdown before collector is running should cause it to return quickly
	require.NotPanics(t, func() { col.Shutdown() })

	// Wait for run to finish
	wg.Wait()

	// Call Run again and expect already shutdown error
	assert.EqualError(t, col.Run(context.Background()), "collector has already been shutdown and cannot be run again")
}
```

Chinwendu20:
Thank you! Nice way around the insufficient collector states. I guess I would modify my test with this and push?

cpheps:
@Chinwendu20 I think modifying the test should be fine.

Chinwendu20:
Okay, @cpheps. All done!

cpheps (approved):

Chinwendu20:
Hello, please any idea about the failed ci. I can see it is related to an ambiguous import error related to batchprocessor but have no idea why

Chinwendu20:
I would just update thiss branch with the most recent commit in the main branch and see if it works

cpheps:
@Chinwendu20 I think try rebasing your branch. The test failure doesn't look related to your code changes.

Chinwendu20:
Thanks @cpheps . I am not exactly sure of what actions that I should carry out with rebase but I have merged the branch with main

cpheps:
> Thanks @cpheps . I am not exactly sure of what actions that I should carry out with rebase but I have merged the branch with main

That's fine too.

jpkrohling (dismissed):
I'm blocking this for now, as I'm not sure the basic premise is valid here. I believe we should be able to start the collector after a shutdown.

jpkrohling:
@open-telemetry/collector-maintainers, could you confirm the premise I have is still accurate? Should collector services be able to start once they've been closed?

jpkrohling:
ping @open-telemetry/collector-maintainers

bogdandrutu:
> https://github.com/orgs/open-telemetry/teams/collector-maintainers, could you confirm the premise I have is still accurate? Should collector services be able to start once they've been closed?

If you think about config reload, we re-create Service :)

jpkrohling:
> If you think about config reload, we re-create Service :)

Alright, then I was wrong: start cannot be called after shutdown, from what I understand then.

PR #6283, opened 2022-10-11
[pdata] Simplify pdatagen to rely more on the proto naming dependency

dmitryax:
Given the documented naming recommendations for pdata in https://github.com/open-telemetry/opentelemetry-collector/pull/6255, simplify pdatagen to encourage the same naming between pdata and proto fields.

### DIFF
--- a/pdata/internal/cmd/pdatagen/internal/base_fields.go
+++ b/pdata/internal/cmd/pdatagen/internal/base_fields.go
@@ -19,9 +19,9 @@ import (
 	"strings"
 )
 
-const accessorSliceTemplate = `// ${fieldName} returns the ${originFieldName} associated with this ${structName}.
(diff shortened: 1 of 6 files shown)


bogdandrutu (approved):

bogdandrutu:
@dmitryax do we have any exception?

dmitryax:
> do we have any exception besides the Value/Slice/Map?

Yes, we have a few:
- OneOf fields like [these](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L359-L360) in proto are shorten to `Int/Double`
- `SpanId/TraceId `in proto -> `SpanID/TraceID`
- `*TimeUnixNano` in proto -> `*Timestamp`

I believe they are all good

bogdandrutu:
@dmitryax debating if they are good or not, we can have later, but for the moment we should document them to the exceptions.

PR #6285, opened 2022-10-12
[chore] Consistent pattern usage for command flags

bogdandrutu:


service/collector_windows.go
```diff
@@ -59,6 +58,11 @@ func (s *windowsService) Execute(args []string, requests <-chan svc.ChangeReques
 		return false, 1501 // 1501: ERROR_EVENTLOG_CANT_START
 	}
 
+	// Parse all the flags manually.
+	if err = s.flags.Parse(args[1:]); err != nil {
+		return false, 1639 // 1639: ERROR_INVALID_COMMAND_LINE
```
jpkrohling, line 63: Is there a list of error codes somewhere?

jpkrohling (approved):
LGTM, but where are the magic numbers coming from?

service/collector_windows.go
bogdandrutu: There are some, here are the links I used:
https://efmsoft.com/what-is/amp/?code=1639

cannot find the whole list, but that is good for this.

PR #6292, opened 2022-10-12
Replace manually adding tags with multimod

bogdandrutu:
Also, only creates tags for the modules not marked as ignored in the version file.

For the moment I cannot find way to push these tags, filed https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/157

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,35 +165,16 @@ ifndef COMPONENT
 	$(error COMPONENT variable was not defined)
 endif
 
-.PHONY: add-tag
-add-tag:
-	@[ "${TAG}" ] || ( echo ">> env var TAG is not set"; exit 1 )
-	@echo "Adding tag ${TAG}"
-	@git tag -a ${TAG} -s -m "Version ${TAG}"
-	@set -e; for dir in $(ALL_MODULES); do \
-	  (echo Adding tag "$${dir:2}/$${TAG}" && \
-	 	git tag -a "$${dir:2}/$${TAG}" -s -m "Version ${dir:2}/${TAG}" ); \
-	done
-
-.PHONY: push-tag
-push-tag:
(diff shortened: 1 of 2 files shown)


codeboten (approved):
Is the tag deletion in this makefile no longer useful to have? I'm not sure how frequently it's been used (if at all)

This change will require an update to https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#releasing-opentelemetry-collector

bogdandrutu:
@codeboten until I find a way to push the tags will hold on this for the moment

bogdandrutu:
@codeboten @dmitryax please take a look.

Makefile
```diff
@@ -165,35 +165,16 @@ ifndef COMPONENT
...
-	done
+COMMIT?=HEAD
+MODSET?=stable
+REMOTE?=git@github.com:open-telemetry/opentelemetry-collector.git
+.PHONY: push-tags
+push-tags:
+	multimod verify
+	set -e; for tag in `multimod tag -m ${MODSET} -c ${COMMIT} --print-tags | grep -v "Using" `; do \
+		echo "pushing tag $${tag}"; \
+		git push ${REMOTE} $${tag}; \
```
dmitryax, line 176: I wish we had a dry-run option here but I think `multimod tag ...` will fail on a second run... Do you think it's something worth adding to multimod tag?

dmitryax (approved):

Makefile
bogdandrutu: we can ask for that, please open an issue in the build-go repos.

dmitryax: Submitted https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/205

Issue #5687, opened 2022-07-14
[Telemetry] Collector telemetry not including opencensus metrics along with internal metrics when feature flag useOtelForInternalMetrics is true


timannguyen:
**Describe the bug**
using flag --feature-gates=telemetry.useOtelForInternalMetrics using [demo](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/examples/demo) or through 

```go
featuregate.Apply(map[string]bool{
	"telemetry.useOtelForInternalMetrics": true,
})
```

, calling http://localhost:8888/metrics does not displays the metrics mentioned in[ monitoring documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md).

**Steps to reproduce**

1. set config yaml with below or through code, set the config map service::telemetry::metrics::address: ":8888"

```yaml
service:
  telemetry:
    metrics:
      address: ":8888"
```

3. set flag with --feature-gates=telemetry.useOtelForInternalMetrics or in code set telemetry.useOtelForInternalMetrics  to true in featuregate.
4.  go to http://localhost:8888/metrics
5. it will return empty 200 or if any metrics was set through featuregate, only those are shown. Ones from [ monitoring documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md) does not appear.

**What did you expect to see?**

I expected metrics from [ monitoring documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md) and any internal metrics added using otel metrics sdk.

EXAMPLE: 

```txt
# HELP otelcol_exporter_enqueue_failed_log_records Number of log records failed to be added to the sending queue.
# TYPE otelcol_exporter_enqueue_failed_log_records counter
otelcol_exporter_enqueue_failed_log_records{exporter="jaeger",service_instance_id="71249946-a6bd-4e88-86e3-6c310b093c54",service_version="latest"} 0
otelcol_exporter_enqueue_failed_log_records{exporter="logging",service_instance_id="71249946-a6bd-4e88-86e3-6c310b093c54",service_version="latest"} 0
otelcol_exporter_enqueue_failed_log_records{exporter="prometheus",service_instance_id="71249946-a6bd-4e88-86e3-6c310b093c54",service_version="latest"} 0
[...]

PR #6297, opened 2022-10-13
Use the same  prometheus registry for both oc and otel exporters

paivagustavo:
**Description:** Use the same `prometheus.Registry` for both OC and OTel SDK exporters.

**Link to tracking Issue:** #5687

**Testing:** Added unit test and did some manual tests with and without the featuregate to ensure that nothing is broken.

**Documentation:** The observed difference between both exporters is how they export Resource Attributes. OpenCensus appends resource attributes to all metrics while [OpenTelemetry advocates](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#resource-attributes-1) for the use of an additional `target_info` metric which contains all resource attributes and other metrics are only appended with `job` and `instance` which can be derived from the `service.name`, `service.namespace`, and `service.instance.id attributes`.

OTel-Go doesn't implement the `target_info` yet, but I've added some notes and TODOs to track those.


service/telemetry.go
```diff
@@ -205,12 +214,13 @@ func (tel *telemetryInitializer) initOpenCensus(cfg telemetry.Config, telAttrs m
...
 	if err := view.Register(views...); err != nil {
-		return nil, err
+		return err
 	}
 
 	// Until we can use a generic metrics exporter, default to Prometheus.
 	opts := ocprom.Options{
-		Namespace: "otelcol",
+		Namespace:  "otelcol",
+		Registerer: promRegistry,
```
bogdandrutu, line 223: Not sure if it is not better to pass the full `Registry` here. Have not checked the code to see if it has any usage of the Gatherer/Registry. I think it is safer to pass the `Registry: registry`

bogdandrutu: And even maybe use for the moment the Handler from the OC, since I am not 100% if there is no side effect there..

paivagustavo: The `Gatherer` is just used if we want to use the exporter itself as the handler.

bogdandrutu: :)) famous last words. I am not confident, so still prefer to send the entire Registry to OC and use the Handler form there, since OC is the current state that guarantees me that no new code is used. Look at the PR I created.

paivagustavo: I did check and couldn't side effects for not using the gatherer. I can double check to make sure, but since we would want to get rid of the oc exporter in the future, we would to create our handler anyway.

paivagustavo: Sure! I understand your point :)

bogdandrutu (approved):

bogdandrutu:
Please update descriptions, and add some unit tests

paivagustavo:
@bogdandrutu I've added some unit tests, not sure if that was the best place, open for suggestions if you have any. 

Also documented on the tests the differences between how both SDK exports Resource attributes. I'm not sure what is the best way for us to document this public-wise, since it can be considered as a breaking change for some.

bogdandrutu:
I would prefer proper unittests, I know we have some kind of "integration" tests there, but they are actually complicated and hard to follow. I would rather remove them and have proper unit-tests.

service/telemetry.go
```diff
@@ -84,22 +83,28 @@ func (tel *telemetryInitializer) init(buildInfo component.BuildInfo, logger *zap
...
-	if cfg.Metrics.Level == configtelemetry.LevelNone || cfg.Metrics.Address == "" {
-		logger.Info(
-			"Skipping telemetry setup.",
-			zap.String(zapKeyTelemetryAddress, cfg.Metrics.Address),
-			zap.String(zapKeyTelemetryLevel, cfg.Metrics.Level.String()),
-		)
-		return nil
-	}
-
+func (tel *telemetryInitializer) initOnce(buildInfo component.BuildInfo, logger *zap.Logger, cfg telemetry.Config) error {
```
paivagustavo, line 107: I've moved some of the code from `initOnce` in order to avoid starting a http server for the prometheus handler, since that is already captured by the tests in `collector_test.go`

service/telemetry_test.go
```diff
@@ -67,3 +87,133 @@ func TestBuildTelAttrs(t *testing.T) {
...
+				require.NoError(t, tel.shutdown())
+			}()
+
+			v := createTestMetrics(t, tel.mp)
+			defer func() {
+				view.Unregister(v)
+			}()
+
+			// Wait some time because OpenCensus can take sometime to set up and collect a measure.
+			time.Sleep(10 * time.Millisecond)
```
paivagustavo, line 162: I had to add this casual wait because sometimes we're not getting OpenCensus metrics. I ran this a few times with `-test.count 1000` and it would just randomly fail. Adding a millisecond wait was enough for me to stop seeing it, but just for safety I bumped it to 10 milliseconds.

I suspect that this does not affect the existing integration simply because they do more work before starting to test.

bogdandrutu: I think if you call `view.RetrieveData(ocPrefix + counterName)` this will work

service/telemetry_test.go
```diff
@@ -67,3 +87,133 @@ func TestBuildTelAttrs(t *testing.T) {
...
+	}
+}
+
+func createTestMetrics(t *testing.T, mp metric.MeterProvider) *view.View {
+	// Creates a OTel Go counter
+	counter, err := mp.Meter("collector_test").SyncInt64().Counter(otelPrefix + counterName)
+	require.NoError(t, err)
+	counter.Add(context.Background(), 13)
+
+	// Creates a OpenCensus counter
```
bogdandrutu, line 191: This is not an opencensus "counter". This is an opencensus measure + sum aggregation view :)

service/telemetry_test.go
paivagustavo: That did the trick, thanks @bogdandrutu!

service/telemetry_test.go
```diff
@@ -67,3 +86,133 @@ func TestBuildTelAttrs(t *testing.T) {
...
+		Description: ocCounter.Description(),
+		Measure:     ocCounter,
+		Aggregation: view.Sum(),
+	}
+	err = view.Register(v)
+	require.NoError(t, err)
+
+	stats.Record(context.Background(), stats.Int64(ocPrefix+counterName, counterName, stats.UnitDimensionless).M(13))
+
+	_, _ = view.RetrieveData(ocPrefix + counterName)
```
bogdandrutu, line 200:
```suggestion
	// Forces a flush for the view data.
	_, _ = view.RetrieveData(ocPrefix + counterName)
```

bogdandrutu (approved):

Issue #5966, opened 2022-08-25
Configuration reloading logic and customizability


swiatekm-sumo:
**Is your feature request related to a problem? Please describe.**
Right now, the collector reloads configuration whenever it [gets a notification](https://github.com/open-telemetry/opentelemetry-collector/blob/7545a294c9cf5f483e7a0f45034dae2c9bd9d089/service/collector.go#L173) from any of its configuration sources. In practice, this never happens, as watching is not implemented in any of the [existing config providers](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider), but it will the moment that is added.

I don't think this is behaviour users would universally want. In many contexts, you only want to reload the configuration on-demand. Kubernetes is one example of this, where mounted ConfigMaps can be remounted if they change, but it's not desirable behaviour for an application to reload the config without the Pod having been recreated. Running as a systemd unit is another example.

I'd like to figure out what is the best way of letting the user control how the collector reloads configuration. I'd like to explore different approaches - for example, many applications do this on receipt of a SIGHUP signal.

**Describe the solution you'd like**
I don't know what the right set of solutions is, but I think it should allow:
* the user to decide if they want automatic configuration reload based on provider notification, maybe per provider
* the user to be able to reload configuration on demand without restarting the collector process
* (maybe?) for the collector to keep using older working configuration if the new one is invalid 

In particular, I'm not sure what the default should be.

**Describe alternatives you've considered**
Some possible solutions:
* let automatic reloading be controlled by command line options
* let automatic reloading be controlled by a section in the configuration
* unconditionally reload configuration on receipt of a SIGHUP signal
* keep the last known working configuration and fall back to it if the new one is invalid

I'm open to other ideas.

**Additional context**
This originally came up because I wanted to add notifications to the fileprovider in #5945 . @TylerHelmuth suggested it should start off behind a feature flag, and I reconsidered whether it should even ultimately be enabled for all users. I looked at how this is done in some other agents, and this issue is the result.

bogdandrutu:
> Kubernetes is one example of this, where mounted ConfigMaps can be remounted if they change, but it's not desirable behaviour for an application to reload the config without the Pod having been recreated. Running as a systemd unit is another example.

I am confused about this, I don't think the remounted happens for the existing pods. Imagine an app crash, then restart will pick the new version? Can you investigate a bit more?

> (maybe?) for the collector to keep using older working configuration if the new one is invalid
[...]

PR #4454, opened 2021-11-18
Add file watch to support config reload on file change

vjsamuel:
**Description:**
Fixes: https://github.com/open-telemetry/opentelemetry-collector/issues/4397

This PR allows the main config.yml to be reloaded each time the config file changes. The entire pipeline gets reloaded.

This PR doesn't use an FS notify/inotify style watcher as we have seen that Kubernetes doesnt support such watches when a config is mounted as a config file. os.Stat given that it is cheap can be run every second to trigger a reload.

PR #6300, opened 2022-10-13
Add a flag for enabling config watch

swiatekm-sumo:
**Description:**
Add a flag to control configuration watching. 

Technically this is a breaking change, as the flag defaults to false, and this is currently unconditionally enabled. In practice, none of the currently available config sources implement notifications, so users shouldn't see any difference unless they're using a distribution which specifically added such a config source. This being disabled by default will also allow changes like #5945 to be merged without affecting users unless they opt-in to the behaviour.

I'm open to suggestions for the flag name.

**Link to tracking Issue:** #5966 

**Testing:**
Added a unit test for the default (disabled), and updated the one checking for the other case.

**Documentation:** 
Do I need to update this in https://github.com/open-telemetry/opentelemetry.io?

### DIFF
--- a/.chloggen/config-hot-reloading.yaml
+++ b/.chloggen/config-hot-reloading.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service/collector
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add a flag enabling configuration watch
+
+# One or more tracking issues or pull requests related to the change
+issues: [5966]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -61,6 +61,8 @@ type Resolver struct {
 
 	closers []CloseFunc
 	watcher chan error
+
+	enableConfigWatch bool
 }
 
 // ResolverSettings are the settings to configure the behavior of the Resolver.
@@ -75,6 +77,9 @@ type ResolverSettings struct {
 
 	// MapConverters is a slice of Converter.
 	Converters []Converter
+
+	// EnableConfigWatch enables reloading configuration when a provider reports a change
+	EnableConfigWatch bool
 }
 
 // NewResolver returns a new Resolver that resolves configuration from multiple URIs.
@@ -132,10 +137,11 @@ func NewResolver(set ResolverSettings) (*Resolver, error) {
 	copy(convertersCopy, set.Converters)
 
 	return &Resolver{
-		uris:       uris,
-		providers:  providersCopy,
-		converters: convertersCopy,
-		watcher:    make(chan error, 1),
+		uris:              uris,
+		providers:         providersCopy,
+		converters:        convertersCopy,
+		watcher:           make(chan error, 1),
+		enableConfigWatch: set.EnableConfigWatch,
 	}, nil
 }
 
@@ -348,5 +354,11 @@ func (mr *Resolver) retrieveValue(ctx context.Context, uri location) (*Retrieved
 	if !ok {
 		return nil, fmt.Errorf("scheme %q is not supported for uri %q", uri.scheme, uri.asString())
 	}
-	return p.Retrieve(ctx, uri.asString(), mr.onChange)
+	var watcherFunc WatcherFunc
+	if mr.enableConfigWatch {
+		watcherFunc = mr.onChange
+	} else {
+		watcherFunc = nil
+	}
+	return p.Retrieve(ctx, uri.asString(), watcherFunc)
 }
--- a/confmap/resolver_test.go
+++ b/confmap/resolver_test.go
@@ -41,8 +41,9 @@ func (m *mockProvider) Retrieve(_ context.Context, _ string, watcher WatcherFunc
 	if m.retM == nil {
 		return NewRetrieved(nil)
 	}
-
-	watcher(&ChangeEvent{Error: m.errW})
+	if watcher != nil {
+		watcher(&ChangeEvent{Error: m.errW})
+	}
 	return NewRetrieved(m.retM, WithRetrievedClose(func(ctx context.Context) error { return m.errC }))
 }
 
@@ -173,7 +174,12 @@ func TestResolverErrors(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			resolver, err := NewResolver(ResolverSettings{URIs: tt.locations, Providers: makeMapProvidersMap(tt.providers...), Converters: tt.converters})
+			resolver, err := NewResolver(ResolverSettings{
+				URIs:              tt.locations,
+				Providers:         makeMapProvidersMap(tt.providers...),
+				Converters:        tt.converters,
+				EnableConfigWatch: true,
+			})
 			if tt.expectBuildErr {
 				assert.Error(t, err)
 				return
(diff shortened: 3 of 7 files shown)


service/flags.go
```diff
@@ -22,13 +22,15 @@ import (
 )
 
 const (
-	configFlag = "config"
-	setFlag    = "set"
+	configFlag    = "config"
+	setFlag       = "set"
+	hotReloadFlag = "enable-hot-reload"
```
bogdandrutu, line 27: "config-autoreload"?

I don't like necessarily the "hot" part of the name. "auto" sounds better to me.

Also important to mention what is reloaded :)

service/collector.go
```diff
@@ -172,6 +172,11 @@ LOOP:
 	for {
 		select {
 		case err := <-col.set.ConfigProvider.Watch():
+			if !col.set.EnableConfigHotReload {
+				col.service.telemetrySettings.Logger.Warn("Config updated, but hot reloading is disabled, ignoring")
+				continue
+			}
```
bogdandrutu, line 178: Not sure about this, it may spam the logging system.

swiatekm-sumo: It might, but I think it's still useful information that config sources keep changing and the collector is ignoring them. In most cases I think this will alert users that they need to enable reloading if they want it.

service/flags.go
swiatekm-sumo: Hm, how about just `--watch-config`? That's consistent with our internal terminology, and reasonably descriptive. I'm fine with `--enable-auto-reload` as well though.

service/flags.go
```diff
@@ -22,13 +22,15 @@ import (
...
 )
 
 var (
 	// Command-line flag that control the configuration file.
-	gatesList = featuregate.FlagValue{}
+	gatesList        = featuregate.FlagValue{}
+	hotReloadEnabled bool
```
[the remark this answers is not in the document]
swiatekm-sumo: Done. Would be less awkward if `flags` exposed its own `boolValue`, but it's not too much additional code.

service/flags.go
```diff
@@ -64,6 +82,12 @@ func flags() *flag.FlagSet {
 	flagSet.Var(featuregate.FlagValue{}, featureGatesFlag,
 		"Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.")
 
+	flagSet.Var(
+		new(boolValue),
+		watchConfigFlag,
+		"Enable config watch for configuration sources.",
+	)
```
bogdandrutu, line 89: flagSet.BoolVar(
		new(bool),
		watchConfigFlag,
		"Enable config watch for configuration sources.",
	)

service/flags.go
```diff
@@ -75,3 +99,7 @@ func getConfigFlag(flagSet *flag.FlagSet) []string {
 func getFeatureGatesFlag(flagSet *flag.FlagSet) featuregate.FlagValue {
 	return flagSet.Lookup(featureGatesFlag).Value.(featuregate.FlagValue)
 }
+
+func getWatchConfigFlag(flagSet *flag.FlagSet) bool {
+	return bool(*flagSet.Lookup(watchConfigFlag).Value.(*boolValue))
```
bogdandrutu, line 104: return flagSet.Lookup(watchConfigFlag).Value.(flag.Getter).Get().(bool)

service/collector.go
bogdandrutu: And if they don't want auto-reload? Do we just spam with "warnings"?

service/collector.go
```diff
@@ -172,6 +172,11 @@ LOOP:
 	for {
 		select {
 		case err := <-col.set.ConfigProvider.Watch():
+			if !col.set.EnableConfigWatch {
```
bogdandrutu, line 175: I thought more about this feature, and found that if we have `EnableConfigWatch` a config in the `ConfigProviderSettings`, then send a "nil" watcher that does disables "watching" completely.

What do you think?

service/collector.go
swiatekm-sumo: Is there any situation where this is actually going to be spam? To begin with, it should be very rare for a config source to change very frequently. My thinking was that this would rarely produce an unwanted message for some users, and catch misconfigurations for others. If you think that's not a good tradeoff, I can remove it, it's not that important.

service/collector.go
swiatekm-sumo: I thought about it, and ended up putting this in the collector loop for two reasons:
* It's more visible and explicit this way. If it was part of `ConfigProvider` logic, then we'd have a case in the collector loop that implicitly wouldn't do anything depending on settings.
* It's possible for distributions to provide their own `ConfigProvider` implementations, and in that case we'd be forcing them to implement this logic on their own.

That said, I don't think it really matters *that* much where it is.

bogdandrutu: I would go this way for the moment... Feels more natural to be there... Happy to hear other opinions.

swiatekm-sumo:
@bogdandrutu are there any other concerns I should address before we can get this merged?

service/collector.go
jpkrohling: INFO level would be more appropriate. On Kubernetes, those events can be quite frequent, as there are 3 or 4 levels of symlink between the mount path and the actual file, and any changes to that would trigger an event.

service/collector.go
jpkrohling: I'm with @bogdandrutu here. Disabling (or not enabling) the config watch should not have the file watched at all. I understand the sentiment, but I also prefer to not have this code executed at all unless the config watch is enabled.

service/flags.go
bogdandrutu: `--watch-config` is fine

service/collector.go
bogdandrutu: @swiatekm-sumo we are blocked on this :)

swiatekm-sumo: I had misunderstood your original comment @bogdandrutu, and thought you wanted to leave this as is. Sorry about that, I'm fine moving this check to a lower level.

If we want to do no additional work at all if this is disabled, then we'd need to pass this value down the the Resolver and indicate to the provider we don't want notifications by passing a nil watcher function here: https://github.com/open-telemetry/opentelemetry-collector/blob/ae077bec51f81e838bacbb522f5711636f7a83e0/confmap/resolver.go#L312. Does that sound reasonable?

service/collector.go
swiatekm-sumo: After moving the change to Resolver, this doesn't exist anymore, as we don't get the notifications to begin with.

jpkrohling (approved):
LGTM

swiatekm-sumo:
I rebased the changes. `codecov` is complaining about the additional lines in the command handler not being covered by tests, which is correct in that there's no test for it, but I'm not sure if I should add it or leave as is.

phrodrigues77:
Hi! Will this be merged soon?
The flag along with the configuration reloading [5966](https://github.com/open-telemetry/opentelemetry-collector/issues/5966) would be much appreciated!

jpkrohling:
ping @open-telemetry/collector-approvers

Aneurysm9 (approved):

dmitryax (approved):
LGTM

dmitryax:
@bogdandrutu are you ok with merging it?

dmitryax:
After giving it a second thought, I don't think this change is necessary. I'd rather lean towards keeping the watch always enabled. But I believe it's important to introduce the ability to watch files as suggested in https://github.com/open-telemetry/opentelemetry-collector/pull/5945 and https://github.com/open-telemetry/opentelemetry-collector/pull/4454.

My reasoning is that I, as a user, may want to watch one file but ignore updates in another. Another concern is that AFAIR, we've been trying to avoid introducing command-line flags and rely on the yaml config as much as possible.

Maybe we can have different config provider prefixes (schema parts) for files that are watched and not? Or have some additional syntax for flags like disable/enable watch in the config provider invocation strings?

@swiatekm-sumo, maybe we bring it for discussion to one of the next SIGs?

swiatekm-sumo:
> After giving it a second thought, I don't think this change is necessary. I'd rather lean towards keeping the watch always enabled. But I believe it's important to introduce the ability to watch files as suggested in #5945 and #4454.
> 
> My reasoning is that I, as a user, may want to watch one file but ignore updates in another. Another concern is that AFAIR, we've been trying to avoid introducing command-line flags and rely on the yaml config as much as possible.
> 
> Maybe we can have different config provider prefixes (schema parts) for files that are watched and not? Or have some additional syntax for flags like disable/enable watch in the config provider invocation strings?
> 
> @swiatekm-sumo, maybe we bring it for discussion to one of the next SIGs?

Sure, let's discuss during the sig. Offline, https://github.com/open-telemetry/opentelemetry-collector/issues/5966 seems to be the best place to talk about this. I actually thought about this as well, and I'm not convinced the automated config reloading is even a good idea (and afaik no other open source monitoring agents do it), and there's a lot of room for different solutions.

Part of the reason the solution in this PR was originally proposed, is that it covers all the providers at once. Even if you introduce new syntax to make this customizable per-file, you still need to cover other providers which may implement notifications in the future.

I'll make a longer and more comprehensive comment about this in the linked issue.

Issue #5878, opened 2022-08-09
exporter/loggingexporter.Config.LogLevel has different behavior after #5677


bogdandrutu:
After https://github.com/open-telemetry/opentelemetry-collector/pull/5677 was merged, we no longer use `LogLevel` the way was before (where we set it for zap.Core). 

We need to look into what that configuration does right now, if we need to deprecate/rename/re-describe or what we need to do.

mx-psi:
The new behavior or the log level option is described on this comment https://github.com/open-telemetry/opentelemetry-collector/pull/5677#pullrequestreview-1036046290.

We have only ever used `debug` and `info` levels, and I doubt we would use `warn` or `error` here. Some possible renames:
- `verbosity`or `verbosity_level` (either keeping `info` and `debug` or maybe with `brief` and `detailed` values)
- `verbose`, `detailed` or `show_payload_contents` (boolean flag, defaults to `false`)

@bogdandrutu (and also @tigrannajaryan ) Do you prefer any particular option among those?

bogdandrutu:
I think I like the "verbosity" level. The reason is that, it may allow us to add new levels if needed in the future. Also look into https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtelemetry/configtelemetry.go#L26 may use that?

mx-psi:
Take a look at #6334

mx-psi:
Fixed by #6334

PR #6334, opened 2022-10-17
[exporter/logging] Deprecate 'loglevel' in favor of 'verbosity'

mx-psi:
**Description:** 

- Add `verbosity` option
- Deprecate `loglevel` in favor of verbosity option
- Add warning log if `loglevel` option is used

**Link to tracking Issue:** Updates #5878


exporter/loggingexporter/config.go
```diff
@@ -15,28 +15,96 @@
...
+	}
+
+	expected, err := mapLevel(cfg.LogLevel)
+	if err != nil {
+		return fmt.Errorf("failed to map 'loglevel': %w", err)
+	}
+
+	if cfg.Verbosity != expected {
+		return fmt.Errorf("verbosity %q does not match loglevel %q", cfg.Verbosity, cfg.LogLevel)
+	}
```
bogdandrutu, line 107: Please add a test where only the "verbosity" is set to see what happens here...

exporter/loggingexporter/testdata/config_verbosity.yaml
```diff
@@ -0,0 +1,3 @@
+verbosity: detailed
+sampling_initial: 10
+sampling_thereafter: 50
```
bogdandrutu, line 3: please add new line, since I don't like this github sign :)

exporter/loggingexporter/config.go
```diff
@@ -15,28 +15,87 @@
...
 
 // Validate checks if the exporter configuration is valid
 func (cfg *Config) Validate() error {
+	switch cfg.Verbosity {
+	case configtelemetry.LevelBasic, configtelemetry.LevelNormal, configtelemetry.LevelDetailed:
+		// These are supported. Do nothing.
+	default:
+		// configtelemetry.LevelNone and future values are not supported.
+		return fmt.Errorf("verbosity level %q is not supported", cfg.Verbosity)
+	}
```
bogdandrutu, line 98: nit: somehow this is difficult to read, I really like more the style "if !ok { return }" ways of checking the conditions. If you have an idea how to do that please do, otherwise it is fine.

exporter/loggingexporter/config.go
```diff
@@ -15,28 +15,87 @@
...
+	Verbosity configtelemetry.Level `mapstructure:"verbosity"`
+
 	// SamplingInitial defines how many samples are initially logged during each second.
 	SamplingInitial int `mapstructure:"sampling_initial"`
 
 	// SamplingThereafter defines the sampling rate after the initial samples are logged.
 	SamplingThereafter int `mapstructure:"sampling_thereafter"`
+
+	// warnLogLevel is set on unmarshaling to warn users about `loglevel` usage.
+	warnLogLevel bool
```
bogdandrutu, line 45: i know that because is private the mapstructure most likely does not unmarshal it, but prefer safe than sorry, can you add a `mapstructure:-` (I think that is the annotation to ignore a field).

bogdandrutu (approved):

exporter/loggingexporter/config.go
mx-psi: Done with a set (map to empty struct)

exporter/loggingexporter/config.go
mx-psi: This does not ignore the field: https://go.dev/play/p/yTXlhIa4IEq it only works with `encoding/json`, but mapstructure does not ignore such a field

mx-psi: Should I still add it, if only for consistency?

bogdandrutu (approved):

exporter/loggingexporter/config.go
```diff
@@ -15,28 +15,91 @@
...
+	}
+
+	return nil
+}
 
 // Validate checks if the exporter configuration is valid
 func (cfg *Config) Validate() error {
+	// Supported configtelemetry.Level values for this exporter.
+	// configtelemetry.LevelNone and future values are not supported.
+	supportedLevels := map[configtelemetry.Level]struct{}{
```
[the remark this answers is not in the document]
mx-psi: done

Issue #6338, opened 2022-10-18
JSON Marshaler must emit enums as ints


bogdandrutu:
See https://github.com/open-telemetry/opentelemetry-specification/pull/2758

PR #6339, opened 2022-10-18
JSON marshaler emits enums as ints per spec reuqirements.

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6338

### DIFF
--- a/.chloggen/enumsasints.yaml
+++ b/.chloggen/enumsasints.yaml
@@ -1,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: pdata
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: JSON marshaler emits enums as ints per spec reuqirements. This may be a breaking change if receivers were not confirming with the spec.
+
+# One or more tracking issues or pull requests related to the change
+issues: [6338]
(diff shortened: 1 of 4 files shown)


.chloggen/enumsasints.yaml
```diff
@@ -0,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
dmitryax, line 2: Isn't this still a breaking change?

dmitryax (commented):
Thanks for fixing it

pdata/plog/json.go
```diff
@@ -26,7 +26,9 @@ import (
 
 // NewJSONMarshaler returns a Marshaler. Marshals to OTLP json bytes.
 func NewJSONMarshaler() Marshaler {
-	return &jsonMarshaler{delegate: jsonpb.Marshaler{}}
+	return &jsonMarshaler{delegate: jsonpb.Marshaler{
+		EnumsAsInts: true,
```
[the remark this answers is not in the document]
bogdandrutu: We have, but we test roundtrip and that works since the unmarshaler supports both.

.chloggen/enumsasints.yaml
bogdandrutu: The protocol required (no longer) to support to unmarshal both, so not sure why this is a "breaking change". Can mark if you feel that way.

pdata/plog/json.go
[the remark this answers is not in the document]
dmitryax: I saw the e2e test. I was thinking about `NewJSONMarshaler` -> json test with enums

.chloggen/enumsasints.yaml
dmitryax: Ok sounds good.

dmitryax (approved):

PR #6340, opened 2022-10-18
Deprecate New[JSON|Proto][Marshaler|Unmarshale], expose the structs

bogdandrutu:
The main motivation is because we want to be able to extend these Marshaler/Unmarshaler implementation with possible options.

Because of that we have 2 options: 1. add Option to each New func, OR 2. expose the structs so that later users can configure different options inside these structs, similar with `jsonpb.Marshaler`.

I implemented the version 2 since it is simpler, and less code, and also common in the industry.

### DIFF
--- a/.chloggen/structs.yaml
+++ b/.chloggen/structs.yaml
@@ -1,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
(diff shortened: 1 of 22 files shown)


bogdandrutu:
> I agree, the only concern that we use initializers all around pdata :)

We don't have any other case like this, we do use initializers for "empty" objects. Also these don't need to have guarantees of not sharing code etc. as the main pdata structs. So I don't see any concern, feel that these structs are completely different than all the others.

dmitryax (approved):

PR #6342, opened 2022-10-18
Add hack to continue to read protobuf id 1000 in the resource messages

bogdandrutu:


dmitryax (approved):
It would be nice to have some test coverage, but given it's hack for to support deprecated fields, It's probably good to go as is

bogdandrutu:
> It would be nice to have some test coverage, but given it's hack for to support deprecated fields, It's probably good to go as is

Done.

dmitryax (approved):

PR #6343, opened 2022-10-18
[wip] split exporters

codeboten:
This PR is a test to see if there are any issues with using the `exporter` module once the other exporters have been split into their own modules. It doesn't appear there are any issues with linting.


.chloggen/split-exporters.yaml
```diff
@@ -0,0 +1,21 @@
...
+note: "Splitting otlp, otlphttp and logging exporters into their own modules"
+
+# One or more tracking issues or pull requests related to the change
+issues: [6343]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |-
+  The import path for these exporters can now be access directly:
```
Aneurysm9, line 17:
```suggestion
  The import path for these exporters can now be accessed directly:
```

or

```suggestion
  The import path for these exporters can now be required directly as modules:
```

exporter/go.mod
```diff
@@ -0,0 +1,47 @@
...
+go 1.18
+
+require (
+	github.com/cenkalti/backoff/v4 v4.1.3
+	github.com/stretchr/testify v1.8.0
+	go.opencensus.io v0.23.0
+	go.opentelemetry.io/collector v0.0.0-00010101000000-000000000000
+	go.opentelemetry.io/collector/pdata v0.62.1
+	go.opentelemetry.io/otel v1.11.0
+	go.opentelemetry.io/otel/sdk v1.11.0
```
Aneurysm9, line 12: Not immediately problematic, but this direct dependency on the OTel Go SDK in an instrumented library that should only use the API is something we recommend avoiding.  In the go-contrib repo we have done this by having a subordinate `test` module that can depend on the SDK tracetest packages and the like.  It presents a good and reliable signal to consumers that an instrumentation library is not going to attempt to modify your SDK configuration when it doesn't require the SDK module at all.

.chloggen/split-exporters.yaml
codeboten, line 17:
```suggestion
  The import path for these exporters can now be required directly as modules:
```

codeboten: 👍🏻

exporter/go.mod
codeboten: `exporter/go.mod` is no longer in this PR, but I see what you're saying. as the exporterhelper is back in the collector module (which depends on the sdk for different reasons), i think it's ok for now, but if the exporterhelper gets split into its own module, we could do what you're suggesting here to avoid taking on a dependency on the sdk for the tests alone.

.chloggen/split-exporters.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - `go.opentelemetry.io/collector/exporters/loggingexporter`
```
jpkrohling, line 18:
```suggestion
  - `go.opentelemetry.io/collector/exporter/loggingexporter`
```

Same for the next two lines.

versions.yaml
```diff
@@ -18,6 +18,9 @@ module-sets:
     modules:
       - go.opentelemetry.io/collector
       - go.opentelemetry.io/collector/cmd/builder
+      - go.opentelemetry.io/collector/exporter/loggingexporter
```
jpkrohling, line 21: Was this added manually? Or is this the one that is managed by multimod?

codeboten: Manually to get `multimod-verify` to pass

.chloggen/split-exporters.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - `go.opentelemetry.io/collector/exporters/otlpexporter`
+  - `go.opentelemetry.io/collector/exporters/otlphttpexporter`
```
codeboten, line 20:
```suggestion
  - `go.opentelemetry.io/collector/exporter/otlpexporter`
  - `go.opentelemetry.io/collector/exporter/otlphttpexporter`
```

.chloggen/split-exporters.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |-
+  The import path for these exporters can now be access directly:
+  - `go.opentelemetry.io/collector/exporter/loggingexporter`
+  - `go.opentelemetry.io/collector/exporter/otlpexporter`
+  - `go.opentelemetry.io/collector/exporter/otlphttpexporter`
+
```
mx-psi, line 21: I think we should explicitly mention how to migrate builder configuration:

```suggestion
  
  If using these exporters, modify your Collector builder configuration to use `gomod` directly, such as:
  -  `gomod: go.opentelemetry.io/collector/exporter/otlpexporter v0.63.0`
```

Maybe even pin an issue with instructions about how to do this in this repository, or a small script to go from the old configuration to the new one

mx-psi: Yes, that makes sense :+1:

codeboten: Issue pinned and description of the issue updated to include the instructions

mx-psi:
How do we make contrib tests pass? I don't see a way out of it at the moment

codeboten:
> How do we make contrib tests pass? I don't see a way out of it at the moment

This may be a good reason to use `exporters` rather than `exporter` in the package name. It would allow us to provide a deprecation path by adding aliases in `exporter` pointing users to migrate to `exporters`

Any thoughts @jpkrohling @bogdandrutu @dmitryax @Aneurysm9?

@mx-psi the alternative is to update contrib in a follow up PR.

exporter/Makefile
```diff
@@ -0,0 +1 @@
+include ../Makefile.Common
```
bogdandrutu, line 1: Why is this file necessary?

codeboten: It is not, it was leftover from the testing of having `exporter` as its own module

mx-psi:
> @mx-psi the alternative is to update contrib in a follow up PR.

IMO this is okay, but we would need a mechanism for skipping contrib tests then (`Skip contrib` label?)

codeboten:
> > @mx-psi the alternative is to update contrib in a follow up PR.
> 
> IMO this is okay, but we would need a mechanism for skipping contrib tests then (`Skip contrib` label?)

I didn't realize "Skip contrib" was a label we had :D

bogdandrutu (approved):

PR #6351, opened 2022-10-19
Added cleanup if service errors on Start

cpheps:
**Description:** Same as #6239. Call `service.Shutdown` if `service.Start` call errors in collector. A failed `service.Start` would leave up the `telemetryInitializer` and the next attempt at starting the collector would cause an error as the `telemetryInitializer` didn't clean up global state.

I'm not sure if this interferes with work on #5564. I moved the `service.telemetryInitializer.Shutdown` into `service.Shutdown`. Currently it the `telemetryInitializer` needs to be shutdown separately from the service. This changes makes it so calling `service.Shutdown` ensures all its state is cleaned up.

Fixes: #6352

### DIFF
--- a/service/collector.go
+++ b/service/collector.go
@@ -153,16 +153,9 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	}
 
 	if err = col.service.Start(ctx); err != nil {
-		errs := err
-		if shutdownErr := col.service.Shutdown(ctx); shutdownErr != nil {
-			errs = multierr.Append(err, fmt.Errorf("failed to shutdown service after error: %w", shutdownErr))
+		if shutdownErr := col.shutdownServiceAndTelemetry(ctx); shutdownErr != nil {
+			return multierr.Append(err, shutdownErr)
 		}
-
-		// TODO: Move this as part of the service shutdown.
-		if shutdownErr := col.service.telemetryInitializer.shutdown(); shutdownErr != nil {
-			errs = multierr.Append(errs, fmt.Errorf("failed to shutdown collector telemetry: %w", shutdownErr))
-		}
-		return errs
 	}
 	col.setCollectorState(Running)
 	return nil
@@ -228,17 +221,30 @@ func (col *Collector) shutdown(ctx context.Context) error {
 		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown config provider: %w", err))
 	}
 
-	if err := col.service.Shutdown(ctx); err != nil {
-		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown service: %w", err))
-	}
-
-	// TODO: Move this as part of the service shutdown.
-	if err := col.service.telemetryInitializer.shutdown(); err != nil {
-		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown collector telemetry: %w", err))
+	if err := col.shutdownServiceAndTelemetry(ctx); err != nil {
+		errs = multierr.Append(errs, err)
 	}
 
 	col.setCollectorState(Closed)
 
+	return errs
+}
+
+// shutdownServiceAndTelemetry bundles shutting down the service and telemetryInitializer.
+// Returned error will be in multierr form and wrapped.
+func (col *Collector) shutdownServiceAndTelemetry(ctx context.Context) error {
+	var errs error
+
+	// shutdown service
+	if err := col.service.Shutdown(ctx); err != nil {
+		errs = multierr.Append(err, fmt.Errorf("failed to shutdown service after error: %w", err))
+	}
+
+	// TODO: Move this as part of the service shutdown.
+	// shutdown telemetryInitializer
+	if err := col.service.telemetryInitializer.shutdown(); err != nil {
+		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown collector telemetry: %w", err))
+	}
 	return errs
 }
 
--- a/service/collector.go
+++ b/service/collector.go
@@ -153,6 +153,9 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	}
 
 	if err = col.service.Start(ctx); err != nil {
+		if shutdownErr := col.service.Shutdown(ctx); shutdownErr != nil {
+			return multierr.Append(err, fmt.Errorf("failed to shutdown service after error: %w", shutdownErr))
+		}
 		return err
 	}
 	col.setCollectorState(Running)
@@ -221,11 +224,6 @@ func (col *Collector) shutdown(ctx context.Context) error {
 
 	if err := col.service.Shutdown(ctx); err != nil {
 		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown service: %w", err))
-	}
-
-	// TODO: Move this as part of the service shutdown.
-	if err := col.service.telemetryInitializer.shutdown(); err != nil {
-		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown collector telemetry: %w", err))
 	}
 
 	col.setCollectorState(Closed)
--- a/service/collector.go
+++ b/service/collector.go
@@ -153,10 +153,16 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	}
 
 	if err = col.service.Start(ctx); err != nil {
+		errs := err
 		if shutdownErr := col.service.Shutdown(ctx); shutdownErr != nil {
-			return multierr.Append(err, fmt.Errorf("failed to shutdown service after error: %w", shutdownErr))
+			errs = multierr.Append(err, fmt.Errorf("failed to shutdown service after error: %w", shutdownErr))
 		}
-		return err
+
+		// TODO: Move this as part of the service shutdown.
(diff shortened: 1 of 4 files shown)


cpheps:
> I think it is because we cannot restart everything, so hence we cannot shutdown everything unless we know for sure we have to turn off completely

@bogdandrutu  I'm not sure I completely follow this. My understanding is within a single process it's ok to stop an instance of the collector and start a new one as many times as you'd like. 

If for some reason an instance of the collector fails to start due to the service it's wrapping failing to init or start it can leave state around that would prevent another instance from starting. 

I actually just noticed too in the loop in `Run` if the `ConfigProvider.Watch` case is triggered currently the `telemetryInitializer` isn't shut down [here](https://github.com/open-telemetry/opentelemetry-collector/blob/bb193a24d0a968027ad48c397899ae3688165756/service/collector.go#L183) so a new service can't be started. It seems like putting the telemetryInitializer shutdown in the service shutdown avoids having to know they both need to be shutdown together.

I can remove it and call it separately if you prefer. Let me know your thoughts.

cpheps:
@bogdandrutu Would you mind reviewing this now that it's out of draft since you took a look earlier?

bogdandrutu (commented):
Please add a test, that with same "collector" instance, you can start the collector, see the metrics, then "reload" config (which stops/starts service) and you can see again metrics.

**Update:** If simpler just one "telemetryInitializer" and start/stop two times, and check that metrics initialization works properly second time.

cpheps:
>If simpler just one "telemetryInitializer" and start/stop two times, and check that metrics initialization works properly second time.

@bogdandrutu I think the test I wrote for #6239 [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/service_test.go#L95-L140) covers this. I wasn't sure how to trigger a service start failure in one run and not in another to target this fix. I figured the tests in #6239 covered the specific issue of the telemetryInitializer colliding with a previous version.

Let me know if you think that test doesn't cover what you're looking for.

bogdandrutu:
>Let me know if you think that test doesn't cover what you're looking for.

Not really, since I want to start/check works/stop/start/check works/stop same instance, not different instance, also not one invalid.

cpheps:
>Not really, since I want to start/check works/stop/start/check works/stop same instance, not different instance, also not one invalid.

@bogdandrutu I just want to make sure I understand. Do you want to test a single instance starting, stopping, restarting? I didn't think restarting an instance was valid from your [comment](https://github.com/open-telemetry/opentelemetry-collector/issues/5084#issuecomment-1091938307) in issue #5084.

```
I think we are going to the direction to follow the language pattern which is after Shutdown you cannot call Run on the same instance, you have to create a new instance.
```

I can do the test I just want to make sure I'm not writing a test that runs the collector in a way that should not be supported.

bogdandrutu (changes requested):
> I think we are going to the direction to follow the language pattern which is after Shutdown you cannot call Run on the same instance, you have to create a new instance.

For sure I stand by my words, these were in the context of `Collector` type. The `Collector` has a capability, which is to reload the configuration, and this is implemented by shutting down the `Service` instance and creating a new one, everything as planned for the moment. The only thing is that the `telemetryInitializer` is shared between these Service instances and not re-created, because of these we either change the design of the `telemetryInitializer` to recreate it every time we re-create the Service, or we test that it is ok to start/stop/start/stop.

Let me know if this is not clear for you.

bogdandrutu:
> The test it seems like you're looking for should target the use case of passing the same telemetryInitializer to multiple services and starting/stopping them, as what happens during a configuration reload. This would verify a single telemetryInitializer is reusable in services.

Correct

bogdandrutu:
Before it was a hack, because `telemetryInitializer` was initialize only first time (see the once variable inside the telemetryInitializer) and was shutdown only once when the collector was shutdown :)) So that is why I am worried (I know) that your change does not work, so to fix this you need way more things :)

service/service_test.go
```diff
@@ -139,6 +139,56 @@ func TestServiceTelemetryCleanupOnError(t *testing.T) {
...
+		Config:    validCfg,
+		telemetry: telemetry,
+	})
+	require.NoError(t, err)
+
+	// Start the service
+	require.NoError(t, srvOne.Start(context.Background()))
+
+	// Shutdown the service
+	require.NoError(t, srvOne.Shutdown(context.Background()))
```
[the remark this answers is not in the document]
cpheps: I see why now the telemetryInitializer shutdown is outside of the service shutdown. So I think to fix this bug as we only want to shut down everything in the error case I move the telemetryInitializer Shutdown back out to the collector level. That was it keeps the same contract before as being reusable but also cleans up in error cases.

cpheps:
@bogdandrutu I updated the test and have this working now. I will admit the current solution is not the most elegant but it does fix the immediate bug. 

Over the past few PRs I've done for this it seems like there is some work to be done on the telemetryInitializer pattern in the case of a single process starting and stopping new instances of the collector. It works fine if there is only ever one instance of the collector in a process.

I would like your opinion on this but due to the fact that the bug this PR fixes can cause a process to stop executing correctly (but still be active), and that there is a release schedule for tomorrow, that it's might be worth merging this PR if there's not concerns with unintended behavior. I'd be happy to file an issue (if there isn't one already) to further discuss the telemetryInitializer lifecycle within a single process as an action item.

service/service_test.go
```diff
@@ -139,6 +141,74 @@ func TestServiceTelemetryCleanupOnError(t *testing.T) {
...
+		telemetry: telemetry,
+	})
+	require.NoError(t, err)
+
+	// For safety ensure everything is cleaned up
+	t.Cleanup(func() {
+		assert.NoError(t, telemetry.shutdown())
+		assert.NoError(t, srvOne.Shutdown(context.Background()))
+		assert.NoError(t, srvTwo.Shutdown(context.Background()))
+	})
```
bogdandrutu, line 197: 1. srvOne already closed
2. I would move the cleanup for telemetry immediately close to the telemetry initialization.
3. remove `srvTwo` shutdown from here.

service/collector.go
```diff
@@ -149,7 +149,9 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	}
 
 	if err = col.service.Start(ctx); err != nil {
-		return err
+		if shutdownErr := col.shutdownServiceAndTelemetry(ctx); shutdownErr != nil {
+			return multierr.Append(err, shutdownErr)
+		}
```
bogdandrutu, line 154:
```suggestion
		return multierr.Append(err, col.shutdownServiceAndTelemetry(ctx))
```

service/collector.go
```diff
@@ -215,17 +217,30 @@ func (col *Collector) shutdown(ctx context.Context) error {
 		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown config provider: %w", err))
 	}
 
+	if err := col.shutdownServiceAndTelemetry(ctx); err != nil {
+		errs = multierr.Append(errs, err)
+	}
```
bogdandrutu, line 222: when using `multierr.Append` unless you want to enhance the err and wrap it with more context (like using `fmt.Errorf`, you can simply do this:

```suggestion
	errs = multierr.Append(errs, col.shutdownServiceAndTelemetry(ctx))
```

bogdandrutu (approved):

bogdandrutu:
> Over the past few PRs I've done for this it seems like there is some work to be done on the telemetryInitializer pattern in the case of a single process starting and stopping new instances of the collector. It works fine if there is only ever one instance of the collector in a process.
> I would like your opinion on this but due to the fact that the bug this PR fixes can cause a process to stop executing correctly (but still be active), and that there is a release schedule for tomorrow, that it's might be worth merging this PR if there's not concerns with unintended behavior. I'd be happy to file an issue (if there isn't one already) to further discuss the telemetryInitializer lifecycle within a single process as an action item.

Let's open an issue to discuss how to fix this. Unfortunately there is a limitation coming from the telemetry solution we use (opencensus) which uses a global state, so cannot initiate multiple instances (and possible cannot initiate multiple times).

service/collector.go
cpheps: Thanks for the suggestion. I'm not using to using this package so forgot it's ok to pass in a nil error.

cpheps:
@bogdandrutu pushed up suggested changes. I'll write up the issue today and link it here.

cpheps:
Created #6407 to discuss possible solutions for the global state kept for the internal telemetry solution.

service/collector.go
```diff
@@ -219,17 +219,28 @@ func (col *Collector) shutdown(ctx context.Context) error {
...
+
+// shutdownServiceAndTelemetry bundles shutting down the service and telemetryInitializer.
+// Returned error will be in multierr form and wrapped.
+func (col *Collector) shutdownServiceAndTelemetry(ctx context.Context) error {
+	var errs error
+
+	// shutdown service
 	if err := col.service.Shutdown(ctx); err != nil {
-		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown service: %w", err))
+		errs = multierr.Append(err, fmt.Errorf("failed to shutdown service after error: %w", err))
```
bogdandrutu, line 236:
```suggestion
		errs = multierr.Append(errs, fmt.Errorf("failed to shutdown service after error: %w", err))
```

bogdandrutu (approved):

PR #6354, opened 2022-10-19
[chore] add note decided in the SIG meeting

bogdandrutu:


tigrannajaryan:
I know I suggested to add a comment in the SIG meeting, but now that I am looked at what there is I don't think this is necessary. I don't see anywhere in the code that we are planning to remove this feature. The existing comment in the code is about the feature gate which is fine, I don't see a problem with that.

I don't think we need to have comments in the code that explain that the particular feature is not going to be removed without justification and proper transition plan. That's implied for every feature.

The misunderstanding seems to have been triggered by _"is still supported (for at least few versions)"_ wording in the PR that added the new feature. Since @bogdandrutu now explained the intent to keep the old feature around and only remove when we agree and have a proper transition plan I don't see a problem anymore.

cc @Aneurysm9

bogdandrutu:
@tigrannajaryan please tell your opinion if the extra "explanation" was a necessary thing, or this was an unnecessary concerned raised in the SIG meeting since the initial PR has clear TODO (for the featuregate) and not a clear plan/comment that this feature will be removed, just mentioned the intent to do that in the next releases.

tigrannajaryan:
> @tigrannajaryan please tell your opinion if the extra "explanation" was a necessary thing, or this was an unnecessary concerned raised in the SIG meeting since the initial PR has clear TODO (for the featuregate) and not a clear plan/comment that this feature will be removed, just mentioned the intent to do that in the next releases.

I think there is no need for any additional explanation. Seem like there was a misunderstanding of what the intent is and now that misunderstanding is cleared.

PR #6358, opened 2022-10-19
[docs] Add a doc for scraping receivers

dmitryax:
This document adds some definitions, stability requirements and guidelines on how to update emitted metrics for scraping receivers.

### DIFF
--- a/docs/scraping-receivers.md
+++ b/docs/scraping-receivers.md
@@ -1,0 +1,150 @@
+# Scraping Metrics Receivers
+
+Scraping metrics receivers are receivers that pull data from external sources at regular intervals and translate it
+into [pdata](../pdata/README.md) which is sent further in the pipeline. The external source of metrics usually is a
+monitored system providing data about itself in some arbitrary format. There are two types of scraping metrics
+receivers:
+
+- **Generic scraping metrics receivers:** The set of metrics emitted by this type of receiver fully depends on the 
+  state of the external source and/or the user settings. Examples:
+  - [Prometheus Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver)
+  - [SQL Query Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/sqlqueryreceiver)
+
+- **Built-in scraping metrics receivers:** Receivers of this type emit a predefined set of metrics. However, the 
+  metrics themselves are configurable via user settings. Examples of scrapings metrics receivers:
+
+- [Redis Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver)
+- [Zookeeper Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/zookeeperreceiver)
+
+This document covers built-in scraping metrics receivers. It defines which metrics these receivers can emit, 
+defines stability guarantees and provides guidelines for metric updates.
+
+## Defining emitted metrics 
+
+Each built-in scraping metrics receiver has a `metadata.yaml` file that MUST define all the metrics emitted by the 
+receiver. The file is being used to generate an API for metrics recording, user settings to customize the emitted 
+metrics and user documentation. The file schema is defined in 
+https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/mdatagen/metric-metadata.yaml. 
+Defining a metric in `metadata.yaml` DOES NOT guarantee that the metric will always be produced by the receiver. In
+some cases it may be impossible to fetch particular metrics from a system in a particular state.
+
+There are two categories of the metrics emitted by scraping receivers:
+
+- **Default metrics**: emitted by default, but can be disabled in user settings.
+
+- **Optional metrics**: not emitted by default, but can be enabled in user settings.
+
+### How to identify if new metric should be default or optional?
+
+There is no strict rule to differentiate default metrics from optional. As a rule of thumb, default metrics SHOULD be 
+treated as metrics essential to determine healthiness of a particular monitored system. Optional metrics on the 
+other hand SHOULD be treated as a source of additional information about the monitored system. 
+
+Additionally, if any of the following conditions can be applied to a metric, it MUST be marked as optional:
+
+- **It is redundant with another metric.** For example system CPU usage can be emitted as two different metrics:
+  `system.cpu.time` (CPU time reported as cumulative sum in seconds) or `system.cpu.utilization` (fraction of CPU 
+  time spent in different states reported as gauge), one of them has to be marked as option.
+- **It creates a disproportionately high cardinality of resources and/or data points.**
+- **There is a notable expected performance impact.**
+- **The source must be configured in an unusual way.**
+- **It requires dedicated configuration in the receiver.**
+
+## Stability levels of scraping receivers
+
+All the requirements defined for components in [the Collector's README](../README.md#stability-levels) are
+applicable to the scraping receivers as well. In addition, the following rules applied specifically to scraping
+metrics receivers:
+
+### In development
+
+The receiver is not ready for use. All the metrics emitted by the receiver are not finalized and can change in any way.
+
+### Alpha
+
+The receiver is ready for limited non-critical workloads. The list of emitted default metrics SHOULD be 
+considered as complete, but any changes to the `metadata.yaml` still MAY be applied.
+
+### Beta
+
+The receiver is ready for non-critical production workloads. The list of emitted default metrics MUST be
+considered as complete. Breaking changes to the emitted metrics SHOULD be applied following [the deprecation 
+process](#changing-the-emitted-metrics).
+
+### Stable
+
+The receiver is ready for production workloads. Breaking changes to the emitted metrics SHOULD be avoided.
+Nevertheless, metrics that are emitted by default MUST be always kept up-to-date with the latest stable version of the 
+monitored system. Given that, occasional breaking changes in the emitted metrics are expected even in the stable 
+receivers. Any breaking change MUST be applied following [the deprecation process](#changing-the-emitted-metrics).
+
+## Changing the emitted metrics
+
+Some changes are not considered breaking and can be applied to metrics emitted by scraping receivers of any 
+stability level:
+
+- Adding a new optional metric.
+
+Most of other changes to the emitted metrics are considered breaking and MUST be handled according to the stability 
+level of the receiver. Each type of breaking change defines a set of steps that MUST (or SHOULD) be applied across 
+several releases for a Stable (or Beta) components. At least 3 versions SHOULD be kept between the steps to give 
+users time to prepare, e.g. if the first step is released in v0.62.0, the second step SHOULD be released not earlier
+than 0.65.0. Any warnings SHOULD include the version starting from which the next step will take effect. If a 
+breaking change is more complicated and many metrics are involved in the change, feature gates SHOULD be used instead.
+
+### Removing an optional metric
+
+Steps to remove an optional metric:
+
+1. Mark the metric as deprecated in `metadata.yaml` by adding "[DEPRECATED]" in its description. Show a warning that 
+   the metric will be removed if the `enabled` option is set explicitly to `true` in user settings. 
+2. Remove the metric.
+
+### Removing a default metric
+
+Steps to remove a default metric:
+
+1. Mark the metric as deprecated in `metadata.yaml` by adding "[DEPRECATED]" in its description. Show a warning that
+   the metric will be removed if the `enabled` option is not explicitly set to `false` in user settings.
+2. Make the metric optional. Show a warning that the metric will be removed if the `enabled` option is set to `true`
+   in user settings.
+3. Remove the metric.
+
+### Making a default metric optional
+
+Steps to turn a metric from default to optional:
+
+1. Add a warning that the metric will be turned into optional if `enabled` field is not set explicitly to any value in
+   user settings. Warning example: "WARNING: Metric `foo.bar` will be disabled by default in v0.65.0. If you want to
+   keep it, please enable it explicitly in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric optional.
+
+### Adding a new default metric or turning an existing optional metric into default
+
+Adding a new default metric is a breaking change for a scraping receiver because it introduces an unexpected output
+for users and additional load on metric backends. Steps to apply such a change:
+
+1. If the metric doesn't exist yet, add one as an optional metric. Add a warning that the metric will be turned into 
+   default if the `enabled` option is not set explicitly to any value in user settings. A warning example: "WARNING: 
+   Metric `foo.bar` will be enabled by default in v0.65.0. If you don't want the metric to be emitted, please
+   disable it in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric default.
+
+### Other changes
+
+Other breaking changes SHOULD follow similar strategies inspecting presence of `enabled` field in user settings. For
+example, if a metric has to be renamed for any reason, the guidelines for "Removing an optional metric" and "Adding a  
+new default metric" SHOULD be followed simultaneously.
+
+Breaking changes that cannot be done through enabling/disabling metrics (e.g. removing or adding an extra attribute)
+SHOULD be applied using a feature gate with the following steps:
+
+1. Add a feature gate that is disabled by default. Enabling the feature gate changes the metrics behavior in a desired 
+   way. For example, if several metrics emitted by host metrics receiver need to be updated to have an additional 
+   `direction` attribute, the following feature gate can be used:
+  `receiver.hostmetricsreceiver.emitMetricsWithDirectionAttribute`. Show user a warning if the feature gate is not 
+   enabled explicitly, for example: "[WARNING] Metrics `system.network.packets` and `system.network.errors` will be 
+   changed in v0.65.0 to emit an additional `direction` attribute, enable a feature gate
+   `receiver.hostmetricsreceiver.emitMetricsWithDirectionAttribute` to apply and test the upcoming changes earlier".
+2. Enable the feature gate by default and update the warning thrown if user disables the feature gate explicitly. 
+3. Remove the feature gate along with the old behavior.


dmitryax:
@open-telemetry/collector-contrib-approvers PTAL

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+- [Redis Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/redisreceiver)
+
+## Defining emitted metrics 
+
+Each scraping receiver has `metadata.yaml` file that MUST define all the metrics emitted by the receiver. The
+file is being used to generate an API for metrics recording, user settings to customize the emitted metrics and user
+documentation. The file schema is defined in 
+https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/mdatagen/metric-metadata.yaml. 
+Defining a metric in `metadata.yaml` DOES NOT guarantee that the metric will always be produced by the receiver. In
+some cases it may be impossible to fetch particular metrics from a system in a particular state. If it's know that
```
codeboten, line 17:
```suggestion
some cases it may be impossible to fetch particular metrics from a system in a particular state. If it's known that 
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+## Defining emitted metrics 
+
+Each scraping receiver has `metadata.yaml` file that MUST define all the metrics emitted by the receiver. The
+file is being used to generate an API for metrics recording, user settings to customize the emitted metrics and user
+documentation. The file schema is defined in 
+https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/mdatagen/metric-metadata.yaml. 
+Defining a metric in `metadata.yaml` DOES NOT guarantee that the metric will always be produced by the receiver. In
+some cases it may be impossible to fetch particular metrics from a system in a particular state. If it's know that 
+the metric is reported only under certain conditions (e.g. a specific version) of the external source, the 
+condition MUST be mentioned in the metric description.
```
codeboten, line 19: Is there a format that's recommended for this? It would be great to be consistent

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+There are two main categories of the metrics emitted by scraping receivers:
```
codeboten, line 21: Are there additional categories? If not, i would remove the work `main` here

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+- **Default metrics**: emitted by default, but can be disabled in user settings.
```
codeboten, line 23: Any thoughts around naming these `Enabled` and `Disabled` metrics? The behaviour between these two is opposite, but the word default doesn't imply it's the opposite of optional

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+- **Optional metrics**: not emitted by default, but can be enabled in user settings.
+
+### How to identify if new metric should be default or optional?
+
+There is no strict rule to differentiate default metrics from optional. As a rule of thumb, default metrics SHOULD be 
+treated as metrics essential to determine healthiness of a particular monitored entity. Optional metrics from the
```
codeboten, line 30:
```suggestion
treated as metrics essential to determine healthiness of a particular monitored entity. Optional metrics on the 
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Alpha
+
+The receiver is ready for limited non-critical workloads. The list of emitted default metrics SHOULD be 
+considered as complete, but any changes to the `metadata.yaml` still MAY be applied.
+
+### Beta
+
+The receiver is ready for non-critical production workloads. The list of emitted default metrics MUST be
+considered as complete. Breaking changes to the emitted metrics SHOULD be applied following the deprecation
```
codeboten, line 60: Please link the "changing emitted metrics" section below here

codeboten (commented):
Thank you for documenting this, just a few comments/questions

docs/scraping-receivers.md
dmitryax: I'm not sure. For me, it feels like Default/Optional are more intuitive even if Default doesn't imply it's the opposite of Optional. But I'm good with Enabled/Disabled as well. The only concern is that Enabled is the same name the settings option which may be ok. 

Let's keep it for now, I can rename them if there is no other feedback from @open-telemetry/collector-contrib-approvers

docs/scraping-receivers.md
dmitryax: removed

docs/scraping-receivers.md
dmitryax: done

docs/scraping-receivers.md
dmitryax: I don't know what format we can prescribe here. Conditions can vary significantly. Do you want to add an example like "The metric is only emitted for Redis version >= 6"?

docs/scraping-receivers.md
djaglowski: Another suggestion: `DefaultOn` / `DefaultOff`. It's more verbose but it makes clear that we are describing the default state only.

`Enabled` is problematic because then when the user disables the metric you have a "disabled Enabled metric"?

`Optional` is also problematic because all metrics are optional.

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+The receiver is ready for limited non-critical workloads. The list of emitted default metrics SHOULD be 
+considered as complete, but any changes to the `metadata.yaml` still MAY be applied.
+
+### Beta
+
+The receiver is ready for non-critical production workloads. The list of emitted default metrics MUST be
+considered as complete. Breaking changes to the emitted metrics SHOULD be applied following [the deprecation 
+process](#changing-the-emitted-metrics). All the metric and resource attribute names SHOULD be released in an 
+[OpenTelemetry Specification](https://github.com/open-telemetry/opentelemetry-specification).
```
djaglowski, line 62: Whether or not these metric names should be tracked is not settled in the specification itself.

https://github.com/open-telemetry/opentelemetry-specification/issues/2610

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+the metric and resource attribute names MUST be released in an [OpenTelemetry 
+Specification](https://github.com/open-telemetry/opentelemetry-specification)
+
+## Changing the emitted metrics
+
+Some changes are not considered breaking and can be applied to metrics emitted by scraping receivers of any 
+stability level:
+
+- Adding a new optional metric.
+- Adding a new resource attribute to existing metrics.
```
djaglowski, line 77: Some backends would consider this breaking because metrics would become associated with a new resource.

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+Steps to replace a default metric with another default metric:
+
+1. Mark the metric as deprecated in `metadata.yaml` by adding "[DEPRECATED]" in its description. Show a warning 
+   that the metric will be replaced if the `enabled` option is not set explicitly to `false` in user settings. Add the
+   new metric as optional for now. Show another warning for the new metric if the `enabled` option is not set 
+   explicitly to any value in user settings.
+2. Turn the old metric into optional. Turn the new metric into default. Show the warning in case if user keeps the 
+   old metric enabled explicitly.
+3. Remove the old metric.
```
djaglowski, line 123: If a metric is being replaced with another, I think we should just refer to "Removing a default metric" and "Adding a default metric"?

That said, I think we should consider whether there is a reasonable process for making breaking changes to a metric (without changing its name). Often, a metric name is very intuitive and really seems like the "right" name for that metric. I think we're creating an undesirable situation if metric names are burned for good due to bugs in the data model that are resolved during beta.

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+1. Add a warning that the metric will be turned into optional if `enabled` field is not set explicitly to any value in
+   user settings. Warning example: "WARNING: Metric `foo.bar` will be disabled by default in v0.65.0. If you want to
+   keep it, please enable if explicitly in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric optional.
+
+### Adding a new default metric or turning an existing optional metric into default
+
+Adding a new default metric is a breaking for a scraping receiver because it introduces an unexpected output for users 
+and additional load on metric backends. Steps to apply such change:
```
djaglowski, line 137: I'm not sure this is really necessary or even reasonable.

With any other signal type, you would expect to get extra data that you do not care about, and either filter it or just ignore it.

Over time metrics will be added for many reasons and our scrapers will become less and less user friendly. 

To separate the concerns listed here:
- Are there backends that are actually "broken" by unexpected metrics?
- As long as additional load is roughly linear, is this really a problem?

djaglowski (commented):
Thanks for writing this up. We definitely need to pin these details down.

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
+# Scraping metrics receivers
+
+Scraping metrics receivers collect predefined set of metrics from externals sources. Typically, the external source
```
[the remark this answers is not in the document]
dmitryax: Makes sense. I'll add the some categorization of scraping receivers. "Predefined scraping metrics receiver" sounds a bit vague. Maybe "Built-in scraping metrics receiver"?

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
...
+
+- **Default metrics**: emitted by default, but can be disabled in user settings.
+
+- **Optional metrics**: not emitted by default, but can be enabled in user settings.
+
+### How to identify if new metric should be default or optional?
+
+There is no strict rule to differentiate default metrics from optional. As a rule of thumb, default metrics SHOULD be 
+treated as metrics essential to determine healthiness of a particular monitored entity. Optional metrics on the 
+other hand SHOULD be treated as a source of additional information about the monitored entity.
```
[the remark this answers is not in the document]
dmitryax: Thanks @djaglowski for providing the list. I agree that we should define it as a minimal requirement for marking a metric as default. But I don't agree that we shouldn't even try to consider "usefulness" of a metric before picking up the category. I'm afraid that if we go with this strict set of rules, a user that needs a rare metric which isn't not needed to anyone else will be encouraged to add it as default one, that will make other users to disable more and more metrics in their settings. It can be especially disruptive over time when receiver is marked beta or stable.

docs/scraping-receivers.md
dmitryax: Ok I'll remove this requirement for now

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
+# Scraping metrics receivers
+
+Scraping metrics receivers collect predefined set of metrics from externals sources. Typically, the external source 
+is a monitored entity providing data about itself in some arbitrary format, scraping receivers parse the data and 
+translate it to the OpenTelemetry Pipeline Data. Examples of the scrapings metrics receivers:
+
+- [Host Metrics Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver)
```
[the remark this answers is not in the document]
dmitryax: Makes sense, replaced with zookeeper receiver

docs/scraping-receivers.md
dmitryax: I'll remove this line for now. But potentially we can consider an ability to hide non-identifying resource attribute in user settings. In that case it'll become:
```
- Adding a new optional resource attribute.
```

docs/scraping-receivers.md
dmitryax: > Are there backends that are actually "broken" by unexpected metrics?
> As long as additional load is roughly linear, is this really a problem?

Not "broken" but can add an unexpected increase in billing at least. For example, in Splunk Infrastructure Monitoring, some metrics are categorized by use case and billed by number of identified entities, e.g. number of hosts or containers. Each category has a predefined set of metrics. Other metrics are going to a "custom" category which is billed by a number of time series. If we don't consider it as a breaking change and don't notify users about a new metric that will be enabled by default, they will get an unexpected increase in the custom time series affecting their billing.

dmitryax: > Over time metrics will be added for many reasons and our scrapers will become less and less user friendly.

Not sure I understand why scrapers will become less and less user friendly?

docs/scraping-receivers.md
```diff
@@ -0,0 +1,143 @@
+# Scraping metrics receivers
+
+Scraping metrics receivers collect predefined set of metrics from externals sources. Typically, the external source 
+is a monitored entity providing data about itself in some arbitrary format, scraping receivers parse the data and 
+translate it to the OpenTelemetry Pipeline Data. Examples of the scrapings metrics receivers:
```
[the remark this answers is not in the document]
dmitryax: Makes sense, updated

dmitryax:
> Tangential to this PR, we should also update https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/mdatagen/README.md to be more accurate. Maybe linking to this doc would be enough?

I believe https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/mdatagen/README.md should describe technical aspects of using the mdatagel tool. I'll update it once this PR is merged

docs/scraping-receivers.md
djaglowski: That would be nice. In general the lack of distinction between identifying and non-identifying resource attributes is problematic.

docs/scraping-receivers.md
djaglowski: > Not sure I understand why scrapers will become less and less user friendly?

Let's say we enable N metrics for a scraper at the time it is marked stable, over time this set of metrics will become outdated, often through events that are outside of our control:
- Source technologies will deprecate and remove metrics from their APIs
- Source technologies will add new metrics, some of which are critical
- We will realize that we missed important metrics
- We will occasionally make a mistake and realize that a metric we added is problematic (e.g. causes excessive load on the source)

Over time, the set of metrics that we launched with grows stale, and in order for users to find the scraper useful, they will have to make more and more modifications to the default set of metrics. 

I think it would be unfortunately if we are unable to make sensible adjustments to the scraper's output from time to time. In my opinion, metrics should each have their own lifecycle so that our scrapers are not frozen in the past forever. This model would reflect the reality of the situation much better. We should look for another way to define and control a static _set_ of metrics.

dmitryax: I completely agree that the set of default metrics should be always up to date and we shouldn't freeze them when we mark the component stable. That's why I mentioned that adding/removing metrics is possible in the stable state, but I don't think we should promote doing it silently. What if we relieve the requirement for such changes in the beta/stable states and mention all the acceptable reasons that you provided? Or you don't agree that we should call them "breaking" at all?

djaglowski: > That's why I mentioned that adding/removing metrics is possible in the stable state, but I don't think we should promote doing it silently. What if we relieve the requirement for such changes in the beta/stable states and mention all the acceptable reasons that you provided?

Ok, I agree with what you're suggesting here. We can keep the metrics up to date by _expecting_ breaking changes, while also communicating them in a way that is easily noticed.

docs/scraping-receivers.md
dmitryax: Great idea! Submitted an issue for this improvement for later https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/15995

dmitryax: I'll remove this requirement from the doc for now

docs/scraping-receivers.md
dmitryax: Updated the definition of stable to expect breaking changes in order to keep the metrics set up to date.

docs/scraping-receivers.md
dmitryax: Makes sense. Thanks for the feedback. I removed the "replacement" sections to not encourage unnecessary renames. Also added a section for other in-place changes using feature gates.

docs/scraping-receivers.md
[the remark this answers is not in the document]
dmitryax: @djaglowski I incorporated the guidelines that you suggested in the doc. PTAL

dmitryax:
@djaglowski @codeboten @evan-bradley thanks for your feedback. I believe everything is addressed now. Please take another look

djaglowski (approved):
LGTM. Thanks for considering my feedback.

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+into the [pdata](../pdata/README.md) which is sent further in the pipeline. The external source of metrics usually 
+is a monitored system providing data about itself in some arbitrary format. There are two types of scraping metrics 
+receivers:
+
+- **Generic scraping metrics receivers:** Set of metrics emitted by this type of receivers fully depend on a state of 
+  the external source and/or the user settings. Examples:
+  - [Prometheus Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver)
+  - [SQL Query Receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/sqlqueryreceiver)
+
+- **Built-in scraping metrics receivers:** Receivers of his type emit a predefined set of metrics. However, the
```
astencel-sumo, line 13:
```suggestion
- **Built-in scraping metrics receivers:** Receivers of this type emit a predefined set of metrics. However, the 
```

astencel-sumo (approved):
This is very educative read, thanks a lot Dmitrii.

My recent [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14927) is a good example of applying the "Changing the emitted metrics" section. One thought related to this is that I will need to add the warning messages after this doc is accepted.

docs/scraping-receivers.md
dmitryax: Good catch. Thanks!

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+  `system.cpu.time` (CPU time reported as cumulative sum in seconds) or `system.cpu.utilization` (fraction of CPU 
+  time spent in different states reported as gauge), one of them has to be marked as option.
+- **It creates a disproportionately high cardinality of resources and/or data points.**
+- **There is a notable expected performance impact.**
+- **The source must be configured in an unusual way.**
+- **It requires dedicated configuration in the receiver.**
+
+## Stability levels of scraping receivers
+
+All the requirements defined for components in [Collector README](../README.md#stability-levels) are applicable to
```
evan-bradley, line 55:
```suggestion
All the requirements defined for components in [the Collector's README](../README.md#stability-levels) are applicable to 
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+Some changes are not considered breaking and can be applied to metrics emitted by scraping receivers of any 
+stability level:
+
+- Adding a new optional metric.
+
+Most of other changes to the emitted metrics are considered breaking and MUST be handled according to the stability 
+level of the receiver. Each type of breaking change defines a set of steps that MUST (or SHOULD) be applied across 
+several releases for a Stable (or Beta) components. At least 3 versions SHOULD be kept between the steps to give 
+users time to prepare, e.g. if the first step is released in v0.62.0, the second step SHOULD be released not earlier 
+than 0.65.0. Any warnings SHOULD include the version starting from which the next step will take action. If
```
evan-bradley, line 91:
```suggestion
than 0.65.0. Any warnings SHOULD include the version starting from which the next step will take effect. If a
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+Steps to turn a metric from default to optional:
+
+1. Add a warning that the metric will be turned into optional if `enabled` field is not set explicitly to any value in
+   user settings. Warning example: "WARNING: Metric `foo.bar` will be disabled by default in v0.65.0. If you want to
+   keep it, please enable it explicitly in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric optional.
+
+### Adding a new default metric or turning an existing optional metric into default
+
+Adding a new default metric is a breaking for a scraping receiver because it introduces an unexpected output for users
```
evan-bradley, line 123:
```suggestion
Adding a new default metric is a breaking change for a scraping receiver because it introduces an unexpected output for users 
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. Add a warning that the metric will be turned into optional if `enabled` field is not set explicitly to any value in
+   user settings. Warning example: "WARNING: Metric `foo.bar` will be disabled by default in v0.65.0. If you want to
+   keep it, please enable it explicitly in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric optional.
+
+### Adding a new default metric or turning an existing optional metric into default
+
+Adding a new default metric is a breaking for a scraping receiver because it introduces an unexpected output for users 
+and additional load on metric backends. Steps to apply such change:
```
evan-bradley, line 124:
```suggestion
and additional load on metric backends. Steps to apply such a change:
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+1. If the metric doesn't exist yet, add one as an optional metric. Add a warning that the metric will be turned into 
+   default if the `enabled` option is not set explicitly to any value in user settings. A warning example: "WARNING: 
+   Metric `foo.bar` will be enabled by default in v0.65.0. If you don't want the metric to be emitted, please
+   disable it in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric default.
+
+### Other changes
+
+Other breaking changes SHOULD follow similar strategies inspecting presence of `enabled` field in user settings. For 
+example, if metric has to be renamed for any reason, guidelines for "Removing an optional metric" and "Adding a
```
evan-bradley, line 135:
```suggestion
example, if a metric has to be renamed for any reason, the guidelines for "Removing an optional metric" and "Adding a 
```

docs/scraping-receivers.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. If the metric doesn't exist yet, add one as an optional metric. Add a warning that the metric will be turned into 
+   default if the `enabled` option is not set explicitly to any value in user settings. A warning example: "WARNING: 
+   Metric `foo.bar` will be enabled by default in v0.65.0. If you don't want the metric to be emitted, please
+   disable it in the receiver settings."
+2. Remove the warning and update `metadata.yaml` to make the metric default.
+
+### Other changes
+
+Other breaking changes SHOULD follow similar strategies inspecting presence of `enabled` field in user settings. For
```
evan-bradley, line 134:
```suggestion
Other breaking changes SHOULD follow similar strategies that modify the `enabled` field in user settings. For 
```

evan-bradley (approved):
Thanks for writing this up.

I left a few wordsmithing suggestions, but overall it looks good to me.

docs/scraping-receivers.md
dmitryax: This correction changes the statement. We don't modify `enabled` field in the settings, we check if it's set only

dmitryax:
@evan-bradley thanks for the review! I applied all the corrections except for the last one

bogdandrutu (approved):

mx-psi (approved):

PR #6359, opened 2022-10-19
[docs] update design.md doc

codeboten:
Started looking at various documents under the `docs` directory that are somewhat outdatted. This PR updates the design document.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,25 +1,31 @@
 # OpenTelemetry Collector Architecture
 
-This document describes the architecture design and implementation of
+This document describes the architecture design and implementation of the
 OpenTelemetry Collector.
 
 ## Summary
 
-OpenTelemetry Collector is an executable that allows to receive telemetry data, optionally transform it and send the data further.
+OpenTelemetry Collector is an executable that can receive telemetry data, optionally process it, and export it further.
 
-The Collector supports several popular open-source protocols for telemetry data receiving and sending as well as offering a pluggable architecture for adding more protocols.
+The Collector supports several popular open-source protocols for receiving and sending telemetry data as well as offering a pluggable architecture for adding more protocols.
 
-Data receiving, transformation and sending is done using Pipelines. The Collector can be configured to have one or more Pipelines. Each Pipeline includes a set of Receivers that receive the data, a series of optional Processors that get the data from receivers and transform it and a set of Exporters which get the data from the Processors and send it further outside the Collector. The same receiver can feed data to multiple Pipelines and multiple pipelines can feed data into the same Exporter.
+Data receiving, processing, and exporting is done using [Pipelines](#pipelines). The Collector can be configured to have one or more pipelines. Each pipeline includes:
+
+- a set of [Receivers](#receivers) that receive the data
+- a series of optional [Processors](#processors) that get the data from receivers and process it
+- a set of [Exporters](#exporters) which get the data from processors and send it further outside the Collector.
+
+The same receiver can feed data to multiple pipelines and multiple pipelines can feed data into the same Exporter.
 
 ## Pipelines
 
-Pipeline defines a path the data follows in the Collector starting from reception, then further processing or modification and finally exiting the Collector via exporters.
+A pipeline defines a path the data follows in the Collector starting from reception, then further processing or modification and finally exiting the Collector via exporters.
 
-Pipelines can operate on 3 telemetry data types: traces, metrics and logs. The data type is a property of the pipeline defined by its configuration. Receivers, exporters and processors used in a pipeline must support the particular data type otherwise `ErrDataTypeIsNotSupported` will be reported when the configuration is loaded. A pipeline can be depicted the following way:
+Pipelines can operate on 3 telemetry data types: traces, metrics, and logs. The data type is a property of the pipeline defined by its configuration. Receivers, processors, and exporters used in a pipeline must support the particular data type otherwise `ErrDataTypeIsNotSupported` will be reported when the configuration is loaded. A pipeline can be depicted the following way:
 
 ![Pipelines](images/design-pipelines.png)
 
-There can be one or more receivers in a pipeline. Data from all receivers is pushed to the first processor, which performs a processing on it and then pushes it to the next processor (or it may drop the data, e.g. if it is a “sampling” processor) and so on until the last processor in the pipeline pushes the data to the exporters. Each exporter gets a copy of each data element. The last processor uses a `FanOutConnector` to fan out the data to multiple exporters.
+There can be one or more receivers in a pipeline. Data from all receivers is pushed to the first processor, which performs a processing on it and then pushes it to the next processor (or it may drop the data, e.g. if it is a “sampling” processor) and so on until the last processor in the pipeline pushes the data to the exporters. Each exporter gets a copy of each data element. The last processor uses a `fanoutconsumer` to fan out the data to multiple exporters.
 
 The pipeline is constructed during Collector startup based on pipeline definition in the config file.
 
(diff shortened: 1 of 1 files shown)


docs/design.md
```diff
@@ -1,25 +1,31 @@
...
 
-Data receiving, transformation and sending is done using Pipelines. The Collector can be configured to have one or more Pipelines. Each Pipeline includes a set of Receivers that receive the data, a series of optional Processors that get the data from receivers and transform it and a set of Exporters which get the data from the Processors and send it further outside the Collector. The same receiver can feed data to multiple Pipelines and multiple pipelines can feed data into the same Exporter.
+Data receiving, processing, and exporting is done using [Pipelines](#pipelines). The Collector can be configured to have one or more pipelines. Each pipeline includes:
+
+- a set of [Receivers](#receivers) that receive the data
+- a series of optional [Processors](#processors) that get the data from receivers and process it
+- a set of [Exporters](#exporters) which get the data from processors and send it further outside the Collector.
+
+The same receiver can feed data to multiple pipelines and multiple pipelines can feed data into the same Exporter.
```
Aneurysm9, line 18:
```suggestion
The same receiver can be included in multiple pipelines and multiple pipelines can include the same Exporter.
```

This read as if receivers and exporters were separate from pipelines, which I don't think is the intent.

docs/design.md
```diff
@@ -1,25 +1,31 @@
...
 
 ![Pipelines](images/design-pipelines.png)
 
-There can be one or more receivers in a pipeline. Data from all receivers is pushed to the first processor, which performs a processing on it and then pushes it to the next processor (or it may drop the data, e.g. if it is a “sampling” processor) and so on until the last processor in the pipeline pushes the data to the exporters. Each exporter gets a copy of each data element. The last processor uses a `FanOutConnector` to fan out the data to multiple exporters.
+There can be one or more receivers in a pipeline. Data from all receivers is pushed to the first processor, which performs a processing on it and then pushes it to the next processor (or it may drop the data, e.g. if it is a “sampling” processor) and so on until the last processor in the pipeline pushes the data to the exporters. Each exporter gets a copy of each data element. The last processor uses a `fanoutconsumer` to fan out the data to multiple exporters.
 
 The pipeline is constructed during Collector startup based on pipeline definition in the config file.
```
Aneurysm9, line 30:
```suggestion
The pipeline is constructed during Collector startup based on pipeline definition in the configuration.
```

Might be good to avoid referring to config "files" as there are now multiple sources of configuration.

docs/design.md
```diff
@@ -44,42 +50,44 @@ Receivers typically listen on a network port and receive telemetry data. Usually
...
       exporters: [jaeger]
     traces/2:  # another pipeline of “traces” type
-      receivers: [opencensus]
+      receivers: [otlp]
       processors: [batch]
       exporters: [opencensus]
 ```
 
-In the above example “opencensus” receiver will send the same data to pipeline “traces” and to pipeline “traces/2”. (Note: the configuration uses composite key names in the form of `type[/name]` as defined in [this document](https://docs.google.com/document/d/1NeheFG7DmcUYo_h2vLtNRlia9x5wOJMlV4QKEK05FhQ/edit#)).
+In the above example `otlp` receiver will send the same data to pipeline `traces` and to pipeline `traces/2`. (Note: the configuration uses composite key names in the form of `type[/name]` as defined in [this document](https://docs.google.com/document/d/1NeheFG7DmcUYo_h2vLtNRlia9x5wOJMlV4QKEK05FhQ/edit#)).
```
Aneurysm9, line 70: Can we get `this document` included in this repo and not on a gdoc of unknown provenance?

docs/design.md
```diff
@@ -44,42 +50,44 @@ Receivers typically listen on a network port and receive telemetry data. Usually
...
+Important: when the same receiver is referenced in more than one pipeline the Collector will create only one receiver instance at runtime that will send the data to a fan out consumer, which in turn will send the data to the first processor of each pipeline. The data propagation from receiver to the fan out consumer and then to processors is done via a synchronous function call. This means that if one processor blocks the call the other pipelines that are attached to this receiver will be blocked from receiving the same data and the receiver itself will stop processing and forwarding newly received data.
 
 ### Exporters
 
-Exporters typically forward the data they get to a destination on a network (but they can also send it elsewhere, e.g “logging” exporter writes the telemetry data to a local file).
+Exporters typically forward the data they get to a destination on a network (but they can also send it elsewhere, e.g `logging` exporter writes the telemetry data to a standard out).
```
Aneurysm9, line 81:
```suggestion
Exporters typically forward the data they get to a destination on a network (but they can also send it elsewhere, e.g `logging` exporter writes the telemetry data to the logging destination).
```

Not sure about the wording here, but IIRC the `logging` exporter will wrap the service logger, which may be configured to write somewhere other than `STDOUT`.

docs/design.md
```diff
@@ -112,9 +120,9 @@ In the above example “jaeger” exporter will get data from pipeline “traces
...
+The same name of the processor can be referenced in the `processors` key of multiple pipelines. In this case the same configuration will be used for each of these processors however each pipeline will always get its own instance of the processor. Each of these processors will have its own state, the processors are never shared between pipelines. For example if `batch` processor is used in several pipelines each pipeline will have its own batch processor (although the batch processor will be configured exactly the same way if the reference the same key in the config file). As an example, given the following config:
```
Aneurysm9, line 125:
```suggestion
The same name of the processor can be referenced in the `processors` key of multiple pipelines. In this case the same configuration will be used for each of these processors however each pipeline will always get its own instance of the processor. Each of these processors will have its own state, the processors are never shared between pipelines. For example if `batch` processor is used in several pipelines each pipeline will have its own batch processor (although each batch processor will be configured exactly the same way if they reference the same key in the configuration). As an example, given the following configuration:
```

docs/design.md
```diff
@@ -166,7 +174,7 @@ drawbacks, for example:
 To resolve the issues above, you can run OpenTelemetry Collector as an Agent.
 The Agent runs as a daemon in the VM/container and can be deployed independent
 of Library. Once Agent is deployed and running, it should be able to retrieve
-spans/stats/metrics from Library, export them to other backends. We MAY also
```
bogdandrutu, line 169: s/spans/traces?

docs/design.md
```diff
@@ -166,7 +174,7 @@ drawbacks, for example:
...
+spans/metrics/logs from Library, export them to other backends. We MAY also
```
codeboten, line 177:
```suggestion
traces/metrics/logs from Library, export them to other backends. We MAY also
```

docs/design.md
```diff
@@ -175,13 +183,13 @@ should also be able to send raw measurements and have Agent do the aggregation.
 ![agent-architecture](images/design-collector-agent.png)
 
 For developers/maintainers of other libraries: Agent can also
-accept spans/stats/metrics from other tracing/monitoring libraries, such as
+accept spans/metrics/logs from other tracing/monitoring libraries, such as
```
codeboten, line 186:
```suggestion
accept traces/metrics/logs from other tracing/monitoring libraries, such as
```

docs/design.md
Aneurysm9: I think "spans" is correct.  The data the collector operates on are individual spans.  The data model doesn't even have a "trace" entity, just a collection of "spans".

codeboten: Right, but the configuration for pipelines references `traces`, which aligns with the signal terminology for spans. I guess this is a bit of a grey area :(

docs/design.md
codeboten: See PR here https://github.com/open-telemetry/opentelemetry-collector/pull/6369

docs/design.md
bogdandrutu: The reason to use "traces" is because that's how we name the pipeline type.

bogdandrutu (approved):

PR #6361, opened 2022-10-20
Generate Export[Logs|Metrics|Traces]PartialSuccess, plugin into the Response

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/6364

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5816
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/5815

PR #6364, opened 2022-10-20
Change to non-nullable PartialSuccess

bogdandrutu:
To avoid the need of checking if the message is present, see https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/trace/v1/trace_service.proto#L58

Split from https://github.com/open-telemetry/opentelemetry-collector/pull/6361


bogdandrutu:
Internal change, no need for changelog I assume.

dmitryax (approved):

PR #6371, opened 2022-10-21
Add tests to the obsreport.Receiver component helper when using otel instrumentation

paivagustavo:
**Description:**  Adds tests for the OTel Go instrumentation in the `obsreport.Receiver` component helper.

This starts a testing framework for the `obsreport` package to assert metrics when the otel instrumentation featuregate is enabled, it can be used for testing other component helpers from `obsreport` and can be extended for testing gauges and histograms in the future. 

Tests are done by parsing and asserting exported metrics from a prometheus handler.

**Testing:**  Unit tests for the `obsreport.Receiver` component helper.

### DIFF
--- a/obsreport/obsreport_test.go
+++ b/obsreport/obsreport_test.go
@@ -25,6 +25,8 @@ import (
 	"go.opentelemetry.io/otel/codes"
 
 	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/featuregate"
+	"go.opentelemetry.io/collector/internal/obsreportconfig"
 	"go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 	"go.opentelemetry.io/collector/obsreport/obsreporttest"
 	"go.opentelemetry.io/collector/receiver/scrapererror"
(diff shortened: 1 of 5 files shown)


obsreport/obsreporttest/otelreporttest.go
```diff
@@ -0,0 +1,137 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.uber.org/multierr"
+
+	"go.opentelemetry.io/collector/config"
+)
+
+// PrometheusChecker is used to assert exported metrics from a prometheus handler.
+type PrometheusChecker struct {
+	promHandler http.Handler
+}
```
bogdandrutu, line 34: I would not want to expose public something tied to Prometheus.

paivagustavo: Have moved everything behind the existing checks and created a `obsreporttest.SetupOpenTelemetry()` which creates an instance of `TestTelemetry` that will use this `prometheusChecker` to validate the created metrics.

obsreport/obsreport_test.go
```diff
@@ -50,149 +50,174 @@ type testParams struct {
...
+			{items: 23, err: errFake},
+			{items: 29, err: nil},
+		}
+		for i, param := range params {
+			rec := NewReceiver(ReceiverSettings{
+				ReceiverID:             receiver,
+				Transport:              transport,
+				ReceiverCreateSettings: tt.ToReceiverCreateSettings(),
+			})
+			if useOtel {
```
jaronoff97, line 187: question: isn't this done already by [the code](https://github.com/open-telemetry/opentelemetry-collector/blob/a9f41a2fbcd713682bc693ce0fba4cd8c948bde7/obsreport/obsreport_receiver.go#L101)? Could we instead adjust the obsreportconfig being used like we do [here](https://github.com/open-telemetry/opentelemetry-collector/blob/bb193a24d0a968027ad48c397899ae3688165756/service/telemetry_test.go#L139)?

jaronoff97 (approved):
only one suggestion here, otherwise looks good – thanks so much!

obsreport/obsreport_test.go
paivagustavo: That is def better! I've created a new registry and passed it during the creation of the Receiver.

obsreport/obsreporttest/obsreporttest.go
```diff
@@ -102,6 +115,30 @@ func SetupTelemetry() (TestTelemetry, error) {
 	return settings, err
 }
 
+// SetupOpenTelemetry does setup the testing environment to check the metrics produced by the OpenTelemetry Go SDk,
+// and recorded by receivers, producers or exporters.
+// The caller should defer a call to Shutdown the returned TestTelemetry.
+func SetupOpenTelemetry() (TestTelemetry, error) {
```
bogdandrutu, line 121: Can you embed this into the `SetupTelemetry`?

obsreport/obsreporttest/obsreporttest.go
```diff
@@ -77,7 +85,12 @@ func (tts *TestTelemetry) ToReceiverCreateSettings() component.ReceiverCreateSet
 // Shutdown unregisters any views and shuts down the SpanRecorder
 func (tts *TestTelemetry) Shutdown(ctx context.Context) error {
 	view.Unregister(tts.views...)
-	return tts.SpanRecorder.Shutdown(ctx)
+	var errors []error
+	errors = append(errors, tts.SpanRecorder.Shutdown(ctx))
+	if tts.meterProvider != nil {
+		errors = append(errors, tts.meterProvider.Shutdown(ctx))
+	}
+	return multierr.Combine(errors...)
```
bogdandrutu, line 93:
```suggestion
	var errs error
	errs = multierr.Append(errs, tts.SpanRecorder.Shutdown(ctx))
	if tts.meterProvider != nil {
		errs = multierr.Append(errs, tts.meterProvider.Shutdown(ctx))
	}
	return errs
```

This way avoid unnecessary slice allocations when only one error. Not performance critical, but good to do the correct/optimal thing every time since people will copy the code :))

obsreport/obsreporttest/obsreporttest.go
paivagustavo: I'm not sure if we can, we need a way for the user of `TestTelemetry` to decide if it should use the OC views checks or OTel prometheus check.

If we embed it to `SetupTelemetry` we would need to add a `TestTelemetry.CheckWithOtel()` method for changing the underlying check, or something like that.

Do you have other suggestion for this?

bogdandrutu: Use the `featuregate` that we have for checking that?

paivagustavo: Sure, but we can't just pass a `featuregate.Registry` to `SetupTelemetry()` because that would be a breaking change, right? Or don't we care about this on the `obsreporttest` package?

Are we okay by just adding a `featuregate.Registry` as a field of the `TestTelemtry`, so we would be using like:
```go
tt, err := SetupTelemetry()
// handle err

// change the TestTelemetry registry
tt.Registry = registry

// use tt to check things based on its registry?
```

bogdandrutu: Can we simply change the entire package (for receiver) to check (for the moment) metrics via prometheus since opencensus also reports there? Then we revert that and check via a TestExporter?

> Are we okay by just adding a featuregate.Registry as a field of the TestTelemtry, so we would be using like:

I think we should better expose a boolean "OtelChecker" and will remove later.

paivagustavo: > Can we simply change the entire package (for receiver) to check (for the moment) metrics via prometheus since opencensus also reports there? 

I didn't want to touch the OC tests at first. But since we're ok with it, we can get rid of this separation oc x otel, and create a prometheus handler with both sdks to do assertions.

obsreport/obsreporttest/otelprometheuschecker.go
```diff
@@ -0,0 +1,138 @@
...
+func (pc *prometheusChecker) getMetric(expectedName string, expectedType io_prometheus_client.MetricType, expectedAttrs []attribute.KeyValue) (*io_prometheus_client.Metric, error) {
+	parsed, err := fetchPrometheusMetrics(pc.promHandler)
+	if err != nil {
+		return nil, err
+	}
+
+	metricFamily, ok := parsed[expectedName]
+	if !ok {
+		// OTel Go adds `_total` suffix for all monotonic sum.
+		// TODO: once remove support for OpenCensus, we should always check for metrics with the `_total` suffix for prometheus counter metrics.
```
paivagustavo, line 86: This feels like cheating but I couldn't find a better way, we can't disable this behavior even for tests.

bogdandrutu (approved):

obsreport/obsreporttest/otelprometheuschecker.go
bogdandrutu: Once we remove support for OC, we should use a "TestExporter" for otel metrics and not check prometheus :))

paivagustavo: Sure! I'll just remove this TODO, since it doesn't really makes sense :)

PR #6372, opened 2022-10-21
Connectors prototype v2

djaglowski:
2nd iteration of connectors prototype. (Follows #6140) 

This PR does the following:
- Formalizes the connector component in a way that does not lean on receiver and exporter implementations.
- Articulates the possible combinations of signal types, and allows each connector to implement any or all combinations. Also allows each connector to declare a stability level for each combination.
- Includes validation that a connector is actually used as a receiver in at least one pipeline and an exporter in at least one.
- Changes `service/internal/pipelines/Pipelines` to an interface with two implementations. The new implementation is available behind a feature gate called: `service.enableConnectors`
- Detects and rejects cycles in the graph.
- Builds, starts, and stops all components in the collector according to a topological ordering, determined by the graph based struct.
- Add a couple simple connectors to demonstrate capabilities. (`nopconnector`, `countconnector`)
- Tests are quite robust. There is new test framework for defining and validating pipelines that include connectors.

---

Here is a sample configuration that uses both the `nop` and `count` connectors to accomplish a couple simple tasks:

```yaml
receivers:
  # Read logs from two files
  filelog/blue:
    include: ./local/blue_green/in/blue.log
    start_at: beginning
  filelog/green:
    include: ./local/blue_green/in/green.log
    start_at: beginning

exporters:
  # Write each log to a corresponding file
  file/logs_blue:
    path: ./local/blue_green/out/logs_blue.jsonl
  file/logs_green:
    path: ./local/blue_green/out/logs_green.jsonl

  # Write a copy of all logs to a single file
  file/logs_all:
    path: ./local/blue_green/out/logs_all.jsonl

  # Count the number of logs from each file
  file/metrics_blue:
    path: ./local/blue_green/out/metrics_blue.jsonl
  file/metrics_green:
    path: ./local/blue_green/out/metrics_green.jsonl

  # Count the number of logs from both files
  file/metrics_all:
    path: ./local/blue_green/out/metrics_all.jsonl

connectors:
  count/blue:
  count/green:
  count/all:
  nop/all:

service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue, nop/all, count/blue, count/all]
    logs/green:
      receivers: [filelog/green]
      exporters: [file/logs_green, nop/all, count/green, count/all]
    logs/all:
      receivers: [nop/all]
      exporters: [file/logs_all]

    metrics/blue:
      receivers: [count/blue]
      exporters: [file/metrics_blue]
    metrics/green:
      receivers: [count/green]
      exporters: [file/metrics_green]
    metrics/all:
      receivers: [count/all]
      exporters: [file/metrics_all]
```

To run this, make sure to enable the feature gate:
```bash
./bin/otelcorecol_darwin_arm64 --config ./local/blue_green/config.yaml --feature-gates=service.enableConnectors
```


config/connector.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Connector is the configuration of a component.Connector. Specific extensions must implement
+// this interface and must embed ConnectorSettings struct or a struct that extends it.
+type Connector interface {
+	identifiable
+	validatable
+
+	// Implement both to ensure:
+	// 1. Only connectors are defined in 'connectors' section
+	// 2. Connectors may be placed in receiver and exporter positions in pipelines
```
bogdandrutu, line 28: Do we want 2.? It may confuse users.

connector/countconnector/count.go
```diff
@@ -0,0 +1,214 @@
...
+
+type Config struct {
+	config.ConnectorSettings `mapstructure:",squash"`
+	LogsConfig               `mapstructure:"logs"`
+}
+
+var _ config.Connector = (*Config)(nil)
+
+// NewFactory returns a ConnectorFactory.
+func NewFactory() component.ConnectorFactory {
```
bogdandrutu, line 67: Cannot find `component.ConnectorFactory`, did you miss it?

component/connector.go
```diff
@@ -0,0 +1,484 @@
...
+// A MetricsToMetricsConnector sends metrics from one pipeline to another.
+// Its purpose is to allow for differentiated processing of metrics.
+// MetricsToMetricsConnector feeds a consumer.Metrics with data.
+//
+// For example metrics could be collected in one pipeline and routed to another metrics pipeline
+// based on criteria such as attributes or other content of the metric. The second pipeline can
+// then process and export the metric to the appropriate backend.
+type MetricsToMetricsConnector interface {
+	Connector
+	ConsumeMetricsToMetrics(ctx context.Context, tm pmetric.Metrics) error
```
bogdandrutu, line 83: Why do we need new func (`ConsumeMetricsToMetrics`) and just implementing `consumer.Metrics` is not enough? Signature is equivalent.

connector/countconnector/count.go
djaglowski: It's in `component/connector.go`

component/connector.go
djaglowski: This was my original implementation but I found it to be confusing in some ways. I don't think there's anything that technically prevents us from doing it as you suggested so I can switch it back if you want. 

To try to summarize the confusion I found though:

Conceptually, any connector that consumes traces is a `consumer.Traces`, regardless of what it does with them. We also need each combination to implement a different interface, and as you pointed out `consumer.Traces` can differentiate `traces->traces` from the other combinations. However, the others ultimately must be wrapped in `consumer.Traces` as well in order to allow other components to send data to them. I found this confusing because it means that `traces->traces` is always `consumer.Traces`, but `traces->metrics` is only sometimes `consumer.Traces`, depending on where in the code. I found it easier to just treat all combinations the same so we can say they each have a dedicated interface, and so they all become `consumer.Traces` at the same time when building the graph. 

I'm not certain, but from a component author's point of view, the same confusion may be relevant.

config/connector.go
djaglowski: Hmm, this is fundamental to the design as I understood it.

The idea was that you _must_ use a connector as both a receiver and an exporter. For example:

```yaml
logs/in:
  receivers: [filelog]
  exporters: [countconnector]

metrics/out:
  receivers: [countconnector]
  exporters: [metricsexporter]
```

I think this can get confusing in very complex cases, but the basic idea is simple enough to start with. Are you thinking something else in particular, such as having dedicated connector sections in pipelines?

djaglowski (commented):
@bogdandrutu, thanks for taking a look.

I responded to your feedback and am continuing to work through tests.

config/connector.go
djaglowski: Ah, that makes sense. I had not considered this - will take a closer look and figure out how to disallow this.

djaglowski: I had the wrong idea about how the config enforced component usage. Fortunately I think this was an easy fix.

component/connector.go
bogdandrutu: Can you give me a bit of code snippet to make sure we understand. I am a bit confused about the explanation.

djaglowski: I'll try to explain better. It may be necessary to first step back and highlight a couple design questions that I worked through in this implementation.
- What constraints should be placed on the input->output types supported by a connector type?
- Based on the input/output constraints, how do users reason about configuration?
- Based on the input/output constraints, how should connectors be managed within the collector - instanced per-pipeline like processors, shared per-signal like receivers/exporters, or something else?

---

> What constraints should be placed on the input->output types supported by a connector type?

I considered the following options:
1. A connector type can support _exactly one input/output combination_. This is too granular and prevents intuitive use cases like `nop` and `count`, instead requiring `noptraces`, `nopmetrics`, etc.
2. A connector type can support _exactly one output type per input type_. This works for `nop` (`traces->traces, metrics->metrics, logs->logs`) and `count` (`traces->metrics, metrics->metrics, logs->logs`).
3. A connector type can support _any or all input/output combinations_. This is the least restrictive but introduces some implementation considerations. Notably though, this. is necessary to directly support behavior like `spanmetricsprocessor`, which _both_ derives metrics from spans _and_ allows the original spans to pass through unmodified. (i.e.`traces->traces, traces->metrics`).

I think option 3 is best because it allows for the most use cases / is the least restrictive.

---

> Based on the input/output constraints, how do users reason about configuration?

In short, a connector will do all the things it can do that are applicable to the pipelines it connects.
- A connector type may consume one or more signal types and emit one or more signal types.
- Each connector type declares support for specific combinations of signal types. 
- Each signal combination supported by a connector is an independent function. That is, the connector does not require that you use all if its capabilities. For example, if the `count` connector is configured to connect from a logs pipeline to a metrics pipeline, it will count logs. Likewise, if you do not want the `spanmetrics` connector to forward traces, you can configure only the `traces->metrics` connection and ignore the `traces->traces` capability.
- A connector will consume/emit for each pair of pipelines that it is connected.
- If a connector configured in such a way that any combination of its input/output pipelines are unsupported, this will result in a configuration error.


---

> Based on the input/output constraints, how should connectors be managed within the collector - instanced per-pipeline like processors, shared per-signal like receivers/exporters, or something else?

A connector is instanced based on each input/output type. This is similar to how receivers and exporters are managed internally. Specially, receivers and exporters are instanced per signal type. (Technically, the component developer can use `sharedcomponent` to unify these instances, but the collector manages them via a separate handle for each signal type.)

[This example](https://github.com/djaglowski/opentelemetry-collector/blob/connectors-prototype-v2/service/internal/pipelines/testdata/pipelines_conn_matrix.yaml) shows a connector that is consuming and emitting all three signal types. Internally, the collector will instantiate 9 instances of this connector, with each instance constructed by the corresponding `Create<DataType>To<DataType>Connector` factory method. Likewise, each instance of the connector will consume signals by performing the corresponding `Consume<DataType>To<DataType>` function.

In such a case, each of the 9 instances must ultimately satisfy one of the `consumer.<DataType>` interfaces. This is done by wrapping each instances via the appropriate `consumer.New<DataType>` function. [Here is the code that handles this](https://github.com/djaglowski/opentelemetry-collector/blob/244af4ce8c9ddc7fc435b7f25417298e0a4552b4/service/internal/pipelines/nodes.go#L299), but as a simplified version:

```go
switch receiverPipelineType {
case component.DataTypeTraces:
	...
	switch exporterPipelineType {
	case component.DataTypeTraces:
		connectorInstance := n.factory.CreateTracesToTracesConnector(ctx, set, cfg, fanoutConsumer)
		tracesConsumer := consumer.NewTraces(connectorInstance.ConsumeTracesToTraces)
		...
	case component.DataTypeMetrics:
		connectorInstance := n.factory.CreateMetricsToTracesConnector(ctx, set, cfg, fanoutConsumer)	
		metricsConsumer := consumer.NewMetrics(connectorInstance.ConsumeMetricsToTraces)
		...
	case component.DataTypeLogs:
		connectorInstance := n.factory.CreateLogsToTracesConnector(ctx, set, cfg, fanoutConsumer)
		logsConsumer := consumer.NewLogs(connectorInstance.ConsumeLogsToTraces)
		...
	}
```

We _could_ rename `ConsumeTracesToTraces` to `ConsumeTraces`, but then it is a special case vs `ConsumeMetricsToTraces` and `ConsumeLogsToTraces`. This is harder to keep track of in my opinion. It becomes especially awkward when trying to unwrap the connector instance, such as is necessary in tests when validating that each component in the graph is as expected. By treating each signal combination as equal, it allows for uniform code [such as this](https://github.com/djaglowski/opentelemetry-collector/blob/244af4ce8c9ddc7fc435b7f25417298e0a4552b4/service/internal/pipelines/graph_test.go#L108-L118).

djaglowski:
I think this is ready for a serious review. Due to the size of the PR, I will assume that it should be broken up into several parts, but I'd appreciate some more scrutiny of the overall implementation first.

connector/nopconnector/nop.go
```diff
@@ -0,0 +1,146 @@
...
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/internal/sharedcomponent"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+const (
+	typeStr = "nop"
```
tigrannajaryan, line 30: I think the name is confusing. I thought "nop" is something like /dev/null, a blackhome, while in reality it is a perfect passthrough.

component/connector.go
tigrannajaryan: @djaglowski this requires every pairwise combination of signals to have an interface. Is this approach going to scale if we add one or two other new signals? It is 3x3 now, if we add 2 more signals (e.g. profiles and entities), we will have 5x5=25 interfaces, 25 methods and 25 switch cases, etc.
Is there a way to use generics here somehow to reduce the code volume? (Probably no, I don't remember Go's genics having much compile-time power to generate combinations of type parameters).

connector/nopconnector/nop.go
djaglowski: As I understand it, the meaning of [nop](https://en.wikipedia.org/wiki/NOP_(code)#Machine_language_instructions) is basically "passthrough". Regardless, I have reservations about the name for another reason. 

Basically, I think users will be less interested in the passthrough behavior than the fact that the connector can be used to fork and merge pipelines. Technically, the fork/merge behavior comes from the connector framework itself, but ultimately this connector will be the one used when forking/merging should be done without altering the data.

I think maybe `hub` is a better name, as it conveys something more directly related to the use cases it support. WDYT?

component/connector.go
djaglowski: 25 combinations seems... manageable, if far from ideal. The scalability issues is _somewhat_ mitigated in that we likely will not exceed 5 signal types any time soon. I agree though - the current implementation will become quite cumbersome when adding additional signals.

I'll spend some time looking into this. There might be a way to decouple the input/output types so that a connector has to implement one consumer type and one "emitter" type, and then rely on composition rather than pairwise combinations.

TylerHelmuth (commented):
Sharing config thoughts based on the discussion during SIG

```yaml
service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue, nop/all, count/blue, count/all]
    logs/green:
      receivers: [filelog/green]
      exporters: [file/logs_green, nop/all, count/green, count/all]
    logs/all:
      receivers: [nop/all]
      exporters: [file/logs_all]

    metrics/blue:
      receivers: [count/blue]
      exporters: [file/metrics_blue]
    metrics/green:
      receivers: [count/green]
      exporters: [file/metrics_green]
    metrics/all:
      receivers: [count/all]
      exporters: [file/metrics_all]
```

I can definitely understand how the green logs get exported to a new file, the "all logs" file, the green count file, and the "all counts" file.  

I like how the basic assumptions for the pipelines aren't broken; when I look at the receivers for a metrics pipeline I can still assume that all the receivers will produce metrics and when I look at the exporters for the logs pipeline I can still assume that all the exporters export logs.  

It is also easy to match up receiver/exporter pairs since the same identifier is used in both places.

One followup question: Do connectors support being used in n -> m number of exporters/receivers?  Can a single connected be used in many exporter pipelines and then also in many receiver pipelines?

connector/nopconnector/nop.go
TylerHelmuth: I would also prefer something more descriptive than `nop`.  `hub` could work.  I also like `link` or `join`.

ahayworth:
👋 Hi - I wanted to make a note or two on this PR, but I will preface it with this statement: I do not have great context on this change, and thus my opinions should be weighted accordingly. ❤️ 

A few things jump out to me at this point, and I wanted to give you my (partially-informed) feedback for consideration:
- I definitely like the bits that focus on connecting pipelines (removing the need for an arbitrary hop on `localhost` from an exporter to a receiver is very nice indeed).
- The idea of doing cycle detection and configuration verification is also quite wonderful.
- I am concerned that the connectors _also_ implementing aggregations (such as the `count` connector) and transformations (such as log-to-metrics) really muddy the waters here.

As the proposal currently stands, as an end-user of the collector I foresee a lot of confusion and duplication in our future... how do I choose between a connector and a processor? Will this result in duplicate, unmaintained code in the collector as the project transitions away from things like the `metricsgenerationprocessor`? If I don't like what the connectors offer, aren't I right back into processor-land anyways? How do I reason about a connector inside of a pipeline (the configuration doesn't make it clear what it's really doing)?

I believe that the pipeline connection capabilities are quite valuable, as well as the sanity-checking that this introduces. From my perspective, focusing on those first (taking great care to really nail down a good way to represent connectors within the configuration file in a way that makes intuitive sense) would likely be really beneficial to people. We can then take a step back and evaluate whether or not we need to burden connectors with additional _processing_ capabilities right now, or if those are better suited where they are today (processors).

Again: take it all with a huge grain of salt - I'm not super well versed in the proposals. But if it landed as-is today in the collector, as an end-user I would really not know what I'm supposed to be doing with my collector configuration at this point 😅 

(And please do not see this as denigrating the work - I think the effort is fantastic, and it is a very difficult problem space to work in. I applaud you for taking it on!)

djaglowski:
> Do connectors support being used in n -> m number of exporters/receivers? Can a single connected be used in many exporter pipelines and then also in many receiver pipelines?

Yes. Connectors take advantage of the same fanout points that already exist in collector pipelines. Namely, 1) a pipeline with multiple exporters will have a fanout point immediately before the exporters, and 2) a receiver used in multiple pipelines will have a fanout point immediately after the receiver.

Since these fanout points exist _before exporters_ and _after receivers_, data replication is correctly managed outside of the connector framework. Basically, to figure out where data is replicated, you can disregard the notion of connectors and just look at each pipeline individually.

In the following, there is a fanout before each of the exporters. Whether `foo` is an exporter or connector does not change the fanout behavior. Therefore, we would expect both `logs/1` and `logs/2` to stream a copy of their data to `foo`.
```yaml
    logs/1:
      receivers: [filelog/1]
      exporters: [foo, file/1]
    logs/2:
      receivers: [filelog/2]
      exporters: [foo, file/2]
```

Independently, we can determine that in the following scenario, `foo` is used as a receiver in two pipelines. Therefore we would expect that a fanout occurs after `foo`, such that each pipeline will get a copy of the data stream. Again, whether `foo` is a receiver or connector does not affect the fanout pattern. 

```yaml
    metrics/a:
      receivers: [foo]
      exporters: [file/a]
    metrics/b:
      receivers: [foo]
      exporters: [file/b]
```


Putting it all together, you can see that a log read by `filelog/1` will be copied and sent to both `file/1` and `foo`. Whatever metrics are consequently emitted by `foo` will be copied and sent to `file/a` and `file/b`. Similarly, `foo` will get a copy of any log read from `filelog/2`, and any resulting metrics emitted by `foo` will be copied and sent to both `file/a` and `file/b`
```yaml
    logs/1:
      receivers: [filelog/1]
      exporters: [foo, file/1]
    logs/2:
      receivers: [filelog/2]
      exporters: [foo, file/2]
    metrics/a:
      receivers: [foo]
      exporters: [file/a]
    metrics/b:
      receivers: [foo]
      exporters: [file/b]
```

djaglowski:
> how do I choose between a connector and a processor

This is a good question, but I believe there is a clear answer. Basically, processors should never behave in a manner that requires awareness of other pipelines. If the functionality requires interaction with another pipeline, then it should be implemented as a connector. 

The caveat to the above is that we already have a few processors that violate this guideline. Namely, `routingprocessor` and `spanmetricsprocessor`. I believe these should be deprecated and replaced by connector implementations.

Notably here, I think it will be a responsibility of the maintainers to ensure that functionality between processors and connectors is not duplicated. If we do this well, I do not think there will be ambiguity for users regarding which component to use.

djaglowski:
> How do I reason about a connector inside of a pipeline (the configuration doesn't make it clear what it's really doing)?

We should be clear that a connector is always a bridge between pipelines. It never exists _entirely_ inside one pipeline. It must be used as an exporter in one or more pipelines _and_ as a receiver in one or more pipelines.

The expected behavior of a connector should be unambiguous, given the _types_ of pipelines that it connects. For example, if a `count` connector is used as an _exporter in a logs pipeline_ and a _receiver in a metrics pipeline_, then I should be able to determine that this will count the logs it consumes and emit metrics describing those logs. Of course some understanding of the particular connector is necessary, but the same is true of all components. The point is, the behavior can be complex, but should never be ambiguous.

connector/nopconnector/nop.go
djaglowski: I don't care for `join` because it does convey the ability to fork pipelines. 

`link` is a bit better, but in my mind this conveys a 1:1 connection between pipelines. 

Another option might be `multiplex`, but this feels less accessible than `hub`.

ahayworth:
> The way that connectors are declared (i.e. the connectors: block) is exactly the same as the way all of the other component types are declared. I assume this part is intuitive.

Yes, I agree - this part isn't confusing.

> The way that connectors are used in pipelines is of course less intuitive

This is the confusing part, yes. Consider the following snippet from above:

```yaml
service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue, nop/all, count/blue, count/all]
```

In this instance, I must simply _know_ that `count/blue` is a connector (and not an exporter). Today, I can rely on the knowledge that the `exporters` array inside a pipeline only references actual exporters; in the future I must keep the entire state of the configuration in my head to know that some exporters are not really exporters, but instead a different thing that connects to a different pipeline entirely. Careful naming can help - and of course a connector _is_ an exporter and a receiver simultaneously - but the complexity goes up regardless. (I should be clear in stating that I don't necessarily have a better idea, but hopefully highlighting it inspires a better answer from someone smarter than I).

> Certainly connectors enable additional complexity in the flow and processing of data, but this is the goal. 

This isn't a bad goal, I think what I'm driving at is the need to avoid any unnecessarily complexity, so that it's ultimately understandable and usable by people who may not be intimately familiar with the collector. 😄

connector/nopconnector/nop.go
djaglowski: One more option - `forward`. I think this is intuitive for both fork and merge cases, and does not imply any alteration of the data.

```yaml
service:
  pipelines:
    logs/in:
      receivers: [filelog]
      exporters: [forward]
    logs/out1:
      receivers: [forward]
      exporters: [backend/1]
    logs/out2:
      receivers: [forward]
      exporters: [backend/2]
```

```yaml
service:
  pipelines:
    logs/in1:
      receivers: [filelog]
      exporters: [forward]
    logs/in2:
      receivers: [filelog]
      exporters: [forward]
    logs/out:
      receivers: [forward]
      exporters: [backend]
```

djaglowski:
I appreciate the feedback here @ahayworth. You seem to have a better understanding of the feature than you give yourself credit for. 

> I must simply _know_ that count/blue is a connector (and not an exporter)

I see the point you are making here, but in practice I am not sure this is really a problem. Two things in particular mitigate my concern here:
1. The lists of connectors, exporters, receivers are easy enough to refer back to if you don't recall immediately. In the example you cited, it's just a few lines above the pipelines section.
```yaml
connectors:
  count/blue:
  count/green:
  count/all:
  nop/all:

service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue, nop/all, count/blue, count/all]
```
2. The _type_ of connector, exporter, or receiver will almost always be a strong indication of whether it is a connector or exporter/receiver. I expect there will be little to no overlap between the set of connector types and the set of exporter/receiver types. For example, we aren't going to have an `otlp` connector, so if you see `otlp` or `otlp/foo` in the list of exporters, it's clearly an exporter and not a connector. Likewise, I think `count` is quite clearly not a terminal destination, and that the name itself implies an input/output (something to be counted / a count to be emitted).

All that said, if there were a clear way to improve this, we can consider it. I've considered the following but would oppose:
- Declare connectors separately in pipelines. We'd have to be explicit about whether the connector is used as a receiver or exporter, so this is two new sections (e.g. `connectors_as_receivers` and `connectors_as_exporters`). This seems unnecessarily verbose and forces the user to reason about new rules. (e.g. "There must be at least one component in `exporters` or `connectors_as_exporters". You can use both lists simultaneously, or just one or the other.")
- Force a naming convention upon connector types, such as starting with a specific character or word. (e.g. `_count/1`, `conn_count/1`). This kind of thing is undesirable in my opinion because it introduces a new layer of complexity on our very simple and clean notion of a component ID (`type/[name]`).

tigrannajaryan:
@djaglowski do you think perhaps you could come up with an alternate way to render the connection graphs in yaml? I don't have good suggestions myself, but you have been thinking about this for a while, so if you had any other options it may be useful to show them to compare the different approaches.

djaglowski:
> @djaglowski do you think perhaps you could come up with an alternate way to render the connection graphs in yaml? I don't have good suggestions myself, but you have been thinking about this for a while, so if you had any other options it may be useful to show them to compare the different approaches.

Here are several alternative ideas. To be clear, I think all of these are inferior designs for various reasons stated below.

---

Roughly as was suggested in a prior comment, pipelines could have explicit sections for `connectors_as_receivers` and `connectors_as_exporters` (alternately, `->connector` and `connector->` but the special characters would probably cause unnecessary trouble in some templatized config system. ). The same rules apply to connector usage, but the config is explicit about what is expected to be a connector vs what is expected to be a receiver or exporter. The downside is that it introduces new rules such as:
 - A pipeline must have at least one `receiver` OR one `->connector`
 - A pipeline must have at least one `exporter` OR one `connector->`
 
```yaml
service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue]
      connectors->: [nop/all]
    logs/green:
      receivers: [filelog/green]
      exporters: [file/logs_green]
      connectors->: [nop/all]
    logs/all:
      ->connectors: [nop/all]
      exporters: [file/logs_all]
```

---

Connectors have one explicit section for connectors, with implication that the connector acts as an exporter from the pipeline in which it is named, and a receivers in the referenced pipeline(s). Pipelines do not need a receiver if a connector is pointed at them. I believe most users would find this more difficult to reason about because the way it connectors as a psuedo-component
```yaml
service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue]
      connectors: # map with keys referring to connectors above
        nop/all: [logs/all]
    logs/green:
      receivers: [filelog/green]
      exporters: [file/logs_green]
      connectors:
        nop/all: [logs/all]
    logs/all:
      exporters: [file/logs_all]
```

---

Instead of considering connectors as a "component" like receivers/processors/exporters, treat them as a special element that exists at the same level as pipelines. In terms of the graph, each pipeline is a node, and each connector is an edge between pipelines. Implicitly, connectors are inserted as an exporter in the 'from' pipeline and receiver in the 'to' pipeline.
```yaml
service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue]
    logs/green:
      receivers: [filelog/green]
      exporters: [file/logs_green]
    logs/all:
      receivers: [nop/all]
      exporters: [file/logs_all]
  connectors: # map with keys referring to connectors above
    nop/all:
      - from: logs/blue
        to: logs/all
      - from: logs/green
        to: logs/all
```

---

Entirely bypass the notion of 'pipelines' and instead just declare edges directly. Clarity of component responsibility is entirely based on references to declarative sections above. The implication here is that pipeline structure is entirely bypassed and replaced by direct node to node data passing. Many other designs are possible with this approach, but I expect we can all agree this is a nonstarter. I am highlighting it as what may be the "nearest" option that is still fully outside the concept of pipelines.
```yaml
service:
  edges:
    - from: filelog/blue
      to: file/logs_blue
    - from: filelog/green
      to: file/logs_green
    - from: filelog/all
      to: file/logs_all
```

ahayworth:
> You seem to have a better understanding of the feature than you give yourself credit for.

Thank you, that's kind. And I appreciate that you are willing to hear feedback on usability aspects of it, which I know can be tedious at times (but so important to get right).

> The lists of connectors, exporters, receivers are easy enough to refer back to if you don't recall immediately. In the example you cited, it's just a few lines above the pipelines section

I do so wish that collector configs I work with would fit on one (or even two) pages in my editor. 😂

But I do see the point you're making. I will concede that perhaps my usability worries are not representative of the what the average person may be working with!

> All that said, if there were a clear way to improve this, we can consider it.

Reading through the rest of your comment and the subsequent ones with the options considered: perhaps this is the best way after all. I appreciate that you walked us through all the options so thoroughly, it really helps me to see that this may not be ideal but is a lot better than many feasible alternatives.

I'll keep thinking about configuration usability in general, and if I come up with anything maybe we can look at that as a general improvement to make for the collector overall. :)

djaglowski:
I've pulled out the skeleton of the new component type for review here: https://github.com/open-telemetry/opentelemetry-collector/pull/6577

tigrannajaryan:
@djaglowski a slight modification of what you suggested:
```yaml
connectors:
  count/1: # converts telemetry to metrics by counting the number of records

service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      exporters: [file/logs_blue]
      connect_to: 
         - pipeline: logs/all # forward from logs/blue pipeline to logs/all pipeline the data as is.
    logs/green:
      receivers: [filelog/green]
      exporters: [file/logs_green]
      connect_to:
        - pipeline: metrics/all # forward from logs/green pipeline to metrics/all pipeline
          via: count/1 # use this connector logic. If omitted uses "forward" ("noop") connector.
    logs/all:
      exporters: [file/logs_all]
    metrics/all:
       ...
```
With this approach there is no need to refer to the same connector twice, we only refer to it in the source pipeline.

djaglowski:
> ```yaml
> connectors:
>   count/1: # converts telemetry to metrics by counting the number of records
> 
> service:
>   pipelines:
>     logs/blue:
>       receivers: [filelog/blue]
>       exporters: [file/logs_blue]
>       connect_to: 
>          - pipeline: logs/all # forward from logs/blue pipeline to logs/all pipeline the data as is.
>     logs/green:
>       receivers: [filelog/green]
>       exporters: [file/logs_green]
>       connect_to:
>         - pipeline: metrics/all # forward from logs/green pipeline to metrics/all pipeline
>           via: count/1 # use this connector logic. If omitted uses "forward" ("noop") connector.
>     logs/all:
>       exporters: [file/logs_all]
>     metrics/all:
>        ...
> ```
> 
> With this approach there is no need to refer to the same connector twice, we only refer to it in the source pipeline.

Is it really a problem to refer to a connector twice? What is the goal in preventing this?

To me, this approach appears to be more verbose anyways, and more importantly I think it blurs some concepts about the pipeline/component model that users already understand.

At the core of our pipeline model is a very simple rule, that _data enters via receivers and exits via exporters_. This presentation conveys that we are sending data "from a pipeline, to a pipeline" and immediately raises questions about whether there is an exporter or receiver involved. (Of course there is, but shouldn't this be apparent to users?)

Arguably this presentation also challenges the notion that, while a pipeline is an arrangement of components, it is the components themselves that handle the data. If connectors are handling data, aren't they necessarily components? Isn't it then more intuitive to clearly represent connectors as components that are inside of pipelines?

I think we need to decide how we want users to understand connectors. Should they understand that a connector _is a component_ and therefore exists inside of pipelines, as they understand with receivers, processors, and exporters? Or, do we want them to adopt a mental model where connectors are a special concept with an ambiguous relationship to the component/pipeline model? I personally have a strong preference for the former, and I think that the original design does the best job of conveying this model to the user.

tigrannajaryan:
>Is it really a problem to refer to a connector twice? What is the goal in preventing this?

It seems slightly less verbose, particularly when using the "nop" connector. The downside is it is more "asymmetric", I somehow like the symmetry in your proposal.

> I personally have a strong preference for the former, and I think that the original design does the best job of conveying this model to the user.

I am OK with your proposal. I do not think what I suggested is much better.

component/factories.go
```diff
@@ -32,6 +32,9 @@ type Factories struct {
 
 	// Extensions maps extension type names in the config to the respective factory.
 	Extensions map[Type]ExtensionFactory
+
+	// Connectors maps connector type names in the config to the respective factory.
+	Connectors map[Type]ConnectorFactory
```
djaglowski, line 37: Since the factory was added to `connectors` package in #6611, this would ideally be changed:
```suggestion
	Connectors map[Type]connector.Factory
```

However, this causes a circular dependency.

@bogdandrutu, is the plan still to move `Factories` to the `service` package, similar to #6552?

bogdandrutu: Yes :) You can create a PR for that, would be great.

jpkrohling:
Just wanted to mention that this is high on my list to review :-)

jpkrohling:
While I love the idea of having a native way to convert signals, I still couldn't think of a use case for the pure connection of pipelines that today's features wouldn't cover. Concretely, I don't understand the advantage of this:

```
    logs/all:
      receivers: [nop/all]
      exporters: [file/logs_all]
```

over this:

```
    logs/all:
      receivers: [filelog/blue, filelog/green]
      exporters: [file/logs_all]
```

djaglowski:
> While I love the idea of having a native way to convert signals, I still couldn't think of a use case for the pure connection of pipelines that today's features wouldn't cover. Concretely, I don't understand the advantage of this:
> 
> ```
>     logs/all:
>       receivers: [nop/all]
>       exporters: [file/logs_all]
> ```
> 
> over this:
> 
> ```
>     logs/all:
>       receivers: [filelog/blue, filelog/green]
>       exporters: [file/logs_all]
> ```

Adding some processors can illustrate the usefulness of same-signal connectors. Below, suppose we need to pre-process the two files in different ways, and then subsequently we wish to batch all the resulting logs together. (I'll call the connector `forward` as discussed elsewhere, but it is the same as `nop`.)

```yaml
service:
  pipelines:
    logs/blue:
      receivers: [filelog/blue]
      processors: [transform/blue] # something particular to blue
      exporters: [backend/blue, forward]
    logs/green:
      receivers: [filelog/green]
      processors: [transform/green] # something particular to green
      exporters: [backend/green, forward]
    logs/all:
      receivers: [forward]
      processors: [batch, transform/common]
      exporters: [backend/all]
```

Another case based only on `forward` - you may wish to replicate a signal for use in two different ways.

```yaml
service:
  pipelines:
    logs:
      receivers: [filelog]
      processor: [transform]
      exporters: [forward]
    logs/all:
      receivers: [forward]
      exporters: [low_cost_backend]
    logs/filtered:
      receivers: [forward]
      processors: [filter]
      exporters: [expensive_backend]
```

--- 

Arguably, the `routing` processor would be a good use case for a same-signal connector. As a connector, we can do more than just immediately export the data. 

```yaml
service:
  pipelines:
    logs:
      receivers: [filelog]
      exporters: [routing] # routes to one pipeline or the other based on some condition
    logs/type1:
      receivers: [routing]
      processors: [transform/type1]
      exporters: [backend]
    logs/type2:
      receivers: [forward]
      processors: [transform/type2]
      exporters: [backend]
```

---

Another example is the `count` connector. Technically, emitting a count of metrics could be done within a processor, but then you've mixed the derived signal with the original data stream. The user may wish to emit to a different backend instead, or process the signal differently. Additionally, if we would have such a connector for counting logs and spans, it would be intuitive to also have it count metrics.

```yaml
service:
  pipelines:
    metrics/in:
      receivers: [otlp]
      exporters: [nop, count]
    metrics/out:
      receivers: [nop]
      processors: [transform] # something particular to the original signals
      exporters: [backend]
    metric/counts:
      receivers: [nop]
      exporters: [backend/counts_only]
```

component/componenttest/nop_connector.go
```diff
@@ -0,0 +1,130 @@
...
+// NewNopConnectorFactory returns a component.ConnectorFactory that constructs nop processors.
+func NewNopConnectorFactory() component.ConnectorFactory {
+	return component.NewConnectorFactory(
+		"nop",
+		func() component.Config {
+			return &nopConnectorConfig{
+				ConnectorSettings: config.NewConnectorSettings(component.NewID("nop")),
+			}
+		},
+		component.WithTracesToTracesConnector(createTracesToTracesConnector, component.StabilityLevelStable),
```
jpkrohling, line 50: It can't be stable yet, from what I remember. It should be Alpha.

component/connector.go
```diff
@@ -0,0 +1,502 @@
...
+
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+var (
+	errDataTypesNotSupported = "connection from %s to %s is not supported"
+	ErrTracesToTraces        = fmt.Errorf(errDataTypesNotSupported, DataTypeTraces, DataTypeTraces)
```
jpkrohling, line 29: It might be clearer later on during the review, but should those be part of the public API?

component/connector.go
jpkrohling: I'll play around with this a bit, but I believe it might be possible to use generics here to alleviate at least part of the combinatorial problem.

jpkrohling (commented):
I didn't finish the review yet but I wanted to release the comments before I leave for the day.

connector/countconnector/count.go
```diff
@@ -0,0 +1,205 @@
...
+	set component.ConnectorCreateSettings,
+	cfg component.Config,
+	nextConsumer consumer.Metrics,
+) (component.TracesToMetricsConnector, error) {
+	comp := connectors.GetOrAdd(cfg.ID(), func() component.Component {
+		return newCountConnector(cfg.(*Config), set)
+	})
+
+	conn := comp.Unwrap().(*countConnector)
+	conn.metricsConsumer = nextConsumer
```
jpkrohling, line 71: Perhaps I'm missing something, but if you have the same connector for different input signals and different output signals, won't the last output override the previous ones? Like:

logs -> connector/1 -> metrics/prometheus1
traces -> connector/1 -> metrics/prometheus2

In the current code, won't this end up with `logs` sending data to `prometheus2`?

connector/countconnector/count.go
```diff
@@ -0,0 +1,205 @@
...
+	return conn, nil
+}
+
+// This is the map of already created count connectors for particular configurations.
+// We maintain this map because the Factory is asked trace, metric, and log receivers
+// separately but they must not create separate objects. When the connector is shutdown
+// it should be removed from this map so the same configuration can be recreated successfully.
+var connectors = sharedcomponent.NewSharedComponents()
+
+// otlpReceiver is the type that exposes Trace and Metrics reception.
```
jpkrohling, line 113:
```suggestion
// countConnector counts the input data points, converting to metrics
```

connector/countconnector/go.mod
```diff
@@ -0,0 +1,55 @@
+module go.opentelemetry.io/collector/connector/countconnector
+
+go 1.19
```
jpkrohling, line 3: Before merging, can you ensure this is aligned with the main go.mod? I think we are still on 1.18

connector/countconnector/go.mod
```diff
@@ -0,0 +1,55 @@
...
+
+require (
+	github.com/stretchr/testify v1.8.1
+	go.opentelemetry.io/collector v0.65.0
```
jpkrohling, line 7: Same for these :-)

connector/nopconnector/README.md
```diff
@@ -0,0 +1,21 @@
+# Nop Connector
+
+The `nopconnector` can be used to chain pipelines of the same type together.
+For example, it can replicate a signal to multiple pipelines so that each pipeline
+can process the signal independently in varying ways. Alternately, it can be used to
+merge two pipelines together so they can be processed as one pipeline.
```
jpkrohling, line 6: A paragraph here with a couple of use-cases would be nice to have. I myself wouldn't know why I would need this.

service/config.go
```diff
@@ -124,9 +145,14 @@ func (cfg *Config) validateService() error {
 		// Validate pipeline receiver name references.
 		for _, ref := range pipeline.Receivers {
 			// Check that the name referenced in the pipeline's receivers exists in the top-level receivers.
-			if cfg.Receivers[ref] == nil {
-				return fmt.Errorf("pipeline %q references receiver %q which does not exist", pipelineID, ref)
+			if cfg.Receivers[ref] != nil {
+				continue
 			}
+			if cfg.Connectors[ref] != nil {
```
jpkrohling, line 151: Same here:

```suggestion
			if _, ok := cfg.Connectors[ref]; ok {
```

service/config.go
```diff
@@ -124,9 +145,14 @@ func (cfg *Config) validateService() error {
 		// Validate pipeline receiver name references.
 		for _, ref := range pipeline.Receivers {
 			// Check that the name referenced in the pipeline's receivers exists in the top-level receivers.
-			if cfg.Receivers[ref] == nil {
-				return fmt.Errorf("pipeline %q references receiver %q which does not exist", pipelineID, ref)
+			if cfg.Receivers[ref] != nil {
```
jpkrohling, line 148: Is there a performance gain if you use the optional second return value (bool) from the map lookup?

```suggestion
			if _, ok := cfg.Receivers[ref]; ok {
```

service/config.go
```diff
@@ -152,15 +178,33 @@ func (cfg *Config) validateService() error {
 		// Validate pipeline exporter name references.
 		for _, ref := range pipeline.Exporters {
 			// Check that the name referenced in the pipeline's Exporters exists in the top-level Exporters.
-			if cfg.Exporters[ref] == nil {
-				return fmt.Errorf("pipeline %q references exporter %q which does not exist", pipelineID, ref)
+			if cfg.Exporters[ref] != nil {
```
jpkrohling, line 181: And here?

```suggestion
			if _, ok := cfg.Exporters[ref]; ok {
```

service/config.go
```diff
@@ -152,15 +178,33 @@ func (cfg *Config) validateService() error {
...
+				continue
+			}
+			if cfg.Connectors[ref] != nil {
```
jpkrohling, line 184: Here as well:

```suggestion
			if _, ok := cfg.Connectors[ref]; ok {
```

component/connector.go
djaglowski: I'll un-export them.

connector/countconnector/count.go
djaglowski: In this case, each instantiation of the connector would emit to both `prometheus1` and `prometheus2` because the connector is acting as a receiver in a pipeline with two exporters. The `service` package is responsible for detecting this during build time and wrapping the exporters into a `fanoutconsumer` when appropriate.

service/config.go
djaglowski: Good point. I think your suggestion is more idiomatic either way.

cmd/otelcorecol/builder-config.yaml
```diff
@@ -17,11 +17,18 @@ extensions:
 processors:
   - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.65.0
   - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.65.0
+connectors:
+  - import: go.opentelemetry.io/collector/connector/countconnector
+    gomod: go.opentelemetry.io/collector/connector/countconnector v0.0.0
```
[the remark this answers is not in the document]
djaglowski: Done

connector/countconnector/count.go
djaglowski: Fixed

connector/countconnector/go.mod
djaglowski: Fixed

connector/countconnector/go.mod
djaglowski: Fixed

connector/nopconnector/README.md
djaglowski: Added

service/config.go
djaglowski: Fixed

service/config.go
djaglowski: Fixed

service/config.go
djaglowski: Fixed

service/config.go
djaglowski: Fixed

djaglowski:
Closing in favor of #6700

PR #6391, opened 2022-10-24
[changelog] Update changelog entry for adding partial success fields

dmitryax:


codeboten (approved):

codeboten:
🤔 hmmm... changelog failed because no new entry was added... but this change was only a change to a changelog

PR #6395, opened 2022-10-25
Move default unmarshaler to service to avoid using deprecate configs

bogdandrutu:
Remove already deprecates `config.Service`.


service/unmarshaler.go
```diff
@@ -0,0 +1,66 @@
...
+
+type configSettings struct {
+	Receivers  *configunmarshaler.Receivers  `mapstructure:"receivers"`
+	Processors *configunmarshaler.Processors `mapstructure:"processors"`
+	Exporters  *configunmarshaler.Exporters  `mapstructure:"exporters"`
+	Extensions *configunmarshaler.Extensions `mapstructure:"extensions"`
+	Service    ConfigService                 `mapstructure:"service"`
+}
+
+// unmarshal the config.Config from a confmap.Conf.
```
codeboten, line 35:
```suggestion
// unmarshal the configSettings from a confmap.Conf.
```

service/internal/configunmarshaler/common.go
```diff
@@ -0,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
```
codeboten, line 1: I'd expect these in a file named `error.go` instead of common

codeboten (approved):
Just a couple of non-blocking comments

service/internal/configunmarshaler/common.go
bogdandrutu: Done.

service/unmarshaler.go
bogdandrutu: Done.

Issue #4554, opened 2021-12-14
Global sampling service telemetry logging configuration options


rmfitzpatrick:
**Is your feature request related to a problem? Please describe.**
Some components use warn and info level statements for paths that can be hit frequently, like the batch processor when configured with a small sending queue: https://github.com/open-telemetry/opentelemetry-collector/blob/9d3a8a4608a7dbd9f787867226a78356ace9b5e4/processor/batchprocessor/batch_processor.go#L185. This can lead to collectors logging hundreds or thousands of messages a second, much of which is redundant information.

**Describe the solution you'd like**
The `ServiceTelemetryLogs` config should provide zap sampling `initial` and `thereafter` fields that are used when instantiating all component loggers, with defaults equivalent to disabling:

```
    sampling_initial: 1
    sampling_thereafter: 1
```

**Describe alternatives you've considered**
Given the nature of zap logger usage in components I'm not sure of an alternative sampling approach within the collector process, or another course other than setting an arbitrarily high log level and missing important statements, or making valid warning/info scenarios have a lower log level (to same effect or no change with low enough configured log level).

**Additional context**
Would potentially resolve https://github.com/open-telemetry/opentelemetry-collector/issues/1061

tigrannajaryan:
> On a quick pass I'm not seeing where the batch processor uses a sampling config

AFAIK, it doesn't. queued_retry does.

I think it is a good idea to have a general sampling for all logs (at least configurable).

Ideally I would prefer that components are given 2 loggers: one to use during startup which will not be sampled (or will have higher sampling thresholds) to make sure all critical messages during startup are printed, and another logger to be used after startup to make sure long-running processes don't generate huge volumes of logs.

PR #6404, opened 2022-10-25
Allow to configure sampling config for logs

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/4554

### DIFF
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -61,6 +61,9 @@ type LogsConfig struct {
 	// (default = false)
 	DisableStacktrace bool `mapstructure:"disable_stacktrace"`
 
+	// Sampling sets a sampling policy. A nil SamplingConfig disables sampling.
+	Sampling *LogsSamplingConfig `mapstructure:"sampling"`
+
 	// OutputPaths is a list of URLs or file paths to write logging output to.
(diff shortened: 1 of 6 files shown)


service/telemetry/config.go
```diff
@@ -91,6 +94,14 @@ type LogsConfig struct {
 	InitialFields map[string]interface{} `mapstructure:"initial_fields"`
 }
 
+// LogsSamplingConfig sets a sampling strategy for the logger. Sampling caps the
+// global CPU and I/O load that logging puts on your process while attempting
+// to preserve a representative subset of your logs.
+type LogsSamplingConfig struct {
```
codeboten, line 100: Is it worth adding a note in the troubleshooting docs as to how to enable sampling? https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/troubleshooting.md#logs

codeboten (approved):
Change looks good, please rebase. Added just one non-blocking question

service/telemetry/config.go
bogdandrutu: We have only minimal examples there, not sure if we should add all of them there, or keep it minimal only the level.

bogdandrutu: We can do it in a separate PR.

PR #6405, opened 2022-10-25
Revert "[config{grpc,http}] Add warning when using unspecified address (#6267)"

codeboten:
This reverts commit 396964d8cfac89cc126901f4b1590bea7b69817b.


codeboten:
Reverting this change as it causes a breaking change preventing users from entering `:6060` as their endpoint configuration. See https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/3323429895/jobs/5493782699#step:8:921 for failure

bogdandrutu (approved):

PR #6421, opened 2022-10-27
[unrevert][config{grpc,http}] Add warning when using unspecified address

### LINKED ISSUE #6406: Parsing hostname:port causes breaking change
The [original change](https://github.com/open-telemetry/opentelemetry-collector/pull/6267) was reverted to unblock the collector/contrib repos. I think the original change is still desirable, but I'd love to hear people's thoughts on what the right thing to do is. Folks that configured endpoint like `:6060` will no longer be able to do so after this change. Is this acceptable? Should we swallow the error and warn for the time being as to not create a breaking change?

_Originally posted by @codeboten in https://github.com/open-telemetry/opentelemetry-collector/issues/6405#issuecomment-1291037953_

mx-psi:
Ex. Fixing a bug - Describe the bug and how this fixes the issue.
Ex. Adding a feature - Explain what this achieves.>

Unrevert #6267 and make it so that we never error out.

Add support for `:` and `:<port>` address types

**Link to tracking Issue:** Fixes #6406


config/internal/warning.go
```diff
@@ -0,0 +1,64 @@
...
+	switch {
+	case endpoint == ":":
+		// : (aka 0.0.0.0:0)
+		return true
+	case strings.HasPrefix(endpoint, ":"):
+		// :<port> (aka 0.0.0.0:<port>)
+		_, err := strconv.ParseInt(endpoint[1:], 10, 64)
+		if err != nil { // Probably invalid, don't warn
+			return false
+		}
```
bogdandrutu, line 35: The switch in golang does not execute things in order. So you may hit any of these branches if `endpoint == ":"`. This does not look like very deterministic to me, so I would prefer you to fix this.

mx-psi: Are you sure? [The spec](https://go.dev/ref/spec#Switch_statements) says
> the case expressions,[...] are evaluated left-to-right and top-to-bottom; the first one that equals the switch expression triggers execution of the statements of the associated case; the other cases are skipped. If no case matches and there is a "default" case, its statements are executed.

and then
> A missing switch expression is equivalent to the boolean value true. 

A quick check on the playground also seems to confirm this https://go.dev/play/p/KBynu4xGhKD

Maybe you were thinking about the `select` statement?

bogdandrutu: Probably I did not have enough coffee :) You are correct..

I still believe that this code is easier to read:

```golang
func shouldWarn(endpoint string) bool {
	if endpoint == ":" {
		// : (aka 0.0.0.0:0)
		return true
	}

	if strings.HasPrefix(endpoint, ":"):
		// :<port> (aka 0.0.0.0:<port>)
		_, err := strconv.ParseInt(endpoint[1:], 10, 64)
		if err != nil { // Probably invalid, don't warn
			return false
		}
		return true
	}

	// <host>:<port>
	host, _, err := net.SplitHostPort(endpoint)
	if err != nil { // Probably invalid, don't warn.
		return false
	}

	ip := net.ParseIP(host)
	return ip != nil && ip.IsUnspecified()
}
```

mx-psi: Done, PTAL

bogdandrutu (approved):
Needs a rebase

codeboten (approved):

Aneurysm9 (approved):

codeboten (approved):

Issue #6187, opened 2022-09-29
Split `component` into its own module


jpkrohling:
From @bogdandrutu:

> One question is if we want to have this package OR move everything into `go.opentelemetry.io/collector/CMP` (where CMP is one of receiver/processor/exporter/extension).

codeboten:
is there an agreement on how to proceed with this split? I guess if we were to split all the type specific stuff out of component, we'd end up with some amount of repetition in the name:

```
import "go.opentelemetry.io/collector/exporter"

exporter.ExporterFactory
exporter.Exporter
exporter.TracesExporter
```
instead of what it is currently:

```
import "go.opentelemetry.io/collector/component"

component.ExporterFactory
component.Exporter
component.TracesExporter
```

It would be possible to remove some of that repetition, but I'm not sure what we would call the `TracesExporter` for example to make it less repetitive. I do like having the components packages separated per type to only include what's relevant in each one, and reduce the amount of code imported for each one.

@open-telemetry/collector-approvers any thoughts?

Aneurysm9:
We could eliminate a bunch of that repetition in the per-component case, I think:

```
import "go.opentelemetry.io/collector/exporter"

exporter.Factory
exporter.Exporter
exporter.Traces
```

Not sure that there's a better name for `exporter.Exporter`, but [`url.URL`](https://pkg.go.dev/net/url#URL) exists and the world hasn't come to an end. v0v

dmitryax:
I don't believe we can move things from `component` to `go.opentelemetry.io/collector/[receiver|processor|exporter|extension]` at least because of `component.Host`. I'm not 100% sure we need to extract those parts from `component`, but, if we want to, maybe we can have the following packages instead:
```
go.opentelemetry.io/collector/component/exporter
go.opentelemetry.io/collector/component/receiver
go.opentelemetry.io/collector/component/processor
go.opentelemetry.io/collector/component/extention
go.opentelemetry.io/collector/component/host
```
And it's not that simple, we will also need to move other things (like `component.Kind/Component`) since there are interdependencies with `Host`.

codeboten:
After further digging into this, the `Host` complication makes the split per component impractical. My plan is to keep `component` as a single package containing the different component types within it.

bogdandrutu:
I made a prototype, and I think it is simple to do the split, see https://github.com/open-telemetry/opentelemetry-collector/pull/6552

[...]

PR #5870, opened 2022-08-08
Deprecate config.[Receiver|Processor|Exporter|Extension].ID() and SetName()

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-collector/issues/4714

Signed-off-by: Bogdan <bogdandrutu@gmail.com>

PR #6422, opened 2022-10-27
Deprecate all types and funcs in config package

bogdandrutu:
The main reason is to remove the circular dependency between the config (including sub-packages) and component. Here is the current state:
* component depends on config
* config/sub-package[grpc, http, etc.] depends on config & component

Because of this "circular" dependency, we cannot split for example "config" into its own module, only if all the other config sub-packages are also split.

```
config.Type => component.Type
config.DataType => component.DataType
config.Receiver => component.ReceiverConfig
config.Processor => component.ProcessorConfig
config.Exporter => component.ExporterConfig
config.Extension => component.ExtensionConfig
```

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/6187


dmitryax:
This seems like a good solution to me. I was thinking about taking a similar approach.

config/moved_config.go
```diff
@@ -14,13 +14,95 @@
...
+type Exporter = component.ExporterConfig
+
+// Deprecated: [v0.64.0] use component.UnmarshalExporterConfig.
+var UnmarshalExporter = component.UnmarshalExporterConfig
+
+// Deprecated: [v0.64.0] use component.ExporterConfigSettings.
+type ExporterSettings = component.ExporterConfigSettings
+
+// Deprecated: [v0.64.0] use component.NewExporterConfigSettings.
+var NewExporterSettings = component.NewExporterConfigSettings
```
dmitryax, line 66: Do you think we can simplify the new names in any way? Maybe reconsidering where we use `settings` vs `config`, or maybe even drop one of them? I'm not sure if we have strict definition for the terms to bring them from `config` package as is.

codeboten (commented):
Still working through reviewing the changes, overall i think this makes sense. Will this change (once the deprecated funcs are removed) completely decouple `config` from `component`?

.chloggen/deprecateallconfig.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - config.Type => component.Type
+  - config.DataType => component.DataType
+  - config.Receiver => component.ReceiverConfig
+  - config.UnmarshalReceiver => component.UnmarshalReceiverConfig
+  - config.[New]ReceiverSettings => component.[New]ReceiverConfigSettings
```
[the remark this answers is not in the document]
bogdandrutu: The "setting" pattern is an old pattern we used to do "inheritance". So not enough since these "settings" are the struct that MUST be embedded in every other struct that implements config.

I am not trying to fix that pattern yet (will discuss later), so for the moment I kept that way since these "settings" belong to the "config".

config/moved_config.go
bogdandrutu: Clearly a bad choice in terms of settings vs config. See also https://github.com/open-telemetry/opentelemetry-collector/pull/6422#discussion_r1009680107.

Short answer: for the moment we need to keep the long name to show that these settings "belong" to this config. We can debate in the future to remove this pattern completely.

dmitryax: Ok, let's keep it consistent with existing pattern and reconsider later 👍

.chloggen/deprecateallconfig.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - config.Processor => component.ProcessorConfig
+  - config.UnmarshalProcessor => component.UnmarshalProcessorConfig
+  - config.[New]ProcessorSettings => component.[New]ProcessorConfigSettings
+  - config.Exporter => component.ExporterConfig
+  - config.UnmarshalExporter => component.UnmarshalExporterConfig
+  - config.[New]ExporterSettings => component.[New]ExporterConfigSettings
+  - config.Extension => component.ExtensionConfig
+  - config.UnmarshalExtension => component.UnmarshalExtensionConfig
+  - config.[New]ExtensionSettings => component.[New]ExtensionConfigSettings
```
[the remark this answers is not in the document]
bogdandrutu: Done.

config/moved_config.go
bogdandrutu: Done.

.chloggen/deprecateallconfig.yaml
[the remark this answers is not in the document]
codeboten: I understand this is an existing pattern here, and i think this change is in the right direction.

I do think that *if* we plan on getting rid of the `ConfigSettings` suffix, making this change in the same version would be preferable, otherwise all the components will have to change from `[Receiver|Exporter|Processor|Extension]Settings` to `[Receiver|Exporter|Processor|Extension]ConfigSettings` and then to whatever else is decided on.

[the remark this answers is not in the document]
bogdandrutu: I think to remove that, we need to remove more functionality from Config. See https://github.com/open-telemetry/opentelemetry-collector/pull/5870, otherwise we need too many duplicate code to implement that interface.

bogdandrutu:
@codeboten @dmitryax PTAL

.chloggen/deprecateallconfig.yaml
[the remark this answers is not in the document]
bogdandrutu: I want to get rid of the `*ConfigSettings` structs not suffix. I am lost about your comment a bit.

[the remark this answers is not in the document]
bogdandrutu: Are you suggesting to not move yet the "*Setting" structs, and better work on removing these? If that is your suggestion, I think it is reasonable, and I can give a try.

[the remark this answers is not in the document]
bogdandrutu: Reverted that change for the moment.

.chloggen/deprecateallconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - config.DataType => component.DataType
+  - config.Receiver => component.ReceiverConfig
+  - config.UnmarshalReceiver => component.UnmarshalReceiverConfig
+  - config.[New]ReceiverSettings => component.[New]ReceiverConfigSettings
+  - config.Processor => component.ProcessorConfig
+  - config.UnmarshalProcessor => component.UnmarshalProcessorConfig
+  - config.Exporter => component.ExporterConfig
+  - config.UnmarshalExporter => component.UnmarshalExporterConfig
+  - config.Extension => component.ExtensionConfig
+  - config.UnmarshalExtension => component.UnmarshalExtensionConfig
```
dmitryax, line 28: Please update this list according to the latest update

dmitryax (commented):
Overall LGTM

Please rebase

.chloggen/deprecateallconfig.yaml
bogdandrutu: Done.

bogdandrutu:
> Please rebase

Done.

component/exporter_test.go
```diff
@@ -12,66 +12,69 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package component
+// TODO: Move tests back to component package after config.*Settings are removed.
+
+package component_test
```
dmitryax, line 17: the package name seems incorrect

dmitryax: I mean now we have `component` and `component_test` defined in `component/`

bogdandrutu: It is, you can read https://go-review.googlesource.com/c/go/+/420415 and https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language/31443271#31443271

It is allowed, and we used this to avoid circular dependency [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/obsreport/obsreporttest/obsreporttest_test.go#L15) as well.

You can also see that I have a TODO to revert this after we remove the "*Settings".

dmitryax: Nice! Thanks for providing the details

dmitryax (approved):

codeboten (approved):

PR #6423, opened 2022-10-27
Instrument `batch` processor with Otel Go

paivagustavo:
**Description:** 
- Instrument `batch` processor with OpenTelemetry Go. 
- Also fixed a small typo on the bucket definition of the `batch_send_size_bytes` view of OpenCensus.

**Link to tracking Issue:** part of https://github.com/open-telemetry/opentelemetry-collector/issues/816

**Testing:** 
- Updated existing tests that validate metrics using the OpenCensus `view`, to test instrumentation made by both OpenCensus and OpenTelemetry SDK. 
- Locally build and run the collector with the `featuregate` enabled/disabled.

**Note:** There are some "whitespace" changes, so I suggest reviewers to hide those when reviewing, it was done to make existing tests to run one time for each SDK.


processor/batchprocessor/factory.go
```diff
@@ -56,7 +57,7 @@ func createTracesProcessor(
 	nextConsumer consumer.Traces,
 ) (component.TracesProcessor, error) {
 	level := set.MetricsLevel
-	return newBatchTracesProcessor(set, nextConsumer, cfg.(*Config), level)
+	return newBatchTracesProcessor(set, nextConsumer, cfg.(*Config), level, featuregate.GetRegistry())
```
codeboten, line 60: why pass in the registry here instead of getting the registry within the new func?

processor/batchprocessor/metrics_test.go
```diff
@@ -32,3 +52,201 @@ func TestBatchProcessorMetrics(t *testing.T) {
...
+
+type expectedMetrics struct {
+	// processor_batch_batch_send_size_count
+	// processor_batch_batch_send_size_bytes_count
+	sendCount float64
+	// processor_batch_batch_send_size_sum
+	sendSizeSum float64
+	// processor_batch_batch_send_size_bytes_sum
+	sendSizeBytesSum float64
+
```
codeboten, line 70: nit: inconsistent spacing

processor/batchprocessor/metrics.go
```diff
@@ -76,3 +99,154 @@ func MetricViews() []*view.View {
...
+	var err error
+	meter := mp.Meter(scopeName)
+
+	bpt.batchSizeTriggerSend, err = meter.SyncInt64().Counter(
+		obsreport.BuildProcessorCustomMetricName(typeStr, "batch_size_trigger_send"),
+		instrument.WithDescription("Number of times the batch was sent due to a size trigger"),
+		instrument.WithUnit(unit.Dimensionless),
+	)
+	if err != nil {
+		return err
```
codeboten, line 184: what conditions would cause an error here and are they worth testing? only asking since it looks like many error conditions are not being tested

processor/batchprocessor/metrics.go
```diff
@@ -65,7 +88,7 @@ func MetricViews() []*view.View {
 		Description: statBatchSendSizeBytes.Description(),
 		TagKeys:     processorTagKeys,
 		Aggregation: view.Distribution(10, 25, 50, 75, 100, 250, 500, 750, 1000, 2000, 3000, 4000, 5000, 6000, 7000, 8000, 9000, 10000, 20000, 30000, 50000,
-			100_000, 200_000, 300_000, 400_000, 500_000, 600_000, 700_000, 800_00, 900_000,
+			100_000, 200_000, 300_000, 400_000, 500_000, 600_000, 700_000, 800_000, 900_000,
```
codeboten, line 91: Good catch!

processor/batchprocessor/metrics.go
paivagustavo: From the otel-go docs:
```
// If the passed instrument would result in an incompatible Aggregator, an
// error is returned and that Aggregator is not inserted or returned.
```

I think for most cases this should not happen, and can only be a problem when using views that would generate incompatible aggregators.

processor/batchprocessor/factory.go
paivagustavo: Mainly because tests uses these `newBatch*Processor` functions, so in order to change the featuregate registry during tests I had to update it. 

No strong feeling about keep this or changing tests to use the code inside the `newBatchProcessor`, I think I prefer the current definition.

processor/batchprocessor/batch_processor_test.go
```diff
@@ -124,114 +124,100 @@ func TestBatchProcessorSpansDeliveredEnforceBatchSize(t *testing.T) {
...
-	sink := new(consumertest.TracesSink)
-	cfg := createDefaultConfig().(*Config)
-	sendBatchSize := 20
-	cfg.SendBatchSize = uint32(sendBatchSize)
-	cfg.Timeout = 500 * time.Millisecond
-	creationSet := componenttest.NewNopProcessorCreateSettings()
-	batcher, err := newBatchTracesProcessor(creationSet, sink, cfg, configtelemetry.LevelDetailed)
-	require.NoError(t, err)
-	require.NoError(t, batcher.Start(context.Background(), componenttest.NewNopHost()))
+	telemetryTest(t, func(t *testing.T, tel testTelemetry, registry *featuregate.Registry) {
```
jmacd, line 127: The changes in this file would probably be easier to ready if you introduced a new top-level function everywhere `telemetryTest` is used, to preserve indentation.

```
func testBatchProcessorSentBySize(t *testing.T, tel testTelemetry, registry *featuregate.Registry) {
   ... mostly same code not re-indented
}
```

processor/batchprocessor/factory.go
bogdandrutu: @codeboten this is a technique that we did in lots of other places. We pass the registry so that unit-tests can run in parallel and not have to deal with cleanups.

codeboten: sounds good.

codeboten (approved):
Thanks for addressing my comments @paivagustavo, please take a look at @jmacd's suggestion

processor/batchprocessor/batch_processor_test.go
paivagustavo: Thanks for the suggestion, hopefully it more readable now :)

jmacd (approved):

PR #6441, opened 2022-10-28
Provide better helpers for configgrpc, consistent with confighttp

bogdandrutu:
Deprecate old `ToDialOptions` and `ToServerOptions`.

### DIFF
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -178,8 +178,25 @@ func (gcs *GRPCClientSettings) isSchemeHTTPS() bool {
 	return strings.HasPrefix(gcs.Endpoint, "https://")
 }
 
-// ToDialOptions maps configgrpc.GRPCClientSettings to a slice of dial options for gRPC.
+// ToClientConn creates a client connection to the given target. By default, it's
+// a non-blocking dial (the function won't wait for connections to be
+// established, and connecting happens in the background). To make it a blocking
+// dial, use grpc.WithBlock() dial option.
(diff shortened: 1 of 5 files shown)


config/configgrpc/configgrpc.go
```diff
@@ -178,8 +178,25 @@ func (gcs *GRPCClientSettings) isSchemeHTTPS() bool {
 	return strings.HasPrefix(gcs.Endpoint, "https://")
 }
 
-// ToDialOptions maps configgrpc.GRPCClientSettings to a slice of dial options for gRPC.
+// ToConn creates a client connection to the given target. By default, it's
+// a non-blocking dial (the function won't wait for connections to be
+// established, and connecting happens in the background). To make it a blocking
+// dial, use grpc.WithBlock() dial option.
+func (gcs *GRPCClientSettings) ToConn(ctx context.Context, host component.Host, settings component.TelemetrySettings, extraOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
```
bogdandrutu, line 185: Should we name this "Dial" to better align with the grpc equivalent? Same ToListener -> Listen?

dmitryax (approved):

config/configgrpc/configgrpc.go
bogdandrutu: Redundant "Client" since the struct is "GRPCClientSettings"... that's why I kept only `ToConn`.

bogdandrutu: Done. Changed since I feel is better to reflect the name.

Issue #6443, opened 2022-10-29
Crash on 32-bit builds when using memory_ballast extension


perher:
**Describe the bug**
Starting the collector with a 32-bit build using the memory_ballast extension without any ballast size specified causes a crash during startup.

**Steps to reproduce**
Start collector
```
./bin/otelcorecol_linux_386 --config examples/local/otel-config.yaml
```

**What did you expect to see?**
Collector to start correctly

**What did you see instead?**
```
Error: failed to start extensions: strconv.Atoi: parsing "9223372036854771712": value out of range; failed to shutdown service after error: failed to shutdown pipelines: no existing monitoring routine is running; no existing monitoring routine is running
```

**What version did you use?**
latest main branch
linux_386 build

**What config did you use?**
```
extensions:
  memory_ballast: {}
  zpages:
    endpoint: 0.0.0.0:55679

receivers:
  otlp:
    protocols:
      grpc:
      http:

processors:
  batch:
  memory_limiter:
    # 75% of maximum memory up to 4G
    limit_mib: 1536
    # 25% of limit up to 2G
    spike_limit_mib: 512
    check_interval: 5s

exporters:
  logging:
    loglevel: debug

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [memory_limiter, batch]
      exporters: [logging]
    metrics:
      receivers: [otlp]
      processors: [memory_limiter, batch]
      exporters: [logging]

  extensions: [memory_ballast, zpages]
```

**Environment**
Ubuntu 22.04 + with CGroup V1 (explicitly enabled)

**Additional context**
Starting the memory_ballast extension with setting explicit size works fine
```
memory_ballast:
  size_mib: 512
```

PR #6444, opened 2022-10-29
Use int64 for cgroup v1 parsing

perher:
**Description:**
Use int64 when parsing memory limits for cgroup V1 code (V2 code already does this)

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector/issues/6443

**Testing:**
Tested on Linux 32-bit build on Ubuntu 22.04 with cgroup v1 enabled


dmitryax (commented):
Thanks for the contribution. Please add a changelog entry file

perher:
Added changelog entry

dmitryax (approved):

bogdandrutu:
/cc @dashpole 

I think this code was copied/imported from somewhere, want to make sure we don't diverge too much from there.

dashpole:
I think it comes from here: https://github.com/uber-go/automaxprocs/blob/master/internal/cgroups/cgroup.go#L73.  I'm not personally worried about diverging, but it may be good to try and upstream this fix, as I assume it would impact their users as well.

dashpole (approved):

Issue #4613, opened 2021-12-29
Configuration Validations must validate all fields before returning an error


MovieStoreGuy:
**Is your feature request related to a problem? Please describe.**
A lot of the configuration validate methods will return early stating a singular problem of the configuration, this can lead to a rather long feedback loop addressing issues within the configuration within complicated configurations.

**Describe the solution you'd like**
Each validate functions must validate all errors and only return an error after it has checked / validated all required fields.

**Describe alternatives you've considered**
I believe the alternative is leave it as it is or RFTM  ?? 

**Additional context**
As someone that struggles with dyslexia, it takes me a few times to make sure what I typed is what I wanted, and having to reread the configuration on each failed attempted is rather pain staking process.

jpkrohling:
Should this be made a tracker, listing all the places where this should be done? There aren't that many components in this repo, but contrib might be affected as well.

bogdandrutu:
@MovieStoreGuy too many errors may also be a problem and confusion, I am not sold on the "print" all errors is the right approach. I remember the old `gcc` which did that and was extremely hard to identify the root cause, since an error may cause a chain of errors.

I would say probably one of the first thing we should add a "--dry run" (flag) or "validate" (command) that can return immediately and be used to validate the config instead of having to run the entire binary.

MovieStoreGuy:
My original thought is that to use `multierr` and expand each error line by line and I can visually understand the amount of things I had gotten wrong so I can see if I need to update one things, or a lot of things (assuming the config is valid yaml).

Something of the vein of:
```text
pipeline: no exporter defined `signalxf`
pipeline: no exporter defined `jager`
exporter: no exporter registered named `statd`
receiver: statsd: listen address is not set
extension: asap: no valid key provided
extension: asap: no keyserver provided
```

I mainly noticed this as I was cleaning up the contrib lifecycle tests trying to figure out what is the minimal required config to make it work.

> I would say probably one of the first thing we should add a "--dry run" (flag) or "validate" (command) that can return immediately and be used to validate the config instead of having to run the entire binary.

I am happy for this to be the first step to getting there, though I would want to make sure it check each entry within the configuration otherwise, it isn't much better than what we have now.

MovieStoreGuy:
> Should this be made a tracker, listing all the places where this should be done? There aren't that many components in this repo, but contrib might be affected as well.

[...]

PR #6322, opened 2022-10-15
Fixes #4671

Chinwendu20:
Signed-off-by: Maureen <amaka013@gmail.com>

**Description:** 
Enhancement: Added build flag to output components in collector distribution in YAML format

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4671

**Documentation:** I would open an issue on opentelemetry.io for the documentation, however, I updated the readme of the service package.

@jpkrohling please check this out

PR #6445, opened 2022-10-30
Created dry run flag for collector

Chinwendu20:
Signed-off-by: Maureen <amaka013@gmail.com>

**Description:** Created dry run flag for the collector

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4613

**Documentation:** Updated readme of service package but would create an issue on opentelemetry.io to update documentation


Chinwendu20:
Heyy, just discovered there were some recent changes to the main branch that would affect this PR. I would make some changes and push

service/config_provider.go
```diff
@@ -60,6 +60,8 @@ type ConfigProvider interface {
 	//
 	// Should never be called concurrently with itself or Get.
 	Shutdown(ctx context.Context) error
+	// DryRunGet validates the configuration, prints out all related errors without running the collector
+	DryRunGet(ctx context.Context, factories component.Factories) (*Config, error, bool)
```
bogdandrutu, line 64: Should "dry run" be a flag or a command? 

/cc @codeboten @jpkrohling who also reviewed https://github.com/open-telemetry/opentelemetry-collector/pull/6322

codeboten: i would expect it to be a flag

config/moved_config.go
```diff
@@ -153,6 +153,97 @@ func (cfg *Config) validateService() error {
 	}
 	return nil
 }
+func (cfg *Config) DryRunValidate() {
+	// Currently, there is no default receiver enabled.
+	// The configuration must specify at least one receiver to be valid.
+	if len(cfg.Receivers) == 0 {
```
jpkrohling, line 159: Each one of the conditions needs its test to ensure its correctness.

service/internal/configunmarshaler/dry_run_unmarshaller.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+var ExtensionValidateErrors []ExtensionValidateError
+var ExportersValidateErrors []ExporterValidateError
+var ReceiversValidateErrors []ReceiverValidateError
+var ProcessorsValidateErrors []ProcessorValidateError
+
+func (ConfigUnmarshaler) DryRunUnmarshal(v *confmap.Conf, factories component.Factories) (*config.Config, error, bool) {
+	errorcheck := false
+	if err := v.Unmarshal(&rawCfg, confmap.WithErrorUnused()); err != nil {
```
jpkrohling, line 43: Using a var that is defined in an unrelated file would be confusing. If you have to use the same var, it might make sense to split this into its own source file.

service/internal/configunmarshaler/dry_run_unmarshaller.go
```diff
@@ -0,0 +1,233 @@
...
+import (
+	"fmt"
+	"reflect"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/confmap"
+)
+
+type ExtensionValidateError struct {
```
jpkrohling, line 12: If those four types have the same definition, perhaps you can have only one? `ComponentValidationError` would probably be a good candidate here.

service/internal/configunmarshaler/dry_run_unmarshaller.go
```diff
@@ -0,0 +1,233 @@
...
+	Err       string
+}
+
+type ProcessorValidateError struct {
+	Component string
+	Id        config.ComponentID
+	Err       string
+}
+
+var ExtensionValidateErrors []ExtensionValidateError
```
jpkrohling, line 36: You can still have four vars, each being a slice of `ComponentValidationError`s

Chinwendu20:
Still have to make some changes as recent changes in the main branch has affected my most recent push as well. I will try to finish this one up as fast as possible.

Chinwendu20:
Had  a working solution earlier today just to find out that there are new changes to the the main branch later today hence this conflict.

Chinwendu20:
So the problem is that I am creating a new function for validate that just prints validation errors when it is encountered as opposed to returning quickly so if there is a change in the implementation of the main validate function I would have to change mine. I do not know if there is a smarter way of going about this

jpkrohling:
> So the problem is that I am creating a new function for validate that just prints validation errors when it is encountered as opposed to returning quickly

It's okay -- I prefer to have all errors listed instead of failing fast in this case.

> if there is a change in the implementation of the main validate function I would have to change mine

Not quite sure what you meant here.

Chinwendu20:
> > So the problem is that I am creating a new function for validate that just prints validation errors when it is encountered as opposed to returning quickly
> 
> It's okay -- I prefer to have all errors listed instead of failing fast in this case.
> 
> > if there is a change in the implementation of the main validate function I would have to change mine
> 
> Not quite sure what you meant here.

I am giving a reason why I have had to keep changing my code and why some checks are failing It is because at the time there were changes to the validate function after submitting my PR . I would go at it again

Issue #6167, opened 2022-09-28
Feature Gate to include issue and removal date


MovieStoreGuy:
**Is your feature request related to a problem? Please describe.**
Trying to read through feature flags and understand the context around them can be problematic at the best of times, let alone any user that just wants to set and forget.

**Describe the solution you'd like**
Feature gates to be extended to include `Issue`, and either `RemovalVersion` or `RemovalDate`.
This would allow operators the ability to quickly, and easily find the linked issue for the FeatureGate with additional context.

**Describe alternatives you've considered**
Currently, FeatureGates are documented in components readme however, it is rather easy to overlook these and challenging to find any further context.

MovieStoreGuy:
The feature has been added into the project, closing off the issue as done since it will be part of the next release.

PR #6447, opened 2022-10-31
[featuregates] Add fields for associated Issues and deprecated version

MovieStoreGuy:
**Description:** 
This PR extends the feature gate to add the fields `IssueLinks` and `RemovalVersion` to help provide context to users on:
- Why a feature gate exists?
- What is the associated issue(s)
- When will the feature gate will be removed

**Link to tracking Issue:** 

https://github.com/open-telemetry/opentelemetry-collector/issues/6167

**Testing:** 

Tests have been updated to include updated to Gate life cycle as per the README

**Documentation:** 

The featureflag README has been updated with an example.

### DIFF
--- a/featuregate/gates.go
+++ b/featuregate/gates.go
@@ -19,12 +19,106 @@ import (
 	"sync"
 )
 
-// Gate represents an individual feature that may be enabled or disabled based
+// Stage represents the Gate's lifecycle and what is the expected state of it.
+type Stage int8
+
+const (
+	// Alpha is used when creating a new feature and the Gate must be explicitly enabled
+	// by the operator.
+	//
+	// The Gate will be disabled by default.
+	Alpha Stage = iota
+	// Beta is used when the feature flag is well tested and is enabled by default,
+	// but can be disabled by a Gate.
+	//
+	// The Gate will be enabled by default.
+	Beta
+	// Stable is used when feature is permanently enabled and can not be disabled by a Gate.
+	// This value is used to provide feedback to the user that the gate will be removed in the next version.
+	//
+	// The Gate will be enabled by default and will return an error if modified.
+	Stable
+)
+
+func (s Stage) String() string {
+	switch s {
+	case Alpha:
+		return "Alpha"
+	case Beta:
+		return "Beta"
+	case Stable:
+		return "Stable"
+	}
+	return "unknown"
+}
+
+// Gate is an immutable object that is owned by the `Registry`
+// to represents an individual feature that may be enabled or disabled based
 // on the lifecycle state of the feature and CLI flags specified by the user.
 type Gate struct {
-	ID          string
+	// Deprecated: [v0.64.0] Use GetID() instead to read,
+	//				use `Registry.RegisterID` to set value.
+	ID string
+	// Deprecated: [v0.64.0] use GetDescription to read,
+	// 				use `WithRegisterDescription` to set using `Registry.RegisterID`.
 	Description string
-	Enabled     bool
+	// Deprecated: [v0.64.0] use `IsEnabled(id)` to read,
+	//              use `Registry.Apply` to set.
+	Enabled        bool
+	stage          Stage
+	referenceURL   string
+	removalVersion string
+}
+
+// RegistryOption allows for configuration additional information
+// about a gate that can be exposed throughout the application
+type RegistryOption func(g *Gate)
+
+// WithRegisterDescription adds the description to the provided `Gate“.
+func WithRegisterDescription(description string) RegistryOption {
+	return func(g *Gate) {
+		g.Description = description
+	}
+}
+
+// WithRegisterReferenceURL adds an URL that has
+// all the contextual information about the `Gate`.
+func WithRegisterReferenceURL(url string) RegistryOption {
+	return func(g *Gate) {
+		g.referenceURL = url
+	}
+}
+
+// WithRegisterRemovalVersion is used when the `Gate` is considered `Stable`,
+// to inform users that referencing the gate is no longer needed.
+func WithRegisterRemovalVersion(version string) RegistryOption {
+	return func(g *Gate) {
+		g.removalVersion = version
+	}
+}
+
+func (g *Gate) GetID() string {
+	return g.ID
+}
+
+func (g *Gate) IsEnabled() bool {
+	return g.Enabled
+}
+
+func (g *Gate) GetDescription() string {
+	return g.Description
+}
+
+func (g *Gate) GetStage() Stage {
+	return g.stage
+}
+
+func (g *Gate) GetReferenceURL() string {
+	return g.referenceURL
+}
+
+func (g *Gate) GetRemovalVersion() string {
+	return g.removalVersion
 }
 
 var reg = NewRegistry()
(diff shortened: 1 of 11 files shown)


MovieStoreGuy:
@bogdandrutu , 

I know it has been some time since we spoken about this, but does to align on how envisioned how featuregates would work?

MovieStoreGuy:
The patch is somewhat misleading here, it is not being directly tested, however, these new fields are being added to the templates which are being accessed.

featuregate/gates.go
```diff
@@ -19,14 +19,54 @@ import (
...
+	// Beta is used when the feature flag is well tested and is enabled by default,
+	// but can be disabled by a Gate.
+	//
+	// The Gate will be enabled by default.
+	Beta
+	// Stable is used when feature is permanently enabled and can not be disabled by a Gate.
+	// This value is used to provide feedback to the operator that the gate will be removed in the next version.
+	//
+	// The Gate will be enabled by default and will return an error if modified.
+	Stable
```
mx-psi, line 40: I am not sure I understand how the operator bits work. Could you elaborate? How does this work?

mx-psi (commented):
I think adding an issue link is a non-controversial change, but the new stage is a bit confusing.

featuregate/gates.go
MovieStoreGuy: I believe I made reference to wrong persona but the featuregate package guidelines on gate state.

Once a flag is marked stable, it can not be modified but is still until the next release so if someone is referencing the flag. The code should error on parsing the flags.

That is what I understood when reading the readme.

featuregate/gates.go
```diff
@@ -19,14 +19,54 @@ import (
...
 // Gate represents an individual feature that may be enabled or disabled based
 // on the lifecycle state of the feature and CLI flags specified by the user.
 type Gate struct {
 	ID          string
 	Description string
-	Enabled     bool
+	// Deprecated: [v0.64.0] Use Stage instead, this field will move to be non exported.
+	Enabled        bool
+	Stage          Stage
+	IssueLinks     []string
```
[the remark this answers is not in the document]
MovieStoreGuy: I think that makes sense, my thoughts were if a flag touches multiple issues then it would be nice to reference each of them but you could do the same on the flag issue as well.

I'm not fixed on the name, so I am happy to change it.

featuregate/gates.go
```diff
@@ -19,14 +19,54 @@ import (
...
+	RemovalVersion string
 }
 
+func (g Gate) IsEnabled() bool { return g.Enabled }
```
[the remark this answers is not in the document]
MovieStoreGuy: I'm marking the field deprecated since the flag stage defines if the gate starts as either enabled or disabled by default.

Setting enabled would be counter productive at that point.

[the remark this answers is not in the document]
mx-psi: What about implementing `IsEnabled()` by looking at the `Stage`?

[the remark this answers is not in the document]
MovieStoreGuy: Not exactly sure how that would work. 

I should say `Enabled` will still be use, it just wont be publicly accessible by the developer

bogdandrutu (commented):
I like the change, but this makes me more nervous about the current API to register a gate. Feels a bit reversed that we construct a Gate pass a copy to the registry and then query the enable status by the registry.IsEnabled.

@MovieStoreGuy @mx-psi I don't know if you realized but the `Enabled` on the Gate does not return if the gate is enable or not in the current process (represents only the developer configuration). To get the runtime "configuration" need to call `Registry.IsEnabled(id)`.

I find this more and more confusing, I wish we have something like:

```golang
type Registry struct {
}

func (r *Registry) MustRegister(id string, stage Stage, opts ..RegisterOptions) *Gate {
}

// Immutable object, owned by the `Registry`.
type Gate struct {
  ...
}

func (g *Gate) ID() string {
}

// Reflect the "runtime" status, so if somewhere else I have `reg.Apply(flags)` this will get changed since it is owned by the registry.
func (g *Gate) IsEnabled() bool {
}
```

MovieStoreGuy:
That makes sense to me, it make the interaction with the registry clear and it's intent.

featuregate/gates.go
```diff
@@ -19,12 +19,77 @@ import (
...
+	Alpha:  "Alpha",
+	Beta:   "Beta",
+	Stable: "Stable",
+}
+
+func (s Stage) String() string {
+	if v, ok := stageNames[s]; ok {
+		return v
+	}
+	return "unknown"
```
bogdandrutu, line 53: For few values a switch statement is easier to read and I believe faster.

featuregate/gates.go
```diff
@@ -19,12 +19,77 @@ import (
...
 type Gate struct {
-	ID          string
-	Description string
-	Enabled     bool
+	ID             string
+	Description    string
+	Enabled        bool
+	Stage          Stage
+	ReferenceURL   string
+	RemovalVersion string
```
bogdandrutu, line 65: Let's start to make it immutable, by providing getters only and deprecate the public fields.

MovieStoreGuy: Since the registry is returning a copy of the type, do you think it is valuable to do so?

I don't mind doing so just wanted to capture your thoughts here.

bogdandrutu: If it is returning a copy, then changes in "IsEnabled" are not reflected correct? Is that important?

MovieStoreGuy: > If it is returning a copy, then changes in "IsEnabled" are not reflected correct? 

No, so if make a call to `Registry.Apply` and update it after the gates have been read from `Registry.List` then you will have an out of date gate.

>  Is that important?

I would like to say no.
For example, if `zpages` lists each of the flags and then flags are then updated, reloading all of the gates to render that information resolves this.

However, it does raise the point of, should a gate have an exported method `IsEnabled` that could provide out of date information?

bogdandrutu (approved):

bogdandrutu:
@MovieStoreGuy please add another chloggen entry for the deprecated fields.

featuregate/gates.go
```diff
@@ -19,12 +19,106 @@ import (
...
+func (g *Gate) GetStage() Stage {
+	return g.stage
+}
+
+func (g *Gate) GetReferenceURL() string {
+	return g.referenceURL
+}
+
+func (g *Gate) GetRemovalVersion() string {
+	return g.removalVersion
```
mx-psi, line 121: nit: in Go (and in this repository in particular) we usually avoid the `Get` prefix for these methods

mx-psi (approved):
LGTM, only a nit

featuregate/gates.go
bogdandrutu: We should add this to our rules, but good point

mx-psi: #6463

MovieStoreGuy: How would you recommend I change this?

The reasoning for the prefix is that `Description` and `Description()` clash with names so it won't compile.

bogdandrutu: Let's keep the Get for the conflicts and next version we change again :)

mx-psi: Ok, marking as resolved then

Issue #6458, opened 2022-11-01
Replace `func obsreport.NewExporter() *Exporter` with `func obsreport.NewExporter() (*Exporter, error)`


bogdandrutu:
Similar change for all types (Receiver/Processor/etc.)

1. First PR deprecate the current func and add a new one `MustNewExporter`
2. After 1 release and all the usages changed in contrib, replace signature for obsreport.NewExporter and "undeprecate" it and deprecate Must version (if we don't want to keep it).
3. Remove Must version if we don't want to keep it.

PR #6464, opened 2022-11-02
deprecate `obsreport.New[Receiver|Scraper|Processor|Exporter]` in favor of `obsreport.MustNew*`

paivagustavo:
**Description:** As the `obsreport.[Receiver|Scraper|Processor|Exporter]` starts to use otel go, we need to return an error if one happens during the creation of otel go metrics. This is the first part for adding `error` as a return of these functions.

**Link to tracking Issue:** #6458

### DIFF
--- a/obsreport/obsreport_exporter.go
+++ b/obsreport/obsreport_exporter.go
@@ -64,8 +64,11 @@ type ExporterSettings struct {
 	ExporterCreateSettings component.ExporterCreateSettings
 }
 
-// NewExporter creates a new Exporter.
-func NewExporter(cfg ExporterSettings) *Exporter {
+// Deprecated: [v0.64.0] use MustNewExporter.
+var NewExporter = MustNewExporter
+
+// MustNewExporter creates a new Exporter.
+func MustNewExporter(cfg ExporterSettings) *Exporter {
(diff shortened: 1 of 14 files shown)


obsreport/obsreport_exporter.go
```diff
@@ -64,8 +64,11 @@ type ExporterSettings struct {
 	ExporterCreateSettings component.ExporterCreateSettings
 }
 
-// NewExporter creates a new Exporter.
-func NewExporter(cfg ExporterSettings) *Exporter {
+// Deprecated: [v0.64.0] use MustNewExporter.
+var NewExporter = MustNewExporter
+
+// MustNewExporter creates a new Exporter.
+func MustNewExporter(cfg ExporterSettings) *Exporter {
```
bogdandrutu, line 71: I would expect that `NewExporter` will log the error, and `MustNewExporter` will panic. Is that not acceptable? (asking for opinions).

paivagustavo: Yes, I think that's acceptable, will update this pr

bogdandrutu (approved):

Aneurysm9 (approved):

PR #6466, opened 2022-11-03
[chore] remove dependency on ballastextension from main core module

bogdandrutu:


processor/memorylimiterprocessor/memorylimiter.go
```diff
@@ -151,7 +150,7 @@ func getMemUsageChecker(cfg *Config, logger *zap.Logger) (*memUsageChecker, erro
 func (ml *memoryLimiter) start(_ context.Context, host component.Host) error {
 	extensions := host.GetExtensions()
 	for _, extension := range extensions {
-		if ext, ok := extension.(*ballastextension.MemoryBallast); ok {
+		if ext, ok := extension.(interface{ GetBallastSize() uint64 }); ok {
```
codeboten, line 153: should this interface be available somewhere?

bogdandrutu: Not sure, if we make it available we depend on that. And contrib is not using it, so I would keep it anonymous for the moment and if someone else comes and ask for it will find a solution.

codeboten (approved):

PR #6469, opened 2022-11-03
adding gosec scan for opentelemetry-collector repo

senanjude0:
adding gosec scan for opentelemetry-collector repo

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -1,0 +1,24 @@
+name: Static Analysis
+on:
+  pull_request:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


mx-psi (commented):
Hi, thanks for your PR. It looks like [we already use gosec](https://github.com/open-telemetry/opentelemetry-collector/blob/bfc45c16c979746c969de91d8d9dfd7037db1ca9/.golangci.yml#L128) via golangci-lint. Does this PR add additional checks? Is this something that we could do through golangci-lint?

bogdandrutu:
Closing, since we are already using gosec.

Issue #6474, opened 2022-11-03
[pdata] Add functions to clear optional fields


dmitryax:
There is no pdata API to clear optional fields like `pmetric.HistogramDataPoint.Min/Max`. We didn't add it before 1.0 because there was no need for that, but apparently it's needed now for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/15658.

Suggestion is to add methods with `Clear` prefix like:
- `pmetric.HistogramDataPoint.ClearMin()`
- `pmetric.HistogramDataPoint.ClearMax()`

PR #6475, opened 2022-11-03
[pdata] Add ability to clear optional fields

dmitryax:
This change adds ability to clear optional fields like pmetric.HistogramDataPoint.Min/Max/Sum. 

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/6474


TylerHelmuth (approved):
Thanks for taking care of this

pdata/internal/cmd/pdatagen/internal/base_fields.go
```diff
@@ -168,14 +168,31 @@ const accessorsOptionalPrimitiveValueTemplate = `// ${fieldName} returns the ${l
...
 func (ms ${structName}) Has${fieldName}() bool {
 	return ms.getOrig().${fieldName}_ != nil
 }
+
 // Set${fieldName} replaces the ${lowerFieldName} associated with this ${structName}.
 func (ms ${structName}) Set${fieldName}(v ${returnType}) {
 	ms.getOrig().${fieldName}_ = &${originStructType}{${fieldName}: v}
+}
+
+// Clear${fieldName} removes the ${lowerFieldName} associated with this ${structName}.
```
bogdandrutu, line 183: This is the only thing I want to make sure we agree on. Do we have some other similar examples?

dmitryax: We have `pcommon.Map.Clear()` which is kind of similar

bogdandrutu (approved):

pdata/internal/cmd/pdatagen/internal/base_fields.go
dmitryax: We can also call it `RemoveMin`, `RemoveMax` etc, given that we have `Map.Remove` and `Slice.RemoveIf`. Let me know if you think that's a better option

TylerHelmuth: What is the different between `Map.Remove` and `Map.Clear`?

dmitryax: `Map.Remove` removes a particular item under a provided key, `Map.Clear` removed all items from the map.

TylerHelmuth: Then maybe `Remove` is better.

bogdandrutu: "Remove" seems to be removing from a collection, clear seems to make everything disappear. Kind of "ClearMin" seems that everything related to "Min" disappears.

dmitryax: > "Remove" seems to be removing from a collection, clear seems to make everything disappear. Kind of "ClearMin" seems that everything related to "Min" disappears.

Yes, that is my thinking as well. In that case `HistogramDataPoint` can be a kind of a collection and Min is a field in that collection

dmitryax: Given that we have `Min`, `HasMin`, and `SetMin`, I'd rather prefer `RemoveMin` than `ClearMin`. @bogdandrutu WDYT?

bogdandrutu: Fine with "RemoveMin" please document in our name conventions :)

dmitryax: @bogdandrutu, updated to Remove and added documented it in the pdata/README.md

bogdandrutu (approved):

PR #6480, opened 2022-11-06
Add DecodeHookFunc opt for confmap

m1k1o:
Adding `mapstructure.DecodeHookFunc` as options to `conf.Unmarshal` and `conf.Marshal`.

For example to allow custom unmarshaling / marshaling of built-in types like `url.URL`.

```go
func MapstructureStringToUrlFunc() mapstructure.DecodeHookFunc {
	return func(
		f reflect.Type,
		t reflect.Type,
		data interface{}) (interface{}, error) {
		if f.Kind() != reflect.String {
			return data, nil
		}
		if t != reflect.TypeOf(url.URL{}) {
			return data, nil
		}

		return url.Parse(data.(string))
	}
}

err := conf.Unmarshal(&config, confmap.WithDecodeHookFunc(
  MapstructureStringToUrlFunc(),
))
```

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -77,9 +77,9 @@ func WithErrorUnused() UnmarshalOption {
 
 // WithDecodeHookFunc allows adding custom mapstucture.DecodeHookFunc to the decoder.
 // Can be called multiple times to add multiple hook functions.
-func WithDecodeHookFunc(fn mapstructure.DecodeHookFunc) UnmarshalOption {
+func WithDecodeHookFunc(fn ...mapstructure.DecodeHookFunc) UnmarshalOption {
 	return unmarshalOptionFunc(func(uo *unmarshalOption) {
-		uo.hookFuncs = append(uo.hookFuncs, fn)
+		uo.hookFuncs = append(uo.hookFuncs, fn...)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


confmap/confmap.go
```diff
@@ -74,6 +75,14 @@ func WithErrorUnused() UnmarshalOption {
 	})
 }
 
+// WithDecodeHookFunc allows adding custom mapstucture.DecodeHookFunc to the decoder.
+// Can be called multiple times to add multiple hook functions.
+func WithDecodeHookFunc(fn ...mapstructure.DecodeHookFunc) UnmarshalOption {
```
bogdandrutu, line 80: Few comments:
1. We should not expose the "mapstructure" dependency public.
2. Is your use-case needed now or it is just a theoretical one?

m1k1o: Its just [`interface{}`](https://github.com/mitchellh/mapstructure/blob/1f85f72c684702c840b9a77616dccbdba63f0b29/mapstructure.go#L185) and should allow everyone who is familiar with mapstructure to add known hooks. I can either create custom `type DecodeHookFunc interface{}` type, or if we settle for just one type of function i can add that, e.g. `func(from reflect.Value, to reflect.Value) (interface{}, error)`.

I do have usecase for this, to decode `url.URL`. Instead of just adding that to existing hook functions I tried to create more general solution that just adds new option. Although the usecase whithin this project might be more theoretical, I thought it might benefit from this addition as that is just adding new option and not creating other change.

mx-psi: > to decode url.URL

If you create a type alias and implement the [`encoding.TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler) interface you can decode the URL without having to modify confmap:

```go

type MyURL url.URL

func (u *MyURL) UnmarshalText(text []byte) err {
    // use UnmarshalBinary based on https://github.com/golang/go/blob/0b7aa9f/src/net/url/url.go#L1178
    return u.UnmarshalBinary(text)
}

type Config struct {
  // ... other fields here ...
  url *MyURL
}

```

This works because we pass a hook to mapstructure that recognizes this interface: https://github.com/open-telemetry/opentelemetry-collector/blob/0b08d531642685c500d69ffc525bc3d641b5f454/confmap/confmap.go#L169

Would that solve your problem?

m1k1o: That would mean all functions from `url.URL` must be "proxied" and for every function that expects `url.URL` it must be casted. With those constrains, yes it does solve this usecase.

PR #6485, opened 2022-11-07
Remove get from newly added funcs on Gate that don't conflict

bogdandrutu:
Because of the conflict between the public member and func name, cannot do the same for ID/Description, will fix in a later version after we remove the public members.

No need for deprecation, they were added in this version.

/cc @mx-psi based on the discussion from https://github.com/open-telemetry/opentelemetry-collector/pull/6447

### DIFF
--- a/featuregate/gates.go
+++ b/featuregate/gates.go
@@ -109,15 +109,15 @@ func (g *Gate) GetDescription() string {
 	return g.Description
 }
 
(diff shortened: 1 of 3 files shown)


MovieStoreGuy (approved):
Makes sense to me :)

mx-psi (approved):

PR #6488, opened 2022-11-07
Change component.DataType enum value names to be consistent with other in the repo

bogdandrutu:
No deprecation, they were added in this cycle. Deprecated old formats are in config package.


dmitryax (approved):
LGTM 👍

PR #6489, opened 2022-11-07
Remove unused member from confmap.Resolver

bogdandrutu:
This may be a breaking change, but should not be really important...

### DIFF
--- a/.chloggen/confmapmu.yaml
+++ b/.chloggen/confmapmu.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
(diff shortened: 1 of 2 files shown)


Aneurysm9 (changes requested):
> This may be a breaking change, but should not be really important...

I don't think we get to make that decision for users of this framework.

bogdandrutu:
@Aneurysm9 who gets?

bogdandrutu:
Extracting @Aneurysm9 comment from the Slack channel

>  I think that means it should be included in the changelog so that users who are trying to figure out what broke when they update aren’t stumped by missing changes in the changelog

I think it is a fair comment, doing that now.

codeboten (approved):

Issue #6420, opened 2022-10-27
Collector constantly breaking down


ambition-consulting:
**Describe the bug**
After a while the pod, on which the collector runs, stops with this error:

```
panic: runtime error: slice bounds out of range [-2:]
32
31
goroutine 137 [running]:
30
go.opentelemetry.io/collector/pdata/internal/data/protogen/common/v1.(*AnyValue).MarshalToSizedBuffer(0xc002dfd810, {0xc001e84000, 0x22, 0x30c16})
29
go.opentelemetry.io/collector/pdata@v0.62.1/internal/data/protogen/common/v1/common.pb.go:483 +0xd0
28
go.opentelemetry.io/collector/pdata/internal/data/protogen/common/v1.(*KeyValue).MarshalToSizedBuffer(0xc002dfd800, {0xc001e84000, 0x22, 0x30c16})
27
go.opentelemetry.io/collector/pdata@v0.62.1/internal/data/protogen/common/v1/common.pb.go:700 +0x3a
26
go.opentelemetry.io/collector/pdata/internal/data/protogen/resource/v1.(*Resource).MarshalToSizedBuffer(0xc000753360, {0xc001e84000, 0x30984?, 0x30c16})
25
go.opentelemetry.io/collector/pdata@v0.62.1/internal/data/protogen/resource/v1/resource.pb.go:146 +0xf0
24
go.opentelemetry.io/collector/pdata/internal/data/protogen/trace/v1.(*ResourceSpans).MarshalToSizedBuffer(0xc000753360, {0xc001e84000, 0x30984, 0x30c16})
23
go.opentelemetry.io/collector/pdata@v0.62.1/internal/data/protogen/trace/v1/trace.pb.go:890 +0x105
22
go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/trace/v1.(*ExportTraceServiceRequest).MarshalToSizedBuffer(0xc003408198, {0xc001e84000, 0x30c16, 0x30c16})
21
go.opentelemetry.io/collector/pdata@v0.62.1/internal/data/protogen/collector/trace/v1/trace_service.pb.go:351 +0xac
20
go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/trace/v1.(*ExportTraceServiceRequest).Marshal(0xc0cebc8861798197?)
19
[...]

PR #6516, opened 2022-11-10
Fix logging exporter to not mutate the data

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6420

Affected versions:
* v0.62.0
* v0.62.1
* v0.63.0
* v0.64.0

If anyone experience this, simple fix is to remove the logging exporter.

### DIFF
--- a/exporter/loggingexporter/internal/otlptext/databuffer.go
+++ b/exporter/loggingexporter/internal/otlptext/databuffer.go
@@ -52,7 +52,7 @@ func (b *dataBuffer) logAttributes(header string, m pcommon.Map) {
 		attrPrefix = headerParts[0] + attrPrefix
 	}
 
-	m.Sort().Range(func(k string, v pcommon.Value) bool {
+	m.Range(func(k string, v pcommon.Value) bool {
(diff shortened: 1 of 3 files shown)


dmitryax (approved):
LGTM. Thanks for fixing it!

codeboten (approved):

exporter/loggingexporter/internal/otlptext/databuffer.go
```diff
@@ -52,7 +52,7 @@ func (b *dataBuffer) logAttributes(header string, m pcommon.Map) {
 		attrPrefix = headerParts[0] + attrPrefix
 	}
 
-	m.Sort().Range(func(k string, v pcommon.Value) bool {
```
tigrannajaryan, line 55: 👍

tigrannajaryan:
Can we have a "debug" mode that detects if exporters mutate data? E.g. clone data before giving it to exporter, then compare after exporting is finished?

bogdandrutu:
> Can we have a "debug" mode that detects if exporters mutate data? E.g. clone data before giving it to exporter, then compare after exporting is finished?

We can implement that, maybe using a build tag.

PR #6522, opened 2022-11-11
Stop notification for signals before shutdown, increase channel size.

bogdandrutu:


service/collector.go
```diff
@@ -99,7 +99,7 @@ func New(set CollectorSettings) (*Collector, error) {
 		set:               set,
 		state:             atomic.NewInt32(int32(Starting)),
 		shutdownChan:      make(chan struct{}),
-		signalsChannel:    make(chan os.Signal, 1),
+		signalsChannel:    make(chan os.Signal, 3),
```
codeboten, line 102: Why increase the channel size? Are we missing signals because of its current size?

bogdandrutu: We don't know if we miss or not (no information about that), but I read the godoc of the Notify:

>Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.

Based on this, I thought that because we are using the channel for 3 signals, a size of 3 should be better.

dmitryax: If it's based on the number of signals, I believe we need to mention it in a comment or move to a constant

bogdandrutu: Done.

> I believe we need to mention it in a comment or move to a constant

We don't need but it is probably good to document.

Issue #6524, opened 2022-11-11
Reload client CAs periodically


akorompai:
**Is your feature request related to a problem? Please describe.**
Need to restart the collector in order to reload the client CAs.

**Describe the solution you'd like**
Make it work like [certificate reloader](https://github.com/open-telemetry/opentelemetry-collector/blob/e07665cacabfdf6769f2c01576abde9786d0160b/config/configtls/configtls.go#L178) in same context.
Could use `GetClientConfig` callback as a hook to reload the client CAs from disk.

atoulme:
The issue appears fixed with #6708. Please reopen if more work is needed.

PR #6525, opened 2022-11-11
Reload client CAs periodically

akorompai:
**Description:** Reload client CAs used for mTLS in sim

**Link to tracking Issue:**  #6524

**Testing:** manual only

**Documentation:** none

### DIFF
--- a/config/configtls/configtls.go
+++ b/config/configtls/configtls.go
@@ -229,6 +229,67 @@ func (c TLSClientSetting) LoadTLSConfig() (*tls.Config, error) {
 	return tlsCfg, nil
 }
 
+// clientCAsReloader is a wrapper object for certificate reloading
+// Its GetClientCAs method will either return the current client CAs or reload from disk
+// if the last reload happened more than ReloadInterval ago
+type clientCAsReloader struct {
+	// Path to the client CAs
+	ClientCAsFile string
+	// ReloadInterval specifies the duration after which the certificate will be reloaded
+	// If not set, it will never be reloaded (optional)
+	ReloadInterval time.Duration
+	nextReload     time.Time
+	certPool       *x509.CertPool
+	lock           sync.RWMutex
+	c              *TLSServerSetting
+}
+
+func newClientCAsReloader(clientCAsFile string, reloadInterval time.Duration, c *TLSServerSetting) (*clientCAsReloader, error) {
+	certPool, err := c.loadCert(clientCAsFile)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


config/configtls/configtls.go
```diff
@@ -229,18 +229,82 @@ func (c TLSClientSetting) LoadTLSConfig() (*tls.Config, error) {
...
+type clientCAsReloader struct {
+	// Path to the client CAs
+	ClientCAsFile string
+	// ReloadInterval specifies the duration after which the certificate will be reloaded
+	// If not set, it will never be reloaded (optional)
+	ReloadInterval time.Duration
+	nextReload     time.Time
+	certPool       *x509.CertPool
+	lock           sync.RWMutex
+	c              *TLSServerSetting
```
bogdandrutu, line 244: We already have a reload mechanism, what does this bring?

bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/configtls.go#L105

akorompai: Hi @bogdandrutu ! 

I'll gladly admit that this is a blatant copy of `certReloader`, but I found that the reloading of `client_ca_file` and `cert_file/key_file` are different enough concerns to have their own reloaders.
One first one is only related to server, the other one is related to both client and server, their lifecycles differ, etc...
Please let me know what would be a better approach!

akorompai: @bogdandrutu  meanwhile can you allow the CI to run, that would be a great feedback for us!

jpkrohling:
Are certs always local files at the moment? If so, why not just watch the file and reload on that event, instead of reloading it periodically?

akorompai:
That would work for us too, this PR extend the existing pattern to reload client CAs.
Would that have a better chance to be accepted?

jpkrohling:
> Would that have a better chance to be accepted?

I would certainly prefer watching files to running a process periodically. We have this pattern elsewhere as well. I believe @frzifus implemented this recently for Jaeger's remote sampling extension.

frzifus:
We implemented something to reload on change here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.66.0/extension/bearertokenauthextension/bearertokenauth.go#L101-L134

akorompai:
Could you check #6708 ? tried to adapt @frzifus 's solution.

akorompai:
closing in favor of #6708

Issue #6490, opened 2022-11-07
String representation of `component.StabilityLevel` and `configtelemetry.Level` is inconsistent with other types


dmitryax:
`String` methods of all types in `pdata` module along with `service.State` return capitalized strings, for example:
- `pmetric.AggregationTemporality.String()`: "Unspecified", "Delta" or "Cumulative"
- `service.State.String()`: "Starting", "Running", "Closing", etc.

But `component.StabilityLevel` and `configtelemetry.Level` return lowercase strings:
- component.StabilityLevel.String(): "unmaintained", "deprecated", "in development", etc.
- `configtelemetry.Level.String()`: "none", "basic", "normal" or"detailed"

Should we align the to the same Capitalized format?

Or there is still a chance to make all other strings lowercase if we think that's a better option.

cc @bogdandrutu @codeboten @Aneurysm9

bogdandrutu:
I think I like the "Capitalized" version more.

Aneurysm9:
I also prefer capitalization.  It aligns with my expectation that Capitalized Terms are well-defined somewhere in the domain and may carry denotations beyond what might otherwise be considered as their plain meaning.

bogdandrutu:
@dmitryax for configs I think we should allow any capitalization? I think it is fine if in the config file (like `configtelemetry.Level`) we allow users to write `"none"`, `"NONE"` or `"None"`.

dmitryax:
> @dmitryax for configs I think we should allow any capitalization? I think it is fine if in the config file (like configtelemetry.Level) we allow users to write "none", "NONE" or "None".

Yes, that's how it's currently done in `configtelemetry.Level`, input of any case is supported

codeboten:
Capitalized is fine with me.

dmitryax:
Sounds good! I'll submit a PR to capitalize `component.StabilityLevel` and `configtelemetry.Level` strings.

bogdandrutu:
For `CompressionType` there are some things:
* Capitalization does not make that much sense
* Maybe rename to `Compression`, but anyway we also need to fix the enum value names.

atoulme:
It looks like the original issue is completed. Bogdan mentions we could look at renaming `CompressionType` `Compression` additionally.

TylerHelmuth:
I've captured the configcompression piece in its own issue.  Closing this as completed.

PR #6531, opened 2022-11-12
[component] Update values returned by `StabilityLevel.String` method

dmitryax:
This change updates values returned by `StabilityLevel.String` to be consistent with other enum types:
  - All returned strings are capitalized.
  - "Undefined" is returned only for `StabilityLevelUndefined`.
  - "" is returned for integers that are out of StabilityLevel enum range.

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/6490

### DIFF
--- a/component/component.go
+++ b/component/component.go
@@ -120,20 +120,22 @@ const (
 
 func (sl StabilityLevel) String() string {
 	switch sl {
+	case StabilityLevelUndefined:
+		return "Undefined"
 	case StabilityLevelUnmaintained:
-		return "unmaintained"
+		return "Unmaintained"
 	case StabilityLevelDeprecated:
-		return "deprecated"
+		return "Deprecated"
 	case StabilityLevelInDevelopment:
-		return "in development"
+		return "In development"
 	case StabilityLevelAlpha:
-		return "alpha"
+		return "Alpha"
 	case StabilityLevelBeta:
-		return "beta"
+		return "Beta"
 	case StabilityLevelStable:
-		return "stable"
+		return "Stable"
 	}
-	return "undefined"
+	return ""
 }
 
 func (sl StabilityLevel) LogMessage() string {
--- a/.chloggen/update-stability-level-string.yaml
+++ b/.chloggen/update-stability-level-string.yaml
@@ -1,0 +1,8 @@
+change_type: breaking
+component: component
+note: Update values returned by `StabilityLevel.String` method.
+issues: [6490]
+subtext: |
+  - All returned strings are capitalized.
+  - "Undefined" is returned only for `StabilityLevelUndefined`.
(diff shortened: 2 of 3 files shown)


component/component.go
```diff
@@ -120,20 +120,22 @@ const (
...
+		return "Undefined"
 	case StabilityLevelUnmaintained:
-		return "unmaintained"
+		return "Unmaintained"
 	case StabilityLevelDeprecated:
-		return "deprecated"
+		return "Deprecated"
 	case StabilityLevelInDevelopment:
-		return "in development"
+		return "In development"
```
bogdandrutu, line 130: One world "InDevelopment"? Or just "Development"?

bogdandrutu (approved):

component/component.go
dmitryax: I like "InDevelopment" but not sure how it'll work if inserted a sentence, maybe " In Development"?

bogdandrutu: "Prototype" :))))

dmitryax: I believe we eventually have to establish a pattern for identifiers with more than one word :)

bogdandrutu: I think `InDevelopment` is the right one here.

codeboten: Could be shortened to "Development".

codeboten (approved):

component/component.go
bogdandrutu: +1

dmitryax: If we choose "Development", we will need to rename the enum to `StabilityLevelDevelopment` as well. IMO `InDevelopment` is a bit more clear. And I'm good with sticking the words together, but we need to agree to this as a common practice going forward

dmitryax: If we choose "Development", we will need to rename the enum to `StabilityLevelDevelopment` as well. IMO `InDevelopment` is a bit more clear. And I'm good with sticking the words together, but we need to agree to this as a common practice going forward

bogdandrutu: Agree on the rename if we go with `Development`. Not strong opinion.

dmitryax: Ok, let's call it "Development"

dmitryax: I looked at a few other projects that have this stability level and found "Experimental", "Dev", "Devel", "Development"

bogdandrutu: "Development" is better, "Alpha" == "Experimental" for me, sort of :))

dmitryax: Ok, I submitted a separate PR for that https://github.com/open-telemetry/opentelemetry-collector/pull/6561

bogdandrutu:
I merged this, since we have a followup for the "development" name.

PR #6533, opened 2022-11-12
[configtelemetry] Update strings returned by `Level.[String|MarshalText]`

dmitryax:
- All returned strings are capitalized.
  - "" is returned for integers that are out of Level enum range.
  - It also affects `Level.MarshalText` output, but it's not a problem because `UnmarshalText` method accepts strings in all cases, e.g. "normal", "Normal" and "NORMAL".

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/6490

### DIFF
--- a/config/configtelemetry/configtelemetry.go
+++ b/config/configtelemetry/configtelemetry.go
@@ -31,10 +31,10 @@ const (
 	// LevelDetailed adds dimensions and views to the previous levels.
 	LevelDetailed
 
-	levelNoneStr     = "none"
-	levelBasicStr    = "basic"
-	levelNormalStr   = "normal"
-	levelDetailedStr = "detailed"
+	levelNoneStr     = "None"
+	levelBasicStr    = "Basic"
+	levelNormalStr   = "Normal"
+	levelDetailedStr = "Detailed"
 )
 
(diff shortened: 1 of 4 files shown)


config/configtelemetry/configtelemetry.go
```diff
@@ -71,16 +66,16 @@ func (l *Level) UnmarshalText(text []byte) error {
...
-	case levelBasicStr:
+	case strings.ToLower(LevelBasic.String()):
 		*l = LevelBasic
 		return nil
-	case levelNormalStr:
+	case strings.ToLower(LevelNormal.String()):
 		*l = LevelNormal
 		return nil
-	case levelDetailedStr:
+	case strings.ToLower(LevelDetailed.String()):
```
bogdandrutu, line 78: I know that it is not performance critical this code, but does not seem nice to call toLower every time, also call String which is another switch every time...

Maybe keep the constants as capitalized at least and here you have `strings.ToLower(levelNoneStr)`?

dmitryax: > Maybe keep the constants as capitalized at least and here you have strings.ToLower(levelNoneStr)?

Ok, applied this suggestion

bogdandrutu (approved):

Issue #6507, opened 2022-11-08
[Service] Partial start can cause panic


MovieStoreGuy:
**Describe the bug**

On service start up, if a component was to error, the start up process is stopped and then the shutdown method is invoked.

If a component requires `Start` to be called in order to populate internal values (ie, storing a reference to a context.CancelFunc), the `service.Shutdown` will panic and the original error is lost.

**Steps to reproduce**
If possible, provide a recipe for reproducing the error.

Have two components, one that errors on start up and another that requires `Start` to be called for the `Shutdown` method to be valid.

For example the prometheus receiver requires `Start` to be called in order for `Shutdown` to be valid.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/metrics_receiver.go#L307-L312

However, I believe that is a seperate concern.

**What did you expect to see?**

I think there is a few things that are part of this is clarify the expected behaviour for `component.Componet` to include if `Shutdown` can be called without a call to `Start`.

I think resolve that expectation will dictate what would be the following steps.

**What did you see instead?**

As the bug describes, it is possible for a graceful shutdown to panic.

**What version did you use?**
Version: v0.63.0

However, this impacts all versions I believe

**What config did you use?**
NA

**Environment**
OS: Darwin/arm & linux/arm (within a docker container)
Compiler(if manually compiled): go 1.18.4 and go 1.19.2

bogdandrutu:
> On service start up, if a component was to error, the start up process is stopped and then the shutdown method is invoked.

Should we call "Shutdown" only on the components that we started?

MovieStoreGuy:
I know we had spoken about this over slack but just to recap:

> Should we call "Shutdown" only on the components that we started?

I don't think we should, mainly because that would imply that the service is managing that state and I think would over complicate cognitive model, and potential break the OpAMP. 

However, I do think that following the example of the behaviour for `http.Server` as `Shutdown` is safe to call regardless of starting however, once Shutdown correctly, `nil` is returned along with any repeated called to `Shutdown` and if the shutdown errors then the original error is returned for each repeated calls. 

What I would expect as a definition of done:

- The existing components description is updated to clarify behaviour
- A component helper method to wrap this functionality for the components
- Updated lifecycle tests to include validating shutting down a component is safe without being started

cc: @tigrannajaryan since this may impact work with OpAMP

tigrannajaryan:
[...]

PR #6535, opened 2022-11-12
[component] clarifying lifecycle expectation

MovieStoreGuy:
**Description:**

Extended the documentation of the component to include safe `Shutdown` expectations.

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector/issues/6507

**Testing:** 
N/A

**Documentation:** 
N/A


MovieStoreGuy:
I've include a change log for this PR since this documentation impacts all components of the project and contrib.

.chloggen/msg_update-docs-for-component.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
codeboten, line 2: I don't think this qualifies as a breaking change, since nothing is changing here.

component/component.go
```diff
@@ -57,6 +57,9 @@ type Component interface {
 	// Shutdown is invoked during service shutdown. After Shutdown() is called, if the component
 	// accepted data in any way, it should not accept it anymore.
 	//
+	// This method must be safe to call without having Start()
+	// being called or being in a shutdown state.
```
codeboten, line 61:
```suggestion
	// This method must be safe to call:
	//   - without Start() having been called
	//   - if the component is in a shutdown state already
```

codeboten (commented):
Clarified the comment a bit, as i found it confusing to read, PTAL

.chloggen/msg_update-docs-for-component.yaml
MovieStoreGuy: Sorry, I was thinking from this perspective the perspective of existing components that didn't expect to be `Stop` without being `Start`ed.

bogdandrutu (approved):

.chloggen/msg_update-docs-for-component.yaml
dmitryax: I don't believe that a doc string change needs a changelog entry at all. I would like to keep the changelog limited to user facing changes. Doc updates shouldn't be included IMO. Otherwise it'll be hard to find significant changes in the list

MovieStoreGuy: Let me remove this :)

codeboten (approved):

PR #6540, opened 2022-11-14
[pipelines] Change test to not reuse same processor twice in one pipeline

djaglowski:
There is documentation [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/design.md?plain=1#L125) that explains how a single processor can be specified in multiple pipelines, with the understanding that each pipeline will get its own instance. However, there is nothing specifically stated about the case where a single processor is specified multiple times in a single pipeline.

This PR is as much a question as a proposal. Is there an explicit intention to support this use case, or should we clarify and disallow it?

### DIFF
--- a/.chloggen/processor-duplication-per-pipeline.yaml
+++ b/.chloggen/processor-duplication-per-pipeline.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Disallow duplicate references to processors within a single pipeline
+
+# One or more tracking issues or pull requests related to the change
+issues: [6540]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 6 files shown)


djaglowski:
cc: @bogdandrutu, @tigrannajaryan

tigrannajaryan:
I think there can be valid reasons to use the same processor _type_ twice in one pipeline (each processor having a different config). I am not sure though if there are valid use cases for using the same processor type with the same config twice.

bogdandrutu:
> This PR is as much a question as a proposal. Is there an explicit intention to support this use case, or should we clarify and disallow it?

I don't necessary know about an use-case, so we can disallow this for the moment. If we allow it we should document that a new instance is created.

djaglowski:
Ok, sounds like we are in agreement that this is not a supported configuration. I've added a note to the documentation. 

Should I also explicitly disallow this in code?

mx-psi (commented):
I agree with the docs change (I don't see much value in supporting this) and I think we should fix it in the code too

djaglowski:
I've added a check in code. Should we consider this a breaking change or just a bug fix?

adam-kiss-sg:
I just run into this error message. My use-case is using multiple `batch` processor in the same pipeline. From the batch processor docs:

```
The batch processor should be defined in the pipeline after the memory_limiter as well as any sampling processors.
```

So based on that i have something like:

```
processors: [memory_limiter, batch, filter, batch]
```

where filter filters out a lot of metrics. 

Am i using the batch processor wrong? What should be the best practice here?

mx-psi:
@adam-kiss-sg 

> What should be the best practice here?

IMO that's acceptable use of the batch processor. You just need to define two instances of the processor:

```yaml
processors:
  batch:
    # maybe some configuration here
  batch/2:
    # maybe some configuration here

#  Omitted rest of the config

service:
  pipelines:
    traces:
      receiver: [otlp]
      processors: [memory_limiter, batch, filter, batch/2]
      exporters: [otlp]
```

Issue #6508, opened 2022-11-09
Adding new data types to pdata module


dmitryax:
This is part of https://github.com/open-telemetry/opentelemetry-collector/issues/4705. 

Going forward we will need to add new signal types that are initially considered experimental. Before releasing pdata 1.0, we need to decide how we will be doing that given that the new signals probably shouldn't be part of pdata 1.0 right when they are added.

Currently, we have `pdata` as one module. This issue is intended to decide if we want to keep it this way or if we want to break it down into 4 modules. Let's consider both approaches applying profiling data signal as a possible candidate for the new experimental data type.

### Breaking down pdata into more modules

Breaking down pdata into more modules makes the addition of the new experimental signals straightforward. Let's imagine we add profiling data when stable modules are at the `1.2.0` version and beta modules are at `0.70.0`. The pdata modules structure would look like this:

- `go.opentelemetry.io/collector/pdata/pcommon 1.2.0`
- `go.opentelemetry.io/collector/pdata/plog 1.2.0`
- `go.opentelemetry.io/collector/pdata/pmetric 1.2.0`
- `go.opentelemetry.io/collector/pdata/ptrace 1.2.0`
- `go.opentelemetry.io/collector/pdata/pprofiling 0.70.0`

Once the profiling signal is stabilized, it's just upgraded to the current stable 1.x version

### Keeping one module

If we keep `pdata` as one module, things become a bit more complicated. According to the [go package resolving process](https://go.dev/ref/mod#resolve-pkg-mod):

> If one or more modules contain the requested package, the module with the longest path is used.

It means that we cannot introduce `go.opentelemetry.io/collector/pdata/pprofiling 0.70.0` and then make it part of `go.opentelemetry.io/collector/pdata 1.x` without changing the package path. So we will need to introduce an experimental module like `go.opentelemetry.io/collector/pdata/experimental 0.70.0` where we can put `pprofiling` until it's stabilized and moved to `go.opentelemetry.io/collector/pdata/pprofiling`

---

Given all the complications with adding experimental new signals to pdata, I'd lean towards further breaking down `pdata` package. We can break it down after 0.64.0 release and set the version of the new modules to `1.0.0-rc.1`. Looking for other opinions. 

cc @open-telemetry/collector-approvers

dmitryax:
> If I remember correctly it is backwards compatible to break the pdata into smaller packages later (after 1.0). If that is the case why not doing that when there is a need for a new package?

[...]

PR #6543, opened 2022-11-14
split component into its own module

### LINKED ISSUE #6187: Split `component` into its own module

codeboten:
Fixes #6187


.chloggen/split-components.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
bogdandrutu, line 2: Based on our definition, this is NOT a breaking change.

bogdandrutu (approved):

.chloggen/split-components.yaml
codeboten: Since we're asking users of the component package to import a different module, isn't it breaking? The other splits were added to the changelog as breaking.

bogdandrutu: > Since we're asking users of the component package to import a different module, isn't it breaking?

We don't ask, if they run `go mod tidy` is done (which they should to do the go.sum)

bogdandrutu: With that in mind, we also made this decision https://github.com/open-telemetry/opentelemetry-collector/issues/6508#issuecomment-1309342642 

So we should be ok to do this.

bogdandrutu: My point is, if we disagree and you believe this is a breaking change, lots of previous decisions should be revisited like the one I linked.

codeboten: Since with `component` module, users of the builder wouldn't be impacted, I think this is ok to mark as enhancement. Once this is merged I will do further testing w/ contrib to ensure this is true.

codeboten: Confirmed this is not a breaking change as per https://github.com/open-telemetry/opentelemetry-collector/actions/runs/3474306055/jobs/5807319018

Thanks for helping me work through this 👍🏻

PR #6544, opened 2022-11-15
Remove Validate() from Config interfaces and make it optional interface

bogdandrutu:
The reason is to remove the need that every Config includes the "config.*Settings" structs, and remove these structs completely.

One of the functionality they offer right now is that they provide the "default" implementation for `Validate`, and this PR provides a solution to that, by making the `Validate` a proper optional functionality to implement instead of optionally override it.

Alternative is to make mandatory for every `component.*Config` to implement `Validate` and not provide a default in the `config.*Settings`, that also solves the dependency problem (between component and config), but it comes with a small caveat, since if we want to extend the capability to validate sub-configs then for these will be optionally implemented functionality, because of that I believe it is more consistent to make it optionally implementable everywhere.

### DIFF
--- a/.chloggen/rmvalidate.yaml
+++ b/.chloggen/rmvalidate.yaml
@@ -1,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove `Validate()` from component.*Config interfaces and make it optional interface
+
+# One or more tracking issues or pull requests related to the change
+issues: [6544]
--- a/component/config.go
+++ b/component/config.go
@@ -20,8 +20,6 @@ import (
 
 type Config interface {
 	identifiable
-	// Deprecated: [v0.65.0] use component.ValidateConfig.
-	Validate() error
 
 	privateConfig()
 }
--- a/config/exporter.go
+++ b/config/exporter.go
@@ -44,8 +44,3 @@ func (es *ExporterSettings) ID() component.ID {
 func (es *ExporterSettings) SetIDName(idName string) {
 	es.id = component.NewIDWithName(es.id.Type(), idName)
 }
-
-// Deprecated: [v0.65.0] use component.ValidateConfig.
-func (es *ExporterSettings) Validate() error {
-	return nil
-}
--- a/config/extension.go
+++ b/config/extension.go
@@ -44,8 +44,3 @@ func (es *ExtensionSettings) ID() component.ID {
 func (es *ExtensionSettings) SetIDName(idName string) {
 	es.id = component.NewIDWithName(es.id.Type(), idName)
 }
-
-// Deprecated: [v0.65.0] use component.ValidateConfig.
-func (es *ExtensionSettings) Validate() error {
-	return nil
-}
--- a/config/processor.go
+++ b/config/processor.go
@@ -44,8 +44,3 @@ func (ps *ProcessorSettings) ID() component.ID {
 func (ps *ProcessorSettings) SetIDName(idName string) {
 	ps.id = component.NewIDWithName(ps.id.Type(), idName)
 }
-
-// Deprecated: [v0.65.0] use component.ValidateConfig.
-func (ps *ProcessorSettings) Validate() error {
-	return nil
-}
--- a/config/receiver.go
+++ b/config/receiver.go
@@ -44,8 +44,3 @@ func (rs *ReceiverSettings) ID() component.ID {
 func (rs *ReceiverSettings) SetIDName(idName string) {
 	rs.id = component.NewIDWithName(rs.id.Type(), idName)
 }
-
-// Deprecated: [v0.65.0] Not needed anymore since the Validate() will be moved from Config interface to the optional ConfigValidator interface.
-func (rs *ReceiverSettings) Validate() error {
-	return nil
-}


.chloggen/rmvalidate.yaml
```diff
@@ -0,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
bogdandrutu, line 2: It is "breaking" mostly for test (may be other cases as well), where we do `factory.DefaultConfig().Validate()` which no longer compiles.

component/receiver.go
```diff
@@ -24,10 +24,7 @@ import (
 // ReceiverConfig is the configuration of a component.Receiver. Specific extensions must implement
```
codeboten, line 24:
```suggestion
// ReceiverConfig is the configuration of a component.Receiver. Specific receivers must implement
```

component/processor.go
```diff
@@ -24,10 +24,7 @@ import (
 // ProcessorConfig is the configuration of a component.Processor. Specific extensions must implement
```
codeboten, line 24:
```suggestion
// ProcessorConfig is the configuration of a component.Processor. Specific processors must implement
```

component/exporter.go
```diff
@@ -24,10 +24,7 @@ import (
 // ExporterConfig is the configuration of a component.Exporter. Specific extensions must implement
```
codeboten, line 24:
```suggestion
// ExporterConfig is the configuration of a component.Exporter. Specific exporters must implement
```

component/config.go
```diff
@@ -21,12 +21,29 @@ import (
...
-type validatable interface {
-	// Validate validates the configuration and returns an error if invalid.
+type Config interface {
+	identifiable
+
+	privateConfig()
+}
+
+// ConfigValidator defines an optional interface for configurations to implement to do validation.
+type ConfigValidator interface {
```
codeboten, line 31: Why make this optional? I thought we wanted to ensure every component implemented this, otherwise, as we saw in contrib, each component implemented validated in a slightly different way, and some components did not implement this at all.

codeboten (commented):
> The reason is to remove every Config to include the "config.*Settings" structs, which is one of the functionality they offer.

Please clarify the description, i'm not sure I follow

component/processor.go
bogdandrutu: Done separately since not related to this PR. https://github.com/open-telemetry/opentelemetry-collector/pull/6551

component/receiver.go
bogdandrutu: Done separately since not related to this PR. https://github.com/open-telemetry/opentelemetry-collector/pull/6551

component/exporter.go
bogdandrutu: Done separately since not related to this PR. https://github.com/open-telemetry/opentelemetry-collector/pull/6551

bogdandrutu:
> Please clarify the description, i'm not sure I follow

Tell me what you understand to see where the disconnect is.

component/config.go
bogdandrutu: Right now is optional anyway since we are implementing it in the "*Settings", so only some components override it. This PR does not change that it any way, it changes only on how the "optionality" is implemented.

If we want to move to enforce "Validate" it is fine, but will be a big breaking change I think. Happy to do that way as well, as long as we no longer rely on "Settings" to have the "default" implementation for this func.

codeboten:
> > Please clarify the description, i'm not sure I follow
> 
> Tell me what you understand to see where the disconnect is.

> > The reason is to remove every Config to include the "config.*Settings" structs, which is one of the functionality they offer.

Are you saying that we're removing `Config` from every `config.[Exporter|Extension|Processor|Receiver]Settings` struct?

bogdandrutu:
> Are you saying that the goal is to remove Config from every config.[Exporter|Extension|Processor|Receiver]Settings struct?

I am saying that the goal is to remove the `config.[Exporter|Extension|Processor|Receiver]Settings` structs. Right now one reason to have them is because the have the default implementation of the "Validate" func.

component/config.go
codeboten: Any thoughts on how big of a breaking change? I recall adding Validate to *Settings caused a big breaking change because so many components didn't have a Validate func. I'd be curious if there are many more left that do not.

What will be the way to replace *Settings? Is it to embed `component.ID` into the specific component struct directly?

codeboten:
> > Are you saying that the goal is to remove Config from every config.[Exporter|Extension|Processor|Receiver]Settings struct?
> 
> I am saying that the goal is to remove the `config.[Exporter|Extension|Processor|Receiver]Settings` structs. Right now one reason to have them is because they have the default implementation of the "Validate" func.

Thanks for clarifying, this makes more sense to me now :)

component/config.go
bogdandrutu: > Any thoughts on how big of a breaking change? I recall adding Validate to *Settings caused a big breaking change because so many components didn't have a Validate func. I'd be curious if there are many more left that do not.

Let me google(grep) that for you:

```bash
➜  opentelemetry-collector-contrib git:(main) grep -R "type Config struct" | wc -l
     212
➜  opentelemetry-collector-contrib git:(main) grep -R "func (.*Config) Validate() error" | wc -l 
     106
```

> What will be the way to replace *Settings? Is it to embed component.ID into the specific component struct directly?

`config.*Settings` are used for 2 reasons (you can also see in the code):
* Implement the `identifiable` interface to avoid duplicate code of implementing that.
* Implement the `validatable` interface to not force everyone to have to implement it.
  
For the `validatable` this is the proposed solution; For `identifiable` I actually think to not pass the ID via the Config between the Host/Service and the `Component` but to pass it via the `CreateSettings` (see https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/receiver.go#L136)

dmitryax:
I like the idea of removing `config.[Exporter|Extension|Processor|Receiver]Settings` structs.

Somewhat related question: I was looking into adding a built-in way to report deprecated fields in user settings. This should happen during the same time when the validation is done. This can be done by calling another method that component developers can optionally add to their configs. I'm wondering if we can have another optional config interface with a method like `Inspect(component.TelemetrySettings)` or `Inspect(zap.Logger)`, or maybe with a more specific method like `Deprecations() error` that would return deprecations as errors. Another option is to change the existing `Validate` method and pass `component.TelemetrySettings` or `zap.Logger` to it, but this will be a significant breaking change. Please let me know what you think about it and if you have a better idea for that.

bogdandrutu:
@dmitryax maybe we can define a new error type WarningError and not fail if you return only that? Give a try to this idea if you want, I haven't prove it :)

dmitryax:
> @dmitryax maybe we can define a new error type WarningError and not fail if you return only that? Give a try to this idea if you want, I haven't prove it :)

I was considering this option but didn't want to mix the errors. I'll give it a try. Thanks!

dmitryax (commented):
LGTM. Can you resolve the issues on the contrib side first to make contrib-test pass?

codeboten (approved):

component/config.go
bogdandrutu: That note was resolved offline, because the mapping (ID -> Config) still exists in the service.Config.

bogdandrutu:
@codeboten @dmitryax managed to find a way to initially deprecate the method on the interface, this way we can merge this PR next release and contrib will be happy.

See https://github.com/open-telemetry/opentelemetry-collector/pull/6572

dmitryax:
> @codeboten @dmitryax managed to find a way to initially deprecate the method on the interface, this way we can merge this PR next release and contrib will be happy.

But do we need https://github.com/open-telemetry/opentelemetry-collector/pull/6572 to make contrib tests happy? Can't we just remove unnecessary `Validate()` calls that are removed from this PR?

bogdandrutu:
> But do we need https://github.com/open-telemetry/opentelemetry-collector/pull/6572 to make contrib tests happy? Can't we just remove unnecessary Validate() calls of the methods that are removed in this PR?

Unfortunately I cannot understand your comment, sorry. Can you clarify what you are looking for? As mentioned in my comment, we are going to merge #6572 first.

bogdandrutu:
Converting to draft, until #6572 is merged and released.

dmitryax:
> Unfortunately I cannot understand your comment, sorry. Can you clarify what you are looking for? As mentioned in my comment, we are going to merge https://github.com/open-telemetry/opentelemetry-collector/pull/6572 first.

I mean that contrib tests are failing because of unnecessary calls of noop `Validate()` methods which are removed in this PR, the calls can be removed from the contrib instead to make the tests happy I believe. That was my original suggestion, but I'm happier with https://github.com/open-telemetry/opentelemetry-collector/pull/6572 :)

bogdandrutu:
@codeboten @dmitryax as expected with the preliminary step, this becomes no breaking change for contrib.

bogdandrutu:
@codeboten @dmitryax PTAL, now that release is done.

dmitryax:
LGTM 👍

PR #6552, opened 2022-11-15
Draft changes to show how we can split component

bogdandrutu:

PR #6553, opened 2022-11-15
Minimum changes to component.Host, to allow split of the component package

bogdandrutu:


component/componenttest/nop_host.go
```diff
@@ -32,10 +32,10 @@ func (nh *nopHost) GetFactory(_ component.Kind, _ component.Type) component.Fact
 	return nil
 }
 
-func (nh *nopHost) GetExtensions() map[component.ID]component.Extension {
+func (nh *nopHost) GetExtensions() map[component.ID]component.Component {
```
codeboten, line 35: This would impact anyone who's implemented their own Host correct? Any thoughts on the correct deprecation path for this?

codeboten (commented):
Just one question, one way I could imagine doing this gracefully is by providing a new `HostWithComponent` interface and deprecating `Host` in favour of it, any thoughts?

bogdandrutu:
@codeboten the idea behind this draft PR was mostly to agree on the path forward that it is possible to split `component` as I proposed initially, if we do this change (if we are ok with the result). We can debate how to get there, but important is the part that we have a path forward for the split of component, and I showed that in #6552

codeboten:
> @codeboten the idea behind this draft PR was mostly to agree on the path forward that it is possible to split `component` as I proposed initially, if we do this change (if we are ok with the result). We can debate how to get there, but important is the part that we have a path forward for the split of component, and I showed that in #6552

I'm ok with this plan, assuming others are as well. I think the changes to Host are minimal enough and help us untangle all the other components into more meaningful packages. I'm in favour of #6552

component/componenttest/nop_host.go
dmitryax: Can we convert  `component.[Extension|Processor|...]` into deprecated aliases of `component.Component` for some time?

dmitryax:
Do we to remove `component.Exporter` and `component.Extension` only? What about `component.Receiver` and `component.Processor`?

component/componenttest/nop_host.go
codeboten: This wouldn't change the breakage that would happen when the update to the `Host` interface happens... short of adding a new `WithComponent` interface and a new `Start` func, i'm not sure how to move forward without breaking this. I believe @tigrannajaryan hit something similar here: https://github.com/open-telemetry/opentelemetry-collector/pull/6560#discussion_r1025300926

dmitryax: I actually missed that only `Exporter` is removed in this PR. I assumed the `Extension` going away as well

dmitryax: > This wouldn't change the breakage that would happen when the update to the Host interface happens...

It's not a breaking change when parts of API are replaced with identical (aliased) types. We did that a lot in pdata.

bogdandrutu: > It's not a breaking change when parts of API are replaced with identical (aliased) types. We did that a lot in pdata.

It is, because now we will have `type Extension = Component` then I want to revert that to 

```golang
type Extension interface {
  Component
}
```

bogdandrutu: The reason it is, is the same reason why `GetExtension` change in this PR, is a breaking change.

dmitryax:
> Extension is the only one that needs to stay as a type

Can you please point me to why? I'm not sure splitting `component` package justifies removing only `component.Exporter` and leaving others as is

bogdandrutu:
> Can you please point me to why? I'm not sure splitting component package justifies removing only component.Exporter and leaving others as is

Not sure I understand, I said that only "Extension" needs to stay as a type since we don't have a `TracesExtension`, all the other will go, as you see in the PR.

component/componenttest/nop_host.go
bogdandrutu: We do the alias now, and try to figure out how to "unalias" without breaking, may be impossible, but should be minimal impact if any.

component/extension.go
```diff
@@ -36,9 +36,7 @@ func UnmarshalExtensionConfig(conf *confmap.Conf, cfg ExtensionConfig) error {
 // Extension is the interface for objects hosted by the OpenTelemetry Collector that
 // don't participate directly on data pipelines but provide some functionality
 // to the service, examples: health check endpoint, z-pages, etc.
-type Extension interface {
-	Component
-}
+type Extension = Component
```
dmitryax, line 39: What do you think if we do the following?

1. Introduce `Extension` alias to `Component` and add another Extension interface with a temporary name like `TempExtension`

```
type TempExtension interface {
	Component
}

// Deprecated: [0.65.0] Use Component or TempExtension depending on the use case:
// Use Component in Host.GetExtensions() method: GetExtensions() map[component.ID]component.Component
// Use TempExtension in any other case.
type Extension = Component
```

2. Remove the `Extension` alias and move `TempExtension` to `extension.Extension` by deprecating the former.

3. Remove `TempExtension`

This approach may be a bit more disruptive but aligns with our policy for breaking changes.

bogdandrutu: What would we do for `ExtensionFactory.CreateExtension`? Cannot return `TempExtension`.

dmitryax: Yes, you're right, I missed that part. It'll still be a breaking change in any way. 

Probably going forward we can notify users about possible issues through a deprecation message for `component.Extension` to make sure that they don't use `component.Component` and `extension.Extension` interchangeably.

bogdandrutu: Will do that.

.chloggen/draftminhost.yaml
```diff
@@ -0,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: component.Extension will be reverted to not be an alias. Change your `component.Host` implementation.
```
dmitryax, line 8:
```suggestion
note: `component.Extension` is temporarily set to be an alias of `component.Component` which will be reverted once it's moved to the `extension` package. Change your `component.Host.GetExtensions()` implementation to return `map[ID]component.Component` instead of `map[ID]component.Extension`
```

dmitryax (approved):
Could you also please provide some reasoning behind this change in the PR description

PR #6560, opened 2022-11-15
Introduce component status reporting

tigrannajaryan:
This is an alternate to https://github.com/open-telemetry/opentelemetry-collector/pull/6550

## Summary of changes

- Add component status concept. Components can report their status via Host.ReportComponentStatus(). Interested extensions can watch status events if they implement StatusWatcher interface. This is similar to how PipelineWatcher works today.
- Deprecate Host.ReportFatalError() in favour of Host.ReportComponentStatus().

## TODO after this is merged

- healthcheck extension must implement StatusWatcher.
- Replace all ReportFatalError() calls by ReportComponentStatus() calls in core and contrib.

## Open Questions

- StatusWatchers need to be able to tell if all current components are healthy. It is assumed that the listeners need to maintain a map of components and track the status of each component. This works only if we assume that the set of components cannot change during the lifetime of the listener. This assumption is true today but can change in the future if we introduce partial pipeline restarts where only modified/added/removed components are recreated (this will break listener's assumption and the map will become invalid). Should we instead keep track of this entire status map in the Host and broadcast the entire status to the listeners as a whole instead of (or in addition to) individual component events?

We are currently preparing for the upcoming 1.0 GA release. Pull requests that are not aligned with
the current roadmap https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/roadmap.md
and are not aimed at stabilizing and preparing the Collector for the release will not be accepted.


component/status.go
```diff
@@ -0,0 +1,106 @@
...
+// limitations under the License.
+
+package component
+
+import (
+	"errors"
+	"time"
+)
+
+type Status int
```
bogdandrutu, line 22: nit: use int32 to guarantee size independent of the platform.

component/status.go
```diff
@@ -0,0 +1,106 @@
...
+const (
+	StatusOK Status = iota
+	StatusError
+)
+
+// StatusSource component that reports a status about itself.
+// The implementation of this interface must be comparable to be useful as a map key.
+type StatusSource interface {
+	ID() ID
+}
```
bogdandrutu, line 33: Why do you need this interface at all? Simply pass the ID?

component/status.go
```diff
@@ -0,0 +1,106 @@
...
+	if ev.timestamp.IsZero() {
+		ev.timestamp = time.Now()
+	}
+
+	return &ev, nil
+}
+
+// StatusWatcher is an extra interface for Extension hosted by the OpenTelemetry
+// Collector that is to be implemented by extensions interested in changes to component
+// status.
```
bogdandrutu, line 102: Any Component or just extensions?

component/status.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Timestamp returns the time of the event.
+func (ev *StatusEvent) Timestamp() time.Time {
+	return ev.timestamp
+}
+
+// Err returns the error associated with the ComponentEvent.
+func (ev *StatusEvent) Err() error {
+	return ev.err
+}
```
bogdandrutu, line 53: Can this be a "Description" instead?

component/status.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (ev *StatusEvent) Status() Status {
+	return ev.status
+}
+
+// Timestamp returns the time of the event.
+func (ev *StatusEvent) Timestamp() time.Time {
+	return ev.timestamp
+}
```
bogdandrutu, line 48: Do you have use-case for this? If yes, do we need to allow this to be configured or we always do Time.Now in the `NewStatusEvent`.

bogdandrutu (commented):
I like this better, much simpler change and cleaner design (at least for me).

component/status.go
tigrannajaryan: We need a key that uniquely identifies the component. ID is not a unique identification. You may have a processor and a receiver that have the same ID (we don't enforce that receivers and processors don't have the same type string - although in reality they are all unique in core and contrib).

Why it needs to be unique? So that the StatusWatcher can maintain a `map[StatusSource]Status`.

Alternatively we should enforce uniqueness of ID across all component types.

component/status.go
tigrannajaryan: Only extensions can watch. Any component can be a source.

We can add support for any component to watch but that requires more work and I don't know if we need it.

component/status.go
bogdandrutu: Interesting point... I understand now the need, let me think a bit more, probably better to have this as a struct if we can define it.

tigrannajaryan: Yet another alternate is to require the Host to track the aggregate component status. In that case StatusWatcher will only be notified of that single aggregate bool status. But in that case we have to hard-code the aggregation logic in the Host and lose flexibility in allowing the extension to decide what to do with individual component statuses.

component/status.go
tigrannajaryan: No, I think this can be deleted for now.

component/status.go
tigrannajaryan: Why do we need to guarantee the size?

component/status.go
tigrannajaryan: Do you mean to rename to `func (ev *StatusEvent) Description() error`?

tigrannajaryan:
> I like this better, much simpler change and cleaner design (at least for me).

Yeah, I also like this one more.

tigrannajaryan:
@open-telemetry/collector-approvers please take a look at this alternate.

component/status.go
bogdandrutu: For the "global component ID", I think we need something like:

```
package component

type GlobalID struct {
  ID ID
  Kind Kind
  PipelineID ID  // Not empty only if the Kind is Processor
}
```

component/status.go
bogdandrutu: +1

component/status.go
bogdandrutu: `func (ev *StatusEvent) Description() string` not sure we need `error` as type, just a `Description` string. Idea being that `Description` is a bit more generic, can apply to things like `StatusRecovering` or other status type.

component/status.go
bogdandrutu: Is not about guaranteeing the size, is about best practice, we definitely don't need this to be int64 on a 64bit arch. And in general we use int32 for enums, gives enough room to grow and saves memory.

component/status.go
tigrannajaryan: This also requires https://github.com/open-telemetry/opentelemetry-collector/pull/6540 otherwise you can still have duplicates even with GlobalID.

component/status.go
bogdandrutu: I know that it does not happen in this case of alignment situation, but in lots of cases this can be paired with another enum/int32 so may save memory there.

I think in vast majority of the cases struct members should be order by size and in that case the alignment rules should allow us to save that memory.

bogdandrutu: It is a best practice in general "." Not wast memory, and indeed you don't always save, but with `int` you always waste :)

component/status.go
bogdandrutu: Good point about that. Now I am more confused after I looked at your implementation https://github.com/open-telemetry/opentelemetry-collector/pull/6560/files#diff-bcb3fc758a54acf0dda21a0d9c79b6cf6e6adcf58d0a282c93be50553523b09dR41 and https://github.com/open-telemetry/opentelemetry-collector/pull/6560/files#diff-19abf8e6f60ec275051d1447d4d97f4c55b449192f2cb946e3e4ea44ff49fcc0R69

At least in the `statusReportingComponent` is not unique, so my suggestion at least improves that :))

component/status.go
tigrannajaryan: Changed to int32.

component/status.go
tigrannajaryan: I think `&statusReportingComponent{}` is unique. We create a new instance for every created component. It is not a value of `statusReportingComponent{}` struct, we create a pointer to a new struct every time.

component/status.go
tigrannajaryan: I deleted timestamp from everywhere.

component/status.go
tigrannajaryan: Added a test that verifies the uniqueness of `StatusSource`: https://github.com/tigrannajaryan/opentelemetry-collector/blob/2ac7d388cf548e538601796ec9fff25ae5eb6a78/service/service_test.go#L229

component/host.go
```diff
@@ -23,8 +23,14 @@ type Host interface {
 	//
 	// ReportFatalError should be called by the component anytime after Component.Start() ends and
 	// before Component.Shutdown() begins.
+	// Deprecated: [0.65.0] Use ReportComponentStatus instead (with an event of type status.ComponentError)
 	ReportFatalError(err error)
 
+	// ReportComponentStatus can be used by a component to communicate its status to the Host.
+	// The Host implementations may broadcast this information to interested parties via
+	// StatusWatcher interface.
+	ReportComponentStatus(event *StatusEvent)
```
tigrannajaryan, line 32: Addition of this method is a breaking change. See failure on Jaeger repo: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/3481700391/jobs/5823123326

dmitryax: I submitted [a contrib PR](https://github.com/open-telemetry/opentelemetry-collector/pull/6553) that should remove the dependency on Jaeger collector, so we can make contrib tests pass, but I believe we will still likely need to run through a breaking change.

I see a few other problems with the current state of this interface:
1. The set of methods doesn't seem to be complete and potentially can change after we release 1.0, e.g. we might want to introduce getters for receivers and processors, not only exporters and extensions.
2. Most of the methods start with `Get` which goes against Go recommendations, maybe we want to change them and remove `Get` prefix.

I believe we need to take the opportunity and make a bigger refactoring that would allow us to solve all the outlined problems at once. We discussed the problem with @bogdandrutu and came up with the following plan:
Keep only one "core" method `ReportComponentStatus` the `Host` interface and move all other "getter" methods to optional interfaces. The optional interfaces for exporter and extension getters can be moved to their own packages (exporter/extension), so we don't need to change returned values from `Extension|Exporter` to`Component` as done in https://github.com/open-telemetry/opentelemetry-collector/pull/6553 anymore.

I'll play with it, submit a draft PR, then we can combine it with this PR and try to migrate to the new interfaces as gracefully as feasible.

codeboten (commented):
@tigrannajaryan what's the status of this PR? Was there ever a workaround the breaking interface change or is it still blocked on that?

tigrannajaryan:
> @tigrannajaryan what's the status of this PR? Was there ever a workaround the breaking interface change or is it still blocked on that?

I am not working on this anymore. If anyone is willing to take over that would be great, otherwise I can close the PR.

jaronoff97:
@tigrannajaryan I'd be willing to take this on. What's left to do for this PR? Probably update from latest changes at the least haha

tigrannajaryan:
> @tigrannajaryan I'd be willing to take this on. What's left to do for this PR? Probably update from latest changes at the least haha

@jaronoff97 this is what I got stopped by, not sure what's a good solution: https://github.com/open-telemetry/opentelemetry-collector/pull/6560#discussion_r1025300926

component/host.go
djaglowski: I'm catching up on this PR. It looks like this is where it stalled out. Some thoughts on how to move forward:

> The set of methods doesn't seem to be complete and potentially can change after we release 1.0, e.g. we might want to introduce getters for receivers and processors, not only exporters and extensions.

I'm proposing that we deprecate `GetExporters` (https://github.com/open-telemetry/opentelemetry-collector/issues/7370). I don't believe there are valid reasons to need this function or others like it, such as `GetReceivers`, `GetProcessors`, etc.

> Most of the methods start with Get which goes against Go recommendations, maybe we want to change them and remove Get prefix.

Between renaming `GetExtensions` to `Extensions`, deprecating `ReportFatalError` and `GetExporters`, and adding `ReportComponentStatus`, it looks like we want quite a different interface. Hard to believe we will just live with this as is, so ideally we just need to identify the best migration option. It looks like we no longer need to coordinate with Jaeger but presumably this will affect other components defined elsewhere.

I also tried to identify a graceful option. The best I came up with is quite ugly, but would require no unannounced breaking changes does not require changes to any other interfaces (i.e. `component.Component`):

1. Add an optional interface so it is possible to use the new functions. I'll call it `component.Host2`.
2. Deprecate `ReportFatalError`, `GetExporters`, and `GetExtensions` to alert users of these specific functions. Recommend the ugly solution of asserting the optional interfaces (temporarily).
3. After some time Deprecate `component.Host` to alert all all affected. However, clearly indicate that this interface will be _changed_  to `component.Host`, rather than removed.
4. Change `component.Host` to match `component.Host2`. Undeprecate `component.Host` and deprecate `component.Host2`

jaronoff97:
(still want to work on this, was on vacation for quite some time)

tigrannajaryan:
Closing this since I don't work on it anymore.

PR #6562, opened 2022-11-15
[component helper] Adding lifecycle state management to *helper.

MovieStoreGuy:
**Description:** 

Extending the current helpers to adhere to the current component expectations

**Link to tracking Issue:** 
https://github.com/open-telemetry/opentelemetry-collector/issues/6507

**Testing:**

Haven't extended the tests yet, but my intention is to extend the lifecycle test to check if a component can:
- Allow for partial shutdown
- is restartable 

**Documentation:** 
Haven't added it yet, just wanting to get feedback to check the approach.

### DIFF
--- a/internal/lifecycle/state.go
+++ b/internal/lifecycle/state.go
@@ -1,0 +1,55 @@
+// Copyright  The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lifecycle // import "go.opentelemetry.io/collector/internal/lifecycle"
+
+import "go.uber.org/atomic"
+
+// Lifecycle represents the current
+// running state of the component
+// that is intended to be concurrency safe.
+type State interface {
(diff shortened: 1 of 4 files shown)


MovieStoreGuy:
The current components `Processor, Extension` don't have helpers at this point, is worth me adding it as part of this PR?

internal/lifecycle/state.go
```diff
@@ -0,0 +1,55 @@
...
+// that is intended to be concurrency safe.
+type State interface {
+	// Start atomically updates the state and returns true if
+	// the previous state was not running.
+	Start() (started bool)
+
+	// Stop atomically updates the state and returns true if
+	// the previous state was not shutdown.
+	Stop() (stopped bool)
+}
```
bogdandrutu, line 30:
```suggestion
// This will implement component.Component.
type Component struct {
	current *atomic.Int64
	component.StartFunc
	component.ShutdownFunc
}

func NewComponent(start component.StartFunc, shutdown component.ShutdownFunc) *Component {
	return &Component{
		current: atomic.NewInt64(shutdown),
		StartFunc: start,
		ShutdownFunc: shutdown
	}
}

func (c *Component) Start(ctx context.Context, host component.Host) error {
	if s.current.Swap(running) != shutdown {
		return nil
	}
	c.StartFunc.Start(ctx, host)
}

func (c *Component) Shutdown(ctx context.Context) error {
	if s.current.Swap(running) != running {
		return nil
	}
	c.ShutdownFunc.Shutdown(ctx)
}
```

bogdandrutu: Then in the exporterhelper can have 

```
type baseExporter struct {
	*lifecyle.Component
}
```

bogdandrutu:
> The current components Processor, Extension don't have helpers at this point, is worth me adding it as part of this PR?

https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/processorhelper/traces.go#L39

PR #6572, opened 2022-11-17
Deprecate Config.Validate() in favor of component.ValidateConfig

bogdandrutu:
This PR is "split" from https://github.com/open-telemetry/opentelemetry-collector/pull/6544, to avoid breaking changes and go via deprecation.

It is a bit hacky, but better than breaking everyone who uses this interface.

### DIFF
--- a/config/receiver.go
+++ b/config/receiver.go
@@ -45,7 +45,7 @@ func (rs *ReceiverSettings) SetIDName(idName string) {
 	rs.id = component.NewIDWithName(rs.id.Type(), idName)
 }
 
-// Deprecated: [v0.65.0] use component.ValidateConfig.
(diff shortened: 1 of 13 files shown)


config/receiver.go
```diff
@@ -45,7 +45,7 @@ func (rs *ReceiverSettings) SetIDName(idName string) {
 	rs.id = component.NewIDWithName(rs.id.Type(), idName)
 }
 
-// Validate validates the configuration and returns an error if invalid.
+// Deprecated: [v0.65.0] use component.ValidateConfig.
```
dmitryax, line 48: Maybe something like this?

```suggestion
// Deprecated: [v0.65.0] Not needed anymore since the Validate() will be moved from Config interface to the optional ConfigValidator interface.
```

codeboten (approved):

PR #6576, opened 2022-11-18
Add recursive validation to config

### LINKED ISSUE #4584: `exporterhelper` should reject a queue size of `0`
**Describe the bug**

`exporterhelper` allows for a `QueueSize` of `0`, which results in a permanently blocked exporter that drops all data (see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/6718#issuecomment-994715344).

**Steps to reproduce**

Create a pipeline with an exporter that exposes the `QueueSettings` struct and set the queue size to 0.

The Collector would fail to start and provide a descriptive error.

**What did you see instead?**

The Collector fails each time a payload gets to the exporter with a message similar to:

```
exporterhelper/queued_retry.go:99       Dropping data because sending_queue is full. Try increasing queue_size.   {"kind": "exporter", "name": "otlp", "dropped_items": 3}
```

**What version did you use?**
Version: (e.g., `v0.4.0`, `1eb551b`, etc)

Latest commit reproduces 81ab024486f067433d025fe0b25f8b532520ebe0

**What config did you use?**
Config: (e.g. the yaml config file)

<details>

<summary>Example config with contrib distro</summary>

```yaml
receivers:
  hostmetrics:
    scrapers:
      load:
    collection_interval: 2s

exporters:
  otlp:
    endpoint: 0.0.0.0:4317
    sending_queue:
      queue_size: 0

  
service:
  pipelines:
    metrics:
      receivers: [hostmetrics]
      exporters: [otlp]
```

</details>

**Environment**

not relevant

mattsains:
Closes #4584

**Description:** 
This PR, which is heavily based on @FreakyNobleGas 's work, allows for children of config structs to define their own Validate method. All nodes' Validate methods will be checked instead of just the root of the config element.

If a validation error is found within the config tree, it will be surfaced with a message like this:

> extension "ExtensionName" has invalid configuration in "ChildList[1].AnElementWithValidator": The error message returned by the validator within AnElementWithValidator

**Link to tracking Issue:** 
#4584

**Testing:** 
I added two new tests to check that children of config are validated. I ran these tests through `make all` in the root of the project.

I think the testing I've done is a bit weak just because of my unfamiliarity with the project, so I'm happy to do more testing if needed.

Open question: If a config struct has a Validate method implemented, but that validation succeeds, should we continue checking inside the children of that object, or should we consider the entire object "checked" by that implementation? Right now, I've chosen the former option.

### DIFF
--- a/service/config.go
+++ b/service/config.go
@@ -17,6 +17,7 @@ package service // import "go.opentelemetry.io/collector/service"
 import (
 	"errors"
 	"fmt"
+	"reflect"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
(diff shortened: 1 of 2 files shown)


bogdandrutu:
Please take a look at https://github.com/open-telemetry/opentelemetry-collector/pull/6545, where we want to add the validation.

mattsains:
Damn, that sucks, looks like my PR is useless then (I'll close it). @bogdandrutu is any help needed on the other PR? Otherwise I'll just find something else to do

Issue #6566, opened 2022-11-16
`Size()` for `pdata.ResourceLogs`, `pdata.InstrumentationLibraryLogs` and `pdata.LogRecords`


lprakashv:
**Is your feature request related to a problem? Please describe.**
My use case is that I want to split (or limit) my `*pdata.Logs` packet created for a list of log events if the size of `*pdata.Logs` exceeds some `maxBytes`. This is to avoid downstream otel-collector to get a very large size batch for a batch of fixed size log-event. This requirement comes due to the logs event size can var a lot.

As part of the existing system, we are building the packet from a list of events where we keep appending to `ResourceLogs`, `InstrumentationLibraryLogs` and `LogRecords`.

To achieve our goal we are planning to add the size calculation using the `LogSizer` implementation provided by the `LogsMarshaler`, this poses a problem that we will only be able to get the size for the `pdata.Logs` packet after adding/appending `ResourceLogs`/`LogRecords` etc. and see if it exceeds the `maxBytes`. But, in this case we would need to remove the last appended element in the Slices if the size does exceed.

A better approach would be if we are able to calculate the size (in bytes) for `LogRecord`, `ResouceLog` etc elements of the `pdata.Logs` data structure as it would save 2 operations : Adding and Removing, out of which, for removal we need to use the ` `RemoveIf` which will iterate all the elements in the Slice.

**Describe the solution you'd like**
If the `LogSizer` (or some other interface) provides API/method to get the size for other elements (LogRecords, ResourceLogs, InstrumentationLibraryLogs etc), this would help us optimize our solution. 

I can see that there is a `Size()` method defined for all the data structure mentioned above, example : https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/data/protogen/logs/v1/logs.pb.go#L863
We can just expose this via either the `LogSizer` interface or via an altogether a new interface.

**Describe alternatives you've considered**
We cannot access the internal methods to calculate the Size for internal data sctructures of `pdata.Logs`.

**Additional context**

PR #6597, opened 2022-11-22
Exposed Size, Marshal and Unmarshall operations for ResourceLogs and LogRecord

lprakashv:
**Description:** Added exposed Size() computation methods for ResourceLogs and LogRecord in the Sizer interface, consequently changes in Marshaller, Unmarshaller as well. Added unit tests also

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/6566

**Testing:** Only added the unit tests in `pdata/plog/pb_test.go`. The internal methods for size computation were already present and should work as expected.

**Documentation:** Haven't added it yet, just wanting to get feedback to check the approach.

### DIFF
--- a/pdata/internal/wrapper_logs.go
+++ b/pdata/internal/wrapper_logs.go
@@ -44,3 +44,27 @@ func LogsFromProto(orig otlplogs.LogsData) Logs {
 		ResourceLogs: orig.ResourceLogs,
 	}}
 }
+
+// ResourceLogsToProto internal helper to convert ResourceLogs to protobuf representation.
+func ResourceLogsToProto(rl ResourceLogs) otlplogs.ResourceLogs {
+	return *rl.orig
+}
+
+// ResourceLogsFromProto internal helper to convert protobuf representation to ResourceLogs.
+func ResourceLogsFromProto(orig otlplogs.ResourceLogs) ResourceLogs {
+	return ResourceLogs{
+		orig: &orig,
+	}
+}
+
+// LogRecordToProto internal helper to convert LogRecord to protobuf representation.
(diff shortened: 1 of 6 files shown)


lprakashv:
Hi @bogdandrutu I agree but this will help us solve one of our problem while creating `pdata.Logs` from a batch of internal log-events.

We want to split the batch of events based on the wire-size of the formed `pdata.Logs`, for example if my `maxBytes = 100`, we will create `pdata.Logs` packets out of our `var batch []Events` of size no more than 100 bytes.

Our algo currently looks like this:

```go
var batch []Event // some batch of events

logs := CreateNewPdataLogs()
for _, event := range events {
    if someCondition {
        resourceLogs := CreateAResourceLogs()
        AppendResourceLogs(&logs, resourceLogs)

        if logSizer.LogSize(logs) > maxBytes {
            RemoveLastAppendedResourceLog(&logs) // this will use `RemoveIf` operation on `ResourceLogsSlice`
            return logs
        }
    } else {
        logRecord := CreateLogRecord()
        AppendLogRecordAtCurrentResourceIndex(&logs, logRecord)

        if logSizer.LogSize(logs) > maxBytes {
           RemoveLastAppendedLogRecordAtResourceIndex(&logs) // this will use `RemoveIf` operation on `LogRecordSlice`
           return logs
        }
    }
}

return logs
```

But if we have `Size` API available for `ResourceLogs` and `LogRecord`, we can calculate the size upfront before appending either of them to `pdata.Logs`, so that there will be no need for iterating through `ResourceLogsSlice` or `LogRecordSlice` for removing the last element.

Issue #5653, opened 2022-07-08
[proposal] Add string type alias for sensitive fields


mx-psi:
**Is your feature request related to a problem? Please describe.**

There have been several asks for some way of printing the full Collector configuration (see #5223, #1791 and https://github.com/open-telemetry/opentelemetry-collector/issues/5615#issuecomment-1172609483). This will also be more important as the ability to merge multiple configuration files and use different configuration providers gains user adoption. To support this we need a way to censor sensitive information (tokens, passwords, API keys...) before printing the config.

In principle, we only need to censor strings (numbers, enums, booleans are generally not sensitive).

**Describe the solution you'd like**

Introduce a string type alias that must be used for sensitive fields:

```go
type OpaqueString string

var _ encoding.TextMarshaler = (*OpaqueString)(nil)

func (s OpaqueString) MarshalText() ([]byte, error) {
	return bytes.Repeat([]byte("*"), len(s)), nil
}
```

This censors fields when marshaling and is transparent when unmarshaling ([playground](https://go.dev/play/p/wILtOsL82SN)) while allowing to retrieve the value if casting to string.

Since it is an alias the casting is implicit in some cases ([playground](https://go.dev/play/p/XL9tba-Rfge)), therefore the breakage implied by changing a field from `string` to `OpaqueString` is reduced (it's still is a breaking change though).

**Describe alternatives you've considered**

Originally (see https://github.com/open-telemetry/opentelemetry-collector/issues/5223#issuecomment-1099957554) I considered doing this via a struct tag (`sensitive=true/false`). There are some tradeoffs between the `OpaqueString` approach when compared with using struct tags:

1. Struct tags can be applied to fields of any type instead of only strings (so you can censor integers, enums... without introducing new types).
2. `OpaqueString` allows for more granular types: for a field like `Headers`, we can change it to `map[string]OpaqueString`, signaling that while the header names are not sensitive, its values are. With struct tags we would have an all-or-nothing situation: we either censor all the headers information or none of it.
3. Adding a type alias implies _some_ breakage: while you can set `a.OpaqueField = <string>`, you can't pass `a.OpaqueField` to a `func f(string)` function without explicit casting. Struct tags do not imply a breaking change of any kind.

[...]

PR #6637, opened 2022-11-30
[confighttp] Change Headers field type to have opaque values

mx-psi:
**Description:**

Change `Headers` field on HTTP client config to have opaque values and transparent keys.

Depends on #6470

**Link to tracking Issue:** Updates #5653

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -17,7 +17,6 @@ package confighttp // import "go.opentelemetry.io/collector/config/confighttp"
 import (
 	"crypto/tls"
 	"errors"
-	"fmt"
 	"net"
 	"net/http"
 	"time"
@@ -55,13 +54,10 @@ type HTTPClientSettings struct {
 	// Timeout parameter configures `http.Client.Timeout`.
 	Timeout time.Duration `mapstructure:"timeout"`
 
-	// Deprecated: [v0.67.0] Use OpaqueHeaders instead.
-	Headers map[string]string `mapstructure:"-"`
-
 	// Additional headers attached to each HTTP request sent by the client.
 	// Existing header values are overwritten if collision happens.
 	// Header values are opaque since they may be sensitive.
-	OpaqueHeaders map[string]configopaque.String `mapstructure:"headers"`
+	Headers map[string]configopaque.String `mapstructure:"headers"`
 
 	// Custom Round Tripper to allow for individual components to intercept HTTP requests
 	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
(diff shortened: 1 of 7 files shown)


config/confighttp/confighttp.go
```diff
@@ -132,11 +138,28 @@ func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.
...
+		headers = hcs.Headers
+		settings.Logger.Warn("confighttp.HTTPClientSettings.Headers is deprecated in favor of confighttp.HTTPClientSettings.OpaqueHeaders")
+	}
+
+	clientTransport := (http.RoundTripper)(transport)
+	if len(headers) > 0 {
 		clientTransport = &headerRoundTripper{
 			transport: transport,
-			headers:   hcs.Headers,
+			headers:   headers,
```
jpkrohling, line 162: Would it make sense to convert the string headers to opaque headers here?

mx-psi: Sure, it doesn't hurt to use the opaque type there too :)

config/confighttp/confighttp.go
```diff
@@ -53,9 +55,13 @@ type HTTPClientSettings struct {
 	// Timeout parameter configures `http.Client.Timeout`.
 	Timeout time.Duration `mapstructure:"timeout"`
 
+	// Deprecated: [v0.67.0] Use OpaqueHeaders instead.
+	Headers map[string]string `mapstructure:"-"`
```
bogdandrutu, line 59: Implement "Unmarshal" func on the struct, and during that initialize Headers with the values from `OpaqueHeaders`? That way at least the "unmarshalling" is backwards compatible?

config/confighttp/confighttp.go
```diff
@@ -53,9 +55,13 @@ type HTTPClientSettings struct {
...
+
 	// Additional headers attached to each HTTP request sent by the client.
 	// Existing header values are overwritten if collision happens.
-	Headers map[string]string `mapstructure:"headers"`
+	// Header values are opaque since they may be sensitive.
+	OpaqueHeaders map[string]configopaque.String `mapstructure:"headers"`
```
bogdandrutu, line 64: Not convinced that I want to endup with this name. What do you think? Is there a way to use the previous name, maybe a few versions plan?

mx-psi: My idea was to follow this plan: https://github.com/open-telemetry/opentelemetry-collector/pull/6470#issuecomment-1332094426 with which we would end up back with `Headers`

config/confighttp/confighttp.go
mx-psi: > Implement "Unmarshal" func on the struct

Does this work with structs that are not component config structs? I wasn't aware, I can do that.

Note that with the current code unmarshaling is kept backwards compatible in the sense that for a given confmap we end up building the same client as before. But if we care about people accessing the individual fields we should do the unmarshal thing, yes, will fix that

bogdandrutu: Yes works, since 1-2 versions ago :)

config/confighttp/confighttp.go
bogdandrutu: I know that just doing `Headers map[string]configopaque.String` is a "backwards" incompatible change, but it is very limited incompatibility which I believe we can make an exception for to speed this up.

mx-psi: I am fine with that, but I don't think @jpkrohling is, so I am trying to reach a compromise here :)

config/confighttp/confighttp.go
mx-psi: I think it does not work for embedded structs? See #6668, will open issue if you confirm this is indeed a bug

bogdandrutu: This is not embedded, but that is a bug indeed. Do you plan to fix it?

config/confighttp/confighttp.go
bogdandrutu: Ping @jpkrohling

config/confighttp/confighttp.go
mx-psi: > This is not embedded, but that is a bug indeed

I ran into the issue because the client struct is embedded on the OTLP HTTP exporter https://github.com/open-telemetry/opentelemetry-collector/blob/fb0842a5a200ad1bc7475a4f69baa994310792a4/exporter/otlphttpexporter/config.go#L29 and some tests were failing

> Do you plan to fix it?

Not today, but I can do it next week

config/confighttp/confighttp.go
mx-psi: I think the usage outside of contrib is minimal and the error and fix is quite clear (just change the type), so I change to do it in a single step. We can consider case by case on other fields

mx-psi:
I changed this PR to do a breaking change, since the breakage seems limited enough to do in a single step. We will probably need to skip contrib tests and fix things over there in a version bump (I promise to do it ASAP once this is merged within my working hours :))

config/confighttp/confighttp.go
mx-psi: This no longer blocks this PR, so we can continue the discussion on #6671. It sounds like it's not easy to fix though (may need vendoring the Decode function?)

bogdandrutu (approved):

PR #6640, opened 2022-12-01
Prefix auth funcs/types with Server to be consistent with Client.

bogdandrutu:
No need to deprecate/break things since these new types were added in this release.

The reason for Option -> ServerOption is to be consistent with ClientOption.

The reason for AuthenticateFunc -> ServerAuthenticateFunc is because we may have the same Authenticate func on the client in the future and to avoid name conflicts.


dmitryax (approved):
LGTM

PR #6666, opened 2022-12-02
[chore] combine auth files, rename old names related to authenticator

bogdandrutu:


extension/auth/client.go
```diff
@@ -20,9 +20,21 @@ import (
 	"google.golang.org/grpc/credentials"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/extension"
 )
 
-var _ Client = (*defaultClient)(nil)
```
jpkrohling, line 25: Where did this go?

bogdandrutu: Not needed (in my opinion) because this is compile time check that defaultClient implements Client, but the same check (compe time) is in NewClient since we return defaultClient and return type is Client

jpkrohling (approved):

Issue #6660, opened 2022-12-01
[pdata] Reconsider [Map|...Slice].Sort methods 


dmitryax:
Currently we have two types of Sort method in pdata:
1. `Map.Sort() Map`
2. `...Slice.Sort(func(a, b ...) bool) ...Slice` e.g. `func (es SummaryDataPointSlice) Sort(less func(a, b SummaryDataPoint) bool) SummaryDataPointSlice`

We don't have any other methods which mutate and return the same objects in pdata. Documentation for (2) says

```golang
// Returns the same instance to allow nicer code like:
//
//	lessFunc := func(a, b SummaryDataPoint) bool {
//	  return a.Name() < b.Name() // choose any comparison here
//	}
//	assert.Equal(t, expected.Sort(lessFunc), actual.Sort(lessFunc))
```

but we don't have any usages like that for (2). All the usages of (2) are applied on the contrib tests and all of them don't use the returned object.

Contrib usages of (1) do rely on the returned object, but most of them are done in tests. Places where the returned object is used in production code usually use that object only and don't expect the original one to be mutated. This assumption was recently found in multiple exporters claimed as not mutating, and it caused several issues:
- https://github.com/open-telemetry/opentelemetry-collector/issues/6420
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16499

The suggestion is to change both types of the Sort method to not return any value. It'll complicate some tests relying on (1) but will make it clear that it's a mutating method to avoid the mentioned issues. We can also consider https://pkg.go.dev/sort#Slice as a reference that doesn't return any value. Later, maybe we can introduce some helpers for checking the Map equality without doing the sort. That will make the contrib tests cleaner.

Given that most of the usages are done in test code, believe we can make an exception here and introduce a breaking change just by changing the function signature to not return any value.

bogdandrutu:
+1

PR #6667, opened 2022-12-02
[pdata] Change [...Slice|Map].Sort methods to not return any value

dmitryax:
Returning the same object from the `Sort` methods leads to subtle bugs when non-mutating components use the returned object assuming it's a copy. Making the methods not return anything makes it clear that they are mutating.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/6660

### DIFF
--- a/pdata/pcommon/common.go
+++ b/pdata/pcommon/common.go
@@ -717,18 +717,14 @@ func (m Map) PutEmptySlice(k string) Slice {
 }
 
 // Sort sorts the entries in the Map so two instances can be compared.
-// Returns the same instance to allow nicer code like:
-//
-//	assert.EqualValues(t, expected.Sort(), actual.Sort())
 //
 // IMPORTANT: Sort mutates the data, if you call this function in a consumer,
 // the consumer must be configured that it mutates data.
-func (m Map) Sort() Map {
+func (m Map) Sort() {
 	// Intention is to move the nil values at the end.
 	sort.SliceStable(*m.getOrig(), func(i, j int) bool {
 		return (*m.getOrig())[i].Key < (*m.getOrig())[j].Key
 	})
-	return m
 }
 
 // Len returns the length of this map.
--- a/.chloggen/pdata-sort-no-return.yaml
+++ b/.chloggen/pdata-sort-no-return.yaml
@@ -1,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: pdata
+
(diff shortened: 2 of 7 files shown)


bogdandrutu (approved):

pdata/pcommon/common.go
```diff
@@ -717,18 +717,14 @@ func (m Map) PutEmptySlice(k string) Slice {
...
 
 // Sort sorts the entries in the Map so two instances can be compared.
-// Returns the same instance to allow nicer code like:
-//
-//	assert.EqualValues(t, expected.Sort(), actual.Sort())
 //
 // IMPORTANT: Sort mutates the data, if you call this function in a consumer,
 // the consumer must be configured that it mutates data.
-func (m Map) Sort() Map {
+func (m Map) Sort() {
```
codeboten, line 723: This should follow the normal deprecation guideline no?

bogdandrutu: Have you read the issue?

codeboten: > Given that most of the usages are done in test code, believe we can make an exception here and introduce a breaking change just by changing the function signature to not return any value.

I did, though the comment ^^ excludes any usage outside of the repos we know about.

I understand this is important to fix and don't want to block that, however it's unclear to me what differentiates this from other breaking changes that require a deprecation.

codeboten: There's still 2 releases this year :)

bogdandrutu: > There's still 2 releases this year :)

That's what I said, 1 month :))

bogdandrutu: Also, a "proper" deprecation is a 3 release step, since we want to get back to have Sort as name (SortWithoutReturn is added and Sort is deprecated; SortWithoutReturn is deprecated and Sort becomes no return; SortWithoutReturn is removed)

dmitryax: I think that making it the "right" way significantly complicates the process for downstream distributions. Most of the usages currently don't utilize the returned value (and none of ...Slice.Sort usages do that). So if we go with this change, It'll be mostly no-op or changes like https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16575. If we go with the 2-step deprecation approach, downstream distributions will need to replace _all_ the usages two times: Sort -> Sort<Something> -> Sort. I'm not against that, I just wanted to make sure we all agree on introducing this extra work to the API users and that it's worth avoiding the exception.

bogdandrutu: I am all in favor of the "short" path for this. But looks like @codeboten it is not, so unless he agrees we should do the "right" way.

codeboten (approved):

PR #6700, opened 2022-12-07
Connectors prototype v3

djaglowski:
Variation of #6372, with reduced connector types.

This simplifies code in the core connector packages, but pushes some complexity into individual implementations. In particular, connectors that can consume a signal more than one way must ensure that their constructors set up the proper consume function.


djaglowski:
Not stale. Waiting on #6867, #6764, and #6789 before rebasing.

bogdandrutu:
Closing since we are almost done in https://github.com/open-telemetry/opentelemetry-collector/pull/7045

PR #6701, opened 2022-12-07
[chore] update builder to use non deprecated API

bogdandrutu:


Makefile
```diff
@@ -185,6 +185,7 @@ otelcorecol:
 .PHONY: genotelcorecol
 genotelcorecol:
 	pushd cmd/builder/ && $(GOCMD) run ./ --skip-compilation --config ../otelcorecol/builder-config.yaml --output-path ../otelcorecol && popd
+	$(MAKE) -C cmd/otelcorecol fmt
```
codeboten, line 188: this seems unrelated to the change

bogdandrutu: It is :)... the builder does not generate the imports in order, or at least I don't know how to do that in the template. because of that lint fails, so I decided that this is a nice fix.

bogdandrutu: I can try to have:

```tmpl
	"go.opentelemetry.io/collector/exporter"
	{{- range .Exporters}}
	{{.Name}} "{{.Import}}"
	{{- end}}
	"go.opentelemetry.io/collector/extension"
	{{- range .Extensions}}
	{{.Name}} "{{.Import}}"
	{{- end}}
```

This will work for core since exporters are all part of this repo so will start with go.opentelemetry.io/collector/exporter/... but does not work for contrib for example.

codeboten (approved):

PR #6718, opened 2022-12-09
Remove deprecated comonent.Config.[ID|SetIDName]; Deprecate config.*Settings

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-collector/issues/4714


component/config.go
```diff
@@ -29,15 +29,7 @@ import (
...
-type Config interface {
-	// Deprecated: [v0.67.0] use Settings.ID.
-	ID() ID
-
-	// Deprecated: [v0.67.0] use Settings.ID.
-	SetIDName(idName string)
-
-	privateConfig()
-}
+type Config interface{}
```
dmitryax, line 32: > // Config defines the configuration for a component.Component.

This description is probably not correct anymore, it can be applied to any configuration part now, right?

Do you have any further plans for this type? Do we even need it if it's just `interface{}` now?

Also `UnmarshalConfig` doesn't seems to be applicable to `compoonent` module anymore. Do you think we should keep it here?

dmitryax (commented):
Do we have an issue for the `config` package refactoring? It would be nice to provide reasoning for these changes. I can help with that if we don't have it.

bogdandrutu:
> Do we have an issue for the config package refactoring? It would be nice to provide reasoning for these changes. I can help with that if we don't have it.

https://github.com/open-telemetry/opentelemetry-collector/issues/4714

.chloggen/rmcfgid-1.yaml
```diff
@@ -0,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove deprecated `comonent.Config.[ID|SetIDName]`.
```
mx-psi, line 8:
```suggestion
note: Remove deprecated `component.Config.[ID|SetIDName]`.
```

component/config.go
bogdandrutu: > This description is probably not correct anymore, it can be applied to any configuration part now, not component.Component only, right?

Not sure I understand. The "component.Factory.CreateDefaultConfig" returns this object. If others are using it somewhere else that is their problem... What are you looking for here?

> Do you have any further plans for this type? Do we even need it if it's just interface{} now?

Good discussion if we need it. We can think in a future PR about that. 

> Also UnmarshalConfig doesn't seems to be applicable to compoonent module anymore. Do you think we should keep it here?

Same as previous, we can followup with more changes, for the moment the purpose of https://github.com/open-telemetry/opentelemetry-collector/issues/4714 was to make sure we don't mutate the `Config` and that becomes an opaque object from the core collector perspective.

dmitryax: > Not sure I understand. The "component.Factory.CreateDefaultConfig" returns this object. If others are using it somewhere else that is their problem... What are you looking for here?

Makes sense, so we have the Config still mostly applied to Component even if it can be used with any other sub-configuration. That's what I was looking for

dmitryax (approved):

PR #6731, opened 2022-12-12
[memory limiter processor] Refactor internal structure to improve testability

MovieStoreGuy:
**Description:** 

This refactor includes several major improvements:
- chore: Mocked memory with expectations to allow for improved validation on tests
- bugfix: ensures the background task is correctly shutdown since `ticker.C` doesn't actually close
- enhancement:  moves configuration checking into `Config.Validate()` so it doesn't require the pipeline to start
- chore: using virtual time to validate scheduler and expected GC calls (soft vs hard)
- chore: split memory observer and memory limiter from each other to improve maintainability 

**Link to tracking Issue:** 
NA

**Testing:** 

I've updated the testing throughout.

**Documentation:** 

There is no user facing changes present by this, so none are required.

### DIFF
--- a/processor/memorylimiterprocessor/config.go
+++ b/processor/memorylimiterprocessor/config.go
@@ -51,9 +51,34 @@ type Config struct {
 	MemorySpikePercentage uint32 `mapstructure:"spike_limit_percentage"`
 }
 
-var _ component.Config = (*Config)(nil)
+var (
+	_ component.Config = (*Config)(nil)
+)
 
 // Validate checks if the processor configuration is valid
 func (cfg *Config) Validate() error {
+	if cfg.CheckInterval <= 0 {
+		return errCheckIntervalOutOfRange
+	}
+	switch {
+	case cfg.MemoryLimitMiB != 0:
+		if cfg.MemorySpikeLimitMiB > cfg.MemoryLimitMiB {
+			return errMemSpikeLimitOutOfRange
+		}
+	case cfg.MemoryLimitPercentage != 0:
+		if cfg.MemorySpikePercentage > cfg.MemoryLimitPercentage {
+			return errPercentageLimitOutOfRange
+		}
+		for _, value := range []uint32{
+			cfg.MemoryLimitPercentage,
+			cfg.MemorySpikePercentage,
+		} {
+			if value <= 0 || value > 100 {
+				return errPercentageLimitOutOfRange
+			}
+		}
+	default:
+		return errLimitOutOfRange
+	}
 	return nil
 }
--- a/processor/memorylimiterprocessor/factory_test.go
+++ b/processor/memorylimiterprocessor/factory_test.go
@@ -24,6 +24,7 @@ import (
 
 	"go.opentelemetry.io/collector/component/componenttest"
 	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/processor/processortest"
 )
 
 func TestCreateDefaultConfig(t *testing.T) {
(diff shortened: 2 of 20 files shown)


processor/memorylimiterprocessor/factory_test.go
```diff
@@ -42,51 +41,39 @@ func TestCreateProcessor(t *testing.T) {
...
-	assert.Error(t, err, "created processor with invalid settings")
-
-	mp, err := factory.CreateMetricsProcessor(context.Background(), processortest.NewNopCreateSettings(), cfg, consumertest.NewNop())
-	assert.Nil(t, mp)
-	assert.Error(t, err, "created processor with invalid settings")
-
-	lp, err := factory.CreateLogsProcessor(context.Background(), processortest.NewNopCreateSettings(), cfg, consumertest.NewNop())
-	assert.Nil(t, lp)
-	assert.Error(t, err, "created processor with invalid settings")
+	if conf, ok := cfg.(interface{ Validate() error }); ok {
```
MovieStoreGuy, line 44: Not exactly sure what happened to the validatable config, if someone could link me the actual interface I would appreciate it.

processor/memorylimiterprocessor/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/stretchr/testify v1.8.1
+	github.com/tilinna/clock v1.1.0
```
bogdandrutu, line 7: Let's avoid non standard dependencies (with super limited usage/contributions) like github.com/tilinna/clock

MovieStoreGuy: Would github.com/benbjohnson/clock be an reasonable alternative?

The reason why I've used this original package is that it includes support for:
- Setting/Getting clock via context (Useful for not breaking interface abstractions)
- Includes a check to see how many active timers / tickers.

We've used this rather heavily in [gostatsd](https://github.com/atlassian/gostatsd) and some other internal projects, so I can at least speak to its production readiness.

processor/memorylimiterprocessor/config.go
```diff
@@ -51,9 +51,34 @@ type Config struct {
 	MemorySpikePercentage uint32 `mapstructure:"spike_limit_percentage"`
 }
 
-var _ component.Config = (*Config)(nil)
+var (
+	_ component.Config = (*Config)(nil)
+)
 
 // Validate checks if the processor configuration is valid
 func (cfg *Config) Validate() error {
```
mx-psi, line 59: Were all this conditions being checked at runtime, or is this a breaking change?

processor/memorylimiterprocessor/config.go
```diff
@@ -51,9 +51,34 @@ type Config struct {
...
+	if cfg.CheckInterval <= 0 {
+		return errCheckIntervalOutOfRange
+	}
+	switch {
```
mx-psi, line 63: Should we error out also if the memory limit is set in two different ways (via MiB and percentage)?

processor/memorylimiterprocessor/config.go
MovieStoreGuy: Yeah, they were being checked within the `Start` function

processor/memorylimiterprocessor/config.go
MovieStoreGuy: Perhaps, I was keeping the existing logic but moving it to the validate.

processor/memorylimiterprocessor/config.go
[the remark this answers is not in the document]
MovieStoreGuy: That is fair, there is a few changes that I am sceptical on in this and would be best to break them up.

MovieStoreGuy:
@mx-psi I will close up this PR but reference this in future PRs so there is context.

processor/memorylimiterprocessor/config.go
mx-psi: Okay, that's fair, since it would be a breaking change we can just warn for now

PR #6751, opened 2022-12-12
[chore] dependabot pr workflow needs workflow token

codeboten:
This is the same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16857

Signed-off-by: Alex Boten <aboten@lightstep.com>


bogdandrutu:
Are you sure about this change? I just ran the action https://github.com/open-telemetry/opentelemetry-collector/actions/runs/3677937016

codeboten:
As per my comment in the other PR, the permissions issue has resolved itself...

PR #6764, opened 2022-12-12
Extract service.pipelines interface

djaglowski:
Subset of #6700

### DIFF
--- a/service/graph.go
+++ b/service/graph.go
@@ -1,0 +1,45 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package service // import "go.opentelemetry.io/collector/service"
+
+import (
+	"context"
+	"net/http"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+var _ pipelines = (*pipelinesGraph)(nil)
+
+type pipelinesGraph struct{}
+
+func (g *pipelinesGraph) StartAll(ctx context.Context, host component.Host) error {
+	// TODO actual implementation
+	return nil
+}
+
+func (g *pipelinesGraph) ShutdownAll(ctx context.Context) error {
+	// TODO actual implementation
+	return nil
+}
+
+func (g *pipelinesGraph) GetExporters() map[component.DataType]map[component.ID]component.Component {
+	// TODO actual implementation
+	return make(map[component.DataType]map[component.ID]component.Component)
+}
+
+func (g *pipelinesGraph) HandleZPages(w http.ResponseWriter, r *http.Request) {
+	// TODO actual implementation
+}
--- a/service/pipelines_test.go
+++ b/service/pipelines_test.go
@@ -181,7 +181,7 @@ func TestBuildPipelines(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			// Build the pipeline
-			pipelines, err := buildPipelines(context.Background(), pipelinesSettings{
+			pips, err := buildPipelines(context.Background(), pipelinesSettings{
 				Telemetry: componenttest.NewNopTelemetrySettings(),
 				BuildInfo: component.NewDefaultBuildInfo(),
 				Receivers: receiver.NewBuilder(
@@ -211,7 +211,9 @@ func TestBuildPipelines(t *testing.T) {
 				PipelineConfigs: test.pipelineConfigs,
 			})
 			assert.NoError(t, err)
+			assert.IsType(t, &builtPipelines{}, pips)
 
+			pipelines := pips.(*builtPipelines)
 			assert.NoError(t, pipelines.StartAll(context.Background(), componenttest.NewNopHost()))
 
 			for dt, pipeline := range test.pipelineConfigs {
--- a/service/host.go
+++ b/service/host.go
@@ -36,7 +36,7 @@ type serviceHost struct {
 
 	buildInfo component.BuildInfo
 
-	pipelines         *builtPipelines
+	pipelines
 	serviceExtensions *extensions.Extensions
 }
 
--- a/service/pipelines.go
+++ b/service/pipelines.go
@@ -23,6 +23,7 @@ import (
 	"go.uber.org/multierr"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/connector"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/processor"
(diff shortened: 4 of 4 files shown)


djaglowski:
cc @bogdandrutu

djaglowski:
This PR establishes the "fork" between the current way of managing pipelines and the way of doing it via a graph that includes connectors. At the appropriate time, we can use a feature gate to toggle between the two implementations.

.chloggen/component-nodes.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Extract service.pipelines interface, add skeleton graph implementation
+
+# One or more tracking issues or pull requests related to the change
+issues: [6764]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
bogdandrutu, line 16: I don't think we need changelog entries for this internal things that do not touch the logic. It will make the CHANGELOG very hard to digest since this is not important for anyone using the core collector artifacts.

bogdandrutu: cc @codeboten @dmitryax

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	go.uber.org/zap v1.24.0
 	golang.org/x/net v0.0.0-20220805013720-a33c5aa5df48
 	golang.org/x/sys v0.3.0
+	gonum.org/v1/gonum v0.12.0
```
bogdandrutu, line 41: Are we ok this this dependency? I never heard of it, so curios if it is a super used library...

service/graph.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph/topo"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+var _ pipelines = (*serviceGraph)(nil)
+
+type serviceGraph struct {
```
bogdandrutu, line 29: bad name for a type in the service package. maybe consider `pipelinesGraph`

service/graph.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// All component instances represented as nodes, with directed edges indicating data flow.
+	componentGraph *simple.DirectedGraph
+}
+
+func (g *serviceGraph) StartAll(ctx context.Context, host component.Host) error {
+	nodes, err := topo.Sort(g.componentGraph)
```
bogdandrutu, line 36: Should we do this in the "constructor" to avoid doing it during start and shutdown?

service/graph.go
```diff
@@ -0,0 +1,80 @@
...
+			continue
+		}
+		if compErr := comp.Shutdown(ctx); compErr != nil {
+			return compErr
+		}
+	}
+	return nil
+}
+
+func (g *serviceGraph) GetExporters() map[component.DataType]map[component.ID]component.Component {
```
bogdandrutu, line 73: Should this return Connectors as well when used in place of exporters? Just curious of this, not to block this PR, but would like to know the plan.

service/pipelines_test.go
```diff
@@ -182,7 +182,7 @@ func TestBuildPipelines(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			// Build the pipeline
-			pipelines, err := buildPipelines(context.Background(), pipelinesSettings{
+			pipelinesInterface, err := buildPipelines(context.Background(), pipelinesSettings{
```
bogdandrutu, line 185: s/pipelinesInterface/pips

service/pipelines_test.go
```diff
@@ -210,6 +210,9 @@ func TestBuildPipelines(t *testing.T) {
 			})
 			assert.NoError(t, err)
 
+			pipelines, ok := pipelinesInterface.(*builtPipelines)
+			require.True(t, ok)
```
bogdandrutu, line 214: assert.IsType(t, &builtPipelines{}, pipelinesInterface)

Reads easier what you are checking.

go.mod
djaglowski: It's part of a [larger suite of mathematical packages](https://www.gonum.org/) which seems to be quite professionally maintained, and appears to have a pretty good number of users (6.2k stars).

service/graph.go
djaglowski: We could but in my opinion I prefer this way because then we have a single source of truth, instead of maintaining a separate representation. We have a handle to the graph and a very easy way to sort it on demand. Hypothetically, this design is more resilient in case we later support dynamically updating the graph.

service/graph.go
djaglowski: At a high level, I think this function only exists for the sake of "processors that should be connectors", so my thinking is that we should be looking to deprecate this eventually (if at all possible), rather than make it more useful. 

More concretely, there is a very good reason not to include connectors here - that it would allow users to create cycles in the graph. By including only exporters here, components that use this function can only route to exporters, and therefore the signals are immediately offloaded.

.chloggen/component-nodes.yaml
djaglowski: Removed

service/graph.go
djaglowski: Renamed to `pipelinesGraph`

service/pipelines_test.go
djaglowski: Fixed

service/graph.go
bogdandrutu: > More concretely, there is a very good reason not to include connectors here - that it would allow users to create cycles in the graph. By including only exporters here, components that use this function can only route to exporters, and therefore the signals are immediately offloaded.

I understand this, thanks.

> processors that should be connectors

I don't think you can remove "routing/loadbalancing" processor with this.

djaglowski: > I don't think you can remove "routing/loadbalancing" processor with this.

I've been working through what it would take. Proof of concept for `routingconnector` is [here](https://github.com/djaglowski/opentelemetry-collector/pull/73). Limitations and caveats apply, but I think it is possible if we want it.

bogdandrutu:
Need rebase.

djaglowski:
I've rebased, but this basically required adding `connector.Builder`, which I've now spun off into #6867. Also added tests for other component builders in #6862.

gbbr:
Hi Daniel & Bogdan! It would be very nice if we could add descriptions to these PRs

djaglowski:
This is rebased onto https://github.com/open-telemetry/opentelemetry-collector/pull/6867 and ready to go otherwise.

TylerHelmuth (approved):
LGTM

gbbr:
Hi Daniel! @djaglowski 

Since this is an open-source project, it would be nice if the community would be welcoming to all contributors, despite whether we agree with them or not. For this reason, I'd appreciate it if my comments would not be ignored ([example 1](https://github.com/open-telemetry/opentelemetry-collector/pull/6764#issuecomment-1369736331), [example 2](https://github.com/open-telemetry/opentelemetry-collector/issues/2336#issuecomment-1219109643)).

I don't expect you to agree with them, nor to follow them, nor to take any action on them. But I do expect you to acknowledge them because I did spend time to try and contribute and my intention is to _add_ value, not to _block_ or take away.

jpkrohling (commented):
Looks promising, but what's up with the tests? It would be nice to have a few tests demonstrating the graph. For instance, in the look from the last to first after sorting the graph, I assume the exporters are all located towards the end of the slice, with processors next, and receivers first. This is something that could be exemplified (and proven) in the tests.

djaglowski:
Yeah, test coverage is poor on this PR. I think things are covered quite thoroughly on #6700, but some lower level tests would be better here. For now, what do you think about me removing the implementations of `Start` and `Shutdown` in this PR and then adding tests alongside the implementations later?

djaglowski:
I've removed the implementations entirely, so now this is just adding the interface and a stub for the graph-based implementation. @open-telemetry/collector-approvers, please take a look.

service/graph.go
```diff
@@ -0,0 +1,45 @@
...
+	return nil
+}
+
+func (g *pipelinesGraph) ShutdownAll(ctx context.Context) error {
+	// TODO actual implemenation
+	return nil
+}
+
+func (g *pipelinesGraph) GetExporters() map[component.DataType]map[component.ID]component.Component {
+	// TODO actual implemenation
```
jpkrohling, line 39: If you are changing this PR for any other reason, you could fix this typo :-)

```suggestion
	// TODO actual implementation
```

jpkrohling (approved):

djaglowski:
@open-telemetry/collector-maintainers, I believe this is ready to merge.

jpkrohling (approved):

PR #6785, opened 2022-12-13
Add count connector

djaglowski:


bogdandrutu:
Please help me understand why do we need this?

djaglowski:
The reason I propose it now is that it is the simplest example I can imagine for a cross-signal connector. It seems worth having one of these to tinker with while we work through the feature set.

Strictly speaking, it is not necessary. In the longer term such a connector would likely belong in contrib and would have more sophisticated features such as the ability to count based on one or more conditions. (e.g. count error logs, count logs that have a particular attribute, count spans from a particular service, etc)

djaglowski:
I'm fine with holding this back for now and proposing it to contrib as soon as it could be used there.

Issue #6688, opened 2022-12-06
pdata Map.Sort() leaks underlying storage behavior and prevents future optimizations


tigrannajaryan:
The Map.Sort() contract is that it makes the data comparable using reflection-based equality methods, primarily used for testing.

Unfortunately the Sort() method is abused in many places with the expectation that it impacts the behavior of a subsequent Range() operation. See for example https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1ad54c003dcdd21d7f0bf44a67adf2862ddb9b49/processor/cumulativetodeltaprocessor/internal/tracking/identity.go#L45 or https://github.com/influxdata/influxdb-observability/blob/694f0a4cf4906008a896b7ac8a154fa033bf7b89/influx2otel/metrics_telegraf_prometheus_v2.go#L65

The pattern Sort/Range makes an assumption that the Range honours the sorted state. This leaks the implementation and can prevent future optimizations (for example we may want to make the map storage an open addressing hash map instead of unsorted vector).

I suggest to deprecate the Sort function and replace it with:
- IsEqual(map1, map2) or map1.IsEqual(map2)
- Optionally also add Map.SortedRange or Map.OrderedRange if think this is important. Or, we don't provide this and whoever needs an ordered range can do a Range() and copy from the map to a temp slice and sort the slice.

cc @open-telemetry/collector-maintainers

dmitryax:
We discussed that during the SIG call and agreed that is an important issue to fix before releasing pdata 1.0, but we will release 1.0.0-rc1 as planned and do it after

dmitryax:
For the (1) use case, where map equality check is needed in tests, we can use `AsRaw()` for direct comparison. But, sometimes users need to compare pdata instances that include several maps and slices in it. Currently, Sort() can help with that as done in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/551e5c97b32588c33a3852f2635052813e727c2e/pkg/stanza/adapter/frompdataconverter_test.go#L116. Removing it makes the workaround much more complicated. And I think there can be use cases like this outside of contrib.

There is a great module for this particular need: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/scrapertest. It should be expanded to work with traces and logs. And I think it's a good candidate to be moved to pdata as part of `pdata/pmetric/pmetrictest`, `pdata/ptrace/ptracetest` etc. It should make writing test much easier for pdata users.

[...]

PR #6786, opened 2022-12-13
[pdata] Add pcommon.Map.Equal method

dmitryax:
It's one of the measures towards deprecation of pcommon.Map.Sort

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/6688

Performance of Value.Equal with a map isn't affected.

Benchmark results on:
goos: darwin
goarch: arm64

Before:

```
BenchmarkValueEqual-10    	 7666512	       153.0 ns/op	       0 B/op	       0 allocs/op
```

After:
```
BenchmarkValueEqual-10    	 7906196	       149.6 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/.chloggen/map-equal.yaml
+++ b/.chloggen/map-equal.yaml
@@ -1,0 +1,11 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: pdata
+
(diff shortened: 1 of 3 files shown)


djaglowski (approved):

dmitryax:
We discussed this PR with @bogdandrutu and decided not to introduce any new API in pdata. Instead, we improve `consumertest` in contrib and make sure it's used by more test cases that need such functionality. Potentially that can be added to pdata as tests packages. We will also deprecate Value.Equal to be consistent.

PR #6789, opened 2022-12-13
Add support for connector configurations

djaglowski:
Subset of #6700

### DIFF
--- a/cmd/builder/internal/config/default.yaml
+++ b/cmd/builder/internal/config/default.yaml
@@ -17,4 +17,5 @@ extensions:
 processors:
   - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.69.1
   - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.69.1
-
+connectors:
+  - gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.69.0
--- a/cmd/builder/internal/config/default.yaml
+++ b/cmd/builder/internal/config/default.yaml
@@ -18,4 +18,4 @@ processors:
   - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.69.1
   - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.69.1
 connectors:
-  - gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.69.0
+  - gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.69.1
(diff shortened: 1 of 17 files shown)


djaglowski:
Integration test failure addressed in #6797

djaglowski:
cc @bogdandrutu

otelcol/factories.go
```diff
@@ -16,8 +16,14 @@ package otelcol // import "go.opentelemetry.io/collector/otelcol"
...
 
 // Factories struct holds in a single type all component factories that
 // can be handled by the Config.
-type Factories = component.Factories // nolint:staticcheck
+type Factories struct {
+	component.Factories // nolint:staticcheck
+
+	// Connectors maps connector type names in the config to the respective factory.
+	Connectors map[component.Type]connector.Factory
+}
```
bogdandrutu, line 29: We may have to wait to do this until next week. now `otelcol.Factories` is a different type than `component.Factories` so this is a breaking change for `otelcol.CollectorSettings.Factories` element.

djaglowski: Alright, makes sense.

bogdandrutu: This is unblocked if you rebase

djaglowski: I've rebased to a tentative state, with the assumption that #6867 will provide the builder.

cmd/builder/internal/config/default.yaml
```diff
@@ -17,4 +17,6 @@ extensions:
 processors:
   - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.68.0
   - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.68.0
-
+connectors:
+  - import: go.opentelemetry.io/collector/connector/forwardconnector
+    gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.68.0
```
Aneurysm9, line 22: Is the `import` entry needed here?  If the package import path and module identifier are equal I think you only need `gomod`.

djaglowski: Good call. Fixed.

djaglowski:
Thanks @jpkrohling. I've added a changelog entry and updated the title.

cmd/otelcorecol/builder-config.yaml
```diff
@@ -17,12 +17,16 @@ extensions:
 processors:
   - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.69.0
   - gomod: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.69.0
+connectors:
+  - import: go.opentelemetry.io/collector/connector/forwardconnector
+    gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.69.0
```
TylerHelmuth, line 22: Can we use only gomod here as well?

TylerHelmuth (approved):

cmd/otelcorecol/builder-config.yaml
djaglowski: Good catch, thanks. Done.

kovrus (approved):

jpkrohling (approved):

PR #6792, opened 2022-12-13
Move Factories to otelcol, delete otelcoltest

### LINKED ISSUE #6766: Proper division between "service" & "otelcol" packages
In #6723, I attempted to move `component.Factories` and `componenttest.NopFactories` to `otelcol` and `otelcoltest`, respectively. However, I was not able to identify a clear separation point between `service` and `otelcol` packages. 

Additional context: https://github.com/open-telemetry/opentelemetry-collector/pull/6723#discussion_r1044763514

The immediate problems are:
1. Both `otelcol` and `service` contain many references to `Factories`.
2. `otelcol` contains many references to `service`. 

Therefore, `Factories` can exist in `service` without causing import cycles, but the opposite is not true. 

In order to move `Factories` to `otelcol`, we must determine which of the two points above should change. Should `otelcol` not depend on `service`, or should `service` not depend on `Factories`?

djaglowski:
Resolves #6766

`NopFactories` is moved to `otelcoltest`. However, `otelcol` has its own private implementation in order to avoid a dependency cycle.

### DIFF
--- a/.chloggen/mv-factories-otelcol.yaml
+++ b/.chloggen/mv-factories-otelcol.yaml
@@ -1,14 +1,14 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: deprecation
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
-component: component
+component: componenttest
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Deprecate component.Factories in favor of otelcol.Factories
+note: Move NopFactories to otelcoltest
 
 # One or more tracking issues or pull requests related to the change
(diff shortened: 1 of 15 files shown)


bogdandrutu (commented):
I was expecting some deletions in the service package of the things you just added.

djaglowski:
`service.Factories` and `servicetest.NopFactories` were (re)moved. What else did you expect to see removed from `service`?

bogdandrutu:
Where?

Screen Shot 2022-12-13 at 1 10 57 PM

bogdandrutu:
These are all the files in this PR (at least this is what I see)

Screen Shot 2022-12-13 at 1 13 27 PM

djaglowski:
These files were moved from `service` / `servicetest`.

image

bogdandrutu (approved):

PR #6798, opened 2022-12-14
[DRAFT] [pdata] Prototype immutable/mutable pdata

dmitryax:


pdata/pmetric/generated_metrics.go
```diff
@@ -35,19 +35,29 @@ import (
 // Important: zero-initialized instance is not valid for use.
 type ResourceMetricsSlice internal.ResourceMetricsSlice
 
+type MutableResourceMetricsSlice internal.MutableResourceMetricsSlice
```
bogdandrutu, line 38: Try this:

```
type ResourceMetricsSlice internal.ResourceMetricsSlice

// This way MutableResourceMetricsSlice inherits all the getters from the non mutable.
type MutableResourceMetricsSlice struct {
    ResourceMetricsSlice
}
```

bogdandrutu: Discussed offline that this has a downside because it allows users to convert between the two types and allowing to convert from immutable to mutable is not intended.

bogdandrutu:
Closing in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/7088? Feel free to re-open if you think we still need.

PR #6800, opened 2022-12-14
Remove unnecessary configunmarshaler

bogdandrutu:


otelcol/internal/configunmarshaler/processors.go
```diff
@@ -15,40 +15,37 @@
...
 	}
 
 	// Prepare resulting map.
-	p.procs = make(map[component.ID]component.Config)
+	p.cfgs = make(map[component.ID]component.Config)
 	// Iterate over processors and create a config for each.
 	for id, value := range rawProcs {
 		// Find processor factory based on "type" that we read from config source.
 		factory := p.factories[id.Type()]
 		if factory == nil {
```
bogdandrutu, line 47: I tried to use generics with `Configs[F component.Factory]` but I got an error at this line `mismatched types F and untyped nil`. If you have an idea how to fix that, will remove the need of `toFactoryMap` (small func anyway)

djaglowski: This should work: https://github.com/open-telemetry/opentelemetry-collector/pull/6801

PR #6803, opened 2022-12-14
Add receiver.Builder to help with creating components form a set of configs and factories

bogdandrutu:
The main idea is that we pass a `receiver.Builder` (a combo of configs and factories that can create Receivers to the Service) instead of configs and factories separately. This way we help other distributions to re-implement service/host.

### DIFF
--- a/receiver/receiver.go
+++ b/receiver/receiver.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 
+	"go.uber.org/zap"
+
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 )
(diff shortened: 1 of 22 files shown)


service/service_test.go
```diff
@@ -395,9 +395,10 @@ func assertZPages(t *testing.T, zpagesAddr string) {
 
 func newNopSettings() Settings {
 	return Settings{
-		BuildInfo:          component.NewDefaultBuildInfo(),
-		ReceiverFactories:  map[component.Type]receiver.Factory{"nop": receivertest.NewNopFactory()},
-		ReceiverConfigs:    map[component.ID]component.Config{component.NewID("nop"): receivertest.NewNopFactory().CreateDefaultConfig()},
+		BuildInfo: component.NewDefaultBuildInfo(),
+		ReceiverBuilder: receiver.NewBuilder(
+			map[component.ID]component.Config{component.NewID("nop"): receivertest.NewNopFactory().CreateDefaultConfig()},
+			map[component.Type]receiver.Factory{"nop": receivertest.NewNopFactory()}),
```
djaglowski, line 401: `receivertest.NewNopBuilder`?

djaglowski (commented):
Looks like a nice improvement

service/service_test.go
bogdandrutu: Done.

receiver/receiver.go
```diff
@@ -195,3 +197,77 @@ func MakeFactoryMap(factories ...Factory) (map[component.Type]Factory, error) {
 	}
 	return fMap, nil
 }
+
+// Builder receiver is a helper struct that given a set of Configs and Factories helps with creating receivers.
+type Builder struct {
+	cfgs      map[component.ID]component.Config
+	factories map[component.Type]Factory
```
dmitryax, line 204: Should disallow implementing the method to leave us room for improvement given that we have the `NewBuilder`? Otherwise, NewBuilder may be redundant?

dmitryax (commented):
I like the idea.

dmitryax (approved):

dmitryax:
And we will do that for other component types, right?

bogdandrutu:
@dmitryax @djaglowski I am not convinced about the name of the type (Builder), here are some ideas:

* Receivers
* Builder
* please add more ideas...

Issue #6771, opened 2022-12-13
Persistent queue can have data loss if last send was in error state before restarting/shutdown


splunkericl:
**Describe the bug**
Recently, a [new fix](https://github.com/open-telemetry/opentelemetry-collector/issues/6692) was added to prevent duplicate data by allowing writing to storage during shutdown. However, this has caused a side effect of data loss if the last send in the queue was in an error state. As `OnProcessingFinished` is always executed and the item is always removed from dispatched list even if the send was an error.

Examples of shutdown that can cause the problem:
- shut down would cause queue to stop looping processing of the item: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L438
- the error will propagate to the callback function but ignored: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L205
- OnProcessingFinished will be triggered:
  - persistent queue callback function: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/persistent_storage.go#L272
  - execution will remove the item from `currentlyDispatchedItems`
- After restarting, the item won't be processed again since it was considered dispatched 

**Steps to reproduce**
1. Create a pipeline with an exporter that is always failing to export
2. send a log entry to the pipeline
3. restart the pipeline with a new settings with an exporter that is able to send data

**What did you expect to see?**
After step 3 restart, it is expected the log is sent to the destination

**What did you see instead?**
The item was not seen in the exporter destination again. 

**What version did you use?**
Version: v0.67.0

**What config did you use?**
N/A

**Environment**
OS: mac
Compiler(if manually compiled): go 1.18

**Additional context**
A suggestion to fix the problem:
- Update `OnProcessingFinished` parameter to contain `error`
- pass in error from `consumerSender.send` to `OnProcessingFinished`
- Update persistent queue to not update storage if error is returned.

```
	qrs.queue.StartConsumers(qrs.cfg.NumConsumers, func(item internal.Request) {
		err := qrs.consumerSender.send(item)
		item.OnProcessingFinished(err)
	})
```

bogdandrutu:
cc @dmitryax since you approved that PR you own it :)

splunkericl:
I will be happy to make the changes as well if you guys don't have bandwidth :)

dmitryax:
@splunkericl Of course, please feel free to take it

[...]

PR #6815, opened 2022-12-17
Propagate shutdown error from retry sender to fix data loss problem in persistent storage

splunkericl:
Fix a data loss bug occuring in persistent queue when otel collector restart.
When otel collector restart, the retry_sender will mark a request as finished and cause persistent storage to remove it from currentlyDispatchedItems. This causes data loss.

This change propagates an interrupted error when shutdown is called. The error propagates to the consumer callback function and each consumer can determine action against the error. The persistent storage is updated to check and ensure those requests are not marked as finished.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/6771

**Testing:**
- Added unit test around retry sender and persistent storage

**Documentation:**
Updated documentation for release note


splunkericl:
cc @dmitryax @pmm-sumo @swiatekm-sumo

exporter/exporterhelper/queued_retry.go
```diff
@@ -435,7 +435,7 @@ func (rs *retrySender) send(req internal.Request) error {
 		case <-req.Context().Done():
 			return fmt.Errorf("Request is cancelled or timed out %w", err)
 		case <-rs.stopCh:
-			return fmt.Errorf("interrupted due to shutdown %w", err)
+			return internal.NewRetryInterruptedError(fmt.Errorf("interrupted due to shutdown %w", err))
```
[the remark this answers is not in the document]
splunkericl: If we try to requeue the data, there isn't guaranteed that the data is processed since the storage can be shutdown and the loop exits already. So the data can be requeued but not written as currentlyDispatched items in persistent storage, which still causes data loss. 

Another way we can do this is add some kinds of flushing in the persistent_queue and persistent_storage during `Stop` and make sure these queued data are processed(we can add them as currentlyDispatched items during flush). 

Do you think this way would be better?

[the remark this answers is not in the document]
swiatekm-sumo: I'm not sure I follow. When you stop the queue, it [waits](https://github.com/open-telemetry/opentelemetry-collector/blob/cde44b428672a228ef7ee72e24762ada21a0d66c/exporter/exporterhelper/internal/persistent_queue.go#L92) for all consumers to exit before stopping the storage, so we should always be able to requeue successfully. Also, when you say the loop exits, which loop are you talking about? Putting items in the queue is synchronous, it doesn't depend on any loops or channels.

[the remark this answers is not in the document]
splunkericl: The loop i was referring was the [loop](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/persistent_storage.go#L185) in persistentStorage that process next items. 

So double checking the code again, i think i might understood it differently earlier. Do you mean this?
- consumer stop would [wait](https://github.com/open-telemetry/opentelemetry-collector/blob/cde44b428672a228ef7ee72e24762ada21a0d66c/exporter/exporterhelper/internal/persistent_queue.go#L92) for all consumer to finish.
- for each consumer right before it shuts down, if we return temporary error, they would be requeued and [produced](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/persistent_queue.go#L83) in persistent storage.
- in persistent storage, two scenarios can happen:
  - the write index can be incremented so new items can be read after restart
  - or the loop can call `[getNextItem](https://github.com/open-telemetry/opentelemetry-collector/blob/cde44b428672a228ef7ee72e24762ada21a0d66c/exporter/exporterhelper/internal/persistent_storage.go#L185)` and item will be part of `currentlyDispatchedItems`
- Then after restart, the persistent storage is able to process these items

If I understand this correctly it seems cleaner than adding a new request level error.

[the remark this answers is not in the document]
swiatekm-sumo: Yes, that's my understanding of it as well. The loop in the persistent storage only really exists so consumers can wait on a channel. Technically, an item that was removed from the storage and put in the channel by the loop counts as `dispatched`, and will be treated the same as any other dispatched items at storage start: it's going to be requeued.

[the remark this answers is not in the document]
splunkericl: Thanks for the suggestion. I updated the PR and it looks much cleaner.

swiatekm-sumo (approved):
This looks good to me. I think you may want to rename `test_util.go` to `util_test.go` so coverage ignores it. Alternatively, move it back to `persistent_storage_test.go`.

exporter/exporterhelper/queued_retry.go
```diff
@@ -435,7 +435,7 @@ func (rs *retrySender) send(req internal.Request) error {
 		case <-req.Context().Done():
 			return fmt.Errorf("Request is cancelled or timed out %w", err)
```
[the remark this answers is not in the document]
splunkericl: yeah it is a small change that is related so i can add it also.

splunkericl:
> This looks good to me. I think you may want to rename `test_util.go` to `util_test.go` so coverage ignores it. Alternatively, move it back to `persistent_storage_test.go`.



> This looks good to me. I think you may want to rename `test_util.go` to `util_test.go` so coverage ignores it. Alternatively, move it back to `persistent_storage_test.go`.

The mockClient is being used in both `internal` package and `exporterhelper` packages. So this is in a shared util file. It can't be imported with `_test.go` file. I can try to exclude the file from coverage.

exporter/exporterhelper/queued_retry.go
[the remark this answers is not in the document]
splunkericl: After building the test I actually found another issue regarding the context expiration. So to make this change we actually need to fix this bug first: https://github.com/open-telemetry/opentelemetry-collector/issues/6878

I will leave the change out as the change can't be tested right now due to request context always lost in the process of serialization.

crobert-1 (approved):

exporter/exporterhelper/queued_retry_test.go
```diff
@@ -582,6 +583,67 @@ func TestQueuedRetryPersistenceEnabledStorageError(t *testing.T) {
...
+	// replace nextSender inside retrySender to always return error so it doesn't exit send loop
+	castedSender, ok := be.qrSender.consumerSender.(*retrySender)
+	require.True(t, ok, "consumerSender should be a retrySender type")
+	castedSender.nextSender = &ErrorRequestSender{
+		errToReturn: errors.New("some error"),
+	}
+
+	go func() {
+		require.NoError(t, be.sender.send(req))
+	}()
```
MovieStoreGuy, line 637: This could be potentially hazardous since if it tries to report an error once the test has finished, it can cause the test suite to panic. 

Ideally there should be a `sync.Waitgroup`, `chan` or `context` to help join the main test routine to spawned sub routines.

testutils/storage_client.go
```diff
@@ -0,0 +1,95 @@
...
+func NewMockStorageClient() storage.Client {
+	return &MockStorageClient{
+		st: map[string][]byte{},
+	}
+}
+
+// MockStorageClient is a mocked in memory storage client designed to be used in tests
+type MockStorageClient struct {
+	st           map[string][]byte
+	mux          sync.Mutex
```
MovieStoreGuy: If the `mutex` is protecting `st`, then I would suggest reordering these so that it is:

```golang
	mu          sync.Mutex
	st           map[string][]byte
```

Also, I wouldn't us `mux` since that is a common name for http routers and could be confusing, its typical shortening is `mu`

testutils/storage_client.go
```diff
@@ -0,0 +1,95 @@
...
+	defer m.mux.Unlock()
+
+	delete(m.st, s)
+	return nil
+}
+
+func (m *MockStorageClient) Close(_ context.Context) error {
+	m.closeCounter++
+	return nil
+}
```
MovieStoreGuy: Should this method be safe for concurrent access?

testutils/storage_client.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"go.opentelemetry.io/collector/extension/experimental/storage"
+)
+
+// NewMockStorageClient creates a new instance of MockStorageClient
+func NewMockStorageClient() storage.Client {
```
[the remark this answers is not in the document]
splunkericl: That makes sense. I will move it inside the internal package in exporterhelper.

Makefile.Common
```diff
@@ -11,14 +11,15 @@ IMPI=impi
 GOOS := $(shell $(GOCMD) env GOOS)
 GOARCH := $(shell $(GOCMD) env GOARCH)
 GH := $(shell which gh)
+COVERAGE_PKGS=$(shell go list ./... | grep -v testutils)
```
MovieStoreGuy, line 14: Not sure I agree with `testutils` being ignored here, I understand it is a test package and testing tests feels redundant however, they do server as excellent examples for those important and using the package.

Furthermore, it helps check if things change, would the interactions with the library change?

MovieStoreGuy (commented):
Just a few things from me.

Makefile.Common
splunkericl: When you mean things change as in things like interface change, etc? All the changes should really be reflected on the test that uses the library. So if things change, those tests should fail and if they don't the tests were written properly.

Unless this is a widely used test library across a lot of packages internally/externally I don't think it really makes sense to have tests on them.

dmitryax: I also don't see a reason to exclude it. If the package contains test helpers, it means that it must be 100% covered. Otherwise, some helpers are not being used

exporter/exporterhelper/internal/persistent_storage_test.go
```diff
@@ -16,17 +16,16 @@ package internal
...
 	"reflect"
-	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter/exporterhelper/internal/testutils"
```
dmitryax, line 28: Let's call it `testutil` since singular form for packages is usually recommended in go

```suggestion
	"go.opentelemetry.io/collector/exporter/exporterhelper/internal/testutil"
```

Makefile.Common
splunkericl: `MockStorageClient` implements the interface of `Client` and there are 5 methods inside the interface. It is possible the tests are not using all methods but implementation for all methods are required.

dmitryax: I believe it's ok to keep them uncovered if they are noop and added just to implement the interface

splunkericl: I tried it again and the the method is still flagged as not covered.

I notice another package `internal/testutil` with some other testutil codes that have unit test coverage. I will just add some basic test to make this easier. Also renaming the struct from `MockStorageClient` to `InMemoryStorageClient`

dmitryax: > Also renaming the struct from MockStorageClient to InMemoryStorageClient

Why? `InMemoryStorageClient` sounds like a production-use client. I think `MockStorageClient` is a better name if it's intended to be used in tests

splunkericl: `MockStorageClient` isn't explicit on what the storage client is doing. The word `Mock` can confuse whoever is using it. For example, if they have to do specific setups on the Mock? Do they have to mock input/output like the `mockery` library?

`InMemoryStorageClient` represents exactly what the struct is doing and being in a commonly used shared package it is important to have names that people can understand immediately. The package name is `testutil` and it is under an internal package so there are some safe guards to prevent people to mistake it as a production-use client.

dmitryax: Why do we need to mock persistent storage with in-memory storage if in-memory storage is the default anyway? Can we just don't set up the persistent storage in this test? It looks like the only side effect we need here is setting `requeuingEnabled` on `queuedRetrySender` which I believe can be set right in the test. In that case, we don't need the `testutil` package.

splunkericl: The default persistent storage is the file storage client but the test is using in memory storage client. I am not sure why the initial setup was this way but my guess is the test runs don't necessarily have correct files permission access?

`queued_retry` does initialize [storage](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L165) so we have to set some kinds of storage to test the scenarios. I think using the persistent_queue directly can tests these data loss scenarios nicely but it requires a storage client.
Another alternative is to write some mock queue so we can over write the queue in retry_sender. This way it is more unit test like but at the same time we can miss testing these functionalities all together.

splunkericl: This is what i mean by the alternative. The test is written directly against the retrySender. Then we don't have to do anything with the storageClient. However, I was hoping we can have some kinds of tests that involve multiple components so we know data loss can't happen with persistent queue. Maybe there can be some kinds of functional tests instead of unit tests.

```
func TestRetrySender_shutdown_tempError(t *testing.T) {
	req := newMockRequest(context.Background(), 3, errors.New("some error"))

	qCfg := NewDefaultQueueSettings()
	qCfg.NumConsumers = 1
	rCfg := NewDefaultRetrySettings()
	rCfg.InitialInterval = time.Hour // set high initial retry interval so it only waits
	rCfg.MaxElapsedTime = 0          // retry infinitely so shutdown can be triggered

	be, err := newBaseExporter(defaultSettings, fromOptions(WithRetry(rCfg), WithQueue(qCfg)), "", nopRequestUnmarshaler())
	require.NoError(t, err)
	require.NoError(t, be.Start(context.Background(), &mockHost{}))

	// replace nextSender inside retrySender to always return error so it doesn't exit send loop
	castedSender, ok := be.qrSender.consumerSender.(*retrySender)
	require.True(t, ok, "consumerSender should be a retrySender type")
	castedSender.nextSender = &ErrorRequestSender{
		errToReturn: errors.New("some error"),
	}
	onTempFailureErrorChan := make(chan error, 1)
	castedSender.onTemporaryFailure = func(_ *zap.Logger, _ internal.Request, err error) error {
		onTempFailureErrorChan <- err
		return nil
	}

	sendErrChan := make(chan error, 1)
	go func() {
		sendErrChan <- be.sender.send(req)
	}()
	require.NoError(t, <-sendErrChan)

	// shuts down so onTemporaryFailure is invoked
	require.NoError(t, be.Shutdown(context.Background()))
	require.ErrorContains(t, <-onTempFailureErrorChan, "interrupted due to shutdown")
}
```

dmitryax: Can't you use your original test, but instead of:

```golang
	storageID := component.NewIDWithName("file_storage", "storage")

	qCfg := NewDefaultQueueSettings()
	qCfg.NumConsumers = 1
	qCfg.StorageID = &storageID // enable persistence
	rCfg := NewDefaultRetrySettings()
	rCfg.InitialInterval = time.Millisecond
	rCfg.MaxElapsedTime = 0 // retry infinitely so shutdown can be triggered

	be, err := newBaseExporter(defaultSettings, fromOptions(WithRetry(rCfg), WithQueue(qCfg)), "", nopRequestUnmarshaler())
	require.NoError(t, err)

	var extensions = map[component.ID]component.Component{
		storageID: &mockStorageExtension{
			mockClient: testutil.NewInMemoryStorageClient(),
		},
	}
	host := &mockHost{ext: extensions}

	// we start correctly with a file storage extension
	require.NoError(t, be.Start(context.Background(), host))
```

do this:

```
	qCfg := NewDefaultQueueSettings()
	qCfg.NumConsumers = 1
	rCfg := NewDefaultRetrySettings()
	rCfg.InitialInterval = time.Millisecond
	rCfg.MaxElapsedTime = 0 // retry infinitely so shutdown can be triggered

	be, err := newBaseExporter(defaultSettings, fromOptions(WithRetry(rCfg), WithQueue(qCfg)), "", nopRequestUnmarshaler())
	require.NoError(t, err)
	be.qrSender.requeuingEnabled = true

	require.NoError(t, be.Start(context.Background(), componenttest.NewNopHost()))
```

splunkericl: Yeah actually this works. I can make the change

exporter/exporterhelper/queued_retry_test.go
```diff
@@ -582,6 +582,60 @@ func TestQueuedRetryPersistenceEnabledStorageError(t *testing.T) {
...
+	require.True(t, ok, "consumerSender should be a retrySender type")
+	castedSender.nextSender = &ErrorRequestSender{
+		errToReturn: errors.New("some error"),
+	}
+
+	sendErrChan := make(chan error, 1)
+	go func() {
+		sendErrChan <- be.sender.send(req)
+	}()
+	require.NoError(t, <-sendErrChan)
```
dmitryax, line 629: Why channel?

```suggestion
	require.NoError(t, be.sender.send(req))
```

splunkericl: The channel can block until the send operation completes in the goroutine. We can also extract the error returned and verify the error is nil.

exporter/exporterhelper/queued_retry_test.go
```diff
@@ -582,6 +582,60 @@ func TestQueuedRetryPersistenceEnabledStorageError(t *testing.T) {
...
+	be := dataRequeuedTest(t, req, produceCounter)
+
+	// shuts down and ensure the item is produced in the queue again
+	require.NoError(t, be.Shutdown(context.Background()))
+	assert.Eventually(t, func() bool {
+		return produceCounter.Load() == uint32(2)
+	}, time.Second, 1*time.Millisecond)
+}
+
+func dataRequeuedTest(t *testing.T, req internal.Request, produceCounter *atomic.Uint32) *baseExporter {
```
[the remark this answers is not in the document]
splunkericl: removed.

exporter/exporterhelper/queued_retry_test.go
dmitryax: `require.NoError(t, be.sender.send(req))` is a blocking call and also verifies the error

splunkericl: @MovieStoreGuy suggested in [thread](https://github.com/open-telemetry/opentelemetry-collector/pull/6815#discussion_r1068754947) that if the program exits before the goroutine completes, the error can cause a panic in the test program. Having the channel blocking outside can help.

As for the need to send the data in a goroutine is that the send operation puts the handling in an for loop. So we have to call `send` in `shutdown` in different goroutine.

dmitryax: But why was the goroutine needed in the first place?

splunkericl: The `send` operation has a for loop to keep processing the request until the processing has permanently failed or the program is exiting. While our test wants to test out the shutdown path, which is triggered when `shutdown` is called.

So if we don't put `send` in a different goroutine, the test program can't call `shutdown`. As the call to `send` will be waiting until it finishes but it won't finish until `shutdown` is called.

dmitryax: > As the call to send will be waiting until it finishes but it won't finish until shutdown is called.

But it finishes right after spawning the goroutine, you're waiting for it to be finished with no error `require.NoError(t, <-sendErrChan)`. This asynchronous block doesn't make sense. It does the same work as the one-liner that I suggested

dmitryax: Please make sure your test does what you expect

splunkericl: After double checking the code, it seems like i have confused `queuedRetrySender.Send` with `retrySender.Send`. In the test when `be.sender.send(req)` is called, it invokes `queuedRetrySender.Send`, which puts the request on a channel so the consumer will poll and use `retrySender` to `Send`.

I will update the call to not use goroutine. Apology for the confusion.

dmitryax (approved):

PR #6827, opened 2022-12-19
Make NewDefaultConfigProviderSettings public

ttomsu:
Signed-off-by: Travis Tomsu travis.tomsu@gmail.com

**Description:** Makes the function to create default config provider settings public, enabling using this function in e2e unit tests. Not sure why this is private, but the other functions to make e2e unit tests work are public (service.New -> service.CollectorSettings struct -> service.ConfigProvider). 

I currently have a copy of this function in my codebase, but I'd like to be able to use the one that's included by default to confirm everything starts up as intended.

**Testing:** Passed locally.

### DIFF
--- a/otelcol/collector_test.go
+++ b/otelcol/collector_test.go
@@ -42,7 +42,7 @@ func TestCollectorStartAsGoRoutine(t *testing.T) {
 	factories, err := nopFactories()
 	require.NoError(t, err)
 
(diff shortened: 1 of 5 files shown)


bogdandrutu (changes requested):
It is private because we are not yet willing to guarantee that we will not add new providers/etc.

Issue #6596, opened 2022-11-21
Implement ability for extensions to be notified about effective configuration


tigrannajaryan:
To implement OpAMP for the Collector we need a way for extensions to know what is the current effective configuration of the Collector, see the [design document](https://docs.google.com/document/d/1KtH5atZQUs9Achbce6LiOaJxLbksNJenvgvyKLsJrkc/edit#heading=h.ioikt02qpy5f.

This can be implemented as an optional interface that extensions can implement to be notified of the current config.

The config must be reported immediately after the extensions are started and then every time the config changes as a result of ConfigProvider reloads.

evan-bradley:
I can work on this.

tigrannajaryan:
@evan-bradley feel free to add me a as a reviewer on all OpAMP-related PRs.

bogdandrutu:
> This can be implemented as an optional interface that extensions can implement to be notified of the current config.

Why not an optional interface on the Host? I would like to determine when to do one or the other.

evan-bradley:
>Why not an optional interface on the Host? I would like to determine when to do one or the other.

I would like to use a push-based model for this so in the future the Host can update extensions that persist across configuration updates. Even if that doesn't happen anytime soon, having the Host push updates to extensions is not much more complex than having extensions call a method on the Host to get the effective configuration.

To notify extensions, we will either need to provide an interface on the Host that allows extensions to register message handlers, or allow the Host to call a method on extensions that want to be notified of updates. I think having an optional interface for extensions will provide the simplest solution. This also matches the approach to host-component communication for component status reporting in https://github.com/open-telemetry/opentelemetry-collector/pull/6560.

PR #6833, opened 2022-12-19
Notify extensions of the Collector's effective configuration

evan-bradley:
**Description:**

This adds an optional `ConfigWatcher` interface that extensions can implement if they want to be notified of the effective configuration that is used by the Collector.

I don't feel very strongly about any of the decisions I made in this PR, so I am open to input if we would like to take a different approach anywhere. I will leave some comments to explain the decisions I made.

**Link to tracking Issue:**
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/6596

**Testing:**

I've made minimal unit test changes, but I expect to write more tests once there is consensus on the direction for implementing this functionality. I have done some manual testing to show that an extension can get a YAML representation of the effective config using two YAML input files.

### DIFF
--- a/.chloggen/servicesettings-confmap.yaml
+++ b/.chloggen/servicesettings-confmap.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `CollectorConfig` field to `service.Settings`
+
+# One or more tracking issues or pull requests related to the change
+issues: [6596]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This field is intended to be used by the Collector to pass its effective configuration to the service.
--- a/.chloggen/servicesettings-confmap.yaml
+++ b/.chloggen/servicesettings-confmap.yaml
@@ -1,11 +1,11 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: breaking
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: service
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add `CollectorConfig` field to `service.Settings`
+note: Add `CollectorConf` field to `service.Settings`
 
 # One or more tracking issues or pull requests related to the change
 issues: [6596]
--- a/extension/extension.go
+++ b/extension/extension.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/confmap"
 )
 
 // Extension is the interface for objects hosted by the OpenTelemetry Collector that
@@ -30,6 +31,10 @@ type PipelineWatcher interface {
 	// This is sent before receivers are stopped, so the Extension can take any
 	// appropriate actions before that happens.
 	NotReady() error
+}
+
+type ConfigWatcher interface {
+	ConfigResolved(ctx context.Context, conf *confmap.Conf) error
 }
 
 // CreateSettings is passed to Factory.Create(...) function.
--- a/extension/extension.go
+++ b/extension/extension.go
@@ -33,8 +33,11 @@ type PipelineWatcher interface {
 	NotReady() error
 }
 
+// ConfigWatcher is an interface that should be implemented by an extension that
+// wishes to be notified of the Collector's effective configuration.
 type ConfigWatcher interface {
-	ConfigResolved(ctx context.Context, conf *confmap.Conf) error
+	// NotifyConfig notifies the extension of the Collector's current effective configuration.
+	NotifyConfig(ctx context.Context, conf *confmap.Conf) error
 }
 
 // CreateSettings is passed to Factory.Create(...) function.
--- a/otelcol/collector.go
+++ b/otelcol/collector.go
@@ -144,7 +144,7 @@ func (col *Collector) Shutdown() {
 func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	col.setCollectorState(StateStarting)
 
-	conf := &confmap.Conf{}
+	conf := confmap.New()
 
 	if cp, ok := col.set.ConfigProvider.(ConfmapProvider); ok {
 		var err error
--- a/otelcol/collector.go
+++ b/otelcol/collector.go
@@ -144,7 +144,7 @@ func (col *Collector) Shutdown() {
 func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	col.setCollectorState(StateStarting)
 
-	conf := confmap.New()
+	var conf *confmap.Conf
 
 	if cp, ok := col.set.ConfigProvider.(ConfmapProvider); ok {
 		var err error
--- a/otelcol/collector.go
+++ b/otelcol/collector.go
@@ -18,6 +18,7 @@ import (
 	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/confmap"
 	"go.opentelemetry.io/collector/connector"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/extension"
@@ -143,6 +144,17 @@ func (col *Collector) Shutdown() {
 func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	col.setCollectorState(StateStarting)
 
+	conf := &confmap.Conf{}
+
+	if cp, ok := col.set.ConfigProvider.(ConfmapProvider); ok {
+		var err error
+		conf, err = cp.GetConfmap(ctx)
+
+		if err != nil {
+			return fmt.Errorf("failed to resolve config: %w", err)
+		}
+	}
+
 	cfg, err := col.set.ConfigProvider.Get(ctx, col.set.Factories)
 	if err != nil {
 		return fmt.Errorf("failed to get config: %w", err)
@@ -154,6 +166,7 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 
 	col.service, err = service.New(ctx, service.Settings{
 		BuildInfo:         col.set.BuildInfo,
+		CollectorConf:     conf,
 		Receivers:         receiver.NewBuilder(cfg.Receivers, col.set.Factories.Receivers),
 		Processors:        processor.NewBuilder(cfg.Processors, col.set.Factories.Processors),
 		Exporters:         exporter.NewBuilder(cfg.Exporters, col.set.Factories.Exporters),
@@ -174,6 +187,7 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 		return multierr.Combine(err, col.service.Shutdown(ctx))
 	}
 	col.setCollectorState(StateRunning)
+
 	return nil
 }
 
--- a/otelcol/configprovider.go
+++ b/otelcol/configprovider.go
@@ -50,9 +50,22 @@ type ConfigProvider interface {
 	Shutdown(ctx context.Context) error
 }
 
+// ConfmapProvider is an optional interface to be implemented by ConfigProviders
+// to provide confmap.Conf objects representing a marshaled version of the
+// Collector's configuration.
+type ConfmapProvider interface {
+	// GetConfmap resolves the Collector's configuration and provides it as a confmap.Conf object.
+	//
+	// Should never be called concurrently with itself or any ConfigProvider method.
+	GetConfmap(ctx context.Context) (*confmap.Conf, error)
+}
+
 type configProvider struct {
 	mapResolver *confmap.Resolver
 }
+
+var _ ConfigProvider = &configProvider{}
+var _ ConfmapProvider = &configProvider{}
 
 // ConfigProviderSettings are the settings to configure the behavior of the ConfigProvider.
 type ConfigProviderSettings struct {
@@ -104,6 +117,15 @@ func (cm *configProvider) Watch() <-chan error {
 
 func (cm *configProvider) Shutdown(ctx context.Context) error {
 	return cm.mapResolver.Shutdown(ctx)
+}
+
+func (cm *configProvider) GetConfmap(ctx context.Context) (*confmap.Conf, error) {
+	conf, err := cm.mapResolver.Resolve(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("cannot resolve the configuration: %w", err)
+	}
+
+	return conf, nil
 }
 
 func newDefaultConfigProviderSettings(uris []string) ConfigProviderSettings {
--- a/otelcol/configprovider.go
+++ b/otelcol/configprovider.go
@@ -53,6 +53,9 @@ type ConfigProvider interface {
 // ConfmapProvider is an optional interface to be implemented by ConfigProviders
 // to provide confmap.Conf objects representing a marshaled version of the
 // Collector's configuration.
+//
+// The purpose of this interface is that otelcol.ConfigProvider structs do not
+// necessarily need to use confmap.Conf as their underlying config structure.
 type ConfmapProvider interface {
 	// GetConfmap resolves the Collector's configuration and provides it as a confmap.Conf object.
 	//
--- a/otelcol/configprovider_test.go
+++ b/otelcol/configprovider_test.go
@@ -11,71 +11,12 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/zap/zapcore"
 	"gopkg.in/yaml.v3"
 
-	"go.opentelemetry.io/collector/component"
-	"go.opentelemetry.io/collector/config/configtelemetry"
 	"go.opentelemetry.io/collector/confmap"
 	"go.opentelemetry.io/collector/confmap/provider/fileprovider"
 	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
-	"go.opentelemetry.io/collector/connector/connectortest"
-	"go.opentelemetry.io/collector/exporter/exportertest"
-	"go.opentelemetry.io/collector/extension/extensiontest"
-	"go.opentelemetry.io/collector/processor/processortest"
-	"go.opentelemetry.io/collector/receiver/receivertest"
-	"go.opentelemetry.io/collector/service"
-	"go.opentelemetry.io/collector/service/pipelines"
-	"go.opentelemetry.io/collector/service/telemetry"
 )
-
-var configNop = &Config{
-	Receivers:  map[component.ID]component.Config{component.NewID("nop"): receivertest.NewNopFactory().CreateDefaultConfig()},
-	Processors: map[component.ID]component.Config{component.NewID("nop"): processortest.NewNopFactory().CreateDefaultConfig()},
-	Exporters:  map[component.ID]component.Config{component.NewID("nop"): exportertest.NewNopFactory().CreateDefaultConfig()},
-	Connectors: map[component.ID]component.Config{component.NewIDWithName("nop", "con"): connectortest.NewNopFactory().CreateDefaultConfig()},
-	Extensions: map[component.ID]component.Config{component.NewID("nop"): extensiontest.NewNopFactory().CreateDefaultConfig()},
-	Service: service.Config{
-		Extensions: []component.ID{component.NewID("nop")},
-		Pipelines: pipelines.Config{
-			component.NewID("traces"): {
-				Receivers:  []component.ID{component.NewID("nop")},
-				Processors: []component.ID{component.NewID("nop")},
-				Exporters:  []component.ID{component.NewID("nop"), component.NewIDWithName("nop", "con")},
-			},
-			component.NewID("metrics"): {
-				Receivers:  []component.ID{component.NewID("nop")},
-				Processors: []component.ID{component.NewID("nop")},
-				Exporters:  []component.ID{component.NewID("nop")},
-			},
-			component.NewID("logs"): {
-				Receivers:  []component.ID{component.NewID("nop"), component.NewIDWithName("nop", "con")},
-				Processors: []component.ID{component.NewID("nop")},
-				Exporters:  []component.ID{component.NewID("nop")},
-			},
-		},
-		Telemetry: telemetry.Config{
-			Logs: telemetry.LogsConfig{
-				Level:       zapcore.InfoLevel,
-				Development: false,
-				Encoding:    "console",
-				Sampling: &telemetry.LogsSamplingConfig{
-					Initial:    100,
-					Thereafter: 100,
-				},
-				OutputPaths:       []string{"stderr"},
-				ErrorOutputPaths:  []string{"stderr"},
-				DisableCaller:     false,
-				DisableStacktrace: false,
-				InitialFields:     map[string]any(nil),
-			},
-			Metrics: telemetry.MetricsConfig{
-				Level:   configtelemetry.LevelBasic,
-				Address: "localhost:8888",
-			},
-		},
-	},
-}
 
 func newConfig(yamlBytes []byte, factories Factories) (*Config, error) {
 	var stringMap = map[string]interface{}{}
@@ -96,6 +37,7 @@ func newConfig(yamlBytes []byte, factories Factories) (*Config, error) {
 		Receivers:  cfg.Receivers.Configs(),
 		Processors: cfg.Processors.Configs(),
 		Exporters:  cfg.Exporters.Configs(),
+		Connectors: cfg.Connectors.Configs(),
 		Extensions: cfg.Extensions.Configs(),
 		Service:    cfg.Service,
 	}, nil
--- a/otelcol/configprovider_test.go
+++ b/otelcol/configprovider_test.go
@@ -98,3 +98,33 @@ func TestConfigProviderFile(t *testing.T) {
 
 	assert.EqualValues(t, configNop, cfg)
 }
+
+func TestGetConfmap(t *testing.T) {
+	configPath := filepath.Join("testdata", "otelcol-nop.yaml")
+	uriLocation := "file:" + configPath
+	provider := fileprovider.New()
+	set := ConfigProviderSettings{
+		ResolverSettings: confmap.ResolverSettings{
+			URIs:      []string{uriLocation},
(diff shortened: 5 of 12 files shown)


evan-bradley:
cc @tigrannajaryan @portertech

service/service.go
```diff
@@ -172,6 +172,13 @@ func (srv *Service) Shutdown(ctx context.Context) error {
 	return errs
 }
 
+func (srv *Service) ConfigResolved(_ context.Context, effectiveConfig map[string]interface{}) error {
```
evan-bradley, line 175: I decided to make a new method on the service for notifying extensions of changes in the configuration since we would also need a method to notify the service of more granular changes to the configuration if the Collector gains that capability. If we only expect the Collector to do full reloads of the service, then we could likely remove this method.

otelcol/config.go
```diff
@@ -131,3 +152,8 @@ func (cfg *Config) Validate() error {
 
 	return nil
 }
+
+// ToStringMap returns the map used to create this Config
+func (cfg *Config) ToStringMap() map[string]interface{} {
```
evan-bradley, line 157: Providing the map representation of the Config struct from the struct itself from seems like the best solution here. I made this a method to differentiate it from the other data properties on the struct, which then necessitated making a constructor for the struct to keep the field private.

otelcol/collector.go
```diff
@@ -181,6 +181,11 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 		return multierr.Combine(err, col.service.Shutdown(ctx))
 	}
 	col.setCollectorState(StateRunning)
+
+	if err = col.service.ConfigResolved(ctx, cfg.ToStringMap()); err != nil {
```
evan-bradley, line 185: Since this functionality will be used by the OpAMP extension, it is worth noting that the extension will probably have to be restarted every time the configuration changes. I think this is acceptable if we plan for the Collector to only restart the components affected by a configuration change.

service/service.go
bogdandrutu: You cannot remove something but you can always add. So please don't add anything until is necessary.

extension/extension.go
```diff
@@ -43,6 +43,10 @@ type PipelineWatcher interface {
 	NotReady() error
 }
 
+type ConfigWatcher interface {
+	ConfigResolved(data map[string]interface{}) error
```
bogdandrutu, line 47: Why not passing context? Why not passing the confmap.Conf?

bogdandrutu (changes requested):

extension/extension.go
evan-bradley: > Why not passing context?

I didn't pass context since I was mostly following the approach used by the `PipelineWatcher` interface, but I have no issue doing it here. I will add it.

> Why not passing the confmap.Conf?

I went for just passing a map since I don't think extensions will need any functionality from confmap.Conf. I see the value in using a Collector-specific configuration structure though, so I can change to that for now.

service/service.go
```diff
@@ -69,9 +70,10 @@ type Service struct {
 	telemetrySettings    component.TelemetrySettings
 	host                 *serviceHost
 	telemetryInitializer *telemetryInitializer
+	collectorConf        *confmap.Conf
 }
 
-func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
+func New(ctx context.Context, set Settings, cfg Config, collectorConf *confmap.Conf) (*Service, error) {
```
bogdandrutu, line 76: Please follow the conventions we have, conf should be part of settings. Also this is a breaking change.

tigrannajaryan (approved):

otelcol/configprovider_test.go
```diff
@@ -37,51 +38,66 @@ import (
...
-		Telemetry: telemetry.Config{
-			Logs: telemetry.LogsConfig{
-				Level:       zapcore.InfoLevel,
-				Development: false,
-				Encoding:    "console",
-				Sampling: &telemetry.LogsSamplingConfig{
-					Initial:    100,
-					Thereafter: 100,
+func newConfigNop() (*Config, error) {
+	yamlBytes, err := os.ReadFile(filepath.Join("testdata", "otelcol-nop.yaml"))
```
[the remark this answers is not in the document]
evan-bradley: Thanks for calling this out. I agree we should just use the YAML file, I had a change to these tests that implemented that but left it out to reduce the number of lines changed before we agree on an approach. Once we're satisfied with the direction I'll put that back that as I add more tests.

evan-bradley:
@bogdandrutu This should be ready for another look.

otelcol/config.go
```diff
@@ -60,6 +61,9 @@ type Config struct {
 	Extensions map[component.ID]component.Config
 
 	Service service.Config
+
+	// Conf is the confmap.Conf used to create this Config
+	Conf *confmap.Conf
```
bogdandrutu, line 66: Not ok, this is not parsed from the config, you break the design here since everything else is parsed from the config.

bogdandrutu (changes requested):

otelcol/config.go
evan-bradley: I agree, it's not ideal. I've pushed a commit that makes this field a method to separate it from the public fields on this struct. Let me know what you think.

If we don't want the source `confmap.Conf` struct available from `Config`, I see two possibilities:
* We could change the `ConfigProvider.Get` method to also return `*confmap.Conf`. To me, this muddies the method signature a little bit, but would provide a clear separation between the two types.
* We could add a new method to `ConfigProvider` to only resolve and return a `confmap.Conf` struct. I opted against this since we are already resolving the config in `Get` and I think another method would be unnecessary on the `ConfigProvider` interface, but this would also allow clearly separating `confmap.Conf` and `Config`.

Let me know if there's another option I've missed.

evan-bradley: @bogdandrutu I'm sorry, I think I misunderstood what you meant.

If we are looking to have both `Config` and `ConfigProvider` be agnostic of the underlying configuration structure, then I see a few options:

* Provide an optional interface for config providers to implement that resolves the configuration and translates it (if necessary) to a `confmap.Conf` object. This seems like it would probably be the most straightforward solution.
* Provide another type that can be marshaled into YAML from either the `Config` or `ConfigProvider` types that is used in `extension.ConfigWatcher`, in place of `confmap.Conf`. 
* Provide a way to translate `Config` instances back into `confmap.Conf`.

What do you think?

evan-bradley: @bogdandrutu I've pushed a commit including a rough draft for an optional interface to provide `confmap.Conf` objects from a Config Provider. Let me know what you think.

tigrannajaryan:
@bogdandrutu PTAL.

extension/extension.go
```diff
@@ -43,6 +44,10 @@ type PipelineWatcher interface {
 	NotReady() error
 }
 
+type ConfigWatcher interface {
+	ConfigResolved(ctx context.Context, conf *confmap.Conf) error
```
bogdandrutu, line 48: Is not "ConfigResolved" since that happened before the extension was created, not sure what a good name is...

service/extensions/extensions.go
```diff
@@ -83,6 +84,17 @@ func (bes *Extensions) NotifyPipelineNotReady() error {
 	return errs
 }
 
+func (bes *Extensions) NotifyConfigResolved(ctx context.Context, conf *confmap.Conf) error {
```
bogdandrutu, line 87: Instead of doing this design where extension defines an optional interface to be implemented, we define in the service an optional interface for the Host, and extensions that need this can check if the Host passed at Start time implements that interface and uses it? What do you think?

evan-bradley: I think the advantage of defining an optional interface for Hosts would be that the configuration would be available during extension creation time, as where an extension interface would only provide the configuration to the extension in a subsequent method call. My goal has been to leave a door open for components to change during the lifetime of the Host, which I think an extension interface provides for while not complicating the design too much. If we don't want to take that into consideration at this point or we'd like to delay that decision and make breaking changes later, then I think an interface for the Host will work.

tigrannajaryan:
@dmitryax I know Bogdan's availability is limited, it would be great if you could review this.

service/service.go
```diff
@@ -155,6 +161,10 @@ func (srv *Service) Start(ctx context.Context) error {
 		return err
 	}
 
+	if err := srv.host.serviceExtensions.NotifyConfigResolved(ctx, srv.collectorConf); err != nil {
+		return err
+	}
```
srikanthccv, line 166: Can `NotifyConfigResolved` be called right after the extensions are started? Or does it need to wait for pipelines to start?

service/service.go
```diff
@@ -47,6 +48,9 @@ type Settings struct {
 	// BuildInfo provides collector start information.
 	BuildInfo component.BuildInfo
 
+	// CollectorConf contains the Collector's current configuration
+	CollectorConf *confmap.Conf
```
srikanthccv, line 52: Does this allow one to mistakenly provide the `Conf` that doesn't match the other components, such as `Receivers`/`Exporters`? Probably not a big deal, but I wanted to bring this up since it is part of public API.

service/service.go
evan-bradley: My goal originally was to ensure that the pipeline successfully starts before notifying the extension of the config. Thinking on it, we should record the config even for failure cases, so I will update this to notify extensions as soon as they are started.

service/service.go
evan-bradley: This is provided to the service [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6833/files#diff-3c058aa5dde6d9b2b413b65dce28fe97e9cf859a0f03629639843aac6bc26e2aR178) and from there is directly passed to extensions. So the `Conf` should reflect the actual configuration and should be the same for each extension that receives it. I'm not sure I fully understand your question, does that address your concern?

srikanthccv: That should be fine.

srikanthccv (approved):

extension/extension.go
evan-bradley: I've updated the method name to `NotifyConfig`.

evan-bradley:
@dmitryax Your review would be greatly appreciated to get this closer to the finish line.

I will try to summarize the discussion so far, I'd welcome your input on these points.

**Getting the config**

The extension needs a serialized version of the Collector's effective configuration, which we determined is best provided as a `confmap.Conf` object. Since there's currently no way to go from an unmarshaled config (`otelcol.Config`) to a marshaled version, I've created `otelcol.ConfmapProvider`, an optional interface for `otelcol.ConfigProvider` structs to provide the unmarshaled config.

**Communicating the config**

We want to provide an optional interface to facilitate communicating the effective config from the Host to Extensions. The communication flow will look slightly differently depending on what interface we provide.

We can provide an interface to be implemented by the Host, e.g. with a `GetConfig` method, and have Extensions "pull" the configuration. I see the following benefits to this approach:
1. The extension can call this in its `Start` method, and obtaining the config is synchronous. This isn't particularly helpful in the context of OpAMP, but may be useful to other extensions.
2. Extensions can query the Host for the interface to determine whether the Host supports providing config. It's unclear whether this would be useful.
3. Other component classes could use the Host interface. It's not clear to me this is desirable.

We can also provide an interface for Extensions, which the Host will query for and "push" the config to the Extensions. This is the approach currently implemented in this PR. I think this approach has the following benefits:
1. If a `confmap.Provider` that dynamically updates the config were used, it would be feasible for Extensions to persist across config reloads and for the Host to notify them of configuration updates. This would add some complexity and isn't part of this change, but would be possible later on or for custom distributions to implement.
2. Restricts config notifications to Extensions.
3. Follows a similar pattern to the `PipelineWatcher` interface.

Overall providing an interface for extensions opens the possibilities for additional functionality in the future while not being much more complex. However, I think both approaches would work fine and the differences between them aren't substantial right now.

mwear (approved):

service/service.go
```diff
@@ -68,6 +72,7 @@ type Service struct {
 	telemetrySettings    component.TelemetrySettings
 	host                 *serviceHost
 	telemetryInitializer *telemetryInitializer
+	collectorConf        *confmap.Conf
 }
 
 func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
```
djaglowski, line 78: We are passing in two different representations of the same config, one explicitly as "the config", the other riding along in the "settings".

I think @srikanthccv made a good point [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6833/files#r1246509584) that since these two representations are both passed in through the public API, we are introducing an opportunity for them to be out of sync.

Since we are proposing a breaking change to the way we create a Service, I think we should consider a change that does not introduce this possibility.

I am new to this section of the code so may be overlooking some details but perhaps it would be cleaner to pass the `ConfigProvider` into `service.New`, _instead_ of `service.Config`? This would require moving `otelcol.ConfigProvider` and the `configProvider` implementation into the `service` package, but I think it would be more logically consistent and would remove the need for the new `ConfmapProvider` interface.

otelcol/configprovider.go
```diff
@@ -50,10 +50,23 @@ type ConfigProvider interface {
...
+// ConfmapProvider is an optional interface to be implemented by ConfigProviders
+// to provide confmap.Conf objects representing a marshaled version of the
+// Collector's configuration.
+type ConfmapProvider interface {
+	// GetConfmap resolves the Collector's configuration and provides it as a confmap.Conf object.
+	//
+	// Should never be called concurrently with itself or any ConfigProvider method.
+	GetConfmap(ctx context.Context) (*confmap.Conf, error)
+}
+
```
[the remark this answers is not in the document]
evan-bradley: The goal of the interface was to not require structs implementing `ConfigProvider` to have any dependencies on `confmap.Conf`. As for the name, I agree that it is a bit unfortunate. I am open to suggestions here, my intent was to signify that `otelcol.ConfigProvider` provides `Config` structs while `otelcol.ConfmapProvider` provides `confmap.Conf` structs.

otelcol/collector.go
```diff
@@ -143,6 +144,17 @@ func (col *Collector) Shutdown() {
 func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	col.setCollectorState(StateStarting)
 
+	conf := &confmap.Conf{}
```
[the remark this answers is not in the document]
evan-bradley: This is a good catch, this will leave an uninitialized pointer that very well may cause issues. I'll switch this to use `confmap.New`.

.chloggen/servicesettings-confmap.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
dmitryax, line 2: Technically adding a field to a struct is not a breaking change. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/VERSIONING.md, we follow the go compatibility promise. We don't need it to be required, right?

service/extensions/extensions.go
```diff
@@ -72,6 +73,17 @@ func (bes *Extensions) NotifyPipelineNotReady() error {
 	return errs
 }
 
+func (bes *Extensions) NotifyConfig(ctx context.Context, conf *confmap.Conf) error {
```
dmitryax, line 76: Can we check if `srv.collectorConf != nil` here so we can keep the `Settings.CollectorConf` optional?

dmitryax (commented):
I'll make another thorough review tomorrow

.chloggen/servicesettings-confmap.yaml
evan-bradley: That's a good point, I think I may have forgot to change this when updating the `service.New` call. Updated.

service/extensions/extensions.go
evan-bradley: That's a good idea. I did this in `service.Start`.

otelcol/collector.go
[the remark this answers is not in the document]
evan-bradley: Updated this to make it optional per feedback: https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1267595234.

service/service.go
evan-bradley: >This would require moving otelcol.ConfigProvider and the configProvider implementation into the service package, but I think it would be more logically consistent and would remove the need for the new ConfmapProvider interface.

I've spent some time looking into this, and I think an argument could be made for moving these into the `service` package. The `otelcol` package is primarily tasked with setting up the Collector process: OS signal handling, environment variables, command line flag parsing, etc. Output from `otelcol.ConfigProvider` is just passed to the service and isn't used by `otelcol.Collector`, so encapsulating it in the `service` package seems like it would result in a cleaner API. I'll do some additional prototyping to get a sense for whether this is a viable approach.

That said, I have had some difficulties fully understanding the design decisions behind the current API boundaries, so I may be missing something essential. @open-telemetry/collector-maintainers what do you think?

> We are passing in two different representations of the same config, one explicitly as "the config", the other riding along in the "settings".
>
> I think @srikanthccv made a good point [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6833/files#r1246509584) that since these two representations are both passed in through the public API, we are introducing an opportunity for them to be out of sync.

I missed the original point of that comment, but it's a good point, thanks for calling that out. My original goal was to avoid this situation by giving `Config` a method that would convert it to a map-like structure, but that isn't possible without giving it a dependency on `confmap.Conf`. If moving `otelcol.ConfigProvider` to `service` doesn't pan out I'll see if there's a way this can be addressed.

djaglowski: > design decisions behind the current API boundaries

I'm was not tuned in when these changes were made but my understanding of the boundary between `otelcol` and `service` is roughly that `otelcol` should be able to run multiple instances of `service.Service` in sequence. I think the changes we are talking about are in line with this intention.

dmitryax: `service` module should represent the `service` part of the configuration see https://github.com/open-telemetry/opentelemetry-collector/issues/5564. `ConfigProvider` was there before moving to `otelcol`, see https://github.com/open-telemetry/opentelemetry-collector/pull/6768. Maybe we can move to a new separate module given that it only depends on `confmap` so both otelcol and service can use it, but I cannot find any good place for it.

dmitryax (commented):
> We can provide an interface to be implemented by the Host, e.g. with a GetConfig method, and have Extensions "pull" the configuration.

@evan-bradley have you considered an optional host interface with a function taking a callback that extensions can provide to be notified? With that, we should get the "push" model

service/service.go
djaglowski: Maybe [`confmap/provider`](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider)? There's no code in the package currently but it seems like a natural place.

otelcol/configprovider.go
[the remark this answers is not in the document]
tigrannajaryan: >The goal of the interface was to not require structs implementing ConfigProvider to have any dependencies on confmap.Conf.

I am probably missing something, but isn't `configProvider` struct in this same file the only implementation of `ConfigProvider`? And the `configProvider` struct in this same PR adds implementation of `GetConfmap()` func to `configProvider` struct , so it does take dependency `confmap.Conf`, right?

Are there are other implementations of `ConfigProvider` elsewhere which will not implement `GetConfmap()` func?

Which "structs implementing ConfigProvider" do you refer to?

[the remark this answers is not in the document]
evan-bradley: The current design is in response to this feedback: https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1097855015. It's possible I've misinterpreted the feedback.

I'll try to answer your questions by explaining my understanding of the situation and how my approach fits in.

As far as I'm aware, `configProvider` is the only implementation of `ConfigProvider` in any upstream Collector repository. Furthermore, it is only currently used inside the package where it is defined. However, the existence of `ConfigProvider` as a public interface implies that anyone who wants to reimplement this part of the Collector can provide their own implementation. At present, `configProvider` implements `ConfigProvider` using `confmap.Conf` as its underlying format, but `ConfigProvider` doesn't specify any usage of `confmap.Conf`. Anyone who reimplements this part of the Collector could pass `otelcol.NewCollector` a `ConfigProvider` that knows nothing about `confmap.Conf`.

This works, but looking at it again now that I've spent a lot more time in this part of the code, I see a few things I don't like about this approach:
1. It's not clear why someone would want to avoid `confmap.Conf` when it is used across the rest of the Collector.
2. Interacting with `confmap.Conf` is required for the rest of the functionality in this PR, so realistically it's necessary regardless.
3. As pointed out [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1246509584) and [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1267318565), the configuration represented by `Config` and `confmap.Conf` may not be in sync.

I would still argue that my original approach of providing `Config.ToConfmap` is probably the best approach. I think it makes the most sense to marshal data directly from the data itself, and the best way to do that here would be to store the marshaled representation on `Config`. `confmap.Conf` offers this with `NewFromStringMap` and `Conf.ToStringMap`, so there is also precedence for this design.

service/service.go
evan-bradley: >`service` module should represent the `service` part of the configuration see https://github.com/open-telemetry/opentelemetry-collector/issues/5564. `ConfigProvider` was there before moving to `otelcol`, see https://github.com/open-telemetry/opentelemetry-collector/pull/6768.

Thank you for calling these out. I think most of my confusion around this stems from the fact that it seems like the `service` part of the config is just the whole config. At present, the entirety of the config represented by `otelcol.Config` ultimately only goes to configure the service or one of its subcomponents. Is it expected that it may be possible to configure other aspects of the Collector that aren't in the service? Furthermore, since the service contains all of the Collector's prominent features (reporting self-telemetry, running extensions and pipelines), what would we expect might live outside it?

>Maybe we can move to a new separate module given that it only depends on confmap so both otelcol and service can use it, but I cannot find any good place for it.

>Maybe [confmap/provider](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider)? There's no code in the package currently but it seems like a natural place.

I think `ConfigProvider` could possibly make sense in `confmap/provider`, though in that case I would prefer to abstract away the details around the Collector's configuration format since that feels like an orthogonal concern to the `confmap` package.

evan-bradley: I have a sketch of what moving `ConfigProvider` out of the `otelcol` package would look like here: https://github.com/evan-bradley/opentelemetry-collector/commit/0364f7a031048c681d350e56862967786ff0a441. I've implemented it by moving it to the `service` package, but I think some of the design would still stand.

The assumption I'm making in this is that the distinction between `otelcol` and `service` is that `otelcol` provides the Collector as an application, as where `service` provides it as a library. Depending on the answers to my questions in my previous comment, this way of looking at things may entirely miss the mark.

otelcol/configprovider.go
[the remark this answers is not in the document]
tigrannajaryan: >It's not clear why someone would want to avoid confmap.Conf when it is used across the rest of the Collector.

+1. I don't understand what problem the avoidance solves.

evan-bradley:
>@evan-bradley have you considered an optional host interface with a function taking a callback that extensions can provide to be notified? With that, we should get the "push" model

I did think about this a little, I should have called this out as an option when summarizing possible approaches. For the most part this is functionally equivalent to an interface for extensions. Where I think it would get more complex is in the case where a Collector distro has a custom service that supports some variation of hot reloading, we would need to provide a mechanism for registering and unregistering callbacks. I've looked at [#6550](https://github.com/open-telemetry/opentelemetry-collector/pull/6550/files#diff-a82b69bc08c50d251a8b5e1f66810b29336d3016afcaed1bc05371c18bfb9b62) and [#6560](https://github.com/open-telemetry/opentelemetry-collector/pull/6560/files#diff-bcb3fc758a54acf0dda21a0d9c79b6cf6e6adcf58d0a282c93be50553523b09d) for ideas on how the two might compare. Overall, I think the interface for extensions ends up feeling conceptually simpler since it's just a method call and doesn't require any additional structures (callback lists, message buses, etc.).

dmitryax:
> Where I think it would get more complex is in the case where a Collector distro has a custom service that supports some variation of hot reloading, we would need to provide a mechanism for registering and unregistering callbacks

Why would we need unregistering of callbacks?

> Overall, I think the interface for extensions ends up feeling conceptually simpler since it's just a method call and doesn't require any additional structures (callback lists, message buses, etc.).

With the callback, I think we solve the problem pointed out in https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1267318565. We won't need `ConfmapProvider` and `ConfigWatcher` interfaces in that case, just another host interface that can look something like this:

```
type ConfmapNotifier interface {
	Subscribe(func(context.Context) (*confmap.Conf, error)) error
}
```

What other additional structures do you think we need? Maybe I miss something.

evan-bradley:
I apologize in advance for the wall of text, but unfortunately these issues are really about the details.

> Why would we need unregistering of callbacks?

If someone were to implement a service that allows reloading an extension only when its configuration changes, they would need to unregister callbacks for extensions that will be restarted. Thinking on it more though, the Host could handle this itself, so I think we are good here.

 >We won't need ConfmapProvider and ConfigWatcher interfaces in that case, just another host interface that can look something like this:

I would expect the interface to look something more like the following, with `confmap.Conf` as a parameter instead of a return value in the callback function. Is this correct, or am I missing how the interface in your comment would work?

```golang
type ConfmapNotifier interface {
	Subscribe(func(context.Context, *confmap.Conf) error) error
}
```

>What other additional structures do you think we need?

It's not a big deal, but with an interface for extensions, we just loop through the service's set of extensions and query for the interface. With callbacks, we have to keep a list/map of callbacks and maintain it as extensions subscribe, and if a custom Service implements it, as they are hot reloaded.

>We won't need ConfmapProvider and ConfigWatcher interfaces in that case, just another host interface that can look something like this:

I see how it would replace `ConfigWatcher`, but it's not clear to me that we could replace `ConfmapProvider` with a Host interface. Could you detail that a little more?

We need to pull the config out of `ConfigProvider`, then have the `Collector` pass it to the `Service`. `ConfmapProvider` is one way, but there are two other possible options I see to resolve that:
1. Just return it with `ConfigProvider.Get`. This feels a little strange to me, but would work.
2. Implement `Config.ToConfmap`. I had originally implemented it this way, but revised my approach to address this feedback: https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1097855015.

We could also use a callback here, but I feel like we shouldn't need to since `ConfigProvider` has a `Watch` method.

As for https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1267318565, the core of the issue as I see it is passing two representations of the config across public API boundaries, I only see two solutions here:
1. Pass a single object from `ConfigProvider` to `service.New` so that the service gets all of its config from a single source. This would probably be `otelcol.Config`. I'm a little hesitant about this since it would confuse the function signature in my opinion, as fields from `otelcol.Config` are passed into two arguments in `service.New`.
2. Give the service its own `ConfigProvider`, which I played around with here: https://github.com/open-telemetry/opentelemetry-collector/pull/6833#discussion_r1270873077.

dmitryax (approved):
I think this is a good solution. I played with it and didn't find any better way to expose this functionality.

dmitryax:
@djaglowski, please take if it's good to go from your perspective

djaglowski (approved):

Issue #6845, opened 2022-12-22
[exporter/otlphttp] Retry only on status code 429/502/503/504


newly12:
## Important (read before submitting)
Update retry based on spec change https://github.com/open-telemetry/opentelemetry-specification/pull/3028

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #6846, opened 2022-12-22
[exporter/otlphttp] Retry only on status code 429/502/503/504

newly12:
#6845

**Description:** 
Update retry based on spec change

#6845 

### DIFF
--- a/.chloggen/otlpthttp-retry.yaml
+++ b/.chloggen/otlpthttp-retry.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporter/otlphttp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Retry only on status code 429/502/503/504
+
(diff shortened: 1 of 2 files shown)


newly12:
Greetings, could you please take a look at this PR?

codeboten (approved):

kgoralski:
Hey @newly12, what if the sender will receive just 500 Internal Server Errors? It will not retry on it but I think it should? 
Especially in the case of using daemonsets -> gateway and gateway responding with a high memory limit. It will respond with:

```shell
Exporting failed. The error is not retryable. Dropping data.
Permanent error: error exporting items, request to http://gateway.otel.svc.cluster.local:4318/v1/logs responded with HTTP Status Code 500, Message=data dropped due to high memory usage, Details=[]
```
I think it looks like a retriable error?

newly12:
I think it makes sense to retry on 500 error. I think maybe we can discuss more in https://github.com/open-telemetry/opentelemetry-specification/issues/2915

kgoralski:
> I think it makes sense to retry on 500 error. I think maybe we can discuss more in [open-telemetry/opentelemetry-specification#2915](https://github.com/open-telemetry/opentelemetry-specification/issues/2915)

Found this issue about it:
* https://github.com/open-telemetry/opentelemetry-specification/issues/3314

PR #6860, opened 2022-12-27
Remove deprecated pcommon.Value.Equal

bogdandrutu:


djaglowski (approved):

dmitryax (approved):

dmitryax:
It looks like it's actually being used for filtering in the contrib production code. Let me handle that first

dmitryax:
@bogdandrutu I have the following PRs to remove the method usage from contrib:

- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/17527 can be applied with performance improvement.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/17521 doesn't provide any performance benefits.

I think https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/17521 proves that we probably need to keep Value.Equal method.

Issue #6868, opened 2022-12-29
[config] configopaque discloses the length of the secret


atoulme:
**Describe the bug**
See https://github.com/open-telemetry/opentelemetry-collector/blob/e4e76dec22ecf8cf1d5b70af6d74215f0e05f0c0/config/configopaque/opaque.go#L27
`MarshalText marshals into a string of '*' of length equal to the opaque string.`

This inadvertently discloses the length of the secret. It should be of fixed length instead.

mx-psi:
Looks like this was fixed by #6869

PR #6869, opened 2022-12-29
[config/configopaque] fix secret length disclosure with configopaque String

atoulme:
**Description:**
Fix secret length disclosure with configopaque String. Use a 8 byte length consistently.

**Link to tracking Issue:**
#6868

**Testing:**
Added tests to show how the same string is outputted no matter what length of the secret

### DIFF
--- a/config/configopaque/opaque.go
+++ b/config/configopaque/opaque.go
@@ -21,13 +21,13 @@ import (
 // String alias that is marshaled in an opaque way.
 type String string
 
-const maskedString = "********"
+const maskedString = "[REDACTED]"
 
 var maskedBytes = []byte(maskedString)
 
 var _ encoding.TextMarshaler = String("")
 
-// MarshalText marshals into a string of 8 '*' characters.
+// MarshalText marshals the string as `[REDACTED]`.
 func (s String) MarshalText() ([]byte, error) {
 	return maskedBytes, nil
 }
--- a/config/configopaque/opaque.go
+++ b/config/configopaque/opaque.go
@@ -23,11 +23,9 @@ type String string
 
 const maskedString = "[REDACTED]"
 
-var maskedBytes = []byte(maskedString)
-
 var _ encoding.TextMarshaler = String("")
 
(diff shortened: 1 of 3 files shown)


config/configopaque/opaque.go
```diff
@@ -15,16 +15,19 @@
...
 
 import (
-	"bytes"
 	"encoding"
 )
 
 // String alias that is marshaled in an opaque way.
 type String string
 
+const maskedString = "********"
```
Aneurysm9, line 24:
```suggestion
const maskedString = "[REDACTED]"
```

I think I would prefer to have a placeholder that makes clear no length information is carried.  Users may still assume that `********` masks an eight character string.

atoulme: Done.

config/configopaque/opaque.go
```diff
@@ -15,16 +15,19 @@
...
+
+var maskedBytes = []byte(maskedString)
+
 var _ encoding.TextMarshaler = String("")
 
-// MarshalText marshals into a string of '*' of length equal to the opaque string.
+// MarshalText marshals the string as `[REDACTED]`.
 func (s String) MarshalText() ([]byte, error) {
-	return bytes.Repeat([]byte("*"), len(s)), nil
+	return maskedBytes, nil
```
bogdandrutu, line 32: Not sure if the contract of this func is that the caller cannot change the return type. I would suggest:
```suggestion
	return []byte(maskedString), nil
```

bogdandrutu (approved):

config/configopaque/opaque.go
atoulme: Is there no perf gain in caching the byte array as a var?

bogdandrutu: May be, but it is incorrect since if caller does:

```
t := str.MarshalText()
t[2] = 0xFF
```

The result for future calls will be affected.

atoulme: Good point, will change.

bogdandrutu (approved):

PR #6876, opened 2023-01-03
Fix: Force usage of case-sensitive keys in configurations

rapphil:
Signed-off-by: Raphael Silva <rapphil@gmail.com>

**Description:** Fix bug that allowed users to specify invalid configurations that would pass validation but would not work properly or fail silently. E.g.: `loglevel`. If you specify `logLevel`, the validation would pass but the collector logging exporter would not produce logs.

**Testing:** Added unit test before changing the code.

### DIFF
--- a/.chloggen/case-sensitive-configuration.yaml
+++ b/.chloggen/case-sensitive-configuration.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix bug in confmap validation that allowed the usage of case-insensitive keys in the configurations, leading to unexpected behavior.
+
+# One or more tracking issues or pull requests related to the change
+issues: [6876]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/case-sensitive-configuration.yaml
+++ b/.chloggen/case-sensitive-configuration.yaml
@@ -5,7 +5,7 @@ change_type: 'bug_fix'
 component: confmap
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix bug in confmap validation that allowed the usage of case-insensitive keys in the configurations, leading to unexpected behavior.
+note: Fix bug in confmap validation that allowed the usage of case-insensitive keys in the configurations, despite them failing silently.
 
 # One or more tracking issues or pull requests related to the change
 issues: [6876]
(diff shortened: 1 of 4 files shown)


.chloggen/case-sensitive-configuration.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix bug in confmap validation that allowed the usage of case-insensitive keys in the configurations, leading to unexpected behavior.
```
gbbr, line 8:
```suggestion
note: Fix bug in confmap validation that allowed the usage of case-insensitive keys in the configurations, despite them failing silently.
```
I would recommend making this change, otherwise one might be led to believe that this is a breaking change where we start disallowing case-insensitive config options. The real bug here is the silent failure.

confmap/confmap.go
```diff
@@ -189,6 +190,13 @@ func encoderConfig(rawVal interface{}) *encoder.EncoderConfig {
 	}
 }
 
+// case-sensitive version of the callback to be used in the MatchName property
+// of the DecoderConfig. The default for MatchEqual is to use strings.EqualFold,
+// which is case-insensitive.
+func caseSensitiveMatchName() func(mapKey, keyName string) bool {
+	return func(mapKey, keyName string) bool { return mapKey == keyName }
```
gbbr, line 197:
```suggestion
func caseSensitiveMatchName() func(a, b string) bool {
	return func(a, b string) bool { return a == b }
```
nit: this is a generic function that works for any strings. There isn't that much value in using such long descriptive variable names that are only used on two lines.

gbbr (commented):
I hope you'll permit a stranger to start getting involved with nitty reviews :)

.chloggen/case-sensitive-configuration.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
```
bryan-aguilar, line 2: Is this a breaking change? Users who have the wrong case will now receive an error when starting their collector.

rapphil: possible. Is there any implication besides the changelog? 

I'm a bit resistant on changing though because essentially every bug is a breaking change. Users might rely on buggy behavior and when you fix it you can cause a regression.

Aneurysm9: There's always a [relevant XKCD](https://xkcd.com/1172/).

I don't think this is a breaking change.  It will cause configurations that are already failing to fail loudly and quickly.

Aneurysm9 (approved):

confmap/confmap.go
```diff
@@ -189,6 +190,13 @@ func encoderConfig(rawVal interface{}) *encoder.EncoderConfig {
 	}
 }
 
+// case-sensitive version of the callback to be used in the MatchName property
```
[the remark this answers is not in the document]
rapphil: I was following the convention of the functions used in `DecodeHook`, including `expandNilStructPointersHookFunc`. But I guess this does not make sense since `caseSensitiveMatchName` is not part of that and is just a function with defined signature.
I will change.

Aneurysm9 (approved):

PR #6889, opened 2023-01-04
Add connector.MakeFactoryMap

djaglowski:
This adds the `connector.MakeFactoryMap` function, similar to that found in other component type packages.

### DIFF
--- a/connector/connector.go
+++ b/connector/connector.go
@@ -453,3 +453,16 @@ func NewFactory(cfgType component.Type, createDefaultConfig component.CreateDefa
 	}
 	return f
 }
+
+// MakeFactoryMap takes a list of connector factories and returns a map with factory type as keys.
+// It returns a non-nil error when there are factories with duplicate type.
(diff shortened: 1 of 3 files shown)


connector/connector.go
```diff
@@ -453,3 +453,16 @@ func NewFactory(cfgType component.Type, createDefaultConfig component.CreateDefa
 	}
 	return f
 }
+
+// MakeFactoryMap takes a list of connector factories and returns a map with factory type as keys.
+// It returns a non-nil error when there are factories with duplicate type.
+func MakeFactoryMap(factories ...Factory) (map[component.Type]Factory, error) {
```
MovieStoreGuy, line 459: Not sure if other `MakeFactoryMap` functions do so but is it worth checking if the factory is `nil`?

djaglowski: Since it's a variadic function, the input can be looped over safely, even if empty. Whether it should ever be empty is another question but it seems reasonable to let the caller handle that.

bogdandrutu (approved):

PR #6893, opened 2023-01-05
Add testcomponents.StatefulComponent

djaglowski:
This is a small part of #6700. The main idea is to provide a common way to validate components, especially in complex pipelines. Current tests rely on uniform expectations such as all exporters receiving 1 signal. However, testing the connector framework would ideally include expectations on a per-component basis.

This PR adds only the basic start/stop functionality, which itself reduces repetitive code somewhat. In the future, I would like to validating expected signals observed by each component.

### DIFF
--- a/service/internal/testcomponents/stateful_component.go
+++ b/service/internal/testcomponents/stateful_component.go
@@ -12,25 +12,33 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package testcomponents
+package testcomponents // import "go.opentelemetry.io/collector/service/internal/testcomponents"
 
 import (
 	"context"
-	"testing"
 
-	"github.com/stretchr/testify/assert"
-
-	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/component"
 )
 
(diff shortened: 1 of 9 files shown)


jpkrohling (approved):
Nice, it looks cleaner now.

service/internal/testcomponents/stateful_component.go
```diff
@@ -0,0 +1,50 @@
...
+
+package testcomponents // import "go.opentelemetry.io/collector/service/internal/testcomponents"
+
+import (
+	"context"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+type StatefulComponent interface {
```
bogdandrutu, line 23: I know that it is internal package, but do you need to expose this public? And more than that, do you need it at all?

bogdandrutu (approved):

service/internal/testcomponents/stateful_component.go
djaglowski: I believe it will be useful when validating complex component graphs, because each component can be treated as an interface. But it is not needed in this PR so I've removed it.

PR #6895, opened 2023-01-06
Add component node types

djaglowski:
Part of #6700

In order to manage components in a directed graph, each node must be associated with a unique identifier (`int64`) in the case of the `gonum` library. This PR defines a simple hashing system for all types of components, as well as a struct type for each component that contains all of the context necessary to reason about the nodes within the graph.

### DIFF
--- a/service/nodes.go
+++ b/service/nodes.go
@@ -1,0 +1,120 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// nolint:unused
+package service // import "go.opentelemetry.io/collector/service"
+
+import (
+	"hash/fnv"
+	"strings"
+
+	"gonum.org/v1/gonum/graph"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+type nodeID int64
+
+func (n nodeID) ID() int64 {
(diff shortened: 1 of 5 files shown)


service/nodes.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (n componentNodeID) ID() int64 {
+	return int64(n)
+}
+
+func newComponentNodeID(parts ...string) componentNodeID {
+	h := fnv.New64a()
+	for _, part := range parts {
+		_, _ = h.Write([]byte(part))
+	}
```
Aneurysm9, line 36:
```suggestion
	h.Write([]byte(strings.Join(parts, "|"))
```

A delimiter is needed between parts to avoid accidental collision between `[]string{"ab"}` and `[]string{"a", "b"}`.

djaglowski:
@open-telemetry/collector-approvers, please take a look.

service/nodes.go
```diff
@@ -0,0 +1,120 @@
...
+type nodeID int64
+
+func (n nodeID) ID() int64 {
+	return int64(n)
+}
+
+func newNodeID(parts ...string) nodeID {
+	h := fnv.New64a()
+	h.Write([]byte(strings.Join(parts, "|")))
+	return nodeID(int64(h.Sum64()))
```
jpkrohling, line 36: can't you just `nodeID(h.Sum64())` ?

service/nodes.go
```diff
@@ -0,0 +1,120 @@
...
+	"go.opentelemetry.io/collector/component"
+)
+
+type nodeID int64
+
+func (n nodeID) ID() int64 {
+	return int64(n)
+}
+
+func newNodeID(parts ...string) nodeID {
```
jpkrohling, line 33: This looks a bit "lost" in this PR -- where is it being used?

service/nodes.go
djaglowski: Good catch, fixed.

service/nodes.go
djaglowski: It's used by each of the `new<ComponentType>Node` functions, to generate their unique ids.

jpkrohling (approved):

codeboten (approved):

PR #6924, opened 2023-01-10
Fix undefined symbol errors on building otelcorecol for Plan 9

lufia:
**Description:**

Building **otelcorecol** was broken for Plan 9.

```console
$ cd cmd/otelcorecol
$ GOOS=plan9 go build
# go.opentelemetry.io/collector/exporter/loggingexporter
../../exporter/loggingexporter/known_sync_error.go:29:10: undefined: syscall.ENOTSUP
../../exporter/loggingexporter/known_sync_error.go:31:10: undefined: syscall.ENOTTY
../../exporter/loggingexporter/known_sync_error.go:33:10: undefined: syscall.EBADF
```

**Testing:**
Though I'm not sure yet there are more other issues when otelcorecol is running on Plan 9, at least, this PR is fixed undefined symbol errors and I checked **otelcorecol** shows a help successfully with `-h` option.

### DIFF
--- a/.chloggen/otelcol-plan9build.yaml
+++ b/.chloggen/otelcol-plan9build.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: loggingexporter
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix undefined symbol errors on building otelcorecol for Plan 9
+
+# One or more tracking issues or pull requests related to the change
+issues: [6924]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/exporter/loggingexporter/known_sync_error_plan9.go
+++ b/exporter/loggingexporter/known_sync_error_plan9.go
@@ -12,35 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build linux || darwin
-// +build linux darwin
+//go:build plan9
+// +build plan9
 
 package loggingexporter // import "go.opentelemetry.io/collector/exporter/loggingexporter"
 
-import (
-	"errors"
-	"syscall"
-)
-
-var knownSyncErrors = []error{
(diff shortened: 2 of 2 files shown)


mx-psi (commented):
Hey @lufia, thanks for the PR. While we don't currently have a well defined policy on when/how to add support to new OS or architectures, I personally would be fine accepting this change with the understanding that
- we can't ensure we won't break the Plan 9 build in the future
- we probably will not be able to address any bug reports that are Plan 9 specific
- we won't provide official builds or run the test suite on CI for Plan 9
- a Plan 9 build may not have feature parity with our official builds
- If there are issues that interfere with development we may need to remove support for Plan 9 entirely if we don't get help from your or other contributors

(where 'we' here is to be understood as 'the current OTel Collector community')

The above is my personal take on this, other contributors may disagree. I can work on getting agreement from the community and making this informal policy more well defined if this seems acceptable to you.

lufia:
Hi @mx-psi thank you for your review!

I have agree all of them.

I'd like to serve otelcontribcol service in my own Plan 9 box so far. I would make an effort to keep Plan 9 build healthy as far as I possible so please feel free to tell me if someone notices any issues in the future.

exporter/loggingexporter/known_sync_error.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !windows
-// +build !windows
+//go:build !windows && !plan9
+// +build !windows,!plan9
```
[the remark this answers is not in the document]
lufia: I fixed this in  9879a6c

exporter/loggingexporter/known_sync_error.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !windows
-// +build !windows
+//go:build unix
+// +build unix
```
mx-psi, line 16: :facepalm: sorry, the `unix` build tag was added on Go 1.19 so we can't use this yet since we support Go 1.18.

I believe something like this should make the build happy until we drop support for Go 1.18
```suggestion
//go:build linux || darwin
// +build linux darwin
```

lufia: :relaxed: 
question: It is better to merge this suggestion? or is better to fix my local repo and push force?

mx-psi: Either is fine, whatever is more comfortable for you :)

mx-psi (approved):
LGTM :rabbit: :astronaut:

lufia:
Thats glenda!

PR #6925, opened 2023-01-10
[chore] Move component loggers into service/internal/components

djaglowski:
This reduces complexity of `service/pipelines.go` and can be reused such as in #6700


service/internal/components/loggers.go
```diff
@@ -0,0 +1,42 @@
...
+func ProcessorLogger(logger *zap.Logger, id component.ID, pipelineID component.ID) *zap.Logger {
+	return logger.With(
+		zap.String(ZapKindKey, ZapKindProcessor),
+		zap.String(ZapNameKey, id.String()),
+		zap.String(ZapKindPipeline, pipelineID.String()))
+}
+
+func ExporterLogger(logger *zap.Logger, id component.ID, dt component.DataType) *zap.Logger {
+	return logger.With(
+		zap.String(ZapKindKey, ZapKindExporter),
```
bogdandrutu, line 39: I think the constants can be private now.

bogdandrutu (approved):

service/internal/components/loggers.go
djaglowski: Done. I also pulled in the `extensionLogger` function.

bogdandrutu (approved):

PR #6926, opened 2023-01-11
Parameterize connector data type errors

djaglowski:
Part of #6700

### DIFF
--- a/connector/connector.go
+++ b/connector/connector.go
@@ -24,19 +24,6 @@ import (
 	"go.opentelemetry.io/collector/consumer"
 )
 
-var (
-	errDataTypesNotSupported = "connection from %s to %s is not supported"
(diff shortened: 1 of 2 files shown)


jpkrohling (approved):

connector/connector.go
```diff
@@ -641,3 +628,7 @@ func logStabilityLevel(logger *zap.Logger, sl component.StabilityLevel) {
 		logger.Info(sl.LogMessage())
 	}
 }
+
+func errDataTypes(id component.ID, from, to component.DataType) error {
```
bogdandrutu, line 632: Based on https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L29 I think this should be:

```golang
	return fmt.Errorf("connector %q cannot connect from %s to %s: %w", id, from, to, component. ErrDataTypeIsNotSupported)
```

djaglowski: Thanks, I've updated this

bogdandrutu (approved):

PR #6927, opened 2023-01-11
Add Config method to pipeline component builders

djaglowski:
Part of #6700

This method is necessary so that connectors can be differentiated from receivers and exporters when building a component graph.


jpkrohling:
> This method is necessary so that connectors can be differentiated from receivers and exporters when building a component graph.

Could you please expand on that a bit? Not quite sure I understand the problem.

djaglowski:
Basically, we need to be able to check if a given component ID is a connector, as opposed to a receiver or exporter. 

To understand, note a couple rules about the config:
1. A `component.Type` is not necessarily unique to a class of components. `otlp` is a valid type for both a receiver or exporter.
2. Generally, a `component.ID` is not necessarily unique to a particular component. It is ok to have a receiver and exporter, each called `otlp/foo`.
3. With connectors, it is necessary to have a rule that you may _not_ have a connector that shares a `component.ID` with a receiver or exporter. Otherwise, the configuration would be ambiguous. For example, the following is valid, but would become ambiguous if the exporter and connector ID's were not differentiated:

```yaml
receivers:
  otlp:
exporters:
  foo:
connectors:
  foo/conn:

service:
  pipelines:
    logs/in:
      receivers: [nop]
      exporters: [foo, foo/conn] # [foo, foo] would be ambiguous
    logs/out:
      receivers: [foo/conn]
      exporters: [foo]
```

The ambiguity problem is the same at the code level. For each receiver and exporter, we need to establish whether it is actually a connector. All we have to work with is a `component.ID`, and a set of builders (`receiver.Builder`, `exporter.Builder`, `connector.Builder`) The `Factory` methods can tell us whether or not `otlp/foo` _can be_ a receiver, exporter, or connector, but we need the `Config` methods to check if it actually _is_ a connector vs the others.

bogdandrutu:
I understand the motivation, but would like to think if there is a way to avoid this. I have some ideas in mind:

* Define special place for "connectors" in the pipelines by adding 2 new entries "rconnectors" and "econnectors" (names can be improved)?
* Define a special case for connectors IDs in the pipelines, e.g. `exporters: [conntecor/foo, connector/foo/bar]`.
* Happy to hear other options to distinguish in the config between the conflicted type.

djaglowski:
@bogdandrutu, what do you think of my rationale?

jpkrohling:
I prefer your original proposal, @djaglowski.

jpkrohling (approved):
Given the alternatives, this is the cleanest solution with the best UX, in my opinion.

djaglowski:
I think then we can compare the two options like this:

Option 1: A connector cannot have the same `component.ID` as a receiver or exporter.
✅ Does not affect most users of connectors.
✅ No difference between declared connector ID and referenced connector ID.
❌ In rare cases user must resolve ID collision.

Option 2: Prepend `connector/` onto connector IDs when using them in pipelines.
❌ Affects all users of connectors.
❌ Implied relationship between declared connector ID and referenced connector ID.
✅ ID collisions are impossible.

To me, I still think option 1 is much cleaner.

exporter/exporter.go
```diff
@@ -276,6 +276,10 @@ func (b *Builder) CreateLogs(ctx context.Context, set CreateSettings) (Logs, err
 	return f.CreateLogsExporter(ctx, set, cfg)
 }
 
+func (b *Builder) Config(componentID component.ID) component.Config {
```
bogdandrutu, line 279: Independent of the config solution we discuss, I am curious why you need this:
1. Can you just check if a factory for the `componentID.Type()` is present?
2. I think you don't need access to the config. Should you then just have a `IsConfigured(componentId) bool` if the first solution is not enough?

djaglowski: I explained in more detail [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6927#issuecomment-1380799272), but the minimal need here is to identify the connectors in the `receivers` and `exporters` sections of each pipeline. 

1. This would work if a we had a guarantee that connector types cannot overlap with receiver or exporter types. This is possibly a reasonable rule, but I think we'd have to enforce this during factory registration. Not sure we care but I think this would be incompatible with an asynchronous/dynamic factory registration pattern. 
2. `IsConfigured` is enough. In fact, I only need it for connectors specifically. I can switch to that - I just thought it was oddly specific and that in the end supporting `Config` alongside `Factory` was natural.

I don't feel strongly about the implementation, so either of these approaches is fine with me.

djaglowski: Ok, I've reworked #6700 to prove to myself that `connector.Builder.IsConfigured` is the only thing absolutely necessary here, so I'm going to close this PR and include that in another PR that uses it.

djaglowski:
Closing in favor of minimal change described here: https://github.com/open-telemetry/opentelemetry-collector/pull/6927#discussion_r1085921377

Issue #6908, opened 2023-01-09
Release v0.69.0


djaglowski:
Like #4522, but for v0.69.0

**Performed by collector release manager**

- [x] #6911
- [x] #6919
- [x] #6912 
- [x] #6920
- [x] #6913
- [x] #6914
- [x] #6915
- [x] #6916

**Performed by operator maintainers**

- [ ] https://github.com/open-telemetry/opentelemetry-operator/issues/1356

**Performed by helm chart maintainers**

- [x] https://github.com/open-telemetry/opentelemetry-helm-charts/issues/587

MovieStoreGuy:
Hey @djaglowski ,

It looks like the main module wasn't updated to reference the modules of the same version:

https://github.com/open-telemetry/opentelemetry-collector/blob/v0.69.0/go.mod

It looks like the binary release worked because `cmd/otelcol` has a seperate module that is referencing `v0.69.0` excluding https://github.com/open-telemetry/opentelemetry-collector/blob/v0.69.0/cmd/otelcorecol/go.mod#L10 which I don't believe depends on the root project.

For those trying to reference things declared within the root module using `v0.69.0` are faced with the following error:

```
test imports
	go.opentelemetry.io/collector/pdata: module go.opentelemetry.io/collector/pdata@latest found (v0.66.0), but does not contain package go.opentelemetry.io/collector/pdata
test imports
	go.opentelemetry.io/collector/otelcol/otelcoltest imports
	go.opentelemetry.io/collector/otelcol imports
	go.opentelemetry.io/collector/service tested by
	go.opentelemetry.io/collector/service.test imports
	go.opentelemetry.io/collector/extension/zpagesextension imports
	go.opentelemetry.io/collector/config: module go.opentelemetry.io/collector@latest found (v0.69.0), but does not contain package go.opentelemetry.io/collector/config
```

I haven't tried it but for those who are facing this issue now, they be able to work around be adding a replace statement to their go.mod like so:

```
replace (
    go.opentelemetry.io/collector v0.68.0 => go.opentelemetry.io/collector v0.69.0
    go.opentelemetry.io/collector/pdata v1.0.0-rc2 => go.opentelemetry.io/collector/pdata v1.0.0-rc3
)
```

MovieStoreGuy:
It looks like the issue https://github.com/open-telemetry/opentelemetry-collector/issues/6929 is related to this.

PR #6933, opened 2023-01-12
Update all modules to use v0.69.0 of collector modules

djaglowski:
Fixes incorrect references described in https://github.com/open-telemetry/opentelemetry-collector/issues/6908#issuecomment-1380135220


consumer/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/stretchr/testify v1.8.1
-	go.opentelemetry.io/collector v0.68.0
+	go.opentelemetry.io/collector v0.69.0
 	go.opentelemetry.io/collector/pdata v1.0.0-rc2
```
mx-psi, line 8: We should also bump this to rc3, right?

component/go.mod
```diff
@@ -4,9 +4,8 @@ go 1.18
 
 require (
 	github.com/stretchr/testify v1.8.1
-	go.opentelemetry.io/collector v0.68.0
-	go.opentelemetry.io/collector/confmap v0.68.0
-	go.opentelemetry.io/collector/consumer v0.68.0
```
mx-psi, line 9: This one was removed, I guess that means that the module was not tidied before. This should not block the PR, but it's something we should investigate.

mx-psi (approved):

Issue #6940, opened 2023-01-13
Allow for passing ldflags to builder


kristinapathak:
**Is your feature request related to a problem? Please describe.**
I have written a collector component and it has its own version and build information separate from the collector that I want to pass in as `ldflag`s.  Currently, the builder binary doesn't have a way of accepting ldflags to pass to the `go build` command that it executes.

**Describe the solution you'd like**
`ldflag` added as an option to the builder command and passed through to the `go build` command.
https://github.com/open-telemetry/opentelemetry-collector/blob/f2a0133bf677b9fe2e18b7c02eb1b6bf5e849999/cmd/builder/internal/builder/main.go#L89

**Describe alternatives you've considered**
- Creating a small go program that sets up my component factories and then runs the collector directly.

cforce:
Can this be closed, as it seems it is merged, isn't it?

axw:
_A few moments later..._

PR #6946, opened 2023-01-14
add ldflags builder cmd option

kristinapathak:
**Description:** Adding a feature - `ldflags` option in builder cmd, to be passed to `go build`

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/6940

**Testing:** manually built a collector with custom components who have build variables.  Passing `ldflags` set the values of these variables.

**Documentation:** I added a description for the `--help` command, but I can add a small explanation to the README if people would prefer.

### DIFF
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -34,8 +34,9 @@ var ErrInvalidGoMod = errors.New("invalid gomod specification for module")
 // Config holds the builder's configuration
 type Config struct {
 	Logger          *zap.Logger
-	SkipCompilation bool `mapstructure:"-"`
-	SkipGetModules  bool `mapstructure:"-"`
+	SkipCompilation bool   `mapstructure:"-"`
+	SkipGetModules  bool   `mapstructure:"-"`
+	LDFlags         string `mapstructure:"-"`
 
 	Distribution Distribution `mapstructure:"dist"`
 	Exporters    []Module     `mapstructure:"exporters"`
--- a/cmd/builder/internal/builder/main.go
+++ b/cmd/builder/internal/builder/main.go
@@ -84,15 +84,14 @@ func Compile(cfg Config) error {
 		return nil
 	}
 
-	cfg.Logger.Info("Compiling")
-
 	var ldflags = "-s -w"
-
 	args := []string{"build", "-trimpath", "-o", cfg.Distribution.Name}
 	if cfg.Distribution.DebugCompilation {
 		cfg.Logger.Info("Debug compilation is enabled, the debug symbols will be left on the resulting binary")
-		ldflags = ""
+		ldflags = cfg.LDFlags
 		args = append(args, "-gcflags=all=-N -l")
+	} else if len(cfg.LDFlags) > 0 {
+		ldflags = fmt.Sprintf("%s %s", ldflags, cfg.LDFlags)
 	}
 	args = append(args, "-ldflags="+ldflags)
 	if cfg.Distribution.BuildTags != "" {
--- a/cmd/builder/internal/builder/main.go
+++ b/cmd/builder/internal/builder/main.go
@@ -94,7 +94,7 @@ func Compile(cfg Config) error {
 		ldflags = cfg.LDFlags
 		args = append(args, "-gcflags=all=-N -l")
 	} else if len(cfg.LDFlags) > 0 {
-		ldflags = fmt.Sprintf("%s %s", ldflags, cfg.LDFlags)
+		ldflags += " " + cfg.LDFlags
 	}
 	args = append(args, "-ldflags="+ldflags)
 	if cfg.Distribution.BuildTags != "" {
--- a/cmd/builder/internal/builder/main.go
+++ b/cmd/builder/internal/builder/main.go
@@ -84,7 +84,10 @@ func Compile(cfg Config) error {
 		return nil
 	}
 
+	cfg.Logger.Info("Compiling")
+
 	var ldflags = "-s -w"
+
 	args := []string{"build", "-trimpath", "-o", cfg.Distribution.Name}
 	if cfg.Distribution.DebugCompilation {
 		cfg.Logger.Info("Debug compilation is enabled, the debug symbols will be left on the resulting binary")
--- a/cmd/builder/internal/builder/main_test.go
+++ b/cmd/builder/internal/builder/main_test.go
@@ -83,6 +83,16 @@ func TestGenerateAndCompile(t *testing.T) {
 			},
 		},
 		{
+			testCase: "LDFlags Compilation",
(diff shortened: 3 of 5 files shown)


jpkrohling (commented):
The code looks alright, but this PR is missing the tests.

cmd/builder/internal/builder/main_test.go
```diff
@@ -80,6 +80,43 @@ func TestGenerateAndCompileDefault(t *testing.T) {
 	time.Sleep(1 * time.Second)
 }
 
+func TestGenerateAndCompileWithLDFlags(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping the test on Windows, see https://github.com/open-telemetry/opentelemetry-collector/issues/5403")
```
codeboten, line 85: This issue has been closed, is this skip still needed?

codeboten (commented):
Thanks for this change, just one question

cmd/builder/internal/builder/main_test.go
kristinapathak: I'm not sure - I'll try removing it.  It's in other main_test.go tests so I will be removing it there as well.

kristinapathak: It still failed so I reverted.  It seems like we still need to skip these tests on windows.

codeboten (approved):

cmd/builder/internal/builder/config.go
```diff
@@ -34,8 +34,9 @@ var ErrInvalidGoMod = errors.New("invalid gomod specification for module")
 // Config holds the builder's configuration
 type Config struct {
 	Logger          *zap.Logger
-	SkipCompilation bool `mapstructure:"-"`
-	SkipGetModules  bool `mapstructure:"-"`
+	SkipCompilation bool   `mapstructure:"-"`
+	SkipGetModules  bool   `mapstructure:"-"`
+	LDFlags         string `mapstructure:"-"`
```
bogdandrutu, line 39: How do you expect this to be configured?

kristinapathak: As a command line argument. Ex:
```
builder --ldflags='-X "<pkg>.gitVersion=d49d45c039a2a89a26526afa3658de6429099bef" -X "<pkg>.gitTag=local-testing" -X "<pkg>.whoami=kristina.pathak"'
```

cmd/builder/internal/builder/main.go
```diff
@@ -91,8 +91,10 @@ func Compile(cfg Config) error {
 	args := []string{"build", "-trimpath", "-o", cfg.Distribution.Name}
 	if cfg.Distribution.DebugCompilation {
 		cfg.Logger.Info("Debug compilation is enabled, the debug symbols will be left on the resulting binary")
-		ldflags = ""
+		ldflags = cfg.LDFlags
 		args = append(args, "-gcflags=all=-N -l")
+	} else if len(cfg.LDFlags) > 0 {
+		ldflags = fmt.Sprintf("%s %s", ldflags, cfg.LDFlags)
```
bogdandrutu, line 97:
```suggestion
		ldflags += " " + cfg.LDFlags
```

cmd/builder/internal/builder/config.go
bogdandrutu: where is the "--ldflags" flag defined?

bogdandrutu, line 39: This should be part of "Distribution" and we should not have a flag but be part of the yaml config.

kristinapathak: In `cmd/builder/internal/command.go`:
https://github.com/open-telemetry/opentelemetry-collector/pull/6946/files#diff-6019321caab5c53bd0ae7d59974bf883a6101939b98e3ca4d585c7018fa34207R90

kristinapathak: Why?

bogdandrutu: Because there we have other flags, and because we want to go with "configuration as a code". Which means to pass the configuration to the builder as a config file merged into the code repository.

kristinapathak: In my example [here](https://github.com/open-telemetry/opentelemetry-collector/pull/6946/files#r1080692017), you can see that ldflags include git information.  The moment these values are committed and pushed, they are incorrect since it will be a new commit hash with new git tags (our tags include datetimes). If ldflags become a configuration value, my automation would have to add it as part of the build process, and the value won't be able to be committed with the rest of the configuration file.

cmd/builder/internal/builder/main_test.go
```diff
@@ -80,6 +80,43 @@ func TestGenerateAndCompileDefault(t *testing.T) {
...
+	cfg.Replaces = append(cfg.Replaces, fmt.Sprintf("go.opentelemetry.io/collector/processor/memorylimiterprocessor => %s/processor/memorylimiterprocessor", workspaceDir))
+	cfg.Replaces = append(cfg.Replaces, fmt.Sprintf("go.opentelemetry.io/collector/receiver/otlpreceiver => %s/receiver/otlpreceiver", workspaceDir))
+	cfg.Replaces = append(cfg.Replaces, fmt.Sprintf("go.opentelemetry.io/collector/pdata => %s/pdata", workspaceDir))
+	cfg.Replaces = append(cfg.Replaces, fmt.Sprintf("go.opentelemetry.io/collector/semconv => %s/semconv", workspaceDir))
+
+	assert.NoError(t, cfg.Validate())
+	assert.NoError(t, cfg.SetGoPath())
+	require.NoError(t, GenerateAndCompile(cfg))
+
+	// Sleep for 1 second to make sure all processes using the files are completed
```
jpkrohling, line 115: What do you mean?

kristinapathak: This logic was taken from pre-existing main tests:
https://github.com/open-telemetry/opentelemetry-collector/blame/80cabdd33c50eaf386293f90fd8ac57cb3a10dc9/cmd/builder/internal/builder/main_test.go#L106
I see that now they have been refactored so will rebase and include my test in the table, but this logic still exists in those tests.

kristinapathak:
`TestServiceTelemetryRestart` has [failed](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/3999205223/jobs/6862825101#step:6:1267) in the test-coverage github action, which is unrelated to my changes in this PR.

jpkrohling:
I just triggered the failed test again.

cmd/builder/internal/builder/config.go
kristinapathak: @bogdandrutu, would you be open to moving this value under "Distribution" in the yaml config and allowing it to be set through a flag?

PR #6958, opened 2023-01-17
[chore] Retract failed release v0.69.0

MovieStoreGuy:
**Description:**
Ensuring users can't use v0.69.0 since it failed, should use fixed version v0.69.1

**Link to tracking Issue:** 
https://github.com/open-telemetry/opentelemetry-collector/issues/6908

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -96,7 +96,7 @@ replace go.opentelemetry.io/collector/pdata => ./pdata
 replace go.opentelemetry.io/collector/extension/zpagesextension => ./extension/zpagesextension
 
 retract (
-	v0.69.0 // Release failed, use v0.69.1 + v1.0.0-rc3
+	v0.69.0 // Release failed, use v0.69.1 
 	v0.57.1 // Release failed, use v0.57.2
 	v0.57.0 // Release failed, use v0.57.2
 	v0.32.0 // Contains incomplete metrics transition to proto 0.9.0, random components are not working.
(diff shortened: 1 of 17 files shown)


MovieStoreGuy:
Not sure if pdata `v1.0.0-rc2` should also be included.

dmitryax:
> Not sure if pdata v1.0.0-rc2 should also be included.

pdata modules are all good

go.mod
```diff
@@ -96,6 +96,7 @@ replace go.opentelemetry.io/collector/pdata => ./pdata
 replace go.opentelemetry.io/collector/extension/zpagesextension => ./extension/zpagesextension
 
 retract (
+	v0.69.0 // Release failed, use v0.69.1 + v1.0.0-rc3
```
dmitryax, line 99:
```suggestion
	v0.69.0 // Release failed, use v0.69.1 
```

dmitryax (commented):
I believe it should be added in all beta modules

MovieStoreGuy:
> I believe it should be added in all beta modules

I will add them in a moment, so it is basically all modules that aren't pdata right?

dmitryax:
> I will add them in a moment, so it is basically all modules that aren't pdata right?

Yes

MovieStoreGuy:
I couldn't retract `v0.69.0` from `go.opentelemetry.io/collector/cmd/otelcorecol` since the builder does't support `retract`.

Not sure if that is going to be an issue.

dmitryax:
> since the builder does't support retract.

What do you mean? How doesn't it support retract?

bogdandrutu (approved):

bogdandrutu:
> What do you mean? How doesn't it support retract?

The go.mod file is generated by the otelcol builder, and that does not support retract statements.

PR #6960, opened 2023-01-17
add github action to automate the release parts of the process

codeboten:
This PR adds a github action that can be triggered manually to do the following:

- creates tracking issue
- checks for blockers in core
- checks for blockers in contrib
- checks status of build-and-test in core
- checks status of build-and-test in contrib
- runs chlog update
- runs make-prepare for both stable/beta
- creates a PR

See the tracking issue created [here](https://github.com/codeboten/opentelemetry-collector/issues/175)
You can see the PR produced [here](https://github.com/codeboten/opentelemetry-collector/pull/176)

Running the workflow looks like this:
Screen Shot 2023-01-17 at 10 41 17 AM

Signed-off-by: Alex Boten <aboten@lightstep.com>

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -33,7 +33,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CANDIDATE_BETA: ${{ inputs.candidate-beta }}
           CANDIDATE_STABLE: ${{ inputs.candidate-stable }}
-          REPO: codeboten/opentelemetry-collector
+          REPO: open-telemetry/opentelemetry-collector
         run: |
           EXISTING_ISSUE=$( gh issue list --search "Release v${CANDIDATE_BETA}/v${CANDIDATE_STABLE}" --json url --jq '.[].url' --repo ${REPO} )
           if [ "${EXISTING_ISSUE}" != "" ]; then
@@ -132,7 +132,6 @@ jobs:
           CURRENT_BETA: ${{ inputs.current-beta }}
           CURRENT_STABLE: ${{ inputs.current-stable }}
         run: |
-          make chlog-install
           make chlog-update VERSION=${CANDIDATE_STABLE}/${CANDIDATE_BETA}
           git config user.name "$GITHUB_ACTOR"
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -8,6 +8,7 @@ on:
       candidate-stable:
         required: true
         description: Release candidate version (stable, like 1.0.0-rc4)
+
       current-stable:
         required: true
         description: Current version (stable, like 1.0.0-rc3)
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -105,15 +105,6 @@ jobs:
               echo "$(gh run list --branch main --json status,url --jq '[.[] | select(.status != "queued" and .status != "in_progress")][0].url' --workflow build-and-test --repo ${REPO} )"
               exit 1
           fi
-      # Update Contrib to use the latest in development version of Core. Run make update-otel in
-      # Contrib root directory and if it results in any changes, submit a PR. Open this PR as draft.
-      # This is to ensure that the latest core does not break contrib in any way. We’ll update it
-      # once more to the final release number later.
-      - name: Make update-otel in contrib
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          REPO: open-telemetry/opentelemetry-collector-contrib
-        run: "echo TODO" # TODO: This may require a webhook in contrib
       - uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v3
@@ -148,9 +139,3 @@ jobs:
           - make prepare-release GH=none PREVIOUS_VERSION=${CURRENT_STABLE} RELEASE_CANDIDATE=${CANDIDATE_STABLE} MODSET=stable
           - make prepare-release GH=none PREVIOUS_VERSION=${CURRENT_BETA} RELEASE_CANDIDATE=${CANDIDATE_BETA} MODSET=beta
           "
-
-# Still TODO manually:
-#     Create a branch named release/<release-series> (e.g. release/v0.45.x) from the changelog update commit and push to open-telemetry/opentelemetry-collector.
-#     Tag all the module groups (stable, beta) with the new release version by running the make push-tags command (e.g. make push-tags MODSET=stable and make push-tags MODSET=beta). Wait for the new tag build to pass successfully.
-#     The release script for the collector builder should create a new GitHub release for the builder. This is a separate release from the core, but we might join them in the future if it makes sense.
-#     A new v0.55.0 release should be automatically created on Github by now. Edit it and use the contents from the CHANGELOG.md as the release's description. At the top of the release's changelog, add a link to the releases repository where the binaries and other artifacts are landing, like:
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -13,10 +13,10 @@ on:
         description: Current version (stable)
       candidate-beta:
         required: true
-        description: Release candidate version (stable)
+        description: Release candidate version (beta)
       current-beta:
         required: true
-        description: Current version (stable)
+        description: Current version (beta)
 jobs:
   # Releasing opentelemetry-collector
   prepare-release:
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,154 @@
+name: Automation - Prepare Release
+
+on:
+  workflow_dispatch:
+    # Determine the version number that will be assigned to the release. During the beta phase, we increment
+    # the minor version number and set the patch number to 0.
+    inputs:
+      candidate-stable:
+        required: true
+        description: Release candidate version (stable)
+      current-stable:
+        required: true
+        description: Current version (stable)
+      candidate-beta:
+        required: true
+        description: Release candidate version (stable)
+      current-beta:
+        required: true
+        description: Current version (stable)
+jobs:
+  # Releasing opentelemetry-collector
+  prepare-release:
+    runs-on: ubuntu-latest
+    steps:
+      # To keep track of the progress, it might be helpful to create a tracking issue similar to #6067. You are responsible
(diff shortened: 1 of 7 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+    inputs:
+      candidate-stable:
+        required: true
+        description: Release candidate version (stable)
+      current-stable:
+        required: true
+        description: Current version (stable)
+      candidate-beta:
+        required: true
+        description: Release candidate version (stable)
```
djaglowski, line 16:
```suggestion
        description: Release candidate version (beta)
```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+          fi
+      # Update Contrib to use the latest in development version of Core. Run make update-otel in
+      # Contrib root directory and if it results in any changes, submit a PR. Open this PR as draft.
+      # This is to ensure that the latest core does not break contrib in any way. We’ll update it
+      # once more to the final release number later.
+      - name: Make update-otel in contrib
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: open-telemetry/opentelemetry-collector-contrib
+        run: "echo TODO" # TODO: This may require a webhook in contrib
```
MovieStoreGuy, line 113: Should this be done as part of this PR? Or is it fine to make it part of a future PR?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+      - name: Prepare release for core
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: open-telemetry/opentelemetry-collector
+          CANDIDATE_BETA: ${{ inputs.candidate-beta }}
+          CANDIDATE_STABLE: ${{ inputs.candidate-stable }}
+          CURRENT_BETA: ${{ inputs.current-beta }}
+          CURRENT_STABLE: ${{ inputs.current-stable }}
+        run: |
+          make chlog-install
```
MovieStoreGuy, line 132: With the changes made to tooling, you shouldn't need to run `*install` anymore. Running the commands should check if the required command is built and build as required.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+          make prepare-release GH=none PREVIOUS_VERSION=${CURRENT_BETA} RELEASE_CANDIDATE=${CANDIDATE_BETA} MODSET=beta
+          git push origin ${BRANCH}
+          gh pr create --title "[chore] Prepare release ${CANDIDATE_BETA}/${CANDIDATE_STABLE}" --body "
+          The following commands were run to prepare this release:
+          - make chlog-update VERSION=${CANDIDATE_BETA}/${CANDIDATE_STABLE}
+          - make prepare-release GH=none PREVIOUS_VERSION=${CURRENT_STABLE} RELEASE_CANDIDATE=${CANDIDATE_STABLE} MODSET=stable
+          - make prepare-release GH=none PREVIOUS_VERSION=${CURRENT_BETA} RELEASE_CANDIDATE=${CANDIDATE_BETA} MODSET=beta
+          "
+
+# Still TODO manually:
```
MovieStoreGuy, line 150: Since this is added in the release docs, does need to be documented here also?

MovieStoreGuy (commented):
I have a few questions on this:

- What happens if someone chooses an existing version?
- Would it be possible to offer a dry run mode (similar to how terraform plan would do)

On my last question, it does feel a little wasteful and I am happy for it not to be included. I just really like the idea that releases have a reviewed process to help validate the release.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
+name: Automation - Prepare Release
+
+on:
+  workflow_dispatch:
+    # Determine the version number that will be assigned to the release. During the beta phase, we increment
+    # the minor version number and set the patch number to 0.
+    inputs:
+      candidate-stable:
+        required: true
+        description: Release candidate version (stable)
```
jpkrohling, line 10: Can you add an example, to make abundantly clear which kind of version this is about?

```suggestion
        description: Release candidate version (stable, like v1.0.0-rc4)
```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+      current-stable:
+        required: true
+        description: Current version (stable)
```
jpkrohling, line 13:
```suggestion
        description: Current version (stable, like v1.0.0-rc3)

```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+      candidate-beta:
+        required: true
+        description: Release candidate version (beta)
```
jpkrohling, line 16:
```suggestion
        description: Release candidate version (beta, like v0.70.0)

```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+      current-beta:
+        required: true
+        description: Current version (beta)
```
jpkrohling, line 19:
```suggestion
        description: Current version (beta, like v0.69.1)
```

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+      # To keep track of the progress, it might be helpful to create a tracking issue similar to #6067. You are responsible
+      # for all of the steps under the "Performed by collector release manager" heading. Once the issue is created, you can
+      # create the individual ones by hovering them and clicking the "Convert to issue" button on the right hand side.
+      - name: Create issue for tracking release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CANDIDATE_BETA: ${{ inputs.candidate-beta }}
+          CANDIDATE_STABLE: ${{ inputs.candidate-stable }}
+          REPO: codeboten/opentelemetry-collector
+        run: |
```
jpkrohling, line 34: How do you feel about having this in an actual source file?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+    steps:
+      # To keep track of the progress, it might be helpful to create a tracking issue similar to #6067. You are responsible
+      # for all of the steps under the "Performed by collector release manager" heading. Once the issue is created, you can
+      # create the individual ones by hovering them and clicking the "Convert to issue" button on the right hand side.
+      - name: Create issue for tracking release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CANDIDATE_BETA: ${{ inputs.candidate-beta }}
+          CANDIDATE_STABLE: ${{ inputs.candidate-stable }}
+          REPO: codeboten/opentelemetry-collector
```
jpkrohling, line 33: Do you want to have your repo here for the first version, or will you change before merging?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+        run: |
+          EXISTING_ISSUE=$( gh issue list --search "Release v${CANDIDATE_BETA}/v${CANDIDATE_STABLE}" --json url --jq '.[].url' --repo ${REPO} )
+          if [ "${EXISTING_ISSUE}" != "" ]; then
+            echo "Issue already exists: ${EXISTING_ISSUE}"
+            exit 0
+          fi
+          gh issue create -a ${GITHUB_ACTOR} --repo ${REPO} --label release --title "Release v${CANDIDATE_BETA}/v${CANDIDATE_STABLE}" --body "Like #4522, but for v${CANDIDATE_BETA}/v${CANDIDATE_STABLE}
+          **Performed by collector release manager**
+
+          - [ ] Prepare stable core release v${CANDIDATE_STABLE}
```
jpkrohling, line 43: The release manager would then still be required to create the individual issues, right?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,154 @@
...
+
+          **Performed by operator maintainers**
+
+          - [ ] Release the operator v${CANDIDATE_BETA}
+
+          **Performed by helm chart maintainers**
+
+          - [ ] Update the opentelemetry-collector helm chart to use v${CANDIDATE_BETA}"
+      # Make sure that there are no open issues with release:blocker label in Core. The release has to be delayed until they are resolved.
+      - name: Check blockers in core
```
jpkrohling, line 60: Should this happen even before creating the tracker issue?

.github/workflows/prepare-release.yml
jpkrohling, line 113: Or perhaps an approver's PAT as a secret?

.github/workflows/prepare-release.yml
jpkrohling, line 150: Is this pending because of technical challenges? Or because it just wasn't done yet?

.github/workflows/prepare-release.yml
codeboten: This has been removed.

.github/workflows/prepare-release.yml
codeboten: Unless we needed to re-use this code, I'm not sure if there's a value in having a separate script file, though I don't have a strong opinion on this.

.github/workflows/prepare-release.yml
codeboten: I've updated it

.github/workflows/prepare-release.yml
codeboten: I think it should be fine either ways, since the issue will only get created once.

.github/workflows/prepare-release.yml
codeboten: Removed

.github/workflows/prepare-release.yml
codeboten: I've removed it, the release note capture this step anyways, it can be automated later.

.github/workflows/prepare-release.yml
codeboten: That's correct. My hope is that through automation, the need for the tracking issue will be limited to an audit trail rather than a comprehensive reminder list :)

.github/workflows/prepare-release.yml
codeboten: It's possible, I've removed that step for now as it's still documented in the release doc as manual

.github/workflows/prepare-release.yml
jpkrohling: Being in a script means it could be used directly in case of necessity. It would also allow us to use a linter, like the shellcheck linter that is/was proposed as part of another PR.

jpkrohling (approved):

PR #6964, opened 2023-01-18
Add connectors to otelcol config

djaglowski:
Part of #6700 

This PR adds connectors to the `otelcol.Config` and updates associated config provider and unmarshaler.

### DIFF
--- a/otelcol/config.go
+++ b/otelcol/config.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/featuregate"
 	"go.opentelemetry.io/collector/service"
 )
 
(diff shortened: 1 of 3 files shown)


otelcol/config.go
```diff
@@ -38,6 +38,9 @@ type Config struct {
 	// Processors is a map of ComponentID to Processors.
 	Processors map[component.ID]component.Config
 
+	// Connectors is a map of ComponentID to connectors.
+	Connectors map[component.ID]component.Config
```
bogdandrutu, line 42: Let's add the "featuregate" for the connectors now, and in the Config.Validate fail if len(Connectors) != 0 and the gate is disable. The reason is because I want to make sure we fail fast and people don't try to use connectors if they are not fully implemented/enabled.

djaglowski: I've added the feature gate, and a check in the validate func.

codeboten (commented):
Change looks good, needs a rebase

djaglowski:
@codeboten, thanks for the review. Rebased.

jpkrohling (approved):

Issue #6686, opened 2022-12-06
Handle partial success responses from OTLP export services


joaopgrassi:
### Problem Statement

Following the https://github.com/open-telemetry/opentelemetry-specification/pull/2696, each OTel SDK is encouraged to handle the resulting error message string in an appropriate way, considering existing norms. This requires OTLP v0.19. If I got it correct, this PR already added support for it in pdata https://github.com/open-telemetry/opentelemetry-collector/pull/6361

### Proposed Solution

For each of the `Export*ServiceResponse` methods used by OTLP Trace and Metrics exporters, construct an error and call the appropriate handler with the error message string and the number of spans/points(/logs) dropped. Future OTel specifications may call for optional treatment of the number of dropped items, but presently that is just additional information to include in the handled error.

For context, here's a few SDKs that already implemented handling partial success responses: [Go](https://github.com/open-telemetry/opentelemetry-go/issues/3102), [PHP](https://github.com/open-telemetry/opentelemetry-php/issues/802)

cc @jmacd @tigrannajaryan @evan-bradley

evan-bradley:
I'll work on implementing this.

jmacd:
I would say that ideally the return from a PartialSuccess would factor into the counts of dropped points/spans/logs vs accepted points/spans/logs. Was this considered? The way the PRs #6971 and #6970 were written, these become total errors and the result is `exporterhelper` counts them as such, meaning:

1. The metrics are wrong in a way that PartialSuccess was meant to preserve.
2. The producer of the metrics doesn't get propagated results telling them how many of their points dropped.

I think the collector's metrics should factor the meaning of PartialSuccess in and also propagate the partial success to the original producer. Can we leave this open, hoping for improvements?

PR #6970, opened 2023-01-19
Handle partial success responses in the OTLP HTTP exporter

evan-bradley:
**Description:**

Handle partial success messages returned from OTLP HTTP backends.

**Link to tracking Issue:**

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6686

**Testing:**

Unit tests. I was able to manually test partial success messages when submitting metrics to a vendor backend. I'd welcome any additional places to test this.

### DIFF
--- a/exporter/otlphttpexporter/otlp.go
+++ b/exporter/otlphttpexporter/otlp.go
@@ -133,12 +133,7 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
 	}()
 
 	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
-		if err := handlePartialSuccessResponse(resp, partialSuccessHandler); err != nil {
-			return err
-		}
-
-		// Request is successful.
-		return nil
+		return handlePartialSuccessResponse(resp, partialSuccessHandler)
 	}
 
 	respStatus := readResponseStatus(resp)
--- a/exporter/otlphttpexporter/otlp.go
+++ b/exporter/otlphttpexporter/otlp.go
@@ -141,7 +141,7 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
 		return nil
 	}
 
-	respStatus := readResponse(resp)
+	respStatus := readResponseStatus(resp)
 
 	// Format the error message. Use the status if it is present in the response.
 	var formattedErr error
@@ -192,27 +192,49 @@ func isRetryableStatusCode(code int) bool {
 	}
 }
 
+func readResponseBody(body io.ReadCloser) ([]byte, error) {
+	// Read the maximum number of bytes allowed in a request. This avoids
+	// issues with missing or invalid Content-Length headers.
+	protoBytes := make([]byte, maxHTTPResponseReadBytes)
+	n, err := io.ReadFull(body, protoBytes)
+
+	// No bytes read and an EOF error indicates there is no body to read.
+	if n == 0 && (err == nil || errors.Is(err, io.EOF)) {
+		return nil, nil
+	}
+
+	// io.ReadFull will return io.ErrorUnexpectedEOF in most cases since there
+	// will usually be a mismatch between the length of the byte slice and the
+	// size of the body, so we ignore that error.
+	if err != nil && !errors.Is(err, io.ErrUnexpectedEOF) {
+		return nil, err
+	}
+
+	// The pdata unmarshaling methods check for the length of the slice
+	// when unmarshaling it, so we have to trim down the length to the
+	// actual size of the data.
+	return protoBytes[:n], nil
+}
+
 // Read the response and decode the status.Status from the body.
 // Returns nil if the response is empty or cannot be decoded.
-func readResponse(resp *http.Response) *status.Status {
+func readResponseStatus(resp *http.Response) *status.Status {
 	var respStatus *status.Status
 	if resp.StatusCode >= 400 && resp.StatusCode <= 599 {
 		// Request failed. Read the body. OTLP spec says:
 		// "Response body for all HTTP 4xx and HTTP 5xx responses MUST be a
 		// Protobuf-encoded Status message that describes the problem."
-		maxRead := resp.ContentLength
-		if maxRead == -1 || maxRead > maxHTTPResponseReadBytes {
-			maxRead = maxHTTPResponseReadBytes
+		respBytes, err := readResponseBody(resp.Body)
+
+		if err != nil {
+			return nil
 		}
-		respBytes := make([]byte, maxRead)
-		n, err := io.ReadFull(resp.Body, respBytes)
-		if err == nil && n > 0 {
-			// Decode it as Status struct. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#failures
-			respStatus = &status.Status{}
-			err = proto.Unmarshal(respBytes, respStatus)
-			if err != nil {
-				respStatus = nil
-			}
+
+		// Decode it as Status struct. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#failures
+		respStatus = &status.Status{}
+		err = proto.Unmarshal(respBytes, respStatus)
+		if err != nil {
+			return nil
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


exporter/otlphttpexporter/otlp.go
```diff
@@ -235,3 +239,81 @@ func readResponse(resp *http.Response) *status.Status {
...
+
+func tracesPartialSuccessHandler(protoBytes []byte) error {
+	exportResponse := ptraceotlp.NewExportResponse()
+	err := exportResponse.UnmarshalProto(protoBytes)
+	if err != nil {
+		return err
+	}
+	partialSuccess := exportResponse.PartialSuccess()
+	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedSpans() == 0) {
+		return consumererror.NewPermanent(fmt.Errorf("OTLP partial success: %s (%d rejected)", partialSuccess.ErrorMessage(), partialSuccess.RejectedSpans()))
```
joaopgrassi, line 290: The spec also says servers can use the partial success to convey warnings/suggestions, even when all spans/metrics/logs were ingested: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#partial-success

>Servers MAY also make use of the partial_success field to convey warnings/suggestions to clients even when the request was fully accepted. In such cases, the rejected_<signal> field MUST have a value of 0 and the error_message field MUST be non-empty.

With the current handling we would miss such cases right?

joaopgrassi: The same applies to gRPC.

joaopgrassi: Never mind, I get it now. You can mark it as resolved!

djaglowski (approved):
LGTM

djaglowski:
I think we need a rebase in order to pass the mandatory 1.20 check.

exporter/otlphttpexporter/otlp.go
```diff
@@ -228,3 +232,81 @@ func readResponse(resp *http.Response) *status.Status {
...
+	// is the case, the body will still have the full message in it, so we want to
+	// ignore the error and parse the message.
+	if err != nil && !errors.Is(err, io.ErrUnexpectedEOF) {
+		return err
+	}
+
+	// The pdata unmarshaling methods check for the length of the slice
+	// when unmarshaling it, so we have to trim down the length to the
+	// actual size of the data.
+	if needsResize {
```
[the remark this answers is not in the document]
evan-bradley: The goal here was to use the Content Length header if it was present, otherwise try to adapt to an invalid/missing header. If `resp.ContentLength != n`, then the proto unmarshaling will fail. Since we're limiting the body to 64 KiB, I'm fine just allocating the byte slice and resizing it, which would also remove the check for the Content Length header.

exporter/otlphttpexporter/otlp_test.go
```diff
@@ -677,3 +641,231 @@ func TestUserAgent(t *testing.T) {
 		}
 	})
 }
+
+func TestPartialSuccess(t *testing.T) {
+	addr := testutil.GetAvailableLocalAddress(t)
+	set := exportertest.NewNopCreateSettings()
+
+	t.Run("traces", func(t *testing.T) {
```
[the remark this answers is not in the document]
evan-bradley: My goal was to group them into suites, but I'm good with that. Done.

exporter/otlphttpexporter/otlp_test.go
```diff
@@ -677,3 +641,231 @@ func TestUserAgent(t *testing.T) {
...
+				Body:          io.NopCloser(body),
+			}
+			err := handlePartialSuccessResponse(resp, tt.handler)
+			assert.Error(t, err)
+		})
+	}
+
+}
+
+func createBackend(addr string, endpoint string, handler func(writer http.ResponseWriter, request *http.Request)) (*http.Server, error) {
```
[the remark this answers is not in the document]
evan-bradley: Thanks for the tip. I've adapted the tests to use `httptest.Server`, but it felt like a wash overall since the server URL is used when checking error conditions. Let me know what you think.

exporter/otlphttpexporter/otlp.go
```diff
@@ -228,3 +232,81 @@ func readResponse(resp *http.Response) *status.Status {
 
 	return respStatus
 }
+
+func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler partialSuccessHandler) error {
```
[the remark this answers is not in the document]
evan-bradley: Sure, most of the core logic should be in the `readResponseBody` function now.

evan-bradley:
@dmitryax Could you please take a look at this one as well? I believe this PR is also ready to go.

exporter/otlphttpexporter/otlp.go
```diff
@@ -199,29 +203,102 @@ func isRetryableStatusCode(code int) bool {
...
-func readResponse(resp *http.Response) *status.Status {
+func readResponseStatus(resp *http.Response) *status.Status {
 	var respStatus *status.Status
 	if resp.StatusCode >= 400 && resp.StatusCode <= 599 {
 		// Request failed. Read the body. OTLP spec says:
 		// "Response body for all HTTP 4xx and HTTP 5xx responses MUST be a
 		// Protobuf-encoded Status message that describes the problem."
-		maxRead := resp.ContentLength
-		if maxRead == -1 || maxRead > maxHTTPResponseReadBytes {
-			maxRead = maxHTTPResponseReadBytes
```
dmitryax, line 212: Why is this logic dropped in `readResponseBody`? I don't think we want to allocate maxHTTPResponseReadBytes's long slices every time

exporter/otlphttpexporter/otlp.go
```diff
@@ -199,29 +203,102 @@ func isRetryableStatusCode(code int) bool {
...
+	// The pdata unmarshaling methods check for the length of the slice
+	// when unmarshaling it, so we have to trim down the length to the
+	// actual size of the data.
+	return protoBytes[:n], nil
+}
+
 // Read the response and decode the status.Status from the body.
 // Returns nil if the response is empty or cannot be decoded.
-func readResponse(resp *http.Response) *status.Status {
+func readResponseStatus(resp *http.Response) *status.Status {
```
dmitryax, line 232: Are changes in this function related to handling partial request?

exporter/otlphttpexporter/otlp.go
evan-bradley: This was removed in response to earlier feedback: https://github.com/open-telemetry/opentelemetry-collector/pull/6970#discussion_r1143533440. The logic for determining the correct slice size is a little complicated, and 64 KiB doesn't feel like a substantial amount of memory. If there's consensus on allocating smaller slices, I'm fine with that. We could also allocate a single maxHTTPResponseReadBytes-sized slice (or pool of slices) within this component and reuse it if that would be preferable.

exporter/otlphttpexporter/otlp.go
evan-bradley: Not directly, this was a change requested in earlier feedback around some code duplication that was introduced: https://github.com/open-telemetry/opentelemetry-collector/pull/6970#discussion_r1143536486. We can make this change in a follow-up PR if you'd prefer, I'm fine either way.

exporter/otlphttpexporter/otlp.go
dmitryax: What's the typical OTLP response size?

I think the comment was about introducing `needsResize` not calculating the slice size. If the typical OTLP response size is much smaller than 64k, I don't think we should always allocate 64k.

evan-bradley: I would estimate that for most cases it's well below 1 KiB, the largest responses will likely be partial success responses with details in the message. I'm also fine with saving memory where we can, I've added back the logic to use the Content-Length header to size the slice.

I re-read the comment and you're right, I think I misread it. Providing a smaller slice through re-slicing is an extremely cheap operation anyway now that I'm thinking of it, so I've removed the `needsResize` variable.

exporter/otlphttpexporter/otlp.go
```diff
@@ -133,11 +133,15 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte) e
 	}()
 
 	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
+		if err := handlePartialSuccessResponse(resp, partialSuccessHandler); err != nil {
+			return err
+		}
+
 		// Request is successful.
 		return nil
```
dmitryax, line 141:
```suggestion
		return handlePartialSuccessResponse(resp, partialSuccessHandler)
```

dmitryax (approved):
LGTM. Just one minor suggesting

PhilWillMill:
LGTM?  We're currently hitting partial success responses and this will definitely help us investigate

PR #6971, opened 2023-01-19
Handle partial success responses in the OTLP exporter

evan-bradley:
**Description:**

Handle partial success messages returned from OTLP gRPC backends.

**Link to tracking Issue:**

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6686

**Testing:**

Unit tests. I haven't been able to test this against any live backends, any recommendations would be welcome.


exporter/otlpexporter/otlp.go
```diff
@@ -98,20 +98,41 @@ func (e *baseExporter) shutdown(context.Context) error {
...
 	req := ptraceotlp.NewExportRequestFromTraces(td)
-	_, err := e.traceExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
-	return processError(err)
+	resp, respErr := e.traceExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
+	if err := processError(respErr); err != nil {
+		return err
+	}
+	partialSuccess := resp.PartialSuccess()
+	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedSpans() == 0) {
+		return consumererror.NewPermanent(fmt.Errorf("OTLP partial success: \"%s\" (%d rejected)", resp.PartialSuccess().ErrorMessage(), resp.PartialSuccess().RejectedSpans()))
```
bogdandrutu, line 107: I am not the biggest fun of this "text like" error. Should we return a "structured" error where we can have access to the "rejected" value in case we want to record metrics for example.

One option is to change Permanent error to include a "failed" int value.

evan-bradley: I agree that providing some structure would be beneficial here.

One issue I see with making a `PermanentTraces` type that is completely equivalent to `consumererror.Traces` is that at least with OTLP partial success messages, we can't reliably determine which spans failed. One way around this could be to use the options pattern for adding any metadata on top of the wrapped error (failed span count, failed spans, etc.), and the caller can add what it has.

Alternatively, just adding a `failed` field to the Permanent error type might suffice. Access to the actual spans from `consumererror.Traces` is only used to retry exports from what I can tell, so the spans themselves may not be useful in the case of a permanent error.

evan-bradley: I think this could be covered by https://github.com/open-telemetry/opentelemetry-collector/issues/7047 and improved here in a follow-up PR. Would that work for you?

joaopgrassi: It would be really great to have _something_, even though it's not perfect for now I feel a text-like message is already much better than not getting any feedback on what went wrong and how I can fix it.

evan-bradley:
@bogdandrutu @jpkrohling could one of you please take a look?

I will look into the decreased test coverage again today. Last time I checked, it wanted me to test what was essentially an impossible condition, but I will see if I can cover it.

jpkrohling:
@bogdandrutu seems to have had concerns with the log statement. Perhaps @dmitryax or @codeboten could take a look as well, in case @bogdandrutu is not available?

exporter/otlpexporter/otlp.go
codeboten: Would a `NewPartial` with an int represented failed/rejected data points be an option? Since returning a `NewPermanent` doesn't feel like the right thing here.

evan-bradley: @codeboten Could you elaborate more on why a permanent error wouldn't be a good fit here?

The reason I've used it so far is because the [OTLP specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#partial-success) says that clients should not retry requests after receiving partial success responses, which makes a permanent error seem appropriate here. That said, I think a `Partial` error type would still work so long it can be considered non-retryable when exporting OTLP.

If it would help, I've created a draft PR related to this conversation that shows how I think the consumer errors could be updated to help handle partial successes: https://github.com/open-telemetry/opentelemetry-collector/pull/7439.

codeboten: After re-reading the spec and our definition of Permanent, i think fine to use it as @bogdandrutu suggests. The approach in #7439 makes sense to me.

dmitryax:
@evan-bradley do you think we need to merge this now or wait until we have a long-term plan as discussed in https://github.com/open-telemetry/opentelemetry-collector/pull/7439#discussion_r1178092780?

evan-bradley:
@dmitryax I think we can merge this as-is, I don't expect any functional changes to how partial success responses are handled even after we change the errors.

evan-bradley:
@dmitryax Could you give this a look? I think it is ready to go, and would like to be able to use this.

dmitryax (approved):

dmitryax (approved):

dmitryax:
> @dmitryax I think we can merge this as-is, I don't expect any functional changes to how partial success responses are handled even after we change the errors.

Ok, maybe it'll change the messaging only

evan-bradley:
>Ok, maybe it'll change the messaging only

I expect that to be the only thing that changes, since we will still be returning some form of permanent error.

dmitryax (approved):
LGTM as a temporary solution until we have structured permanent errors

PR #6975, opened 2023-01-19
[chore] Simplify contribution / Usage

MovieStoreGuy:
**Description:** 
Looking to reduce contribution toil and reducing onboarding efforts, this removes any additional once of steps that could cause issues with tooling versioning mismatches.

**Link to tracking Issue:**
N/A


MovieStoreGuy:
So the api diff build is failing due to the fact that it using main instead of my branch as the reference.

codeboten (approved):

PR #6980, opened 2023-01-19
Reduce contention in featuregate by using sync.Map instead of mutex.

bogdandrutu:


codeboten (approved):
Change looks good, any idea what the resulting difference in performance is?

codeboten (commented):
Please rebase now that build-tools have been updated :)

bogdandrutu:
> Change looks good, any idea what the resulting difference in performance is?

If I do a benchmark without concurrency the results is almost the same, with concurrency it really depends on how many go routines try to access them in the same time.

PR #6988, opened 2023-01-21
Revert "Fix: Force usage of case-sensitive keys in configurations"

bogdandrutu:
Reverts open-telemetry/opentelemetry-collector#6876

This PR has some unexpected side effects (at least for @bogdandrutu). When updating the contrib it failed because (I also realize that our contrib-test artifact does not work as expected, we can discuss that separately) https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/3971425778/jobs/6808376208.

Then when debugging, I found that was caused be https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/carbonreceiver/protocol/regex_parser.go#L78 which has a typo on the tag, which means the tag is not configured. That implies that after this PR the "default" tag value will be capitalized and no longer work, is this expected?


dmitryax (approved):

bogdandrutu:
@open-telemetry/collector-approvers please let me know your preference here.

mx-psi (approved):
The link on the description is a bit confusing since the issue was fixed on main already, here is a permalink showing the typo: https://github.com/open-telemetry/opentelemetry-collector-contrib/blame/2c17456b77a3322da694ab137b7ff2d5266ff26f/receiver/carbonreceiver/protocol/regex_parser.go#L78

Given we have a release today I think we should revert this and give us more time to discuss what to do and how to do this change properly (original PR + custom unmarshalling on the Carbon exporter sounds like a nice solution to me, but I want to give some time for others to participate in the discussion)

gbbr:
Just my 2c having been involved in the original PR review: isn't this a positive thing? The change caught a bug. Maybe ok to revert it for now, but I'd ultimately put it back...

codeboten (approved):

dmitryax:
I'd say let's revert it before the release and rethink after

rapphil:
Carbon receiver already got the fix for the typo https://github.com/open-telemetry/opentelemetry-collector-contrib/blame/main/receiver/carbonreceiver/protocol/regex_parser.go#L78 which means that collector-contrib v0.70 will have this fix. after this point it is safe to change the collector validation logic to be case sensitive no?

PR #7001, opened 2023-01-23
Add connector validation to otelcol

djaglowski:
Part of #6700 

This PR finishes adding connectors to the otelcol config including validation. Validation for connectors is the same as with other components, except two additional checks are included:
1. That each connector is used as an exporter in at least one pipeline, and also as a receiver in at least one pipeline.
2. That each connector ID is distinct from receivers and exporters. (Depends on decision on https://github.com/open-telemetry/opentelemetry-collector/pull/6927#issuecomment-1400791808)

### DIFF
--- a/otelcol/config.go
+++ b/otelcol/config.go
@@ -108,7 +108,7 @@ func (cfg *Config) Validate() error {
 		}
 
 		if _, ok := cfg.Exporters[connID]; ok {
-			return fmt.Errorf("connectors::%s: cannot have same id as exporter", connID)
+			return fmt.Errorf("connectors::%s: there's already an exporter named %q", connID, connID)
 		}
 		if _, ok := cfg.Receivers[connID]; ok {
 			return fmt.Errorf("connectors::%s: cannot have same id as receiver", connID)
--- a/otelcol/config.go
+++ b/otelcol/config.go
@@ -101,6 +101,20 @@ func (cfg *Config) Validate() error {
 		}
 	}
 
+	// Validate the connector configuration.
+	for connID, connCfg := range cfg.Connectors {
+		if err := component.ValidateConfig(connCfg); err != nil {
+			return fmt.Errorf("connectors::%s: %w", connID, err)
+		}
+
+		if _, ok := cfg.Exporters[connID]; ok {
+			return fmt.Errorf("connectors::%s: cannot have same id as exporter", connID)
+		}
+		if _, ok := cfg.Receivers[connID]; ok {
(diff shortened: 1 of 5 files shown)


otelcol/config.go
```diff
@@ -101,6 +101,20 @@ func (cfg *Config) Validate() error {
...
 	}
 
+	// Validate the connector configuration.
+	for connID, connCfg := range cfg.Connectors {
+		if err := component.ValidateConfig(connCfg); err != nil {
+			return fmt.Errorf("connectors::%s: %w", connID, err)
+		}
+
+		if _, ok := cfg.Exporters[connID]; ok {
+			return fmt.Errorf("connectors::%s: cannot have same id as exporter", connID)
```
jpkrohling, line 111: Putting myself in the shoes of someone who's looking at this message for the first time, I think the version below might be easier to grasp what's going on.

```suggestion
			return fmt.Errorf("connectors::%s: there's already an exporter named %q", connID, connID)
```

jpkrohling (commented):
LGTM, just perhaps one error message could be improved.

otelcol/config_test.go
```diff
@@ -59,6 +62,14 @@ func (nc *nopProcConfig) Validate() error {
 	return nc.validateErr
 }
 
+type nopConnConfig struct {
+	validateErr error
+}
+
+func (nc *nopConnConfig) Validate() error {
+	return nc.validateErr
+}
```
bogdandrutu, line 71: Separate PR, please merge all these config structs. We merged the config some time ago, so they don't need to be a struct per component type.

bogdandrutu (approved):

otelcol/otelcoltest/config_test.go
```diff
@@ -70,6 +71,7 @@ func TestLoadConfigAndValidate(t *testing.T) {
 	factories, err := NopFactories()
 	assert.NoError(t, err)
 
+	require.NoError(t, featuregate.GlobalRegistry().Apply(map[string]bool{"otelcol.enableConnectors": true}))
```
bogdandrutu, line 74: Should you `defer` to the previous state?

bogdandrutu (approved):

otelcol/otelcoltest/config_test.go
djaglowski: Done.

otelcol/config_test.go
djaglowski: Ok, I will work on this next.

PR #7013, opened 2023-01-24
Remove unnecessary interface in service/nodes.go

djaglowski:
This interface was part of #6700 but has been refactored out in the latest implementation.

### DIFF
--- a/service/nodes.go
+++ b/service/nodes.go
@@ -19,8 +19,6 @@ import (
 	"hash/fnv"
 	"strings"
 
-	"gonum.org/v1/gonum/graph"
-
 	"go.opentelemetry.io/collector/component"
 )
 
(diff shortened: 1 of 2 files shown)


djaglowski:
This would remove the `gonum` library, so rather than remove and readd, I'll rebase after https://github.com/open-telemetry/opentelemetry-collector/pull/7016 is merged.

service/nodes.go
```diff
@@ -36,13 +34,6 @@ func newNodeID(parts ...string) nodeID {
...
-type componentNode interface {
-	component.Component
-	graph.Node
-}
-
-var _ componentNode = &receiverNode{}
-
 // A receiver instance can be shared by multiple pipelines of the same type.
 // Therefore, nodeID is derived from "pipeline type" and "component ID".
 type receiverNode struct {
```
bogdandrutu, line 39: Should you remove the `component.Component` from the nodes?

djaglowski: I think it should stay so that we can iterate over nodes and Start/Stop them without needing to resolve a more specific type.

bogdandrutu: But is not initialized correct?

djaglowski: Not yet but it should be very soon.

djaglowski: #7026 can supersede this PR, but I'm curious whether you will allow temporarily untested code for the sake of having reasonably sized PRs.

bogdandrutu (approved):

PR #7016, opened 2023-01-24
Add implementation of pipelineGraph StartAll and ShutdownAll

djaglowski:
Part of #6700 

This adds graph-based implementations for `StartAll` and `Shutdown` all. It also adds a very lightweight test framework that validates the expected start & stop ordering of generic nodes in the graph.

### DIFF
--- a/service/graph.go
+++ b/service/graph.go
@@ -54,9 +54,7 @@ func (g *pipelinesGraph) ShutdownAll(ctx context.Context) error {
 	// Stop receivers first, and work towards exporters
 	var errs error
 	for i := 0; i < len(nodes); i++ {
-		if compErr := nodes[i].(component.Component).Shutdown(ctx); compErr != nil {
-			errs = multierr.Append(errs, compErr)
-		}
+		errs = multierr.Append(errs, nodes[i].(component.Component).Shutdown(ctx))
 	}
 	return errs
 }
--- a/service/graph.go
+++ b/service/graph.go
@@ -18,21 +18,47 @@ import (
 	"context"
 	"net/http"
 
+	"go.uber.org/multierr"
+	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph/topo"
+
 	"go.opentelemetry.io/collector/component"
 )
 
 var _ pipelines = (*pipelinesGraph)(nil)
 
-type pipelinesGraph struct{}
+type pipelinesGraph struct {
+	// All component instances represented as nodes, with directed edges indicating data flow.
+	componentGraph *simple.DirectedGraph
+}
 
 func (g *pipelinesGraph) StartAll(ctx context.Context, host component.Host) error {
-	// TODO actual implementation
+	nodes, err := topo.Sort(g.componentGraph)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


service/graph.go
```diff
@@ -18,20 +18,44 @@ import (
...
 
 func (g *pipelinesGraph) ShutdownAll(ctx context.Context) error {
-	// TODO actual implementation
+	nodes, err := topo.Sort(g.componentGraph)
+	if err != nil {
+		return err
+	}
+	// Stop receivers first, and work towards exporters
+	for i := 0; i < len(nodes); i++ {
+		if compErr := nodes[i].(component.Component).Shutdown(ctx); compErr != nil {
```
bogdandrutu, line 55: I think we should not stop if error, instead accumulate them using `multierr.Append` and continue.

djaglowski: Makes sense. Fixed.

service/graph.go
```diff
@@ -18,21 +18,47 @@ import (
...
+	nodes, err := topo.Sort(g.componentGraph)
+	if err != nil {
+		return err
+	}
+	// Stop receivers first, and work towards exporters
+	var errs error
+	for i := 0; i < len(nodes); i++ {
+		if compErr := nodes[i].(component.Component).Shutdown(ctx); compErr != nil {
+			errs = multierr.Append(errs, compErr)
+		}
```
bogdandrutu, line 59:
```suggestion
		errs = multierr.Append(errs, nodes[i].(component.Component).Shutdown(ctx))
```

service/graph_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+var _ component.Component = &testNode{}
+
+type testNode struct {
+	id          component.ID
+	startErr    error
+	shutdownErr error
+}
+
+func (n *testNode) ID() int64 {
```
bogdandrutu, line 39: Unused?

service/graph_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+			pg := &pipelinesGraph{componentGraph: simple.NewDirectedGraph()}
+			for _, edge := range tt.edges {
+				f, t := &testNode{id: edge[0]}, &testNode{id: edge[1]}
+				pg.componentGraph.SetEdge(simple.Edge{F: f, T: t})
+			}
+
+			require.NoError(t, pg.StartAll(ctx, componenttest.NewNopHost()))
+			for _, edge := range tt.edges {
+				assert.Greater(t, ctx.order[edge[0]], ctx.order[edge[1]])
```
bogdandrutu, line 146: To make sure we don't by mistake get this right, can we have a different map for start order and stop order?

service/graph_test.go
djaglowski: This fulfills the `graph.Node` interface, so that we can use `testNode` in the `simple.DirectedGraph`.

service/graph_test.go
djaglowski: Good call. Now resetting the map before calling ShutdownAll.

service/graph_test.go
bogdandrutu: Add a comment, for someone unfamiliar with that.

bogdandrutu (approved):

PR #7021, opened 2023-01-25
Revert "Revert "Fix: Force usage of case-sensitive keys in configurations (#6876)" (#6988)"

### LINKED ISSUE #7002: Collector supports case insensitive configuration but it is not properly handled across the codebase.
**Describe the bug**
The collector supports case insensitive configuration, allowing typos in the configuration to change the behavior of the code. This can happen in the code because in several instances we are not relaying on the de-serialized datastructure, but checking the existing of keys in the code. 

E.g.: 
* https://github.com/open-telemetry/opentelemetry-collector/blob/2b96397073bccb68a967ca549f780a98b87c067e/exporter/loggingexporter/config.go#L75
* https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0801b5f3aefdf920cfda914d93d23bff14bd5778/receiver/jaegerreceiver/config.go#L157

There was a [fix](https://github.com/open-telemetry/opentelemetry-collector/pull/6876) for this issue but this was [reverted](https://github.com/open-telemetry/opentelemetry-collector/pull/6988).

**Steps to reproduce**

Create a configuration with a typo, replacing a lower case by an uppercase in the logging exporter loglevel -> logLevel.

loglevel to be properly set or an error message that the configuration is invalid.

**What did you see instead?**
loglevel set incorrectly.

**What version did you use?**
v0.69.1

**Environment**
Centos 7

gbbr:
This reverts commit 80cabdd33c50eaf386293f90fd8ac57cb3a10dc9.

Closes #7002

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -161,6 +161,7 @@ func decodeConfig(m *Conf, result interface{}, errorUnused bool) error {
 		Result:           result,
 		TagName:          "mapstructure",
 		WeaklyTypedInput: true,
+		MatchName:        caseSensitiveMatchName,
 		DecodeHook: mapstructure.ComposeDecodeHookFunc(
 			expandNilStructPointersHookFunc(),
 			mapstructure.StringToSliceHookFunc(","),
@@ -187,6 +188,13 @@ func encoderConfig(rawVal interface{}) *encoder.EncoderConfig {
 			marshalerHookFunc(rawVal),
 		),
 	}
+}
+
+// case-sensitive version of the callback to be used in the MatchName property
+// of the DecoderConfig. The default for MatchEqual is to use strings.EqualFold,
+// which is case-insensitive.
+func caseSensitiveMatchName(a, b string) bool {
+	return a == b
 }
 
 // In cases where a config has a mapping of something to a struct pointers
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -116,7 +116,6 @@ func (l *Conf) Get(key string) interface{} {
 }
 
 // IsSet checks to see if the key has been set in any of the data locations.
-// IsSet is case-insensitive for a key.
 func (l *Conf) IsSet(key string) bool {
 	return l.k.Exists(key)
 }
--- a/.chloggen/case-sensitive-configuration.yaml
+++ b/.chloggen/case-sensitive-configuration.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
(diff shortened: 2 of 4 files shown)


bogdandrutu:
Some things that I have in mind and will like answers/someone to think about:
* We have a test helper for configs that disallow uppercase, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/componenttest/configtest.go#L34. Please consider to come with a fix there as well if needed.
* By default because fields need to be public they are uppercase. In case of not having a "mapstructure" tag they will be uppercased in the config, is that expected? Do we want to be lowercase by default if no tag?

gbbr:
@rapphil I think you may want to jump in here considering you have opened to [original PR](https://github.com/open-telemetry/opentelemetry-collector/pull/6876).

rapphil:
@bogdandrutu 
Based on https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/componenttest/configtest.go#L34 can we say that we only support case sensitive configurations with with lowercase keys? (otherwise why would we enforce the tags to be lowercase?)

So answering your questions:

>We have a test helper for configs that disallow uppercase, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/componenttest/configtest.go#L34. Please consider to come with a fix there as well if needed.
We don't need to fix anything as this check function is case sensitive, so it aligns with the validation that will happen in mapstructrure when the configuration is Unmarshalled into a struct.

> By default because fields need to be public they are uppercase. In case of not having a "mapstructure" tag they will be uppercased in the config, is that expected? Do we want to be lowercase by default if no tag?

This will only happen if a config is not validated using the `CheckConfigStruct` function. yes, we should check this case.

bogdandrutu:
@gbbr you need to also remove the wrong comment from `IsSet`

gbbr:
Thanks for flagging that Bogdan, done.

bogdandrutu (approved):

confmap/confmap.go
```diff
@@ -189,6 +189,13 @@ func encoderConfig(rawVal interface{}) *encoder.EncoderConfig {
 	}
 }
 
+// case-sensitive version of the callback to be used in the MatchName property
+// of the DecoderConfig. The default for MatchEqual is to use strings.EqualFold,
+// which is case-insensitive.
+func caseSensitiveMatchName(a, b string) bool {
+	return a == b
```
bogdandrutu, line 196: Should we trim the strings at least?

rapphil: Why?  To allow keys with spaces to work as well?
https://github.com/mitchellh/mapstructure/blob/v1.5.0/mapstructure.go#L275

rapphil (approved):

confmap/confmap.go
gbbr: I agree with Raphael on this one, we should be strict.

bogdandrutu: I think " foo" == "foo " in this case. Spaces in a config file will be hard to see if extra.

PR #7023, opened 2023-01-25
Add entry point to build graph-based pipelines

djaglowski:
Part of #6700

The remainder of the connectors implementation is concerned with setting up the component graph and building each node's associated component. I believe the best way to add these capabilities incrementally will be to so in a top-down way. 

- This PR adds the entry point to building the component graph. 
- Next, outline the steps to set up and build the graph.
- Then add nodes, then edges. 
- Finally, build each node's associated component.

#6700 contains extensive testing of the graph as a whole. However, these tests interface primarily with the graph's built components. I'm not seeing a reasonable way to include coverage for a partially built graph, which I expect we would have if we follow the contribution path I've described above. My hope is that we can move forward for a few PRs based on the understanding that test coverage will be comprehensive in the end.

---

This PR adds the entry point to building the component graph. This entry point is hidden behind the same feature gate that was already to the `otelcol` package in #6964. However, it moves the gate to the `service` so that both packages can use it.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -29,6 +29,7 @@ import (
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/extension"
 	"go.opentelemetry.io/collector/internal/obsreportconfig"
+	"go.opentelemetry.io/collector/internal/sharedgate"
 	"go.opentelemetry.io/collector/processor"
 	"go.opentelemetry.io/collector/receiver"
 	"go.opentelemetry.io/collector/service/extensions"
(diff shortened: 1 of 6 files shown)


service/service.go
```diff
@@ -36,6 +37,12 @@ import (
 	"go.opentelemetry.io/collector/service/telemetry"
 )
 
+var ConnectorsFeatureGate = featuregate.GlobalRegistry().MustRegister(
```
mx-psi, line 40: Can this be moved to some `internal` package to avoid having it be part of the public API?

djaglowski: That's a good idea. I'll move it.

bogdandrutu:
Please rebase

djaglowski:
Thanks, this has been rebased

jpkrohling (approved):

mx-psi (approved):

djaglowski:
Can this be merged?

dmitryax (approved):

Issue #5300, opened 2022-04-29
HowTo: Telemetry Logging to File when running as windows service


Mario-Hofstaetter:
**Describe the bug**

This is no traditional bug report, more a documentation how to solve the described issue.

If maintainers would like a PR to expand the docs anywhere to contain information about this (not sure what from it), I could try to contribute.
If not, this could also considered a rant and be closed. Maybe it helps somebody google landing here.

I spent several HOURS to get a solution for something that should by trivial to do.

### The Situation:

- Running `otelcol.exe` on windows as a windows-service (supported natively by the binary)
- Otelcol spams windows application eventlog with messages (is this config documented anywhere?), 
  - The event logs are somewhat broken? (see screenshots below)
- It is not possible to "bypass" otelcol telemetry logs to an `exporter` to send them off directly? (e.g. [using `loki`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/lokiexporter))

### The Goal:

- Write logmessages of `otelcol` to files (to tail and ship to central logging server)
  - This is necessary to observe issues with otelcol
- Use `json` as log format
- Use Log rotation so the file does not get huge and fills up harddisk in any case
  - Log Roration is a must have feature on production ⚡ 

----

<details>
<summary>SCREENSHOTS: Buggy windows application logs from otelcol (CLICK ME)</summary>

this:
![grafik](https://user-images.githubusercontent.com/33002073/166070856-6dfecbd6-47fa-4bd0-a7eb-f7b78dd0bb73.png)
is actually:
![grafik](https://user-images.githubusercontent.com/33002073/166070885-d623a0a9-30e5-43f7-9c7b-3ee044a89e28.png)

and this:
![grafik](https://user-images.githubusercontent.com/33002073/166070907-f08771b3-e9fe-4727-b0a4-a6f50806e52e.png)
is actually:
![grafik](https://user-images.githubusercontent.com/33002073/166070924-137d16e3-e352-4f5c-b2d2-0b6c1862f19c.png)

</details>

----

### The journey:

The docs here https://opentelemetry.io/docs/collector/configuration/#service describe the `telemetry` logs and link to here: https://github.com/open-telemetry/opentelemetry-collector/blob/7666eb04c30e5cfd750db9969fe507562598f0ae/config/service.go

```
	// Encoding sets the logger's encoding.
	// Example values are "json", "console".
	Encoding string `mapstructure:"encoding"`

[...]

PR #7025, opened 2023-01-25
Enable service logging options when running as windows service

xborder:
**Description:** 

This PR partially addresses the issues described in https://github.com/open-telemetry/opentelemetry-collector/issues/5300.

When running as a windows service, the collector's logger zapcore.Core is replaced by a customized zapcore.Core that writes the logs to the EventViewer ignoring what is configured in the yaml.
Instead of replacing it, the PR creates a Tee that preserves the original Core and its configuration, and adds the customized Core for EventViewer. 
This way logs are duplicated and flushed by both Cores.
Since there isn't a specific `telemetry` configuration for windows the logging level is shared between both Cores.

As an example the following configuration:
```
service:
  telemetry:
    logs:
      level: warn
      output_paths: ["C:\\otel-collector.log"]
```
Would enable logs with a level equal or bigger to `warn` to be sent to EventViewer and the file `otel-collector.log`

This partially addresses this [issue](https://github.com/open-telemetry/opentelemetry-collector/issues/5300).

**Link to tracking Issue:**
* https://github.com/open-telemetry/opentelemetry-collector/issues/5300

**Testing:**
* Tested locally on a windows machine


codeboten (commented):
Will the existing testsuite for windows (https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/build-and-test-windows.yaml) be enough coverage to ensure this test solves part of the issue described?

Gaps in windows testing for this repo has historically caused problems.

xborder:
@codeboten From my understanding, there is not specific test suite for windows. The workflow you references just runs the tests on a windows container.

I only found [1 windows-specific test](https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/collector_windows_test.go) that tests the service change of status. In this PR case, it would require mocking the event viewer somehow and change the code to make stuff public in order to mock it. 
Being something so specific to the OS and only when it runs as a window service, I find it somewhat hard to justify. 

I'm not well-versed in the golang ways so let me know if I could improve it somehow.

PR #7026, opened 2023-01-25
Add component node build methods

djaglowski:
Part of #6700

This PR continues with the bottom up approach started in #6895. It defines a `build` method for each node type. The receiver, processor, and exporter nodes are able to wrap the preexisting `build<Type>` functions from `pipelines.go`. The connector node has a fully new implementation since it will not be shared by the other `pipelines` implementation.


dmitryax (approved):
LGTM. Will this code be covered with tests as part of other PRs?

djaglowski:
> LGTM. Will this code be covered with tests as part of other PRs?

Yes indeed. 👍

PR #7045, opened 2023-01-27
Add full component graph build

djaglowski:
This PR contains the bulk of what remains of #6700

It is a large PR, but mostly because of tests.

I'm presenting this as an alternative to the incremental top-down approach described in #7023. The main benefit of this would be to prevent a temporary lapse in test coverage.


service/nodes.go
```diff
@@ -182,3 +209,69 @@ func (n *connectorNode) build(
...
+
+func newFanInNode(pipelineID component.ID) *fanInNode {
+	return &fanInNode{
+		nodeID:       newNodeID("fanin_to_processors", pipelineID.String()),
+		pipelineID:   pipelineID,
+		Capabilities: consumer.Capabilities{},
+	}
+}
+
+func (n *fanInNode) build(nexts []baseConsumer, processors []*processorNode) {
```
bogdandrutu, line 232: Same thing, why passing a slice when only the first is used.

service/graph.go
```diff
@@ -31,21 +32,236 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+				connectorsAsExporter[exprID] = append(connectorsAsExporter[exprID], pipelineID)
+				continue
+			}
+			g.addExporter(pipelineID, exprID)
+		}
+	}
+
+	if len(connectorsAsExporter) != len(connectorsAsReceiver) {
+		return fmt.Errorf("each connector must be used as both receiver and exporter")
+	}
```
bogdandrutu, line 92: Is this not checked in the "Validation" of the config? Why duplicate the logic here?

service/graph.go
```diff
@@ -31,21 +32,236 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+
+func (g *pipelinesGraph) addReceiver(pipelineID, recvID component.ID) {
+	node := newReceiverNode(pipelineID, recvID)
+	if rcvrNode := g.componentGraph.Node(node.ID()); rcvrNode != nil {
+		g.pipelines[pipelineID].addReceiver(rcvrNode)
+		return
+	}
+	g.pipelines[pipelineID].addReceiver(node)
+	g.componentGraph.AddNode(node)
+}
```
bogdandrutu, line 115: I would refactor this to be just "createReceiver"
```suggestion
func (g *pipelinesGraph) createReceiver(pipelineID, recvID component.ID) *receiverNode {
	node := newReceiverNode(pipelineID, recvID)
	if rcvrNode := g.componentGraph.Node(node.ID()); rcvrNode != nil {
		return
	}
	g.componentGraph.AddNode(node)
}
```

And on line 70, to separate the creation of the individual nodes with the adding to the pipelines:


```golang
	for pipelineID, pipelineCfg := range set.PipelineConfigs {
		pipe := g.pipelines[pipelineID]
		for _, recvID := range pipelineCfg.Receivers {
			if set.Connectors.IsConfigured(recvID) {
				connectorsAsReceiver[recvID] = append(connectorsAsReceiver[recvID], pipelineID)
				continue
			}
			pipe.addReceiver(g.createReceiver(pipelineID, recvID))
		}
		...
	}
```

service/graph.go
```diff
@@ -31,21 +32,236 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+		g.pipelines[exprPipelineID].addExporter(connNode)
+		g.pipelines[rcvrPipelineID].addReceiver(connNode)
+		return
+	}
+	g.pipelines[exprPipelineID].addExporter(node)
+	g.pipelines[rcvrPipelineID].addReceiver(node)
+	g.componentGraph.AddNode(node)
+}
+
+func (g *pipelinesGraph) createEdges() {
```
bogdandrutu, line 145: I feel that this belongs to `pipelineNodes`. I feel that you are trying to create a global "graph", but I think it is simpler to read/implement if we create a graph where the pipeline is the Node and the Edges are the "Connectors".

So essentially we can probably keep almost all the previous logic to build a pipeline (builtPipeline), only add support for connectors to be receiver/exporters, then add them to a graph and set the edges where we have connectors to determine the initialization order.

service/graph.go
djaglowski: You're right. This was just defensive code, but it is unnecessary. I will remove.

service/graph.go
djaglowski: I've worked out the approximate logic. Forgive the mixed psuedocode format, but I think it would look like this:

### Start (reverse for Shutdown)
- Iterate over global list of exporters, start each one
- Sort the graph topologically. For each node (pipeline) in appropriate order
  - Start each adjacent edge (connector)
  - Start each processor in appropriate order
- Iterate over global list of receivers, start each one

### Build

Some things we use along the way:

```go
// Global indices for quick lookup of pipelines where each connector is used as receiver and exporter
var connectorsAsExporters, connectorsAsReceivers map[component.ID][]component.ID

// Global indices for built receivers, exporters, and connectors, to ensure correct instancing
allReceivers map[component.DataType]map[component.ID]component.Component
allExporters map[component.DataType]map[component.ID]component.Component
allConnectors  map[component.DataType]map[component.DataType]map[component.ID]component.Component

// Basic graph structs & utilities
type Edge struct {
	graph.Edge
	component.ID // connector ID
}
type Node struct {
	component.ID // pipeline ID
}
func(component.ID) int64

// Helper to get first processor or fanout to exporters
func (p *builtPipeline) getFirstConsumer()
```

Before we can build anything, we need to populate the graph so we can reason about topological order. Roughly:
```go
// Cross reference connector indices. Each match represents an edge
for connID, exprPipelineIDs := range connectorsAsExporter {
	for _, eID := range exprPipelineIDs {
		for _, rID := range connectorsAsReceiver[connID] {
			// build Edge, add to graph. Nodes added automatically
		}
	}
}
```

Then we can sort the graph topologically and iterate through pipelines from end to beginning. 

Building exporters is the same, other than identifying the connectors and building those separately.

Building connectors is where it gets a bit ugly. Each connector requires accessing the graph to find the downstream pipeline(s) in which that connector is used as a receiver. This is not necessarily all the edges going out from a pipeline, as there may be multiple connectors in the pipeline. Once we've got the right set of edges, we need to get the appropriate consumer from each pipeline (first processor or fanout), then wrap those together into a fanout. Once built, we can iterate through the edges again and add the connector to all associated `builtPipeline`s, each in the appropriate bucket according to either side of the edge.

```
for each comp in pipelineCfg.exporters
  if comp is a connector
    if conn := allConnectors(comp); conn != nil
      continue
    else
      firstConsumers := []consumers // first consumer of each downstream pipeline
      edgesFromPipeline := graph.From(connector.ID)
        for each edge in edgesFromPipeline
          if edge.ID == connector.ID
            downstreamPipeline := // TODO get pipeline on other side of edge
            firstConsumers.append(downstreamPipeline.getFirstConsumer())
      conn := buildConn(..., fanout(firstConsumers))
      allConnectors.add(conn)
      // iterate through edges again and add to once to each node's builtPipeline.[exporter|receiver]
  else if exp := allExporters(comp); exp != nil
    thisPipeline.exporters.add(exp)
  else
    exp := builtExporter
    allExporters.add(exp)
    thisPipeline.exporters.add(exp)
```

Processors are built the same.
- Build last processor with fanout to `builtPipeline.exporters`
- Build remaining processors in appropriate order


Receivers also need to be separated into receivers and connectors, but otherwise building them is the same. 

When discovering a connector here, we can ignore it because it will be built by another pipeline iteration and added appropriately. (This can probably be inverted so it is the exporters that are deferred.)
```
for each comp in pipelineCfg.receivers
  if comp is a connector
    continue
  else if rcvr := allReceivers(comp); rcvr != nil
    thisPipeline.receivers.add(rcvr)
  else
    rcvr := buildReceiver(...)
    allReceivers.add(rcvr)
    thisPipeline.receivers.add(rcvr)
```

I think this is doable, and probably can be further streamlined. I agree it would be able to use a lot more of the existing code. That said, I think it may not be as straightforward as it first appears.

Personally, I find the global graph easy to reason about because it is built around the actual ontology of the collector - components, pipelines, and a global representation of actual data flow. This other way requires management of artificial indices and lurching between abstraction layers. For example, start and shutdown require working with a global set of one component type, then working within pipelines, then components globally again. And of course building connectors appears to be quite convoluted this way as well.

In any case, I'm far enough into this that I think I can probably get a working implementation together early next week if that's how you'd like to proceed.

djaglowski: I started working through the above implementation but it appears that it would be a massive overhaul that is only worse than expected.

What I've proposed in this PR is a clean and working solution that is as well tested as the production code. I've openly pursued this design for three months now and have iterated on it relentlessly to simplify it down to what it is today. I think the suggested overhaul would result in a worse implementation with no meaningful difference in behavior and that it would add unnecessary delay to a feature set that is ready for use by those who opt in. I feel that I have made reasonable efforts to share aspects of the previous implementation but we should expect a temporary state with parallel implementations that diverge in sensible ways.

If there is no other feedback on this PR, I propose that we move forward with my implementation, with the understanding that it may be necessary to work through stylistic changes prior to removal of the feature gate / reduction to one implementation.

bogdandrutu (commented):
I forgot to press submit review, I am sorry

service/graph.go
```diff
@@ -31,21 +31,220 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+		switch n := node.(type) {
+		case *receiverNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Receivers, g.nextConsumers(n.ID()))
+		case *processorNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Processors, g.nextConsumers(n.ID())[0])
+		case *connectorNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Connectors, g.nextConsumers(n.ID()))
+		case *exporterNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Exporters)
+		case *fanInNode:
```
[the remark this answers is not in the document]
djaglowski: 👍 I'll rename it.

service/graph.go
```diff
@@ -31,21 +31,220 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+		case *exporterNode:
+			nexts = append(nexts, next.Component.(baseConsumer))
+		case *connectorNode:
+			nexts = append(nexts, next.Component.(baseConsumer))
+		case *fanInNode:
+			nexts = append(nexts, next.baseConsumer)
+		case *fanOutNode:
+			nexts = append(nexts, next.baseConsumer)
+		}
+	}
```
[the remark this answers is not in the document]
djaglowski: Good idea. This worked as described.

service/graph.go
```diff
@@ -31,21 +31,220 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+	nodes, err := topo.Sort(g.componentGraph)
+	if err != nil {
+		return err // TODO clean up error message
+	}
+
+	for i := len(nodes) - 1; i >= 0; i-- {
+		node := nodes[i]
+		switch n := node.(type) {
+		case *receiverNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Receivers, g.nextConsumers(n.ID()))
```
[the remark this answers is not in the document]
djaglowski: I created https://github.com/open-telemetry/opentelemetry-collector/issues/7089 to track this and assigned to myself.

service/graph.go
```diff
@@ -31,21 +31,220 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+	p.exporters[node.ID()] = node
 }
 
 func (g *pipelinesGraph) StartAll(ctx context.Context, host component.Host) error {
 	nodes, err := topo.Sort(g.componentGraph)
 	if err != nil {
 		return err
 	}
+
 	// Start exporters first, and work towards receivers
```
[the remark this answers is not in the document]
djaglowski: It's slightly inaccurate, but I thought helpful to be a bit more on topic. Technically, the first component to start will always be an exporter and the last will be a receiver, but it does not guarantee that all exporters are started before anything else or that all receivers are started after everything else. I'll clarify.

service/graph.go
```diff
@@ -31,21 +31,220 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+		case *receiverNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Receivers, g.nextConsumers(n.ID()))
+		case *processorNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Processors, g.nextConsumers(n.ID())[0])
+		case *connectorNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Connectors, g.nextConsumers(n.ID()))
+		case *exporterNode:
+			err = n.build(ctx, set.Telemetry, set.BuildInfo, set.Exporters)
+		case *fanInNode:
+			n.build(g.nextConsumers(n.ID())[0], g.nextProcessors(n.ID()))
```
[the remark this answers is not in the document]
djaglowski: 👍 Good point, done.

service/graph.go
```diff
@@ -31,21 +31,209 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+}
+
+// A node-based representation of a pipeline configuration.
+type pipelineNodes struct {
+	// Use maps for receivers and exporters to assist with deduplication of connector instances.
+	receivers map[int64]graph.Node
+	exporters map[int64]graph.Node
+
+	// The order of processors is very important. Therefore use a slice for processors.
+	processors []graph.Node
```
bogdandrutu, line 208: Why not having this as `*processorNode` (avoid casting) in the nextProcessors?

service/graph.go
```diff
@@ -31,21 +31,209 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+	}
+	for pipelineID := range set.PipelineConfigs {
+		pipelines.pipelines[pipelineID] = &pipelineNodes{
+			receivers: make(map[int64]graph.Node),
+			exporters: make(map[int64]graph.Node),
+		}
+	}
+	pipelines.createNodes(set)
+	pipelines.createEdges()
+	return pipelines, pipelines.buildNodes(ctx, set)
```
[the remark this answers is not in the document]
djaglowski: `buildComponents`?

service/graph.go
```diff
@@ -31,21 +31,209 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+			}
+		}
+	}
+}
+
+func (g *pipelinesGraph) createReceiver(pipelineID, recvID component.ID) *receiverNode {
+	rcvrNode := newReceiverNode(pipelineID, recvID)
+	if node := g.componentGraph.Node(rcvrNode.ID()); node != nil {
+		return node.(*receiverNode)
+	}
```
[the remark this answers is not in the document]
djaglowski: Strictly speaking it is not required that the id is reproducible, but I think the code is cleaner if it is.

Either way, I will split these funcs if you still want.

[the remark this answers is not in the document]
bogdandrutu: As I said "Happy to address in a separate PR to see if it improves readability." :) Let's play with this after this PR is merged, we can debate there :)

service/graph.go
```diff
@@ -31,21 +31,209 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+	connNode := newConnectorNode(exprPipelineID.Type(), rcvrPipelineID.Type(), connID)
+	if node := g.componentGraph.Node(connNode.ID()); node != nil {
+		return node.(*connectorNode)
+	}
+	g.componentGraph.AddNode(connNode)
+	return connNode
+}
+
+func (g *pipelinesGraph) createEdges() {
+	for pipelineID, pg := range g.pipelines {
```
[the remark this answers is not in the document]
djaglowski: Will rework it

service/graph.go
```diff
@@ -31,21 +31,209 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+		fanOutToExporters := newFanOutNode(pipelineID)
+		for _, exporter := range pg.exporters {
+			g.componentGraph.SetEdge(g.componentGraph.NewEdge(fanOutToExporters, exporter))
+		}
+
+		if len(pg.processors) == 0 {
```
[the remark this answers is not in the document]
djaglowski: Not sure it's quite what you were looking for, but I changed it so that every pipeline has a capabilities node and fanout node. Now creating edges has no special cases.

> n the building make its base consumer == { the processor if only one processor; the fanOutToExporters if no processor; 

This is already the case because the edge from capabilities node points to the correct next consumer, either the first processor or the fanout node.

> create a real capability consumer if more than one processor and at least one mutates

You might be referring to something more specific here, but I think the capabilities node already _is_ a real capability consumer in the sense that it is a base consumer and also can describe its capabilities (aggregated from the processors).

djaglowski:
I think I've addressed all feedback. Most items were addressed in separate commits for the sake of readability. I can squash when you think it's ready.

service/graph.go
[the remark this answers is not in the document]
bogdandrutu: Mark as resolved since we decided to look into this later.

service/graph.go
```diff
@@ -31,21 +31,202 @@ var _ pipelines = (*pipelinesGraph)(nil)
...
+		for _, exporter := range pg.exporters {
+			g.componentGraph.SetEdge(g.componentGraph.NewEdge(pg.fanOutNode, exporter))
+		}
+	}
+}
+
+func (g *pipelinesGraph) buildComponents(ctx context.Context, set pipelinesSettings) error {
+	nodes, err := topo.Sort(g.componentGraph)
+	if err != nil {
+		return err // TODO clean up error message
```
bogdandrutu, line 159: Can you be more specific about what is left here?

service/nodes.go
```diff
@@ -182,3 +240,82 @@ func (n *connectorNode) build(
...
+		pipelineID:   pipelineID,
+		Capabilities: consumer.Capabilities{},
+	}
+}
+
+func (n *capabilitiesNode) getConsumer() baseConsumer {
+	return n.baseConsumer
+}
+
+func (n *capabilitiesNode) build(next baseConsumer, processors []*processorNode) {
```
bogdandrutu, line 270: I think this logic is wrong since you don't construct a different baseConsumer some time.

service/nodes.go
```diff
@@ -85,12 +135,31 @@ type exporterNode struct {
...
+}
+
+func (n *exporterNode) build(
+	ctx context.Context,
+	tel component.TelemetrySettings,
+	info component.BuildInfo,
+	builder *exporter.Builder,
+) error {
+	set := exporter.CreateSettings{ID: n.componentID, TelemetrySettings: tel, BuildInfo: info}
+	set.TelemetrySettings.Logger = components.ExporterLogger(set.TelemetrySettings.Logger, n.componentID, n.pipelineType)
```
bogdandrutu, line 155: This code is duplicate with the old pipelines code. May worth changing it to:

```
func buildReceiver(ctx context.Context,
	tel component.TelemetrySettings,
	info component.BuildInfo,
	builder *receiver.Builder,
	pipelineID component.ID,
	nexts []baseConsumer,
) (recv component.Component, err error) {
	set := exporter.CreateSettings{ID: n.componentID, TelemetrySettings: tel, BuildInfo: info}
	set.TelemetrySettings.Logger = components.ExporterLogger(set.TelemetrySettings.Logger, n.componentID, n.pipelineType)
	...
}
```

service/graph.go
djaglowski: When there is a cycle in the graph, there is enough information within the error to construct a better error message that indicates exactly the components that are in a cycle. It's non-trivial logic though, so I think it's best for another PR.

service/nodes.go
djaglowski: I don't understand what's wrong. The node always passes data directly to the next consumer, so its baseConsumer is the same as the next consumer.

service/nodes.go
djaglowski: [This commit](https://github.com/open-telemetry/opentelemetry-collector/pull/7045/commits/fce05a8fc53b86ed8c98566f776c766b8d5c0feb) contains this change.

service/nodes.go
bogdandrutu: Then why do you create them?

bogdandrutu: Asked you a separate question, maybe that clarifies.

service/nodes.go
```diff
@@ -103,82 +128,120 @@ type connectorNode struct {
...
+// Every pipeline has a "virtual" capabilities node immediately after the receiver(s).
+// There are two purposes for this node:
+// 1. Present aggregated capabilities to receivers, such as whether the pipeline mutates data.
+// 2. Present a consistent "first consumer" for each pipeline.
+// The nodeID is derived from "pipeline ID".
+type capabilitiesNode struct {
+	nodeID
+	pipelineID component.ID
+	baseConsumer
+	consumer.Capabilities
```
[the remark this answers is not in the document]
djaglowski: We build the graph in such a way that each `receiverNode` has edges _only_ to one or more `capabilitiesNode`.

Before we build a receiver component, we call `componentGraph.nextConsumers` to get the `[]capabilitiesNode` for that receiver. (Returned as `[]baseConsumer` but actually is always `[]capabilitiesNode` for a receiver.)

Then in `buildReceiver`, we build a new `fanoutconsumer` that takes into account `Capabilities.MutatesData`. This `fanoutconsumer` is given to the new receiver component as its next consumer.

All of the above is true for connectors too - they always emit to `fanoutconsumer([]capabilitiesNode)` which accounts for `MutatesData`.

Separately, `HandleZPages` describes each pipeline including whether it `MutatesData`. This can easily be determined from the `capabilitiesNode` in each pipeline. (Equivalent to `builtPipeline.lastConsumer`.)

service/graph.go
djaglowski: Just saw your comment. You probably saw by now that I refactored this like you suggested.

service/nodes.go
djaglowski: Hopefully explained well here: https://github.com/open-telemetry/opentelemetry-collector/pull/7045#discussion_r1095108288

service/nodes.go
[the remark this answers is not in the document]
djaglowski: @bogdandrutu and I discussed briefly offline and determined that there was indeed a problem with the capabilities node. I've believe I've addressed it in [this commit](https://github.com/open-telemetry/opentelemetry-collector/pull/7045/commits/63976de092281c9aceee32ee2ec88717732ced99).

bogdandrutu (approved):

Issue #6932, opened 2023-01-11
confmap expand feature fails in prometheusremotewrite


kago-dk:

mx-psi:
I moved this to core since based on the comment above this is not specific to the PRW exporter.

cc @bogdandrutu

kago-dk:
> endpoint: "${env:PROMETHEUS_FULL_URL}"

It seems to work.

**Background**
I was trying to follow the examples shown in:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/17300/files
<img width="851" alt="image" src="https://user-images.githubusercontent.com/49529296/212086851-39f0086a-e27e-4509-a70f-9598cd328177.png">

mx-psi:
I will revert the examples that use `${env:...}` in the middle of a string until this is solved.

PR #7055, opened 2023-01-29
Add support for concatenating multiple embedded uris, or uris with other string parts

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/6932

### DIFF
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -18,7 +18,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"go.uber.org/multierr"
@@ -26,15 +28,21 @@ import (
 	"go.opentelemetry.io/collector/featuregate"
 )
 
+// schemePattern defines the regexp pattern for scheme names.
+// Scheme name consist of a sequence of characters beginning with a letter and followed by any
+// combination of letters, digits, plus ("+"), period ("."), or hyphen ("-").
+const schemePattern = `[A-Za-z][A-Za-z0-9+.-]+`
+
 var (
 	// follows drive-letter specification:
 	// https://datatracker.ietf.org/doc/html/draft-kerwin-file-scheme-07.html#section-2.2
 	driverLetterRegexp = regexp.MustCompile("^[A-z]:")
 
-	// Scheme name consist of a sequence of characters beginning with a letter and followed by any
-	// combination of letters, digits, plus ("+"), period ("."), or hyphen ("-").
 	// Need to match new line as well in the OpaqueValue, so setting the "s" flag. See https://pkg.go.dev/regexp/syntax.
-	locationRegexp = regexp.MustCompile(`(?s:^(?P<Scheme>[A-Za-z][A-Za-z0-9+.-]+):(?P<OpaqueValue>.*)$)`)
+	uriRegexp = regexp.MustCompile(`(?s:^(?P<Scheme>` + schemePattern + `):(?P<OpaqueValue>.*)$)`)
+
+	// embeddedURI matches "embedded" provider uris into a string value.
+	embeddedURI = regexp.MustCompile(`\${` + schemePattern + `:.*?}`)
 
 	errTooManyRecursiveExpansions = errors.New("too many recursive expansions")
 )
@@ -43,7 +51,7 @@ var (
 var expandEnabledGauge = featuregate.GlobalRegistry().MustRegister(
 	"confmap.expandEnabled",
 	featuregate.StageBeta,
-	featuregate.WithRegisterDescription("controls whether expending embedded external config providers URIs"))
+	featuregate.WithRegisterDescription("controls whether expanding embedded external config providers URIs"))
 
 // Resolver resolves a configuration as a Conf.
 type Resolver struct {
@@ -168,6 +176,7 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 		}
 		retMap = NewFromStringMap(cfgMap)
 	}
+
 	// Apply the converters in the given order.
 	for _, confConv := range mr.converters {
 		if err := confConv.Convert(ctx, retMap); err != nil {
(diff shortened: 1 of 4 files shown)


confmap/resolver.go
```diff
@@ -234,28 +241,38 @@ func (mr *Resolver) expandValueRecursively(ctx context.Context, value interface{
...
+			nchanged = nchanged || changed
+			nerr = multierr.Append(nerr, err)
+			if err != nil {
+				return ""
+			}
+			switch val := ret.(type) {
+			case string:
+				return val
+			default:
+				nerr = multierr.Append(nerr, fmt.Errorf("expending %v, expected string value type, got %T", s, val))
```
mx-psi, line 269:
```suggestion
				nerr = multierr.Append(nerr, fmt.Errorf("expanding %v, expected string value type, got %T", s, val))
```

confmap/resolver_test.go
```diff
@@ -423,6 +423,45 @@ func TestResolverExpandMapAndSliceValues(t *testing.T) {
...
 
+func TestResolverExpandStringValues(t *testing.T) {
+	provider := newFakeProvider("input", func(context.Context, string, WatcherFunc) (*Retrieved, error) {
+		return NewRetrieved(map[string]any{
+			"test_no_match_old":             "${HOST}:${PORT}",
+			"test_no_match_old_no_brackets": "${HOST}:$PORT",
+			"test_match_value":              "${env:COMPLEX_VALUE}",
+			"test_match_embedded":           "${env:HOST}:3043",
+			"test_match_embedded_multi":     "${env:HOST}:${env:PORT}",
+		})
```
mx-psi, line 434: Should we add a test case with a literal both at the beginning and at the end of the string? Say, `http://${env:HOST}:3043`

confmap/resolver_test.go
```diff
@@ -507,6 +546,55 @@ func TestResolverExpandInvalidOpaqueValue(t *testing.T) {
...
+
+	testProvider := newFakeProvider("test", func(context.Context, string, WatcherFunc) (*Retrieved, error) {
+		return NewRetrieved(3043)
+	})
+
+	resolver, err := NewResolver(ResolverSettings{URIs: []string{"input:"}, Providers: makeMapProvidersMap(provider, testProvider), Converters: nil})
+	require.NoError(t, err)
+
+	_, err = resolver.Resolve(context.Background())
+	assert.EqualError(t, err, `expending ${test:PORT}, expected string value type, got int`)
```
mx-psi, line 578:
```suggestion
	assert.EqualError(t, err, `expanding ${test:PORT}, expected string value type, got int`)
```

bogdandrutu: Done.

confmap/resolver.go
bogdandrutu: Done.

confmap/resolver_test.go
bogdandrutu: Done.

confmap/resolver.go
```diff
@@ -234,28 +241,38 @@ func (mr *Resolver) expandValueRecursively(ctx context.Context, value interface{
...
+			nerr = multierr.Append(nerr, err)
+			if err != nil {
+				return ""
+			}
+			switch val := ret.(type) {
+			case string:
+				return val
+			default:
+				nerr = multierr.Append(nerr, fmt.Errorf("expanding %v, expected string value type, got %T", s, val))
+				return v
```
dmitryax, line 270: Should we make the best effort to convert the value to string here? I'm not sure we can expand something like `${env:HOST}:${env:PORT}` this way because the port probably will be a integer

bogdandrutu: That is not true, every env var is a string see https://pkg.go.dev/os#Getenv

dmitryax: Getenv returns string, but we apply yaml.Unmarshal on it :) https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider/internal/provider.go#L28

dmitryax: We should probably do the same even if it's just wrapped into the string quotes`"${env:PORT}"`

dmitryax (approved):
I'm good with merging as is, but we will need to figure out how to enforce env var strings going forward. Maybe as a separate issue

confmap/resolver_test.go
```diff
@@ -423,6 +423,120 @@ func TestResolverExpandMapAndSliceValues(t *testing.T) {
...
+			})
+
+			testProvider := newFakeProvider("env", func(_ context.Context, uri string, _ WatcherFunc) (*Retrieved, error) {
+				switch uri {
+				case "env:COMPLEX_VALUE":
+					return NewRetrieved([]any{"localhost:3042"})
+				case "env:HOST":
+					return NewRetrieved("localhost")
+				case "env:PORT":
+					return NewRetrieved("3044")
```
dmitryax, line 482: I don't think this is what we get from `${env:HOST}`. See https://go.dev/play/p/0mEbAsegFKe

confmap/resolver.go
bogdandrutu: Done. Added support to all supported values in retrieve as well.

confmap/resolver.go
```diff
@@ -234,26 +243,44 @@ func (mr *Resolver) expandValueRecursively(ctx context.Context, value any) (any,
...
+			switch val.Kind() {
+			case reflect.String:
+				return val.String()
+			case reflect.Int, reflect.Int32, reflect.Int64:
+				return strconv.FormatInt(val.Int(), 10)
+			case reflect.Float32, reflect.Float64:
+				return strconv.FormatFloat(val.Float(), 'f', -1, 64)
+			case reflect.Bool:
+				return strconv.FormatBool(val.Bool())
+			default:
```
dmitryax, line 278: Should we also do `nil` -> `""`? This would make `"${env:UNSET}"` to behave the same as `"${UNSET}"`.

dmitryax: Otherwise, using `${env:UNSET}` would fail, which is probably fine. I just wanted to confirm

bogdandrutu: I would rather fail than using "empty".

PR #7084, opened 2023-01-31
[chore] [pdata] Split generated code into more files

dmitryax:


bogdandrutu (approved):
This is great PR, but I think would be better to have one struct per file and remove the "file" concept from the pdata gen.

dmitryax:
Makes sense. I'll apply what you suggested

dmitryax:
@bogdandrutu , updated

bogdandrutu (approved):

Issue #7078, opened 2023-01-30
Improve logging for component start errors


swiatekm:
Errors from components starting are logged in an awkward and confusing way. As the error is propagated all the way up to the command, it's only logged at the end after the collector stops. For example:

```text
2024-06-19T14:15:30.822+0200    info    service@v0.103.0/service.go:115 Setting up own telemetry...
2024-06-19T14:15:30.822+0200    info    service@v0.103.0/telemetry.go:96        Serving metrics {"address": ":8888", "level": "Normal"}
2024-06-19T14:15:30.823+0200    info    service@v0.103.0/service.go:182 Starting otelcorecol... {"Version": "0.103.0-dev", "NumCPU": 32}
2024-06-19T14:15:30.823+0200    info    extensions/extensions.go:35     Starting extensions...
2024-06-19T14:15:30.823+0200    info    service@v0.103.0/service.go:245 Starting shutdown...
2024-06-19T14:15:30.823+0200    info    extensions/extensions.go:61     Stopping extensions...
2024-06-19T14:15:30.823+0200    info    service@v0.103.0/service.go:259 Shutdown complete.
Error: cannot start pipelines: failed to resolve authenticator "oauth2client": authenticator not found
2024/06/19 14:15:30 collector server run finished with error: cannot start pipelines: failed to resolve authenticator "oauth2client": authenticator not found
```

Could we log these *before* we shut down, in addition to the current behaviour? I swear it used to work like this in the past, not sure when it changed.

PR #7086, opened 2023-01-31
Improve logging of component start errors

swiatekm-sumo:
**Description:**
Improved logging of component start errors, so the log lines appear where they chronologically happen, rather than at the end. They also use the component logger, which makes them look nicer.

Output before change:
```text
2023-01-31T18:40:52.517+0100    info    service/pipelines.go:86 Starting exporters...
2023-01-31T18:40:52.517+0100    info    service/pipelines.go:90 Exporter is starting... {"kind": "exporter", "data_type": "traces", "name": "otlp"}
2023-01-31T18:40:52.518+0100    info    service/service.go:155  Starting shutdown...
...
Error: cannot start pipelines: failed to resolve authenticator "not_there": authenticator not found; failed to shutdown pipelines: no existing monitoring routine is running; no existing monitoring routine is running
2023/01/31 18:40:52 collector server run finished with error: cannot start pipelines: failed to resolve authenticator "not_there": authenticator not found; failed to shutdown pipelines: no existing monitoring routine is running; no existing monitoring routine is running
```

Output after change:
```text
2023-01-31T18:35:14.471+0100    info    service/pipelines.go:87 Starting exporters...
2023-01-31T18:35:14.471+0100    info    service/pipelines.go:91 Exporter is starting... {"kind": "exporter", "data_type": "metrics", "name": "otlp"}
2023-01-31T18:35:14.471+0100    error   service/pipelines.go:93 Failed to start exporter        {"kind": "exporter", "data_type": "metrics", "name": "otlp", "error": "failed to resolve authenticator \"not_there\": authenticator not found"}
2023-01-31T18:35:14.471+0100    info    service/service.go:155  Starting shutdown...
...
Error: cannot start pipelines: failed to resolve authenticator "not_there": authenticator not found; failed to shutdown pipelines: no existing monitoring routine is running; no existing monitoring routine is running
2023/01/31 18:35:14 collector server run finished with error: cannot start pipelines: failed to resolve authenticator "not_there": authenticator not found; failed to shutdown pipelines: no existing monitoring routine is running; no existing monitoring routine is running
```

**Link to tracking Issue:** #7078

### DIFF
--- a/service/extensions/extensions.go
+++ b/service/extensions/extensions.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 
 	"go.uber.org/multierr"
+	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/extension"
@@ -43,6 +44,7 @@ func (bes *Extensions) Start(ctx context.Context, host component.Host) error {
 		extLogger := components.ExtensionLogger(bes.telemetry.Logger, extID)
 		extLogger.Info("Extension is starting...")
 		if err := ext.Start(ctx, components.NewHostWrapper(host, extLogger)); err != nil {
+			extLogger.WithOptions(zap.AddStacktrace(zap.DPanicLevel)).Error("Failed to start extension", zap.Error(err))
 			return err
 		}
 		extLogger.Info("Extension started.")
--- a/service/pipelines.go
+++ b/service/pipelines.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 
 	"go.uber.org/multierr"
+	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/connector"
@@ -89,6 +90,7 @@ func (bps *builtPipelines) StartAll(ctx context.Context, host component.Host) er
 			expLogger := components.ExporterLogger(bps.telemetry.Logger, expID, dt)
 			expLogger.Info("Exporter is starting...")
 			if err := exp.Start(ctx, components.NewHostWrapper(host, expLogger)); err != nil {
+				expLogger.WithOptions(zap.AddStacktrace(zap.DPanicLevel)).Error("Failed to start exporter", zap.Error(err))
 				return err
 			}
 			expLogger.Info("Exporter started.")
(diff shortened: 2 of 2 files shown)


service/extensions/extensions.go
```diff
@@ -43,6 +44,7 @@ func (bes *Extensions) Start(ctx context.Context, host component.Host) error {
 		extLogger := components.ExtensionLogger(bes.telemetry.Logger, extID)
 		extLogger.Info("Extension is starting...")
 		if err := ext.Start(ctx, components.NewHostWrapper(host, extLogger)); err != nil {
+			extLogger.WithOptions(zap.AddStacktrace(zap.DPanicLevel)).Error("Failed to start extension", zap.Error(err))
```
swiatekm-sumo, line 47: The reason we're adding the `AddStacktrace` option here, is to avoid adding the stacktrace, as it makes the log less readable and doesn't help much in most cases.

I haven't found a more idiomatic way of doing this with zap, and this looks a bit ugly.

swiatekm-sumo:
Do I need a changelog entry for this?

service/extensions/extensions.go
bogdandrutu: But on the line 45 we don't have it. Why do we need it here?

bogdandrutu: Also, instead of logging it here, we should just enhance the error:

return fmt.Errorf("failed to start exporter %q: %w", extID.String(), err)

swiatekm-sumo: > But on the line 45 we don't have it. Why do we need it here?

The logger is configured so errors add stacktraces, which I want to avoid here. I could also log an Info or Warning, but that feels like a lie. It's an error, I just don't want to put a stacktrace in the middle of the output from other components.

> Also, instead of logging it here, we should just enhance the error:
> 
> return fmt.Errorf("failed to start exporter %q: %w", extID.String(), err)

That will only enhance the error we see at the very end. The error we return here is ultimately packaged with errors from `.Shutdown()` and returned from `cobra.Command`. That'd still be an improvement, but in my opinion this way makes clear where the error happened, which helps new users in particular.

bogdandrutu: I don't think we should "log" every error where it happens, instead we should return better errors that describe what happened. That is my philosophy in general, to avoid logging all over the place...

swiatekm-sumo: How about logging the error from `.Start()` before we call `.Shutdown()`? Part of the problem here is that if we error while starting components, then we don't have telemetry set up, `.Shutdown()` returns its own (spurious) errors, and we end up with a monster like this:
```
2023/01/31 18:40:52 collector server run finished with error: cannot start pipelines: failed to resolve authenticator "not_there": authenticator not found; failed to shutdown pipelines: no existing monitoring routine is running; no existing monitoring routine is running
```

> I don't think we should "log" every error where it happens, instead we should return better errors that describe what happened. That is my philosophy in general, to avoid logging all over the place...

I don't think it needs to happen in the same place *in the code*, but I do think that if we want to log an error, we should do it before we start logging other unrelated information. My problem with this situation is that it proceeds as follows:

1. Get an error during component start
2. Shutdown, producing more log output unrelated to the error above
3. Log the component start error along with other errors encountered during shutdown

PR #7091, opened 2023-02-02
[pdata] Deprecate [Metrics|Logs|Traces].MoveTo methods

dmitryax:
The methods don't provide any value and can be replaced with simple assignments.

This also unblocks https://github.com/open-telemetry/opentelemetry-collector/pull/7088


bogdandrutu:
> This also unblocks https://github.com/open-telemetry/opentelemetry-collector/pull/7088

Not true, I was wrong yesterday, indeed unblocks on the source side, but destination wise you have same problem also for CopyTo.

bogdandrutu (approved):

PR #7094, opened 2023-02-02
[telemetry] Add disabled flag on each telemetry type, and sync logger on shutdown.

atoulme:
**Description:** 
Add disabled flag on each telemetry type, and sync logger on shutdown.

Remove TODO and apply sync on telemetry shutdown

### DIFF
--- a/service/service_test.go
+++ b/service/service_test.go
@@ -263,19 +263,13 @@ func testCollectorStartHelper(t *testing.T, useOtel bool, useGraph bool, tc ownM
 	cfg := newNopConfig()
 	cfg.Telemetry.Logs.Level = zapcore.InfoLevel
 	cfg.Telemetry.Logs.Encoding = "console"
-	cfg.Telemetry.Logs.Enabled = func() *bool {
-		falsePtr := true
-		return &falsePtr
-	}()
+	cfg.Telemetry.Logs.Disabled = false
 	cfg.Telemetry.Logs.ErrorOutputPaths = []string{}
 	cfg.Telemetry.Logs.OutputPaths = []string{}
 
 	cfg.Extensions = []component.ID{component.NewID("zpages")}
 	cfg.Telemetry.Metrics.Address = metricsAddr
-	cfg.Telemetry.Metrics.Enabled = func() *bool {
-		truePtr := true
-		return &truePtr
-	}()
+	cfg.Telemetry.Metrics.Disabled = false
 
 	cfg.Telemetry.Resource = make(map[string]*string)
 	// Include resource attributes under the service::telemetry::resource key.
@@ -304,10 +298,7 @@ func testCollectorStartHelper(t *testing.T, useOtel bool, useGraph bool, tc ownM
 func TestServiceTelemetryRestart(t *testing.T) {
 	// Create a service
 	cfg := newNopConfig()
-	cfg.Telemetry.Metrics.Enabled = func() *bool {
-		truePtr := true
-		return &truePtr
-	}()
+	cfg.Telemetry.Metrics.Disabled = false
 	srvOne, err := New(context.Background(), newNopSettings(), cfg)
 	require.NoError(t, err)
 
@@ -447,13 +438,12 @@ func newNopConfig() Config {
 }
 
 func newNopConfigPipelineConfigs(pipelineCfgs map[component.ID]*PipelineConfig) Config {
-	enabledPtr := false
 	return Config{
 		Extensions: []component.ID{component.NewID("nop")},
 		Pipelines:  pipelineCfgs,
 		Telemetry: telemetry.Config{
 			Logs: telemetry.LogsConfig{
-				Enabled:     &enabledPtr,
+				Disabled:    true,
 				Level:       zapcore.InfoLevel,
 				Development: false,
 				Encoding:    "console",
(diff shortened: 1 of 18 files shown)


service/service_test.go
```diff
@@ -425,8 +425,8 @@ func newNopConfig() Config {
 					Initial:    100,
 					Thereafter: 100,
 				},
-				OutputPaths:       []string{"stderr"},
-				ErrorOutputPaths:  []string{"stderr"},
```
djaglowski, line 429: It's not obvious to me how this is related. Do you mind clarifying for me?

atoulme: Tests fail if we try to sync, because /dev/stderr is not available, most likely because the process doesn’t run with tty. Removing the output fixed the issue.

djaglowski:
Looks like the test failures may be related to the change.

atoulme:
I'll take a look, thanks.

bogdandrutu (approved):

atoulme:
Yes, I'm looking more into it. There is a bit of an issue with how we manage situations where we test with this logger because stderr is not available from within tests.
I tried to disable going to stderr with an empty array, but in that case the default applies.
I have some code towards disabling logging explicitly (level: none). It expands the scope of the PR a bit and I have at least one test that depends on logger emitting at least once that fails. I'll keep looking.

service/service_test.go
atoulme: So that was a bit painful... going to make a bigger comment on the PR itself with a follow up.

atoulme:
It turns out our tests here would fail because stderr is set by default, if no value is provided. I have developed therefore a different approach for testing: creating a log level of none where we just return a zap.NewNop() instead of building the logger.

With that change, and applying that change to the yaml config files used in tests, we get all tests to pass.

djaglowski (approved):

bogdandrutu:
> It turns out our tests here would fail because stderr is set by default, if no value is provided. I have developed therefore a different approach for testing: creating a log level of none where we just return a zap.NewNop() instead of building the logger.

Do I understand correctly that we added the new public API just to fix tests?

atoulme:
> Do I understand correctly that we added the new public API just to fix tests?
I guess there isn't any way to disable logging entirely, so maybe there's value to this beyond tests here.

bogdandrutu (commented):
Please check what is the "standard" we use enable vs disable in other places.

service/telemetry/config.go
```diff
@@ -107,6 +111,10 @@ type LogsSamplingConfig struct {
 // MetricsConfig exposes the common Telemetry configuration for one component.
 // Experimental: *NOTE* this structure is subject to change or removal in the future.
 type MetricsConfig struct {
+	// Enabled enables metrics collection.
+	// (default = true)
+	Enabled *bool `mapstructure:"enabled"`
```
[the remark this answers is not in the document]
atoulme: I'll fix it.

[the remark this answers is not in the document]
atoulme: Done.

codeboten:
> @codeboten do we go with "Enable" or "Disable" int he config otep?

@bogdandrutu the otep currently proposes `disabled`

atoulme:
I didn't know there was an otep. Got a link?

Mind approving the PR?

dmitryax:
We don't use `disable` anywhere in the collector repos. I don't think we should start using `disabled` without planning the migration if we want to migrate. I commented on the Otep https://github.com/open-telemetry/oteps/pull/225/files#r1139485969

jpkrohling:
Just to record somewhere: I prefer "enabled" instead of "disabled", as "disabled" requires a double-negative to have something activated (disabled=false means enabled).

atoulme:
Trying to understand what is blocking there. Are we stuck on disabled/enabled?

dmitryax:
> Trying to understand what is blocking there. Are we stuck on disabled/enabled?

I think tho. If we want to be consistent within Collector, we need to use `enabled`. But if the Configuration SIG chooses a different strategy, we would need to change it in all the places. Having one more of them is probably not that critical. I'd say we can go with `enabled` here. @codeboten @jpkrohling @bogdandrutu WDYT?

atoulme:
I think enabled makes more sense. To keep backwards compatibility, we should assume enabled : true by default. The issue with that is we then need to use a pointer boolean, which earlier in this PR gave pause to Bogdan. We use a pointer boolean to find whether someone set the boolean expressly to false or if the value is not set (nil pointer).

dmitryax:
> I think enabled makes more sense. To keep backwards compatibility, we should assume enabled : true by default. The issue with that is we then need to use a pointer boolean, which earlier in this PR gave pause to Bogdan. We use a pointer boolean to find whether someone set the boolean expressly to false or if the value is not set (nil pointer).

We can keep backward compatibility for end users without pointer. Default values are set https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/unmarshaler.go#L43, not ideal, but that's what we have

jpkrohling (commented):
I agree that we should pursue `enabled` instead of `disabled` unless absolutely necessary.

tigrannajaryan:
We have `logs.level` and `metrics.level`. `metrics.level` already supports `none`. We can add `none` to `logs.level` too and then enabled/disabled won't be necessary anymore.

atoulme:
> We have `logs.level` and `metrics.level`. `metrics.level` already supports `none`. We can add `none` to `logs.level` too and then enabled/disabled won't be necessary anymore.

That was discussed and was my first array into this work, but unfortunately there was pushback as maintainers said they want log level to be zap log levels.

PR #7096, opened 2023-02-02
Simplify forward connector by removing sharedcomponent

djaglowski:
There isn't really any reason to use `sharedcomponent` in the forward connector because there is no shared state or interaction between signals.


bogdandrutu (approved):

djaglowski:
I need to remove a few more bits here.

djaglowski:
Skipped contrib tests because they are failing for an unrelated reason. This is ready to go otherwise.

Issue #6331, opened 2022-10-17
queue retry: support tiered queue


hanjm:
**Is your feature request related to a problem? Please describe.**
I'm adopting queue retry with [filestorage](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage) in collector, but i found significant performance downgrade when enable filestorage, every otel data will write to filestorage and read from filestorage.

**Describe the solution you'd like**
Currently implementation is one of memory queue and disk queue.
https://github.com/open-telemetry/opentelemetry-collector/blob/ee02bc7b2bbc71781cb96bcca79c9b387096dcf7/exporter/exporterhelper/queued_retry.go#L123-L126
https://github.com/open-telemetry/opentelemetry-collector/blob/ee02bc7b2bbc71781cb96bcca79c9b387096dcf7/exporter/exporterhelper/queued_retry.go#L155-L166
Should we implement a tiered queue, memory queue as primary queue, disk queue as secondary queue.  

if the memory queue is full when export endpoint down, will use disk queue.

**Describe alternatives you've considered**

**Additional context**
I known [nsq](https://nsq.io/overview/internals.html#topics-and-channels) use tiered queue. 
![image](https://user-images.githubusercontent.com/7621241/196080357-a0ad27db-23f2-417f-ac29-4e54a86a5124.png)

jefchien:
Please assign this to me. I was going to open a similar issue and have already written something which defaults to the current behavior (one queue if enabled), but is configurable and can have a separate "backlog" queue enabled. It also handles sending overflow requests between the two.

swiatekm-sumo:
This setup compromises on durability for performance, right? Items in the main queue will be lost if the collector is suddenly killed. I think this is fine, but should be documented.

jefchien:
If the collector is shutdown gracefully, then the items in the main queue (assuming the main queue is the in-memory queue) will drain and get sent to the secondary backlog one, which will write it to disk. You're correct though that it is a compromise.

dmitryax:
Instead of losing durability, maybe we can introduce a configurable memory buffer in the persistent queue for now? We will still be writing into the storage before accepting the data, but at least we won't be reading/unmarshalling from the persistence storage while the memory buffer has capacity. It should improve the performance. We'll can if that is enough...

cc @jefchien @swiatekm-sumo

jefchien:
[...]

PR #7098, opened 2023-02-02
[exporterhelper] Add optional tiered sender to coordinate requests between two queues

jefchien:
**Description:**
The base exporter used in the `exporterhelper` has been changed to use a tiered sender instead of a queued retry sender. The tiered sender is optionally comprised of a primary queued retry sender (configured by `WithQueue`) and a backlog queued retry sender (configured by `WithBacklog`). Each request it receives gets sent to the primary queue first and if the primary queue has overflow, the earliest requests will be removed and sent to the backlog. If requeuing is enabled, the requeued requests will also be sent to the backlog.

The primary use-case for this is to use an in-memory queue for the primary and a persistent queue for the backlog. On shutdown, the in-memory queue will flush and send all of its requests to the backlog persistent queue. This way the primary queue doesn't need to be persistent, which can improve performance.

The default behavior remains the same. If the backlog queue is not enabled through configuration, then it will behave the same as before based on the queue and retry settings.

Example configuration:
```yaml
exporters:
  otlp:
    sending_queue:
      enabled: true
      num_consumers: 10
      queue_size: 50
    backlog_queue:
      enabled: true
      num_consumers: 1
      queue_size: 100
      storage: file_storage/otc
extensions:
  file_storage/otc:
    directory: /var/lib/storage/otc
    timeout: 10s
```

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/6331

**Testing:** Added unit tests to test overflow behavior. Added tests for tiered sender.

**Documentation:** N/A

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -18,6 +18,9 @@ import (
 	"context"
 	"time"
 
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/exporter"
@@ -74,6 +77,7 @@ type baseSettings struct {
 	consumerOptions []consumer.Option
 	TimeoutSettings
 	QueueSettings
+	BacklogSettings QueueSettings
 	RetrySettings
 }
 
@@ -83,7 +87,8 @@ func fromOptions(options ...Option) *baseSettings {
 	opts := &baseSettings{
 		TimeoutSettings: NewDefaultTimeoutSettings(),
 		// TODO: Enable queuing by default (call DefaultQueueSettings)
-		QueueSettings: QueueSettings{Enabled: false},
+		QueueSettings:   QueueSettings{Enabled: false},
+		BacklogSettings: QueueSettings{Enabled: false},
 		// TODO: Enable retry by default (call DefaultRetrySettings)
 		RetrySettings: RetrySettings{Enabled: false},
 	}
@@ -138,6 +143,15 @@ func WithQueue(queueSettings QueueSettings) Option {
 	}
 }
 
+// WithBacklog overrides the default BacklogSettings for an exporter.
+// The default BacklogSettings is to disable the secondary backlog queue.
+// Requires QueueSettings to be enabled.
+func WithBacklog(queueSettings QueueSettings) Option {
+	return func(o *baseSettings) {
+		o.BacklogSettings = queueSettings
+	}
+}
+
 // WithCapabilities overrides the default Capabilities() function for a Consumer.
 // The default is non-mutable data.
 // TODO: Verify if we can change the default to be mutable as we do for processors.
@@ -151,9 +165,8 @@ func WithCapabilities(capabilities consumer.Capabilities) Option {
 type baseExporter struct {
 	component.StartFunc
 	component.ShutdownFunc
-	obsrep   *obsExporter
-	sender   requestSender
-	qrSender *queuedRetrySender
+	obsrep *obsExporter
+	sender *tieredSender
 }
 
 func newBaseExporter(set exporter.CreateSettings, bs *baseSettings, signal component.DataType, reqUnmarshaler internal.RequestUnmarshaler) (*baseExporter, error) {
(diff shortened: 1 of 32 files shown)


jefchien:
@codeboten Sorry about all of the commits. Had to fix code coverage and a failing test. Please take a look when you get the chance.

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -200,6 +211,28 @@ func TestZeroSize(t *testing.T) {
...
+}
+
+func TestOverflow(t *testing.T) {
+	q := NewBoundedMemoryQueue(1)
+	cs := newConsumerState(t)
+	q.OnOverflow(func(item Request) {
+		cs.record(item.(stringRequest).str)
+	})
+	assert.True(t, q.Produce(newStringRequest("a")))
+	assert.True(t, q.Produce(newStringRequest("b")))
```
codeboten, line 232: What's the expectation of what happens to this request? Should the assert here be false as the queue should be overflowing?

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -74,7 +75,17 @@ func (q *boundedMemoryQueue) Produce(item Request) bool {
...
-		return false
+		if q.overflow == nil {
+			return false
+		}
+		select {
+		case old := <-q.items:
+			q.size.Add(^uint32(0))
+			q.overflow(old)
+		default:
+			q.overflow(item)
```
codeboten, line 86: The docstring for this method says it returns false if the queue overflows, that's no longer the case here. The docs should be updated

exporter/exporterhelper/tiered.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+	if ts.backlog != nil {
+		if err := ts.backlog.start(ctx, host); err != nil {
+			return err
+		}
+		ts.backlog.queue.OnOverflow(func(req internal.Request) {
+			ts.logger.Error(
+				"Backlog is full. Dropping data.",
+				zap.Int("dropped_items", req.Count()),
```
codeboten, line 79: This should be captured in a metric available through obsreport

exporter/exporterhelper/tiered.go
```diff
@@ -0,0 +1,116 @@
...
+			)
+		})
+		ts.primary.queue.OnOverflow(func(req internal.Request) {
+			if err := ts.backlog.send(req); err != nil {
+				ts.logger.Error(
+					"Unable to add overflow to backlog. Dropping data.",
+					zap.Error(err),
+					zap.Int("dropped_items", req.Count()),
```
codeboten, line 87: This should also be available through obsreport

codeboten (commented):
Overall the code looks good, just a few comments/questions.

exporter/exporterhelper/common.go
```diff
@@ -138,6 +143,15 @@ func WithQueue(queueSettings QueueSettings) Option {
 	}
 }
 
+// WithBacklog overrides the default BacklogSettings for an exporter.
+// The default BacklogSettings is to disable the secondary backlog queue.
+// Requires QueueSettings to be enabled.
```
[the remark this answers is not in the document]
jefchien: Yes that's correct. If you configure the backlog queue without a queue enabled, it'll ignore it. There's a unit test for this scenario.

exporter/exporterhelper/internal/bounded_memory_queue_test.go
jefchien: It overflows, but because we have an `OnOverflow` callback configured (which is only configured if there's a backlog), the head of the queue is removed and sent to it. So, the second request is successfully added to the queue and `Produce` returns true. In this test, we record the request that the overflow callback receives and assert that the correct one overflows.

```
// q.Produce(newStringRequest("a"))
capacity: 1
insert: "a"
before:
  queue: []
after:
  queue: ["a"]

// q.Produce(newStringRequest("b"))
capacity: 1
insert: "b"
before:
  queue: ["a"]
after:
  queue: ["b"]
  overflow: "a"
```

exporter/exporterhelper/tiered.go
jefchien: Should they be separate metrics or is a dropped items count sufficient?

exporter/exporterhelper/tiered.go
jefchien: Added a dropped items metric to the obsreport.

dmitryax:
Can you please update https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md ?

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -72,7 +74,17 @@ func (q *boundedMemoryQueue) Produce(item Request) bool {
...
 	// should match the capacity of the new queue
 	if q.size.Load() >= q.capacity {
-		return false
+		if q.overflow == nil {
+			return false
+		}
+		select {
+		case old := <-q.items:
+			q.size.Add(^uint32(0))
+			q.overflow(old)
```
dmitryax, line 83: Can you please clarify why this is needed?

jefchien: If the queue is full and a new request is added to the queue, the head of the queue will be removed and sent through the overflow callback. This prioritizes live requests in the primary queue and relegates older requests and retry requests to the backlog in the case of queue overflow.

tigrannajaryan:
> This way the primary queue doesn't need to be persistent, which can improve performance.

I am curious if there are any benchmarks comparing the old and new behaviors.

swiatekm-sumo:
> > This way the primary queue doesn't need to be persistent, which can improve performance.
> 
> I am curious if there are any benchmarks comparing the old and new behaviors.

I don't have benchmarks, but I do believe this setup would allow us to have persistence in situations where the storage is either expensive or not very performant.

In the current persistent queue, every batch needs to be written to disk before it can be sent. This makes it durable, but causes a lot of write activity, which makes it difficult to use on networked file systems, for example. But if we could avoid writing to disk if we don't need to (as in, the consumers are faster than the producers), then a non-performant storage would be much less of a problem. You might even be able to use something like a S3 storage, which is impossible with the current design.

jefchien:
> > This way the primary queue doesn't need to be persistent, which can improve performance.
> 
> I am curious if there are any benchmarks comparing the old and new behaviors.

Sorry was out and just got some time to look at this again. I also don't have any benchmarks. Do you have any suggestions/ideas for what would be sufficient or useful?

tigrannajaryan:
> Sorry was out and just got some time to look at this again. I also don't have any benchmarks. Do you have any suggestions/ideas for what would be sufficient or useful?

It would be very useful benchmarks that compare performance of old (before this PR) and new (after this PR) implementations :-)

I don't have specific suggestions on what scenario exactly the benchmark should perform.

dmitryax:
What if we deprecate `sending_queue.storage` and just introduce `persistent_sending_queue` option with the same parameters? In that case, if both available, this behavior will be triggered? The example would be:
```
exporters:
  otlp:
    sending_queue:
      enabled: true
      num_consumers: 10
      queue_size: 50
    persistent_sending_queue:
      enabled: true
      num_consumers: 1
      queue_size: 100
      storage: file_storage/otc
extensions:
  file_storage/otc:
    directory: /var/lib/storage/otc
    timeout: 10s
```

PR #7106, opened 2023-02-03
[telemetry] allow to filter logs using zapfilter rules or a regular expression

atoulme:
**Description:**
Add 2 new config options for logs:
* `filter_rules`: taken after the zapfilter library, allows to apply filtering to logs.
* `filter_regex`: a regular regular expression. Any matched log (message only) is excluded.

**Link to tracking Issue:**
#7107

**Testing:**
Unit tests.

**Documentation:**
Added to go docs.

### DIFF
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -94,6 +94,23 @@ type LogsConfig struct {
 	//
 	// By default, there is no initial field.
 	InitialFields map[string]any `mapstructure:"initial_fields"`
+
+	// Filtering rules to exclude log statements. See https://github.com/moul/zapfilter
+	// for documentation.
+	//
+	// Example:
+	//
+	//      filter_rules: "*:myns info,warn:myns.* error:*"
+	// By default, there is no filter.
+	FilterRules string `mapstructure:"filter_rules"`
+
+	// Filtering regular expression to exclude log statements.
+	//
+	// Example:
+	//
+	//      filter_regex: "partition.*"
+	// By default, there is no regular expression filter.
+	FilterRegex string `mapstructure:"filter_regex"`
 }
 
 // LogsSamplingConfig sets a sampling strategy for the logger. Sampling caps the
--- a/.chloggen/filter-log.yaml
+++ b/.chloggen/filter-log.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: telemetry
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add zapfilter and regular expression filtering capabilities for logs
+
+# One or more tracking issues or pull requests related to the change
+issues: [7107]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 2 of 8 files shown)


atoulme:
This is built on top of #7094, which should be looked at first.

dmitryax:
Please update the issue link, it points to the PR itself.

atoulme:
I assumed that was ok based on [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/18205#issuecomment-1410974587) - will fix.

dmitryax:
> I assumed that was ok based on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/18205#issuecomment-1410974587 - will fix.

That link is in the changelog, not PR description. This change introduces a bunch of public API that we are trying to reduce and stabilize for 1.0, so we need to understand the problem that is intended to be solved by this.

atoulme:
No worries. I have created the issue and updated links.

atoulme:
Closing as per SIG discussion, we will use #7223's approach instead.

atoulme:
Reopening since https://github.com/open-telemetry/opentelemetry-collector/issues/7223 is abandoned.

service/telemetry/config.go
```diff
@@ -94,6 +94,23 @@ type LogsConfig struct {
...
 	InitialFields map[string]any `mapstructure:"initial_fields"`
+
+	// Filtering rules to exclude log statements. See https://github.com/moul/zapfilter
+	// for documentation.
+	//
+	// Example:
+	//
+	//      filter_rules: "*:myns info,warn:myns.* error:*"
+	// By default, there is no filter.
+	FilterRules string `mapstructure:"filter_rules"`
```
dmitryax, line 105: This field name seems a bit unclear, and we don't have a pattern like this in other collector configurations. The closest one is stanza include/exclude https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver#configuration. Maybe we adopt the same `exclude` name here?

service/telemetry/config.go
```diff
@@ -94,6 +94,23 @@ type LogsConfig struct {
...
+
+	// Filtering regular expression to exclude log statements.
+	//
+	// Example:
+	//
+	//      filter_regex: "partition.*"
+	// By default, there is no regular expression filter.
+	FilterRegex string `mapstructure:"filter_regex"`
```
dmitryax, line 113: Do we need regexp? Can we keep the blob only for now? If we will need regexp, we can apply the same approach as in the signalfx exporter and read values like `/regexp/` from the same `exclude` field.

atoulme: We don't have a blob approach right now, did I miss something?

service/telemetry/config.go
atoulme: This is exposing a zapfilter ruleset. From their [README](https://github.com/moul/zapfilter):
```
*                            everything
        *:*                          everything
        info:*                       level info;  any namespace
        info+:*                      levels info, warn, error, dpanic, panic, and fatal; any namespace
        info,warn:*                  levels info, warn; any namespace
        ns1                          any level; namespace 'ns1'
        *:ns1                        any level; namespace 'ns1'
        ns1*                         any level; namespaces matching 'ns1*'
        *:ns1*                       any level; namespaces matching 'ns1*'
        *:ns1,ns2                    any level; namespaces 'ns1' and 'ns2'
        *:ns*,-ns3*                  any level; namespaces matching 'ns*' but not matching 'ns3*'
        info:ns1                     level info; namespace 'ns1'
        info,warn:ns1,ns2            levels info and warn; namespaces 'ns1' and 'ns2'
        info:ns1 warn:n2             level info + namespace 'ns1' OR level warn and namespace 'ns2'
        info,warn:myns* error+:*     levels info or warn and namespaces matching 'myns*' OR levels error, dpanic, panic or fatal for any namespace
```

This is not a glob approach - this is specifically allowing the namespaces and levels set in the rules.

dwalczakfx:
Can we reopen this ?

PR #7110, opened 2023-02-03
Disambiguate component ID in 'connectortest' package

djaglowski:
Use a different ID than `receivertest` and `exportertest` to avoid ambiguous configuration scenarios. Ambiguous IDs are detected in the `otelcol` package, but lower level packages such as `service` assume that IDs are disambiguated.

To demonstrate this working with the connectors feature gate enabled, I've temporarily updated the gate level to beta. Once tests pass, I will revert the gate level to alpha.


djaglowski:
All checks have passed with the feature gate enabled: https://github.com/open-telemetry/opentelemetry-collector/pull/7110/checks?check_run_id=11095874904

bogdandrutu (approved):

PR #7109, opened 2023-02-03
Add graph-based implemenation of HandleZPages

djaglowski:
Follows #7108

PR #7114, opened 2023-02-03
Update connectors flag and documentation

djaglowski:
This PR is preparation for the v0.71.0 release, which would be the first with a functional implementation of the connectors framework.
- Changes the feature gate from `enableConnectors` to `connectors`, which makes more sense in the long run because the enabled status is independent of the gate's _name_.
- Adds a note to the connectors README about the status of the feature set and how to enable.
- Adds a changelog entry with a basic description of the feature set and a link to the README.

### DIFF
--- a/internal/sharedgate/sharedgate.go
+++ b/internal/sharedgate/sharedgate.go
@@ -18,7 +18,7 @@ package sharedgate // import "go.opentelemetry.io/collector/internal/sharedgate"
 import "go.opentelemetry.io/collector/featuregate"
 
 var ConnectorsFeatureGate = featuregate.GlobalRegistry().MustRegister(
-	"enableConnectors",
+	"connectors",
 	featuregate.StageAlpha,
 	featuregate.WithRegisterDescription("Enables 'connectors', a new type of component for transmitting signals between pipelines."),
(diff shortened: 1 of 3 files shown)


djaglowski:
This should not be merged before #7109

dmitryax (approved):

internal/sharedgate/sharedgate.go
```diff
@@ -18,7 +18,7 @@ package sharedgate // import "go.opentelemetry.io/collector/internal/sharedgate"
 import "go.opentelemetry.io/collector/featuregate"
 
 var ConnectorsFeatureGate = featuregate.GlobalRegistry().MustRegister(
-	"enableConnectors",
+	"connectors",
```
bogdandrutu, line 21: I think `service.connectors` is more appropriate?

djaglowski: Updated.

djaglowski:
It would be great to get this merged before the release Monday. Ideally #7109 would also be merged so we have a complete initial implementation, but it may not be necessary as it would only affect those who are both enabling connectors and using zpages extension.

PR #7119, opened 2023-02-05
Split graph based pipelines implementation from connectors functionality

bogdandrutu:
Add a separate featuregate to control the new graph base pipelines implementation release. This way we can enable the graph base implementation now, and keep connectors still in alpha (since the validation of the otelcol Config will fail) to limit the number of changes necessary to enable connectors.


djaglowski (approved):
This doesn't need to block the PR, but it looks like we no longer need the `sharedgate` package and don't need to export the gate either.

bogdandrutu:
We still use it in otelcol and otelcoltest.

Issue #6794, opened 2022-12-13
Proposal: New way of handling data mutability


bogdandrutu:
## Background

Currently we ask every component to declare their "Capabilities". Capabilities are part of the `consumer.Consumer*` interface, to allow "helper consumer" to also declare their own capabilities see in contrib [batchperesourceattr](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/batchperresourceattr/batchperresourceattr.go#L32)

This design proved in the recent weeks to be very limited, see issues/PRs (most of the recent issues are related to sort, but this is not the only mutable func that may cause this):
* https://github.com/open-telemetry/opentelemetry-collector/issues/6420
* https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16506
* https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/16498

There are few issues with this design:
1. Currently there is no way to test components. An possible solution is to provide a different pdata package maybe using a build tag that checks for incorrect access.
2. In case of crashes like https://github.com/open-telemetry/opentelemetry-collector/issues/6420 (I was lucky that I am very familiar with the pdata and collector code to find) could take a long time to find an the root cause with the current design since we do not crash/log when UB happens (this is an UB since concurrent access to non thread safe data is defined as UB).

## Proposal

The current issue/document proposes to replace the current "Capabilities" model from the `consumer.Consumer*` interface and add a notion of "Shared(ReadOnly)" and "Exclusive(Mutable)" (Shared/Exclusive are nice since they are also used in MESI protocol) to the top signal objects `ptrace.Traces`/`pmetric.Metrics`/`plog.Logs`. Here is the example API for traces (similar changes will be applied to all):

```golang

type Traces struct {
  // state is propagated to every "sub struct" only to enforce correctness.
  state pcommon.State
}

// State returns the current access state of the data. If StateShared the object cannot be mutated, and any call that mutates 
// the current instance or a sub-struct (Span, SpanEvent, etc.) will panic. If mutability is required a copy of the object should
// be made using CopyTo (TODO: may consider a Clone func since this will become used quite some time)
func (t Traces) State() pcommon.State {
  return t.state
}

// Changes the state of the StateShared. Must be called only when the object is fan-out to multiple consumers, or in tests.
func (t *Traces) MarkShared() {
  t.state = pcommon.StateShared;
}

func (ms Traces) MoveTo(dest Traces) {
[...]

PR #7140, opened 2023-02-06
[DRAFT] [pdata] Mutable/immutable pdata with structs

dmitryax:
This is a PR alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/7088. The goal is the same: a compile-time solution to https://github.com/open-telemetry/opentelemetry-collector/issues/6794.

This approach doesn't use interfaces as suggested by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/pull/7088#discussion_r1095082529


pdata/plog/generated_scopelogs.go
```diff
@@ -28,34 +28,60 @@ import (
...
+	preventConversion struct{}
+}
+
+type commonScopeLogs struct {
 	orig *otlplogs.ScopeLogs
 }
 
-func newScopeLogs(orig *otlplogs.ScopeLogs) ScopeLogs {
-	return ScopeLogs{orig}
+func newImmutableScopeLogs(orig *otlplogs.ScopeLogs) ScopeLogs {
```
bogdandrutu, line 46: s/newImmutableScopeLogs/newScopeLogs

pdata/plog/generated_scopelogs.go
```diff
@@ -28,34 +28,60 @@ import (
...
-// Must use NewScopeLogs function to create new instances.
+// Must use NewExemplar function to create new instances.
 // Important: zero-initialized instance is not valid for use.
 type ScopeLogs struct {
+	commonScopeLogs
+}
+
+type MutableScopeLogs struct {
+	commonScopeLogs
+	preventConversion struct{}
```
bogdandrutu, line 39: What does this prevent? Want to learn this trick..

pdata/plog/generated_scopelogs.go
```diff
@@ -28,34 +28,60 @@ import (
...
+}
+
+type commonScopeLogs struct {
```
bogdandrutu, line 42: commonScopeLogs/immutableScopeLogs

pdata/plog/generated_scopelogs.go
dmitryax: it prevents the conversion between ScopeLogs MutableScopeLogs

pdata/plog/generated_scopelogs.go
dmitryax: We need to export the same helper for the common structures in the internal package where we would have `NewImmutableInstrumentationScope(orig)` -> `NewInstrumentationScope(orig)` that will conflict with `NewInstrumentationScope()`. So I kept the names consistent across pdata both for common structs (exported) and not exported.

pdata/plog/generated_scopelogs.go
dmitryax: Do you mean renaming it?

pdata/plog/generated_scopelogs.go
bogdandrutu: Sure, but people don't have access to `immutableScopeLogs` since is private. Who can do that conversion?

pdata/plog/generated_scopelogs.go
bogdandrutu: yes..

pdata/plog/generated_scopelogs.go
bogdandrutu: But it is not consistent: func new**Immutable**ScopeLogs(orig *otlplogs.ScopeLogs) ScopeLogs -> here the Immutable is extra....

dmitryax: I know. I added the redundant `Immutable` part of all the structs. We can remove in for this struct but cannot remove in from:

```
func NewImmutableInstrumentationScope(orig *otlpcommon.InstrumentationScope) InstrumentationScope
```

because it'll conflict with

```
func NewInstrumentationScope() InstrumentationScope
```

pdata/plog/generated_scopelogs.go
dmitryax: By `immutableScopeLogs` I meant an instance of `ScopeLogs`

pdata/plog/generated_scopelogs.go
dmitryax: Not sure I follow why we want to rename this?

dmitryax: Even if we don't have methods on `commonScopeLogs` in this particular struct, it's not the case for other structs. Maybe better to keep it consistent and not introduce any additional logic to the generator looking at the set of struct fields and see if the `common` struct can be removed?

jpkrohling:
Did I understand it right that @dmitryax will break this down in smaller PRs? I'm in favor of this change, but will wait to review the smaller pieces.

dmitryax:
> Did I understand it right that @dmitryax will break this down in smaller PRs? I'm in favor of this change, but will wait to review the smaller pieces.

Yes, this PR will keep the end state, and I'll break it down into smaller PRs with deprecations.

djaglowski:
I really like the design.

You've got me thinking about whether there is a way to have compile-time guarantees at the factory API level, rather than the consumer implementation level. This is not at all meant to block this PR, especially because I have not followed previous discussions on this effort. My thoughts on this are focused on consumers, so I may be missing or misunderstanding the needs for other use cases.

That said, I'm am wondering if the mutable / immutable structs defined in this PR can solve our data consistency problem without needing to expose the "get mutable" methods.

Basically, the idea would be to have the service call "get mutable" methods when necessary, based on `MutatesData`, and to do so in a way that forces components to set the flag correctly. Something like:

Keep the mutable / immutable structs from this PR, but minimize access to the "get mutable" methods.

In `service` we define a new type of skinny wrapper consumer (like `fanoutconsumer`). We wrap _every_ consumer with this. The wrapper just calls `ToImmutable` or `getMutable` as necessary according to the consumer's `MutatesData` before passing along.

Effectively, a consumer will always be given immutable data unless it has correctly set `MutatesData: true`.
- When a mutable consumer precedes an immutable one, the service correctly calls `ToImmutable` prior to handing it off. 
- When a a mutable consumer follows an immutable one, "get mutable" was going to be called anyways, but now is handled within the service rather than every consumer implementation. 
- It is still possible for a consumer to _unnecessarily_ declare `MutatesData: true`, but this is basically a performance optimization problem rather than a data integrity / collector stability problem.

Again, I am likely missing some nuance here, but I'm sharing because it might be a way to solve the correctness problem while exposing less additional API.

dmitryax:
@djaglowski that was my initial thought as well, but there are a couple of problems with that:
1. We would need to update all the consumer interfaces to have another method like `ConsumeMutableTraces`.
2. With the approach in this PR, we give an opportunity to the component authors to delay access to the mutable pdata and avoid unnecessary copying. For example, filter or transform processors don't need to mutate pdata if it doesn't match the conditions.

djaglowski:
@dmitryax, makes sense. Thanks for explaining.

pdata/internal/generated_instrumentationscope.go
```diff
@@ -0,0 +1,134 @@
...
+
+func NewMutableInstrumentationScope(orig *otlpcommon.InstrumentationScope) MutableInstrumentationScope {
+	return MutableInstrumentationScope{commonInstrumentationScope: commonInstrumentationScope{orig}}
+}
+
+// NewInstrumentationScope creates a new empty InstrumentationScope.
+//
+// This must be used only in testing code. Users should use "AppendEmpty" when part of a Slice,
+// OR directly access the member if this is embedded in another struct.
+func NewInstrumentationScope() MutableInstrumentationScope {
```
bogdandrutu, line 56: We should stick with our guideline/rule `NewMutableInstrumentationScope` since we return `MutableInstrumentationScope`

pdata/ptrace/traces.go
```diff
@@ -24,16 +24,23 @@ import (
...
 
 // NewTraces creates a new Traces struct.
 func NewTraces() Traces {
-	return newTraces(&otlpcollectortrace.ExportTraceServiceRequest{})
+	orig := &otlpcollectortrace.ExportTraceServiceRequest{}
+	state := internal.StateExclusive
+	return Traces(internal.NewTraces(orig, &state))
+}
+
+func (ms Traces) MarkShared() {
```
bogdandrutu, line 42: MarkImmutable? or AsImmutable

pdata/internal/wrapper_traces.go
```diff
@@ -20,15 +20,24 @@ import (
 )
 
 type Traces struct {
-	orig *otlpcollectortrace.ExportTraceServiceRequest
+	orig  *otlpcollectortrace.ExportTraceServiceRequest
+	state *State
```
bogdandrutu, line 24: state State

pdata/ptrace/traces.go
```diff
@@ -46,7 +53,7 @@ func (ms Traces) MoveTo(dest Traces) {
 
 // CopyTo copies the Traces instance overriding the destination.
 func (ms Traces) CopyTo(dest Traces) {
-	ms.ResourceSpans().CopyTo(dest.ResourceSpans())
+	ms.ResourceSpans().CopyTo(dest.MutableResourceSpans())
 }
```
bogdandrutu, line 57: Clone.

pdata/plog/generated_scopelogsslice.go
```diff
@@ -31,24 +31,41 @@ import (
...
 // NewScopeLogsSlice creates a ScopeLogsSlice with 0 elements.
 // Can use "EnsureCapacity" to initialize with a given capacity.
-func NewScopeLogsSlice() ScopeLogsSlice {
+func NewScopeLogsSlice() MutableScopeLogsSlice {
 	orig := []*otlplogs.ScopeLogs(nil)
-	return newScopeLogsSlice(&orig)
+	return newMutableScopeLogsSlice(&orig)
+}
+
+func (es MutableScopeLogsSlice) ToImmutable() ScopeLogsSlice {
```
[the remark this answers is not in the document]
dmitryax: Makes sense. updated

pdata/internal/generated_instrumentationscope.go
dmitryax: We probably should, yes. Updated

pdata/ptrace/traces.go
dmitryax: I think using `Immutable` word can be confusing here. I changed it to `AsShared` and added a comment. Let me know WDYT

pdata/internal/wrapper_traces.go
dmitryax: Resolved. I applied the changes that you suggested on the call

pdata/ptrace/traces.go
dmitryax: After wrapping the `internal.Traces` into another struct, this func should work as expected. Do you think we still need to change it to `Clone`?

pdata/plog/generated_scopelogs.go
dmitryax: Discussed this offline, resolved

pdata/plog/generated_scopelogs.go
dmitryax: Discussed offline, resolved

dmitryax:
This draft PR now represents the end state that I envision with passing the internal CI

I'll prepare a plan how this can be gradually applied

cc @bogdandrutu @djaglowski @jpkrohling

PR #7148, opened 2023-02-07
Fix API to be compatible with flagset, missed the comment initially

bogdandrutu:
Not breaking change, not released yet.

For details see https://pkg.go.dev/flag#VisitAll and https://pkg.go.dev/flag#Visit


dmitryax (approved):
Looks like contrib already adopted the new interface, please update it after

Issue #7147, opened 2023-02-07
[connectors] Enable routing to specific pipelines


djaglowski:
**Is your feature request related to a problem? Please describe.**

Although the connectors framework is nearly ready for use, it does yet not provide connectors the ability to emit data to specific pipelines. Currently, connectors are given a single consumer and they must emit to it. This issue proposes a small enhancement to the framework which would enable this functionality. Several potential connectors could use it:
- A `routing` connector, ported from the [`routing` processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/routingprocessor/README.md). As a processor, it is currently necessary to issue several [caveats](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e9e4eee56e1ecf3e076e9b4e9d7bb97bd175c202/processor/routingprocessor/README.md?plain=1#L13-L15) about correct usage in a pipeline. A connector implementation could provide the same functionality in more natural way.
- A `failsafe` connector, which reacts to backpressure by emitting to backup pipeline, and potentially emits corresponding events.
- A `loadbalancing` connector, which distributes data across multiple pipelines. This also could be sensitive to backpressure.
- A `statemachine` connector, where each state may emit to a different set of pipelines.

**Describe the solution you'd like**

In order to expose a simple, safe, and efficient API to connector implementations, I believe we should define a special type of consumer whose `Consume` function optionally accepts a list of pipeline IDs. e.g. `ConsumeLogs(context.Context, plog.Logs, ...component.ID) error`. For clarity, I'll refer to this as `connector.LogsRouter` and assume that each signal would use the same pattern.

This would be passed into `connector.Factory.CreateLogs...` in place of `consumer.Logs`. We could consider passing it _as_ `consumer.Logs`, with the understanding that the type may be asserted as necessary. This would present the familiar API to the majority of connector implementations that do not require routing functionality.

Tentatively, the `connector.LogsRouter` would be built with a `map[component.ID]consumer.Logs`, where a key is a pipeline ID and the corresponding value is the first consumer in that pipeline (i.e. `capabilitiesNode.getConsumer()`). When `connector.LogsRouter.ConsumeLogs()` is called, it will either retrieve or build a handle to the appropriate consumer, which may be a fanoutconsumer. We can consider some options here, either pre-building all possible fanouts, caching fanouts that were previously built, or just building a temporary fanout for each call. The exact implementation would be an internal concern for the framework, so could be reworked as necessary.

[...]

PR #7152, opened 2023-02-07
Allow connectors to route data to specific consumers

djaglowski:
This is a proposed solution for #7147. 
- Adds `connector.<Type>Router` interface to extend `consumer.<Type>` with a `Route<Type>` method.
- Adds implementations of `connector.<Type>Router` in `service/internal/fanoutconsumer` 
  - `Route<Type>` methods build a fanout consumer every call. I think this likely needs to be more efficient, but can be addressed later if the overall design looks good.
- Getting the first consumer from a pipeline is a lot easier if the `capabilitiesNode` is a proper consumer, so I restored this, hopefully correctly this time.
- Adds an example router in `service/internal/testcomponents`. To simulate routing, it keeps a counter and uses parity to route one way or the other.
- Adds a test to `service/graph_test.go` that validates expected behavior of the example router.

### DIFF
--- a/connector/connector.go
+++ b/connector/connector.go
@@ -22,6 +22,9 @@ import (
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
 // A Traces connector acts as an exporter from a traces pipeline and a receiver
@@ -41,6 +44,12 @@ type Traces interface {
 	consumer.Traces
 }
 
+// TracesRouter feeds the first consumer.Traces in each of the specified pipelines.
+type TracesRouter interface {
+	consumer.Traces
+	RouteTraces(context.Context, ptrace.Traces, ...component.ID) error
+}
+
 // A Metrics connector acts as an exporter from a metrics pipeline and a receiver
 // to one or more traces, metrics, or logs pipelines.
 // Metrics feeds a consumer.Traces, consumer.Metrics, or consumer.Logs with data.
@@ -57,6 +66,12 @@ type Metrics interface {
 	consumer.Metrics
 }
 
+// MetricsRouter feeds the first consumer.Metrics in each of the specified pipelines.
+type MetricsRouter interface {
+	consumer.Metrics
+	RouteMetrics(context.Context, pmetric.Metrics, ...component.ID) error
+}
+
 // A Logs connector acts as an exporter from a logs pipeline and a receiver
 // to one or more traces, metrics, or logs pipelines.
 // Logs feeds a consumer.Traces, consumer.Metrics, or consumer.Logs with data.
@@ -70,6 +85,12 @@ type Metrics interface {
 type Logs interface {
 	component.Component
 	consumer.Logs
+}
+
+// LogsRouter feeds the first consumer.Logs in each of the specified pipelines.
+type LogsRouter interface {
+	consumer.Logs
+	RouteLogs(context.Context, plog.Logs, ...component.ID) error
 }
 
 // CreateSettings configures Connector creators.
--- a/.chloggen/connectors-routing.yaml
+++ b/.chloggen/connectors-routing.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: connectors
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Provide connectors with a mechanism to route data to specific pipelines
+
+# One or more tracking issues or pull requests related to the change
+issues: [7152]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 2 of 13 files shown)


connector/connector.go
```diff
@@ -41,6 +44,12 @@ type Traces interface {
 	consumer.Traces
 }
 
+// TracesRouter feeds the first consumer.Traces in each of the specified pipelines.
+type TracesRouter interface {
+	consumer.Traces
+	RouteTraces(context.Context, ptrace.Traces, ...component.ID) error
```
bogdandrutu, line 50: Feel like a better way is to give access to `map[PipelineID]consumer.Traces` (first consumer in each pipeline):
* Gives access to all pipelines: component knows how many and what are the IDs available, in this case the component has to be configured with the IDs. This is important for a load-balancer component where you want to do Round-Robin for example. 
* Avoid a slice creation because of the vararg (the compiler may be smart in some cases to stack allocate, but not sure).

connector/connector.go
```diff
@@ -41,6 +44,12 @@ type Traces interface {
 	consumer.Traces
 }
 
+// TracesRouter feeds the first consumer.Traces in each of the specified pipelines.
+type TracesRouter interface {
```
bogdandrutu, line 48: There are few places where we do fan-out:
* Receivers
* Connector as receiver
* Last processor

We need to clarify in which case we expect this interface to work (what is the expected goal).

connector/connector.go
djaglowski: That's a good point about needing access to all pipelines, even when not sending a given request to all.

Do you agree that `map[PipelineID].consumer.Traces` fits better into https://github.com/open-telemetry/opentelemetry-collector/pull/7179, where this signature would be `Consumers() map[PipelineID]consumer.Traces`, rather than as a request to handle routing directly?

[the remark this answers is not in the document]
djaglowski: Makes sense. Pipeline IDs are component IDs, but with a restriction on DataType. Similar to the relationship between `component.Type` and `component.DataType`.

connector/connector.go
djaglowski: My goal was only with connectors, so I defined the interface in the connectors package. We could allow it in the other cases as well, but I think any use case that would require it could be implemented as a connector. In my opinion, this should remain a connectors-only interface unless there is a clear need.

connector/connector.go
bogdandrutu: Yes... Sorry will keep a bit the conversation here, or you can close this and copy the threads and continue there.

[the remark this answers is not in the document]
bogdandrutu: If that is the case what about the Connector factory to accept a "TraceRouter" (probably name can be better):

```golang

type Factory interface {
...

CreateTracesToLogs(ctx context.Context, set CreateSettings, cfg component.Config, next TraceRouter) (Traces, error)

...
}
```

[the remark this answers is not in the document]
bogdandrutu: Not sure, if this is a better idea, but want to make sure we thought about this.

connector/connector.go
bogdandrutu: "A last processor" use-case may also be ok. It simplifies a bit the configuration, for use-cases like the routing processor which can remain a processor.

If we don't allow that, what about we disallow this by doing some wrapping in case of processors/receivers to not be able to convert the next to this interface.

connector/connector.go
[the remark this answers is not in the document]
bogdandrutu: Actually, we can even limit this to "CreateTracesToTraces" to allow fanout routing.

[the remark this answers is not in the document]
djaglowski: > If that is the case what about the Connector factory to accept a "TraceRouter" (probably name can be better):
> 
> ```go
> type Factory interface {
> ...
> 
> CreateTracesToLogs(ctx context.Context, set CreateSettings, cfg component.Config, next TraceRouter) (Traces, error)
> 
> ...
> }
> ```

This is effectively what I tried to accomplish. I allows the signature to remain with `consumer.Traces`, because most connectors will only care about one consumer. But it would be understood to actually be a `connector.TraceRouter` so some connectors can assert the more specific type if they need.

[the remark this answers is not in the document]
djaglowski: > Actually, we can even limit this to "CreateTracesToTraces" to allow fanout routing.

I can think of cross-signal use cases for routing (generating signal but also sending to specific consumers based on configuration). That said, we could limit for now since obvious use cases seem to be same signal.

connector/connector.go
djaglowski: > "A last processor" use-case may also be ok. It simplifies a bit the configuration, for use-cases like the routing processor which can remain a processor. Not proposing, want to hear ideas. I also believe that it is clear that all "routing" happens in the connectors.

I think there is a better case to do routing as a connector instead of a processor. For example, with logs it is sometimes necessary to read from one file but then separate based on the structure. Then process each type differently based on that. A connector can support this easily but a processor would have to send to an exporter that _is_ a connector, so I think it's easier to have only connectors behave this way.

Similarly with metrics, maybe you want to process metrics differently for some reason - name, type, attributes - so you send to different pipelines for differentiated processing. A processor can only send to different exporters, so you would have to sent to exporters that are connectors anyways.

> If we don't allow that, what about we disallow this by doing some wrapping in case of processors/receivers to not be able to convert the next to this interface.

I agree. This is what I have intended with this implementation. The service only passes the `connector.TracesRouter` type to create connector methods, so there is no worry about this for other components. Even if we need to change the signature for creating connectors, we would still make sure in the service to only pass in the routing types to connectors.

connector/connector.go
djaglowski: No problem. Will keep discussion here for now at least.

djaglowski: @bogdandrutu, I believe this one is what you suggested: https://github.com/open-telemetry/opentelemetry-collector/pull/7226

connector/connector.go
djaglowski: One more reason to only allow routing from connectors is that accessing a fanout is only possible for the last processor in a pipeline. Users would have to take this into account in order to determine valid placement of the processor. Connectors are always in the correct position to handle this, so the only consideration for users is which pipelines they are emitting to, which is a standard concern for connectors.

djaglowski:
Closing in favor of #7179

PR #7158, opened 2023-02-08
Fix typo in docs/performance.md

imomaliev:
**Description:**
fixed RAM columns for results with tail-based sampling. Looks like were entered wrong values for Zipkin and Jaeger benchmarks

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -58,10 +58,10 @@ service
 | :---:             | :---:           | :---:       | :---:             | :---:                  |
 | OpenTelemetry        | 1               | 2           | ~9K               | 8K                     |
 | OpenTelemetry        | 2               | 4           | ~18K              | 16K                    |
-| Jaeger Thrift     | 1               | 6           | ~11.5K            | 10K                    |
-| Jaeger Thrift     | 2               | 8           | ~23K              | 20K                    |
-| Zipkin v2 JSON    | 1               | 6           | ~8.5K             | 7K                     |
-| Zipkin v2 JSON    | 2               | 8           | ~16K              | 14K                    |
+| Jaeger Thrift     | 1               | 2           | ~11.5K            | 10K                    |
+| Jaeger Thrift     | 2               | 4           | ~23K              | 20K                    |
+| Zipkin v2 JSON    | 1               | 2           | ~8.5K             | 7K                     |
+| Zipkin v2 JSON    | 2               | 4           | ~16K              | 14K                    |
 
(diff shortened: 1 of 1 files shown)


docs/performance.md
```diff
@@ -58,10 +58,10 @@ service
 | :---:             | :---:           | :---:       | :---:             | :---:                  |
 | OpenTelemetry        | 1               | 2           | ~9K               | 8K                     |
 | OpenTelemetry        | 2               | 4           | ~18K              | 16K                    |
-| Jaeger Thrift     | 1               | 6           | ~11.5K            | 10K                    |
-| Jaeger Thrift     | 2               | 8           | ~23K              | 20K                    |
-| Zipkin v2 JSON    | 1               | 6           | ~8.5K             | 7K                     |
-| Zipkin v2 JSON    | 2               | 8           | ~16K              | 14K                    |
+| Jaeger Thrift     | 1               | 2           | ~11.5K            | 10K                    |
```
jpkrohling, line 61: Looking at the history of this line, I would guess the OpenTelemetry one is the one that should be changed: https://github.com/open-telemetry/opentelemetry-collector/commit/1ce867d2716d40b24e614ea87be183c087a9ca30

I wouldn't change this without you having executed new performance tests or confirming with the original author if the numbers are wrong.

jpkrohling (approved):

jpkrohling (changes requested):
Sorry, I didn't mean to approve. Quite the opposite. See my previous message.

docs/performance.md
imomaliev: @jpkrohling You are right. I apologize, it this was a little hasty of me not to check previous history and assume there was a typo. 

I will draft this PR for now and check with author and previous contributors. If currently everything is correct, I will update the docs with an explanation.

imomaliev:
@flands Hi! I am trying to understand why there is a mismatch in amount of RAM for tail-sampling benchmarks between OpenTelemetry (2 and 4) and Zipkin with Jaeger (6 and 8). Could you please provide more context why there is a difference? I think it would be better to have similar amount of RAM for all protocols, as well as having matching matrixes for head and tail based sampling.

I see that you have left a [note](https://github.com/open-telemetry/opentelemetry-collector/commit/1ce867d2716d40b24e614ea87be183c087a9ca30#diff-118f0ef20c6b4cf62284949363699de7c7412fddb16c693a22f07610ee14fec8R52)
```
> Note: Additional memory is required for tail-based sampling
```
I assume this is why initial benchmarks for tail-sampling had higher RAM amount

Issue #7048, opened 2023-01-27
Get prefix is not idiomatic in Golang  


bogdandrutu:
See https://github.com/open-telemetry/opentelemetry-collector/blob/main/consumer/consumererror/signalerrors.go#L39

PR #7172, opened 2023-02-09
[consumererror] Remove Get prefix from consumer package functions

evan-bradley:
**Description:**

Remove all `Get` prefixes from functions in the consumer package so that they are in line with our [naming conventions](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#naming-convention).

**Link to tracking Issue:**
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/7048


bogdandrutu (commented):
Since we are here, let's start using generics to avoid duplicate core:

```golang
type retryable[V ptrace.Traces|pmetric.Metrics|plog.Logs] struct {
	error
	failed V
}

// Unwrap returns the wrapped error for functions Is and As in standard package errors.
func (err Traces) Unwrap() error {
	return err.error
}


// Traces is an error that may carry associated Trace data for a subset of received data
// that failed to be processed or sent.
type Traces retryable[ptrace.Traces]

// NewTraces creates a Traces that can encapsulate received data that failed to be processed or sent.
func NewTraces(err error, failed ptrace.Traces) error {
	return Traces{
		error:  err,
		failed: failed,
	}
}

// GetTraces returns failed traces from the associated error.
func (err Traces) GetTraces() ptrace.Traces {
	return err.failed
}

```

consumer/consumererror/signalerrors.go
```diff
@@ -35,8 +35,8 @@ func NewTraces(err error, failed ptrace.Traces) error {
 	}
 }
 
-// GetTraces returns failed traces from the associated error.
-func (err Traces) GetTraces() ptrace.Traces {
+// Traces returns failed traces from the associated error.
+func (err Traces) Traces() ptrace.Traces {
```
bogdandrutu, line 39: Need to follow the depreciation policy in the contribution readme

evan-bradley: Whoops, not sure what happened there. Fixed.

evan-bradley:
>Since we are here, let's start using generics to avoid duplicate core

I'm good with that. I pushed a commit that approaches it using struct embedding instead of type aliasing, which I think should help reduce duplication if we add more common methods to the types. I'm also fine with just aliasing each signal type to the retryable type, so let me know if you think that will be the better path.

consumer/consumererror/signalerrors.go
```diff
@@ -20,77 +20,90 @@ import (
...
 func (err Logs) GetLogs() plog.Logs {
-	return err.failed
+	return err.Logs()
 }
 
-// Unwrap returns the wrapped error for functions Is and As in standard package errors.
-func (err Logs) Unwrap() error {
-	return err.error
+// Logs returns failed logs from the associated error.
+func (err Logs) Logs() plog.Logs {
```
bogdandrutu, line 81: `func (err Logs) Logs() plog.Logs`

OR

`func (err Logs) Data() plog.Logs` (or other generic name)

Advantage is that all types will have "Data" as the accessor?

consumer/consumererror/signalerrors.go
```diff
@@ -20,77 +20,90 @@ import (
...
+	return err.failed
 }
 
 // Metrics is an error that may carry associated Metrics data for a subset of received data
 // that failed to be processed or sent.
 type Metrics struct {
-	error
-	failed pmetric.Metrics
+	retryable[pmetric.Metrics]
 }
```
bogdandrutu, line 89: I think you can do 

```golang
type Metrics retryable[pmetric.Metrics]
```

bogdandrutu: Hmm.. not sure if that does not expose too much about the internal implementation. Let's keep it as is.

consumer/consumererror/signalerrors.go
evan-bradley: I'm having trouble finding a strong advantage to either choice, but I do see the value in having a consistent method to get the telemetry data, and it does reduce some duplication. I added a commit to test out `Data` to see if we like it.

bogdandrutu (approved):

PR #7179, opened 2023-02-13
Allow connectors to request consumers to specific routes

### LINKED ISSUE #7147: [connectors] Enable routing to specific pipelines

djaglowski:
Resolves #7147 

Alternative to #7152

This allows individual connector implementations more responsibility in managing routes. Specifically, connectors can request a consumer to a specific set of pipelines.

There are two apparent advantages to this approach:
1. We don't necessarily need to recreate the fanout consumer every time data is passed to a route. Rather, the connector has the opportunity to retain the handle and reuse it. In most cases, I would expect the connector to initialize all consumers (one per route) in the create func.
2. Error handling is much more natural. If a connector requests a consumer for an invalid pipeline, the error is returned and can be handled by the connector at create / start time. In the other implementation, the errors would be returned only when attempting to send data.


djaglowski:
[Here is the diff](https://github.com/djaglowski/opentelemetry-collector/compare/connectors-routing...djaglowski:opentelemetry-collector:connectors-routing-v2) vs #7152

connector/connector.go
```diff
@@ -41,6 +41,13 @@ type Traces interface {
 	consumer.Traces
 }
 
+// TracesRouter feeds the first consumer.Traces in each of the specified pipelines.
+// The router will create a fanout consumer for the set of pipelines and return a uuid
+type TracesRouter interface {
```
djaglowski, line 46: @bogdandrutu, one note about this implementation vs exporting `fanoutconsumer`. Here we add a new interface, but it is not a interface that connectors need to implement. The oimplementation is within the internal fanoutconsumer package. The interface just makes it simple for connectors to get consumers without having to worry about proper implementation.

connector/connector.go
```diff
@@ -41,6 +41,13 @@ type Traces interface {
...
+	Consumer(...component.ID) (consumer.Traces, error)
+	Consumers() []component.ID
```
bogdandrutu, line 48: Consumers don't have an ID. Maybe "Pipelines"? or "PipelineIDs"?

service/internal/fanoutconsumer/logs.go
```diff
@@ -80,3 +83,50 @@ func (lsc *logsConsumer) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
...
 }
+
+var _ connector.LogsRouter = (*logsRouter)(nil)
+
+type logsRouter struct {
+	consumer.Logs
+	consumers map[component.ID]consumer.Logs
+}
+
+func NewLogsRouter(cm map[component.ID]consumer.Logs) consumer.Logs {
```
bogdandrutu, line 94: Confused why you need a new class here vs changing NewLogs to return this?

djaglowski: This means that receiver constructors can assert `connector.LogsRouter` as well. Maybe this is ok but I wasn't expecting to enable this.

djaglowski: It also turns out to be quite an overhaul of `NewLogs` and its tests, because we need to pass in `map[component.ID]consumer.Logs`, instead of just `[]consumer.Logs`. Doable, but seems to generate enough unnecessary code that it's worse than having a second class in my opinion.

service/internal/fanoutconsumer/logs.go
```diff
@@ -80,3 +83,50 @@ func (lsc *logsConsumer) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
...
+	consumers := make([]consumer.Logs, 0, len(pipelineIDs))
+	var errors error
+	for _, pipelineID := range pipelineIDs {
+		c, ok := r.consumers[pipelineID]
+		if ok {
+			consumers = append(consumers, c)
+		} else {
+			errors = multierr.Append(errors, fmt.Errorf("missing consumer: %q", pipelineID))
+		}
+	}
```
bogdandrutu, line 126: I would go with simpler solution and just return at first error. Since this is not a "user input" showing just the first error is enough.

service/internal/fanoutconsumer/logs.go
djaglowski: Are you ok with leaving these as separate structs?

service/internal/fanoutconsumer/logs.go
djaglowski: Is it ok if we leave it so that users can see all errors?

bogdandrutu: I think "it is ok". But the question I have, is it worth it? Does it provide any benefit? It is hard to imagine how you will get to this, since the components that use this API have access to all "piepeline IDs", so this to me seem like a bug in that component not a user error.

bogdandrutu (approved):
Last question before merging since all other comments are about internal things that we can change.

Where should this optional interface be defined?

service/internal/fanoutconsumer/logs.go
djaglowski: Here is an example of when a user would see this error:

```yaml
receivers:
  foo:
exporters:
  bar/left:
  bar/right:
connectors:
  routing:
    table:
      - statement: ... 
        pipelines: [logs/left]
      - statement: ...
        pipelines: [logs/right, logs/fake, logs/oops]
service:
  pipelines:
    logs/in:
      receivers: [foo]
      exporters: [routing]
    logs/left:
      receivers: [routing]
      exporters: [bar/left]
    logs/right:
      receivers: [routing]
      exporters: [bar/right]
```

Routing connector constructor will ask for consumers and this will detect invalid user config when validating the second route. The error informs the user which of the specified pipelines are invalid.
```go
func createLogs(_ context.Context, _ connector.CreateSettings, cfg component.Config, logs consumer.Logs) (connector.Logs, error) {
	c := cfg.(RoutingConnectorConfig)
	logsRouter := logs.(connector.LogsRouter)
	
	// generate consumer for each route only once
	routeConsumers := make([]consumer.Logs, 0, len(c.Table))
	for i, route := range c.Table {
		cons, err := logsRouter.Consumer(route.Pipelines...)
		if err != nil {
			return nil, err // missing consumer: "logs/fake", missing consumer: "logs/oops"
		}
		routeConsumers[i] = cons 
	}
	...
	return &logsRoutingConnector{
		routeConsumers: routeConsumers,
		...
	}
}

```

djaglowski:
> Last question before merging since all other comments are about internal things that we can change.
> 
> Where should this optional interface be defined?
> 
> * `component` the lowest level package that can declare it?
> * `service/pipeline[s]` since it is a property of our pipelines?
> * keep it in `connector` since in general only connectors will have this logic? Not sure if we don't want ever to say that even processor can access this (last processor in the pipeline).

I think `connector` is a reasonable place for the reasons you cited. It's also a brand new package so _if_ we need to deprecate it, I think it is not that big of a problem yet. `component` on the other hand would be risky because any change would be breaking a mature package. `service/pipeilne[s]` could make sense if it existed, but I would just as soon not create a new package for this.

service/internal/fanoutconsumer/logs.go
bogdandrutu: If you really want that you can check the IDs since you have access to all of them and error across all "tables". Still not convinced it is worth it.

PR #7202, opened 2023-02-15
Move marshal to persistent storage, preparation to use generics

bogdandrutu:


codeboten:
Some tests will need fixing :)

PR #7206, opened 2023-02-16
Improve error message when cycle is detected in component graph

djaglowski:
See https://github.com/open-telemetry/opentelemetry-collector/pull/7045#discussion_r1094843917

### DIFF
--- a/service/graph.go
+++ b/service/graph.go
@@ -162,7 +162,7 @@ func (g *pipelinesGraph) createEdges() {
 func (g *pipelinesGraph) buildComponents(ctx context.Context, set pipelinesSettings) error {
 	nodes, err := topo.Sort(g.componentGraph)
 	if err != nil {
-		return cycleErr(err)
+		return cycleErr(err, topo.DirectedCyclesIn(g.componentGraph))
 	}
 
 	for i := len(nodes) - 1; i >= 0; i-- {
@@ -357,14 +357,18 @@ func (p *pipelineNodes) mutatesData() bool {
 	return p.capabilitiesNode.getConsumer().Capabilities().MutatesData
 }
 
-func cycleErr(err error) error {
+func cycleErr(err error, cycles [][]graph.Node) error {
 	var topoErr topo.Unorderable
-	if !errors.As(err, &topoErr) {
+	if !errors.As(err, &topoErr) || len(cycles) == 0 || len(cycles[0]) == 0 {
 		return err
 	}
 
 	// There may be multiple cycles, but report only the first one.
-	cycle := topoErr[0]
+	cycle := cycles[0]
+
+	// The last node is a duplicate of the first node.
+	// Remove it because we may start from a different node.
+	cycle = cycle[:len(cycle)-1]
 
 	// A cycle always contains a connector. For the sake of consistent
 	// error messages report the cycle starting from a connector.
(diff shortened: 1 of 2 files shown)


service/graph.go
```diff
@@ -356,3 +356,36 @@ func (p *pipelineNodes) exporterIDs() []string {
...
 	return p.capabilitiesNode.getConsumer().Capabilities().MutatesData
 }
+
+func cycleErr(err error) error {
+	var topoErr topo.Unorderable
+	if !errors.As(err, &topoErr) {
+		return err
+	}
+
+	// There may be multiple cycles, but report only the first one.
```
codeboten, line 366: Why only the first one?

codeboten (commented):
Thanks for the fix, just one question

service/graph_test.go
```diff
@@ -1217,7 +1216,7 @@ func TestGraphBuildErrors(t *testing.T) {
 					Exporters:  []component.ID{component.NewIDWithName("nop", "conn")},
 				},
 			},
-			expectedStartsWith: "topo: no topological ordering: cyclic components:",
+			expected: `cycle detected: connector "nop/conn" -> processor "nop" -> connector "nop/conn"`,
```
bogdandrutu, line 1219: Print incoming pipeline type and outgoing for conn?

bogdandrutu (commented):
Do we have

service/graph.go
djaglowski: It's just an attempt to limit the size of the error message.

djaglowski:
This is working as intended now.

service/graph_test.go
djaglowski: I added this info. Also, since processors are associated with a specific pipeline, I added that ID as well.

Let me know if you have suggestions for the format.

service/graph.go
```diff
@@ -356,3 +356,43 @@ func (p *pipelineNodes) exporterIDs() []string {
...
+		switch n := node.(type) {
+		case *processorNode:
+			componentDetails = append(componentDetails, fmt.Sprintf("processor %q in pipeline %q", n.componentID, n.pipelineID))
+		case *connectorNode:
+			componentDetails = append(componentDetails, fmt.Sprintf("connector %q (%s to %s)", n.componentID, n.exprPipelineType, n.rcvrPipelineType))
+		default:
+			continue // skip capabilities/fanout nodes
+		}
+	}
+	return fmt.Errorf("cycle detected: %s", strings.Join(componentDetails, " -> "))
```
mx-psi, line 397: To make tests more robust you can make a (unexported) custom error type that has the component list as a field and handles this formatting on the `Error` method. That way we can check the component list on tests instead of the error itself, and we won't have to update the tests in the future if we only update the error wording.

djaglowski: I can appreciate what you're going for here but I'm not sure it's worth it given the tradeoffs.
- Validating the actual error message ensures we know it looks as expected.
- Updating the format is pretty easy. You can just run the tests, ensure the actual result looks correct, and if so replace the expected result with it.
- The error could use `[]graph.Node`, but then we have to type switch again. It's not difficult but seems like we end up duplicating most of this code in tests anyways, which seems more complex and less robust to me.

mx-psi: Fair enough, those are good counterarguments :) We can leave it as is then and refactor later if the need arises

djaglowski:
This is ready for another review

mx-psi (approved):

djaglowski:
Can this be merged @open-telemetry/collector-maintainers?

codeboten (approved):

PR #7210, opened 2023-02-17
Ensure workflows use Go >=1.19.6

Aneurysm9:


.github/workflows/api-compatibility.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19
+          go-version: ~1.19.6
```
codeboten, line 34: shouldn't this be `>=` instead of `~`? Looking at https://github.com/actions/setup-go#v3

codeboten (approved):

PR #7226, opened 2023-02-21
Allow connectors to access all consumers

djaglowski:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/7147.

Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/7152 and https://github.com/open-telemetry/opentelemetry-collector/pull/7179.

This implementation adds a less limited interface to the connectors package. The expectation is that connector implementations that require routing capabilities will request the full map of consumers and build fanout consumers as appropriate. Correspondingly, the `fanoutconsumer` package is moved from `service/internal/` to `consumer/`.


connector/consumer.go
```diff
@@ -0,0 +1,132 @@
...
+// Unlike other consumers, the connector consumer will expose its underlying consumers.
+type TracesConsumer interface {
+	consumer.Traces
+	Consumers() map[component.ID]consumer.Traces
+}
+
+var _ TracesConsumer = (*tracesConsumer)(nil)
+
+type tracesConsumer struct {
+	consumer.ConsumeTracesFunc
```
bogdandrutu, line 33: Why not a consumer.Traces here, then no need to implement Capabilities.

connector/consumer.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/consumer/fanoutconsumer"
+)
+
+// TracesConsumer feeds the first consumer.Traces in each of the specified pipelines.
+// Unlike other consumers, the connector consumer will expose its underlying consumers.
+type TracesConsumer interface {
```
bogdandrutu, line 25: I think you can define only:

```golang
// MultiTracesConsumer is an optional interface implemented by consumer.Traces which fanout to multiple other consumers. (or FanoutTracesConsumer)
type MultiTracesConsumer interface {
  Consumers() map[component.ID]consumer.Traces
}
```

Since that is what go recommends for this cases.

connector/consumer.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+var _ TracesConsumer = (*tracesConsumer)(nil)
+
+type tracesConsumer struct {
+	consumer.ConsumeTracesFunc
+	consumers map[component.ID]consumer.Traces
+}
+
+func NewTracesConsumer(cm map[component.ID]consumer.Traces) TracesConsumer {
```
bogdandrutu, line 37: I would definitely make this internal. Find a place somewhere in service/internal so no need on exposing the fanout components.

connector/consumer.go
bogdandrutu: Separate question: 
Should this be here? Should we have it as part of "component" maybe?

connector/consumer.go
bogdandrutu: > For example, the current routing processor allows to send to multiple routes per entry in the routing table. This is ok in the current implementation because we route directly to exporters, which we assume do not mutate the data.

That is a wrong assumption.

We can expose fanout, but still don't have to expose the implementation of the `TracesRouter`.

connector/consumer.go
bogdandrutu:
```
// TracesRouter is an optional interface implemented by consumer.Traces which fanout to multiple other consumers
type TracesRouter interface {
  Consumers() map[component.ID]consumer.Traces
}
```

> If we remove the consumer.Traces, then non-routing connectors will always have to handle their own fanout to all consumers. I think it's better if we can pass into the connector constructors the typical consumer.Traces interface, and require only non-routing connectors to be concerned with this optional method.

No because as the text said, that is an optional interface implemented by `consumer.Traces`. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/connector/forwardconnector/forward.go#LL45-L53C2 you still receive `consumer.Traces`.

```golang
// createTracesToTraces creates a trace receiver based on provided config.
func createTracesToTraces(
	_ context.Context,
	set connector.CreateSettings,
	cfg component.Config,
	nextConsumer consumer.Traces,
) (connector.Traces, error) {
	router, ok = nextConsumer.(TracesRouter)
	if ok {
		// It means the next component is a router, so if I want to do manual routing I can access
		// router.Consumers()
	} else {
		next consumer is not a router means component is connected to only 1 pipeline.
	}
}
```

djaglowski: > It means the next component is a router, so if I want to do manual routing I can access

We're trying to build a routing connector. The `TracesRouter` interface is meant to allow that, not indicate anything about the next component.

The comment about the interface being optional is not well stated, but my intention was to indicate that this interface can optionally be asserted, so that only routing connectors would assert this in order to gain the ability to route to specific consumers.

connector/consumer.go
djaglowski: The assumption may be bad, but it is a good example how easy it is to ignore mutability because the current routing processor does not account for it.

Exposing fanout would make it _easier_ to do the right thing in such cases, but we will still have to enforce it better than we did with the current routing processor. I'm fine with this, but still think https://github.com/open-telemetry/opentelemetry-collector/pull/7179 is a better solution because it does not allow connector authors to make for such errors.

connector/consumer.go
bogdandrutu: But if user connected to only one pipeline?

connector/consumer.go
bogdandrutu: Related to https://github.com/open-telemetry/opentelemetry-collector/pull/7179 we agreed that is hard to use for the fact that needs to know IDs in advance.

connector/consumer.go
djaglowski: > But if user connected to only one pipeline?

If we are building a routing connector for example, my idea looks like this:
```go
func createTracesToTraces(
	_ context.Context,
	set connector.CreateSettings,
	cfg component.Config,
	nextConsumer consumer.Traces,
) (connector.Traces, error) {
	router, ok = nextConsumer.(TracesRouter)
	if !ok {
		// impossible because service ensures this is always `TracesRouter`, even though it passes in as `consumer.Traces`
	}
	if len(router.Consumers() < 2 {
		return fmt.Errorf("routing connector requires at least two downstream pipelines")
	// or maybe we decide it's ok to have only one. point is it can be checked either way
	}
}
```

connector/consumer.go
djaglowski: > Related to #7179 we agreed that is hard to use for the fact that needs to know IDs in advance.

Yeah, that's true. I think we can adapt the interface to provide this though.

connector/consumer.go
bogdandrutu:
```
func createTracesToTraces(
	_ context.Context,
	set connector.CreateSettings,
	cfg component.Config,
	nextConsumer consumer.Traces,
) (connector.Traces, error) {
	router, ok = nextConsumer.(TracesRouter)
	if !ok {
		// Why not meaning it is only 1?
	}
	if len(router.Consumers() < 2 {
		return fmt.Errorf("routing connector requires at least two downstream pipelines")
		// or maybe we decide it's ok to have only one. point is it can be checked either way
	}
}
```

djaglowski:
Closing in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/7179

PR #7205, opened 2023-02-16
[WIP] [pdata] Copy-on-write pdata

dmitryax:
Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/7140 with copy-on-write approach

With this approach, we can keep the existing pdata interface and remove Capabilities. 

Thanks to @bogdandrutu for bringing the idea.

WIP: Performance optimization.

main branch:

```
BenchmarkLogsUsage-10                       6967	    168206 ns/op	   26321 B/op	    1645 allocs/op
BenchmarkMetricsUsage-10                     650	   1694838 ns/op	  115760 B/op	    8946 allocs/op
BenchmarkTracesUsage-10                     2197	    476184 ns/op	   50630 B/op	    3409 allocs/op
BenchmarkBatchMetricProcessor-10           10000	    107410 ns/op	     250 B/op	       7 allocs/op
```

current state of this PR:

```
BenchmarkLogsUsage-10                       4628	    256751 ns/op	  125050 B/op	    5026 allocs/op
BenchmarkMetricsUsage-10                     421	   2727828 ns/op	  935812 B/op	   39081 allocs/op
BenchmarkTracesUsage-10                     2089	    574658 ns/op	  120796 B/op	    5957 allocs/op
BenchmarkBatchMetricProcessor-10            8367	    263180 ns/op	   85008 B/op	    3016 allocs/op
```

[7329: no orig/state on slices](https://github.com/open-telemetry/opentelemetry-collector/pull/7329):

```
BenchmarkLogsUsage-10                       4342	    237057 ns/op	  100635 B/op	    4011 allocs/op
BenchmarkMetricsUsage-10                     488	   2402435 ns/op	  722830 B/op	   29295 allocs/op
BenchmarkTracesUsage-10                     2098	    544608 ns/op	   95206 B/op	    4893 allocs/op
BenchmarkBatchMetricProcessor-10            9355	    209421 ns/op	   56333 B/op	    2011 allocs/op
```

[7315: no orig/state on any child structs](https://github.com/open-telemetry/opentelemetry-collector/pull/7315):

```
BenchmarkLogsUsage-10                       5530	    214248 ns/op	   66977 B/op	    2954 allocs/op
BenchmarkMetricsUsage-10                     526	   2231044 ns/op	  388111 B/op	   18242 allocs/op
BenchmarkTracesUsage-10                     2296	    520569 ns/op	   56787 B/op	    3640 allocs/op
[...]

PR #7241, opened 2023-02-26
[chore] [pdata] Migrate the generator to go templates

dmitryax:
This change migrates the env var based templates to go templates to allow more complicated logic in them needed for https://github.com/open-telemetry/opentelemetry-collector/pull/7205


pdata/internal/cmd/pdatagen/internal/base_fields.go
```diff
@@ -16,272 +16,284 @@ package internal // import "go.opentelemetry.io/collector/pdata/internal/cmd/pda
...
+	{{- if .isCommon }}
+	return {{ .packageName }}{{ .returnType }}(internal.New{{ .returnType }}(&ms.{{ .origAccessor }}.{{ .fieldName }}))
+	{{- else }}
+	return new{{ .returnType }}(&ms.{{ .origAccessor }}.{{ .fieldName }})
+	{{- end }}
 }`
 
-const accessorSliceTemplateCommon = `// ${fieldName} returns the ${fieldName} associated with this ${structName}.
-func (ms ${structName}) ${fieldName}() ${packageName}${returnType} {
-	return ${packageName}${returnType}(internal.New${returnType}(&ms.${origAccessor}.${fieldName}))
```
djaglowski, line 30: If I understand correctly, we're now autodetecting whether or not each struct is common based on the package, and this is necessarily part of this change in order to build the map expected by go templates. Is that right?

djaglowski (approved):

bogdandrutu (approved):

pdata/internal/cmd/pdatagen/internal/base_fields.go
dmitryax: > now autodetecting whether or not each struct is common based on the package

We've been doing that for a while already. But it's still can be changed if even needed

PR #7201, opened 2023-02-15
Set service.graph feature gate to stable

djaglowski:
Follows #7119

PR #7268, opened 2023-02-28
Remove code path to unused service implementation

djaglowski:
Follows #7201 

The `service.graph` feature gate is now stable, so it cannot be changed. This PR removes the code path to the previous implementation of the service pipelines since it is now unused. It does not contain any user facing changes. There will be further code to remove but this PR is limited to these initial steps:
- Remove usage of service.graph feature gate (without removing the feature gate itself)
- Remove service.pipelines interface

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -37,7 +37,7 @@ import (
 	"go.opentelemetry.io/collector/service/telemetry"
 )
 
-var graphFeatureGate = featuregate.GlobalRegistry().MustRegister(
+var graphFeatureGate = featuregate.GlobalRegistry().MustRegister( // nolint:unused
 	"service.graph",
 	featuregate.StageStable,
 	featuregate.WithRegisterRemovalVersion("v0.74.0"),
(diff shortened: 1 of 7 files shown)


service/service.go
```diff
@@ -37,7 +37,7 @@ import (
 	"go.opentelemetry.io/collector/service/telemetry"
 )
 
-var graphFeatureGate = featuregate.GlobalRegistry().MustRegister(
+var graphFeatureGate = featuregate.GlobalRegistry().MustRegister( // nolint:unused
```
codeboten, line 40: Will this no longer be used only after version 0.74.0?

bogdandrutu: var _ = ?

djaglowski: My understanding is that it is effectively not used as of v0.72.0 (only used in the sense that it is registered so users can see an error if they try to disable it), and that it will be removed entirely in v0.74.0.

bogdandrutu: my suggestion is to allow you to remove the nolint.

djaglowski: Thanks @bogdandrutu. My page hadn't reloaded before I commented so I missed your suggestion and meant to address @codeboten's question. I've removed the nolint as suggested.

bogdandrutu (approved):

codeboten (approved):

PR #7269, opened 2023-02-28
Remove zpagesPipeline interface

djaglowski:
Follows #7201

The service.graph feature gate is now stable, so it cannot be changed. This PR removes the `zpagesPipeline` interface that supported the dual implementations of `service.pipelines`. It does not contain any user facing changes.

### DIFF
--- a/service/graph.go
+++ b/service/graph.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


djaglowski:
> I feel a bigger cleanup can be done between graph/pipelines/zpages/nodes files.

I agree. this is just an incremental step.

PR #7275, opened 2023-03-02
Consolidate zpages constants

djaglowski:
This is a minor cleanup of the zpages constants in the service package.
- Move zpages constants from pipelines.go into zpages.go
- Use common naming format for all zpages constants

### DIFF
--- a/service/zpages.go
+++ b/service/zpages.go
@@ -26,17 +26,23 @@ import (
 )
 
 const (
-	servicezPath   = "servicez"
-	pipelinezPath  = "pipelinez"
-	extensionzPath = "extensionz"
-	featurezPath   = "featurez"
+	// Paths
+	zServicePath   = "servicez"
+	zPipelinePath  = "pipelinez"
+	zExtensionPath = "extensionz"
+	zFeaturePath   = "featurez"
+
+	// URL Params
+	zPipelineName  = "pipelinenamez"
+	zComponentName = "componentnamez"
+	zComponentKind = "componentkindz"
 )
 
 func (host *serviceHost) RegisterZPages(mux *http.ServeMux, pathPrefix string) {
-	mux.HandleFunc(path.Join(pathPrefix, servicezPath), host.zPagesRequest)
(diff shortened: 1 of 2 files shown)


service/zpages.go
```diff
@@ -26,17 +26,20 @@ import (
...
-	pipelinezPath  = "pipelinez"
-	extensionzPath = "extensionz"
-	featurezPath   = "featurez"
+	zPipelineName  = "zpipelinename"
+	zComponentName = "zcomponentname"
+	zComponentKind = "zcomponentkind"
+	zService       = "servicez"
+	zPipeline      = "pipelinez"
+	zExtension     = "extensionz"
+	zFeature       = "featurez"
```
djaglowski, line 35: I don't know enough about zpages - Can we rename the values to use the same format? Or will this be a breaking change for users?
```suggestion
	zPipelineName  = "zpipelinename"
	zComponentName = "zcomponentname"
	zComponentKind = "zcomponentkind"
	zService       = "zservice"
	zPipeline      = "zpipeline"
	zExtension     = "zextension"
	zFeature       = "zfeature"
```

bogdandrutu: the `zService` (a.k.a the one that finish in z) are paths (so will be breaking change). The other are used as URL params so you can change. So if you want to consolidate, you do it the other way around.

bogdandrutu (approved):

service/zpages.go
```diff
@@ -26,17 +26,20 @@ import (
...
-	pipelinezPath  = "pipelinez"
-	extensionzPath = "extensionz"
-	featurezPath   = "featurez"
+	zPipelineName  = "pipelinenamez"
+	zComponentName = "componentnamez"
+	zComponentKind = "componentkindz"
+	zService       = "servicez"
+	zPipeline      = "pipelinez"
+	zExtension     = "extensionz"
+	zFeature       = "featurez"
```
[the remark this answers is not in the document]
djaglowski: Updated. Should be good to go.

bogdandrutu (approved):

PR #7295, opened 2023-03-02
Prototype moving graph & nodes into internal package

djaglowski:
This is a prototype refactoring of the service package, which would move the graph and nodes into an internal package. I'm looking for high level feedback on whether this makes sense.


djaglowski:
Closing in favor of #7365

Issue #4544, opened 2021-12-13
Change batch processor to take client info into consideration


jpkrohling:
Currently, the client info is propagated down from receivers to other components as part of the incoming context. Some processors, like the batch, discard the context, building a new one. This issue is to add the possibility to group data points by client characteristics at the batch processor.

graemechristie:
Is there any progress on this ? Currently we cannot use the batch processor with the routing or header setter processors ... this seems a pretty essential feature for any scenarios involving headers or attributes ?

jpkrohling:
There hasn't been any progress on this yet.

jdgeisler:
We are also wondering if there are any plans for this on the roadmap? 

For Tempo, we are looking to dynamically set the tenant from a header or attribute like we do with Loki using the attribute hints, but without the batch processor the current solution with the headers_setter extension is not working well at scale.

jmacd:
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/4646?

AlexDHoffer:
We would also really like this feature! It really impacts scalability

jpkrohling:
I understand this is only waiting on a maintainer to push the button. 

cc @open-telemetry/collector-maintainers

atoulme:
#7578 is merged, can this issue be marked as completed?

Yangningli:
The receiver adds label with key node_identifier_host_name to the collected data samples, and its value is from net.host.name (or host.name for some OTLP versions) resource attributes defined in OpenTelemetry proto, for identification of the metric data.
the question: how to configuration "node_identifier_host_name"  receiver in OpenTelemetry Collector?

PR #7325, opened 2023-03-06
Add batchprocessor support for client metadata

jmacd:
**Description:** Add support for batching by metadata keys. We are aware that this has been frequently requested.

This work was prioritized on our team because it is a prerequisite for the OTel-Arrow compression bridge, see https://github.com/open-telemetry/community/issues/1332 and https://github.com/open-telemetry/oteps/pull/171

**Link to tracking Issue:** #4544 

**Testing:** One new test was added.

**Documentation:** The README was updated.


bogdandrutu (commented):
Question: Do you need to batch by metadata or you need the batch to preserve the metadata? Are both needed or just the preservation?

processor/batchprocessor/batch_processor.go
```diff
@@ -41,21 +44,57 @@ import (
...
+	// processor refers to this processor, for access to common
+	// configuration.
+	processor *batchProcessor
+
+	// exportCtx is a context with the metadata key-values
+	// corresponding with this batcher set.
+	exportCtx context.Context
+
+	// timer informs the batcher send a batch.
+	timer *time.Timer
```
bogdandrutu, line 87: Would having a large number of timers (cardinality of the metadata) be a serious problem for the performance?

processor/batchprocessor/batch_processor.go
```diff
@@ -23,12 +23,14 @@ import (
...
 	"go.uber.org/zap"
 
+	"go.opentelemetry.io/collector/client"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/collector/processor"
+	"go.opentelemetry.io/otel/attribute"
```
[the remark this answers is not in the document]
jmacd: I can, and I will, but I in passing I want to state that:

- why not use this (stable, high-performance, battle-tested API in the OTel-Go repo)?
- it's outrageous not to have a utility function available for use across the collector code base for exactly this purpose. ESPECIALLY since you've removed the ability to stable-sort a pcommon.Map in place, IMO.

processor/batchprocessor/batch_processor.go
jmacd: It's a documented feature of this implementation.  I don't think you're asking whether I want to add a limit, and I don't think you're suggesting I should implement a single timer across all the batchers.

I think this approach is good enough to start with, especially since this is a common feature request. If the user implements batching-by-metadata correctly, it will not be a problem. If necessary, the user should install an Auth extension to limit the cardinality of the input--that would be my recommendation if this becomes a problem. 

If pressed, I would implement a hard limit, one where the collector refuses to batch above a hard limit in terms of the cardinality seen so far. WDYT?

processor/batchprocessor/batch_processor.go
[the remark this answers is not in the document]
jmacd: [fd9a671](https://github.com/open-telemetry/opentelemetry-collector/pull/7325/commits/fd9a671049de8ecbf12036546465c67269a02c6e)

jmacd:
> Question: Do you need to batch by metadata or you need the batch to preserve the metadata? Are both needed or just the preservation?

I need both for the OTel-Arrow compression bridge, see https://github.com/open-telemetry/community/issues/1332 and https://github.com/open-telemetry/oteps/pull/171 because I want the use of a bridge to be transparent with regards to selected metadata.

processor/batchprocessor/README.md
```diff
@@ -29,6 +29,9 @@ regardless of size.
   `0` means no upper limit of the batch size.
   This property ensures that larger batches are split into smaller units.
   It must be greater than or equal to `send_batch_size`.
+- `metadata_keys` (default = empty): When set, this processor will
```
djaglowski, line 32: I'm missing a lot of context here, but it's unclear to me how "metadata keys" correlates to the OTel data model. Are these resource attribute keys?

processor/batchprocessor/batch_processor.go
```diff
@@ -31,6 +34,15 @@ import (
 	"go.opentelemetry.io/collector/processor"
 )
 
+// attributeSet is a stand-in for otel-go's attribute.Set.  See
+// https://github.com/open-telemetry/opentelemetry-collector/pull/7325#discussion_r1126972549.
+type attributeSet string
+
+// attributeSet is a stand-in for otel-go's attribute.KeyValue, which
```
ericmustin, line 41:
```suggestion
// attributeKeyValue is a stand-in for otel-go's attribute.KeyValue, which
```

processor/batchprocessor/batch_processor.go
```diff
@@ -71,147 +119,235 @@ var _ consumer.Traces = (*batchProcessor)(nil)
...
+		// Lookup the value in the incoming metadata, copy it
+		// into the outgoing metadata, and create a unique
+		// value for the attributeSet.
+		vs := info.Metadata.Get(k)
+		md[k] = vs
+		if len(vs) == 1 {
+			attrs = append(attrs, attributeString(k, vs[0]))
+		} else {
+			attrs = append(attrs, attributeStringSlice(k, vs))
+		}
```
ericmustin, line 303: whats the behavior when info.Metadata.Get() returns an empty slice, ie `k` isn't present on the Metadata?

processor/batchprocessor/batch_processor.go
astencel-sumo: In my opinion, the added documentation does a good job at explaining the possible risk.

Perhaps what would help is a benchmark of sorts, showing the consumed memory for different cardinalities of metadata. Not sure if it makes sense.

processor/batchprocessor/config.go
```diff
@@ -42,5 +53,13 @@ func (cfg *Config) Validate() error {
 	if cfg.SendBatchMaxSize > 0 && cfg.SendBatchMaxSize < cfg.SendBatchSize {
 		return errors.New("send_batch_max_size must be greater or equal to send_batch_size")
 	}
+	uniq := map[string]bool{}
+	for _, k := range cfg.MetadataKeys {
+		l := strings.ToLower(k)
+		if _, has := uniq[l]; has {
+			return errors.New("duplicate entry in metadata_keys")
```
astencel-sumo, line 60: Not a big deal, but perhaps it would be nice to point to the duplicate key in the log message.

processor/batchprocessor/config.go
```diff
@@ -33,6 +34,16 @@ type Config struct {
...
 	// Default value is 0, that means no maximum size.
 	SendBatchMaxSize uint32 `mapstructure:"send_batch_max_size"`
+
+	// MetadataKeys is a list of client.Metadata keys that will be
+	// used to form distinct batchers.  If this setting is empty,
+	// a single batcher instance will be used.  When this setting
+	// is not empty, one batcher will be used per distinct
+	// combination of values for the listed metadata keys.
+	//
+	// Entries are case-insensitive.  Duplicated entries will be
```
astencel-sumo, line 44: "Duplicate entries will be ignored" is not true according to the current implementation.

processor/batchprocessor/batch_processor.go
codeboten: It would be good to add a limit that's configurable to guard against unbound number of keys in the metadata. The number of batchers may be useful to track in a metric.

processor/batchprocessor/batch_processor.go
ericmustin: thanks for the clarification 👍

processor/batchprocessor/batch_processor.go
jmacd: Done with `MetadataCardinalityLimit` and an otel metric name `otelcol_batch_cardinality_limit`. Note I've added only an OTel instrument, not an OpenCensus instrument.

processor/batchprocessor/batch_processor.go
```diff
@@ -98,120 +199,188 @@ func (bp *batchProcessor) Capabilities() consumer.Capabilities {
...
+
+func (bp *batchProcessor) findBatcher(ctx context.Context) (*batcher, error) {
+	if bp.singleton != nil {
+		return bp.singleton, nil
+	}
+
+	// Get each metadata key value, form the corresponding
+	// attribute set for use as a map lookup key.
+	info := client.FromContext(ctx)
+	md := map[string][]string{}
```
dmitryax, line 300: Given that `metadataKeys` is a sorted list of unique keys that we know in advance, can't we just use an array of values as the key `[len(bp.metadataKeys)]string` instead of constructing the `aset`?

processor/batchprocessor/batch_processor.go
```diff
@@ -98,120 +199,188 @@ func (bp *batchProcessor) Capabilities() consumer.Capabilities {
...
+	bp.lock.Lock()
+	defer bp.lock.Unlock()
+
+	b, ok := bp.batchers[aset]
+	if ok {
+		return b, nil
+	}
+
+	if limit := bp.metadataLimit; limit != 0 && len(bp.batchers) >= int(limit) {
+		return nil, errTooManyBatchers
```
dmitryax, line 325: What do we do with batchers that don't receive any data for a while? Shouldn't we add some expiration mechanism for them?

processor/batchprocessor/batch_processor.go
dmitryax: Sorry, disregard this comment. Array size must be set from a constant

processor/batchprocessor/batch_processor.go
jmacd: Adding an expiration mechanism is not out of the question, but it raises new synchronization issues that I wanted to bypass. I do not expect that expiration would matter in the cases that we care about.

processor/batchprocessor/README.md
```diff
@@ -43,6 +46,34 @@ processors:
...
+authorization metadata.  For example
+
+```yaml
+processors:
+  batch:
+    # batch data by tenant-id
+    metadata_keys:
+    - tenant_id
+
+    # limit to 100 batcher processes before raising errors
```
codeboten, line 62:
```suggestion
    # limit to 10 batcher processes before raising errors
```

processor/batchprocessor/factory.go
```diff
@@ -44,8 +44,9 @@ func NewFactory() processor.Factory {
 
 func createDefaultConfig() component.Config {
 	return &Config{
-		SendBatchSize: defaultSendBatchSize,
-		Timeout:       defaultTimeout,
+		SendBatchSize:            defaultSendBatchSize,
+		Timeout:                  defaultTimeout,
+		MetadataCardinalityLimit: 100,
```
codeboten, line 49: nit: define this in the const above
```suggestion
		MetadataCardinalityLimit: defaultMetadataCardinalityLimit,
```

codeboten (approved):
Just a couple of non blocking comments on my end, PTAL

codeboten:
@bogdandrutu @dmitryax @astencel-sumo @ericmustin please review and approve if your comments have been addressed

ericmustin (approved):
lgtm

processor/batchprocessor/README.md
```diff
@@ -29,6 +29,9 @@ regardless of size.
   `0` means no upper limit of the batch size.
   This property ensures that larger batches are split into smaller units.
   It must be greater than or equal to `send_batch_size`.
+- `metadata_keys` (default = empty): When set, this processor will
+  create one batcher instance per distinct combination of values in
+  the `client.Metadata`.  See notes about metadata batching below.
```
astencel-sumo, line 34: Can you add `metadata_cardinality_limit` to this list of options?

astencel-sumo (approved):

processor/batchprocessor/batch_processor.go
```diff
@@ -98,120 +199,188 @@ func (bp *batchProcessor) Capabilities() consumer.Capabilities {
...
+	bp.lock.Lock()
+	defer bp.lock.Unlock()
+	return len(bp.batchers)
 }
 
 // ConsumeTraces implements TracesProcessor
-func (bp *batchProcessor) ConsumeTraces(_ context.Context, td ptrace.Traces) error {
-	bp.newItem <- td
+func (bp *batchProcessor) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
+	b, err := bp.findBatcher(ctx)
```
jpkrohling, line 343: I feel like there's room for a small improvement here: given that the decision to use either a singleton batcher or individual batchers is known when the processor is built, perhaps each could have its own implementation of `findBatcher` and this would be set as part of the processor? The hot path here would then just call the right implementation. 

In the current implementation, it saves only a conditional check, which might not justify the improvement though.

processor/batchprocessor/batch_processor_test.go
```diff
@@ -819,3 +822,168 @@ func TestShutdown(t *testing.T) {
...
+	spansPerRequest := 33
+	sentResourceSpans := ptrace.NewTraces().ResourceSpans()
+	for requestNum := 0; requestNum < requestCount; requestNum++ {
+		td := testdata.GenerateTraces(spansPerRequest)
+		spans := td.ResourceSpans().At(0).ScopeSpans().At(0).Spans()
+		for spanIndex := 0; spanIndex < spansPerRequest; spanIndex++ {
+			spans.At(spanIndex).SetName(getTestSpanName(requestNum, spanIndex))
+		}
+		td.ResourceSpans().At(0).CopyTo(sentResourceSpans.AppendEmpty())
+		// Choose a random context
```
jpkrohling, line 908: Is this only so that all contexts have some data? Personally, I try to make the tests predictable and reproducible, so that it gets easier to debug if it fails. If I'm reading this right, you could replace the random bits with `requestNum % len(callCtxs)`.

processor/batchprocessor/config.go
```diff
@@ -42,5 +61,13 @@ func (cfg *Config) Validate() error {
 	if cfg.SendBatchMaxSize > 0 && cfg.SendBatchMaxSize < cfg.SendBatchSize {
 		return errors.New("send_batch_max_size must be greater or equal to send_batch_size")
 	}
+	uniq := map[string]bool{}
+	for _, k := range cfg.MetadataKeys {
+		l := strings.ToLower(k)
+		if _, has := uniq[l]; has {
+			return fmt.Errorf("duplicate entry in metadata_keys: %s (case-insensitive)", l)
```
jpkrohling, line 68:
```suggestion
			return fmt.Errorf("duplicate entry in metadata_keys: %q (case-insensitive)", l)
```

processor/batchprocessor/factory.go
```diff
@@ -28,8 +28,9 @@ const (
 	// The value of "type" key in configuration.
 	typeStr = "batch"
 
-	defaultSendBatchSize = uint32(8192)
-	defaultTimeout       = 200 * time.Millisecond
+	defaultSendBatchSize            = uint32(8192)
+	defaultTimeout                  = 200 * time.Millisecond
+	defaultMetadataCardinalityLimit = 100
```
jpkrohling, line 33: I can see users asking already for guidance on setting this value. Is there a reason why this is 100 and not 1000? I have a _feeling_ that 100 is too low, but I don't know the impact of having 1000 as the default. Have you played with the settings, and can you write a couple of sentences on picking the right value?

jmacd:
I will process all the feedback above, today. Thanks reviewers!

jmacd:
@jpkrohling I want to draw a connection between your remarks:
- @bogdandrutu [asked me to try not to use the OTel-Go `attribute.Set`](https://github.com/open-telemetry/opentelemetry-collector/pull/7325#discussion_r1126972549), so I replaced it quickly with inferior code, which you noticed in several ways. [I put `attribute.Set` back.](https://github.com/open-telemetry/opentelemetry-collector/pull/7325/commits/de3503242dac6eca0093d14da66371e8e7d4dd1f). I filed https://github.com/open-telemetry/opentelemetry-collector/issues/7455 to describe how we really ought to have a core component for computing attribute sets directly from pdata and how without that, it's tempting to use inferior code. Note that since `pcommon` removed the ability to `(pcommon.Map).Sort()` it's really not a trivial task to compute a distinct attribute set from pdata.
- The reason I chose to limit the number of lifetime batchers is that it avoids a complex synchronization problem, one that could be taken up in a separate PR. Note that it's exactly the same problem facing a metrics SDK in general, one I've worked several ways. There's a connected discussion in OTel-Go about [how many memory allocations are required to (safely) perform an aggregate-by-attribute-set step in a metrics pipeline](https://github.com/open-telemetry/opentelemetry-go/issues/3943), and the use of `attribute.Set` found in this is PR not ideal because it's at least two allocations per data item that passes through, that's one to form a slice of attributes (for sorting), and one to form a distinct map key (an array of attributes, in the case of `attribute.Set`). As discussed in that thread, we can optimize this to get to one allocation (or zero  allocations using a sync.Pool of temporary attribute slices), but it comes with substantial complexity, especially so if the implementation also must support eliminating stale attribute sets. https://github.com/lightstep/otel-launcher-go/blob/main/lightstep/sdk/metric/README.md demonstrates how this can be done, for the record.

Here's why I think it's OK to go with a simple approach for now and defer complex discussions about how expensive the process is and how quickly it can recycle memory when batchers go idle. I expect this batching process to be performed on the customer side of an OTel collector pipeline. The customer will be responsible for setting this number to at most the number of distinct metadata keys that they are using themselves, for it is they who benefit from batching before sending data across an expensive network connection. In this case, the customer will have as many distinct metadata sets as they have teams, or possibly environments*teams. This should be a fixed number for the customer.

This is not a feature meant to help a vendor install a batching process in their own pipeline, where the number of distinct attribute sets might be determined by the number of customers or expected to be large and variable.

processor/batchprocessor/README.md
```diff
@@ -43,6 +46,34 @@ processors:
...
+```
+
+Receivers should be configured with `include_metadata: true` so that
+metadata keys are available to the processor.
+
+Note that each distinct combination of metadata triggers the
+allocation of a new background task in the Collector that runs for the
+lifetime of the process.  The maximum number of distinct combinations
+is limited to the configured `metadata_cardinality_limit`, which
+defaults to 100.  Users of the batching processor configured with
```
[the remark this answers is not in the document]
jmacd: Now documented. (It's `otelcol_processor_batch_metadata_cardinality`.)

processor/batchprocessor/batch_processor_test.go
jmacd: Fixed.

processor/batchprocessor/factory.go
jmacd: I am not sure what kind of user will think 100 is too low. In the scenario that Lightstep cares about, the number of metadata combinations is small, it's the number of projects per customer. The customer might have multiple projects in use, but they'll know how many that is. I wrote:
```
	// defaultMetadataCardinalityLimit should be set to the number
	// of metadata configurations the user expects to submit to
	// the collector.
```

processor/batchprocessor/batch_processor.go
```diff
@@ -18,19 +18,35 @@ import (
...
 	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/consumer/consumererror"
 	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/collector/processor"
 )
 
+// errTooManyBatchers is returned when the MetadataCardinalityLimit has been reached.
+var errTooManyBatchers = consumererror.NewPermanent(fmt.Errorf("too many batcher metadata-value combinations"))
```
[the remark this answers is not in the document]
jmacd: Done.

processor/batchprocessor/batch_processor.go
```diff
@@ -41,21 +57,60 @@ import (
...
+	batchFunc func() batch
+
+	// metadataKeys is the configured list of metadata keys.  When
+	// empty, the `singleton` batcher is used.  When non-empty,
+	// each distinct combination of metadata keys and values
+	// triggers a new batcher, counted in `goroutines`.
+	metadataKeys []string
+
+	// metadataLimit is the limiting size of the batchers map.
+	metadataLimit uint32
```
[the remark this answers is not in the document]
jmacd: The config structs are uint32 to avoid negative values, so I thought it made sense to keep that type. I'll switch to int here.

processor/batchprocessor/batch_processor.go
```diff
@@ -71,24 +126,70 @@ var _ consumer.Traces = (*batchProcessor)(nil)
...
-		batch:            batch,
+		batchFunc:        batchFunc,
 		shutdownC:        make(chan struct{}, 1),
-	}, nil
+		metadataKeys:     mks,
+		metadataLimit:    cfg.MetadataCardinalityLimit,
+	}
+	bpt, err := newBatchProcessorTelemetry(set, bp.currentMetadataCardinality, useOtel)
+	if err != nil {
+		return nil, fmt.Errorf("error to create batch processor telemetry %w", err)
```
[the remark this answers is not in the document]
jmacd: (It was pre-existing this way -- now fixed.)

processor/batchprocessor/batch_processor.go
```diff
@@ -98,120 +199,188 @@ func (bp *batchProcessor) Capabilities() consumer.Capabilities {
...
+
+	// aset.ToSlice() returns the sorted, deduplicated,
+	// and name-downcased list of attributes.
+	b = bp.newBatcher(md)
+	bp.batchers[aset] = b
+	return b, nil
+}
+
+func (bp *batchProcessor) currentMetadataCardinality() int {
+	bp.lock.Lock()
```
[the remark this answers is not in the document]
jmacd: This lock will be contended by writers on lines 294-295 and will only be read occasionally, so I don't think that would be an improvement. [I've commented on how the need to maintain a map of attribute-sets while avoiding lock contention, allocations and also supporting recycling is a tricky problem, one faced by a metrics SDK.](https://github.com/open-telemetry/opentelemetry-collector/pull/7325#issuecomment-1497905039)

processor/batchprocessor/README.md
jmacd: Done.

processor/batchprocessor/batch_processor.go
jmacd: Done.

processor/batchprocessor/batch_processor.go
```diff
@@ -71,24 +126,70 @@ var _ consumer.Traces = (*batchProcessor)(nil)
...
+	return b
+}
+
+// newAttributeSet is like otel-go's attribute.NewSet(attrs...).
+func newAttributeSet(attrs ...attributeKeyValue) attributeSet {
+	// Note: Key uniqueness is ensured in (Config).Validate()
+	// and sorted order is ensured in newBatchProcessor().
+	var aset attributeSet
+	for _, attr := range attrs {
+		aset = attributeSet(fmt.Sprint(aset, ";", attr))
```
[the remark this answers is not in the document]
jmacd: See https://github.com/open-telemetry/opentelemetry-collector/pull/7325#issuecomment-1497905039

processor/batchprocessor/batch_processor.go
```diff
@@ -71,24 +126,70 @@ var _ consumer.Traces = (*batchProcessor)(nil)
...
+	}
+	return aset
+}
+
+// attributeString is like otel-go's attribute.String, for input
+// to newAttributeSet.
+func attributeString(k string, v string) attributeKeyValue {
+	return attributeKeyValue(fmt.Sprint(k, "=", v))
```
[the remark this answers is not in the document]
jmacd: https://github.com/open-telemetry/opentelemetry-collector/pull/7325#issuecomment-1497905039

processor/batchprocessor/batch_processor.go
```diff
@@ -71,24 +126,70 @@ var _ consumer.Traces = (*batchProcessor)(nil)
...
+}
+
+// attributeStringSlice is like otel-go's attribute.StringSlice, for
+// input to newAttributeSet.
+func attributeStringSlice(k string, v []string) attributeKeyValue {
+	return attributeKeyValue(fmt.Sprint(k, "∈", strings.Join(v, ",")))
```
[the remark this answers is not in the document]
jmacd: This was the "inferior" code described in https://github.com/open-telemetry/opentelemetry-collector/pull/7325#issuecomment-1497905039.

jmacd:
image
I've looked at which lines are untested.and think this bar may be too high.

processor/batchprocessor/batch_processor.go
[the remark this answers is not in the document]
jpkrohling: Thank you for making this consistent with the current code, even though uint32 makes sense for the new (and even existing) values.

processor/batchprocessor/factory.go
jpkrohling: My view needs to be adjusted to the reality of common users of this, but when I think about metadata combinations, I think of tenant-id and cluster-name, for instance. In that case, 5 clusters and 20 tenants would already be 100 potential combinations. Going back to what I had originally in mind: is there guidance we can provide on the costs of 1000 vs. 100? Is it mostly about memory? There might also be a processing/scheduling penalty, but I believe this will be mostly memory, right?

jpkrohling (approved):
LGTM, perhaps a specific readme doc can be improved, but this is not blocking.

jmacd:
I have re-opened this PR https://github.com/open-telemetry/opentelemetry-collector/pull/7578. I had accidentally force-pushed the branch and it's easier now to create a new PR, sorry. Identical code as this pr had through https://github.com/open-telemetry/opentelemetry-collector/pull/7578/commits/9ab2157e3805e08aa62ee363b6abdbe749c1cc49, followed by merge with main, conflict resolution, and then https://github.com/open-telemetry/opentelemetry-collector/pull/7578/commits/e7a79198de6dba8928807650efe0e3133de15a82

Issue #7042, opened 2023-01-27
Review featuregate.FlagValue


bogdandrutu:
Few caveats that I found so far:

1. The need of applying the the flag to the registry. Would be good if we can pass a registry to the flag constructor (or Registry.ToFlag something like this API) and when the flags are parsed the values are Set to the Registry.
2. The "String" represents only the values set in the Flag not all the gates. This can be solved similarly with 1.
3. This [logic](https://github.com/open-telemetry/opentelemetry-collector/blob/main/featuregate/flags.go#L67) seems very strange. I think we should remove this.

Possible solutions for 1.0:
* Move it from the current package to `otelcol` and delay the fixes.
* Fix now.

PR #7356, opened 2023-03-11
Replace existing featuregate.FlagValue with a direct apply version

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/7042

Uses direct values to/from a given registry. Simplifies usage, defers the "set" logic to the Registry.


djaglowski:
👍 I see it now. The referenced code is deprecated, not changed.

Issue #7370, opened 2023-03-14
Deprecate component.Host.GetExporters


djaglowski:
**Is your feature request related to a problem? Please describe.**

[`component.Host.GetExporters`](https://github.com/open-telemetry/opentelemetry-collector/blob/ff09ab196c59c1049469419d00bd37ca132e5580/component/host.go#L57-L61) has a comment that states `This is an experimental function that may change or even be removed completely.` With the addition of connectors we should further question whether this function should exist.

**Describe the solution you'd like**
I'm proposing that we deprecate the function once the `service.connectors` featuregate is marked stable. (Currently proposed to become beta in #7369).

This function is not only unnecessary, but potentially problematic for at least the following reasons:
1. It provides any component with the ability to start and stop any exporter. While it's not expected that this capability would be misused, there doesn't appear to be any valid reason why this should be possible, nor is it clear that this would ever be safe.
2. It allows any component to emit data directly to any exporter. This is not the intention of the collector's pipeline model. Rather, it _was_ a necessary workaround to solve various use cases that can now be solved with connectors.
3. Components that currently rely on this functionality may violate subtle concerns, such as data mutability, which would otherwise not be a concern of any component. The same concerns should be handled by the service itself, which is necessarily the case with an implementation that requires data to flow through pipelines.

The following components currently use this function:
- No uses in core, other than self-references and examples.
- `routingprocessor` routes data directly to exporters. This component would be much more naturally implemented as a connector.  
- `spanmetricsprocessor` generates data and emits it directly to an exporter. Already ported to a connector.
- `datadogprocessor` generates data and emits it directly to an exporter. Could easily be ported to a connector.
- `servicegraphprocessor` generates data and emits it directly to an exporter. Already ported to a connector.
- A couple other receivers use the function when mocking a `component.Host`. Wouldn't be necessary at all.
- `k8sclusterreceiver` sends metadata directly to exporters. This is the only usage that is not replaced by connectors. However, the implementation is quite unusual and seems like it should be implemented as an extension anyways.

mx-psi:
I support this, if there are no objections, should we create issues on contrib for the individual components that need some work?

djaglowski:
@mx-psi, I've created tracking issues in contrib and updated my proposal to include them in a checklist.

In my opinion, we are better off deprecating the function sooner rather than later. This will prevent new usages and give any external users as much notification as possible. I will also create a PR for this, but of course this can be merged if/when more approvers support it.

[...]

PR #7390, opened 2023-03-15
Deprecate component.Host.GetExporters

djaglowski:
See #7370

### DIFF
--- a/component/host.go
+++ b/component/host.go
@@ -48,7 +48,7 @@ type Host interface {
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
 	//
-	// Deprecated: [0.75.0] This function will be removed no earlier than 0.80.0.
+	// Deprecated: [0.75.0] This function will be removed in the future.
 	// Several components in the contrib repository use this function so it cannot be removed
 	// before those cases are removed. In most cases, use of this function can be replaced by a
 	// connector. See https://github.com/open-telemetry/opentelemetry-collector/issues/7370 and
--- a/component/host.go
+++ b/component/host.go
@@ -47,6 +47,12 @@ type Host interface {
 	//
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
-	// Deprecated: [0.74.0] Use a connector or extension to send data between pipelines.
+	//
+	// Deprecated: [0.75.0] This function will be removed no earlier than 0.80.0.
+	// Several components in the contrib repository use this function so it cannot be removed
+	// before those cases are removed. In most cases, use of this function can be replaced by a
+	// connector. See https://github.com/open-telemetry/opentelemetry-collector/issues/7370 and
+	// https://github.com/open-telemetry/opentelemetry-collector/pull/7390#issuecomment-1483710184
+	// for additional information.
 	GetExporters() map[DataType]map[ID]Component
 }
--- a/component/host.go
+++ b/component/host.go
@@ -48,7 +48,7 @@ type Host interface {
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
 	//
-	// Deprecated: [0.75.0] This function will be removed in the future.
+	// Deprecated: [0.79.0] This function will be removed in the future.
 	// Several components in the contrib repository use this function so it cannot be removed
 	// before those cases are removed. In most cases, use of this function can be replaced by a
 	// connector. See https://github.com/open-telemetry/opentelemetry-collector/issues/7370 and
--- a/component/host.go
+++ b/component/host.go
@@ -47,5 +47,6 @@ type Host interface {
 	//
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
+	// Deprecated: [0.74.0] Use a connector or extension to send data between pipelines.
 	GetExporters() map[DataType]map[ID]Component
 }
--- a/service/host.go
+++ b/service/host.go
@@ -57,7 +57,7 @@ func (host *serviceHost) GetExtensions() map[component.ID]component.Component {
 	return host.serviceExtensions.GetExtensions()
 }
 
-// Deprecated: [0.75.0] This function will be removed no earlier than 0.80.0.
+// Deprecated: [0.79.0] This function will be removed in the future.
 // Several components in the contrib repository use this function so it cannot be removed
 // before those cases are removed. In most cases, use of this function can be replaced by a
 // connector. See https://github.com/open-telemetry/opentelemetry-collector/issues/7370 and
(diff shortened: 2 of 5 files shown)


mx-psi (approved):
LGTM, should we also deprecate https://github.com/open-telemetry/opentelemetry-collector/blob/042b8763cc7e25cd5dc8a520887822c10bc774a9/service/host.go#L70 and https://github.com/open-telemetry/opentelemetry-collector/blob/042b8763cc7e25cd5dc8a520887822c10bc774a9/service/graph.go#L318

to avoid further internal usage of these methods?

Also, we may need to add exceptions for existing components on contrib that use this method while we migrate them

component/host.go
```diff
@@ -58,5 +58,6 @@ type Host interface {
 	//
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
+	// Deprecated: [0.74.0] Use a connector or extension to send data between pipelines.
```
djaglowski, line 61:
```suggestion
	// Deprecated: [0.75.0] Use a connector or extension to send data between pipelines.
```

service/host.go
```diff
@@ -68,6 +68,7 @@ func (host *serviceHost) GetExtensions() map[component.ID]component.Component {
 	return host.serviceExtensions.GetExtensions()
 }
 
+// Deprecated: [0.74.0] Use a connector or extension to send data between pipelines.
```
djaglowski, line 71:
```suggestion
// Deprecated: [0.75.0] Use a connector or extension to send data between pipelines.
```

service/internal/graph/graph.go
```diff
@@ -321,6 +321,7 @@ func (g *Graph) ShutdownAll(ctx context.Context) error {
 	return errs
 }
 
+// Deprecated: [0.74.0] Use a connector or extension to send data between pipelines.
```
djaglowski, line 324:
```suggestion
// Deprecated: [0.75.0] Use a connector or extension to send data between pipelines.
```

dmitryax:
> There are some usages that cannot be converted I believe, one is in the signalfx receiver. cc @dmitryax

At Splunk, we rely on metadata updates sent directly from receivers to exporters. The same approach is implemented in k8scluster, which we have an issue for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/19741. The effort to replace that mechanism is significant. So maybe we can wait until for alternative implementation before the deprecation?

dmitryax:
Also, we usually resolve all the to-be-deprecated usages in contrib before the actual deprecation in the core. Then, the deprecated API is usually removed on the next release. That way, we avoid adding `//nolint` comments in contrib. This deprecation should definitely stay longer than one version, but I think deprecating it now is too aggressive anyway. I'd like to have at least some solution for the metadata updates first.

djaglowski:
> There are some usages that cannot be converted I believe, one is in the signalfx receiver. cc @dmitryax

https://github.com/open-telemetry/opentelemetry-collector/issues/7370 details all uses in core and contrib. `k8sclusterreciever -> signalfxexporter` is the only case that cannot be solved with a connector, but it can be solved with an extension.

> we usually resolve all the to-be-deprecated usages in contrib before the actual deprecation in the core.

In this case at least, I wonder if by waiting to deprecate we could be inviting additional novel uses that would have to be untangled. If you think it'll be a while before `k8sclusterreciever -> signalfxexporter` can use an extension, perhaps we could deprecate when that is the only remaining case? Then we only have one `//nolint` but in exchange we could discourage misuse.

dmitryax:
> https://github.com/open-telemetry/opentelemetry-collector/issues/7370 details all uses in core and contrib. k8sclusterreciever -> signalfxexporter is the only case that cannot be solved with a connector, but it can be solved with an extension.

We also use that in some receivers in Splunk distribution, but it's the same metadata use case.

> In this case at least, I wonder if by waiting to deprecate we could be inviting additional novel uses that would have to be untangled. If you think it'll be a while before k8sclusterreciever -> signalfxexporter can use an extension, perhaps we could deprecate when that is the only remaining case? Then we only have one //nolint but in exchange we could discourage misuse.

Sounds good to me. Let's migrate at least a couple of components before deprecating. And I'll start planning the migration for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/19741. I'm good with deprecating it earlier, but we need a more descriptive message telling users it'll stay until all the usages in contrib are migrated with a link to https://github.com/open-telemetry/opentelemetry-collector/issues/7370. Currently, it seems like it'll be dropped next release as it's usually done for other deprecations.

djaglowski:
> I'm good with deprecating it earlier, but we need a more descriptive message telling users it'll stay until all the usages in contrib are migrated with a link to #7370. Currently, it seems like it'll be dropped next release as it's usually done for other deprecations.

Good points. I'll update the comment and will continue pushing forward with the other uses.

component/host.go
```diff
@@ -58,5 +58,12 @@ type Host interface {
 	//
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
+	//
+	// Deprecated: [0.75.0] This function will be removed no earlier than 0.80.0.
+	// Several components in the contrib repository use this function so it cannot be removed
```
dmitryax, line 63:
```suggestion
	// Deprecated: [0.75.0] This function will be removed in the future.
	// Several components in the contrib repository use this function so it cannot be removed
```

dmitryax (approved):
I think we can merge it. Let's just remove the target version for now?

dmitryax (approved):
I think we can merge it. Let's just remove the target version for now?

component/host.go
```diff
@@ -58,5 +58,12 @@ type Host interface {
 	//
 	// GetExporters can be called by the component anytime after Component.Start() begins and
 	// until Component.Shutdown() ends.
+	//
+	// Deprecated: [0.75.0] This function will be removed in the future.
```
djaglowski, line 62:
```suggestion
	// Deprecated: [0.79.0] This function will be removed in the future.
```

service/host.go
```diff
@@ -68,6 +68,12 @@ func (host *serviceHost) GetExtensions() map[component.ID]component.Component {
 	return host.serviceExtensions.GetExtensions()
 }
 
+// Deprecated: [0.75.0] This function will be removed no earlier than 0.80.0.
```
djaglowski, line 71:
```suggestion
// Deprecated: [0.79.0] This function will be removed in the future.
```

service/internal/graph/graph.go
```diff
@@ -321,6 +321,12 @@ func (g *Graph) ShutdownAll(ctx context.Context) error {
 	return errs
 }
 
+// Deprecated: [0.75.0] This function will be removed no earlier than 0.80.0.
```
djaglowski, line 324:
```suggestion
// Deprecated: [0.79.0] This function will be removed in the future.
```

mx-psi (approved):

dmitryax (approved):

PR #7397, opened 2023-03-20
[chore] update documentation to use feature gate instead of flag

codeboten:


CONTRIBUTING.md
```diff
@@ -457,13 +457,13 @@ func DoFoo() {}
...
-flag](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
+gate](https://github.com/open-telemetry/opentelemetry-collector/blob/6b5a3d08a96bfb41a5e121b34f592a1d5c6e0435/service/featuregate/)
 in a subsequent version, and eventually removing it after yet another version. This is how it would look like, considering
 that each of the following steps is done in a separate version:
 
-1. Mark the feature as deprecated, add a short lived feature flag with the feature enabled by default
-1. Change the feature flag to disable the feature by default, deprecating the flag at the same time
-1. Remove the feature and the flag
+1. Mark the feature as deprecated, add a short lived feature gate with the feature enabled by default
+1. Change the feature gate to disable the feature by default, deprecating the gate at the same time
```
bogdandrutu, line 465: "Deprecating" -> this is marking as stable correct? Also should this be a separate step since we do not allow changing the gate at that point?

bogdandrutu, line 465: "Deprecating" -> this is marking as stable correct? Also should this be a separate step since we do not allow changing the gate at that point?

bogdandrutu (approved):

CONTRIBUTING.md
codeboten: Right, I guess this should be re-written to include the stages of the gate. I was just renaming flag to gate :D

bogdandrutu: Sure, let's do that :)

bogdandrutu (approved):

PR #7420, opened 2023-03-23
[chore] update merged pull request w/ milestone (#293)

codeboten:
Following the conversation in the SIG call on March 22, the github action in this PR will automatically add a merged PR to the first open milestone with `next version` name it finds.

I've tested this code in my fork

### DIFF
--- a/.github/workflows/milestone-add-to-pr.yml
+++ b/.github/workflows/milestone-add-to-pr.yml
@@ -1,0 +1,33 @@
+# This action adds the latest milestone to a pull request
+# when it is merged
+
+name: 'Project: Add PR to Milestone'
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  if_merged:
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            const milestones = await github.rest.issues.listMilestones({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: "open"
+            })
+            for (const milestone of milestones.data) {
+              if (milestone.title.startsWith("v")) {
+                await github.rest.issues.update({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: ${{ github.event.pull_request.number }},
+                  milestone: milestone.number
+                });
+                return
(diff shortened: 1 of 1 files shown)


.github/workflows/milestone-add-to-pr.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            const milestones = await github.rest.issues.listMilestones({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: "open"
+            })
+            for (const milestone of milestones.data) {
+              if (milestone.title.startsWith("v")) {
```
TylerHelmuth, line 24: It would be create if we could have a more rigorous identifier.  Any chance we could look for the milestone with a specific label like `next release`?

codeboten: I find having the milestone be deterministic and kept for posterity useful.

If we decided to go w/ `next release` then we'd remove all the items tagged as next release w/ each release. This would remove the ability from someone to look at a pr and see where it landed. I suppose they could always get that by looking at the merged commit instead.

Additionally, milestones do give users an easy view into what changes made it into them. Tags don't quite do the same.

This is all my opinion of course, if folks feel strongly otherwise, i'm happy to agree w/ whatever the consensus is.

TylerHelmuth: Oh I don't mean tagging PRs with the label, I mean the milestone itself.  Instead of hoping the only open milestone to start with `v` is the `v0.*.0` release milestone, when we create the new milestone as part of the release process we could add a `Next Release` tag to the new milestone and remove it from the old.  I might be being too paranoid though about the `v`.

codeboten: They can definitely be renamed. This could work, and it would save me the trouble of figuring out how to auto-increment the milestone for the next future release 👍🏻

TylerHelmuth: Turns out you can't label milestones anyways so everything I said was wrong lol I like @Aneurysm9 idea.

codeboten: Updated to use `next version`

Aneurysm9 (approved):

PR #7434, opened 2023-03-27
[service/proctelemetry] Unset HOST_PROC to make sure we use and report the actual process and introduce an option  to allow user to overwrite the unset logic and use the value set in the environment variable

dloucasfx:
**Description:** 

While debugging the below error in k8s env
````
Error: failed to register process metrics: process does not exist
2023/03/23 03:44:47 main.go:115: application run finished with error: failed to register process metrics: process does not exist
````
I have noticed that the metric server is calling GOPSUTIL while the HOST_PROC variable is set , this causes gopsutil `PidExistsWithContext ` to retrieve the process from the host instead from the container 

````
func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
	if pid <= 0 {
		return false, fmt.Errorf("invalid pid %v", pid)
	}
	proc, err := os.FindProcess(int(pid))
	if err != nil {
		return false, err
	}

	if isMount(common.HostProc()) { // if /<HOST_PROC>/proc exists and is mounted, check if /<HOST_PROC>/proc/<PID> folder exists
		_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))
		if os.IsNotExist(err) {
			return false, nil
		}
		return err == nil, err
	}
````
This PR unsets and resets the host_proc variable and introduces an option to allow the use of host_proc if for whatever reason they need to

unit tests

### DIFF
--- a/service/internal/proctelemetry/process_telemetry.go
+++ b/service/internal/proctelemetry/process_telemetry.go
@@ -63,7 +63,7 @@ type processMetrics struct {
 
 // RegisterProcessMetrics creates a new set of processMetrics (mem, cpu) that can be used to measure
 // basic information about this process.
-func RegisterProcessMetrics(ocRegistry *metric.Registry, mp otelmetric.MeterProvider, useOtel bool, ballastSizeBytes uint64) error {
+func RegisterProcessMetrics(ocRegistry *metric.Registry, mp otelmetric.MeterProvider, useOtel bool, ballastSizeBytes uint64, useHostProcEnvVar bool) error {
 	var err error
 	pm := &processMetrics{
 		startTimeUnixNano: time.Now().UnixNano(),
@@ -71,9 +71,18 @@ func RegisterProcessMetrics(ocRegistry *metric.Registry, mp otelmetric.MeterProv
 		ms:                &runtime.MemStats{},
 	}
 
+	hostproc, exists := os.LookupEnv("HOST_PROC")
+	// unset HOST_PROC env variable so the process search occurs locally
+	if !useHostProcEnvVar && exists {
+		os.Unsetenv("HOST_PROC")
+	}
 	pm.proc, err = process.NewProcess(int32(os.Getpid()))
 	if err != nil {
 		return err
+	}
+	// restore immediately if it was previously set, don't use defer
+	if !useHostProcEnvVar && exists {
+		os.Setenv("HOST_PROC", hostproc)
 	}
 
 	if useOtel {
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -16,6 +16,7 @@ package telemetry // import "go.opentelemetry.io/collector/service/telemetry"
 
 import (
 	"fmt"
+	"os"
 
 	"go.uber.org/zap/zapcore"
 
@@ -116,6 +117,10 @@ type MetricsConfig struct {
 
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
+
+	// UseHostProcEnvVar when set to true, the metric server, through gopsutil,
+	// lookup for the otel process in the proc path defined in the HOST_PROC environment variable.
+	UseHostProcEnvVar bool `mapstructure:"useHostProcEnvVar"`
 }
 
 // TracesConfig exposes the common Telemetry configuration for collector's internal spans.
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: service/internal/proctelemetry/process_telemetry.go [hunk 1]
```diff
@@ -63,18 +63,27 @@ type processMetrics struct {
...
+	if !useHostProcEnvVar && exists {
+		os.Unsetenv("HOST_PROC")
+	}
 	pm.proc, err = process.NewProcess(int32(os.Getpid()))
 	if err != nil {
 		return err
 	}
+	// restore immediately if it was previously set, don't use defer
+	if !useHostProcEnvVar && exists {
+		os.Setenv("HOST_PROC", hostproc)
```
rmfitzpatrick, line 85: Do you know if there's a chance of memoization or is there some guarantee from gopsutil that subsequent lookups w/ the restored env var will be reliably using the hostfs procfs?

rmfitzpatrick: Also this should be probably be deferred so that it always happens moving forward regardless of changes in error handling.

rmfitzpatrick: Sorry, just realized that subsequent stat updates are all though gopsutil so this would need to happen for each update cycle, which is almost certainly going to cause race conditions w/ the host metrics receiver.

service/internal/proctelemetry/process_telemetry.go
```diff
@@ -63,18 +63,27 @@ type processMetrics struct {
...
 		ballastSizeBytes:  ballastSizeBytes,
 		ms:                &runtime.MemStats{},
 	}
 
+	hostproc, exists := os.LookupEnv("HOST_PROC")
+	// unset HOST_PROC env variable so the process search occurs locally
+	if !useHostProcEnvVar && exists {
+		os.Unsetenv("HOST_PROC")
+	}
 	pm.proc, err = process.NewProcess(int32(os.Getpid()))
```
rmfitzpatrick, line 79: I think the issue is that this pid needs to be from the hostfs and not gathered from the container's syscall.

service/telemetry/config.go
```diff
@@ -116,6 +117,10 @@ type MetricsConfig struct {
 
 	// Address is the [address]:port that metrics exposition should be bound to.
 	Address string `mapstructure:"address"`
+
+	// UseHostProcEnvVar when set to true, the metric server, through gopsutil,
+	// lookup for the otel process in the proc path defined in the HOST_PROC environment variable.
+	UseHostProcEnvVar bool `mapstructure:"useHostProcEnvVar"`
```
mx-psi, line 123: Do we actually need to expose this? If it's unclear, I would remove this from the PR to avoid extending the public API surface

service/internal/proctelemetry/process_telemetry.go
```diff
@@ -63,18 +63,27 @@ type processMetrics struct {
...
 	if err != nil {
 		return err
 	}
+	// restore immediately if it was previously set, don't use defer
```
mx-psi, line 83: If we don't use `defer` and `process.NewProcess` panics we won't set the environment variable back, so I don't think this approach is acceptable. What are the challenges with using `defer`?

I think a better approach for this would be for `gopsutil` to allow us specifying the use or not of `HOST_PROC` instead of us modifying the environment variables. Have we asked the `gopsutil` maintainer about this?

mx-psi (changes requested):

service/internal/proctelemetry/process_telemetry.go
dloucasfx: yep, this is a bust! gopsutil re-use the hostproc on every call :( 
```
func (p *Process) fillFromTIDStatWithContext(ctx context.Context, tid int32) (uint64, int32, *cpu.TimesStat, int64, uint32, int32, *PageFaultsStat, error) {
	pid := p.Pid
	var statPath string

	if tid == -1 {
		statPath = common.HostProc(strconv.Itoa(int(pid)), "stat")
	} else {
		statPath = common.HostProc(strconv.Itoa(int(pid)), "task", strconv.Itoa(int(tid)), "stat")
	}
```

dloucasfx:
Thanks @rmfitzpatrick  @mx-psi for reviewing.
I based my PR on the assumption that gopsutil will only need the `host_proc` to locate the files once, but after @rmfitzpatrick pointed it out, I also noticed that it reuses `host_proc` on every call:
example when getting CPU time:
```
func (p *Process) fillFromTIDStatWithContext(ctx context.Context, tid int32) (uint64, int32, *cpu.TimesStat, int64, uint32, int32, *PageFaultsStat, error) {
	pid := p.Pid
	var statPath string

	if tid == -1 {
		statPath = common.HostProc(strconv.Itoa(int(pid)), "stat")
	} else {
		statPath = common.HostProc(strconv.Itoa(int(pid)), "task", strconv.Itoa(int(tid)), "stat")
	}

```

so looks like the fix is more involved , so we either need to:
1) wait for https://github.com/shirou/gopsutil/pull/1439 to be merged
2) or implement our own package to pull the data/or use our own fork of gopsutil that includes https://github.com/shirou/gopsutil/pull/1439 

cc: @dmitryax

dloucasfx:
closing PR and moving discussion to https://github.com/open-telemetry/opentelemetry-collector/issues/7435

Issue #7047, opened 2023-01-27
Investigate how to expose exporterhelper.NewThrottleRetry in the consumererror


bogdandrutu:

bogdandrutu:
An idea is to merge the current `consumererror.[Traces|Metrics|Logs]` with the ThrottleRetry and have a

```golang
// which contains the data that need to be retried, and the potential `delay` if any.
type consumererror.Retriable[Traces|Metrics|Logs] 
```

I am not sure we need an error per signal or we can use generics, or simply have all possible signals as part of the error. Both these options will create a `consumererror.Retriable` type which has a nice name. 

If we do this, we should also include a "failed" uint64 to the `consumererror.Permanent` that represents the number of signal entries (spans/dps (or metrics)/log records) dropped. The reason to not return like the Retirable the `pdata` is because this is permanent error, so it is useless to filter/construct a pdata with the wrong data.

dashpole:
Side note: the googlecloud exporter likely won't utilize this.  The underlying client libraries have relatively intelligent retry mechanisms built-in, and we are in the process of making a fix to the trace retry logic before we switch away from the collector's retry mechanism.

But in general, I think https://github.com/open-telemetry/opentelemetry-collector/issues/7047#issuecomment-1407161974 makes a lot of sense.

evan-bradley:
> I am not sure we need an error per signal or we can use generics, or simply have all possible signals as part of the error.

I think an error type per signal that is aliased or extended from a generic base class would keep duplication to a minimum while not requiring users to instantiate the generic types. Just including all possible signals could work, but most types are specific to a signal, so this seems unconventional. Are there any other places where this is done?

> If we do this, we should also include a "failed" uint64 to the consumererror.Permanent that represents the number of signal entries (spans/dps (or metrics)/log records) dropped.

I agree about adding a `failed` field. Are there other fields we would want? One that that comes to mind is a `total` that counts of the total number of telemetry items sent in the payload, or alternatively a `successful` to count successful items. Another would be a reason for failure (rejected by backend, backend is down, processing failure, etc.). These are both immediately useful for OTLP partial success responses, but I think would be broadly useful for any component.

jamesmoessis:
Stumbled upon this issue after encountering an issue with the load balancing exporter. Since load balancing exporter delegates to sub-exporters, only one of them needs to fail and it will retry all. In our case, this causes a lot of duplicated exporting.

[...]

PR #7439, opened 2023-03-28
[consumer] Add retryable error types and add counts to permanent errors

evan-bradley:
**Description:**
This will replace `exporterhelper.throttleRetry` with `consumererror.Retryable[Traces|Metrics|Logs]` types and add success/fail counts to the permanent error types.

Right now this PR is just intended to be a discussion point for how these types should look. I will complete a full implementation once there is consensus on how these should look.

**Link to tracking Issue:**

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/7047

### DIFF
--- a/consumer/consumererror/permanent.go
+++ b/consumer/consumererror/permanent.go
@@ -14,27 +14,60 @@
 
 package consumererror // import "go.opentelemetry.io/collector/consumer/consumererror"
 
-import "errors"
+import (
+	"errors"
+	"strconv"
+)
 
-// permanent is an error that will be always returned if its source
+// NO_DATA_COUNT indicates an unknown quantity of telemetry items.
+const NO_DATA_COUNT = -1
+
+// Permanent is an error that will always be returned if its source
 // receives the same inputs.
-type permanent struct {
-	err error
+type Permanent struct {
+	error
+	successful int
+	failed     int
 }
 
-// NewPermanent wraps an error to indicate that it is a permanent error, i.e. an
-// error that will be always returned if its source receives the same inputs.
+// Deprecated [v0.75.0] Use consumererror.NewPermanentWithCounts instead.
 func NewPermanent(err error) error {
-	return permanent{err: err}
+	return Permanent{error: err, successful: NO_DATA_COUNT, failed: NO_DATA_COUNT}
 }
 
-func (p permanent) Error() string {
-	return "Permanent error: " + p.err.Error()
+// NewPermanentWithCounts wraps an error to indicate that it is a permanent error, i.e. an
+// error that will be always returned if its source receives the same inputs.
+// The error should also include the number of successful and failed telemetry items
+// if the counts are known. Unknown counts should be set to consumererror.NO_DATA_COUNT.
+func NewPermanentWithCounts(err error, successful, failed int) error {
+	return Permanent{error: err, successful: successful, failed: failed}
 }
 
-// Unwrap returns the wrapped error for functions Is and As in standard package errors.
-func (p permanent) Unwrap() error {
-	return p.err
+func (p Permanent) Error() string {
+	successful := "unknown"
+	failed := "unknown"
+
+	if p.successful != NO_DATA_COUNT {
+		successful = strconv.Itoa(p.successful)
+	}
+	if p.failed != NO_DATA_COUNT {
+		failed = strconv.Itoa(p.failed)
+	}
+
+	return "Permanent error (" + successful + " successful, " + failed + " failed): " + p.error.Error()
+}
+
+// Unwrap returns the wrapped error for use by `errors.Is` and `errors.As`.
+func (p Permanent) Unwrap() error {
+	return p.error
+}
+
+func (p Permanent) Successful() int {
+	return p.successful
+}
+
+func (p Permanent) Failed() int {
+	return p.failed
 }
 
 // IsPermanent checks if an error was wrapped with the NewPermanent function, which
@@ -44,5 +77,5 @@ func IsPermanent(err error) bool {
 	if err == nil {
 		return false
 	}
-	return errors.As(err, &permanent{})
+	return errors.As(err, &Permanent{})
 }
--- a/consumer/consumererror/permanent.go
+++ b/consumer/consumererror/permanent.go
@@ -19,42 +19,36 @@ import (
 	"strconv"
 )
 
-// NO_DATA_COUNT indicates an unknown quantity of telemetry items.
-const NO_DATA_COUNT = -1
+// noDataCount indicates an unknown number of records.
+const noDataCount = -1
 
 // Permanent is an error that will always be returned if its source
 // receives the same inputs.
 type Permanent struct {
 	error
-	successful int
-	failed     int
+	rejected int
 }
 
-// Deprecated [v0.75.0] Use consumererror.NewPermanentWithCounts instead.
+// Deprecated [v0.76.0] Use consumererror.NewPermanentWithCount instead.
 func NewPermanent(err error) error {
-	return Permanent{error: err, successful: NO_DATA_COUNT, failed: NO_DATA_COUNT}
+	return Permanent{error: err, rejected: noDataCount}
 }
 
-// NewPermanentWithCounts wraps an error to indicate that it is a permanent error, i.e. an
+// NewPermanentWithCount wraps an error to indicate that it is a permanent error, i.e. an
 // error that will be always returned if its source receives the same inputs.
-// The error should also include the number of successful and failed telemetry items
-// if the counts are known. Unknown counts should be set to consumererror.NO_DATA_COUNT.
-func NewPermanentWithCounts(err error, successful, failed int) error {
-	return Permanent{error: err, successful: successful, failed: failed}
+// The error should also include the number of rejected records.
+func NewPermanentWithCount(err error, rejected int) error {
+	return Permanent{error: err, rejected: rejected}
 }
 
 func (p Permanent) Error() string {
-	successful := "unknown"
-	failed := "unknown"
+	rejected := "unknown"
 
-	if p.successful != NO_DATA_COUNT {
-		successful = strconv.Itoa(p.successful)
-	}
-	if p.failed != NO_DATA_COUNT {
-		failed = strconv.Itoa(p.failed)
+	if p.rejected != noDataCount {
+		rejected = strconv.Itoa(p.rejected)
 	}
 
-	return "Permanent error (" + successful + " successful, " + failed + " failed): " + p.error.Error()
+	return "Permanent error (" + rejected + " rejected): " + p.error.Error()
 }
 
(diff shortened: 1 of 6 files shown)


consumer/consumererror/permanent.go
```diff
@@ -14,27 +14,60 @@
...
+// NO_DATA_COUNT indicates an unknown quantity of telemetry items.
+const NO_DATA_COUNT = -1
+
+// Permanent is an error that will always be returned if its source
 // receives the same inputs.
-type permanent struct {
-	err error
+type Permanent struct {
+	error
+	successful int
```
evan-bradley, line 29: I added this since I think it would provide good contextual information for the failed batch.

I think `total` could also work here, but `successful` seems like a better match for a `failed` count.

consumer/consumererror/permanent.go
```diff
@@ -14,27 +14,60 @@
...
+	failed     int
```
evan-bradley, line 30: `accepted`/`rejected` could also be good terminology here, though to me those terms feel like they apply more to ingesting the data into a backend, where `successful`/`failed` I think apply more broadly to failures anywhere in the pipeline.

consumer/consumererror/retryable.go
```diff
@@ -15,17 +15,24 @@
...
 	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
 type retryable[V ptrace.Traces | pmetric.Metrics | plog.Logs] struct {
 	error
-	data V
+	delay time.Duration
+	data  V
```
evan-bradley, line 28: We could also potentially add a `successful` count to this to indicate that it is a retryable partial success-style error. While this wouldn't be useable for OTLP partial success messages, other formats that support a partial success concept while allowing retries of some or all of the data could use it.

jpkrohling (commented):
LGTM. Given the conversation we had today, I think there are a few things that can be changed here:

1) add the `toHTTP` from #7486 somewhere here, so that gRPC <-> HTTP translation can happen
2) add new constructors to be used with exporters, like `NewFromGRPCStatus(*status.Status)` and `NewFromHTTPStatus(int)`. Perhaps also the other way around would make sense, to be used in receivers: `ToHTTPStatus()` and `ToGRPCStatus()`.

consumer/consumererror/permanent.go
dmitryax: Why do we need both successful and failed? Caller always know what's the total number of items. I think we can follow what's defined in [the proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/logs/v1/logs_service.proto#L69) and keep just one `rejected`

consumer/consumererror/permanent.go
joaopgrassi: Yeah I agree. With total you can derive "successful"

consumer/consumererror/permanent.go
```diff
@@ -14,27 +14,60 @@
...
-	return permanent{err: err}
+	return Permanent{error: err, successful: NO_DATA_COUNT, failed: NO_DATA_COUNT}
+}
+
+// NewPermanentWithCounts wraps an error to indicate that it is a permanent error, i.e. an
+// error that will be always returned if its source receives the same inputs.
+// The error should also include the number of successful and failed telemetry items
+// if the counts are known. Unknown counts should be set to consumererror.NO_DATA_COUNT.
+func NewPermanentWithCounts(err error, successful, failed int) error {
+	return Permanent{error: err, successful: successful, failed: failed}
```
joaopgrassi, line 43: Would you pass `total` when creating this and then derive successful from total - failed?

consumer/consumererror/retryable.go
```diff
@@ -35,22 +42,35 @@ func (err retryable[V]) Data() V {
...
 // that failed to be processed or sent.
-type Traces struct {
+type RetryableTraces struct {
 	retryable[ptrace.Traces]
 }
 
-// NewTraces creates a Traces that can encapsulate received data that failed to be processed or sent.
-func NewTraces(err error, data ptrace.Traces) error {
-	return Traces{
+// Deprecated [v0.75.0] Use `RetryableTraces` instead
```
codeboten, line 56: This will need updating to 0.76.0

consumer/consumererror/permanent.go
```diff
@@ -14,27 +14,60 @@
...
 
-// permanent is an error that will be always returned if its source
+// NO_DATA_COUNT indicates an unknown quantity of telemetry items.
+const NO_DATA_COUNT = -1
+
+// Permanent is an error that will always be returned if its source
 // receives the same inputs.
-type permanent struct {
-	err error
+type Permanent struct {
```
codeboten, line 27: Does this need to be public? Would it be possible to return the count of rejected as part of `IsPermanent` and keep this private? Not sure if the ergonomics make sense that way

consumer/consumererror/permanent.go
```diff
@@ -14,27 +14,60 @@
...
+	error
+	successful int
+	failed     int
 }
 
-// NewPermanent wraps an error to indicate that it is a permanent error, i.e. an
-// error that will be always returned if its source receives the same inputs.
+// Deprecated [v0.75.0] Use consumererror.NewPermanentWithCounts instead.
```
codeboten, line 33:
```suggestion
// Deprecated [v0.76.0] Use consumererror.NewPermanentWithCounts instead.
```

codeboten (commented):
I think this approach makes sense, just a couple of questions. I agree w/ @dmitryax that going w/ rejected makes the most sense here.

consumer/consumererror/retryable.go
evan-bradley: Following the discussion on adding retryable partial success responses to OTLP, I think we should leave this out until there's a concrete need.

consumer/consumererror/permanent.go
evan-bradley: You're right. I took a look at where this is used, and having just the `rejected` field is sufficient. I've update the fields accordingly.

consumer/consumererror/permanent.go
evan-bradley: It looks like the successful records are calculated by the `obsreport` package and the totals are probably best left to be recorded by the caller, so I'm going to leave this out for now.

consumer/consumererror/permanent.go
evan-bradley: It looks like the `obsreport` package that records the metrics we can use these counts for does this and the total is best calculated by the caller, so I'm leaving this out for now.

consumer/consumererror/permanent.go
evan-bradley: Most uses of `IsPermanent` take advantage of the `bool` return type and look like `if IsPermanent(err) { ... }`, so I think this would ultimately reduce the ergonomics of the function. I think having the type public with a method to get the rejected items count is likely the best approach, but we can investigate alternatives if it will present a problem.

obsreport/obsreport_exporter.go
```diff
@@ -251,6 +252,10 @@ func endSpan(ctx context.Context, err error, numSent, numFailedToSend int64, sen
 
 func toNumItems(numExportedItems int, err error) (int64, int64) {
 	if err != nil {
+		if e, ok := err.(consumererror.Permanent); ok {
```
dmitryax, line 255:
```suggestion
		var permErr consumererror.Permanent
		if errors.As(err, &permErr) { {
```

Otherwise we ignore wrapped errors, right?

obsreport/obsreport_exporter.go
```diff
@@ -251,6 +252,10 @@ func endSpan(ctx context.Context, err error, numSent, numFailedToSend int64, sen
...
+			rejected := int64(e.Rejected())
```
dmitryax, line 256: this will become
```suggestion
			rejected := int64(permErr.Rejected())
```

evan-bradley:
@jpkrohling Yes, that was my intent. Sorry for the delay, I was looking into that and got stopped just short of being able to implement it. I hope to have something this week.

obsreport/obsreport_exporter.go
evan-bradley: Yes, that's a good point. I was making the assumption that we're calling this right after `ConsumeTraces`, but we should assume that the error may be wrapped.

consumer/consumererror/permanent.go
```diff
@@ -14,27 +14,50 @@
...
+// Deprecated [v0.76.0] Use consumererror.NewPermanentWithCount instead.
 func NewPermanent(err error) error {
-	return permanent{err: err}
+	return Permanent{error: err, rejected: noDataCount}
+}
+
+// NewPermanentWithCount wraps an error to indicate that it is a permanent error, i.e. an
+// error that will be always returned if its source receives the same inputs.
+// The error should also include the number of rejected records.
+func NewPermanentWithCount(err error, rejected int) error {
```
dmitryax, line 40: I'm a bit worried that we require providing the rejected field here. I'm wondering if we can keep this field optional. If not provided, the caller can assume that all the items got rejected. My concern is that some exporters return this error when there is no access to the original pdata that was translated to some other schema. It'll be difficult for them to fetch the number of the items. One option is to have
```
NewPermanent(err error, otps ...PermanentErrorOptions)
```

and supply `rejected` via `WithRejectedCount(int)` option.

Or we don't even need `Permanent` part here and assume it's always permanent unless `Retryable(ptrace.Traces)` option is provided. Later we might have some other options.

These are just thoughts, not suggestions for changes. I believe we need to think more about an overarching design for this API that would cover all the use cases, including embedding the GRPC/HTTP and possible future additions. Maybe having some small design doc would be beneficial.

evan-bradley: I agree that having a design doc would be good, I'll think through some of your points and add them to a design doc here. If it looks like the PR is growing we can move the document to a separate PR.

jpkrohling:
Folks, what's the state of this? This is blocking #7486 and I would really love to move forward with that, perhaps even with a suboptimal solution at first, and a proper solution in the longer term.

evan-bradley:
@jpkrohling Sorry, I had to focus on a few other things and haven't made as much progress on this as I would have liked. In my opinion, if we can keep the errors you have in #7486 internal, we could move to a long-term solution in the OTLP receiver and exporter once we've agreed on a design here. That way the functionality wouldn't change from user's perspective (with respect to error code propagation) and we wouldn't have to make any breaking changes in the API. What do you think?

jpkrohling:
Sounds good -- that error is internal already anyway, usable only by collector components in the core repository. Or do you mean making them even more private?

evan-bradley:
> that error is internal already anyway, usable only by collector components in the core repository. Or do you mean making them even more private?

I think the internal scope you have in that PR is good.

dmitryax:
Reopening so it's still on our radar

PR #7440, opened 2023-03-28
[test] Break public API and check if apidiff complains

mx-psi:
**Description:** <Describe what has changed. 
Ex. Fixing a bug - Describe the bug and how this fixes the issue.
Ex. Adding a feature - Explain what this achieves.>

This is a test to see if the "Inform incompatible PRs" check actually works. It should fail on this PR.

PR #7442, opened 2023-03-28
[test] Test apidiff on main module

mx-psi:
**Description:** 

Same as #7440, but this time on a package on the main module.


mx-psi:
Same as with #7440, this also passes when it shouldn't https://github.com/open-telemetry/opentelemetry-collector/actions/runs/4546896410/jobs/8016097952?pr=7442

PR #7447, opened 2023-03-29
[chore] Add omitempty for marshalling of logexporter

chapmanc:
**Description:** Fixing a bug: Add omitempty to the LogLevel and Verbosity fields so that marshalling them drop deprecated fields

We are importing the logexporter to build our config but leaving the loglevel empty is not removing it from the marshalling since it doesn't have omitempty. Viceversa since the code only allows either LogLevel of Verbosity it makes sense to also omitempty Verbosity as well. 

The default values should handle cases where users don't specify these values. Once LogLevel is fully removed the Verbosity omitempty could be removed if so desired. 

**Testing:** Unit testing added

### DIFF
--- a/exporter/loggingexporter/config.go
+++ b/exporter/loggingexporter/config.go
@@ -38,10 +38,10 @@ var (
 type Config struct {
 	// LogLevel defines log level of the logging exporter; options are debug, info, warn, error.
 	// Deprecated: Use `Verbosity` instead.
-	LogLevel zapcore.Level `mapstructure:"loglevel"`
+	LogLevel zapcore.Level `mapstructure:"loglevel,omitempty"`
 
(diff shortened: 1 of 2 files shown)


mx-psi (approved):

chapmanc:
@mx-psi thanks for the review. I pushed a change to re-run the changelog validation but it seems they got stuck. Are you able to bump this to get the tests to run? I don't have any access to your CI.

mx-psi:
I was able to approve it now, CI should start running :)

chapmanc:
Thank you! I assume you have a release cadence for merging these. Thanks for all of your help @mx-psi 🙇

codeboten (approved):

Issue #1084, opened 2020-06-04
Clarify memory limiter operation


tigrannajaryan:
It is not entirely clear what currently happens when memory limiter hits the limit. Do we continue receiving data and we drop it? Or we stop receiving data? Do receivers indicate to their senders to stop sending, i.e. backpressure/throttle them?

TODO:
1. We need to document the behavior clearly and make any corresponding changes needed in the memory limiter code.
2. We need to make sure all receivers consistently follow the documented behavior.
3. We ~may~ likely want to make the behavior configurable, i.e. drop vs backpressure. Depending on the use-case the desired behavior may be different.

nilebox:
Currently Memory Limiter drops the data and returns an error back to receiver: https://github.com/open-telemetry/opentelemetry-collector/blob/f134f50df819ea09b0274e5242352d46ccba8afa/processor/memorylimiter/memorylimiter.go#L166, so the action is actually specific to the receiver implementation - it may ignore the error, log it, retry etc.

We should probably at least define a public type / const for this error to be able to consistently process it everywhere.

Additionally, it may be useful to implement some callback interface to be called by MemoryLimiter or any other component to backpressure or stop the pipeline from receiving data, or perform some other action. It may even be useful for exporter receiving API throttling errors like in #1121 to stop the pipeline.

Definining such interface will allows to directly control the pipeline.

tigrannajaryan:
>     3. We may want to make the behavior configurable, i.e. drop vs backpressure. Depending on the use-case the desired behavior may be different.

@flands here is the issue I mentioned.

flands:
Definitely need 3. I also still confused why it is a processor :) you can only define once per instance right? it applies to all pipelines (or the least defined one wins for all pipelines. Then we also have: https://github.com/open-telemetry/opentelemetry-collector/issues/1078

tigrannajaryan:
It has to be a processor. Today it is the only way you can affect the pipeline (drop the data or return errors to receivers).
We can turn into a special concept that does not fit our receiver/exporter/processor/extension component model but I am not sure it is necessary.

owais:
@flands  It might not conceptually be a "processor" as it does not process data but it is part of the data pipeline. Data has to flow through it in order for it to be able to drop it. This is a very simple design that fits quite well with the rest of the system. Data goes into the memory limiter "processor" and depending on memory pressure, it'll either come out on the other side or never be seen again.

tigrannajaryan:
[...]

Issue #669, opened 2020-03-20
Add throttling / backpressure to OTLP receiver


tigrannajaryan:
OTLP must throttle the clients when it receives a backpressure from the pipeline (e.g. memory_limiter).

Must comply with OTLP spec and use gRPC status codes + RetryInfo (see the spec).

tigrannajaryan:
@pjanotti can you please add details here about how exactly receivers are informed about backpressure?

tigrannajaryan:
This is a nice-to-have, removing from 1.0.

tigrannajaryan:
We also need to test this for OTLP receiver's HTTP transport and ensure we return the correct backpressure code according to OTLP/HTTP spec.

Preferably these all should have automated tests.

punya:
Is this specific to OTLP, or does it apply to other push-based receivers too?

tigrannajaryan:
Related or duplicate of https://github.com/open-telemetry/opentelemetry-collector/issues/4335 ?

PR #7459, opened 2023-03-31
Clarify that memory limiter refuses data, doesn't drop it

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opentelemetry-collector/issues/1084

- Clarify what the memory limiter does.
- Set expectations from receivers, how they are supposed to react
  when the memory limiter refuses the data.
- Add a test that demonstrates that memory limiter does not lose data
  if the receiver and exporter behave according to the contract.

All receivers must adhere to this contract. See for example
an issue opened against filelog receiver:
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/20511

Note that there are no functional changes to the memory limiter.

Future work: one additional thing we can do is implement a backoff
logic in the memory limiter. When in memory limited mode the processor
can introduce pauses before it returns from the ConsumeLogs/Traces/Metrics
call. This will allow to slow down the inflow of data into the Collector
and give time for the pipeline to clear and memory usage to return to the
normal. This needs to be explored further.

### DIFF
--- a/processor/memorylimiterprocessor/README.md
+++ b/processor/memorylimiterprocessor/README.md
@@ -13,21 +13,32 @@ on the configured processors, it is important to put checks in place regarding
 memory usage.
  
 The memory_limiter processor allows to perform periodic checks of memory
-usage if it exceeds defined limits will begin dropping data and forcing GC to reduce
+usage if it exceeds defined limits will begin refusing data and forcing GC to reduce
 memory consumption.
 
 The memory_limiter uses soft and hard memory limits. Hard limit is always above or equal
 the soft limit.
 
-When the memory usage exceeds the soft limit the processor will start dropping the data and
-return errors to the preceding component it in the pipeline (which should be normally a
-receiver).
+When the memory usage exceeds the soft limit the processor will enter the memory limited
+mode and will start refusing the data by returning errors to the preceding component
+in the pipeline that made the ConsumeLogs/Trace/Metrics function call.
+The preceding component should be normally a receiver.
 
-When the memory usage is above the hard limit in addition to dropping the data the
+In memory limited mode the error returned by ConsumeLogs/Trace/Metrics function is a
+non-permanent error. When receivers see this error they are expected to retry sending
+the same data. The receivers may also apply a backpressure to their data sources
+in order to slow down the inflow of data into the Collector and allow the memory usage
+to go below the limits.
+
+>Warning: if the component preceding the memory limiter in the pipeline does not correctly
+retry and send the data again after ConsumeLogs/Trace/Metrics functions return then that
+data will be permanently lost. We consider such components incorrectly implemented.
+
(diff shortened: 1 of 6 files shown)


processor/memorylimiterprocessor/README.md
```diff
@@ -13,21 +13,32 @@ on the configured processors, it is important to put checks in place regarding
...
-When the memory usage is above the hard limit in addition to dropping the data the
+In memory limited mode the error returned by ConsumeLogs/Trace/Metrics function is a
+non-permanent error. When receivers see this error they are expected to retry sending
+the same data. The receivers may also apply a backpressure to their data sources
+in order to slow down the inflow of data into the Collector and allow the memory usage
+to go below the limits.
+
+>Warning: if the component preceding the memory limiter in the pipeline does not correctly
+retry and send the data again after ConsumeLogs/Trace/Metrics functions return then that
+data will be permanently lost. We consider such components incorrectly implemented.
```
dmitryax, line 35: Do you mean that receivers need to keep the data for retries? Or applying back-pressure by push-based receivers and responding with 429 applicable here as well? We don't have any receiver that can be called as properly retrying at the moment. I'm not sure we should call them incorrectly implemented until we provide best practices applied in the OTLP receiver.

tigrannajaryan: >Do you mean that receivers need to keep the data for retries? Or applying back-pressure by push-based receivers and responding with 429 applicable here as well

Both. It depends on the receiver. For example filelog receiver can simply keep and re-send the last plog.Log, while otlp receiver can drop the received request and simply signal to its client that the client must retry.

It is the receiver's responsibility to ensure the data is retried. How exactly the retrying happens, the mechanics of it are not memory limiter's concern.

>We don't have any receiver that can be called as properly retrying at the moment. I'm not sure we should call them incorrectly implemented until we provide best practices and apply them in the OTLP receiver.

Yes, the goal of this PR is to define best practices :-)

The otlp receiver is [already implemented incorrectly](https://github.com/open-telemetry/opentelemetry-collector/issues/669) regardless of this PR. It does not return the correct status codes to the client, the throttling logic is not implemented correctly.

I can delete the last sentence but it won't change the fact that (as you rightfully call out) most receivers are not correctly implemented :-)

dmitryax: Yes, I agree. That documentation should be updated if we keep this warning

dmitryax (approved):

processor/memorylimiterprocessor/README.md
tigrannajaryan: I will submit a separate PR.

dmitryax:
This PR was merged https://github.com/open-telemetry/opentelemetry-collector/commit/b1426afc1bccd45b0d171c27b3e05ec31937394e. Not sure why it's still open. Must be a GitHub glitch

Issue #7176, opened 2023-02-10
Allow using env variables in confmap.Provider URI


tigrannajaryan:
## Problem Description

It is currently impossible to use env variable in confmap.Provider.Resolve() URIs. Trying something like this:

```yaml
receivers: ${file:${HOME}myreceivers.yaml}
```

Results in an error message: `Error: failed to get config: cannot resolve the configuration: the uri "file:${HOME" contains unsupported characters ('$')
2023/02/10 16:04:08 collector server run finished with error: failed to get config: cannot resolve the configuration: the uri "file:${HOME" contains unsupported characters ('$')`

## Proposed Solution

Allowing using env variables in the URL of confmap.Provider. 

This will enable interesting use cases. For example assuming [OTELCOL_CFG_LOCATION_BASE](https://github.com/open-telemetry/opentelemetry-collector/issues/7175) proposal is implemented we can reference config files relative to each other like this.

```yaml
# this is the main config file.

# Include other config files from the same directory where the main config file is located.
receivers: ${file:${OTELCOL_CFG_LOCATION_BASE}receivers.yaml}
processors: ${file:${OTELCOL_CFG_LOCATION_BASE}processors.yaml}
exporters: ${file:${OTELCOL_CFG_LOCATION_BASE}exporters.yaml}
```

mx-psi:
I think this is a duplicate of #7117

tigrannajaryan:
https://github.com/open-telemetry/opentelemetry-collector/issues/7117 also says "For the moment probably we should document that this is not supported and forbidden". 

If that's the conclusion then let's keep this one open to be done after #7117 is closed with documenting and forbidding.

bogdandrutu:
It is still duplicate, just that my propose solution is not what you expect, which is fine and I am happy to have someone implementing that.

tigrannajaryan:
> It is still duplicate, just that my propose solution is not what you expect, which is fine and I am happy to have someone implementing that.

Sure, feel free to close this one if you think https://github.com/open-telemetry/opentelemetry-collector/issues/7117 captures all that is needed.

mx-psi:
Resolved by #7504

Issue #7117, opened 2023-02-05
Investigate to support embedded in embedded locations


bogdandrutu:
e.g. 
```yaml
test: "${http:example.com?os=${env:OS}}"
```

Right now we fail because we disallow `$` inside the location. The idea was to do this to allow us in the future to be able to extend this functionality. We need to try to see how hard will be to implement this, and if not that complicated, to add support for this. For the moment probably we should document that this is not supported and forbidden.

FYI: This is less important than all other issues in the stability milestone.

PR #7504, opened 2023-04-05
confmap: Add support for nested URIs

mackjmr:
**Description:**
This PR adds support for nesting URIs, e.g.
```
test: "${http://example.com/?os=${env:OS}}"
```
Breaking change: In the case of an InvalidScheme, error `invalid uri: "g_c_s:VALUE"` will be thrown. Previously, no error was thrown (see [test case](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.74.0/confmap/resolver_test.go#L623-L625)).

Other than the above, there will be no breaking changes. Although the parser now provides the ability to understand why parsing a URI fails (e.g. missing opening `${`), no errors will be thrown to preserve the previous behaviour.

The old way to expand env vars (e.g. `${HOST}`) is not supported in nested URIs, as expanding the old way is done in a converter. This has been documented.

**Link to tracking Issue:** #7117

**Testing:** Extensive unit test cases have been added to test the parser.


mackjmr:
cc @songy23

confmap/resolver_test.go
```diff
@@ -620,9 +742,8 @@ func TestResolverExpandInvalidScheme(t *testing.T) {
 	require.NoError(t, err)
 
 	_, err = resolver.Resolve(context.Background())
-	// TODO: Investigate how to return an error like `invalid uri: "g_c_s:VALUE"` and keep the legacy behavior
-	//  for cases like "${HOST}:${PORT}"
-	assert.NoError(t, err)
```
bogdandrutu, line 625: Please read this TODO :)

bogdandrutu, line 625: Please read this TODO :)

confmap/resolver.go
```diff
@@ -308,6 +286,90 @@ func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, erro
...
+	remaining := input[closeIndex(input)+1:]
+
+	if strings.Contains(remaining, "}") {
+		expandedRemaining, changed, err = mr.expandURIs(ctx, remaining)
+		return noExpand + expandedRemaining, changed, err
+	}
+	return input, changed, err
+}
+
+func (mr *Resolver) expandURI(ctx context.Context, value string) (string, bool, error) {
```
tigrannajaryan, line 353: Can we document the newly added functions a bit more? I find the existence of `expandURI` and `expandURIs` functions that accept and return exactly the same parameters confusing. Perhaps rename to make the purpose clearer?

Generally I find it hard to reason about how this new set of functions call recursively and how the recursion is supposed to be finite. Documentation may help with this too.

confmap/resolver.go
```diff
@@ -242,44 +239,25 @@ func (mr *Resolver) expandValueRecursively(ctx context.Context, value any) (any,
...
 	case string:
-		// If no embedded "uris" no need to expand. embeddedURI regexp matches uriRegexp as well.
-		if !embeddedURI.MatchString(v) {
+		// No URIs to expand.
+		if !strings.Contains(v, "${") || !strings.Contains(v, "}") {
 			return value, false, nil
 		}
 
+		// Don't expand. Too many closing brackets.
+		if strings.Count(v, "}") > 100 {
```
tigrannajaryan, line 248: Is this to protect from too deep recursion?

tigrannajaryan:
This PR will also resolve https://github.com/open-telemetry/opentelemetry-collector/issues/7176

confmap/resolver.go
```diff
@@ -308,6 +286,90 @@ func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, erro
...
+	return input, changed, err
+}
+
+func (mr *Resolver) expandableURI(ctx context.Context, input string, uri string) (string, bool, error) {
+	repl, changed, err := mr.expandURI(ctx, uri)
+	return strings.ReplaceAll(input, uri, repl), changed, err
+}
+
+// nonExpandableURI finds the next expandable URI in input and expands it.
+func (mr *Resolver) nonExpandableURI(ctx context.Context, input string) (string, bool, error) {
```
songy23, line 338: nit: `nonExpandableURI` sounds a bit confusing, maybe `nextExpandableURI`?

confmap/resolver_test.go
```diff
@@ -506,6 +507,119 @@ func TestResolverExpandStringValues(t *testing.T) {
 			input:  "test_${env:BOOL}",
 			output: "test_true",
 		},
+		// Nested.
```
songy23, line 510: Add a test on having `$` in the URI?

confmap/resolver.go
```diff
@@ -308,6 +286,90 @@ func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, erro
 	return value, false, nil
 }
 
+// findURI will find the URI corresponding to the first closing bracket in input.
+// findURI is only called when input contains a }.
+func (mr *Resolver) findURI(input string) (string, bool) {
```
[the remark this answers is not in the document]
bogdandrutu: Also not needed to be a func on Resolver since does not use anything from the struct.

confmap/resolver.go
```diff
@@ -308,6 +286,90 @@ func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, erro
...
+
+func closeIndex(s string) int {
+	return strings.Index(s, "}")
+}
+
+func openIndex(s string) int {
+	return strings.LastIndex(s, "${")
+}
+
+func (mr *Resolver) expandURIs(ctx context.Context, input string) (string, bool, error) {
```
bogdandrutu, line 317: Same, I feel it does not need to be a func on the struct.

bogdandrutu (commented):
To simplify the review. I would recommend to possibly extract the "expand*" funcs in a separate file with separate unit tests and good comments, so I don't need to know the internals about "expanding" when I code against confmap :)

bogdandrutu:
@mackjmr thanks for the PR, this is a great work, we just need to make sure we structure the code in a way that helps with future development since this code became a bit too complicated. Thanks a lot.

confmap/resolver.go
mackjmr: I also thought of throwing an error here, but this would be a breaking change, as previously using > 100 `}` (or > 100 embedded URIs) wouldn't throw an error. I've documented the 100 URI limit [here](https://github.com/open-telemetry/opentelemetry-collector/blob/e283dced6215def786b51fd91261e9e7a16914bf/confmap/README.md#configuration-resolving) (in limitation).

confmap/expand.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func openIndex(s string) int {
+	return strings.LastIndex(s, "${")
+}
+
+// findAndExpandURI attempts to find and expand the first occurrence of an expandable URI in input.
+func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (output string, changed bool, err error) {
+	var repl string
+	URI := findURI(input)
```
bogdandrutu, line 121: Why not `uri`?

confmap/resolver_test.go
songy23: OK, that looks good

confmap/expand.go
```diff
@@ -0,0 +1,213 @@
...
+		// Should not expand because there is a missing ${.
+		return ""
+	}
+
+	URI := input[openIndex : closeIndex+1]
+
+	if !strings.Contains(URI, ":") {
+		// Should not expand. This is expanded in the expandconverter.
+		return ""
+	}
```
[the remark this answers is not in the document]
mackjmr: > That checks for the uri structure.

Do you mean by calling function `newLocation` which uses the `uriRegexp` to check the URI structure ?

I don't believe we should check the uri structure through `uriRegexp` for this case ? The `!strings.Contains(URI, ":")` is to check if this is the old way to expand env vars (e.g. `${HOST}`). If I pass this over to function `newLocation`, then it will throw an invalid uri error, although this syntax is currently expected.

Am I misunderstanding the suggestion ?

confmap/expand.go
```diff
@@ -0,0 +1,213 @@
...
+			// No URIs to expand.
+			return value, false, nil
+		}
+
+		if strings.Count(v, "}") > 100 {
+			// Too many closing brackets. Don't expand to protect from too deep recursion.
+			return value, false, nil
+		}
+
+		URI := findURI(v)
```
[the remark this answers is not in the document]
mackjmr: My mistake, changed to lowercase in all places 👍

confmap/expand.go
```diff
@@ -0,0 +1,213 @@
...
+	return URI
+}
+
+func closeIndex(s string) int {
+	return strings.Index(s, "}")
+}
+
+func openIndex(s string) int {
+	return strings.LastIndex(s, "${")
+}
```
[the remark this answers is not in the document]
mackjmr: Removed 👍

confmap/expand.go
```diff
@@ -0,0 +1,213 @@
...
+
+func newLocation(uri string) (location, error) {
+	submatches := uriRegexp.FindStringSubmatch(uri)
+	if len(submatches) != 3 {
+		return location{}, fmt.Errorf("invalid uri: %q", uri)
+	}
+	return location{scheme: submatches[1], opaqueValue: submatches[2]}, nil
+}
+
+func (mr *Resolver) retrieveValue(ctx context.Context, uri location) (*Retrieved, error) {
```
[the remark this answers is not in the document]
mackjmr: Moved back to resolver file.

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, error) {
+	switch v := value.(type) {
+	case string:
+		if !strings.Contains(v, "${") || !strings.Contains(v, "}") {
+			// No URIs to expand.
+			return value, false, nil
+		}
+
+		if strings.Count(v, "}") > 100 {
```
bogdandrutu, line 47: What do we protect against? Also, return error to signal to the user that something did not work.

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+	if strings.Contains(remaining, "}") {
+		expandedRemaining, changed, err = mr.findAndExpandURI(ctx, remaining)
+		return noExpand + expandedRemaining, changed, err
+	}
+	return input, changed, err
+}
+
+// expandURI tries to expand uri. If an expandable URI is found, it returns the uri expanded, true and nil.
+// Otherwise, it returns the unchanged input, false and the expanding error.
+func (mr *Resolver) expandURI(ctx context.Context, uri string) (string, bool, error) {
```
bogdandrutu, line 145: s/expandURI/expandURIAsString

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+	case reflect.Float32, reflect.Float64:
+		return strconv.FormatFloat(val.Float(), 'f', -1, 64), changed, err
+	case reflect.Bool:
+		return strconv.FormatBool(val.Bool()), changed, err
+	default:
+		return uri, changed, fmt.Errorf("expanding %v, expected string value type, got %T", uri, expanded)
+	}
+}
+
+func (mr *Resolver) expandStringURI(ctx context.Context, uri string) (any, bool, error) {
```
bogdandrutu, line 165: s/expandStringURI/expandURI

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+	return uri
+}
+
+// findAndExpandURI attempts to find and expand the first occurrence of an expandable URI in input.
+func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (output string, changed bool, err error) {
+	var repl string
+	uri := findURI(input)
+	if uri != "" {
+		repl, changed, err = mr.expandURI(ctx, uri)
+		input = strings.ReplaceAll(input, uri, repl)
```
bogdandrutu, line 116: You should just return instead of having an else.

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+	} else {
+		// Check if the next URI in input is expandable.
+		input, changed, err = mr.nextExpandableURI(ctx, input)
```
bogdandrutu, line 119: merge `nextExpandableURI` into this func, hard to have recursion split across multiple funcs, it is much clear to see that the func calls itself.

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+	scheme      string
+	opaqueValue string
+}
+
+func (c location) asString() string {
+	return c.scheme + ":" + c.opaqueValue
+}
+
+func newLocation(uri string) (location, error) {
+	submatches := uriRegexp.FindStringSubmatch(uri)
```
bogdandrutu, line 192: I think uriRegexp is only used in this file, so move it here.

confmap/expand.go
```diff
@@ -0,0 +1,197 @@
...
+func (mr *Resolver) nextExpandableURI(ctx context.Context, input string) (string, bool, error) {
+	var err error
+	var expandedRemaining string
+	var changed bool
+
+	closeIndex := strings.Index(input, "}")
+	noExpand := input[:closeIndex+1]
+	remaining := input[closeIndex+1:]
+
+	// if remaining does not contain }, there are no URIs left: stop recursion.
```
bogdandrutu, line 135: This is not what the code does :) it is opposite, but I would like the code to be like this:

 ```golang
	if !strings.Contains(remaining, "}") {
		return input, changed, err
	}
	expandedRemaining, changed, err = mr.findAndExpandURI(ctx, remaining)
	return noExpand + expandedRemaining, changed, err
```

File: confmap/expand.go [hunk 20]
```diff
@@ -0,0 +1,197 @@
...
+
+// findAndExpandURI attempts to find and expand the first occurrence of an expandable URI in input.
+func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (output string, changed bool, err error) {
+	var repl string
+	uri := findURI(input)
+	if uri != "" {
+		repl, changed, err = mr.expandURI(ctx, uri)
+		input = strings.ReplaceAll(input, uri, repl)
+	} else {
+		// Check if the next URI in input is expandable.
```
bogdandrutu, line 118: A bit confused about this comment. If you did not find any uri (since uri == "") how could you find a next one? I think this code works because of the recursion in `expandValueRecursively` and this branch is a no-op always. If that is the case then you can remove it and also the `nextExpandableURI`.

confmap/expand.go
mackjmr: Swapped the function names 👍

confmap/expand.go
mackjmr: Agreed -- merged the two together, the func calls itself now 👍

confmap/expand.go
mackjmr: Also moved other vars which were only used in this file.

confmap/expand.go
mackjmr: Yeah, good catch :) Made the code as suggested 👍

confmap/resolver.go
```diff
@@ -225,123 +208,6 @@ func (mr *Resolver) closeIfNeeded(ctx context.Context) error {
...
-		var nerr error
-		var nchanged bool
-		nv := embeddedURI.ReplaceAllStringFunc(v, func(s string) string {
-			ret, changed, err := mr.expandStringURI(ctx, s)
-			nchanged = nchanged || changed
-			nerr = multierr.Append(nerr, err)
-			if err != nil {
-				return ""
-			}
-			// This list must be kept in sync with checkRawConfType.
```
bogdandrutu, line 266: This comment should be preserved.

confmap/expand.go
```diff
@@ -0,0 +1,206 @@
...
+		return noExpand + expandedRemaining, changed, err
+	}
+	repl, changed, err = mr.expandStringURI(ctx, uri)
+	input = strings.ReplaceAll(input, uri, repl)
+	return input, changed, err
+}
+
+// expandStringURI tries to expand uri as a string. If an expandable URI is found and it can be converted to a string,
+// it returns the expanded uris string value, true and nil. Otherwise, it returns the unchanged input, false and the expanding error.
+func (mr *Resolver) expandStringURI(ctx context.Context, uri string) (string, bool, error) {
```
bogdandrutu, line 154: If I see correctly, only one caller of this. Consider to make this just a "toString" func and move the call to `expandURI` to the caller func.

confmap/expand.go
```diff
@@ -0,0 +1,206 @@
...
+	var repl string
+	var expandedRemaining string
+
+	uri := findURI(input)
+	if uri == "" {
+		// The first URI in input is not expandable. Strip the first URI from input and check if
+		// other URIs are expandable.
+		closeIndex := strings.Index(input, "}")
+		noExpand := input[:closeIndex+1]
+		remaining := input[closeIndex+1:]
```
bogdandrutu, line 137: This code is duplicate with findURI. Let's do your suggestion to change findURI to be recursive.

confmap/expand.go
bogdandrutu: I think this is artificial, we should not care because it can be a value like `${}}}...}` (101 `}`) but only one URI. The other limit 100 that you see in `expandValueRecursively` is to avoid deadlocks not to avoid "too deep" recursions.

confmap/expand.go
```diff
@@ -0,0 +1,206 @@
...
+			return value, false, nil
+		}
+
+		if strings.Count(v, "}") > 100 {
+			// Too many closing brackets. Don't expand to protect from too deep recursion.
+			return "", false, errURILimit
+		}
+
+		uri := findURI(v)
+		if uri != "" && uri == value {
```
bogdandrutu, line 69: This logic can be moved to `findAndExpandURI` to avoid duplicate calls to findURI.

confmap/expand.go
mackjmr: Changed to `toString` func.

confmap/expand.go
mackjmr: Moved to avoid duplicate calls.

confmap/expand.go
mackjmr: Moving to resolved as recursion was moved to `findURI`.

confmap/expand.go
```diff
@@ -0,0 +1,194 @@
...
+	case reflect.String:
+		return val.String(), nil
+	case reflect.Int, reflect.Int32, reflect.Int64:
+		return strconv.FormatInt(val.Int(), 10), nil
+	case reflect.Float32, reflect.Float64:
+		return strconv.FormatFloat(val.Float(), 'f', -1, 64), nil
+	case reflect.Bool:
+		return strconv.FormatBool(val.Bool()), nil
+	default:
+		return "", fmt.Errorf("expanding %v, expected string value type, got %T", strURI, input)
```
bogdandrutu, line 158:
```suggestion
		return "", fmt.Errorf("expanding %v, expected convertable to string value type, got %q(%T)", strURI, input, input)
```

confmap/expand.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return input[openIndex : closeIndex+1]
+}
+
+// findAndExpandURI attempts to find and expand the first occurrence of an expandable URI in input. If an expandable URI is found it
+// returns the input with the URI expanded, true and nil. Otherwise, it returns the unchanged input, false and the expanding error.
+func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (output any, changed bool, err error) {
+	var expanded any
+	var repl string
+
```
bogdandrutu, line 123: Remove these declared vars with `expanded, changed, err := mr.expandURI(ctx, uri)` on line 133. Or move the declaration closer to that place if you want to keep it this way. It is a best practice to declare vars close to where you use them.

confmap/expand_test.go
```diff
@@ -0,0 +1,476 @@
...
+					return NewRetrieved(64)
+				case "env:FLOAT32":
+					return NewRetrieved(float32(3.25))
+				case "env:FLOAT64":
+					return NewRetrieved(float64(6.4))
+				case "env:BOOL":
+					return NewRetrieved(true)
+				}
+				return nil, errors.New("impossible")
+			})
```
bogdandrutu, line 325: Consider to move this as a separate func to simplify the reading of the test case.

confmap/expand.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	expanded, changed, err = mr.expandURI(ctx, uri)
+	if err != nil {
+		return input, false, err
+	}
+	repl, err = toString(uri, expanded)
+	if err != nil {
+		return input, false, err
+	}
+	return strings.ReplaceAll(input, uri, repl), changed, err
```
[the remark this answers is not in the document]
mackjmr: Added a test case.

confmap/expand.go
mackjmr: Removed 👍

confmap/expand.go
mackjmr: Removed the declared vars ✅

confmap/expand_test.go
```diff
@@ -0,0 +1,476 @@
...
+
+func TestResolverExpandStringValues(t *testing.T) {
+	tests := []struct {
+		name   string
+		input  string
+		output any
+	}{
+		// Embedded.
+		{
+			name:   "test_no_match_old",
```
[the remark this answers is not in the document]
mackjmr: Removed "test_" prefix and re-worked the test names 👍

confmap/expand.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		return findURI(remaining)
+	}
+
+	return input[openIndex : closeIndex+1]
+}
+
+// findAndExpandURI attempts to find and expand the first occurrence of an expandable URI in input. If an expandable URI is found it
+// returns the input with the URI expanded, true and nil. Otherwise, it returns the unchanged input, false and the expanding error.
+func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (output any, changed bool, err error) {
```
bogdandrutu, line 112:
```suggestion
func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (any, bool, error) {
```

confmap/expand.go
```diff
@@ -0,0 +1,183 @@
...
+	uri := findURI(input)
+	if uri == "" {
+		return input, false, err
```
bogdandrutu, line 115:
```suggestion
		// No URI found, return.
		return input, false, err
```

bogdandrutu (approved):

songy23 (approved):
LGTM

Issue #7508, opened 2023-04-05
Batch processor: support zero timeout (split only)


jmacd:
**Is your feature request related to a problem? Please describe.**
I have an application where it would be nice to limit requests without introducing delay. The batchprocessor contains exactly the code necessary, but the configuration with zero timeout and/or zero SendBatchSize does not work correctly. 

**Describe the solution you'd like**
It is relatively easy to fix this. When `timeout` is zero, no timer is needed and all requests are sent immediately. In this configuration, only the `send_batch_max_size` configuration matters and (I propose) `send_batch_size` should be ignored.

**Describe alternatives you've considered**
Make it an error if `send_batch_size != 1` when `timeout == 0`, since if the timeout is zero there the minimum batch size is irrelevant.

I think we should not send an empty batch just because `send_batch_size` is set to zero.

dmitryax:
Resolved by https://github.com/open-telemetry/opentelemetry-collector/pull/7509

PR #7509, opened 2023-04-06
Batch processor support for split-only configuration

jmacd:
**Description:** Implement the fixes proposed in #7508. Otherwise, using a `batchprocessor` with `Config{}` passes validation but acts badly at runtime there is both a zero-duration timer and the call to `sendItems` never returns, constantly sending empty batches.

**Link to tracking Issue:** #7508.

**Testing:** Two new tests.

**Documentation:** Updated.

### DIFF
--- a/processor/batchprocessor/batch_processor.go
+++ b/processor/batchprocessor/batch_processor.go
@@ -113,8 +113,16 @@ func (bp *batchProcessor) Shutdown(context.Context) error {
 
 func (bp *batchProcessor) startProcessingCycle() {
 	defer bp.goroutines.Done()
-	bp.timer = time.NewTimer(bp.timeout)
+	if bp.timeout != 0 && bp.sendBatchSize != 0 {
+		bp.timer = time.NewTimer(bp.timeout)
+	}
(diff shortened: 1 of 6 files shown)


processor/batchprocessor/batch_processor.go
```diff
@@ -113,8 +113,16 @@ func (bp *batchProcessor) Shutdown(context.Context) error {
...
+	if bp.timeout != 0 && bp.sendBatchSize != 0 {
+		bp.timer = time.NewTimer(bp.timeout)
+	}
 	for {
+		// timerCh ensures we only block when there is a
+		// timer, since <- from a nil channel is blocking.
+		var timerCh <-chan time.Time
+		if bp.timer != nil {
+			timerCh = bp.timer.C
+		}
```
dmitryax, line 125: Shouldn't this be done outside the for loop close where the `bp.timer` is initialized?

jmacd: Done.

dmitryax (approved):

PR #7510, opened 2023-04-06
Initial support for server interceptors

jpkrohling:
This PR adds support to interceptors, to be used on HTTP clients and servers.

We talked during the SIG meeting that I would implement server interceptors first, but given that the same interface would apply to clients, I think it was worth implementing both on this first PR.


jpkrohling:
cc @sh0rez

Issue #7478, opened 2023-04-03
Add test helper to verify receiver behavior on errors


tigrannajaryan:
The receivers are required to follow the [Error Handling contract](https://github.com/open-telemetry/opentelemetry-collector/blob/1c47d89fc38df12133adc78c4ea7bc4d3374e6b6/receiver/doc.go#L35) which says:

```
// If the error is non-Permanent then the nextConsumer.Consume*() call should be retried
// with the same data.
```

and 

```
// If the error is Permanent, then the nextConsumer.Consume*() call should not be
// retried with the same data.
```

We need a test helper that makes testing for this part of the contract easy.

## Helper Inputs

The helper should accept:
- A receiver factory.
- A config for the receiver. We may want to make it easy to allocate a listening port for http receivers.
- A request generator. E.g. for `otlp` receiver testing we need `otlp` request generator that behaves correctly according to OTLP protocol spec and retries the requests when instructed so by the receiver.

## Events to Test for

It is important that all 3 event types happen during the test at least once:
- ConsumeLogs/Traces/Metrics call succeeds at the first try.
- ConsumeLogs/Traces/Metrics call fails once with non-permanent error, then succeeds on the next try.
- ConsumeLogs/Traces/Metrics call fails more than once with non-permanent error, then eventually succeeds.
- ConsumeLogs/Traces/Metrics call fails once with permanent error.

The test helper should provide a mock consumer that can accept data and return permanent and non-permanent errors (or success) as necessary. 

## Test Scenarios

The helper should be possible to configure for testing for permanent errors scenario and must verify:
- No data is lost if ConsumeLogs/Traces/Metrics call succeeds.
- No data is duplicated.
- Data is not retried when permanent error is returned.
 
The helper should be possible to configure for testing for non-permanent errors scenario and must verify:
- No data is lost.
- No data is duplicated.
- Generated and delivered data content match exactly (but requests may be re-ordered).

## Other Requirements

- The test helper should support concurrent calls from the receiver.
- The helper should support all 3 signal types: logs, traces, metrics.

PR #7516, opened 2023-04-11
Add receiver/consumer contract test helper

tigrannajaryan:
- Added a CheckConsumeContract() helper func that can be used to test the contract that receivers are expected to maintain for Consume() calls.
- Added an example usage of CheckConsumeContract() helper.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/7478

### DIFF
--- a/receiver/receivertest/contract_checker.go
+++ b/receiver/receivertest/contract_checker.go
@@ -33,13 +33,19 @@ const UniqueIDAttrName = "test_id"
 type UniqueIDAttrVal = string
 
 type Generator interface {
+	// Start the generator and prepare to generate. Will be followed by calls to Generate().
+	// Start() may be called again after Stop() is called to begin a new test scenario.
 	Start()
+
+	// Stop generating. There will be no more calls to Generate() until Start() is called again.
+	Stop()
 
 	// Generate must generate and send at least one data element (span, log record or metric data point)
 	// to the receiver and return a copy of generated element ids.
 	// The generated data must contain uniquely identifiable elements, each with a
 	// different value of attribute named UniqueIDAttrName.
 	// CreateOneLogWithID() can be used a helper to create such logs.
+	// May be called concurrently from multiple goroutines.
 	Generate() []UniqueIDAttrVal
 }
 
@@ -53,7 +59,8 @@ type CheckConsumeContractParams struct {
 	Config component.Config
 	// Generator that can send data to the receiver.
 	Generator Generator
-	// GenerateCount specifies the number of times to call the generator.Generate().
+	// GenerateCount specifies the number of times to call the generator.Generate()
+	// for each test scenario.
 	GenerateCount int
 }
 
@@ -66,12 +73,12 @@ func CheckConsumeContract(params CheckConsumeContractParams) {
 	// success case or for error case or a mix of both). See for example randomErrorsConsumeDecision.
 	scenarios := []struct {
 		name         string
-		decisionFunc func(ids IDSet) error
+		decisionFunc func(ids idSet) error
 	}{
 		{
 			name: "always_succeed",
 			// Always succeed. We expect all data to be delivered as is.
-			decisionFunc: func(ids IDSet) error { return nil },
+			decisionFunc: func(ids idSet) error { return nil },
 		},
 		{
 			name:         "random_non_permanent_error",
@@ -96,7 +103,7 @@ func CheckConsumeContract(params CheckConsumeContractParams) {
 	}
 }
 
-func checkConsumeContractScenario(params CheckConsumeContractParams, decisionFunc func(ids IDSet) error) {
+func checkConsumeContractScenario(params CheckConsumeContractParams, decisionFunc func(ids idSet) error) {
 	consumer := &mockConsumer{t: params.T, consumeDecisionFunc: decisionFunc}
 	ctx := context.Background()
 
@@ -122,17 +129,17 @@ func checkConsumeContractScenario(params CheckConsumeContractParams, decisionFun
 
 	// Begin generating data to the receiver.
 
-	var generatedIds IDSet
+	var generatedIds idSet
 	var generatedIndex int64
 	var mux sync.Mutex
 	var wg sync.WaitGroup
 
 	const concurrency = 4
 
+	params.Generator.Start()
+
 	// Create concurrent goroutines that use the generator.
 	// The total number of generator calls will be equal to params.GenerateCount.
-
-	params.Generator.Start()
 
 	for j := 0; j < concurrency; j++ {
 		wg.Add(1)
@@ -156,10 +163,12 @@ func checkConsumeContractScenario(params CheckConsumeContractParams, decisionFun
 	// Wait until all generator goroutines are done.
 	wg.Wait()
 
+	params.Generator.Stop()
+
 	// Wait until all data is seen by the consumer.
 	assert.Eventually(params.T, func() bool {
 		// Calculate the union of accepted and dropped data.
-		acceptedAndDropped, duplicates := consumer.acceptedIds.union(consumer.droppedIds)
+		acceptedAndDropped, duplicates := consumer.acceptedAndDropped()
 		if len(duplicates) != 0 {
 			assert.Failf(params.T, "found duplicate elements in received and dropped data", "keys=%v", duplicates)
 		}
@@ -169,7 +178,7 @@ func checkConsumeContractScenario(params CheckConsumeContractParams, decisionFun
 	}, 5*time.Second, 10*time.Millisecond)
 
 	// Do some final checks. Need the union of accepted and dropped data again.
-	acceptedAndDropped, duplicates := consumer.acceptedIds.union(consumer.droppedIds)
+	acceptedAndDropped, duplicates := consumer.acceptedAndDropped()
 	if len(duplicates) != 0 {
 		assert.Failf(params.T, "found duplicate elements in accepted and dropped data", "keys=%v", duplicates)
 	}
@@ -197,11 +206,11 @@ func checkConsumeContractScenario(params CheckConsumeContractParams, decisionFun
 	)
 }
 
-// IDSet is a set of unique ids of data elements used in the test (logs, spans or metric data points).
-type IDSet map[UniqueIDAttrVal]bool
+// idSet is a set of unique ids of data elements used in the test (logs, spans or metric data points).
+type idSet map[UniqueIDAttrVal]bool
 
 // compare to another set and calculate the differences from this set.
-func (ds IDSet) compare(other IDSet) (missingInOther, onlyInOther []UniqueIDAttrVal) {
+func (ds idSet) compare(other idSet) (missingInOther, onlyInOther []UniqueIDAttrVal) {
 	for k := range ds {
 		if _, ok := other[k]; !ok {
 			missingInOther = append(missingInOther, k)
@@ -216,7 +225,7 @@ func (ds IDSet) compare(other IDSet) (missingInOther, onlyInOther []UniqueIDAttr
 }
 
 // merge another set into this one and return a list of duplicate ids.
-func (ds *IDSet) merge(other IDSet) (duplicates []UniqueIDAttrVal) {
+func (ds *idSet) merge(other idSet) (duplicates []UniqueIDAttrVal) {
 	if *ds == nil {
 		*ds = map[UniqueIDAttrVal]bool{}
 	}
@@ -231,7 +240,7 @@ func (ds *IDSet) merge(other IDSet) (duplicates []UniqueIDAttrVal) {
 }
 
 // merge another set into this one and return a list of duplicate ids.
-func (ds *IDSet) mergeSlice(other []UniqueIDAttrVal) (duplicates []UniqueIDAttrVal) {
+func (ds *idSet) mergeSlice(other []UniqueIDAttrVal) (duplicates []UniqueIDAttrVal) {
 	if *ds == nil {
 		*ds = map[UniqueIDAttrVal]bool{}
 	}
@@ -247,7 +256,7 @@ func (ds *IDSet) mergeSlice(other []UniqueIDAttrVal) (duplicates []UniqueIDAttrV
 
 // union computes the union of this and another sets. A new set if created to return the result.
 // Also returns a list of any duplicate ids found.
-func (ds *IDSet) union(other IDSet) (union IDSet, duplicates []UniqueIDAttrVal) {
+func (ds *idSet) union(other idSet) (union idSet, duplicates []UniqueIDAttrVal) {
 	union = map[UniqueIDAttrVal]bool{}
 	for k, v := range *ds {
 		union[k] = v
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
Let me know if a changelog entry is needed for this.

tigrannajaryan:
@open-telemetry/collector-approvers please review.

tigrannajaryan:
@open-telemetry/collector-approvers please review.

tigrannajaryan:
https://github.com/orgs/open-telemetry/teams/collector-approvers please review.

tigrannajaryan:
@open-telemetry/collector-approvers please review.

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,382 @@
...
+		len(consumer.acceptedIds.m),
+		len(consumer.droppedIds.m),
+		consumer.nonPermanentFailures,
+	)
+}
+
+// IDSet is a set of unique ids of data elements used in the test (logs, spans or metric data points).
+type IDSet struct {
+	m map[UniqueIDAttrDataType]bool
+}
```
djaglowski, line 168: nit: Can this just be an alias?
```suggestion
type IDSet map[UniqueIDAttrDataType]bool
```

tigrannajaryan:
> This generally looks good to me. In addition to adding support for metrics and filling in test coverage somewhat, do you think it would make sense that this PR should include a real usage of this functionality, perhaps in testing the otlp receiver?

I was trying to keep the PR small. Definitely planning to add an otlp receiver test in a future PR, but can add to this one if that's what we want to see.

djaglowski:
It's fine with me to limit this to only the receivertest package.

tigrannajaryan:
> It's fine with me to limit this to only the receivertest package.

There is obviously the risk that the receivertest may need modifications to work for otlp receiver. I tried to foresee how it should work but no guarantees.

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,382 @@
...
+
+// Compare to another set and calculate the differences from this set.
+func (ds *IDSet) Compare(other IDSet) (missingInOther, onlyInOther []UniqueIDAttrDataType) {
```
dmitryax, line 171: Can we reduce the exposed API for now while this functionality is not stabilized? We have a policy that any change of the public API has to go through a deprecation period https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes. For example, I'm not sure why `IDSet` methods need to be exposed to the end user.

receiver/receivertest/contract_checker_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Different scenarios to test for.
+	decisionFuncs := []func(ids IDSet) error{
+		// Always succeed. We expect all data to be delivered as is.
+		func(ids IDSet) error { return nil },
+
+		// Various scenarios with errors injected into the consumer's decision making.
+		RandomNonPermanentErrorConsumeDecision,
+		RandomPermanentErrorConsumeDecision,
+		RandomErrorsConsumeDecision,
```
dmitryax, line 127: Do you think users will necessarily need to add custom decision-making functions? If not, maybe we can unexport these functions for now and run them all inside `CheckConsumeContract`?

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,382 @@
...
+// CheckConsumeContract checks the contract between the receiver and its next consumer. For the contract
+// description see ../doc.go. The checker will detect violations of contract on success, on permanent
+// and non-permanent errors based on decision-making done by consumeDecisionFunc.
+// It is advised to run CheckConsumeContract with a variety of decision-making functions.
+func CheckConsumeContract(params CheckConsumeContractParams) {
+	consumer := &mockConsumer{t: params.T, consumeDecisionFunc: params.ConsumeDecisionFunc}
+	ctx := context.Background()
+
+	// Create and start the receiver.
+	receiver, err := params.Factory.CreateLogsReceiver(ctx, NewNopCreateSettings(), params.Config, consumer)
```
dmitryax, line 80: This `CheckConsumeContract` function is applied to logs only, so we probably need to rename it or specify a signal type in `CheckConsumeContractParams`?

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,382 @@
...
+		// Compare accepted+dropped with generated. Once they are equal it means all data is seen by the consumer.
+		missingInOther, onlyInOther := generatedIds.Compare(acceptedAndDropped)
+		return len(missingInOther) == 0 && len(onlyInOther) == 0
+	}, 5*time.Second, 10*time.Millisecond)
+
+	// Do some final checks. Need the union of accepted and dropped data again.
+	acceptedAndDropped, duplicates := consumer.acceptedIds.Union(consumer.droppedIds)
+	if len(duplicates) != 0 {
+		assert.Failf(params.T, "found duplicate elements in accepted and dropped data", "keys=%v", duplicates)
+	}
```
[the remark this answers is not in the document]
tigrannajaryan: This is just to print a nicer error message explaining what exactly failed. The "eventually" function above may fail for a few different reasons. I think it is useful to print the exact reason why it failed.

receiver/receivertest/contract_checker_test.go
tigrannajaryan: Moved scenarios inside `CheckConsumeContract`. The decision funcs are no longer exported.

receiver/receivertest/contract_checker.go
tigrannajaryan: Done.

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,382 @@
...
+
+	for j := 0; j < concurrency; j++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				if atomic.AddInt64(&generatedIndex, 1) >= int64(params.GenerateCount) {
+					// Generated as many as was requested. We are done.
+					return
+				}
```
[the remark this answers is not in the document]
tigrannajaryan: Done.

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,405 @@
...
+		ils := rss.At(i).ScopeSpans()
+		for j := 0; j < ils.Len(); j++ {
+			ss := ils.At(j).Spans()
+			for k := 0; k < ss.Len(); k++ {
+				elem := ss.At(k)
+				key, exists := elem.Attributes().Get(UniqueIDAttrName)
+				if !exists {
+					return ds, fmt.Errorf("invalid data element, attribute %q is missing", UniqueIDAttrName)
+				}
+				if key.Type() != pcommon.ValueTypeInt {
```
dmitryax, line 326: I'm not sure that all the receivers can produce integer attributes. Maybe we should use string representation for `UniqueIDAttrDataType`?

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,405 @@
...
+	// Generator that can send data to the receiver.
+	Generator Generator
+	// GenerateCount specifies the number of times to call the generator.Generate().
+	GenerateCount int
+}
+
+// CheckConsumeContract checks the contract between the receiver and its next consumer. For the contract
+// description see ../doc.go. The checker will detect violations of contract on different scenarios: on success,
+// on permanent and non-permanent errors and mix of error types.
+func CheckConsumeContract(params CheckConsumeContractParams) {
```
dmitryax, line 61: I tried this on filelog receiver in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/22833 and run into problems with reusing the same generator and config across multiple tests.
I fixed it by changing this signature to access a function returning `CheckConsumeContractParams`, but I don't think it's the best idea. Another, maybe a better, option is to accept a generator factory instead of the generator itself. Then, I think we need an additional `Stop()` method on it.

Otherwise, it worked pretty well. Thank you for adding these tests!

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,405 @@
...
+// UniqueIDAttrDataType is the value type of the UniqueIDAttrName.
+type UniqueIDAttrDataType int64
+
+type Generator interface {
+	// Generate must generate and send at least one data element (span, log record or metric data point)
+	// to the receiver and return a copy of generated element ids.
+	// The generated data must contain uniquely identifiable elements, each with a
+	// different value of attribute named UniqueIDAttrName.
+	// CreateOneLogWithID() can be used a helper to create such logs.
+	Generate() IDSet
```
dmitryax, line 41: If we change this signature to `Generate() []UniqueIDAttrDataType`, we don't need to expose the `IDSet` anymore, which seems not very useful to the end user at this point. WDYT?

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,405 @@
...
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.opentelemetry.io/collector/receiver"
+)
+
+// UniqueIDAttrName is the attribute name that is used in log records/spans/datapoints as the unique identifier.
+const UniqueIDAttrName = "test_id"
+
+// UniqueIDAttrDataType is the value type of the UniqueIDAttrName.
+type UniqueIDAttrDataType int64
```
dmitryax, line 33: Maybe `UniqueIDAttrVal` given that it's a type itself?

receiver/receivertest/contract_checker.go
dmitryax: Maybe `Generator.Generate` and `IDSetFromLogs` (or `IDsFromLogs`) can work on `[]UniqueIDAttrDataType` instead, so we don't need to expose the IDSet. The slices seem to be more user-friendly. WDYT?

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,405 @@
...
+	const concurrency = 4
+
+	// Create concurrent goroutines that use the generator.
+	// The total number of generator calls will be equal to params.GenerateCount.
+
+	for j := 0; j < concurrency; j++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for atomic.AddInt64(&generatedIndex, 1) < int64(params.GenerateCount) {
```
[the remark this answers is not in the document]
tigrannajaryan: Good catch. AddInt64 returns the value _after_ incrementing.

receiver/receivertest/contract_checker.go
tigrannajaryan: Can you clarify what the problem is with reusing the config? Do you need to have different configs for different subtests? If you don't need different configs we can add Start() method to Generator, where you can re-create the test input file.

tigrannajaryan: I added Start() to Generator. See if it helps the filelog use case.

receiver/receivertest/contract_checker.go
tigrannajaryan: Changed to string.

receiver/receivertest/contract_checker.go
tigrannajaryan: Changed to a slice.

receiver/receivertest/contract_checker.go
dmitryax: The problem is we create a new receiver for every subtest, in which I need to set up a file to send the logs through. I have two options:
1. The first option is to use a single file, pass it to the filelog configuration, and append data to the file for each subtest on top of the existing data. This approach would be feasible if the filelog receiver could be configured to read from the end of the same file in every test case. However, the filelog receiver begins polling files only after [a certain interval](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3fb4d9963cc5e8f01c0f7c6894b1248f48162396/pkg/stanza/fileconsumer/file.go#L87-L95), by which time all the logs have already been written by the generator.
2. The second option is to generate a new file for each subtest in the `Generator.Start()` method and set up the filelog receiver to read from the beginning of all files that match a specific file pattern. This option would work, but I need to be able to remove the file after each subtest is finished. Otherwise, subsequent subtests would end up reading data from files created for previous subtests.

I would go with the (2), but I need `Generator.Stop()` or `Shutdown()` method additionally to `Start()`.

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,424 @@
...
+			m.nonPermanentFailures++
+		}
+		// Return the error to the receiver.
+		return err
+	}
+
+	// The decision is a success. Remember the ids of the data in the accepted list.
+	duplicates := m.acceptedIds.merge(ids)
+	if len(duplicates) > 0 {
+		require.FailNow(m.t, "elements that were accepted previously were sent again: %v", duplicates)
```
dmitryax, line 415: Should it be `require.FailNowf`?

Or we can have instead of the condition:

```
require.Emptyf(m.t, duplicates, "elements that were accepted previously were sent again: %v", duplicates)
```

I believe it'll be the same effect

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,424 @@
...
+}
+
+func (m *mockConsumer) ConsumeTraces(_ context.Context, data ptrace.Traces) error {
+	ids, err := IDSetFromTraces(data)
+	require.NoError(m.t, err)
+	return m.consume(ids)
+}
+
+// IDSetFromTraces computes an IDSet from given ptrace.Traces. The IDSet will contain ids of all spans.
+func IDSetFromTraces(data ptrace.Traces) (IDSet, error) {
```
dmitryax, line 332: I think we don't need to expose this function along with `IDSet` anymore

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,424 @@
...
+		}()
+	}
+
+	// Wait until all generator goroutines are done.
+	wg.Wait()
+
+	// Wait until all data is seen by the consumer.
+	assert.Eventually(params.T, func() bool {
+		// Calculate the union of accepted and dropped data.
+		acceptedAndDropped, duplicates := consumer.acceptedIds.union(consumer.droppedIds)
```
dmitryax, line 162: I think we can run into a read/write race condition here. `consumer.mux` should probably be used here. Or maybe better to use `sync.Map` everywhere?

receiver/receivertest/contract_checker.go
tigrannajaryan: Sounds good, I think it is reasonable for Generator to have a Start/Stop. I will add.

receiver/receivertest/contract_checker.go
tigrannajaryan: Good catch. This can indeed race. The only reason it didn't is because my test receiver doesn't do async consumption. Will fix.

receiver/receivertest/contract_checker.go
tigrannajaryan: Added Stop().

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,437 @@
...
+				// generated data set.
+				require.Empty(params.T, duplicates)
+			}
+		}()
+	}
+
+	// Wait until all generator goroutines are done.
+	wg.Wait()
+
+	params.Generator.Stop()
```
dmitryax, line 166: I think this should be put after the next `assert.Eventually` block. In the case of filelog receiver, we can remove the logs file only after the receiver has received all the data.

receiver/receivertest/contract_checker.go
```diff
@@ -0,0 +1,437 @@
...
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.opentelemetry.io/collector/receiver"
+)
+
+// UniqueIDAttrName is the attribute name that is used in log records/spans/datapoints as the unique identifier.
+const UniqueIDAttrName = "test_id"
+
+// UniqueIDAttrVal is the value type of the UniqueIDAttrName.
+type UniqueIDAttrVal = string
```
dmitryax, line 33: Should we apply the type definition to enforce it? Otherwise just `string` can be used instead

```suggestion
type UniqueIDAttrVal string
```

tigrannajaryan: I went back and forth on this, chose an alias to have less typing in the Generator, but no strong opinion. If you feel a stronger type could help we can change to that.

receiver/receivertest/contract_checker.go
tigrannajaryan: OK, moved to `defer` so that it is done at the very end.

receiver/receivertest/contract_checker.go
tigrannajaryan: Changed.

dmitryax (approved):
Thank you!

PR #7526, opened 2023-04-12
[chore] Add govulncheck scan (as a new job in build-and-test workflow)

lootek:
**Description:**
Enable [govulncheck](https://go.dev/blog/vuln) linter

It's a great tool, a part of the official Go toolset. As they state:
> Go provides tooling to analyze your codebase and surface known vulnerabilities. This tooling is backed by the Go vulnerability database, which is curated by the Go security team. Go’s tooling reduces noise in your results by only surfacing vulnerabilities in functions that your code is actually calling.

Also, please note it's NOT a part of golangci-lint (yet): https://github.com/golangci/golangci-lint/issues/3094.

**Inspired by/relates to:**
* https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/19154
* https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20342#issuecomment-1504689049
* https://github.com/open-telemetry/opentelemetry-collector-releases/pull/319

**Testing:**
We use Go 1.19.6 here and since there were 2 patch releases recently that addressed several CVEs found the new job is expected to fail (if only we really *use* that vulnerable code here) until I add a next commit that'll bump Go to 1.19.8.


lootek:
cc @dmitryax

lootek:
Can someone approve the updated workflow please? I'd like to see the newly added job to fail (due to the outdated Go SDK being used) to get a proof my addition works. Then I'll bump Go to make it pass.

codeboten (commented):
@lootek it looks like the check failed as expected 👍🏻

.github/workflows/build-and-test.yml
```diff
@@ -55,6 +55,30 @@ jobs:
       - name: goimpi
         run: make goimpi
 
+  govulncheck:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
```
TylerHelmuth, line 60: Can you bump this up to 30 mins?  We have this workflow in Contrib and if it ever can't grab the cache correctly and has to redownload everything it can run long.

lootek: changed

dmitryax (approved):

Issue #6599, opened 2022-11-22
Make service.instance.id and other telemetry attributes available to extensions


tigrannajaryan:
OpAMP extension will need to know the [generated service.instance.id](https://github.com/open-telemetry/opentelemetry-collector/blob/e98116922cf6c079c246fa78eef59dfd9c2e4867/service/telemetry.go#L131) and other telemetry attributes to use it in the communication with the OpAMP Server. 

This can be implemented by making it available on the component.Host interface or by adding a field to xCreateSettings.

For reference the [OpAMP spec](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes) requires that in OpAMP protocol the following is set:
```
For standalone running Agents (such as OpenTelemetry Collector) the following attributes SHOULD be specified:
- service.name should be set to the same value that the Agent uses in its own telemetry.
- service.namespace if it is used in the environment where the Agent runs.
- service.version should be set to version number of the Agent build.
- service.instance.id should be set. It may be be set equal to the Agent's instance uid (equal to ServerToAgent.instance_uid field) or any other value that uniquely identifies the Agent in combination with other attributes.
```

bogdandrutu:
Why do you need to know only the `service.instance.id` and not all the service parts? Per spec `service.instance.id` is not unique, the combination of the 3 `service` parts are unique.

tigrannajaryan:
> Why do you need to know only the `service.instance.id` and not all the service parts? Per spec `service.instance.id` is not unique, the combination of the 3 `service` parts are unique.

That's a good point. I was thinking that the extension will get the service name from the BuildInfo (we set service.name to BuildInfo.Command), but this is not the right way, since we now allow it to be overriden.

So you are right, it is best that we make all telemetry attributes available to extensions.

bogdandrutu:
Who provides the configuration? Is not the "OpAMP"? If that is the case you have access to them... Also, they are in the config under `service::telemetry::resource`, so if you have access to final config you have access to them.

tigrannajaryan:
>Who provides the configuration? Is not the "OpAMP"?

No, "opamp" extension does not provide configuration.

>Also, they are in the config under service::telemetry::resource, so if you have access to final config you have access to them.

They are not in the config if the user does not specify them. In that case the Collector generates the id and sets the service name equal to BuildInfo.Command. The "opamp" extension needs access to both of these values.

[...]

PR #7531, opened 2023-04-12
Enable create settings to access resource attributes

jaronoff97:
**Description:** 

Adds the ability for create settings to access instantiated resource attributes

**Link to tracking Issue:** #6599 

**Testing:**

Added tests to ensure that resource attributes are being set upon service instantiation, added test to ensure they're not being added when not instantiated

**Documentation:**

Added comments to explain purpose


component/telemetry.go
```diff
@@ -36,4 +36,7 @@ type TelemetrySettings struct {
 	// MetricsLevel controls the level of detail for metrics emitted by the collector.
 	// Experimental: *NOTE* this field is experimental and may be changed or removed.
 	MetricsLevel configtelemetry.Level
+
+	// ResourceAttrs contains the resource attributes for the collector's telemetry.
+	ResourceAttrs map[string]string
```
codeboten, line 41: Could this be a Resource instead of a map?

service/service.go
```diff
@@ -99,14 +101,18 @@ func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
...
 		TracerProvider: srv.telemetry.TracerProvider(),
 		MeterProvider:  metric.NewNoopMeterProvider(),
 		MetricsLevel:   cfg.Telemetry.Metrics.Level,
+
+		// Construct telemetry attributes from build info and config's resource attributes.
+		ResourceAttrs: buildTelAttrs(set.BuildInfo, cfg.Telemetry),
 	}
 
-	if err = srv.telemetryInitializer.init(set.BuildInfo, srv.telemetrySettings.Logger, cfg.Telemetry, set.AsyncErrorChannel); err != nil {
+	if err = srv.telemetryInitializer.init(srv.telemetrySettings.ResourceAttrs, srv.telemetrySettings.Logger, cfg.Telemetry, set.AsyncErrorChannel); err != nil {
```
codeboten, line 115: this could be simplified to only pass in `telemetrySettings`

bogdandrutu:
Confused about this vs https://github.com/open-telemetry/opentelemetry-collector/pull/6833. Which direction are we going?

jaronoff97:
@bogdandrutu this is solving a different issue – #6833 is allowing extensions to read the current configuration of a collector whereas this PR is solely allowing components to see the set resource attributes. I don't believe you would be able to get the resources of a collector given only the configuration of said collector because it lacks information like command, instance id, etc.

service/telemetry_test.go
```diff
@@ -44,19 +44,23 @@ const (
 	counterName  = "test_counter"
 )
 
-func TestBuildTelAttrs(t *testing.T) {
+func TestBuildres(t *testing.T) {
```
codeboten, line 47: nit
```suggestion
func TestBuildResource(t *testing.T) {
```

codeboten (commented):
Just one comment and will need a rebase.

component/telemetry.go
```diff
@@ -16,6 +16,7 @@ package component // import "go.opentelemetry.io/collector/component"
 
 import (
 	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
```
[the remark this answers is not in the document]
codeboten: That's a fair concern, i don't know that we expose the otel sdk anywhere else in the collector API... I'm ok with using `pcommon.Resource` here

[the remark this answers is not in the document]
jaronoff97: sdk's resource is used to initialize the telemetry [here](https://github.com/open-telemetry/opentelemetry-collector/pull/7531/files#diff-52da5566fa270d6a253f7a301b66f4a1dbda7c2c42f3781f8357efe7f08dff10L210) which is why I went with this resource to start with – it allows us to only initialize the sdk resource once and pass it down. Would it be straightforward to convert `pcommon.Resource` to the sdk's resource for initializing the otel sdk?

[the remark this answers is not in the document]
evan-bradley: I'd agree we should avoid making a dependency on the Go SDK in the Collector API since Collector distributions should be free to choose their own SDK. I don't see any Collector APIs where the SDK is currently used, either.

Converting from `pcommon.Resource` to an SDK resource should be fairly easy by just iterating through the underlying map, e.g. through `pResource.Attributes().Range(...)`.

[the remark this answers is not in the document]
jaronoff97: @evan-bradley @bogdandrutu are there any examples of initializing a `pcommon.Resource` every one i've found so far are using this method, which has a note to only use it in tests.
```
// This must be used only in testing code. Users should use "AppendEmpty" when part of a Slice,
// OR directly access the member if this is embedded in another struct.
func NewResource() Resource {
```

Everything else would require me to directly access the internal package which is forbidden. What should i be doing here?
Screen Shot 2023-04-26 at 3 22 24 PM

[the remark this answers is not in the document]
jaronoff97: i ended up using this method as that's the only way to do this as far as i can tell. Please let me know if this is okay, and i'll update the docs.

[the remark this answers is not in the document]
evan-bradley: I think the comment should likely be changed. The apparent reason for the comment is that anything constructing pdata should use e.g. `ptrace.NewTraces()` and just use `ResourceSpans().Attributes()` to set attributes. However, that's not possible if the resource attributes are signal-agnostic, as they are here or in some of the places where `pcommon.NewResource` is used, such as in the resource detection processor.

Someone else can feel free to correct me if I've missed something, but I think this usage is okay.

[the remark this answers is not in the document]
jaronoff97: oy, looks like changing this comment is going to be a PITA... I'll comment on the usage in buildResource.

service/telemetry.go
```diff
@@ -224,9 +182,19 @@ func (tel *telemetryInitializer) initOpenTelemetry(attrs map[string]string, prom
 	if err != nil {
 		return fmt.Errorf("error creating otel prometheus exporter: %w", err)
 	}
+	var resAttrs []attribute.KeyValue
+	res.Attributes().Range(func(k string, v pcommon.Value) bool {
+		resAttrs = append(resAttrs, attribute.String(k, v.AsString()))
+		return true
+	})
+
+	otelRes, err := resource.New(context.Background(), resource.WithAttributes(resAttrs...))
```
evan-bradley, line 191: I'm wondering if we should instead use the OTel Go SDK to construct the resource then convert the SDK resource to pcommon.Resource for use in TelemetrySettings. This would let us use the SDK's resource detectors (e.g. environment variables) in addition to just the attributes specified in the Collector config.

Doing that would be an internal refactoring, so not necessarily a blocking issue here.

evan-bradley (commented):
Overall this looks good to me, thanks for implementing this. Just one question.

service/telemetry.go
jaronoff97: I dig that idea, currently the way we get in resource attributes is using the resource detector processor in the metrics pipeline. If people are open to me making that change, I'd be happy to. cc @codeboten @bogdandrutu

codeboten: I think it's a good idea to allow users to configure telemetry for the collector to be done via SDK configuration.

jaronoff97:
severely messed up my merge... fixing now.

service/telemetry.go
jaronoff97: @evan-bradley @codeboten lmk if the latest change is what you're looking for!

service/service.go
```diff
@@ -29,8 +32,10 @@ import (
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/extension"
 	"go.opentelemetry.io/collector/internal/obsreportconfig"
+	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/processor"
 	"go.opentelemetry.io/collector/receiver"
+	semconv "go.opentelemetry.io/collector/semconv/v1.5.0"
```
bogdandrutu, line 38: We have newer versions of the semver, should we use that?

service/service.go
```diff
@@ -237,3 +251,46 @@ func getBallastSize(host component.Host) uint64 {
...
+		telAttrs = append(telAttrs, attribute.String(semconv.AttributeServiceInstanceID, instanceID))
+	}
+
+	if _, ok := cfg.Resource[semconv.AttributeServiceVersion]; !ok {
+		// AttributeServiceVersion is not specified in the config. Use the actual
+		// build version.
+		telAttrs = append(telAttrs, attribute.String(semconv.AttributeServiceVersion, buildInfo.Version))
+	}
+	return resource.New(ctx,
+		resource.WithFromEnv(),
```
bogdandrutu, line 283: Why this?

bogdandrutu (changes requested):
Please don't try to add side changes like enabling `resource.WithFromEnv()` in a change that tries to do other things.

service/service.go
jaronoff97: i did in a previous version, rebasing blew it away.

bogdandrutu (approved):

Issue #7541, opened 2023-04-18
Builder: add option to skip generate


hypnoce:
**Is your feature request related to a problem? Please describe.**
The builder currently offers skipping building and resolving modules. We have a case where we would like to skip generated sources as we need to version it but then resolve modules and build in the CI.

**Describe the solution you'd like**
Add an option to skip generate.

**Describe alternatives you've considered**
I did not find any alternative as I cannot hook into the builder lifecycle.

bogdandrutu:
To build and resolve modules you don't need this binary, you can use a simple `go build ...` command.

hypnoce:
I agree, but some flags (linking for instance) can be defined in the builder config file which has different owner in my case. The builder provides also out of the box default build configuration that we would like to inherit and follow based on builder evolution.

jpkrohling:
I'm also a bit confused: if you don't have the sources, what do you have? Only the manifest?

hypnoce:
As I understood, there are 3 steps in the builder: code generation, module resolution, build

In my scenario, the 2 first steps are executed in a 'generate source' step and committed to a git repo (skipping compilation).
Then a CI syncs the code and execute the 3rd step (skipping code generation and mod resolution).

It's today possible to skip compilation and module resolution, but not code generation.

jpkrohling:
I see, thank you for the clarification. While I enjoy seeing this tool being used in different manners than what I originally planned, I wouldn't use it this way myself. Besides, I think the value the tool would provide without its main purpose (generating code) would be too low to justify having this option.

On the other hand, I believe it should be relatively easy to implement this solution, as you proved with your PR.

jmacd:
I came looking for a similar issue and found this one. I would like to support use of the builder within an existing monorepo, meaning to support updating an existing go.mod instead somewhere outside the generated code.

See this branch https://github.com/open-telemetry/opentelemetry-collector/compare/main...jmacd:opentelemetry-collector:jmacd/skip_generate_builder

I added one commit on top of the one in #7542 with a second new option related to skipping. The new option `SkipNewGoMod` and command-line flag `--skip-new-go-mod` causes two behaviors:

1. The `go.mod` file will not be generated.
2. The `go mod tidy` step is replaced instead with `go get` for each module except those with `v0.0.0` (indicating they are part of the main module)

[...]

PR #7542, opened 2023-04-18
[cmd/builder] Add option to skip generating sources

hypnoce:
This option makes the builder skip generating sources.

**Link to tracking Issue:** #7541

**Testing:** Unit tests

**Documentation:** as part of the CLI help menu

### DIFF
--- a/cmd/builder/README.md
+++ b/cmd/builder/README.md
@@ -112,3 +112,24 @@ replaces:
   # a list of "replaces" directives that will be part of the resulting go.mod
   - github.com/open-telemetry/opentelemetry-collector-contrib/internal/common => github.com/open-telemetry/opentelemetry-collector-contrib/internal/common v0.40.0
 ```
+
+## Steps
+
+The builder has 3 steps:
+
+* Generate: generates the golang source code
+* Get modules: generates the go.mod file based on the imported modules in the generated golang source code
+* Compilation: builds the OpenTelemetry Collector executable
+
+Each step can be skipped independently: `--skip-generate`, `--skip-get-modules` and `--skip-compilation`.
+
+For instance, a code generation step could execute
+
+```console
+ocb --skip-compilation --config=config.yaml
+```
+then commit the code in a git repo. A CI can sync the code and execute
+```console
+ocb --skip-generate --skip-get-modules --config=config.yaml
+```
+to only execute the compilation step.
--- a/cmd/builder/internal/command.go
+++ b/cmd/builder/internal/command.go
@@ -20,6 +20,7 @@ import (
 )
 
 const (
+	skipGenerateFlag               = "skip-generate"
 	skipCompilationFlag            = "skip-compilation"
 	skipGetModulesFlag             = "skip-get-modules"
 	ldflagsFlag                    = "ldflags"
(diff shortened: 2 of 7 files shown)


cmd/builder/internal/command.go
```diff
@@ -85,6 +86,7 @@ configuration is provided, ocb will generate a default Collector.
 	cmd.Flags().StringVar(&cfgFile, "config", "", "build configuration file")
 
 	// the distribution parameters, which we accept as CLI flags as well
+	cmd.Flags().BoolVar(&cfg.SkipGenerate, skipGenerateFlag, false, "Whether builder should skip generating go code (default false)")
```
bogdandrutu, line 89: If it skips generating go code, what does it do?

hypnoce: Depending on the other options (SkipCompilation and SkipGetModules), it can fetch modules and or build.

bogdandrutu: But that is a "go" command. If anything this tool should not build at all.

hypnoce: You suggest to remove building from the builder ?
Besides, do you know why it's possible to skip module sync but still compiling ? I guess that's also an invalid use case since the go.mod file is not in sync with the generated code in that case.

jpkrohling (commented):
This change deserves some good documentation as part of the readme, especially around the target use case. Other than that, it looks good to me, but I'll let the maintainers judge whether this is a feature worth having.

hypnoce:
> This change deserves some good documentation as part of the readme, especially around the target use case. Other than that, it looks good to me, but I'll let the maintainers judge whether this is a feature worth having.

I've updated the README with clear use cases. Let me know if that's clear enough.

jpkrohling (approved):

jpkrohling (approved):

hypnoce:
@jpkrohling I've added a unit test for SkipGenerate. Let me know what else I can do. Thanks

cmd/builder/README.md
```diff
@@ -112,3 +112,24 @@ replaces:
   # a list of "replaces" directives that will be part of the resulting go.mod
   - github.com/open-telemetry/opentelemetry-collector-contrib/internal/common => github.com/open-telemetry/opentelemetry-collector-contrib/internal/common v0.40.0
 ```
+
+## Steps
+
+The builder have 3 steps:
```
jpkrohling, line 118:
```suggestion
The builder has 3 steps:
```

cmd/builder/README.md
```diff
@@ -112,3 +112,24 @@ replaces:
...
+
+* Generate: generates the golang source code
+* Get modules: generates the go.mod file based on the imported modules in the generated golang source code
+* Compilation: builds the otel executable
```
jpkrohling, line 122:
```suggestion
* Compilation: builds the OpenTelemetry Collector executable
```

jpkrohling (commented):
This looks good, apart from a couple of typos, but we still need @bogdandrutu's blessing, given he had concerns previously about this.

codeboten (approved):
It's a small enough change, it solves a use-case, and it's the only step that is currently not skippable. I'm ok w/ this change.

Issue #7517, opened 2023-04-11
High cardinality metric observed from grpc instrumentation


jaronoff97:
**Describe the bug**
When the flag `telemetry.useOtelForInternalMetrics` is enabled, metrics for gRPC now come through because of [this line](https://github.com/open-telemetry/opentelemetry-collector/blob/a2f0153679220d5de7b2a2ac79a5bebfca90cf89/config/configgrpc/configgrpc.go#L263). When running a collector with a sufficiently high number of connections, this floods the prometheus exporter with a metric with `net.sock.peer.addr` attributes. This resulted in a prometheus scrape being a 33 MB file.

This issue has been reported and discussed [here](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3071). I'm moving the discussion here because this is going to cause any user enabling this flag to potentially experience this cardinality explosion. Temporarily, it would be ideal if we could disable the line I linked to remediate this problem.

**Steps to reproduce**
Create a collector with the `telemetry.useOtelForInternalMetrics` flag enabled, load test it, curl the metrics endpoint.

**What did you expect to see?**
A stable collector that doesn't constantly OOM.

**What did you see instead?**
Cardinality explosion

**What version did you use?**
v0.74.0

**What config did you use?**
Config:
```
    extensions:
      health_check:
        check_collector_pipeline:
          enabled: false
          exporter_failure_threshold: 5
          interval: 5m
        endpoint: 0.0.0.0:13133
        path: /
      pprof: null
    exporters:
      otlp/selfreport:
        endpoint: XXXXXXXXXXXXXXXXX
        headers:
          lightstep-access-token: ${LS_TOKEN}
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
            include_metadata: true
      prometheus:
        config:
          scrape_configs:
          - job_name: spaningest
            scrape_interval: 5s
            static_configs:
            - labels:
                collector_name: ${KUBE_POD_NAME}
              targets:
                - 0.0.0.0:8888
    processors:
      batch:
        send_batch_max_size: 1500
        send_batch_size: 1000
        timeout: 1s
      memory_limiter:
        check_interval: 1s
        limit_percentage: 85
        spike_limit_percentage: 10
    service:
      extensions:
      - health_check
      - pprof
      pipelines:
        metrics:
          exporters:
          - otlp/selfreport
          processors:
          - memory_limiter
          - batch
          receivers:
[...]

PR #7483, opened 2023-04-03
use the otel opencensus bridge

codeboten:
This change configures the opencensus bridge when useOtelForInternalMetrics is enabled. This allows us to stop relying on prometheus as the bridge.

PR #7543, opened 2023-04-18
Put otel meter for http/grpc behind a FF

jaronoff97:
**Description:** Puts the grpc meter provider behind a feature flag for controlling high cardinality metrics.

**Link to tracking Issue:** #7517 

**Testing:** Tested locally and ensured the metric didn't appear when using otel metrics enabled.

### DIFF
--- a/.chloggen/7517-behind-ff.yaml
+++ b/.chloggen/7517-behind-ff.yaml
@@ -1,11 +1,11 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: breaking
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: observavibility
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Puts the meter provider for otlp grpc and otlp http receivers behind a feature flag due to high cardinality issues.
+note: Allows users to disable high cardinality OTLP attributes behind a feature flag.
 
 # One or more tracking issues or pull requests related to the change
 issues: [7517]
--- a/.chloggen/7517-behind-ff.yaml
+++ b/.chloggen/7517-behind-ff.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: observavibility
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Puts the meter provider for otlp grpc and otlp http receivers behind a feature flag due to high cardinality issues.
+
+# One or more tracking issues or pull requests related to the change
+issues: [7517]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/7517-behind-ff.yaml
+++ b/.chloggen/7517-behind-ff.yaml
@@ -2,7 +2,7 @@
 change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
-component: observavibility
+component: service
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Allows users to disable high cardinality OTLP attributes behind a feature flag.
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -45,6 +45,7 @@ import (
 	"go.opentelemetry.io/collector/config/configtls"
 	"go.opentelemetry.io/collector/config/internal"
 	"go.opentelemetry.io/collector/extension/auth"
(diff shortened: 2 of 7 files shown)


mx-psi (changes requested):

config/configgrpc/configgrpc.go
```diff
@@ -260,10 +261,13 @@ func (gcs *GRPCClientSettings) toDialOptions(host component.Host, settings compo
 
 	otelOpts := []otelgrpc.Option{
 		otelgrpc.WithTracerProvider(settings.TracerProvider),
-		otelgrpc.WithMeterProvider(settings.MeterProvider),
 		otelgrpc.WithPropagators(otel.GetTextMapPropagator()),
 	}
 
+	if obsreportconfig.EnableHighCardinalityMetricsfeatureGate.IsEnabled() {
+		otelOpts = append(otelOpts, otelgrpc.WithMeterProvider(settings.MeterProvider))
```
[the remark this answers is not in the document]
jaronoff97: Will adding a view to filter out this metric still allocate memory for all of the potential attribute values? Are there any known performance hits we would take by adding a view?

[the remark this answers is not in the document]
mx-psi: My understanding is that memory would not be allocated and that the performance hit would be negligible (it just means testing if the view matches against instruments at creation time). More concretely, adding a Drop view means that 
- if the instrument is synchronous, the [`aggregators` array on the instrument](https://github.com/open-telemetry/opentelemetry-go/blob/d1959c9239ae1559e828e54cfd05c7bd4adb9c56/sdk/metric/instrument.go#L207) would be `nil`, which would remove any work associated with the metric (it will not be aggregated, nor will it show up in the Prometheus scrape). I guess after looking at the code my question would be if [creating the new set](https://github.com/open-telemetry/opentelemetry-go/blob/d1959c9239ae1559e828e54cfd05c7bd4adb9c56/sdk/metric/instrument.go#L206) allocates any memory (I don't think so from a brief look at the implementation but if it did it could be solved in OTel Go side),
- if the instrument is asynchronous, the callback would not be registered at all, so it would never be called

I guess @Aneurysm9 can settle this more definitively than I can, but in summary I think it should be fine :)

internal/obsreportconfig/obsreportconfig.go
```diff
@@ -31,6 +31,14 @@ var UseOtelForInternalMetricsfeatureGate = featuregate.GlobalRegistry().MustRegi
 	featuregate.StageAlpha,
 	featuregate.WithRegisterDescription("controls whether the collector uses OpenTelemetry for internal metrics"))
 
+// EnableHighCardinalityMetricsfeatureGate is the feature gate that controls whether the collecor should enable
+// potentially high cardinality metrics.
+var EnableHighCardinalityMetricsfeatureGate = featuregate.GlobalRegistry().MustRegister(
```
codeboten, line 36: If this was `DisableHighCardinalityMetricsfeatureGate` instead, we could not have a breaking change immediately, and instead follow the feature gate workflow, where it's off by default on the first release, then on by default w/ the gate still available for anyone that wants it in the next release

config/configgrpc/configgrpc.go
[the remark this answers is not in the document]
codeboten: I agree w/ @mx-psi, I'd expect this to be a view. In a future where views are configurable by end users, this is likely how we would expect it to be filtered out.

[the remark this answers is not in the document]
Aneurysm9: +1 to using a view.

.chloggen/7517-behind-ff.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: observavibility
```
mx-psi, line 5:
```suggestion
component: observability
```

or better yet, for consistency with #7483, 

```suggestion
component: service
```

mx-psi (approved):

internal/obsreportconfig/obsreportconfig.go
```diff
@@ -31,6 +31,14 @@ var UseOtelForInternalMetricsfeatureGate = featuregate.GlobalRegistry().MustRegi
 	featuregate.StageAlpha,
 	featuregate.WithRegisterDescription("controls whether the collector uses OpenTelemetry for internal metrics"))
 
+// DisableHighCardinalityMetricsfeatureGate is the feature gate that controls whether the collector should enable
+// potentially high cardinality metrics.
+var DisableHighCardinalityMetricsfeatureGate = featuregate.GlobalRegistry().MustRegister(
```
bogdandrutu, line 36: Feature flags are not for configuration. Please describe how you intend to remove this?

mx-psi: This was discussed here https://github.com/open-telemetry/opentelemetry-collector/issues/7517#issuecomment-1511168350 it is meant as an interim solution until we can configure views on the OTel configuration, at which point we can remove this.

bogdandrutu: Please document that in the gate description to set the expectations.

Aneurysm9 (approved):

codeboten (approved):
Thanks for addressing my comments.

PR #7545, opened 2023-04-18
Clarify checkpointing order for receivers

tigrannajaryan:
This expands the delivery guarantees introduced for network-based receivers to receivers that use checkpointing.

The purpose of the requirement is the same: to prevent data losses. See previous PR: https://github.com/open-telemetry/opentelemetry-collector/pull/4262

### DIFF
--- a/receiver/doc.go
+++ b/receiver/doc.go
@@ -39,7 +39,7 @@
 // retried. In case of OTLP/HTTP for example, this means that HTTP 429 or 503 response
 // is returned.
 //
-// # Acknowledgment Handling
+// # Acknowledgment and Checkpointing
 //
 // The receivers that receive data via a network protocol that support acknowledgments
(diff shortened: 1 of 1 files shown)


atoulme (approved):

atoulme:
Should this PR skip changelog? I think so since there are no functional changes involved. Please add "Skip changelog" to the PR if so.

tigrannajaryan:
> Should this PR skip changelog? I think so since there are no functional changes involved. Please add "Skip changelog" to the PR if so.

@open-telemetry/collector-approvers what do you think? Do we need a changelog?

dmitryax:
> https://github.com/orgs/open-telemetry/teams/collector-approvers what do you think? Do we need a changelog?

No, we don't need changelog entries for doc updates

dmitryax (approved):

PR #7570, opened 2023-04-25
Add path prefix for otlp http receiver

### LINKED ISSUE #7511: Allow users to specify a path prefix for OTLP HTTP receiver
When running the OTLP receiver behind a regular proxy or gateway, it's often easier to have consistent URLs between the proxy and the backend (OTLP receiver, in this case). Typically, those proxies would be used for multiple backends, each one under a specific path prefix. For instance, `/otlp` would be used by the proxy to pick the collector with the OTLP HTTP receiver at the backend. In practice this means URLs like `https://proxy.example.com/otlp/v1/traces`.

To make it easier to configure an environment like that, I would like to propose a new configuration option, allowing users to specify a path prefix for the OTLP HTTP receiver, likely in the following places:

https://github.com/open-telemetry/opentelemetry-collector/blob/7c982345763f963f9fecf65423096b18890696e6/receiver/otlpreceiver/otlp.go#L203

https://github.com/open-telemetry/opentelemetry-collector/blob/7c982345763f963f9fecf65423096b18890696e6/receiver/otlpreceiver/otlp.go#L228

https://github.com/open-telemetry/opentelemetry-collector/blob/7c982345763f963f9fecf65423096b18890696e6/receiver/otlpreceiver/otlp.go#L253

fredthomsen:
**Description:** 
Supports multiple otlp http receivers running behind a reverse proxy/ingress.

Fixes #7511 

**Testing:** 
Updated existing tests to account for new parameter.

**Documentation:**
Added documentation for configuring prefix in the receiver README.


receiver/otlpreceiver/README.md
```diff
@@ -31,6 +31,11 @@ The following settings are configurable:
 - `endpoint` (default = 0.0.0.0:4317 for grpc protocol, 0.0.0.0:4318 http protocol):
   host:port to which the receiver is going to receive data. The valid syntax is
   described at https://github.com/grpc/grpc/blob/master/doc/naming.md.
+- `path_prefix` (default = "" and only applicable to http protocol):
```
atoulme, line 34: what is the behavior if grpc is used? Is it ignored?

atoulme: Should it be `http/path_prefix` ?

atoulme: OK, I would move this down after CORS and provide an example.

fredthomsen: > what is the behavior if grpc is used? Is it ignored?

Yes.  I'll update the README to be more explicit.

fredthomsen: > Should it be `http/path_prefix` ?

It is already under the http settings.  Let me try and make that clearer.

fredthomsen: > OK, I would move this down after CORS and provide an example.

I thought under the HTTP/JSON section might be better given what's there.  If this still doesn't pass muster, I can move to a new section under CORS.

fredthomsen: @atoulme are you in agreement with the changes that I made above?

atoulme:
LGTM. The only thing is maybe this should be offered at the confighttp.HTTPServerSettings level? A maintainer should have a better view.

atoulme (approved):

fredthomsen:
> LGTM. The only thing is maybe this should be offered at the confighttp.HTTPServerSettings level? A maintainer should have a better view.

Looking over this, yeah I agree.  This would be better served having the `PathPrefix` as a field inside of `confighttp.HTTPServerSettings`.  I'll make those updates.

receiver/otlpreceiver/config_test.go
```diff
@@ -126,18 +126,21 @@ func TestUnmarshalConfig(t *testing.T) {
...
+						CORS: &confighttp.CORSSettings{
+							AllowedOrigins: []string{"https://*.test.com", "https://test.com"},
+							MaxAge:         7200,
 						},
 					},
-					CORS: &confighttp.CORSSettings{
-						AllowedOrigins: []string{"https://*.test.com", "https://test.com"},
-						MaxAge:         7200,
-					},
+					PathPrefix: "testprefix",
```
jpkrohling, line 143: Could you please add a test for "/"? We should end with a normalized path with a single slash instead of two.

receiver/otlpreceiver/testdata/config.yaml
```diff
@@ -39,3 +39,4 @@ protocols:
         - https://*.test.com # Wildcard subdomain. Allows domains like https://www.test.com and https://foo.test.com but not https://wwwtest.com.
         - https://test.com # Fully qualified domain name. Allows https://test.com only.
       max_age: 7200
+    path_prefix: "testprefix"
```
jpkrohling, line 42:
```suggestion
    path_prefix: "/otlp"
```

I think this is easier for readers to reason about.

jpkrohling (commented):
I like this change, especially as it's forward-compatible with a possible generalization of the path prefix into the HTTPServerSettings. Given that no other receivers would use the path prefix, I'm against adding it directly there.

Let's add it only for OTLP Receiver, see the feedback, and generalize it if needed.

fredthomsen:
@jpkrohling ok, I've made the updates you requested.

receiver/otlpreceiver/otlp.go
```diff
@@ -200,7 +202,8 @@ func (r *otlpReceiver) registerTraceConsumer(tc consumer.Traces) error {
 	r.tracesReceiver = trace.New(tc, r.obsrepGRPC)
 	httpTracesReceiver := trace.New(tc, r.obsrepHTTP)
 	if r.httpMux != nil {
-		r.httpMux.HandleFunc("/v1/traces", func(resp http.ResponseWriter, req *http.Request) {
+		urlPath := filepath.Join("/", url.PathEscape(r.cfg.HTTP.PathPrefix), "v1/traces")
```
bogdandrutu, line 205: `url.PathJoin`?

bogdandrutu:
@jpkrohling @atoulme @fredthomsen please make sure we are consistent with the exporter option do configure this, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/config.go#L26

I think we should do something like that instead, since the next request will be to change the "fixed" path. So I think in terms of something like having the full path `traces_path`, `metrics_path`, `logs_path` which will replace the entire path and default to otlp defaults. What do you think?

fredthomsen:
> @jpkrohling @atoulme @fredthomsen please make sure we are consistent with the exporter option do configure this, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/config.go#L26
> 
> I think we should do something like that instead, since the next request will be to change the "fixed" path. So I think in terms of something like having the full path `traces_path`, `metrics_path`, `logs_path` which will replace the entire path and default to otlp defaults. What do you think?

So, I think there is two options here:
* Just having endpoints per signal type and dropping a command path prefix
* Keep common path prefix, and then allow configuration to override http paths per signal type.
I lean towards the second option, given I think that setting individual http signal paths will be more infrequent .

jpkrohling:
I agree that the ideal is to have three endpoints for the HTTP receiver as well.

fredthomsen:
> I agree that the ideal is to have three endpoints for the HTTP receiver as well.

@jpkrohling @bogdandrutu I've moved to three separate url path configurations.  I was waffling between validating the path config up front or not given that there is a panic caused by an invalid path.

receiver/otlpreceiver/config.go
```diff
@@ -66,6 +81,20 @@ func (cfg *Config) Unmarshal(conf *confmap.Conf) error {
 
 	if !conf.IsSet(protoHTTP) {
 		cfg.HTTP = nil
+	} else {
+		// Verify URL path sanity
+		signalURLPaths := []*string{&cfg.HTTP.TracesURLPath, &cfg.HTTP.MetricsURLPath, &cfg.HTTP.LogsURLPath}
+		for i, urlPath := range signalURLPaths {
+			u, err := url.Parse(*urlPath)
+			if err != nil {
+				return errors.New("Invalid HTTP URL path set for signal")
```
jpkrohling, line 90: I think you should wrap the original error, so that users would know what's wrong.

```suggestion
				return fmt.Errorf("invalid HTTP URL path set for signal: %w", err)
```

receiver/otlpreceiver/config.go
```diff
@@ -66,6 +81,20 @@ func (cfg *Config) Unmarshal(conf *confmap.Conf) error {
 
 	if !conf.IsSet(protoHTTP) {
 		cfg.HTTP = nil
+	} else {
+		// Verify URL path sanity
+		signalURLPaths := []*string{&cfg.HTTP.TracesURLPath, &cfg.HTTP.MetricsURLPath, &cfg.HTTP.LogsURLPath}
```
jpkrohling, line 86: It's probably a matter of taste, but I would the sanitization on its own function, and call it three times. Like this (untested):

```go
var err error

if cfg.HTTP.TracesURLPath, err = sanitizeURLPath(cfg.HTTP.TracesURLPath); err != nil {
  return err
}

if cfg.HTTP.MetricsURLPath, err = sanitizeURLPath(cfg.HTTP.MetricsURLPath); err != nil {
  return err
}

if cfg.HTTP.LogsURLPath, err = sanitizeURLPath(cfg.HTTP.LogsURLPath); err != nil {
  return err
}

func sanitizeURLPath(urlPath string) string {
  u, err := url.Parse(*urlPath)
  if err != nil { 
    return fmt.Errorf("invalid HTTP URL path set for signal: %w", err)
  }

  if !path.IsAbs(u.Path) {
    u.Path = "/" + u.Path
  }

  return u.Path
}

```

receiver/otlpreceiver/otlp_test.go
```diff
@@ -178,10 +178,11 @@ func TestJsonHttp(t *testing.T) {
 		},
 	}
 	addr := testutil.GetAvailableLocalAddress(t)
+	tracesURLPath := "/v1/traceingest"
 
 	// Set the buffer count to 1 to make it flush the test span immediately.
 	sink := &errOrSinkConsumer{TracesSink: new(consumertest.TracesSink)}
-	ocr := newHTTPReceiver(t, addr, sink, nil)
+	ocr := newHTTPReceiver(t, addr, tracesURLPath, defaultMetricsURLPath, sink, nil)
```
jpkrohling, line 185: I might find the answer later in this PR, but why is it missing the logs URL path?

receiver/otlpreceiver/otlp_test.go
```diff
@@ -941,10 +961,13 @@ func newGRPCReceiver(t *testing.T, endpoint string, tc consumer.Traces, mc consu
...
 }
 
-func newHTTPReceiver(t *testing.T, endpoint string, tc consumer.Traces, mc consumer.Metrics) component.Component {
+func newHTTPReceiver(t *testing.T, endpoint string, tracesURLPath string, metricsURLPath string, tc consumer.Traces, mc consumer.Metrics) component.Component {
 	factory := NewFactory()
 	cfg := factory.CreateDefaultConfig().(*Config)
 	cfg.HTTP.Endpoint = endpoint
+	cfg.HTTP.TracesURLPath = tracesURLPath
+	cfg.HTTP.MetricsURLPath = metricsURLPath
+	cfg.HTTP.LogsURLPath = defaultLogsURLPath
```
jpkrohling, line 970: Why is logs different here?

receiver/otlpreceiver/otlp_test.go
fredthomsen: Don't think logs were ever really leverage in this test file.  Of course, them being present in the config below proves there is some need, so I'll just go ahead and add this additional parameter to avoid confusion.

receiver/otlpreceiver/otlp_test.go
fredthomsen: I've changed the log path difference so it doesn't cause any confusion.

jpkrohling (approved):
LGTM, I tested with a config like this:

```
receivers:
  otlp/1:
    protocols:
      grpc:

  otlp/2:
    protocols:
      http:
        traces_url_path: /traces

exporters:
  otlphttp/2:
    traces_endpoint: http://localhost:4318/traces
  logging:

service:
  pipelines:
    traces/1:
      receivers: [ otlp/1 ]
      processors:
      exporters: [ otlphttp/2 ]
    traces/2:
      receivers: [ otlp/2 ]
      processors:
      exporters: [ logging ]
```

I did have some trouble in finding the right config, so perhaps the following could be added to the readme:

1) explicitly mention in the readme how and where to set the endpoints. This is optional, as people can look into the test data for reference
2) state that, on the other side, the OTLP HTTP exporter should use the similar attributes for an end-to-end solution.

I wasn't able to get the SDK to work with this, including the telemetrygen. If you can make it work with that, it would be worth adding instructions here as well. 

Those are things that can be done in a follow-up PR. This one here LGTM.

fredthomsen:
>     1. explicitly mention in the readme how and where to set the endpoints. This is optional, as people can look into the test data for reference

There was information in the README present already, but it had the incorrect configuration parameter names.   I have updated them now.

>     2. state that, on the other side, the OTLP HTTP exporter should use the similar attributes for an end-to-end solution.

Agreed.  I have added this information.

> I wasn't able to get the SDK to work with this, including the telemetrygen. If you can make it work with that, it would be worth adding instructions here as well.
> 
> Those are things that can be done in a follow-up PR. This one here LGTM.

I am not too familiar w/telemetrygen, so yeah let's tackle that in a follow up.

jpkrohling (approved):

fredthomsen:
Surprised to see these unit test failures as locally my runs are working fine.

fredthomsen:
I stand corrected. I see it now.

> Surprised to see these unit test failures as locally my runs are working fine.

jpkrohling:
@open-telemetry/collector-maintainers, would one of you please review/merge this one?

receiver/otlpreceiver/factory_test.go
```diff
@@ -162,8 +175,11 @@ func TestCreateMetricReceiver(t *testing.T) {
 			cfg: &Config{
 				Protocols: Protocols{
 					GRPC: defaultGRPCSettings,
-					HTTP: &confighttp.HTTPServerSettings{
-						Endpoint: "327.0.0.1:1122",
+					HTTP: &httpServerSettings{
+						HTTPServerSettings: &confighttp.HTTPServerSettings{
+							Endpoint: "327.0.0.1:1122",
```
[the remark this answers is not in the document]
fredthomsen: This was for an invalid endpoint test so there is an invalid IP here.

receiver/otlpreceiver/factory_test.go
```diff
@@ -238,8 +259,11 @@ func TestCreateLogReceiver(t *testing.T) {
...
```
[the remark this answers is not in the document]
fredthomsen: Again invalid endpoint test.

receiver/otlpreceiver/factory_test.go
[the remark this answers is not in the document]
fredthomsen: These were here prior to my changes, I just indented them to put them in the new structure.  This is just to test for an invalid endpoint.

receiver/otlpreceiver/factory_test.go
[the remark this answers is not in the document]
fredthomsen: same answer as above.

receiver/otlpreceiver/factory_test.go
```diff
@@ -251,8 +275,10 @@ func TestCreateLogReceiver(t *testing.T) {
...
```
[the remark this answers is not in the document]
fredthomsen: Alright this one may actually be a copy and paste error from the test author.  I'll update this guy as I don't think there is a need for an invalid IP here.

codeboten (approved):

PR #7573, opened 2023-04-26
Retract pdata v1.0.0-rc10 and several modules v0.76.0 which depended on it

djaglowski:
Our versioning scheme for release candidate modules depends on alphabetical ordering. As a result `rc10` is interpreted as being less then `rc2` through `rc9`.

This causes all sorts of problems with dependency management. e.g.:
`go: downgraded [go.opentelemetry.io/collector/pdata](http://go.opentelemetry.io/collector/pdata) v1.0.0-rc9.0.20230424163446-8e33ded10872 => v1.0.0-rc10`

As a result, I suggest we retract `pdata v1.0.0-rc10` and all `v0.76.0` modules that depend on it.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -14,7 +14,7 @@
 
 module-sets:
   stable:
-    version: v1.0.0-rc10
+    version: v1.0.0-rc9
     modules:
       - go.opentelemetry.io/collector/pdata
   beta:
--- a/cmd/otelcorecol/go.mod
+++ b/cmd/otelcorecol/go.mod
@@ -71,7 +71,7 @@ require (
 	go.opentelemetry.io/collector/confmap v0.76.0 // indirect
 	go.opentelemetry.io/collector/consumer v0.76.0 // indirect
 	go.opentelemetry.io/collector/featuregate v0.76.0 // indirect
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 <!-- next version -->
 
-## v1.0.0-rc10/v0.76.0
+## v1.0.0-rc9/v0.76.0
```
djaglowski, line 7: This line should probably be relabeled to `v1.0.0-rc9/v0.76.1`, but maybe after that release is created?

versions.yaml
```diff
@@ -14,11 +14,11 @@
 
 module-sets:
   stable:
-    version: v1.0.0-rc10
+    version: v1.0.0-rc9
     modules:
       - go.opentelemetry.io/collector/pdata
   beta:
-    version: v0.76.0
+    version: v0.75.0
```
djaglowski, line 21: I don't know what is appropriate to do with these prior to cutting another release.

djaglowski:
Unless others think we should merge this sooner rather than later, we may want to hold off for a bit.

I'm looking into rolling this into a new release.

djaglowski:
Closing in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/7577

PR #7574, opened 2023-04-26
[config/confighttp] Ensure Auth happens after compression

Aneurysm9:
When request signing-based authentication extensions are used it is necessary to ensure that the Auth RoundTripper is the innermost as changes to the request by other RoundTrippers (such as compression or headers) may invalidate the request signature.  This change will ensure that the Auth RoundTripper is invoked after all others and is followed directly by the http.Transport.

### DIFF
--- a/config/confighttp/confighttp_test.go
+++ b/config/confighttp/confighttp_test.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/config/configtelemetry"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest/observer"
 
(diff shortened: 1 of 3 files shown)


config/confighttp/confighttp_test.go
```diff
@@ -336,15 +366,34 @@ func TestHTTPClientSettingWithAuthConfig(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			client, err := test.settings.ToClient(test.host, componenttest.NewNopTelemetrySettings())
+			// Omit TracerProvider and MeterProvider in TelemetrySettings as otelhttp.Transport cannot be introspected
+			client, err := test.settings.ToClient(test.host, component.TelemetrySettings{Logger: zap.NewNop(), MetricsLevel: configtelemetry.LevelNone})
```
bogdandrutu, line 370: use componenttest to create the TelemetrySettings?

bogdandrutu (approved):

config/confighttp/confighttp_test.go
Aneurysm9: As noted in the comment the `otelhttp.Transport` doesn't export the `RoundTripper` that it wraps and so there would be no way to test the containment order beyond that transport if the `componenttest.NewNopTelemetrySettings()` were used.

codeboten (approved):

PR #7579, opened 2023-04-26
Mark forward connector as beta

djaglowski:
This also updates the readme to indicate correctly the distributions that already contain the connector.

### DIFF
--- a/.chloggen/forward-beta.yaml
+++ b/.chloggen/forward-beta.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: forwardconnector
+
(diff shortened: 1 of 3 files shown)


.chloggen/forward-beta.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
codeboten, line 2: Since this is a docs only change, i would remove the changelog here and skip the check

djaglowski: Isn't it a user facing concern though?

codeboten: Oh... maybe i'm misunderstanding. Should this change also update `forward.go`? https://github.com/open-telemetry/opentelemetry-collector/blob/3b62976f35387b150244390e8f8286a820966963/connector/forwardconnector/forward.go#L34-L36

bogdandrutu: +1 @codeboten

codeboten (approved):

codeboten (approved):

Issue #7043, opened 2023-01-27
Review featuregate States and removal version


bogdandrutu:
There are some caveats that I found while working with this API/functionality:

- [ ] #7621
- [x] I saw a need of having a "deprecated" State since sometimes a feature may fail and we want to mark the end of life for a gate that is not "stable" (successful). This state should be equivalent in capabilities (not allowing to be set, etc.) with Stable, but force the gate to be "false" (unsuccessful).
- [x] It will be good to have a "from" date as well to mark when a gate was added. This way at the beginning of each release (or right before) we can build a table like https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/#feature-gates-that-are-removed

bogdandrutu:
Marked as resolve, extracted what is left in a separate issue.

PR #7586, opened 2023-05-01
[featuregate] Fix issue where StageDeprecated cause an error on register

TylerHelmuth:
**Description:**
Updates the new `StageDeprecated` stage to work like `StageStable` but false.  `StageDeprecated` is now no longer changeable and expects a `ToVersion` value to be included when used.  Also fixes a defect where `StageDeprecated` caused an `unknown stage` error on register.

**Link to tracking Issue:** 
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/7043

**Testing:** 
Added tests for `StageDeprecated`

**Documentation:** 
Updated featuregate README to include how `deprecated` is used when a feature is abandoned in `alpha`.

### DIFF
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -71,6 +71,7 @@ without proceeding to the `beta` stage. Instead, they will proceed to the
 `deprecated` stage, which will feature is permanently disabled. A feature gate will
 be removed once it has been `deprecated` for at least 2 releases of the collector.
 
-Features that make it to the `beta` stage will not be dropped and will eventually
-reach general availability where the `Gate` that allowed them to be disabled during
-the `beta` stage will be removed.
+Features that make it to the `beta` stage are intended to reach general availability but may still be discontinued.
+If, after wider use, it is determined that the gate should be discontinued it will be reverted to the `alpha` stage
+for 2 releases and then proceed to the `deprecated` stage. If instead it is ready for general availability it will
+proceed to the `stable` stage.
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -67,7 +67,10 @@ modeled after the [system used by Kubernetes](https://kubernetes.io/docs/referen
 4. A `stable` feature gate will be removed in the version specified by its `ToVersion` value.
 
 Features that prove unworkable in the `alpha` stage may be discontinued 
-without proceeding to the `beta` stage.  Features that make it to the `beta` 
-stage will not be dropped and will eventually reach general availability 
-where the `Gate` that allowed them to be disabled during the `beta` stage 
-will be removed.
+without proceeding to the `beta` stage. Instead, they will proceed to the
+`deprecated` stage, which will feature is permanently disabled. A feature gate will
+be removed once it has been `deprecated` for at least 2 releases of the collector.
+
+Features that make it to the `beta` stage will not be dropped and will eventually
+reach general availability where the `Gate` that allowed them to be disabled during
+the `beta` stage will be removed.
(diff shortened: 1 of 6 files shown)


featuregate/README.md
```diff
@@ -67,7 +67,10 @@ modeled after the [system used by Kubernetes](https://kubernetes.io/docs/referen
...
 Features that prove unworkable in the `alpha` stage may be discontinued 
-without proceeding to the `beta` stage.  Features that make it to the `beta` 
-stage will not be dropped and will eventually reach general availability 
-where the `Gate` that allowed them to be disabled during the `beta` stage 
-will be removed.
+without proceeding to the `beta` stage. Instead, they will proceed to the
+`deprecated` stage, which will feature is permanently disabled. A feature gate will
+be removed once it has been `deprecated` for at least 2 releases of the collector.
+
+Features that make it to the `beta` stage will not be dropped and will eventually
```
bogdandrutu, line 74: Are we sure that we never revert this? Just curios, what is then the purpose of beta if no roll-back?

featuregate/registry.go
```diff
@@ -110,8 +110,8 @@ func (r *Registry) Register(id string, stage Stage, opts ...RegisterOption) (*Ga
 	default:
 		return nil, fmt.Errorf("unknown stage value %q for gate %q", stage, id)
 	}
-	if g.stage == StageStable && g.toVersion == "" {
-		return nil, fmt.Errorf("no removal version set for stable gate %q", id)
+	if (g.stage == StageStable || g.stage == StageDeprecated) && g.toVersion == "" {
```
bogdandrutu, line 113: Sorry for creating a bit more work to you, but would really appreciate a conclusion on https://github.com/open-telemetry/opentelemetry-collector/issues/7621, do you mind getting that to the final line?

featuregate/README.md
TylerHelmuth: We can definitely change its meaning, but the current README says this: `Features that prove unworkable in the alpha stage may be discontinued without proceeding to the beta stage.  Features that make it to the beta stage will not be dropped and will eventually reach general availability`.

If we don't want to be so locked in we could remove that wording, but I think the original intent was that if we are going to send it to beta and enable it by default we are going to keep the feature.  We use beta instead of going straight to stable to give users time to adjust to the new feature, not because we think we might get rid of it.

featuregate/registry.go
TylerHelmuth: Yup I was planning to work on that issue

TylerHelmuth: Lets merge https://github.com/open-telemetry/opentelemetry-collector/pull/7626 before this PR.

bogdandrutu: Please rebase

featuregate/README.md
evan-bradley: I agree that in general sending a feature gate to beta signals an intent to go forward with it, but it's always possible that widespread use while in beta reveals that a feature is untenable or otherwise needs major rework. I think it makes sense to indicate here that beta features are intended to become stable and likely won't be deprecated, but I'm not sure we should lock ourselves into guaranteeing that they become stable.

TylerHelmuth: Updated the stance on beta to be less strict and document the rollback procedure.

bogdandrutu (approved):

bogdandrutu (approved):

bogdandrutu (approved):

PR #7589, opened 2023-05-01
[chore] add license-check

codeboten:
First pass at cleaning up the license in files.

### DIFF
--- a/.github/workflows/scripts/dependabot-pr.sh
+++ b/.github/workflows/scripts/dependabot-pr.sh
@@ -1,4 +1,5 @@
 #!/bin/bash -ex
+#
 # Copyright The OpenTelemetry Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/.github/workflows/scripts/release-check-blockers.sh
+++ b/.github/workflows/scripts/release-check-blockers.sh
@@ -1,4 +1,5 @@
 #!/bin/bash -ex
+#
 # Copyright The OpenTelemetry Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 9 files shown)


dmitryax:
We already have `addlicense` and `checklicense` targets in Makefile.common. Do we really need another one?

I think the problem with the existing `checklicense` is that it checks only for _presence_ of the license but doesn't compare the header line with what we want. I think we need to remove `-check` from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/Makefile.Common#L131 and use the diff approach that we have for other generated stuff. WDYT?

codeboten:
> We already have `addlicense` and `checklicense` targets in Makefile.common. Do we really need another one?
> 

We likely don't need another one. I just used the check license from the opentelemetry-go-build-tools repo. I don't care which way we end up going, but i do like the simplicity of the check w/ awk

codeboten:
After testing this out, i don't think addlicense will tell us if a license is different than what's expected. I ran it on the repo as is, and although it added licenses to any missing files, it left these in place (note the 2 spaces after Copyright):

```
// Copyright  The OpenTelemetry Authors
```

It's possible I misused `addlicence` though

.github/workflows/scripts/dependabot-pr.sh
```diff
@@ -1,4 +1,17 @@
 #!/bin/bash -ex
+# Copyright The OpenTelemetry Authors
```
dmitryax, line 2: We have a line break here in contrib scripts. Should we do the same in core?

codeboten: Done

.github/workflows/scripts/dependabot-pr.sh
```diff
@@ -1,4 +1,18 @@
 #!/bin/bash -ex
+#
+# Copyright The OpenTelemetry Authors
```
bogdandrutu, line 3: Should this be before the first line?

bogdandrutu (commented):
LGTM on the fixes, I will let you folks @dmitryax and @codeboten to decide which tool to use.

dmitryax (approved):

Issue #7359, opened 2023-03-13
[exporterhelper] Default queue size is too large


swiatekm-sumo:
**Is your feature request related to a problem? Please describe.**
Exporterhelper has a default queue size of 5000. If we use the batch processor with default settings as well, that results in a maximum size of 5000 * 8192 ~= 41 000 000 spans. At a conservative 500 bytes per span, we'd consume ~20 GB of memory, and most likely be killed by either the orchestrator or the OS itself.

This is way too much in my opinion. The default size of this queue should allow for enough buffer space to absorb temporary increases in data volume and to feed the 10 consumer threads without issue. It shouldn't try to cover for a prolonged remote unavailability - if the user wants that, they should probably use the persistent queue instead.

**Describe the solution you'd like**
I'd like the value to be reduced to the low 100s, both 100 and 200 seem fine given the above napkin math.

**Describe alternatives you've considered**
If the default stays as is, I think we should include a warning about this in the documentation.

**Additional context**
This is especially dangerous given that the sending queue is enabled by default for a lot of exporters, including otlpexporter.

swiatekm-sumo:
From the SIG meeting at 26.04.2023:
* reduce this a bit more gently, starting with 5000 -> 1000
* look into reducing the default batch size as well

mx-psi:
Fixed by #7592

dmitryax:
I believe it's not resolved yet. 1000 was the first step, but we want to reduce it further

swiatekm-sumo:
As per the SIG discussion on 21.06.2023, we'll leave this value alone, as it's likely to be replaced by a different key as we move batching to exporterhelper.

PR #7592, opened 2023-05-02
[exporterhelper]: reduce default queue size to 1000

swiatekm-sumo:
**Description:**
Reduce the default queue size to 1000. Combined with the defaults on batchprocessor, the current default of 5000 is too much and will lead to an OOMKill in most environments if actually filled.

1000 is probably too high still, but we don't want to make too drastic of a change here at once. See linked issue for specifics.

**Link to tracking Issue:** #7359

**Documentation:** Updated the relevant README

### DIFF
--- a/exporter/exporterhelper/queued_retry.go
+++ b/exporter/exporterhelper/queued_retry.go
@@ -34,6 +34,8 @@ import (
 	"go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 )
 
+const defaultQueueSize = 1000
+
 var (
 	errSendingQueueIsFull = errors.New("sending_queue is full")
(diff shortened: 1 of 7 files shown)


exporter/exporterhelper/queued_retry.go
```diff
@@ -34,6 +34,8 @@ import (
 	"go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 )
 
+const DefaultQueueSize = 1000
```
bogdandrutu, line 37: Why exposing this public?

swiatekm-sumo: No reason in the end, made it private.

dmitryax (approved):

PR #7635, opened 2023-05-07
[scrapehelper] Start scraping from component start

MovieStoreGuy:
**Description:** 
This change will allow scrapers to start gather values on scrape controller start instead of having to wait for scrape controller duration to perform the initial scrape.

This is particularly useful for components such as `hostmetrics receiver` that generate `utilization` metrics that require an initial state to compute that value. With this change, it means that this metrics will start being sent from `Start + Duration` instead of `Start + 2*Duration` which for scrape durations like `30s` means that this value won't be sent til a minute has elapsed. 

**Link to tracking Issue:** 
N/A 
Mostly because I forgot to raise an issue before I did this.

**Testing:** 

Updated the test cases where appropriate. 

**Documentation:** 

I don't believe any user / dev docs need to be updated beyond the change log since the only behavioural change is when the scrapers are called which shouldn't impact any behaviours scrapers.


MovieStoreGuy:
One thing worth pointing out for this change is that having scrapers starting on collector / component start will have an increase on resource usage (CPU, Mem, I/O, etc ...) since scraper start will happen shortly after create and the missed interval for `hostmetrics` will now be included in the metrics results.

dmitryax:
With this change, the utilization metrics will still be misaligned with others, right? Maybe we can better fix those instead? We can record the initial utilization values on Start.

I like the current state of reporting metrics on `Start + Duration` because it's easy to differentiate start issues from scraping issues by looking at the logs. And, as you mentioned, it puts more memory/cpu pressure on the start procedure, which already can be pretty busy.

MovieStoreGuy:
> With this change, the utilization metrics will still be misaligned with others, right?

Yes, they will continue to be offset by one collection (specifically for the hostmetric receiver, not sure if other receivers also experience this).

> Maybe we can better fix those instead? We can record the initial utilization values on Start.

I like this idea, perhaps worth raising an issue in contrib for the host receiver (since I only noticed it there.)

> I like the current state of reporting metrics on Start + Duration because it's easy to differentiate start issues from scraping issues by looking at the logs. And, as you mentioned, it puts more memory/cpu pressure on the start procedure, which already can be pretty busy.

My only hesitation on keeping as it is with `Start + Duration` is that it means that scrapers have this min time to start producing data which is problematic for scrapers that can't fully validate the configuration until they are started. 
This could lead to setting collection interval to `1s` then potentially flooding the database for example instead of doing one scrape `30s` and failing fast on component start.

bogdandrutu:
I like this, just brainstorming an idea: maybe we have a config for "initial scrape delay" and have that set to 1sec or something like that, then we do the interval from there?

MovieStoreGuy:
> I like this, just brainstorming an idea: maybe we have a config for "initial scrape delay" and have that set to 1sec or something like that, then we do the interval from there?

Yeah that makes sense to me, we can keep the current behaviour (ie: `initial_delay = collection_duration`) but expose it as a field.

receiver/scraperhelper/scrapercontroller.go
```diff
@@ -34,7 +34,8 @@ import (
 // configuration. Scraper controller receivers can embed this struct, instead
 // of receiver.Settings, and extend it with more fields if needed.
 type ScraperControllerSettings struct {
-	CollectionInterval time.Duration `mapstructure:"collection_interval"`
+	CollectionInterval time.Duration  `mapstructure:"collection_interval"`
+	InitialDelay       *time.Duration `mapstructure:"initial_delay"`
```
mx-psi, line 38: We should add a comment explaining why we use a pointer here (to distinguish between the zero value and the field being unset?). We don't generally use pointer fields, so I think we need to have a good reason to add it here

MovieStoreGuy: That is a good point, I would if it is worth a type addition to confmap?

MovieStoreGuy: I guess what would be better to ask is, what would make more sense:

- Define the method for `confmap.Marshaller` and see if it is being explicitly set like the mdatagen package does?
- Continue using the pointer method and just let it be set by `mapstructure`
- Using a value that isn't the zero value like `-1` that would imply isn't set and should default to X behaviour
- Include a new type in confmap that is safe to consume the zero value, doesn't require a pointer and is alligned to the sql package?

mx-psi: > Define the method for confmap.Marshaller and see if it is being explicitly set like the mdatagen package does?

This sounds like the best option to me

receiver/scraperhelper/scrapercontroller.go
```diff
@@ -184,7 +191,16 @@ func (sc *controller) startScraping() {
 
 			sc.tickerCh = ticker.C
 		}
+		if sc.initialDelay > 0 {
+			timer := time.NewTimer(sc.collectionInterval)
+			<-timer.C
+			timer.Stop()
```
bogdandrutu, line 197:
```suggestion
			<-time.After(sc.initialDelay)
```

receiver/scraperhelper/scrapercontroller.go
```diff
@@ -184,7 +191,16 @@ func (sc *controller) startScraping() {
 
 			sc.tickerCh = ticker.C
 		}
+		if sc.initialDelay > 0 {
+			timer := time.NewTimer(sc.collectionInterval)
```
bogdandrutu, line 195: wrong duration

receiver/scraperhelper/scrapercontroller.go
bogdandrutu: I want to understand the motivation in the first place? Are we not ok to change the default behavior to be an immediate scrape by default?

receiver/scraperhelper/scrapercontroller.go
MovieStoreGuy: I had thought about doing that but `time.After` explicitly calls out that it leaks the underlying go channel.

I know it will only happen once but I didn't want to give the wrong impression that it should be used more regularly throughout the collector.

receiver/scraperhelper/scrapercontroller.go
MovieStoreGuy: Yeah, I was thinking from a user's perspective that behaviour suddenly changes and mislead to believe that the collector is now suddenly using more resources due to the potential overlap of collector getting ready and starting.

However, making this an opt out does make this a lot easier.

receiver/scraperhelper/scrapercontroller.go
bogdandrutu: Are you talking about the fact that the timer is not collected until it fires https://pkg.go.dev/time#After? Or do I miss something else?

MovieStoreGuy: Yes, though I remembered it differently. Considering I am not waiting on a secondary channel it should be fine.

MovieStoreGuy:
I've updated the PR to be more of an opt out approach so users are required to explicitly set `initial_delay:0` within the configuration instead of going for an opt in approach which convoluted the design a bit.

receiver/scraperhelper/scrapercontroller.go
```diff
@@ -184,7 +188,14 @@ func (sc *controller) startScraping() {
 
 			sc.tickerCh = ticker.C
 		}
+		if sc.initialDelay > 0 {
```
astencel-sumo, line 191: If I understand correctly, this means that if I configure CollectionInterval to 10s and InitialDelay to 2s, the scrape will occur at the following times since start of collector: 2s, 10s, 20s, 30s, ...

Shouldn't the initial delay also delay the following scrapes? I would assume the scrapes would occur at: 2s, 12s, 22s, 32s, ...

MovieStoreGuy: Yes, you're example is correct. 

I think the ordering should be fixed up to address that to align with what I imagine most users would expect and make it clear in the change log.

MovieStoreGuy: Updated :)

astencel-sumo: Great, thanks!

receiver/scraperhelper/scrapercontroller_test.go
```diff
@@ -352,12 +363,83 @@ func TestSingleScrapePerTick(t *testing.T) {
...
+		receivertest.NewNopCreateSettings(),
+		new(consumertest.MetricsSink),
+		AddScraper(scp),
+	)
+	require.NoError(t, err, "Must not error when creating scrape controller")
+
+	assert.NoError(t, r.Start(context.Background(), componenttest.NewNopHost()), "Must not error on start")
+	<-time.After(500 * time.Nanosecond)
+	assert.NoError(t, r.Shutdown(context.Background()), "Must not have errored on shutdown")
+	assert.Equal(t, tsm.timesScrapeCalled, 1, "Must have been called as soon as the controller started")
```
astencel-sumo, line 408: Expected value should be first

```suggestion
	assert.Equal(t, 1, tsm.timesScrapeCalled, "Must have been called as soon as the controller started")
```

astencel-sumo (approved):
I love this enhancement.

I wonder if/what changes we need to make in contrib to make sure this change does not surprise users. I noticed that the ScraperControllerSettings struct is often initialized like this:

```
ScraperControllerSettings: scraperhelper.ScraperControllerSettings{
    CollectionInterval: defaultCollectionInterval,
},
```

In those cases, the InitialDelay will be initialized to `0`, right? Should we make sure is it initializes with the scraper controller's default value of 1 second?

MovieStoreGuy:
> I love this enhancement.
> 
> I wonder if/what changes we need to make in contrib to make sure this change does not surprise users. I noticed that the ScraperControllerSettings struct is often initialized like this:
> 
> ```
> ScraperControllerSettings: scraperhelper.ScraperControllerSettings{
>     CollectionInterval: defaultCollectionInterval,
> },
> ```
> 
> In those cases, the InitialDelay will be initialized to `0`, right? Should we make sure is it initializes with the scraper controller's default value of 1 second?

Mmm, this is a good point.

Let me go find those receivers referencing the struct directly and I will update them to use the function so it shouldn't come as a _complete_ surprise.

MovieStoreGuy:
@bogdandrutu , @mx-psi , @astencel-sumo 

This change should be safe to merge in now without any issues.

dmitryax (approved):

astencel-sumo (approved):
This is a great enhancement!

We're going to need to update all the affected contrib components' README files with the new `initial_delay` option, I think it's going to be done in this PR:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/23030.

MovieStoreGuy:
> This is a great enhancement!
> 
> We're going to need to update all the affected contrib components' README files with the new `initial_delay` option, I think it's going to be done in this PR:
> 
> * [[chore] Updates to scrape duration opentelemetry-collector-contrib#23030](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/23030).

Should be done :)

djaglowski (approved):

dmitryax (approved):

Issue #7632, opened 2023-05-05
[confighttp] Support snappy and zstd compression on the server side


swiatekm-sumo:
We should support the same compression methods on the client and server sides of confighttp. Right now we have:

* Client side: `gzip`, `zlib`, `deflate`, `zstd`, `snappy`
* Server side: `gzip`, `zlib`, `deflate`

If there isn't a good reason we don't support zstd and snappy on the server side, then I can work on adding them.

atoulme:
Please take a look at https://github.com/open-telemetry/opentelemetry-collector/pull/7636

tigrannajaryan:
Ideally we also want a compression method negotiation added to OTLP spec so that we can have an "auto" setting.

atoulme:
How would that work? Would you use a similar mechanism as Accept-Encoding? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

tigrannajaryan:
> How would that work? Would you use a similar mechanism as Accept-Encoding? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

Yes, with 2-step negotiation for the server's accepting side. The initial client request should be in old encoding until it gets a response from the server about acceptance of the new encoding. Client should also be ready to go back (can happen in load-balanced deployments, mid new version rollout).

atoulme:
That sounds like it warrants its own issue. I will create one and link it here.

swiatekm-sumo:
`zstd` was added in #7927, but snappy is still missing.

swiatekm-sumo:
Looks like this can be closed with #8983.

PR #7636, opened 2023-05-08
[config/confighttp] Support zstd and snappy decompress

zdyj3170101136:
fix https://github.com/open-telemetry/opentelemetry-collector/issues/7632
**Description:** 
bug: the otlphttp exporter zstd and snappy compression is not work.
how to fix: add zstd and snappy decompress reader.

### DIFF
--- a/config/confighttp/compression.go
+++ b/config/confighttp/compression.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/golang/snappy"
 	"github.com/klauspost/compress/zstd"
+	"go.uber.org/multierr"
 
 	"go.opentelemetry.io/collector/config/configcompression"
 )
@@ -146,6 +147,7 @@ func (d *decompressor) wrap(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		newBody, err := newBodyReader(r)
 		if err != nil {
+			r.Body.Close()
 			d.errorHandler(w, r, err.Error(), http.StatusBadRequest)
 			return
 		}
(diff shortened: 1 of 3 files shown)


config/confighttp/compression.go
```diff
@@ -165,13 +165,22 @@ func (d *decompressor) wrap(h http.Handler) http.Handler {
 
 func newBodyReader(r *http.Request) (io.ReadCloser, error) {
 	switch r.Header.Get("Content-Encoding") {
-	case "gzip":
+	case string(configcompression.Gzip):
```
atoulme, line 168: do you need to convert every time? Is that needed?

config/confighttp/compression.go
```diff
@@ -165,13 +165,22 @@ func (d *decompressor) wrap(h http.Handler) http.Handler {
...
-	case "deflate", "zlib":
+	case string(configcompression.Snappy):
+		sr := snappy.NewReader(r.Body)
+		return io.NopCloser(sr), nil
+	case string(configcompression.Zstd):
+		zr, err := zstd.NewReader(r.Body)
+		if err != nil {
+			return nil, err
+		}
+		return zr.IOReadCloser(), nil
```
atoulme, line 182: Please add tests for those new compression encodings.

atoulme:
Please review and sign the CLA.

atoulme:
Please make sure to link to an issue. https://github.com/open-telemetry/opentelemetry-collector/pull/7632 seems related to this issue.

config/confighttp/compression.go
swiatekm-sumo: I don't think these enums are supposed to represent HTTP content encodings in the abstract anyway? I'd prefer to either keep the existing strings, or add new constants specifically for the encodings, or a `GetHTTPEncoding` method on `CompressionType` which would return the string directly.

config/confighttp/compression.go
```diff
@@ -146,6 +147,7 @@ func (d *decompressor) wrap(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		newBody, err := newBodyReader(r)
 		if err != nil {
+			r.Body.Close()
```
bogdandrutu, line 150: This is a separate issue, would be nice if we fix this in a separate PR.

config/confighttp/compression.go
```diff
@@ -146,6 +147,7 @@ func (d *decompressor) wrap(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		newBody, err := newBodyReader(r)
```
bogdandrutu, line 148: Not sure I understand why we close the newBody on line 155. @codeboten @dmitryax do you have any idea? Is that unnecessary and possible bug? If the server wants to hold a reference to the Body and call delayed, we should not call it for them.

config/confighttp/compression.go
zdyj3170101136: as @atoulme  said, this pr would also test the compression reader close the real http.body.
so i add a closeOnceReader in TestHTTPContentDecompressionHandler.
and i need line 150 to pass the test.
截屏2023-05-15 上午10 10 48

bogdandrutu: But I don't see any check that it is really closed (in the tests). I do see something that you return error on multiple calls (but the return error is ignored on line 150 and 155 so that check is not useful.

zdyj3170101136: > But I don't see any check that it is really closed (in the tests). I do see something that you return error on multiple calls (but the return error is ignored on line 150 and 155 so that check is not useful.

in compression_test, i implement a closeOnceReader, which would return net.ErrClosed when closed multiple time.
and in line 269, i check the real http.body is already closed whenever the compression reader create successfully or not.

config/confighttp/compression.go
```diff
@@ -146,6 +147,7 @@ func (d *decompressor) wrap(h http.Handler) http.Handler {
...
 		if err != nil {
+			r.Body.Close()
 			d.errorHandler(w, r, err.Error(), http.StatusBadRequest)
 			return
 		}
```
bogdandrutu, line 153: Is there not a bug at line 155? Why closing the newBody?

atoulme:
Please resolve conflicts.

config/confighttp/compression.go
atoulme: Please address this comment?

PR #7638, opened 2023-05-08
[chore] Prepare release v1.0.0-rcv0011/v0.77.0

opentelemetrybot:
The following commands were run to prepare this release:
- make chlog-update VERSION=v1.0.0-rcv0011/v0.77.0
- make prepare-release GH=none PREVIOUS_VERSION=0.76.1 RELEASE_CANDIDATE=0.77.0 MODSET=beta


CHANGELOG.md
```diff
@@ -4,6 +4,27 @@
 
 <!-- next version -->
 
+## v1.0.0-rcv0011/v0.77.0
```
codeboten, line 7: This should probably only show v0.77.0

codeboten (approved):
The bit that creates the description for the PR mentions `GH=none` which is a bit confusing, doesn't have to be fixed in this PR, but maybe as a separate item

CHANGELOG.md
dmitryax: Why do you think so? This is what we will show in the release. Maybe repeating the stable version is good to highlight the latest version?

codeboten: I would find it confusing as an end user if `v1.0.0-rcv0011` is called out on a release that doesn't include any changes for that version. Then I would have to go back until the earliest mention of that version to find what i'm looking for

TylerHelmuth: This is hard.  I agree that listing the same version multiple places make it harder to find the actual changes associated with it, but also having the changelog synced with the Release also makes sense.  

I trend towards dropping the unchanged RC candidate from the changelog so that the changelog stays exactly accurate with what changed in that release.

dmitryax: Ok, sounds good. Let's drop the stable version here. I'll do it manually for now, and will update the automation after

dmitryax: Ok, I'll update the automation rather than dealing with the updating this branch https://github.com/open-telemetry/opentelemetry-collector/pull/7640

PR #7644, opened 2023-05-08
[wip] add support for console/otlp metric exporter

### LINKED ISSUE #7641: Support additional exporters for exporting collector metrics

codeboten:
This is a prototype for adding support for metric readers to service.telemetry.metrics. The parsing code was generated using the json schema generator go-jsonschema, with the source schema in this PR: https://github.com/open-telemetry/opentelemetry-configuration/pull/15

There are many open questions still. For this to move forward, i propose it should be behind a separate feature gate to allow users who want to experiment with it to do so.

Fix #7641


service/telemetry.go
```diff
@@ -118,18 +98,71 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings component
...
+	if !tel.useOtel {
+		// when still using opencensus, return here as there's no need to configure readers below
+		_, err := tel.initPrometheus(res, settings.Logger, cfg.Metrics.Address, cfg.Metrics.Level, asyncErrorChannel)
+		if err != nil {
+			return err
+		}
+	}
+
+	if len(cfg.Metrics.Address) > 0 {
+		settings.Logger.Warn("service.telemetry.metrics.address is being deprecated in favor of service.telemetry.metrics.metric_readers")
```
TylerHelmuth, line 110: Do we need to somehow feature gate this behavior where the user has enabled the UseOTel feature gate but hasn't started using the Reader config yet?  Sounds like we eventually we want users to configure the Readers and won't support `address`, but that would be a breaking change.

service/internal/proctelemetry/config.go
```diff
@@ -0,0 +1,244 @@
...
+
+func newOTLPGRPCMetricExporter(ctx context.Context, args map[string]interface{}) (sdkmetric.Exporter, error) {
+	opts := []otlpmetricgrpc.Option{}
+	for k, v := range args {
+		switch k {
+		case "endpoint":
+			opts = append(opts, otlpmetricgrpc.WithEndpoint(fmt.Sprintf("%s", v)))
+		case "certificate":
+		case "client_key":
+		case "client_certificate":
```
TylerHelmuth, line 89: Do we need these cases if we don't do anything with them?

service/internal/proctelemetry/config.go
```diff
@@ -0,0 +1,244 @@
...
+	for k, v := range in {
+		out[k] = fmt.Sprintf("%v", v)
+	}
+	return out
+}
+
+func newOTLPGRPCMetricExporter(ctx context.Context, args map[string]interface{}) (sdkmetric.Exporter, error) {
+	opts := []otlpmetricgrpc.Option{}
+	for k, v := range args {
+		switch k {
```
TylerHelmuth, line 84: IDK what the configuration file spec looks like, but if a user configures an unsupported value should we error?

service/telemetry/generated_schema.go
```diff
@@ -0,0 +1,176 @@
+// Code generated by github.com/atombender/go-jsonschema, DO NOT EDIT.
```
TylerHelmuth, line 1: What command generated this?  Do we need it in the makefile?

TylerHelmuth (commented):
I think the pattern in `proctelemetry` is reasonable given the Go SDK doesn't support this type of configuration input yet.  For our implementation of the mapping _maybe_ generics could be used somewhere, but it probably isn't worth it.

service/telemetry.go
codeboten: Yup, I'm planning on putting this all behind this feature gate: https://github.com/open-telemetry/opentelemetry-collector/pull/7678

service/internal/proctelemetry/config.go
codeboten: These were here as a reminder we need to do something with them :) once i split this PR into smaller ones, i'll fill in these details

service/internal/proctelemetry/config.go
codeboten: i think it would make sense for an error to be returned for unsupported values

service/telemetry/generated_schema.go
codeboten: yup! will add go-jsonschema to the internal tools in a separate PR

service/internal/proctelemetry/config.go
```diff
@@ -0,0 +1,244 @@
...
+	HTTPUnacceptableKeyValues = []attribute.KeyValue{
+		attribute.String(semconv.AttributeNetHostName, ""),
+		attribute.String(semconv.AttributeNetHostPort, ""),
+	}
+)
+
+func toStringMap(in map[string]interface{}) map[string]string {
+	out := map[string]string{}
+	for k, v := range in {
+		out[k] = fmt.Sprintf("%v", v)
```
MovieStoreGuy, line 76: Would it be easier to call `fmt.Sprint` instead?

service/internal/proctelemetry/config.go
```diff
@@ -0,0 +1,244 @@
...
+	opts := []otlpmetrichttp.Option{}
+	for k, v := range args {
+		switch k {
+		case "endpoint":
+			opts = append(opts, otlpmetrichttp.WithEndpoint(fmt.Sprintf("%s", v)))
+		case "certificate":
+		case "client_key":
+		case "client_certificate":
+		case "compression":
+			switch fmt.Sprintf("%s", v) {
```
MovieStoreGuy, line 123: At this point you can't assert that `v` is a string due to args being a map of string to any, switching to `fmt.Sprint` should work without issue here.

service/internal/proctelemetry/config.go
```diff
@@ -0,0 +1,244 @@
...
+			case compressionGzip:
+				opts = append(opts, otlpmetrichttp.WithCompression(otlpmetrichttp.GzipCompression))
+			}
+		case "timeout":
+			timeout, ok := v.(int)
+			if !ok {
+				return nil, fmt.Errorf("invalid timeout for otlp exporter: %s", v)
```
MovieStoreGuy, line 130:
```suggestion
				return nil, fmt.Errorf("invalid timeout for otlp exporter: %v", v)
```

service/telemetry.go
```diff
@@ -118,18 +98,71 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings component
 		return err
 	}
 
-	return tel.initPrometheus(res, settings.Logger, cfg.Metrics.Address, cfg.Metrics.Level, asyncErrorChannel)
+	if !tel.useOtel {
+		// when still using opencensus, return here as there's no need to configure readers below
+		_, err := tel.initPrometheus(res, settings.Logger, cfg.Metrics.Address, cfg.Metrics.Level, asyncErrorChannel)
+		if err != nil {
+			return err
+		}
```
MovieStoreGuy, line 106: Should this return in general instead of just if an error occurred?

MovieStoreGuy (commented):
Just a few comments from me.

service/telemetry/generated_schema.go
codeboten: Added here https://github.com/open-telemetry/opentelemetry-collector/pull/7679

codeboten (commented):
Superseeded by https://github.com/open-telemetry/opentelemetry-collector/pull/8094 & https://github.com/open-telemetry/opentelemetry-collector/pull/8097

Issue #7672, opened 2023-05-11
add ability to create `connector.*Routers` for tests


mwear:
In order to test routing based connectors we need to be able to construct `connector.*Router`s in order to build connectors via their factory methods. The [connector internals](https://github.com/open-telemetry/opentelemetry-collector/blob/9ef6429bd0fa1eb101deed570d5177c0ecd31030/service/internal/graph/nodes.go#L252-L258) handles this for us in the real world. 

As suggested in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/21498#issuecomment-1542682841 we can move the fanoutconsumer from `service/internal` to `internal` and add methods to aid construction of `connector.*Router`s in `connectortest`.

Please assign this to me. I have PR to add this functionality that I will open shortly.

atoulme:
Can we close this issue as complete?

mwear:
We can!

PR #7673, opened 2023-05-11
Add helpers to construct `connector.*Router`s for tests

mwear:
**Description:** 
This PR adds helpers to `connectortest` to aid the construction of `connector.*Router`s for testing connectors. This was implemented based on @djaglowski's  [comment here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/21498#issuecomment-1542682841), with some slight modifications. I found it more ergonomic to pass the sink into the `WithTracesSink` (and similar) options. You usually want a handle on the sink after creating the router. While it's possible to get the sink out of the router after the fact, it's a little cumbersome.

These helpers will be useful for https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/21498 and future connectors that need use of routers.

For example, here is a test with the consumer passed in:

```go
func TestFanoutTracesWithSink(t *testing.T) {
    var sink0, sink1 consumertest.TracesSink

    tr, err := NewTracesRouterSink(
        WithTracesSink(component.NewIDWithName(component.DataTypeTraces, "0"), &sink0),
        WithTracesSink(component.NewIDWithName(component.DataTypeTraces, "1"), &sink1),
    )

    require.NoError(t, err)
    require.Equal(t, 0, sink0.SpanCount())
    require.Equal(t, 0, sink1.SpanCount())

    td := testdata.GenerateTraces(1)
    err = tr.(consumer.Traces).ConsumeTraces(context.Background(), td)

    require.NoError(t, err)
    require.Equal(t, 1, sink0.SpanCount())
    require.Equal(t, 1, sink1.SpanCount())
}

```

The same test having to extract the consumer out after the fact:

```go
func TestFanoutTracesWithSink(t *testing.T) {
    traces0 := component.NewIDWithName(component.DataTypeTraces, "0")
    traces1 := component.NewIDWithName(component.DataTypeTraces, "1")

    tr, err := NewTracesRouterSink(
        WithTracesSink(traces0),
        WithTracesSink(traces1),
    )
   
    require.NoError(t, err)
   
    cons0, _ := tr.Consumer(traces0)
    sink0 := cons0.(*consumertest.TracesSink)
    cons1, _ := tr.Consumer(traces1)
    sink1 := cons1.(*consumertest.TracesSink)

    require.Equal(t, 0, sink0.SpanCount())
    require.Equal(t, 0, sink1.SpanCount())

    td := testdata.GenerateTraces(1)
    err = tr.(consumer.Traces).ConsumeTraces(context.Background(), td)

    require.NoError(t, err)
    require.Equal(t, 1, sink0.SpanCount())
    require.Equal(t, 1, sink1.SpanCount())}
}
```

**Link to tracking Issue:** 
#7672

**Testing:**
Unit tests

**Documentation:**
Source code comments


connector/connectortest/fanoutconsumer.go
```diff
@@ -0,0 +1,108 @@
+// Copyright The OpenTelemetry Authors
```
djaglowski, line 1: I suggest we call this file `router.go`. Technically it wraps fanoutoutconsumer, but we're still only exposing it as a "router"

connector/connectortest/fanoutconsumer.go
```diff
@@ -0,0 +1,108 @@
...
+	"errors"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/connector"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/internal/fanoutconsumer"
+)
+
+var errTooFewConsumers = errors.New("fanoutconsumer requires a mininum of 2 consumers")
```
djaglowski, line 27:
```suggestion
var errTooFewConsumers = errors.New("router requires a minimum of 2 consumers")
```

connector/connectortest/fanoutconsumer.go
```diff
@@ -0,0 +1,108 @@
...
+
+type TracesRouterTestOption struct {
```
djaglowski, line 29:
```suggestion
type tracesRouterTestOption struct {
```

Can we get away with not exporting this?

djaglowski (commented):
Generally looks good to me. Just a couple nits.

connector/connectortest/fanoutconsumer.go
mwear: We can. I changed it because the linter was complaining with:

> unexported-return: exported func WithNopTracesSink returns unexported type connectortest.tracesRouterTestOption, which can be annoying to use (revive)

Should I add an override for the lint instead?

djaglowski: Let's leave it then. Thanks for confirming.

djaglowski (approved):
LGTM.

@bogdandrutu, do you care to review this?

connector/connectortest/router.go
```diff
@@ -0,0 +1,108 @@
...
+	return TracesRouterTestOption{id: id, cons: consumertest.NewNop()}
+}
+
+// WithTracesSink adds a consumer to a connector.TracesRouter
+func WithTracesSink(id component.ID, sink *consumertest.TracesSink) TracesRouterTestOption {
+	return TracesRouterTestOption{id: id, cons: sink}
+}
+
+// NewTracesRouterSink returns a connector.TracesRouter with sinks based on the options provided
+func NewTracesRouterSink(opts ...TracesRouterTestOption) (connector.TracesRouter, error) {
```
bogdandrutu, line 45: `NewTracesRouterSink` -> `NewTracesRouter`?

connector/connectortest/router.go
```diff
@@ -0,0 +1,108 @@
...
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/connector"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/internal/fanoutconsumer"
+)
+
+var errTooFewConsumers = errors.New("router requires a minimum of 2 consumers")
+
+type TracesRouterTestOption struct {
```
bogdandrutu, line 29: `TracesRouterTestOption` -> `TracesRouterOption`

connector/connectortest/router.go
```diff
@@ -0,0 +1,108 @@
...
+	cons consumer.Traces
+}
+
+// WithNopTracesSink creates a nop consumer for a connector.TracesRouter
+func WithNopTracesSink(id component.ID) TracesRouterTestOption {
+	return TracesRouterTestOption{id: id, cons: consumertest.NewNop()}
+}
+
+// WithTracesSink adds a consumer to a connector.TracesRouter
+func WithTracesSink(id component.ID, sink *consumertest.TracesSink) TracesRouterTestOption {
```
bogdandrutu, line 40: Why not just `WithTraces` and accept a `consumer.Traces`

connector/connectortest/router.go
```diff
@@ -0,0 +1,108 @@
...
+
+var errTooFewConsumers = errors.New("router requires a minimum of 2 consumers")
+
+type TracesRouterTestOption struct {
+	id   component.ID
+	cons consumer.Traces
+}
+
+// WithNopTracesSink creates a nop consumer for a connector.TracesRouter
+func WithNopTracesSink(id component.ID) TracesRouterTestOption {
```
bogdandrutu, line 35: `WithNopTracesSink` -> `WithNopTraces`

bogdandrutu (commented):
Same changes to all signal types.

connector/connectortest/router.go
djaglowski: This was my suggestion. The idea was to limit the usefulness of the TracesRouter to tests only. 

If we accept any `consumer.Traces`, then we might want to just export fanoutconsumer and let people build routers directly.

codeboten (approved):
This looks good to me, happy to merge once @bogdandrutu's confirmed his feedback has been addressed

connector/connectortest/router.go
bogdandrutu: May not be a bad idea (not the whole `fanoutconsumer`) only the router part correct? You don't need the `fanoutconsumer.NewLogs`.

mwear: Is there any reason why someone would need to create a router other than for testing purposes?

djaglowski: @bogdandrutu, WDYT?

bogdandrutu:
Alternative that I have in my mind is:

Make `TracesRouter` a proper struct and change the factory funcs like `CreateMetricsToTraces(ctx context.Context, set CreateSettings, cfg component.Config, router *TracesRouter) (Metrics, error)`


```goland
type TracesRouter struct {
...
}

func (*TracesRouter) Consumer(...component.ID) (consumer.Traces, error) {}
func (*TracesRouter) PipelineIDs() []component.ID {}
```

Main idea is that we don't expect alternative implementation of the interface, and we also solve the inconsistency between 1 vs multiple pipelines. Another benefit is that right now we will have 2 implementations of the interface (test and non-test).

djaglowski:
> change the factory funcs like `CreateMetricsToTraces(ctx context.Context, set CreateSettings, cfg component.Config, router *TracesRouter) (Metrics, error)`

I like the idea of being more explicit about what is being provided to the factory funcs but I think this signature implies that all connectors are routers and forces all connector authors to work with this new struct. Many (maybe most) connectors should not be concerned with routing and only need a "consumer.Traces" like all other pipeline components.

The other problem is that this would be a breaking change to the `connector` package, which I think we should avoid even more than _potential_ breaking changes to `connectortest`.

---

What if instead, we define a set of factory funcs specifically for routers? e.g. `CreateMetricsToTracesRouter(ctx context.Context, set CreateSettings, cfg component.Config, router *TracesRouter) (Metrics, error)` This exposes more surface area, but does so in a cleaner way and without breaking changes.

---

Another option is:
1. Solve the inconsistency problem by always passing in a fanout consumer, even if the consumer only emits to one pipeline. 
2. Accept this PR or something similar (e.g. `connectortest` exposes a more generalized `NewTracesRouter` function that we are more confident will not need to change in the future)

mwear:
I'm struggling to pick a clear winner between these options. I agree with @djaglowski that passing a router into the factory will be confusing for connectors that do not do routing (which is probably the majority use case). I like the option where we keep the factory functions as is, but always pass in a fanoutconsumer or the status quo. In both cases we'll need expose the necessary types for tests. I'm not completely opposed to adding additional factory functions for routers, but I think we still end up in the same situation where the connector internals will need to decide which factory function call, likely based on the number of consumers (as it is today). If we can agree on an approach, I'd be happy to see the work through.

djaglowski:
> 1. Solve the inconsistency problem by always passing in a fanout consumer, even if the consumer only emits to one pipeline.
> 2. Accept this PR or something similar (e.g. connectortest exposes a more generalized NewTracesRouter function that we are more confident will not need to change in the future)

This is my preferred approach. It's simple, non-breaking, and solves both problems. @bogdandrutu, can we move forward with this?

mwear:
@bogdandrutu, if you get a chance, could you take another look at this. It goes along with #7840 and solves all of the issues I ran into in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/21498.

codeboten (approved):
I believe all the concerns were addressed w/ the discussion around https://github.com/open-telemetry/opentelemetry-collector/pull/7840 and this PR from last week's SIG call.

PR #7678, opened 2023-05-15
add feature gate for supporting additional exporters

codeboten:
This adds the feature-gate `telemetry.useOtelWithExtendedConfigurationForInternalTelemetry` to start adding support for additional exporters for the collector's internal telemetry.

This is part of of https://github.com/open-telemetry/opentelemetry-collector/pull/7644

### DIFF
--- a/.chloggen/codeboten_add-featuregate-telemetry.yaml
+++ b/.chloggen/codeboten_add-featuregate-telemetry.yaml
@@ -5,7 +5,7 @@ change_type: enhancement
 component: service
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: "Add feature gate `telemetry.useOtelWithExtendedConfigurationForInternalTelemetry` to start adding support for exporting internal telemetry"
+note: "Add feature gate `telemetry.useOtelWithSDKConfigurationForInternalTelemetry` that will add support for configuring the export of internal telemetry to additional destinations in future releases"
 
 # One or more tracking issues or pull requests related to the change
 issues: [7641]
--- a/.chloggen/codeboten_add-featuregate-telemetry.yaml
+++ b/.chloggen/codeboten_add-featuregate-telemetry.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add feature gate `telemetry.useOtelWithExtendedConfigurationForInternalTelemetry` to start adding support for exporting internal telemetry"
+
+# One or more tracking issues or pull requests related to the change
+issues: [7641]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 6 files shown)


internal/obsreportconfig/obsreportconfig.go
```diff
@@ -39,6 +39,14 @@ var DisableHighCardinalityMetricsfeatureGate = featuregate.GlobalRegistry().Must
 	featuregate.WithRegisterDescription("controls whether the collector should enable potentially high"+
 		"cardinality metrics. The gate will be removed when the collector allows for view configuration."))
 
+// UseOtelWithExtendedConfigurationForInternalTelemetryFeatureGate is the feature gate that controls whether the collector
+// supports configuring the OpenTelemetry SDK via configuration
+var UseOtelWithExtendedConfigurationForInternalTelemetryFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"telemetry.useOtelWithExtendedConfigurationForInternalTelemetry",
```
TylerHelmuth, line 45: I like how the comment above calls out that this feature gate allows configuring the SDK, can the gate name also reflect that?  Maybe something like `telemetry.useOtelWithSDKConfigurationForInternalTelemetry`

service/telemetry/config.go
```diff
@@ -104,6 +104,18 @@ type LogsSamplingConfig struct {
 	Thereafter int `mapstructure:"thereafter"`
 }
 
+// MetricReader exposes configuration of metric readers to end users.
+// TODO: replace this temporary struct w/ auto-generated struct from jsonschema
+//
+// Experimental: *NOTE* this structure is subject to change or removal in the future.
```
TylerHelmuth, line 110: Is there anything from the configuration SIG we can link to from here?

internal/obsreportconfig/obsreportconfig.go
```diff
@@ -39,6 +39,14 @@ var DisableHighCardinalityMetricsfeatureGate = featuregate.GlobalRegistry().Must
...
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("controls whether the collector supports extended OpenTelemetry"+
+		"configuration for internal telemetry"))
+
```
TylerHelmuth, line 49: Is there any reference URL that would be useful here?

.chloggen/codeboten_add-featuregate-telemetry.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add feature gate `telemetry.useOtelWithExtendedConfigurationForInternalTelemetry` to start adding support for exporting internal telemetry"
```
TylerHelmuth, line 8: Probably worth noting that the gate doesn't do anything yet (unless you think you'll get in all the related PRs in the same release that this PR would go).

internal/obsreportconfig/obsreportconfig.go
codeboten: Updated

service/telemetry/config.go
codeboten: Added a link to the configuration repo, note the follow up PR adds the generated code

.chloggen/codeboten_add-featuregate-telemetry.yaml
codeboten: updated to note that configuration will be supported in future releases

TylerHelmuth (approved):

internal/obsreportconfig/obsreportconfig.go
codeboten: Not yet, i think once we have some configuration supported, i'll update the doc and add a link here

PR #7696, opened 2023-05-18
Customising service metrics with additional views, reader and server

ptodev:
There are two changes in the PR:
* Passing a custom metric reader and metric views to the `service.New` function. 
* Adding a configuration option to disable the Collector's `/metric` server.

This could be useful to users of the Collector who want to leverage the functionality of the service module, but who need to extend it with their own metric server, views and readers. 

The change in this PR was discussed recently on the CNCF slack channel for the collector, on [this](https://cloud-native.slack.com/archives/C01N6P7KR6W/p1684321333902779) thread.

Alternative solutions to this problem have already been discussed in [#4970](https://github.com/open-telemetry/opentelemetry-collector/issues/4970). So far they have mostly been turned down because overriding telemetry settings could contradict other settings in the service [config](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.68.0/service/config.go#L35).

There is one potential issue with this change. Unfortunately, the `useExternalMetricsServer` could indeed contradict the config. If the config specifies a telemetry metrics [address](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.68.0/service/telemetry/config.go#L118) and `useExternalMetricsServer = true` , then Collector wouldn't start a Prometheus server. However, if you think passing views and metric readers is ok, I could think of a way to minimise the contradictions which `useExternalMetricsServer` would introduce.

On Slack it was also mentioned that there is [work](https://github.com/open-telemetry/oteps/blob/6b62348b280b82036e897781d49ac4f2c3a9d95e/text/assets/0225-config.yaml#L310) underway to include metric readers and views in the config, and that this change could potentially contradict them (or at least provide another way to configure the same thing):
* I agree that if views are configurable through the config, the "views" part of this PR may not be necessary.
* Even if metric readers are configurable through config, there is still value in being able to pass custom ones as this change does, because then the Collector and the code outside of it could share the same Prometheus Registerer. Also, I don't think it would contradict the config in any way.

I would be happy to add a changelog entry, tests and documentation if needed. There is currently no GitHub issue created for this particular approach, but I would be happy to create one if needed as well.


ptodev:
Hello, if a maintainer has a chance to review and opine whether this approach has potential, it would be much appreciated please :)

mx-psi:
@open-telemetry/collector-maintainers any thoughts?

mx-psi:
ping @open-telemetry/collector-maintainers

Issue #7328, opened 2023-03-07
HTTPServerSettings: Support for additional headers


ledor473:
**Is your feature request related to a problem? Please describe.**
When deploying an OpenTelemetry Collector with `httpforwarder` or `healthcheckextension` behind a caching proxy or a CDN, the GET requests gets cached and there's currently no options to pass additional headers on HTTPServerSettings.  

Here are some links for various caching proxy that documents the headers that can be used to configure the cache behavior:
- https://www.varnish-software.com/developers/tutorials/http-caching-basics/
- https://docs.fastly.com/en/guides/caching-best-practices#cache-control-headers
- https://www.nginx.com/blog/nginx-caching-guide/

**Describe the solution you'd like**
Add a `headers` configuration to HTTPServerSettings. 
  
It would be similar to the `headers` property on HTTPClientSettings:
https://github.com/open-telemetry/opentelemetry-collector/blob/687d5b77926f55d2f8e20eddd3f184f5434dc5c3/config/confighttp/confighttp.go#L57-L60

**Describe alternatives you've considered**
I've looked at CORSSettings that is exposed but it only exposes `MaxAge` that ends up in Access-Control-Max-Age

atoulme:
@ledor473 would you like to work on this issue? I'd be happy to assign it to you.

ledor473:
If the proposal makes sense, I'd be happy to contribute it yes.

atoulme:
not sure, you can bring it to the SIG meeting for discussion or open a PR to provoke thoughts.

atoulme:
@ledor473 ready to close? This was shipped in 0.81.0.

ledor473:
Yep! I thought it would auto-close. Closing..

atoulme:
It closes only if in the PR you use the keyword `Fixes`, as in `Fixes #foo`

PR #7697, opened 2023-05-19
Add headers to HTTPServerSettings

ledor473:
**Description:**
Add `headers` configuration option on HTTPServerSettings. It allows for additional headers to be attached to each HTTP response sent to the client

**Link to tracking Issue:**
#7328

**Testing:**
I've added a test case that ensures the code still works with `nil` or an empty map along with one with a set of headers 

**Documentation:**
I've added comments in the struct, I imagine it will automatically be included in the doc?

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -219,7 +219,7 @@ type HTTPServerSettings struct {
 
 	// Additional headers attached to each HTTP response sent to the client.
 	// Header values are opaque since they may be sensitive.
-	Headers map[string]configopaque.String `mapstructure:"headers"`
+	ResponseHeaders map[string]configopaque.String `mapstructure:"response_headers"`
 }
 
 // ToListener creates a net.Listener.
@@ -297,8 +297,8 @@ func (hss *HTTPServerSettings) ToServer(host component.Host, settings component.
 		handler = cors.New(co).Handler(handler)
 	}
 
-	if hss.Headers != nil {
-		handler = headerHandler(handler, hss.Headers)
+	if hss.ResponseHeaders != nil {
+		handler = responseHeadersHandler(handler, hss.ResponseHeaders)
 	}
 
 	// Enable OpenTelemetry observability plugin.
@@ -325,7 +325,7 @@ func (hss *HTTPServerSettings) ToServer(host component.Host, settings component.
 	}, nil
 }
 
-func headerHandler(handler http.Handler, headers map[string]configopaque.String) http.Handler {
+func responseHeadersHandler(handler http.Handler, headers map[string]configopaque.String) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		h := w.Header()
 
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -216,6 +216,10 @@ type HTTPServerSettings struct {
 	// IncludeMetadata propagates the client metadata from the incoming requests to the downstream consumers
 	// Experimental: *NOTE* this option is subject to change or removal in the future.
 	IncludeMetadata bool `mapstructure:"include_metadata"`
+
+	// Additional headers attached to each HTTP response sent to the client.
+	// Header values are opaque since they may be sensitive.
+	Headers map[string]configopaque.String `mapstructure:"headers"`
 }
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -510,7 +510,7 @@ An entry into the changelog is required for the following reasons:
 - Changes to default settings
 - New components being added
 
-It is reasonable to omit an entry to the changelog under these circuimstances:
+It is reasonable to omit an entry to the changelog under these circumstances:
```
atoulme, line 513: Can you please open a separate PR for this typo?

ledor473: I mean, yes. But that seems a bit extra?

ledor473: Removed it from this PR

atoulme: That’s just how it’s done for this repo.

config/confighttp/confighttp.go
```diff
@@ -216,6 +216,10 @@ type HTTPServerSettings struct {
 	// IncludeMetadata propagates the client metadata from the incoming requests to the downstream consumers
 	// Experimental: *NOTE* this option is subject to change or removal in the future.
 	IncludeMetadata bool `mapstructure:"include_metadata"`
+
+	// Additional headers attached to each HTTP response sent to the client.
+	// Header values are opaque since they may be sensitive.
+	Headers map[string]configopaque.String `mapstructure:"headers"`
```
bogdandrutu, line 222: Should it be response_headers to clarify?

ledor473: I’m not opposed to it. Since HTTPClientSettings is using `headers` rather than `request _headers`, I went with that. It can be changed easily if that’s preferred.

bogdandrutu: Indeed we don't use `request_headers`, but the request headers are the most common, that's kind of my thinking... On the server I can see a functionality to say route by this headers (accept requests only if these request headers are present) so I can see some usages for "headers". @open-telemetry/collector-approvers any idea?

jpkrohling (changes requested):
A maintainer can override my block at any time, but I wanted to mention that the header setter extension could also be used for this purpose and is more powerful. It might be too much for this case, but it's worth considering it instead of adding a new feature here.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/headerssetterextension/README.md

cc @kovrus

ledor473:
@jpkrohling This PR is about the ability to configure extra headers to be returned by the OpenTelemetry Collector to a client.

When I looked at the headersetter extension, it appears that we can only use it to set extra headers on request made by the OpenTelemetry Collector.

Something like this:
```
                                           headersetter                                        
                                                |                                              
                                                |                                              
                                                |                                              
+--------------+    req    +--------------+    req    +--------------+                         
|              |-----------|              |-----------|              |                         
|    Client    |-----------|   OTel Col   |-----------|   Some API   |                         
+--------------+    resp   +--------------+    resp   +--------------+                         
                     |                                                                         
                     |                                                                         
                     |                                                                         
                  This PR                                                                                                                                   
```

config/confighttp/confighttp.go
ledor473: @bogdandrutu I'm not sure if the `@open-telemetry/collector-approvers` ping worked, but do let me know and I can revisit this

jpkrohling:
Yes, sorry!

config/confighttp/confighttp.go
dmitryax: I'd also lean towards `response_headers` even if it's inconsistent with the `HTTPClientSettings`'s `headers` (we can consider renaming them as well, but I'm not sure it's necessary). My reasoning is that request headers are more commonly discussed and come to mind first if referred by HTTP headers.

ledor473: Sounds good! I’ll make the change.

ledor473:
@bogdandrutu @dmitryax I've resolved the conversation above by changing it to `response_headers`. 

Let me know if there's anything else. Hopefully, we can work through the review and have it merged before it's closed by the `stale` label

dmitryax (approved):

ledor473:
@dmitryax @codeboten looks like it didn't make the cut for the 0.80 release. Is there any change we can have it merged for the next one?

Issue #7619, opened 2023-05-03
Document tlsconfig reload_interval data type


jroper:
The documentation for the tlsconfig `reload_interval` field is:

> Additionally certificates may be reloaded by setting the below configuration.
>
> * `reload_interval` (optional) : ReloadInterval specifies the duration after which the certificate will be reloaded. If not set, it will never be reloaded.

So, what do I actually put in it? `10`? Is that 10 seconds or 10 days? Or does it accept units, like `10s`, `10ns`, `10y`? And if units are allowed, are valid units documented anywhere? I'm guessing it uses the [Go `time.ParseDuration` function](https://pkg.go.dev/time#ParseDuration), but I only know that because I know OpenTelemetry uses Go and I'm familiar with Go. Importantly, `1d` is not valid for the Go parser, even though that's a pretty reasonable period to reload certificates that are renewed every month for example. I think it would help, if it hasn't already been documented, if somewhere in the OpenTelemetry documentation, the format for interval fields is documented, so that users don't have to read programming language reference documentation to find out, and also to link to it from documentation on places that use it.

astencel-sumo:
Fixed in #7700

PR #7700, opened 2023-05-20
document duration format

seankhliao:
**Description:** Document the format used for durations

**Link to tracking Issue:** #7619


djaglowski:
Since the format is defined by the `time` package, should we also link to an "official" description there? This is the best I could find: https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/time/format.go;l=1584-1588

seankhliao:
I've added links.
I considered it an implementation detail, but I see there are already other links into Go stdlib source.

codeboten (approved):
👍🏻

bogdandrutu (approved):

PR #7703, opened 2023-05-21
[scraper helper] Include support for timeout

MovieStoreGuy:
**Description:** 
This introduces a change that each scraper context is cancelled by the next collection period.

**Link to tracking Issue:** 
NA

**Testing:**
It requires integration tests against both core and contrib.

**Documentation:** 
Add comments to the collection interval to explain usage.

**Alternative**

I had considered adding a new field named `Timeout` which would require to be equal or less than collection interval.
However, since a lot of scrapers already have adopted a timeout fields, I worry about potential conflicts so I opted for this approach.

### DIFF
--- a/receiver/scraperhelper/scrapercontroller.go
+++ b/receiver/scraperhelper/scrapercontroller.go
@@ -19,23 +19,6 @@ import (
 	"go.opentelemetry.io/collector/receiver/scrapererror"
 )
 
-// ScraperControllerSettings defines common settings for a scraper controller
-// configuration. Scraper controller receivers can embed this struct, instead
-// of receiver.Settings, and extend it with more fields if needed.
-type ScraperControllerSettings struct {
-	CollectionInterval time.Duration `mapstructure:"collection_interval"`
-	InitialDelay       time.Duration `mapstructure:"initial_delay"`
-}
-
-// NewDefaultScraperControllerSettings returns default scraper controller
-// settings with a collection interval of one minute.
-func NewDefaultScraperControllerSettings(component.Type) ScraperControllerSettings {
-	return ScraperControllerSettings{
-		CollectionInterval: time.Minute,
-		InitialDelay:       time.Second,
-	}
-}
-
 // ScraperControllerOption apply changes to internal options.
 type ScraperControllerOption func(*controller)
 
@@ -189,7 +172,9 @@ func (sc *controller) startScraping() {
 		for {
 			select {
 			case <-sc.tickerCh:
-				sc.scrapeMetricsAndReport(context.Background())
+				ctx, done := context.WithTimeout(context.Background(), sc.collectionInterval)
+				sc.scrapeMetricsAndReport(ctx)
+				done()
 			case <-sc.done:
 				sc.terminated <- struct{}{}
 				return
--- a/receiver/scraperhelper/scrapercontroller.go
+++ b/receiver/scraperhelper/scrapercontroller.go
@@ -47,6 +47,7 @@ type controller struct {
 	logger             *zap.Logger
 	collectionInterval time.Duration
 	initialDelay       time.Duration
+	timeout            time.Duration
 	nextConsumer       consumer.Metrics
 
 	scrapers    []Scraper
@@ -91,6 +92,7 @@ func NewScraperControllerReceiver(
 		logger:             set.Logger,
 		collectionInterval: cfg.CollectionInterval,
 		initialDelay:       cfg.InitialDelay,
+		timeout:            cfg.Timeout,
 		nextConsumer:       nextConsumer,
 		done:               make(chan struct{}),
 		terminated:         make(chan struct{}),
@@ -172,7 +174,7 @@ func (sc *controller) startScraping() {
 		for {
 			select {
 			case <-sc.tickerCh:
-				ctx, done := context.WithTimeout(context.Background(), sc.collectionInterval)
+				ctx, done := context.WithTimeout(context.Background(), sc.timeout)
 				sc.scrapeMetricsAndReport(ctx)
 				done()
 			case <-sc.done:
--- a/receiver/scraperhelper/settings.go
+++ b/receiver/scraperhelper/settings.go
@@ -28,7 +28,7 @@ type ScraperControllerSettings struct {
 	CollectionInterval time.Duration `mapstructure:"collection_interval"`
 	// InitialDelay sets the initial start delay for the scraper,
 	// any non positive value is assumed to be immediately.
-	InitialDelay time.Duration `mapstructure:"intial_delay"`
+	InitialDelay time.Duration `mapstructure:"initial_delay"`
 	// Timeout is used to set scraper's context deadline, it must be within
 	// the range of (0, CollectionInterval]
 	Timeout time.Duration `mapstructure:"timeout"`
--- a/receiver/scraperhelper/settings.go
+++ b/receiver/scraperhelper/settings.go
@@ -14,7 +14,8 @@ import (
 )
 
 var (
-	ErrNonPositiveInterval = errors.New("requires positive value")
+	ErrNonPositiveInterval    = errors.New("requires positive value")
+	ErrTimeoutExceedsInterval = errors.New("timeout value exceeds collection interval")
 )
 
 // ScraperControllerSettings defines common settings for a scraper controller
@@ -26,6 +27,9 @@ type ScraperControllerSettings struct {
 	// to ensure that scrapers don't exceed the interval.
 	CollectionInterval time.Duration `mapstructure:"collection_interval"`
 	InitialDelay       time.Duration `mapstructure:"intial_delay"`
+	// Timeout is used to set scraper's context deadline, it must be within
+	// the range of (0, CollectionInterval]
+	Timeout time.Duration `mapstructure:"timeout"`
 }
 
 // NewDefaultScraperControllerSettings returns default scraper controller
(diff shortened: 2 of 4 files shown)


receiver/scraperhelper/scrapercontroller.go
```diff
@@ -177,7 +162,9 @@ func (sc *controller) startScraping() {
 		for {
 			select {
 			case <-sc.tickerCh:
-				sc.scrapeMetricsAndReport(context.Background())
+				ctx, done := context.WithTimeout(context.Background(), sc.collectionInterval)
```
MovieStoreGuy, line 165: I need to add more tests around this, I realised I only validate the collection interval is a positive value.

bogdandrutu:
> I had considered adding a new field named Timeout which would require to be equal or less than collection interval.
However, since a lot of scrapers already have adopted a timeout fields, I worry about potential conflicts so I opted for this approach.

I think if anything this is the right time to fix this. Do we know which use `Timeout`? If most uses timeout it will be a code conflict yes, but will be no change to the user.

MovieStoreGuy:
> > I had considered adding a new field named Timeout which would require to be equal or less than collection interval.
> > However, since a lot of scrapers already have adopted a timeout fields, I worry about potential conflicts so I opted for this approach.
> 
> I think if anything this is the right time to fix this. Do we know which use `Timeout`? If most uses timeout it will be a code conflict yes, but will be no change to the user.

I will need to take a look, I at least know of two components that I own that define their own timeout within the configuration.

MovieStoreGuy:
So running the following shell expression:

```shell
find . -name 'config.go' | xargs grep "mapstructure:\"timeout\"" | sort
```

The impacted components would be:
```
./aerospikereceiver/config.go:	Timeout                                 time.Duration                 `mapstructure:"timeout"`
./chronyreceiver/config.go:	Timeout time.Duration `mapstructure:"timeout"`
./collectdreceiver/config.go:	Timeout          time.Duration `mapstructure:"timeout"`
./dockerstatsreceiver/config.go:	Timeout time.Duration `mapstructure:"timeout"`
./memcachedreceiver/config.go:	Timeout time.Duration `mapstructure:"timeout"`
./mongodbreceiver/config.go:	Timeout                       time.Duration       `mapstructure:"timeout"`
./podmanreceiver/config.go:	Timeout time.Duration `mapstructure:"timeout"`
./zookeeperreceiver/config.go:	Timeout time.Duration `mapstructure:"timeout"`
```

codeboten (approved):
👍🏻 to the explicit timeout.

codeboten (commented):
Please address the conflicts

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,58 @@
...
+func NewDefaultScraperControllerSettings(component.Type) ScraperControllerSettings {
+	return ScraperControllerSettings{
+		CollectionInterval: time.Minute,
+		InitialDelay:       time.Second,
+		Timeout:            time.Minute,
+	}
+}
+
+func (set *ScraperControllerSettings) Validate() (errs error) {
+	if set.CollectionInterval < 1 {
```
bogdandrutu, line 48: Less than 1 what? Second?

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,58 @@
...
+		errs = multierr.Append(errs, fmt.Errorf(`"collection_interval": %w`, ErrNonPositiveInterval))
+	}
+	if set.Timeout < 1 {
```
bogdandrutu, line 51: Same, Timeout is duration you should compare with another duration.

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,58 @@
...
+	"time"
+
+	"go.uber.org/multierr"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+var (
+	ErrNonPositiveInterval    = errors.New("requires positive value")
+	ErrTimeoutExceedsInterval = errors.New("timeout value exceeds collection interval")
```
bogdandrutu, line 18: why public?

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// ScraperControllerSettings defines common settings for a scraper controller
+// configuration. Scraper controller receivers can embed this struct, instead
+// of receiver.Settings, and extend it with more fields if needed.
+type ScraperControllerSettings struct {
```
bogdandrutu, line 24: Separate issue to rename this to maybe `ScraperControllerConfig`? or just `Config`.

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,58 @@
...
+	// any non positive value is assumed to be immediately.
+	InitialDelay time.Duration `mapstructure:"initial_delay"`
+	// Timeout is used to set scraper's context deadline, it must be within
+	// the range of (0, CollectionInterval]
+	Timeout time.Duration `mapstructure:"timeout"`
+}
+
+// NewDefaultScraperControllerSettings returns default scraper controller
+// settings with a collection interval of one minute.
+func NewDefaultScraperControllerSettings(component.Type) ScraperControllerSettings {
```
bogdandrutu, line 39: The Type seem to not be used, a separate issue to remove it.

receiver/scraperhelper/settings.go
MovieStoreGuy: It is fine as is, it is wordy, but I don't believe there is value in changing the name since it is already adopted within Contrib.

receiver/scraperhelper/settings.go
MovieStoreGuy: Yeah, @astencel-sumo were talking about how it is redundant now due to `component.Type` not needing to be embedded into the config. 

What are your thoughts on having available options for setting the default collection interval and initial delay? This would save the awkward initialling of the value to set your component's default collection interval.

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (set *ScraperControllerSettings) Validate() (errs error) {
+	if set.CollectionInterval <= 0 {
+		errs = multierr.Append(errs, fmt.Errorf(`"collection_interval": %w`, errNonPositiveInterval))
+	}
+	if set.Timeout <= 0 {
+		errs = multierr.Append(errs, fmt.Errorf(`"timeout": %w`, errNonPositiveInterval))
+	}
+	if set.Timeout > set.CollectionInterval {
+		errs = multierr.Append(errs, errTimeoutExceedsInterval)
```
astencel-sumo, line 55: This means that when a user of a scraping component wants to decrease the `collection_interval`, they will also _have_ to decrease the `timeout` accordingly. For example this configuration will not work:

```yaml
receivers:
  hostmetrics:
    collection_interval: 10s
    scrapers:
      memory:
```

Because user will also need to set the `timeout` accordingly:

```yaml
receivers:
  hostmetrics:
    collection_interval: 10s
    timeout: 10s
    scrapers:
      memory:
```

I'm not in favor of such an API.

How about instead of erroring out when the timeout is higher than the collection interval, we only log a warning and adjust the timeout to collection interval automatically? Something like "warning: `timeout` is 1m, which is higher than `collection_interval` 10s, resetting `timeout` to 10s" or similar. What do you think?

bogdandrutu: Fail fast should be the right approach, especially because this can be caught in validating configuration. I am 100% in favor of erroring here instead of doing magic.

receiver/scraperhelper/settings.go
bogdandrutu: The value is for consistency with other places, since this is a very experimental package. Anyway if you don't want to do it it is ok, will do later.

receiver/scraperhelper/settings.go
bogdandrutu: I don't like that extra public API, if you want to set values, use `ScraperControllerSettings { Timeout: 5s, ...}` no need to use this. We should always avoid extra unnecessary public APIs, because that means extra maintenance for us.

bogdandrutu (approved):

receiver/scraperhelper/settings.go
astencel-sumo: I can understand that the "magic" approach has its disadvantages. But I still don't think it's a good API design to force a user to change two properties (`collection_interval` and `timeout`) when they only want to change one (`collection_interval`).

Also in the current state, this change that is an `enhancement`, will be a breaking change in contrib, as it will break existing user configurations for scraping components. This configuration:

```yaml
receivers:
  hostmetrics:
    collection_interval: 10s
    scrapers:
      memory:
```

is valid in current versions of contrib, but it will be invalid after this change is included in contrib. Users will need to supplement their configurations with `timeout` property reflecting the value of `collection_interval`.

I believe we should try harder to find a solution that allows this feature to exist without troubling users.

Perhaps another approach is to have a special (default) value that means "make timeout equal to collection interval"?

receiver/scraperhelper/settings.go
astencel-sumo: > I don't like that extra public API, if you want to set values, use ScraperControllerSettings { Timeout: 5s, ...} no need to use this. 

My first thought was to disagree, but thinking about this more, the `NewDefaultScraperControllerSettings` function defines a default set of settings for the ScraperController that is common to all components. But what is the value of having such a common default set of settings? Shouldn't it be each component defining its own default settings? 🤔 

I wonder if this could be extrapolated to other common functionalities - can we remove exporter helper's [NewDefaultQueueSettings](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.80.0/exporter/exporterhelper/queued_retry.go#L48), [NewDefaultRetrySettings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L258), [NewDefaultTimeoutSettings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/common.go#L24)? Can we remove [NewDefaultHTTPClientSettings](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.80.0/config/confighttp/confighttp.go#L82)?

receiver/scraperhelper/settings.go
MovieStoreGuy: > Perhaps another approach is to have a special (default) value that means "make timeout equal to collection interval"?

I would prefer not to have special values defining if a timeout should default to the collection interval.

I think a reasonable alternative is if the zero value is used (since zero is not a valid timeout), it can use the default interval.

I believe that satisfies the requirement of not having a scraper exceed its collection interval and listening to the timeout?

Does that work for you @astencel-sumo ?

djaglowski (commented):
I strongly disagree with this being the default behavior. Especially I think we should not _require_ this behavior. This will cause much unnecessary fragility.

I propose that a timeout is optional, and that if there is a default, it should be only kill a collection attempt when it is running excessively long (N * collection interval).

PR #7708, opened 2023-05-22
Fix start/stop logic for batch processor

bogdandrutu:


processor/batchprocessor/batch_processor.go
```diff
@@ -182,15 +182,11 @@ func (bp *batchProcessor) Capabilities() consumer.Capabilities {
 
 // Start is invoked during service startup.
 func (bp *batchProcessor) Start(context.Context, component.Host) error {
-	bp.goroutines.Add(1)
```
bogdandrutu, line 185: No go-routine started, no need to Add then Done in Shutdown, which will cause issues if Shutdown is called without Start for example.

jmacd (approved):

dmitryax (approved):

PR #7714, opened 2023-05-23
Change multiBatcher to use sync.Map, avoid global lock on fast path

bogdandrutu:

PR #7717, opened 2023-05-23
add benchmark test for multi batch processing

codeboten:
This refactors the batch processor tests to make testing multiple configuration scenarios easier. It adds a testcase for a batch processor w/ metadata keys.

### DIFF
--- a/processor/batchprocessor/batch_processor_test.go
+++ b/processor/batchprocessor/batch_processor_test.go
@@ -611,6 +611,20 @@ func BenchmarkBatchMetricProcessor(b *testing.B) {
 		Timeout:       100 * time.Millisecond,
 		SendBatchSize: 2000,
 	}
+	runMetricsProcessorBenchmark(b, cfg)
+}
+
+func BenchmarkMultiBatchMetricProcessor(b *testing.B) {
+	b.StopTimer()
+	cfg := Config{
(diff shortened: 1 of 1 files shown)


processor/batchprocessor/batch_processor_test.go
```diff
@@ -611,6 +611,20 @@ func BenchmarkBatchMetricProcessor(b *testing.B) {
...
+}
+
+func BenchmarkMultiBatchMetricProcessor(b *testing.B) {
+	b.StopTimer()
+	cfg := Config{
+		Timeout:       100 * time.Millisecond,
+		SendBatchSize: 2000,
+		MetadataKeys:  []string{"test", "test2"},
+	}
+	runMetricsProcessorBenchmark(b, cfg)
```
bogdandrutu, line 624: We should change the `runMetricsProcessorBenchmark` to use `b.RunParralel(func(pb *testing.PB) {...})`, because for example in case of https://github.com/open-telemetry/opentelemetry-collector/pull/7714 that was the main issue.

bogdandrutu (approved):

processor/batchprocessor/batch_processor_test.go
codeboten: Switched the method to use RunParallel. PTAL

Issue #7769, opened 2023-05-24
Should the logging exporter be renamed?


codeboten:
Please vote in the comment below https://github.com/open-telemetry/opentelemetry-collector/issues/7769#issuecomment-1601379447

atoulme:
Let's rename it before we mark the component away from development - See #7768

codeboten:
Fwiw, we [moved away](https://github.com/open-telemetry/opentelemetry-collector/commit/2363b528407133471a9d32468b1fe5286932fd1f) from from loglevel back in october and users are still running into issues w/ this due to documentation floating around that uses the incorrect config. 3 releases may not be enough for this change

astencel-sumo:
I agree `logging` is not a good name for this exporter. This issue has been raised before, last time in Collector SIG meeting on [December 21, 2022](https://docs.google.com/document/d/1r2JC5MB7GupCE7N32EwGEXs9V_YIsPgoFiLP4VWVMkE/edit#heading=h.tdupcqrlnbe5). Here are two related issues:
- https://github.com/open-telemetry/opentelemetry-collector/issues/3524 from June 2021
- https://github.com/open-telemetry/opentelemetry-specification/issues/1931 from September 2021

If we do decide to rename, does it mean that rather than strictly renaming, we would need to create a new exporter under a different name that would be a copy of the current `logging` exporter, and deprecate the `logging` exporter? Possibly keeping the deprecated one "forever", or at least for a year.

codeboten:
@astencel-sumo we could add a method to the logging exporter to make it instantiable under a different configuration option, something like https://github.com/open-telemetry/opentelemetry-collector/pull/7773

DewaldDeJager:
> 1. Deprecate the go.opentelemetry.io/collector/exporter/loggingexporter module and add some mechanism in ocb to notify users about deprecated modules

I agree with this approach. I'm not sure I agree with wrapping one exporter from the other (as per discussions in the SIG meeting). It feels similar to when the `spanmetrics` exporter was deprecated and replaced by the `spanmetrics` connector. The deprecated component should have no further development and users should be nudged to switch to the new component. The downside of this is that the code will be duplicated initially.

I also agree that it's a good idea for `ocb` to log warnings when trying to include any deprecated modules so that the distribution maintainers can look at replacing it. We should consider logging a warning about any deprecated components when the Collector starts if we don't have that already. This will help the distribution maintainers as their end-users will be nudged to move over to the new component.

[...]

PR #7773, opened 2023-05-25
Add 'debug' exporter

codeboten:
This PR does the following:

* make a copy of exporter/loggingexporter -> exporter/debugexporter
* adds a deprecation notice to loggingexporter
* updates documentation + examples to use `debug` instead of `logging`

I would suggest to move forward w/ this, we could release v0.81.0 w/ both exporters, and then as of v0.82.0, we could only release the debug exporter, but including the logging exporter at v0.81.0 in the release file. Any thoughts?

FIxes https://github.com/open-telemetry/opentelemetry-collector/issues/7769


bogdandrutu (commented):
Alternative, implement a https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/converter.go#L12 that converts the config `exporters::logging[/.*]?` to `exporters::debug[/.*]?`. Then we move the code to debug.

codeboten (commented):
> Alternative, implement a https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/converter.go#L12 that converts the config exporters::logging[/.*]? to exporters::debug[/.*]?. Then we move the code to debug.

@bogdandrutu Two questions around this:
- Would you end up creating a copy of the entire config map to do the replace? I couldn't find a way to delete a key other than by creating another map and omitting it
- This would also require changing any `logging/*` in the configured pipelines, would you still want to go this route?

exporter/debugexporter/README.md
```diff
@@ -0,0 +1,46 @@
+# Logging Exporter
```
djaglowski, line 1:
```suggestion
# Debug Exporter
```

codeboten: Addressed. I've also removed `loglevel` from the readme as the `debug` exporter may allow us to remove `loglevel` altogether

connector/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.19
 
 require (
 	github.com/stretchr/testify v1.8.4
+	go.opentelemetry.io/collector v0.80.0
```
jpkrohling, line 7: Why this change?

docs/troubleshooting.md
```diff
@@ -161,21 +161,21 @@ $ curl -X POST localhost:9411/api/v2/spans -H'Content-Type: application/json' -d
 You should see a log entry like the following from the Collector:
 
 ```
-2020-11-11T04:12:33.089Z	INFO	loggingexporter/logging_exporter.go:296	TraceExporter	{"#spans": 1}
+2020-11-11T04:12:33.089Z	INFO	debugexporter/debug_exporter.go:296	TraceExporter	{"#spans": 1}
```
jpkrohling, line 164: Call me paranoid, but can you bump the date to a date closer to today? It might be just me, but I would be really confused seeing this timestamp and no versions of the collector around that date having this component.

exporter/loggingexporter/factory.go
```diff
@@ -32,9 +32,9 @@ func NewFactory() exporter.Factory {
 	return exporter.NewFactory(
 		typeStr,
 		createDefaultConfig,
-		exporter.WithTraces(createTracesExporter, component.StabilityLevelDevelopment),
-		exporter.WithMetrics(createMetricsExporter, component.StabilityLevelDevelopment),
-		exporter.WithLogs(createLogsExporter, component.StabilityLevelDevelopment),
+		exporter.WithTraces(createTracesExporter, component.StabilityLevelDeprecated),
```
jpkrohling, line 35: I think the logging exporter will live on for quite a while, right? Would it make sense to get rid of the code for this exporter and embed the new exporter here? You could probably just return the debug exporter's outcome to NewFactory here, no?

codeboten: We'll need to release a new version of the exporter to mark it as "deprecated", so i figured i would remove the code on the next release.

I would remove the code completely and continue to include the module in the built distros, WDYT?

connector/go.mod
codeboten: i ran `make gotidy` and this happened

exporter/loggingexporter/factory.go
jpkrohling: > I would remove the code completely and continue to include the module in the built distros, WDYT?

Can you expand on that? Are you going to keep including the same version of the logging exporter? I don't think that's what you meant, but it's not clear to me what the plan is.

codeboten: > keep including the same version of the logging exporter?
This would be one option.

Another would be to go w/ the wrapper approach, i originally proposed this. A third option would be to do what @bogdandrutu suggested [here](https://github.com/open-telemetry/opentelemetry-collector/pull/7773#pullrequestreview-1447018212) and implement a converter.

I could go with any of those, but either ways we'll need both modules published for at least one version to deprecate the logging exporter

jpkrohling: > Another would be to go w/ the wrapper approach

That's what I also had in mind as well, I think it's the cleanest. The new exporter gets the code, the old exporter wraps the new exporter, so that it's code base is minimal and can continue to exist with almost no maintenance for a year before we delete it from the code base.

> we'll need both modules published for at least one version to deprecate the logging exporter

If the cut off date is July 2024, we should continue publishing new versions of the logging exporter until then. Wrapping the debug exporter within the logging exporter allows that with minimal overhead.

codeboten:
Reopening this PR. Followed the suggestion from @jpkrohling on aliasing the logging exporter as a debug exporter

exporter/loggingexporter/factory.go
codeboten: Finally getting back to this PR. Updated it to follow your suggestion, PTAL

codeboten:
Well apparently github didn't like me trying to re-open this PR so i opened a new one here: https://github.com/open-telemetry/opentelemetry-collector/pull/8375

Issue #7776, opened 2023-05-28
forwardconnector shouldn't set MutatesData: false


seankhliao:
**Describe the bug**
A filter processor in a second pipeline affects export data in a previous pipeline connected by forwardconnector

**Steps to reproduce**
Try running the config provided below multiple times, it sometimes works, sometimes breaks.

**What did you expect to see?**
Original pipeline unaffected by second pipeline, 
logging/1: 17 metrics
logging/2: 1 metric

log output:

```
2023-05-28T17:05:15.454+0100	info	service/telemetry.go:104	Setting up own telemetry...
2023-05-28T17:05:15.454+0100	info	service/telemetry.go:127	Serving Prometheus metrics	{"address": ":8888", "level": "Basic"}
2023-05-28T17:05:15.454+0100	info	exporter@v0.78.2/exporter.go:275	Development component. May change in the future.	{"kind": "exporter", "data_type": "metrics", "name": "logging/2"}
2023-05-28T17:05:15.456+0100	info	exporter@v0.78.2/exporter.go:275	Development component. May change in the future.	{"kind": "exporter", "data_type": "metrics", "name": "logging/1"}
2023-05-28T17:05:15.459+0100	info	connector/connector.go:636	Development component. May change in the future.	{"kind": "exporter", "name": "forward", "exporter_in_pipeline": "metrics", "receiver_in_pipeline": "metrics"}
2023-05-28T17:05:15.464+0100	info	service/service.go:131	Starting otelcol-custom...	{"Version": "1.0.0", "NumCPU": 16}
2023-05-28T17:05:15.464+0100	info	extensions/extensions.go:30	Starting extensions...
2023-05-28T17:05:15.464+0100	info	prometheusreceiver@v0.77.0/metrics_receiver.go:255	Starting discovery manager	{"kind": "receiver", "name": "prometheus_simple", "data_type": "metrics"}
2023-05-28T17:05:15.466+0100	info	prometheusreceiver@v0.77.0/metrics_receiver.go:243	Scrape job added	{"kind": "receiver", "name": "prometheus_simple", "data_type": "metrics", "jobName": "prometheus_simple/127.0.0.1:8888"}
2023-05-28T17:05:15.467+0100	info	service/service.go:148	Everything is ready. Begin running and processing data.
[...]

PR #7777, opened 2023-05-28
forwardconnector: mark as mutates data

seankhliao:
**Description:** Pessimistically mark the forward connector as `MutatesData: true`, as the connected output pipeline(s) may mutate

**Link to tracking Issue:** #7776

### DIFF
--- a/.chloggen/forward-copy.yaml
+++ b/.chloggen/forward-copy.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: forwardconnector
+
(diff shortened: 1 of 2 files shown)


djaglowski (commented):
I don't think there is a problem with the forward connector. It doesn't mutate data so the current setting is correct. Importantly, this setting controls whether or not the connector will _always_ make a copy of the data it forwards, which may have a major performance impact.

I think there must be a problem internally with the way the pipelines are built. Specifically, in the example config in #7776, the filter processor should receive data from a "fanout consumer" that is aware of it's ability to mutate data. I believe this must not be working as intended. I will look into this today.

djaglowski:
I've opened https://github.com/open-telemetry/opentelemetry-collector/pull/7779, which I believe addresses the problem correctly. Before I mark the PR as ready for review, I'll try to demonstrate the issue in unit tests.

djaglowski:
#7779 has been merged and should resolve the issue for all connectors. I'm closing this but please let us know if you find the problem unresolved.

PR #7802, opened 2023-05-31
[docs] fix code reference in configtls readme

krsmanovic:
**Documentation:** Fix code reference to upstream Go TLS package code line.

The link to upstream Golang crypto/tls code line is wrong, because it was changed since Go release 1.19.

I suggest to use commit reference instead of volatile master branch.

```
https://github.com/golang/go/blob/ed9db1d36ad6ef61095d5941ad9ee6da7ab6d05a/src/crypto/tls/common.go#L700
```

Alternative fix would be to use release branch reference (a bit riskier).

```
https://github.com/golang/go/blob/release-branch.go1.19/src/crypto/tls/common.go#L700
```

Let me know if you prefer this to be referenced using Go release branch so I change the PR.


codeboten (approved):
👍🏻

Issue #7646, opened 2023-05-09
`otlpreceiver` OTLP receiver is not throwing a valid `grpc-status` for authentication-related errors or in successful scenario


shalithakaru:
**Describe the bug**
OTLP receiver is not throwing a valid `grpc-status` for authentication-related errors.

**Steps to reproduce**
1. Enable the OTLP receiver configuration as below for the Collector 
(Please note you need this [extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/oidcauthextension) to enable authentication with the receiver)
```
receivers:
  otlp:
    protocols:
      grpc:
        auth:
          authenticator: oidc
      http: 
        auth: 
          authenticator: oidc
```

2. Send sample traces to the collector endpoints in each port using both HTTP and gRPC SDK implementations with a sample JWT token.
OTLP/HTTP is 4318
OTLP/gRPC is 4317

**What did you expect to see?**
`grpc-status:16` for unauthenticated requests and `grpc-status:0` for successful authentication.

**What did you see instead?**
HTTP will throw proper status codes like 401(Unauthorized) for errors and 200 for successful authentication.
But with gRPC, it will throw `grpc-status:2` for errors and nothing for successful authentication.

**What version did you use?**
Version: v0.76.1

**What config did you use?**
Config: N/A

**Environment**
N/A

**Additional context**
Found a temporary fix by customizing the this [line](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ca15907071b8d24e03f1ae03dfb71abe4ae68440/extension/oidcauthextension/extension.go#L48) for error scenarios
```
errNoAudienceProvided                = errors.New("no Audience provided for the OIDC configuration")
....
```
to 
```
import (
        ....
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
)

errNoAudienceProvided                = status.Error(codes.Unauthenticated, "failed to verify the token")
....
```

But I was not able to find a any solution for the successful scenario

[...]

PR #7805, opened 2023-06-01
Return Unauthenticated GRPC code on error

richard-bain:
**Description:**
Because `configgrpc` doesn't explicitly set an GRPC status code, `grpc-go` will return the `Unknown` code if there is an error. This doesn't match the spec and can lead to issues in clients that expect the "correct" status code and throw errors in the case of an "unknown" code. All GRPC authentication errors should use the `Unauthenticated (16)` code.

This updated behaviour matches [confighttp.go](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/confighttp.go#L346) that always returns `Unauthorized (401)` for errors.

See the [GRPC Status code spec](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) and the [GRPC <-> HTTP code mapping](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md).

**Link to tracking Issue:** #7646

**Testing:** 
- Updated unit tests to explicitly check for error codes and compare to GRPC error types.

**Documentation:** Updated `README.md` to include details on error codes

### DIFF
--- a/config/configgrpc/README.md
+++ b/config/configgrpc/README.md
@@ -107,3 +107,13 @@ see [confignet README](../confignet/README.md).
 - [`read_buffer_size`](https://godoc.org/google.golang.org/grpc#ReadBufferSize)
 - [`tls`](../configtls/README.md)
 - [`write_buffer_size`](https://godoc.org/google.golang.org/grpc#WriteBufferSize)
+
+## GRPC Status codes for authorization extensions
+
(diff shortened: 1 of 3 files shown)


richard-bain:
We may want to potentially throw away the error messages and force a response of `not authorized` or similar so that authentication details are not leaked by extensions under different failure scenarios.

Issue #7778, opened 2023-05-30
Triager role request for @astencel-sumo


astencel-sumo:
Hi, I'd like to help with triaging issues in this repo. I'm an approver for the [contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/) repo.

Just today I stumbled upon a couple issues that should be closed:
- https://github.com/open-telemetry/opentelemetry-collector/issues/7623
- https://github.com/open-telemetry/opentelemetry-collector/issues/7619

jpkrohling:
We commented on Slack already, but I also wanted to tell you that we have seen this request and are evaluating it. You should be hearing more about this soon.

PR #7807, opened 2023-06-01
Propose @astencel-sumo as triager

codeboten:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/7778


astencel-sumo (approved):
Thank you

PR #7840, opened 2023-06-06
Always pass fanout consumers to connector factory functions

mwear:
**Description:**
This is an implementation of https://github.com/open-telemetry/opentelemetry-collector/pull/7673#issuecomment-1572126437 which resolves some inconsistencies regarding fanout consumers and connectors. It updates the connector code to always pass a fanout consumer to connector factory functions (even if the connector emits to a single pipeline). This will likely pair with the work in #7673 to add helpers to construct connector routers (when needed) for tests.

**Link to tracking Issue:** #7672, #7673

**Testing:**
Unit testing

cc: @bogdandrutu @djaglowski

### DIFF
--- a/service/internal/graph/nodes.go
+++ b/service/internal/graph/nodes.go
@@ -238,7 +238,7 @@ func (n *connectorNode) buildComponent(
 
 	switch n.rcvrPipelineType {
 	case component.DataTypeTraces:
-		capability := nexts[0].Capabilities()
+		capability := consumer.Capabilities{MutatesData: false}
 		consumers := make(map[component.ID]consumer.Traces, len(nexts))
 		for _, next := range nexts {
 			consumers[next.(*capabilitiesNode).pipelineID] = next.(consumer.Traces)
@@ -271,7 +271,7 @@ func (n *connectorNode) buildComponent(
 		}
 
 	case component.DataTypeMetrics:
-		capability := nexts[0].Capabilities()
+		capability := consumer.Capabilities{MutatesData: false}
 		consumers := make(map[component.ID]consumer.Metrics, len(nexts))
 		for _, next := range nexts {
 			consumers[next.(*capabilitiesNode).pipelineID] = next.(consumer.Metrics)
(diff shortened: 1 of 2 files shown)


djaglowski:
Thanks for working this out @mwear.

> the concrete type of the consumer passed in to the factory functions could differ from the real world if the helpers are not used

I may be overlooking something but ideally this is where we are able to lean on the interfaces (`consumer.*` and `connector.*Router`). I think it would be reasonable for a test scenario to provide its own implementation of either interface, which might behave differently that the fanout consumer. The most important thing is whether or not the interface is satisfied, but beyond that test expectations should generally be concerned with whether or not data was emitted to the correct consumer. Very few tests should specifically rely on the implementation details of the fanout consumer - I think maybe only integration tests, which may mutate the data _after_ the connector has emitted it. 

If the above makes sense, I am curious whether we need to make all these changes in one PR. In other words, is there anything technically that prevents us from _only_ switching to always pass a fanout consumer into connector factory functions?

mwear:
For this PR the only changes we really need are those made in `graph` to always pass in a fanout consumer. Separately we will need helpers create a fanout consumer (aka router) for connectors that need them. I am probably over doing (over thinking) things by worrying about connectors potentially being built with a consumer implementation that could differ from what the service will provide in the real world. I agree in most situations just relying on the regular consumer interfaces should be fine. Given that, we do not need to make the changes in the forward_connector_test and the changes made in the example router test are optional.

I think we can pare this PR down to just the changes to graph and omit the changes to the tests and we can work on the helpers separately in #7673 (I think I prefer the original naming).

service/internal/graph/nodes.go
```diff
@@ -238,16 +238,13 @@ func (n *connectorNode) buildComponent(
 
 	switch n.rcvrPipelineType {
 	case component.DataTypeTraces:
-		next := nexts[0].(consumer.Traces)
 		capability := nexts[0].Capabilities()
```
bogdandrutu, line 241: start with false?

bogdandrutu:
overall I think this is a good PR, LGTM after fixing the comment.

service/internal/graph/nodes.go
mwear: @bogdandrutu: Could I ask you to clarify what you're asking for? Previously this code used to assign `next := nexts[0].(consumer.Traces)` and then reassign it to a fanoutconsumer `if len(nexts) > 1`. Now it unconditionally assigns next to a fanoutconsumer. Looking at the original code, it appears that nexts should always have a length of at least 1, is there a case that I am missing?

bogdandrutu: Just do:

```
capability := consumer.Capabilities{MutatesData: false}
```

bogdandrutu (approved):

bogdandrutu (approved):

PR #7852, opened 2023-06-07
[chore] remove unnecessary additional tools

codeboten:
This is replicating the setup in the contrib repo that doesn't use go-acc & gcovmerge for coverage. It uses tools built into go 1.20 instead


codeboten:
Will spend some time trying to figure out why the coverage dropped from this change.

codeboten:
> Will spend some time trying to figure out why the coverage dropped from this change.

Found the missing coverage, needed to pass in `coverpkg`

djaglowski (approved):

PR #7854, opened 2023-06-07
Add pipelines.Config to remove duplicate of the pipelines configuration

bogdandrutu:


service/config.go
```diff
@@ -27,61 +23,16 @@ type Config struct {
...
-
-	// Validate no processors are duplicated within a pipeline.
-	procSet := make(map[component.ID]struct{}, len(cfg.Processors))
-	for _, ref := range cfg.Processors {
-		// Ensure no processors are duplicated within the pipeline
-		if _, exists := procSet[ref]; exists {
-			return fmt.Errorf("references processor %q multiple times", ref)
-		}
-		procSet[ref] = struct{}{}
+		fmt.Printf("service::telemetry config validation failed: %v\n", err)
```
codeboten, line 35: this should return fmt.Errorf()

codeboten (approved):

service/config.go
bogdandrutu: I want to do it in a separate PR, since it changes previous behavior.

PR #7859, opened 2023-06-08
Avoid re-creating the encoders WriteCloser for every request

bogdandrutu:


config/confighttp/compressor.go
```diff
@@ -0,0 +1,78 @@
...
+
+		if closeErr != nil {
+			return closeErr
+		}
+	}
+
+	if err := writer.Close(); err != nil {
+		return err
+	}
+	p.pool.Put(writer)
```
dmitryax, line 76: Shouldn't we have `defer p.pool.Put(writer)` given that we have returns above?

bogdandrutu: I thought about this, but I don't know if it is safe to return a failed writer... If we don't return GC will collect this as any other object.

bogdandrutu: Tried to read some of the implementations gzip and snappy and seem to be fine...

dmitryax (approved):
LGTM

PR #7867, opened 2023-06-09
Mark fanout consumer error as permanent

jpkrohling:
Based on discussions from the SIG's call, I changed one of the fanout consumers to mark the returned error as permanent if at least one consumer succeeded and one failed, as we don't want retries to be performed on those situations.

Once this is done, #7486 will be changed to return a 500 in case the error it receives is a permanent error.


jpkrohling:
If the general approach from this PR looks good, I'll apply the same to the other signals.

PR #7868, opened 2023-06-09
[fanoutconsumer] Mark fanout consumer error as permanent

jpkrohling:
Based on discussions from the SIG's call, I changed one of the fanout consumers to mark the returned error as permanent if at least one consumer succeeded and one failed, as we don't want retries to be performed on those situations.

Once this is done, #7486 will be changed to return a 500 in case the error it receives is a permanent error.

Signed-off-by: Juraci Paixão Kröhling juraci@kroehling.de

### DIFF
--- a/.chloggen/fanoutconsumer-permanenterror.yaml
+++ b/.chloggen/fanoutconsumer-permanenterror.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: fanoutconsumer
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: The fanoutconsumer now returns a permanent error if at least one consumer succeeded and one failed.
+
+# One or more tracking issues or pull requests related to the change
+issues: [7868]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/service/internal/fanoutconsumer/traces.go
+++ b/service/internal/fanoutconsumer/traces.go
@@ -12,6 +12,7 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/connector"
 	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/consumer/consumererror"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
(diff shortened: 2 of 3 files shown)


astencel-sumo:
I'm not convinced that the fan-out consumer is the right place to arbitrarily decide whether a payload should be retried or not. The existing approach of the fan-out consumer collecting all errors from consumers and passing them back to the sender seems enough for the sender to be able to decide by itself whether to retry or not, perhaps based on the errors returned. For example this existing [OTLP receiver code](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.79.0/receiver/otlpreceiver/otlphttp.go#L37-L41):

```go
	otlpResp, err := tracesReceiver.Export(req.Context(), otlpReq)
	if err != nil {
		for 
		writeError(resp, encoder, err, http.StatusInternalServerError)
		return
	}
```

could be changed to something similar to this:

```go
	otlpResp, err := tracesReceiver.Export(req.Context(), otlpReq)
	if err != nil {
		isPermanent := false
		for _, individualConsumerError := range multierr.Errors(err) {
			if consumererror.IsPermanent(individualConsumerError) {
				isPermanent = true
				break
			}
		}
		if isPermanent {
			writeError(resp, encoder, err, http.StatusInternalServerError)
		} else {
			writeError(resp, encoder, err, http.StatusServiceUnavailable)
		}
		return
	}
```

jpkrohling:
I think I agree with @astencel-sumo, except that we don't need to go over all errors in multierr: `consumererror.IsPermanent` will use `IsA`, which will unwrap all errors in the chain. So, if anything in the multierr is a permanent error, the function will return true. 

I just confirmed this by reverting the traces.go from this PR to the latest state from main and changing the test to:

```go
	p1 := mutatingErr{Consumer: consumertest.NewErr(myErr)}
	p2 := consumertest.NewErr(consumererror.NewPermanent(myErr))
	p3 := new(consumertest.TracesSink)
```

One permanent error from at least one exporter is enough to have IsPermanent to return true. If you all agree with @astencel-sumo's approach, I think I'll change this PR to just include the new test, and go back to the otlp receiver doing something similar to @astencel-sumo's proposal.

bogdandrutu:
I think the problem was not finding if one pipeline permanently errored, but what to do in the cases like one pipeline succeed the other has a retriable error? Do we return to the client a retriable error and they can retry? That will cause duplicate data on one path, what do you do then?

astencel-sumo:
> I think the problem was not finding if one pipeline permanently errored, but what to do in the cases like one pipeline succeed the other has a retriable error? Do we return to the client a retriable error and they can retry? That will cause duplicate data on one path, what do you do then?

I agree it's important to discuss all possible scenarios. Let's discuss in the original PR about propagating errors if that's OK? [Propagate errors from exporters to receivers #7486](https://github.com/open-telemetry/opentelemetry-collector/pull/7486)

PR #7870, opened 2023-06-09
[chore]: Avoid private types in public API, remove options on internal funcs

bogdandrutu:


codeboten (approved):
Needs a rebase, otherwise looks good

codeboten (approved):

PR #7679, opened 2023-05-15
add auto-generated configuration code

codeboten:
This is generated from the opentelemetry-configuration repository using gojsonschema. Updated makefile to support this w/ `make genjsonschema`

Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/7678

PR #7871, opened 2023-06-09
[wip] add ability to configure prometheus export via `readers`

codeboten:
This enables end users to configure additional prometheus exporters for the collector's telemetry
via the `readers` configuration option. Configuring prometheus through the existing method
of setting the service::metrics::address will continue to work, and only log a warning for users
who have enabled the `telemetry.useOtelWithSDKConfigurationForInternalTelemetry` feature gate.

Linked issue: #7641

This is a follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/7679

### DIFF
--- a/service/internal/proctelemetry/config.go
+++ b/service/internal/proctelemetry/config.go
@@ -4,7 +4,17 @@
 package proctelemetry // import "go.opentelemetry.io/collector/service/internal/proctelemetry"
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/bridge/opencensus"
+	otelprom "go.opentelemetry.io/otel/exporters/prometheus"
 	"go.opentelemetry.io/otel/sdk/instrumentation"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/metric/aggregation"
@@ -12,6 +22,7 @@ import (
 
 	"go.opentelemetry.io/collector/obsreport"
 	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
+	"go.opentelemetry.io/collector/service/telemetry"
 )
 
 const (
@@ -21,6 +32,12 @@ const (
 
 	// http Instrumentation Name
 	HTTPInstrumentation = "go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
+	// supported metric readers
+	pullMetricReader = "pull"
+
+	// supported exporters
+	prometheueExporter = "prometheus"
 )
 
 var (
@@ -38,6 +55,20 @@ var (
 	}
 )
 
+func InitMetricReader(ctx context.Context, name string, reader any, asyncErrorChannel chan error) (sdkmetric.Reader, *http.Server, error) {
+	readerType := strings.Split(name, "/")[0]
+	switch readerType {
+	case pullMetricReader:
+		r, ok := reader.(telemetry.PullMetricReader)
+		if !ok {
+			return nil, nil, fmt.Errorf("invalid metric reader configuration: %v", reader)
+		}
+		return initExporter(ctx, r.Exporter, asyncErrorChannel)
+	default:
+		return nil, nil, fmt.Errorf("unsupported metric reader type: %s", readerType)
+	}
+}
+
 func InitOpenTelemetry(res *resource.Resource, options []sdkmetric.Option, disableHighCardinality bool) (*sdkmetric.MeterProvider, error) {
 	opts := []sdkmetric.Option{
 		sdkmetric.WithResource(res),
@@ -48,6 +79,21 @@ func InitOpenTelemetry(res *resource.Resource, options []sdkmetric.Option, disab
 	return sdkmetric.NewMeterProvider(
 		opts...,
 	), nil
+}
+
+func InitPrometheusServer(registry *prometheus.Registry, address string, asyncErrorChannel chan error) *http.Server {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{}))
+	server := &http.Server{
+		Addr:    address,
+		Handler: mux,
+	}
+	go func() {
+		if serveErr := server.ListenAndServe(); serveErr != nil && !errors.Is(serveErr, http.ErrServerClosed) {
+			asyncErrorChannel <- serveErr
+		}
+	}()
+	return server
 }
 
 func batchViews(disableHighCardinality bool) []sdkmetric.View {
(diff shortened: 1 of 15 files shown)


service/telemetry/config.go
```diff
@@ -137,9 +127,45 @@ type TracesConfig struct {
 func (c *Config) Validate() error {
 
 	// Check when service telemetry metric level is not none, the metrics address should not be empty
-	if c.Metrics.Level != configtelemetry.LevelNone && c.Metrics.Address == "" {
+	if c.Metrics.Level != configtelemetry.LevelNone && c.Metrics.Address == "" && len(c.Metrics.Readers) == 0 {
 		return fmt.Errorf("collector telemetry metric address should exist when metric level is not none")
```
TylerHelmuth, line 131: Is this message a little misleading when `c.Metrics.Readers` is the culprit?

service/telemetry/config.go
```diff
@@ -137,9 +127,45 @@ type TracesConfig struct {
...
 	}
 
 	return nil
 }
+
+func (mrs *MeterProviderJsonMetricReaders) Unmarshal(conf *confmap.Conf) error {
+	if err := conf.Unmarshal(mrs); err != nil {
+		return err
+	}
+	if *mrs == nil {
```
TylerHelmuth, line 141: Should this be `if mrs == nil`?

TylerHelmuth: Also, can we add a test case that can hit this? Is it even possible for `conf.Unmarshal` to not return an error but `mrs` be nil?

service/telemetry.go
```diff
@@ -86,44 +85,73 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings component
...
-		zap.String(zapKeyTelemetryAddress, address),
-		zap.String(zapKeyTelemetryLevel, level.String()),
-	)
+	metricproducer.GlobalManager().AddProducer(tel.ocRegistry)
+	opts := []sdkmetric.Option{}
+	for name, reader := range cfg.Metrics.Readers {
+		// @codeboten: server returned here only happens when a pull based metric
+		// reader is configured, this could be refactored to pass in a
+		// func to add the server to the list of servers. another thing that would
+		// be nice is not to have to pass down the asyncErrorChannel
```
mx-psi, line 129: Should we file an issue for this?

service/telemetry/config_test.go
```diff
@@ -37,6 +38,18 @@ func TestLoadConfig(t *testing.T) {
 			},
 			success: false,
 		},
+		{
+			name: "valid metric telemetry with metric readers",
+			cfg: &Config{
+				Metrics: MetricsConfig{
+					Level: configtelemetry.LevelBasic,
+					Readers: MeterProviderJsonMetricReaders{
+						"pull/prometheus": PullMetricReader{},
```
mx-psi, line 47: Can you add a test where we don't have a `/` in the name? What should happen then?

service/telemetry/config.go
```diff
@@ -137,9 +127,45 @@ type TracesConfig struct {
...
+			for key, exporter := range r.Exporter {
+				switch key {
+				case "prometheus":
+					var promExporter Prometheus
+					if err := mapstructure.Decode(exporter, &promExporter); err != nil {
+						return fmt.Errorf("invalid exporter configuration: %w", err)
+					}
+					r.Exporter[key] = promExporter
+				default:
+					return fmt.Errorf("unsupported metric exporter type: %s", key)
```
mx-psi, line 163: nit (for consistency with other errors such as https://github.com/open-telemetry/opentelemetry-collector/blob/bc9421c14b0e856b36d9ad9545e423bf8112f99f/config/configtelemetry/configtelemetry.go#L76)
```suggestion
					return fmt.Errorf("unsupported metric exporter type %q", key)
```

service/telemetry/config.go
```diff
@@ -137,9 +127,45 @@ type TracesConfig struct {
...
+				}
+			}
+		default:
+			return fmt.Errorf("unsupported metric reader type: %s", readerType)
```
mx-psi, line 167:
```suggestion
			return fmt.Errorf("unsupported metric reader type %q", readerType)
```

service/telemetry.go
```diff
@@ -86,44 +85,73 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings component
...
+func (tel *telemetryInitializer) initMetrics(res *resource.Resource, logger *zap.Logger, cfg telemetry.Config, asyncErrorChannel chan error) error {
+	// Initialize the ocRegistry, still used by the process metrics.
+	tel.ocRegistry = ocmetric.NewRegistry()
+	if !tel.useOtel && !tel.extendedConfig {
+		return tel.initOpenCensus(res, logger, cfg.Metrics.Address, cfg.Metrics.Level, asyncErrorChannel)
+	}
+
+	if len(cfg.Metrics.Address) != 0 {
+		if tel.extendedConfig {
+			logger.Warn("service.telemetry.metrics.address is being deprecated in favor of service.telemetry.metrics.metric_readers")
```
mx-psi, line 100: nit, use `::` as the separator (using `::` will actually work)
```suggestion
			logger.Warn("service::telemetry::metrics::address is being deprecated in favor of service::telemetry::metrics::metric_readers")
```

service/telemetry/config.go
```diff
@@ -121,7 +111,7 @@ type MetricsConfig struct {
 
 	// Readers allow configuration of metric readers to emit metrics to
 	// any number of supported backends.
-	Readers []MetricReader `mapstructure:"metric_readers"`
+	Readers MeterProviderJsonMetricReaders `mapstructure:"metric_readers"`
```
mx-psi, line 114: Maybe too late/this comes from the Configuration WG, but it seems a bit weird to me to have `metrics::metric_readers` instead of just `metrics::readers`

service/internal/proctelemetry/config.go
```diff
@@ -21,6 +32,12 @@ const (
 
 	// http Instrumentation Name
 	HTTPInstrumentation = "go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+
+	// supported metric readers
+	pullMetricReader = "pull"
+
+	// supported exporters
+	prometheueExporter = "prometheus"
```
mx-psi, line 40:
```suggestion
	prometheusExporter = "prometheus"
```

service/internal/proctelemetry/config.go
```diff
@@ -38,6 +55,20 @@ var (
...
+	readerType := strings.Split(name, "/")[0]
+	switch readerType {
+	case pullMetricReader:
+		r, ok := reader.(telemetry.PullMetricReader)
+		if !ok {
+			return nil, nil, fmt.Errorf("invalid metric reader configuration: %v", reader)
+		}
+		return initExporter(ctx, r.Exporter, asyncErrorChannel)
+	default:
+		return nil, nil, fmt.Errorf("unsupported metric reader type: %s", readerType)
```
mx-psi, line 68:
```suggestion
		return nil, nil, fmt.Errorf("unsupported metric reader type %q", readerType)
```

service/internal/proctelemetry/config.go
```diff
@@ -92,3 +138,44 @@ func cardinalityFilter(kvs ...attribute.KeyValue) attribute.Filter {
...
+	if prometheusConfig.Host == nil {
+		return nil, nil, fmt.Errorf("host must be specified")
+	}
+	if prometheusConfig.Port == nil {
+		return nil, nil, fmt.Errorf("port must be specified")
+	}
+	wrappedRegisterer := prometheus.WrapRegistererWithPrefix("otelcol_", promRegistry)
+	// We can remove `otelprom.WithoutUnits()` when the otel-go start exposing prometheus metrics using the OpenMetrics format
+	// which includes metric units that prometheusreceiver uses to trim unit's suffixes from metric names.
+	// https://github.com/open-telemetry/opentelemetry-go/issues/3468
```
mx-psi, line 153: Should we have a tracking issue for this on this repository (linking the one on opentelemetry-go)?

service/internal/proctelemetry/config.go
```diff
@@ -92,3 +138,44 @@ func cardinalityFilter(kvs ...attribute.KeyValue) attribute.Filter {
...
+	return exporter, InitPrometheusServer(promRegistry, fmt.Sprintf("%s:%d", *prometheusConfig.Host, *prometheusConfig.Port), asyncErrorChannel), nil
+}
+
+func initExporter(_ context.Context, exporters telemetry.MetricExporter, asyncErrorChannel chan error) (sdkmetric.Reader, *http.Server, error) {
+	for exporterType, exporter := range exporters {
+		switch exporterType {
+		case prometheueExporter:
+			e, ok := exporter.(telemetry.Prometheus)
+			if !ok {
+				return nil, nil, fmt.Errorf("prometheus exporter invalid: %v", exporter)
```
mx-psi, line 173: This error message is a bit confusing IMO, should we show the type (`%T`) of the `exporter` here and say the type we expected it to have?

service/internal/proctelemetry/config.go
```diff
@@ -92,3 +138,44 @@ func cardinalityFilter(kvs ...attribute.KeyValue) attribute.Filter {
...
+			}
+			return initPrometheusExporter(e, asyncErrorChannel)
+		default:
+			return nil, nil, fmt.Errorf("unsupported metric exporter type: %s", exporterType)
```
mx-psi, line 177:
```suggestion
			return nil, nil, fmt.Errorf("unsupported metric exporter type %q", exporterType)
```

service/internal/proctelemetry/config.go
codeboten: 👍🏻 fixed

service/telemetry/config.go
codeboten: I think you're right, this check doesn't make sense here. Looking through the rest of the code it does look like there should be a check for `conf == nil` before line 138 though.

service/telemetry/config.go
mx-psi: Filed open-telemetry/opentelemetry-configuration/issues/20

service/telemetry/config_test.go
codeboten: I added a test case for this in `TestUnmarshalMetricReaders`

service/internal/proctelemetry/config.go
codeboten: Done https://github.com/open-telemetry/opentelemetry-collector/issues/8043

service/telemetry.go
```diff
@@ -86,44 +85,73 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings component
...
-		zap.String(zapKeyTelemetryAddress, address),
-		zap.String(zapKeyTelemetryLevel, level.String()),
-	)
+	metricproducer.GlobalManager().AddProducer(tel.ocRegistry)
+	opts := []sdkmetric.Option{}
+	for name, reader := range cfg.Metrics.Readers {
+		// @codeboten: server returned here only happens when a pull based metric
+		// reader is configured, this could be refactored to pass in a
+		// func to add the server to the list of servers. another thing that would
+		// be nice is not to have to pass down the asyncErrorChannel
```
codeboten, line 129:
```suggestion
		// https://github.com/open-telemetry/opentelemetry-collector/issues/8045
```

service/telemetry.go
```diff
@@ -86,44 +85,70 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings component
...
-	} else {
-		if err := tel.initOpenCensus(level, res, promRegistry); err != nil {
+		portInt, err := strconv.Atoi(port)
+		if err != nil {
 			return err
 		}
+		if cfg.Metrics.Readers == nil {
+			cfg.Metrics.Readers = telemetry.MeterProviderJsonReaders{}
+		}
+		cfg.Metrics.Readers["pull/serviceaddress"] = telemetry.PullMetricReader{
```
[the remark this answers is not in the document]
TylerHelmuth: Is there a downside to `pull/prometheus`?  That feels like a clearer key, even if the key name doesn't matter.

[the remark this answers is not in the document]
codeboten: there isn't. Updated the name. Note there's a propsal to change from a map to an array, which may cause this to change in the future anyways :)

Issue #4646, opened 2022-01-05
[Proposal] Move batching in `exportehelper`


bogdandrutu:
Right now batching capability is offered as a processor, and this is extremely generic in our current pipeline model, which is a great thing, but also comes with limitations:
1. Batch processor uses a queue (implemented as a channel), which is not the same as the one used in the queue retry, that can lose data when the binary crashes even after the client was informed that data were accepted. There is already work to offer a persistent queue mechanism which will not be available in the processor.
2. If users want to use batching with the routing processor: will need to fix the #4544 issue, AND this will batch data that later will be split again. Very unnecessary.
3. SignalFx/SplunkHec receiver/exporter uses a "hack" that adds one header from the incoming request to the resource, batches things, then has to split based on that header. Batching inside the exporter helper can be configured with possible custom logic much easier, by making the batching library accepting a custom "batcher" func.
4. Allow customization of the "size" function. There were requests to support batching by the serialized size, we can offer this in the exporter helper much nicer since the custom `sizer` can actually work using the exporter wire format. So batching can happen after data serialization.
5. When multiple exporters configured, they may have different batching requirements. This can be achieved today with multiple pipelines, but that causes code duplicate between the pipelines.

This will also fix some more problems:
* https://github.com/open-telemetry/opentelemetry-collector/issues/4544
* https://github.com/open-telemetry/opentelemetry-collector/issues/4459 - by using the same workers defined in the queue_retry for this, hence we allow users to control this and we can still default on NumCpu.
* https://github.com/open-telemetry/opentelemetry-collector/issues/4554 - nothing to fix :)

tigrannajaryan:
A couple quick counterarguments:
1. Dedicated batch processor keeps the batch in memory once even when there are multiple exporters in the pipeline. If we move the equivalently sized batch to queuedretry of every exporter we will slightly increase memory requirements (not by much since most of pdata is going to be shared assuming exporters' output throughput more or less keeps up with the input).
1. Virtually the same batching work needs to happen for every exporter, which adds some CPU work per exporter. It would be useful to benchmark to see how much this difference is when there is more than one exporter.
1. Batching _probably_ helps to improve the performance of subsequent processors in the pipeline. We haven't benchmarked this, but it is possible that there is non-zero per-pdata cost that is amortized better if the batches are larger. This may be negligible, but I would like to see some benchmarks measuring it.

These are not necessarily strong enough to block the proposal, but I think it is worth thinking about these.

[...]

PR #7874, opened 2023-06-12
[exporterhelper] New exporter helper for custom requests

dmitryax:
Introduce a new exporter helper that operates over client-provided requests instead of pdata. The helper user now have to provide:
- Converter: an interface with a function implementing translation of pdata Metrics/Traces/Logs into a user-defined Request.
- Request sender: a function that will be called to send the Request

It opens a door for moving batching to the exporter, where batches will be built from client data format, instead of pdata. The batches can be properly sized by custom request size, which can be different from OTLP. The same custom request sizing will be applied to the sending queue. It will also improve performance of the sending queue retries for non-OTLP exporters, they don't need to translate pdata on every retry. 

This is an experimental API. Once stabilized, it's intended to replace the existing helpers.

Tracking Issue: https://github.com/open-telemetry/opentelemetry-collector/issues/8122

Related issue: https://github.com/open-telemetry/opentelemetry-collector/issues/4646

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -75,12 +75,14 @@ type baseSettings struct {
 	queueSettings
 	RetrySettings
 	RequestSender
+	requestExporter bool
 }
 
-// fromOptions returns the internal options starting from the default and applying all configured options.
-func fromOptions(options ...Option) *baseSettings {
-	// Start from the default options:
-	opts := &baseSettings{
+// fromOptions returns the baseSettings starting from the default and applying all configured options.
+// requestExporter indicates whether the base settings are for a new request exporter or not.
+func newBaseSettings(requestExporter bool, options ...Option) *baseSettings {
+	bs := &baseSettings{
+		requestExporter: requestExporter,
 		TimeoutSettings: NewDefaultTimeoutSettings(),
 		// TODO: Enable queuing by default (call DefaultQueueSettings)
 		queueSettings: queueSettings{
@@ -91,10 +93,10 @@ func fromOptions(options ...Option) *baseSettings {
 	}
 
 	for _, op := range options {
-		op(opts)
+		op(bs)
 	}
 
-	return opts
+	return bs
 }
 
 // Option apply changes to baseSettings.
@@ -137,6 +139,9 @@ func WithRetry(retrySettings RetrySettings) Option {
 // Permanent queue is not yet available for exporter helpers V2, they ignore StorageID field.
 func WithQueue(config QueueSettings) Option {
 	return func(o *baseSettings) {
+		if o.requestExporter {
+			panic("queueing is not supported for request exporters yet")
+		}
 		o.queueSettings.config = config
 	}
 }
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -56,14 +56,25 @@ func (req *baseRequest) OnProcessingFinished() {
 	}
 }
 
+type queueSettings struct {
+	config      QueueSettings
+	marshaler   internal.RequestMarshaler
+	unmarshaler internal.RequestUnmarshaler
+}
+
+func (qs *queueSettings) persistenceEnabled() bool {
+	return qs.config.StorageID != nil && qs.marshaler != nil && qs.unmarshaler != nil
+}
+
 // baseSettings represents all the options that users can configure.
 type baseSettings struct {
 	component.StartFunc
 	component.ShutdownFunc
 	consumerOptions []consumer.Option
 	TimeoutSettings
-	QueueSettings
+	queueSettings
 	RetrySettings
+	RequestSender
 }
 
 // fromOptions returns the internal options starting from the default and applying all configured options.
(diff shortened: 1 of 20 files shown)


jmacd (approved):
Looking forward! I see this as an important step for batching by request size.

dmitryax:
@open-telemetry/collector-approvers, please review whenever you have a chance. I'm planning to put more PR's on top of it next week

exporter/exporterhelper/common.go
```diff
@@ -121,9 +134,10 @@ func WithRetry(retrySettings RetrySettings) Option {
 
 // WithQueue overrides the default QueueSettings for an exporter.
 // The default QueueSettings is to disable queueing.
-func WithQueue(queueSettings QueueSettings) Option {
+// Permanent queue is not yet available for exporter helpers V2, they ignore StorageID field.
```
mx-psi, line 137: I would at least raise a warning about this, possibly just error out

exporter/exporterhelper/request.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	"go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+)
+
+// Request represents a single request that can be sent to the endpoint.
+type Request interface {
+	// ItemsCount returns the count basic item in the request, the smallest peaces of data that can be sent to the endpoint.
```
mx-psi, line 14:
```suggestion
	// ItemsCount returns the count basic item in the request, the smallest pieces of data that can be sent to the endpoint.
```

exporter/exporterhelper/request.go
```diff
@@ -0,0 +1,39 @@
...
+package exporterhelper // import "go.opentelemetry.io/collector/exporter/exporterhelper"
+
+import (
+	"context"
+
+	"go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+)
+
+// Request represents a single request that can be sent to the endpoint.
+type Request interface {
```
mx-psi, line 13: Should we make this interface unimplementable and have some public base request struct to provide default implementations for future methods?

exporter/exporterhelper/logs.go
```diff
@@ -114,6 +122,70 @@ func NewLogsExporter(
 	}, err
 }
 
+type LogsConverter interface {
+	// RequestFromLogs converts plog.Logs data into a request.
+	RequestFromLogs(context.Context, plog.Logs) (Request, error)
+}
+
+// NewLogsExporterV2 creates new logs exporter based on custom LogsConverter and RequestSender.
+func NewLogsExporterV2(
```
mx-psi, line 131: nit: We could name this `NewLogsRequestExporter` instead of using the `V2` suffix

mx-psi (commented):
Apologies for the delay in reviewing this, I missed this

exporter/exporterhelper/logs.go
bogdandrutu: We have not stabilized the previous API, so I agree with @mx-psi on this.

exporter/exporterhelper/logs.go
```diff
@@ -58,7 +64,7 @@ func (req *logsRequest) Marshal() ([]byte, error) {
 	return logsMarshaler.MarshalLogs(req.ld)
```
bogdandrutu, line 64: Should you delete this func?

exporter/exporterhelper/logs.go
bogdandrutu: Also would be good if we can eventually make old func call into the new.

exporter/exporterhelper/request.go
bogdandrutu: Alternative is to have also `RequestItemer` (or something like that) so we just don't need an interface and we use `any`. And ask users to provide funcs that work with the request. Another option is to ask user to implement the Send on that request as well, but right now `ItemCount` is on the request but Send is not, so inconsistent.

dmitryax: > Alternative is to have also RequestItemer (or something like that) so we just don't need an interface and we use `any`. And ask users to provide funcs that work with the request

In that case, do you suggest we make this an optional interface? We could do that, but the new itemized queue and batching exporter helpers won't work. We can probably allow that in case clients don't need that.

> Another option is to ask user to implement the Send on that request as well, but right now ItemCount is on the request but Send is not, so inconsistent.

What if we ask clients to implement `Request` with only one required `Send` function, and if they want itemized queue or batching helpers, they implement the optional interface `RequestItemer`. Then, we can possibly introduce bytes-sized queue or batching helpers if they implement `RequestSizer`.

```golang
type Request interface {
	Send(ctx context.Context) error
}

type RequestItemer interface {
	ItemsCount()  int
}

type RequestSizer interface {
	BytesSize() int
}
```

WDYT?

exporter/exporterhelper/common.go
dmitryax: The new exporter helper will get another option for the persistent queue with marshaller and unmarshaller as required params, and maybe another option for in-memory queue when the config doesn't have the `storage` param. So we won't be able to reuse this option anyway. I'll just put a panic here if it's used with the new exporter helper.

exporter/exporterhelper/logs.go
dmitryax: SGTM. Updated

exporter/exporterhelper/logs.go
dmitryax: Right. It's not needed anymore. Removed

exporter/exporterhelper/logs.go
dmitryax: > Also would be good if we can eventually make old func call into the new.

I didn't add queue options for the new helper in this PR yet. Wanted to do it as the next step. Once it's done, we can make the old func call the new one.

jpkrohling:
I would like to review this before it gets merged, especially if it conflicts with https://github.com/open-telemetry/opentelemetry-collector/pull/7510.

exporter/exporterhelper/request.go
dmitryax: I submitted another PR with the suggested approach

Issue #7054, opened 2023-01-29
Split core into more granular modules (some added after previous split)


bogdandrutu:
- [ ] obsreport, OR consider if splitting it as part of each component type.
- [ ] extension
- [ ] receiver
- [ ] processor
- [ ] exporter
- [ ] connector
- [ ] config
- [ ] service
- [ ] otelcol

mx-psi:
I think we should also split client and obsreport before closing this

evan-bradley:
I think we're done here, all components have been checked off and client and obsreport have been split off. Feel free to reopen (or just create a new issue and link this one) if you feel like this isn't true.

PR #7883, opened 2023-06-13
[extension/auth] split into its own module

codeboten:
Linked issue #7054


extension/ballastextension/go.mod
```diff
@@ -53,6 +53,8 @@ replace go.opentelemetry.io/collector/exporter => ../../exporter
 
 replace go.opentelemetry.io/collector/extension => ../
 
+replace go.opentelemetry.io/collector/extension/auth => ../auth
```
jpkrohling, line 56: Interesting, why does the ballast extension has a dependency on the auth package?

jpkrohling (approved):

extension/ballastextension/go.mod
codeboten: it does not... crosslink adds the replace in here automatically

bryan-aguilar: Is there any chance that this is a transient dependency?

jpkrohling: I removed the replace statement, ran the tests, and it worked without implying further changes to go.sum nor go.mod. Meaning: it's not a transitive dependency.

PR #7884, opened 2023-06-13
Add an option to skip generating the go.mod in the builder

jmacd:
**Description:** 
Adds to #7542, making this PR add two options, one to skip generation of all code (meaning to verify the config only), and one to skip generation of the go.mod file. 

Skipping the generation of the go.mod file means that instead of calling `go mod tidy` on a newly-generated `go.mod` file, we instead call `go get` on each module to update an existing go.mod file. This allows the builder to be integrated into a Golang monorepo where the collector is part of a larger build.

**Link to tracking Issue:** #7541 

**Testing:** This has been tested manually. There are no existing tests for the GetModules code path that needed to be tested here. Willing to add more tests if this PR is something the maintainers will accept.

**Documentation:** TODO

### DIFF
--- a/.chloggen/builder-skip-generate.yaml
+++ b/.chloggen/builder-skip-generate.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: cmd/builder
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add "--skip-generate" option to make builder skip source generation
+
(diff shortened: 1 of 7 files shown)


bogdandrutu:
@codeboten please make sure we are consistent with the dry run approach for collector.

jmacd:
Re: "meaning to verify the config" vs. --dry-run

The --dry-run flag is about verifying the collector's config. The stuff being discussed in this PR and #7541 is about validating a builder config.

bogdandrutu:
>The --dry-run flag is about verifying the collector's config. The stuff being discussed in this PR and https://github.com/open-telemetry/opentelemetry-collector/issues/7541 is about validating a builder config.

If I understand correctly both are to validate/verify the command config, so we should name the properties the same (flag dry-run, or sub-command verify, etc.).

Issue #7895, opened 2023-06-14
Split core into more granular modules: config


codeboten:

codeboten:
For the config, i think it makes sense for each directory under `config` (ie. `configgrpc`, `confighttp`) to be separate modules. I'll open a few PRs for separate modules and if there's any feedback against this, we can always go w/ the full `config` module, but if possible i'd rather keep each package small. The only shared code i've come across thus far has been the warning method in config/internal, but for now I propose duplicating the code in the two places using it since it's pretty minimal code

mx-psi:
From private convo, we can avoid sharing code by using a module that has `internal` in its path, like opentelemetry-go does here: https://github.com/open-telemetry/opentelemetry-go/blob/b757c7083baeece1be36caa4b1c102a3ddcdfa4b/exporters/otlp/otlpmetric/go.mod#L9

codeboten:
Closed the wrong issue by accident

PR #7906, opened 2023-06-15
 split config/configopaque into its own module

codeboten:
Part of #7895 

Follow up to #7905


jpkrohling (approved):

jpkrohling:
I reviewed only the last two commits from this PR. The other ones were reviewed in the PR this was initially based on.

PR #7927, opened 2023-06-19
Add zstd decompression support to HTTPServerSettings

tigrannajaryan:
This adds ability to decompress zstd-compressed HTTP requests to all receivers that use HTTPServerSettings.

Also added missing error handling for the case when an unsupported compression type was used in the request. Now it correctly returns 400 Bad Request. Also added a unit test to verify this case.

Once this is merged I will submit a PR in contrib repo to add end-to-end tests that use zstd compression in the testbed.

### DIFF
--- a/.chloggen/add-zstd-decompression.yaml
+++ b/.chloggen/add-zstd-decompression.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: HTTPServerSettings
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add zstd support to HTTPServerSettings
+
+# One or more tracking issues or pull requests related to the change
+issues: [7927]
+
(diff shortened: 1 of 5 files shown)


atoulme (approved):

tigrannajaryan:
> Change looks good. Please add a second changelog entry file for the bugfix

Done.

bogdandrutu:
Duplicate of https://github.com/open-telemetry/opentelemetry-collector/pull/7636?

bogdandrutu:
Please look at the other PR and incorporate that change as well.

tigrannajaryan:
I looked at https://github.com/open-telemetry/opentelemetry-collector/pull/7636 and it seems to do 2 additional things:
1. Adds snapy decompression. I think this can be added in a separate PR.
1. It closes the request Body after handling the request, which is not necessary (Go docs only [require it for clients, not for servers](https://pkg.go.dev/net/http#Request), which is where the confusion probably comes from). However reading the code helped me discover that I did not close zstd Decoder correctly. Fixed now in [this commit](https://github.com/open-telemetry/opentelemetry-collector/pull/7927/commits/2ac04a0ddb46b40382a861624639959500eaf09d), so the PR was useful! :-)

codeboten (approved):
Thanks for addressing my comments @tigrannajaryan

PR #7946, opened 2023-06-20
enhancement: record internal telemetry spans from all components, not just receivers

mattsains:
**Description:**

Configured the internal tracer to record child spans. This effectively extends span recording to capture processor and exporter-reported spans, and not just receiver spans.

The reason this works is because of this code: https://github.com/open-telemetry/opentelemetry-go/blob/fc88a6a25b9ccd8cb370947dd913898e9c37f838/sdk/trace/sampling.go#L276-L279

If the parent span is not set to sampled, child spans (like the ones created for subsequent component in the otel pipeline) will be non-recording spans and won't be recorded.

**Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/7947

**Testing:**
I wrote a custom extension to capture the internal traces and send to a local jaeger instance I created.

Telemetry before:

image

Telemetry after:

image

### DIFF
--- a/service/telemetry/otel_trace_sampler.go
+++ b/service/telemetry/otel_trace_sampler.go
@@ -10,7 +10,7 @@ import (
 type recordSampler struct{}
 
 func (r recordSampler) ShouldSample(_ sdktrace.SamplingParameters) sdktrace.SamplingResult {
-	return sdktrace.SamplingResult{Decision: sdktrace.RecordOnly}
+	return sdktrace.SamplingResult{Decision: sdktrace.RecordAndSample}
 }
 
 func (r recordSampler) Description() string {
--- a/service/telemetry/otel_trace_sampler.go
+++ b/service/telemetry/otel_trace_sampler.go
@@ -10,7 +10,7 @@ import (
 type recordSampler struct{}
 
 func (r recordSampler) ShouldSample(_ sdktrace.SamplingParameters) sdktrace.SamplingResult {
-	return sdktrace.SamplingResult{Decision: sdktrace.RecordAndSample}
+	return sdktrace.SamplingResult{Decision: sdktrace.RecordOnly}
 }
 
 func (r recordSampler) Description() string {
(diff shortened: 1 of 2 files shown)


service/telemetry/otel_trace_sampler.go
```diff
@@ -10,7 +10,7 @@ import (
 type recordSampler struct{}
 
 func (r recordSampler) ShouldSample(_ sdktrace.SamplingParameters) sdktrace.SamplingResult {
-	return sdktrace.SamplingResult{Decision: sdktrace.RecordOnly}
+	return sdktrace.SamplingResult{Decision: sdktrace.RecordAndSample}
```
bogdandrutu, line 13: We should respect parent sampling decision and not always sample blindly.

mattsains: Let me know if I'm misunderstanding something, but I think this is what this code does. This recordSampler type is used lower down in the method called "alwaysRecord" which returns a sampler. This is used to make a tracer provider with a sampler that always decides to record.

However, if a parent span is created with this sampler, none of its children are recorded. This doesn't seem to respect the "always record" contract to me.

From a purely functionality perspective, it seems like a bug that the collector only collects telemetry spans about its receivers, and not for any other component. This means that there is little point in having traces because there's only one span in them.

bogdandrutu: Use this https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace#ParentBased

mattsains: Would it make more sense to add a line here:

https://github.com/open-telemetry/opentelemetry-collector/blob/c6b5e77f06f6d98091490927276111806a8682eb/service/telemetry/otel_trace_sampler.go#L26

like:

```
sdktrace.WithLocalParentNotSampled(sdktrace.AlwaysSample()),
```

?

mattsains:
@bogdandrutu could you take another look at this please?

mattsains:
it's not stale, I'm waiting on @bogdandrutu or someone else to review this PR

mattsains:
it's not stale, I'm waiting on @bogdandrutu or someone else to review this PR.

mattsains:
it's not stale, I'm waiting on @bogdandrutu or someone else to review this PR.

mattsains:
it's not stale

mattsains:
it's not stale, it's just waiting for review

PR #7951, opened 2023-06-21
[scraper helper] improve ease of use of scraper controller settings

MovieStoreGuy:
**Description:** 

- Reduce the complexity required in order to override values within components
- Address feedback from https://github.com/open-telemetry/opentelemetry-collector/pull/7703#discussion_r1235257542

**Link to tracking Issue:** 
N/A

**Testing:** 

Added tests for additional options

**Documentation:** 

Documentation to be added to affected components in contrib

### DIFF
--- a/.chloggen/msg_cleanup-scraper-helper.yaml
+++ b/.chloggen/msg_cleanup-scraper-helper.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: scraperhelper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adding optional timeout field to scrapers
+
+# One or more tracking issues or pull requests related to the change
+issues: [7951]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/receiver/scraperhelper/scrapercontroller.go
+++ b/receiver/scraperhelper/scrapercontroller.go
@@ -19,23 +19,6 @@ import (
 	"go.opentelemetry.io/collector/receiver/scrapererror"
 )
 
-// ScraperControllerSettings defines common settings for a scraper controller
-// configuration. Scraper controller receivers can embed this struct, instead
-// of receiver.Settings, and extend it with more fields if needed.
-type ScraperControllerSettings struct {
-	CollectionInterval time.Duration `mapstructure:"collection_interval"`
-	InitialDelay       time.Duration `mapstructure:"initial_delay"`
-}
-
-// NewDefaultScraperControllerSettings returns default scraper controller
-// settings with a collection interval of one minute.
-func NewDefaultScraperControllerSettings(component.Type) ScraperControllerSettings {
-	return ScraperControllerSettings{
-		CollectionInterval: time.Minute,
-		InitialDelay:       time.Second,
-	}
-}
-
 // ScraperControllerOption apply changes to internal options.
 type ScraperControllerOption func(*controller)
 
@@ -64,6 +47,7 @@ type controller struct {
 	logger             *zap.Logger
 	collectionInterval time.Duration
 	initialDelay       time.Duration
+	timeout            time.Duration
 	nextConsumer       consumer.Metrics
 
 	scrapers    []Scraper
@@ -108,6 +92,7 @@ func NewScraperControllerReceiver(
 		logger:             set.Logger,
 		collectionInterval: cfg.CollectionInterval,
 		initialDelay:       cfg.InitialDelay,
+		timeout:            cfg.Timeout,
 		nextConsumer:       nextConsumer,
 		done:               make(chan struct{}),
 		terminated:         make(chan struct{}),
@@ -182,14 +167,19 @@ func (sc *controller) startScraping() {
 			sc.tickerCh = ticker.C
 		}
 
+		ctx, done := withScrapeContext(sc.timeout)
+
 		// Call scrape method on initialision to ensure
 		// that scrapers start from when the component starts
 		// instead of waiting for the full duration to start.
-		sc.scrapeMetricsAndReport(context.Background())
+		sc.scrapeMetricsAndReport(ctx)
+		done()
 		for {
 			select {
 			case <-sc.tickerCh:
-				sc.scrapeMetricsAndReport(context.Background())
+				ctx, done = withScrapeContext(sc.timeout)
+				sc.scrapeMetricsAndReport(ctx)
+				done()
 			case <-sc.done:
 				sc.terminated <- struct{}{}
 				return
@@ -229,4 +219,14 @@ func (sc *controller) scrapeMetricsAndReport(ctx context.Context) {
 // stopScraping stops the ticker
 func (sc *controller) stopScraping() {
 	close(sc.done)
+}
+
+// withScrapeContext will return a context that has no deadline if timeout is 0
+// which implies no explicit timeout had occurred, otherwise, a context
+// with a deadline of the provided timeout is returned.
+func withScrapeContext(timeout time.Duration) (context.Context, context.CancelFunc) {
+	if timeout == 0 {
+		return context.WithCancel(context.Background())
+	}
+	return context.WithTimeout(context.Background(), timeout)
 }
--- a/receiver/scraperhelper/scrapercontroller.go
+++ b/receiver/scraperhelper/scrapercontroller.go
@@ -166,20 +166,14 @@ func (sc *controller) startScraping() {
 
 			sc.tickerCh = ticker.C
 		}
-
-		ctx, done := withScrapeContext(sc.timeout)
-
 		// Call scrape method on initialision to ensure
 		// that scrapers start from when the component starts
 		// instead of waiting for the full duration to start.
-		sc.scrapeMetricsAndReport(ctx)
-		done()
+		sc.scrapeMetricsAndReport()
 		for {
 			select {
 			case <-sc.tickerCh:
-				ctx, done = withScrapeContext(sc.timeout)
-				sc.scrapeMetricsAndReport(ctx)
-				done()
+				sc.scrapeMetricsAndReport()
 			case <-sc.done:
 				sc.terminated <- struct{}{}
 				return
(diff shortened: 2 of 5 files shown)


MovieStoreGuy:
With making a breaking change that impacts contrib, what is the best way to order the PRs?

Should those be raised after this has been merged?
Should it be done beforehand? If so, what is the best way to reference this change to validate it?

mx-psi:
I think we should follow example 3 from https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#example-3---changing-the-arguments-of-a-function

receiver/scraperhelper/option.go
```diff
@@ -0,0 +1,42 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package scraperhelper // import "go.opentelemetry.io/collector/receiver/scraperhelper"
+
+import "time"
+
+// Option allows for overriding the defaults in
+// a developer friendly way
+type Option interface {
```
mx-psi, line 10: We can already set the settings manually on the struct since they are public fields, I don't think we should add a second way of doing the same thing; does this approach allow you to do anything you couldn't do before?

receiver/scraperhelper/scrapercontroller.go
```diff
@@ -78,6 +78,11 @@ func NewScraperControllerReceiver(
 		return nil, errors.New("collection_interval must be a positive duration")
 	}
 
+	timeout := cfg.Timeout
+	if timeout == 0 {
+		timeout = cfg.CollectionInterval
+	}
```
mx-psi, line 84: Docs say the timeout *"must be within the range of (0, CollectionInterval]"*. I think instead of doing this we should enforce this requirement and return an error if the timeout is 0. You can already set the timeout to the CollectionInterval if you want to.

.chloggen/msg_cleanup-scraper-helper.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: scraperhelper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Removing deprecated parameter from `NewDefaultScraperControllerSettings`
```
mx-psi, line 8: The function itself has not been deprecated, so I don't think we should do it until we have properly deprecated things

receiver/scraperhelper/option.go
MovieStoreGuy: This is more of a devx improvement to avoid having simplify setting default configuration.

An example of this can be seen here https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/zookeeperreceiver/factory.go#L33-L34

Not many services set the value, but enough do that it would be nice to have this improvement.

receiver/scraperhelper/scrapercontroller.go
MovieStoreGuy: So this point was raised in a previous PR that was merged that @astencel-sumo highlighted is that this change (not yet made it as part of a release) will enforce users to set both `timeout` and `collection_interval` if they wanted to reduce the `collection_interval` below the default `timeout`.

This allows the zero value (actual 0) to represent, "Use collection interval as timeout", which was the previous behaviour to this change. The expectation will be that `timeout` fields will only be used for scrapers with large collection intervals that doesn't make sense to persist the connection for 10m or 30m.

receiver/scraperhelper/option.go
bogdandrutu: +1 to @mx-psi. I had the same comment. Though I think it is fine to remove the old way with type since is not needed, and we can add Options without having any option defined yet (since adding varargs later is still considered a breaking change).

receiver/scraperhelper/settings.go
```diff
@@ -29,26 +29,41 @@ type ScraperControllerSettings struct {
 	// InitialDelay sets the initial start delay for the scraper,
 	// any non positive value is assumed to be immediately.
 	InitialDelay time.Duration `mapstructure:"initial_delay"`
-	// Timeout is used to set scraper's context deadline, it must be within
-	// the range of (0, CollectionInterval]
+	// Timeout is an optional value used to set scraper's context deadline,
+	// it must be within the range of (0, CollectionInterval].
```
djaglowski, line 33: It's not clear to me why we would constrain the timeout based on the collection interval. I would like to be able to set the timeout to something greater than the collection interval, so that a scrape may take a reasonable amount of additional time if necessary, but is killed in extreme cases. (e.g. 1.2x collection interval is fine, but 10x is not).

MovieStoreGuy: I do think we need a bounded cap that is based collection interval on the timeout should be, however, since we don't have a means of surfacing how long a scrape took, I am happy to leave this unbounded until we can surface that data.

receiver/scraperhelper/option.go
MovieStoreGuy: Just to be sure, @bogdandrutu are you saying that we should remove the `component.Type` parameter and leave the method as `NewDefaultScraperControllerSettings` ?

receiver/scraperhelper/settings.go
MovieStoreGuy: > What's the rationale for needing a bounded cap [?]

I would like to improve the scheduling guarantees for scrapers since we currently don't offer any safeguards.

The concerns I have where scrapers run outside the collection interval:
- Miss-aligned data resolution between configuration and observability system
  - Could be configured to run every 10s but finishes every 30s
- Resource contention created by scraper
  - Some scrapers may invoke a _stop the world_ or a global lock process which could impact the actual function of the service.
- No delays between scrapes
  - In the event that each scrape takes as long as the collection interval to observe a resource, then the observed system's resources are always used by the collector which may have negative side effects.

Now, these are mostly theoretical scenarios that I can think of a bounded execution cap set that would help solve but I'd like to actually get the data to help validate the frequency of these cases happening. 

I am happy to keep it as an opt-in addition since I don't know the likelihood of these happening at this point.

> ... as opposed to letting users decide?

Sorry, not against letting users decide at all as I know this will more than likely be a power user feature.

Let me create an issue to discuss this further to best circulate this around.

djaglowski: I'm happy to discuss on another issue but to clarify my question:

Assuming a user opts-in to a timeout, why shouldn't they be allowed to chose a timeout that is longer than the collection interval? I think this is a valid choice, (and the only way I would ever use it or recommend using it.)

Maybe I missed the reason here, but none of these seem to address the point.

> * Miss-aligned data resolution between configuration and observability system
>   
>   * Could be configured to run every 10s but finishes every 30s

This guarantees when we _start trying_ to collect data, but generally will have no positive effect on the result of collecting data. Consider two cases:

Case 1: `median time required to collect: 30s`, `collection_interval: 10s`, `timeout: 10s`

In this scenario the timeout will end up killing the collection attempt pretty much every time. The user will receive no data. The problem isn't solved but is instead exacerbated.

Case 2: `median time required to collect: 5s`, `collection_interval: 10s`, `timeout: 10s`

Collection will complete within the collection interval nearly every time. However, if it _does_ run long, we till end up with irregularities in the data reported. The timeout can help give a better chance that we _only_ miss one interval, but doesn't give us any guarantees either way.

> * Resource contention created by scraper
>   
>   * Some scrapers may invoke a _stop the world_ or a global lock process which could impact the actual function of the service.

I agree this would be worth killing, but it seems like a bug that needs to be addressed in the scraper. Is there any scraper where this is an actual possibility?

> * No delays between scrapes
>   
>   * In the event that each scrape takes as long as the collection interval to observe a resource, then the observed system's resources are always used by the collector which may have negative side effects.

A timeout that is less than the collection interval would help in this case, but I don't see that it has any bearing on why the timeout _may not_ be greater than the collection interval. Any timeout equal to or greater than the collection interval will result in no improvement in this case, so why allow the "equal" case but not the "greater than".

MovieStoreGuy: Right, I see what you mean now. 

> This guarantees when we start trying to collect data, but generally will have no positive effect on the result of collecting data. 

I was working off the premise that this would happen infrequently but would be consistent intervals that are missed, for example, the docker socket is overwhelmed and needs to have resources freed up so it can continue. In this scenario is better to exit early and come back to retry. However, your point remains true since these intervals would report no values.

> I agree this would be worth killing, but it seems like a bug that needs to be addressed in the scraper. Is there any scraper where this is an actual possibility?

I don't know for certain, But I think the JMX, SQL, and expvar under certain conditions.

> A timeout that is less than the collection interval would help in this case, but I don't see that it has any bearing on why the timeout may not be greater than the collection interval. Any timeout equal to or greater than the collection interval will result in no improvement in this case, so why allow the "equal" case but not the "greater than".

I agree it won't fix the data guarantees that I'd like but I'd like to add timeout support (that has no restrictions on duration besides can't be neg) to help provide some controls around scraper execution.

djaglowski: Thanks for working through the details with me. I think a timeout will be a great feature as long as it's optional and not constrained by the collection interval.

MovieStoreGuy:
Hey @djaglowski ,

I've updated the PR to reflect our discussion above, please let me know if there anything else I've missed.

Once this is merged in, I can create a PR in contrib updating the implementing scrapers' docs.

receiver/scraperhelper/scrapercontroller.go
```diff
@@ -182,14 +167,19 @@ func (sc *controller) startScraping() {
...
 		}
 
+		ctx, done := withScrapeContext(sc.timeout)
+
 		// Call scrape method on initialision to ensure
 		// that scrapers start from when the component starts
 		// instead of waiting for the full duration to start.
-		sc.scrapeMetricsAndReport(context.Background())
+		sc.scrapeMetricsAndReport(ctx)
+		done()
```
djaglowski, line 176: Since we're doing the exact same thing here and within the loop, would it make sense to move the context work into `scrapeMetricsAndReport`? This would allow us to `defer done()`, which I think it more typical.

MovieStoreGuy: ... I should have noticed this

djaglowski (approved):

mx-psi (approved):

codeboten (approved):

receiver/scraperhelper/settings.go
```diff
@@ -0,0 +1,53 @@
...
+func NewDefaultScraperControllerSettings(component.Type) ScraperControllerSettings {
+	return ScraperControllerSettings{
+		CollectionInterval: time.Minute,
+		InitialDelay:       time.Second,
+		Timeout:            0,
+	}
+}
+
+func (set *ScraperControllerSettings) Validate() (errs error) {
+	if set.CollectionInterval <= 0 {
```
jpkrohling, line 46: This makes the PR a breaking change, technically.

PR #7953, opened 2023-06-21
Revert "[scraper helper] Include support for timeout"

djaglowski:
Reverts open-telemetry/opentelemetry-collector#7703

Based on discussion in today's SIG meeting, I believe this change should be reverted. 

My takeaway from the discussion is that the following changes would be well received:
1. Produce a signal describing the duration of the scrape.
2. Provide an optional timeout setting, which is not constrained by the collection interval.


bogdandrutu:
Link the discussion points?

djaglowski:
> Link the discussion points?

I don't think points were transcribed but to the best of my recollection (with a few additions noted):

The primary motivation to limit collection time was so that backends would receive data points that align with the collection interval. There are several reasons why the timeout is not a solution for this.
1. Even when scrapes complete within the collection interval, they cannot guarantee uniformity of data point spacing. Observations may be delayed arbitrarily within the scrape interval, or the system providing the data points may timestamp them arbitrarily. (e.g. a collection interval of 10s may yield points at times 1, 19, 21, 39, 41)
2. (not discussed, but along the lines of point 1) Scrapers do not make guarantees that a given metric or time series is reported every scrape interval. Many systems provide data only when an event occurs and we do not necessarily report "0"'s when an API omits corresponding information.
3. A scraper that fails to complete within a collection interval is likely to _always_ fail to complete during subsequent collection intervals, so the typical result of killing a scrape is that users will see _no_ data rather than better data. When this occurs, users are likely to be confused by the lack of data. 

Killing scrapers mid-scrape is likely to expose unnecessary fragility. (Many code paths involving cancelation will be exercised regularly during steady state of collector, rather than only at shutdown. While this would help us identify any problems, we should expect some issues to be sticky e.g. in client libraries. In the mean time, all users of scrapers are requires to accept this risk.)

Some uses would use a timeout only to kill excessively long scrapes that may be hung due to a problem. (e.g. 10x longer than collection interval) This is different than a scrape that regularly takes 1.2x the collection interval to complete (which I would argue is acceptable to most users.)

If the goal is to inform users that their configured scrape interval is shorter than what is required for a successful scrape, then we should inform them using a signal (specifically a metric was suggested). At the same time, they are still receiving data _and_ sufficient information to inform them about how to adjust the interval, if they even care. (In comparison, if we had killed the interval, they would not know how long is needed.)

MovieStoreGuy:
Hey @djaglowski ,

> A scraper that fails to complete within a collection interval is likely to always fail to complete during subsequent collection intervals, so the typical result of killing a scrape is that users will see no data rather than better data. When this occurs, users are likely to be confused by the lack of data.

Are there any scrapers that are known to be rather expensive in their runtime?

Within our internal CI tests, we've had all of our scrapers complete in their collection interval (we check the number of data points received is equal to the number we are expecting) however this is only a subset of receivers that are available so I am curious what scrapers are likely to be affected.

There is a concern here that even though a user has configured their collection interval to `1s` (for argument's sake) if the scraper exceeds that interval due to the scraper itself overwhelming the resource (ie: expvar receiver DDoS a go app) then I believe the scraper should be terminated in that case. I don't believe that the scraper should be allowed to indefinitely wait until it has completed or errored to continue on the next scrape.

I do believe there is a definite hazard in having this as a default across all scrapers for the points you have highlighted but I do want to invest more into the scraper controller to help surface the issue of too short of a collection interval and provide user's controls to help ensure that observing resources doesn't additional problems for it.

I have created the [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/7951) that addressed the concerns raised in the SIG, mainly to move the timeout to be opt-in field, which I agree makes more sense since I see it as a power user setting as opposed to users who want to set and forget. 

I think it also makes sense to revert the original commit since it does make the changes confusing for users.

I will need to update the change log to correctly reflect the changes since it was originally intended to provide improved devx support.

bogdandrutu (approved):
Though majority of the reasons added by @djaglowski in my opinion are not related to this PR for example 1 and 2, I do believe there are some legit concerns about having this Timeout default to collection interval, and for components/users that change the collection interval to fail validation is bad.

So @MovieStoreGuy I think we should have a default for timeout that if a user was configuring a different collection interval for a component, the validation does not fail. We can start by having timeout disabled or having timeout = 0 means the timeout will be equal to collection interval (up for discussion).

PR #7954, opened 2023-06-21
[chore] mark confignet as stable

codeboten:
This contents of this module have not changed in over 2 years


bogdandrutu:
Can you give me until mid July before marking any config stable? I want to also discuss somewhere and document things like:
* Adding new validation rules means breaking change?
* Should we offer a "NewDefaul...." for each config and say that the behavior of the config is stable only if initialized with NewDefault?
* Adding new fields means breaking change? Let's assume someone "squash" the message into another message, then adding a field with same mapstruct value will be a breaking change, what do we do with that? What are the rules we follow.

There may be other things to document (just added some in my mind) but would like others to think of any special implications for configs that are not defined by the golang conventions of breaking compatibility,.

codeboten (commented):
I think that's fine, i'll move this draft for now. My only motivation for moving this to RC is that there havent been changes in this dir for some time.

bogdandrutu:
@codeboten the time in my comment was to give us time to address my concerns, so I do believe someone (who is working full time, not like me until Sept) should address them and clarify what is possible and what is not in a stable config package.

mx-psi:
Filed #8002 to discuss the stability guarantees for configuration structures. One of these was already discussed on the `VERSIONING.md` document (adding new fields was deemed to be fine).

bogdandrutu:
Another thing is to make sure we pass "context" everywhere. Maybe TelemetrySettings as well

PR #7957, opened 2023-06-23
Fix docs on ca_file

jroper:
The `ca_file` configuration is only used for client TLS configuration, to specify the CA to use to verify the certificate supplied by the server. The existing documentation said it would also be used by the server to specify the CA to use to verify client certificates, but this is not true (verified with testing, setting it for the server has no impact), and contradicts the documentation below which says that `client_ca_file` is used to specify the CA to use to verify client certificates.

So, I've updated the documentation to be correct, and I've also moved  the documentation `ca_file` to the section for client specific properties, since it only applies to the client.

### DIFF
--- a/config/configtls/README.md
+++ b/config/configtls/README.md
@@ -24,20 +24,6 @@ As a result, the following parameters are also required:
   only be used if `insecure` is set to false.
   - `key_pem`: Alternative to `key_file`. Provide the key contents as a string instead of a filepath.
 
-A certificate authority may also need to be defined:
-
-- `ca_file`: Path to the CA cert. For a client this verifies the server
-  certificate. For a server this verifies client certificates. If empty uses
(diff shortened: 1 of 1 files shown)


codeboten (commented):
Thanks for opening this PR, after reviewing the code, you're right that the client verification in a server will only use client CA file. That being said, `ca_file` is [still loaded](https://github.com/open-telemetry/opentelemetry-collector/blob/c3f70bd1f8ea00dd04be45814d02b671064e5aa0/config/configtls/configtls.go#L180) in the TLS configuration of a server, it just isn't used to validate clients connecting to it. I suppose if a server was to make outbound connections, `ca_file` would be applied, but i suspect this is not a normal use-case.

This makes me wonder if this warrants a warning or an error being logged when someone starts a server with a CA file configured.

codeboten (commented):
Please rebase the PR to make CI happy

PR #7965, opened 2023-06-26
Add externally provided Prometheus registry to service.New

eine-doodka:
**Description:**

Service.New() can now be provided with an external reference to a prometheus.Registry.

The rationale behind those changes is that current internal registry cannot be shared with external users of collector.

Example: my team and I are working on a project on top of the OpenTelemetry and, consequently, are building a thing on top of opentelemetry-collector. The thing has two metric endpoints itself, and we do not want to have a third one inside the collector. With external prometheus.Registry we can now provide Service.New() with our own registry and make an endpoint with its metrics by ourselves.

I believe this contribution aligns with Observable and Extensible items from ```vision.md```

**Testing:**
No additional tests were made as the changes seem quite trivial and do not introduce any breaking API changes


bogdandrutu (changes requested):
Please do not merge this since:
1. We are moving away from Prometheus being the way to record metrics.
2. Avoid exposing any telemetry internal components to our public API.

Issue #7920, opened 2023-06-16
importing opentelemetry-collector clobbers previously registered gRPC compressors


mostynb:
**Describe the bug**

Some consumers of this library have registered grpc compressors, and the init() order can cause their compressors to be clobbered by opentelemetry-collector's use of github.com/mostynb/go-grpc-compression, which follows the pattern of https://github.com/grpc/grpc-go/blob/master/encoding/gzip/gzip.go (overwrite any previously registered encoders with the same name).

**Steps to reproduce**
1. Write a program that uses a specific zstd grpc compressor implementation (that isn't github.com/mostynb/go-grpc-compression/zstd).
2. Add opentelemetry-collector to that program.
3. Note that the zstd grpc compressor is now github.com/mostynb/go-grpc-compression/zstd, depending on init() ordering

**What did you expect to see?**
The programs original grpc compressor should be used.

**What did you see instead?**
github.com/mostynb/go-grpc-compression/zstd is used instead of the preferred compressor.

**What version did you use?**
Any version of opentelemetry-collector that uses github.com/mostynb/go-grpc-compression/* should be affected.

**Proposed solution**
https://github.com/mostynb/go-grpc-compression/pull/19 adds "nonclobbering" versions of each compressor. If I were to land that and make a new release, we could change all opentelemetry-collector imports of github.com/mostynb/go-grpc-compression/* to the corresponding github.com/mostynb/go-grpc-compression/nonclobbering/* package.

If you would accept such a change, I will land that PR in github.com/mostynb/go-grpc-compression (reviews of the PR above are welcome too).

mx-psi:
Hi @mostynb thanks for taking the time to open the issue. I think this is a reasonable solution, cc @open-telemetry/collector-approvers

codeboten:
Fixed by https://github.com/open-telemetry/opentelemetry-collector/pull/7966

PR #7966, opened 2023-06-26
Upgrade github.com/mostynb/go-grpc-compression and switch to nonclobbering imports

mostynb:
**Description:** Switch to grpc encoders that won't override the consumer's previously registered codecs.

Some consumers of this library have registered grpc compressors, and the init() order can cause their compressors to be clobbered by opentelemetry-collector's use of github.com/mostynb/go-grpc-compression, which follows the pattern from the grpc-go library (overwrite any previously registered encoders with the same name).

Version 1.2.0 of github.com/mostynb/go-grpc-compression added so-called "nonclobbering" imports, which only register the codec if there is not already a codec registered with that name. This PR switches to use those.

**Link to tracking Issue:** #7920

**Testing:** tests have been added to the upstream library.


codeboten (approved):
Thanks @mostynb

codeboten (commented):
@mostynb can you add a [changelog entry](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#adding-a-changelog-entry) to capture this change?

I suspect it will mostly be a non event for most users, but in case anyone is surprised by their compression options changing, they can reference the changelog to see what changed.

mostynb:
> @mostynb can you add a [changelog entry](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#adding-a-changelog-entry) to capture this change?
> 
> I suspect it will mostly be a non event for most users, but in case anyone is surprised by their compression options changing, they can reference the changelog to see what changed.

I gave it a try- I hope the formatting is OK.

codeboten (approved):
Thanks for rebasing @mostynb!

config/configgrpc/go.sum
```diff
@@ -291,7 +291,6 @@ github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd/go.mod h1:6dJ
 github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742/go.mod h1:bx2lNnkwVCuqBIxFjflWJWanXIb3RllmbCylyMrvgv0=
 github.com/modern-go/reflect2 v1.0.1/go.mod h1:bx2lNnkwVCuqBIxFjflWJWanXIb3RllmbCylyMrvgv0=
 github.com/modern-go/reflect2 v1.0.2 h1:xBagoLtFs94CBntxluKeaWgTMpvLxC4ur3nMaC9Gz0M=
-github.com/modern-go/reflect2 v1.0.2/go.mod h1:yWuevngMOJpCy52FWWMvUC8ws7m/LJsjYzDa0/r8luk=
```
mostynb, line 294: oops, this should not have been removed

PR #7977, opened 2023-06-27
add support for additional decoders

codeboten:
This provides the ability to add support for compression types which are not supported in core, to be supported by individual components.

This change addresses an issue caused by https://github.com/open-telemetry/opentelemetry-collector/pull/7927, which prevents receivers that use confighttp from adding support for their content-encoding types easily. They still can by implementing an error handler, but that seems messy.

An alternative to this PR is to not return an error for an unsupported encoding type as was the case before https://github.com/open-telemetry/opentelemetry-collector/pull/7927

If the approach is acceptable, I will add the unit tests and changelog

### DIFF
--- a/config/confighttp/compression.go
+++ b/config/confighttp/compression.go
@@ -68,25 +68,68 @@ func (r *compressRoundTripper) RoundTrip(req *http.Request) (*http.Response, err
 type decompressor struct {
 	errHandler func(w http.ResponseWriter, r *http.Request, errorMsg string, statusCode int)
 	base       http.Handler
+	decoders   map[string]func(body io.ReadCloser) (io.ReadCloser, error)
 }
 
 // httpContentDecompressor offloads the task of handling compressed HTTP requests
 // by identifying the compression format in the "Content-Encoding" header and re-writing
 // request body so that the handlers further in the chain can work on decompressed data.
 // It supports gzip and deflate/zlib compression.
-func httpContentDecompressor(h http.Handler, eh func(w http.ResponseWriter, r *http.Request, errorMsg string, statusCode int)) http.Handler {
+func httpContentDecompressor(h http.Handler, eh func(w http.ResponseWriter, r *http.Request, errorMsg string, statusCode int), decoders map[string]func(body io.ReadCloser) (io.ReadCloser, error)) http.Handler {
 	errHandler := defaultErrorHandler
 	if eh != nil {
 		errHandler = eh
 	}
-	return &decompressor{
+
+	d := &decompressor{
 		errHandler: errHandler,
 		base:       h,
+		decoders: map[string]func(body io.ReadCloser) (io.ReadCloser, error){
+			"": func(body io.ReadCloser) (io.ReadCloser, error) {
+				// Not a compressed payload. Nothing to do.
+				return nil, nil
+			},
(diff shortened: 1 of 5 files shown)


jmacd (approved):

config/confighttp/confighttp.go
```diff
@@ -262,6 +263,14 @@ func WithErrorHandler(e errorHandler) ToServerOption {
 	}
 }
 
+// WithDecoder provides support for additional decoders to be configured
+// by the caller.
+func WithDecoder(key string, dec decoder) ToServerOption {
```
bogdandrutu, line 268: See https://github.com/open-telemetry/opentelemetry-collector/pull/7870 related to exposing internal types

codeboten: Yup, i was wondering the same thing when i looked at errorHandler :D

codeboten: Removed internal type

tigrannajaryan:
> An alternative to this PR is to not return an error for an unsupported encoding type as was the case before #7927

I think reverting to the wrong behavior is probably not what we want. Let's go forward with your PR. The question I have is whether there is a need for custom decoders or we just need to add "snappy" for all?

tigrannajaryan (approved):

mwear (approved):

codeboten:
> I think reverting to the wrong behavior is probably not what we want. Let's go forward with your PR. The question I have is whether there is a need for custom decoders or we just need to add "snappy" for all?

Adding snappy would address the one case failing in contrib at the moment. I've not looked for additional cases, though @jpkrohling pointed out in a DM that this would allow anyone wanting to offer alternative decoders even for supported encodings could do so w/ this functionality

config/confighttp/compression.go
```diff
@@ -68,25 +68,68 @@ func (r *compressRoundTripper) RoundTrip(req *http.Request) (*http.Response, err
...
+			"zlib": func(body io.ReadCloser) (io.ReadCloser, error) {
+				zr, err := zlib.NewReader(body)
+				if err != nil {
+					return nil, err
+				}
+				return zr, nil
+			},
+		},
+	}
+	d.decoders["deflate"] = d.decoders["zlib"]
```
jpkrohling, line 122: Would it be asking too much to add an explicit test for snappy, as this is the one failing for existing receivers?

jpkrohling (commented):
Looks good to me, but I would love to have another test for snappy.

config/confighttp/compression.go
codeboten: I could add one, but it would essentially be a duplicate of the "UnsupportedCompression" test here: https://github.com/open-telemetry/opentelemetry-collector/blob/ec033ca833d34108142c8a3bdf27ce2f2e517328/config/confighttp/compression_test.go#L185-L189

jpkrohling: My idea is that it would replace the "custom-encoding" test case with a real implementation, and it would be converted into a test that overrides an existing encoding once snappy is added. In any case, this isn't a blocker.

jpkrohling (approved):

config/confighttp/compression.go
codeboten: I'll add this in a follow up PR since this is blocking contrib :)

Issue #7804, opened 2023-06-01
Create a strict mode for feature gates


atoulme:
**Is your feature request related to a problem? Please describe.**
The current behavior of the collector allows folks to assume sane defaults and embraces breaking changes without warning on version upgrades. The collector relies on feature gates that work as feature toggles that enable a breaking change explicitly.
Initially, features gates are considered alpha. After a period of time, feature gates are then turned on by default and declared beta. Finally, feature gates are declared stable and removed altogether.

The system doesn't allow proactive customers to prepare and execute well with feature gates, instead forcing customers to read through changelogs and make sure they understand the implication of a feature gate. It requires to pay attention to new feature changes, change of status of feature gates, and granular updates by keeping up with the collector changes constantly.

**Describe the solution you'd like**
I would like to introduce a strict mode that forces users of the collector to explicitly opt into feature gates without any ambiguity.
* If the feature gate is freshly introduced and considered alpha, the strict mode must make the collector fail to start if the configuration doesn't explicitly list a decision about the feature gate, whether it is enabled or disabled. This forces the customer to adopt the feature gate explicitly.
* If the feature gate is considered beta and therefore enabled by default, the strict mode must fail if the gate is not adopted by the customer, as in the flag is not explicitly set to activate the feature gate. The customer can decide to override this for a specific feature gate, but this must result in loud warnings. This is to ensure that the customer cannot be surprised on adoption of the feature gate by default.
* If the feature gate is stable and removed, the strict mode must fail if the feature gate flag is used. This is to ensure that customers understand that the feature gate can no longer be changed.

The strict mode can be used in testing environments of customers to test and validate configurations and help smooth over migrations between versions of the collector.

TylerHelmuth:
@atoulme are you imaging this mode would be a separate flag to pass into the collector, not a new stability level?

When the collector fails, what information should be provided to the user to ensure they understand that the failure is because of strict mode and a feature gate?

Could this problem be solved if the `--help` flag share more information about the available feature gates, such as the description, `FromVersion`, and reference URL? Would a more explicit list of feature gates, other than the changelog, help ([k8s has this](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features))?

atoulme:
It would be a separate flag. It can also be set in the collector, I hope, for downstream distributions.

[...]

PR #7991, opened 2023-06-28
Use strict mode to drive explicit behaviors when setting feature gates

atoulme:
**Description:** 
Implementation of a new feature gate flag that allows stricter validation of feature gate configuration.
This is useful when testing new versions of the collector to alert downstream maintainers that components have new or changed feature gates.

**Link to tracking Issue:**
#7804 

**Testing:**
Unit tests.

**Documentation:**
README changes.

### DIFF
--- a/.chloggen/featuregate_strict_mode.yaml
+++ b/.chloggen/featuregate_strict_mode.yaml
@@ -5,7 +5,7 @@ change_type: enhancement
 component: featuregate
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Use strict mode to drive explicit behaviors when setting feature gates
+note: Add a new flag to set feature gates with additional validation
 
 # One or more tracking issues or pull requests related to the change
 issues: [7804]
@@ -13,4 +13,8 @@ issues: [7804]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext:
+subtext: |
+  In strict mode, the collector will perform additional checks on startup to validate the feature gates set:
+  * All alpha and beta feature gates must be set explicitly.
+  * All beta feature gates must be explicitly enabled.
+  * All stable feature gates must be left unset.
--- a/.chloggen/featuregate_strict_mode.yaml
+++ b/.chloggen/featuregate_strict_mode.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: featuregate
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use strict mode to drive explicit behaviors when setting feature gates
+
+# One or more tracking issues or pull requests related to the change
+issues: [7804]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.chloggen/featuregate_strict_mode.yaml
+++ b/.chloggen/featuregate_strict_mode.yaml
@@ -1,5 +1,5 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: featuregate
@@ -18,3 +18,4 @@ subtext: |
   * All alpha and beta feature gates must be set explicitly.
   * All beta feature gates must be explicitly enabled.
   * All stable feature gates must be left unset.
+change_logs: [api, user]
--- a/.chloggen/featuregate_strict_mode.yaml
+++ b/.chloggen/featuregate_strict_mode.yaml
@@ -5,7 +5,7 @@ change_type: enhancement
 component: featuregate
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add a new flag to set feature gates with additional validation
+note: "Add new flag `--feature-gates-strict` to set feature gates with additional validation"
 
 # One or more tracking issues or pull requests related to the change
 issues: [7804]
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -56,7 +56,7 @@ This will enable `gate1` and `gate3` and disable `gate2`.
 
 Features gate can be executed in strict mode as well.
 
-To activate, use the flag `--feature-gates-strict=true`.
+To activate, use the flag `--feature-gates-strict`.
 
 When set, the flag adds validation rules:
 
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -52,6 +52,21 @@ otelcol --config=config.yaml --feature-gates=gate1,-gate2,+gate3
 
 This will enable `gate1` and `gate3` and disable `gate2`.
 
+### Strict mode
+
+Features gate can be executed in strict mode as well. This mode uses the flag `--feature-gates-strict`.
+This flag is mutually exclusive flag with `--feature-gates`.
+
+The flag follows the same notation as the `--feature-gates` flag and is interpreted with additional rules by the collector:
+
+* If the feature gate is alpha, the collector will fail to start if the feature gate is not explicitly set, 
+whether it is enabled or disabled.
+* If the feature gate is considered beta and therefore enabled by default, the collector will fail if the gate is not enabled.
+* If the feature gate is stable, the strict mode fails if the feature gate flag is used.
+
+The strict mode is typically used when testing collector upgrades to ensure that changes to feature gates stability levels
+are tracked and explicitly managed as part of the upgrade.
+
 ## Feature Lifecycle
 
 Features controlled by a `Gate` should follow a three-stage lifecycle, 
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -55,7 +55,6 @@ This will enable `gate1` and `gate3` and disable `gate2`.
 ### Strict mode
 
 Features gate can be executed in strict mode as well. This mode uses the flag `--feature-gates-strict`.
-This flag is mutually exclusive flag with `--feature-gates`.
 
 The flag follows the same notation as the `--feature-gates` flag and is interpreted with additional rules by the collector:
 
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -56,7 +56,7 @@ This will enable `gate1` and `gate3` and disable `gate2`.
 
 Features gate can be executed in strict mode as well. This mode uses the flag `--feature-gates-strict`.
 
-The flag follows the same notation as the `--feature-gates` flag and is interpreted with additional rules by the collector:
+When set, the flag adds validation rules:
 
 * If the feature gate is alpha, the collector will fail to start if the feature gate is not explicitly set, 
 whether it is enabled or disabled.
--- a/featuregate/README.md
+++ b/featuregate/README.md
@@ -54,7 +54,9 @@ This will enable `gate1` and `gate3` and disable `gate2`.
 
 ### Strict mode
 
-Features gate can be executed in strict mode as well. This mode uses the flag `--feature-gates-strict`.
+Features gate can be executed in strict mode as well.
+
+To activate, use the flag `--feature-gates-strict=true`.
 
 When set, the flag adds validation rules:
 
--- a/featuregate/flag.go
+++ b/featuregate/flag.go
@@ -74,7 +74,7 @@ func (f *flagValue) Set(s string) error {
 			return
 		}
 		if f.reg.strict && !enabled && gate.stage == StageBeta {
-			errs = multierr.Append(errs, fmt.Errorf("gate %s must be explicitly enabled, remove strict mode to override", gate.id))
+			errs = multierr.Append(errs, fmt.Errorf("gate %q must be explicitly enabled, remove strict mode to override", gate.id))
 		}
 		if f.reg.strict && gate.stage == StageStable {
 			errs = multierr.Append(errs, fmt.Errorf("gate %q must not be set", gate.id))
--- a/featuregate/flag.go
+++ b/featuregate/flag.go
@@ -12,14 +12,14 @@ import (
 )
 
 // NewFlag returns a flag.Value that directly applies feature gate statuses to a Registry.
-func NewFlag(reg *Registry, strict bool) flag.Value {
+func NewFlag(reg *Registry, strict *bool) flag.Value {
 	return &flagValue{reg: reg, strict: strict}
 }
 
 // flagValue implements the flag.Value interface and directly applies feature gate statuses to a Registry.
 type flagValue struct {
 	reg    *Registry
-	strict bool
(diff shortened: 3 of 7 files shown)


mx-psi (changes requested):
I think we still need to clarify the behavior and use case in the issue, but I left an initial review. I feel strongly that we should not make this work via `--feature-gates=strict,[rest of flags]`

otelcol/flags.go
```diff
@@ -51,7 +51,7 @@ func flags(reg *featuregate.Registry) *flag.FlagSet {
 		})
 
 	flagSet.Var(featuregate.NewFlag(reg), featureGatesFlag,
-		"Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.")
+		"Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature. Start with strict to enable strict mode.")
```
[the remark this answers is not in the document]
atoulme: Sure thing - please note as a positional argument it must be placed before feature gates are set though. I can do that or maybe I can use a `-feature-gates-strict` argument that would enforce strictness. `-feature-gates` and `-feature-gates-strict` would be mutually exclusive. Would that work for you?

atoulme:
Thanks for the review @mx-psi and @djaglowski , it looks like the mechanism is overall ok and now I will work on the interfaces, documentation and tests to make this an acceptable feature addition.

featuregate/registry.go
```diff
@@ -18,7 +18,8 @@ func GlobalRegistry() *Registry {
 }
 
 type Registry struct {
-	gates sync.Map
+	gates  sync.Map
+	strict bool
```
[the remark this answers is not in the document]
atoulme: removed from registry and kept it in flag now. I also added docs in the README.

featuregate/flag_test.go
```diff
@@ -106,6 +106,40 @@ func TestNewFlag(t *testing.T) {
...
+		{
+			name:        "strict mode",
+			input:       "strict,alpha,beta,-alpha",
+			expected:    map[string]bool{"alpha": false, "beta": true, "deprecated": false, "stable": true},
+			expectedStr: "strict,-alpha,beta,-deprecated,stable",
+		},
+		{
+			name:           "strict mode but no alpha gate set",
+			input:          "strict,beta",
+			expectedSetErr: true,
```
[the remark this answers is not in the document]
atoulme: Done.

.chloggen/featuregate_strict_mode.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: featuregate
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use strict mode to drive explicit behaviors when setting feature gates
```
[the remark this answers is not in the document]
atoulme: added more information in the changelog.

otelcol/flags.go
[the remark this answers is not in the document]
atoulme: Going with that and let's see how it lands.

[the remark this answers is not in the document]
mx-psi: > please note as a positional argument it must be placed before feature gates are set though

@atoulme could you elaborate on this? I don't understand why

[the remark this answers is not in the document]
atoulme: It’s my guess, but the flags are applied sequentially. Strict mode applies when the flag for feature gates is set.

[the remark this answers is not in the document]
mx-psi: I don't think the fact that flags are applied sequentially means we can't make this work in any order. We can either 
1. delay setting flags until all flags have been parsed (e.g. by using a `PreRun` hook from cobra to actually apply the feature gates)
2. apply feature gates if `--feature-gates` is passed before the strict mode flag, but fail if we don't pass strict mode validation

I guess we should do (2) if people have been complaining that feature gate initialization happens to late as in #4967. The only case where this would make a difference is if there is a feature gate that changes something about strict mode or CLI flag parsing.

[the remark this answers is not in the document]
atoulme: 1. seems like it will clash with #4967 then.
2. seems ok with me and I can do it.
What do you think of my current approach of using `--feature-gates-strict`?

atoulme:
@mx-psi would you like to take another look? I think I addressed your feedback. Thanks again!

.chloggen/featuregate_strict_mode.yaml
```diff
@@ -0,0 +1,20 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: featuregate
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add a new flag to set feature gates with additional validation
```
mx-psi, line 8:
```suggestion
note: "Add new flag `--feature-gates-strict` to set feature gates with additional validation"
```

featuregate/flag.go
```diff
@@ -49,7 +54,28 @@ func (f *flagValue) Set(s string) error {
...
+		gatesEnabled[id] = val
+		if _, ok := f.reg.gates.Load(id); !ok {
+			errs = multierr.Append(errs, fmt.Errorf("no such feature gate %q", id))
+		}
 	}
+	f.reg.VisitAll(func(gate *Gate) {
+		enabled, ok := gatesEnabled[gate.id]
+		if !ok {
+			if *f.strict && (gate.stage == StageAlpha || gate.stage == StageBeta) {
+				errs = multierr.Append(errs, fmt.Errorf("gate %q is in %s and is not explicitly configured", gate.id, gate.stage))
```
mx-psi, line 66: To make the wording match that of the other errors, we can also phrase this as a 'must':
```suggestion
				errs = multierr.Append(errs, fmt.Errorf("gate %q is in %s and must be explicitly configured", gate.id, gate.stage))
```

featuregate/README.md
```diff
@@ -52,6 +52,21 @@ otelcol --config=config.yaml --feature-gates=gate1,-gate2,+gate3
 
 This will enable `gate1` and `gate3` and disable `gate2`.
 
+### Strict mode
+
+Features gate can be executed in strict mode as well. This mode uses the flag `--feature-gates-strict`.
+This flag is mutually exclusive flag with `--feature-gates`.
+
+The flag follows the same notation as the `--feature-gates` flag and is interpreted with additional rules by the collector:
```
mx-psi, line 60: I don't think this corresponds to the implementation, right? The definition here https://github.com/open-telemetry/opentelemetry-collector/blob/53670a347f1c7c4e0399dc76095384743c6adaf2/otelcol/flags.go#L54-L56 is that of a boolean flag, so it does not follow the same notation.

featuregate/README.md
```diff
@@ -52,6 +52,21 @@ otelcol --config=config.yaml --feature-gates=gate1,-gate2,+gate3
 
 This will enable `gate1` and `gate3` and disable `gate2`.
 
+### Strict mode
+
+Features gate can be executed in strict mode as well. This mode uses the flag `--feature-gates-strict`.
+This flag is mutually exclusive flag with `--feature-gates`.
```
mx-psi, line 58: I also believe this sentence is not in line with the implementation
```suggestion
```

featuregate/flag_test.go
```diff
@@ -14,97 +14,151 @@ func TestNewFlag(t *testing.T) {
 	for _, tt := range []struct {
```
mx-psi, line 14: Can we have a test where we test that both 
```
--feature-gates=alpha,beta --feature-gates-strict
```

and

```
--feature-gates-strict --feature-gates=alpha,beta
```

work?

atoulme: Sure thing. I tried manually and it worked, maybe we need an integration test for this.

featuregate/README.md
atoulme: OK, the README needs an update indeed. On it.

featuregate/flag_test.go
mx-psi: > Sure thing. I tried manually and it worked, maybe we need an integration test for this.

I think we can make it work with a unit test: for each `arguments` combination
1. call the `flags` function with a new registry,
2. call [`flagset.Parse`](https://pkg.go.dev/flag#FlagSet.Parse) with `arguments`
3. check that the error is as expected, or that the registry has the values you expect otherwise.

That should do the trick

atoulme: ok I will do that for sure, thanks. Sorry, I'll get it done soon, juggling work.

mx-psi: No problem :) not my intention to introduce any sense of urgency, sorry if it came out that way

atoulme: No, just feeling mad at myself for not getting this right :)

atoulme: Took me a good while but I think I figured it out. There was indeed an ordering issue.

otelcol/flags.go
```diff
@@ -50,9 +51,12 @@ func flags(reg *featuregate.Registry) *flag.FlagSet {
 			return nil
 		})
 
-	flagSet.Var(featuregate.NewFlag(reg), featureGatesFlag,
-		"Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.")
+	gatesFlag, strictFlag := featuregate.NewFlags(reg)
+	flagSet.Var(strictFlag, featureGatesStrictFlag,
```
mx-psi, line 55: Why is this a `Var` instead of a boolean flag?

mx-psi (commented):
I think this looks almost ready to go from my side, just one more (hopefully last) question about the type of the flag

otelcol/flags.go
atoulme: Because a bool flag does not take a function, and instead returns a bool pointer, depriving me from the possibility to react when the flag is set.

mx-psi: Right, thanks! That makes sense, but I think we should follow what is stated on https://pkg.go.dev/flag#Value:

> If a Value has an IsBoolFlag() bool method returning true, the command-line parser makes -name equivalent to -name=true rather than using the next command-line argument.

And add `IsBoolFlag` to `strictFlagValue`

atoulme: Cool, I'll add it. Thanks for teaching me this.

mx-psi (approved):

otelcol/flags.go
bogdandrutu, line 55: Instead of returning the `strictFlag` which also breaks the API for the featuregate why not adding it as as flag to the `gatesFlag` set?

atoulme: It doesn't change much. It pretty much takes this code and moves it inside the featuregate package I guess. If that's what we want to do, sure.

bogdandrutu: I think I was confused by the API, my proposal doesn't make sense. But I still don't like the new API. 

For me this is an option/property of the `featuregate.NewFlag` so, let's change the old API to `func NewFlag(reg *Registry, opts ...FlagOption) flag.Value` and pass the strictness as option.

atoulme: strictness is not an option passed in, it's a flag. This API creates flags for the user to fill in.

atoulme: Where do we go from here? I am not sure how to address this feedback. I can bring the method `NewFlag` to deprecate it.

djaglowski: > For me this is an option/property of the `featuregate.NewFlag` so, let's change the old API to `func NewFlag(reg *Registry, opts ...FlagOption) flag.Value` and pass the strictness as option.

This doesn't make sense to me. The PR proposes to add the concept of strictness as a user-defined, runtime option. If I'm understanding @bogdandrutu's suggestion correctly, it would have to be a a developer-defined compile-time option, which would serve a very different purpose.

djaglowski (approved):

mx-psi (approved):

featuregate/flag.go
```diff
@@ -5,40 +5,88 @@ package featuregate // import "go.opentelemetry.io/collector/featuregate"
...
 	})
-	return strings.Join(ids, ",")
+	return strings.Join(gates, ",")
 }
 
-func (f *flagValue) Set(s string) error {
-	if s == "" {
-		return nil
+func (f *gatesFlagValue) Set(s string) error {
+	f.args.gates = s
```
Aneurysm9, line 45: Does this prevent the `--feature-gates` flag from being provided multiple times?

atoulme: No, though that was never supported from what I can see.

Aneurysm9: Can you add a test that proves this?  Implementing the `flag.Value` interface should allow the flag to be specified multiple times on the CLI at invocation, resulting in a call to `Set()` for each instance of the flag.

otelcol/flags_test.go
```diff
@@ -54,11 +54,28 @@ func TestSetFlag(t *testing.T) {
...
+		{
+			name:        "feature gates and strict",
+			args:        []string{"--feature-gates=alpha,-beta", "--feature-gates-strict=true"},
+			expectedErr: `invalid boolean value "true" for -feature-gates-strict: gate "beta" is in beta and must be explicitly enabled`,
+		},
+		{
+			name:        "strict flag first and feature gates",
+			args:        []string{"--feature-gates-strict=true", "--feature-gates=alpha,-beta"},
+			expectedErr: `invalid value "alpha,-beta" for flag -feature-gates: gate "beta" is in beta and must be explicitly enabled`,
+		},
```
Aneurysm9, line 70: Varying the behavior based on the order of the flags doesn't seem ideal and may confuse users.

atoulme: I could not find a way around this limitation of go flags.

featuregate/flag.go
Aneurysm9: https://go.dev/play/p/wQG2ktm4uQE demonstrates the concept of having a `flag.Value` specified multiple times.

mx-psi:
This is blocking featuregate 1.0. Is there anything we can do to move it forward? What are the main blockers?

featuregate/flag.go
```diff
@@ -5,40 +5,88 @@ package featuregate // import "go.opentelemetry.io/collector/featuregate"
...
 func NewFlag(reg *Registry) flag.Value {
-	return &flagValue{reg: reg}
+	return &gatesFlagValue{args: &gateRegistrationArgs{reg: reg}}
 }
 
-// flagValue implements the flag.Value interface and directly applies feature gate statuses to a Registry.
-type flagValue struct {
-	reg *Registry
+// NewFlags returns two flag.Values: one that registers gates, and one that sets strict mode.
+func NewFlags(reg *Registry) (flag.Value, flag.Value) {
```
mx-psi, line 22: If we ever want to introduce more flags with this approach we would have to deprecate the `NewFlags` function and add a new one that has one more value. What about returning a struct? e.g.

```go
type Flags struct {
    RegistrationFlag flag.Value
    StrictMode flag.Value
}
```

If this PR is still unclear, I would suggest we make just this change and then this PR does not block 1.0 readiness

mx-psi: I filed #8727 to implement this and decouple the 1.0 release from this change

mx-psi (commented):
PTAL at #8727 which is a modified subset of these changes to move things forward with the 1.0 release

atoulme:
Closing as this work has gone off the rails and having a CLI flag for it is involving breaking changes slowing down the development of the collector.

PR #7997, opened 2023-06-28
Update koanf to v2

dbason:
This PR updates the koanf dependency to v2


codeboten (commented):
Is this addressing an issue or limitation with the older koanf library?

mx-psi:
@knadh If you have the time, can you help us understand what changed between koanf v1 and v2 and what would be the benefits and changes if we update?

knadh:
Hi @mx-psi. Functionally, no changes. The package structure has changed where every provider/parser has become its own module (with its own go.mod) to become a one repository-many modules system.

**Benefits:**
This removes references to all indirect dependencies in the main go.mod significantly de-cluttering it. These would never be compiled into the build, but `go get .../koanf` would still pull them in the dev environment in v1. Check out the main package's [go.mod](https://github.com/knadh/koanf/blob/master/go.mod) in v2.

It is safe (and also easy) to upgrade to v2 as it is functionally identical to v1. While this needn't be done on priority, I would recommend moving to v2 as that is where any continued improvements will land. v1 is frozen and will only receive security updates.

codeboten (commented):
Thanks for the explanation @knadh!

codeboten (commented):
Please run `make gotidy` to tidy up the dependencies

dbason:
I suspect the tests failed due to the github outage.

codeboten:
Needs one more `make crosslink && make gotidy`... not sure why crosslink included a bunch of imports in a previous commit

dbason:
I've done that now and squashed into the last commit

codeboten (approved):
Thanks it's looking good now

PR #7998, opened 2023-06-28
[service/proctelemetry] Offer to override the HOST_PROC environment variable with a programmatic value

### LINKED ISSUE #7435: [service/proctelemetry]  when host_proc env variable is set, metric server uses and reports on the wrong process
**Describe the bug**
While debugging the below error in k8s env

````
Error: failed to register process metrics: process does not exist
2023/03/23 03:44:47 main.go:115: application run finished with error: failed to register process metrics: process does not exist
````
I have noticed that the metric server registers its own process irrespective whether `HOST_PROC` environment variable is set or not.
This becomes an issue when `HOST_PROC` is pointing (ex: k8s environment) to the host `proc` on the host filesystem, as gopsutil will be looking at the host processes to identify the OTEL process instead of looking into the container `/proc` .

In a container, OTEL runs as PID1, and as explained above, the metric server will be reporting metrics of the host PID1 instead. 

**Steps to reproduce**
Deploy OTEL in a k8s cluster with `HOST_PROC` env set to a mounted path of the host proc folder inside the container and make sure the telemetry->metric server is enabled.
(`HOST_PROC` is needed to monitor the host itself.)

the metric server identifies the otel process correctly and send metric (ex:cpu)  of the OTEL process usage

**What did you see instead?**
you will either get an error:
````
Error: failed to register process metrics: process does not exist
2023/03/23 03:44:47 main.go:115: application run finished with error: failed to register process metrics: process does not exist
````
or you will get metrics from the wrong process.

**What version did you use?**
ALL

**Environment**
OS: Linux

**Additional context**
I took a stab at this here https://github.com/open-telemetry/opentelemetry-collector/pull/7434 but this solution does not work well as all stats call in gopsutil will re-read the env variable `HOST_PROC`

We have a PR open in gopsutil https://github.com/shirou/gopsutil/pull/1439 this will be the most robust way to address this issue, but not sure how long it will take to get it merged

### PR DESCRIPTION
**Pull request by atoulme**
Reprising https://github.com/open-telemetry/opentelemetry-collector/pull/7434

**Description:** 

While debugging the below error in k8s env
````
Error: failed to register process metrics: process does not exist
2023/03/23 03:44:47 main.go:115: application run finished with error: failed to register process metrics: process does not exist
````
I have noticed that the metric server is calling GOPSUTIL while the HOST_PROC variable is set , this causes gopsutil `PidExistsWithContext ` to retrieve the process from the host instead from the container 

````
func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
	if pid <= 0 {
		return false, fmt.Errorf("invalid pid %v", pid)
	}
	proc, err := os.FindProcess(int(pid))
	if err != nil {
		return false, err
	}

	if isMount(common.HostProc()) { // if /<HOST_PROC>/proc exists and is mounted, check if /<HOST_PROC>/proc/<PID> folder exists
		_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))
		if os.IsNotExist(err) {
			return false, nil
		}
		return err == nil, err
	}
````
This PR unsets and resets the host_proc variable and introduces an option to allow the use of host_proc if for whatever reason they need to

**Link to tracking Issue:**
Fixes #7435

**Testing:**
unit tests

### CODE REVIEW

File: service/internal/proctelemetry/process_telemetry.go [hunk 1]
```diff
@@ -51,15 +52,19 @@ type processMetrics struct {
...
 	pm := &processMetrics{
 		startTimeUnixNano: time.Now().UnixNano(),
 		ballastSizeBytes:  ballastSizeBytes,
 		ms:                &runtime.MemStats{},
 	}
 
-	pm.proc, err = process.NewProcess(int32(os.Getpid()))
+	ctx := context.Background()
+	if !useHostProcEnvVar {
+		ctx = context.WithValue(ctx, common.EnvKey, common.EnvMap{common.HostProcEnvKey: "/proc"})
```
codeboten, line 65: How does this work on non-unix environments?

service/telemetry/config.go
```diff
@@ -122,6 +123,10 @@ type MetricsConfig struct {
 	// Readers allow configuration of metric readers to emit metrics to
 	// any number of supported backends.
 	Readers []MetricReader `mapstructure:"metric_readers"`
+
+	// UseHostProcEnvVar when set to true, the metric server, through gopsutil,
+	// lookup for the otel process in the proc path defined in the HOST_PROC environment variable.
+	UseHostProcEnvVar bool `mapstructure:"useHostProcEnvVar"`
```
codeboten, line 129: This should not be camel case

dloucasfx (approved):
LGTM !!! the tests are failing but not related to the change

service/internal/proctelemetry/process_telemetry.go
atoulme: Non-unix envs don't rely on this env var at all: https://github.com/shirou/gopsutil/blob/fb83fb22d68f468adf2a01a9ed4963e7d980cfd7/process/process_windows.go#L268

service/telemetry/config.go
atoulme: Fixed, moved to use the host_proc value directly.

.chloggen/metric-server-hostproc.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service/proctelemetry
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Make sure OTEL internal metrics is reading and reporting its own process data when HOST_PROC is set set
```
codeboten, line 8:
```suggestion
note: Make sure OTEL internal metrics is reading and reporting its own process data when HOST_PROC is set
```

service/internal/proctelemetry/process_telemetry_test.go
```diff
@@ -137,10 +137,10 @@ func TestOtelProcessTelemetry(t *testing.T) {
 func TestOCProcessTelemetry(t *testing.T) {
 	ocRegistry := metric.NewRegistry()
 
-	require.NoError(t, RegisterProcessMetrics(ocRegistry, noop.NewMeterProvider(), false, 0))
+	require.NoError(t, RegisterProcessMetrics(ocRegistry, noop.NewMeterProvider(), false, 0, ""))
 
 	// Check that the metrics are actually filled.
-	<-time.After(200 * time.Millisecond)
+	time.Sleep(200 * time.Millisecond)
```
codeboten, line 143: Why the change here?

service/telemetry/config.go
```diff
@@ -122,6 +122,11 @@ type MetricsConfig struct {
 	// Readers allow configuration of metric readers to emit metrics to
 	// any number of supported backends.
 	Readers []MetricReader `mapstructure:"metric_readers"`
+
+	// HostProc is the path to the proc folder used to look up for the collector process.
+	// When unset, the collector will use the environment variable HOST_PROC if set, and
+	// default to `/proc` if no value is set.
```
codeboten, line 128: Should this be called out as a unix-only feature? What should happen if it's configured on a windows system? I understand the underlying library will just ignore the value, but it would be good to set expectations for people using the collector.

Should the default value be set somewhere in case the underlying library ever decided to change the default value? Alternatively, a test that would help identify regressions would be useful here

atoulme: This is a Unix specific change, correct. The env vars were only set with Linux, not other OSes.

service/internal/proctelemetry/process_telemetry_test.go
atoulme: because it seemed like a better way to express that we're waiting 200 milliseconds? I can peel that off into a chore if you like.

service/telemetry/config.go
atoulme: We also set the default values ourselves always, and in any case rely on gopsutil otherwise. Do you want me to call out that this config entry is only applicable to Linux?

service/internal/proctelemetry/process_telemetry_test.go
atoulme: also uniformize with https://github.com/open-telemetry/opentelemetry-collector/pull/7998/files#diff-bbd186f2d03f0a4df651e362301bbbe615e892be0b4f3ad8931937b281b374ddR27

service/telemetry/config.go
mx-psi: Would it make sense to have this configuration option only available on Linux via build tags? Something like:

```go
type MetricsConfig struct {
  // Existing fields go here
  // ...

  // OSSpecificMetricsConfig defines OS-specific metrics configuration
  OSSpecificMetricsConfig `mapstructure:",squash"`
}
```

and then we have different definitions of the `OSSpecificMetricsConfig` struct for Linux (having the `HostProc` setting) and Windows/not Linux (where `OSSpecificMetricsConfig` would just be an empty struct). 

This makes the option available only on Linux for both library users and end-users of the binary, and implicitly documents that the feature is Linux-only.

codeboten: I don't know which would be more confusing to users, to have an error when using the same config across OS's or to apply the same config and have one of the options do nothing on some platforms.

I would expect the behaviour to be the same as if someone tried to use config that isn't available on certain platforms in the hostmetrics receiver

mx-psi: Agree that following precedent on hostmetricsreceiver makes sense if we already have OS-specific options. If we are not sure which path to follow, I think raising an error is better in that we can always change our minds if people complain, while we can't easily do that with the "ignore" behavior.

mx-psi: > I am a bit hesitant about this, would it break the collector to run on Windows with a message saying the config is unrecognized? I can log a warning on Windows if you like, but I'm not sure we should use different configs as it might confuse users.

Yes, it would fail on Windows. I would at least raise a warning. A benefit of failing on Windows is that we can change that behavior later, while if we commit to a specific behavior on Windows it will be hard to change later

atoulme: How about I cut this PR in 2 and first introduce this behavior, and then a separate PR to discuss how we expose this in config?

mx-psi: @atoulme makes sense to me!

atoulme: Thanks, updated the PR.

.chloggen/metric-server-hostproc.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
mx-psi, line 2: We are breaking the API, so this should be `breaking` IMO (at least I think that has been the case in opentelemetry-collector historically)

service/internal/proctelemetry/process_telemetry.go
```diff
@@ -51,15 +53,20 @@ type processMetrics struct {
...
 	var err error
 	pm := &processMetrics{
 		startTimeUnixNano: time.Now().UnixNano(),
 		ballastSizeBytes:  ballastSizeBytes,
 		ms:                &runtime.MemStats{},
 	}
 
-	pm.proc, err = process.NewProcess(int32(os.Getpid()))
+	ctx := context.Background()
+	if hostProc != "" {
```
mx-psi, line 65: I am not a fan of having a special meaning when passing an empty string. Since we are breaking the API I think we can do better, either with a struct with all the options or with the `Option` type pattern

atoulme: Option pattern has my preference.

.chloggen/metric-server-hostproc.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
mx-psi, line 2: Sorry for my confusion :facepalm: since this is a small internal change I think we can drop the release note altogether. Apologies, I did not realize this was internal at first.

.chloggen/metric-server-hostproc.yaml
atoulme: Your comment below seems to indicate this comment is now outdated: https://github.com/open-telemetry/opentelemetry-collector/pull/7998#discussion_r1272095494

I will resolve this thread now.

.chloggen/metric-server-hostproc.yaml
atoulme: OK, so I'll delete the changelog entry.

.chloggen/metric-server-hostproc.yaml
mx-psi: Yes, sorry about that

mx-psi (approved):

codeboten (approved):

Issue #7892, opened 2023-06-14
forward connector: fails validation across pipelines


MovieStoreGuy:
**Describe the bug**
it is possible for a shared reference a forward connector to fail validation due it being used across pipelines

**Steps to reproduce**

Have two different telemetry pipelines that share the forward reference, ie:

```yaml
service:
  pipelines:
    metrics/receiving:
      receivers:  [nop]
      processors: []
      exporters:  [forward]
    metrics/exporting:
      receivers:  [forward]
      processors: []
      exporters:  [nop]
    traces/receiving:
      receivers:  [nop]
      processors: []
      exporters:  [forward]
    traces/exporting:
      receivers:  [forward]
      processors: []
      exporters:  [nop]
```

**What did you expect to see?**
This should start up without issue

**What did you see instead?**
An error occurred during the validation step:

```
Error: failed to build pipelines: connector "forward" cannot connect from metrics to logs: telemetry type is not supported
2023/06/14 10:41:50 collector server run finished with error: failed to build pipelines: connector "forward" cannot connect from metrics to logs: telemetry type is not supported
```

**What version did you use?**
Version: v0.79.0

**What config did you use?**
Shared above

**Environment**
OS: docker linux/arm
Compiler: go 1.20.5

mx-psi:
I am putting a p2 priority on this one since we have a workaround (create separate instances of the connector) and it's not (I think) a regression

djaglowski:
Resolved by #8004

PR #8003, opened 2023-06-29
Fix connector validation based on usage in pipelines

djaglowski:
Fixes #https://github.com/open-telemetry/opentelemetry-collector/issues/7892

Validation of connectors was too aggressive such that a connector that was used in any combination of unsupported roles would fail. Instead, validation should pass as long as each use of the connector has a supported corresponding use.

For example, the forward connector may forward traces and metrics at the same time. Previously, validation would fail because it detected that traces->metrics and metrics->traces were possible connections. Now it will pass as long as there is a supported connection type for each pipeline in which the connector is used.


djaglowski:
Closing in favor of #8004 which I believe is a cleaner solution.

PR #8004, opened 2023-06-29
Fix connector validation based on usage in pipelines

djaglowski:
Alternate to https://github.com/open-telemetry/opentelemetry-collector/pull/8003

Fixes #https://github.com/open-telemetry/opentelemetry-collector/issues/7892

Validation of connectors was too aggressive such that a connector that was used in any combination of unsupported roles would fail. Instead, validation should pass as long as each use of the connector has a supported corresponding use.

For example, the forward connector may forward traces and metrics at the same time. Previously, validation would fail because it detected that traces->metrics and metrics->traces were possible connections. Now it will pass as long as there is a supported connection type for each pipeline in which the connector is used.


djaglowski:
Validation of connectors is currently split between `otelcol` and `service/internal/graph` packages, roughly as follows:

`otelcol`: Obvious problems that can be reasoned about just be looking at the config.
- Invalid connector type
- Connector not used as both exporter and receiver

`service/internal/graph`: Less obvious problems that basically require building the graph / components.
- Connector introduces cycle into graph
- Connector config invalid

The validation added in this PR could arguably be added to either package, as it seems to fall somewhere in between, but having tried adding it to `otelcol` in #8003, I believe this is a better solution. I also pulled the "Connector not used as both exporter and receiver" logic from `otelcol` into this solution because it is very closely related.

service/internal/graph/graph.go
```diff
@@ -98,15 +105,70 @@ func (g *Graph) createNodes(set Settings) {
...
+			if supportedUse {
+				continue
+			}
+			// Retain previous error message for special case where connector is used once, but incorrectly.
+			// TODO remove this case if we agree the slightly more generic error message is acceptable.
+			if len(expTypes) == 1 && len(recTypes) == 1 {
+				for recType := range recTypes {
+					return fmt.Errorf("connector %q cannot connect from %s to %s: telemetry type is not supported", connID, expType, recType)
+				}
+			}
```
djaglowski, line 149: I would prefer to remove this and migrate to the slightly more generic error message seen below, but would appreciate other opinions on whether this is informative enough.

connector/connector.go
```diff
@@ -437,6 +439,39 @@ func (f factory) LogsToLogsStability() component.StabilityLevel {
...
+		case component.DataTypeTraces:
+			return f.logsToTracesStabilityLevel
+		case component.DataTypeMetrics:
+			return f.logsToMetricsStabilityLevel
+		case component.DataTypeLogs:
+			return f.logsToLogsStabilityLevel
+		}
+	}
+	return component.StabilityLevelUndefined
+}
```
djaglowski, line 473: This code doesn't strictly need to be part of the API, but it felt silly to reimplement it elsewhere.

service/internal/graph/graph.go
mx-psi: I think it would be fine to remove this special case, we can always add it back if people find the generic error message confusing.

djaglowski: Good catch. Artifacts from #8003. I've removed all changes to this file.

connector/connectortest/connector.go
```diff
@@ -26,57 +26,62 @@ func NewNopCreateSettings() connector.CreateSettings {
 type nopConfig struct{}
 
 // NewNopFactory returns a connector.Factory that constructs nop processors.
-func NewNopFactory() connector.Factory {
+func NewNopFactory(opts ...connector.FactoryOption) connector.Factory {
```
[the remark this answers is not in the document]
djaglowski: Good catch. Artifacts from #8003. I've removed all changes to this file.

service/internal/graph/graph.go
djaglowski: I've removed the special case and updated expected error messages in tests.

mx-psi (approved):
LGTM, one last thought that we don't need to address here:

> Validation of connectors is currently split between otelcol and service/internal/graph packages

I am wondering if everything should be on `service/internal/graph`; if we are to eventually support #4970/#6888 it seems like the best solution is to expose the graph package, in which case all validation should happen there.

We don't need to solve this on this PR necessarily, since moving the validation around is not a breaking change.

djaglowski:
@mx-psi, thanks for the reviews.

w/r/t where validation is located, I think this PR effectively consolidates it into the `graph` package. The only exception now is the very high level checks like "is this a valid type of connector", which we still do in `otelcol` along with the other components.

mx-psi:
>  which we still do in otelcol along with the other components.

alright, sounds like it's something we should solve independently then :)

codeboten (approved):

PR #8005, opened 2023-06-29
[chore] tidy

codeboten:
Running `make gotidy` locally removes these deps


codeboten:
Will be included in https://github.com/open-telemetry/opentelemetry-collector/pull/7997

PR #8017, opened 2023-07-04
[confmap] fix bug of unmarshalling slice values

### LINKED ISSUE #4001: Config unmarshalling doesn't work properly for slice values

fatsheep9146:
**Description:** 
fix the bug of unmarshalling slice values for package confmap

**Link to tracking Issue:** 
fix #4001

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -157,6 +157,7 @@ func decodeConfig(m *Conf, result any, errorUnused bool) error {
 			mapstructure.StringToTimeDurationHookFunc(),
 			mapstructure.TextUnmarshallerHookFunc(),
 			unmarshalerHookFunc(result),
+			zeroSliceHookFunc(),
 		),
 	}
 	decoder, err := mapstructure.NewDecoder(dc)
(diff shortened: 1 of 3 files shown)


fatsheep9146:
@dmitryax could you help review this pr?

dmitryax:
Please address comments from @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/pull/6163. I think we should not override with `nil` but probably should override with `[]`

fatsheep9146:
> Please address comments from @bogdandrutu in #6163. I think we should not override with `nil` but probably should override with `[]`

@dmitryax 

I think it is already doing like that, in unit test, these two cases are also covered

https://github.com/open-telemetry/opentelemetry-collector/blob/4dc71d6716c20667412e3c2262be22b14db1ce8a/confmap/confmap_test.go#L449-L472

- The default value will not be overrided by `nil`.
- The default value will be overrided by `[]` empty slice.

confmap/confmap.go
```diff
@@ -336,3 +337,34 @@ type Marshaler interface {
...
+//
+// 3. configuration have `keys` field and have nil value for this key, the output should be default config
+//   - for example, input is {"keys": nil}, then output is Config{ Keys: ["a", "b"]}
+//
+// 4. configuration have no `keys` field specified, the output should be default config
+//   - for example, input is {}, then output is Config{ Keys: ["a", "b"]}
+func zeroSliceHookFunc() mapstructure.DecodeHookFuncValue {
+	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
+		if to.CanSet() && to.Kind() == reflect.Slice {
+			to.Set(reflect.MakeSlice(reflect.SliceOf(to.Type().Elem()), 0, 0))
```
dmitryax, line 365: Shouldn't this be?
```suggestion
			to.Set(reflect.MakeSlice(to.Type(), to.Len(), to.Cap()))
```

fatsheep9146: I think this should be `to.Set(reflect.MakeSlice(to.Type(), from.Len(), from.Cap()))`, since we use `from` to set `to`

dmitryax: Why `reflect.SliceOf(to.Type().Elem())`? `to.Type()` must be the slice type already

fatsheep9146:
still doing

dmitryax (approved):

fatsheep9146:
@open-telemetry/collector-maintainers 
could you help review about this pr?

codeboten (approved):
Thanks for including the detailed documentation @fatsheep9146

PR #8040, opened 2023-07-05
[pdata] Fix Map.EnsureCapacity bug when the Map is not empty

tigrannajaryan:
EnsureCapacity was incorrectly clearing the Map when increasing the capacity. It will now correctly preserve existing elements.

### DIFF
--- a/pdata/pcommon/map_test.go
+++ b/pdata/pcommon/map_test.go
@@ -375,6 +375,38 @@ func TestMap_EnsureCapacity(t *testing.T) {
 	assert.Equal(t, 8, cap(*am.getOrig()))
 }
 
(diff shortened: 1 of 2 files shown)


pdata/pcommon/map_test.go
```diff
@@ -375,6 +375,38 @@ func TestMap_EnsureCapacity(t *testing.T) {
 	assert.Equal(t, 8, cap(*am.getOrig()))
 }
 
+func TestMap_EnsureCapacity_Exiting(t *testing.T) {
```
codeboten, line 378:
```suggestion
func TestMap_EnsureCapacity_Existing(t *testing.T) {
```

codeboten (approved):
Minor typo, otherwise looks good.

PR #8046, opened 2023-07-10
[chore] update jsonschema

codeboten:
- `make genjsonschema`
- remove duplicate type

### DIFF
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -93,19 +93,6 @@ type LogsSamplingConfig struct {
 	Thereafter int `mapstructure:"thereafter"`
 }
 
-// MetricReader exposes configuration of metric readers to end users.
-// TODO: replace this temporary struct w/ auto-generated struct from jsonschema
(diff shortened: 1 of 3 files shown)


service/telemetry/config.go
```diff
@@ -93,19 +93,6 @@ type LogsSamplingConfig struct {
...
-// TODO: replace this temporary struct w/ auto-generated struct from jsonschema
-// https://github.com/open-telemetry/opentelemetry-configuration/tree/main/schema
-//
-// Experimental: *NOTE* this structure is subject to change or removal in the future.
-type MetricReader struct {
-	// Args corresponds to the JSON schema field "args".
-	Args any `mapstructure:"args"`
-
-	// Type corresponds to the JSON schema field "type".
-	Type string `mapstructure:"type"`
```
dmitryax, line 106: Given that these fields were released previously and removed now, should we still mark it as a breaking change even if it's marked as experimental?

codeboten: Added a changelog.

dmitryax (approved):

PR #8097, opened 2023-07-17
Add support for OTLP export for internal metrics

### LINKED ISSUE #7641: Support additional exporters for exporting collector metrics

codeboten:
Internal collector metrics can now be exported via OTLP
using the otel-go otlpgrpc and otlphttp exporters.

Closes #7641 

Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/8113


service/internal/proctelemetry/config.go
```diff
@@ -171,7 +187,69 @@ func initPeriodicExporter(_ context.Context, exporter telemetry.MetricExporter)
...
+		return sdkmetric.NewPeriodicReader(exp, opts...), nil, nil
 	}
 	return nil, nil, fmt.Errorf("no valid exporter")
 }
+
+func initOTLPgRPCExporter(ctx context.Context, otlpConfig *telemetry.OtlpMetric) (sdkmetric.Exporter, error) {
+	opts := []otlpmetricgrpc.Option{}
+
+	if len(otlpConfig.Endpoint) > 0 {
+		opts = append(opts, otlpmetricgrpc.WithEndpoint(otlpConfig.Endpoint))
```
jaronoff97, line 215: should this be a required option?

TylerHelmuth (commented):
Hype!

jaronoff97 (approved):
🥳

service/internal/proctelemetry/config.go
codeboten: > By default, if an environment variable is not set, and this option is not passed, "localhost:4317" will be used. 

It is not required: https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc#WithEndpoint

service/internal/proctelemetry/config.go
```diff
@@ -171,7 +187,69 @@ func initPeriodicExporter(_ context.Context, exporter telemetry.MetricExporter)
...
+	return otlpmetricgrpc.New(ctx, opts...)
+}
+
+func initOTLPHTTPExporter(ctx context.Context, otlpConfig *telemetry.OtlpMetric) (sdkmetric.Exporter, error) {
+	opts := []otlpmetrichttp.Option{}
+
+	if len(otlpConfig.Endpoint) > 0 {
+		opts = append(opts, otlpmetrichttp.WithEndpoint(otlpConfig.Endpoint))
+	}
+	// if otlpConfig.Compression != nil {
```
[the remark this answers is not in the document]
codeboten: Thanks for the reminder, I'll implement it before moving it out of draft

service/internal/proctelemetry/config_test.go
```diff
@@ -123,6 +123,55 @@ func TestMetricReader(t *testing.T) {
 				},
 			},
 		},
+		{
```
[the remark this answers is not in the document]
codeboten: Added 👍🏻

service/internal/proctelemetry/config.go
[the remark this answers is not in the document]
codeboten: Implemented.

service/internal/proctelemetry/config.go
```diff
@@ -182,5 +191,100 @@ func initPeriodicExporter(_ context.Context, exporter telemetry.MetricExporter,
...
+		if u.Scheme == "http" {
+			opts = append(opts, otlpmetricgrpc.WithInsecure())
+		}
+	}
+
+	if otlpConfig.Compression != nil {
+		opts = append(opts, otlpmetricgrpc.WithCompressor(*otlpConfig.Compression))
+	}
+	if otlpConfig.Timeout != nil {
+		opts = append(opts, otlpmetricgrpc.WithTimeout(time.Millisecond*time.Duration(*otlpConfig.Timeout)))
```
mx-psi, line 238: Why is `time.Millisecond` needed? Other timeouts just use `time.Duration` https://github.com/open-telemetry/opentelemetry-collector/blob/8949d1da72143abe97fa767e90721d85d79fe6b5/exporter/exporterhelper/common.go#L20 and rely on `ParseDuration`. I think we should do that here if possible

mx-psi (commented):
A few comments on the config. Since they were not added by this PR we may consider them separately if this is not the first use of those fields

service/internal/proctelemetry/config.go
codeboten: The spec defines duration values in milliseconds: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#duration

If we were to rely on `ParseDuration`, that would cause us to differ from the generated jsonschema, unless I'm misunderstanding your suggestion. jsonschema does support duration in the format of [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339#appendix-A) but i'm not sure if we'd want to support that.

service/internal/proctelemetry/config_test.go
```diff
@@ -123,6 +124,218 @@ func TestMetricReader(t *testing.T) {
...
+			name: "periodic/otlp-grpc-exporter-no-endpoint",
+			reader: telemetry.MetricReader{
+				Periodic: &telemetry.PeriodicMetricReader{
+					Exporter: telemetry.MetricExporter{
+						Otlp: &telemetry.OtlpMetric{
+							Protocol:    "grpc/protobuf",
+							Compression: strPtr("gzip"),
+							Timeout:     intPtr(1000),
+							Headers: map[string]interface{}{
+								"test": "test1",
```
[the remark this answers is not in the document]
codeboten: Good point, i've opened a PR to the configuration repo to change this to a map of strings: https://github.com/open-telemetry/opentelemetry-configuration/pull/32

To make this a map of configopaque string i would need to run a substitution on the generated code.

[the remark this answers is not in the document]
codeboten: Well it turns out because of a limitation of go-jsonschema, we need a patch anyways. I've opened the PR to make this a map[string]string. I'll take a look at implementing this functionality in the library itself, but for now https://github.com/open-telemetry/opentelemetry-collector/pull/8113

service/internal/proctelemetry/config.go
mx-psi: I see. There are a few principles here:

1. *Internal Collector configuration consistency*: When specifying the same kind of value on the Collector's configuration (e.g. timeouts), the same syntax should be used everywhere
2. *Consistency with specification*: Inherited from Configuration working group. We should be consistent with whatever way of expressing this is stated in the spec.
3. *Ability to implement with JSON schema*: Whatever we choose should be easy to implement with jsonschema without applying patches on top of the generated code.
4. *Ease of usage*: Specifying common timeout values should be as simple as possible.

We cant' pick all (specifically, I think the `time.Duration` syntax addresses (4) but is incompatible with (2), and (1) is incompatible with (2) unless we make a breaking change somewhere).

I was favoring (1) and (4) over (2) and (3), but I am undecided on whether this is the right choice, specially considering (2) seems very important and making non-Go implementations support `time.Duration`'s syntax seems non-trivial. Do you think we should favor (2) and (3)?

codeboten: In general I agree w/ you that 1 & 4 are preferable. I've pinged the configuration wg w/ a question of what to do here. The spec calls for timeout in millis, but java has an [implementation](https://github.com/open-telemetry/opentelemetry-java/blob/2d6b26b10d47f66290ed76ad7f9e4b5ec22dc161/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/internal/DefaultConfigProperties.java#L146-L175) that supports for both millis and a duration that's closer to what the collector supports today.

Since this is all behind a feature gate, we could always support millis to begin and switch it if the decision changes in the future, wdyt?

mx-psi: Okay, let's go with that :)

mx-psi (approved):

PR #8113, opened 2023-07-19
[chore] update jsonschema

codeboten:
This updates the jsonschema to the latest in main. It additionally adds a sed patch for making Headers a map[string]string instead of a map[string]interface.

@mx-psi

### DIFF
--- a/internal/tools/jsonschema_patch.sed
+++ b/internal/tools/jsonschema_patch.sed
@@ -1,0 +1,4 @@
+# go-jsonschema always generates patternProperties as
+# map[string]interface{}, for more specific types, they must
+# be replaced here
+s+type Headers.*+type Headers map[string]string+g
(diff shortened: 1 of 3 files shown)


internal/tools/jsonschema_patch.sed
```diff
@@ -0,0 +1,4 @@
+# go-jsonschema always generates patternProperties as
+# map[string]interface{}, for more specific types, they must
+# be replaced here
+s+type Headers.*+type Headers map[string]string+g
```
mx-psi, line 4: nit: I think an actual patch to be applied by `patch` would be more resilient here (not a fan of sed specially after open-telemetry/opentelemetry-collector-contrib/issues/23981)

mx-psi (approved):
Approving with a suggestion we can deal with now or later

Issue #8106, opened 2023-07-18
Support additional exporters for exporting collector traces


codeboten:

bogdandrutu:
Please document the list of exporters you plan to support. I think we should for the moment only support otlp + stdout.

PR #8117, opened 2023-07-20
add span processor configuration

codeboten:
This PR adds support for configuring batch span processor for internal
collector traces. It adds support to export them via the stdout exporter.

Linked issue https://github.com/open-telemetry/opentelemetry-collector/issues/8106

Follows https://github.com/open-telemetry/opentelemetry-collector/pull/8097

PR #8119, opened 2023-07-20
add OTLP export support for internal traces 

codeboten:
This PR adds support for exporting internal collector traces via the
OTel Go OTLP exporters.

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/8106

~Follows https://github.com/open-telemetry/opentelemetry-collector/pull/8117~

### DIFF
--- a/service/internal/proctelemetry/config.go
+++ b/service/internal/proctelemetry/config.go
@@ -20,6 +20,8 @@ import (
 	"go.opentelemetry.io/otel/bridge/opencensus"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
 	otelprom "go.opentelemetry.io/otel/exporters/prometheus"
 	"go.opentelemetry.io/otel/exporters/stdout/stdoutmetric"
 	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
@@ -83,7 +85,7 @@ func InitMetricReader(ctx context.Context, reader telemetry.MetricReader, asyncE
 	return nil, nil, fmt.Errorf("unsupported metric reader type %v", reader)
 }
 
-func InitSpanProcessor(_ context.Context, processor telemetry.SpanProcessor) (sdktrace.SpanProcessor, error) {
+func InitSpanProcessor(ctx context.Context, processor telemetry.SpanProcessor) (sdktrace.SpanProcessor, error) {
 	if processor.Batch != nil {
 		if processor.Batch.Exporter.Console != nil {
 			exp, err := stdouttrace.New(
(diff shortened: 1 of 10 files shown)


service/telemetry/config.go
```diff
@@ -150,7 +150,7 @@ func (sp *SpanProcessor) Unmarshal(conf *confmap.Conf) error {
 	}
 
 	if sp.Batch != nil {
-		if sp.Batch.Exporter.Console == nil {
+		if sp.Batch.Exporter.Console == nil && sp.Batch.Exporter.Otlp == nil {
```
bogdandrutu, line 153: I think I commented on a different PR and cannot find it. But I will repeat here:
* This should be part of `Validate` and should be belong to the deepest object `sp.Batch.Exporter`.

codeboten: This is done now. I found another couple of validations (for the pull and period metric readers) which I moved into their own validate methods as well

TylerHelmuth (approved):

service/internal/proctelemetry/config.go
```diff
@@ -92,32 +94,23 @@ func InitSpanProcessor(_ context.Context, processor telemetry.SpanProcessor) (sd
...
+		if processor.Batch.Exporter.Otlp != nil {
+			var err error
+			var exp sdktrace.SpanExporter
+			switch processor.Batch.Exporter.Otlp.Protocol {
+			case protocolProtobufHTTP:
+				exp, err = initOTLPHTTPSpanExporter(ctx, processor.Batch.Exporter.Otlp)
+			case protocolProtobufGRPC:
+				exp, err = initOTLPgRPCSpanExporter(ctx, processor.Batch.Exporter.Otlp)
+			default:
+				return nil, fmt.Errorf("unsupported protocol %s", processor.Batch.Exporter.Otlp.Protocol)
```
Aneurysm9, line 108:
```suggestion
				return nil, fmt.Errorf("unsupported protocol %q", processor.Batch.Exporter.Otlp.Protocol)
```

service/internal/proctelemetry/config.go
```diff
@@ -335,3 +328,97 @@ func initOTLPHTTPExporter(ctx context.Context, otlpConfig *telemetry.OtlpMetric)
...
+		switch *otlpConfig.Compression {
+		case "gzip":
+			opts = append(opts, otlptracehttp.WithCompression(otlptracehttp.GzipCompression))
+		case "none":
+			opts = append(opts, otlptracehttp.WithCompression(otlptracehttp.NoCompression))
+		default:
+			return nil, fmt.Errorf("unsupported compression %q", *otlpConfig.Compression)
+		}
+	}
+	if otlpConfig.Timeout != nil {
```
Aneurysm9, line 386:
```suggestion
	if otlpConfig.Timeout != nil && *otlpConfig.Timeout > 0 {
```

service/internal/proctelemetry/config.go
```diff
@@ -335,3 +328,97 @@ func initOTLPHTTPExporter(ctx context.Context, otlpConfig *telemetry.OtlpMetric)
...
+		opts = append(opts, otlptracegrpc.WithEndpoint(u.Host))
+		if u.Scheme == "http" {
+			opts = append(opts, otlptracegrpc.WithInsecure())
+		}
+	}
+
+	if otlpConfig.Compression != nil {
+		opts = append(opts, otlptracegrpc.WithCompressor(*otlpConfig.Compression))
+	}
+	if otlpConfig.Timeout != nil {
```
Aneurysm9, line 349:
```suggestion
	if otlpConfig.Timeout != nil && *otlpConfig.Timeout > 0 {
```

service/internal/proctelemetry/config.go
```diff
@@ -335,3 +328,97 @@ func initOTLPHTTPExporter(ctx context.Context, otlpConfig *telemetry.OtlpMetric)
...
+			return nil, err
+		}
+		opts = append(opts, otlptracegrpc.WithEndpoint(u.Host))
+		if u.Scheme == "http" {
+			opts = append(opts, otlptracegrpc.WithInsecure())
+		}
+	}
+
+	if otlpConfig.Compression != nil {
+		opts = append(opts, otlptracegrpc.WithCompressor(*otlpConfig.Compression))
```
Aneurysm9, line 347: Does this require the same validation of the compression value as is done for HTTP?

service/internal/proctelemetry/config.go
```diff
@@ -335,3 +328,97 @@ func initOTLPHTTPExporter(ctx context.Context, otlpConfig *telemetry.OtlpMetric)
...
+	opts := []otlptracegrpc.Option{}
+
+	if len(otlpConfig.Endpoint) > 0 {
+		u, err := url.ParseRequestURI(normalizeEndpoint(otlpConfig.Endpoint))
+		if err != nil {
+			return nil, err
+		}
+		opts = append(opts, otlptracegrpc.WithEndpoint(u.Host))
+		if u.Scheme == "http" {
+			opts = append(opts, otlptracegrpc.WithInsecure())
```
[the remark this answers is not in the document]
codeboten: There is a insecure env variable, however the [specification does say](https://github.com/open-telemetry/opentelemetry-specification/blob/bc76c97525eef0f81550d939a3f87f44b182cef9/specification/protocol/exporter.md#configuration-options) that:

> A scheme of http indicates an insecure connection and takes precedence over the insecure configuration setting

This is aligned w/ the go implementation that parses the env variables: 
https://github.com/open-telemetry/opentelemetry-go/blob/f15ae160c4acfe3872455f87e6133af87859172b/exporters/otlp/otlptrace/internal/otlpconfig/envconfig.go#L104-L111

service/internal/proctelemetry/config.go
codeboten: Is calling `WithTimeout` with a 0 value not valid?

MovieStoreGuy (approved):
Everything looks fun to me, I do suggest adding the checks for `Timeout > 0` to avoid issues were the gRPC starts correctly but does not work because it will always timeout.

service/internal/proctelemetry/config.go
codeboten: Done 👍🏻

PR #2020, opened 2020-10-28
Improve diagnostic logging for exporters

tigrannajaryan:
This change introduces diagnostic debug, info and error logging in
queued_retry exporter helpers. The logging will be applied to all
exporters which use the helpers.

All logging is heavily sampled to avoid flooding the logs. We will
likely need to revise the contributing guide here to clarify that
such sampled logging is allowed:
https://github.com/open-telemetry/opentelemetry-collector/blob/master/CONTRIBUTING.md#logging

We will likely also want to fix the logging for batch processor
which currently outputs not very user friendly messages on failures.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/2013

Sample output with otlphttp exporter and destination unavailable:

```
{"level":"info","ts":1603846135.431853,"caller":"service/service.go:252","msg":"Everything is ready. Begin running and processing data."}
{"level":"info","ts":1603846140.7589068,"caller":"exporterhelper/queued_retry.go:250","msg":"Exporting failed. Will retry the request after interval.","component_kind":"exporter","component_type":"otlphttp","component_name":"otlphttp","error":"failed to make an HTTP request: Post \"http://localhost:1234\": dial tcp [::1]:1234: connect: connection refused","interval":"3.571319363s"}
{"level":"info","ts":1603846151.5628319,"caller":"exporterhelper/queued_retry.go:250","msg":"Exporting failed. Will retry the request after interval.","component_kind":"exporter","component_type":"otlphttp","component_name":"otlphttp","error":"failed to make an HTTP request: Post \"http://localhost:1234\": dial tcp [::1]:1234: connect: connection refused","interval":"13.463090615s"}
[...]

PR #8134, opened 2023-07-26
[telemetrySettting] Create sampled Logger

antonjim-te:
**Issue**: 

In one of the use cases which we are working on, we have identified that `sampledLogger` mechanism used in the Opentelemetry exporters (`otlpexporter` and `otlphttpexporter`), could become a memory bottleneck, when the number of exports in the same pipeline is high (we are talking about 100s of instances of exporters running on the same machine).

The sampledLogger is created inside the exporterhelper module and the sampling is always activated unless the collector is in debug mode:

- https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L88

```
func newQueuedRetrySender(id component.ID, signal component.DataType, qCfg QueueSettings, rCfg RetrySettings, reqUnmarshaler internal.RequestUnmarshaler, nextSender requestSender, logger *zap.Logger) *queuedRetrySender {
	retryStopCh := make(chan struct{})
	sampledLogger := createSampledLogger(logger)
```
- https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry.go#L269
```
func createSampledLogger(logger *zap.Logger) *zap.Logger {
	if logger.Core().Enabled(zapcore.DebugLevel) {
		// Debugging is enabled. Don't do any sampling.
		return logger
	}

	// Create a logger that samples all messages to 1 per 10 seconds initially,
	// and 1/100 of messages after that.
	opts := zap.WrapCore(func(core zapcore.Core) zapcore.Core {
		return zapcore.NewSamplerWithOptions(
			core,
			10*time.Second,
			1,
			100,
		)
	})
	return logger.WithOptions(opts)
}
```
**Investigation:**

The high memory usage has been discovered using https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/pprofextension

- In the first scenario, default configuration with the `sampledLogger` active, based on the profiled data we could observe that sampling mechanism was responsible for allocating aprox. 75% of the memory of the running collector:
image

- In the second scenario, with the same number of running exporters, but with the `sampledLogger` mechanism disabled (using `debug` level configuration check configuration below), we were able to observe a much smaller memory footprint:
```
    service:
      telemetry:
        logs:
          level: debug
```
image

Given that the logging sampling mechanism could become a bottleneck in certain scenarios (our case), we would like to propose a contribution, which potentially can improve the memory footprint of the exporters in certain use cases (potentially making it more scalable)

**Proposal:**

Create an extra sampled logger in the telemetry configuration used to avoid flooding the logs with messages that are repeated frequently

The sampled logger is configured by `LogsSamplingConfig`
```
service:
    telemetry:
        logs:
            sampling:
```

Default configuration for the sampled logger 

```
service:
    telemetry:
        logs:
            sampling:
	        initial: 1
	        thereafter: 100
	        tick: 10s
```


exporter/exporterhelper/README.md
```diff
@@ -24,6 +24,8 @@ The following configuration options can be modified:
       [the batch processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor)
       is used, the metric `batch_send_size` can be used for estimation)
 - `timeout` (default = 5s): Time to wait per individual attempt to send data to a backend
+- `sampled_logger`: Samples logging messages, which caps the CPU and I/O load of logging while keeping a representative subset of your logs.
```
xvilaca-te, line 27:
```suggestion
- `sampled_logger`: Enables/disables sampling of logging messages, which caps the CPU and I/O load of logging while keeping a representative subset of your logs.
```

exporter/exporterhelper/common.go
```diff
@@ -27,6 +27,21 @@ func NewDefaultTimeoutSettings() TimeoutSettings {
 	}
 }
 
+// SampledLoggerSettings samples logging messages, which caps the CPU and I/O load of logging while keeping a
+// representative subset of your logs.
+// Its purpose is to balance between the need for comprehensive logging and the potential performance impact of logging too much data.
+type SampledLoggerSettings struct {
+	// Enable the sampledLogger
```
xvilaca-te, line 34:
```suggestion
	// Enable/disable the sampledLogger
```

exporter/exporterhelper/common.go
```diff
@@ -27,6 +27,21 @@ func NewDefaultTimeoutSettings() TimeoutSettings {
 	}
 }
 
+// SampledLoggerSettings samples logging messages, which caps the CPU and I/O load of logging while keeping a
```
xvilaca-te, line 30:
```suggestion
// SampledLoggerSettings configure the logging sampler, which caps the CPU and I/O load of logging while keeping a
```

exporter/exporterhelper/queued_retry.go
```diff
@@ -266,12 +266,16 @@ func NewDefaultRetrySettings() RetrySettings {
...
 }
 
-func createSampledLogger(logger *zap.Logger) *zap.Logger {
+func createSampledLogger(logger *zap.Logger, lCfg SampledLoggerSettings) *zap.Logger {
 	if logger.Core().Enabled(zapcore.DebugLevel) {
 		// Debugging is enabled. Don't do any sampling.
 		return logger
 	}
 
+	if !lCfg.Enabled {
```
xvilaca-te, line 275: Should we merge this with the above if?

exporter/exporterhelper/queued_retry_test.go
```diff
@@ -617,6 +619,56 @@ func TestQueuedRetryPersistentEnabled_shutdown_dataIsRequeued(t *testing.T) {
 	}, time.Second, 1*time.Millisecond)
 }
 
+func TestCreateSampledLogger(t *testing.T) {
+	testCases := []struct {
+		desc                 string
```
xvilaca-te, line 624: This is kind of our way of writing parametrised tests. The test cases above seem simpler. Not sure if they will want a test function per test case with a name using snake case.

exporter/otlphttpexporter/README.md
```diff
@@ -31,6 +31,8 @@ The following settings can be optionally configured:
 - `timeout` (default = 30s): HTTP request time limit. For details see https://golang.org/pkg/net/http/#Client
 - `read_buffer_size` (default = 0): ReadBufferSize for HTTP client.
 - `write_buffer_size` (default = 512 * 1024): WriteBufferSize for HTTP client.
+- `sampled_logger`: Samples logging messages, which caps the CPU and I/O load of logging while keeping a representative subset of your logs.
```
xvilaca-te, line 34: Same as above.

exporter/exporterhelper/queued_retry_test.go
antonjim-te: They have a similar test in this file https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queued_retry_test.go#L389
```
func TestGetRetrySettings(t *testing.T) {
	getStorageClientError := errors.New("unable to create storage client")
	testCases := []struct {
		desc           string
		storage        storage.Extension
		numStorages    int
		storageIndex   int
		expectedError  error
		getClientError error
	}{
		{
			desc:          "obtain storage extension by name",
			numStorages:   2,
			storageIndex:  0,
			expectedError: nil,
		},
		{
			desc:          "fail on not existing storage extension",
			numStorages:   2,
			storageIndex:  100,
			expectedError: errNoStorageClient,
		},
		{
			desc:          "invalid extension type",
			numStorages:   2,
			storageIndex:  100,
			expectedError: errNoStorageClient,
		},
		{
			desc:           "fail on error getting storage client from extension",
			numStorages:    1,
			storageIndex:   0,
			expectedError:  getStorageClientError,
			getClientError: getStorageClientError,
		},
	}

	for _, tC := range testCases {
		t.Run(tC.desc, func(t *testing.T) {
			storageID := component.NewIDWithName("file_storage", strconv.Itoa(tC.storageIndex))

			var extensions = map[component.ID]component.Component{}
			for i := 0; i < tC.numStorages; i++ {
				extensions[component.NewIDWithName("file_storage", strconv.Itoa(i))] = &mockStorageExtension{GetClientError: tC.getClientError}
			}
			host := &mockHost{ext: extensions}
			ownerID := component.NewID("foo_exporter")

			// execute
			client, err := toStorageClient(context.Background(), storageID, host, ownerID, component.DataTypeTraces)

			// verify
			if tC.expectedError != nil {
				assert.ErrorIs(t, err, tC.expectedError)
				assert.Nil(t, client)
			} else {
				assert.NoError(t, err)
				assert.NotNil(t, client)
			}
		})
	}
}
```

exporter/exporterhelper/README.md
```diff
@@ -24,6 +24,8 @@ The following configuration options can be modified:
       [the batch processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor)
       is used, the metric `batch_send_size` can be used for estimation)
 - `timeout` (default = 5s): Time to wait per individual attempt to send data to a backend
+- `sampled_logger`: Enables/disables sampling of logging messages, which caps the CPU and I/O load of logging while keeping a representative subset of your logs.
```
sboychen-te, line 27:
```suggestion
- `sampled_logger`: Enables/disables sampling of the log messages.
```

we could also replace `Enables/disables` by `Toggles`

exporter/exporterhelper/common.go
```diff
@@ -27,6 +27,20 @@ func NewDefaultTimeoutSettings() TimeoutSettings {
 	}
 }
 
+// SampledLoggerSettings configures the logger sampler, which caps the CPU and I/O load of logging while keeping a representative subset of your logs.
+// Its purpose is to balance between the need for comprehensive logging and the potential performance impact of logging too much data.
```
sboychen-te, line 31: Is this sentence related to logger sampler? If it is I would remove it, would only focus describing settings

exporter/exporterhelper/queued_retry.go
```diff
@@ -266,8 +266,8 @@ func NewDefaultRetrySettings() RetrySettings {
 	}
 }
 
-func createSampledLogger(logger *zap.Logger) *zap.Logger {
-	if logger.Core().Enabled(zapcore.DebugLevel) {
+func createSampledLogger(logger *zap.Logger, lCfg SampledLoggerSettings) *zap.Logger {
+	if !lCfg.Enabled || logger.Core().Enabled(zapcore.DebugLevel) {
```
sboychen-te, line 270: I feel like there is a bit of contradictory behavior, it makes sense in terms of backwards compatibility and the code, but conceptually I am not sure (we might need to document it at least):

"Sampler is enabled and logs are in debug mode"

In this case the code will not enable the sampler, but since we have a very explicit configuration which literally says "I want sampling", not sure if this combination might confuse people

sboychen-te: I would at least document this behavior in case we will not be proposing the behavior change which will allow the property to overwrite debug mode setting

exporter/otlpexporter/config.go
```diff
@@ -17,7 +17,8 @@ type Config struct {
 	exporterhelper.QueueSettings   `mapstructure:"sending_queue"`
 	exporterhelper.RetrySettings   `mapstructure:"retry_on_failure"`
 
-	configgrpc.GRPCClientSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	configgrpc.GRPCClientSettings        `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	exporterhelper.SampledLoggerSettings `mapstructure:"sampled_logger"`
```
sboychen-te, line 21: Would move this up, next to 
```
exporterhelper.RetrySettings   `mapstructure:"retry_on_failure"`
```

exporter/otlphttpexporter/README.md
```diff
@@ -31,6 +31,8 @@ The following settings can be optionally configured:
 - `timeout` (default = 30s): HTTP request time limit. For details see https://golang.org/pkg/net/http/#Client
 - `read_buffer_size` (default = 0): ReadBufferSize for HTTP client.
 - `write_buffer_size` (default = 512 * 1024): WriteBufferSize for HTTP client.
+- `sampled_logger`: Enables/disables sampling of logging messages, which caps the CPU and I/O load of logging while keeping a representative subset of your logs.
```
sboychen-te, line 34:
```suggestion
- `sampled_logger`: Enables/disables sampling of the log messages.
```

exporter/otlphttpexporter/config.go
```diff
@@ -25,6 +25,8 @@ type Config struct {
 
 	// The URL to send logs to. If omitted the Endpoint + "/v1/logs" will be used.
 	LogsEndpoint string `mapstructure:"logs_endpoint"`
+
+	exporterhelper.SampledLoggerSettings `mapstructure:"sampled_logger"`
```
sboychen-te, line 29: Would move this up

exporter/exporterhelper/README.md
antonjim-te: Keeping `Enables/disables ` for the moment

exporter/exporterhelper/queued_retry.go
```diff
@@ -266,8 +266,8 @@ func NewDefaultRetrySettings() RetrySettings {
...
 		// Debugging is enabled. Don't do any sampling.
```
sboychen-te, line 271:
```suggestion
		// Debugging is enabled or log sampling is disabled. Don't do any sampling.
```

exporter/otlpexporter/config.go
```diff
@@ -13,11 +13,11 @@ import (
...
 type Config struct {
-	exporterhelper.TimeoutSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
-	exporterhelper.QueueSettings   `mapstructure:"sending_queue"`
-	exporterhelper.RetrySettings   `mapstructure:"retry_on_failure"`
-
-	configgrpc.GRPCClientSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	exporterhelper.TimeoutSettings       `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	exporterhelper.QueueSettings         `mapstructure:"sending_queue"`
+	exporterhelper.RetrySettings         `mapstructure:"retry_on_failure"`
+	exporterhelper.SampledLoggerSettings `mapstructure:"sampled_logger"`
```
sboychen-te, line 19: Better to leave one line space here, like it was before

sboychen-te (approved):

xvilaca-te (approved):

pedro-te (approved):

antonjim-te:
Hi @bogdandrutu, could you take a look at the PR please?
Thank you.

antonjim-te:
Hi @jpkrohling @bogdandrutu @mx-psi @jmacd @dmitryax @codeboten @open-telemetry/collector-approvers,
Please review it whenever you have a chance. It will solve a memory issue that we are facing.
Thank you in advance.

exporter/exporterhelper/queued_retry.go
```diff
@@ -273,9 +273,9 @@ func NewDefaultRetrySettings() RetrySettings {
 	}
 }
 
-func createSampledLogger(logger *zap.Logger) *zap.Logger {
-	if logger.Core().Enabled(zapcore.DebugLevel) {
-		// Debugging is enabled. Don't do any sampling.
+func createSampledLogger(logger *zap.Logger, lCfg SampledLoggerSettings) *zap.Logger {
```
jpkrohling, line 276: @codeboten, do you know why we create a sampled logger here? Shouldn't loggers be created based on the telemetry configuration alone? Instead of this PR, I would consider removing this sampled logger altogether.

mx-psi: This was added in #2020 by @tigrannajaryan. The PR description says

> All logging is heavily sampled to avoid flooding the lo

exporter/exporterhelper/README.md
```diff
@@ -24,6 +24,8 @@ The following configuration options can be modified:
       [the batch processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor)
       is used, the metric `batch_send_size` can be used for estimation)
 - `timeout` (default = 5s): Time to wait per individual attempt to send data to a backend
+- `sampled_logger`: Enables/disables sampling of logging messages.
```
tigrannajaryan, line 27: It is not clear what does enabling/disabling sampling do. Please expand the description.

antonjim-te: @tigrannajaryan thank you for your feedback.
Is it now more clear? Am I missing something?

Thanks

exporter/exporterhelper/queued_retry.go
jpkrohling: Right, and shouldn't this be a collector-wide setting, instead of something exclusively to this helper?

mx-psi: I think this was meant to be specific to this helper to avoid flooding the logs in the case where all requests fail, and you log once per request. The [contributing guidelines][1] say

> Use logging carefully for events that can happen frequently to avoid flooding the logs. Avoid outputting logs per a received or processed data item since this can amount to very large number of log entries (Collector is designed to process many thousands of spans and metrics per second). For such high-frequency events instead of logging consider adding an internal metric and increment it when the event happens.

I guess at the time this was written metrics were not really available and thus we went with sampled logs.

(Not sure if this is a good reason today to keep this, just trying to reconstruct the history of this to make a better decision)

[1]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#logging

jpkrohling: Same here: I'm sure this was a good decision by then, not so sure the next iteration should be to expose a setting for this instead of using a metric and/or exposing log sampling as part of the collector's telemetry config.

mx-psi: @tigrannajaryan any thoughts on the above?

tigrannajaryan: I think a possible future option is to supply 2 loggers to every component: one that has no sampling (outputs 100% of logs) and another that is sampled. Components will use the no-sampled logger for everything that is critical to be printed and is known to be non-repeating so it can't flood logs (e.g. startup and shutdown messages. The messages that can repeat for every data item or request will use the sampled logger. The sampled logger can derive its sampling ratio from a central setting (somewhere in the `telemetry` config section).

I do not think merely passing a single sampled logger to every component to be used for every message is good enough since it can result in critical one-off messages being lost.

On a side note: the memory usage by sampled logger that this PR attempts to address is puzzling. Maybe try to figure out why and reduce that instead of disabling sampling?

antonjim-te: Hi @tigrannajaryan answering the side note. As you can see on the graph https://user-images.githubusercontent.com/5960625/255850931-f9b3e891-2695-4aac-b8d6-9d60cb0dfb9d.png the memory problem is coming from the logger sampling. We also tried modifying the configuration of `NewSamplerWithOptions` 
```
zapcore.NewSamplerWithOptions(
			core,
			10*time.Second,
			1,
			100,
		)
```
But it was still a high memory consumption.
So, it would be great to disable the logger sampler somehow from the `telemetry` config (in a long-term solution) or from the `exporterHelper` config (short-term solution)

Thank you.

dmitryax: I agree with @tigrannajaryan. This doesn't seem like a good option to expose in the exporters' configuration interface.

@antonjim-te can you please try initializing one sampled logger instance per service and see if that helps with memory utilization? It can be created in `service.New` function and passed as another field of `TelemetrySettings`  to all components. If it doesn't help with memory utilization, we can still expose a similar configuration for the sampled logger in the `service::telemetry`.

antonjim-te: Hi @dmitryax ,
Thank you for your suggestion. I will work on that in a separate branch, test it and if it solves the problem I will bring those changes to into this PR.

I will create a ticket in our internal process and add it to our next sprint.

I have a good feeling about this suggestion because it will just create one logger sampled for all the exporters instead of creating one per `exporterHelper`.

Thank you, team. I will keep you updated.

antonjim-te: Hi team, 
I have followed the suggestion from @dmitryax and created an extra sampled logger in the `TelemetrySettings` and used it when necessary in our case in the `QueuedRetrySender`

Please take a look at the new PR changes.

Thank you in advance.

component/telemetry.go
```diff
@@ -17,6 +17,10 @@ type TelemetrySettings struct {
 	// component to be used later as well.
 	Logger *zap.Logger
 
+	// SampledLogger passed to the created component.
+	// It will be used to avoid flooding the logs with messages that are repeated frequently.
+	SampledLogger *zap.Logger
```
mx-psi, line 22: `SampledLogger` should always be built from `Logger`. Instead of doing it like this, if we want to have this as part of the public API, I think it is better to have a method on `TelemetrySettings` that generates the `SampledLogger` on first use (e.g. using `sync.Once` to initialize a private field).

Alternatively, I would question that we need this on `TelemetrySettings` (do we have any other use cases for it?). Can we do it so that this is internal to exporterhelper? Without changing the public API, the way to do that would be a private global variable initialized on first use.

antonjim-te: Great feedback. I will create the sampledLogger the first time using it

antonjim-te:
Hi @mx-psi , @dmitryax , @bogdandrutu  , @tigrannajaryan could you please take a look at the PR with the lastest feedback applied.
Thank you.

service/telemetry/telemetry.go
```diff
@@ -59,7 +76,6 @@ func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
 	zapCfg := &zap.Config{
 		Level:             zap.NewAtomicLevelAt(cfg.Level),
 		Development:       cfg.Development,
-		Sampling:          toSamplingConfig(cfg.Sampling),
```
dmitryax, line 62: In this current form, this is a breaking change that drops existing configuration functionality. 

`LogsSamplingConfig` was applied to all the logs. Now, we are changing it to be applied to exporter logs only. There are other places where the logs can be pretty noisy when triggered by data handling errors. This config option was initially introduced to resolve that specific issue in the batch processor: https://github.com/open-telemetry/opentelemetry-collector/issues/4554. And now, we are dropping that functionality.

Until we pass the sampledLogger to all the possibly noisy outputs, we cannot apply this change. But I don't think it's necessary at all.

I suggest we just introduce a reasonable default to the existing sampling configuration and remove the additional hardcoded sampling in the exporterhelper. By default, we can probably sample 100 logs in a second after 10 of the initial logs. That should be enough not to miss important logs and ensure it's applied to the noisy ones only. This will simplify the code and will not introduce any additional configuration options or exposed Go API.

@open-telemetry/collector-approvers WDYT?

antonjim-te: Hi @dmitryax ,
I agree we should not break customer already using `Sampling` in the general logger. I kept this configuration.

Setting a default configuration for the extra sampled logger.

For future work, we should use this sampled logger for noisy outputs. 

Thank you.

jmacd (approved):

service/telemetry/telemetry.go
dmitryax: Discussed the suggested approach during the Collector SIG meeting. No objections to going ahead with it. We need to make sure it's documented with instructions on how to disable the sampling

service/telemetry/telemetry.go
```diff
@@ -16,6 +18,9 @@ import (
 type Telemetry struct {
 	logger         *zap.Logger
 	tracerProvider *sdktrace.TracerProvider
+
+	createSampledLogger sync.Once
+	sampledLogger       *zap.Logger
```
dmitryax, line 23: Based on the recent discussion, we don't need any additional sampling logger. We just need to introduce the default sampling configuration for the `logger`

antonjim-te: Hi @dmitryax ,
Thank you for reviewing this topic in the SIG meeting.

Default configuration
```
					Sampling: &telemetry.LogsSamplingConfig{
						Enabled:    true,
						Tick:       1 * time.Second,
						Initial:    10,
						Thereafter: 100,
					},
```

It can be disabled easily by setting `enabled` to `true`

Thanks

otelcol/unmarshaler.go
```diff
@@ -45,7 +47,9 @@ func unmarshal(v *confmap.Conf, factories Factories) (*configSettings, error) {
 					Development: false,
 					Encoding:    "console",
 					Sampling: &telemetry.LogsSamplingConfig{
-						Initial:    100,
+						Enabled:    true,
+						Tick:       1 * time.Second,
```
dmitryax, line 51: I think let's increase the tick and make it 10 seconds. It'll be more aligned with the current configuration on the exporter side. Otherwise, after this change, it will give us 100x more logs.

service/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"go.opentelemetry.io/collector/config/configtelemetry"
+)
+
+func TestTelemetryConfiguration(t *testing.T) {
```
dmitryax, line 18: Is this test related to the change?

service/telemetry/telemetry.go
```diff
@@ -78,16 +77,23 @@ func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
...
-	if sc == nil {
-		return nil
-	}
-	return &zap.SamplingConfig{
-		Initial:    sc.Initial,
-		Thereafter: sc.Thereafter,
-	}
+func newSampledLogger(logger *zap.Logger, sc *LogsSamplingConfig) *zap.Logger {
+	// Create a logger that samples all messages to sc.Tick per second initially,
+	// and sc.Initial/sc.Thereafter of messages after that.
```
dmitryax, line 89: The comment is a bit confusing. Should be something like this

```suggestion
	// Create a logger that samples every Nth message after the first M messages every S seconds 
	// where N = sc.Thereafter, M = sc.Initial, S = sc.Tick.
```

service/telemetry/config.go
```diff
@@ -91,8 +99,10 @@ type LogsConfig struct {
 // global CPU and I/O load that logging puts on your process while attempting
 // to preserve a representative subset of your logs.
 type LogsSamplingConfig struct {
-	Initial    int `mapstructure:"initial"`
-	Thereafter int `mapstructure:"thereafter"`
+	Enabled    bool          `mapstructure:"enabled"`
+	Tick       time.Duration `mapstructure:"tick"`
+	Initial    int           `mapstructure:"initial"`
+	Thereafter int           `mapstructure:"thereafter"`
```
dmitryax, line 105: Can you please add comment to the each field?

dmitryax (commented):
Thank you, @antonjim-te. It looks good to me. Just minor comment

service/telemetry/telemetry_test.go
antonjim-te: No, it was more in the side of testing the validation of the Telemetry Configuration

service/telemetry/config.go
```diff
@@ -91,7 +99,14 @@ type LogsConfig struct {
...
 // to preserve a representative subset of your logs.
 type LogsSamplingConfig struct {
-	Initial    int `mapstructure:"initial"`
+	// Enabled enable sampling logging
+	Enabled bool `mapstructure:"enabled"`
+	// Tick represents the interval in seconds that the logger apply each sampling.
+	Tick time.Duration `mapstructure:"tick"`
+	// Initial represents the first M messages logged each Tick.
+	Initial int `mapstructure:"initial"`
+	// Thereafter represents the N messages logged after Initial and the rest are dropped, during each Tick.
```
dmitryax, line 108:
```suggestion
	// Thereafter represents the sampling rate, every Nth message will be sampled after Initial messages are logged during each Tick.
```

.chloggen/SampledLoggerTelemetry.yaml
```diff
@@ -0,0 +1,26 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Enable sampling logging by default and apply it to all components."
+
+# One or more tracking issues or pull requests related to the change
+issues: [8134]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: The sampled logger configuration can be disabled easily by setting the 'enabled' field to 'false'.
```
dmitryax, line 18:
```suggestion
subtext: The sampled logger configuration can be disabled easily by setting the `service::telemetry::logs::sampling::enabled` to `false`.
```

dmitryax (approved):

dmitryax:
I'd like to have another look from any of @open-telemetry/collector-approvers before merging

otelcol/unmarshaler.go
```diff
@@ -45,7 +47,9 @@ func unmarshal(v *confmap.Conf, factories Factories) (*configSettings, error) {
 					Development: false,
 					Encoding:    "console",
 					Sampling: &telemetry.LogsSamplingConfig{
-						Initial:    100,
+						Enabled:    true,
+						Tick:       10 * time.Second,
+						Initial:    10,
```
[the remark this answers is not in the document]
antonjim-te: Agree 
Rollback the initial to 100 as it as initially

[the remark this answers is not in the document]
dmitryax: I don't think that's too low. Currently exporters are sampling from the 1 log. So changing is to 10 makes the exporters already 10x chatty than they used to be. I think making them 100x chatty is too much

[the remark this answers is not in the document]
antonjim-te: I see your point. 
@jpkrohling @dmitryax What should be the default? Should the `Initial` be 20?
```
					Sampling: &telemetry.LogsSamplingConfig{
						Enabled:    true,
						Tick:       10 * time.Second,
						Initial:    20,
						Thereafter: 100,
					},
```
Thanks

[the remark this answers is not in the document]
jpkrohling: I guess the docs are confusing to me: is it about logging the _same_ messages 10 times, or _any_ messages 10 times? If it's about skipping the 11th _repeated_ message within any 10s window, I agree that 10 is a good number.

mx-psi (approved):

Issue #7481, opened 2023-04-03
[otlpexporter] Add test to verify behavior on destination errors


tigrannajaryan:
We want to add tests otlpexporter to verify behavior when the destination responds with permanent and non-permanent errors.

Use the test helper after it is implemented here https://github.com/open-telemetry/opentelemetry-collector/issues/7479

Issue #7479, opened 2023-04-03
Add test helper to verify exporter behavior on errors


tigrannajaryan:
The consumers (and all exporters are consumers) are required to follow the [Error Handling contract](https://github.com/open-telemetry/opentelemetry-collector/blob/1c47d89fc38df12133adc78c4ea7bc4d3374e6b6/receiver/doc.go#L35) which says:

```
// If the error is non-Permanent then the nextConsumer.Consume*() call should be retried
// with the same data.
```

and 

```
// If the error is Permanent, then the nextConsumer.Consume*() call should not be
// retried with the same data.
```

We need a test helper that makes testing exporters for this part of the contract easy.

## Helper Inputs

The helper should accept:
- An exporter factory.
- A config for the exporter. We may want to make it easy to allocate a listening port for mock receivers and pass it to the exporter config as the endpoint to send to.
- A mock receiver. E.g. for `otlp` exporter testing we need `otlp` receiver that behaves correctly according to OTLP protocol spec and responds as instructed by the helper.

## Events to Test for

It is important that all 3 event types happen during the test at least once:
- ConsumeLogs/Traces/Metrics call succeeds at the first try.
- ConsumeLogs/Traces/Metrics call fails once with non-permanent error, then succeeds on the next try.
- ConsumeLogs/Traces/Metrics call fails more than once with non-permanent error, then eventually succeeds.
- ConsumeLogs/Traces/Metrics call fails once with permanent error.

The test helper should provide a data generator that can create data to be given to the exporter.

## Test Scenarios

The helper should be possible to configure for testing for permanent errors scenario and must verify:
- No data is lost if mock receiver always succeeds.
- No data is duplicated (in the presence or absence of mock receiver failures).
- Data is not retried by the exporter when the mock receiver indicates a permanent error to the exporter.
 
The helper should be possible to configure for testing for non-permanent errors scenario and must verify:
- No data is lost.
- No data is duplicated.
- Generated and delivered data content match exactly (but requests may be re-ordered).

## Other Requirements

- The test helper should generate concurrent calls to the exporter's ConsumeLogs/Traces/Metrics.
- The mock receiver should be able to accept and respond to concurrent requests from the exporter.
- The helper should support all 3 signal types: logs, traces, metrics.

omrozowicz-splunk:
Hey! I spent some time figuring out this task (mostly in comparision with the receiver test helper) and have a few questions:
[...]

PR #8143, opened 2023-07-28
 [chore] add test helper to verify exporter behavior on errors

omrozowicz-splunk:
**Description:** Adding a test helper to test exporter behaviour on errors.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/7479
Will also fix: https://github.com/open-telemetry/opentelemetry-collector/issues/7481

It is created based on the exporter receiver test: https://github.com/open-telemetry/opentelemetry-collector/pull/7516
The exporter targets `mockReceiver` as an endpoint.

The test cases covered here are:
1. ConsumeLogs/Traces/Metrics call succeeds at the first try
2. ConsumeLogs/Traces/Metrics call fails once with non-permanent error, then succeeds on one of the next tries
3. ConsumeLogs/Traces/Metrics call fails once with permanent error and the data is not retried
4. ConsumeLogs/Traces/Metrics call fails once with permanent error or permanent error (then the test pass requirement is calculated based on number of errors per kind/successfully delivered data/total number of data to be delivered)


exporter/exportertest/contract_checker.go
```diff
@@ -0,0 +1,298 @@
...
+type CheckConsumeContractParams struct {
+	T *testing.T
+	// Factory that allows to create an exporter.
+	Factory exporter.Factory
+	// DataType to test for.
+	DataType component.DataType
+	// Config of the exporter to use.
+	Config               component.Config
+	NumberOfTestElements int
+	MockReceiver         interface{}
```
dmitryax, line 39: This should be an interface providing the functionality needed for the contract checker only. `mockMetricsReceiver` and others should implement this interface in `exporter/otlpexporter` module. 

Probably better to do it thorough a factory
```suggestion
	MockReceiverFactory MockReceiverFactory
```

where 
```golang
type DecisionFunc func() error

type MockReceiverFactory func(DecisionFunc) MockReceiver

type MockReceiver interface {
     ... methods needed for contract checked
}

The main idea is that otlpexporter (or any other exporter) should depend on the contract checker, not the other way.

Comment (omrozowicz-splunk) [reply exporter/exportertest/contract_checker.go]: Hey, I changed it `MockReceiver MockReceiver` for now, I think I'm missing the point of why and how to use `type MockReceiverFactory func(DecisionFunc) MockReceiver`. I keep the decision functions in contract checker to mimic the same behaviour that was implemented in receiver tests: https://github.com/open-telemetry/opentelemetry-collector/pull/7516/files#diff-a246dccd935c825bf517eb70b6476fdeb05f4750d960247b7df7fbf9a17b23aeR85
I would appreciate some more tips on how to use `MockReceiverFactory` ☺️

Comment (dmitryax) [reply exporter/exportertest/contract_checker.go]: The main goal is to move the actual test using the helper to the tested modules, e.g. `TestConsumeContractOtlpLogs` should go to `exporter/otlpexporter` module. Keeping that in mind should influence the right decisions for the `MockReceiver` interface. 

You need to pass DecisionFunc to the exporter test which should give you an implementation of `MockReceiver` in return. `MockReceiverFactory` should be implemented on the exporter side. Please let me know if it makes it any clearer.

Comment (omrozowicz-splunk) [reply exporter/exportertest/contract_checker.go]: Ok, I think now it is more clear, I pushed another commit with MockReceiverFactory and tests moved under `exporter/otlpexproter`. It took some gymnastics with `net.Listen` (as now we create an instance every time depending on decision func). Please take a look and tell me if is it more or less what you meant.

Comment (dmitryax) [reply exporter/exportertest/contract_checker.go]: We are going in the right direction. Half way through :)

File: exporter/exportertest/contract_checker.go [hunk 2]
```diff
@@ -0,0 +1,269 @@
...
+
+	for i := 0; i < params.NumberOfTestElements; i++ {
+		id := UniqueIDAttrVal(strconv.Itoa(i))
+		fmt.Println("Preparing metric number: ", id)
+		data := CreateOneMetricWithID(id)
+
+		err = exp.ConsumeMetrics(ctx, data)
+	}
+
+	reqCounter := mockReceiver.getReqCounter()
```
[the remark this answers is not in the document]
omrozowicz-splunk: For now I set up retrySettings to 10 miliseconds with these settings:

```
// NewTestRetrySettings returns the default settings for otlp exporter test.
func NewTestRetrySettings() exporterhelper.RetrySettings {
	return exporterhelper.RetrySettings{
		Enabled: true,
		// interval is short for the test purposes
		InitialInterval:     10 * time.Millisecond,
		RandomizationFactor: backoff.DefaultRandomizationFactor,
		Multiplier:          1.1,
		MaxInterval:         10 * time.Second,
		MaxElapsedTime:      1 * time.Minute,
	}
}
```

but yes, I can add delay, we might have problems if we're very unlucky and getting error all the time 😅

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,355 @@
...
+var errPermanent = status.Error(codes.Internal, "Permanent error")
+
+type DecisionFunc func() error
+
+type MockReceiver interface {
+	clearCounters()
+	setExportErrorFunction(decisionFunction func() error)
+	srvStop()
+	getListenerAddress() string
+	getReqCounter() requestCounter
```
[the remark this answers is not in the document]
omrozowicz-splunk: ok, thanks for the tips! so I deleted Exports and put `ConsumeX` instead, and created a `mockConsumer`. I reused some of the code from the link. Additionally I keep there `ReceivedTraces/Metrics/Logs`, but not sure if I can make it useful when I already have that requestCounter - most of the logic is based on counting how many errors of a kind do we have. I can use the len of `ReceivedX`, but still I need to keep number of requests ent/failed somewhere.

connector/forwardconnector/go.mod
```diff
@@ -76,3 +76,19 @@ retract (
...
+
+replace go.opentelemetry.io/collector/config/internal => ../../config/internal
+
+replace go.opentelemetry.io/collector/extension/auth => ../../extension/auth
+
+replace go.opentelemetry.io/collector/exporter/otlpexporter => ../../exporter/otlpexporter
+
+replace go.opentelemetry.io/collector/config/configopaque => ../../config/configopaque
+
+replace go.opentelemetry.io/collector/config/configauth => ../../config/configauth
```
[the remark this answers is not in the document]
omrozowicz-splunk: Right, it disappeared 😄

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,355 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+package exportertest // import "go.opentelemetry.io/collector/exporter/exportertest"
+
+import (
+	"context"
+	"fmt"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter"
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,355 @@
...
+// // half of the time and fails with a non-permanent error the rest of the time.
+func randomNonPermanentErrorConsumeDecision() error {
+	if rand.Float32() < 0.5 {
+		return errNonPermanent
+	}
+	return nil
+}
+
+// Define a function that matches the MockReceiverFactory signature
+func CreateMockOtlpReceiver(decisionFunc DecisionFunc, dataType component.DataType, ln net.Listener) MockReceiver {
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

exporter/otlpexporter/consume_contract_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"go.opentelemetry.io/collector/config/configgrpc"
+	"go.opentelemetry.io/collector/config/configtls"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+	"go.opentelemetry.io/collector/exporter/exportertest"
+)
+
+const DefaultOTLPPort = 9999
+
+// NewTestRetrySettings returns the default settings for otlp exporter test.
+func NewTestRetrySettings() exporterhelper.RetrySettings {
```
dmitryax, line 25: There is no need to export this function

exporter/otlpexporter/consume_contract_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/cenkalti/backoff/v4"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config/configgrpc"
+	"go.opentelemetry.io/collector/config/configtls"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+	"go.opentelemetry.io/collector/exporter/exportertest"
+)
+
+const DefaultOTLPPort = 9999
```
dmitryax, line 22: This one should not be exported as well

exporter/otlpexporter/consume_contract_test.go
```diff
@@ -0,0 +1,103 @@
...
+		// interval is short for the test purposes
+		InitialInterval:     10 * time.Millisecond,
+		RandomizationFactor: backoff.DefaultRandomizationFactor,
+		Multiplier:          1.1,
+		MaxInterval:         10 * time.Second,
+		MaxElapsedTime:      1 * time.Minute,
+	}
+}
+
+func ReturnOtlpConfig(port int) component.Config {
```
dmitryax, line 37:
```suggestion
func testConfig(port int) component.Config {
```

exporter/otlpexporter/consume_contract_test.go
dmitryax: btw why 9999? We can extract the real default port from `defaultGRPCEndpoint`

exporter/otlpexporter/mock_receiver.go
```diff
@@ -0,0 +1,78 @@
...
+
+type BaseOTLPDataReceiver struct {
+	DataReceiverBase
+	mockConsumer    *exportertest.MockConsumer
+	traceReceiver   receiver.Traces
+	metricsReceiver receiver.Metrics
+	logReceiver     receiver.Logs
+}
+
+func NewOTLPDataReceiver(port int, mockConsumer *exportertest.MockConsumer) *BaseOTLPDataReceiver {
```
dmitryax, line 32: this should not be exported

exporter/otlpexporter/mock_receiver.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.opentelemetry.io/collector/receiver/receivertest"
+)
+
+// DataReceiverBase implement basic functions needed by all receivers.
+type DataReceiverBase struct {
+	// Port on which to listen.
+	Port int
+}
+
+type BaseOTLPDataReceiver struct {
```
dmitryax, line 24: No need to export publicly 
```suggestion
type mockReceiver struct {
```

exporter/otlpexporter/mock_receiver.go
```diff
@@ -0,0 +1,78 @@
...
+	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/config/confignet"
+	"go.opentelemetry.io/collector/exporter/exportertest"
+	"go.opentelemetry.io/collector/receiver"
+	"go.opentelemetry.io/collector/receiver/otlpreceiver"
+	"go.opentelemetry.io/collector/receiver/receivertest"
+)
+
+// DataReceiverBase implement basic functions needed by all receivers.
+type DataReceiverBase struct {
```
dmitryax, line 19: Why is this needed as a separate struct?

exporter/exportertest/contract_checker.go
```diff
@@ -0,0 +1,260 @@
...
+
+	reqCounter := mockReceiver.RequestCounter()
+	// The overall number of requests sent by exporter
+	fmt.Printf("Number of export tries: %d\n", reqCounter.total)
+	// Successfully delivered items
+	fmt.Printf("Total items received successfully: %d\n", reqCounter.success)
+	// Number of errors that happened
+	fmt.Printf("Number of permanent errors: %d\n", reqCounter.error.permanent)
+	fmt.Printf("Number of non-permanent errors: %d\n", reqCounter.error.nonpermanent)
+	checkIfTestPassed(params.T, params.NumberOfTestElements, reqCounter)
```
dmitryax, line 206: We cannot check this synchronously. The test should pass with the enabled exporter's queue as well. Please use `assert.Eventually` instead

exporter/otlpexporter/consume_contract_test.go
omrozowicz-splunk: done

exporter/otlpexporter/consume_contract_test.go
omrozowicz-splunk: I think it was a leftover form testing, now I left default grpc settings in the constructor, but to use `defaultGRPCEndpoint` in the config passed to exporter I think I'd need to make it exportable, right? I hardcoded it for now, I'm 99% sure it's wrong but don't know what would be clean way to do this.

omrozowicz-splunk: Or do you mean that I should use it like this?

exporter/otlpexporter/consume_contract_test.go
omrozowicz-splunk: done

exporter/otlpexporter/mock_receiver.go
omrozowicz-splunk: done

exporter/otlpexporter/mock_receiver.go
omrozowicz-splunk: Done, but `mockReceiver` already existed in this package in `otlp_test`, so I named it `mockOtlpReceiver`

exporter/otlpexporter/mock_receiver.go
omrozowicz-splunk: I used the structure from contrib's testbed, but it indeed does not help here, I deleted this part and port is set up by the default config of otlp receiver factory

exporter/exportertest/contract_checker.go
omrozowicz-splunk: Done with `assert.EventuallyWithT`, so I didn't need to rewrite functions to return bool

exporter/exportertest/contract_checker.go
```diff
@@ -0,0 +1,271 @@
...
+func CreateOneTraceWithID(id UniqueIDAttrVal) ptrace.Traces {
+	data := ptrace.NewTraces()
+	data.ResourceSpans().AppendEmpty().ScopeSpans().AppendEmpty().Spans().AppendEmpty().Attributes().PutStr(
+		UniqueIDAttrName,
+		string(id),
+	)
+	return data
+}
+
+func CreateOneMetricWithID(id UniqueIDAttrVal) pmetric.Metrics {
```
dmitryax, line 267: Should not be exposed. Limiting the API should the primary goal. Otherwise, it'll be hard to maintain and change after 1.0 release

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,237 @@
...
+	if r < third {
+		return consumererror.NewPermanent(errPermanent)
+	}
+	if r < 2*third {
+		return errNonPermanent
+	}
+	return nil
+}
+
+type BaseMockConsumer interface {
```
dmitryax, line 68: The interface is not being used. I think we should use it instead of exposing the struct `MockConsumer`. The interface will be `MockConsumer`, and struct will be unexported `mockConsumer`

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,237 @@
...
+type MockConsumer struct {
+	reqCounter          RequestCounter
+	mux                 sync.Mutex
+	exportErrorFunction func() error
+	ReceivedTraces      []ptrace.Traces
+	ReceivedMetrics     []pmetric.Metrics
+	ReceivedLogs        []plog.Logs
+}
+
+type MockConsumerFactory func() MockConsumer
```
dmitryax, line 84: This looks like not needed

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,237 @@
...
+
+type DecisionFunc func() error
+
+type MockReceiver interface {
+	Start() error
+	Stop() error
+	RequestCounter() RequestCounter
+}
+
+type MockReceiverFactory func(DecisionFunc) MockReceiver
```
dmitryax, line 33: I think we can avoid exposing `DecisionFunc` by changing this signature to 
```suggestion
type MockReceiverFactory func(MockConsumer) MockReceiver
```

CreateDefaultConsumer will not be needed after that

dmitryax (commented):
Also, please rebase

exporter/otlpexporter/consume_contract_test.go
dmitryax: Keeping it hardcoded is ok

exporter/otlpexporter/consume_contract_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func testConfig() component.Config {
+	return &Config{
+		TimeoutSettings: exporterhelper.TimeoutSettings{},
+		QueueSettings:   exporterhelper.QueueSettings{Enabled: false},
+		RetrySettings:   newTestRetrySettings(),
+		GRPCClientSettings: configgrpc.GRPCClientSettings{
+			Endpoint: confignet.NetAddr{Endpoint: "0.0.0.0:4317", Transport: "tcp"}.Endpoint,
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

exporter/exportertest/contract_checker.go
omrozowicz-splunk: changes, I think nothing left is uppercase unnecessarily

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: I expanded it with all the methods we need:

```
type MockConsumer interface {
	ConsumeLogs(_ context.Context, ld plog.Logs) error
	ConsumeMetrics(_ context.Context, md pmetric.Metrics) error
	ConsumeTraces(_ context.Context, td ptrace.Traces) error
	Capabilities() consumer.Capabilities
	Clear()
	RequestCounter() RequestCounter
}
```

and I use it in `mockOtlpReceiver` like this:

```
type mockOtlpReceiver struct {
	mockConsumer    exportertest.MockConsumer
	traceReceiver   receiver.Traces
	metricsReceiver receiver.Metrics
	logReceiver     receiver.Logs
}
```

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: removed

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: I changed it this way, but I think we still need a way to create MockConsumer to pass decision function based on test suite, so I created `NewMockConsumer`:

```
func NewMockConsumer(decisionFunc func() error) MockConsumer {
	return &mockConsumer{
		reqCounter:          newRequestCounter(),
		mux:                 sync.Mutex{},
		exportErrorFunction: decisionFunc,
		ReceivedTraces:      nil,
		ReceivedMetrics:     nil,
		ReceivedLogs:        nil,
	}
}
```

Didn't figure out how to create `MockConsumer` without an exported function. Let me know if I misunderstood 🙏

omrozowicz-splunk:
> Also, please rebase

done

exporter/otlpexporter/mock_receiver.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (bor *mockOtlpReceiver) Stop() error {
+	bor.mockConsumer.Clear()
+	err := bor.traceReceiver.Shutdown(context.Background())
+	err = errors.Join(err, bor.metricsReceiver.Shutdown(context.Background()))
+	return errors.Join(err, bor.logReceiver.Shutdown(context.Background()))
+}
+
+func (bor *mockOtlpReceiver) RequestCounter() exportertest.RequestCounter {
```
dmitryax, line 62: The MockConsumer has all the information about received/rejected data, right? So we don't need to require implementation of this method and don't need to expose RequestCounter in the `exportertest`

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	return nil
+}
+
+type MockConsumer interface {
+	ConsumeLogs(_ context.Context, ld plog.Logs) error
+	ConsumeMetrics(_ context.Context, md pmetric.Metrics) error
+	ConsumeTraces(_ context.Context, td ptrace.Traces) error
+	Capabilities() consumer.Capabilities
+	Clear()
```
dmitryax, line 71: Clear() is also not needed. It can be called in this package right after MockReceiver.Stop() if needed, instead of having it called by the API user from the Stop() method.

exporter/otlpexporter/mock_receiver.go
omrozowicz-splunk: To clarify - do you mean we don't need a request counter at all? This is the only part that contain information about nonpermanent and permanent errors. According to the implementation I have now we need it exported to check if the test finished successfully:

```
reqCounter := mockReceiver.RequestCounter()
// The overall number of requests sent by exporter
fmt.Printf("Number of export tries: %d\n", reqCounter.total)
// Successfully delivered items
fmt.Printf("Total items received successfully: %d\n", reqCounter.success)
// Number of errors that happened
fmt.Printf("Number of permanent errors: %d\n", reqCounter.error.permanent)
fmt.Printf("Number of non-permanent errors: %d\n", reqCounter.error.nonpermanent)

assert.EventuallyWithT(params.T, func(c *assert.CollectT) {
	checkIfTestPassed(params.T, params.NumberOfTestElements, reqCounter)
}, 2*time.Second, 100*time.Millisecond)
```

MockConsumer indeed has all the information, but it's not exposed at all in `exportertest/contract_checker.go`, that's why I use it through `MockReceiver`. Do you suggest moving this check logic somewhere else?

dmitryax: > MockConsumer indeed has all the information, but it's not exposed at all in exportertest/contract_checker.go, that's why I use it through MockReceiver.

It's available in the `reqCounter` and it's not required to be exposed because it's being used in the same package - the checker, not on the API consumer side. You can convert the `MockConsumer` interface into the `mockConsumer` struct to get access to all the private field. But that maybe too hacky. I think we can expose the struct instead of the interface to make it easier: so remove `MockReceiver` interface and expose the `mockConsumer` struct -> `MockConsumer`. Sorry, looks like I'm asking to undo what I asked before, but we need to get the the point where the API is clean :)

omrozowicz-splunk: Ok, so I deleted MockConsumer interface completely, exposed MockConsumer struct instead, but I didn't delete `MockReceiver` interface as I thought we'd reuse it for otlp http exporter. Now I use request counter like this: `reqCounter := mockConsumer.reqCounter` and the MockReceiver interface is:

```
type MockReceiver interface {
	Start() error
	Stop() error
}
```

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: Now I use `mockConsumer.clear()` directly after 
```
err = mockReceiver.Stop()
require.NoError(params.T, err)
```

receiver/otlpreceiver/go.mod
```diff
@@ -25,7 +25,6 @@ require (
 )
 
 require (
-	cloud.google.com/go/compute/metadata v0.2.4-0.20230617002413-005d2dfb6b68 // indirect
```
omrozowicz-splunk, line 28: Not sure what is wrong, but every time I run `make gotidy` I have this error:
```
go.opentelemetry.io/collector/exporter/otlpexporter imports
        go.opentelemetry.io/collector/config/configgrpc imports
        go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc tested by
        go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc.test imports
        google.golang.org/grpc/interop imports
        golang.org/x/oauth2/google imports
        cloud.google.com/go/compute/metadata: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
        cloud.google.com/go v0.26.0 (/Users/omrozowicz/go/pkg/mod/cloud.google.com/go@v0.26.0/compute/metadata)
        cloud.google.com/go/compute/metadata v0.2.4-0.20230617002413-005d2dfb6b68 (/Users/omrozowicz/go/pkg/mod/cloud.google.com/go/compute/metadata@v0.2.4-0.20230617002413-005d2dfb6b68)
make: *** [tidy] Error 1
```
For both otlp receiver and otlp exporter. When I bring back the changes from `main`, `make gotidy` deletes this automatically.

I inserted this line to make CI green but I am sure it's not a way to go:
```
replace cloud.google.com/go => cloud.google.com/go v0.110.2
```

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+func newRequestCounter() requestCounter {
+	return requestCounter{
+		success: 0,
+		error:   newErrorCounter(),
+		total:   0,
+	}
+}
```
dmitryax, line 109: Please move it after the `MockConsumer` methods

receiver/otlpreceiver/go.mod
dmitryax: `replace cloud.google.com/go => cloud.google.com/go v0.110.2` is good. We have it all around the codebase

dmitryax (commented):
I think we're almost done :) Thank you, @omrozowicz-splunk

exporter/otlpexporter/mock_receiver.go
dmitryax: Sounds great!

receiver/otlpreceiver/go.mod
```diff
@@ -130,6 +129,8 @@ replace go.opentelemetry.io/collector/extension/zpagesextension => ../../extensi
 
 replace go.opentelemetry.io/collector/consumer => ../../consumer
 
+replace cloud.google.com/go => cloud.google.com/go v0.110.2
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+type MockConsumer struct {
+	reqCounter          requestCounter
+	mux                 sync.Mutex
+	exportErrorFunction func() error
+	ReceivedTraces      []ptrace.Traces
+	ReceivedMetrics     []pmetric.Metrics
+	ReceivedLogs        []plog.Logs
```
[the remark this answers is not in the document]
omrozowicz-splunk: done

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: done

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,221 @@
...
+type MockConsumer struct {
+	reqCounter          requestCounter
+	mux                 sync.Mutex
+	exportErrorFunction func() error
+	receivedTraces      []ptrace.Traces
+	receivedMetrics     []pmetric.Metrics
+	receivedLogs        []plog.Logs
+}
+
+func NewMockConsumer(decisionFunc func() error) MockConsumer {
```
dmitryax, line 74: This not needed to be exported I believe.

exporter/exportertest/mock_receiver.go
```diff
@@ -0,0 +1,221 @@
...
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+var errNonPermanent = status.Error(codes.DeadlineExceeded, "non Permanent error")
+var errPermanent = status.Error(codes.Internal, "Permanent error")
+
+type MockReceiver interface {
+	Start() error
+	Stop() error
+}
```
dmitryax, line 28: We can avoid introducing this interface and use `component.Component` instead.

dmitryax (commented):
@omrozowicz-splunk couple last comments. Please rebase and I think we can merge it. Thank you!

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: done

exporter/exportertest/mock_receiver.go
omrozowicz-splunk: done

dmitryax (approved):
LGTM. Thanks @omrozowicz-splunk!

dmitryax:
@omrozowicz-splunk one last thing. Can you please add `contract_checker_test.go` with some basic test coverage. Looks like usage from another module isn't counted and it contract_checker_test.go the project's coverage below the threshold. Similar to `receivertest/contract_checker_test.go`

omrozowicz-splunk:
> @omrozowicz-splunk one last thing. Can you please add `contract_checker_test.go` with some basic test coverage. Looks like usage from another module isn't counted and it contract_checker_test.go the project's coverage below the threshold. Similar to `receivertest/contract_checker_test.go` which should showcase how to use the helper for other exporters

Okaay, I can try to do that, but that will prolong this task even more - it requires a fake receiver and fake exporter and making the fake exporter send data to that fake receiver. Not sure if that will be super helpful, probably looking at the otlp example will be easier to understand. 
But I'll take a look, maybe it is easier than I imagine now 😅

dmitryax:
Unfortunately, if we merge this without the tests, the collector's build will marked as failing because the project's coverage will be below the threshold

omrozowicz-splunk:
Hey @dmitryax, I added `contract_checker_test.go`, it's probably not the most elegant, but works and will increase the coverage. I needed to change `requestCounter` in `MockConsumer` struct to `*requestCounter` to be able to control it from exporter and then passing over to receiver. Please take a look 😄

dmitryax:
Thanks @omrozowicz-splunk! I added another commit to reduce the exposed API more and to simplify some tests. Let me know if something doesn’t look right for you.

Issue #7682, opened 2023-05-16
Component status reporting


djaglowski:
**Is your feature request related to a problem? Please describe.**
We should formalize the notion of component status, so that the state/health of each component in the collector may be understood independently.

**Describe the solution you'd like**
https://github.com/open-telemetry/opentelemetry-collector/pull/6560 is a good starting point.

Beyond establishing a system for publishing and subscribing to component status, we should enumerate the possible states and define a finite state diagram that governs when components transition from one state to another.

seankhliao:
I feel like reporting the resolved/hydrated config should also be part of the status it reports (for zpages / #5223)

mwear:
I can work on this.

PR #8169, opened 2023-08-02
Component Status Reporting

mwear:
**Description:** 
This PR introduces component status reporting. There have been several attempts to introduce this functionality previously, with the most recent being: #6560.

This PR was orignally based off of #6560, but has evolved based on the feedback received and some additional enhancements to improve the ease of use of the `ReportComponentStatus` API.

In earlier discussions (see https://github.com/open-telemetry/opentelemetry-collector/pull/8169#issuecomment-1668367246) we decided to model status as a finite state machine with the following statuses: `Starting`, `OK`, `RecoverableError`, `PermanentError`, `FatalError`. `Stopping`, and `Stopped`. A benefit of this design is that `StatusWatcher`s will be notified on changes in status rather than on potentially repetitive reports of the same status.

With the additional statuses and modeling them using a finite state machine, there are more statuses to report. Rather than having each component be responsible for reporting all of the statuses, I automated status reporting where possible. A component's status will automatically be set to `Starting` at startup. If the components `Start` returns an error, the status will automatically be set to `PermanentError`. A component is expected to report `StatusOK` when it has successfully started (if it has successfully started) and from there can report changes in status as it runs. It will likely be a common scenario for components to transition between `StatusOK` and `StatusRecoverableError` during their lifetime. In extenuating circumstances they can transition into terminal states of `PermanentError` and `FatalError` (where a fatal error initiates collector shutdown). Additionally, during component Shutdown statuses are automatically reported where possible. A component's status is set to `Stopping` when Shutdown is initially called, if Shutdown returns an error, the status will be set to `PermanentError` if it does not return an error, the status is set to `Stopped`. 

In #6560 ReportComponentStatus was implemented on the `Host` interface. I found that few components use the Host interface, and none of them save a handle to it (to be used outside of the `start` method). I found that many components keep a handle to the `TelemetrySettings` that they are initialized with, and this seemed like a more natural, convenient place for the `ReportComponentStatus` API. I'm ultimately flexible on where this method resides, but feel that `TelemetrySettings` a more user friendly place for it.

Regardless of where the `ReportComponentStatus` method resides (Host or TelemetrySettings), there is a difference in the method signature for the API based on whether it is used from the service or from a component. As the service is not bound to a specific component, it needs to take the `instanceID` of a component as a parameter, whereas the component version of the method already knows the `instanceID`. In #6560 this led to having both `component.Host` and `servicehost.Host` versions of the Host interface to be used at the component or service levels. In this version, we have the same for TelemetrySettings. There is a `component.TelemetrySettings` and a `servicetelemetry.Settings` with the only difference being the method signature of `ReportComponentStatus`.

Lastly, this PR sets up the machinery for report component status, and allows extensions to be `StatusWatcher`s, but it does not introduce any `StatusWatcher`s. We expect the OpAMP extension to be a `StatusWatcher` and use data from this system as part of its AgentHealth message (the message is currently being extended to accommodate more component level details). We also expect there to be a non-OpAMP `StatusWatcher` implementation, likely via the HealthCheck extension (or something similiar).

**Testing:** Unit tests

cc: @tigrannajaryan @djaglowski @evan-bradley


component/component.go
```diff
@@ -175,3 +175,10 @@ type CreateDefaultConfigFunc func() Config
 func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
 	return f()
 }
+
+// GlobalID uniquely identifies a component
+type GlobalID struct {
+	ID         ID
+	Kind       Kind
+	PipelineID ID // Not empty only if the Kind is Processor
+}
```
djaglowski, line 184: There are a couple nuances to the way we instance components within the service, which I think likely need to be reconciled in this PR.

Receivers and exporters are instanced _per data type_. For example, users will often configure a single `otlpexporter` for exporting multiple types of data, but the underlying implementation creates separate instances. which would share the same GlobalID.

This is probably the correct model to present externally, because it aligns with the user's understanding, but we will have to define how to reconcile with the underlying implementation. Minimally, instances of the "same" component must be able to safely publish updates concurrently.

This raises a usability complication though, because we may expect status thrashing or burying of problems. e.g. the traces instance of a component reports a problem, but the metrics instance immediately clobbers the status by reporting that it is healthy.

I think we probably need to maintain a separate statuses for each instance of the "same" component, and aggregate them whenever an event is published. e.g. report "lowest" health state of any instance that shares the GlobalID, regardless of which instance sent the most recent event.

To further complicate things, we have the `sharedcomponent` package, which allows component authors to opt in to a unified model. In this case, the service still reasons about separate instances, but there is in fact only one instance after all. For example, the `otlpreceiver` uses this mechanism so that the shared instance can present a single port. I'm not certain how best to handle this, but perhaps the `sharedcomponent` package needs to manage status reporting for the components that use it.

Connectors share the same concerns as receivers and exporters, except instancing is _per ordered pair of data types_. `sharedcomponent` may or may not be used here as well.

mwear: 🤯 this sounds somewhat complicated, but thanks for the explanation.

I'm wondering if we should assume that each instance of a component should report status individually, and if so, perhaps we can rename `GlobalID` to `InstanceID`, where the "ID" would have enough metadata to identify the underlying component. We could add some additional fields to the struct if needed.

I'm also wondering if this is accidentally already happening with the `GlobalID` as it's currently implemented. We are passing around pointers to GlobalID instances, via the host wrapper. Is the code that generates the GlobalIDs in graph generating a global ID per component instance? If so, we would have a GlobalID instance per component, each with a distinct pointer, but possibly the same struct values.

I'd also be interested in any other reasonable ways to approach this.

djaglowski: > I'm also wondering if this is accidentally already happening with the GlobalID as it's currently implemented. We are passing around pointers to GlobalID instances, via the host wrapper. Is the code that generates the GlobalIDs in graph generating a GlobalID per component instance? If so, we would have a GlobalID instance per component, each with a distinct pointer, but possibly the same struct values.

To clarify, are you talking about [`graph.nodeID`](https://github.com/open-telemetry/opentelemetry-collector/blob/8aeb3ab04e695b626ae642c8c06f1d909fc33624/service/internal/graph/nodes.go#L37-L47) as the current implementation of `GlobalID`, and `service.Host`'s `graph.Graph` as the mechanism by which we're passing around pointers to the IDs?

Assuming so, I don't believe there are uniqueness problems there because the `nodeID`'s are hashes that are generated based on the the considerations I mentioned above. [Processor IDs](https://github.com/open-telemetry/opentelemetry-collector/blob/8aeb3ab04e695b626ae642c8c06f1d909fc33624/service/internal/graph/nodes.go#L120) are hashed with a specific pipeline ID. [Receiver IDs](https://github.com/open-telemetry/opentelemetry-collector/blob/8aeb3ab04e695b626ae642c8c06f1d909fc33624/service/internal/graph/nodes.go#L64) and [Exporter IDs](https://github.com/open-telemetry/opentelemetry-collector/blob/8aeb3ab04e695b626ae642c8c06f1d909fc33624/service/internal/graph/nodes.go#L168) are hashed with a data type, and [Connector IDs](https://github.com/open-telemetry/opentelemetry-collector/blob/8aeb3ab04e695b626ae642c8c06f1d909fc33624/service/internal/graph/nodes.go#L218) are hashed with an ordered pair of data types. `nodeID` is also not accessible outside of the graph package, and I don't think there's any way to indirectly reference or modify them.

> I'm wondering if we should assume that each instance of a component should report status individually, and if so, perhaps we can rename GlobalID to InstanceID, where the ID would have enough metadata to identify the underlying component. We could add some additional fields to the struct if needed.

I think it would be very hard for consumers of the events to make sense of this. The instancing is really an internal concern of the collector and not how we expect users to reason about the components.

> I'd also be interested in any other reasonable ways to approach this.

The only way I can think to do this is by modeling _component_ and _instance_ separately:
1. A _component_ model is presented to users, where each component is 1:1 with those defined in the configuration, except for processors which are understood to be unique to each pipeline.
2. An _instance_ model underlies the component model but is not directly exposed to users.
3. Every instance belongs to a component. i.e. A component may have one or more instances.
4. Internally, each instance reports its own status.
5. Externally, each component reports an aggregation of the statuses of its instances. (Often, a component will have only 1 instance, so its status is that of the instance)

Based on this, I think the host's `ReportComponentStatus` should effectively be `ReportInstanceStatus` and the reported status should be logged into a data structure which:
1. Maintains status for all instances
2. Understands the relationship between components and instances
3. Roughly, `map[GlobalComponentID]map[GlobalInstanceID]Status`

After updating the instance status, the `ReportInstanceStatus` function should aggregate the statuses for the relevant component and emit an event to StatusWatchers.

There's some plumbing to figure out, but what do you think about this model conceptually?

mwear: > To clarify, are you talking about [graph.nodeID](https://github.com/open-telemetry/opentelemetry-collector/blob/8aeb3ab04e695b626ae642c8c06f1d909fc33624/service/internal/graph/nodes.go#L37-L47) as the current implementation of GlobalID, and service.Host's graph.Graph as the mechanism by which we're passing around pointers to the IDs?

I made some changes to graph in this PR. I ended up making a `map[graph.nodeID]*component.GlobalID` and due to idiosyncrasies of implementation, the component.GlobalID pointer is effectively an instance id.  That's not super important though.

> There's some plumbing to figure out, but what do you think about this model conceptually?

I'm ultimately open to any and all options, but there is one thing I wanted to bring up that likely complicates the ability to aggregate status at the component level. Right now we only have `StatusOK` and `StatusError`, but I would like a more expressive set of statuses for components to choose from. I was looking at [GRPC status](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) as a possibility. I think there is a status for many, if not all issues a component could encounter. The 16 GRPC status codes are well-defined and carry a lot of information about the health of a component as a simple `code` and `message`. I'm not sure how you would aggregate multiple statuses from a system like GRPC into a single code though.

I wanted to do a little more homework on this idea before bringing it up, but now that I've mentioned it, what are your thoughts about adopting GRPC status for component status reporting? Would this change your thoughts about reporting status per instance vs status per component?

mwear: If we did not want to go with GRPC status, we could come up with our own. At a minimum, I'd like a component to be able to report:

| Status | Description |
|-------|-------------|
| StatusOK | The component is operating as expected |
| StatusError | The component has encountered an "unexpected" error, look at the error for details |
| StatusStartupError | The component failed to start |
| StatusRecoverableError | The component is experiencing a likely transient error |
| StatusPermanentError | The component has encountered an error from which it will not recover; the component is giving up and will stop itself |
| StatusFatalError | The error from this component should trigger a shutdown of the entire collector |

djaglowski:
> I was looking at [GRPC status](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) as a possibility. I think there is a status for many, if not all issues a component could encounter. The 17 GRPC status codes are well-defined and carry a lot of information about the health of a component as a simple code and message. I'm not sure how you would aggregate multiple statuses from a system like GRPC into a single code though.

Setting aside the question of aggregation, I think there is an important distinction between what we are trying to represent and what GRPC is representing with status codes. A GRPC status code describes the response to a specific request. On the other hand, a component is a persistent thing which can be described as being in various states over time, and where the possible future states depend not only on current events but also on the previous state. For example, if we have a "Permanent Error" state, this should be a terminal state where no future event is going to move it back to "OK". Similarly, if we have a "Stopping" state, we would enforce constraints on future possible states.

I think a finite state machine is probably an appropriate model here. The vanilla path might be `Starting -> OK -> Stopping -> Stopped`. We can have various error states, only some of which are be recoverable, etc. 

GRPC codes _might_ be a reasonable set of states to represent, but the question to me is whether we can clearly define the relationships between the states. My intuition is that the full set of GRPC codes may be too verbose.

---

> Right now we only have StatusOK and StatusError, but I would like a more expressive set of statuses for components to choose from.

I made a rough attempt to diagram the simpler set of states that you mentioned and this seems to be a bit more manageable.

image

---

W/r/t aggregating multiple instances of a component into a single component status, perhaps the solution lies in having a well defined state diagram for _component_ status, but which takes into account instance-level information as needed to determine whether and/or when to transition to another status.

For example, let's say an `otlpexporter` is configured for use in both a traces and metrics pipeline, and that everything has started up successfully so we are in an "OK" state. Then the following occurs:
1. The metrics instance reports a recoverable error
2. The traces instance reports a recoverable error
3. The traces instance reports that it has recovered
4. The metrics instance reports that it has recovered

The aggregate behavior would be that the component enters the "Recoverable Error" state at step 1, and returns to the "OK" state at step 4. Optionally, we might want steps 2 and 3 to result in events to the status watcher, but these events are not changing the status, only the error message. E.g.
1. Error: "metrics..."
2. Error: "metrics ..., traces ..."
3. Error: "metrics..."
4. OK

The implementation of this doesn't _necessarily_ have to revolve around instances. For example, every recoverable error could be associated with a callback or ID which can be used to clear the error. Then we only need to track that all errors have been cleared.

Anyways, does this seems like a design worth exploring further? I'm also open to anything that works but it seems like we need some clarity on how instances reporting status-related events will contribute to overall component status.

mwear:
I was considering using GRPC status because many of the components are clients or servers of network requests, and they could return the result of their last request as their status. However, I share some of your concerns about GPRC status possibly being too verbose and not a good fit for modeling stateful components. I think modeling status as a finite state machine makes sense, and probably has some benefits. As an FSM we would only have to report changes in status, rather than reporting the same status (possibly repeatedly). The other big benefit is that we could constrain the set of states to things that will know will be useful for monitoring collector health. A smaller set of states, so long as they effectively communicate component status, will be easier for component authors to use properly, and easier for StatusWatchers to consume. I think this is a design that is worth exploring further.

jaronoff97:
I'm a HUGE fan of this state machine diagram. I think there are really two FSMs here as mentioned: a component and the collector. i.e. would a permanent failure in a processor mean a permanent failure in the collector? not necessarily. I.e. if we permanently fail to apply a transformation would that stop the flow of telemetry? I would definitely like to see further design on this (probably in a separate issue) I think for this PR it's fine to use the system we have today with the option of expanding this in the future.

mwear:
> would a permanent failure in a processor mean a permanent failure in the collector?

A permanent failure in a component would affect the pipeline it's part of, but the collector would continue to run in a degraded mode. A fatal error would be one that would trigger a collector shutdown. I agree we need to hash out the various failure modes, make sure we have statuses for them and that we can surface this information as best as possible.

djaglowski:
I iterated on the initial state machine and came up with the following, which has just a couple changes:
- Clarifies that both "Permanent" and "Fatal" are directly reachable from any state besides "Stopped"
- Allows transition from "Recoverable" directly to "Stopping". The idea being that we still have control so should be able to issue a stop command before it recovers.

image

---

Then I tried to figure out how this would apply to components vs instances. The only way I could make this work was to say that the above diagram defines the lifecycle of an _instance_. Then I worked out some aggregation logic.

To work through this, I imagined two instances moving through the state machine independently, and then represented this as a compound state machine, where each state represents a pair of states (one for each instance) and the transitions between states each represent a change to one of the instances.

I found that the aggregate states seem somewhat intuitive, based on a some rules (see below). Here's the compound machine:

image

---

The "rules" that I think make sense, in order of priority are as follows:
1. If any instance encounters a fatal error, the component is in a Fatal Error state (if this even matters, since the collector will be killed).
2. If any instance is in a Permanent Error state, the component status is Permanent Error.
3. If any instance is Stopping, the component is in a Stopping state. That is, there is a clear intention to stop all instances and so the component is Stopping until it is fully Stopped, or encounters a Fatal or Permanent Error.
4. If any instance is Stopped, but no instances are Stopping, we must be in the process of stopping the component. This is an edge case where one instance has complied with a stop command but the stop command hasn't been issued to another instance yet.
5. If all instances are Stopped, the component is Stopped.
6. If any instance is in a Recoverable Error state, the component status is Recoverable Error.
7. If any instance is Starting, the component status is Starting.
8. None of the above were true, so the component is OK. (In other words, all instances are OK.)

In code:

```go
func (c Component) Status {

  for _, i := c.instances {
    if i.status == Fatal {
      return Fatal
    }
  }

  for _, i := c.instances {
    if i.status == Permanent {
      return Permanent
    }
  }

  for _, i := c.instances {
    if i.status == Stopping {
      return Stopping
    }
  }

  var stopped, nonstopped
  for _, i := c.instances {
    if i.status == Stopped {
      stopped = true
    } else {
      nonstopped = true
    }
  }
  if stopped && nonstopped {
    return Stopping
  } else if stopped {
    return Stopped
  }

  for _, i := c.instances {
    if i.status == Recoverable {
      return Recoverable
    }
  }

  for _, i := c.instances {
    if i.status == Starting {
      return Starting
    }
  }

  return OK
}
```

mwear:
Thanks for working through the gnarly details of how to aggregate status at a component level. The rules you came up with make sense. 

One piece of feedback I received from @mhausenblas at the Agent Management Working Group, is that while users are interested in component status, they are very interested in knowing the health of each pipeline. I'd assume users would want to look at pipeline health first, and then look at component statuses to help diagnose pipeline issues. Is the reason that some components are represented by multiple instances is that they belong to multiple pipelines? If that is the case, would be easier to tie component health to pipeline health if we report status on a instance basis? 

If we wanted to link component health to pipeline health, I'd assume that we'd include the pipeline (or possibly pipelines) a component belongs as part of the status event (likely part of the `InstanceID` of the component). A plausible way for a status watcher to determine pipeline status would be to group by pipeline and follow rules similar to what you came up with for aggregating component status. I'm wondering this scenario is any more difficult with we aggregate status at a component level, before it is sent to a status watcher? I fear that we may lose the ability to map components to pipelines in this situation, but I'm still trying to understand how components, component instances, and pipelines are related under the hood.

djaglowski:
> Is the reason that some components are represented by multiple instances is that they belong to multiple pipelines?

Not exactly but that's almost correct. Each instance of a component corresponds to an independent behavior of the component (e.g. exporting logs vs exporting traces)

> If we wanted to link component health to pipeline health, I'd assume that we'd include the pipeline (or possibly pipelines) a component belongs as part of the status event (likely part of the InstanceID of the component). A plausible way for a status watcher to determine pipeline status would be to group by pipeline

This is a great idea. I think that _all_ instances can be uniquely identified this way, and that it gives a clear way to roll up a status for a pipeline. 

```go
type InstanceID struct {
	ID         ID
	Kind       Kind
	PipelineIDs []ID // order does not matter, so alternately could use map[ID]struct{}
}
```

mwear:
Great. I think my next steps will be to replace `GlobalID` with `InstanceID` and add the additional statuses. I think the goal is to have enough information on the status events so that status watchers can make use of them.

OpAMP will be a consumer of this data and we briefly discussed a rough plan of how this could work going forward. We'll have to modify the `AgentHealth` message to accommodate a map of values so that it can hold health information about multiple entities. The OpAMP extension can implement the StatusWatcher interface and surface health information out of the collector. We can probably apply similar ideas to the health check extension for the non-OpAMP use case.

tigrannajaryan:
> OpAMP will be a consumer of this data and we briefly discussed a rough plan of how this could work going forward. We'll have to modify the `AgentHealth` message to accommodate a map of values so that it can hold health information about multiple entities. The OpAMP extension can implement the StatusWatcher interface and surface health information out of the collector. We can probably apply similar ideas to the health check extension for the non-OpAMP use case.

@mwear Please open an issue in https://github.com/open-telemetry/opamp-spec/issues to discuss and work on this. Depending on what the changes are this may need to happen before OpAMP 1.0 (unless change are additive).

mwear:
@tigrannajaryan, I created https://github.com/open-telemetry/opamp-spec/issues/165 to discuss.

component/component.go
```diff
@@ -175,3 +175,10 @@ type CreateDefaultConfigFunc func() Config
 func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
 	return f()
 }
+
+// InstanceID uniquely identifies a component instance
+type InstanceID struct {
+	ID          ID
+	Kind        Kind
+	PipelineIDs map[ID]struct{}
```
codeboten, line 183: Does an `InstanceID` need a map of its pipelines?

component/status.go
```diff
@@ -0,0 +1,141 @@
...
+// StatusEventOption applies options to a StatusEvent.
+type StatusEventOption func(*StatusEvent) error
+
+// ErrStatusEventInvalidArgument indicates an invalid option was specified when creating a status
+// event. This will happen when using WithError for a non-error status.
+var ErrStatusEventInvalidArgument = errors.New("status event argument error")
+
+// WithError sets the error object of the StatusEvent. It is optional
+// and should only be applied to an Event of type ComponentError.
+func WithError(err error) StatusEventOption {
```
djaglowski, line 85: How does this correspond to the different types of error statuses?

component/host.go
```diff
@@ -12,6 +12,7 @@ type Host interface {
 	//
 	// ReportFatalError should be called by the component anytime after Component.Start() ends and
 	// before Component.Shutdown() begins.
+	// Deprecated: [x.x.x] Use ReportComponentStatus instead (with an event component.StatusFatalError)
```
codeboten, line 15: Worth being more specific on where `ReportComponentStatus` can be found since it's not on the Host

component/status.go
djaglowski: Should we just have `NewRecoverableErrorEvent`, `NewPermanentErrorEvent`, `NewFatalErrorEvent`, each which require an error parameter?

service/host.go
```diff
@@ -66,3 +66,11 @@ func (host *serviceHost) GetExtensions() map[component.ID]component.Component {
 func (host *serviceHost) GetExporters() map[component.DataType]map[component.ID]component.Component {
 	return host.pipelines.GetExporters()
 }
+
+func (host *serviceHost) notifyComponentStatusChange(source *component.InstanceID, event *component.StatusEvent) {
+	// TODO: What should we do if there is an error returned by a StatusWatcher?
```
djaglowski, line 71: Good question. This could cause an explosion of errors if they are reported back to the status watcher. 

Is there anything useful we can do with an error from the watcher? Maybe we just log it in the watcher and never return an error?

service/internal/servicetelemetry/settings.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/service/internal/status"
+)
+
+// Settings mirrors component.TelemetrySettings except for the method signature of
+// ReportComponentStatus. The service level Settings is not bound a specific component, and
+// therefore takes a component.InstanceID as an argument.
+type Settings component.TelemetrySettingsBase[status.ServiceStatusFunc]
```
codeboten, line 14: I don't know if it would be possible at all, but it would be great not to have both servicetelemetry.Settings and component.TelemetrySettings. If both are necessary, I would lean towards keeping the name consistent at least, so that the module could indicate the difference. i.e:

```
service.TelemetrySettings
component.TelemetrySettings
```

or

```
service.Settings
component.Settings
```

Though the term Settings may be the reason why this is hard to grok in the first place. Open to suggestions here.

codeboten (commented):
Thanks for taking this on @mwear! The PR is quite large so it will take some time to get through the entire review. I've added some initial comments

extension/extensiontest/statuswatcher_extension.go
```diff
@@ -0,0 +1,46 @@
...
+		func() component.Config {
+			return &struct{}{}
+		},
+		func(context.Context, extension.CreateSettings, component.Config) (component.Component, error) {
+			return &statusWatcherExtension{onStatusChanged: onStatusChanged}, nil
+		},
+		component.StabilityLevelStable)
+}
+
+// statusWatcherExtension stores consumed traces and metrics for testing purposes.
```
codeboten, line 37: This comment seems incorrect

extension/extensiontest/statuswatcher_extension.go
```diff
@@ -0,0 +1,46 @@
...
+
+// NewStatusWatcherExtensionCreateSettings returns a new nop settings for Create*Extension functions.
+func NewStatusWatcherExtensionCreateSettings() extension.CreateSettings {
+	return extension.CreateSettings{
+		TelemetrySettings: componenttest.NewNopTelemetrySettings(),
+		BuildInfo:         component.NewDefaultBuildInfo(),
+	}
+}
+
+// NewStatusWatcherExtensionFactory returns a component.ExtensionFactory that constructs nop extensions.
```
codeboten, line 22: this isn't really a nop extension correct?

service/internal/status/status.go
```diff
@@ -0,0 +1,167 @@
...
+			component.StatusNone: {
+				component.StatusStarting: {},
+			},
+			component.StatusStarting: {
+				component.StatusOK:               {},
+				component.StatusRecoverableError: {},
+				component.StatusPermanentError:   {},
+				component.StatusFatalError:       {},
+				component.StatusStopping:         {},
+				component.StatusStopped:          {},
```
djaglowski, line 68: I tend to think we shouldn't be able to go directly to Stopped, without going through Stopping.

service/internal/status/status.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+			component.StatusOK: {
+				component.StatusRecoverableError: {},
+				component.StatusPermanentError:   {},
+				component.StatusFatalError:       {},
+				component.StatusStopping:         {},
+				component.StatusStopped:          {},
```
djaglowski, line 75: Same here

service/internal/status/status.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+			component.StatusRecoverableError: {
+				component.StatusOK:             {},
+				component.StatusPermanentError: {},
+				component.StatusFatalError:     {},
+				component.StatusStopping:       {},
+				component.StatusStopped:        {},
```
djaglowski, line 82: And here

djaglowski (commented):
Generally this looks good to me. Just a few minor notes.

component/component.go
djaglowski: I think so. This is what differentiates instances of a component. For example:

```yaml
receivers:
  foo: ...

exporters:
  bar: ...

pipelines:
  logs:
    receivers: foo
    exporters: bar
  logs/2:
    receivers: foo
    exporters: bar
  traces:
    receivers: foo
    exporters: bar
```

The configuration defines two components, but the collector will instantiate four instances:

```go
{
	ID: "foo",
	Kind: "receiver",
	PipelineIDs: { "logs": {}, "logs/2": {} }
},
{
	ID: "foo",
	Kind: "receiver",
	PipelineIDs: { "traces": {}, }
},
{
	ID: "bar",
	Kind: "exporter",
	PipelineIDs: { "logs": {}, "logs/2": {} }
},
{
	ID: "bar",
	Kind: "exporter",
	PipelineIDs: { "traces": {}, }
},
```

This differentiation isn't utilized in this PR but it will be necessary in order to produce an accurate & aggregated health for a component or pipeline. @mwear, do you think this logic should be part of this PR too?

service/internal/status/status.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+			return fsm.transition(status, opts...)
+		}
+
+}
+
+// NewComponentStatusFunc returns a function to be used as ReportComponentStatus for
+// component.TelemetrySettings, which differs from servicetelemetry.Settings in that
+// the component version is tied to specific component instance.
+func NewComponentStatusFunc(id *component.InstanceID, srvStatus ServiceStatusFunc) component.StatusFunc {
```
djaglowski, line 163: One complication I want to call out again is the [`sharedcomponent`](https://github.com/open-telemetry/opentelemetry-collector/blob/2ebe1e501f9a206237de1009a127f1a419ab3696/internal/sharedcomponent/sharedcomponent.go) package. This can be used to combine instances together.

Take for instance the `otlp` receiver which uses this mechanism. The collector will believe that it is creating multiple instances, but the receiver implementation will use `sharedcomoponent` to check if any instance has already been created. If so, it will return that to the collector. The problem here is that whichever instance is created first will be the only one that ever reports status events.

I think it should be possible to fix this within the package. Basically, a shared component could hang on to the `StatusFunc` from each of the instances that it represents. Then whenever it reports an event, it should report it through all of the functions.

service/internal/servicetelemetry/settings.go
mwear: `service.TelemetrySettings` would be possible, with the caveat that it will expand the public API surface of `service`. That was my main reason for putting it in the internal `servicetelemetry`. I'm for moving to service, it does seem like overkill to introduce an internal package to restrict visibility of the telemetry settings.

component/status.go
mwear: The comment is incorrect. WithError should be usable with `StatusRecoverableError`, `StatusPermanentError` and `StatusFatalError`. I prefer to keep the functional option if at all possible since due to the signature of report component status (e.g. `ReportComponentStatus(component.Status, ...component.StatusEventOption)`). Is it fine if I just fix the comment 😅?

service/internal/servicetelemetry/settings.go
mwear: Actually, I don't think we can expose this from the `service` package. `TelemetrySettings` is needed the in the `extensions` package as well as the `service` package and exposing it from the `service` package creates an import cycle.

component/status.go
djaglowski: > due to the signature of report component status

I think the signature may be problematic. It just seems a bit too easy for the caller to use it incorrectly, e.g. report an error status without an error, or report a non-error status with an error. Incorrect usage may not be immediately apparent either because the cases where we report errors may happen very rarely. I'm worried that the effect of misusing it may not be noticed until we're in the exact situation where we really need it to work correctly.

It seems like there must be a better signature that can't be misused. Maybe it should just accept a `StatusEvent`, and we'll make it as easy as possible to construct events appropriately?

service/host.go
mwear: I removed the error return value. It turns out it was never being set anyways. A StatusWatcher should handle its own errors internally.

service/internal/status/status.go
mwear: It was an oversight, and cascaded via copypasta. Thanks for catching this.

component/status.go
mwear: I removed the functional options, replaced them with per event type constructors, and updated `ReportComponentStatus` to take a `StatusEvent`.

service/internal/servicetelemetry/settings.go
mwear: As a half step, I renamed the struct from `Settings` to `TelemetrySettings`. If we have a better name for the package we can rename it also, but for now, it's `servicetelemetry.TelemetrySettings`. This package is basically a place for things that need to be shared between the `service` and `extensions` packages without creating an import cycle. The first and only thing to share is `TelemetrySettings` and the reason for it's current name. We could try to come up with something more general if we'd like.

component/component.go
mwear: I was leaving this for our first StatusWatchers, but I think it does make sense to have a shared method to compute an aggregated status. Would a function, `component.AggregateStatus([]*StatusEvent) *StatusEvent` make sense? It would return the highest priority event from the slice (based on the status value). If there is a tie, it would use the earliest timestamp? Is this something along the lines of what you had in mind?

djaglowski: I think an aggregation function definitely makes sense.

> It would return the highest priority event from the slice (based on the status value).

Based on the analysis I did earlier, I think this logic would fail to account for some notable cases.

For example, say we're in the process of Stopping a pipeline. We could in theory have four instance with the following statuses: `[Stopped, OK, Recoverable, Starting]`. The first instance has presumably been issued a Stop command and immediately complied. The others either haven't yet been issued the command, haven't acked it, or for whatever reason their statuses haven't been updated yet. I think the correct aggregated status here is `Stopping`, even though none of the instances show that status. We have to infer that because one instance is Stopped, we must be in the process of stopping the component or pipeline as a whole.

I had proposed specific logic for aggregation [here](https://github.com/open-telemetry/opentelemetry-collector/pull/8169#issuecomment-1670048722). However, if you disagree and think we should stick with a simple highest priority, then I think the following is the closest order:
```
	StatusNone Status = iota
	StatusOK
	StatusStarting // Aggregate is not OK until all instances are done starting
	StatusRecoverableError
	StatusStopped
	StatusStopping // Aggregate is not Stopped until all instances are done stopping
	StatusPermanentError
	StatusFatalError
```

My proposed logic didn't account for timestamps, but I think there are two aspects to this which we need to handle.
1. Ensure that we only consider the latest event from each instance. I think you could solve this by changing the signature to accept a map: `AggregateStatus(map[InstanceID]*StatusEvent) *StatusEvent`
2. Assuming (1), what is the appropriate timestamp to apply to an aggregated status? We could probably make a case for any of the following:
    - Earliest, because it indicates the "most stale" instance status.
    - Latest, because it indicates the most recently reported instance status.
    - Average, median, or some other statistical representation e.g. excluding outliers
    - Some combination of the above. This would require a slightly different structured though.

component/status.go
```diff
@@ -0,0 +1,113 @@
...
+// status.
+type StatusWatcher interface {
+	// ComponentStatusChanged notifies about a change in the source component status.
+	// Extensions that implement this interface must be ready that the ComponentStatusChanged
+	// may be called  before, after or concurrently with Component.Shutdown() call.
+	// The function may be called concurrently with itself.
+	ComponentStatusChanged(source *InstanceID, event *StatusEvent)
+}
+
+// StatusFunc is the expected type of ReportComponentStatus for compoment.TelemetrySettings
```
mx-psi, line 112:
```suggestion
// StatusFunc is the expected type of ReportComponentStatus for component.TelemetrySettings
```

service/internal/status/status.go
mwear: I made an attempt at this. I modified `SharedComponent.GetOrAdd` to take a reference to a `component.TelemetrySettings` and modified the `ReportComponentStatus` of the `SharedComponent` to call the `ReportComponentStatus` methods of its "instances". I built the collector and checked that it works in the real world with the OTLP receiver.

component/component.go
mwear: After thinking about this a little more, I don't think it's necessary to return a StatusEvent for an aggregate status. An effective status is most likely all that we're going to need. I went ahead and implemented `component.AggregateStatus` based on your rules from https://github.com/open-telemetry/opentelemetry-collector/pull/8169#issuecomment-1670048722. I think it's good that we codified the rules for computing an effective status and ff it turns out we need more later, we can build on this initial implementation. Does that sound reasonable to you?

mwear: I wrote up a proposal for how we can compute effective status for pipelines by aggregating component status, and how we can compute effective status for the collector overall by aggregating pipeline status. The proposal is here: https://github.com/open-telemetry/opamp-spec/issues/165#issuecomment-1722617524. I added two more utility methods so that we can find the most recent status event (`component.LastStatusEvent`) and find the most recent event with a specific status (`component.LastEventByStatus`). Along with `component.AggregateStatus`, I believe we have all the utility methods a `StatusWatcher` would need to report status according to my proposal. Since we need to be able to group by `InstanceID` and pipelineID (which is of type `component.ID`), I implemented the functions with a type parameter for the key of the map of status events passed in. They will work with any comparable.

mwear: I guess I've come full circle and have decided that it would be useful to have a method that returns an effective status (as a status event). I called the method `component.EffectiveStatus`. Based on the OpAMP proposal, the effective status should have:

* a status equal to the aggregated status
* the timestamp of the most recent status event
* for an error status, the error of the most recent event with the same error status

The status is an aggregation of all the current states, the timestamp represents the most recent state change, and in the case of an error status, the error on the event will match the most recent occurrence of the same error state.

mwear:
FYI, I'm on vacation 🌴 until 9/25. Please continue to review and leave feedback. I'll start addressing it when I return.

component/telemetry.go
```diff
@@ -29,4 +29,21 @@ type TelemetrySettings struct {
...
 	Resource pcommon.Resource
+
+	// ReportComponentStatus allows a component to report runtime changes in status. The service
+	// will automatically report status for a component during startup and shutdown. Components can
+	// use this method to report status after start and before shutdown. ReportComponentStatus
+	// will only return errors if the API used incorrectly. The three scenarios where an error will
+	// be returned are:
+	//
+	//   - An illegal state transition
+	//   - Using the WithError() option with a non-error status
```
djaglowski, line 40:
```suggestion
```

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -27,19 +27,37 @@ func NewSharedComponents[K comparable, V component.Component]() *SharedComponent
 
 // GetOrAdd returns the already created instance if exists, otherwise creates a new instance
 // and adds it to the map of references.
-func (scs *SharedComponents[K, V]) GetOrAdd(key K, create func() (V, error)) (*SharedComponent[V], error) {
+func (scs *SharedComponents[K, V]) GetOrAdd(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*SharedComponent[V], error) {
 	if c, ok := scs.comps[key]; ok {
+		// If we haven't already seen this telemetry settings, this shared component represents
+		// another instance. Wrap ReportComponentStatus to report for all instances this shared
+		// component represents.
+		if _, ok := c.seenSettings[telemetrySettings]; !ok {
```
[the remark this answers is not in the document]
mwear: I was being defensive about calling `GetOrAdd` multiple times. If this is something that we don't need to worry about it, I can remove the check (and `seenSettings`).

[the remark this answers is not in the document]
djaglowski: In practice I don't think it's really intended that this function is called more than once per instantiation. That said, you're probably right to check it anyways because it is technically possible.

component/status.go
```diff
@@ -0,0 +1,215 @@
...
+
+	if _, isStopped := seen[StatusStopped]; isStopped {
+		return StatusStopping
+	}
+
+	if _, isRecoverable := seen[StatusRecoverableError]; isRecoverable {
+		return StatusRecoverableError
+	}
+
+	return StatusStarting
```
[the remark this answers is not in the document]
mwear: I added some comments to make it clarify what cases are being handled where. While I was at it, I renamed `EffectiveStatus` to `AggregateStatusEvent` since it returns a status event (and we have the related `AggregateStatus` function that returns a status).

djaglowski (approved):

component/telemetry.go
```diff
@@ -29,4 +29,20 @@ type TelemetrySettings struct {
 
 	// Resource contains the resource attributes for the collector's telemetry.
 	Resource pcommon.Resource
+
+	// ReportComponentStatus allows a component to report runtime changes in status. The service
+	// will automatically report status for a component during startup and shutdown. Components can
+	// use this method to report status after start and before shutdown. ReportComponentStatus
+	// will only return errors if the API used incorrectly. The three scenarios where an error will
```
evan-bradley, line 36:
```suggestion
	// will only return errors if the API used incorrectly. The two scenarios where an error will
```
There are only two scenarios listed, is there one missing or should this be updated?

evan-bradley (commented):
I'm still working through this, but so far it looks good to me. I'll take another look tomorrow.

component/telemetry.go
mwear: There were three scenarios, but after some refactoring, there are only two. Thanks for correcting.

evan-bradley (approved):
Looks good to me, just a few minor things. Thanks for taking this on.

component/status.go
```diff
@@ -0,0 +1,219 @@
...
+func NewFatalErrorEvent(err error) *StatusEvent {
+	ev := NewStatusEvent(StatusFatalError)
+	ev.err = err
+	return ev
+}
+
+// StatusWatcher is an extra interface for Extension hosted by the OpenTelemetry
+// Collector that is to be implemented by extensions interested in changes to component
+// status.
+type StatusWatcher interface {
```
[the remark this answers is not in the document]
mwear: I moved it. The `extension` package is a better place for it. Thanks for the suggestion.

component/status.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// Handle mixed status cases
+	if _, isFatal := seen[StatusFatalError]; isFatal {
+		return StatusFatalError
+	}
+
+	if _, isPermanent := seen[StatusPermanentError]; isPermanent {
+		return StatusPermanentError
+	}
```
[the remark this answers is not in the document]
djaglowski: I think it depends on whether we expose multiple levels of granularity, or just one. As I understand it, the intention here is that one could look at each component independently, or each pipeline independently, or the collector as a whole.

So if for example you have a component in a permanent error state, the collector would have an aggregate status of permanent error, but one could still see each component status independently.

extension/extension.go
```diff
@@ -40,6 +40,17 @@ type ConfigWatcher interface {
...
 
+// StatusWatcher is an extra interface for Extension hosted by the OpenTelemetry
+// Collector that is to be implemented by extensions interested in changes to component
+// status.
+type StatusWatcher interface {
+	// ComponentStatusChanged notifies about a change in the source component status.
+	// Extensions that implement this interface must be ready that the ComponentStatusChanged
+	// may be called before, after or concurrently with calls to Component.Start() and Component.Shutdown().
+	// The function may be called concurrently with itself.
+	ComponentStatusChanged(source *component.InstanceID, event *component.StatusEvent)
```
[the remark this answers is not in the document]
mwear: I think we should only pass the component status to the watcher, and the watcher can aggregate status as it sees fit. Currently I am prototyping using this from the healtcheck extension, and I am aggregating status by pipeline, as well as keeping an aggregate status for the collector overall.

component/status.go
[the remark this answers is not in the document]
mwear: Confirming what @djaglowski says is saying, an aggregate status is useful for the seeing the overall health of a collector, or the health of a pipeline, but we intend to expose the individual component status in addition to the aggregate statuses. A user can look at the aggregate statuses to quickly assess the health of a collector, but then drill down to the component level for more details.

service/service_test.go
```diff
@@ -414,6 +414,29 @@ func TestServiceTelemetryLogger(t *testing.T) {
...
+
+	go func() {
+		ev := component.NewFatalErrorEvent(assert.AnError)
+		srv.host.notifyComponentStatusChange(&component.InstanceID{}, ev)
+	}()
+
+	err = <-srv.host.asyncErrorChannel
+
+	require.Error(t, err)
+	require.ErrorIs(t, err, assert.AnError)
```
codeboten, line 437: is this needed if you're already requiring an error above it?

processor/processortest/unhealthy_processor.go
```diff
@@ -0,0 +1,70 @@
...
+type unhealthyProcessor struct {
+	component.StartFunc
+	component.ShutdownFunc
+	consumertest.Consumer
+	telemetry component.TelemetrySettings
+}
+
+func (p unhealthyProcessor) Start(_ context.Context, _ component.Host) error {
+	go func() {
+		_ = p.telemetry.ReportComponentStatus(component.NewStatusEvent(component.StatusRecoverableError))
```
codeboten, line 67: Would it be useful for the processor to support other events as well?

service/host.go
```diff
@@ -33,10 +33,10 @@ type serviceHost struct {
 // ReportFatalError is used to report to the host that the receiver encountered
 // a fatal error (i.e.: an error that the instance can't recover from) after
 // its start function has already returned.
+// Deprecated: [x.x.x] Replaced by servicetelemetry.Settings.ReportComponentStatus
```
codeboten, line 36:
```suggestion
// Deprecated: [0.87.0] Replaced by servicetelemetry.Settings.ReportComponentStatus
```

service/internal/graph/graph.go
```diff
@@ -45,12 +46,19 @@ type Graph struct {
 
 	// Keep track of how nodes relate to pipelines, so we can declare edges in the graph.
 	pipelines map[component.ID]*pipelineNodes
+
+	// Keep track of status source per node
+	instanceIDs map[int64]*component.InstanceID
+
+	telemetry servicetelemetry.TelemetrySettings
```
codeboten, line 53: why two servicetelemetry.TelemetrySettings fields?

mwear: The `Settings` struct get passed to `Graph.Build` and is passed around from there, but goes out of scope after `Build`. I need access to `TelemetrySettings` later during `Graph.StartAll` and `Graph.ShutdownAll`. There needs to be a way to access the telemetry settings for the lifetime of `Graph`.

processor/processortest/unhealthy_processor.go
mwear: It could be updated to support other events. We use this in a test where we start a collector and observe the results. It is not possible (or at least not easy) to get a handle to a processor to trigger events from a component. We could pass in a static event to the factory, but I don't think we'd take advantage of that functionality. The testcase this is used in verifies that an event make it from a component to a status watcher. It's not necessary to exhaustively test the type of events, we do that at a lower level. In addition, we automatically get `Starting`, `Stopping`, and `Stopped` events from auto status reporting in graph. tl;dr it could be made more generic, but we would not be taking advantage of it currently.

mwear: That said, if you'd like it made more generic, let me know.

service/service_test.go
mwear: Not needed. There were a few more instances of this that I also cleaned up.

codeboten: Thx

codeboten (approved):

service/internal/graph/graph.go
codeboten: Oh I see... I didn't realize the struct in question was Graph, and not Settings. Thanks

processor/processortest/unhealthy_processor.go
codeboten: It can be improved in the future if the need arises

receiver/otlpreceiver/otlp.go
```diff
@@ -43,13 +43,13 @@ type otlpReceiver struct {
 	obsrepGRPC *receiverhelper.ObsReport
 	obsrepHTTP *receiverhelper.ObsReport
 
-	settings receiver.CreateSettings
+	settings *receiver.CreateSettings
```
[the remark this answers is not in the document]
mwear: Yes. There are some complications where the telemetry settings for the shared component must be the same as the component it represents. A copy will not work.

codeboten:
@mwear please resolve the conflict and we can get this merged

PR #8177, opened 2023-08-04
[chore] Use env vars when running shell commands in CI

djaglowski:


djaglowski:
Verified that the changelog workflow runs with the env vars: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/5765664809/job/15632081184?pr=8177#step:5:16

dmitryax (approved):

Issue #8122, opened 2023-07-22
Introduce new exporter helper with batching option


dmitryax:
This is a tracking issue for introducing the new exporter helper and migrating the existing exporters to use it.

The primary reason for introducing the new exporter helper is to move the batching to the exporter side and deprecate the batch processor as part of making the delivery pipeline reliable, as reported in https://github.com/open-telemetry/opentelemetry-collector/issues/7460. More details about moving batching to the exporter helper can be found in https://github.com/open-telemetry/opentelemetry-collector/issues/4646.

Shifting batching to the exporter side grants us the opportunity to leverage the exporter's data model instead of relying on OTLP. As a result, we can achieve the following benefits:
- Ability to place failed requests back into the queue without the need for converting them back to OTLP format.
- Enhanced control in counting queue and batch sizes using basic items (like spans, data points, or log records for OTLP) tailored to different exporters, resolving the concern raised in issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/7134.
- Optional counting of queue and batch sizes in bytes of serialized data.

Adapting to the new exporter helper requires exporter developers to implement at least two functions:

1. Converter: to translate pdata Metrics/Traces/Logs into a user-defined Request.
2. Request sender: to send the user-defined Request

bogdandrutu:
Some feedback about the interface:
1. [Traces|Metrics|Logs]Converter can be replaced with a simple func instead of having an interface. If you need in the future to extend capabilities of that, they can be options. Comparing interfaces with nil is problematic sometimes, see https://mangatmodi.medium.com/go-check-nil-interface-the-right-way-d142776edef1, referring to https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/metrics.go#L142
2. Can you make the old NewMetricsExporter to call into NewMetricsRequestExporter to remove duplicate code and to actually start testing the new path more intensively? Also removes lots of duplicate tests.

dmitryax:
Thanks for the feedback!

> Can you make the old NewMetricsExporter to call into NewMetricsRequestExporter to remove duplicate code and to actually start testing the new path more intensively? Also removes lots of duplicate tests.

That's what I wanted to do after https://github.com/open-telemetry/opentelemetry-collector/pull/8248 is merged 👍

verejoel:
[...]

PR #8178, opened 2023-08-07
[exporterhelper] New exporter helper for custom requests

dmitryax:
Introduce a new exporter helper that operates over client-provided requests instead of pdata. The helper user now has to provide `Converter` - an interface with a function implementing translation of pdata Metrics/Traces/Logs into a user-defined `Request`. `Request`  is an interface with only one required function `Export`.

It opens a door for moving batching to the exporter, where batches will be built from client data format, instead of pdata. The batches can be properly sized by custom request size, which can be different from OTLP. The same custom request sizing will be applied to the sending queue. It will also improve the performance of the sending queue retries for non-OTLP exporters, they don't need to translate pdata on every retry.

This is an implementation alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/7874 as suggested in https://github.com/open-telemetry/opentelemetry-collector/pull/7874#discussion_r1278444371

Tracking Issue: https://github.com/open-telemetry/opentelemetry-collector/issues/8122

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -56,32 +56,46 @@ func (req *baseRequest) OnProcessingFinished() {
 	}
 }
 
+type queueSettings struct {
+	config      QueueSettings
+	marshaler   internal.RequestMarshaler
+	unmarshaler internal.RequestUnmarshaler
+}
+
+func (qs *queueSettings) persistenceEnabled() bool {
+	return qs.config.StorageID != nil && qs.marshaler != nil && qs.unmarshaler != nil
+}
+
 // baseSettings represents all the options that users can configure.
 type baseSettings struct {
 	component.StartFunc
 	component.ShutdownFunc
 	consumerOptions []consumer.Option
 	TimeoutSettings
-	QueueSettings
+	queueSettings
 	RetrySettings
+	requestExporter bool
 }
 
-// fromOptions returns the internal options starting from the default and applying all configured options.
-func fromOptions(options ...Option) *baseSettings {
-	// Start from the default options:
-	opts := &baseSettings{
+// fromOptions returns the baseSettings starting from the default and applying all configured options.
+// requestExporter indicates whether the base settings are for a new request exporter or not.
+func newBaseSettings(requestExporter bool, options ...Option) *baseSettings {
+	bs := &baseSettings{
+		requestExporter: requestExporter,
 		TimeoutSettings: NewDefaultTimeoutSettings(),
 		// TODO: Enable queuing by default (call DefaultQueueSettings)
-		QueueSettings: QueueSettings{Enabled: false},
+		queueSettings: queueSettings{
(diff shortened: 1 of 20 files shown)


jmacd (approved):

exporter/exporterhelper/common.go
```diff
@@ -56,32 +56,46 @@ func (req *baseRequest) OnProcessingFinished() {
...
+	queueSettings
 	RetrySettings
+	requestExporter bool
 }
 
-// fromOptions returns the internal options starting from the default and applying all configured options.
-func fromOptions(options ...Option) *baseSettings {
-	// Start from the default options:
-	opts := &baseSettings{
+// fromOptions returns the baseSettings starting from the default and applying all configured options.
```
codeboten, line 80:
```suggestion
// newBaseSettings returns the baseSettings starting from the default and applying all configured options.
```

exporter/exporterhelper/metrics.go
```diff
@@ -115,6 +118,55 @@ func NewMetricsExporter(
 	}, err
 }
 
+type MetricsConverter interface {
+	// RequestFromMetrics converts pdata.Metrics into a request.
+	RequestFromMetrics(context.Context, pmetric.Metrics) (Request, error)
+}
+
+// NewMetricsRequestExporter creates a new metrics exporter based on a custom TracesConverter and RequestSender.
```
codeboten, line 126:
```suggestion
// NewMetricsRequestExporter creates a new metrics exporter based on a custom MetricsConverter and RequestSender.
```

exporter/exporterhelper/request.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type request struct {
+	Request
+	baseRequest
+}
+
+var _ internal.Request = (*request)(nil)
+
+func (req *request) OnError(_ error) internal.Request {
```
codeboten, line 36: are OnError and Count defined in an interface somewhere?

codeboten (commented):
The change looks ok, just one question and a suggestion

exporter/exporterhelper/request.go
dmitryax: Yes, they are part of the internal.Request interface https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/request.go

codeboten: Ah yes... this was the piece i was missing 👍🏻

codeboten (approved):

jpkrohling (approved):
Looks nice, I prefer this over the previous PR.

dmitryax:
@codeboten, how did it happen to be added to 0.83.0? It should be part of 0.84.0, right?

codeboten:
@dmitryax yeah sorry i tagged it incorrectly.... let me revisit the milestone labeling i did manually yesterday

Issue #5403, opened 2022-05-20
[builder] TestGenerateAndCompileDefault ant TestSkipGenerate are failing on Windows


dmitryax:
```
=== RUN   TestGenerateAndCompileDefault
2022-05-20T00:28:04.056Z	INFO	builder/config.go:99	Using go	{"go-executable": "C:\\hostedtoolcache\\windows\\go\\1.18.2\\x64\\bin\\go.exe"}
2022-05-20T00:28:04.056Z	INFO	builder/main.go:52	You're building a distribution with non-aligned version of the builder. Compilation may fail due to API changes. Please upgrade your builder or API	{"builder-version": "0.51.0"}
2022-05-20T00:28:04.095Z	INFO	builder/main.go:76	Sources created	{"path": "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\default1503028659"}
2022-05-20T00:28:43.056Z	INFO	builder/main.go:108	Getting go modules
2022-05-20T00:28:44.049Z	INFO	builder/main.go:87	Compiling
    main_test.go:51: 
        	Error Trace:	main_test.go:51
        	Error:      	Received unexpected error:
Error:         	            	failed to compile the OpenTelemetry Collector distribution: exit status 2. Output: "# go.opentelemetry.io/collector/cmd/builder\n.\\main_windows.go:45:32: undefined: service.NewSvcHandler\n"
        	Test:       	TestGenerateAndCompileDefault
--- FAIL: TestGenerateAndCompileDefault (77.82s)
```

See https://github.com/open-telemetry/opentelemetry-collector/runs/6517064706?check_suite_focus=true

kristinapathak:
I tried running the `GeneratorAndCompile` unit tests on windows in https://github.com/open-telemetry/opentelemetry-collector/pull/6946 and they failed:
```
--- FAIL: TestGenerateAndCompileDefault (91.37s)
Error:     testing.go:1097: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestGenerateAndCompileDefault3151526655\001\components.go: The process cannot access the file because it is being used by another process.
--- FAIL: TestGenerateAndCompileWithLDFlags (6.99s)
[...]

PR #8208, opened 2023-08-09
Remove go 1.19 support, bump minimum to go 1.20 and add testing for 1.21

### LINKED ISSUE #8207: Add testing for Go 1.21, drop testing for Go 1.19
Same as #7159, but for Go 1.21/Go 1.19.

Go 1.21 was released recently: https://go.dev/doc/go1.21

Per our [Compatibility](https://github.com/open-telemetry/opentelemetry-collector#compatibility) doc, this means we need to drop testing for Go 1.19, add testing for Go 1.21. This also means we can now use Go 1.20+ features.

mx-psi:
Fixes #8207

### DIFF
--- a/.chloggen/mx-psi_refactor-service-telemetry.yaml
+++ b/.chloggen/mx-psi_refactor-service-telemetry.yaml
@@ -14,3 +14,5 @@ issues: [8207]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext:
+
+change_logs: [api]
--- a/.chloggen/mx-psi_refactor-service-telemetry.yaml
+++ b/.chloggen/mx-psi_refactor-service-telemetry.yaml
@@ -1,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: all
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove go 1.19 support, bump minimum to go 1.20 and add testing for 1.21
+
+# One or more tracking issues or pull requests related to the change
+issues: [8207]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: ~1.19.12
+          go-version: ~1.20.7
       - name: Cache Go
         id: go-cache
         uses: actions/cache@v3
@@ -41,7 +41,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: ~1.19.12
+          go-version: ~1.20.7
       - name: Cache Go
         id: go-cache
         uses: actions/cache@v3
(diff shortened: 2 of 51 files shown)


mx-psi:
There is a builder test consistently failing on this PR that is not failing on main

```
--- FAIL: TestSkipGenerate (1.75s)
    testing.go:1206: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestSkipGenerate2118395261\001: The process cannot access the file because it is being used by another process.
FAIL
FAIL	go.opentelemetry.io/collector/cmd/builder/internal/builder	1.836s
FAIL
```

mx-psi:
I think the error is related to actions/runner-images/issues/712, although I am not sure why this only happens with Go 1.20.

mx-psi:
I decided to skip the test and just list it on #5403 together with the other test that is already failing because of the same reason

cmd/builder/internal/builder/main_test.go
```diff
@@ -36,6 +36,10 @@ func TestGenerateInvalidOutputPath(t *testing.T) {
 }
 
 func TestSkipGenerate(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping the test on Windows, see https://github.com/open-telemetry/opentelemetry-collector/issues/5403")
+	}
+
```
djaglowski, line 42: Unrelated - the `LDFlags Compilation` test case in this file contains a reference to go `1.19` which may need to be updated.

mx-psi: I think any value works on this test, it's just setting a linker argument to `1.19` but it should not make any difference to the builder. Do you think we should change it still?

djaglowski: I'm fine with leaving it.

.github/workflows/tidy-dependencies.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           token: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
       - uses: actions/setup-go@v4
         with:
-          go-version: 1.19
+          go-version: 1.20
```
bryan-aguilar, line 21:
```suggestion
          go-version: ~1.20.7
```
This would align this setup go with other worfklows.

.github/workflows/build-and-test.yml
```diff
@@ -133,7 +133,7 @@ jobs:
   unittest:
     strategy:
       matrix:
-        go-version: ["1.20", 1.19] # 1.20 needs quotes otherwise it's interpreted as 1.2
+        go-version: ["1.21", "1.20"] # 1.20 needs quotes otherwise it's interpreted as 1.2
```
bryan-aguilar, line 136: Should this also use the `~` convention to force min version? Aligning across actions could make search and replace easier in the future.

mx-psi: Happy to do that, but I am worried that `~` is not a supported character on cache keys and since we use this variable here https://github.com/open-telemetry/opentelemetry-collector/blob/929cce9f18fb3014385da1d54fc4137efe434aec/.github/workflows/build-and-test.yml#L158 it would break things, so I would prefer to work on this on a separate PR.

dmitryax:
Apparently, I cannot update the protected branch rules. Submitted https://github.com/open-telemetry/community/issues/1636

mx-psi:
This is blocked by #8218

arminru:
> Apparently, I cannot update the protected branch rules. Submitted [open-telemetry/community#1636](https://github.com/open-telemetry/community/issues/1636)

Branch rule updated:
image

.chloggen/mx-psi_refactor-service-telemetry.yaml
```diff
@@ -0,0 +1,16 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove go 1.19 support, bump minimum to go 1.20 and add testing for 1.21
+
+# One or more tracking issues or pull requests related to the change
+issues: [8207]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
dmitryax, line 16: Let's adopt the new changelog format and mark this as an API breaking change. Otherwise, it'll go to the changelog for collector end-users

dmitryax (approved):

PR #8211, opened 2023-08-09
Add omits to confighttp's TLSSetting and CustomRoundTripper fields

JayPolanco:
**Description:**
- Adds `omitempty` mapstructure tag to confighttp's `TLSSetting` field - The [resourcedetection](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md) processor defined in opentelemetry-collector-contrib panics with a nullpointer exception when the default values are used. 
- Adds ignore to CustomRoundtripper - The resourcedetection processor cannot be marshalled due to the `CustomRoundTripper` field being of type function. To allow marshalling we ignore this field.

**Testing:** ensured all current tests pass. `make gotest`

### DIFF
--- a/.chloggen/tlsomit.yaml
+++ b/.chloggen/tlsomit.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confighttp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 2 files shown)


JayPolanco:
@dmitryax Apologies for the delay. If you execute something like the following you will get the panic (i.e `panic: cannot marshal type: func(http.RoundTripper) (http.RoundTripper, error)`):

```
func TestYAMLMarshall(t *testing.T) {
	factory := resourcedetectionprocessor.NewFactory()
	cfg := factory.CreateDefaultConfig().(*resourcedetectionprocessor.Config)

	var buf bytes.Buffer
	enc := yaml.NewEncoder(&buf)
	err := enc.Encode(cfg)
	fmt.Println(err)
}
```

We get the former issue when we set up a resourcedetection processor that does not have a TLSSetting configured.

Issue #8210, opened 2023-08-09
Use Go 1.20 `errors.Join` instead of `go.uber.org/multierr`


mx-psi:
Go 1.20 includes new builtin [error joining support](https://pkg.go.dev/errors#Join) that can replace the usage of `go.uber.org/multierr`. We should switch to using these and ban the usage of `go.uber.org/multierr` via `depguard` (same as [this](https://github.com/open-telemetry/opentelemetry-collector/blob/9d7deb4098e51207ba6e3a95d76c0d3b152a8d11/.golangci.yml#L106-L107)).

This is **blocked by** #8207

shivanshu1333:
@mx-psi kindly assign this one as well

bogdandrutu:
@mx-psi can you document why we stopped this effort until better understand the performance issue.

mx-psi:
I removed the `good first issue` and `easy` labels and added `on hold` instead.

The 'easy' way to change the code to use `errors.Join` means creating a slice of errors, which in the common case will all be nil, and then joining that slice. `multierr.Append` instead avoids this by checking if the error is nil before adding to the internal slice, which is more performant.

We can still use `errors.Join` and avoid this by checking that the errors are not nil before appending to the slice, but it leads to more verbose code.

dmathieu:
Note that `errors.Join` does check if the error is nil, and skips adding those into the internal slice.
https://github.com/golang/go/blob/a9922d096f0de877fba68739b35367d4c25f6ecb/src/errors/join.go#L33-L35
So I'm not sure we do need to check that errors aren't nil before joining them.

I just proposed this additional patch to the Go team which will optimize the case where a single non-nil error is provided, and it can be returned as-is.
https://github.com/golang/go/pull/70770

mx-psi:
@dmathieu The patch looks great, thanks for doing this :) 

I think the problem I was pointing at above is not addressed by existing code. The problem is the difference between these two:
```go
plainJoin := errors.Join(e1, e2, e3)
nestedJoin := errors.Join(errors.Join(e1, e2), e3)

plainAppend := multierr.Append(e1, e2, e3)
nestedAppend := multierr.Append(multierr.Append(e1, e2), e3)
```

[...]

PR #8225, opened 2023-08-14
[chore] migrate  receiver/* to use errors.Join

shivanshu1333:
Part of: https://github.com/open-telemetry/opentelemetry-collector/issues/8210

### DIFF
--- a/receiver/scraperhelper/scrapercontroller.go
+++ b/receiver/scraperhelper/scrapercontroller.go
@@ -143,12 +143,12 @@ func (sc *controller) Shutdown(ctx context.Context) error {
 		<-sc.terminated
 	}
 
-	var errs error
+	var errs []error
 	for _, scraper := range sc.scrapers {
-		errs = errors.Join(errs, scraper.Shutdown(ctx))
+		errs = append(errs, scraper.Shutdown(ctx))
 	}
 
-	return errs
+	return errors.Join(errs...)
 }
 
 // startScraping initiates a ticker that calls Scrape based on the configured
--- a/receiver/scraperhelper/scrapercontroller.go
+++ b/receiver/scraperhelper/scrapercontroller.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"time"
 
-	"go.uber.org/multierr"
 	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
@@ -146,7 +145,7 @@ func (sc *controller) Shutdown(ctx context.Context) error {
 
 	var errs error
 	for _, scraper := range sc.scrapers {
-		errs = multierr.Append(errs, scraper.Shutdown(ctx))
+		errs = errors.Join(errs, scraper.Shutdown(ctx))
 	}
 
 	return errs
--- a/receiver/go.mod
+++ b/receiver/go.mod
@@ -10,7 +10,6 @@ require (
 	go.opentelemetry.io/collector/pdata v1.0.0-rcv0014
 	go.opentelemetry.io/otel v1.16.0
 	go.opentelemetry.io/otel/sdk v1.16.0
-	go.uber.org/multierr v1.11.0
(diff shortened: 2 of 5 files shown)


receiver/scraperhelper/scrapercontroller.go
```diff
@@ -146,7 +145,7 @@ func (sc *controller) Shutdown(ctx context.Context) error {
 
 	var errs error
 	for _, scraper := range sc.scrapers {
-		errs = multierr.Append(errs, scraper.Shutdown(ctx))
+		errs = errors.Join(errs, scraper.Shutdown(ctx))
```
mx-psi, line 148: @seankhliao I see you have participated in some discussions related to the `errors.Join` implementation. Would you mind commenting on whether this pattern should be used? Or should we prefer appending to an `[]error` and only then call `errors.Join`?

codeboten (commented):
Is there a way to override how `errors.Join` outputs the errors? Looking at https://github.com/open-telemetry/opentelemetry-collector/actions/runs/5852149629/job/15867545790?pr=8225#step:6:243, it looks like the output will be changed:

```
--- FAIL: TestScrapeErrorsCombine (0.00s)
    scrapeerror_test.go:95: 
        	Error Trace:	/home/runner/work/opentelemetry-collector/opentelemetry-collector/receiver/scrapererror/scrapeerror_test.go:95
        	Error:      	Error message not equal:
        	            	expected: "bad scrapes; err: bad scrape"
        	            	actual  : "bad scrapes\nerr: bad scrape"
        	Test:       	TestScrapeErrorsCombine
    scrapeerror_test.go:95: 
        	Error Trace:	/home/runner/work/opentelemetry-collector/opentelemetry-collector/receiver/scrapererror/scrapeerror_test.go:95
        	Error:      	Error message not equal:
        	            	expected: "bad regular; err: bad reg"
        	            	actual  : "bad regular\nerr: bad reg"
        	Test:       	TestScrapeErrorsCombine
    scrapeerror_test.go:95: 
        	Error Trace:	/home/runner/work/opentelemetry-collector/opentelemetry-collector/receiver/scrapererror/scrapeerror_test.go:95
        	Error:      	Error message not equal:
        	            	expected: "bad two scrapes; 10 scrapes failed: bad things happened; bad event; event: something happened"
        	            	actual  : "bad two scrapes\n10 scrapes failed: bad things happened\nbad event\nevent: something happened"
        	Test:       	TestScrapeErrorsCombine
FAIL
```

receiver/scraperhelper/scrapercontroller.go
bogdandrutu: I think Join replaces nicely `multierr.Combine`, but does not have a good equivalent for `Append`.

seankhliao: sorry I missed this last week.
I think semantically it would be more appropriate to populate a slice and join that.
On implementation, `errors.Join(slice...)` would inspect (`errors.Is/As`) the errors based on the order they were added, while `err = errors.Join(err, newErr)` would do so in reverse order.

seankhliao:
`errors.Join` doesn't offer any direct control over the output format, it was decided that there are too many opinions to easily support and instead `errors.Join` should remain opinionated.

receiver/scraperhelper/scrapercontroller.go
mx-psi: Thanks @seankhliao! Could you update the PR based on this,@shivanshu1333?

shivanshu1333: Sure!

shivanshu1333: done, sorry for the delay

mx-psi (commented):
Seems like you need to update some tests since errors are now separated by `\n` instead of `;` and we can't control the output format

PR #8228, opened 2023-08-14
Add support for :authority pseudo-header for grpc client

enuret:
This adds support for configuring `authority` dial option for grpc exporter

### DIFF
--- a/.chloggen/otlp_with_authority.yaml
+++ b/.chloggen/otlp_with_authority.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: configgrpc
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add support for :authority pseudo-header in grpc client"
+
+# One or more tracking issues or pull requests related to the change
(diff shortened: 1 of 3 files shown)


codeboten (commented):
Thanks for the contribution @enuret. The change looks ok, i'm not familiar w/ the `:authority` header can you tell me a bit more about the use-case for setting it?

enuret:
> Thanks for the contribution @enuret. The change looks ok, i'm not familiar w/ the `:authority` header can you tell me a bit more about the use-case for setting it?

Hey @codeboten! This header is a destination host-name by default.

My particular case is configuring envoy to balance traffic between two opentelemetry collectors or between opentelemetry collector and data backend and using  envoy virtual hosts to match backends.

So, envoy uses ":authority"  header for http/2 requests to match them to virtual hosts: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/route_matching

In my case the example call will look like:
collector (otlpexporter endpoint = envoyhost) -> envoy -> backend

By default, :authority header will be equal "envoyhost" because it is the endpoint address but if  "backend" will be specified in :authority header it will be possible to match it based on Virtual host domain name in envoy. 

For example, it will let routing based on domain name with configuration like it :

```
route_config:
  name: local_route
  virtual_hosts:
  - name: backend1
     domains: ["backend1"]
     routes:
     - match:
          prefix: /
        route:
          cluster: backend1

  - name: backend2
     domains: ["backend2"]
     routes:
     - match:
          prefix: /
        route:
          cluster: backend2
```

codeboten (approved):
Thanks for the info 👍🏻

Issue #8237, opened 2023-08-15
Allow configuration of collector service telemetry


akats7:
**Is your feature request related to a problem? Please describe.**

In certain cases it would be very useful to have the same control over the collectors own telemetry (mainly metrics) as we do over other OTEL data points. For example, I'd like to be able to modify the aggregation temporality for the metrics as it seems by default they are all cumulative, and for certain metrics like refused spans/metrics delta seems a more intuitive value. It would also be convenient to be able to enable/disable specific metrics the same way that we are able to in other components. 

**Describe the solution you'd like**
I would like to the service telemetry config to allow for configuration of telemetry data

**Describe alternatives you've considered**
Using the filter processor to exclude specific metrics, create queries on backend to extract delta.

**Additional context**
I'm aware this is not critical with the impending GA release, but wanted to bring it up as an eventual enhancement

codeboten:
Hello @akats7! Yes I'm currently working on supporting additional configuration for the Collector as part of https://github.com/open-telemetry/opentelemetry-collector/issues/7532. I've started making metrics more configurable using the scheme defined by the [configuration working group](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml) and would love more help to add additional feature support!

Let me know if this is something you'd like to help with, tackling configuring views was next on my list of priorities.

akats7:
Hi @codeboten, I'd be happy to help

Issue #7532, opened 2023-04-12
Proposed strategy for making the collector an observable application


codeboten:
This is a follow up to various conversations that have been had around this topic. I'd like to propose that we make the configuration for the collector follow the proposed configuration for SDKs (which is still under work). Ideally I'd like to use the collector as a prototype of the configuration as it evolves.

I propose that most of the configuration parsing be done inside the collector for the time being, with the plan to leverage the Go SDK's implementation in the future. In order for this to work, there is quite a bit of work that needs to happen:

- [ ] Update `telemetry` code to support more than a single prometheus instance
- [ ] Support additional exporters for exporting collector metrics
- [ ] Support additional exporters for exporting collector traces
- [ ] Support additional exporters for exporting collector logs
- [ ] Collector `telemetry` stanza must support additional configuration parameters

Related issues:
https://github.com/open-telemetry/opentelemetry-collector/issues/6629
https://github.com/open-telemetry/opentelemetry-collector/issues/7223

mattsains:
Hello, I am interested in helping on this initiative, specifically exporting internal spans. Please let me know what I can do.

For my own testing, I have been writing a hacky extension which registers itself as a span processor for internal traces, converts the sdk span to a ptrace span, and exports it over otlp. However, I think it would be nice to work on a more official feature which could help others as well.

mattsains:
I think it would be worth adding a new issue here, for being able to turn off the existing telemetry metrics prometheus endpoint. Because even if you use the new export functionality for metrics, the collector still tries to open a prometheus port, and this behaviour cannot be disabled (only relocated to a different port). This causes problems if you try to run two collector instances on the same host

kasun-bandara:
> Hello, I am interested in helping on this initiative, specifically exporting internal spans. Please let me know what I can do.
> 
> For my own testing, I have been writing a hacky extension which registers itself as a span processor for internal traces, converts the sdk span to a ptrace span, and exports it over otlp. However, I think it would be nice to work on a more official feature which could help others as well.

Do we have the official processor support for internal traces/metrics yet?

mx-psi:
@codeboten is there anything left that we are tracking on this issue?

mattsains:
@julealgon does this help? https://opentelemetry.io/docs/collector/internal-telemetry/

julealgon:
> [@julealgon](https://github.com/julealgon) does this help? https://opentelemetry.io/docs/collector/internal-telemetry/

[...]

Issue #3039, opened 2021-04-28
component support custom metric view register


shilicqupt:
**Is your feature request related to a problem? Please describe.**
- custom Component need more rich metric to reflect Component internal states, however appTelemetry only support limited component metric views(batchprocessor, kafkareceiver, ....), we need a solution to help custom component to define MetricView and add to appTelemetry

**Describe the solution you'd like**
- in appTelemetry.go add metric register solution

```golang
var customViews = make(map[string][]*view.View)

// RegisterCustomView registers component views.
func RegisterCustomView(typ string, customView []*view.View) {
	if _, ok := customViews[typ]; ok {
		panic(" typ exists:" + typ)
	}
	customViews[typ] = customView
}
```

**Describe alternatives you've considered**
- without register func, we need to modify the source code support

**Additional context**
- self health metric is very import to see collector state

bogdandrutu:
OpenCensus is a singleton, you can do this for the moment in an `init` func and will work until we change to OpenTelemetry. Then we will give an explicit call.

shilicqupt:
> OpenCensus is a singleton, you can do this for the moment in an `init` func and will work until we change to OpenTelemetry. Then we will give an explicit call.

what means change to OpenTelemetry?

PR #8247, opened 2023-08-17
[service] add support for views

codeboten:
This adds support for configuring metric views for internal service telemetry.

This follows #8119

Related to #3039 #8237 #https://github.com/open-telemetry/opentelemetry-collector/issues/7532

### DIFF
--- a/service/telemetry.go
+++ b/service/telemetry.go
@@ -163,6 +163,8 @@ func (tel *telemetryInitializer) initMetrics(res *resource.Resource, logger *zap
 		opts = append(opts, sdkmetric.WithReader(r))
 	}
 
+	opts = append(opts, telemetry.ViewOptionsFromConfig(cfg.Metrics.Views)...)
+
 	mp, err := proctelemetry.InitOpenTelemetry(res, opts, tel.disableHighCardinality)
 	if err != nil {
 		return err
--- a/.chloggen/codeboten_configure-views.yaml
+++ b/.chloggen/codeboten_configure-views.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: add support for metric views configuration
+
+# One or more tracking issues or pull requests related to the change
+issues: [8237]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -76,9 +76,55 @@ For begin/stop events we need to define an appropriate hysteresis to avoid gener
 
 The service should collect host resource metrics in addition to service's own process metrics. This may help to understand that the problem that we observe in the service is induced by a different process on the same host.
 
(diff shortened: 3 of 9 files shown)


service/telemetry.go
```diff
@@ -163,6 +163,8 @@ func (tel *telemetryInitializer) initMetrics(res *resource.Resource, logger *zap
 		opts = append(opts, sdkmetric.WithReader(r))
 	}
 
+	opts = append(opts, telemetry.ViewOptionsFromConfig(cfg.Metrics.Views)...)
```
jpkrohling, line 166: Does this mean that views are only coming from the user-provided config, and not from components? I understand this is not the final state, but could you add some documentation on how to use this?

codeboten: That's a good point. I'll have to spend a bit of time trying to figure out how to give components the ability to create their own views.

codeboten:
One of the issues w/ this change is that a handful of components today configure metric views (using the opencensus registration mechanism available at runtime) wouldn't be able to do so in OTel. Currently there would be no way for a collector component to specify aggregation configuration. This is because there’s no way for components to return views to the service that’s configuring the otel go SDK. 

_Is this something we want components to be able to specify or is this something that should be left to users to configure?_

There are 5 components (loadbalancingexporter, groupbyattrsprocessor, groupbytraceprocessor, tailsamplingprocessor, batchprocessor) that configure `view.Distribution` for their metrics. To get around this problem for batch processor metrics, there is a `batchViews` function called by the service telemetry initialization currently. This is somewhat problematic because now the service has to configure a view for a metric that may or may not be there. I wouldn’t want to configure all of the configuration this way as it would leak information from contrib components into the core.

Here are some of the options we have moving forward.

### Remove component configured views when moving to Otel and only support user configured views
#### Pro
  - Not having to support components configuring views means there's no additional info that needs to be passed from components to the service
  - There are no more views that are automatically configured for end users, this can be somewhat surprising for users?
#### Con
  - This would be a breaking change in the telemetry emitted by anyone using the metrics today. This would all be behind a feature gate so it would allow users to migrate slowly.
  - It would also mean that anyone wanting aggregation other than the default aggregation would need additional configuration in their collector config.

### Leave metrics that require custom aggregation using opencensus. 
#### Pro
  - Least amount of moving pieces
  - Behaviour doesn't change for end users
#### Con
  - This would mean we could never remove OC dependency 

### Add a mechanism for components factories to provide views
#### Pro
  - Gives component authors and users most flexibility
#### Con
  - Adds public API surface

@jpkrohling any thoughts on the best approach here or alternatives?

jpkrohling:
> Is this something we want components to be able to specify or is this something that should be left to users to configure?

I believe component owners are best positioned to configure how the telemetry should be exposed. I think advanced users might be interested in tweaking this, but I don't think the majority of the users of, say, load-balancing exporter would be willing to configure the views in the config to have the same level of metrics that are being exposed today.

> Add a mechanism for components factories to provide views

I'd vote for this. It could be an optional interface, and if implemented by components, it's called when the service is configured.

dashpole:
Other than default buckets for histograms, is there anything else that can be done with views that can't be done with the API itself?  If this is going to be obsolete in ~a month when the hint API is added, it seems potentially worth waiting.  Views being provided by libraries containing the instrumentation seems to go against their intended purpose.

Go already has an RC, so I would expect a stable release in a few weeks.  The `ExplicitBucketBoundaries` config already has the necessary approvals to move to stable: https://github.com/open-telemetry/opentelemetry-specification/pull/3694.  It could possibly be part of the stable Go release if that happens...

codeboten:
@jpkrohling i've filled in a section under the observability doc on how users may enable configuration support. I've not added a section on configuring views specifically, but there is a link to the otel-configuration repo that contains that information.  Here's an example of how this could be used:

```
      views:
      - selector:
          instrument_name: process_memory_rss
          instrument_type: observable_gauge
        stream:
          name: new_instrument_name
          description: new_description
          aggregation:
            last_value: {}
          attribute_keys:
            - key1
            - key2
      - selector:
          instrument_name: "*"
        stream:
          aggregation:
            drop: {}
```

It sounds like the work to include hints in the API will be unblocked soon, it might be worth waiting for this work to move this views PR forward, WDYT?

codeboten:
@jpkrohling yes I expect this to be a more advanced use-case since the hints API should fulfill the more common needs for components

PR #8249, opened 2023-08-18
[chore] remove BUILD_INFO

codeboten:
This is no longer used as the build version information comes from the builder configuration. I've removed the VERSION variable as well, as it's only used in chlog-update and it's passed into the command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,11 +16,6 @@ ALL_MODULES := $(shell find . -type f -name "go.mod" -exec dirname {} \; | sort 
 
 CMD?=
 
-# TODO: Find a way to configure this in the generated code, currently no effect.
-BUILD_INFO_IMPORT_PATH=go.opentelemetry.io/collector/internal/version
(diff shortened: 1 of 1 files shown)


dmitryax (approved):

Makefile
```diff
@@ -16,11 +16,6 @@ ALL_MODULES := $(shell find . -type f -name "go.mod" -exec dirname {} \; | sort
 
 CMD?=
 
-# TODO: Find a way to configure this in the generated code, currently no effect.
-BUILD_INFO_IMPORT_PATH=go.opentelemetry.io/collector/internal/version
```
dmitryax, line 20: Do you remember when this package was removed by chance?

codeboten: Looks like it was removed about a year ago https://github.com/open-telemetry/opentelemetry-collector/commit/c5f118ef8667877edc94d87bc033d125674a5b0b

PR #8270, opened 2023-08-23
[exporter/otlphttpexporter] Partial success HTTP response handling

### LINKED ISSUE #8263: OTLPHTTP exporter ignores response code and content-type, retries and logs anyway
There exists a regression/bug in 0.81.0 (and 82 and 83) of the `otel-collector-contrib`
with the otlphttp exporter. This bug is initially observed when debug logging for
trace export is enabled:

```
{"kind": "exporter", "data_type": "traces", "name": "logging/debug"}
2023-08-22T15:36:59.784-0700	info	exporterhelper/queued_retry.go:423	Exporting failed. Will retry the request after interval.	{"kind": "exporter", "data_type": "traces", "name": "otlphttp", "error": "unexpected EOF", "interval": "7.214643193s"}
```

The collector has sent a successful `otlphttp` payload to a backend and received an HTTP 200
success response, but fails to parse the return body and thus unnecessarily re-queues
the data for retransmission.

The result is both a confusing error message to the user and a duplicate sending
of payload data to the backend. This wastes resources and could potentially cause 
other larger problems for backends.

It is believed that the root cause was introduced when the partial success
behavior was added in 0.81.0. This assumes that a response payload from the backend
always contains protobuf, which is not necessarily true. In this repro case, 
the response is clearly json.

**Steps to reproduce**

See the details in this repro repo: https://github.com/breedx-splk/collector_otlphttp_eof

Successful exports should not log an error and definitely should not retry, regardless of the `content-type` of the response.

**What did you see instead?**

An error and the collector retries, even tho data was exported correctly.

**What version did you use?**

Verified as problematic on 0.81.0, 0.82.0, and 0.83.0

**What config did you use?**

See above repro repo for config.

**Environment**

Tested on MacOS.

**Additional context**

https://github.com/breedx-splk/collector_otlphttp_eof

atoulme:
**Description:**
Fix the handling of the HTTP response to  ignore responses not encoded as protobuf

**Link to tracking Issue:**
Fixes #8263

### DIFF
--- a/exporter/otlphttpexporter/otlp.go
+++ b/exporter/otlphttpexporter/otlp.go
@@ -47,6 +47,8 @@ type baseExporter struct {
 const (
 	headerRetryAfter         = "Retry-After"
 	maxHTTPResponseReadBytes = 64 * 1024
+
+	protobufContentType = "application/x-protobuf"
 )
 
 // Create new exporter.
@@ -118,7 +120,7 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
 	if err != nil {
 		return consumererror.NewPermanent(err)
 	}
-	req.Header.Set("Content-Type", "application/x-protobuf")
+	req.Header.Set("Content-Type", protobufContentType)
 	req.Header.Set("User-Agent", e.userAgent)
 
 	resp, err := e.client.Do(req)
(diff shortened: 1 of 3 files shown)


rmfitzpatrick (approved):

atoulme:
@evan-bradley would you please take a look as well?

dmitryax (approved):

breedx-splk (approved):
👍🏻

exporter/otlphttpexporter/otlp.go
```diff
@@ -270,7 +275,10 @@ func tracesPartialSuccessHandler(protoBytes []byte) error {
 	return nil
 }
 
-func metricsPartialSuccessHandler(protoBytes []byte) error {
+func metricsPartialSuccessHandler(protoBytes []byte, contentType string) error {
+	if contentType != protobufContentType {
+		return nil
```
tigrannajaryan, line 280: @atoulme Why ignore the response if it is not Protobuf? It can be OTLP/JSON which is a valid content type and which can be decoded, right?

tigrannajaryan: Nevermind, I see OTLP/JSON is not yet implemented by this exporter.

Issue #8260, opened 2023-08-22
Add option to disable HTTP keep-alives in OTLP/HTTP Exporter


kkujawa-sumo:
**Is your feature request related to a problem? Please describe.**

In [this](https://github.com/SumoLogic/sumologic-kubernetes-collection/tree/main/docs#log-collection) solution used for logs collection in Kubernetes cluster it is visible that Otel collectors for logs collection (otlphttp exporter is used to send logs) keep persistent connection with Otel collectors for metadata enrichment (otlp receiver is used to get logs) what makes that adding more instances of Otel collectors for metadata enrichment does not help in handling higher amount of logs (using hpa does not make sense in this situation).
The problem was observed in production clusters. 

In single node k8s cluster, which was used only for tests, it is visible that only one instance of Otel collectors for metadata enrichment is used despite the fact that there are 3 instances in the cluster.

<img width="1612" alt="Screenshot 2023-08-22 at 10 49 14" src="https://github.com/open-telemetry/opentelemetry-collector/assets/73836361/d9f4ca8e-8170-4c44-a7b2-159d7675be13">

In the same single node cluster when modified version of Otel collector with disabled HTTP keep-alives is used it is visible that all 3 instances of Otelcol collectors for metadata enrichment are working:

<img width="2028" alt="Screenshot 2023-08-21 at 14 06 07" src="https://github.com/open-telemetry/opentelemetry-collector/assets/73836361/3a30a27f-c297-4e1a-8542-ece2ad256240">

**Describe the solution you'd like**
Add option to disable HTTP keep-alives in OTLP/HTTP Exporter (it will require changes in [HTTPClientSettings](https://github.com/open-telemetry/opentelemetry-collector/blob/3dad181148bdda41bcba2a3f655e7bcc73ee2ccf/config/confighttp/confighttp.go#L31))

**Additional context**
I'd like to prepare pull request with changes to add option to disable HTTP keep-alives.

PR #8274, opened 2023-08-23
[confighttp] Add option to disable HTTP keep-alives

kkujawa-sumo:
**Description:**

Add option to disable HTTP keep-alives which helps in better load balancing in k8s clusters when otlphttp exports are used to send logs to otlp receivers

**Link to tracking Issue:**  #8260

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -74,6 +74,14 @@ type HTTPClientSettings struct {
 	// IdleConnTimeout is the maximum amount of time a connection will remain open before closing itself.
 	// There's an already set value, and we want to override it only if an explicit value provided
 	IdleConnTimeout *time.Duration `mapstructure:"idle_conn_timeout"`
+
+	// DisableKeepAlives, if true, disables HTTP keep-alives and will only use the connection to the server
+	// for a single HTTP request.
+	//
+	// WARNING: enabling this option can result in significant overhead establishing a new HTTP(S)
+	// connection for every request. Before enabling this option please consider whether changes
+	// to idle connection settings can achieve your goal.
+	DisableKeepAlives bool `mapstructure:"disable_keep_alives"`
 }
 
 // NewDefaultHTTPClientSettings returns HTTPClientSettings type object with
@@ -123,6 +131,8 @@ func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.
 	if hcs.IdleConnTimeout != nil {
 		transport.IdleConnTimeout = *hcs.IdleConnTimeout
 	}
+
+	transport.DisableKeepAlives = hcs.DisableKeepAlives
 
 	clientTransport := (http.RoundTripper)(transport)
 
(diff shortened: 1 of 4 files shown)


atoulme:
Please add a changelog entry.

config/confighttp/confighttp.go
```diff
@@ -74,6 +74,10 @@ type HTTPClientSettings struct {
 	// IdleConnTimeout is the maximum amount of time a connection will remain open before closing itself.
 	// There's an already set value, and we want to override it only if an explicit value provided
 	IdleConnTimeout *time.Duration `mapstructure:"idle_conn_timeout"`
+
+	// DisableKeepAlives, if true, disables HTTP keep-alives and will only use the connection to the server
+	// for a single HTTP request.
+	DisableKeepAlives *bool `mapstructure:"disable_keep_alives"`
```
atoulme, line 80: Why are you using a boolean pointer? Can a boolean do the job?

kkujawa-sumo: No special reason, boolean is also ok.

kkujawa-sumo:
@bogdandrutu Could you take a look at this pull request?

astencel-sumo (approved):
This makes sense to me, the change seems pretty straightforward too.

atoulme (approved):
LGTM. Please rebase off main to fix the conflict.

jpkrohling (approved):

jpkrohling:
@open-telemetry/collector-maintainers, I believe this is ready to be merged.

config/confighttp/confighttp.go
```diff
@@ -74,6 +74,10 @@ type HTTPClientSettings struct {
 	// IdleConnTimeout is the maximum amount of time a connection will remain open before closing itself.
 	// There's an already set value, and we want to override it only if an explicit value provided
 	IdleConnTimeout *time.Duration `mapstructure:"idle_conn_timeout"`
+
+	// DisableKeepAlives, if true, disables HTTP keep-alives and will only use the connection to the server
+	// for a single HTTP request.
```
Aneurysm9, line 79:
```suggestion
	// for a single HTTP request.
	//
	// WARNING: enabling this option can result in significant overhead establishing a new HTTP(S)
	// connection for every request.  Before enabling this option please consider whether changes
	// to idle connection settings can achieve your goal.
```

kkujawa-sumo: @Aneurysm9 I added this warning

jpkrohling: I wonder if this should be a warning for the user of the collector as well, visible in the collector's logs when this option is enabled.

codeboten: I feel like this warning, if it was visible in the collector logs, would potentially just be noise to someone who's already aware of the configuration they're using and need to have this option disabled.

codeboten (commented):
@Aneurysm9 please take a look, the warning has been added.

codeboten (approved):

Aneurysm9:
> @Aneurysm9 please take a look, the warning has been added.

I would prefer not to add this configuration option at all, but don't feel strongly enough to block this PR, so I'll simply decline to approve it.  

I'm not sure that logging a warning when this option is enabled would be more than noise.  Hopefully someone enabling this option has a better understanding of their environment than we do and understands what they're buying.

PR #8284, opened 2023-08-25
[exporterhelper] Refactor queue initialization

dmitryax:
Make the queue initialization process consistent for both queue types. Instead of different workflows for memory and persistent queues, this change breaks it into two generic steps:
1. Queue factory
2. Start method

This change:
- reduces coupling;
- allows future refactoring of queuedRetrySender;
- allows extracting of the queue package from the exporterhelper;
- makes it possible to have `WithRequestQueue` option for the new exporter helper API.


dmitryax:
Hi @swiatekm-sumo, can you please take a look at this PR?

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -6,49 +6,55 @@
...
+		size:         &atomic.Uint32{},
+		capacity:     uint32(capacity),
+		numConsumers: numConsumers,
 	}
 }
 
 // StartConsumers starts a given number of goroutines consuming items from the queue
 // and passing them into the consumer callback.
-func (q *boundedMemoryQueue) StartConsumers(numWorkers int, callback func(item Request)) {
+func (q *boundedMemoryQueue) Start(_ context.Context, _ component.Host, set QueueSettings) error {
```
bogdandrutu, line 42: Why not passing the callback as well in the ctor vs passing set here?

exporter/exporterhelper/internal/producer_consumer_queue.go
```diff
@@ -19,4 +32,8 @@ type ProducerConsumerQueue interface {
 	// Stop stops all consumers, as well as the length reporter if started,
 	// and releases the items channel. It blocks until all consumers have stopped.
 	Stop()
+	// Capacity returns the capacity of the queue.
+	Capacity() int
+	// IsPersistent returns true if the queue is persistent.
+	IsPersistent() bool
```
bogdandrutu, line 38: Add a TODO to not expose this public if the struct moves to a public package.

exporter/exporterhelper/queued_retry.go
```diff
@@ -193,30 +134,35 @@ func (qrs *queuedRetrySender) onTemporaryFailure(logger *zap.Logger, req interna
...
-	qrs.queue.StartConsumers(qrs.queueSettings.config.NumConsumers, func(item internal.Request) {
-		_ = qrs.consumerSender.send(item)
-		item.OnProcessingFinished()
+	err := qrs.queue.Start(ctx, host, internal.QueueSettings{
+		CreateSettings: set,
+		DataType:       qrs.signal,
+		Callback: func(item internal.Request) {
+			_ = qrs.consumerSender.send(item)
+			item.OnProcessingFinished()
+		},
```
bogdandrutu, line 148: Same as previous comment, better to pass all of these in the ctor if possible or at least pass only the callback.

bogdandrutu (approved):

exporter/exporterhelper/queued_retry.go
dmitryax: The idea is to expose the constructors to the user, as you suggested in https://github.com/open-telemetry/opentelemetry-collector/pull/8248#discussion_r1302106280. I wanted to reuse the same structures in the public and private packages. If we want to move all these arguments that users should not be providing, we will need different public structs encapsulating the internal ones. Also, we would likely need to change the signature of the Option from `type Option func(*baseSettings)` to `type Option func(*baseExporter)`, which should not be considered as a breaking change I believe. Let me try this out

exporter/exporterhelper/queued_retry.go
```diff
@@ -225,7 +171,7 @@ func (qrs *queuedRetrySender) start(ctx context.Context, host component.Host) er
 // shutdown is invoked during service shutdown.
 func (qrs *queuedRetrySender) shutdown() {
 	// Cleanup queue metrics reporting
-	if qrs.queueSettings.config.Enabled {
```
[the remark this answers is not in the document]
dmitryax: done

exporter/exporterhelper/internal/producer_consumer_queue.go
dmitryax: added

exporter/exporterhelper/queued_retry.go
dmitryax: This involves too many additional changes. I would like to merge this one and try it in a separate PR

Issue #5059, opened 2022-03-22
`[receiver/scraperhelper]`: Allow Scraper errors to be logged at `DEBUG` level


braydonk:
**Is your feature request related to a problem? Please describe.**
Scrapers do not currently have any control over what level their errors are logged at. Sometimes, for example in the [hostmetrics process scraper](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver/internal/scraper/processscraper), the scraper will encounter errors that ideally should not be logged at `ERROR` level. Currently the solution was to completely suppress an error that was particularly problematic, however nearly all errors can be encountered in cases where it would be great to keep them logged at a `DEBUG` level, rather than being limited to completely muting them. 

**Describe the solution you'd like**
(Note: Names are demonstrative for this issue and not necessarily final)
Add a new method to the Scraper interface that will check if the scraper’s errors should be logged at DEBUG level called Scraper.LogErrorsAtDebug() and returns a bool. The baseScraper will get a new bool field scraper.logErrorsAtDebug. A new ScraperOption function will be exported from scraperhelper called WithDebugErrorLogs, which will return a ScraperOption that sets baseScraper.logErrorsAtDebug to true. In [scrapercontroller.go:scrapeMetricsAndReport() when scraper errors are received](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/scraperhelper/scrapercontroller.go#L197-L198) a new check will be added using `Scraper.LogErrorsAtDebug()` to check whether Zap should log the error at DEBUG or ERROR level.

See a first draft of the proposed solution here:
https://github.com/braydonk/opentelemetry-collector/tree/ScraperErrorDebug/receiver/scraperhelper

**Describe alternatives you've considered**
* Change the scrapercontroller to log every error at DEBUG 
   * This would be a change for every existing scraper, so a fair bit more revolutionary than the existing one which is an optional feature. There is a good chance that this is the right direction overall though.
* Provide a way to configure a particular log level
    * While a case could maybe be made for WARN, it seems pretty unlikely that errors would ever need to be logged at anything but ERROR or DEBUG
* Specifically fix the problem that inspired this request, and make the hostmetrics process scraper mute all errors, similar to the [first PR submitted to mitigate this problem](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7176)
[...]

PR #8292, opened 2023-08-29
[receiver/scrapercontroller] add ScrapeLevelError

braydonk:
**Description:**
This PR adds a new error type called ScrapeLevelError. This error is optional, does not affect any default behaviour, and can be used by scrapers to produce errors that should be logged as Warning or Debug. The Scraper Controller can handle this error if a scraper produces it and log accordingly through the error rather than unilaterally logging an error like in the alternate case.

**Link to tracking Issue:** 
#5059
#8293

**Testing:** 
`make gotest`
[ ] run a test with a build of `otelcol` and a config that will scrape and produce `ScrapeLevelErrors`

**Documentation:** 
[ ] Add proper godoc comments

### DIFF
--- a/receiver/scrapererror/scrapelevelerror.go
+++ b/receiver/scrapererror/scrapelevelerror.go
@@ -1,0 +1,92 @@
+package scrapererror
+
+import (
+	"errors"
+
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+)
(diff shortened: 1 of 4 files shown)


braydonk:
keep open

braydonk:
keep open

PR #8306, opened 2023-08-29
[chore] Update the release doc

dmitryax:
The following changes are applied:
1. Move the contrib draft PR creation to the first step. This is required to be done before the core release, because some changes may be needed in the core codebase to resolve possible core/contrib integration issues which was the case with the 0.84.0 release.
2. Add a missing `v` prefix for OTEL_RC_VERSION env var in step 8.
3. Rotate the schedule
4. Update 0.55.0 version in the examples to 0.85.0. It was inconsistent with using the latest version for the stable module set
5. Mention CHANGELOG-API in step 7

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -25,18 +25,18 @@ It is possible that a core approver isn't a contrib approver. In that case, the 
 
 ## Releasing opentelemetry-collector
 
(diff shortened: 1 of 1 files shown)


codeboten (approved):
👍🏻 Thanks @dmitryax

PR #8369, opened 2023-09-06
[exporterhelper] Refactor options and baseExporter

dmitryax:
Separate all the parts of the baseExporter into an explicit chain of request senders. It makes it easier to follow the data flow and add additional senders. 

This change also removes the baseSettings, because keeping the settings is not needed anymore. All the options update the internal senders in place.

This change also removes confusing error messages like "Exporting failed. Dropping data. Try enabling sending_queue to survive temporary failures" when the Queue is not even available in the exporter (WithQueue option is not applied) which means users don't `sending_queue` config option. Now, such messages are only shown when the sending_queue (or retry_on_failure) is available, but not enabled by the user.

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -7,6 +7,9 @@ import (
 	"context"
 	"time"
 
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/exporter"
@@ -29,8 +32,33 @@ func NewDefaultTimeoutSettings() TimeoutSettings {
 
 // requestSender is an abstraction of a sender for a request independent of the type of the data (traces, metrics, logs).
 type requestSender interface {
+	start(ctx context.Context, host component.Host, set exporter.CreateSettings) error
+	shutdown()
 	send(req internal.Request) error
+	setNextSender(nextSender requestSender)
 }
+
+type baseRequestSender struct {
+	nextSender requestSender
+}
+
+var _ requestSender = (*baseRequestSender)(nil)
+
+func (b *baseRequestSender) start(context.Context, component.Host, exporter.CreateSettings) error {
+	return nil
+}
+
+func (b *baseRequestSender) shutdown() {}
+
+func (b *baseRequestSender) send(req internal.Request) error {
+	return b.nextSender.send(req)
+}
+
+func (b *baseRequestSender) setNextSender(nextSender requestSender) {
+	b.nextSender = nextSender
+}
+
+type obsrepSenderFactory func(obsrep *obsExporter) requestSender
 
 // baseRequest is a base implementation for the internal.Request.
 type baseRequest struct {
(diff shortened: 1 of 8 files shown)


exporter/exporterhelper/common.go
```diff
@@ -56,99 +89,68 @@ func (req *baseRequest) OnProcessingFinished() {
...
-	for _, op := range options {
-		op(bs)
-	}
-
-	return bs
-}
-
 // Option apply changes to baseSettings.
-type Option func(*baseSettings)
+type Option func(*baseExporter)
```
dmitryax, line 93: Changing this signature should not be considered a breaking change since the argument is of unexported type, so users cannot implement it directly

bogdandrutu: would suggest in a separate PR to change this to an interface with a private func. I think this is kind of the recommended new way of doing options.

exporter/exporterhelper/common.go
```diff
@@ -157,48 +157,106 @@ func WithCapabilities(capabilities consumer.Capabilities) Option {
...
+	// If no error then start the queueSender.
+	return be.queueSender.start(ctx, host, be.set)
+}
+
+func (be *baseExporter) Shutdown(ctx context.Context) error {
+	// First shutdown the retry sender
+	be.retrySender.shutdown()
+
+	// Then shutdown the queue sender
+	be.queueSender.shutdown()
```
bogdandrutu, line 244: Random senders that you start/stop. If the interface is like that, you should call start on all. Also should it call start on the next sender itself and same for stop?

exporter/exporterhelper/common.go
```diff
@@ -213,3 +271,22 @@ func (ts *timeoutSender) send(req internal.Request) error {
 	}
 	return req.Export(ctx)
 }
+
+func createSampledLogger(logger *zap.Logger) *zap.Logger {
```
bogdandrutu, line 275: We need to revisit this, seems very arbitrary to have a sampler logger only for some cases. Maybe just better document as comment of the func why we have this and where should be used, but it is still random for me.

exporter/exporterhelper/common.go
```diff
@@ -29,9 +32,37 @@ func NewDefaultTimeoutSettings() TimeoutSettings {
...
+	return nil
+}
+
+func (b *baseRequestSender) shutdown() {}
+
+func (b *baseRequestSender) send(req internal.Request) error {
+	if b.nextSender != nil {
+		return b.nextSender.send(req)
+	}
+	return req.Export(req.Context())
```
bogdandrutu, line 57: Very confusing this. If next is not initialized then exports it... Should we separate this into 2 structs on that always calls next and the other that calls export?

exporter/exporterhelper/common.go
dmitryax: I agree. We have another effort to address this https://github.com/open-telemetry/opentelemetry-collector/pull/8134. I'm just copying it from one file to another here

exporter/exporterhelper/common.go
dmitryax: The shutdown sequence is a bit random for a reason. Ideally, it could've been the same sequence as in the pipelines, the same as the data flow: queueSender -> retrySender -> user's exporter. But we need to stop the `retrySender` first so it can push an unfinished request back to the queue.

The start sequence should be applied in reverse order.

exporter/exporterhelper/common.go
dmitryax: If there is no next sender, it means it's the latest in the chain. This line is actually never called. The last sender `timeoutSender` explicitly calls `req.Export`. I left it here just to avoid potential panics in case the last sender is ever changed. I can remove it to avoid confusion.

bogdandrutu (approved):

exporter/exporterhelper/common.go
antonjim-te: Thank you @dmitryax ,
Right as part of #8134 we want to have a Sampled logger as part of the TelemetrySetting and use it when needed. 

Please @bogdandrutu @dmitryax take a look at my PR #8134. Thanks :)

PR #8375, opened 2023-09-07
add debug exporter

codeboten:
This PR does the following:

* creates a new debugexporter that instantiates the loggingexporter initially
* adds a method in the logging exporter to make it possible for the debug exporter to instantiate it w/ a new name
* adds a deprecation notice to loggingexporter
* updates documentation + examples to use `debug` instead of `logging`

I would suggest to move forward w/ this, we could release v0.85.0 w/ both exporters.

FIxes https://github.com/open-telemetry/opentelemetry-collector/issues/7769


codeboten:
Pinging @jpkrohling as you'd reviewed the PR this replaces

exporter/debugexporter/factory.go
```diff
@@ -0,0 +1,21 @@
...
+const (
+	// The value of "type" key in configuration.
+	typeStr                   = "debug"
+	defaultSamplingInitial    = 2
+	defaultSamplingThereafter = 500
+)
+
+// NewFactory creates a factory for Logging exporter
+func NewFactory() exporter.Factory {
+	return loggingexporter.NewFactoryWithName(typeStr)
```
dmitryax, line 20: I think it's better to move all the code to the new exporter and make this call from the old one. WDYT? This PR can be kept as is to reduce the change set though.

codeboten: yeah I could go either way. I figured with this change, we could leave the code as is until its time to remove the logging exporter. at that point all the code could be moved. i just wanted to prevent having two copies of the same code.

One nice thing is with this change, once loggingexporter goes away, the NewFactoryWithName goes with it.

dmitryax: > One nice thing is with this change, once loggingexporter goes away, the NewFactoryWithName goes with it.

I see. Ok, keeping it there is probably a better choice than 👍

exporter/debugexporter/factory.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/loggingexporter"
+)
+
+const (
+	// The value of "type" key in configuration.
+	typeStr                   = "debug"
+	defaultSamplingInitial    = 2
+	defaultSamplingThereafter = 500
```
dmitryax, line 15: These two consts aren't used, are they?

codeboten: Good point, removed 👍🏻

jpkrohling (approved):
LGTM. A minor comment on how to migrate from one component to another, but it can be done in a follow-up PR.

exporter/loggingexporter/README.md
```diff
@@ -2,10 +2,12 @@
 
 | Status                   |                       |
 | ------------------------ |-----------------------|
-| Stability                | [Development]         |
+| Stability                | [Deprecated]         |
 | Supported pipeline types | traces, metrics, logs |
 | Distributions            | [core], [contrib]     |
 
+This exporter is being deprecated in favour of the [debug exporter]. It will be removed in September 2024.
```
codeboten, line 9:
```suggestion
This exporter is being deprecated in favour of the [debug exporter]. The debug exporter has backward compatibility with the logging exporter. To migrate, replace `logging` in the `exporters` section of your Collector's config with `debug`. The logging exporter will be removed in September 2024.
```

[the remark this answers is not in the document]
codeboten: Added this to the readme 👍🏻

exporter/loggingexporter/factory.go
```diff
@@ -32,6 +32,16 @@ func NewFactory() exporter.Factory {
...
+		exporter.WithTraces(createTracesExporter, component.StabilityLevelDeprecated),
+		exporter.WithMetrics(createMetricsExporter, component.StabilityLevelDeprecated),
+		exporter.WithLogs(createLogsExporter, component.StabilityLevelDeprecated),
+	)
+}
+
+func NewFactoryWithName(s component.Type) exporter.Factory {
+	return exporter.NewFactory(
+		s,
+		createDefaultConfig,
```
dmitryax, line 44: Instead of introducing the exporter `NewFactoryWithName` and keep it for a year, maybe we can move `createDefaultConfig` to `exporter/internal/` package and have both debug and logging exporters use it?

dmitryax:
The loggingexporter module has a public type `Config`, which the debug exporter doesn't have yet. This made me think that moving all the code to the new module is still worth doing after this PR is released. I believe it's better to have all the code in the new module to avoid making changes to the deprecated one.

exporter/loggingexporter/factory.go
codeboten: Let me try that. I'll move this PR back to draft while I try this out.

codeboten:
Superseeded by https://github.com/open-telemetry/opentelemetry-collector/pull/8378

PR #8378, opened 2023-09-08
add debug exporter, move code from logging exporter into internal

codeboten:
Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/8375

This moves most of the code from the logging exporter into two internal packages that will eventually end up in the `debugexporter` once the `loggingexporter` has been removed. 

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/7769


codeboten:
@dmitryax @jpkrohling please have a look at this PR as an alternative to "aliasing" the logging exporter.

Pros:
- not causing the debug exporter to depend on the logging exporter
- no longer adds a new public method to the logging exporter.

Cons:
- code that was internal to the logging exporter is now in an `exporter/internal` package
- there is some amount of duplication in the code

exporter/debugexporter/config.go
```diff
@@ -0,0 +1,44 @@
...
+	// configtelemetry.LevelNone and other future values are not supported.
+	supportedLevels map[configtelemetry.Level]struct{} = map[configtelemetry.Level]struct{}{
+		configtelemetry.LevelBasic:    {},
+		configtelemetry.LevelNormal:   {},
+		configtelemetry.LevelDetailed: {},
+	}
+)
+
+// Config defines configuration for logging exporter.
+type Config struct {
```
dmitryax, line 24: You didn't move the config to the internal to avoid adding the deprecated `logLevel` field?

exporter/debugexporter/factory.go
```diff
@@ -0,0 +1,99 @@
...
+	cfg := config.(*Config)
+	exporterLogger := createLogger(cfg, set.TelemetrySettings.Logger)
+	s := newDebugExporter(exporterLogger, cfg.Verbosity)
+	return exporterhelper.NewTracesExporter(ctx, set, cfg,
+		s.pushTraces,
+		exporterhelper.WithCapabilities(consumer.Capabilities{MutatesData: false}),
+		// Disable Timeout/RetryOnFailure and SendingQueue
+		exporterhelper.WithTimeout(exporterhelper.TimeoutSettings{Timeout: 0}),
+		exporterhelper.WithRetry(exporterhelper.RetrySettings{Enabled: false}),
+		exporterhelper.WithQueue(exporterhelper.QueueSettings{Enabled: false}),
```
dmitryax, line 57: No need to add these options if they are not exposed in the config

exporter/debugexporter/debug_exporter.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/config/configtelemetry"
+	"go.opentelemetry.io/collector/exporter/internal/otlptext"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+type debugExporter struct {
```
dmitryax, line 18: Why can't we move this one to internal?

exporter/debugexporter/config.go
codeboten: Right, i figured there's no need to bring over loglevel in the debug exporter

exporter/debugexporter/debug_exporter.go
codeboten: We could, it just didn't seem to really fit, i guess i could create an `exporter/internal/loggingcommonexporter` package or something like it

dmitryax: Yeah, something like that. I'm just thinking of how we can reduce the code duplication. Looks like it's technically possible. If so, I'd merge this PR and do that later rather than https://github.com/open-telemetry/opentelemetry-collector/pull/8375

jpkrohling:
Is this to address this comment?

>  I believe it's better to have all the code in the new module to avoid making changes to the deprecated one.

If so, wouldn't it be sufficient to move the code from logging to debugging, keeping logging as the wrapper to the new one?

codeboten:
>    - not causing the debug exporter to depend on the logging exporter
>    - no longer adds a new public method to the logging exporter.

@jpkrohling the advantages are really not having to create a dependency between the two modules moving forward and avoiding having to add a public method. With the latest change, there's little code duplication. PTAL

jpkrohling:
> the advantages are really not having to create a dependency between the two modules moving forward

Is it a big problem? I mean, you are in the end creating the same dependency through a proxy, which is the common library, no? If you change the debug exporter, we need to either change the common library or fork the code, which is pretty much what we'd have to do if we had the logging exporter wrapping the debug exporter.

> and avoiding having to add a public method

It was caused by the desire to share the createDefaultConfig implementation in the factory, right? It's such a small function. Wouldn't it be worth just having two implementations of it and of the factory? The logging exporter's factory could just use the debug exporter's factory. Wouldn't it work?

exporter/debugexporter/go.mod
```diff
@@ -0,0 +1,83 @@
...
+
+replace go.opentelemetry.io/collector/extension/zpagesextension => ../../extension/zpagesextension
+
+replace go.opentelemetry.io/collector/extension => ../../extension
+
+replace go.opentelemetry.io/collector/connector => ../../connector
+
+replace go.opentelemetry.io/collector/config/confignet => ../../config/confignet
+
+replace go.opentelemetry.io/collector/exporter/loggingexporter => ../loggingexporter
```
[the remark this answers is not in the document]
codeboten: Correct. I thought crosslink added it, but i just removed it and it didn't complain

djaglowski (approved):

cmd/builder/README.md
```diff
@@ -14,7 +14,7 @@ dist:
   output_path: /tmp/dist
 exporters:
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.69.0
-  - gomod: go.opentelemetry.io/collector/exporter/loggingexporter v0.69.1
+  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.69.1
```
mx-psi, line 17: We should probably bump the version here as well, otherwise the example won't work (there is no debug exporter at v0.69.0)

mx-psi (approved):

cmd/builder/README.md
codeboten: Good point, i've updated it to show 0.86.0, though folks reading this may run into the problem that it wont work until the next release. I think that's ok, just something to keep in mind.

PR #8349, opened 2023-09-01
[chore] [exporterhelper] Cleanup mock storage test helpers

dmitryax:
Remove redundant test helpers

PR #8379, opened 2023-09-09
[extension/experimental/storage] [chore] Deprecate redundant test helper

dmitryax:
It is not used in this repo after https://github.com/open-telemetry/opentelemetry-collector/pull/8349 and should not be exported given that we have helpers close to the implementation in contrib

### DIFF
--- a/extension/experimental/storage/nop_client.go
+++ b/extension/experimental/storage/nop_client.go
@@ -10,6 +10,7 @@ type nopClient struct{}
 var nopClientInstance Client = &nopClient{}
 
 // NewNopClient returns a nop client
+// Deprecated: [0.85.0] Use helpers from github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage/storagetest instead.
 func NewNopClient() Client {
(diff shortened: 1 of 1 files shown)


extension/experimental/storage/nop_client.go
```diff
@@ -10,6 +10,7 @@ type nopClient struct{}
 var nopClientInstance Client = &nopClient{}
 
 // NewNopClient returns a nop client
+// Deprecated: [0.85.0] Use helpers from github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage/storagetest instead.
```
djaglowski, line 13: It's used in non-test code [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/789811ca8c728b7684bc8f80cd44cfe9523c2b2a/pkg/stanza/adapter/storage.go#L16).

Personally I find this useful as a default implementation of the interface. Code that would interact with storage _if_ it was specified doesn't need to be concerned with whether or not it actually was.

Issue #3040, opened 2021-04-28
Add gRPC health check


flands:
**Is your feature request related to a problem? Please describe.**
If you want to deploy the OpenTelemetry Collector in gateway-mode behind an AWS Application Load Balancer (ALB) to support gRPC (e.g. OTLP) then ALB requires a gRPC health check. Today, we have a health check extension, however it only support HTTP.

**Describe the solution you'd like**
Either extend the health check extension or add health check endpoints to gRPC receivers.

**Describe alternatives you've considered**
Switch from gRPC to HTTP for receiving data

rakyll:
Might be related to https://github.com/open-telemetry/opentelemetry-collector/issues/3002.

ligeogeorge:
+1 

Unable to set up ALB->(gRPC)->ECS because there is no support for gRPC health checks as is required by AWS ALB. This would be a common use case, IMO.

#3002 is only talking about enhancing HTTP health checks.

obitech:
You can specify the allowed return codes on the ALB, as a workaround just set them to `0-99` and leave the health check path as default.

g3kr:
@flands @obitech can you share your collector config? I am running into problems using the ALB as the otlp exporter endpoint.  or could you throw some light on #5246 ? Much appreciated. Thank you

jangaraj:
I use this config for ALB for OTLP/GRPC, which works like a charm:
```
 TargetGroupGRPC:
    Type: AWS::ElasticLoadBalancingV2::TargetGroup
    Properties:
      VpcId: !Ref VpcId
      Port: 4317
      Protocol: HTTP
      ProtocolVersion: GRPC
      HealthCheckIntervalSeconds: 5
      HealthCheckPath: /
      HealthCheckProtocol: HTTP
      Matcher:
        GrpcCode: 0-99
      HealthCheckTimeoutSeconds: 4
      HealthyThresholdCount: 2
      UnhealthyThresholdCount: 2
      TargetType: ip
```
Unfortunately, I have a problem with OTLP/HTTP. It works fine with JSON payload, but it has problem with protobuf payloads sometimes (SSL handshake timeout, client headers timeout):
```
 TargetGroupHTTP:
    Type: AWS::ElasticLoadBalancingV2::TargetGroup
    Properties:
      VpcId: !Ref VpcId
      Port: 4318
      Protocol: HTTP
      HealthCheckIntervalSeconds: 5
      HealthCheckPath: /
[...]

PR #8397, opened 2023-09-12
add gRPC health check

seankhliao:
**Description:** added an option to gRPC server for serving the standard health check service grpc.health.v1.Health

**Link to tracking Issue:** #3040

**Testing:** added test to check that the gRPC service was registered

**Documentation:** added entry in readme for `healthcheck` key

### DIFF
--- a/.chloggen/grpc-healthcheck.yaml
+++ b/.chloggen/grpc-healthcheck.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: configgrpc
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: gRPC servers now have the option to register the standard grpc.health.v1.Health healthcheck service.
+
+# One or more tracking issues or pull requests related to the change
+issues: [3040]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/config/configgrpc/README.md
+++ b/config/configgrpc/README.md
@@ -110,4 +110,5 @@ see [confignet README](../confignet/README.md).
 - [`read_buffer_size`](https://godoc.org/google.golang.org/grpc#ReadBufferSize)
 - [`tls`](../configtls/README.md)
 - [`write_buffer_size`](https://godoc.org/google.golang.org/grpc#WriteBufferSize)
+- [`healthcheck`](https://github.com/grpc/grpc/blob/master/doc/health-checking.md): registers the health check service
 - [`auth`](../configauth/README.md)
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -21,6 +21,8 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
+	"google.golang.org/grpc/health"
+	"google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
@@ -142,6 +144,10 @@ type GRPCServerSettings struct {
 
 	// Keepalive anchor for all the settings related to keepalive.
 	Keepalive *KeepaliveServerConfig `mapstructure:"keepalive"`
+
+	// HealthCheck service for the gRPC server. See gRPC health checking.
+	// (https://github.com/grpc/grpc/blob/master/doc/health-checking.md)
+	HealthCheck bool `mapstructure:"healthcheck"`
 
 	// Auth for this receiver
 	Auth *configauth.Authentication `mapstructure:"auth"`
@@ -280,7 +286,15 @@ func (gss *GRPCServerSettings) ToServer(host component.Host, settings component.
 		return nil, err
 	}
 	opts = append(opts, extraOpts...)
-	return grpc.NewServer(opts...), nil
+	grpcServer, err := grpc.NewServer(opts...), nil
+	if err != nil {
+		return nil, err
+	}
+	if gss.HealthCheck {
+		healthServer := health.NewServer()
+		grpc_health_v1.RegisterHealthServer(grpcServer, healthServer)
+	}
+	return grpcServer, nil
 }
 
 func (gss *GRPCServerSettings) toServerOption(host component.Host, settings component.TelemetrySettings) ([]grpc.ServerOption, error) {
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -286,10 +286,7 @@ func (gss *GRPCServerSettings) ToServer(host component.Host, settings component.
 		return nil, err
 	}
 	opts = append(opts, extraOpts...)
-	grpcServer, err := grpc.NewServer(opts...), nil
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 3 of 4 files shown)


seankhliao:
I think this should be part of the gRPC server settings rather than an extension to allow load balancers to properly observe if the listener is ready.

dmitryax (approved):

codeboten (commented):
> I think this should be part of the gRPC server settings rather than an extension to allow load balancers to properly observe if the listener is ready.

The change looks good. But I'm curious if the healthcheck extension could be used here instead. With the work to have status reporting per component, the healthcheck extension could be modified to report the health of components and the load balancer could point to that extension's endpoint. Would that address the same use-case?

bogdandrutu:
Also confused about the need to be on the same server with the traffic.

bpalermo:
imho from the load balancer perspective is always better to health check the traffic port directly instead of another that may or may not reflect the state of the underlying service. Specially if we are exposing multiple otel collector ports (e.g. otlp/grpc + otlp/http).

gjshao44:
What is blocking this change? I don't see a way to make AWS ALB health check work without this change

gjshao44:
what is preventing the merge if all checks have passed and changes approved?

SurenNihalani:
How can I help prioritize merge this?

gjshao44:
Suren, i can't setup an AWS ALB for OTEL Collector GRPC endpoint without this feature. If you can merge this, I will greatly appreciate it.We are trying to standardize on OpenTelemetry. I have some other less disirable solutions by bypassing OpenTelemetry, but that will defeat the whole purpose.

atoulme (approved):

gjshao44:
Suren, I may be confused about the merging process, seems we have two reviewers with write access approved, can the PR be merged now? Can you push the button? Thanks!

yurishkuro:
@codeboten is your concern a blocker?

pellepelle3:
I too really need this, can we get this merged.

pellepelle3 (approved):

pellepelle3:
> > I think this should be part of the gRPC server settings rather than an extension to allow load balancers to properly observe if the listener is ready.
> 
> The change looks good. But I'm curious if the healthcheck extension could be used here instead. With the work to have status reporting per component, the healthcheck extension could be modified to report the health of components and the load balancer could point to that extension's endpoint. Would that address the same use-case?

The healthcheck extension as is would not work as in aws load balancers require it to be GRPC response not a http, the same as what is being served, the expectation is that should be on the same port as the traffic. You can override the port but even so, it needs to be a GRPC response.

atoulme:
We also implemented receiver specific health checks - see for example splunk_hec, which exposes a health check for specific HEC clients.

gjshao44:
This issue now has 3 reviewers approved, one with a question, not necessarily a blocking one in my opinion. Is that sufficient to get this issue move forward, @SurenNihalani, are you able to merge now? Hope everyone has a good holiday!

codeboten:
> The healthcheck extension as is would not work as in aws load balancers require it to be GRPC response not a http, the same as what is being served, the expectation is that should be on the same port as the traffic. You can override the port but even so, it needs to be a GRPC response.

Thanks for your response @pellepelle3, I missed the notification of your response. To be clear the requirement here is to support a gRPC specific response, regardless of the port this is on correct?

This was discussed at this week's SIG (Dec-20), my opinion was that the best way to implement this would be to support a gRPC response in the healthcheck extension. This would prevent users from having to check multiple ports for each component they have configured of a collector.

Using the healthcheck extension would also provide a single source of truth, as I would find it confusing if the healthcheck extension was to return one status code and individual components could return different ones. If this is acceptable as a solution, then I would suggest adding this as a feature of the healthcheck extension that @mwear has been re-writing (to leverage component status reporting)

One question that @kentquirk brought up in the SIG meeting was whether there was a need for the port to be the same, in the case of the ALB mentioned in this PR, it is definitely possible to [configure alternative ports](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html#health-check-settings) (this is often used to avoid making healthcheck visible externally for services)

bpalermo:
@codeboten I'm not familiar with the health check extension internals, but I wonder how it can reflect the healthiness of multiple receivers.
Let's say I have otlp/grpc (4317) + otlp/http (4318), and one of them becomes unhealthy. I assume the extension will reporting the service as unhealthy and the load balancer will take this endpoint out of serving, although the other port would still be able to server.
Is my understanding correct?

codeboten:
> @codeboten I'm not familiar with the health check extension internals, but I wonder how it can reflect the healthiness of multiple receivers. Let's say I have otlp/grpc (4317) + otlp/http (4318), and one of them becomes unhealthy. I assume the extension will reporting the service as unhealthy and the load balancer will take this endpoint out of serving, although the other port would still be able to server. Is my understanding correct?

That's a great question. I think there should be a mechanism in the healthcheck extension to differentiate between the health of different pipelines. There could also be an aggregate status for the overall health of the collector, but in more specific use-cases, like the one you mention, it would make sense to have a separate path (or port, but i feel like this would get confusing fast) for each pipeline

gjshao44:
@bpalermo, @codeboten, interesting discussion. I am planning to have a different AWS ALB endpoint to serve http and grpc endpoint for an otel collector. The http endpoint is the only one that is working for me currently . Here is a sample configuration annotations that I lifted from our Thanos grpc endpoint for AWS ALB, and I am hoping the otel collector can deliver the same working mechanism:

  default = {
    "kubernetes.io/ingress.class" : "alb",
    "alb.ingress.kubernetes.io/scheme" : "internal",
    "alb.ingress.kubernetes.io/target-type" : "ip",
    "alb.ingress.kubernetes.io/backend-protocol-version" : "GRPC",
    "alb.ingress.kubernetes.io/healthcheck-path" : "/grpc.health.v1.Health.Check",
    "alb.ingress.kubernetes.io/healthcheck-protocol" : "HTTP",
    "alb.ingress.kubernetes.io/success-codes" : "1-99",
    "alb.ingress.kubernetes.io/backend-protocol" : "HTTP",
    "alb.ingress.kubernetes.io/listen-ports" : "[{\"HTTPS\": 443}]",
  }

gjshao44:
@codeboten, follow up on your write-up about the SIG meeting and adding feature to the healthcheck extension, is there an issue tracking the feature?

mwear:
Hi @gjshao44. Here is the tracking issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26661. I am actively working on the new version of the healthcheck extension that is based on component status reporting. It has an implementation of the grpc.v1.healthservice: https://github.com/grpc/grpc-proto/blob/master/grpc/health/v1/health.proto as an option. I expect to have a PR up sometime next week.

vynu:
Hello @mwear thanks for working on this ... waiting for your commit .. when can we expect the changes to be part of the release ?

mwear:
I have a PR open for the health check extension here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30673. We will also need some version of https://github.com/open-telemetry/opentelemetry-collector/pull/8684 or https://github.com/open-telemetry/opentelemetry-collector/pull/8788 to provide the extension with status events for exporters. I'll revisit those PRs shortly. I'd appreciate any feedback on any of the PRs I've mentioned.

TRAD-Anthony-CKO:
Any updates on this? I see the PR as closed but multiple people are suffering from this issue.

tmegow:
I am attempting to setup a centralized otel collector which serves via an ALB and gRPC. I find that my target group health checks are failing and it seems this topic might be relevant to my goals. 

May I inquire about the status of this PR/issue?

mwear:
I've written an alternative health check extension that supports both HTTP and gRPC health checks. It's in the process of being reviewed. The full version is here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30673. I am slicing it up into smaller PRs to facilitate review. The current chunk is here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33528. Based on this users experience: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30673#issuecomment-2039223996, I think it will work for your use case when the extension becomes available.

yurishkuro:
@mwear why is healthcheck extension a replacement for this PR? Extension will run on a different port, which is very different from what was asked in the original ticket.

mwear:
@yurishkuro the healthcheck extension provides a grpc health server that is based on component status reporting. Component status reporting, while still a work in progress, allows us to derive collector health based on the health of the individual components. As you point out, it does not run on the same port. The need for the health check to be on the same port was debated in this PR and it wasn't clear if it was a hard requirement. I know that a user successfully used the extension on an AWS ALB, which is one of the use cases discussed, see: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30673#issuecomment-2039223996.

PR #8435, opened 2023-09-13
[exporterhelper] Convert internal request interface into a struct

dmitryax:
At this point, we have two interfaces for request: public and internal. There is no need to have the internal interface. Replacing it with a struct makes the logic easier to follow.

This change replaces the internal interface with a struct embedding the public interface. This requires moving the public interface to a separate package which makes the module structure cleaner anyway. 

It also adds a new optional request interface to handle partial failures.


.chloggen/exporter-helper-add-request-onerror.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterhgelper
```
atoulme, line 5:
```suggestion
component: exporterhelper
```

.chloggen/exporter-helper-move-request.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterhgelper
```
atoulme, line 5:
```suggestion
component: exporterhelper
```

.chloggen/exporter-helper-add-request-onerror.yaml
dmitryax: Thanks 👍

.chloggen/exporter-helper-move-request.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterhelper
```
bogdandrutu, line 5: You can avoid this for the moment, by duplicating the interfaces for the moment since that works in go. I think is better to postpone a bit this, since I think with the new Queue interface we can make the current internal depend on the main package not the other way around.

bogdandrutu: If you don't like the duplicate code, using aliases also is an option.

PR #8439, opened 2023-09-13
fix incorrect use of fsnotify in configtls

### LINKED ISSUE #8438: Incorrect use of fsnotify in configtls
**Describe the bug**
When working on #8395, I observed another testing failure:

```
=== RUN   TestErrorRecordedIfFileDeleted
Event: fsnotify.Event{Name:"/var/folders/9l/f2cn0_n52w38y4m77wms60gc0000gn/T/clientCAs.crt1808838314", Op:0x12}    clientcasfilereloader_test.go:65:
        	Error Trace:	/Users/haoqixu/code/personal/wip/otel/opentelemetry-collector/config/configtls/clientcasfilereloader_test.go:65
        	Error:      	Condition never satisfied
        	Test:       	TestErrorRecordedIfFileDeleted
    clientcasfilereloader_test.go:70:
        	Error Trace:	/Users/haoqixu/code/personal/wip/otel/opentelemetry-collector/config/configtls/clientcasfilereloader_test.go:70
        	Error:      	Not equal:
        	            	expected: "test error on reload"
        	            	actual  : "<nil>"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-test error on reload
        	            	+<nil>
        	Test:       	TestErrorRecordedIfFileDeleted
--- FAIL: TestErrorRecordedIfFileDeleted (5.00s)
```

**Steps to reproduce**
Run `go test -v -run TestErrorRecordedIfFileDeleted -race -count 1000 .` under `config/configtls`.

**Environment**
OS: macOS 13.2.1
Compiler(if manually compiled): go1.20.3

**Additional context**

The `Op` field of [`fsnotify.Event`](https://pkg.go.dev/github.com/fsnotify/fsnotify#Event) is a bitmask and some systems may send multiple operations at once. We should use the `Event.Has()` method instead of comparing with `==`.

The failure above was caused by an multiple operations event.

haoqixu:
**Description:** fix #8438

**Link to tracking Issue:**  #8438

### DIFF
--- a/config/configtls/clientcasfilereloader.go
+++ b/config/configtls/clientcasfilereloader.go
@@ -112,7 +112,7 @@ func (r *clientCAsFileReloader) handleWatcherEvents() {
 			// NOTE: k8s configmaps uses symlinks, we need this workaround.
 			// original configmap file is removed.
 			// SEE: https://martensson.io/go-fsnotify-and-kubernetes-configmaps/
-			if event.Op == fsnotify.Remove || event.Op == fsnotify.Chmod {
+			if event.Has(fsnotify.Remove) || event.Has(fsnotify.Chmod) {
 				// remove the watcher since the file is removed
 				if err := r.watcher.Remove(event.Name); err != nil {
 					r.lastReloadError = err
@@ -123,7 +123,7 @@ func (r *clientCAsFileReloader) handleWatcherEvents() {
 				}
 				r.reload()
 			}
-			if event.Op == fsnotify.Write {
+			if event.Has(fsnotify.Write) {
 				r.reload()
 			}
 		}
--- a/.chloggen/fix-incorrect-use-of-fsnotify.yaml
+++ b/.chloggen/fix-incorrect-use-of-fsnotify.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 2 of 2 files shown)


codeboten (commented):
Thanks for the fix! Please add a changelog entry for the bug fix 👍🏻

config/configtls/clientcasfilereloader.go
```diff
@@ -112,7 +112,7 @@ func (r *clientCAsFileReloader) handleWatcherEvents() {
 			// NOTE: k8s configmaps uses symlinks, we need this workaround.
 			// original configmap file is removed.
 			// SEE: https://martensson.io/go-fsnotify-and-kubernetes-configmaps/
-			if event.Op == fsnotify.Remove || event.Op == fsnotify.Chmod {
+			if event.Has(fsnotify.Remove) || event.Has(fsnotify.Chmod) {
```
bogdandrutu, line 115: My confusion here is why we need to monitor for the chmod...

codeboten: Possibly because of this https://github.com/fsnotify/fsnotify/issues/421?

haoqixu: Inotify generates `IN_DELETE_SELF` for the watched file only when its link count is reduces to 0.  Looks like it is used to handle this situation. :thinking:
https://github.com/fsnotify/fsnotify/blob/v1.6.0/backend_inotify.go#L26-L33

codeboten (approved):
Thanks @haoqixu!

PR #8478, opened 2023-09-16
Fix featuregate late initialization

frzifus:
**Description:** 
The factories for the components are created before the CLI flags are evaluated. Therefore, featuregate flags are ignored during the early startup phase.
~This PR simply shifts the time of creation of the factory map after the CLI flag evaluation.~

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/4967

**Testing:**
Short manual testing via cli. Adding the following line in `cobra.Command.RunE`.
```golang
featuregate.GlobalRegistry().VisitAll(func(fg *featuregate.Gate) { fmt.Println(fg.ID(), fg.IsEnabled()) })
```
Changing `+` & `-`:
```bash
./bin/otelcorecol_linux_amd64 --config=config.yaml --feature-gates=+telemetry.disableHighCardinalityMetrics
```

### DIFF
--- a/.chloggen/fix_featuregate_late_initialization.yaml
+++ b/.chloggen/fix_featuregate_late_initialization.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: bug_fix
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: builder
--- a/.chloggen/fix_featuregate_late_initialization.yaml
+++ b/.chloggen/fix_featuregate_late_initialization.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: builder
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix featuregate late initialization
+
+# One or more tracking issues or pull requests related to the change
+issues: [4967]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/cmd/otelcorecol/main.go
+++ b/cmd/otelcorecol/main.go
@@ -11,22 +11,13 @@ import (
 )
 
 func main() {
-	if err := otelcol.NewCommandFeatureGate().Execute(); err != nil {
-		log.Fatalf("failed to validate featuregates: %v", err)
-	}
-
-	factories, err := components()
-	if err != nil {
-		log.Fatalf("failed to build components: %v", err)
-	}
-
 	info := component.BuildInfo{
 		Command:     "otelcorecol",
 		Description: "Local OpenTelemetry Collector binary, testing only.",
 		Version:     "0.88.0-dev",
 	}
 
-	if err := run(otelcol.CollectorSettings{BuildInfo: info, Factories: factories}); err != nil {
+	if err := run(otelcol.CollectorSettings{BuildInfo: info, Factories: components}); err != nil {
 		log.Fatal(err)
 	}
 }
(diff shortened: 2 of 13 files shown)


cmd/otelcorecol/main.go
```diff
@@ -11,8 +11,7 @@ import (
 )
 
 func main() {
-	factories, err := components()
-	if err != nil {
+	if _, err := components(); err != nil {
```
codeboten, line 14: With `components()` called twice, it's possible the second call will swap out a component the second time around if a feature gate is set and end up error'ing differently at line 25 right? Would it be possible to move the initialization of the feature gates up instead of calling this twice?

frzifus: > With `components()` called twice, it's possible the second call will swap out a component the second time around if a feature gate is set and end up error'ing differently at line 25 right?

I did not consider that. But yes, that might be the case. Making the first call somehow useless.

> Would it be possible to move the initialization of the feature gates up instead of calling this twice?

Defiantly, that is what I tried first. Any recommendations how to do that? What I had in mind was creating another root command and parse only the featuregate flags. This only worked, when I ignored the config flag using https://github.com/spf13/cobra/pull/662.

But since I moved the featureflag evaluation out, the original root command started to fail.

otelcol/command.go
```diff
@@ -13,13 +13,20 @@ import (
 )
 
 // NewCommand constructs a new cobra.Command using the given CollectorSettings.
-func NewCommand(set CollectorSettings) *cobra.Command {
+func NewCommand(set CollectorSettings, createFactories func() (Factories, error)) *cobra.Command {
```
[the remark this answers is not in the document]
frzifus: Changed it back :x:

otelcol/command.go
```diff
@@ -12,6 +12,14 @@ import (
 	"go.opentelemetry.io/collector/featuregate"
 )
 
+// NewCommandFeatureGate constructs a new cobra.Command used to parse given FeatureGates.
+func NewCommandFeatureGate() *cobra.Command {
+	flagSet := flags(featuregate.GlobalRegistry())
```
frzifus, line 17: I guess here we could add something like `flagFeatureGate()` instead and add `FParseErrWhitelist: cobra.FParseErrWhitelist{UnknownFlags: false}` to the root command and add a No-op FeatureGate flag to the actual root command? Or simply parse them twice. Wdyt?

frzifus:
Ping @codeboten @jpkrohling

jpkrohling (approved):

bogdandrutu (changes requested):
This seem to me as a big hack, I would prefer the version that I think @dashpole proposed, to provide a func that returns the factories to the settings instead. 

But still not sure what factories need this to justify this change. I understand that it is a problem/limitation but I don't understand why is important.

dashpole:
> But still not sure what factories need this to justify this change. I understand that it is a problem/limitation but I don't understand why is important.

My memory is fuzzy, but from what I recall, I wanted to use a feature gate for a refactor that entirely replaced the factory implementation. The natural place to do that was in NewFactory(), but it took a long time to realize that the feature gate wasn't being respected. The eventual workaround: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9116. I would be satisfied if there was an easy way to detect that case.

It is also definitely possible it isn't worth the effort if no one else has encountered this case.  We could just document this shortcoming, and close the issue if that is our conclusion.

frzifus:
> This seem to me as a big hack, I would prefer the version that I think @dashpole proposed, to provide a func that returns the factories to the settings instead.

@bogdandrutu I had chosen this way, because the first approach to pass a constructor func through changes the signature of the `otelcon.NewCommand`. This means that new collector versions can only be built with the upcoming builder version. (Was mentioned by @codeboten  https://github.com/open-telemetry/opentelemetry-collector/pull/8478#discussion_r1330680598)

> But still not sure what factories need this to justify this change. I understand that it is a problem/limitation but I don't understand why is important.

From my point of view, it is somewhat misleading that the feature gates in the factory do not work. I have run into the same problems as @dashpole. We also wanted to use a feature gate for an entirely replaced factory implementation. (Jaeger Remote Sampling and Jaeger Receiver)

--

I think some kind of tradeoff has to be made. Leaving it at that it does not work and documenting it is already one. It seems like other alternatives are a lot of refactor work, breaking downward compatibility or double parsing the feature gate flags.

frzifus:
@bogdandrutu based on yesterday's conversation, does it match the concept?

I checked the initialization of featuregates and found none inside a factory.

.chloggen/fix_featuregate_late_initialization.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
sfc-gh-bdrutu, line 4: It is also a breaking change for the otelcol.Settings.

sfc-gh-bdrutu (approved):

.chloggen/fix_featuregate_late_initialization.yaml
bogdandrutu, line 4: It is also a breaking change for the otelcol.Settings.

bogdandrutu (approved):

codeboten (approved):

PR #8344, opened 2023-08-31
Add template provider and converter

djaglowski:
### Example

`my_template.yaml`
```yaml
type: my_otlp_template
template: |
  receivers:
    otlp:
      protocols:
        http:
          endpoint: localhost:{{ .port }}
  pipelines:
    traces:
      receivers: [otlp]
    metrics:
      receivers: [otlp]
    logs:
      receivers: [otlp]

```

`config.yaml`
```yaml
receivers:
  template/my_otlp_template:
    port: 4317

exporters:
  logging:
    loglevel: debug

service:
  pipelines:
    logs:
      receivers: [template/my_otlp_template]
      exporters: [logging]
```

`./otelcorecol --config config.yaml --config template:my_template.yaml`

PR #8491, opened 2023-09-19
[confmap] Add Conf.Delete

djaglowski:
Working on https://github.com/open-telemetry/opentelemetry-collector/pull/8344 has convinced me that this is a reasonable function to expose because it would greatly simplify the way we work with the Conf.


djaglowski:
I've reworked my PR to convert to and from a raw map, with all modifications made against the raw map.

PR #8494, opened 2023-09-20
[pdata] Introduce runtime safeguards to catch incorrect pdata mutations

dmitryax:
This change introduces an option to enable runtime assertions to catch unintentional pdata mutations in components that are claimed as non-mutating pdata. Without these assertions, runtime errors may still occur, but thrown by unrelated components, making it very difficult to troubleshoot.

For now, this doesn't change the default behavior. It just introduces a new method on `[Metrics/Traces|Logs].Shared()` that returns pdata marked as shared. The method will be applied to fan-out consumers in the next PR.

Later, if we want to remove the need of `MutatesData` capability, we can introduce another method `[Metrics/Traces|Logs].Exclusive()` which returns a copy of the pdata if it's shared.

This change unblocks the 1.0 release by implementing the original solution proposed by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/issues/6794. Going forward, we may introduce a copy-on-write solution that doesn't require the runtime assertions. That will likely be part of the 2.0 release.

### DIFF
--- a/pdata/internal/cmd/pdatagen/internal/base_slices.go
+++ b/pdata/internal/cmd/pdatagen/internal/base_slices.go
@@ -17,17 +17,19 @@ const sliceTemplate = `// {{ .structName }} logically represents a slice of {{ .
 // Important: zero-initialized instance is not valid for use.
 type {{ .structName }} struct {
 	orig *[]{{ .originElementType }}
+	state *internal.State
 }
 
-func new{{ .structName }}(orig *[]{{ .originElementType }}) {{ .structName }} {
-	return {{ .structName }}{orig}
+func new{{ .structName }}(orig *[]{{ .originElementType }}, state *internal.State) {{ .structName }} {
+	return {{ .structName }}{orig: orig, state: state}
 }
 
 // New{{ .structName }} creates a {{ .structName }} with 0 elements.
 // Can use "EnsureCapacity" to initialize with a given capacity.
 func New{{ .structName }}() {{ .structName }} {
 	orig := []{{ .originElementType }}(nil)
-	return new{{ .structName }}(&orig)
+	state := internal.StateMutable
+	return new{{ .structName }}(&orig, &state)
 }
 
 // Len returns the number of elements in the slice.
@@ -60,6 +62,7 @@ func (es {{ .structName }}) At(i int) {{ .elementName }} {
 //       // Here should set all the values for e.
 //   }
 func (es {{ .structName }}) EnsureCapacity(newCap int) {
+	es.state.AssertMutable()
 	oldCap := cap(*es.orig)
 	if newCap <= oldCap {
 		return
@@ -73,6 +76,7 @@ func (es {{ .structName }}) EnsureCapacity(newCap int) {
 // AppendEmpty will append to the end of the slice an empty {{ .elementName }}.
 // It returns the newly added {{ .elementName }}.
 func (es {{ .structName }}) AppendEmpty() {{ .elementName }} {
+	es.state.AssertMutable()
 	*es.orig = append(*es.orig, {{ .emptyOriginElement }})
 	return es.At(es.Len() - 1)
 }
@@ -80,6 +84,8 @@ func (es {{ .structName }}) AppendEmpty() {{ .elementName }} {
 // MoveAndAppendTo moves all elements from the current slice and appends them to the dest.
 // The current slice will be cleared.
 func (es {{ .structName }}) MoveAndAppendTo(dest {{ .structName }}) {
+	es.state.AssertMutable()
+	dest.state.AssertMutable()
 	if *dest.orig == nil {
 		// We can simply move the entire vector and avoid any allocations.
 		*dest.orig = *es.orig
@@ -92,6 +98,7 @@ func (es {{ .structName }}) MoveAndAppendTo(dest {{ .structName }}) {
 // RemoveIf calls f sequentially for each element present in the slice.
 // If f returns true, the element is removed from the slice.
 func (es {{ .structName }}) RemoveIf(f func({{ .elementName }}) bool) {
+	es.state.AssertMutable()
 	newLen := 0
 	for i := 0; i < len(*es.orig); i++ {
 		if f(es.At(i)) {
@@ -112,6 +119,7 @@ func (es {{ .structName }}) RemoveIf(f func({{ .elementName }}) bool) {
 
 // CopyTo copies all elements from the current slice overriding the destination.
 func (es {{ .structName }}) CopyTo(dest {{ .structName }}) {
+	dest.state.AssertMutable()
 	srcLen := es.Len()
 	destCap := cap(*dest.orig)
 	if srcLen <= destCap {
@@ -119,7 +127,7 @@ func (es {{ .structName }}) CopyTo(dest {{ .structName }}) {
 
 	{{- if eq .type "sliceOfPtrs" }}
 		for i := range *es.orig {
-			new{{ .elementName }}((*es.orig)[i]).CopyTo(new{{ .elementName }}((*dest.orig)[i]))
+			new{{ .elementName }}((*es.orig)[i], es.state).CopyTo(new{{ .elementName }}((*dest.orig)[i], dest.state))
 		}
 		return
 	}
@@ -127,7 +135,7 @@ func (es {{ .structName }}) CopyTo(dest {{ .structName }}) {
 	wrappers := make([]*{{ .originName }}, srcLen)
 	for i := range *es.orig {
 		wrappers[i] = &origs[i]
-		new{{ .elementName }}((*es.orig)[i]).CopyTo(new{{ .elementName }}(wrappers[i]))
+		new{{ .elementName }}((*es.orig)[i], es.state).CopyTo(new{{ .elementName }}(wrappers[i], dest.state))
 	}
 	*dest.orig = wrappers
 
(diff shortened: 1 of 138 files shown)


codeboten (commented):
Thanks for taking this on @dmitryax, have you run benchmarks to compare the results of this change?

pdata/ptrace/traces.go
```diff
@@ -46,5 +46,10 @@ func (ms Traces) SpanCount() int {
 
 // ResourceSpans returns the ResourceSpansSlice associated with this Metrics.
 func (ms Traces) ResourceSpans() ResourceSpansSlice {
-	return newResourceSpansSlice(&ms.getOrig().ResourceSpans)
+	return newResourceSpansSlice(&ms.getOrig().ResourceSpans, internal.GetTracesState(internal.Traces(ms)))
+}
+
+// Shared returns a Traces that shares underlying resources with ms.
+func (ms Traces) Shared() Traces {
```
bogdandrutu, line 53: Few things to consider here:
1. Why return a new Traces vs changing the current Traces? Since you don't do a deep copy even the current Traces is shared.
2. I did propose Shared as a name but I am thinking if ReadOnly is easier for people to understand.
3. Do we need a IsShared()? If we have that we can get rid of the "Capabilities" in the consumers, and the consumers that need to mutate will check if IsShared first thing then will do a Clone if they are "mutable" and object is shared/readonly. Not ideal solution but is better than Capabilities I believe.
4. We should look a bit at the interaction of this concept with the Request.Traces(). That Request.Traces() may have to change to better fit this. Just something to make sure we check before final design.
5. I also want to think if we should also implement a ref-counting (separate of sharable) and optimize our allocs by using pools of allocated pointers. The reason I am mentioning this is just I want you to think if that will still be possible and the two concepts can live separately and be clearly defined independently.

dmitryax: > 3. Do we need a IsShared()? If we have that we can get rid of the "Capabilities" in the consumers, and the consumers that need to mutate will check if IsShared first thing then will do a Clone if they are "mutable" and object is shared/readonly. Not ideal solution but is better than Capabilities I believe.

I don't think we need it. As I mentioned in the issue description, in order to get rid of the capabilities, we can just introduce another method `func (t Traces) Exclusive() Traces` which will return either the same object. if it's already exclusive or a copy if it's shared. That's why I thought it's better to have `func (t Traces) Shared() Traces` for consistency, replying to your (1) comment. If you agree with this, we need another name for that method if we switch to `ReadOnly` (2).

> 4. We should look a bit at the interaction of this concept with the Request.Traces(). That Request.Traces() may have to change to better fit this. Just something to make sure we check before final design.

I think I addressed that in this PR. The state can be set both on `Traces` and `ExportRequest`, so if you convert from one another, it's preserved. Unmarshalling gives a request with the Exclusive state.

> 5. I also want to think if we should also implement a ref-counting (separate of sharable) and optimize our allocs by using pools of allocated pointers. The reason I am mentioning this is just I want you to think if that will still be possible and the two concepts can live separately and be clearly defined independently.

I don't think it'll conflict with this implementation, but can dig deeper. I think this is something that we can do along with the copy-on-write after the 1.0 release. Based on our latest discussion, the pointer-based CoW solution can be a significant change for the clients, so, if it works out, I think it should go to 2.0. That way, we can provide a shim to support both during the transition period. WDYT?

dmitryax: Maybe we still need `IsShared` in case components want to do different things based on that.

dmitryax:
@codeboten the performance should not be affected because the additional field is a part of a struct that stays on the memory stack. I added some benchmarks from my previous work to confirm it:

Before:
```
BenchmarkTracesUsage-10    	    1832	    613221 ns/op	   50631 B/op	    3409 allocs/op
BenchmarkMetricsUsage-10    	     506	   2247812 ns/op	  115778 B/op	    8946 allocs/op
BenchmarkLogsUsage-10    	    5146	    224939 ns/op	   26323 B/op	    1645 allocs/op
```

After:
```
BenchmarkMetricsUsage-10    	     512	   2288719 ns/op	  115777 B/op	    8946 allocs/op
BenchmarkTracesUsage-10    	    1795	    615730 ns/op	   50629 B/op	    3409 allocs/op
BenchmarkLogsUsage-10    	    5469	    217728 ns/op	   26321 B/op	    1645 allocs/op
```

dmitryax:
Updated PR to add coverage for panics. Had to make some semi-related chores in pdatagen that I moved to another PR https://github.com/open-telemetry/opentelemetry-collector/pull/8537

pdata/ptrace/traces.go
bogdandrutu: I kind of don't understand why you need the return value even in the Exclusive case, why not cloning internally the referenced data? The reason I am asking because if I have an "exclusive" pdata and I mark it as "shared" with the current API, I can end up with one object exclusive and one shared that point to the same data:

```golang
td := ptrace.NewTraces()
std := td.Shared()

td.ResourceSpans().AppendEmpty() // works fine
std.ResourceSpans().AppendEmpty() // panics

// Hence if I pass `td` to a pipeline, and `std` to a different one I get UB.
```

dmitryax: This makes sense. I liked `Shared()` because we don't have to introduce a method with a pointer receiver. Linter might be complaining that we have both methods on both value and pointer receivers, but I  think that's okay. I don't believe we can migrate other methods to pointers without affecting end users.

dmitryax: After giving it a second thought, I don't think we are safe from this UB even with this solution. I still can do this:

```golang
td := ptrace.NewTraces()
tds := td
td.MarkShared()

// td is shared, but tds is still exclusive
```

I implemented your suggestion before realizing it. Now, I think `Shared()` is still a cleaner option.

pdata/ptrace/ptraceotlp/response.go
```diff
@@ -8,18 +8,23 @@ import (
...
+	"go.opentelemetry.io/collector/pdata/internal"
 	otlpcollectortrace "go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/trace/v1"
 	"go.opentelemetry.io/collector/pdata/internal/json"
 )
 
 // ExportResponse represents the response for gRPC/HTTP client/server.
 type ExportResponse struct {
-	orig *otlpcollectortrace.ExportTraceServiceResponse
+	orig  *otlpcollectortrace.ExportTraceServiceResponse
+	state internal.State
```
bogdandrutu, line 19: If this is a pointer you can make changes to reflect to copied instances.

dmitryax: Yeah, that's what I thought as well. It'll introduce one allocation but I think it's fine

dmitryax: Migrated all the states to a pointer 👍

pdata/ptrace/traces.go
dmitryax: As discussed in another thread, I migrated the state to a pointer, so the concern above is not valid anymore.

dmitryax: @bogdandrutu speaking about `ReadOnly` vs `Shared`. What do you think of the following API:

```
func (ms [Logs|Traces|Metrics]) MarkAsReadOnly()
func (ms [Logs|Traces|Metrics]) IsReadOnly()
```

This can be enough for now, users will be able to use `CopyTo` to get writable. Or, optionally, we could add the following later:

```
func (ms [Logs|Traces|Metrics]) MarkAsMutable()
```

sfc-gh-bdrutu: Not a fan of the "MarkAs". What about just `ReadOnly` or `MakeReadOnly` since will make more sense `MakeMutable`.

bogdandrutu: Not a fan of the "MarkAs". What about just `ReadOnly` or `MakeReadOnly` since will make more sense `MakeMutable`.

dmitryax: `MakeReadOnly` sounds good to me. I believe it has to involve an action as it's a mutating function. 

Then we should have `IsMutable` and `MakeMutable`? I think `IsMutable` is something that users want to know essentially, so it's better than writing `!logs.IsReadOnly()`.

bogdandrutu: `ToReadOnly` or `MakeReadOnly` both are good for the moment. Let's do the `Is` and `Mutable` in the next PR.

bogdandrutu: Another option, could be to have `MarkReadOnly` and `Clone`? Then `Clone` can return a value.

dmitryax: `MarkReadOnly` is also a good option. Not sure I understand why do we need `Clone`? We used to have it and replaced it with `CopyTo`

sfc-gh-bdrutu: Let's go with `MarkReadOnly` for the moment, and will discuss in the next PR the next change.

bogdandrutu: Let's go with `MarkReadOnly` for the moment, and will discuss in the next PR the next change.

dmitryax: Applied `MarkReadOnly` 👍

pdata/internal/cmd/pdatagen/internal/base_slices.go
```diff
@@ -17,17 +17,25 @@ const sliceTemplate = `// {{ .structName }} logically represents a slice of {{ .
...
 	orig := []{{ .originElementType }}(nil)
-	return new{{ .structName }}(&orig)
+	state := internal.StateMutable
+	return new{{ .structName }}(&orig, &state)
+}
+
+func (es {{ .structName }}) assertMutable() {
+	if *es.state == internal.StateReadOnly {
+		panic("invalid access to shared data")
+	}
```
bogdandrutu, line 38: Instead of making every struct to have this assertMutable it is easier to make assertMutable on the state object so you will call `es.state.assertMutable()`

dmitryax: Good idea. I updated the code

bogdandrutu (approved):

PR #8499, opened 2023-09-20
rename exporterhelper obsreport method

codeboten:
This aligns better with the other new methods in the various helper modules.


dmitryax (approved):
LGTM

PR #8501, opened 2023-09-21
move obsreporttest from obsreport to internal/obsreportconfig

### LINKED ISSUE #8492: obsreport, OR consider if splitting it as part of each component (exporter, extension,...) module.

codeboten:
This will allow the complete removal of obsreport in the next release.

Fixes #8492


obsreport/obsreporttest/deprecated.go
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package obsreporttest // import "go.opentelemetry.io/collector/obsreport/obsreporttest"
+
+import (
+	"go.opentelemetry.io/collector/component"
+	ortest "go.opentelemetry.io/collector/internal/obsreportconfig/obsreporttest"
+)
+
+// Deprecated: [0.85.0] Use go.opentelemetry.io/collector/internal/obsreportconfig/obsreporttest.TestTelemetry instead.
```
dmitryax, line 11: `go.opentelemetry.io/collector/internal/obsreportconfig/obsreporttest ` cannot be used from the outside, it's internal

codeboten: do you have any suggestions where this should live? i dont really want to keep `obsreport` just for a testing subpackage

dmitryax: `TestTelemetry` maybe can go to `componenttest` along with `SetupTelemetry`, but all its methods have to go to `[exporter|receiver|processor]test` modules as functions, similar to `CheckScraperMetrics`, which should go to `receivertest` or new `scrapertest` package I believe

dmitryax: Btw seems like `CheckScraperMetrics` is not being used. It would be a nice to have an issue to whether adopt it or remove it.

codeboten: Hmmm... you're right it doesn't look like it's used outside of the package itself... i'll mark it as deprecated and we can remove it next release after that unless there's any pushback

djaglowski (approved):
LGTM. I agree if there are no known external uses of the deprecated functions we don't need to recommend alternatives.

dmitryax:
> LGTM. I agree if there are no known external uses of the deprecated functions we don't need to recommend alternatives

I was talking about `CheckScraperMetrics` only. Other API is being used in contrib

obsreport/obsreporttest/deprecated.go
codeboten: nm it's used in the scraperhelper. i'll move it to `componenttest` as well

dmitryax: It cannot go to `componenttest` because is depends on `receiver`. Should go to `receivertest` or `scrapertest`

dmitryax (approved):

Issue #8372, opened 2023-09-06
Template provider


djaglowski:
**Is your feature request related to a problem? Please describe.**

Configuration of the collector is a major barrier to entry for users because the process of "developing" a configuration solution often requires detailed knowledge of one or more collector components, a sophisticated understanding of how to interface with an external technology, or just a non-trivial amount of effort working through necessary data manipulations.

**Describe the solution you'd like**

We should provide an abstraction mechanism that allows expert users to abstract away complex "configuration solutions" and provide novice users with a simplified configuration experience.

I propose that expert users should write templated configuration files which are natively recognized by the collector. Novice users may then include a templated solution in their configuration by defining only a simplified set of parameters.

For example, consider the following configuration ([source](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/16db6ce12fbcda090811c0bb878065c8dfc58f98/examples/couchbase/otel-collector-config.yaml)) for scraping and normalizing metrics from couchbase. 

```yaml
receivers:
  prometheus/couchbase:
    config:
      scrape_configs:
        - job_name: 'couchbase'
          scrape_interval: 5s
          static_configs:
            - targets: ['couchbase:8091']
          basic_auth:
            username: 'otelu'
            password: 'otelpassword'
          metric_relabel_configs:
            # Include only a few key metrics
            - source_labels: [ __name__ ]
              regex: "(kv_ops)|\
                (kv_vb_curr_items)|\
                (kv_num_vbuckets)|\
                (kv_ep_cursor_memory_freed_bytes)|\
                (kv_total_memory_used_bytes)|\
                (kv_ep_num_value_ejects)|\
                (kv_ep_mem_high_wat)|\
                (kv_ep_mem_low_wat)|\
                (kv_ep_tmp_oom_errors)|\
                (kv_ep_oom_errors)"
              action: keep

processors:
  filter/couchbase:
    # Filter out prometheus scraping meta-metrics.
    metrics:
      exclude:
        match_type: strict
        metric_names:
          - scrape_samples_post_metric_relabeling
          - scrape_series_added
          - scrape_duration_seconds
          - scrape_samples_scraped
          - up

  metricstransform/couchbase:
    transforms:
      # Rename from prometheus metric name to OTel metric name.
      # We cannot do this with metric_relabel_configs, as the prometheus receiver does not
      # allow metric renames at this time.
      - include: kv_ops
        match_type: strict
        action: update
[...]

PR #8504, opened 2023-09-22
Add template provider

djaglowski:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/8372

This PR defines a new `template` provider. It acts very similarly to the `file` provider but expects a "template" file as defined in #8372. It is not yet useful without an associated converter which will be needed to expand the template.

Subset of #8344


confmap/provider/templateprovider/provider.go
```diff
@@ -0,0 +1,109 @@
...
+	// }
+	// This allows multiple templates to be aggreated under the same global key.
+	allTemplatesKey = "templates"
+)
+
+type provider struct{}
+
+// New returns a new confmap.Provider that reads the template from a file.
+//
+// This Provider supports "file" scheme, and can be called with a "uri" that follows:
```
evan-bradley, line 37: Is there any way to take advantage of the other providers to handle reading the file contents? Since templates feel more like another type of config than a config "source" (file, http, s3, etc.), it would be unfortunate if templates could either only be obtained from select sources or that getting them from additional sources would require updates to the provider.

djaglowski: Thanks for taking a look at this @evan-bradley. I worked out an implementation based only on a converter. (See https://github.com/open-telemetry/opentelemetry-collector/pull/8507) 

I actually like this a lot better.
- It has the advantage you mentioned, of allowing any provider to provide templates.
- It removes the awkward coupling between a provider and converter.
- It allows `templates` to appear directly in a config file (though I think typically they would not be used this way.) Still, it is nice to be able to reason about them in a more natural way.

djaglowski:
Closing in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/8507

PR #8507, opened 2023-09-22
Add template converter

djaglowski:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/8372

Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/8504

---

This approach uses only a template _converter_. This means that any provider may provide templates as part of the raw configuration.

In the simplest case, this looks like this, where the config and a template are defined in the same file:

```yaml
receivers:
  template/my_filelog_template:
    my_file: foo.log
processors:
  batch:
exporters:
  otlp:
    endpoint: localhost:4317

service:
  pipelines:
    logs:
      receivers: [template/my_filelog_template]
      processors: [batch]
      exporters: [otlp]

templates:
  receivers:
    my_filelog_template: |
      receivers:
        filelog:
          include: {{ .my_file }}
```

Note that this it is not really the intention of templates. Typically, the `templates` section would be provided outside of the main config.

```yaml
# config.yaml
receivers:
  template/my_filelog_template:
    my_file: foo.log
processors:
  batch:
exporters:
  otlp:
    endpoint: localhost:4317

service:
  pipelines:
    logs:
      receivers: [template/my_filelog_template]
      processors: [batch]
      exporters: [otlp]
```

```yaml
# my_template.yaml
templates:
  receivers:
    my_filelog_template: |
      receivers:
        filelog:
          include: {{ .my_file }}
```

Additional notes:
- A template would typically be more complex, such that the user is saved some difficulty - this example only shows how the template is defined relative to its usage. Move complex examples are covered in the tests.
- The `templates` section is deleted by the converter after expanding any templates, similar to how environment variable expansion deletes the original reference to the env var. So although it appears to be part of the config, it is should be considered a precursor to the actual collector configuration.

### DIFF
--- a/confmap/converter/templateconverter/config.go
+++ b/confmap/converter/templateconverter/config.go
@@ -1,0 +1,121 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package templateconverter // import "go.opentelemetry.io/collector/confmap/converter/templateconverter"
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+	"text/template"
+
+	"gopkg.in/yaml.v3"
+)
+
+type instanceID struct {
+	tmplType string
+	instName string
+}
+
+func newInstanceID(id string) (*instanceID, error) {
+	parts := strings.SplitN(id, "/", 3)
+	switch len(parts) {
+	case 2: // template/type
+		return &instanceID{
+			tmplType: parts[1],
+		}, nil
+	case 3: // template/type/name
+		return &instanceID{
+			tmplType: parts[1],
+			instName: parts[2],
+		}, nil
+	}
+	return nil, fmt.Errorf("'template' must be followed by type")
+}
+
+func (id *instanceID) withPrefix(prefix string) string {
+	if id.instName == "" {
+		return prefix + "/" + id.tmplType
+	}
+	return prefix + "/" + id.tmplType + "/" + id.instName
+}
+
+type templateConfig struct {
+	Receivers  map[string]any             `yaml:"receivers,omitempty"`
+	Processors map[string]any             `yaml:"processors,omitempty"`
+	Pipelines  map[string]partialPipeline `yaml:"pipelines,omitempty"`
+}
+
+type partialPipeline struct {
+	Receivers  []string `yaml:"receivers,omitempty"`
+	Processors []string `yaml:"processors,omitempty"`
+}
+
+func newTemplateConfig(id *instanceID, tmpl *template.Template, parameters any) (*templateConfig, error) {
+	var rendered bytes.Buffer
+	var err error
+	if err = tmpl.Execute(&rendered, parameters); err != nil {
+		return nil, fmt.Errorf("render: %w", err)
+	}
+
+	cfg := new(templateConfig)
+	if err = yaml.Unmarshal(rendered.Bytes(), cfg); err != nil {
+		return nil, fmt.Errorf("malformed: %w", err)
+	}
+
+	if len(cfg.Receivers) == 0 {
+		return nil, errors.New("must have at least one receiver")
+	}
+
+	if len(cfg.Processors) > 0 && len(cfg.Pipelines) == 0 {
+		return nil, errors.New("template containing processors must have at least one pipeline")
+	}
+	cfg.applyScope(id.tmplType, id.instName)
+	return cfg, nil
+}
+
+func (cfg *templateConfig) applyScope(tmplType, instName string) {
+	// Apply a scope to all component IDs in the template.
+	// At a minimum, the this appends the template type, but will
+	// also apply the template instance name if possible.
+	scopeMapKeys(cfg.Receivers, tmplType, instName)
+	scopeMapKeys(cfg.Processors, tmplType, instName)
+	for _, p := range cfg.Pipelines {
+		scopeSliceValues(p.Receivers, tmplType, instName)
+		scopeSliceValues(p.Processors, tmplType, instName)
+	}
+}
+
+// In order to ensure the component IDs in this template are globally unique,
+// appending the template instance name to the ID of each component.
+func scopeMapKeys(cfgs map[string]any, tmplType, instName string) {
+	// To avoid risk of collision, build a list of IDs,
+	// sort them by length, and then append starting with the longest.
+	componentIDs := make([]string, 0, len(cfgs))
+	for componentID := range cfgs {
+		componentIDs = append(componentIDs, componentID)
+	}
+	sort.Slice(componentIDs, func(i, j int) bool {
+		return len(componentIDs[i]) > len(componentIDs[j])
+	})
+	for _, componentID := range componentIDs {
+		cfg := cfgs[componentID]
+		delete(cfgs, componentID)
+		cfgs[scopedID(componentID, tmplType, instName)] = cfg
+	}
+}
+
+func scopeSliceValues(componentIDs []string, tmplType, instName string) {
+	for i, componentID := range componentIDs {
+		componentIDs[i] = scopedID(componentID, tmplType, instName)
+	}
+}
+
+func scopedID(componentID string, templateType, instanceName string) string {
+	if instanceName == "" {
+		return componentID + "/" + templateType
+	}
+	return componentID + "/" + templateType + "/" + instanceName
+}
--- a/confmap/converter/templateconverter/converter.go
+++ b/confmap/converter/templateconverter/converter.go
@@ -50,7 +50,14 @@ func (c *converter) Convert(_ context.Context, conf *confmap.Conf) error {
 	}
 
 	c.cfg = conf.ToStringMap()
-	for templateID, parameters := range c.cfg["receivers"].(map[string]any) {
+	if c.cfg["receivers"] == nil {
+		return nil // invalid, but let the unmarshaler handle it
+	}
+	receiverCfgs, ok := c.cfg["receivers"].(map[string]any)
+	if !ok {
+		return nil // invalid, but let the unmarshaler handle it
+	}
+	for templateID, parameters := range receiverCfgs {
 		if !strings.HasPrefix(templateID, "template") {
 			continue
 		}
--- a/confmap/converter/templateconverter/converter.go
+++ b/confmap/converter/templateconverter/converter.go
@@ -1,0 +1,262 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package templateconverter // import "go.opentelemetry.io/collector/confmap/converter/templateconverter"
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"text/template"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+// Define a template type by adding a "templates" section to the config.
+//
+// templates:
+//
+//	receivers:
+//	  my_template: |
+//	    receivers: ... # required, one or more
+//	    processors: ... # optional
+//	    pipelines: ...  # required if processors are used, otherwise optional
+//
+// Instantiate a templated receiver:
+//
+// receivers:
+//
+//	template/my_template[/name]:
+//	  parameter_one: value_one
+//	  parameter_two: value_two
+type converter struct {
+	cfg               map[string]any
+	receiverTemplates map[string]*template.Template
+}
+
+// New returns a confmap.Converter, that renders all templates and inserts them into the given confmap.Conf.
+func New() confmap.Converter {
+	return &converter{
+		receiverTemplates: make(map[string]*template.Template),
+	}
+}
+
+func (c *converter) Convert(_ context.Context, conf *confmap.Conf) error {
+	if err := c.parseTemplates(conf); err != nil {
+		return err
+	} else if len(c.receiverTemplates) == 0 {
+		return nil
+	}
+
+	c.cfg = conf.ToStringMap()
+	for templateID, parameters := range c.cfg["receivers"].(map[string]any) {
+		if !strings.HasPrefix(templateID, "template") {
+			continue
+		}
+
+		id, err := newInstanceID(templateID)
+		if err != nil {
+			return err
+		}
+
+		tmpl, ok := c.receiverTemplates[id.tmplType]
+		if !ok {
+			return fmt.Errorf("template type %q not found", id.tmplType)
+		}
+
+		cfg, err := newTemplateConfig(id, tmpl, parameters)
+		if err != nil {
+			return err
+		}
+
+		c.expandTemplate(cfg, id)
+	}
+	delete(c.cfg, "templates")
+
+	*conf = *confmap.NewFromStringMap(c.cfg)
+	return nil
+}
+
+func (c *converter) parseTemplates(conf *confmap.Conf) error {
+	if !conf.IsSet("templates") {
+		return nil
+	}
+
+	templatesMap, ok := conf.ToStringMap()["templates"].(map[string]any)
+	if !ok {
+		return fmt.Errorf("'templates' must be a map")
+	}
+	if templatesMap["receivers"] == nil {
+		return fmt.Errorf("'templates' must contain a 'receivers' section")
+	}
+
+	receiverTemplates, ok := templatesMap["receivers"].(map[string]any)
+	if !ok {
+		return fmt.Errorf("'templates::receivers' must be a map")
+	}
+
+	for templateType, templateVal := range receiverTemplates {
+		templateStr, ok := templateVal.(string)
+		if !ok {
+			return fmt.Errorf("'templates::receivers::%s' must be a string", templateType)
+		}
+		parsedTemplate, err := template.New(templateType).Parse(templateStr)
+		if err != nil {
+			return err
+		}
+		c.receiverTemplates[templateType] = parsedTemplate
+	}
+	return nil
+}
+
+func (c *converter) expandTemplate(cfg *templateConfig, id *instanceID) {
+	// Delete the reference to this template instance and
+	// replace it with the rendered receiver(s).
+	delete(c.cfg["receivers"].(map[string]any), id.withPrefix("template"))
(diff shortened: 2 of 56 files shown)


mx-psi (commented):
How does this feature interact with the existing `${...}` syntax? For example, is this supported? If not, what kind of error do we output?

```yaml
receivers:
  template/my_contrived_example:
    my_env_var_name: ENV_VAR_NAME
processors:
  batch:
exporters:
  otlp:
    endpoint: localhost:4317

service:
  pipelines:
    logs:
      receivers: [template/my_filelog_template]
      processors: [batch]
      exporters: [otlp]

templates:
  receivers:
    my_contrived_example: |
      receivers:
        contrived_receiver:
          setting: ${env:{{ .my_env_var_name }}}
```

I am not claiming that this should be supported (sounds like a contrived edge case better addressed through different means), just that we should error out with a clear error if not supported and document the limitations

djaglowski:
> How does this feature interact with the existing `${...}` syntax? For example, is this supported? If not, what kind of error do we output?

I actually think this would potentially be quite useful. In my mind, templates should be expanded first, but from there the content should be a valid config as we would otherwise define it. 

That said, I misunderstood where expansion is handled, thinking it was within the expand converter. It appears we have a sort of hardcoded psuedo-converter block [here](https://github.com/open-telemetry/opentelemetry-collector/blob/b8eb08342e0b18bd5f19b9b2c502797ec0317c5f/confmap/resolver.go#L140-L147), and then immediately call the other converters. I think if we move the expansion code into a converter, we could handle this correctly.

Currently, it looks like we currently get an ugly error like this: `Error: failed to resolve config: cannot resolve the configuration: expanding ${env:{{ .my_env_var_name }, expected convertable to string value type, got %!q(<nil>)(<nil>)`

If we really don't care to support this, it's simple enough to add some clearer context to the error message:

```go
if err != nil {
	if strings.HasPrefix(k, "templates::") {
		return nil, fmt.Errorf("cannot expand value in template %q: %w", k, err)
	}
	return nil, err
}
```

I'll play around with my idea and provide an update as soon as possible.

djaglowski:
Looking into this more, I think this is going to be quite difficult to support.

First of all, I think it's desirable that `${...}` expansion could be used to pull a template into the config file, so I was incorrect to suggest that templates should be rendered first. There are also some expectations that the template converter needs to validate at some point, e.g. that a template must contain a `receivers` map. It should be possible that this map is provided via `${...}` expansion, so again, template expansion cannot necessary happen first.

If template expansion happens after `${...}` expansion, then a template which contains `${...}` syntax will error during rendering. 

Intuitively, I think it might be possible to recursively alternating between the two types of expansions, but I can't quite see how to manage this in a reasonable way. It might have to be error driven. For now, I think it probably just makes sense to prohibit the use of `${...}` syntax within templates and provide a cleaner error. I've pushed a change to apply this additional context to the error.

djaglowski:
@open-telemetry/collector-maintainers, any thoughts on this feature?

confmap/converter/templateconverter/converter.go
```diff
@@ -0,0 +1,262 @@
...
+		c.cfg["processors"] = make(map[string]any, len(cfg.Processors))
+	}
+	for processorID, processorCfg := range cfg.Processors {
+		c.cfg["processors"].(map[string]any)[processorID] = processorCfg
+	}
+
+	// Add a dedicated forward connector for this template instance.
+	// This will consume data from the partial pipelines and emit to
+	// any top-level pipelines that used the template as a receiver.
+	connectorID := id.withPrefix("forward")
```
evan-bradley, line 175: Do we have the ability to validate that the forward connector is available here? I realize it's outside the responsibility of a converter to validate the config this way, but I worry about the case where a user of a custom distribution attempts to use a template and receives an error saying that the forward connector isn't a known component when they never specified it in any of their configs.

Alternatively, could we check if it's available somewhere in `otelcol` if a template is provided?

confmap/converter/templateconverter/config.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, fmt.Errorf("render: %w", err)
+	}
+
+	cfg := new(templateConfig)
+	if err = yaml.Unmarshal(rendered.Bytes(), cfg); err != nil {
+		return nil, fmt.Errorf("malformed: %w", err)
+	}
+
+	if len(cfg.Receivers) == 0 {
+		return nil, errors.New("must have at least one receiver")
```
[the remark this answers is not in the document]
djaglowski: I fully agree with the other use cases as well. I would describe your example (1) as a "processor template", and (2) as an "exporter template", and this PR as supporting "receiver templates".

I think it's really important to look at templates from two distinct points of view:
1. A novice user who benefits from not knowing the details of the template.
2. An experienced user who captures and shares domain knowledge by creating a template.

For the novice user, the purpose of a template is to be an abstraction. In order to do this effectively, I think a template needs to _feel_ as much as possible like a regular component, meaning it has both a class (i.e. receiver, processor, exporter) and type (e.g. filelog, my_template), and its use in the configuration is almost exactly like any other component. If they can think of it as a receiver, then it's much clearer where to define an instance (in the `receivers` section) and where to use it (in a pipeline's `receivers` section).

For the experienced users who create templates, and for us as maintainers, there are other reasons to constrain each template to a particular class. There are several natural differences which I believe would be very difficult to manage otherwise:
1. Requirements for what is optional and required within the template. (A receiver template must contain a receiver because otherwise using the template in place of a receiver would have no effect.)
2. Rules for how a template _instance_ is defined (in the `receivers` section) and how it must be used in the service (in a pipeline's `receivers` section).
3. Logic for how the template is expanded and how it hooks into the global component graph.

confmap/converter/templateconverter/converter.go
djaglowski: This is a good question. I haven't found a good way to do this. Any more specific ideas?

Generally I agree it would be nice to have a better error message but maybe documentation would be sufficient.

confmap/converter/templateconverter/converter.go
```diff
@@ -0,0 +1,262 @@
...
+			return newReceiverIDs[i].(string) < newReceiverIDs[j].(string)
+		})
+		pipelineMap.(map[string]any)["receivers"] = newReceiverIDs
+	}
+}
+
+// Any partial pipelines defined in a template will be expanded into full pipelines.
+// These will all use a forward connector to emit data to the pipelines in which
+// the template was used as a receiver.
+func (c *converter) expandComplex(cfg *templateConfig, id *instanceID) {
```
evan-bradley, line 162: How do we want to handle exporters in receiver templates? Should we explicitly forbid them?

confmap/converter/templateconverter/config.go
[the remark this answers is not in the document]
evan-bradley: That's a really insightful answer, thanks for the details. I really like the idea of ascribing a class to templates, I agree with all your points on how it would simplify both the validation and usage of a template. I'm seeing now that classifying templates is already present in the existing implementation, I've left a few comments/questions around the implementation side of that.

confmap/converter/templateconverter/converter.go
evan-bradley: I think it will have to just be documentation for now. Any code aware of the available components only deals with a resolved config that has already had the `templates` section removed, so there's no reliable way to determine if templates were used.

confmap/converter/templateconverter/converter.go
djaglowski: Yes, I think they should be forbidden because a receiver should be focused on ingesting data. The consumer of the data is defined by the pipeline(s) in which the receiver template is used.

confmap/converter/templateconverter/converter.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		id, err := newInstanceID(templateID)
+		if err != nil {
+			return err
+		}
+
+		tmpl, ok := c.receiverTemplates[id.tmplType]
+		if !ok {
+			return fmt.Errorf("template type %q not found", id.tmplType)
```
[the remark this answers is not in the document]
djaglowski: I think it'll be easy enough to based on the structure I've proposed for defining templates:
```yaml
templates:
  receivers:
    my_receiver_template: ...
```

Since we know the context in which the user is referencing the template, and we know the class of template based on where it was defined, it should be very easy to cross-reference.

That said, I don't think there is anything to do about it until we add support for an additional class of templates.

confmap/converter/templateconverter/converter.go
```diff
@@ -0,0 +1,262 @@
...
+	// Delete the reference to this template instance and
+	// replace it with the rendered receiver(s).
+	delete(c.cfg["receivers"].(map[string]any), id.withPrefix("template"))
+	for receiverID, receiverCfg := range cfg.Receivers {
+		c.cfg["receivers"].(map[string]any)[receiverID] = receiverCfg
+	}
+
+	// Special case where the template only contains receivers. In this case,
+	// we can just substitute the rendered receivers in place of the template ID.
+	if len(cfg.Processors) == 0 && len(cfg.Pipelines) == 0 {
```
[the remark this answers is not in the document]
djaglowski: I don't have a particular design in mind for connector templates, or for using connectors in other classes of templates. My intuition is that it may make sense eventually but I think someone will have to propose a detailed set of rules for how this would work. I'd rather leave this for later unless we think there's a risk of designing ourselves into a corner by not solving it now.

---

Purely as speculation, I think would could eventually allow connectors within receiver templates by introducing a special key (maybe just `forward`) which MUST be used at least once as an exporter and MAY NOT be used as a receiver in the template. Basically, we'd relax the constraints on what can be used within a receiver template and simply require that the user indicate how data flows out of it. (Again, I'd rather leave this for later, but I'm including this to illustrate that it should be possible to extend the framework if necessary.)

```yaml
templates:
  receivers:
    my_receiver_template: |
      receivers:
        jaeger: ...
      connectors:
        spanmetrics: ...
      pipelines:
        traces:
          receivers: [ jaeger ]
          exporters: [ spanmetrics ] # current proposal does not allow exporters but we could relax the rule
        metrics:
          receivers: [ spanmetrics ]
          exporters: [ forward ] # "forward" is automatically defined and must be used at least once as an exporter
```

confmap/converter/templateconverter/converter.go
evan-bradley, line 1: It occurred to me while playing around with this that we have an implicit dependency on `otelcol.Config` here. Everything up to the point where we're running converters on a confmap is only dealing with maps without any assumptions about their structure, but to render the templated components we need to understand the structure of a config. Should we consider adding a new stage of config processing that processes the config after we have a semantic understanding of the config and expand templates there, or is this an acceptable trade-off?

confmap/converter/templateconverter/converter.go
```diff
@@ -0,0 +1,262 @@
...
+
+func (c *converter) Convert(_ context.Context, conf *confmap.Conf) error {
+	if err := c.parseTemplates(conf); err != nil {
+		return err
+	} else if len(c.receiverTemplates) == 0 {
+		return nil
+	}
+
+	c.cfg = conf.ToStringMap()
+	for templateID, parameters := range c.cfg["receivers"].(map[string]any) {
```
evan-bradley, line 53: It's unfortunate, but we will need to validate that this key actually exists in the map since the config hasn't been validated before this converter runs. We will likely need to validate the `service.pipelines` section as well.

confmap/converter/templateconverter/converter.go
evan-bradley: That makes sense to me. I suppose my follow-up suggestion would be that we should validate no exporters have been defined in the template within this function to enforce that.

confmap/converter/templateconverter/converter.go
djaglowski: This is a really interesting idea and I spent some time trying to make it work. Unfortunately, it appears to lead us to a chicken/egg scenario.

It's certainly much easier and more concise to work with the `otelcol.Config` struct directly. However, the problem is that by the time we have this struct, we've already missed our chance to properly unmarshal any components which were contained in templates.

Basically, this is what happens when converting each component from `confmap.Conf` to `component.Config`:
1. Look up the type and make sure there is a factory for it. (I had to add a carve out for `if type == "template"` to make this work.)
2. Use the factory to instantiate a default config.
3. Unmarshal the `confmap.Conf` into the new instance of the default config.

After doing this for each component, we eventually have an `otelcol.Config`, but now if we expand the templates we end up adding new components which were not unmarshaled in the same way.

Technically, we could do steps 1-3 again for only these components but it looks like a substantial overhaul to separate this functionality from the unmarshaler. Or, if we had a way to convert `otelcol.Config` back to `confmap.Conf`, we could just unmarshal the whole thing again. Neither of these seem like good options and I'm not currently seeing any other ideas on how to approach this.

confmap/converter/templateconverter/converter.go
djaglowski: `templateConfig` is a struct which doesn't contain an exporters field, so the user would see a standard message about `exporters` not being a valid field.

confmap/converter/templateconverter/converter.go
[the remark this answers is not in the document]
djaglowski: To me it seems like these are the same errors one would get if they try to define or use any receiver in a `processors` section. e.g.
- `error decoding 'processors': unknown type: "filelog" for id: "filelog/foo"`
- `service::pipelines::logs: references processor "filelog/foo" which is not configured `

In my opinion this isn't really a special case that requires different handling. It really only becomes confusing if we get away from the notion that receiver templates _are_ receivers. If that's happening a lot, then I think we've failed to create an abstraction.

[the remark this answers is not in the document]
evan-bradley: I agree, we should avoid adding special handling here where possible. It does get a bit weird here though since the config unmarshaler doesn't know about templates. As a novice user, it wouldn't be clear to me why `template` isn't a valid value in the full error message:

```
unknown type: "template" for id: "template/my_filelog_template" (valid values: [batch memory_limiter])
```

All that said, I'm remembering this is behind a feature gate, so I'm okay with handling it later.

confmap/converter/templateconverter/converter.go
evan-bradley: You're right, I missed that. I think we're good here, thanks.

confmap/converter/templateconverter/converter.go
evan-bradley: > Technically, we could do steps 1-3 again for only these components but it looks like a substantial overhaul to separate this functionality from the unmarshaler.

That was more or less the approach I had in mind. I agree it would probably require a substantial overhaul, but I thought about it some more after reading your summary and I still think it is possibly worth the effort. I've attempted to draft a solution that would solve all our issues with minimal downsides, but let me know if there's some fatal flaw that I've missed.

I think the problems with implementing templating within our current unmarshaling facilities are that they expect a one-to-one mapping between component configs and factories, and that all component unmarshaling is done within a section of the config instead of in the scope of the whole config.

We would probably still want to take advantage of `configunmarshaler.Configs` to unmarshal individual components, but would want to process the whole config after that to do template expansion. I would see a "config processor" being something similar to a converter that implements the following interface. A list of these would be registered with `otelcol.ConfigProvider` and would be used during config unmarshaling.

```golang
type ConfigProcessor interface {
	// Prefix for configs like "template/"
	Prefix() string

	// Config type for component configs that need
	// processing like `template/my_receiver`
	ComponentConfig() any

	// Key and type for the processor's config (e.g. "templates" and `templatesConfig`)
	ProcessorConfig() (string, any)

	// Handles configs starting with the processor's prefix
	// given the available factories and its config.
	// All instances of the ComponentConfig type should be
	// replaced by component.Config when this function returns.
	Process(pc *PartialConfig, factories Factories, config any) error
}
```

We could then split config unmarshaling into stages:

First, unmarshal the config as much as possible, considering during unmarshaling that there may be some keys that need further processing. So we could unmarshal components, but any config keys prepended with `template/` would need to be handled by a processor. I think this first step would produce a struct like:

```golang
type PartialConfig struct {
	Receivers: map[string]any // Factory or types returned by ConfigProcessor.ComponentConfig()
	[...]
	Service: PartialService // Some type that enables accessing/creating pipelines for the case of templating 
}
```

We would then run the `Process` method of each `ConfigProcessor` on the `PartialConfig`, with the expectation that after all processors have run, we're left with `map[string]component.Config` for each component class. I think the template converter would run in a similar fashion to how it runs now with this: it would remove the component template from the config, then add all components and add them to the pipelines.

I didn't specify a mechanism here, but at least for templates we would also want to provide a common function to take a `confmap.Conf` and return a `component.Config`. Component config unmarshaling would still happen at two different steps, but would at least be done consistently.

I think this design would also provide a straightforward path for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18509, which would probably work similar to templates but be slightly simpler.

I see a few downsides to this route:
1. We currently only have two use-cases in mind, which I'm not sure is enough to say that this design will serve all our needs going forward or will be used enough to justify the additional complexity.
2. Overall I think this will be cleaner and for templating will enable things like checking whether the forward connector is available. However, at least for our current use-cases, this doesn't enable anything that we can't already do with a Converter.
3. We would need to meddle with `service.Config` (or produce it from another struct), for which the whole point is to be directly unmarshaled from a `confmap.Conf`. We'd likely have to refactor something here.

djaglowski: I think what you're proposing probably makes sense. Unfortunately, I'm finding myself too short on time to spearhead this overhaul. If you have a clear idea of what should change and have the time, I can make time in the next few weeks to review PRs. Either way, I will hopefully pick this back up later in November.

jsoriano: Thinking about this `ConfigProcessor` interface, it would not be enough for templates that provide pipelines of receivers and processors, right?

In https://github.com/open-telemetry/opentelemetry-collector/issues/8372, the example includes a template with receivers and processors, and is expected that the processors are expanded as part of the pipeline of the receiver.

Following the approach of the `ConfigProcessor`, `Process` would need to receive the whole config, or some interface to append components beyond the `PartialConfig`.

PR #8525, opened 2023-09-26
[docs/release] Specify platform tier for bugfix releases

mx-psi:
**Description:** 

Since we now have support tiers defined, we can clarify what platforms allow for a bugfix release

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -116,7 +116,7 @@ When considering making a bugfix release on the `v0.N.x` release cycle, the bug 
     - Reverting a feature gate.
     - Changing the configuration to an easy to find value.
 2. The bug happens in common setups. To gauge this, maintainers can consider the following:
-    - The bug is not specific to a platform on [Tier 2 or below](docs/platform-support.md#tiered-platform-support-model).
+    - The bug is not specific to a platform on [Tier 2 or below](../docs/platform-support.md#tiered-platform-support-model).
     - The bug happens with the default configuration or with a commonly used one (e.g. has been reported by multiple people)
 3. The bug is sufficiently severe. For example (non-exhaustive list):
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -116,7 +116,7 @@ When considering making a bugfix release on the `v0.N.x` release cycle, the bug
     - Reverting a feature gate.
     - Changing the configuration to an easy to find value.
 2. The bug happens in common setups. To gauge this, maintainers can consider the following:
-    - The bug is not specific to an uncommon platform
+    - The bug is not specific to a platform on [Tier 2 or below](../docs/platform-support.md#tiered-platform-support-model).
```
djaglowski, line 119: The intent here is to say that we will only do a bug fix for tier 1 platforms, right? Should we just say that instead of the double negative?

mx-psi: The intent is to say that "if the bug is platform-specific, then we will only fix it for Tier 1 platforms". I guess I can word it just like that instead of the double negative

docs/release.md
```diff
@@ -116,7 +116,7 @@ When considering making a bugfix release on the `v0.N.x` release cycle, the bug
     - Reverting a feature gate.
     - Changing the configuration to an easy to find value.
 2. The bug happens in common setups. To gauge this, maintainers can consider the following:
-    - The bug is not specific to an uncommon platform
+    - If the bug is specific to a certain platform, then the platform is [Tier 1](../docs/platform-support.md#tiered-platform-support-model).
```
codeboten, line 119:
```suggestion
    - If the bug is specific to a certain platform, and if that platform is in [Tier 1](../docs/platform-support.md#tiered-platform-support-model).
```

codeboten (approved):

PR #8538, opened 2023-09-26
[obsreport] deprecate test funcs/structs

codeboten:
This deprecates the remaining code under the obsreport package.

Follows https://github.com/open-telemetry/opentelemetry-collector/commit/bf141227c5b3eb95e849067b33a67cf344992985


codeboten:
> @codeboten I think if we move `CheckExporterTraces` (and other methods) to `exportertest` (and others) as functions, we will be able to avoid all the additional dependencies on `componenttest`

All those method rely on the internal otelPromChecker... which is a member of the TestTelemetry anyways, wouldn't that end up with the same dependency? and the exporterhelper (for example) already depends on componenttest... I guess i'm not sure how much more untangling we could do there, short of removing the promChecker (or including it in each component type helper)

component/componenttest/obsreporttest.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package obsreporttest // import "go.opentelemetry.io/collector/obsreport/obsreporttest"
+package componenttest // import "go.opentelemetry.io/collector/component/componenttest"
```
bogdandrutu, line 4: I think we should remove line https://github.com/open-telemetry/opentelemetry-collector/pull/8538/files#diff-40b33886588167c233289a650794771a6d5b9fec56b4af3ab12f1de66d82e42eL40 and change it to a func

dmitryax:
My concern is that we are making `component` depend on `exporter`, `receiver` and `processor` modules. So far, it's been the other way around: `exporter/receiver/processor` -> `component`. It looks like these dependencies are brought by the tests for `CheckExporterTraces` and other functions. That's probably why we don't run into circular dependency failure.

What if we move `otelPromChecker` to another package in `go.opentelemetry.io/collector/internal`? I think that should allow us to move `CheckExporterTraces` and other functions with their tests to `exportertest` and other packages and avoid introducing those additional dependencies.

component/componenttest/obsreporttest.go
codeboten: @bogdandrutu added TelemetrySettingsFunc, the struct member is used in a few different places which I updated to the use the func instead. PTAL

component/go.mod
```diff
@@ -47,3 +67,25 @@ retract (
...
+
+replace go.opentelemetry.io/collector => ../
+
+replace go.opentelemetry.io/collector/receiver => ../receiver
+
+replace go.opentelemetry.io/collector/extension/zpagesextension => ../extension/zpagesextension
+
+replace go.opentelemetry.io/collector/extension => ../extension
+
+replace go.opentelemetry.io/collector/consumer => ../consumer
```
dmitryax, line 91: We don't need the replaces anymore, right?

codeboten: this is the result from running `make crosslink` 🤷🏻

codeboten: re-ran crosslink after rebasing from main, these added lines are gone

connector/go.mod
```diff
@@ -53,16 +72,6 @@ replace go.opentelemetry.io/collector/confmap => ../confmap
...
-
 replace go.opentelemetry.io/collector/featuregate => ../featuregate
 
 replace go.opentelemetry.io/collector/pdata => ../pdata
-
-replace go.opentelemetry.io/collector/processor => ../processor
-
-replace go.opentelemetry.io/collector/receiver => ../receiver
-
-replace go.opentelemetry.io/collector/exporter => ../exporter
```
dmitryax, line 66: This is nice 👍

component/componenttest/obsreporttest.go
```diff
@@ -118,6 +118,12 @@ func (tts *TestTelemetry) CheckReceiverMetrics(protocol string, acceptedMetricPo
 	return tts.prometheusChecker.checkReceiverMetrics(tts.id, protocol, acceptedMetricPoints, droppedMetricPoints)
 }
 
+// CheckScraperMetrics checks that for the current exported values for metrics scraper metrics match given values.
+// When this function is called it is required to also call SetupTelemetry as first thing.
+func (tts *TestTelemetry) CheckScraperMetrics(receiver component.ID, scraper component.ID, scrapedMetricPoints, erroredMetricPoints int64) error {
+	return tts.otelPrometheusChecker.checkScraperMetrics(receiver, scraper, scrapedMetricPoints, erroredMetricPoints)
```
dmitryax, line 124: This should probably help with the failing CI
```suggestion
	return tts.prometheusChecker.checkScraperMetrics(receiver, scraper, scrapedMetricPoints, erroredMetricPoints)
```

component/componenttest/obsreporttest.go
```diff
@@ -38,6 +37,7 @@ const (
 )
 
 type TestTelemetry struct {
+	// Deprecated: [0.93.0] use TelemetrySettingsFunc() instead.
 	component.TelemetrySettings
```
dmitryax, line 41: Given that users need to adapt to the move it anyway, should we hide this field in the new package and expose the function under this name instead of `TelemetrySettingsFunc`?

connector/go.mod
dmitryax: Removal of the deprecated `obsreport/obsereporttest` in the next release should give us a huge reduction go.sum in the diff 🎉

component/componenttest/obsreporttest.go
codeboten: Thx, fixed

connector/go.mod
codeboten: yes! less interdependencies is exciting

component/componenttest/obsreporttest.go
codeboten: Good point, I've updated the code

dmitryax (approved):

Issue #8510, opened 2023-09-23
Why is 0.0.0.0 the default host?


yurishkuro:
I am running OTLP receivers with default settings:
```yaml
receivers:
  otlp:
    protocols:
      grpc:
      http:
```

The start up logs are littered with these warnings:

`internal@v0.85.0/warning.go:40	Using the 0.0.0.0 address exposes this server to every network interface, which may facilitate Denial of Service attacks	{"kind": "receiver", "name": "otlp", "data_type": "traces", "documentation": "https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/security-best-practices.md#safeguards-against-denial-of-service-attacks"}`

Why isn't the default config for those exporters is such that it follows the mentioned best practices?

mx-psi:
Previous discussion can be found on #6151. If I recall correctly ultimately we decided on a SIG meeting around that time not to change the default for backwards compatibility reasons.

mx-psi:
If we were to change it, what default value would you suggest? `localhost`? Another problem here is that there is no good default value that works in all cases. Specifically, the containerized cases usually depend on the network configuration, we are able to provide a good default on the Helm chart but not in other cases.

yurishkuro:
If we believe the default value is unsafe, it should be changed (despite backwards compatibility concerns). If it is safe (and appropriate in some scenarios) it should not log the warning. 

The current situation is a bad user experience.

atoulme:
I believe 127.0.0.1 should be default. There is no good reason to have warnings show up on the default setting.
We should have an info message indicating we're using 127.0.0.1.
For our distros and Docker image, the default config should have a way to override the default by using an environment variable. That should be documented. The env var should be a well known name that we standardize on.

mx-psi:
>  the default config should have a way to override the default by using an environment variable

We don't support this (maybe we should), see #4384 for a related feature request.

mx-psi:
Filed open-telemetry/opentelemetry-collector-releases#408 so we can address the Docker case independently from the general default.

A [quick search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-collector-contrib+0.0.0.0+factory.go&type=code) on contrib reveals at least the following components are affected:

- [awsproxy extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/39d1e5398ff9f20681adb1c46f667b8f61c74143/extension/awsproxy/factory.go#L19)
[...]

PR #8622, opened 2023-10-04
[receiver/otlp, internal, featuregate] Add featuregate to switch to localhost defaults for server-like components

mx-psi:
**Description:** 

- Define `component.UseLocalHostAsDefaultHost` in the `internal/localhostgate` package. 
- Define `featuregate.ErrIsAlreadyRegistered` error, returned by `Register` when a gate is already registered.
- Adds support for the localhost gate on the OTLP receiver.

This PR does not remove the current warning in any way, we can remove this separately.

**Link to tracking Issue:** Updates #8510

**Testing:** Adds unit tests

**Documentation:** Document on OTLP receiver template and add related logging.


mx-psi:
To see the usage of this API on contrib take a look at open-telemetry/opentelemetry-collector-contrib/pull/27439

component/featuregate.go
```diff
@@ -0,0 +1,40 @@
...
+// server-like receivers and extensions such as the OTLP receiver use localhost as the default host for their endpoints.
+var UseLocalHostAsDefaultHostfeatureGate = featuregate.GlobalRegistry().MustRegister(
+	useLocalHostAsDefaultHostID,
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("controls whether server-like receivers and extensions such as the OTLP receiver use localhost as the default host for their endpoints"))
+
+// EndpointForPort gets the endpoint for a given port using localhost or 0.0.0.0 depending on the feature gate.
+func EndpointForPort(port int) string {
+	host := "0.0.0.0"
+	if UseLocalHostAsDefaultHostfeatureGate.IsEnabled() {
```
iblancasa, line 26: Instead of setting this to `localhost`, would it make sense to allow setting whatever value from the feature flag?

iblancasa (approved):

component/featuregate.go
mx-psi: The feature gate is meant to ease migration to the new defaults, but I wouldn't want people relying on it to set the endpoint. For existing users, if they want a different endpoint then they have already modified the YAML, so I don't think it affects them. For new users, we want them to modify the YAML instead of using the gate. So I don't think we should add this capability.

mx-psi:
Adding some people that commented on the original issue as reviewers :) PTAL!

atoulme (approved):

receiver/otlpreceiver/README.md
```diff
@@ -30,7 +30,9 @@ The following settings are configurable:
 
 - `endpoint` (default = 0.0.0.0:4317 for grpc protocol, 0.0.0.0:4318 http protocol):
   host:port to which the receiver is going to receive data. The valid syntax is
-  described at https://github.com/grpc/grpc/blob/master/doc/naming.md.
+  described at https://github.com/grpc/grpc/blob/master/doc/naming.md. The 
+  `component.UseLocalHostAsDefaultHost` feature gate changes these to localhost:4317 and 
+  localhost:4318 respectively.
```
jpkrohling, line 35: Should it say that this will be the default behavior in the future?

receiver/otlpreceiver/otlp.go
```diff
@@ -163,6 +166,9 @@ func (r *otlpReceiver) startProtocolServers(host component.Host) error {
 // Start runs the trace receiver on the gRPC server. Currently
 // it also enables the metrics receiver too.
 func (r *otlpReceiver) Start(_ context.Context, host component.Host) error {
+	onceLogLocalHost.Do(func() {
+		component.LogAboutUseLocalHostAsDefault(r.settings.Logger)
```
jpkrohling, line 170: No big deal, but I think I would log it once per receiver instance instead of once per receiver type. The reason is that the r.settings.Logger has the component name in it, which is specific to the receiver instance.

component/featuregate.go
```diff
@@ -0,0 +1,40 @@
...
+		host = "localhost"
+	}
+	return fmt.Sprintf("%s:%d", host, port)
+}
+
+// LogAboutUseLocalHostAsDefault logs about the upcoming change from 0.0.0.0 to localhost on server-like components.
+func LogAboutUseLocalHostAsDefault(logger *zap.Logger) {
+	if !UseLocalHostAsDefaultHostfeatureGate.IsEnabled() {
+		logger.Info(
```
jpkrohling, line 35: I think this should only log something if the config value is set to empty or to 0.0.0.0 (current default).

receiver/otlpreceiver/otlp.go
jpkrohling: Alternative: if `component.LogAboutUseLocalHostAsDefault` is always logging a message anyway (see my other comment there), then the sync.Once should be at the feature flag level, and not per component?

mx-psi:
@jpkrohling after the last commit we now warn per component and only if the endpoint is not set. The approach is similar to what I did for the `logging` exporter: https://github.com/open-telemetry/opentelemetry-collector/blob/5c225d9c11ad55680f93267a169c835969b9e61f/exporter/loggingexporter/config.go#L41-L42

receiver/otlpreceiver/README.md
mx-psi: Done!

jpkrohling (approved):

bogdandrutu (approved):

component/featuregate.go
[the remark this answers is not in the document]
bogdandrutu: I don't think we intend to have "feature-gate" shared code-wise that much. Definitely a 1.0 package is not the right place, but I don't have a better idea. Can we have 2 different gates for core vs contrib so that will be an internal package?

bogdandrutu (approved):
I just have concern with sharing the gate (code) and then removing the code.

component/featuregate.go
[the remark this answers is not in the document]
mx-psi: I would like to avoid two feature gates if possible, I think it could be confusing. One alternative to move to internal is to try to register on both core and contrib and ignore errors if related to already being registered:

```go
var UseLocalHostAsDefaultHostfeatureGate featuregate.Gate

func init() {
   var err error
   UseLocalHostAsDefaultHostfeatureGate, err = featuregate.GlobalRegistry().Register(...)
    
   if !errors.Is(err, featuregate.ErrAlreadyRegistered) { // new error type, could also just check error message
       panic(err)
   }
}
```

Does this sound okay?

mx-psi:
@bogdandrutu see https://github.com/open-telemetry/opentelemetry-collector/pull/8622#discussion_r1374312483 for an alternative way of doing this

receiver/otlpreceiver/otlp.go
```diff
@@ -44,6 +44,8 @@ type otlpReceiver struct {
 	obsrepHTTP *receiverhelper.ObsReport
 
 	settings *receiver.CreateSettings
+
+  onceLogLocalHost sync.Once
```
bogdandrutu, line 48: strange indentation?

component/featuregate.go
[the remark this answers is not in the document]
bogdandrutu: I like this, the setting the address code will be duplicated but I think that is fine.

component/featuregate.go
```diff
@@ -0,0 +1,40 @@
...
+// UseLocalHostAsDefaultHostfeatureGate is the feature gate that controls whether
+// server-like receivers and extensions such as the OTLP receiver use localhost as the default host for their endpoints.
+var UseLocalHostAsDefaultHostfeatureGate = featuregate.GlobalRegistry().MustRegister(
+	useLocalHostAsDefaultHostID,
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("controls whether server-like receivers and extensions such as the OTLP receiver use localhost as the default host for their endpoints"))
+
+// EndpointForPort gets the endpoint for a given port using localhost or 0.0.0.0 depending on the feature gate.
+func EndpointForPort(port int) string {
+	host := "0.0.0.0"
```
bogdandrutu, line 25: Just to make sure we "encourage" the right before, what about you start with "localhost" and change if the gate is disabled.

bogdandrutu (approved):

receiver/otlpreceiver/otlp.go
mx-psi: Fixed

mx-psi:
## Summary of major changes to this PR 

1. Added `featuregate.ErrIsAlreadyRegistered` to the `featuregate` module to be able to check if a gate is already registered.
2. Moved the gate definition from `component` to a new `internal/localhostgate` package. This package will be duplicated in  contrib. This avoids having this as part of the public API.
3. **Kept** the feature gate in the `component.` namespace. I can't think of a better name, and this one seems good enough.

mx-psi:
@bogdandrutu final look?

internal/localhostgate/featuregate.go
```diff
@@ -0,0 +1,80 @@
...
+// UseLocalHostAsDefaultHostfeatureGate is the feature gate that controls whether
+// server-like receivers and extensions such as the OTLP receiver use localhost as the default host for their endpoints.
+var UseLocalHostAsDefaultHostfeatureGate *featuregate.Gate
+
+// registerOrLoad tries to register the feature gate and loads it if it already exists.
+func registerOrLoad(reg *featuregate.Registry, id string, stage featuregate.Stage, opts ...featuregate.RegisterOption) (*featuregate.Gate, error) {
+	gate, err := reg.Register(id, stage, opts...)
+
+	if err != nil {
+		switch {
```
bogdandrutu, line 30: nit: personally I don't like switch/one cond/default instead of `if`.

receiver/otlpreceiver/otlp.go
```diff
@@ -163,6 +166,11 @@ func (r *otlpReceiver) startProtocolServers(host component.Host) error {
 // Start runs the trace receiver on the gRPC server. Currently
 // it also enables the metrics receiver too.
 func (r *otlpReceiver) Start(_ context.Context, host component.Host) error {
+	r.onceLogLocalHost.Do(func() {
```
bogdandrutu, line 169: What is the purpose of this log? Why not log once per process if the gate is disabled?

mx-psi: Now logging once when the Collector starts

mx-psi:
@bogdandrutu (and others) will merge this by Wednesday EU morning unless somebody objects before then :)

Issue #8632, opened 2023-10-06
[receiver] Create helper for server-like receivers to replace memory_limiter processor


mx-psi:
The `memory_limiter` processor allows end-users to drop telemetry data when a memory limit is reached. It is up to receivers to apply backpressure and handle retrying of errors. 

A large class of receivers spin up one or more servers that listen for telemetry in a particular format, map to pdata and send it along to the next consumer. For protocols like HTTP and gRPC there are standard ways ([429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429), [RESOURCE_EXHAUSTED](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)) to report the memory has been exhausted and apply backpressure to clients.

We should build a server receiver helper that handles this. In particular it should:
- Provide an easy way to build receivers based on one or more servers that wrap handlers to handle memory limits
- Support at least HTTP and gRPC
- Be extensible to add other capabilities to the helper in the future (e.g. we could tackle #4951 generally for all gRPC-based receivers?)
- Be extensible to support fairness across multiple clients reporting to the same receiver endpoint when it comes to throttling them
- Be configurable in the same manner as the memory limiter currently is

Having this helper available and tackling this for scraper helpers may remove the need of a `memory_limiter` processor.

splunkericl:
Hey @dmitryax and @mx-psi ,we are helping out with the http use cases and want to confirm a few things:

the suggestion was to build a receiver helper method to wrap the log consumption with memory limiter. But this means the memory limiting is handled post HTTP connections creation and receiving the data on the server. This still poses problem on too many resources used.

Instead, there should be some kinds of http interceptor(similar to [authInterceptor](https://github.com/open-telemetry/opentelemetry-collector/blob/766c1022d032006e3da55abdaceec01032c7afae/config/confighttp/confighttp.go#L309)) with memory limiter baked in. So users can just plug it in when they spawn the http server. Is my understanding correct?

mx-psi:
An interceptor sounds like a good approach. We should keep the requirements in mind (I wrote a list on the original post, we can discuss them if they are too stringent but they seem like a reasonable list for starting to think about this)

bogdandrutu:
@mx-psi I am not sure if this is the right approach, here are some use-cases that would be good the new design (if any) should cover:
[...]

PR #8633, opened 2023-10-06
[confighttp] add max_concurrent_connections to HTTPServerSettings

timannguyen:
**Description:** 

Adding `max_concurrent_connections` to allow http server to limits http connection. this protects the server from being bombarded by tons of connection in-process and result in OOM.

telemetry is an optional metrics to instrument httpserver with opencensus. 

**Link to tracking Issue:** 

https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27066

**Testing:** 

unit test

**Documentation:** 

updated confighttp README with `max_concurrent_connections` and `telemetry`


timannguyen:
@dmitryax do you happen to know who has capacity to review this?

splunkericl:
adding @atoulme as well. We want to introduce this for splunkhecreceiver but this will modify the httpserversettings for every component that uses it. Do you know who is the right POC to consult about this change?

config/confighttp/metrics.go
```diff
@@ -0,0 +1,82 @@
...
+	defaultMaxHttpConnectionsMetricName = metricPrefix + "http_max_connections"
+	defaultHttpConnectionsMetricName    = metricPrefix + "http_connections"
+)
+
+// HttpServerTelemetryConfig defines the telemetry configuration for HTTPServerSettings
+type HttpServerTelemetryConfig struct {
+	// Enabled is set to enable opencensus metrics
+	Enabled bool `mapstructure:"enabled"`
+	// ExtraAttributes is the extra attributes for metrics
+	ExtraAttributes map[string]string `mapstructure:"extra_attributes"`
```
dmitryax, line 24: Why do you need this parameter? Collector provides a way to add extra resource attributes to the emitted metrics in the `service::telemetry::resource` config.

dmitryax (commented):
Please break this PR down:
1. Limit config option
2. Telemetry (it has to be emitted in both opencensus and otel as all other metrics)

dmitryax:
Let's hold off until discussion https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27066#issuecomment-1769740566 is resolved

timannguyen:
it looks that [mem](https://github.com/open-telemetry/opentelemetry-collector/issues/8632) is what we are expecting. this would overall resolve the underlying issue. i've talked to @splunkericl. we're likely to close this out and use https://github.com/open-telemetry/opentelemetry-collector/issues/8632 when it is avail

PR #8634, opened 2023-10-07
[pdata] Enable the pdata mutation safeguards in the fanout consumers

dmitryax:
This change enables the runtime assertions to catch unintentional pdata mutations in components claiming as non-mutating pdata. Without these assertions, runtime errors may still occur, but thrown by unrelated components, making it very difficult to troubleshoot.

This required introducing extra API to get the pdata mutability state:
- p[metric|trace|log].[Metrics|Traces|Logs].IsReadOnly()

Resolves: https://github.com/open-telemetry/opentelemetry-collector/issues/6794

### DIFF
--- a/internal/fanoutconsumer/logs.go
+++ b/internal/fanoutconsumer/logs.go
@@ -20,36 +20,22 @@ import (
 // NewLogs wraps multiple log consumers in a single one.
 // It fanouts the incoming data to all the consumers, and does smart routing:
 //   - Clones only to the consumer that needs to mutate the data.
-//   - If all consumers needs to mutate the data one will get the original data.
+//   - If all consumers needs to mutate the data one will get the original mutable data.
 func NewLogs(lcs []consumer.Logs) consumer.Logs {
-	if len(lcs) == 1 {
-		// Don't wrap if no need to do it.
-		return lcs[0]
-	}
-	var pass []consumer.Logs
-	var clone []consumer.Logs
-	for i := 0; i < len(lcs)-1; i++ {
-		if !lcs[i].Capabilities().MutatesData {
-			pass = append(pass, lcs[i])
+	lc := &logsConsumer{}
+	for i := 0; i < len(lcs); i++ {
+		if lcs[i].Capabilities().MutatesData {
+			lc.mutable = append(lc.mutable, lcs[i])
 		} else {
-			clone = append(clone, lcs[i])
+			lc.readonly = append(lc.readonly, lcs[i])
 		}
 	}
-	// Give the original data to the last consumer if no other read-only consumer,
-	// otherwise put it in the right bucket. Never share the same data between
-	// a mutating and a non-mutating consumer since the non-mutating consumer may process
-	// data async and the mutating consumer may change the data before that.
-	if len(pass) == 0 || !lcs[len(lcs)-1].Capabilities().MutatesData {
-		pass = append(pass, lcs[len(lcs)-1])
-	} else {
-		clone = append(clone, lcs[len(lcs)-1])
-	}
-	return &logsConsumer{pass: pass, clone: clone}
+	return lc
 }
 
 type logsConsumer struct {
-	pass  []consumer.Logs
-	clone []consumer.Logs
+	mutable  []consumer.Logs
+	readonly []consumer.Logs
 }
 
 func (lsc *logsConsumer) Capabilities() consumer.Capabilities {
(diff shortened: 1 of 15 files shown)


internal/fanoutconsumer/logs.go
```diff
@@ -59,17 +45,30 @@ func (lsc *logsConsumer) Capabilities() consumer.Capabilities {
...
+			ld.CopyTo(clonedLogs)
+			errs = multierr.Append(errs, lc.ConsumeLogs(ctx, clonedLogs))
+		} else {
+			errs = multierr.Append(errs, lc.ConsumeLogs(ctx, ld))
+		}
 	}
-	for _, lc := range lsc.pass {
+
+	// Mark the data as read-only if it was sent to more than one read-only consumer.
+	if len(lsc.readonly) > 1 {
```
bogdandrutu, line 65: and data not already ReadOnly?

internal/fanoutconsumer/logs.go
```diff
@@ -59,17 +45,30 @@ func (lsc *logsConsumer) Capabilities() consumer.Capabilities {
...
+			clonedLogs := plog.NewLogs()
+			ld.CopyTo(clonedLogs)
+			errs = multierr.Append(errs, lc.ConsumeLogs(ctx, clonedLogs))
+		} else {
+			errs = multierr.Append(errs, lc.ConsumeLogs(ctx, ld))
+		}
 	}
-	for _, lc := range lsc.pass {
+
+	// Mark the data as read-only if it was sent to more than one read-only consumer.
```
bogdandrutu, line 64: // Mark the data as read-only if it will be sent to more than one read-only consumer.

internal/fanoutconsumer/metrics.go
```diff
@@ -18,36 +18,22 @@ import (
...
 }
 
 type metricsConsumer struct {
-	pass  []consumer.Metrics
-	clone []consumer.Metrics
+	mutable  []consumer.Metrics
+	readonly []consumer.Metrics
 }
 
 func (msc *metricsConsumer) Capabilities() consumer.Capabilities {
```
bogdandrutu, line 39: This is a bug. Can we fix this first? Since data are mutated with current implementation for example if we have 2 mutable consumers.

internal/fanoutconsumer/logs.go
```diff
@@ -59,17 +45,30 @@ func (lsc *logsConsumer) Capabilities() consumer.Capabilities {
...
-		ld.CopyTo(clonedLogs)
-		errs = multierr.Append(errs, lc.ConsumeLogs(ctx, clonedLogs))
+
+	// Clone the data before sending to mutable consumers.
+	// The only exception is the last consumer which is allowed to mutate the data only if there are no
+	// other non-mutating consumers and the data is mutable. Never share the same data between
+	// a mutating and a non-mutating consumer since the non-mutating consumer may process
+	// data async and the mutating consumer may change the data before that.
+	for i, lc := range lsc.mutable {
+		if i < len(lsc.mutable)-1 || ld.IsReadOnly() || len(lsc.readonly) > 0 {
```
bogdandrutu, line 55: Don't like that we do too many checks and complicated logic on the critical path. Can this be simplified?

internal/fanoutconsumer/metrics.go
dmitryax: Should we consider changing the state a mutation? The original data isn't touch. Anyway, it I don't think it should matter because the Capabilities of the fanout consumer are not being used in the components graph building process if I'm not mistaken.

internal/fanoutconsumer/logs.go
dmitryax: If it's already ReadOnly, `MarkReadOnly` function is a no-op as it updates the underlying data using the pointer. Do you think we should add an additional condition here?

internal/fanoutconsumer/metrics.go
bogdandrutu: We need to check that :)

dmitryax: Sure. We add a capability processor in front of processors [here](https://github.com/open-telemetry/opentelemetry-collector/blob/53615832e65416a89667611481b8765ab244021c/service/internal/graph/graph.go#L291-L300) without looking at the fanout consumer at the end of the processors' chain. When building connector, [we look at capability nodes](https://github.com/open-telemetry/opentelemetry-collector/blob/53615832e65416a89667611481b8765ab244021c/service/internal/graph/nodes.go#L241-L258) exposed on the receiver side and wrap the fanout consumer with capability consumer based on that. The regular receiver use case is straight forward. So I don't see any place where capabilities of the fanout consumer is used.

internal/fanoutconsumer/logs.go
dmitryax: I believe this is the most concise solution but maybe not the easiest to read. I tried expanding this a bit. Appiled on logs only for now. PTAL if that's better

internal/fanoutconsumer/logs.go
bogdandrutu: But I think code is cleaner if you check here.

dmitryax: Added

bogdandrutu (approved):
I like more the logs code.

PR #8674, opened 2023-10-11
[exporterhelper] make enqueue failures available

### LINKED ISSUE #8673: enqueue_failed counters not available in otel metrics
**Describe the bug**
The following metrics are not available via opentelemetry metrics:

```
otelcol_exporter_enqueue_failed_log_records_total{exporter="otlp"} 0
otelcol_exporter_enqueue_failed_metric_points_total{exporter="otlp"} 0
otelcol_exporter_enqueue_failed_spans_total{exporter="otlp"} 0
```

This is because they're registered with a separate registry than the one used by the otel OC bridge.

**Steps to reproduce**
Enabled the following feature gates:
```
--feature-gates=telemetry.useOtelForInternalMetrics
--feature-gates=telemetry.useOtelWithSDKConfigurationForInternalTelemetry
```

Expected to see those metrics emitted via OTLP

**What did you see instead?**
Those metrics did not appear

**What version did you use?**
v0.87.0

codeboten:
These metrics were only exporter either via OC or via the prometheus exporter. Fixes #8673

### DIFF
--- a/exporter/exporterhelper/obsexporter.go
+++ b/exporter/exporterhelper/obsexporter.go
@@ -34,14 +34,17 @@ type ObsReport struct {
 	tracer         trace.Tracer
 	logger         *zap.Logger
 
-	useOtelForMetrics        bool
-	otelAttrs                []attribute.KeyValue
-	sentSpans                metric.Int64Counter
-	failedToSendSpans        metric.Int64Counter
-	sentMetricPoints         metric.Int64Counter
-	failedToSendMetricPoints metric.Int64Counter
-	sentLogRecords           metric.Int64Counter
-	failedToSendLogRecords   metric.Int64Counter
+	useOtelForMetrics           bool
+	otelAttrs                   []attribute.KeyValue
+	sentSpans                   metric.Int64Counter
+	failedToSendSpans           metric.Int64Counter
+	failedToEnqueueSpans        metric.Int64Counter
+	sentMetricPoints            metric.Int64Counter
+	failedToSendMetricPoints    metric.Int64Counter
+	failedToEnqueueMetricPoints metric.Int64Counter
+	sentLogRecords              metric.Int64Counter
+	failedToSendLogRecords      metric.Int64Counter
+	failedToEnqueueLogRecords   metric.Int64Counter
 }
 
 // ObsReportSettings are settings for creating an ObsReport.
@@ -96,6 +99,12 @@ func (or *ObsReport) createOtelMetrics(cfg ObsReportSettings) error {
 		metric.WithUnit("1"))
 	errors = multierr.Append(errors, err)
 
+	or.failedToEnqueueSpans, err = meter.Int64Counter(
+		obsmetrics.ExporterPrefix+obsmetrics.FailedToEnqueueSpansKey,
+		metric.WithDescription("Number of spans failed to be added to the sending queue."),
+		metric.WithUnit("1"))
+	errors = multierr.Append(errors, err)
+
 	or.sentMetricPoints, err = meter.Int64Counter(
 		obsmetrics.ExporterPrefix+obsmetrics.SentMetricPointsKey,
 		metric.WithDescription("Number of metric points successfully sent to destination."),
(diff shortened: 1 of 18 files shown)


jmacd (commented):
🥇 💯 🚀 👍

exporter/exporterhelper/obsexporter.go
```diff
@@ -252,3 +273,43 @@ func toNumItems(numExportedItems int, err error) (int64, int64) {
...
+	}
+	if failed > 0 {
+		_ = stats.RecordWithTags(
+			ctx,
+			or.mutators,
+			failedMeasure.M(failed))
+	}
+}
+
+func (or *ObsReport) recordEnqueueFailureWithOtel(ctx context.Context, dataType component.DataType, failed int64) {
```
moh-osman3, line 303: I'm wondering if this new otel metric can be updated externally through the API? This package allows an external package to call `exporterhelper.ObsReport.StartTracesOp`  and `exporterhelper.ObsReport.EndTracesOp`. And `EndTracesOp` will call `recordMetrics` which will record `sent` and `failedToSend` - i.e. can the API also update `failedToEnqueue`?

moh-osman3 (commented):
Thanks for making this change!

exporter/exporterhelper/obsexporter.go
jmacd: IMO StartTracesOp and EndTracesOp should be able to catch the error from the queue sender, but the individual sub-components are not connected in the right order for this. I do not want to replicate existing functionality, for the record, @moh-osman3, I think we should focus on a new spec: https://github.com/open-telemetry/semantic-conventions/pull/184

codeboten: > will call recordMetrics which will record sent and failedToSend - i.e. can the API also update failedToEnqueue?

I originally considered going the route of recordMetrics updating all 3 metrics, but sent/failedToSend are just opposites of each other, which makes it easy to record together. I suppose recording enqueue failed could be done by checking the error type in `End*Op`. Whether we decide to do this or not (based on @jmacd's comment) I would do this as a follow up PR, as that's a change in behaviour. This PR is really just about ensuring that the metric is emitted in both the OC and the Otel case :)

Would be happy to open an issue to follow up @moh-osman3

moh-osman3: That makes sense thanks @codeboten @jmacd! I was mainly asking because I was writing some code in the otel-arrow repository and realized I wouldn't have access to that metric if I tried to access obsreport externally. But as Josh pointed out we might just do a simpler instrumentation from scratch instead of depending on that obsreport package

obsreport/obsreporttest/otelprometheuschecker.go
```diff
@@ -77,20 +77,32 @@ func (pc *prometheusChecker) checkExporterLogs(exporter component.ID, sent, send
...
-func (pc *prometheusChecker) checkExporterMetrics(exporter component.ID, sentMetricPoints, sendFailedMetricPoints int64) error {
-	return pc.checkExporter(exporter, "metric_points", sentMetricPoints, sendFailedMetricPoints)
+func (pc *prometheusChecker) checkExporterMetrics(exporter component.ID, sent, sendFailed int64) error {
+	return pc.checkExporter(exporter, "metric_points", sent, sendFailed)
 }
 
 func (pc *prometheusChecker) checkExporter(exporter component.ID, datatype string, sent, sendFailed int64) error {
 	exporterAttrs := attributesForExporterMetrics(exporter)
-	errs := pc.checkCounter(fmt.Sprintf("exporter_sent_%s", datatype), sent, exporterAttrs)
+	var errs error
```
codeboten, line 86:
```suggestion
	errs := pc.checkCounter(fmt.Sprintf("exporter_sent_%s", datatype), sent, exporterAttrs)
```

obsreport/obsreporttest/otelprometheuschecker.go
```diff
@@ -77,20 +77,32 @@ func (pc *prometheusChecker) checkExporterLogs(exporter component.ID, sent, send
...
+
+	errs = multierr.Append(errs,
+		pc.checkCounter(fmt.Sprintf("exporter_sent_%s", datatype), sent, exporterAttrs))
+
```
codeboten, line 90:
```suggestion
	errs := pc.checkCounter(fmt.Sprintf("exporter_sent_%s", datatype), sent, exporterAttrs)
```

obsreport/obsreporttest/otelprometheuschecker.go
```diff
@@ -77,20 +77,34 @@ func (pc *prometheusChecker) checkExporterLogs(exporter component.ID, sent, send
...
```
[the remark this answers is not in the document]
codeboten: I've reverted this change

dmitryax (approved):

PR #8684, opened 2023-10-13
Automatic status reporting for exporterhelper and core exporters

mwear:
**Description:**
This PR adds a `WithStatusReporting` option to `exporterhelper` to opt-in to automatic status reporting `Consume*` calls and updates the core exporters to use it.

The behavior for `Consume*` is to report `StatusOK` when it returns without an error and `StatusRecoverableError` when it returns with an error. Users can report a more severe status (e.g. `StatusPermanentError`, `StatusFatalError`) from `Consume` from within the component, if they wish, and it will effectively override the automatic status reporting. 
The OTLP exporters demonstrate this usage by reporting permanent errors for certain statuses in this PR.

A component should report `StatusPermanentError` for error conditions that require human intervention to fix. Typically this is a configuration issue that is discovered at run time. A 401 (Unauthorized) status is an example of such an error. The collector config was structurally valid, but it has become clear at runtime that credentials are missing or invalid. Note: once a component transitions into `StatusPermanentError` it cannot make any further transitions. This is enforced by the finite state machine used by the underlying status reporting system.

The following HTTP status codes are treated as `PermanentError`s for the OTLP/HTTP exporter:

- http.StatusUnauthorized (401)
- http.StatusForbidden (403)
- http.StatusNotFound (404)
- http.StatusMethodNotAllowed (405)
- http.StatusRequestEntityTooLarge (413)
- http.StatusRequestURITooLong (414)
- http.StatusRequestHeaderFieldsTooLarge (431)

The following gRPC codes are treated as `PermanentError`s for the OTLP/gRPC exporter:
- codes.NotFound (5)
- codes.PermissionDenied (7)
- codes.Unauthenticated (16)

The choice of status codes we treat as permanent errors are negotiable.

**Testing:** Units, manual

**Documentation:** Comments

**Edit**
I opened #8788 for comparison and as a manual alternative to this PR. Please compare this with #8788 and weigh in on what approach you prefer, and/or additional suggestions.


mwear:
I'm closing this for the same reason as mentioned here: https://github.com/open-telemetry/opentelemetry-collector/pull/8709#issuecomment-1779817651

mwear:
I closed this PR after our discussion on #8709 last week, but after revisiting the problem of reporting status for exporters, I think this approach is viable and worth further consideration. This is a limited version of #8709 that optionally wraps the `Consume*` and `Start` methods for exporters and will report status based on the returned error. Since an exporter sits at the end of a pipeline, the errors that it emits are related to the status of the exporter itself. By intercepting the errors we are able to reduce the amount of boilerplate code for reporting `StatusOK` and `StatusRecoverableError` and leave the component to handle more severe errors (e.g. `StatusPermanentError`).

PR #8685, opened 2023-10-16
[exporterhelper] Introduce batching functionality

dmitryax:
This change introduces new experimental batching functionality to the exporter helper. The batch sender is fully concurrent and synchronous. It's set after the queue sender, which, if enabled, introduces the asynchronous behavior and ensures no data loss with the permanent queue. 

Follow-up TODO list:
- Add pre-built merge funcs for pdata
- Handle partial errors
- A missing part compared to the batch processor is the ability to shard the batches by context value.

### DIFF
--- a/exporter/exporterbatcher/batch_func.go
+++ b/exporter/exporterbatcher/batch_func.go
@@ -1,0 +1,24 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterbatcher // import "go.opentelemetry.io/collector/exporter/exporterbatcher"
+
+import "context"
+
+// BatchMergeFunc is a function that merges two requests into a single request.
+// Do not mutate the requests passed to the function if error can be returned after mutation or if the exporter is
+// marked as not mutable.
+// Experimental: This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type BatchMergeFunc[T any] func(context.Context, T, T) (T, error)
+
+// BatchMergeSplitFunc is a function that merge and/or splits one or two requests into multiple requests based on the
+// configured limit provided in MaxSizeConfig.
+// All the returned requests MUST have a number of items that does not exceed the maximum number of items.
+// Size of the last returned request MUST be less or equal than the size of any other returned request.
+// The original request MUST not be mutated if error is returned after mutation or if the exporter is
+// marked as not mutable. The length of the returned slice MUST not be 0. The optionalReq argument can be nil,
+// make sure to check it before using.
+// Experimental: This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type BatchMergeSplitFunc[T any] func(ctx context.Context, cfg MaxSizeConfig, optionalReq T, req T) ([]T, error)
--- a/exporter/exporterbatcher/config.go
+++ b/exporter/exporterbatcher/config.go
@@ -1,0 +1,70 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterbatcher // import "go.opentelemetry.io/collector/exporter/exporterbatcher"
+
+import (
+	"errors"
+	"time"
+)
+
+// Config defines a configuration for batching requests based on a timeout and a minimum number of items.
+// MaxSizeItems defines batch splitting functionality if it's more than zero.
+// Experimental: This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Config struct {
+	// Enabled indicates whether to not enqueue batches before sending to the consumerSender.
+	Enabled bool `mapstructure:"enabled"`
+
+	// FlushTimeout sets the time after which a batch will be sent regardless of its size.
+	FlushTimeout time.Duration `mapstructure:"flush_timeout"`
+
+	MinSizeConfig `mapstructure:",squash"`
+	MaxSizeConfig `mapstructure:",squash"`
+}
+
+// MinSizeConfig defines the configuration for the minimum number of items in a batch.
+// Experimental: This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type MinSizeConfig struct {
+	// MinSizeItems is the number of items (spans, data points or log records for OTLP) at which the batch should be
+	// sent regardless of the timeout. There is no guarantee that the batch size always greater than this value.
+	// This option requires the Request to implement RequestItemsCounter interface. Otherwise, it will be ignored.
+	MinSizeItems int `mapstructure:"min_size_items"`
+}
+
+// MaxSizeConfig defines the configuration for the maximum number of items in a batch.
+// Experimental: This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type MaxSizeConfig struct {
+	// MaxSizeItems is the maximum number of the batch items, i.e. spans, data points or log records for OTLP.
+	// If the batch size exceeds this value, it will be broken up into smaller batches if possible.
+	// Setting this value to zero disables the maximum size limit.
+	MaxSizeItems int `mapstructure:"max_size_items"`
+}
+
+func (c Config) Validate() error {
+	if c.MinSizeItems < 0 {
+		return errors.New("min_size_items must be greater than or equal to zero")
+	}
+	if c.MaxSizeItems < 0 {
+		return errors.New("max_size_items must be greater than or equal to zero")
+	}
+	if c.MaxSizeItems != 0 && c.MaxSizeItems < c.MinSizeItems {
+		return errors.New("max_size_items must be greater than or equal to min_size_items")
+	}
+	if c.FlushTimeout <= 0 {
+		return errors.New("timeout must be greater than zero")
+	}
+	return nil
+}
+
+func NewDefaultConfig() Config {
+	return Config{
+		Enabled:      true,
+		FlushTimeout: 200 * time.Millisecond,
+		MinSizeConfig: MinSizeConfig{
+			MinSizeItems: 8192,
+		},
+	}
+}
--- a/exporter/exporterhelper/batch_sender.go
+++ b/exporter/exporterhelper/batch_sender.go
@@ -1,0 +1,218 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterhelper // import "go.opentelemetry.io/collector/exporter/exporterhelper"
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+)
+
+// batchSender is a component that places requests into batches before passing them to the downstream senders.
+// Batches are sent out with any of the following conditions:
+// - batch size reaches cfg.SendBatchSize
+// - cfg.Timeout is elapsed since the timestamp when the previous batch was sent out.
+// - concurrencyLimit is reached.
+type batchSender struct {
+	baseRequestSender
+	cfg            exporterbatcher.Config
+	mergeFunc      exporterbatcher.BatchMergeFunc[Request]
+	mergeSplitFunc exporterbatcher.BatchMergeSplitFunc[Request]
+
+	// concurrencyLimit is the maximum number of goroutines that can be created by the batcher.
+	// If this number is reached and all the goroutines are busy, the batch will be sent right away.
+	// Populated from the number of queue consumers if queue is enabled.
+	concurrencyLimit uint64
+	activeRequests   atomic.Uint64
+
+	resetTimerCh chan struct{}
+
+	mu          sync.Mutex
+	activeBatch *batch
+
+	logger *zap.Logger
+
+	shutdownCh chan struct{}
+	stopped    *atomic.Bool
+}
+
+// newBatchSender returns a new batch consumer component.
+func newBatchSender(cfg exporterbatcher.Config, set exporter.CreateSettings) *batchSender {
+	bs := &batchSender{
+		activeBatch:  newEmptyBatch(),
+		cfg:          cfg,
+		logger:       set.Logger,
+		shutdownCh:   make(chan struct{}),
+		stopped:      &atomic.Bool{},
+		resetTimerCh: make(chan struct{}),
+	}
+	return bs
+}
+
+func (bs *batchSender) Start(_ context.Context, _ component.Host) error {
+	timer := time.NewTimer(bs.cfg.FlushTimeout)
+	go func() {
+		for {
+			select {
+			case <-bs.shutdownCh:
+				bs.mu.Lock()
+				if bs.activeBatch.request != nil {
+					bs.exportActiveBatch()
+				}
+				bs.mu.Unlock()
+				if !timer.Stop() {
+					<-timer.C
+				}
+				return
+			case <-timer.C:
+				bs.mu.Lock()
+				if bs.activeBatch.request != nil {
+					bs.exportActiveBatch()
+				}
+				bs.mu.Unlock()
+				timer.Reset(bs.cfg.FlushTimeout)
+			case <-bs.resetTimerCh:
+				if !timer.Stop() {
+					<-timer.C
+				}
+				timer.Reset(bs.cfg.FlushTimeout)
+			}
+		}
+	}()
+
+	return nil
+}
+
+type batch struct {
+	ctx     context.Context
+	request Request
+	done    chan struct{}
+	err     error
+}
+
+func newEmptyBatch() *batch {
+	return &batch{
+		ctx:  context.Background(),
+		done: make(chan struct{}),
+	}
+}
+
+// exportActiveBatch exports the active batch asynchronously and replaces it with a new one.
+// Caller must hold the lock.
+func (bs *batchSender) exportActiveBatch() {
+	go func(b *batch) {
+		b.err = b.request.Export(b.ctx)
+		close(b.done)
+	}(bs.activeBatch)
+	bs.activeBatch = newEmptyBatch()
+}
+
+// isActiveBatchReady returns true if the active batch is ready to be exported.
+// The batch is ready if it has reached the minimum size or the concurrency limit is reached.
+// Caller must hold the lock.
+func (bs *batchSender) isActiveBatchReady() bool {
+	return bs.activeBatch.request.ItemsCount() >= bs.cfg.MinSizeItems ||
+		(bs.concurrencyLimit > 0 && bs.activeRequests.Load() >= bs.concurrencyLimit)
+}
+
+func (bs *batchSender) send(ctx context.Context, req Request) error {
+	// Stopped batch sender should act as pass-through to allow the queue to be drained.
+	if bs.stopped.Load() {
+		return bs.nextSender.send(ctx, req)
+	}
+
+	bs.activeRequests.Add(1)
+	defer bs.activeRequests.Add(^uint64(0))
+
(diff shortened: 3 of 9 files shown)


exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,443 @@
...
+	for _, r := range rs {
+		reqs = append(reqs, &request{
+			baseRequest: baseRequest{ctx: req.Context()},
+			Request:     r,
+		})
+	}
+	return reqs, nil
+}
+
+func NewMergeBatcher(cfg MergeBatcherConfig, mf BatchMergeFunc, opts ...BatcherOption) *Batcher {
```
mx-psi, line 177: Instead of passing a function directly I think we should have an interface like it is done for `http.Handler` and `http.HandlerFunc` in the standard library

exporter/exporterhelper/common.go
```diff
@@ -132,6 +132,15 @@ func WithCapabilities(capabilities consumer.Capabilities) Option {
 	}
 }
 
+func WithBatcher(batcher *Batcher) Option {
+	return func(o *baseExporter) {
+		if !o.requestExporter {
+			panic("batching is only available for the new request exporters")
```
mx-psi, line 138: Instead of panicking we could add a private field for errors that we can check and use to report errors through `newBaseExporter`

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,443 @@
...
+
+// MergeSplitBatcherConfig defines batching configuration for merging or splitting requests based on a timeout and
+// minimum and maximum number of items.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type MergeSplitBatcherConfig struct {
+	MergeBatcherConfig `mapstructure:",squash"`
+	// MaxSizeItems is the maximum number of the batch items, i.e. spans, data points or log records for OTLP,
+	// but can be anything else for other formats. If the batch size exceeds this value,
+	// it will be broken up into smaller batches if possible.
```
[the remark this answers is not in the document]
dmitryax: Makes sense. addressed

exporter/exporterhelper/batch_sender.go
dmitryax: I'm not sure I understand the benefits of using the interfaces here. I don't think we will have a need to add other functions as optional interfaces if that's your concern. Even if we will, it's possible to do with another `BatcherOption`. For me, the function seems to be easier to understand and implement. And we are asking users to provide functions in most of the other exporter helper interfaces.

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,461 @@
...
+		batcher:    b,
+		newRequest: make(chan *request, runtime.NumCPU()),
+	}
+	b.goroutines.Add(1)
+	go s.start()
+	return s
+}
+
+func (b *Batcher) mergeRequests(req1 *request, req2 *request) (*request, error) {
+	r, err := b.mergeFunc(req1.Context(), req1.Request, req2.Request)
```
michalpristas, line 155: i wonder if it would be valuable to merge req1 and req2 Contexts into a single contex. here behavior may not be deterministic when considering cancelable contexts or contexts with Deadlines

dmitryax: @michalpristas, this is a good question. In the current implementation (which is similar to the current batch processor), we should pass only one context from the request that is currently being added to the batch and is blocking the current `Consume` call. We can ignore the context associated with the existing batch because all the other requests are already accumulated in the batch and are not active anymore.

But if the queue is not enabled, we probably want to block all the incoming requests until a batch is ready, as discussed in https://github.com/open-telemetry/opentelemetry-collector/pull/8762. In that case, we probably need some kind of context merging. That option is not implemented yet.

exporter/exporterhelper/batch_sender.go
mx-psi: The intent is to be consistent with the standard library, since many users will be familiar with that. I agree it's not particularly likely that we will add other functions, but being similar to the standard library is, I believe, a good goal to have by itself.

mx-psi: This is debatable but to me it's also slightly cleaner to write batch merge procedures that keep state if using an interface.

If you have code like this:

```go
type Merger struct {
    state State
}

func (m *Merger) BatchMerge(context.Context, Request, Request) (Request, error) {
     // check or update m.state here
     // ...
}

```

with the current code you have to do:

```go
merger := NewMerger()
mergeBatcher := NewMergeBatcher(cfg, merger.BatchMerge) // merger is passed implicitly
```

while with the interface you directly pass the merger object:

```go
merger := NewMerger()
mergeBatcher := NewMergeBatcher(cfg, merger)
```

Not sure how common this would be, and again, it is debatable, but it seems cleaner with an interface.

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,461 @@
...
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer/consumererror"
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+)
+
+// MergeBatcherConfig defines a basic configuration for batching requests based on a timeout and a minimum number of
+// items.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
```
mx-psi, line 24: IMO we should start designing things with exporterhelper with 1.0 in mind. This doesn't mean that it will happen any time soon, but at least we should not make it harder to publish exporterhelper 1.0.

To this end, I think we should publish this functionality as a separate module (`exporter/exporterhelper/batcher`). This needs a bit of a refactor to be able to define `WithBatcher` on a separate module (we likely need an exporter/exporterhelper/internal` module to be able to have options), but allows us to eventually declare `exporter/exporterhelper` as 1.0 without declaring this (or other functionality) as part of 1.0.

What do you think? I can try and do this for other experimental functionality on other modules and see how it works out, but here I think it should be doable.

exporter/exporterhelper/batch_sender.go
bogdandrutu: To not forget @dmitryax please be consistent and apply the same logic https://github.com/open-telemetry/opentelemetry-collector/pull/8764. 

I will think more about this.

bogdandrutu: I personally don't mind any of the solutions, but if we do `Interface` please offer also a `MergerFunc` that helps with implementing the interface simpler. I would also rename `BatchMerge` -> `Merge`.

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,461 @@
...
+		reqs = append(reqs, &request{
+			baseRequest: baseRequest{ctx: req.Context()},
+			Request:     r,
+		})
+	}
+	return reqs, nil
+}
+
+// NewMergeBatcher creates a new Batcher that merges requests.
+func NewMergeBatcher(cfg MergeBatcherConfig, mf BatchMergeFunc, opts ...BatcherOption) *Batcher {
```
bogdandrutu, line 190: Do we need `NewMergeBatcher` and the `Batcher` struct to be public? Why not `WithBatcher` accepts everything that `NewMergeBatcher` accepts?

exporter/exporterhelper/batch_sender.go
dmitryax: @mx-psi, what kind of `State` do you expect the `Merger` to have? The provided function is expected to be stateless by the API. If we ask users to provide an interface, it makes it not that clear as a stateless function IMO.

exporter/exporterhelper/batch_sender.go
dmitryax: I agree, I was planning to move it to a sub-package, but I'm not sure if `exporter/exporterhelper/batcher` deserves to be a module. I think we can wait for the transition and release exporter 1.0

exporter/exporterhelper/batch_sender.go
dmitryax: The idea is to provide 2 options for the exporter authors:
- merge only batching 
- merge+split batching

Probably we can move `split` option to another `BatcherOption` and keep one `WithBatcher`

exporter/exporterhelper/batch_sender.go
mx-psi: >  what kind of State do you expect the Merger to have?

A logger or metric instruments could be common examples (potentially also something specific to how you handle requests, but I don't have a concrete example here)

exporter/exporterhelper/batch_sender.go
mx-psi: :+1: Package is fine, we can do module in the future if we want to release 1.0 before we stabilize this

exporter/exporterhelper/batch_sender.go
dmitryax: I don't think we should encourage users to write logs and emit metrics themselves in the batch function. The function is expected to be slim. All the metrics reporting will be done by the exporter helper. Returned errors should be logged by the exporter helper as well. If there is a real need to do so, it's probably better to keep them in the custom exporter. Then, they will provide `myExporter.mergeBatches` func.

BTW I got feedback from a user of the new interface that having `Export()` function as part of the `Request` interface is pretty inconvenient for the same reason of having to pass a state. Probably, we should consistently move towards functions instead of interfaces so users can provide methods of their custom exporter like `myExporter.convert`, `myExporter.mergeBatches` and `myExporter.export` if needed. The interface would be

```
func NewLogsRequestExporter(
	_ context.Context,
	set exporter.CreateSettings,
	convertFunc ConvertLogsFunc,
	exportFunc ExportFunc,
	options ...Option,
) (exporter.Logs, error)
```

cc @bogdandrutu WDYT?

exporter/exporterhelper/batch_sender.go
dmitryax: Applied splitting as an option

exporter/exporterhelper/batch_sender.go
dmitryax: Given that we already have a bunch of API related to the exporterhelper v2 in this module, I would leave this here just for now, marked as experimental the same way as the existing experimental API. We can move this out before 1.0 or agree to have `experimental` API parts in 1.x. The problem is that we have to move `Request` API from here to be able to move the batcher API. We cannot agree with @bogdandrutu to move the `Request` API yet :) See https://github.com/open-telemetry/opentelemetry-collector/pull/8867#discussion_r1391802064

mx-psi: Okay, makes sense, no need to do this right now, let's continue with the status quo of doing things and change after we have general rules for how to add experimental APIs

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,327 @@
...
+		if batch.err != nil {
+			return batch.err
+		}
+		reqs = reqs[1:]
+	} else {
+		bs.mu.Unlock()
+	}
+
+	// Intentionally do not put the last request in the active batch to not block it.
+	// TODO: Consider including the partial request in the error to avoid double publishing.
```
jmacd, line 287: If I'm reading the code and TODO correctly -- it looks like the batch `Export()` error may or may not be returned to the caller depending on whether the export is triggered by a timer or by a full batch.  When one caller sees the error, it's because a batch that (may have?) included its own data failed synchronously while trying to send. The error handling looks inaccurate.

I'm looking to replicate the functionality in https://github.com/open-telemetry/otel-arrow/blob/main/collector/processor/concurrentbatchprocessor/README.md, which has each caller block until the data they entered into one or more batches has been decisively exported or not. Each caller ends up with different partials depending on whether the data they submitted lands in one or more batches. If the data they submitted ends in exactly one batch, it will not get a partial error in that case, it will get a total error describing the uniform result. 

The solution I'm referring to blocks each caller until each caller's data has been processed and returns one of three outcomes to each caller: total error, partial error, and no error. It looks like the solution here may or may not block the caller until their data is processed.

The reason this matters is that I want to have accurate success/failure counts from the producer perspective. If the batch processor does not return accurate error information to the SDKs that produced the data, then the SDK-level metrics become not very useful. It's not only to avoid double publishing, it's to improve SDK-level metrics accuracy.

dmitryax: The current solution always blocks all the callers until the batch is complete or timed out. 

This particular comment is about splitting behavior. If one of the batch splits fails to be sent, we return an error as the whole batch failed. We should return a partial error, letting the caller know what portion failed. I can take care of this TODO in this PR. Addressing it should cover all your needs

dmitryax: BTW the line above is not related to the TODO. It just means that we send the remainder of a split right away, even if the size of it is smaller than the minimum threshold. It's needed to avoid stacking blocked requests on top of each other if they are not part of one batch.

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,327 @@
...
+		ctx:  context.Background(),
+		done: make(chan struct{}),
+	}
+}
+
+// exportActiveBatch exports the active batch and creates a new one.
+// Caller must hold the lock.
+func (bs *batchSender) exportActiveBatch() {
+	go func(b *batch) {
+		b.err = b.request.Export(b.ctx)
```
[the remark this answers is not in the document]
dmitryax: Sorry for the delayed response. I'm not sure I fully understand your concern. Without batching, if the timeout is not configured, a request will be blocked until the request is successfully sent or canceled. With the enabled batching, it's the same behavior: all incoming requests are blocked until the batched request is successfully sent or any of the initial requests are canceled. The latter is not fully implemented. I'll take care of it, but I believe that resolves [your previous comment](https://github.com/open-telemetry/opentelemetry-collector/pull/8685#discussion_r1378642907), not this one. Also, there is no leak, we wait until all batched requests are complete on shutdown. Please elaborate, if I misunderstood.

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,327 @@
...
+	timer := time.NewTimer(bs.mergeCfg.Timeout)
+	go func() {
+		for {
+			select {
+			case <-bs.shutdownCh:
+				bs.mu.Lock()
+				if bs.activeBatch.request != nil {
+					bs.exportActiveBatch()
+				}
+				bs.mu.Unlock()
```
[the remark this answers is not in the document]
dmitryax: Good point. addressed

exporter/exporterhelper/batch_sender.go
dmitryax: I looked into merging contexts of requests that create a batch.

There was a proposal to implement this capability natively in go https://github.com/golang/go/issues/36503, but it was declined. 

We could manually merge the contexts, but it would require spawning another goroutine specifically to handle the context cancellations. I think this is an overkill for this task. I think it's ok just to take the context from the first request, assuming that it has the shortest timeout. I updated the code accordingly. Let me know WDYT.

exporter/exporterbatcher/batch_func.go
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterbatcher // import "go.opentelemetry.io/collector/exporter/exporterbatcher"
+
+import "context"
+
+// BatchMergeFunc is a function that merges two requests into a single request.
+// Do not mutate the requests passed to the function if error can be returned after mutation or if the exporter is
+// marked as not mutable.
+// This API is at the early stage of development and may change without backward compatibility
```
mx-psi, line 11: Can you put an `Experimental` at the start? That way we can easily grep for experimental APIs (e.g. when we work on stabilizing the `exporterhelper` module

```suggestion
// Experimental: This API is at the early stage of development and may change without backward compatibility
```

exporter/exporterbatcher/batch_func.go
```diff
@@ -0,0 +1,24 @@
...
+type BatchMergeFunc[T any] func(context.Context, T, T) (T, error)
+
+// BatchMergeSplitFunc is a function that merge and/or splits one or two requests into multiple requests based on the
+// configured limit provided in MaxSizeConfig.
+// All the returned requests MUST have a number of items that does not exceed the maximum number of items.
+// Size of the last returned request MUST be less or equal than the size of any other returned request.
+// The original request MUST not be mutated if error is returned after mutation or if the exporter is
+// marked as not mutable. The length of the returned slice MUST not be 0. The optionalReq argument can be nil,
+// make sure to check it before using.
+// This API is at the early stage of development and may change without backward compatibility
```
mx-psi, line 22:
```suggestion
// Experimental: This API is at the early stage of development and may change without backward compatibility
```

exporter/exporterbatcher/config.go
```diff
@@ -0,0 +1,70 @@
...
+package exporterbatcher // import "go.opentelemetry.io/collector/exporter/exporterbatcher"
+
+import (
+	"errors"
+	"time"
+)
+
+// Config defines a configuration for batching requests based on a timeout and a minimum number of items.
+// MaxSizeItems defines batch splitting functionality if it's more than zero.
+// This API is at the early stage of development and may change without backward compatibility
```
mx-psi, line 13:
```suggestion
// Experimental: This API is at the early stage of development and may change without backward compatibility
```

exporter/exporterbatcher/config.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// FlushTimeout sets the time after which a batch will be sent regardless of its size.
+	FlushTimeout time.Duration `mapstructure:"flush_timeout"`
+
+	MinSizeConfig `mapstructure:",squash"`
+	MaxSizeConfig `mapstructure:",squash"`
+}
+
+// MinSizeConfig defines the configuration for the minimum number of items in a batch.
+// This API is at the early stage of development and may change without backward compatibility
```
mx-psi, line 27:
```suggestion
// Experimental: This API is at the early stage of development and may change without backward compatibility
```

exporter/exporterbatcher/config.go
```diff
@@ -0,0 +1,70 @@
...
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type MinSizeConfig struct {
+	// MinSizeItems is the number of items (spans, data points or log records for OTLP) at which the batch should be
+	// sent regardless of the timeout. There is no guarantee that the batch size always greater than this value.
+	// This option requires the Request to implement RequestItemsCounter interface. Otherwise, it will be ignored.
+	MinSizeItems int `mapstructure:"min_size_items"`
+}
+
+// MaxSizeConfig defines the configuration for the maximum number of items in a batch.
+// This API is at the early stage of development and may change without backward compatibility
```
mx-psi, line 37:
```suggestion
// Experimental: This API is at the early stage of development and may change without backward compatibility
```

exporter/exporterhelper/batch_sender.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+)
+
+// batchSender is a component that accepts places requests into batches before passing them to the downstream senders.
+//
+// batch_processor implements consumer.Traces and consumer.Metrics
+//
```
mx-psi, line 22: I think this is a leftover from copying from https://github.com/open-telemetry/opentelemetry-collector/blob/fc08135156f6de4d553c29f0f5de2b51addbc9c8/processor/batchprocessor/batch_processor.go#L34-L36, so we can remove it, right?
```suggestion
```

exporter/exporterbatcher/batch_func.go
dmitryax: Make sense. Updated 👍

exporter/exporterhelper/batch_sender.go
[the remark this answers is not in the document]
djaglowski: > all incoming requests are blocked until the batched request is successfully sent or any of the initial requests are canceled. The latter is not fully implemented.

@dmitryax, I'm troubleshooting a customer issue where the sending queue is full and the collector will not shut down. Best I can tell it may be hanging here because it is retrying requests indefinitely. I found this comment and also noticed that https://github.com/open-telemetry/opentelemetry-collector/pull/9738 added functionality which breaks linearity between the initial requests. I don't understand the code deeply enough yet but my intuition is that context cancelation is not propagating correctly. Is that still unimplemented?

PR #8709, opened 2023-10-19
Automated status seporting via typed errors

mwear:
**Description:** 
This PR is an alternative to #8684. It's currently a rough draft to get feedback on whether or not this is a good direction to pursue. 

This PR adds default status reporting to processors and exporters by wrapping the `Consume*` methods to report status based on the returned error type. To give users more control over automated status reporting, it introduces `componenterror` types that match the error status types. It also improves the existing automated status reporting for component.Start by allowing a component to report StatusRecoverableError, and not halt startup, as well as automating reporting StatusOK. The idea is that we can wrap calls to well known methods for components, consumers, and exporters and use the return type of the error as the reported status value. This will provide out of the box default status reporting that can be overidden as needed. The behavior is as follows:

* Consume
  * If err == nil report `StatusOK`
  * If err matches a `componenterror` report the corresponding status for the error type
  * If err isn't a `componenterror` report `StatusRecoverableError`
* Start
  * If err == nil report `StatusOK`
  * If err is `componenterror.RecoverableError`, continue starting other components and report `StatusRecoverableError`
  * If err is `componenterror.Permament` or `componenterror.Fatal` report `StatusPermanentError` or `StatusFatalError` respectively
  * If err is not a `componenterror`, report `StatusPermanentError` (this will preserve the current behavior and halt startup)
* Stop
  * If err matches a `componenterror` report the corresponding status
  * If err is not a `componenterror`, report `StatusPermanentError`  

From within one of these well defined methods, returning different `componenterror` types can control the automated status reporting. From outside these methods, the `TelemetrySettings.ReportComponentStatus` API can be used directly.


mwear:
I'm closing this as it appears to be a bad idea based on our discussion at the collector SIG earlier. Thanks to all who were involved in the discussion, I found it very helpful. For those not at the SIG, the tl;dr is that we cannot assume that errors from `Consume*` relate to error statuses for components and `Start` is riddled with edge cases. While I was hoping to find more ways to automate status reporting, it seems like there are not any further reporting we can automate at this level. We may find that there could be shareable behavior for classes of components, but for now each component will be responsible for reporting its status from the period after `starting` until `stopping`.

PR #8713, opened 2023-10-20
[cmd/builder] Fix otelcol_version being ignored

### LINKED ISSUE #8692: Breaking change in 0.87.0? Build fails and demands component.TelemetrySettings argument for call to receivers.GetOrAdd
**Describe the bug**
before 0.87.0 was released my setup was working, using 0.86.0. After release of 0.87.0 I get a build error saying:
```yaml
not enough arguments in call to receivers.GetOrAdd
  have (*Config, func() (*otlpReceiver, error))
  want (*Config, func() (*otlpReceiver, error), *component.TelemetrySettings)
```
See images in bottom for more details.

If I change my config to match the latest version it builds. But that I assume will only last until the next version is released.
I tried downloading the binary version of the builder, just to make sure I did not have any issues there with versions etc and I got the same behavior.

**Steps to reproduce**
go install go.opentelemetry. io/collector/cmd/builder@v0.86.0
builder --config="path to yaml defined below" --output-path="some output path"

A successful build

**What did you see instead?**
image

**What version did you use?**
Version: v0.86.0

**What config did you use?**
Config: yaml
```yaml
dist:
  name: otel-collector.exe
  otelcol_version: 0.86.0

receivers:
  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.86.0

processors:
  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.86.0
  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/redactionprocessor v0.86.0
  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributesprocessor v0.86.0

exporters:
  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.86.0
  - gomod: go.opentelemetry.io/collector/exporter/otlpexporter v0.86.0
```

**Environment**
OS: Win 11 Version 10.0.22621
Compiler: go 1.21.3

### Discussed in https://github.com/open-telemetry/opentelemetry-collector/discussions/8668

<div type='discussions-op-text'>

<sup>Originally posted by **robinlindberg** October 11, 2023</sup>
This might be two separate bugs.

I am unable to install a specific version of the builder. I always get "dev" which seems to be the default value the version.go file sets.

image-2023-10-11-15-59-19-589

I suspected the builder version being wrong was the cause of the go.mod being filled with both 0.86.0 and 0.87.0 versions. But now I tried using the binaries of the builder directly, instead of doing Go install. I got the same result when building for the non-latest version, e.g. 0.86.0. Building for the latest version works in both scenarios.
image
image

Does anyone have any clue what this might be caused by?

What I have tried:
- Uninstalling Go
- Clearing out all files from GOPATH
- If I instead refer to the latest version, it builds just fine
- Running everything as a Github action and try building there, completely separate from my machine. Same behavior.
- Using the builder binary directly instead of using Go install

</div>

mx-psi:
**Description:** 

Adds `go.opentelemetry.io/collector/otelcol` to `go.mod` template for the builder. This fixes #8692.

Since #8443, the `otelcol` folder is its own component. Before this change, `otelcol_version` was enforced by the `collector` module dependency:
https://github.com/open-telemetry/opentelemetry-collector/blob/287b98f6973fd6baa278150b9fca8c83abea0af4/cmd/builder/internal/builder/templates/go.mod.tmpl#L23

After this change, the `go mod tidy` step here:
https://github.com/open-telemetry/opentelemetry-collector/blob/287b98f6973fd6baa278150b9fca8c83abea0af4/cmd/builder/internal/builder/main.go#L115

will add the latest available version for the `otelcol` module since none of the components actually depend on it. For example, with the `v0.86.0` builder config the output is as follows:

```
❯ go mod tidy -v -compat=1.20
go: finding module for package go.opentelemetry.io/collector/otelcol
go: found go.opentelemetry.io/collector/otelcol in go.opentelemetry.io/collector/otelcol v0.87.0
```

Explicitly adding `otelcol` makes it so the `otelcol_version` is correctly honored.

**Link to tracking Issue:** Fixes #8692

### DIFF
--- a/cmd/builder/internal/builder/templates/go.mod.tmpl
+++ b/cmd/builder/internal/builder/templates/go.mod.tmpl
@@ -21,7 +21,9 @@ require (
 	{{if .GoMod}}{{.GoMod}}{{end}}
 	{{- end}}
 	go.opentelemetry.io/collector v{{.Distribution.OtelColVersion}}
+	{{if .Distribution.RequireOtelColModule}}
 	go.opentelemetry.io/collector/otelcol v{{.Distribution.OtelColVersion}}
+	{{- end}}
 )
 
 require (
--- a/cmd/builder/internal/builder/templates/go.mod.tmpl
+++ b/cmd/builder/internal/builder/templates/go.mod.tmpl
@@ -20,10 +20,7 @@ require (
 	{{- range .Processors}}
 	{{if .GoMod}}{{.GoMod}}{{end}}
 	{{- end}}
-	go.opentelemetry.io/collector v{{.Distribution.OtelColVersion}}
-	{{if .Distribution.RequireOtelColModule}}
-	go.opentelemetry.io/collector/otelcol v{{.Distribution.OtelColVersion}}
-	{{- end}}
+	go.opentelemetry.io/collector{{if .Distribution.RequireOtelColModule}}/otelcol{{end}} v{{.Distribution.OtelColVersion}}
 )
 
 require (
--- a/cmd/builder/internal/builder/templates/go.mod.tmpl
+++ b/cmd/builder/internal/builder/templates/go.mod.tmpl
@@ -21,6 +21,7 @@ require (
 	{{if .GoMod}}{{.GoMod}}{{end}}
 	{{- end}}
 	go.opentelemetry.io/collector v{{.Distribution.OtelColVersion}}
+	go.opentelemetry.io/collector/otelcol v{{.Distribution.OtelColVersion}}
 )
 
 require (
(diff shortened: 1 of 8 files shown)


cmd/builder/internal/builder/templates/go.mod.tmpl
```diff
@@ -21,6 +21,7 @@ require (
 	{{if .GoMod}}{{.GoMod}}{{end}}
 	{{- end}}
 	go.opentelemetry.io/collector v{{.Distribution.OtelColVersion}}
+	go.opentelemetry.io/collector/otelcol v{{.Distribution.OtelColVersion}}
```
mx-psi, line 24: This breaks compatibility of the new builder with versions older than v0.86.0. Is that okay?

mx-psi: Changed to conditionally add it and be backwards compatible

codeboten (approved):
Thanks for fixing @mx-psi

PR #8726, opened 2023-10-23
[cmd/builder] (docs) Fix link to collector builder releases

myieye:
**Description:**
With the collector builder now in the same repo as the collector itself, `/latest` is not a good place to find the latest release of the collector builder. This change provides a link that finds the desired release.


jpkrohling (approved):

codeboten (approved):
Thanks @myieye

Issue #8220, opened 2023-08-11
Release `featuregate` 1.0


mx-psi:
The [`featuregate`][1] module [milestone][2] has been completed for three months at time of writing. None of the remaining [open issues related to featuregate][3] seem like a blocker for GA for me.

This issue is for discussing if there are any blockers not listed on the issue tracker to release a 1.0 version of this module, and, if not, release this module. At time of writing this would be our first 1.0 module.

A 1.0 release would [honor our stability guarantees][4].

[1]: https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate
[2]: https://github.com/open-telemetry/opentelemetry-collector/milestone/30
[3]: https://github.com/open-telemetry/opentelemetry-collector/labels/area%3Afeaturegate
[4]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/VERSIONING.md

atoulme:
I would like to see https://github.com/open-telemetry/opentelemetry-collector/issues/7804 addressed first as https://github.com/open-telemetry/opentelemetry-collector/pull/7991 introduces an API breaking change.

mx-psi:
#8766 may be necessary for 1.0

mx-psi:
I am going to close this one in favor of #8946, which follows the template from #8935

PR #8727, opened 2023-10-23
[featuregate] Deprecate NewFlag in favor of RegisterFlags

mx-psi:
**Description:** 

Deprecate `featuregate.NewFlag` in favor of `Registry.RegisterFlags`, which registers flags into a provided flagset.

The intent is to unblock future development of new features such as #7991 without incurring into deprecations or breaking changes to implement these.
This will allow us to do a 1.0 release.

**Link to tracking Issue:** Relates to #8220


djaglowski (approved):

TylerHelmuth (approved):

MovieStoreGuy (approved):

featuregate/flag.go
```diff
@@ -10,11 +10,25 @@ import (
 	"go.uber.org/multierr"
 )
 
+// Flags related to feature gates.
+type Flags struct {
```
bogdandrutu, line 14: Question, why just having Options in the `NewFlag` is not enough to extend capabilities?

mx-psi: >  why just having Options in the NewFlag is not enough to extend capabilities?

How would you add the new flag from #7991 in this way? One would have to define the strict flag before defining the usual flag, I am not sure if that is even possible.

I am also not sure what the benefit of `Options` over this approach is. It is still [a breaking change](https://go.dev/play/p/LYJ7VL_kXwP).

bogdandrutu: I am not arguing about breaking change, but about the fact that now this API is more complicated for normal users. Returning a struct that then I need decide which flag to install, etc.

If we want to allow use to construct multiple flags, then pass a FlagSet to the func and do that. Also still we should add Options to allow extensions in the future.

atoulme:
I think I need to simplify things. I will drop this requirement to run this with a CLI flag since it's so impractical and work my way with a programmatic change so I can test strict mode in a test. We can stop this, we're trying to make it work but it's just impractical and I have exhausted enough good will with this attempt.

bogdandrutu:
@atoulme I understand your frustration, but please don't close others PRs :)

atoulme:
@bogdandrutu Oh I meant to help close this because @mx-psi only opened it to make my own PR whole as we're going for 1.0. No harm done. Just trying to make it explicit to @mx-psi his help is deeply appreciated, but I need to relieve him from his duty of pushing this work in ahead of the 1.0 release of featuregate. I'll go close my original PR as well.

featuregate/flag.go
bogdandrutu: I would do something like:

```golang
// NewFlags returns Flags that control building the global Registry.
func RegisterFlag(flag.FlagSet fs, reg *Registry, opts ...Option) Flags {
	...
        // One option can be to enable strict mode/to add a flag for that.
       // Though, the problem with flag is that it has to be defined first.
}
```

bogdandrutu:
@mx-psi @atoulme I still believe the current `NewFlag` is not very extensible, so having a way to extend that may help in the future, see my comments.

atoulme:
Understood! I'll leave it be then. But I can decorrelate the work on strict mode from flags, for now at least.

featuregate/flag.go
mx-psi: PTAL at 1fb858c67, is that what you had in mind? This works for me

bogdandrutu (approved):

featuregate/flag.go
```diff
@@ -10,11 +10,31 @@ import (
...
 
+// FlagsOption is an option for RegisterFlags.
+type FlagsOption interface {
+	private()
+}
+
+// RegisterFlags that directly applies feature gate statuses to a Registry.
+func RegisterFlags(flagSet *flag.FlagSet, reg *Registry, opts ...FlagsOption) {
+	flagSet.Var(newFeatureGateValue(reg), featureGatesFlag, featureGatesFlagDescription)
+}
```
bogdandrutu, line 36: @mx-psi sorry for this :).

Do you think we can simplify this API and make this a func on the Registry struct? Makes things simpler. 

Also separate concern: FlagsOption -> RegisterFlagsOption

mx-psi: Tried that out on f20fbf9e80216aa468b541614b1cdc494856b36e, looks nice :)

mx-psi:
@atoulme @TylerHelmuth @djaglowski PR changed a bit after back and forth with Bogdan, mind taking another look?

bogdandrutu (approved):

Issue #8732, opened 2023-10-24
Start order of extensions is not guaranteed?


yurishkuro:
**Describe the bug**
Maybe I am misreading the situation, but this code iterates over map when starting extensions, so it cannot guarantee in which order they are started: https://github.com/open-telemetry/opentelemetry-collector/blob/e97ceca47f9eacf50688a5b2e40c0ad22261579b/service/extensions/extensions.go#L34

Even though the documentation implies that the order is define by the declaration order: https://github.com/open-telemetry/opentelemetry-collector/blob/e97ceca47f9eacf50688a5b2e40c0ad22261579b/docs/service-extensions.md?plain=1#L63-L66

I observed some spontaneous failures in my experiments with JaegerV2 which declares extensions with an order expectation: `extensions: [jaeger_storage, jaeger_query]` but the logs show differently:
```
2023-10-24T01:17:01.143Z	info	extensions/extensions.go:33	Starting extensions...
2023-10-24T01:17:01.143Z	info	extensions/extensions.go:36	Extension is starting...	{"kind": "extension", "name": "jaeger_query"}
```

**Steps to reproduce**
It's random behavior, hard to reproduce reliably.

**What did you expect to see?**
Extensions should be starting in the order defined in the config.

**What did you see instead?**
Seeing different order.

**What version did you use?**
Version: `v0.87.0`

**What config did you use?**
https://github.com/jaegertracing/jaeger/blob/28520b31471b36d9e949de4ffaa016c2640a3be7/cmd/jaeger-v2/internal/all-in-one.yaml#L1

**Environment**
OS: MacOS, Linux
Compiler(if manually compiled): go 1.20.x

atoulme:
I'd be happy to work on this one.

dmitryax:
I'm not sure this is a bug. I think this was implemented intentionally. All extensions that we have in core and contrib do not depend on each other. This assumption is used in some discussions around config grouping/merging, e.g.: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18509 and https://github.com/open-telemetry/opentelemetry-collector/issues/8394.

I think we need a strong argument to introduce this behavior.

@yurishkuro do you have an extension that depends on another extension? Can you please let us know why it's needed in your use case?

yurishkuro:
[...]

PR #8733, opened 2023-10-24
[service/extensions] extension lifecycle order

atoulme:
**Description:**
Enforce order of start and shutdown of extensions according to configuration

**Link to tracking Issue:**
Fixes issue #8732

**Testing:**
Unit tests

### DIFF
--- a/service/extensions/extensions.go
+++ b/service/extensions/extensions.go
@@ -80,7 +80,6 @@ func (bes *Extensions) NotifyPipelineReady() error {
 }
 
 func (bes *Extensions) NotifyPipelineNotReady() error {
-	// Notify extensions in reverse order.
 	var errs error
 	for _, extID := range bes.extensionIDs {
 		ext := bes.extMap[extID]
--- a/service/extensions/extensions.go
+++ b/service/extensions/extensions.go
@@ -23,18 +23,20 @@ const zExtensionName = "zextensionname"
 
 // Extensions is a map of extensions created from extension configs.
 type Extensions struct {
-	telemetry   servicetelemetry.TelemetrySettings
-	extMap      map[component.ID]extension.Extension
-	instanceIDs map[component.ID]*component.InstanceID
+	telemetry    servicetelemetry.TelemetrySettings
+	extensionIDs []component.ID
+	extMap       map[component.ID]extension.Extension
+	instanceIDs  map[component.ID]*component.InstanceID
 }
 
 // Start starts all extensions.
 func (bes *Extensions) Start(ctx context.Context, host component.Host) error {
 	bes.telemetry.Logger.Info("Starting extensions...")
-	for extID, ext := range bes.extMap {
+	for _, extID := range bes.extensionIDs {
 		extLogger := components.ExtensionLogger(bes.telemetry.Logger, extID)
 		extLogger.Info("Extension is starting...")
 		instanceID := bes.instanceIDs[extID]
+		ext := bes.extMap[extID]
 		_ = bes.telemetry.ReportComponentStatus(instanceID, component.NewStatusEvent(component.StatusStarting))
 		if err := ext.Start(ctx, components.NewHostWrapper(host, extLogger)); err != nil {
 			_ = bes.telemetry.ReportComponentStatus(instanceID, component.NewPermanentErrorEvent(err))
@@ -49,8 +51,9 @@ func (bes *Extensions) Start(ctx context.Context, host component.Host) error {
 func (bes *Extensions) Shutdown(ctx context.Context) error {
 	bes.telemetry.Logger.Info("Stopping extensions...")
 	var errs error
-	for extID, ext := range bes.extMap {
+	for _, extID := range bes.extensionIDs {
 		instanceID := bes.instanceIDs[extID]
+		ext := bes.extMap[extID]
 		_ = bes.telemetry.ReportComponentStatus(instanceID, component.NewStatusEvent(component.StatusStopping))
 		if err := ext.Shutdown(ctx); err != nil {
 			_ = bes.telemetry.ReportComponentStatus(instanceID, component.NewPermanentErrorEvent(err))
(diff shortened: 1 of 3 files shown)


yurishkuro (approved):
Thanks!

dmitryax (changes requested):
Just to hold off until discussion in https://github.com/open-telemetry/opentelemetry-collector/issues/8732 is resolved

atoulme:
I received a comment that I should stop extensions in reverse order that they are started, I'll push that. After that, we can discuss the merits of this PR.

yurishkuro:
Another implementation option discussed on the issue is to build support for extension dependencies in the graph/ module. Then an extension can programmatically declare a dependency on another extension and the graph would guarantee correct execution order. This would be a larger change, but preferable over explicit config order where users need to worry about listing extensions in the right order.

dmitryax:
> Another implementation option discussed on the issue is to build support for extension dependencies in the graph/ module. Then an extension can programmatically declare a dependency on another extension and the graph would guarantee correct execution order. This would be a larger change, but preferable over explicit config order where users need to worry about listing extensions in the right order.

Makes sense, but it would require another predefined way to specify dependencies on the extensions in the configuration interface. The current dependencies topology is constructed directly from the pipeline YAML definition. Introducing an additional source for building the graph seems like an excessive complication of the user interface. And some extensions naturally do not provide any capabilities to other components, only to the end-user. I think it's ok to keep it simple like implemented in this PR going forward.

yurishkuro:
> Introducing an additional source for building the graph seems like an excessive complication of the user interface.

Don't get me wrong, I am happy with this PR. But when we talk about complicating user interface, I think it's important to remember _which users are exposed to such interface_. With dependencies expressed programmatically, a very small number of users who write their own extensions are affected. With dependencies expressed via config order, _all_ users of the collector are exposed to the notion that there are dependencies between extensions, even though there's very little they can actually do about it. E.g. `jaeger_query` extension simply will never work if it's placed ahead of `jaeger_storage` in the configuration. So in my documentation I will have to mention "always put it last", even though users didn't really need to be exposed to this accidental complexity.

dmitryax:
@yurishkuro, how do you envision a programmatic way to define the dependencies? Something like an optional interface for extensions to define a dependency on another extension interface? I believe that doesn't require integration with the pipeline graph. We would still start all the extensions before the pipelines.

yurishkuro:
> how do you envision a programmatic way to define the dependencies?

I'd add an optional argument `dependsOn []component.ID` here:

https://github.com/open-telemetry/opentelemetry-collector/blob/6405e152a2d9a08de1fe1b70fd4ac775a12a3cda/extension/extension.go#L103-L107

I'm not sure why these constructor APIs are currently not made extensible by taking args as a struct instead of fixed positional args.

dmitryax:
> I'd add an optional argument dependsOn []component.ID here

In that case, you can only define a hardcoded dependency on a particular component. There should be a way to define a dependency on a class of extensions like storage or encoding, which is defined by an interface an extension implements.

yurishkuro:
> In that case, you can only define a hardcoded dependency on a particular component. There should be a way to define a dependency on a class of extensions like storage or encoding, which is defined by an interface an extension implements.

I personally would avoid duck-typing dependencies, it's better to have explicit ones by ID. There is only one pool of extensions, so what if I have some components depending on one kind of encoding extension and other components depending on another? When they refer to those by ID there is no ambiguity, but duck-typing only works when there is exactly one instance defined for a given interface.

dmitryax:
Extensions should provide the functionality to components via interfaces like [storage](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/experimental/storage/storage.go#L14). Components dependent on that interface can choose any storage extension even if it's implemented outside of OTel core/contrib repos. Choosing a particular storage implementation is exposed to the user with a config option. Having a hardcoded set of IDs means that custom extensions of particular classes cannot be implemented outside, and every extension dependent on a particular class of extension has to maintain a set of all instances, e.g., each time a new storage/encoding extension is added, all dependent components have to be updated.

> There is only one pool of extensions, so what if I have some components depending on one kind of encoding extension and other components depending on another? 

They should be two classes of extensions that implement two different (or additional) interfaces.

yurishkuro:
@dmitryax I didn't mean that the dependency **has** to be hardcoded: for some extensions it may be sufficient, others may need to express their dependencies after reading their own config. That means the dependencies should not be exposed via `NewFactory`, but as part of `CreateExtension` constructor. You're right, in this case a runtime interface check is a good solution - here's a prototype #8768 (to be combined with this PR for the actual ordered start/shutdown)

service/extensions/extensions.go
```diff
@@ -77,7 +82,8 @@ func (bes *Extensions) NotifyPipelineReady() error {
 func (bes *Extensions) NotifyPipelineNotReady() error {
 	// Notify extensions in reverse order.
```
yurishkuro, line 83: is this accurate?

dmitryax:
@yurishkuro, I think we can merge this one and rebase https://github.com/open-telemetry/opentelemetry-collector/pull/8768 on top of it since it won't cause any breaking changes, right?

service/extensions/extensions.go
dmitryax: Looks like it's not. @atoulme can you please address this?

yurishkuro:
> I think we can merge this one and rebase https://github.com/open-telemetry/opentelemetry-collector/pull/8768 on top of it since it won't cause any breaking changes, right?

@dmitryax my PR is already based on this one (it includes the original commits). I don't have a strong objection to merging each one separately, but it means the 1st PR will introduce one way of ordering (based purely on config order) and 2nd will override that in favor of the order defined by the DependentExtension interface (with other order still being random). In theory it's possible to support both at the same time: by default start with config-based order as documentation implies, but if DependentExtension interfaces are detected in some extensions move them to the end and apply topological sort (seems a bit complicated). I think just the 2nd approach is cleaner - no order guarantees unless explicitly requested via interface.

service/extensions/extensions.go
atoulme: How would you like me to address this? It feels like this comment needs to be removed.

dmitryax:
> I think just the 2nd approach is cleaner - no order guarantees unless explicitly requested via interface.

Ok, makes sense to me. Especially if that is simpler

dmitryax:
Superseded by https://github.com/open-telemetry/opentelemetry-collector/pull/8768

Issue #8752, opened 2023-10-25
[confighttp] Support http+unix 


atoulme:
**Is your feature request related to a problem? Please describe.**
In some cases, I want to offer HTTP connection over a socket rather than a tcp port.

**Describe the solution you'd like**
Support unix socket style server support, such as 
```
http+unix:/var/run/otel.socket:/v1/traces
```

mx-psi:
What component(s) do we want to use this in?

atoulme:
I would like to use this with the internal metrics endpoint so it doesn't use a port on the host. I also want to use this for OTLP on hosts.

atoulme:
This was just asked about on the CNCF slack :)

utezduyar:
Can loopback (localhost) not work? Or is it the case that you don't want to use any kind of port?

atoulme:
localhost is not as secure as using a socket bound to file permissions, and we run into port conflicts when we have multiple collectors running - we have the case where the collector is now so popular it is used by different products for different purposes, and tends to clash over port 8888.

PR #8753, opened 2023-10-25
[confighttp] support http+unix

atoulme:
**Description:**
Support binding HTTP servers to a unix socket, and HTTP clients to connect to unix sockets.

**Link to tracking Issue:**
#8752 

**Testing:**
Unit tests.

**Documentation:**
Not yet


config/confighttp/confighttp.go
```diff
@@ -238,7 +246,14 @@ type HTTPServerSettings struct {
 
 // ToListener creates a net.Listener.
 func (hss *HTTPServerSettings) ToListener() (net.Listener, error) {
-	listener, err := net.Listen("tcp", hss.Endpoint)
+	network := hss.Network
+	if network == "" {
+		network = "tcp"
+	}
+	if network == "http+unix" || network == "https+unix" || network == "unixpacket" {
```
bogdandrutu, line 253: is this a standard? if yes, can you link to a doc?

config/confighttp/confighttp.go
```diff
@@ -255,6 +270,15 @@ func (hss *HTTPServerSettings) ToListener() (net.Listener, error) {
 	return listener, nil
 }
 
+func (hss *HTTPServerSettings) Validate() error {
+	switch hss.Network {
+	case "tcp", "tcp4", "tcp6", "unix", "unixpacket", "http+unix", "https+unix", "":
+	default:
+		return fmt.Errorf("invalid network %q", hss.Network)
+	}
+	return nil
```
bogdandrutu, line 279: nit: personal preference, not liking this way of saying not a value, but others may also comment.

config/confighttp/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.20
 require (
 	github.com/golang/snappy v0.0.4
 	github.com/klauspost/compress v1.17.1
+	github.com/peterbourgon/unixtransport v0.0.3
```
bogdandrutu, line 8: Do we trust a random 0.0.3 package? Do we have other alternative?

config/confighttp/confighttp.go
atoulme: would you return nil in the first case statement instead?

config/confighttp/go.mod
atoulme: We could inline that package into the code here, given that the code itself is very small. It would probably be refactored a bit through that process to make it easier to maintain.

config/confighttp/confighttp.go
atoulme: No, I looked around and I didn't see a good standard scheme for HTTP over unix. I saw a number of approaches like this one. I would solicit anyone with expertise to help here.

bogdandrutu: Then why supporting these values? Can we start only with "unix"?

atoulme: sure

sinkingpoint:
@atoulme Any particular reason that this fell away? I can pick it up if you need

atoulme:
Please do!

PR #8763, opened 2023-10-26
[builder] remove the need for adding a replace statement

codeboten:
The builder inserted a replace statement for the cloud.google.com/go to address an ambiguous import issue. This has the negative side effect of breaking go install for any modules that use the builder to generate a go.mod. This change removes that replace statement. Instead, the builder explicitly runs `go get cloud.google.com/go` as part of the generation process.


dmitryax:
@codeboten PTAL at the linter failures

jmacd (approved):

cmd/builder/internal/builder/main.go
```diff
@@ -111,8 +111,15 @@ func GetModules(cfg Config) error {
 		return nil
 	}
 
+	// ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules
+	cmd := exec.Command(cfg.Distribution.Go, "get", "cloud.google.com/go")
```
codeboten, line 115:
```suggestion
	// ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules
	// #nosec G204 -- cfg.Distribution.Go is trusted to be a safe path
	cmd := exec.Command(cfg.Distribution.Go, "get", "cloud.google.com/go")
```

dmitryax (approved):

PR #8766, opened 2023-10-27
[featuregate] Validate arguments to `Register`

mx-psi:
**Description:**
- Validate Gate IDs: ids must be alphanumeric with dots. At a minimum we have to disallow `,` to avoid ambiguity in the CLI, but I am being intentionally more restrictive here
- Validate URLs: we check that a valid URL can be built from it using the `net/url` stdlib package
- Validate versions: we check that a valid version is passed to the `To/FromVersion` options using hashicorp's libraries and also that the implicit version range, when defined, is not empty.

This does not change the public API types, it only adds validation.

Relates to #8220


atoulme (approved):

featuregate/registry.go
```diff
@@ -28,43 +38,66 @@ func NewRegistry() *Registry {
...
 	})
 }
 
 // WithRegisterToVersion is used to set the Gate "ToVersion".
 // The "ToVersion", if not empty, contains the last Collector release in which you can still use a feature gate.
 // If the feature stage is either "Deprecated" or "Stable", the "ToVersion" is the Collector release when the feature is removed.
 func WithRegisterToVersion(toVersion string) RegisterOption {
-	return registerOptionFunc(func(g *Gate) {
+	return registerOptionFunc(func(g *Gate) error {
+		if err := validateVersion(toVersion); err != nil {
```
bogdandrutu, line 95: Should we store the parsed version? Just a question :)

bogdandrutu (approved):

featuregate/registry.go
mx-psi: It simplifies the code a bit so I went for it. This now means some normalization happens (e.g. `v0.088.0` gets transformed into `v0.88.0`) but I think that's okay.

featuregate/registry.go
```diff
@@ -5,12 +5,22 @@ package featuregate // import "go.opentelemetry.io/collector/featuregate"
...
+	"github.com/hashicorp/go-version"
 )
 
-var globalRegistry = NewRegistry()
+var (
+	globalRegistry = NewRegistry()
+
+	// idRegexp is used to validate the ID of a Gate.
+	// IDs' characters must be alphanumeric or dots.
+	idRegexp = regexp.MustCompile(`^[0-9a-zA-Z\.]*$`)
```
bogdandrutu, line 22: Do we not allow "-" or "_"? Just a question, this may fail some already created featuregates. Happy to not allow that initially if we at least check contrib.

featuregate/registry.go
```diff
@@ -77,14 +105,32 @@ func (r *Registry) MustRegister(id string, stage Stage, opts ...RegisterOption)
...
+	}
+
+	if !idRegexp.MatchString(id) {
+		return fmt.Errorf("invalid character(s) in ID")
+	}
+	return nil
+}
+
 // Register a Gate and return it. The returned Gate can be used to check if is enabled or not.
 func (r *Registry) Register(id string, stage Stage, opts ...RegisterOption) (*Gate, error) {
```
bogdandrutu, line 120: Please describe in the comment the rules for the ID string.

featuregate/registry.go
```diff
@@ -28,43 +38,61 @@ func NewRegistry() *Registry {
...
+
+		g.fromVersion = from
+		return nil
 	})
 }
 
 // WithRegisterToVersion is used to set the Gate "ToVersion".
 // The "ToVersion", if not empty, contains the last Collector release in which you can still use a feature gate.
 // If the feature stage is either "Deprecated" or "Stable", the "ToVersion" is the Collector release when the feature is removed.
 func WithRegisterToVersion(toVersion string) RegisterOption {
```
bogdandrutu, line 87: Please explain in the comment what is a valid version.

featuregate/registry.go
```diff
@@ -28,43 +38,61 @@ func NewRegistry() *Registry {
...
+
+		g.referenceURL = referenceURL
+		return nil
 	})
 }
 
 // WithRegisterFromVersion is used to set the Gate "FromVersion".
 // The "FromVersion" contains the Collector release when a feature is introduced.
 func WithRegisterFromVersion(fromVersion string) RegisterOption {
-	return registerOptionFunc(func(g *Gate) {
```
bogdandrutu, line 57: Please explain in the comment what is a valid version.

featuregate/registry.go
```diff
@@ -28,43 +38,61 @@ func NewRegistry() *Registry {
...
+	return registerOptionFunc(func(g *Gate) error {
 		g.description = description
+		return nil
 	})
 }
 
 // WithRegisterReferenceURL adds a URL that has all the contextual information about the Gate.
-func WithRegisterReferenceURL(url string) RegisterOption {
-	return registerOptionFunc(func(g *Gate) {
```
bogdandrutu, line 49: Please explain in the comment what is a valid url (refer to the url Parse documentation or the rfc).

bogdandrutu (approved):

featuregate/registry.go
mx-psi: > Do we not allow "-" or "_"? Just a question, this may fail some already created featuregates.

I think `-` can be confusing, at least on the first value (since we use `-` on the CLI to say when we disable a gate). We don't have feature gates with `-` or `_` on contrib (I manually checked and also would expect contrib tests to fail otherwise).

> Happy to not allow that initially if we at least check contrib.

I would lean towards being conservative here at first if that's okay with everybody else

featuregate/registry.go
mx-psi: Done

featuregate/registry.go
mx-psi: Done

featuregate/registry.go
mx-psi: Done, refererred to `url.Parse`

mx-psi:
Can you take one (hopefully) final look @bogdandrutu ?

codeboten (approved):

dmitryax (approved):

featuregate/registry.go
```diff
@@ -28,43 +38,66 @@ func NewRegistry() *Registry {
...
 // The "FromVersion" contains the Collector release when a feature is introduced.
+// fromVersion must be a valid version string: it may start with 'v' and must be in the format Major.Minor.Patch[-PreRelease].
+// PreRelease is optional and may have dashes, tildes and ASCII alphanumeric characters.
 func WithRegisterFromVersion(fromVersion string) RegisterOption {
-	return registerOptionFunc(func(g *Gate) {
-		g.fromVersion = fromVersion
+	return registerOptionFunc(func(g *Gate) error {
+		from, err := version.NewVersion(fromVersion)
+		if err != nil {
+			return fmt.Errorf("WithRegisterFromVersion: %w", err)
```
bogdandrutu, line 79: Curious why decided to record reference URL but not the version? Any reason?

bogdandrutu (approved):
Would be good to have the same setup in core like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.gitattributes to not expand all the sum files.

featuregate/registry.go
mx-psi: No particular reason, I added them now

mx-psi:
> Would be good to have the same setup in core like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.gitattributes to not expand all the sum files.

See #8810

PR #8767, opened 2023-10-27
[extension/experimental/storage] Deprecate in favor of non-experimental package

djaglowski:
Storage extensions have been widely used for quite some time now. I propose the base package should no longer be considered experimental.


extension/storage/README.md
```diff
@@ -0,0 +1,39 @@
+# Storage
+
+A storage extension persists state beyond the collector process. Other components can request a storage client from the storage extension and use it to manage state.
```
yurishkuro, line 3: how does one configure this extension? It seems to have no factory / configuration like normal extensions do.

atoulme (approved):

extension/storage/README.md
djaglowski: This package is intended only to define the API for storage extensions. There are a couple concrete implementations in contrib which each have their own factory, config, etc. The `file_storage` extension is used quite frequently: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage/filestorage

yurishkuro: I suggest making this explicit in the readme and in the code. Right now it looks nothing like an extension. Maybe even rename `storageapi` or have some common directory structure for APIs

djaglowski: I've added a note to the top of the readme and verified that we have code comments to this effect.

I think the package naming is consistent with for example the `extension/auth` package, which serves a similar purpose.

yurishkuro: thanks.

astencel-sumo (approved):
Is there a time frame when we want to remove the experimental version?

We probably don't want to rush it, as this requires a change in any component that directly uses this extension. In contrib, this is [a couple receivers](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-collector-contrib%20%22extension%2Fexperimental%2Fstorage%22&type=code) :
- Azure Event Hub receiver
- MongoDB Atlas receiver
- SQL Query receiver

There are probably other uses "out there" in other distros. Sumo Logic also has one.

djaglowski:
@astencel-sumo, what time frame do you think would be appropriate for external users of the experimental package? For contrib, I think this will be resolved with a single PR.

extension/storage/nop_client.go
```diff
@@ -0,0 +1,40 @@
...
+package storage // import "go.opentelemetry.io/collector/extension/storage"
+
+import "context"
+
+type nopClient struct{}
+
+var nopClientInstance Client = &nopClient{}
+
+// NewNopClient returns a nop client
+func NewNopClient() Client {
```
bogdandrutu, line 13: In general we have these no-op in a ...test package, should we do the same?

extension/storage/storage.go
```diff
@@ -0,0 +1,96 @@
...
+type operation struct {
+	// Key specifies key which is going to be get/set/deleted
+	Key string
+	// Value specifies value that is going to be set or holds result of get operation
+	Value []byte
+	// Type describes the operation type
+	Type opType
+}
+
+type Operation *operation
```
bogdandrutu, line 74: Very confused about this. Why not simply have `Operation` as the struct and return pointers? I don't think this stops users to write `return Operation{...}` (I may be wrong).

extension/storage/storage.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Extension is the interface that storage extensions must implement.
+type Extension interface {
+	extension.Extension
+
+	// GetClient will create a client for use by the specified component.
+	// Each component can have multiple storages (e.g. one for each signal),
+	// which can be identified using storageName parameter.
+	// The component can use the client to manage state
+	GetClient(ctx context.Context, kind component.Kind, id component.ID, storageName string) (Client, error)
```
bogdandrutu, line 21: NewClient?

extension/storage/nop_client.go
djaglowski: This is not intended specifically as a test component. Previous discussion on the purpose:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2883#discussion_r606296192
- https://github.com/open-telemetry/opentelemetry-collector/pull/8379#discussion_r1320537822

extension/storage/storage.go
djaglowski: I agree. I don't recall how we wound up with this implementation. I think perhaps `Operation` was originally an interface and then we decided to switch to a concrete implementation. Either way, might as well take the opportunity to update since we're changing packages.

djaglowski:
I've moved this back to draft until I have a chance to debug the additional changes.

astencel-sumo:
> @astencel-sumo, what time frame do you think would be appropriate for external users of the experimental package? For contrib, I think this will be resolved with a single PR.

Hm actually, reading again [https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes](https://github.com/open-telemetry/opentelemetry-collector/blob/74b690263bbcc00ed49bd6642e9330d1dddd615f/CONTRIBUTING.md#breaking-changes), I'm leaning towards the regular single release schedule - deprecate in version X, remove in version X+1. I suppose this is what we've been doing with any API deprecations, right?

djaglowski:
@astencel-sumo, that's fine with me. At the same time, I don't see any reason to rush the removal of the experimental package, so long as it's clearly deprecated, so I'd have no objections to keeping it for a bit longer if that would be helpful.

dmitryax (approved):

mx-psi (approved):

extension/storage/storage.go
```diff
@@ -0,0 +1,94 @@
...
+//   - Get doesn't error if a key is not found - it just returns nil.
+//   - Delete doesn't error if the key doesn't exist - it just no-ops.
+//
+// Similarly:
+//   - Batch doesn't error if any of the above happens for either retrieved or updated keys
+//
+// This also provides a way to differentiate data operations
+//
+//	[overwrite | not-found | no-op] from "real" problems
+type Client interface {
```
bogdandrutu, line 37: One thing that may be useful to think is to offer "Append" as well?

extension/storage/storage.go
```diff
@@ -0,0 +1,94 @@
...
+	Delete
+)
+
+type Operation struct {
+	// Key specifies key which is going to be get/set/deleted
+	Key string
+	// Value specifies value that is going to be set or holds result of get operation
+	Value []byte
+	// Type describes the operation type
+	Type opType
```
bogdandrutu, line 71: No need to be public the fields correct?

extension/storage/storage.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func SetOperation(key string, value []byte) *Operation {
```
bogdandrutu, line 74: Can we avoid pointers here and return `Operation`?

extension/storage/storage.go
bogdandrutu: We can also make it as an optional interface to be implemented. Want to hear opinions.

extension/storage/storage.go
```diff
@@ -0,0 +1,94 @@
...
+	Batch(ctx context.Context, ops ...*Operation) error
+
+	// Close will release any resources held by the client
+	Close(ctx context.Context) error
+}
+
+type opType int
+
+const (
+	Get opType = iota
```
bogdandrutu, line 60: A bit strange that the `opType` is private and `Get` is public. I may miss something, is this helping with not allowing "unknown" types?

extension/storage/storage.go
djaglowski: I don't understand what you're suggesting. Can you elaborate?

Is it necessary to establish the package?

extension/storage/storage.go
djaglowski: Users must be able to specify `Get`, `Set` or `Delete` for `Batch`, but I don't think there's a direct reason to refer to `OpType`. I guess it should be exported though in case they want to write some logic around it? WDYT?

extension/storage/storage.go
djaglowski: I suppose not, but `Get` populates `Value` and expect the user to retrieve from there, so I'll add `Value()` method to compensate then.

extension/storage/storage.go
djaglowski: The problem is that `Get` populates the `Value` field, so we need a pointer there, so we can have all the same or just `Get` different.

extension/storage/storage.go
bogdandrutu: We are using the current storage to implement a "WAL", and for every message we right now write a new entry in the storage. The idea was, can we instead append to an entry or we always have to create a new entry or provide the full value for an entry?

djaglowski: Doesn't `Batch` give you this ability, you just have to choose when to execute it?

djaglowski:
@bogdandrutu, could you take another look at this?

extension/storage/storage.go
dmitryax: Maybe `Result()`?

extension/storage/storage.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Set will store data. The data can be retrieved by the same
+	// component after a process restart, using the same key
+	Set(ctx context.Context, key string, value []byte) error
+
+	// Delete will delete data associated with the specified key
+	Delete(ctx context.Context, key string) error
+
+	// Batch handles specified operations in batch. Get operation results are put in-place
+	Batch(ctx context.Context, ops ...*Operation) error
```
dmitryax, line 51: I didn't check all the implementations in our code base, but are we sure it's enough to provide a list of independent operations? Do we have a use case when one operation depends on another? Did you consider using some sort of a callback instead of the operations? It might complicate the implementations and usage of the interface, but the interface itself will probably be cleaner and more capable. I'm not saying that that's a better option, just wanted to validate other options.

djaglowski:
My goal was to rename a package which appeared to be a misnomer, but there is a lot of good feedback which basically constitutes a redesign. I don't have a complete picture in mind so I'm closing this and perhaps someone else will propose a completely new API.

PR #8768, opened 2023-10-28
[service/extensions] extension lifecycle order

### LINKED ISSUE #8732: Start order of extensions is not guaranteed?

yurishkuro:
**Description**:
Enforce order of start and shutdown of extensions according to their internally declared dependencies

**Link to tracking Issue**:
Resolves #8732

**Motivation**:
This is an alternative approach to #8733 which uses declaration order in the config to start extensions. That approach (a) enforces order when it's not always necessary to enforce, and (b) exposes unnecessary complexity to the user by making them responsible for the order.

This PR instead derives the desired order of extensions based on the dependencies they declare by implementing a `DependentExtension` interface. That means that extensions that must depend on others can expose this interface and be guaranteed to start after their dependencies, while other extensions can be started in arbitrary order (same as happens today because of iterating over a map).

The extensions that have dependencies have two options to expose them:
  1. if the dependency is always static (e.g. `jaeger_query` extension depending on `jaeger_storage` as in the OP), the extension can express this statically as well, by returning a predefined ID of the dependent extension
  2. in cases where dependencies are dynamic, the extension can read the names of the dependencies from its configuration.

The 2nd scenario is illustrated by the following configuration. Here each complex extension knows that it needs  dependencies that implement `storage` and `encoding` interfaces  (both existing APIs in collector & contrib), but does not know statically which instances of those, the actual names are supplied by the user in the configuration.

```yaml
extensions:
  complex_extension_1:
    storage: filestorage
    encoding: otlpencoding
  complex_extension_2:
    storage: dbstorage
    encoding: jsonencoding
  filestorage:
    ...
  dbstorage:
    ...
  otlpencoding:
  jsonencoding:
```

**Changes**:
* Introduce `DependentExtension` optional interface 
* Change `Extensions` constructor to derive the required order using a directed graph (similar to pipelines)
* Inherited from #8733 - use new ordered list of IDs to start/stop/notify extensions in the desired order (previously a map was used to iterate over, which resulted in random order).
* Tests

**Testing**:
Unit tests

### DIFF
--- a/docs/service-extensions.md
+++ b/docs/service-extensions.md
@@ -1,37 +1,37 @@
 # OpenTelemetry Collector: Extensions
 
 Besides the pipeline elements (receivers, processors, and exporters) the Collector
-uses various service extensions (e.g.: healthcheck, z-pages, etc). 
+uses various service extensions (e.g.: healthcheck, z-pages, etc).
 This document describes the “extensions” design and how they are implemented.
 
 ## Configuration and Interface
 
-The configuration follows the same pattern used for pipelines: a base 
-configuration type and the creation of factories to instantiate the extension 
+The configuration follows the same pattern used for pipelines: a base
+configuration type and the creation of factories to instantiate the extension
 objects.
 
-In order to support generic service extensions an interface is defined 
+In order to support generic service extensions an interface is defined
 so the service can interact uniformly with these. At minimum service extensions
-need to implement the interface that covers Start and Shutdown. 
+need to implement the interface that covers Start and Shutdown.
 
-In addition to this base interface there is support to notify extensions when 
-pipelines are “ready” and when they are about to be stopped, i.e.: “not ready” 
-to receive data. These are a necessary addition to allow implementing extensions 
-that indicate to LBs and external systems if the service instance is ready or 
-not to receive data 
-(e.g.: a [k8s readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes)). 
-These state changes are under the control of the service server hosting 
+In addition to this base interface there is support to notify extensions when
+pipelines are “ready” and when they are about to be stopped, i.e.: “not ready”
+to receive data. These are a necessary addition to allow implementing extensions
+that indicate to LBs and external systems if the service instance is ready or
+not to receive data
+(e.g.: a [k8s readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes)).
+These state changes are under the control of the service server hosting
 the extensions.
 
-There are more complex scenarios in which there can be notifications of state 
-changes from the extensions to their host. These more complex cases are not 
+There are more complex scenarios in which there can be notifications of state
+changes from the extensions to their host. These more complex cases are not
 supported at this moment, but this design doesn’t prevent such extensions in the
 future[^1].
 
 
 ## Collector State and Extensions
 
-The diagram below shows the basic state transitions of the OpenTelemetry Collector 
+The diagram below shows the basic state transitions of the OpenTelemetry Collector
 and how it will interact with the service extensions.
 
 ![ServiceLifeCycle](images/design-service-lifecycle.png)
@@ -39,9 +39,9 @@ and how it will interact with the service extensions.
 
 ## Configuration
 
-The config package will be extended to load the service extensions when the 
-configuration is loaded. The settings for service extensions will live in the 
-same configuration file as the pipeline elements. Below is an example of how 
+The config package will be extended to load the service extensions when the
+configuration is loaded. The settings for service extensions will live in the
+same configuration file as the pipeline elements. Below is an example of how
 these sections would look like in the configuration file:
 
 ```yaml
@@ -61,17 +61,18 @@ extensions:
 # The service lists extensions not directly related to data pipelines, but used
 # by the service.
 service:
-  # extensions lists the extensions added to the service. They are started
-  # in the order presented below and stopped in the reverse order.
+  # extensions lists the extensions added to the service. The start-up order respects
+  # any internal dependencies between extensions, but otherwise is arbitrary.
+  # The stop order is always the reverse of the start-up order.
   extensions: [health_check, pprof, zpages]
 ```
 
 The configuration base type does not share any common fields.
 
 The configuration, analogous to pipelines, allows to have multiple extensions of
-the same type. Implementers of extensions need to take care to return error 
+the same type. Implementers of extensions need to take care to return error
 if it can only execute a single instance. (Note: the configuration uses composite
-key names in the form of `type[/name]` 
+key names in the form of `type[/name]`
 as defined in this [this document](https://docs.google.com/document/d/1NeheFG7DmcUYo_h2vLtNRlia9x5wOJMlV4QKEK05FhQ/edit#)).
 
 The factory follows the same pattern established for pipeline configuration:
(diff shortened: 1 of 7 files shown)


extension/extension.go
```diff
@@ -16,6 +16,11 @@ import (
 // to the service, examples: health check endpoint, z-pages, etc.
 type Extension = component.Component
 
+// TODO this can even be DependentComponent, but not sure if there's a use case for that.
+type DependentExtension interface {
+	Dependencies() []component.ID
```
dmitryax, line 21: As I mentioned in https://github.com/open-telemetry/opentelemetry-collector/pull/8733#issuecomment-1783637186, this capability is only applicable for the Jager-specific use case. It doesn't cover the way components depend on extensions in the Collector core/contrib repos.

yurishkuro: Sorry but I don't understand how you see the duck-typing solution work when you have more than one dependency implementing the same interface.

With this PR you can use this config:
```
extensions:
  complex_extension_1:
    storage: filestorage
  complex_extension_2:
    storage: dbstorage
  filestorage:
    ...
  dbstorage:
    ...
```
The "complex" extensions here will implement DependentExtension and return an ID of the other extension they depend on, which they can read from their user-provided config.

dmitryax: > The "complex" extensions here will implement DependentExtension and return an ID of the other extension they depend on, which they can read from their user-provided config.

After this change, the `complex` extensions have to know about every implementation of the storage extension interface: `filestorage` and `dbstorage`. If I have another storage extension in my own distribution of OTel collector, the `complex` won't be able to work with it.

dmitryax: Another simple option is to introduce some kind of priority class for the extensions that would be considered in the order of initialization

yurishkuro: > After this change, the complex extensions have to know about every implementation of the storage extension interface: filestorage and dbstorage. If I have another storage extension in my own distribution of OTel collector, the complex won't be able to work with it.

On the contrary, it does not need to know anything about implementations at all, only the interface they implement. The actual IDs of the dependent extensions are provide by the user in the configuration.

dmitryax: I see what you mean. I missed the fact that user config is already available at this point. Sorry for the confusion. This approach looks good to me.

yurishkuro:
@atoulme @dmitryax can we agree/confirm that we can take the approach in this PR? I am happy to finish it off. Without a resolution for the ordering we're blocked on jaeger-v2 work.

atoulme:
Are we blocked on this comment?  https://github.com/open-telemetry/opentelemetry-collector/pull/8733#discussion_r1378241672

yurishkuro:
a) I don't know if the comment right or not or if the order even matters for notify
b) that comment is a minor issue, the main question is about the approach itself: https://github.com/open-telemetry/opentelemetry-collector/pull/8733#issuecomment-1784329406

atoulme:
OK so let's keep ourselves nimble and ready for merge - I'll remove the comment. You can do the same in your PR. Dmitrii and other maintainers can make a call on the approach.

atoulme:
Feel free to pick up my commit 84826a34f or apply the diff on top, either works. If you want this PR to be considered for merge, can you up the coverage enough to pass coverage? Remove WIP and draft and let's make this convincing so maintainers can weigh in.

atoulme:
There is a `discussion-needed` label on this PR - do you want to talk it out tomorrow during the SIG meeting?

yurishkuro:
> do you want to talk it out tomorrow during the SIG meeting?

@atoulme I won't be able to make it. I updated the description with details and motivation.

docs/service-extensions.md
```diff
@@ -61,17 +61,18 @@ extensions:
 # The service lists extensions not directly related to data pipelines, but used
 # by the service.
 service:
-  # extensions lists the extensions added to the service. They are started
-  # in the order presented below and stopped in the reverse order.
+  # extensions lists the extensions added to the service. The start-up order respects
+  # any internal dependencies between extensions, but otherwise is arbitrary.
+  # The stop order is always the reverse of the start-up order.
   extensions: [health_check, pprof, zpages]
```
yurishkuro, line 67: here is the actual change, ignore whitespace auto-fixes

dmitryax:
I'm good with this approach 👍

extension/extension.go
```diff
@@ -16,6 +16,13 @@ import (
 // to the service, examples: health check endpoint, z-pages, etc.
 type Extension = component.Component
 
+// DependentExtension is an optional interface that can be implemented by extensions
+// that depend on other extensions and must be started only after their dependencies.
+// See https://github.com/open-telemetry/opentelemetry-collector/pull/8768 for examples.
+type DependentExtension interface {
```
dmitryax, line 22: I don't think we need `Extension` suffix with the given package name, `extension.Dependent` should be enough. We usually follow this practice in this codebase. Also, we should include `extension.Extension` to avoid misusing it by the clients.

djaglowski (commented):
I think this is a great solution. The optional interface is an elegant way to manage ordering concerns.

service/extensions/graph.go
```diff
@@ -0,0 +1,58 @@
...
+					graph.SetEdge(graph.NewEdge(d, n))
+				} else {
+					return nil, fmt.Errorf("unable to find extension %s on which extension %s depends", depID, extID)
+				}
+			}
+		}
+	}
+	orderedNodes, err := topo.Sort(graph)
+	if err != nil {
+		return nil, fmt.Errorf("unable to sort the extenions dependency graph: %w", err)
```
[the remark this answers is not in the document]
yurishkuro: That's a lot of code to copy - is there a good place to share it?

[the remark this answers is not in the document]
djaglowski: Export a shared function in `service/internal/graph` would make sense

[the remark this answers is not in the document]
yurishkuro: turns out that function depends on private struct as a type of node, so I add a simpler version here.

extension/extension.go
```diff
@@ -16,6 +16,13 @@ import (
 // to the service, examples: health check endpoint, z-pages, etc.
 type Extension = component.Component
 
+// Dependent is an optional interface that can be implemented by extensions
+// that depend on other extensions and must be started only after their dependencies.
+// See https://github.com/open-telemetry/opentelemetry-collector/pull/8768 for examples.
+type Dependent interface {
+	Dependencies() []component.ID
```
dmitryax, line 23: As I mentioned in another comment, I think we should add the `Extension` interface to avoid misuse. We do this for other extension types
```suggestion
	Extension
	Dependencies() []component.ID
```

yurishkuro: Sorry, I forgot to add it. Done

dmitryax (approved):
LGTM

yurishkuro:
🎉

PR #8788, opened 2023-11-01
Manually report status for core exporters

mwear:
**Description:**
This is a PR is an alternative to #8684. It adds manual status reporting to the otlp, otlphttp, and debug exporters. Reviewers, please compare with #8684 and weigh in on what approach you prefer, and/or additional suggestions.

**Testing:** Units, manual.

### DIFF
--- a/.chloggen/core-exporters-manual-status.yaml
+++ b/.chloggen/core-exporters-manual-status.yaml
@@ -7,7 +7,7 @@ change_type: 'enhancement'
 component: core exporters
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: This PR adds status reporting to the core exporters (otlp, otlphttp, and debug)
+note: This PR adds status reporting to the core exporters (otlp and otlphttp)
 
 # One or more tracking issues or pull requests related to the change
 issues: [7682]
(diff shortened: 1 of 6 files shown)


codeboten (commented):
Looking at this in comparison with https://github.com/open-telemetry/opentelemetry-collector/pull/8684, I like the approach in the other PR better, it feels like the opportunities for most of the cases is covered automatically for components, while still giving enough flexibility for edge cases.

With the approach in this PR, it feels like most of the code will end up being quite repetitive across components. I'm thinking of how much code this would mean for the contrib repo especially.

mwear:
If I were only reporting status for exporters, I would choose #8684 because it introduces a flexible opt-in approach that reduces the overall amount of code that needs to be written. My only reservation is that that approach wouldn't apply fully to other component types. We could add similar `WithStatusReporting` options to the `processorhelper` and `receiverhelper`, but they would likely only allow you to wrap start. Receivers are not consumers. While processors are consumers, they are in the middle of a pipeline, and errors returned from consume do not necessarily belong to the processor and might be propagated from further down the pipeline. That is to say, status reporting for receivers and processors will be more manual, so it might make sense to do things manually for exporters so they aren't a special case.

Issue #8343, opened 2023-08-31
Consider removing `memory_ballast` extension


mx-psi:
The `memory_ballast` extension allows defining a 'heap ballast' to make the behavior of Go's garbage collector more predictable. Quoting [this post](https://web.archive.org/web/20210929130001/https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap-26c2462549a2/),
>  the ballast increases the base size of the heap so that our GC triggers are delayed and the number of GC cycles over time is reduced.

This functionality was added in #45 but a few years and Go versions have passed since and there has since been discussion (e.g. see https://github.com/open-telemetry/opentelemetry-collector/issues/7512#issuecomment-1499605455 ) on the need of this extension today. In particular, tweaking `GOMEMLIMIT` and `GOGC` may be enough to tackle the issues the extension was meant to address.
It also seems like the extension [may be risky](https://groups.google.com/g/golang-nuts/c/66d0cItfkjY/m/3NvgzL_sAgAJ) in some cases and increase memory usage (this may be the cause of #7512).

The most recent issue on the matter upstream is golang/go/issues/42430.

This issue is to decide whether we should remove the memory ballast extension or not.

cc @Aneurysm9 @dmitryax

dmitryax:
@mx-psi thank you for summarizing the issue.

I don't see any good reason to keep this extension. `GOMEMLIMIT` env var is much more intuitive to use. The only configuration that the user can potentially miss is `size_in_percentage`, but I doubt it's something we should keep this extension for.

I'd suggest we just go ahead and deprecate it with clear instructions on how to utilize `GOMEMLIMIT` instead.

Let's discuss it in the next SIG meeting.

mx-psi:
One starting point to ensure we can do this correctly is to replace the memory ballast extension by an equivalent mechanism using `GOMEMLIMIT` and `GOGC` in the Collector Helm chart and ensure that it works correctly in that setting. We can use that experience to aid us in writing the documentation.

MovieStoreGuy:
From Atlassian's experience, we've not needed the ballast extension for a deployment that is strictly observing the host (doesn't expose any endpoints for services to send data, and prom is used to monitor the collector itself).

On average, our deployments will use:

- 50Mb of memory with a hard limit 300Mb (set by docker aka cgroups)
- 3% CPU utilisation with a hard limit of 1 full core. 

[...]

Issue #8342, opened 2023-08-31
Change `process_runtime_heap_alloc_bytes` metric to include memory ballast size


mx-psi:
The [`process/runtime/heap_alloc_bytes`][1]/[`process_runtime_heap_alloc_bytes`][2] metric is defined as 
> Bytes of allocated heap objects (see 'go doc runtime.MemStats.HeapAlloc')

but, since #45, the actual value is `runtime.MemStats.HeapAlloc` minus the memory ballast size if the `memory_ballast` extension is being used:

https://github.com/open-telemetry/opentelemetry-collector/blob/7030388dde3f765af60cf8431ea54340194d9f18/service/internal/proctelemetry/process_telemetry.go#L283-L293

This delays the process metrics initialization until extensions are running: https://github.com/open-telemetry/opentelemetry-collector/blob/7030388dde3f765af60cf8431ea54340194d9f18/service/service.go#L223

We should remove this special case and change the metric to report the raw value of `runtime.MemStats.HeapAlloc` instead, which will simplify both the metric's semantics as well as the implementation of telemetry initialization. If we keep using the memory ballast extension (see the separate issue #8343 for discussion on this, which may happen after this change), we can add a metric to the `memory_ballast` extension to indicate the memory ballast size and be able to recover the previous value reported by this metric.

If we find out it is not desirable to just report `runtime.MemStats.HeapAlloc`, we should instead change the metric description to state that the memory ballast value is removed.

[1]: https://github.com/open-telemetry/opentelemetry-collector/blob/7030388dde3f765af60cf8431ea54340194d9f18/service/internal/proctelemetry/process_telemetry.go#L119-L122
[2]: https://github.com/open-telemetry/opentelemetry-collector/blob/7030388dde3f765af60cf8431ea54340194d9f18/service/internal/proctelemetry/process_telemetry.go#L186-L193

mx-psi:
> This change would make sense to me. How do you propose introducing the change? I'd expect anyone that monitors using this value today will be surprised to see that value change by the ballast size after a release.

We can add the `memory_ballast` metric without breaking anyone. For the `process_runtime_heap_alloc_bytes`, we can do the change through a feature gate to allow users to migrate at their own pace.

codeboten:
Sounds good. Pinging @open-telemetry/collector-approvers for more eyes on this

dmitryax:
[...]

PR #8803, opened 2023-11-06
[extension/ballastextension] Deprecate memory_ballast extension

mx-psi:
**Description:** 

Based on user reports on open-telemetry/opentelemetry-helm-charts/issues/891 and the discussion on #8343, we can deprecate the memory ballast extension in favor of using `GOMEMLIMIT`. This PR:

- Deprecates the memory ballast extension in the README
- Removes references to the memory ballast extension on docs
- Updates k8s example to use `GOMEMLIMIT` with the same approach as in the Helm chart (80% of memory limit)
- Deprecates the memory ballast extension Go module

Once this PR is accepted, open-telemetry/opentelemetry-helm-charts/issues/891 can move ahead with enabling `useGOMEMLIMIT` by default on the Helm chart.

Other issues will be opened for opentelemetry.io, the Opentelemetry Operator and other parts of the OpenTelemetry project to remove references to this extension once the PR is merged.

No explicit timeline is given for removal of the extension.

**Link to tracking Issue:** Updates #8343


mx-psi:
@dmitryax @Aneurysm9 @TylerHelmuth @JaredTan95

dloucasfx (commented):
Thanks @mx-psi for pushing this and getting us closer to GOMEMLIMIT !

Not sure if this was already discussed, the ballast extension has been used in conjunction with the memory-limiter processor to control GC; 

With `GOMEMLIMIT` do we need to change the behavior of memory-limiter ? 
The memory-limiter currently provides 2 main tasks:
- signal the pipeline to drop or stop accepting new data
- call GC explicitly

With `GOMEMLIMIT` we should no longer worry about explicitly calling GC anymore, however, we might still need to keep the drop or stop accepting data? 
I also understand that @dmitryax suggested to move the memory-limiter as a receiver helper.

extension/ballastextension/README.md
```diff
@@ -1,8 +1,12 @@
+> [!WARNING]
+> The memory ballast extension is deprecated in favor of using the `GOMEMLIMIT` environment variable.
+> This environment variable is available on any Collector built with Go 1.19 or higher. Official binary releases are built with Go 1.19 since v0.61.0. See [issue 8343](https://github.com/open-telemetry/opentelemetry-collector/issues/8343) for the deprecation timeline.
```
codeboten, line 3: Should this include a recommendation of what a user should configure this based on their previous setting?

bogdandrutu:
@mx-psi would be good to look into the memory limiter as well.

mx-psi:
@bogdandrutu @dloucasfx There are a couple dependencies on the memory ballast in other parts of the Collector:
- Issue #8632 summarizes the work needed to get rid of the memory limiter and moving to receivers
- Internal metrics also have a dependency on the memory ballast, which is summarized on #8342

IMO, we should first
1. deprecate the memory ballast
2. update all our documentation and examples to stop mentioning the memory ballast
3. update the Helm chart, operator and Docker default configuration to use GOMEMLIMIT
4. (can happen in parallel with 2 & 3) wait for a couple of releases

After these three happen, we can think about changing the behavior of other parts of the Collector that depend on the memory ballast, including the memory limiter (if the receiver part) and the metrics. Does that make sense? Do you think there is specific parts we should do now as part of this PR?

mx-psi:
@bogdandrutu @dloucasfx can you take another look?

mx-psi:
Looks like we are first going to enable `GOMEMLIMIT` by default in the Helm Chart for a couple of versions before doing this, you can follow along at open-telemetry/opentelemetry-helm-charts/issues/891

bogdandrutu (approved):

Issue #4947, opened 2022-03-02
Collector Shutdown should block until Run cleans up


cpheps:
**Is your feature request related to a problem? Please describe.**
Discussion from [4878](https://github.com/open-telemetry/opentelemetry-collector/pull/4878#issuecomment-1055438081) led to a decision to design `Shutdown` to follow language patterns for how it and `Run` should operate. Following `http.Server` as an example the following behaviors should be met:

1. `Shutdown` be safe to be called at any point. Even if there is no action to take
2. If `Shutdown` is called before `Run` it will exit cleanly and the next call to `Run` will exit due to the previous `Shutdown`
3. If `Shutdown` is called while `Run` is active it should block until `Run` has confirmed shutdown.

Point 3 isn't currently implemented but desired behavior. 

**Describe the solution you'd like**
`Shutdown` will block until `Run` has confirmed cleanup. Whatever `Shutdown` checks for should not stop it from cleanly exiting if `Run` hasn't been called yet or after `Run` was called.

**Describe alternatives you've considered**
Alternative is don't block `Shutdown` and just exit like it does now after closing the channel.

**Additional context**
Action item from #4878 

If we want to go this route I think there needs to be some discussion on what `Shutdown` should block on. It seems like it should be when the state is `Closing` but that won't allow `Shutdown` to cleanly exit if called before or after `Run`.

dmitryax:
@djaglowski said it can be upstreamed

djaglowski:
Following up on this - I misunderstood the context of this issue, #4946, and 4939, and mistakenly believed they may have been resolved as part of #4878.

deepto98:
Can I work on this?

dmitryax:
@deepto98 sure. Assigned to you. Thanks

Chinwendu20:
Hi @deepto98, are you still working on this? I would like to take this up

Chinwendu20:
Hi @cpheps I guess 1 and 2 has been addressed by other the other issues

michalpristas:
going to take a look at this

dridi-walid:
Hi, I wanted to check if point 3 (Shutdown blocking until Run confirms cleanup) has been addressed in any issue or PR. If not, I’d like to work on implementing this behavior. Please let me know!

dmitryax:
@dridi-walid i don't think it was ever addressed. Feel free to look into that

david-sevcik:
[...]

PR #8811, opened 2023-11-07
Synchronized Run and Shutdown for otelcol/Collector

michalpristas:
**Description:** 
Adding some synchronization for `Run` and `Shutdown` commands in `otelcol/collector.go`
Should deal with all 3 requirements from the issue:

1. `Shutdown` be safe to be called at any point. Even if there is no action to take
2. If `Shutdown` is called before `Run` it will exit cleanly and the next call to `Run` will exit due to the previous `Shutdown`
3. If `Shutdown` is called while `Run` is active it should block until `Run` has confirmed shutdown.

I went with waitgroup approach because i could think of 3 options how Run and Shutdown can be called.
- `Shutdown` before `Run`: no need to wait for confirmation pipelines not set up
- `Shutdown` during `Run` we wait for all run to finish work
- `Run` finished before `Shutdown` No need to wait as nobody will handle shutdown signal until next `Run` is called for same collector. Also in this case i'm not sure it's viable case. It looks like Run on exit stops the service.

From these situations it looks that it's enough to wait for Runs to finish.
Other way would be pulling state of Config provider and Service until we have a Shutdown match. This could possibly block in case of no Run command running with the same effect as WaitGroup approach but it would take longer. On top of that we would need to have 2 or 3 states for each of these components (not started[optional], running, stopped)

**Link to tracking Issue:** #4947 

**Testing:** Unit tests

As for incoming release, I'm raising this mainly for discussion and I'll keep it up to date until it's possible to merge


michalpristas:
hey @bogdandrutu any feedback would be appreciated. does direction took seem ok?

otelcol/collector.go
```diff
@@ -226,6 +231,16 @@ func (col *Collector) DryRun(ctx context.Context) error {
 // Run starts the collector according to the given configuration, and waits for it to complete.
 // Consecutive calls to Run are not allowed, Run shouldn't be called once a collector is shut down.
 func (col *Collector) Run(ctx context.Context) error {
+	col.runWg.Add(1)
+	defer col.runWg.Done()
```
bogdandrutu, line 235: I think you create a deadlock here https://github.com/open-telemetry/opentelemetry-collector/pull/8811/files#diff-3c058aa5dde6d9b2b413b65dce28fe97e9cf859a0f03629639843aac6bc26e2aR286

michalpristas: havent run into a deadlock but i will for sure verify

michalpristas: yeah i think that's right. will need to figure something else

michalpristas: still looking at it at first i thought `Shutdown` is called within `Run` but this is not the case so this deadlock is not there.
i'm drawing some flow charts and it should be fine from what i'm seeing. @bogdandrutu  can you elaborate on where you think the deadlock is?

changes are only to exported funcs which are not calling each other
- when Runs are running Shutdown waits for them
- when they are not waitgroup does not block
- and each run should finish as `.shutdown` is triggered from Shutdown which escapes run loop and decrement waitgroup counter.

Issue #7625, opened 2023-05-03
[featuregate] Include the description, fromVersion, toVersion, and reference URL in String()


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**
Although we set description, fromVersion, toVersion, and reference url in the gates today, users cannot see those values unless they use zpagesextension.  Only the default value based on the stage is included in the flag's `String()` implementation.  As a result, users can see the available flags when using the `--help` command, but not when the flags represent. This can cause confusion and extra burden on feature gate users to ensure the gate is documented somewhere else.

**Describe the solution you'd like**
Each feature gate's description, toVersion, fromVersion, and referenceURL is included in string so that users know what they can use.

**Describe alternatives you've considered**
zpagesextension can display all this information, but the collector must have that extension and be running in order to be useful.

mx-psi:
Should this block featuregate 1.0? I don't think so, since the `String` method should not have any guarantees as to what its output is other than "human readable description of the feature gate".

Issue #7095, opened 2023-02-02
[pdata] Find replacement for now deprecated gogo/protobuf


yurishkuro:
I don't know the full background of using [gogo/protobuf](https://github.com/gogo/protobuf), I suspect it was similar to Jaeger's use of gogoproto for serialization efficiency (in particular, the ability to embed structs by value, as in `[]KeyValue` instead of `[]*KeyValue`, which significantly reduces memory allocations).

Well, the bad news, which was long coming, is that GoGo is officially deprecated since Oct'22. Continue to depend on it going forward presents security and compatibility risks.

This ticket is to kick off a discussion for a viable path forward, since I don't have a solution in mind right now.

Some past related tickets:
  * https://github.com/open-telemetry/opentelemetry-collector/issues/4258

yurishkuro:
As far as replacements, there are other projects that seem to provide similar capabilities to what Gogo had:
  * https://github.com/planetscale/vtprotobuf
  * https://github.com/CrowdStrike/csproto

At minimum, they seem to be maintained at the moment, and hopefully are more narrow in scope than gogo.

jpkrohling:
What is Jaeger using nowadays? Any experience you can share with us?

yurishkuro:
Jaeger hasn't changed from gogo/proto

codeboten:
I've not tried using `csproto`, but there were a few investigations I'd done in the past including trying to use `vtproto` with the standard google protobuf library here: https://github.com/open-telemetry/opentelemetry-collector/pull/4658

It's probably worth another go at it now that it's been over a year.

yurishkuro:
@codeboten I saw that PR, but can you share some findings / conclusions? It was closed without any explanation, so not clear if you ran into a dead end with something.

codeboten:
The scope of the work then was to add support in the collector for `optional` fields which was achieved through a workaround (https://github.com/open-telemetry/opentelemetry-collector/issues/4258#issuecomment-1049379599). That was the reason this PR was closed.

From what I recall the performance numbers were not encouraging at first. Additionally I ran into issues w/ lack of support for`oneof` fields.

yurishkuro:
https://go.dev/blog/protobuf-opaque looks very promising as a path towards removing gogo/protobuf

tigrannajaryan:
> https://go.dev/blog/protobuf-opaque looks very promising as a path towards removing gogo/protobuf

Yes, let's see if performance is on par.

[...]

PR #8812, opened 2023-11-07
[VERSIONING.md] Reorganize and clarify compatibility guarantees

### LINKED ISSUE #8002: Document Go API stability guarantees
To start marking more modules as stable, we want to discuss the following open questions about configuration structs:

1. Does adding new validation rules mean a breaking change?
2. Should we offer a "NewDefault...." for each config and say that the behavior of the config is stable only if initialized with NewDefault?
3. Is adding new fields means breaking change? Let's assume someone "squash" the message into another message, then adding a field with same mapstruct value will be a breaking change, what do we do with that? What are the rules we follow.

We generally also need to think of any special implications for configs that are not defined by the golang conventions of breaking compatibility.

_Reworded from post by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/issues/7954#issuecomment-1601902350_

mx-psi:
**Description:** Reorganize `VERSIONING.md`, specify compatibility guarantees that have already been discussed and answer questions on #8002.

Three new guarantees in the document had already been discussed elsewhere:

- String representation: https://github.com/open-telemetry/opentelemetry-collector/issues/7625#issuecomment-1674789807
- Go version compatibility: Mentioned [on README.md](https://github.com/open-telemetry/opentelemetry-collector?tab=readme-ov-file#compatibility)
- Dependency updates: Discussed in private, most recently in relation to whether #7095 should block `pdata`'s 1.0 (consensus seems to be that it should not).

Regarding the questions mentioned on #8002:

> Does adding new validation rules mean a breaking change?

Generally, yes except when the configuration was already invalid (i.e. the Collector did not work properly with it).

> Should we offer a "NewDefault...." for each config and say that the behavior of the config is stable only if initialized with NewDefault?

I did not add anything for this one. I think we should discuss it, but:
- I don't think there is a sensible output for `NewDefault...` for all configurations (e.g. for `confignet.TCPAddr`, what would the default be?)
- It seems to me that we should handle configuration validity through `Validate`, and not through `NewDefault...`. `NewDefault` may help but ultimately we need to verify through `Validate`.

> Is adding new fields means breaking change? Let's assume someone "squash" the message into another message, then adding a field with same mapstruct value will be a breaking change, what do we do with that? What are the rules we follow.

This was already in this document, where we had decided that adding new fields was okay. I think it would be too stringent to bar us from adding new fields.

**Link to tracking Issue:** Fixes #8002


VERSIONING.md
```diff
@@ -72,14 +107,8 @@ is designed so that the following goal can be achieved:
           whenever you are using the module name).
         * If a module is version `v0` or `v1`, do not include the major version
           in either the module path or the import path.
-    * Configuration structures should be considered part of the public API and backward
-      compatibility maintained through any changes made to configuration structures.
-        * Because configuration structures are typically instantiated through unmarshalling
-          a serialized representation of the structure, and not through structure literals,
-          additive changes to the set of exported fields in a configuration structure are
-          not considered to break backward compatibility.
     * Modules will be used to encapsulate receivers, processor, exporters,
```
songy23, line 110:
```suggestion
    * Modules will be used to encapsulate receivers, processors, exporters,
```

VERSIONING.md
```diff
@@ -5,7 +5,52 @@ is designed so that the following goal can be achieved:
...
+  break the struct by creating a conflict with an existing method of the other embedded type. We cannot protect 
+  against this rare case and do not guarantee compatibility in such scenarios.
+* **Dot imports.** If a program imports a package using `import .`, additional names defined in the imported package
+  in future releases may conflict with other names defined in the program. We do not recommend the use of
+  `import .` with OpenTelemetry Collector modules.
+
+Unless otherwise specified in the documentation, the following may change in any way between minor versions:
+* **String representation**. The `String` method of any struct is intended to be human-readable and may change its output
+  in any way.
+* **Go version compatibility**. Removing support for an unsupported Go version is not considered a breaking change.
```
songy23, line 31: How about OS compatibility?

VERSIONING.md
```diff
@@ -5,7 +5,52 @@ is designed so that the following goal can be achieved:
...
+Unless otherwise specified in the documentation, the following may change in any way between minor versions:
+* **Adding new fields to configuration structures**. Because configuration structures are typically instantiated through 
+unmarshalling a serialized representation of the structure, and not through structure literals, additive changes to 
+the set of exported fields in a configuration structure are not considered to break backward compatibility.
+
+The following are explicitly considered to be breaking changes:
+* **Modifying struct tags related to serialization**. Struct tags used to configure serialization mechanisms (`yaml:`, 
+`mapstructure:`, etc) are part of the structure definition and must maintain compatibility to the same extent as the 
+structure.
+* **Making validation rules more strict**. A valid configuration struct as defined by its `Validate` method return value
```
songy23, line 49: Do you want to mention the opposite i.e. relaxing validation rules is considered backwards-compatible?

VERSIONING.md
mx-psi: Added a bullet point for this

VERSIONING.md
mx-psi: Added it, this one seemed a bit more obvious to me, but I guess it doesn't hurt to add it

songy23 (approved):
LGTM

atoulme (approved):

VERSIONING.md
```diff
@@ -5,7 +5,55 @@ is designed so that the following goal can be achieved:
...
+  break the struct by creating a conflict with an existing method of the other embedded type. We cannot protect 
+  against this rare case and do not guarantee compatibility in such scenarios.
+* **Dot imports.** If a program imports a package using `import .`, additional names defined in the imported package
+  in future releases may conflict with other names defined in the program. We do not recommend the use of
+  `import .` with OpenTelemetry Collector modules.
+
+Unless otherwise specified in the documentation, the following may change in any way between minor versions:
+* **String representation**. The `String` method of any struct is intended to be human-readable and may change its output
+  in any way.
+* **Go version compatibility**. Removing support for an unsupported Go version is not considered a breaking change.
```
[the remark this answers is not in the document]
mx-psi: Supported by the Go team. This is mostly a rephrasing of what is already on the README https://github.com/open-telemetry/opentelemetry-collector?tab=readme-ov-file#compatibility

TylerHelmuth (approved):

dmitryax (approved):
LGTM. Thanks for putting this together.

PR #8827, opened 2023-11-08
[chore] [confighttp] Deprecate HTTPClientSettings.CustomRoundTripper

### LINKED ISSUE #8627: [confmap] Should not marshal functions/channels by default
**Describe the bug**
`mapstructure` will take any exported field and attempt to decode it. This includes non-tagged fields.

> When decoding to a struct, mapstructure will use the field name by default to perform the mapping. For example, if a struct has a field "Username" then mapstructure will look for a key in the source value of "username" (case insensitive).
> ```go
>type User struct {
>    Username string
>}
> ```

https://pkg.go.dev/github.com/mitchellh/mapstructure

Similarly, the `mapstructure` encoder will also take any exported field and attempt to encode it into a form that `go-yaml` can use. The encoder does not handle function or channel types. Any exported function/channel which will use the default handler, which will fall back on leaving the field as is. `go-yaml` [cannot handle the function/channel](https://github.com/go-yaml/yaml/blob/v3/encode.go#L182) and panics.

One example of an exported function in the configuration is the `confighttp.HTTPClientSettings`, which exports the CustomRoundTripper function:
https://github.com/open-telemetry/opentelemetry-collector/blob/b5635a7a90d204453915db36e4e29182ee2dddb2/config/confighttp/confighttp.go#L52-L53

**Steps to reproduce**
```go
func TestMarshal(t *testing.T) {
    settings := confighttp.NewDefaultHTTPClientSettings()
    conf := confmap.New()
    assert.NoError(t, conf.Marshal(settings))
    out, err := yaml.Marshal(conf.ToStringMap())
    assert.NoError(t, err)
    t.Log(string(out))
}
```

That the function would be skipped. `customroundtripper` is not present in the output YAML.

```yaml
        auth: null
        compression: ""
        disable_keep_alives: false
        endpoint: ""
        headers: {}
        idle_conn_timeout: 1m30s
        max_conns_per_host: null
        max_idle_conns: 100
        max_idle_conns_per_host: null
        read_buffer_size: 0
        timeout: 0s
        tls:
            ca_file: ""
            ca_pem: '[REDACTED]'
            cert_file: ""
            cert_pem: '[REDACTED]'
            insecure: false
            insecure_skip_verify: false
            key_file: ""
            key_pem: '[REDACTED]'
            max_version: ""
            min_version: ""
            reload_interval: 0s
            server_name_override: ""
        write_buffer_size: 0
```

**What did you see instead?**
```
panic: cannot marshal type: func(http.RoundTripper) (http.RoundTripper, error)

panic({0x1779440, 0xc0003ef630})
	/path/to/go/src/runtime/panic.go:884 +0x213
gopkg.in/yaml%2ev3.(*encoder).marshal(0x179b500?, {0x0, 0x0}, {0x17988c0?, 0x0?, 0x10?})
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:182 +0xa4b
gopkg.in/yaml%2ev3.(*encoder).marshal(0x17b13e0?, {0x0, 0x0}, {0x179b500?, 0xc0003ef620?, 0x98?})
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:162 +0x85b
gopkg.in/yaml%2ev3.(*encoder).mapv.func1()
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:192 +0xf4
gopkg.in/yaml%2ev3.(*encoder).mappingv(0xc0003e4400, {0x0?, 0x0}, 0xc0001e1890)
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:265 +0x146
gopkg.in/yaml%2ev3.(*encoder).mapv(0x17b13e0?, {0x0?, 0xc0003d6960?}, {0x17b13e0?, 0xc0003d6960?, 0x0?})
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:187 +0x5f
gopkg.in/yaml%2ev3.(*encoder).marshal(0xc0003e4400?, {0x0, 0x0}, {0x17b13e0?, 0xc0003d6960?, 0xc0001e1a60?})
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:164 +0x88b
gopkg.in/yaml%2ev3.(*encoder).marshalDoc(0xc0003e4400, {0x0, 0x0}, {0x17b13e0?, 0xc0003d6960?, 0x1021445?})
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/encode.go:105 +0x12b
gopkg.in/yaml%2ev3.Marshal({0x17b13e0?, 0xc0003d6960})
	/path/to/go/pkg/mod/gopkg.in/yaml.v3@v3.0.1/yaml.go:222 +0x389
go.opentelemetry.io/collector/config/confighttp.TestMarshal(0xc00019d520)
```

**What version did you use?**
Version: v0.86.0

**What config did you use?**
Config: N/A

**Environment**
OS: (e.g., "Ubuntu 20.04") macOS
Compiler(if manually compiled): Go 1.20.8

**Additional context**

haoqixu:
**Description:**
`HTTPClientSettings.CustomRoundTripper` is not an end-user option and may cause failure when marshaling/unmarshaling config.

This PR adds an optional argument to `ToClient()` and deprecates the `HTTPClientSettings.CustomRoundTripper` field.

**Link to tracking Issue:**
resolve #8627

**Testing:**
add `TestHTTPClientSettingWithRoundTripper()` to test the new approach with cases in `TestAllHTTPClientSettings()`.

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -54,6 +54,7 @@ type HTTPClientSettings struct {
 	Headers map[string]configopaque.String `mapstructure:"headers"`
 
 	// Custom Round Tripper to allow for individual components to intercept HTTP requests
+	// Deprecated: [0.89.0] Replaced by confighttp.WithRoundTripper
 	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
 
 	// Auth configuration for outgoing HTTP calls.
@@ -103,8 +104,31 @@ func NewDefaultHTTPClientSettings() HTTPClientSettings {
 	}
 }
 
+// toClientOptions has options that change the behavior of the HTTP client
+// returned by HTTPClientSettings.ToClient().
+type toClientOptions struct {
+	customRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
+}
+
+// ToClientOption is an option to change the behavior of the HTTP client
+// returned by HTTPClientSettings.ToClient().
+type ToClientOption func(opts *toClientOptions)
+
+// WithRoundTripper allow for individual components to intercept HTTP requests
+func WithRoundTripper(customRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)) ToClientOption {
+	return func(opts *toClientOptions) {
+		opts.customRoundTripper = customRoundTripper
+	}
+}
+
 // ToClient creates an HTTP client.
-func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.TelemetrySettings) (*http.Client, error) {
+func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.TelemetrySettings, opts ...ToClientOption) (*http.Client, error) {
+
+	clientOpts := &toClientOptions{}
+	for _, o := range opts {
+		o(clientOpts)
+	}
+
 	tlsCfg, err := hcs.TLSSetting.LoadTLSConfig()
 	if err != nil {
 		return nil, err
@@ -196,7 +220,14 @@ func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.
 	}
 
 	if hcs.CustomRoundTripper != nil {
+		settings.Logger.Warn("You are using a deprecated `CustomRoundTripper` field which will be removed soon; use `WithRoundTripper` instead")
 		clientTransport, err = hcs.CustomRoundTripper(clientTransport)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if clientOpts.customRoundTripper != nil {
+		clientTransport, err = clientOpts.customRoundTripper(clientTransport)
 		if err != nil {
 			return nil, err
 		}
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -55,7 +55,7 @@ type HTTPClientSettings struct {
 
 	// Custom Round Tripper to allow for individual components to intercept HTTP requests
 	// Deprecated: [0.89.0] Replaced by confighttp.WithRoundTripper
-	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
+	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error) `mapstructure:"-"`
 
 	// Auth configuration for outgoing HTTP calls.
 	Auth *configauth.Authentication `mapstructure:"auth"`
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -112,21 +112,29 @@ type toClientOptions struct {
 
 // ToClientOption is an option to change the behavior of the HTTP client
 // returned by HTTPClientSettings.ToClient().
-type ToClientOption func(opts *toClientOptions)
+type ToClientOption interface {
+	apply(*toClientOptions)
+}
+
+type toClientOptionFunc func(*toClientOptions)
+
+func (fn toClientOptionFunc) apply(opts *toClientOptions) {
+	fn(opts)
+}
 
 // WithRoundTripper allow for individual components to intercept HTTP requests
 func WithRoundTripper(customRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)) ToClientOption {
-	return func(opts *toClientOptions) {
+	return toClientOptionFunc(func(opts *toClientOptions) {
 		opts.customRoundTripper = customRoundTripper
-	}
+	})
 }
 
 // ToClient creates an HTTP client.
 func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.TelemetrySettings, opts ...ToClientOption) (*http.Client, error) {
 
 	clientOpts := &toClientOptions{}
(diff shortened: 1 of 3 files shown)


config/confighttp/confighttp.go
```diff
@@ -50,6 +50,7 @@ type HTTPClientSettings struct {
 	Headers map[string]configopaque.String `mapstructure:"headers"`
 
 	// Custom Round Tripper to allow for individual components to intercept HTTP requests
+	// Deprecated: [0.89.0] Replaced by confighttp.WithRoundTripper
 	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
```
bogdandrutu, line 54: You can also set a `mapstructure:"-"`.

config/confighttp/confighttp.go
```diff
@@ -99,8 +100,31 @@ func NewDefaultHTTPClientSettings() HTTPClientSettings {
...
 
+// toClientOptions has options that change the behavior of the HTTP client
+// returned by HTTPClientSettings.ToClient().
+type toClientOptions struct {
+	customRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
+}
+
+// ToClientOption is an option to change the behavior of the HTTP client
+// returned by HTTPClientSettings.ToClient().
+type ToClientOption func(opts *toClientOptions)
```
bogdandrutu, line 111: I think we decided to go with the model of having Options as interface with a private func `apply`

haoqixu: I also updated `ToServerOption` to make the code consistent.

config/confighttp/confighttp_test.go
```diff
@@ -644,7 +759,7 @@ func TestHttpReception(t *testing.T) {
 					return rt, nil
 				}
 			}
-			client, errClient := hcs.ToClient(componenttest.NewNopHost(), component.TelemetrySettings{})
+			client, errClient := hcs.ToClient(componenttest.NewNopHost(), component.TelemetrySettings{Logger: zap.NewNop()})
```
bogdandrutu, line 762: Why not using `componenttest.NewNopTelemetrySettings()`?

config/confighttp/confighttp_test.go
```diff
@@ -156,6 +156,121 @@ func TestAllHTTPClientSettings(t *testing.T) {
...
+			},
+			customRoundTripper: func(next http.RoundTripper) (http.RoundTripper, error) { return nil, errors.New("error") },
+			shouldError:        true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			tt := componenttest.NewNopTelemetrySettings()
+			tt.TracerProvider = nil
```
bogdandrutu, line 251: why?

haoqixu: This is actually copied from `TestAllHTTPClientSettings()`. There are several `tt.TracerProvider = nil` lines in the unittests. I have dug through the Git history and think it's used to avoid the client's transport being wrapped in `otelhttp.Transport`. 🤔

haoqixu: This is required by some testcases which need to check or modify the fields of `http.Transport`. This line is unnecessary here.

config/confighttp/confighttp_test.go
haoqixu: The reason is same as the `tt.TracerProvider = nil`. This tests needs to modify the fields of `http.Transport`. I think I can update the tests to make them more consistent.

haoqixu: Do we need to provide a way for those usecases to access the underlying`http.Transport`? Omitting `TracerProvider` or `MetircsProvider` to access `http.Transport` in the `CustomRoundTripper` seems depending on implementation details of `ToClient()`.

Maybe add an option like `WithHTTPTransport(t *Transport)`. But I'm not sure if there are any other use cases besides testing.

config/confighttp/confighttp.go
```diff
@@ -103,8 +104,39 @@ func NewDefaultHTTPClientSettings() HTTPClientSettings {
...
+}
+
+type toClientOptionFunc func(*toClientOptions)
+
+func (fn toClientOptionFunc) apply(opts *toClientOptions) {
+	fn(opts)
+}
+
+// WithRoundTripper allow for individual components to intercept HTTP requests
+func WithRoundTripper(customRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)) ToClientOption {
```
bogdandrutu, line 126: With[Client]RoundTripper to make sure we start a pattern we add Client/Server after with.

bogdandrutu (approved):

config/confighttp/confighttp.go
bogdandrutu: ping @haoqixu

haoqixu: Sorry for the late reply. Please correct me if I've misunderstood.

PR #8829, opened 2023-11-08
Simplify logic in boundedMemoryQueue, use channels len/cap

bogdandrutu:


exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -246,12 +246,3 @@ func (s *consumerState) assertConsumed(expected map[string]bool) {
 	}
 	assert.Equal(s.t, expected, s.snapshot())
 }
-
-func TestZeroSize(t *testing.T) {
```
bogdandrutu, line 250: This is covered by not allowing users to configure 0 size, see QueueConfig.Validate()

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -247,11 +247,24 @@ func (s *consumerState) assertConsumed(expected map[string]bool) {
 	assert.Equal(s.t, expected, s.snapshot())
 }
 
-func TestZeroSize(t *testing.T) {
+func TestZeroSizeWithConsumers(t *testing.T) {
```
bogdandrutu, line 250: There is a small behavior change here if capacity/size is 0 and numConsumers >0. But this is not a valid case in QueueConfig, see Validate so we should not worry.

yurishkuro (approved):

dmitryax (approved):

dmitryax:
Submitted https://github.com/open-telemetry/opentelemetry-collector/pull/8832 to make the test pass

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -91,11 +75,11 @@ func (q *boundedMemoryQueue) Stop() {
 
 // Size returns the current size of the queue
 func (q *boundedMemoryQueue) Size() int {
-	return int(q.size.Load())
+	return len(q.items)
```
atoulme, line 78: If we are waiting on the call a back on a queue item, is it still in the queue?

dmitryax:
@bogdandrutu please rebase

PR #8830, opened 2023-11-09
persistentstorage: Simplify usage of storage client, avoid unnecessary allocations

bogdandrutu:


dmitryax (approved):
Nice simplification and optimization!

Issue #8804, opened 2023-11-06
Add tracing information to all components


codeboten:
Now that it is possible to emit traces from the Collector, it would be great produce spans consistently across components. This issue is a general issue that will include a checklist for all components in this repo.

**Describe the solution you'd like**
Each component produces a span using the `StartTracesOp` and `EndTracesOp` methods available in obsreport.

atoulme:
Cross-reference this issue https://github.com/open-telemetry/opentelemetry-collector/issues/6368

iblancasa:
What needs to be done? Any hints? I would love to contribute.

TylerHelmuth:
I still like the solution that implemented it in consumer so it automatically happened everywhere: https://github.com/open-telemetry/opentelemetry-collector/pull/8833

bogdandrutu:
I would be concern to add traces for each individual component. Think about a simple processor that would not benefit.

mackjmr:
@codeboten @iblancasa @TylerHelmuth I am interested in helping this issue move forward. 

There is already some tracing enabled in some parts of core:
- exporterhelper: In the [Send](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/traces.go#L165)
- otlpreceiver: In the [Export](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/internal/trace/otlp.go#L41)
- Some autoinstrumentation in [configgrpc](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/configgrpc.go#L480-L491) and [confighttp](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/confighttp.go#L462-L473)

Seems like there are currently 2 POCs to add traces to all components:

- Add obsreport to consumer (#8833, #10847)

The benefits of this approach is that this adds tracing to all components, specifically to the `Consume<Signal>Func` which actually does the work. 

One point I am unclear on in these POCs is that the obsreport is expected to be created and implemented in each of the components using the consumer. Why not create the obsreport within the consumer package ?
Is this made in this way so that the individual components could add some custom logic to the tracing ? If so, perhaps this could be done via configuration options of the obsreport within the consumer. Another benefit of creating the obsreport within the consumer would mean it would require exposing less APIs.

The main downside of this approach is that it exposes APIs/objects:
- Exposes `ObsReport` interface
[...]

PR #8833, opened 2023-11-09
[consumer] add obsreport to consumer

codeboten:
This change provides an option to the consumer (`WithObsReport`) to set an `obsreport` variable, which is used for adding spans at the beginning and end of each consume operation. If the option is not set, a no-op implementation of the ObsReport interface is used.

The follow up to this PR is to use the `WithObsReport` to set the option in various components.

Part of #8804


codeboten:
Pinging @TylerHelmuth as we talked about this at kubecon

TylerHelmuth (commented):
It's so beautiful

consumer/logs.go
```diff
@@ -35,8 +35,17 @@ func NewLogs(consume ConsumeLogsFunc, options ...Option) (Logs, error) {
 	if consume == nil {
 		return nil, errNilFunc
 	}
+
+	baseImpl := newBaseImpl(options...)
+	fn := func(ctx context.Context, ld plog.Logs) error {
+		baseImpl.obsreport.StartTracesOp(ctx)
+		err := consume(ctx, ld)
+		baseImpl.obsreport.EndTracesOp(ctx, ld.LogRecordCount(), err)
```
djaglowski, line 43: Worth looking at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/29274.

It's arguably a problem with the API here but at least we should capture the count prior to calling ConsumeLogs.

PR #8834, opened 2023-11-09
Fix internal/tools to not depend on any module that requires go 1.21

bogdandrutu:


yurishkuro:
how did these upgrades get in in the first place and succeeded in the CI?

PR #8836, opened 2023-11-09
Automate status reporting on start

mwear:
**Description:** 
This is part of the continued component status reporting effort. Currently we have automated status reporting for the following component lifecycle events: `Starting`, `Stopping`, `Stopped` as well as definitive errors that occur in the starting or stopping process (e.g. as determined by an error return value). This leaves the responsibility to the component to report runtime status after start and before stop. We'd like to be able to extend the automatic status reporting to report `StatusOK` if `Start` completes without an error. One complication with this approach is that some components spawn async work (via goroutines) that, depending on the Go scheduler, can report status before `Start` returns. As such, we cannot assume a nil return value from `Start` means the component has started properly. The solution is to detect if the component has already reported status when start returns, if it has, we will use the component-reported status and will not automatically report status. If it hasn't, and `Start` returns without an error, we can report `StatusOK`. Any subsequent reports from the component (async or otherwise) will transition the component status accordingly. 

The tl;dr is that we cannot control the execution of async code, that's up to the Go scheduler, but we can handle the race, report the status based on the execution, and not clobber status reported from within the component during the startup process. That said, for components with async starts, you may see a `StatusOK` before the component-reported status, or just the component-reported status depending on the actual execution of the code. In both cases, the end status will be same.

The work in this PR will allow us to simplify #8684 and #8788 and ultimately choose which direction we want to go for runtime status reporting.

**Testing:** units / manual

### DIFF
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -31,8 +31,8 @@ type TelemetrySettingsBase struct {
 	Resource pcommon.Resource
 }
 
-// TelemetrySettings and servicetelemetry.Settings differ in the method signature for
-// ReportComponentStatus
+// TelemetrySettings and servicetelemetry.TelemetrySettings differ in their
+// mechanism for reporting component status.
 type TelemetrySettings struct {
 	*TelemetrySettingsBase
 
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -12,7 +12,7 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 )
 
-type TelemetrySettingsBase[T any] struct {
+type TelemetrySettingsBase struct {
 	// Logger that the factory can use during creation and can pass to the created
 	// component to be used later as well.
 	Logger *zap.Logger
@@ -29,6 +29,12 @@ type TelemetrySettingsBase[T any] struct {
 
 	// Resource contains the resource attributes for the collector's telemetry.
 	Resource pcommon.Resource
+}
+
+// TelemetrySettings and servicetelemetry.Settings differ in the method signature for
+// ReportComponentStatus
+type TelemetrySettings struct {
+	*TelemetrySettingsBase
 
 	// ReportComponentStatus allows a component to report runtime changes in status. The service
 	// will automatically report status for a component during startup and shutdown. Components can
@@ -40,9 +46,5 @@ type TelemetrySettingsBase[T any] struct {
 	//   - Calling this method before component startup
 	//
 	// If the API is being used properly, these errors are safe to ignore.
-	ReportComponentStatus T
+	ReportComponentStatus StatusFunc
 }
-
-// TelemetrySettings and servicetelemetry.Settings differ in the method signature for
-// ReportComponentStatus
-type TelemetrySettings TelemetrySettingsBase[StatusFunc]
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -12,7 +12,12 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 )
 
-type TelemetrySettingsBase struct {
+// TelemetrySettings provides components with APIs to report telemetry.
+//
+// Note: there is a service version of this struct, servicetelemetry.TelemetrySettings, that mirrors
+// this struct with the exception of ReportComponentStatus. When adding or removing anything from
+// this struct consider whether or not the same should be done for the service version.
+type TelemetrySettings struct {
 	// Logger that the factory can use during creation and can pass to the created
 	// component to be used later as well.
 	Logger *zap.Logger
@@ -29,12 +34,6 @@ type TelemetrySettingsBase struct {
 
 	// Resource contains the resource attributes for the collector's telemetry.
 	Resource pcommon.Resource
-}
-
-// TelemetrySettings and servicetelemetry.TelemetrySettings differ in their
-// mechanism for reporting component status.
-type TelemetrySettings struct {
-	TelemetrySettingsBase
 
 	// ReportComponentStatus allows a component to report runtime changes in status. The service
 	// will automatically report status for a component during startup and shutdown. Components can
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -48,7 +48,7 @@ type TelemetrySettings struct {
 	ReportComponentStatus StatusFunc
 }
 
-// Deprecated: [0.90.0] Use TelemetrySettings directly
+// Deprecated: [0.91.0] Use TelemetrySettings directly
 type TelemetrySettingsBase[T any] struct {
 	// Logger that the factory can use during creation and can pass to the created
 	// component to be used later as well.
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -47,3 +47,35 @@ type TelemetrySettings struct {
 	// If the API is being used properly, these errors are safe to ignore.
 	ReportComponentStatus StatusFunc
 }
+
+// Deprecated: [0.90.0] Use TelemetrySettings directly
+type TelemetrySettingsBase[T any] struct {
+	// Logger that the factory can use during creation and can pass to the created
+	// component to be used later as well.
+	Logger *zap.Logger
+
+	// TracerProvider that the factory can pass to other instrumented third-party libraries.
+	TracerProvider trace.TracerProvider
+
+	// MeterProvider that the factory can pass to other instrumented third-party libraries.
+	MeterProvider metric.MeterProvider
+
+	// MetricsLevel controls the level of detail for metrics emitted by the collector.
+	// Experimental: *NOTE* this field is experimental and may be changed or removed.
+	MetricsLevel configtelemetry.Level
+
+	// Resource contains the resource attributes for the collector's telemetry.
+	Resource pcommon.Resource
+
+	// ReportComponentStatus allows a component to report runtime changes in status. The service
+	// will automatically report status for a component during startup and shutdown. Components can
+	// use this method to report status after start and before shutdown. ReportComponentStatus
+	// will only return errors if the API used incorrectly. The two scenarios where an error will
+	// be returned are:
+	//
+	//   - An illegal state transition
+	//   - Calling this method before component startup
+	//
+	// If the API is being used properly, these errors are safe to ignore.
+	ReportComponentStatus T
+}
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -34,7 +34,7 @@ type TelemetrySettingsBase struct {
 // TelemetrySettings and servicetelemetry.TelemetrySettings differ in their
 // mechanism for reporting component status.
 type TelemetrySettings struct {
-	*TelemetrySettingsBase
+	TelemetrySettingsBase
 
 	// ReportComponentStatus allows a component to report runtime changes in status. The service
 	// will automatically report status for a component during startup and shutdown. Components can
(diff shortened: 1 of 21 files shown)


component/telemetry.go
```diff
@@ -12,7 +12,7 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 )
 
-type TelemetrySettingsBase[T any] struct {
+type TelemetrySettingsBase struct {
```
yurishkuro, line 15: what's the value of having this as a separate type?

jmacd (approved):
Looks good to me. Thanks for the detailed PR description @mwear.

mwear:
The test failures in contrib are due to embedding `TelemetrySettingsBase` in `TelemetrySettings`. This PR as-is, will require some changes there. If that's not an option, I can do away with TelemetrySettingsBase altogether and duplicate code between `component.TelemetrySettings` and `servicetelemetry.TelemetrySettings`.

service/internal/graph/graph.go
```diff
@@ -386,12 +386,24 @@ func (g *Graph) StartAll(ctx context.Context, host component.Host) error {
...
+				component.NewPermanentErrorEvent(compErr),
+			)
 			return compErr
 		}
+
+		_ = g.telemetry.Status.ReportComponentStatusIf(
+			instanceID,
+			component.NewStatusEvent(component.StatusOK),
+			func(st component.Status) bool { return st == component.StatusStarting },
+		)
```
djaglowski, line 406: I think this solution works and won't hold this up, but I want to point out a minor drawback of this approach and a possible alternative.

The drawback here is that we're exposing an incongruity where a developer could apply reasoning about statuses and status transitions in an unintended way that works against the design. For example, one could suggest to transition to OK only if it is in a PermanentError state. (Obviously this won't work, but kind of weird that our API would allow the developer to suggest something which directly subverts the intended model.)

---

The alternative I would suggest is in the end I think quite simple but requires a bit of explanation.

Basically, I think we have slightly convoluted the notions of _Status_ and _StatusEvent_ by making them share the same enumeration. If you look at the FSM which governs status, we are describing _everything_ in terms of _Status_, but I think it's worth looking at what it would mean to support a separate enumeration for _StatusEvent_.

Currently, when updating Status, we look at the type of StatusEvent (a Status) and interpret it to mean "go to this Status". Then we apply some logic, "if this is a valid transition from the current state, do it".

What I'm proposing would be to look at the type of StatusEvent (an enum separate from Status) and interpret it as "an event happened, react accordingly". Then we apply some logic, "if this is a relevant event while in the current state, trigger a corresponding transition to another state".
 
The reason we've gotten away without this distinction is that if we did have a separate StatusEvent enum, it would be almost identical to the current Status enum - so much so that I think this particular case may be the only one where is doesn't work smoothly.

---

To show this in more detail, I've added a notion of StatusEvent types to our FSM. The only difference here is that I've disambiguated "OK" into "Started" and "Recovered".

image

As to how this specifically helps in this situation, first just assume StatusEvents are _automatically_ reported as either "Started" or "Fatal" based on the return from Start. (Same as this PR except replace "OK" with "Started"). Then we can consider the possible cases where an async routine also reports a StatusEvent.

There are eight possible cases, based on:
- Which happens first, Start returns, or the async routine reports a StatusEvent
- Whether Start returns nil or an error
- Whether the async routine reports "StatusEvent: Started" or "StatusEvent: Recoverable" (other error types follow the same pattern here so I'm ignoring them)

All of these cases working naturally within the updated FSM without any need for the user of the API to be concerned with current state or order of events.

In the first two cases, Start returns nil first, which triggers an automatic "StatusEvent: Started". This is a relevant event while "Status: Starting", so a transition to "Status: OK" occurs. Then:
    - Case 1: The async routine reports "StatusEvent: Started". This is _not_ a relevant event while "Status: OK", so no transition is triggered.
    - Case 2: The async routine reports "StatusEvent: Recoverable". This is a relevant event while "Status: OK", so a transition to "Status: Recoverable" occurs.

In the next two cases, Start returns error first, which triggers an automatic "StatusEvent: Fatal". This is a relevant event while "Status: Starting", so a transition to "Status: Fatal" occurs. Then:
    - Case 3: The async routine reports "StatusEvent: Started". This is _not_ a relevant event while "Status: Fatal", so no transition is triggered.
    - Case 4: The async routine reports "StatusEvent: Recoverable". This is _not_ a relevant event while "Status: Fatal", so no transition is triggered.

In the next two cases, the async routine reports "StatusEvent: Started" first. This is a relevant event while "Status: Starting", so a transition to "Status: OK" occurs. Then:
    - Case 5: Start returns nil, which triggers an automatic "StatusEvent: Started". This is _not_ a relevant event while "Status: OK", so no transition is triggered.
    - Case 6: Start returns an error, which triggers an automatic "StatusEvent: Fatal". This is a relevant event while "Status: OK", so a transition to "Status: Fatal" occurs.

In the next two cases, the async routine reports "StatusEvent: Recoverable" first. This is a relevant event while "Status: Starting", so a transition to "Status: Recoverable" occurs. Then:
    - Case 7: Start returns nil, which triggers an automatic "StatusEvent: Started". This is _not_ a relevant event while "Status: Recoverable", so no transition is triggered.
    - Case 8: Start returns an error, which triggers an automatic "StatusEvent: Fatal". This is a relevant event while "Status: Recoverable", so a transition to "Status: Fatal" occurs.

This design changes very little else about how the system works:
- We need to enumerate the types of StatusEvent, and update the StatusEvent structure to carry it instead of Status.
- We need to rewrite the FSD somewhat, to switch based on StatusEvent, but the rules remain nearly unchanged.
- Users of the API need to report "StatusEvent: Started" or "StatusEvent: Recovered" instead of "Status: OK".

--- 

Anyways, this is a really long explanation for something which may not be strictly necessary, but I wanted to share the thought in case you agree this addresses the root cause of our problem in a more consistent way.

mwear: I'd like to first discuss the following comment and then move on to discuss the proposal.

> The drawback here is that we're exposing an incongruity where a developer could apply reasoning about statuses and status transitions in an unintended way that works against the design. For example, one could suggest to transition to OK only if it is in a PermanentError state. (Obviously this won't work, but kind of weird that our API would allow the developer to suggest something which directly subverts the intended model.)

I'd to clarify that this PR does not expose the `ReportComponentStatusIf` method to components (via  `component.TelemetrySettings`). It's only available to the service (on `servicetelemetry.TelemetrySettings`). That said, it doesn't mean it couldn't be abused in the service, but component authors won't have access to it. If we are worried about it being abused in the service, we could change the method to `ReportComponentStatusIfStarting` effectively hardcoding the predicate.


If I understand the new proposal properly, we would still have our current statuses:

```go
type Status int32
const (
    StatusNone Status = iota
    StatusStarting
    StatusOK
    StatusRecoverableError
    StatusPermanentError
    StatusFatalError
    StatusStopping
    StatusStopped
)
```

And introduce the following status event enums

```go
type StatusEventType int32
const(
    StatusEventNone StatusEventType = iota
    StatusEventStarting
    StatusEventStarted
    StatusEventRecovered
    StatusEventRecoverableError
    StatusEventPermanentError
    StatusEventFatalError
    StatusEventStopping
    StatusEventStopped
)    
```

`StatusOK` is split into `StatusEventStarted` and `StatusEventRecovered`. API users will report `StatusEvent`s with `StatusEventType`, and we'll map them to the corresponding `Status`. In order to keep our event aggregation code working as is, we'll need to retain the `component.Status`.

Currently a `StatusEvent` is defined as:

```go
type StatusEvent struct {
    status    Status
    err       error
    timestamp time.Time
}
```

It would change to
```go
type StatusEvent struct {
    eventType StatusEventType
    status    Status
    err       error
    timestamp time.Time
}
```

We'd have to account for aggregating the `eventType`. For the most part, it matches the status aggregation we have, but we'd have to handle the `StatusEventStarted` / `StatusEventRecovered` case for the aggregate event.

One big benefit of the current design is that it does not require a component to keep any state about its current status. A component can report status changes as they happen and the FSM takes care of the rest. Splitting `StatusOK` into `StatusEventStarted` and `StatusEventRecovered` will complicate reporting runtime status. 

As an example, let's consider an exporter in the current system, after it has started and is running successfully. It then:
- successfully exports a batch of spans and reports `StatusOK` (it's already OK, so no state change occurs)
- it has a timeout exporting spans and reports `StatusRecoverableError` (the state is changed to `StatusRecoverableError`)
- it successfully exports a batch of spans and reports `StatusOK` (the state is changed to `StatusOK`); the component has recovered

In the new system, since we lose `StatusOK`, we need to know previous state to know if the component has recovered or not. Alternatively we could preemptively report `StatusEventRecovered` in place of `StatusOK`.

Following the exporter example, suppose the exporter has been successfully started and then it:
- successfully exports a batch of spans. It needs to either a) unconditionally report `StatusEventRecovered` or b) keep track of state internally (recoverable or ok), and conditionally report `StatusEventRecovered` based on that state; it would report nothing in this case. 
- it has a timeout exporting spans and reports `StatusEventRecoverableError` (the state is changed to `StatusRecoverableError`)
- it successfully exports a batch of spans. It needs to either a) unconditionally report StatusEventRecovered or b) keep track of state internally (recoverable or ok), and conditionally report StatusEventRecovered based on that state; in this case, it would report `StatusEventRecovered` and the status would become `StatusOK`

Runtime status reporting is my biggest concern with this change. I think it's natural to report `StatusOK` or `StatusRecoverable` as operations succeed or fail, but very unnatural to report `StatusEventRecovered` unconditionally, or have to maintain state and conditionally report it.

While I think that splitting `StatusOK` into `StatusEventStarted` and `StatusEventRecovered` smooths over some issues during startup, it complicates runtime status reporting and to some degree event aggregation. A middle ground might be to introduce `StatusStarted` as a pseudo-status that will be converted to `StatusOK` if the current status is `StatusStarting`. I could prototype how this could work if it seems worthwhile. I'll also reiterate the current solution, which is an API that is only available at the service level, not the component level, to conditionally report status (based on a predicate). It could be improved by removing the predicate to prevent misuse. 

Let me know if this makes any sense or if I've misunderstood the proposal in some way. If any of the options discussed stand out as a clear winner, also let me know.

djaglowski: Thanks for considering my suggestion here. I think you've pointed out some things I hadn't considered. I was thinking it buttoned up some loose ends but it adds complexity elsewhere, so not worth it.

I think some form of `ReportComponentStatusIfStarting` which only allows what is necessary is a good enough solution.

djaglowski (approved):

component/telemetry.go
```diff
@@ -12,7 +12,12 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 )
 
-type TelemetrySettingsBase[T any] struct {
```
codeboten, line 15: This should be deprecated rather than removed since its a public API

service/service.go
```diff
@@ -151,7 +148,7 @@ func (srv *Service) Start(ctx context.Context) error {
 	)
 
 	// enable status reporting
-	srv.statusInit()
+	srv.telemetrySettings.Status.Ready()
```
codeboten, line 151: do we need to worry about status ever being `nil` in any of the places where its called without a check for nil first?

component/telemetry.go
mwear: This was introduced as part of https://github.com/open-telemetry/opentelemetry-collector/pull/8169. It only existed to be a base type for `component.TelemetrySettings` and `servicetelemetry.TelemetrySettings`. I would not expect any usage of it outside this repo. I think it's safe to remove, but if we want to leave here for now, deprecate it, and remove it later, that's fine with me.

service/service.go
mwear: This is intialized in `service.New(...)` and is in the same situation as `telemetrySettings.TracerProvider`, `telemetrySettings.MeterProvider` etc. The only way it would not be initialized would be skip the constructor, which I don't we need to be worried about.

PR #8840, opened 2023-11-11
cleaning up some lint complaints

JohnRoesler:
Huge fan of this project (but I haven't looked at the codebase before)! So, I was just poking around in the code to get familiar and made some tweaks along the way. Hopefully you find it helpful.

**Description:**  fixing various linting complaints
- structs should have either pointer or value receivers but not both (some can't be adjusted due to interfaces that are being implemented that require value receivers)
- variable name collisions with imported packages
- explicitly discard errors when not checking

- Tests are passing locally (20,641 of them)
- golanglint passing


atoulme:
How did you generate those lint errors? Any chance you can offer CI changes to generate the same coverage?

config/configtls/configtls_test.go
```diff
@@ -513,11 +513,11 @@ func TestCertificateReload(t *testing.T) {
 			// Copy certs into a temp dir so we can safely modify them
 			certFile, err := os.CreateTemp("", "cert")
 			assert.NoError(t, err)
-			defer os.Remove(certFile.Name())
+			defer func() { _ = os.Remove(certFile.Name()) }()
```
bogdandrutu, line 516:
```suggestion
			t.Cleanup(func() { assert.NoError(t, os.Remove(certFile.Name()) })
```

bogdandrutu: Same everywhere.

JohnRoesler:
@atoulme I will do some looking into how intellij is generating the lint complaints - it would be nice there are rules in golangci-lint that can catch them

confmap/provider/internal/configurablehttpprovider/provider.go
```diff
@@ -96,7 +96,7 @@ func (fmp *provider) Retrieve(_ context.Context, uri string, _ confmap.WatcherFu
 	if err != nil {
 		return nil, fmt.Errorf("unable to download the file via HTTP GET for uri %q: %w ", uri, err)
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
bogdandrutu, line 99: Does this impact performance on the critical path for no good reasons? I would retain from doing it unless a good reason to do so.

bogdandrutu: Same everywhere in no-test code.

bogdandrutu (commented):
I would suggest:
1. Split from test vs non-test code;
2. For non-test let's see what impact performance what does not;

confmap/provider/internal/configurablehttpprovider/provider.go
JohnRoesler: I was curious, so created a benchmark to compare the two behaviors:

and in this simple example, explicitly ignoring the error in a closure is actually more performant (now i want to know why that is):

https://gist.github.com/JohnRoesler/d3bf89d60630bc60e6afd8d14ab88e90

Issue #8844, opened 2023-11-11
[exporter/debug] default verbosity should to be changed from `normal` to `basic`.


codeboten:
This is a followup from #8841

> Although I believe this description fits better into the `basic` section and it's the `normal` behavior that is somehow incorrect and subject to be changed. I also beileve the default verbosity also needs to be changed from `normal` to `basic`. We can talk about this in future pull requests/issues though. :slightly_smiling_face:

_Originally posted by @astencel-sumo in https://github.com/open-telemetry/opentelemetry-collector/pull/8841#discussion_r1390228280_

codeboten:
fixed by https://github.com/open-telemetry/opentelemetry-collector/pull/8846

Issue #7806, opened 2023-06-01
[exporter/debug] change behavior of "normal" verbosity to be different from "basic"


andrzej-stencel:
EDIT: Updated the issue to change the `debug` exporter instead of `logging` exporter.

Collector version: v0.88.0

**Is your feature request related to a problem? Please describe.**

Currently there are three verbosity levels available for the [Debug exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.88.0/exporter/debugexporter/README.md): `basic`, `normal`, `detailed`. The `basic` and `normal` level result in the same output, which only displays the number of records:

```yaml
exporters:
  logging:
    verbosity: normal
```

```console
2023-06-01T13:26:16.044+0200    info    MetricsExporter {"kind": "exporter", "data_type": "metrics", "name": "logging", "resource metrics": 1, "metrics": 1, "data points": 1}
```

On the other hand, the `detailed` verbosity displays all the details about telemetry, with multiple lines per span/data point/log record:

```console
2023-06-01T13:30:22.669+0200    info    MetricsExporter {"kind": "exporter", "data_type": "metrics", "name": "logging", "resource metrics": 1, "metrics": 1, "data points": 1}
2023-06-01T13:30:22.669+0200    info    ResourceMetrics #0
Resource SchemaURL: 
ScopeMetrics #0
ScopeMetrics SchemaURL: 
InstrumentationScope otelcol/httpcheckreceiver 0.78.0
Metric #0
Descriptor:
     -> Name: httpcheck.duration
     -> Description: Measures the duration of the HTTP check.
     -> Unit: ms
     -> DataType: Gauge
NumberDataPoints #0
Data point attributes:
     -> http.url: Str(https://opentelemetry.io)
StartTimestamp: 2023-06-01 11:30:18.60847008 +0000 UTC
Timestamp: 2023-06-01 11:30:21.630471795 +0000 UTC
Value: 1038
```

**Describe the solution you'd like**

I propose to change the behavior for the `normal` verbosity to display telemetry details **one line per span/data point/log record**:

```yaml
exporters:
  logging:
    verbosity: normal
```

```console
2023-06-01T13:30:22.669+0200    info    MetricsExporter {"kind": "exporter", "data_type": "metrics", "name": "logging", "resource metrics": 1, "metrics": 1, "data points": 1}
[...]

PR #8846, opened 2023-11-11
[exporter/debug] change default verbosity from `normal` to `basic`

astencel-sumo:
**Description:**

This change has currently no effect, as `basic` and `normal` verbosity share the same behavior. This might change in the future though, with the `normal` verbosity being more verbose than it currently is (see https://github.com/open-telemetry/opentelemetry-collector/issues/7806). This is why we are changing the default to `basic`, which is expected to stay at the current level of verbosity (one line per batch).

**Link to tracking Issue:**

- https://github.com/open-telemetry/opentelemetry-collector/issues/8844


.chloggen/debug-exporter-default-verbosity-basic.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
codeboten, line 4: Since the change is essentially a no-op, would this really be a breaking change?

astencel-sumo: Hmm yeah maybe not. I'll change it to enhancement then.

codeboten (approved):

PR #8853, opened 2023-11-12
[exporterhelper] Add WithRequestQueue option to the exporter

dmitryax:
Introduce a way to enable queue in the new exporter helper with a developer interface suggested in https://github.com/open-telemetry/opentelemetry-collector/pull/8248#discussion_r1302102261. 

The new configuration interface for the end users provides a new `queue_size_items` option to limit the queue by a number of spans, log records, or metric data points. The previous way to limit the queue by number of requests is preserved under the same field, `queue_size,` which will later be deprecated through a longer transition process.

Tracking issue: https://github.com/open-telemetry/opentelemetry-collector/issues/8122

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -13,6 +13,7 @@ import (
 	"go.opentelemetry.io/collector/config/configretry"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exporterqueue"
 )
 
 // requestSender is an abstraction of a sender for a request independent of the type of the data (traces, metrics, logs).
@@ -85,17 +86,43 @@ func WithRetry(config configretry.BackOffConfig) Option {
 func WithQueue(config QueueSettings) Option {
 	return func(o *baseExporter) {
 		if o.requestExporter {
-			panic("queueing is not available for the new request exporters yet")
+			panic("WithQueue option is not available for the new request exporters, use WithRequestQueue instead")
 		}
 		if !config.Enabled {
 			o.exportFailureMessage += " Try enabling sending_queue to survive temporary failures."
 			return
 		}
-		consumeErrHandler := func(err error, req Request) {
-			o.set.Logger.Error("Exporting failed. Dropping data."+o.exportFailureMessage,
-				zap.Error(err), zap.Int("dropped_items", req.ItemsCount()))
+		qf := exporterqueue.NewPersistentQueueFactory[Request](config.StorageID, exporterqueue.PersistentQueueSettings[Request]{
+			Marshaler:   o.marshaler,
+			Unmarshaler: o.unmarshaler,
+		})
+		q := qf(context.Background(), exporterqueue.Settings{
+			DataType:         o.signal,
+			ExporterSettings: o.set,
+		}, exporterqueue.Config{
+			Enabled:      config.Enabled,
+			NumConsumers: config.NumConsumers,
+			QueueSize:    config.QueueSize,
+		})
+		o.queueSender = newQueueSender(q, o.set, config.NumConsumers, o.exportFailureMessage)
+	}
+}
+
+// WithRequestQueue enables queueing for an exporter.
+// This option should be used with the new exporter helpers New[Traces|Metrics|Logs]RequestExporter.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+func WithRequestQueue(cfg exporterqueue.Config, queueFactory exporterqueue.Factory[Request]) Option {
+	return func(o *baseExporter) {
+		if !cfg.Enabled {
+			o.exportFailureMessage += " Try enabling sending_queue to survive temporary failures."
+			return
 		}
-		o.queueSender = newQueueSender(config, o.set, o.signal, o.marshaler, o.unmarshaler, consumeErrHandler)
+		set := exporterqueue.Settings{
+			DataType:         o.signal,
+			ExporterSettings: o.set,
+		}
+		o.queueSender = newQueueSender(queueFactory(context.Background(), set, cfg), o.set, cfg.NumConsumers, o.exportFailureMessage)
 	}
 }
 
@@ -114,8 +141,8 @@ type baseExporter struct {
 	component.ShutdownFunc
 
 	requestExporter bool
-	marshaler       RequestMarshaler
-	unmarshaler     RequestUnmarshaler
+	marshaler       exporterqueue.Marshaler[Request]
+	unmarshaler     exporterqueue.Unmarshaler[Request]
 	signal          component.DataType
 
 	set    exporter.CreateSettings
@@ -136,8 +163,8 @@ type baseExporter struct {
 }
 
 // TODO: requestExporter, marshaler, and unmarshaler arguments can be removed when the old exporter helpers will be updated to call the new ones.
-func newBaseExporter(set exporter.CreateSettings, signal component.DataType, requestExporter bool, marshaler RequestMarshaler,
-	unmarshaler RequestUnmarshaler, osf obsrepSenderFactory, options ...Option) (*baseExporter, error) {
+func newBaseExporter(set exporter.CreateSettings, signal component.DataType, requestExporter bool,
+	marshaler exporterqueue.Marshaler[Request], unmarshaler exporterqueue.Unmarshaler[Request], osf obsrepSenderFactory, options ...Option) (*baseExporter, error) {
 
 	obsReport, err := NewObsReport(ObsReportSettings{ExporterID: set.ID, ExporterCreateSettings: set})
 	if err != nil {
--- a/exporter/exporterqueue/queue.go
+++ b/exporter/exporterqueue/queue.go
@@ -1,0 +1,96 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterqueue // import "go.opentelemetry.io/collector/exporter/exporterqueue"
+
+import (
+	"context"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/internal/queue"
+)
+
+// Queue defines a producer-consumer exchange which can be backed by e.g. the memory-based ring buffer queue
+// (boundedMemoryQueue) or via a disk-based queue (persistentQueue)
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Queue[T any] queue.Queue[T]
+
+// Settings defines settings for creating a queue.
+type Settings struct {
+	DataType         component.DataType
+	ExporterSettings exporter.CreateSettings
+}
+
+// Marshaler is a function that can marshal a request into bytes.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Marshaler[T any] func(T) ([]byte, error)
+
+// Unmarshaler is a function that can unmarshal bytes into a request.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Unmarshaler[T any] func([]byte) (T, error)
+
+// Factory is a function that creates a new queue.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Factory[T any] func(context.Context, Settings, Config) Queue[T]
+
+// NewMemoryQueueFactory returns a factory to create a new memory queue.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+func NewMemoryQueueFactory[T itemsCounter]() Factory[T] {
+	return func(_ context.Context, _ Settings, cfg Config) Queue[T] {
+		return queue.NewBoundedMemoryQueue[T](queue.MemoryQueueSettings[T]{
+			Sizer:    sizerFromConfig[T](cfg),
+			Capacity: capacityFromConfig(cfg),
+		})
+	}
+}
+
+// PersistentQueueSettings defines developer settings for the persistent queue factory.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type PersistentQueueSettings[T any] struct {
+	// Marshaler is used to serialize queue elements before storing them in the persistent storage.
+	Marshaler Marshaler[T]
+	// Unmarshaler is used to deserialize requests after reading them from the persistent storage.
+	Unmarshaler Unmarshaler[T]
+}
+
+// NewPersistentQueueFactory returns a factory to create a new persistent queue.
+// If cfg.StorageID is nil then it falls back to memory queue.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+func NewPersistentQueueFactory[T itemsCounter](storageID *component.ID, factorySettings PersistentQueueSettings[T]) Factory[T] {
+	if storageID == nil {
+		return NewMemoryQueueFactory[T]()
+	}
+	return func(_ context.Context, set Settings, cfg Config) Queue[T] {
+		return queue.NewPersistentQueue[T](queue.PersistentQueueSettings[T]{
+			Sizer:            sizerFromConfig[T](cfg),
+			Capacity:         capacityFromConfig(cfg),
+			DataType:         set.DataType,
+			StorageID:        *storageID,
+			Marshaler:        factorySettings.Marshaler,
+			Unmarshaler:      factorySettings.Unmarshaler,
+			ExporterSettings: set.ExporterSettings,
+		})
+	}
+}
+
+type itemsCounter interface {
+	ItemsCount() int
+}
+
+func sizerFromConfig[T itemsCounter](Config) queue.Sizer[T] {
+	// TODO: Handle other ways to measure the queue size once they are added.
+	return &queue.RequestSizer[T]{}
+}
+
+func capacityFromConfig(cfg Config) int {
+	// TODO: Handle other ways to measure the queue size once they are added.
+	return cfg.QueueSize
+}
--- a/.chloggen/exporter-helper-v2.yaml
+++ b/.chloggen/exporter-helper-v2.yaml
@@ -1,0 +1,32 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporter/exporterhelper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add API for enabling queue in the new exporter helpers.
+
+# One or more tracking issues or pull requests related to the change
+issues: [7874]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The following experimental API is introduced in exporter package:
+    - `exporterhelper.WithRequestQueue`: a new exporter helper option for using a queue.
+    - `exporterqueue.Queue`: an interface for queue implementations.
+    - `exporterqueue.Factory`: a queue factory interface, implementations of this interface are intended to be used with WithRequestQueue option.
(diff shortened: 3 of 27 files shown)


exporter/exporterhelper/internal/request/request.go
```diff
@@ -0,0 +1,69 @@
...
+// then 0 is returned.
+func (req *Request) Count() int {
+	if counter, ok := req.Request.(request.ItemsCounter); ok {
+		return counter.ItemsCount()
+	}
+	return 0
+}
+
+// OnProcessingFinished calls the optional callback function to handle cleanup after all processing is finished
+func (req *Request) OnProcessingFinished() {
```
dmitryax, line 54: If we expose the queue interface, this probably can be exposed as an optional request interface as well. Then the queue `Produce` method can accept a public `Request` instead of the internal one

exporter/exporterhelper/internal/request/request.go
```diff
@@ -0,0 +1,69 @@
...
+	if req.onProcessingFinishedCallback != nil {
+		req.onProcessingFinishedCallback()
+	}
+}
+
+// SetOnProcessingFinished allows to set an optional callback function to do the cleanup (e.g. remove the item from persistent queue)
+func (req *Request) SetOnProcessingFinished(callback func()) {
```
dmitryax, line 61: If we expose the queue interface now, this can be exposed as an optional Request interface as well. Then the `Queue.Produce` method can accept the public Request type instead of the internal one

dmitryax: Not relevant anymore after recent changes

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,92 @@
...
+func (rcl *requestsCapacityLimiter[T]) Claim(_ T) bool {
+	if rcl.size.Load() >= rcl.capacity {
+		return false
+	}
+	rcl.size.Add(1)
+	return true
+}
+
+func (rcl *requestsCapacityLimiter[T]) Release(_ T) {
+	rcl.size.Add(^uint64(0))
```
splunkericl, line 68: This can overflow from 0 to 4294967295. 

I tested the new changes with persistent queue and saw that:
- when persistent queue initialize, it loads some `currentlyDispatchedItems` and consume it. Once an item is consumed, this number becomes overflow and all future `Claim` becomes blocked.

Would it make sense to safe guard it from overflowing(always above 0)? Otherwise it is hard to ensure callers to not accidentally call `Release`.

dmitryax: Thanks for the reports! It's a bug indeed. Let me fix

dmitryax: I implemented the more resilient queue size tracking for the persistent queue

exporter/exporterhelper/queue.go
```diff
@@ -0,0 +1,104 @@
...
+	// StorageID if not empty, enables the persistent storage and uses the component specified
+	// as a storage extension for the persistent queue
+	StorageID *component.ID `mapstructure:"storage"`
+}
+
+// NewPersistentQueueFactory returns a factory to create a new persistent queue.
+// If cfg.StorageID is nil then it falls back to memory queue.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+func NewPersistentQueueFactory(storageID *component.ID, marshaler RequestMarshaler, unmarshaler RequestUnmarshaler) QueueFactory {
```
bogdandrutu, line 97: Provide a "PersistentQueueSettings" to allow extending. Include the Marshaler/Unmarshaler for the moment.

exporter/exporterhelper/queue.go
```diff
@@ -0,0 +1,104 @@
...
+	Enabled bool `mapstructure:"enabled"`
+	// NumConsumers is the number of consumers from the queue.
+	NumConsumers int `mapstructure:"num_consumers"`
+	// QueueSizeItems is the maximum number of items (spans, metric data points or log records)
+	// allowed in queue at any given time.
+	QueueSizeItems int `mapstructure:"queue_size_items"`
+	// QueueSize is the maximum number of requests allowed in queue at any given time.
+	// This option is left for backward compatibility and will be deprecated in the future.
+	// It's recommended to use QueueSizeItems instead.
+	QueueSize int `mapstructure:"queue_size"`
```
bogdandrutu, line 29: Should we offer 2 sizes or a size + enum saying count num requests or count num items eventually count bytes if we want?

exporter/exporterhelper/queue.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+)
+
+// QueueConfig defines configuration for queueing requests before exporting.
+// It's supposed to be used with the new exporter helpers New[Traces|Metrics|Logs]RequestExporter.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+// TODO: Consider moving it to config/configqueue module.
```
bogdandrutu, line 17: Let's probably start there?

exporter/exporterhelper/queue.go
dmitryax: I was thinking about it. From the user's perspective, specifying the enum field might be less obvious. Also, setting `queue_element: item` (assuming we find a better field name) where the item represents a span or metric data point seems more confusing than having a set of predefined documented fields. Let me know if you have any other specific suggestions. For now, I think `queue_size`, `queue_size_items`, and potentially `queue_size_mib` is an easier to use interface.

bogdandrutu: maybe use suffix? `queue_size: 10[B|KB|MB..]` (applies to size in mb) then `queue_size: 10[requests]?` (requests) and `queue_size: 10items`

exporter/exporterhelper/queue.go
dmitryax: Makes sense. Applied

exporter/exporterhelper/queue.go
dmitryax: Ok. Created the `config/configqueue` module

exporter/exporterhelper/queue.go
dmitryax: I'm still hesitant that it'll make the user experience easier with `items` being ambiguous.

Having it as a part of the field names gives some flexibility to the component implementations, like they can have different name for the `items` field where it's not directly mapped to spans, datapoint log records. Having the parsing logic hardcoded wouldn't allow that. Also, I can imagine a use case when only one type would be supported by a particular exporter. With different fields, it's possible to restrict the options on the exporter side.

dmitryax: If we agree to keep different fields, maybe we can also rename them by removing `queue_` prefix since it's supposed be under `sending_queue` group anyway. Can be 
- `max_size_items` and `max_size_mib`
- `size_limit_items` and `size_limit_mib`
- `items_limit` and `mib_limit`

config/configqueue/queue.go
```diff
@@ -0,0 +1,69 @@
...
+	return nil
+}
+
+// PersistentQueueConfig defines configuration for queueing requests in a persistent storage.
+// The struct is provided to be added in the exporter configuration as one struct under the "sending_queue" key.
+// The exporter helper Go interface requires the fields to be provided separately to WithRequestQueue and
+// NewPersistentQueueFactory.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type PersistentQueueConfig struct {
```
bogdandrutu, line 64: Would be interested to see how people decide to use this vs QueueConfig in their Configs. When do I pick one or the other?

config/configqueue/queue.go
```diff
@@ -0,0 +1,69 @@
...
+	"errors"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+// QueueConfig defines configuration for queueing requests before exporting.
+// It's supposed to be used with the new exporter helpers New[Traces|Metrics|Logs]RequestExporter.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type QueueConfig struct {
```
bogdandrutu, line 16: second thoughts: for retry we put it under config because receivers/connectors may also retry adding to a pipeline. For queue do we see an use case to be used outside exporter? If not, consider to have a `exporter/exporterqueue` where we declare the config and the Queue interface?

consumer/go.mod
```diff
@@ -53,3 +53,5 @@ retract (
 replace go.opentelemetry.io/collector/config/configtelemetry => ../config/configtelemetry
 
 replace go.opentelemetry.io/collector/config/configretry => ../config/configretry
+
+replace go.opentelemetry.io/collector/config/configqueue => ../config/configqueue
```
bogdandrutu, line 57: How does this gets here? Seems that we have some sort of circular dependency via the `go.opentelemetry.io/collector` module.

exporter/debugexporter/go.mod
```diff
@@ -28,15 +30,14 @@ require (
...
+	go.opentelemetry.io/collector/config/configqueue v0.0.0-00010101000000-000000000000 // indirect
 	go.opentelemetry.io/collector/config/configretry v0.0.0-20231221085427-9027a8d9cc3f // indirect
 	go.opentelemetry.io/collector/consumer v0.91.0 // indirect
 	go.opentelemetry.io/collector/extension v0.91.0 // indirect
 	go.opentelemetry.io/collector/featuregate v1.0.0 // indirect
 	go.opentelemetry.io/collector/pdata v1.0.0 // indirect
 	go.opentelemetry.io/collector/receiver v0.91.0 // indirect
 	go.opentelemetry.io/otel v1.21.0 // indirect
-	go.opentelemetry.io/otel/metric v1.21.0 // indirect
-	go.opentelemetry.io/otel/trace v1.21.0 // indirect
```
bogdandrutu, line 39: Interesting change... Do you know why? This is just educational.

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -22,10 +22,10 @@ type boundedMemoryQueue[T any] struct {
 
 // NewBoundedMemoryQueue constructs the new queue of specified capacity, and with an optional
 // callback for dropped items (e.g. useful to emit metrics).
-func NewBoundedMemoryQueue[T any](sizer Sizer[T], capacity int) Queue[T] {
+func NewBoundedMemoryQueue[T any](set QueueCreateSettings[T]) Queue[T] {
```
bogdandrutu, line 25: The CreateSettings change can be extracted in a separate PR since we need that anyway.

exporter/exporterhelper/internal/persistent_queue.go
```diff
@@ -87,19 +87,19 @@ var (
 )
 
 // NewPersistentQueue creates a new queue backed by file storage; name and signal must be a unique combination that identifies the queue storage
-func NewPersistentQueue[T any](sizer Sizer[T], capacity int, dataType component.DataType, storageID component.ID,
-	marshaler func(req T) ([]byte, error), unmarshaler func([]byte) (T, error), set exporter.CreateSettings) Queue[T] {
-	_, isRequestSized := sizer.(*RequestSizer[T])
+func NewPersistentQueue[T any](set QueueCreateSettings[T], storageID component.ID, marshaler func(req T) ([]byte,
```
bogdandrutu, line 90: Why not having a PersistentQueueCreateSettings? Also, to be honest not sure if we should use `Create` here, we use in the Factory case since that is a `CreateProcessor` function, but in case of the `service` objects I think we don't use.

exporter/exporterhelper/common.go
```diff
@@ -111,7 +113,47 @@ func WithQueue(config QueueSettings) Option {
...
+		}
+		o.queueSender = newQueueSender(queue, o.set, config.NumConsumers)
+	}
+}
+
+// WithRequestQueue enables queueing for an exporter.
+// This option should be used with the new exporter helpers New[Traces|Metrics|Logs]RequestExporter.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+func WithRequestQueue(cfg configqueue.QueueConfig, queueFactory QueueFactory) Option {
```
bogdandrutu, line 136: Instead of having a new public `QueueFactory` what about we change this to `func WithRequestQueue(queue Queue, numConsumers int)`

config/configqueue/queue.go
dmitryax: `QueueConfig` is used only for exporters that don't support persistent queue. And `PersistentQueueConfig` for exporters that support both. Maybe we don't need to support memory-queue-only exporters. My idea was that some exporters don't want to provide serializers but still want to use custom requests.

config/configqueue/queue.go
dmitryax: Sounds good to me. I don't think there is any use case outside exporters

consumer/go.mod
dmitryax: It comes from `obsreport/obsreporttest` which is not a separate module. We need to get rid of that package

exporter/debugexporter/go.mod
dmitryax: I believe it should be part of https://github.com/open-telemetry/opentelemetry-collector/pull/9178. Not sure why it wasn't caught there

exporter/exporterhelper/common.go
dmitryax: Not sure how to pass `QueueCreateSettings` then. Should we move it to the Start method?

bogdandrutu: `QueueCreateSettings` - What is in there that user doesn't have access to or you think is "duplicate"?

config/configqueue/queue.go
bogdandrutu: >  My idea was that some exporters don't want to provide serializers but still want to use custom requests.

We can have some default serializers using pdata maybe in that case? May complicate too much the implementation design...

dmitryax: That would require asking user for Request->pdata translator. Right now we ask for pdata->Request only

exporter/exporterhelper/common.go
dmitryax: Yes, we would need to ask the user to pass at least `component.DataType` and `exporter.CreateSettings` to the persistent queue constructor in addition to the exporter helper itself

exporter/exporterhelper/internal/bounded_memory_queue.go
dmitryax: Submitted https://github.com/open-telemetry/opentelemetry-collector/pull/9197

exporter/exporterhelper/internal/persistent_queue.go
dmitryax: Makes sense. I renamed it to `QueueSettings/PersistentQueueSettings` in https://github.com/open-telemetry/opentelemetry-collector/pull/9197

config/configqueue/queue.go
dmitryax: I moved the new API to `exporter/exporterqueue`. Given that both `exporter/exporterhelper` and `exporter/exporterqueue` depend on `exporter/exporterqueue/internal`, I moved it as is to `exporter/internal` for now to keep the changeset simple. Once this PR is merged, I'll distribute that code.

config/configqueue/queue.go
dmitryax: Otherwise, we can apply the conversion after the queue. This can significantly complicate the implementation, as you mentioned.

exporter/exporterhelper/queue.go
dmitryax: @open-telemetry/collector-approvers, can you please share your thoughts on the user interface? The ask is to provide a way to configure queue size in different measures:
1. Number of basic items: log records, spans, data points - this should be the default measure going forward to be consistent with config for batching
2. Number of incoming requests - this is the current measure set in `queue_size`. We likely should keep it as an option even though it's not required.
3. Size in MiB - will be nice to have in the future.

**Options:**

A. Provide different fields for each option
- `max_size_items`, `max_size_requests` and `max_size_mib`
- `size_limit_items`, `size_limit_requests` and `size_limit_mib`
- `items_limit`, `requests_limit` and `mib_limit`

B. Keep `queue_size` and introduce another field for the `queue_measure` enum which can be `items`, `requests` or `MiB`.

C. Support suffixes like `10requests`, `10items` `10[B|KB|MB..]` and keep the `queue_size` field. 

Both _B_ and _C_ would require a plan on how to migrate from the default measure from `requests` to `items`.

Feel free to suggest other ideas or names for the fields.

exporter/exporterqueue/queue.go
```diff
@@ -0,0 +1,69 @@
...
+		})
+	}
+}
+
+// PersistentQueueSettings defines developer settings for the persistent queue factory.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type PersistentQueueSettings[T any] struct {
+	// Marshaler is used to serialize queue elements before storing them in the persistent storage.
+	Marshaler func(req T) ([]byte, error)
```
bogdandrutu, line 45: Extract "Marshaler" as a type and eventually deprecate `RequestMarshaler` in the exporterhelper?

exporter/internal/err.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package internal // import "go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+package internal // import "go.opentelemetry.io/collector/exporter/internal"
```
bogdandrutu, line 4: put it in a `internal/queue` package please.

exporter/exporterqueue/queue.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Queue defines a producer-consumer exchange which can be backed by e.g. the memory-based ring buffer queue
+// (boundedMemoryQueue) or via a disk-based queue (persistentQueue)
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Queue[T any] internal.Queue[T]
+
+// Settings defines settings for creating a queue.
+type Settings[T any] struct {
+	Sizer            internal.Sizer[T]
```
bogdandrutu, line 20: Anti pattern to expose internal types like this. Consider to move the definition of Sizer at least.

Also, is the sizer/type not defined by the config arguments? Why do you need to pass it here?

exporter/exporterqueue/queue.go
```diff
@@ -0,0 +1,69 @@
...
+	ExporterSettings exporter.CreateSettings
+}
+
+type Factory[T any] func(Settings[T]) Queue[T]
+
+// NewMemoryQueueFactory returns a factory to create a new memory queue.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+func NewMemoryQueueFactory[T any]() Factory[T] {
+	return func(set Settings[T]) Queue[T] {
```
bogdandrutu, line 32: Should accept Context, Settings and Config. See pattern here https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/extensions/extensions.go#L174

Where Settings are things like TelemetrySettings, etc. In your case Sizer and Capacity should be passed and calculated from Config.

exporter/exporterhelper/queue.go
bogdandrutu: Personally I like in this order (from most liked to least like: C > B > A.

> Both B and C would require a plan on how to migrate from the default measure from requests to items without affecting configs applied by users.

Why would you want to change the default measure?

dmitryax: > Main reasons (though we can also validate to not set multiple values) it is confusing if users configure multiple sizes which one to be used, and trying to implement multiple sizes is out of the scope and an overkill.

Right, I wanted to validate that only one field is specified.

> Why would you want to change the default measure?

Because the batch processor is configured in items, I think it should be the same default for the batcher sender. So, it may be confusing to have different measures by default in the ` sending_queue` and the `batching` options.

exporter/exporterqueue/queue.go
dmitryax: Defined Marshaler in this package and deprecated `RequestMarshaler`

exporter/internal/err.go
dmitryax: done

exporter/exporterqueue/queue.go
dmitryax: done

exporter/exporterqueue/queue.go
dmitryax: Removed this field 👍

exporter/exporterhelper/queue.go
dmitryax: I removed the `queue_size_items` and left only the backward compatible `queue_size` until we agree on the user config and a migration path. I'll create another issue to move this discussion

bogdandrutu (approved):

exporter/exporterhelper/queue.go
dmitryax: Submitted https://github.com/open-telemetry/opentelemetry-collector/issues/9462

PR #8857, opened 2023-11-12
[chore] Simplify initialization logic of the not dispached requests

bogdandrutu:


exporter/exporterhelper/internal/persistent_storage.go
```diff
@@ -210,17 +183,16 @@ func (pcs *persistentContiguousStorage) put(req Request) error {
 
 	reqBuf, err := pcs.marshaler(req)
 	if err != nil {
-		return err
+		return false
```
dmitryax, line 186: Why do we need to suppress the error here? I know that Produce eventually does that, but I don't think that's the right behavior. I think we should propagate the error to the queue sender

bogdandrutu: That is a valid approach and maybe we should do that. this was just trying to consolidate on one interface.

exporter/exporterhelper/internal/persistent_storage.go
```diff
@@ -183,7 +162,11 @@ func (pcs *persistentContiguousStorage) put(req any) error {
 
 	pcs.mu.Lock()
 	defer pcs.mu.Unlock()
+	return pcs.putInternal(req)
+}
 
+// putInternal is the internal version that requires caller to hold the
```
dmitryax, line 168:
```suggestion
// putInternal is the internal version that requires caller to hold the mutex lock.
```

dmitryax (approved):

PR #8858, opened 2023-11-12
[chore] Remove a time.Sleep used in testing

atoulme:


exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -127,9 +129,15 @@ func TestShutdownWhileNotEmpty(t *testing.T) {
 	q.Produce(newStringRequest("i"))
 	q.Produce(newStringRequest("j"))
 
+	go func() {
+		require.EventuallyWithT(t, func(c *assert.CollectT) {
+			assert.False(c, q.Produce(newStringRequest("x")), "cannot push to closed queue")
```
dmitryax, line 134: This `Produce` always returns false due to queue overflow. We probably need to significantly increase the capacity to ensure it's rejected due to the stopped queue

atoulme: The queue has one slot open because one item is considered read as it is blocked in processing. 9 items are in the channel. The workaround imo is to add a few more consumers, and we can even assert on the size of the queue based of those. Please see my changes.

dmitryax: Since we send a request every 10 milliseconds for 1 second (100 possible requests in total), even with 5 workers, we potentially can run into the queue overflow, especially on slow windows. I'd suggest setting 1000 queue capacity to make sure it's always rejected due to stopped queue

atoulme: that makes sense, done.

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -127,9 +129,20 @@ func TestShutdownWhileNotEmpty(t *testing.T) {
 	q.Produce(newStringRequest("i"))
 	q.Produce(newStringRequest("j"))
 
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(c, 5, q.Size())
+	}, 1*time.Second, 10*time.Millisecond)
```
dmitryax, line 134: I don't think we need this. it's not related to the test case goal

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -127,9 +129,16 @@ func TestShutdownWhileNotEmpty(t *testing.T) {
 	q.Produce(newStringRequest("i"))
 	q.Produce(newStringRequest("j"))
 
+	go func() {
```
bogdandrutu, line 132: Please comment why async, if it takes more than few seconds to understand why it needs a comment :) (and yes, it took me some time to understand why executing async).

atoulme: sure thing

dmitryax (approved):

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -107,13 +107,15 @@ func TestBoundedQueue(t *testing.T) {
 // only after Stop will mean the consumers are still locked while
 // trying to perform the final consumptions.
 func TestShutdownWhileNotEmpty(t *testing.T) {
-	q := NewBoundedMemoryQueue(10, 1)
+	q := NewBoundedMemoryQueue(1000, 5)
 
 	consumerState := newConsumerState(t)
 
+	var wg sync.WaitGroup
+	wg.Add(1)
```
bogdandrutu, line 115: For a single producer/consumer blocking code, use a channel, requires 1 less line :)

atoulme: 👍

dmitryax (approved):

PR #8859, opened 2023-11-13
[chore] remove time.Sleep from benchmarks

atoulme:
Refactor benchmarks to run as tests as well, remove use of time.Sleep in benchmarks.

### DIFF
--- a/exporter/exporterhelper/internal/bounded_memory_queue_test.go
+++ b/exporter/exporterhelper/internal/bounded_memory_queue_test.go
@@ -7,8 +7,12 @@ package internal
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -146,60 +146,79 @@ func TestShutdownWhileNotEmpty(t *testing.T) {
...
+
+	q := NewBoundedMemoryQueue(capacity, numConsumers)
+	var wg sync.WaitGroup
+	err := q.Start(context.Background(), componenttest.NewNopHost(), newNopQueueSettings(func(item Request) {
+		wg.Done()
+	}))
+	wg.Add(numberOfItems)
+	require.NoError(tb, err)
+	for j := 0; j < numberOfItems; j++ {
+		if !q.Produce(newStringRequest(fmt.Sprintf("%d", j))) {
```
dmitryax, line 216: Why do we need to benchmark queue overflow rejects? Shouldn't we have assert here instead?

atoulme: We can have an assert too. The waitGroup waits at the end for completion of all items, or for all the rejections to be processed.

dmitryax: I still don't get the reason of running benchmarks for queue overflow rejects. Can you please replace this with
```
assert.True(t, q.Produce(newStringRequest(fmt.Sprintf("%d", j)))
```
and update/remove the callers causing this assert to fail

atoulme: I'll just remove benchmarks causing overflows then, since they add no practical value.

bogdandrutu (approved):

bogdandrutu:
Please resolve conflicts.

atoulme:
Done. conflicts are resolved. Thanks!

PR #8862, opened 2023-11-13
Avoid nil queue in queue_sender, consolidate implementation

bogdandrutu:
Also, fixes `WithRetry` called with a config that is disabled.

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -40,13 +40,14 @@ func (b *baseRequestSender) setNextSender(nextSender requestSender) {
 
 type errorLoggingRequestSender struct {
 	baseRequestSender
-	logger *zap.Logger
+	logger  *zap.Logger
+	message string
 }
 
 func (l *errorLoggingRequestSender) send(req internal.Request) error {
 	err := l.baseRequestSender.send(req)
 	if err != nil {
-		l.logger.Error("Exporting failed", zap.Error(err))
+		l.logger.Error(l.message, zap.Int("dropped_items", req.Count()), zap.Error(err))
 	}
 	return err
 }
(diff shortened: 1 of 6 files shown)


exporter/exporterhelper/common.go
```diff
@@ -185,9 +184,15 @@ func newBaseExporter(set exporter.CreateSettings, signal component.DataType, req
...
-		retrySender:   &errorLoggingRequestSender{logger: set.Logger},
+		queueSender: &errorLoggingRequestSender{
+			logger:  set.Logger,
+			message: "Exporting failed. Dropping data. Try enabling sending_queue to survive temporary failures.",
+		},
+		obsrepSender: osf(obsReport),
+		retrySender: &errorLoggingRequestSender{
+			logger:  set.Logger,
+			message: "Exporting failed. Dropping data. Try enabling retries to survive temporary failures.",
+		},
```
dmitryax, line 195: We will be showing these messages in the exporters that don't have an option to enable sending_queue or retries (no `WithQueue`/`WithRetry` used). That would be confusing

dmitryax: We need to keep separate behaviors for unavailable or disabled

bogdandrutu: Good point. Let's do that then: if "disabled" we print these messages, if not called at all (a.k.a. not available) we don't print anything.

exporter/exporterhelper/queue_sender.go
```diff
@@ -211,17 +211,6 @@ func (qs *queueSender) Shutdown(ctx context.Context) error {
 
 // send implements the requestSender interface
 func (qs *queueSender) send(req internal.Request) error {
-	if qs.queue == nil {
```
dmitryax, line 214: Not sure if there a bug to be fixed. retry sender has a similar check https://github.com/open-telemetry/opentelemetry-collector/blob/23500dda16b565d5f15c535e5a9d22927e4700d2/exporter/exporterhelper/retry_sender.go#L110

exporter/exporterhelper/common.go
```diff
@@ -106,9 +107,16 @@ func WithTimeout(timeoutSettings TimeoutSettings) Option {
...
 // WithRetry overrides the default RetrySettings for an exporter.
 // The default RetrySettings is to disable retries.
-func WithRetry(retrySettings RetrySettings) Option {
+func WithRetry(config RetrySettings) Option {
 	return func(o *baseExporter) {
-		o.retrySender = newRetrySender(o.set.ID, retrySettings, o.set.Logger, o.onTemporaryFailure)
+		if !config.Enabled {
+			o.retrySender = &errorLoggingRequestSender{
+				logger:  o.set.Logger,
+				message: "Exporting failed. Dropping data. Try enabling retries to survive temporary failures.",
```
dmitryax, line 115: It's not necessary dropping data. If the batch processor is not enabled, the error can be propagated to the receiver and can be retried

exporter/exporterhelper/queue_sender.go
dmitryax: If that's the case, can you please remove this block ^ and the change log entry

bogdandrutu: Correct.

exporter/exporterhelper/common.go
bogdandrutu: I did not change the messages, as you can see they are previous messages when component were disabled. We can revisit them in a separate PR>

dmitryax (approved):

PR #8828, opened 2023-11-08
Restructure Queue to not have to deal with consumers

bogdandrutu:
## Important (read before submitting)
We are currently preparing for the upcoming 1.0 GA release. Pull requests that are not aligned with
the current roadmap https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/roadmap.md
and are not aimed at stabilizing and preparing the Collector for the release will not be accepted.

_Delete this paragraph before submitting._

**Description:** <Describe what has changed. 
Ex. Fixing a bug - Describe the bug and how this fixes the issue.
Ex. Adding a feature - Explain what this achieves.>

**Link to tracking Issue:** <Issue number if applicable>

**Testing:** < Describe what testing was performed and which tests were added.>

**Documentation:** < Describe the documentation added.>

_Please delete paragraphs that you did not use before submitting._

PR #8867, opened 2023-11-13
[exporterhelper] Remove redundant request interfaces

dmitryax:
Pass request+context through the senders pipeline similar to what we do in the collector's pipeline. Use a helper QueueRequest struct for passing requests through queues

This changes also moves the experimental Request interface to a separate package.

### DIFF
--- a/exporter/exporterhelper/internal/request.go
+++ b/exporter/exporterhelper/internal/request.go
@@ -5,32 +5,27 @@ package internal // import "go.opentelemetry.io/collector/exporter/exporterhelpe
 
 import "context"
 
-// Request defines capabilities required for persistent storage of a request
-type Request interface {
-	// Context returns the context.Context of the requests.
-	Context() context.Context
-
-	// SetContext updates the context.Context of the requests.
-	SetContext(context.Context)
-
-	Export(ctx context.Context) error
-
-	// OnError returns a new Request may contain the items left to be sent if some items failed to process and can be retried.
-	// Otherwise, it should return the original Request.
-	OnError(error) Request
-
-	// Count returns the count of spans/metric points or log records.
-	Count() int
-
-	// OnProcessingFinished calls the optional callback function to handle cleanup after all processing is finished
-	OnProcessingFinished()
-
-	// SetOnProcessingFinished allows to set an optional callback function to do the cleanup (e.g. remove the item from persistent queue)
-	SetOnProcessingFinished(callback func())
+// QueueRequest defines a request coming through a queue.
+type QueueRequest struct {
+	Request                  any
+	Context                  context.Context
+	onProcessingFinishedFunc func()
 }
 
-// RequestUnmarshaler defines a function which takes a byte slice and unmarshals it into a relevant request
-type RequestUnmarshaler func([]byte) (Request, error)
+func newQueueRequest(ctx context.Context, req any) QueueRequest {
+	return QueueRequest{
+		Request: req,
+		Context: ctx,
+	}
+}
 
-// RequestMarshaler defines a function which takes a request and marshals it into a byte slice
-type RequestMarshaler func(Request) ([]byte, error)
+// OnProcessingFinished calls the optional callback function to handle cleanup after all processing is finished
+func (qr *QueueRequest) OnProcessingFinished() {
+	if qr.onProcessingFinishedFunc != nil {
+		qr.onProcessingFinishedFunc()
+	}
+}
+
+type QueueRequestMarshaler func(req any) ([]byte, error)
+
+type QueueRequestUnmarshaler func(data []byte) (any, error)
--- a/exporter/exporterhelper/queue_sender.go
+++ b/exporter/exporterhelper/queue_sender.go
@@ -87,12 +87,13 @@ type queueSender struct {
 }
 
 func newQueueSender(config QueueSettings, set exporter.CreateSettings, signal component.DataType,
-	marshaler internal.RequestMarshaler, unmarshaler internal.RequestUnmarshaler) *queueSender {
+	marshaler RequestMarshaler, unmarshaler RequestUnmarshaler) *queueSender {
 	var queue internal.Queue
 	if config.StorageID == nil {
 		queue = internal.NewBoundedMemoryQueue(config.QueueSize, config.NumConsumers)
 	} else {
-		queue = internal.NewPersistentQueue(config.QueueSize, config.NumConsumers, *config.StorageID, marshaler, unmarshaler, set)
+		queue = internal.NewPersistentQueue(config.QueueSize, config.NumConsumers, *config.StorageID,
+			queueRequestMarshaler(marshaler), queueRequestUnmarshaler(unmarshaler), set)
 	}
 	return &queueSender{
 		fullName:       set.ID.String(),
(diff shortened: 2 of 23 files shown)


exporter/exporterhelper/request/request.go
```diff
@@ -0,0 +1,42 @@
...
+package request // import "go.opentelemetry.io/collector/exporter/exporterhelper/request"
+
+import (
+	"context"
+)
+
+// Request represents a single request that can be sent to an external endpoint.
+// This API is at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Request interface {
```
bogdandrutu, line 13: For the moment can we keep this internal, since the usage is still internal anyway.

exporter/exporterhelper/internal/request.go
```diff
@@ -3,34 +3,29 @@
...
+// OnProcessingFinished calls the optional callback function to handle cleanup after all processing is finished
+func (qr *QueueRequest) OnProcessingFinished() {
+	if qr.onProcessingFinishedFunc != nil {
+		qr.onProcessingFinishedFunc()
+	}
+}
 
-// RequestMarshaler defines a function which takes a request and marshals it into a byte slice
-type RequestMarshaler func(Request) ([]byte, error)
+func (qr *QueueRequest) Context() context.Context {
```
bogdandrutu, line 26: Why do you need this if you can access the internal member?

exporter/exporterhelper/internal/request.go
```diff
@@ -3,34 +3,29 @@
...
+	if qr.ctx == nil {
```
bogdandrutu, line 27: Always initialize it with Background?

exporter/exporterhelper/request/request.go
bogdandrutu: Maybe keep it in internal for the moment.

exporter/exporterhelper/request/request.go
```diff
@@ -0,0 +1,42 @@
...
+	// Export exports the request to an external endpoint.
+	Export(ctx context.Context) error
+	// ItemsCount returns a number of basic items in the request where item is the smallest piece of data that can be
+	// sent. For example, for OTLP exporter, this value represents the number of spans,
+	// metric data points or log records.
+	ItemsCount() int
```
bogdandrutu, line 19: One thing that you need to make sure, especially for batching is that it is very bad for some systems to split the metric across requests. Even if the limit may be at data points, you may want to create an issue for your batcher work.

exporter/exporterhelper/request/request.go
dmitryax: The usage is not only internal. This is a public interface used by the new experimental exporterhelper. We are  moving to another place

bogdandrutu: Let't do that in a separate PR then?

exporter/exporterhelper/internal/request.go
dmitryax: This is being used by the queue sender for sending the context to the next sender

exporter/exporterhelper/request/request.go
bogdandrutu: I don't think we need it in a separate package, that's why I don't want to have it.

exporter/exporterhelper/internal/request.go
bogdandrutu: Why not passing it in the callback?

exporter/exporterhelper/request/request.go
dmitryax: In this PR both public and internal packages need `Request` and other related API. If we leave it in the `exporterhelper` package, we could use `any` in the `internal` package instead of `Request`, but we also need to translate marshal/unmarshal funcs. Is that what you think of?

I like a separate package for `Request` because it makes the naming a bit nicer. Instead of `exporterhelper.RequestUnmarshaler` we get `request.Unmarshaler`

exporter/exporterhelper/internal/request.go
dmitryax: I tried that. In that case, the QueueRequest is left with `onProcessingFinishedFunc` func only. So we can pass them as three fields, but the callback function gets pretty complicated.

```
Callback func(ctx context.Context, req request.Request, onProcessingFinishedFunc func())
```

Not sure if keeping the onProcessingFinishedFunc field only warrants having the struct like 

```
type QueueRequest struct {
	request.Request
	onProcessingFinishedFunc func()
}
```

Let me know WDYT

exporter/exporterhelper/request/request.go
dmitryax: Makes sense. We can probably add comments to the batcher config recommending users to keep the difference between `MinSizeItems` and `MaxSizeItems` high and implement batching for OTLP exporter to make the best effort not to split metrics with those restrictions. WDYT?

exporter/exporterhelper/internal/request.go
bogdandrutu: Maybe keep `onProcessingFinishedFunc` as a prop on the Request and wrap the original request in case of the PersistenQueue?

exporter/exporterhelper/request/request.go
bogdandrutu: I think we already had everywhere using the `internal.Request`, and I am confused why we had to change too many things?

exporter/exporterhelper/internal/request.go
dmitryax: This is essentially the same as what's implemented in this PR, just with removed Context field, right?

```
type QueueRequest struct {
	request.Request
	onProcessingFinishedFunc func()
}
```

exporter/exporterhelper/request/request.go
dmitryax: What I'm suggesting in this PR is to stop using the `internal.Request` and use the original request in the senders along with the context. I think this change significantly simplifies code. The only place where we need to wrap it is the queue/internal package.

exporter/exporterhelper/internal/request.go
bogdandrutu: Yes and now. Also change https://github.com/open-telemetry/opentelemetry-collector/pull/8867/files#diff-667c34fb5331f087f510bdf2b6f39e626e6a9e6d9003425b26236b660e3a5be2R17 to be `Callback func(context.Context, Request)`

exporter/exporterhelper/request/request.go
bogdandrutu: What I am saying is that I don't believe long term we want request in a separate package, so better not to put it there now.

bogdandrutu: Keep in mind that go is duck typing, which means for the moment you can define the interface twice one in exporter and one in internal and everything works.

dmitryax: Ok, I moved the Request back to the exporterhelper package

exporter/exporterhelper/internal/request.go
dmitryax: I realized that we need the context to be embedded for the channel interface we want to have in https://github.com/open-telemetry/opentelemetry-collector/pull/8828

exporter/exporterhelper/internal/request.go
dmitryax: Sure, updated

exporter/exporterhelper/queue_sender.go
```diff
@@ -135,9 +136,10 @@ func (qs *queueSender) onTemporaryFailure(logger *zap.Logger, req internal.Reque
 func (qs *queueSender) Start(ctx context.Context, host component.Host) error {
 	err := qs.queue.Start(ctx, host, internal.QueueSettings{
 		DataType: qs.signal,
-		Callback: func(item internal.Request) {
-			_ = qs.nextSender.send(item)
-			item.OnProcessingFinished()
+		Callback: func(qr internal.QueueRequest) {
+			_ = qs.nextSender.send(qr.Context, qr.Request.(Request))
+			// TODO: Update OnProcessingFinished to accept error and remove the retry->queue sender callback.
```
bogdandrutu, line 141: Nice :)

bogdandrutu (approved):

PR #8878, opened 2023-11-14
Fix connector logger zap kind key

bogdandrutu:


service/internal/components/loggers.go
```diff
@@ -25,34 +22,34 @@ const (
 
 func ReceiverLogger(logger *zap.Logger, id component.ID, dt component.DataType) *zap.Logger {
 	return logger.With(
-		zap.String(zapKindKey, zapKindReceiver),
+		zap.String(zapKindKey, strings.ToLower(component.KindReceiver.String())),
```
bogdandrutu, line 25: Using ToLower just to make sure we are not changing the previous strings. The `String()` method returns upper case to be consistent with other funcs like StabilityLevel.

dmitryax (approved):

PR #8879, opened 2023-11-14
Do not add Logger field queue name, already configured

bogdandrutu:
These data are already logger fields, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/components/loggers.go#L40

* ID is under `name`
* DataType is under `data_type`


dmitryax (approved):
Nice!

PR #8885, opened 2023-11-14
[chore] remove duplicate members and funcs between persistent queue and storage

bogdandrutu:


exporter/exporterhelper/internal/persistent_storage.go
```diff
@@ -164,18 +158,31 @@ func (pcs *persistentContiguousStorage) get() (QueueRequest, bool) {
...
 	pcs.logger.Debug("Stopping persistentContiguousStorage")
 	return pcs.client.Close(ctx)
 }
 
-// put marshals the request and puts it into the persistent queue
-func (pcs *persistentContiguousStorage) put(req any) error {
+// Offer inserts the specified element into this queue if it is possible to do so immediately
+// without violating capacity restrictions. If success returns no error.
+// It returns ErrQueueIsFull if no space is currently available.
+func (pcs *persistentContiguousStorage) Offer(ctx context.Context, req any) error {
```
dmitryax, line 185: Why do we need to move Offer to the storage? I think hiding it in the embedded struct makes it less readable. I would expect to see this method on `persistentQueue` even if it's just calling the storage.put(). The terms "Offer" and "put" semantically align better with "queue" and "storage" names respectively.

bogdandrutu: I think after your change to remove the consumers from the queue, you will see that queue == storage for persistent structs.

dmitryax (approved):

PR #8896, opened 2023-11-14
Change Queue to use generics, removes some unsafe conversions

bogdandrutu:


exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -16,25 +16,25 @@ import (
 // boundedMemoryQueue implements a producer-consumer exchange similar to a ring buffer queue,
 // where the queue is bounded and if it fills up due to slow consumers, the new items written by
 // the producer are dropped.
-type boundedMemoryQueue struct {
+type boundedMemoryQueue[T comparable] struct {
```
yurishkuro, line 19: what is `comparable`?

bogdandrutu: Fixed, was trying to see if with comparable can compare with nil, but cannot and forgot to revert.

atoulme (approved):

dmitryax (approved):
Nice 👍

PR #8910, opened 2023-11-14
[exporterhelper] alternative tracing of components

codeboten:
Instead of providing an interface that is then implemented by each helper, this proposes using the existing components as is and wrapping Consume<Signal>Func in each helper. This reduces new APIs, but it does mean that the wrapping code will need to be implemented inside each helper.

Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/8833

Part of https://github.com/open-telemetry/opentelemetry-collector/issues/8804

### DIFF
--- a/exporter/exporterhelper/logs.go
+++ b/exporter/exporterhelper/logs.go
@@ -94,11 +94,13 @@ func NewLogsExporter(
 	}
 
 	lc, err := consumer.NewLogs(func(ctx context.Context, ld plog.Logs) error {
+		ctx = be.obsrep.StartLogsOp(ctx)
 		req := newLogsRequest(ld, pusher)
 		serr := be.send(ctx, req)
 		if errors.Is(serr, internal.ErrQueueIsFull) {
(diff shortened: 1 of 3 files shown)


TylerHelmuth:
@codeboten with this solution, any component that is not using the helper would have to manually instrument themselves to show up in the trace correct?

codeboten:
@TylerHelmuth correct

dmitryax:
The persistent queue drops the context, so this will break tracing if persistent queue is enabled

codeboten:
@bogdandrutu @TylerHelmuth @dmitryax @djaglowski is there a preference between this and #8833? This one has less surface API, but I guess it does mean that anyone who isn't using helpers would have to implement this call themselves.

TylerHelmuth:
@codeboten the helpers are pretty ubiquitous right?  All the components in Core and Contrib would be instrumented?

dmitryax:
@codeboten, unfortunately, neither of them will work with the enabled persistent queue. I think we need to keep it after the queue until we find a way to propagate the context through the persistent queue if we want to. But it may introduce a significant overhead.

Issue #8063, opened 2023-07-11
Publish stable modules with otel release versions and not just release candidates


braydonk:
**Is your feature request related to a problem? Please describe.**
When I want to update the `go.opentelemetry.io/collector` dependencies in my project, I'd like to just `go get` all the dependencies `@latest`. With `featuregate` and `pdata` being marked as stable, they are only being published as release candidates. This means `@latest` doesn't work properly; while all the other `collector` latest dependencies are `v0.81.0` as of writing, `featuregate` for example `latest` resolves to `v0.77.0`. This ruins the process of just updating all collector dependencies, because featuregate requires lower versions of the other depedencies meaning they resolve in various incorrect ways.

**Describe the solution you'd like**
It would be great if `latest` could resolve to the same thing for these two stable modules as all the others, even if the tags aren't exactly the same thing.

**Describe alternatives you've considered**
The current strategy will be to exclude these from the automatic upgrade process and manually grab the latest tags for these modules, and presumably all future stable modules. This is manual work in what in the past has always been a simple automatic upgrade. This also means I'll presumably need to do this with any future modules being marked stable.

**Additional context**
I am able to provide screenshots or pastebins similar to https://cloud-native.slack.com/archives/C01N6P7KR6W/p1689092712052069 if it will help.

mx-psi:
> It would be great if latest could resolve to the same thing for these two stable modules as all the others, even if the tags aren't exactly the same thing.

I think unfortunately this is how the `go` command works and there is not much we can do about it. Once we start releasing `1.x` versions for `pdata` and `featuregate` this should go away but for now I don't see a way to solve this.

braydonk:
> I think unfortunately this is how the go command works and there is not much we can do about it.

Yeah I understand that, but imo the fact that this is how the `go` command works exposes a flaw in the release process. The fact that the "latest release" in `go mod`'s eyes is not the same for the modules marked stable is pretty unintuitive in my opinion.

> Once we start releasing 1.x versions for pdata and featuregate this should go away

Is there a timeline for this? And will this same release candidate interim happen for each module being marked stable? If not, I have a proposed solution that would mitigate the problem.

If these modules that have "release candidate" versions were also simultaneously published under a "release" tag that synced up with all the other modules, this wouldn't be a problem. The restriction of not keeping release versions updated for modules that are marked stable feels like an adverse side effect.

[...]

Issue #4705, opened 2022-01-18
[Decision to be made] Approach to handle experimental types of telemetry


dmitryax:
Before declaring Collector GA we need to decide on how we handle new experimental telemetry types. Currently we have logs signal which is in Beta, and pdata parts related to logs can be changed. 

We need to decide if we can:

- **Mark unstable parts of the public API as _Experimental_.** This is the easiest approach and allow us to proceed with no changes to existing public API just marking API related to log as experimental. The same approach can be applied to any other signals that can be added in future.

- **Split all the existing packages based on telemetry type: pdata, consumer, component.** This seems to be the the most disruptive way and requires many changes. Another downside of this approach is that we will not be able add support of experimental data type in stable components like batch processor, memorylimiter processor, otlp receiver/collector with the current design. It means that those components cannot be marked as GA while logs signal is still in Beta unless we want rework them as well or separate out the logs support.

Aneurysm9:
> * **Mark unstable parts of the public API as _Experimental_.** This is the easiest approach and allow us to proceed with no changes to existing public API just marking API related to log as experimental. The same approach can be applied to any other signals that can be added in future.

Would this be done in conjunction with a change to the version of the collector and pdata modules to `v1.0.0` or not?  I do not believe that it is appropriate to use a `v1.0.0` module version but say "some parts of the public API of this module may change at any time".  The spec for [Go module versions](https://go.dev/ref/mod#versions) is rather clear:

> Each part of a version indicates whether the version is stable and whether it is compatible with previous versions.
>
>    * The major version must be incremented and the minor and patch versions must be set to zero after a backwards incompatible change is made to the module’s public interface or documented functionality, for example, after a package is removed.

I would prefer that we take the approach of separating stable modules from unstable modules.  We have already introduced the use of the [`multimod` tool](https://github.com/open-telemetry/opentelemetry-go-build-tools/tree/main/multimod) that enabled us to manage this split in the Go SDK, so I think that all of the tooling needed to make this workable already exists and is integrated.

> Another downside of this approach is that we will not be able add support of experimental data type in stable components like batch processor, memorylimiter processor, otlp receiver/collector with the current design. It means that those components cannot be marked as GA while logs signal is still in Beta unless we want rework them as well or separate out the logs support.

[...]

PR #8935, opened 2023-11-16
[docs, .github] Add new issue template for stabilizing a module

mx-psi:
**Description:** 

This adds a new release template for stabilizing a module.
The intent is to try this out for pdata and featuregate, and iterate on the template over time.

### DIFF
--- a/.github/ISSUE_TEMPLATE/stabilization.md
+++ b/.github/ISSUE_TEMPLATE/stabilization.md
@@ -13,7 +13,7 @@ Before stabilizing a module, an approver or maintainer must make sure that the f
 - [ ] No TODOs in the module code that would require breaking changes
 - [ ] No deprecated symbols in the module
 - [ ] No symbols marked as experimental in the module
-- [ ] The module follows the [Coding guidelines](CONTRIBUTING.md#coding-guidelines)
+- [ ] The module follows the [Coding guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md)
 
 Please also make sure to publicly announce our intent to stabilize the module on:
 
--- a/.github/ISSUE_TEMPLATE/stabilization.md
+++ b/.github/ISSUE_TEMPLATE/stabilization.md
@@ -1,0 +1,24 @@
+---
+name: Module stabilization
+about: Stabilize a module before a 1.0 release
+title: 'Stabilize module X'
+labels: 'stabilization'
+assignees: ''
+---
+
+Before stabilizing a module, an approver or maintainer must make sure that the following criteria are met:
+
+- [ ] One RC release or more have been done of this module
+- [ ] No open issues or PRs in the module that would require breaking changes
+- [ ] No TODOs in the module code that would require breaking changes
+- [ ] No deprecated symbols in the module
+- [ ] No symbols marked as experimental in the module
+- [ ] The module follows the [Coding guidelines](CONTRIBUTING.md#coding-guidelines)
+
+Please also make sure to publicly announce our intent to stabilize the module on:
+
+- [ ] The #otel-collector CNCF Slack Channel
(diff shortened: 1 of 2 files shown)


codeboten (approved):
Nice, I like this idea

djaglowski (approved):

.github/ISSUE_TEMPLATE/stabilization.md
```diff
@@ -0,0 +1,24 @@
...
+assignees: ''
+---
+
+Before stabilizing a module, an approver or maintainer must make sure that the following criteria are met:
+
+- [ ] One RC release or more have been done of this module
+- [ ] No open issues or PRs in the module that would require breaking changes
+- [ ] No TODOs in the module code that would require breaking changes
+- [ ] No deprecated symbols in the module
+- [ ] No symbols marked as experimental in the module
```
songy23, line 15: Curious why? Seems common for stable module to have experimental APIs, e.g. grpc-go.

.github/ISSUE_TEMPLATE/stabilization.md
```diff
@@ -0,0 +1,24 @@
...
+name: Module stabilization
+about: Stabilize a module before a 1.0 release
+title: 'Stabilize module X'
+labels: 'stabilization'
+assignees: ''
+---
+
+Before stabilizing a module, an approver or maintainer must make sure that the following criteria are met:
+
+- [ ] One RC release or more have been done of this module
```
dmitryax, line 11: Do we want to do rc releases going forward, given the reported issues?

dmitryax:
Overall LGTM. Thanks for putting this together!

.github/ISSUE_TEMPLATE/stabilization.md
mx-psi: We discussed a bit about this on #4705 and #4832. Personally, I think grpc-go's approach has caused a lot of pain to end-users; I gathered some issues related to it on https://github.com/open-telemetry/opentelemetry-collector/issues/4832#issuecomment-1039394229.

We also have the experience of opentelemetry-go, where their decision to have interfaces that can change across minor versions has caused some pain (see open-telemetry/opentelemetry-go/issues/3277)

We can discuss this more explicitly (I think having a more explicit guideline as to how to implement experimental methods is important), but for now I think requiring the 1.0 version to not have experimental symbols is not too onerous

.github/ISSUE_TEMPLATE/stabilization.md
mx-psi: I think doing one or two RC releases makes sense, but I agree that definitely 18 RC releases is too much :smile: We can add some guidance to this document as part of solving #8063 (something like "if we do more than two RCs go back to minor versions until we solve the issue(s) in question")

mx-psi:
Accidentally re-requested Alex's review, feel free to ignore :)

.github/ISSUE_TEMPLATE/stabilization.md
songy23: Thanks for the context, I agree the implication of experimental APIs can be a separate discussion. Makes sense to me to not have them for 1.0 to begin with (since relaxing this constraint in the future is feasible).

songy23 (approved):
LGTM

.github/ISSUE_TEMPLATE/stabilization.md
dmitryax: Maybe we can just require two 0.x versions with all other checkboxes marked instead of jumping back and forth to RC

codeboten: unless there's a strong argument for a RC tag, i would agree that we could call a package out as RC as of version 0.x and avoid unnecessary process and pain

codeboten (approved):
Re-approved 👍🏻

PR #8937, opened 2023-11-16
[cmd/builder] add a flag to skip generating main files

atoulme:
**Description:**
We have the need to generate list of components in tooling adjacent to the collector, such as configschema. For those, it makes sense to generate a list of components per type that can be used by the program in a programmatic way that differs from running the collector.

**Testing:**
Unit tests.

### DIFF
--- a/cmd/builder/internal/command.go
+++ b/cmd/builder/internal/command.go
@@ -23,6 +23,7 @@ const (
 	skipGenerateFlag               = "skip-generate"
 	skipCompilationFlag            = "skip-compilation"
 	skipGetModulesFlag             = "skip-get-modules"
+	skipGenerateMainFlag           = "skip-generate-main"
 	ldflagsFlag                    = "ldflags"
 	distributionNameFlag           = "name"
 	distributionDescriptionFlag    = "description"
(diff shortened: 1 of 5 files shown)


cmd/builder/internal/command.go
```diff
@@ -82,6 +83,7 @@ configuration is provided, ocb will generate a default Collector.
 	// the distribution parameters, which we accept as CLI flags as well
 	cmd.Flags().BoolVar(&cfg.SkipGenerate, skipGenerateFlag, false, "Whether builder should skip generating go code (default false)")
 	cmd.Flags().BoolVar(&cfg.SkipCompilation, skipCompilationFlag, false, "Whether builder should only generate go code with no compile of the collector (default false)")
+	cmd.Flags().BoolVar(&cfg.SkipGenerateMain, skipGenerateMainFlag, false, "Whether builder should skip generating main.go, main_others.go and main_windows.go (default false)")
```
bogdandrutu, line 86: Do we need a flag? Why?

atoulme: > We have the need to generate list of components in tooling adjacent to the collector, such as configschema.

I also need this for distros where main.go is managed manually.

jpkrohling:
@atoulme, is this still up to date? I thought we discussed this during a SIG Collector call and decided to go another route?

atoulme:
During the SIG meeting I think we said we'd apply a different route to configschema, but I'm not sure I've had time to evaluate that route.

atoulme:
With https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30187, I no longer need this. Closing.

Issue #8927, opened 2023-11-15
[pdata] Add ForEach methods to reduce deeply nested looping


djaglowski:
**Is your feature request related to a problem? Please describe.**
Drilling through pdata structs in order to perform work on each items is very common but generally requires deeply nested and repetitive code.

```go
func doSomethingWithLogs(ld plog.Logs) {
	rls := ld.ResourceLogs()
	for i := 0; i < rls.Len(); i++ {
		rl := rls.At(i)
		sls := rl.ScopeLogs()
		for j := 0; j < sls.Len(); j++ {
			sl := sls.At(j)
			for k := 0; k < sl.LogRecords().Len(); k++ {
				lr := sl.LogRecords().At(k)
				// do something with rl, sl, lr
			}
		}
	}
}
```

**Describe the solution you'd like**
We should consider adding `ForEach*` methods at various levels in the pdata hierarchy.

The same code above could look something like:

```go
func doSomethingWithLogs(ld plog.Logs) {
	ld.ForEachLogRecord(func(rl plog.ResourceLogs, sl plog.ScopeLogs, lr plog.LogRecord) {
		// do something with rl, sl, lr
	})
}
```

This could be implemented at the level of each "Slice" struct. e.g.

```go
func (es ResourceLogsSlice) ForEachResourceLogs(func(rl plog.ResourceLogs))
func (es ResourceLogsSlice) ForEachScopeLogs(func(rl plog.ResourceLogs, sl plog.ScopeLogs))
func (es ResourceLogsSlice) ForEachLogRecord(func(rl plog.ResourceLogs, sl plog.ScopeLogs, lr plog.LogRecord))

func (es ScopeLogsSlice) ForEachScopeLogs(func(sl plog.ScopeLogs))
func (es ScopeLogsSlice) ForEachLogRecord(func(sl plog.ScopeLogs, lr plog.LogRecord))

func (es LogRecordSlice) ForEachLogRecord(func(lr plog.LogRecord))
```

Optionally, we could go one step further and expose these on the structs which contain Slices. e.g.

```go
func (ms Logs) ForEachResourceLogs(func(rl plog.ResourceLogs))
func (ms Logs) ForEachScopeLogs(func(rl plog.ResourceLogs, sl plog.ScopeLogs))
func (ms Logs) ForEachLogRecord(func(rl plog.ResourceLogs, sl plog.ScopeLogs, lr plog.LogRecord))

[...]

PR #8938, opened 2023-11-16
Add Range methods to pdata slices

djaglowski:
**Description:**

This PR adds a `Range` method to each pdata slice struct.

I updated many loops throughout the codebase to demonstrate usage of the functions. I may have missed a few or applied the pattern unnecessarily in simple situations, but primarily am seeking feedback on the implementation at this point.

**Link to tracking Issue:** #8927.

**Testing:**

Tests are added for generated methods.

### DIFF
--- a/exporter/internal/otlptext/databuffer.go
+++ b/exporter/internal/otlptext/databuffer.go
@@ -89,7 +89,7 @@ func (b *dataBuffer) logMetricDataPoints(m pmetric.Metric) {
 }
 
 func (b *dataBuffer) logNumberDataPoints(ps pmetric.NumberDataPointSlice) {
-	ps.ForEachIndex(func(i int, p pmetric.NumberDataPoint) {
+	ps.Range(func(i int, p pmetric.NumberDataPoint) {
 		b.logEntry("NumberDataPoints #%d", i)
 		b.logDataPointAttributes(p.Attributes())
 
@@ -107,7 +107,7 @@ func (b *dataBuffer) logNumberDataPoints(ps pmetric.NumberDataPointSlice) {
 }
 
 func (b *dataBuffer) logHistogramDataPoints(ps pmetric.HistogramDataPointSlice) {
-	ps.ForEachIndex(func(i int, p pmetric.HistogramDataPoint) {
+	ps.Range(func(i int, p pmetric.HistogramDataPoint) {
 		b.logEntry("HistogramDataPoints #%d", i)
 		b.logDataPointAttributes(p.Attributes())
 
@@ -140,7 +140,7 @@ func (b *dataBuffer) logHistogramDataPoints(ps pmetric.HistogramDataPointSlice) 
 }
 
 func (b *dataBuffer) logExponentialHistogramDataPoints(ps pmetric.ExponentialHistogramDataPointSlice) {
-	ps.ForEachIndex(func(i int, p pmetric.ExponentialHistogramDataPoint) {
+	ps.Range(func(i int, p pmetric.ExponentialHistogramDataPoint) {
 		b.logEntry("ExponentialHistogramDataPoints #%d", i)
 		b.logDataPointAttributes(p.Attributes())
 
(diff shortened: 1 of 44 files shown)


djaglowski:
@bogdandrutu, @mx-psi, I'm curious your thoughts on this, both for the `ForEach` mentioned in #8938 and the variants which may also be useful.

exporter/internal/otlptext/databuffer.go
```diff
@@ -89,8 +89,7 @@ func (b *dataBuffer) logMetricDataPoints(m pmetric.Metric) {
 }
 
 func (b *dataBuffer) logNumberDataPoints(ps pmetric.NumberDataPointSlice) {
-	for i := 0; i < ps.Len(); i++ {
-		p := ps.At(i)
+	ps.ForEachIndex(func(i int, p pmetric.NumberDataPoint) {
```
bogdandrutu, line 92: Not sure I like the "Index" suffix.

djaglowski: Just the name, or the entire function?

If just the name, some alternatives:
- `ForEachWithIndex` (not _just_ for each index)
- `Range` (perhaps implying similarity to go `range`'s `index, item` pattern)

Will try to think of other options..

djaglowski: Alternately, we could just consolidate on this signature for `ForEach` and expect often the index is ignored.

mx-psi: I would just drop the `Index` suffix. I am also wondering if we can just have a single `ForEach` method that is just the `ForEachWhile` one (although that means a bit more verbosity with an extra `return true`)

djaglowski: I played around with this some more and found that a single `Range` function should be sufficient.
- Renamed `ForEachIndex` to `Range`
- Deleted `ForEach` in favor of `Range`. Ignoring the index is easy enough and doesn't substantially cluster the code.
- Deleted `ForEachWhile` in favor of `Range`. The loop can be short circuited easily enough by saving a value outside the loop and checking it as necessary. e.g. `var err error` outside the loop and `if err != nil { return }` inside the loop. Technically the loop still runs for each item but the cost is minimal. I think we could look at adding a `RangeWhile` function later if necessary.

mx-psi: I really like it this way, if performance is a concern one can go back to a loop so that edge case is not very concerning to me. 

Would be great to see this in other components from contrib before merging IMO (e.g. something more complex like the transform processor or prometheus exporter?) to make sure we cover possible edge cases there that are not present in components in core.

djaglowski: I looked again at simply adding a boolean return value to `Range`, but there downsides seem too much in my opinion:
- When there is no intention of short circuiting, it feels very unnatural to have to worry about returning a value
- The name `Range` implies nothing about how to interpret the return value (Should true or false mean stop iterating?)

I looked at what would change if we added a `RangeIf` (or could be `RangeWhile`, or inverse as `RangeUntil`): https://github.com/djaglowski/opentelemetry-collector-contrib/compare/pdata-range-demo...djaglowski:opentelemetry-collector-contrib:pdata-rangeif-demo

It's definitely an improvement over using a single function that cannot truly be short circuited. I also think it's much nicer having separate functions, to allow using the appropriate one depending on whether short-circuiting may be necessary.

mx-psi: Thanks for testing this! I can see how "`Range` with a return value" does not lead to a good UX. An alternative option to adding `Range[If/While/Until]` is only adding `Range` and using the for-loop style for these short-circuiting cases. So my question is: Is just having `Range` and using the current for-loop style for the short-circuiting cases good enough? If not, why not? Is mixing styles too bad or hard for contributors writing code using pdata?

mx-psi: > That said, I think that adding only Range is still a clear benefit and we could revisit Range[If/While/Until] if prompted by feedback.

If other @open-telemetry/collector-maintainers are okay with this and accept that we will add `Range[If/While/Until]` in the future if need be, I would like to do it this way.

bogdandrutu: I still believe that `"Range with a return value"` is the right API for use, because it is the golang standard way (see https://pkg.go.dev/sync#Map.Range) to deal with this, and users should be familiar with that API.

mx-psi: We have that kind of range with our own `pcommon.Map` https://pkg.go.dev/go.opentelemetry.io/collector/pdata/pcommon#Map.Range. It's unclear to me if this is the Golang standard just for maps or generally ranging over a data structure.

mx-psi:
@bogdandrutu how can we unblock this?

PR #8942, opened 2023-11-17
[chore] [exporterhelper] Remove retry sender -> queue sender callback

dmitryax:
Use returned error instead to simplify the senders feedback loop

### DIFF
--- a/.chloggen/introduce-reenque-option.yaml
+++ b/.chloggen/introduce-reenque-option.yaml
@@ -1,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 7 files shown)


dmitryax:
The coverage will be improved after https://github.com/open-telemetry/opentelemetry-collector/pull/8960 is merged

dmitryax:
This PR introduces a significant change in behavior: re-enqueueing in the persistent queue will be always enabled even if the retry sender is disabled. I'm turning it into a draft for now until it's properly addressed.

dmitryax:
> This PR introduces a significant change in behavior: re-enqueueing in the persistent queue will be always enabled even if the retry sender is disabled. I'm turning it into a draft for now until it's properly addressed.

This is now resolved by introducing a separate config option for re-enqueue.

dmitryax:
I'm going to break this PR down

dmitryax:
Closing in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/8985

PR #8947, opened 2023-11-17
[.github] Add support for stable-v1 modset

mx-psi:
**Description:** 

Adds support for a new, `stable-v1` module set. This is the module we will use for 1.0.0 releases.

The current `stable` module set is kept for release candidates (we can rename to `stable-rc` if that makes sense?).

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -5,6 +5,12 @@ on:
     # Determine the version number that will be assigned to the release. During the beta phase, we increment
     # the minor version number and set the patch number to 0.
     inputs:
+      candidate-stable-v1:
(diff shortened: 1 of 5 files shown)


mx-psi:
Hm, multimod doesn't like this. It seems like the approach from opentelemetry-go is that they do a release candidate for all stable modules (see e.g. for v1.19.0-rc.1 https://github.com/open-telemetry/opentelemetry-go/blob/v1.19.0-rc.1/versions.yaml). I feel like we can do the same here. **This means that we can't release featuregate and pdata independently, we would need to move both from RC to v1.0.0.**

Anyway, I am closing this since it doesn't work, we can try and do this if we change multimod, but we don't *really* need it this time around.

Issue #8115, opened 2023-07-20
Persistent queue fails to delete items from disk in some circumstances


swiatekm:
**Describe the bug**
The persistent queue pulls items from the on-disk database, but fails to delete them after they've been successfully sent. The result is that the db has orphaned data in it that never gets deleted.

**Steps to reproduce**
I haven't been able to reproduce this, but I have had a user send me their DB file, and have confirmed it contained items with indices less than the saved read index.

**What version did you use?**
 0.77.0

**What config did you use?**
The relevant part is:

```yaml
extensions:
  file_storage:
    directory: /var/lib/storage/otc
    timeout: 10s
    compaction:
      on_rebound: true
      directory: /tmp
```

**Environment**
The report I have is from AWS EKS 1.23 with the default storage driver.

**Additional context**
#7396 should technically help address similar problems, but this has happened quite consistently for this specific user in low disk space conditions, and the issue that change addresses should be fairly difficult to trigger.

I'm filing this in the hope that I can get more reports of this behaviour to help verify that it still occurs, and hopefully get to the bottom of the issue.

swiatekm-sumo:
@jpkrohling @djaglowski @frzifus you asked to be included during yesterday's SIG meeting.

johngmyers:
We're seeing something that looks like this with splunk-otel-collector 0.102.0, which claims to have changes from opentelemetry-collector 0.102.0. The persistent queue file is upwards of 100GB, yet the Prometheus metrics show no backlog.

swiatekm:
@johngmyers do you have access to the queue file? It's possible to inspect it with bbolt's CLI to check if there are keys outside of the queue's normal range.

johngmyers:
```
$ sudo ~/bbolt keys exporter_splunk_hec_platform_logs_logs default
254468
254469
[...]
263190
263191
263192
di
ri
wi
$ sudo ~/bbolt get -format hex exporter_splunk_hec_platform_logs_logs default di
0300000004e203000000000005e203000000000006e2030000000000
$ sudo ~/bbolt get -format hex exporter_splunk_hec_platform_logs_logs default ri
07e2030000000000
ubuntu@ip-10-93-178-122:/var/addon/splunk/exporter_queue/agent$ sudo ~/bbolt get -format hex exporter_splunk_hec_platform_logs_logs default wi
1904040000000000
```

johngmyers:
`ri` decodes to 254471, so there are three keys before that.

[...]

PR #8963, opened 2023-11-20
[exporterhelper] Fix invalid write index updates in the persistent queue

swiatekm-sumo:
**Description:**
Fixing a bug where the in-memory value of the persistent queue's write index would be updated even if writing to the storage failed. This normally wouldn't have any negative effect other than inflating the queue size temporarily, as the read loop would simply skip over the nonexistent record. However, in the case where the storage doesn't have any available space, the in-memory and in-storage write index could become significantly different, at which point a collector restart would leave the queue in an inconsistent state.

Worth noting that the same issue affects reading from the queue, but in that case the writes are very small, and in practice the storage will almost always have enough space to carry them out.

**Link to tracking Issue:** #8115

**Testing:**
The `TestPersistentQueue_StorageFull` test actually only passed by accident. Writing would leave one additional item in the put channel, then the first read would fail (as there is not enough space to do the read index and dispatched items writes), but subsequent reads would succeed, so the bugs would cancel out. I modified this test to check for the number of items in the queue after inserting them, and also to expect one fewer item to be returned.

### DIFF
--- a/exporter/exporterhelper/internal/persistent_queue.go
+++ b/exporter/exporterhelper/internal/persistent_queue.go
@@ -213,10 +213,12 @@ func (pq *persistentQueue[T]) putInternal(ctx context.Context, req T) error {
 	}
 
 	// Carry out a transaction where we both add the item and update the write index
-	setWriteIndexOp := storage.SetOperation(writeIndexKey, itemIndexToBytes(newIndex))
-	setItemOp := storage.SetOperation(itemKey, reqBuf)
-	if err := pq.client.Batch(ctx, setWriteIndexOp, setItemOp); err != nil {
-		return err
+	ops := []storage.Operation{
+		storage.SetOperation(writeIndexKey, itemIndexToBytes(newIndex)),
+		storage.SetOperation(itemKey, reqBuf),
+	}
+	if storageErr := pq.client.Batch(ctx, ops...); storageErr != nil {
+		return storageErr
 	}
 
 	pq.writeIndex = newIndex
--- a/exporter/exporterhelper/internal/persistent_queue.go
+++ b/exporter/exporterhelper/internal/persistent_queue.go
@@ -205,20 +205,25 @@ func (pq *persistentQueue[T]) putInternal(ctx context.Context, req T) error {
 	}
 
 	itemKey := getItemKey(pq.writeIndex)
-	pq.writeIndex++
+	newIndex := pq.writeIndex + 1
 
 	reqBuf, err := pq.marshaler(req)
 	if err != nil {
 		return err
 	}
-	err = pq.client.Batch(ctx,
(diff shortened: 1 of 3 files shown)


dmitryax:
Both changes make sense to me, but can you please break them down in two PRs with changelog entries?

dmitryax:
Thanks for the bug fix. 

About the second change, I’m not 100% sure that the proposed mode of handling erroneous situations is always better than the existing one. 

Can you please submit separate PRs?

swiatekm-sumo:
> Thanks for the bug fix.
> 
> About the second change, I’m not 100% sure that the proposed mode of handling erroneous situations is always better than the existing one.
> 
> Can you please submit separate PRs?

Can do, but I need to modify the test for full storage in a somewhat unintuitive way. I've done this and left a comment explaining the reasoning, let me know if it's clear.

exporter/exporterhelper/internal/persistent_queue.go
```diff
@@ -205,20 +205,25 @@ func (pq *persistentQueue[T]) putInternal(ctx context.Context, req T) error {
...
-	err = pq.client.Batch(ctx,
-		storage.SetOperation(writeIndexKey, itemIndexToBytes(pq.writeIndex)),
-		storage.SetOperation(itemKey, reqBuf))
 
+	// Carry out a transaction where we both add the item and update the write index
+	setWriteIndexOp := storage.SetOperation(writeIndexKey, itemIndexToBytes(newIndex))
+	setItemOp := storage.SetOperation(itemKey, reqBuf)
+	if err := pq.client.Batch(ctx, setWriteIndexOp, setItemOp); err != nil {
+		return err
+	}
```
bogdandrutu, line 220: I understand the return error, but why changing the way operations were constructed?

swiatekm-sumo: I wanted to write this by assigning to `err` in the condition, which is more readable with a shorter statement imo. So I defined the operations separately. Do you think it's worse now?

bogdandrutu: I think it just complicates the review, especially understanding what changed. Writing the code as you suggested makes sense in general, but here I feel that it complicates a bit (especially that you still shadow err). I would write maybe like:

```golang
ops := []storage.Operations{ storage.SetOperation(writeIndexKey, itemIndexToBytes(newIndex),  storage.SetOperation(itemKey, reqBuf)}
if err := pq.client.Batch(ctx, ops...); err != nil {
	return err
}
```

bogdandrutu (approved):

exporter/exporterhelper/internal/persistent_queue.go
swiatekm-sumo: Yeah, that looks better. Applied this change and also renamed the error variable to avoid shadowing.

dmitryax (approved):

PR #8975, opened 2023-11-21
Clarify our stance re: release candidates

### LINKED ISSUE #8063: Publish stable modules with otel release versions and not just release candidates

mx-psi:
**Description:** 

Clarifies our stance regarding RC releases, inspired by https://github.com/open-telemetry/opentelemetry-collector/pull/8935#discussion_r1397887303. In a nutshell:

- Stabilization criteria have to be met for at least two minor version releases before moving to the stable module.
- We treat these two (or more) minor releases as release candidates and do not release under the `-rc` release family.
- After these releases, we move directly to the `1.x` release family.

**Link to tracking Issue:** Fixes #8063 (together with the upcoming 1.0 release for pdata and featuregate)

cc @braydonk

### DIFF
--- a/.github/ISSUE_TEMPLATE/stabilization.md
+++ b/.github/ISSUE_TEMPLATE/stabilization.md
@@ -6,9 +6,8 @@ labels: 'stabilization'
 assignees: ''
 ---
 
-Before stabilizing a module, an approver or maintainer must make sure that the following criteria are met:
+Before stabilizing a module, an approver or maintainer must make sure that the following criteria have been met for at least two minor version releases:
 
-- [ ] One RC release or more have been done of this module
 - [ ] No open issues or PRs in the module that would require breaking changes
 - [ ] No TODOs in the module code that would require breaking changes
 - [ ] No deprecated symbols in the module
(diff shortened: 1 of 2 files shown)


braydonk (approved):
Looks like a great change, thanks!

dmitryax (approved):

dmitryax:
Fix for the failing test: https://github.com/open-telemetry/opentelemetry-collector/pull/8980

jpkrohling (approved):

codeboten (approved):
Thanks @mx-psi

PR #8982, opened 2023-11-22
[chore] [exporterhelper] Simplify queue consumption logic

dmitryax:
Remove the QueueRequest wrapper. Pass consumeFunc to the queue instead of asking for the QueueRequest wrapper with callback. The callback isn't needed for the memory queue, and the field for the callback is occupying the memory even it's always empty. The bounded queue benchmark results show the RAM savings:

Before:
```
Benchmark_QueueUsage_50000_1_50000
Benchmark_QueueUsage_50000_1_50000-10    	       1	69813548417 ns/op	 2806232 B/op	   99753 allocs/op
```

After:
```
Benchmark_QueueUsage_50000_1_50000
Benchmark_QueueUsage_50000_1_50000-10    	       1	65498055709 ns/op	 2404832 B/op	   99754 allocs/op
```

Cut from https://github.com/open-telemetry/opentelemetry-collector/pull/8942


exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -37,17 +37,22 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
...
 	default:
 		return ErrQueueIsFull
 	}
 }
 
-// Poll returns a request from the queue once it's available. It returns false if the queue is stopped.
-func (q *boundedMemoryQueue[T]) Poll() (QueueRequest[T], bool) {
+// Consume applies the provided function on the head of queue. The call blocks until there is an item available.
+// It returns false if the queue is stopped.
+func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T)) bool {
```
sfc-gh-bdrutu, line 49: return error for consume.

bogdandrutu, line 49: return error for consumeFunc.

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -37,17 +37,22 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
...
 		return nil
 	default:
 		return ErrQueueIsFull
 	}
 }
 
-// Poll returns a request from the queue once it's available. It returns false if the queue is stopped.
-func (q *boundedMemoryQueue[T]) Poll() (QueueRequest[T], bool) {
+// Consume applies the provided function on the head of queue. The call blocks until there is an item available.
+// It returns false if the queue is stopped.
```
bogdandrutu, line 48: // It returns false if the queue is stopped.

The function returns true when an item is available or false if the queue is stopped.

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -37,17 +37,22 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
...
+	case q.items <- queueRequest[T]{ctx: ctx, req: req}:
 		return nil
 	default:
 		return ErrQueueIsFull
 	}
 }
 
-// Poll returns a request from the queue once it's available. It returns false if the queue is stopped.
-func (q *boundedMemoryQueue[T]) Poll() (QueueRequest[T], bool) {
+// Consume applies the provided function on the head of queue. The call blocks until there is an item available.
```
bogdandrutu, line 47: // The call blocks until there is an item available.

bogdandrutu (approved):

exporter/exporterhelper/internal/bounded_memory_queue.go
dmitryax: What would we do with that? Quque.Consume reports a bool when the queue is stopped. My idea is that it's the responsibility of the client to handle the consumeErr before in a wrapped `consumeFunc`

bogdandrutu: Don't you need that, to get rid of the `https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queue_sender.go#L117`?

dmitryax: No. I initially tried to move the error handling inside the queue but realized that it would be a significant change in the behavior: we will always re-enqueue to the persistent queue even if the retry sender is disabled. I believe this should be configurable on the queue sender, that way, we need to keep the error handling in the queue sender. I updated https://github.com/open-telemetry/opentelemetry-collector/pull/8942 with this approach

bogdandrutu: `we will always re-enqueue to the persistent queue even if the retry sender is disabled` only if it is a not persistent error.

exporter/exporterhelper/internal/bounded_memory_queue.go
dmitryax: Updated 👍

PR #8985, opened 2023-11-22
[chore] [exporterhelper] Remove retry sender -> queue sender callback

dmitryax:
Use returned error instead to simplify the senders feedback loop. This change preserves the behavior. Re-enqueueing of the temporary failures depends on the enabled retry sender. This will be changed in the next step when re-queueing becomes a configurable option

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -91,7 +91,7 @@ func WithRetry(config RetrySettings) Option {
 			}
 			return
 		}
-		o.retrySender = newRetrySender(config, o.set, o.onTemporaryFailure)
+		o.retrySender = newRetrySender(config, o.set)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


exporter/exporterhelper/common.go
```diff
@@ -181,6 +176,15 @@ func newBaseExporter(set exporter.CreateSettings, signal component.DataType, req
...
 
+	// If retry sender is disabled then disable requeuing in the queue sender.
+	// TODO: Make re-enqueuing configurable on queue sender instead of relying on retry sender.
+	if qs, ok := be.queueSender.(*queueSender); ok {
+		// if it's not retrySender, then it is disabled.
+		if _, ok = be.retrySender.(*retrySender); !ok {
+			qs.requeuingEnabled = false
+		}
+	}
+
```
bogdandrutu, line 187: I am not sure that this was intentional or a bug of the previous implementation. Happy to keep it as it is now, but my bet is that it was just a bug in the previous implementation.

bogdandrutu (approved):

PR #8988, opened 2023-11-23
[service] use `WithNamespace` instead of `WrapRegistererWithPrefix`

codeboten:
Using this functionality in the otel prom exporter fixes a bug where the target_info was prefixed as otelcol_target_info previously.

### DIFF
--- a/service/internal/proctelemetry/config.go
+++ b/service/internal/proctelemetry/config.go
@@ -203,9 +203,8 @@ func initPrometheusExporter(prometheusConfig *config.Prometheus, asyncErrorChann
 	if prometheusConfig.Port == nil {
 		return nil, nil, fmt.Errorf("port must be specified")
 	}
-	wrappedRegisterer := prometheus.WrapRegistererWithPrefix("otelcol_", promRegistry)
 	exporter, err := otelprom.New(
-		otelprom.WithRegisterer(wrappedRegisterer),
+		otelprom.WithRegisterer(promRegistry),
 		// https://github.com/open-telemetry/opentelemetry-collector/issues/8043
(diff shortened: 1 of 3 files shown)


service/internal/proctelemetry/config.go
```diff
@@ -203,16 +203,16 @@ func initPrometheusExporter(prometheusConfig *config.Prometheus, asyncErrorChann
...
-		otelprom.WithRegisterer(wrappedRegisterer),
+		otelprom.WithRegisterer(promRegistry),
 		// https://github.com/open-telemetry/opentelemetry-collector/issues/8043
 		otelprom.WithoutUnits(),
 		// Disabled for the moment until this becomes stable, and we are ready to break backwards compatibility.
 		otelprom.WithoutScopeInfo(),
 		otelprom.WithProducer(opencensus.NewMetricProducer()),
 		// This allows us to produce metrics that are backwards compatible w/ opencensus
 		otelprom.WithoutCounterSuffixes(),
+		otelprom.WithNamespace("otelcol_"),
```
bogdandrutu, line 215: Do you need the "_" suffix? Is it not added automatically?

codeboten: You're right, it's added automatically https://github.com/open-telemetry/opentelemetry-go/blob/47ba653e6962948dab3a481231c469387f11f184/exporters/prometheus/config.go#L144

codeboten, line 215:
```suggestion
		otelprom.WithNamespace("otelcol"),
```

mx-psi (approved):

Issue #6624, opened 2022-11-25
Flaky test: TestQueuedRetry_RequeuingEnabled


codeboten:
Seen this failure a couple of times: 

https://github.com/open-telemetry/opentelemetry-collector/actions/runs/3524647585/jobs/5910347724
https://github.com/open-telemetry/opentelemetry-collector/actions/runs/3521003167/jobs/5951483003

```
=== RUN TestQueuedRetry_RequeuingEnabled
panic: sync: negative WaitGroup counter
goroutine 178 [running]:
sync.(*WaitGroup).Add(0xc000046730, 0xffffffffffffffff)
/opt/hostedtoolcache/go/1.18.7/x64/src/sync/waitgroup.go:80 +0x1f8
sync.(*WaitGroup).Done(...)
/opt/hostedtoolcache/go/1.18.7/x64/src/sync/waitgroup.go:105
go.opentelemetry.io/collector/exporter/exporterhelper.(*observabilityConsumerSender).send(0xc000094e40, {0x1296340, 0xc00015bd80})
/home/runner/work/opentelemetry-collector/opentelemetry-collector/exporter/exporterhelper/queued_retry_test.go:683 +0x108
go.opentelemetry.io/collector/exporter/exporterhelper.(*queuedRetrySender).start.func1({0x1296340, 0xc00015bd80})
/home/runner/work/opentelemetry-collector/opentelemetry-collector/exporter/exporterhelper/queued_retry.go:205 +0x65
go.opentelemetry.io/collector/exporter/exporterhelper/internal.(*boundedMemoryQueue).StartConsumers.func1()
/home/runner/work/opentelemetry-collector/opentelemetry-collector/exporter/exporterhelper/internal/bounded_memory_queue.go:61 +0x144
created by go.opentelemetry.io/collector/exporter/exporterhelper/internal.(*boundedMemoryQueue).StartConsumers
/home/runner/work/opentelemetry-collector/opentelemetry-collector/exporter/exporterhelper/internal/bounded_memory_queue.go:56 +0x68

FAIL go.opentelemetry.io/collector/exporter/exporterhelper 0.421s
```

Saggi432:
@codeboten I am new to Opentelemetry .. but willing to work on this issue.. pls lmk.. if I can.

codeboten:
Thanks @Saggi432! I've assigned it to you

Abhishek-569:
[...]

PR #8989, opened 2023-11-23
bug: fix flaky test TestQueuedRetry_RequeuingEnabled

berkeli:
**Description:** fixes a flaky test 
Fixing a bug - TestQueuedRetry_RequeuingEnabled has been failing occasionally. Upon investigation, `be.send` calls the underlying `observabilityConsumerSender.send` which calls `wg.Done()`, so we need to add `wg.Add(1)` before that

**Link to tracking Issue:** #6624

**Testing:** Ran the same test multiple times before/after change: `go test -race -run TestQueuedRetry_RequeuingEnabled -count 1000`

**Documentation:** N/A


codeboten (approved):
awesome! thanks for the fix @berkeli

Issue #8987, opened 2023-11-23
[exporterhelper] Make the re-enqueuing configurable on the queue sender 


dmitryax:
Currently, the exporter helper puts failed (non-permanent error) requests back to the beginning of the queue only if retry_on_failure and persistent queue are enabled. There are a few problems with that:
1. There is no way to use the persistent queue and drop requests that the retry sender failed to deliver after several attempts
2. There is no way to configure the persistent queue with re-enqueueing but with disabled exponential backoff retry.
3. There is no way to configure re-enqueueing with the memory queue, which potentially still can be useful for someone who wants to make sure that requests are never dropped after reaching the queue as long as collector is up and running.
4. The queue sender (`sending_queue` configuration section) behavior is affected by the retry sender (`retry_on_failure` configuration section).

Proposed solution is to add an option to the `sending_queue` section:
```golang
	// RequeueOnFailure indicates whether requests are requeued or dropped on non-permanent failures.
	// If true, a request failed with a non-permenent error will be put back in the queue and 
	// retried after the current queue is drained.
	// If false (by default), items will be dropped after any failed send operation.
	// Setting this option to true means that the request will never leave the exporter queue until it's 
	// successfully exported or failed with permanent error.
	// Keep in mind that this option is independent of the `retry_on_failure`, which consumes 
	// requests from the queue and apply blocking exponential backoff retry mechanism.
	RequeueOnFailure bool `mapstructure:"requeue_on_failure"`
```

Introducing this option also requires making sure that the re-enqueue mechanism works well with the memory queue shutdown, which currently can lead to data races. See https://github.com/open-telemetry/opentelemetry-collector/issues/8124

dmitryax:
Not needed anymore. See https://github.com/open-telemetry/opentelemetry-collector/issues/8382

PR #8993, opened 2023-11-27
[exporterhelper] Make the re-enqueue behavior configurable

dmitryax:
Introduce a new option `sending_queue::requeue_on_failure` to control the requeuing independently of the retry sender. This can be useful for users who don't want the blocking exponential retry and just want to put the failed request back in the queue. This option can also be enabled with the memory queue now, which means that the data will never be dropped after getting to the queue as long as the collector is up and running.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/8987

### DIFF
--- a/exporter/exporterhelper/queue_sender.go
+++ b/exporter/exporterhelper/queue_sender.go
@@ -41,6 +41,14 @@ type QueueSettings struct {
 	// StorageID if not empty, enables the persistent storage and uses the component specified
 	// as a storage extension for the persistent queue
 	StorageID *component.ID `mapstructure:"storage"`
+	// RequeueOnFailure indicates whether requests are requeued or dropped on non-permanent failures.
+	// Do not confuse this option with the retry_on_failure which consumes
+	// requests from the queue and apply blocking exponential backoff retry mechanism.
+	// This option is applied after all the retries configured with retry_on_failure are exhausted.
+	// If true, a request failed with a non-permanent error will be put back in the queue and
+	// retried after the current queue is drained.
+	// If false (default), all failed requests will be dropped.
+	RequeueOnFailure bool `mapstructure:"requeue_on_failure"`
 }
 
 // NewDefaultQueueSettings returns the default settings for QueueSettings.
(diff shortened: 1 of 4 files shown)


exporter/exporterhelper/queue_sender.go
```diff
@@ -41,6 +41,14 @@ type QueueSettings struct {
...
 	// as a storage extension for the persistent queue
 	StorageID *component.ID `mapstructure:"storage"`
+	// RequeueOnFailure indicates whether requests are requeued or dropped on non-permanent failures.
+	// Do not confuse this option with the retry_on_failure which consumes
+	// requests from the queue and apply blocking exponential backoff retry mechanism.
+	// This option is applied after all the retries configured with retry_on_failure are exhausted.
+	// If true, a request failed with a non-permanent error will be put back in the queue and
+	// retried after the current queue is drained.
+	// If false (default), all failed requests will be dropped.
+	RequeueOnFailure bool `mapstructure:"requeue_on_failure"`
```
splunkericl, line 51: Any reasons we want `requeue_on_failure` rather than `drop_data_on_failure`?
- By default, users don't like data loss. If they miss out on this configuration they would have data loss unexpectedly.
- Having `drop_data_on_failure` with default to false would make this change backward compatible.

dmitryax: But that's the case for the persistent queue only. Memory queue has the opposite behavior.

dmitryax: It's not even always applicable to the persistent queue. Only if `retry_on_failure::enabled=true`

dmitryax: We can probably do some transition period when this option is preset if persistent queue + retry_on_failure are enabled with warning asking users to set this explicitly

splunkericl: hmm i see. so this is also aligning the behavior between memory/persistent queue. yeah having some WARN logs would probably be helpful in this transition time.

dmitryax:
https://github.com/open-telemetry/opentelemetry-collector/pull/9054 should be merged first, otherwise requeue won't work very well with the memory queue

dmitryax:
We decided to remove requeuing instead https://github.com/open-telemetry/opentelemetry-collector/pull/9090

Issue #8770, opened 2023-10-30
[cmd/builder] Use runtime.DebugInfo to show version


mx-psi:
As suggested in https://github.com/gordonklaus/ineffassign/issues/53#issuecomment-753688189 we can use `runtime.DebugInfo` to get the version when installing via `go install`. This avoids the confusion of #8691. We should check our official builds to avoid breaking the version on non-`go install` binaries (in the worst case, we can check which of the two we are in)

devoc09:
@mx-psi Hi! Is it okay if I take on this issue?

mx-psi:
@devoc09 assigned to you :)

atoulme:
Closing as completed, thanks!

PR #8994, opened 2023-11-27
refactor(builder): change output of `version` subcommand.

devoc09:
With this change, running `builder version` on binaries installed with `go install` will output the version specified at the suffix.

If it's under development, the following output will be obtained.
```
$ GO111MODULE=on CGO_ENABLED=0 go run . version
ocb version (devel)
```

**Link to tracking Issue:** #8770


cmd/builder/internal/version.go
```diff
@@ -5,22 +5,37 @@ package internal // import "go.opentelemetry.io/collector/cmd/builder/internal"
...
+}
+
 func versionCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "version",
 		Short: "Version of ocb",
 		Long:  "Prints the version of the ocb binary",
 		Run: func(cmd *cobra.Command, args []string) {
-			cmd.Println(fmt.Sprintf("%s version %s", cmd.Parent().Name(), version))
+			cmd.Println(fmt.Sprintf("%s version %s", cmd.Parent().Name(), binVersion()))
```
bogdandrutu, line 38: Use `RunE` and return an error?

cmd/builder/internal/version.go
```diff
@@ -5,22 +5,37 @@ package internal // import "go.opentelemetry.io/collector/cmd/builder/internal"
...
 
+// binVersion returns the version of the binary.
+// If the version is not set, it attempts to read the build information.
+func binVersion() string {
+	if version != "" {
+		return version
+	}
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		log.Fatal("Failed to read build info")
```
bogdandrutu, line 27: Return an error instead of crashing.

cmd/builder/internal/version.go
devoc09: Thank you for the review.
I have changed it to use `RunE` in order to return an error.

https://github.com/open-telemetry/opentelemetry-collector/pull/8994/files#diff-c3eff1de420450a694b19343e4543ca062f42962c968262c00d776504810861cR37-R41

cmd/builder/internal/version.go
devoc09: Changed to return errro.

https://github.com/open-telemetry/opentelemetry-collector/pull/8994/files#diff-c3eff1de420450a694b19343e4543ca062f42962c968262c00d776504810861cR21-R29

devoc09:
I have combined the commits into one and pushed it, so please check again.

bogdandrutu (approved):

bogdandrutu:
Please add a changelog entry.

devoc09:
> Please add a changelog entry.

Added changelog. Please review again 🙏

cmd/builder/internal/version.go
```diff
@@ -5,22 +5,42 @@ package internal // import "go.opentelemetry.io/collector/cmd/builder/internal"
...
 import (
 	"fmt"
+	"runtime/debug"
 
 	"github.com/spf13/cobra"
 )
 
 var (
-	version = "dev"
+	version = ""
```
TylerHelmuth, line 14: Do we still need this variable?  I don't believe it is ever set, only read.  If thats true then the

```
if version != "" {
    return version, nil
}
```

is unneeded.

TylerHelmuth: I think we can replace any use of this variable with `binVersion()`

Issue #4885, opened 2022-02-18
Investigate forking-renovate instead of dependabot


bogdandrutu:
Thanks to @dyladan who suggested to use https://github.com/apps/forking-renovate for dependency upgrades, since this bot creates PRs in a separate fork, which will pass the default branch protection rule added by the CNCF.

bogdandrutu:
Also, in the meantime asked for an exception to remove easyCLA status check for branches created by dependabot, which will be merged in the `main` branch anyway. See https://cloud-native.slack.com/archives/CK8QVEXQB/p1645216299079989

codeboten:
After investigating using renovatebot, i think there are some things we could benefit from here. The main use of dependabot today is for it to create a noisy list of PRs that are then used to create a single PR to update all the dependencies. This is a problem because:

* many PRs  every week that cannot be acted on is noisy to maintainers/approvers
* it creates an unnecessary strain on github actions as each PR triggers a series of checks (even if it skips most)
* the creation of the PR to update dependencies is done by manually triggering a gh action which creates unnecessary toil

I propose we start using renovatebot with the following configuration. This allows to start using renovabot for all other dependencies than gomod dependencies. This means that approvers/maintainers can at least update github actions and docker dependencies by approving and merging renovatebot PRs.

Since we cannot merge gomod updates as is (because we require `make gotidy` to run, this requires a standalone runner either for dependabot or renovatebot) i propose we enable gomod checks later and group them as a single update. Although this will still require an update to the PR to run `make gotidy`, the work required to do this should be more minimal than what's required today.

```json
{
  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
  "labels": [
    "dependencies"
  ],
  "extends": [
    "config:base"
  ],
  "packageRules": [
    {
      "matchManagers": ["dockerfile"],
      "groupName": "dockerfile deps"
    },
    {
      "matchManagers": ["docker-compose"],
      "groupName": "docker-compose deps"
    },
    {
      "matchManagers": ["github-actions"],
      "groupName": "github-actions deps"
    },
    {
      "matchManagers": ["gomod"],
      "enabled": false
    }
  ]
}
```

PR #9009, opened 2023-11-28
[chore] remove dependabot for gomod

### LINKED ISSUE #8936: Enable renovatebot for this repo (and contrib)
After spending more time testing out renovatebot for this repository, I think it's an overall improvement over what we currently have.

### Dependabot workflow

1. dependabot opens a new PR for every dependency for every module in the repo (this is super noisy)
2. an approver/maintainer kicks off the "Automation - Dependabot PR" action
3. that action gets a list of all dependabot PRs, runs update-dep and gotidy

This works in some cases, but often times, the PR produced runs into a variety of issues.

#### Issues w/ this workflow
1. any of the dependency requires code changes causing a failure in the CI
2. some deps are currently known to cause issues (exp update for example causes prometheus issues)
3. dependabot runs into a limit of components it supports, which we run into in the contrib repo.
4. it's annoying to have someone remember to trigger a CI manually, this could be automated

An alternative would be to use Renovatebot.

### Renovatebot workflow
1. A single PR per dependency or per monorepo that combines multiple deps is opened for all components.
2. The workflow "Project: Tidy" is run as part of the CI for the PR above to run gotidy and pushes a change to the PR's branch

#### Issues w/ this workflow
Now ideally, this is where an approver could review the PR and approve it to get it merged. Unfortunately, it isn't quite so simple. Once the "Project: Tidy" has completed, the change does *not* trigger CI again. This is because the standard token (GITHUB_TOKEN) used to push the change does not have the ability to trigger github actions. This is a known limitation of the token. The solution is to obtain a token with write access to the repository. However, there isn't a way to limit write access of a token to only a branch.

A write token has access to various other things, including publishing release. This was specifically called out as a problem when I opened a [community request](https://github.com/open-telemetry/community/issues/1503) to obtain such a token. So, there are a few options we have here and I would like other contributors' input on what steps we should follow here:

#### Options
1. we could obtain a write token, but this require us to not publish any binary artifacts in this repository. Currently the only thing we publish is the builder, which I've proposed we should move to releases some time ago.
2. we keep the current dependabot workflow
3. we switch to renovatebot, and reviewers have to close/re-open (or we could do things like label/unlabel the PR) to trigger CI

codeboten:
Finally rid us of the hundreds of notifications.

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/4885
Closes #8936


dmitryax (approved):
Sweet! 🎉 Thank you, @codeboten!

PR #9014, opened 2023-11-29
[docs] Document max_request_body_size setting of HTTPServerSettings

andreasgerstmayr:
**Documentation:**
Document `max_request_body_size` setting of `HTTPServerSettings`: https://github.com/open-telemetry/opentelemetry-collector/blob/daef971a0b3087ed76ebe9fd4793877b41b65306/config/confighttp/confighttp.go#L241

PR #9020, opened 2023-11-30
[chore] make genotelcorecol

atoulme:


mx-psi:
I don't understand what is broken nor why does this fix it. The first PR where this started failing is #9014, yet this is a docs only PR, so I assume something changed upstream. Do we know what happened? I am kind of hesitant to merge this without understanding it better

atoulme:
No idea, but this is similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/29570

dmitryax:
this is a result of removing the replace statements for the google sdk. there is an issue for that https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/28839

cc @codeboten , we need to see how we can fix this

dmitryax (approved):

PR #9022, opened 2023-11-30
Introduce HTTP2 health check transport options

dloucasfx:
**Description:** 
This PR introduces options to enable `http/2` health check by exposing `HTTP2ReadIdleTimeout` and `HTTP2PingTimeout`  

The golang issues are:
https://github.com/golang/go/issues/59690
https://github.com/golang/go/issues/36026

In summary, if due to environmental issue the underlying tcp connection used by the http/2 client in the exporter became unstable/unusable/unreachable, unlike http/1, the http/2 client does not forcibly close the connection and redial a new one, instead it keeps using it for 15 minutes  (default value of OS `tcp_retries2` ) until the OS cleans it up and a new tcp connection gets established. 
From OTEL user perspective, one will see a spike in export failures/timeouts for ~15 minutes, this will happen for every connection that got into a bad state, after 15 minutes things will recover until next time the tcp connection gets into a bad state.

**Testing:**
- Run OTEL with one of the exporters that uses HTTP/2 client, example `signalfx` exporter
- For simplicity use a single pipeline/exporter
- In a different shell, run this to watch the tcp state of the established connection
```
 while (true); do echo date; sudo netstat -anp | grep -E '<endpoin_ip_address(es)>' | sort -k 5; sleep 2; done
 ```  
- From the netstat, take a note of the source port and the source IP address
- replace <> from previous step
`sudo iptables -A OUTPUT -s <source_IP> -p tcp --sport <source_Port> -j DROP`
- Note how the OTEL exporter export starts timing out

Expected Result:
- A new connection should be established, similarly to http/1 and exports should succeed

Actual Result: 
- The exports keep failing for  ~ 15 minutes or for whatever the OS `tcp_retries2` is configured to
- After 15 minutes, a new tcp connection is created and exports start working

**Documentation:** 
Readme is updated with new settings

### CODE REVIEW

File: config/confighttp/confighttp.go [hunk 1]
```diff
@@ -86,6 +87,16 @@ type HTTPClientSettings struct {
 	// connection for every request. Before enabling this option please consider whether changes
 	// to idle connection settings can achieve your goal.
 	DisableKeepAlives bool `mapstructure:"disable_keep_alives"`
+
+	// This is needed in case you run into
+	// https://github.com/golang/go/issues/59690
+	// https://github.com/golang/go/issues/36026
+	// HTTP2ReadIdleTimeout if the connection has been idle for the configured value send a ping frame for health check
+	// 0s means no health check will be performed.
```
atoulme, line 95: this is also using the idle conn timeout default, is that right?

.chloggen/http2ping.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: [config/confighttp]
```
atoulme, line 7:
```suggestion
component: config/confighttp
```

config/confighttp/confighttp.go
```diff
@@ -147,6 +158,17 @@ func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.
 
 	transport.DisableKeepAlives = hcs.DisableKeepAlives
 
+	if hcs.HTTP2ReadIdleTimeout != nil && hcs.HTTP2ReadIdleTimeout.Seconds() > 0 {
```
rmfitzpatrick, line 161: rather than silently discarding negative values does returning an error make more sense?

rmfitzpatrick (approved):

config/confighttp/confighttp.go
```diff
@@ -147,6 +158,17 @@ func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.
...
+		transport2, err := http2.ConfigureTransports(transport)
+		if err != nil {
+			return nil, fmt.Errorf("failed to configure http2 transport: %w", err)
+		}
+		transport2.ReadIdleTimeout = *hcs.HTTP2ReadIdleTimeout
+		if hcs.HTTP2PingTimeout != nil {
```
rmfitzpatrick, line 167: these are ignored if the idle timeout isn't also set, which functionally is the same tmk but may be more reflective to configure an http2 transport if either value isn't nil.

config/confighttp/confighttp.go
dloucasfx: It's configurable and I think it should be different from the idle connection timeout; our sfx idle timeout it 30 seconds, this should be equal or close to the regular timeout , ie:10 seconds
```
        // ReadIdleTimeout is the timeout after which a health check using ping
	// frame will be carried out if no frame is received on the connection.
	// Note that a ping response will is considered a received frame, so if
	// there is no other traffic on the connection, the health check will
	// be performed every ReadIdleTimeout interval.
	// If zero, no health check is performed.
```

.chloggen/http2ping.yaml
dloucasfx: Thanks!

config/confighttp/confighttp.go
dloucasfx: Actually I was not thinking of negative values; I just needed to check if it's not 0, as 0 means the healthcheck is not performed so no need to execute that block.
```
        // ReadIdleTimeout is the timeout after which a health check using ping
	// frame will be carried out if no frame is received on the connection.
	// Note that a ping response will is considered a received frame, so if
	// there is no other traffic on the connection, the health check will
	// be performed every ReadIdleTimeout interval.
	// If zero, no health check is performed.
```

config/confighttp/confighttp.go
dloucasfx: @rmfitzpatrick 
> these are ignored if the idle timeout isn't also set

are you referring to the "read idle timeout" ? if yes, these comes together, in other words, if `ReadIdleTimeout` is set to 0 or not set, we should not set the ping timeout as the health check will not be performed

config/confighttp/confighttp.go
rmfitzpatrick: If a user manually sets a valid value, as `0s` is even though it's effectively a noop, there's no need to restrict its application in config imo.

config/confighttp/confighttp.go
rmfitzpatrick: >we should not set the ping timeout 

This is the flow of the config application I'm referring to, but it may be a matter of preference. If the aim is to "respect user input" then it'd make sense to set the values specified and leave the semantics to the underlying transport.

dloucasfx: same as previous comment, the idea why adding the extra call, ie `ConfigureTransports` , if knowing that it will be ignored

config/confighttp/confighttp.go
rmfitzpatrick: not certain but I think the default transport already calls it via the h2_bundle since `ForceAttemptHTTP2` is true and it specifies a dialer. I think we're having a preferential discussion so it's not a big deal to me (approved and appreciate the fix).

config/confighttp/confighttp.go
rmfitzpatrick: likely covered by https://github.com/open-telemetry/opentelemetry-collector/pull/9022#discussion_r1412504732, but because the user specified it

config/confighttp/confighttp.go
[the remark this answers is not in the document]
dloucasfx: yes, makes sense. PR is updated.
Thanks

Issue #7454, opened 2023-03-29
Work needed to enable `useOtelForInternalMetrics` by default


codeboten:
This issue is to discuss the work needed to enable using the OTel SDK by default. 

Related issues:

- [ ] https://github.com/open-telemetry/opentelemetry-collector/issues/816

codeboten:
@jmacd I believe this PR will help w/ #3: https://github.com/open-telemetry/opentelemetry-collector/pull/7483

The issue is that currently both OC and OT are using prometheus as a bridge to allow components to continue using  opencensus instrumentation. The change I'm proposing disables the OC configuration when `useOtelForInternalMetrics` is enabled and relies on the oc to otel bridge instead.

mx-psi:
We may need to address #7517 before closing this

mx-psi:
The `check_collector_pipeline` functionality in the health check extension depends on OpenCensus, see [here]( https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e42f386e07d6b2099ff82d664dc7e54d6f2af0b4/extension/healthcheckextension/exporter.go#L14). We would need to either deprecate this functionality or update it to use OTel before enabling `useOtelForInternalMetrics`

codeboten:
With the PR for `_total` suffix merging yesterday, I started looking at the second issue @jmacd mentioned in his comment above. With opencensus telemetry, service labels are applied on each metric:

```
otelcol_exporter_queue_capacity{exporter="otlp",service_instance_id="aae63e2f-cad2-4224-bbb3-806689b0a2d0",service_name="otelcontribcol",service_version="0.89.0-dev"} 1000
```

With `otelForInternalMetrics` enabled, all that information is set in the target_info:

```
otelcol_target_info{service_instance_id="73a547db-f135-45c4-870e-cde704b3cf7d",service_name="otelcontribcol",service_version="0.89.0-dev"} 1
```

I'm not sure that there's a good mechanism today for applying info from target_info as resource attributes. @dashpole pointed me to a spec change proposed to address this for prometheus exporters https://github.com/open-telemetry/opentelemetry-specification/pull/3761 (thanks David!)

I don't know what the right way to move forward with this change is with the current state. I did test the OTLP exporter for internal metrics and the resource attributes contained the service information I expected. The following is the output from the Go stdout exporter:

```
  "Resource": [
    {
      "Key": "service.instance.id",
      "Value": {
        "Type": "STRING",
[...]

PR #9037, opened 2023-12-04
make useOtelForInternalMetrics as beta

codeboten:
The metrics are now consistent with the metrics produced by OpenCensus. We should move the featuregate forward.

Note that the OpenTelemetry generated metrics includes grpc client/server metrics (for receivers/exporters that use grpc) and `target_info` metrics

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/7454

### DIFF
--- a/.chloggen/codeboten_make-otel-default.yaml
+++ b/.chloggen/codeboten_make-otel-default.yaml
@@ -1,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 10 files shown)


codeboten:
While testing this change I came across a discrepancy in how the `otelcol_exporter_sent_spans` counter is measured with and without the feature gate enabled. Will leave this in draft until I get to the bottom of it.

codeboten:
No longer blocked, it was decided to move ahead without #9065

bogdandrutu (approved):

TylerHelmuth (approved):
🎉

PR #9048, opened 2023-12-05
[exporterhelper] fix missed metric aggregations

codeboten:
This PR ensures that context cancellation in the exporter doesn't interfere with metric aggregation. The OTel SDK currently returns if there's an error in the context used in `Add` (see https://github.com/open-telemetry/opentelemetry-go/blob/6cee2b4a4c76b581115d0d0ca150ad8b2e683db6/sdk/metric/instrument.go#L241-L243). This means that if there's a cancelled context in an export, the metrics are not recorded.

### DIFF
--- a/exporter/exporterhelper/obsexporter.go
+++ b/exporter/exporterhelper/obsexporter.go
@@ -154,7 +154,7 @@ func (or *ObsReport) StartTracesOp(ctx context.Context) context.Context {
 // EndTracesOp completes the export operation that was started with StartTracesOp.
 func (or *ObsReport) EndTracesOp(ctx context.Context, numSpans int, err error) {
 	numSent, numFailedToSend := toNumItems(numSpans, err)
-	or.recordMetrics(ctx, component.DataTypeTraces, numSent, numFailedToSend)
+	or.recordMetrics(context.WithoutCancel(ctx), component.DataTypeTraces, numSent, numFailedToSend)
 	endSpan(ctx, err, numSent, numFailedToSend, obsmetrics.SentSpansKey, obsmetrics.FailedToSendSpansKey)
 }
 
@@ -169,7 +169,7 @@ func (or *ObsReport) StartMetricsOp(ctx context.Context) context.Context {
 // StartMetricsOp.
 func (or *ObsReport) EndMetricsOp(ctx context.Context, numMetricPoints int, err error) {
 	numSent, numFailedToSend := toNumItems(numMetricPoints, err)
-	or.recordMetrics(ctx, component.DataTypeMetrics, numSent, numFailedToSend)
+	or.recordMetrics(context.WithoutCancel(ctx), component.DataTypeMetrics, numSent, numFailedToSend)
 	endSpan(ctx, err, numSent, numFailedToSend, obsmetrics.SentMetricPointsKey, obsmetrics.FailedToSendMetricPointsKey)
 }
 
@@ -183,7 +183,7 @@ func (or *ObsReport) StartLogsOp(ctx context.Context) context.Context {
 // EndLogsOp completes the export operation that was started with StartLogsOp.
 func (or *ObsReport) EndLogsOp(ctx context.Context, numLogRecords int, err error) {
 	numSent, numFailedToSend := toNumItems(numLogRecords, err)
-	or.recordMetrics(ctx, component.DataTypeLogs, numSent, numFailedToSend)
+	or.recordMetrics(context.WithoutCancel(ctx), component.DataTypeLogs, numSent, numFailedToSend)
 	endSpan(ctx, err, numSent, numFailedToSend, obsmetrics.SentLogRecordsKey, obsmetrics.FailedToSendLogRecordsKey)
 }
 
--- a/exporter/exporterhelper/obsexporter.go
+++ b/exporter/exporterhelper/obsexporter.go
@@ -5,7 +5,6 @@ package exporterhelper // import "go.opentelemetry.io/collector/exporter/exporte
 
 import (
 	"context"
-	"time"
 
 	"go.opencensus.io/stats"
 	"go.opencensus.io/tag"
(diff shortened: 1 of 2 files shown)


exporter/exporterhelper/obsexporter.go
```diff
@@ -154,7 +154,7 @@ func (or *ObsReport) StartTracesOp(ctx context.Context) context.Context {
 // EndTracesOp completes the export operation that was started with StartTracesOp.
 func (or *ObsReport) EndTracesOp(ctx context.Context, numSpans int, err error) {
 	numSent, numFailedToSend := toNumItems(numSpans, err)
-	or.recordMetrics(ctx, component.DataTypeTraces, numSent, numFailedToSend)
+	or.recordMetrics(context.Background(), component.DataTypeTraces, numSent, numFailedToSend)
```
TylerHelmuth, line 157: Is it true that we'd never need the original context inside `recordMetrics`?

codeboten: go 1.21 has a WithoutCancel func I would suggest we move to in the future (whenever 1.22 is out)

TylerHelmuth: If we go forward with this can we add a TODO or an issue to track that?

codeboten: Yeah i was planning on filing an issue to track this change

TylerHelmuth (approved):

exporter/exporterhelper/obsexporter.go
codeboten: added an issue #9049

fatsheep9146:
I also have met this problem when I implemented the grpc instrumentation library, here is the comment

https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4356#discussion_r1360132514

and @pellared from sig go sdk have created an issue talked about this behavior in go sdk sig meeting https://github.com/open-telemetry/opentelemetry-go/pull/4671

The conclusion is go sdk will keep the behavior. And suggest user to use `WithoutCancel` to escape this, and before go 1.20 we could solve like we do in grpc instrumentation library , which basically is a copy of `WithCancel` in go 1.21
https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6009b3188da4f25eb8d2e65ffeefb3bf4b7c3415/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L216-L245

@codeboten

exporter/exporterhelper/obsexporter.go
jmacd: Suggest use of `context.TODO()` here. I understand this solves a short-term problem, but I'd like to ask a bigger question about why we would bypass metric events on canceled contexts. It seems like a recipe for lost time and user confusion, as happened here. If we could convince the Otel-Go SDK to modify this behavior, perhaps restore the `context.TODO()` to the original context.

As you suggested, an alternate SDK with an extension for interpreting context will lose access to baggage, something we could fix in the long run. We have a few internal uses where exporterhelper metrics such as these are being extended from context with header values from the gRPC metadata. (Using [this](https://github.com/lightstep/otel-launcher-go/tree/main/lightstep/sdk/metric#measurementprocessor).)

codeboten:
thanks @fatsheep9146, copied the implementation of withoutCancel in this PR

exporter/exporterhelper/obsexporter.go
codeboten: @jmacd i followed @fatsheep9146's suggestion and implemented withoutCancel temporarily in this code

exporter/exporterhelper/obsexporter.go
```diff
@@ -313,3 +314,34 @@ func (or *ObsReport) recordEnqueueFailureWithOtel(ctx context.Context, dataType
 
 	enqueueFailedMeasure.Add(ctx, failed, metric.WithAttributes(or.otelAttrs...))
 }
+
+func withoutCancel(parent context.Context) context.Context {
```
bogdandrutu, line 318: Duplicate of https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queue_sender.go#L236

Why not using that?

codeboten: good catch... i didn't see it there

pellared:
> @pellared from sig go sdk have created an issue talked about this behavior in go sdk sig meeting https://github.com/open-telemetry/opentelemetry-go/pull/4671

I am thinking about "reviving" the issue. Especially in the context of Logs API+SDK where I do not want the same behavior for Logs SDK. I want to be inline with e.g. https://pkg.go.dev/log/slog#Handler which says:

> Canceling the context should not affect record processing. (Among other things, log messages may be necessary to debug a cancellation-related problem.)

Changing the metrics SDK behavior would be a breaking change, but I think it would help 90% of the use cases and I doubt it will cause any harm to anyone. Even if so I think it may be an acceptable breaking behavioral change.

Please let me know what you think. I think that  OTel GO "consumers" point of view is very important here.

Because of the current behavior we have to deal with issues/PRs like this one: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4634

jmacd:
> I think it would help 90% of the use cases and I doubt it will cause any harm to anyone.

Personally, I can't imagine a user who expects to skip metric events because of context cancellation. Has the group heard any user stories about that? Mainly, I agree with the comment quoted above:

> (Among other things, log messages may be necessary to debug a cancellation-related problem.)

The way I think about it, context cancellation is a mechanism to stop blocking, avoid waiting, or skip useless work. So, I expect a user to inspect the context and avoid instrumented code, not to avoid instrumentation events. It is true that the aggregator could acquire a mutex, and that could briefly block the application. However, so does a call to `ctx.Err()` imply taking a mutex that could briefly block the application.

In my implementation, the context is last used and processed in the synchronous instrument path, before an aggregator sees the event. If the SDK wants to inspect the baggage or other contextual information--to me--it probably wants to be done at a higher level, so before the aggregators, so that context-related behavior is aggregator-independent.

PR #9051, opened 2023-12-07
[chore] pass shutdown context when closing batch senders

atoulme:
**Description:**

Resolves a TODO in the batch processor. Pass the context of the `Shutdown` method to the shard sender methods.


TylerHelmuth (approved):
LGTM

processor/batchprocessor/batch_processor.go
```diff
@@ -171,9 +170,9 @@ func (bp *batchProcessor) Start(context.Context, component.Host) error {
 }
 
 // Shutdown is invoked during service shutdown.
-func (bp *batchProcessor) Shutdown(context.Context) error {
+func (bp *batchProcessor) Shutdown(ctx context.Context) error {
+	bp.shutdownCtx = ctx
```
dmitryax, line 174: This context passed to Shutdown isn't going to be always cancelled. We shouldn't use it

Also, you are replacing the `bp.shutdownCtx` with another instance. So the old context passed to the downstream consumer is unchanged.

processor/batchprocessor/batch_processor.go
```diff
@@ -195,6 +196,8 @@ func (b *shard) start() {
 		DONE:
 			for {
 				select {
+				case <-b.processor.shutdownCtx.Done():
```
bogdandrutu, line 199: Is this thread-safe?

atoulme: Yes, I believe it is. The code only enters this loop after the `shutdownCtx` field is set, after `shutdownC` is closed.

bogdandrutu: That is not true, you read the `shutdownCtx` variable which is written unprotected. Do you guarantee that shard.start(..) is called only after shutdown somewhere else?

atoulme: ok, so I changed things around again, with channels to propagate shutdown context signaling it's done.

atoulme:
This PR is ready for another look - please let me know. I ran both new tests a million times to make sure neither are flaky.

PR #9053, opened 2023-12-07
Fix bug where MutatesData would not correctly propogate through connectors

djaglowski:
This fixes two closely related problems. 
1. While fanoutconsumers do not themselves mutate data, they should expose whether or not they are handing data off to consumers which may do so. Otherwise, the service cannot correctly determine how to fan out after a receiver. e.g. a receiver shared between two pipelines, one of which contains an exporter or connector which mutates data.
2. Connectors can themselves mutate data but we were not taking this into account when building the graph.

### DIFF
--- a/internal/fanoutconsumer/metrics.go
+++ b/internal/fanoutconsumer/metrics.go
@@ -42,7 +42,7 @@ type metricsConsumer struct {
 }
 
 func (msc *metricsConsumer) Capabilities() consumer.Capabilities {
-	return consumer.Capabilities{MutatesData: false}
+	return consumer.Capabilities{MutatesData: len(msc.mutable) > 0}
 }
 
 // ConsumeMetrics exports the pmetric.Metrics to all consumers wrapped by the current one.
--- a/service/internal/graph/graph.go
+++ b/service/internal/graph/graph.go
@@ -288,7 +288,10 @@ func (g *Graph) buildComponents(ctx context.Context, set Settings) error {
 		case *connectorNode:
 			err = n.buildComponent(ctx, telemetrySettings, set.BuildInfo, set.ConnectorBuilder, g.nextConsumers(n.ID()))
 		case *capabilitiesNode:
-			capability := consumer.Capabilities{MutatesData: false}
+			capability := consumer.Capabilities{
+				// The fanOutNode represents the aggregate capabilities of the exporters in the pipeline.
+				MutatesData: g.pipelines[n.pipelineID].fanOutNode.getConsumer().Capabilities().MutatesData,
+			}
 			for _, proc := range g.pipelines[n.pipelineID].processors {
 				capability.MutatesData = capability.MutatesData || proc.getConsumer().Capabilities().MutatesData
 			}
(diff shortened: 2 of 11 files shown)


service/internal/graph/graph.go
```diff
@@ -288,7 +288,9 @@ func (g *Graph) buildComponents(ctx context.Context, set Settings) error {
 		case *connectorNode:
 			err = n.buildComponent(ctx, telemetrySettings, set.BuildInfo, set.ConnectorBuilder, g.nextConsumers(n.ID()))
 		case *capabilitiesNode:
-			capability := consumer.Capabilities{MutatesData: false}
+			capability := consumer.Capabilities{
+				MutatesData: g.pipelines[n.pipelineID].fanOutNode.getConsumer().Capabilities().MutatesData,
```
bogdandrutu, line 292: Is the exporter fanout? If that is the case please comment in the code or rename the variable to clarify it.

service/internal/graph/nodes.go
```diff
@@ -253,6 +253,10 @@ func (n *connectorNode) buildComponent(
 				return err
 			}
 			n.Component = conn
+
+			// The connector may mutate data itself
```
bogdandrutu, line 257: Very interesting that you do this only for traces (same for the capability consumer initialization). Is this not a bug? Can we also start extracting this in a separate func? (hard to follow)

djaglowski: I'll more comments to clarify this, but note the comment above the next line:
```
// When connecting pipelines of the same data type, the connector must
// inherit the capabilities of pipelines in which it is acting as a receiver.
n.baseConsumer = capabilityconsumer.NewTraces(conn, capability)
```

What's going on here is that when we connect pipelines of the same type, the original data may be mutated later on. However, when we are _not_ connecting pipelines of the same type, the connector does not pass the original data along, so there is no way the data is mutated downstream.

So in the different-type cases, we already correctly represent capabilities by setting `n.baseConsumer` as the connector.

But in the same-type cases we are building a new `capabilityconsumer`, but were forgetting to include the capability of the connector itself.

service/internal/graph/graph.go
djaglowski: The fanOutNode always comes after processors and before exporters, so should represent the aggregate capabilities of the exporter in the pipeline. I'll add a comment.

djaglowski: I've added a comment.

service/internal/graph/nodes.go
djaglowski: I moved this line and combined the comment with the one below, since they are so closely related.

internal/fanoutconsumer/metrics.go
```diff
@@ -42,7 +42,7 @@ type metricsConsumer struct {
 }
 
 func (msc *metricsConsumer) Capabilities() consumer.Capabilities {
-	return consumer.Capabilities{MutatesData: false}
+	return consumer.Capabilities{MutatesData: len(msc.mutable) > 0}
```
dmitryax, line 45: If there is a readonly receiver in the fan-out, it's not mutating. The mutable data will be copied
```suggestion
	return consumer.Capabilities{MutatesData: len(msc.mutable) > 0 && len(msc. readonly) == 0}
```

bogdandrutu (approved):

Issue #7388, opened 2023-03-15
[exporterhelper] boundedMemoryQueue may send on closed channel


atingchen:
**Describe the bug**
When the gorutine prepares to submit new item to the queue, `q.stopped.Load()`  returns true. The gorutine has not finished sending items to the channel.
```go
func (q *boundedMemoryQueue) Produce(item Request) bool {
	if q.stopped.Load() {
		return false
	}

	if q.size.Load() >= q.capacity {
		return false
	}

	q.size.Add(1)
	select {
	case q.items <- item:
		return true
	default:
		// should not happen, as overflows should have been captured earlier
		q.size.Add(^uint32(0))
		return false
	}
}
```
At the same time, service shutdown and  invokes  `stop()` to close the channel. The gorutine may send item to the closed channel.
```GO 
func (q *boundedMemoryQueue) Stop() {
	q.stopped.Store(true) // disable producer
	close(q.items)
	q.stopWG.Wait()
}
```
**What did you expect to see?**
thread-safe

**What did you see instead?**

**What version did you use?**
v0.73.0

yutingcaicyt:
The collector will stop in topological order so that upstream components are stopped before downstream components.  This ensures that each component has a chance to drain to its consumer before the consumer is stopped. Therefore, under normal circumstances, before the exporter is closed, the receiver and processor have guaranteed to complete the data transmission and close. When the exporter is closing, no data will enter the exporter.

atingchen:
> The collector will stop in topological order so that upstream components are stopped before downstream components. This ensures that each component has a chance to drain to its consumer before the consumer is stopped. Therefore, under normal circumstances, before the exporter is closed, the receiver and processor have guaranteed to complete the data transmission and close. When the exporter is closing, no data will enter the exporter.

Thanks for your reply @yutingcaicyt. `boundedMemoryQueue` will be used as a retry queue. When the upstream component is closed, data may still re-enter the queue due to the retry strategy.

When `queuedRetrySender` calls `shutdown`, it will first close `retryStopCh`, and then close the retry queue.
`OnTemporaryFailure` is called between these two steps, which has a small chance of putting the data back into the retry queue.

```go
		select {
		case <-req.Context().Done():
			return fmt.Errorf("Request is cancelled or timed out %w", err)
		case <-rs.stopCh:
			return rs.onTemporaryFailure(rs.logger, req, fmt.Errorf("interrupted due to shutdown %w", err))
[...]

PR #9054, opened 2023-12-08
[chore] [exporterhelper] Do not requeue on shutdown

dmitryax:
This change unblocks adding the `enqueue_on_failure` option to the queue sender by removing the requeue behavior on the shutdown. If we don't remove requeue on shutdown, it's possible to run into a situation described in https://github.com/open-telemetry/opentelemetry-collector/issues/7388. After the recent refactoring, the chance of running into it is pretty small, but it's still possible.

The only reason to requeue on shutdown is to make sure there is no data loss with the persistent queue enabled. The persistent queue captures all the inflight requests in the persistent storage anyway, so there is no reason to requeue an inflight request. The only downside is it potentially can cause sending duplicate data on the collector restart in case of a partially failed request during shutdown.

Another option would be to rework the memory queue to never close the channel but still ensure draining.

### DIFF
--- a/exporter/exporterhelper/internal/persistent_queue.go
+++ b/exporter/exporterhelper/internal/persistent_queue.go
@@ -141,7 +141,7 @@ func (pq *persistentQueue[T]) initPersistentContiguousStorage(ctx context.Contex
 // Consume applies the provided function on the head of queue.
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped.
-func (pq *persistentQueue[T]) Consume(consumeFunc func(context.Context, T)) bool {
+func (pq *persistentQueue[T]) Consume(consumeFunc func(context.Context, T) bool) bool {
 	var (
 		req                  T
 		onProcessingFinished func()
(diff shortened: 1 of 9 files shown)


exporter/exporterhelper/internal/queue.go
```diff
@@ -30,7 +30,8 @@ type Queue[T any] interface {
 	// Consume applies the provided function on the head of queue.
 	// The call blocks until there is an item available or the queue is stopped.
 	// The function returns true when an item is consumed or false if the queue is stopped.
-	Consume(func(ctx context.Context, item T)) bool
+	// The provided callback function returns true if the item was consumed or false if the consumer is stopped.
+	Consume(func(ctx context.Context, item T) bool) bool
```
bogdandrutu, line 34: Please return `error` for signaling success or failure.

exporter/exporterhelper/internal/persistent_queue.go
```diff
@@ -141,7 +141,7 @@ func (pq *persistentQueue[T]) initPersistentContiguousStorage(ctx context.Contex
 // Consume applies the provided function on the head of queue.
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped.
-func (pq *persistentQueue[T]) Consume(consumeFunc func(context.Context, T)) bool {
+func (pq *persistentQueue[T]) Consume(consumeFunc func(context.Context, T) bool) bool {
```
bogdandrutu, line 144: Per the same logic that we never call into this after stop you should remove stopChan from here as well.

bogdandrutu: May not be possible this...

bogdandrutu (approved):

PR #9057, opened 2023-12-08
[chore] add semconvkit to reduce toil

codeboten:
This code, copied from the opentelemetry-go repo, allows to generate schema.go automatically as well, reducing effort needed to generate semantic convention packages in the future

### DIFF
--- a/internal/tools/tools.go
+++ b/internal/tools/tools.go
@@ -24,6 +24,7 @@ import (
 	_ "go.opentelemetry.io/build-tools/crosslink"
 	_ "go.opentelemetry.io/build-tools/multimod"
 	_ "go.opentelemetry.io/build-tools/semconvgen"
+	_ "go.opentelemetry.io/collector/internal/tools/semconvkit"
 	_ "golang.org/x/exp/cmd/apidiff"
 	_ "golang.org/x/tools/cmd/goimports"
 	_ "golang.org/x/vuln/cmd/govulncheck"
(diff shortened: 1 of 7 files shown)


internal/tools/tools.go
```diff
@@ -27,4 +27,6 @@ import (
 	_ "golang.org/x/exp/cmd/apidiff"
 	_ "golang.org/x/tools/cmd/goimports"
 	_ "golang.org/x/vuln/cmd/govulncheck"
+
+	_ "go.opentelemetry.io/collector/internal/tools/semconvkit"
```
TylerHelmuth, line 31: Should this live in go-build-tools?

codeboten: Maybe, but I didn't want to block on this. I'm happy to open a follow up issue

TylerHelmuth: Ya followup thing

dmitryax (approved):

semconv/README.md
```diff
@@ -16,8 +16,3 @@ Generating semantic convention constants from specification version v1.22.0 at /
 .tools/semconvgen -o semconv/v1.22.0 -t semconv/template.j2 -s v1.22.0 -i /tmp/semantic-conventions/model/. --only=event -p conventionType=event -f generated_event.go
 .tools/semconvgen -o semconv/v1.22.0 -t semconv/template.j2 -s v1.22.0 -i /tmp/semantic-conventions/model/. --only=span -p conventionType=trace -f generated_trace.go
 ```
-
-When generating the constants for a new version ot the specification it is important to note that only
-`generated_trace.go` and `generated_resource.go` are generated automatically. The `schema.go` and `nonstandard.go`
```
dmitryax, line 21: Looks like we lost `nonstandard.go` at some point, not sure it we need it though

codeboten: I didn't see it in the latest version of the semconv package, so i left it out. If need be we can always add another template 🤷🏻

PR #9062, opened 2023-12-11
Fanout consumer does not need to mutate in some cases

djaglowski:
Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/9053.

@dmitryax pointed out [here](https://github.com/open-telemetry/opentelemetry-collector/pull/9053#discussion_r1420871665) that the fanout consumer will pass original data to a non-mutating consumer if any is available. This PR incorporates that point and updates test expectations accordingly.

### DIFF
--- a/internal/fanoutconsumer/logs.go
+++ b/internal/fanoutconsumer/logs.go
@@ -44,7 +44,8 @@ type logsConsumer struct {
 }
 
 func (lsc *logsConsumer) Capabilities() consumer.Capabilities {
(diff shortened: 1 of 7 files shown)


service/internal/graph/__debug_bin1931707248
TylerHelmuth, line 1: Looks like something got commit by mistake

djaglowski: Thanks, removed it.

TylerHelmuth (approved):

PR #9065, opened 2023-12-11
record a metric with open census to ensure backwards compat

codeboten:
The healthcheck extension currently relies on an OpenCensus view to determine the health of the collector. This behaviour is not great as it only looks at span exports, but it exists today. To avoid breaking it, I suggest this PR that adds the same OC view even on the otel path, and records the same metric.

Note to reviewers: This is only a hack that's needed until the new healthcheck extension that uses component status is ready to go. cc @mwear

### DIFF
--- a/.chloggen/codeboten_use-otel-with-healthcheck-view.yaml
+++ b/.chloggen/codeboten_use-otel-with-healthcheck-view.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: telemetry
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add OpenCensus view to useOtelForInternalMetrics path
+
+# One or more tracking issues or pull requests related to the change
+issues: [9065]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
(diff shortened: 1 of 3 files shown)


mwear (approved):

TylerHelmuth:
Oh I see now that we already have a way forward for the healthcheckextension, and this is a hack until then

TylerHelmuth:
@mwear @codeboten you got a link to the healthcheckextension issue that, when closed, means we can remove this hack?  Can we also open an issue to track the removal of this hack?

codeboten:
@TylerHelmuth This issue captures most of the problems w/ the healthcheck extension with this issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26661 capturing the refactoring of it. I can create an issue to track removing this hack once https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26661 is done

codeboten:
@TylerHelmuth added https://github.com/open-telemetry/opentelemetry-collector/issues/9069

TylerHelmuth (approved):

codeboten:
Closing this PR. After further investigation, the behaviour in the current healthcheck extension doesn't work as expected. It registers itself as an OpenCensus exporter to be called when census wants to emit a metric. For the `exporter/send_failed_requests`, this export is triggered every 10s, rather than every time a failure is recorded.

In addition, the extension does not read the value of the metric, which means the threshold it calculates is solely based on the number of times the view is exported, which has no correlation to failures to export.

Since the view is configured only to record measurements for span exports, the healthcheck extension only ever triggered for a single signal, which doesn't make it overly useful.

After discussing this with other collector approvers and maintainers, we've decided to flag the feature as not working as expected in the healthcheck extension. Since this extension is being revamped to use component status, it isn't worth fixing in the short term. I'll open a PR to update the README to warn users of this issue, this is only to prevent a breaking change multiple times in short successions for users. cc @TylerHelmuth @mx-psi

Issue #4331, opened 2021-11-01
Add timeout to confignet


BinaryFissionGames:
**Is your feature request related to a problem? Please describe.**
When dialing, there should be a configurable timeout in these configs. Without these, no timeout is used and depending on the protocol and OS, may take a large amount of time to timeout or not timeout at all.

**Describe the solution you'd like**
Add a `timeout` parameter and use it with net.DialTimeout, allowing a configurable timeout when dialing.

TylerHelmuth:
Please assign this to me

PR #9066, opened 2023-12-11
[confignet] Add DialerTimeout config option

TylerHelmuth:
**Description:** 
Adds a new `dialer_timeout` config option to `confignet` to allow configuring the timeout of the connection.

**Link to tracking Issue:** 
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/4331

**Testing:** 
Added new unit tests

**Documentation:** 
Updated readme

### DIFF
--- a/config/confignet/confignet.go
+++ b/config/confignet/confignet.go
@@ -8,6 +8,13 @@ import (
 	"time"
 )
 
+// Dialer contains options for connecting to an address.
+type Dialer struct {
+	// Timeout is the maximum amount of time a dial will wait for
+	// a connect to complete. The default is no timeout.
+	Timeout time.Duration `mapstructure:"timeout"`
+}
+
 // NetAddr represents a network endpoint address.
 type NetAddr struct {
 	// Endpoint configures the address for this network connection.
@@ -21,14 +28,13 @@ type NetAddr struct {
 	// "udp6" (IPv6-only), "ip", "ip4" (IPv4-only), "ip6" (IPv6-only), "unix", "unixgram" and "unixpacket".
 	Transport string `mapstructure:"transport"`
 
-	// DialerTimeout is the maximum amount of time a dial will wait for
-	// a connect to complete. The default is no timeout.
-	DialerTimeout time.Duration `mapstructure:"dialer_timeout"`
+	// Dialer contains options for connecting to an address.
+	Dialer Dialer `mapstructure:"dialer"`
 }
 
 // Dial equivalent with net.Dial for this address.
 func (na *NetAddr) Dial() (net.Conn, error) {
-	return net.DialTimeout(na.Transport, na.Endpoint, na.DialerTimeout)
+	return net.DialTimeout(na.Transport, na.Endpoint, na.Dialer.Timeout)
 }
 
 // Listen equivalent with net.Listen for this address.
@@ -45,14 +51,13 @@ type TCPAddr struct {
 	// "[fe80::1%zone]:80". The zone specifies the scope of the literal IPv6 address as defined in RFC 4007.
 	Endpoint string `mapstructure:"endpoint"`
 
-	// DialerTimeout is the maximum amount of time a dial will wait for
-	// a connect to complete. The default is no timeout.
-	DialerTimeout time.Duration `mapstructure:"dialer_timeout"`
+	// Dialer contains options for connecting to an address.
+	Dialer Dialer `mapstructure:"dialer"`
 }
 
 // Dial equivalent with net.Dial for this address.
 func (na *TCPAddr) Dial() (net.Conn, error) {
-	return net.DialTimeout("tcp", na.Endpoint, na.DialerTimeout)
+	return net.DialTimeout("tcp", na.Endpoint, na.Dialer.Timeout)
 }
 
 // Listen equivalent with net.Listen for this address.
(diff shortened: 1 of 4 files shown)


codeboten (commented):
Thanks for submitting this, I was looking at this Friday, and I'm curious what your thoughts are around whether or not this feature will allow the removal of dialer timeouts configuration in other components.

please take a look at the failing tests

TylerHelmuth:
> I'm curious what your thoughts are around whether or not this feature will allow the removal of dialer timeouts configuration in other components.

If a component already allowed configuring a `dialer_timeout` then they must already be doing their own `net.DialTimeout`, circumventing `TCPAddr.Dial`/`NetAddr.Dial`, so they can take advantage of their timeout value. In these cases I suspect the `TCPAddr` and `NetAddr` are only being used for their config and not their methods.

The [carbonexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/carbonexporter/config.go) is an example of a Contrib component that does [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/95cb7336823185453e7a6538dc38988ee6596042/exporter/carbonexporter/exporter.go#L148).

It looks like some signalfx exporter also does this: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/95cb7336823185453e7a6538dc38988ee6596042/exporter/signalfxexporter/internal/dimensions/dimclient.go#L92-L96. But in this case they also take advantage of additional `net.Dialer` settings that `confignet` does not support.

This kinda gets to a question about how much of a wrapper `confignet` needs to be.

config/confignet/confignet.go
```diff
@@ -39,11 +44,15 @@ type TCPAddr struct {
 	// If the host is a literal IPv6 address it must be enclosed in square brackets, as in "[2001:db8::1]:80" or
 	// "[fe80::1%zone]:80". The zone specifies the scope of the literal IPv6 address as defined in RFC 4007.
 	Endpoint string `mapstructure:"endpoint"`
+
+	// DialerTimeout is the maximum amount of time a dial will wait for
+	// a connect to complete. The default is no timeout.
+	DialerTimeout time.Duration `mapstructure:"dialer_timeout"`
 }
```
bogdandrutu, line 51: Should we consider having "Dialer" as a separate struct with only Timeout for the moment in case we add more options there?

So essentially is a matter of writing:
```
dialer:
  timeout: 5s
  foo: ...
```
vs
```
dialer_timeout: 5s
dialer_foo: ...
```

config/confignet/confignet.go
```diff
@@ -39,11 +44,15 @@ type TCPAddr struct {
...
 
 // Dial equivalent with net.Dial for this address.
 func (na *TCPAddr) Dial() (net.Conn, error) {
```
bogdandrutu, line 54: If we are here, should we also change this to accept context?

config/confignet/confignet.go
TylerHelmuth: I like this idea, especially if we want to keep our options open for settings like `KeepAlive` or `Dealine`, etc.

config/confignet/confignet.go
TylerHelmuth: It looks like the `net` authors let `net.Dial` and `net.DialTimeout` use `context.Background()` and expect users to use `net.DialContext()`.  Instead of our wrapper merging the concepts, would it be better to expose a `TCPAddr.DialContext()` that calls `net.DialContext`? That would let us avoid a breaking change.

bogdandrutu: > It looks like the net authors let net.Dial and net.DialTimeout use context.Background() and expect users to use net.DialContext()

They are legacy, because net was added before context. I think we should have a rule of thumb that all our public APIs should accept Context. This is not yet 1.0 so we can fix that here.

TylerHelmuth: @bogdandrutu do you want me to make that change in this PR or in a separate PR?

bogdandrutu: Separate PR please.

config/confignet/confignet.go
TylerHelmuth: @bogdandrutu updated

config/confignet/confignet.go
TylerHelmuth: Created https://github.com/open-telemetry/opentelemetry-collector/issues/9105 to track this topic

config/confignet/confignet.go
TylerHelmuth: Instead of making our own struct should we embedded `net.Dialer`?

TylerHelmuth: NVM, we need mapstructure

atoulme (approved):

bogdandrutu (approved):

config/confignet/confignet.go
```diff
@@ -5,8 +5,16 @@ package confignet // import "go.opentelemetry.io/collector/config/confignet"
 
 import (
 	"net"
+	"time"
 )
 
+// Dialer contains options for connecting to an address.
+type Dialer struct {
```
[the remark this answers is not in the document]
TylerHelmuth: Renamed

bogdandrutu (approved):

PR #9089, opened 2023-12-12
Add RetrySettings validation function

bogdandrutu:


exporter/exporterhelper/retry_sender.go
```diff
@@ -39,6 +39,25 @@ type RetrySettings struct {
 	MaxElapsedTime time.Duration `mapstructure:"max_elapsed_time"`
 }
 
+func (cfg *RetrySettings) Validate() error {
```
TylerHelmuth, line 42: Can we add some tests in `retry_sender_test.go` for these different scenarios?

bogdandrutu: Done.

TylerHelmuth (approved):

exporter/exporterhelper/retry_sender.go
```diff
@@ -39,6 +39,28 @@ type RetrySettings struct {
...
+	if !cfg.Enabled {
+		return nil
+	}
+	if cfg.InitialInterval < 0 {
+		return errors.New("initial interval must be non-negative")
```
djaglowski, line 47: I'm not sure how consistently we follow this or not, but typically I advise that we use the exact keys which the user would need to work with. If this goes against established patterns, feel free to ignore. Otherwise, this could be applied to all error messages below as well.
```suggestion
		return errors.New("'initial_interval' must be non-negative")
```

djaglowski (approved):

dmitryax (approved):

Issue #8382, opened 2023-09-11
reenqueue may not be needed


yutingcaicyt:
**Is your feature request related to a problem? Please describe.**
When I used the persistent queue, I found that the strategy dealing with requests that retry failed is different. The persistent queue enqueues these requests but the memory queue does not. In fact, users can configure the retry time themselves, if timeout, the request can be discarded intuitively. If reenqueue, the time a request tries to send is hard to know. So why not keep it simple and Let the user decide how long a request can be retried at most?

**Describe the solution you'd like**
We can remove the reenqueue feature to make it simple and intuitive so users can decide the retry time with the configuration.

dmitryax:
I very much agree with the suggestion. I don't know why it was introduced in the first place. I think the only scenario when we need to re-enqueue is during shutdown.

@swiatekm-sumo @bogdandrutu WDYT?

swiatekm-sumo:
Requeueing as implemented is how a lot of users intuitively expect the queue to work, in my experience. That is, they expect no data to be dropped if the queue still has room. And this behaviour is difficult to reproduce by adjusting retry settings.

I'd prefer to add it as an option to the in-memory queue, with a default that is consistent between the two queue types.

yutingcaicyt:
So how should we deal with it better?

swiatekm-sumo:
Like I said, add the configuration option so both queues behave the same way. Then you can decide which behaviour you prefer.

yutingcaicyt:
With the reenqueue option, users have two ways to control the retry of the data. Is there an alternative relationship between these two ways? If the user does not want to discard data, can it be enough to increase the retry time?

dmitryax:
I'm trying to introduce this config option in the scope of https://github.com/open-telemetry/opentelemetry-collector/issues/8987 and running into a couple of issues:
1. The requeuing doesn't work well with the memory queue on shutdown. The memory queue channel is closed on shutdown to drain the queue. So if we introduce this option, there is a chance of running into https://github.com/open-telemetry/opentelemetry-collector/issues/7388. I tried to avoid requeuing on shutdown but doesn't seem to be a bulletproof solution https://github.com/open-telemetry/opentelemetry-collector/pull/9054. So making sure we are not requeuing into a closed memory channel introduces unnecessary complexity.
2. As pointed out in https://github.com/open-telemetry/opentelemetry-collector/pull/8993#issuecomment-1828375507, introducing another "retry" option might be more confusing to the users.

[...]

PR #9090, opened 2023-12-12
[exporterhelper] Do not re-enqueue failed requests

dmitryax:
The current re-enqueuing behavior is not obvious and cannot be configured. It takes place only for persistent queue and only if `retry_on_failure::enabled=true` even if `retry_on_failure` is a setting for a different backoff retry strategy. 

This change removes the re-enqueuing behavior in favor of the `retry_on_failure` option. Consider increasing `retry_on_failure::max_elapsed_time` to reduce chances of data loss.

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/8382

### DIFF
--- a/exporter/exporterhelper/internal/bounded_memory_queue.go
+++ b/exporter/exporterhelper/internal/bounded_memory_queue.go
@@ -40,12 +40,13 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
 // Consume applies the provided function on the head of queue.
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
-func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) bool) bool {
+func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
 	item, ok := <-q.items
 	if !ok {
 		return false
 	}
-	consumeFunc(item.ctx, item.req)
+	// the memory queue doesn't handle consume errors
+	_ = consumeFunc(item.ctx, item.req)
 	return true
 }
 
--- a/exporter/exporterhelper/queue_sender.go
+++ b/exporter/exporterhelper/queue_sender.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync/atomic"
 	"time"
 
 	"go.opencensus.io/metric/metricdata"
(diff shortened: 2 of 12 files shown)


dmitryax:
@swiatekm-sumo, PTAL

exporter/exporterhelper/queue_sender.go
```diff
@@ -105,50 +102,22 @@ func newQueueSender(config QueueSettings, set exporter.CreateSettings, signal co
...
 			zap.Int("dropped_items", req.ItemsCount()),
 		)
-		return true
-	}
-
-	if qs.queue.Offer(ctx, extractPartialRequest(req, err)) == nil {
-		qs.logger.Error(
-			"Exporting failed. Putting back to the end of the queue.",
-			zap.Error(err),
-		)
```
swiatekm-sumo, line 143: This is actually a change in behaviour, right? Before, with the persistent queue, we'd requeue partial failures, whereas now we drop them. And it doesn't look like old behaviour can't be reproduced by messing with retry sender settings.

I think this is fine, but I'd like to have clarity on what the actual effect of this change is, so we can call it out in the changelog.

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -40,12 +40,13 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
...
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
-func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) bool) bool {
+func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
 	item, ok := <-q.items
 	if !ok {
 		return false
 	}
-	consumeFunc(item.ctx, item.req)
+	// the memory queue doesn't handle consume errors
+	_ = consumeFunc(item.ctx, item.req)
```
swiatekm-sumo, line 49: I know this isn't a change, but having to ignore the error here feels like our logging architecture is messy. But I suppose the new exporter helper will address this.

dmitryax: It's already logged in the retry sender. We pass the error here in case queue implementations want to do some action on it like acknowledgement

exporter/exporterhelper/queue_sender.go
dmitryax: It's called out in the changelog https://github.com/open-telemetry/opentelemetry-collector/pull/9090/files#diff-318157536dd6308f2a06e09664dc9bf14f8646cf979de227ec0a7c02317e16b9R19-R22. Let me know if it's not enough

swiatekm-sumo (approved):

bogdandrutu (approved):

PR #9095, opened 2023-12-13
Remove circular dependency between default otel and connector

bogdandrutu:


connector/metrics_router.go
```diff
@@ -0,0 +1,70 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package connector // import "go.opentelemetry.io/collector/connector"
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/internal/fanoutconsumer"
```
dmitryax, line 9: the linter should be failing here

connector/logs_router_test.go
```diff
@@ -0,0 +1,156 @@
...
+				assert.Len(t, logs, expected[id])
+				for n := 0; n < len(logs); n++ {
+					assert.EqualValues(t, ld, logs[n])
+				}
+			}
+		}
+	}
+}
+
+func TestLogsGetConsumers(t *testing.T) {
```
dmitryax, line 107: nit: the test names are inconsistent between between data types. Should be `Test[Logs|Metrics|Traces]RouterConsumer`?

dmitryax (approved):

connector/metrics_router.go
bogdandrutu: Done.

connector/logs_router_test.go
bogdandrutu: Done.

dmitryax (approved):

djaglowski (commented):
I'm working on something closely related and would like some time to reconcile possible changes with this PR. I may have some suggestions based on this perspective. I will follow up here today or tomorrow.

djaglowski (approved):
This looks good to me. The changes I am looking at elsewhere would fit into this just as well either way.

codeboten (commented):
I believe this change should be done by first deprecating the original interfaces no? https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes

djaglowski:
Hmm, the intention isn't to remove the interfaces, but changing them does break them. Do we deprecate, then remove the deprecation while making the change?

codeboten:
> Hmm, the intention isn't to remove the interfaces, but changing them does break them. Do we deprecate, then remove the deprecation while making the change?

I believe this was done in the past by providing a temporary interface name, that will then be renamed again

bogdandrutu:
@codeboten I think this has a very limited use, and the benefit of doing the deprecation is limited. I would actually ask to go with an exception if that is the case.

PR #9096, opened 2023-12-13
[chore] remove unused default.otel.yaml file

bogdandrutu:


dmitryax (approved):
I usually use it for test runs :) but probably no need to keep it

bogdandrutu:
It is same as core.yaml, you can use that witch is also used in tests or rename it to default

Issue #9099, opened 2023-12-13
Leaking goroutine in memorylimiterprocessor


jskiba:
**Describe the bug**
Memorylimiter processor never leaves that loop https://github.com/open-telemetry/opentelemetry-collector/blob/1d0d1ef21baebfcc9ed1d67f0237437ceb9a0049/processor/memorylimiterprocessor/memorylimiter.go#L230C1-L230C1. In result this goroutine or many of them leak every time after shutdown procedure.

**Steps to reproduce**
https://github.com/jskiba/opentelemetry-collector/commit/2f8ee15b03a84739acbdd62bc87c4691b703800b
The only leaking goroutine is the one from this bug https://github.com/open-telemetry/opentelemetry-collector/issues/8945.

**What did you expect to see?**
No leaking goroutine and correct exit from for loop.

**What did you see instead?**
Goroutine https://github.com/open-telemetry/opentelemetry-collector/blob/1d0d1ef21baebfcc9ed1d67f0237437ceb9a0049/processor/memorylimiterprocessor/memorylimiter.go#L229 never exits. When a ticker is stopped it does not return information that it was closed, different mechanism should be used.

**What version did you use?**
[1d0d1ef](https://github.com/open-telemetry/opentelemetry-collector/commit/1d0d1ef21baebfcc9ed1d67f0237437ceb9a0049)

**What config did you use?**
not applicable

**Environment**
OS: Debian 12.4 bookworm
Compiler(if manually compiled): go 1.21.5"

**Additional context**
---

dmitryax:
@jskiba thanks for reporting. Do you have a chance to help with a PR?

jskiba:
@dmitryax it's already approved https://github.com/open-telemetry/opentelemetry-collector/pull/9100

PR #9100, opened 2023-12-13
Fix leaking goroutine in memorylimiter

jskiba:
**Description:**
Fixing a bug - described here https://github.com/open-telemetry/opentelemetry-collector/issues/9099

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector/issues/9099

**Testing:**
No tests added.
Unit tests were run

**Documentation:**
None


processor/memorylimiterprocessor/memorylimiter.go
```diff
@@ -141,6 +145,10 @@ func (ml *memoryLimiter) shutdown(context.Context) error {
 		return errShutdownNotStarted
 	} else if ml.refCounter == 1 {
 		ml.ticker.Stop()
+		ml.closeOnce.Do(func() {
+			close(ml.closed)
+		})
```
bogdandrutu, line 150: Don't need this, we guarantee to close only once.

bogdandrutu (approved):

processor/memorylimiterprocessor/memorylimiter.go
jskiba: without this https://github.com/open-telemetry/opentelemetry-collector/blob/1d0d1ef21baebfcc9ed1d67f0237437ceb9a0049/processor/memorylimiterprocessor/factory_test.go#L28 that test fails

jskiba:
@bogdandrutu I forgot about the changelog

processor/memorylimiterprocessor/memorylimiter.go
bogdandrutu: I personally think that this test doesn't make sense. I would like @open-telemetry/collector-approvers to say their opinion.

dmitryax: I think we can remove the second `Shutdown` call from the test. We don't have guarantees that it's allowed.

dmitryax: Oh, it's Start->Shutdown->Start-> causing the failure... This change breaks that workflow even without `ml.closeOnce`. I don't think we have all other components supporting this support.

jskiba: so maybe let's move `closed` and `closeOnce` initialization to `Start` method, the behavior will stay the same

TylerHelmuth: > Start->Shutdown->Start

@dmitryax is calling a processor's `Start` after calling `Shutdown` an unsupported action? Would OPAMP ever need to restart a shutdown component?

If that pattern is not allowed I agree that the test is invalid and we can treat the code in `shutdown` as only ever being invoked once.

processor/memorylimiterprocessor/memorylimiter.go
```diff
@@ -226,8 +234,16 @@ func (ml *memoryLimiter) startMonitoring() {
...
 	if ml.refCounter == 1 {
+		ml.waitGroup.Add(1)
 		go func() {
-			for range ml.ticker.C {
+			defer ml.waitGroup.Done()
+
+			for {
+				select {
+				case <-ml.ticker.C:
+				case <-ml.closed:
```
TylerHelmuth, line 244: Is it possible to get into some weird case (since the ticker is closed before `ml.closed`) where we get an empty value from `m.ticker.C` and run `ml.checkMemLimits()` when we're trying to shutdown? Should we do an `ok` check?

processor/memorylimiterprocessor/memorylimiter.go
dmitryax: > so maybe let's move closed and closeOnce initialization to Start method, the behavior will stay the same

We can move `closed` channel creation to Start. `closeOnce` will not be needed not needed.

> @dmitryax is calling a processor's Start after calling Shutdown an unsupported action? Would OPAMP ever need to restart a shutdown component?

I'm not sure

> If that pattern is not allowed I agree that the test is invalid and we can treat the code in shutdown as only ever being invoked once.

I'm not saying it's not allowed. We just don't explicitly support it. Likely some components will be failing.

Probably better not not break this pattern here with this change and just move `closed` channel creation to Start.

processor/memorylimiterprocessor/memorylimiter.go
dmitryax: > since the ticker is closed before ml.closed

It's not closed. Stopping the ticker doesn't close its channel

yurishkuro:
Has anyone considered enforcing https://github.com/uber-go/goleak in all tests?

processor/memorylimiterprocessor/memorylimiter.go
jskiba: ok, I moved `closed` channel to `Start` and removed `closeOnce`

jskiba: I moved it to `startMonitoring`, so the `closed` channel is created when the goroutine is created. When there are metrics, logs and traces in the pipeline it would be called 3 times and `closed` would be created 3 times too resulting in data race

dmitryax: > When there are metrics, logs and traces in the pipeline it would be called 3 times and closed would be created 3 times too resulting in data race

Why? They are synchronized via `refCounterLock`

dmitryax (approved):

dmitryax:
> Has anyone considered enforcing https://github.com/uber-go/goleak in all tests?

It would be nice to have. I think I saw some leaking goroutines with the open census instrumentation. So we probably need to wait until we migrate from it. I'll create an issue anyway

Issue #8962, opened 2023-11-20
recording metrics with OTel SDK doesn't produce metrics


codeboten:
**Describe the bug**
Using the OpenTelemetry SDK (via the telemetry settings' MeterProvider) doesn't generate any metrics if `useOtelForInternalMetrics` isn't enabled.

**Steps to reproduce**
Configured a local collector to use the routingprocessor (which is instrumented using otel) and started the collector w/o enabling `useOtelForInternalMetrics`

**What did you expect to see?**
Metrics generated

**What did you see instead?**
No metrics generated

**What version did you use?**
v0.89.0

codeboten:
This will be marked as fixed once useOtelForInternalMetrics is updated to stable

PR #9029, opened 2023-12-01
[service] add resource attributes as labels to otel metrics

codeboten:
This ensures backwards compatibility with OpenCensus metrics.

### Collector metrics with OpenCensus:

```
# HELP otelcol_exporter_queue_capacity Fixed capacity of the retry queue (in batches)
# TYPE otelcol_exporter_queue_capacity gauge
otelcol_exporter_queue_capacity{exporter="otlp",service_instance_id="e46d41fd-810a-4685-82d0-b01e177cfe7b",service_name="otelcorecol",service_version="0.90.0-dev"} 1000
# HELP otelcol_exporter_queue_size Current size of the retry queue (in batches)
# TYPE otelcol_exporter_queue_size gauge
otelcol_exporter_queue_size{exporter="otlp",service_instance_id="e46d41fd-810a-4685-82d0-b01e177cfe7b",service_name="otelcorecol",service_version="0.90.0-dev"} 0
# HELP otelcol_process_cpu_seconds Total CPU user and system time in seconds
# TYPE otelcol_process_cpu_seconds counter
otelcol_process_cpu_seconds{service_instance_id="e46d41fd-810a-4685-82d0-b01e177cfe7b",service_name="otelcorecol",service_version="0.90.0-dev"} 0.03
# HELP otelcol_process_memory_rss Total physical memory (resident set size)
# TYPE otelcol_process_memory_rss gauge
otelcol_process_memory_rss{service_instance_id="e46d41fd-810a-4685-82d0-b01e177cfe7b",service_name="otelcorecol",service_version="0.90.0-dev"} 2.70336e+07
# HELP otelcol_process_runtime_heap_alloc_bytes Bytes of allocated heap objects (see 'go doc runtime.MemStats.HeapAlloc')
# TYPE otelcol_process_runtime_heap_alloc_bytes gauge
otelcol_process_runtime_heap_alloc_bytes{service_instance_id="e46d41fd-810a-4685-82d0-b01e177cfe7b",service_name="otelcorecol",service_version="0.90.0-dev"} 2.445496e+06
# HELP otelcol_process_runtime_total_alloc_bytes Cumulative bytes allocated for heap objects (see 'go doc runtime.MemStats.TotalAlloc')
# TYPE otelcol_process_runtime_total_alloc_bytes counter
[...]

PR #9102, opened 2023-12-13
[telemetry] mark useOtelForInternalMetrics stable

codeboten:
This marks the flag as stable. Leaving this as a draft until v0.92.0 is released.

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/8962
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/816


TylerHelmuth (commented):
It's beautiful.  What impact does this have on components in Contrib, like [tailsamplingprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/metrics.go) or [filterprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/telemetry.go), that use OC?

mwear (approved):

atoulme:
Same question as Tyler regarding the Splunk HEC exporter: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/splunkhecexporter/heartbeat.go#L58

codeboten:
Should have no impact, the opencensus bridge is still in place to receive OC metrics

TylerHelmuth (approved):

codeboten:
@TylerHelmuth @atoulme the global OC registry is still usable by components: https://github.com/open-telemetry/opentelemetry-collector/blob/49e14a1480a180acd05d8139d1752b1943b2a92e/service/telemetry.go#L137

atoulme (approved):

splunkericl:
Can this PR wait for a couple releases? Our team relies on OpenCensus reporting and the metrics generated. Some metrics in queue_sender like queue size and queue capacity are important for our business.

We only notice the OpenCensus metrics is being removed because of the feature gate change in V0.92. I can imagine there are other users in the same boat. Would it make sense to wait for a couple releases so users have time to migrate off OpenCensus metrics?

codeboten:
>  Some metrics in queue_sender like queue size and queue capacity are important for our business.

Thanks for the feedback @splunkericl! Can you specify which metrics? Queue size and capacity were added to the opentelemetry generated metrics some time ago. See https://github.com/open-telemetry/opentelemetry-collector/pull/8716

mx-psi (approved):

codeboten:
> Our team relies on OpenCensus reporting and the metrics generated

@splunkericl just more follow up on this, any components that's using OpenCensus to generate metrics can continue to do so even with this change. The telemetry configuration for the Collector still supports the OpenCensus bridge.

splunkericl:
@codeboten ah I see. So the application generating open census metrics will see them in the prometheus server spawned? If that is the case that should be fine.

Our application has been scraping against the prometheus metrics endpoint and use these metrics for business logics. As long as the workflow is still the same it should be fine and gives us more time to migrate to start using OTEL metrics framework.

codeboten:
@splunkericl right. The metrics that are being exposed via prometheus today will continue to be exposed in the same manner, regardless of how those metrics are being recorded inside the collector itself. 

You can see a sample output of the metrics before and after the flag is used here: https://github.com/open-telemetry/opentelemetry-collector/pull/9029#issue-2021348799

Any component (in core or in contrib) that is still recording its metrics via opencensus will continue to see its metrics exported via the collector's internal prometheus endpoint. Note there is an open issue to migrate components away in the future though: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29867. This would allow us to remove a dependency on the opencensus library.

splunkericl:
Got it. Thanks for the explanations!

Issue #9070, opened 2023-12-11
[pdata] Inspect and update the pdata comments


dmitryax:
Inspect all the pdata comments related to the invalid usage and update them if needed. Similar to how it was resolved for `pcommon.Value` in Similar to https://github.com/open-telemetry/opentelemetry-collector/issues/9025, do the same for other API parts.

berkeli:
I'm happy to look into this one

PR #9107, opened 2023-12-13
[chore] [pdata] update comments to make it clear which functions panic

berkeli:
**Description:** 
This updates comments and adds tests based on https://github.com/open-telemetry/opentelemetry-collector/issues/9070

**Link to tracking Issue:** #9070

**Testing:** This only updates comments, but I have added unit tests to verify the behaviour described in comments


pdata/pcommon/map.go
```diff
@@ -55,8 +55,7 @@ func (m Map) EnsureCapacity(capacity int) {
 // It is allowed to modify the returned value using Value.Set* functions.
 // Such modification will be applied to the value stored in this map.
 //
-// If the key does not exist returns an invalid instance of the KeyValue and false.
-// Calling any functions on the returned invalid instance will cause a panic.
+// If the key does not exist returns a zero-initialized KeyValue and false.
```
berkeli, line 58: I have removed "Calling any functions on the returned invalid instance will cause a panic." because not all functions cause a panic. Methods on the value have relevant comments, so we can leave for the user to check docs per method instead of listing them here

TylerHelmuth: To make the risks immediately clear, how about 

`Calling any functions on the returned invalid instance may cause a panic.`

berkeli: Thanks I have added that comment!

TylerHelmuth (approved):

bogdandrutu (approved):

pdata/pcommon/slice.go
```diff
@@ -42,6 +42,7 @@ func NewSlice() Slice {
 // Len returns the number of elements in the slice.
 //
 // Returns "0" for a newly instance created with "NewSlice()".
+// Calling this function on zero-initialized Slice will cause a panic.
```
bogdandrutu, line 45: Instead of a comment on all funcs should we have a comment at struct level?

berkeli: So `Value` has this comment: 
https://github.com/open-telemetry/opentelemetry-collector/blob/6b12cc3203488429a3e3d101ee9e9dd152936c6a/pdata/pcommon/value.go#L68-L69

It also has method specific comments specifying when they panic. Happy to update, but when it comes to panics I think it doesn't hurt to have it in both places (I can add it on struct level for `Map`, `Slice`, etc).

dmitryax: Given that all the methods in the Slice cause panic while it's not the case for Value, I think it's fine to have it in the Slice's struct comment only.

```
// Must use NewSlice function to create new instances.
// Important: zero-initialized instance is not valid for use, any method call will result in a panic.
```

pdata/pcommon/trace_state.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 // TraceState represents the trace state from the w3c-trace-context.
+//
+// Must use NewTraceState function to create new instances.
+// Important: zero-initialized instance is not valid for use.
```
dmitryax, line 13: We can put the same comment as in `Map` given that all the methods panic?

dmitryax:
@berkeli can you please resolve the linter failures?

berkeli:
> @berkeli can you please resolve the linter failures?

Apologies, didn't notice failures. Fixed now 🙇

dmitryax (approved):

dmitryax:
Thanks @berkeli!

PR #9109, opened 2023-12-14
exporterhelper: simplify initialization and starting logic

bogdandrutu:


exporter/exporterhelper/common.go
```diff
@@ -27,195 +26,213 @@ type baseRequestSender struct {
...
+	// Create senders in reverse order:
+	// Start with the basic exporter wrapper.
+	be.exporterSender = &exporterSender{StartFunc: sets.startFunc, ShutdownFunc: sets.shutdownFunc}
+	be.addSender(be.exporterSender)
+	// Before that add the timeout sender.
+	be.addSender(newTimeoutSender(sets.timeoutConfig, be.lastSender))
+	// Before that add the retry sender if configured.
+	if sets.retryConfig != nil {
+		if !sets.retryConfig.Enabled {
+			// TODO: Simplify logic for error logging request. Consider to add it always if both retry_on_failure is not enabled, not only when configured but disabled.
```
dmitryax, line 158: > Consider to add it always if both retry_on_failure is not enabled, not only when configured but disabled.

I think it'd be pretty confusing to log "Try enabling retry_on_failure" if an exporter doesn't support it.

exporter/exporterhelper/common.go
```diff
@@ -27,195 +26,213 @@ type baseRequestSender struct {
...
+			be.retrySender = newRetrySender(*sets.retryConfig, set, be.lastSender)
+			be.addSender(be.retrySender)
+		}
+	}
+	// Before that add the observability sender.
+	be.addSender(osf(obsReport, be.lastSender))
+	// Before that add the queue sender if configured.
+	if sets.queueConfig != nil {
+		if !sets.queueConfig.Enabled {
+			// TODO: Consider to remove this. Not sure how this helps with dropping data.
```
dmitryax, line 174: I think we can remove this once https://github.com/open-telemetry/opentelemetry-collector/pull/9090 is merged

dmitryax (commented):
The change looks good to me. What I like about the existing code though is that it's pretty clear how the senders pipeline look like. This code makes it not that obvious.

bogdandrutu:
> The change looks good to me. What I like about the existing code though is that it's pretty clear how the senders pipeline look like. This code makes it not that obvious, but we can probably just add another comment with all the senders together.

Where do you get that from? I think actually this is cleaner in a way that you can see the sequence of constructing the chain independent of when the user adds the config.

exporter/exporterhelper/common.go
```diff
@@ -27,195 +26,213 @@ type baseRequestSender struct {
...
-func (be *baseExporter) send(ctx context.Context, req Request) error {
-	return be.queueSender.send(ctx, req)
-}
-
-// connectSenders connects the senders in the predefined order.
-func (be *baseExporter) connectSenders() {
-	be.queueSender.setNextSender(be.obsrepSender)
-	be.obsrepSender.setNextSender(be.retrySender)
-	be.retrySender.setNextSender(be.timeoutSender)
+func (be *baseExporter) addSender(sender requestSender) {
```
dmitryax, line 194: It'd be good to have a comment here that the sender is set in front of the current chain.

dmitryax:
> Where do you get that from? I think actually this is cleaner in a way that you can see the sequence of constructing the chain independent of when the user adds the config.

[The connectSenders](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/common.go#L197) seems to play that role. Maybe just having a comment highlighting a typical chain of senders like `queueSender -> retrySender -> timeoutSender -> exporterSenedr` is just enough.

dmitryax:
Also having to call `addSender` func and pass the `nextSender` to the constructor introduces two ways to link the sender, which needs to be kept in sync. Not very convenient. Maybe we can drop `addSender/allSenders` and call the Start methods explicitly the same as Shutdown?

PR #9121, opened 2023-12-15
[chore] remove TODO from confighttp tests

atoulme:
Since the listener is already listening, the Serve call will not introduce a delay in processing requests.
It is then fair to block the main thread on waiting the goroutine to be started, and unblock before Serve is called.

### DIFF
--- a/config/confighttp/confighttp_test.go
+++ b/config/confighttp/confighttp_test.go
@@ -817,13 +817,14 @@ func TestHttpCors(t *testing.T) {
 				}))
 			require.NoError(t, err)
 
+			waitStart := make(chan struct{})
 			go func() {
+				close(waitStart)
 				_ = s.Serve(ln)
 			}()
 
-			// TODO: make starting server deterministic
 			// Wait for the servers to start
-			<-time.After(10 * time.Millisecond)
+			<-waitStart
 
 			url := fmt.Sprintf("http://%s", ln.Addr().String())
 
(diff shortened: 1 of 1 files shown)


config/confighttp/confighttp_test.go
```diff
@@ -817,13 +817,14 @@ func TestHttpCors(t *testing.T) {
...
+			waitStart := make(chan struct{})
 			go func() {
+				close(waitStart)
 				_ = s.Serve(ln)
 			}()
 
-			// TODO: make starting server deterministic
 			// Wait for the servers to start
-			<-time.After(10 * time.Millisecond)
+			<-waitStart
```
bogdandrutu, line 828: We usually do a WaitGroup for go routines, and I am not saying this is not correct. But let's be consistent and use that everywhere, simplifies things.

Also, this is still not guaranteed to have the server started unfortunately.

atoulme: ok sure thing

atoulme: I think there are some good guarantees this will work. I have tried and reproduced the scenario with this Go playground: https://go.dev/play/p/yrBE49_-cbk
I have based my work off this SO thread: https://stackoverflow.com/questions/32738188/how-can-i-start-the-browser-after-the-server-started-listening/32742904#32742904

atoulme: Based on those findings, I can even try to remove all the time.Sleep statements as we shouldn't have to wait at all.

bogdandrutu (approved):

PR #9123, opened 2023-12-15
[chore] remove confighttp TODO

atoulme:
Remove a TODO recommending to look at set the operation name prefix as the component ID.
We already set the exporter ID as the name of the tracer in [receiverhelper](https://github.com/open-telemetry/opentelemetry-collector/blob/b5635a7a90d204453915db36e4e29182ee2dddb2/receiver/receiverhelper/obsreport.go#L79).
The component ID is also not available in this section of the code.

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -358,7 +358,6 @@ func (hss *HTTPServerSettings) ToServer(host component.Host, settings component.
 	}
 
 	// Enable OpenTelemetry observability plugin.
-	// TODO: Consider to use component ID string as prefix for all the operations.
 	handler = otelhttp.NewHandler(
 		handler,
 		"",


codeboten (commented):
If we're not setting the component ID, would it be worth at least recording some information about the operation? I'm not familiar with what this data looks like today.

> We already set the exporter ID as the name of the tracer in [exporterhelper]

`ToServer` code is in in the receiver path right? Do we do the same for receivers? If so, is that enough information on the trace?

atoulme:
> ToServer code is in in the receiver path right? Do we do the same for receivers? If so, is that enough information on the trace?
do you mean the same for exporters? Yes, for exporters the tracer provider sets the name of the tracer according to the exporter ID.

codeboten:
> do you mean the same for exporters?

The description for the PR references exporters, but the TODO is in the `ToServer` which is generally used by receivers (except for pull exporters) which is why i was asking about receivers

atoulme:
> > do you mean the same for exporters?
> 
> The description for the PR references exporters, but the TODO is in the `ToServer` which is generally used by receivers (except for pull exporters) which is why i was asking about receivers

Sorry, updated the PR title. Here is where the receiverhelper code sets up the tracer: https://github.com/open-telemetry/opentelemetry-collector/blob/b5635a7a90d204453915db36e4e29182ee2dddb2/receiver/receiverhelper/obsreport.go#L79

bogdandrutu:
>  Here is where the receiverhelper code sets up the tracer:

This code does not use that, we pass directly the providers to the http integration.

atoulme:
Understood and apologies for my bad reading. I'll close this PR.

Issue #9129, opened 2023-12-18
otlp: balancer_name: round_robin doesn't make it round_robin


sokoide:
**Describe the bug**
In the following network,

```
OTELCOL --(gRPC)--> foo.com ---> receiver1.com
                            |--> receiver2.com
```

When the OTELCOL sends to foo.com who has 2 DNS A records for reciver1.com and receier2.com, the following configuration doesn't make the export round-robin. It always export to receiver1.com

```yaml
exporters:
  otlp:
    endpoint: "foo.com:4317"
    tls:
      insecure: true
    balancer_name: round_robin
```

**Steps to reproduce**

1. Run OTELCOL with the config above including `balancer_name: round_robin`
2. Run OTLP receivers on receiver1.com and receiver2.com with TLS off (TLS doesn't matter though)
3. Send traces to the OTELCOL over OTLP

**What did you expect to see?**
The first batch of the traces goes to receiver1, second goes to receiver2, the third goes to receiver1... (round_robin)

**What did you see instead?**
All trace buckets go to receiver1.

**What version did you use?**
v0.91.0

**What config did you use?**

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: ":4317"
exporters:
  otlp:
    endpoint: "foo.com:4317"
    tls:
      insecure: true
    balancer_name: round_robin

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: []
      exporters: [otlp]
```

**Environment**

OS:  Ubuntu 22.04 x64
Compiler: Go 1.21.5

**Additional context**

`balancer_name: round_robin` should be supported according to [the spec](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/README.md#client-configuration)

A question asked as #8000.

sokoide:
I confirmed the round_robin feature works with `dns:` prefix in the endpoint. Please refer to #9130 for more details.

PR #9130, opened 2023-12-18
fix: otlp exporter round_robin

sokoide:
**Description:**
Fix a bug #9129

**Link to tracking Issue:**
#9129

**Testing:**
Tested with 2 DNS A records for the export OTLP targets. Confirmed the round_robin behavior with out decreasing the throughput.

```yaml
exporters:
  otlp:
    endpoint: "foo.com:4317"
    tls:
      insecure: true
    balancer_name: round_robin
```

DNS
```
dig foo.com  
;; QUESTION SECTION:  
;foo.com.             IN      A  
   
;; ANSWER SECTION:  
foo.com. 2       IN      A 192.168.2.201  
foo.com. 2       IN      A 192.168.2.202  
```

Confirmed the traces went to 192.168.2.201 first, then 202, then 201, ... (round robin)

**Documentation:**
The [document (spec) is already available here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/README.md#client-configuration)
The patch will enable the `balancer_name: round_robin`.

### DIFF
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -18,6 +18,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/balancer"
+	"google.golang.org/grpc/balancer/roundrobin"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
@@ -181,7 +182,17 @@ func (gcs *GRPCClientSettings) ToClientConn(ctx context.Context, host component.
 		return nil, err
 	}
 	opts = append(opts, extraOpts...)
-	return grpc.DialContext(ctx, gcs.SanitizedEndpoint(), opts...)
+
+	// gRPC round_robin is only enabled when the DNS resolver is used.
+	// There are 2 options to use it.
+	// This function takes #2.
+	// 1. configure the global resolver by `resolver.SetDefaultScheme("dns")`
+	// 2. add "dns:[//authority/]" prefix in grpc.DialContext
+	var endpoint string = gcs.SanitizedEndpoint()
+	if gcs.BalancerName == roundrobin.Name {
+		endpoint = "dns:" + endpoint
+	}
+	return grpc.DialContext(ctx, endpoint, opts...)
 }
 
 func (gcs *GRPCClientSettings) toDialOptions(host component.Host, settings component.TelemetrySettings) ([]grpc.DialOption, error) {
@@ -245,7 +256,8 @@ func (gcs *GRPCClientSettings) toDialOptions(host component.Host, settings compo
 		if !valid {
 			return nil, fmt.Errorf("invalid balancer_name: %s", gcs.BalancerName)
 		}
-		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingPolicy":"%s"}`, gcs.BalancerName)))
+
+		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingConfig": [{"%s":{}}]}`, gcs.BalancerName)))
 	}
 
 	if gcs.Authority != "" {


config/configgrpc/configgrpc.go
```diff
@@ -245,7 +246,11 @@ func (gcs *GRPCClientSettings) toDialOptions(host component.Host, settings compo
 		if !valid {
 			return nil, fmt.Errorf("invalid balancer_name: %s", gcs.BalancerName)
 		}
-		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingPolicy":"%s"}`, gcs.BalancerName)))
+
+		if gcs.BalancerName == "round_robin" {
+			resolver.SetDefaultScheme("dns")
+			opts = append(opts, grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`))
+		}
```
TylerHelmuth, line 253: This will break `BalancerName`s that are not `round_robin`.

TylerHelmuth:
Please sign the CLA

config/configgrpc/configgrpc.go
sokoide: @TylerHelmuth , thank you for your review.

If you have `balancer_name: foo`, otelcorecol fails to start with the following error.
```
Error: cannot start pipelines: invalid balancer_name: foo
```

If you have `balancer_name:` (empty), or doesn't have the `balancer_name:` node, it defaults to `pick_first`.  Are there other patterns I should test and confirm?

sokoide:
> Please sign the CLA

I am a member of a corporate contributor, but I forgot to make the membership public. I made the membership public and commented "/easycla" to retrigger the scan, but it looks it's not reevaluated. Is there anything I can try?

config/configgrpc/configgrpc.go
TylerHelmuth: I mean to say that previously if `BalancerName` was configured we explicitly set `loadBalancingPolicy` on the service config.  Although `pick_first` may be the default (and currently only other option), I like continuing to explicitly set a value whenever `BalancerName` is set.  This defensive code ensures we are compatible with any `BalancerName` options added in the future.

sokoide: @TylerHelmuth , thank you for the clear explanation. It makes sense. Updated the PR.

config/configgrpc/configgrpc.go
```diff
@@ -245,7 +246,9 @@ func (gcs *GRPCClientSettings) toDialOptions(host component.Host, settings compo
 		if !valid {
 			return nil, fmt.Errorf("invalid balancer_name: %s", gcs.BalancerName)
 		}
-		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingPolicy":"%s"}`, gcs.BalancerName)))
+
+		resolver.SetDefaultScheme("dns")
```
yurishkuro, line 250: Changing a global setting like this has much larger implications than fixing a specific issue. Isn't there more local way of doing it?

config/configgrpc/configgrpc.go
```diff
@@ -181,7 +182,17 @@ func (gcs *GRPCClientSettings) ToClientConn(ctx context.Context, host component.
...
-	return grpc.DialContext(ctx, gcs.SanitizedEndpoint(), opts...)
+
+	// gRPC round_robin is only enabled when the DNS resolver is used.
+	// There are 2 options to use it.
+	// This function takes #2.
+	// 1. configure the global resolver by `resolver.SetDefaultScheme("dns")`
+	// 2. add "dns:[//authority/]" prefix in grpc.DialContext
+	var endpoint string = gcs.SanitizedEndpoint()
+	if gcs.BalancerName == roundrobin.Name {
+		endpoint = "dns:" + endpoint
```
yurishkuro, line 193: Why should this be hardcoded instead of user-provided in the config? What if another user wants to use a different resolver?

sokoide: @yurishkuro, thank you for your comment! I confirmed the following config let Traefik use the DNS resolver and round_robin option worked. 

```yaml
exporters:
  otlp:
    endpoint: "dns://foo.com:4317"
  balancer_name: round_robin
```

Because it looks changing from `loadBalancingPolicy` to `loadBalancingConfig` is not required yet, I'll close the PR.

sokoide: correction.
```yaml
exporters:
  otlp:
    endpoint: "dns:///foo.com:4317"
  balancer_name: round_robin
```

or
```yaml
exporters:
  otlp:
    endpoint: "dns:foo.com:4317"
  balancer_name: round_robin
```

PR #9131, opened 2023-12-18
[service] remove code to configure otel SDK

codeboten:
This code will live in the otel-go-contrib `config` package instead. This PR depends on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4617

~Leaving this as draft until config package is updated. Ideally I'd like to get the MeterProvider configured in this PR as well~

Addressing only TracerProvider in this initial PR.


mx-psi (approved):
Should we wait for the next `config` release?

codeboten:
> Should we wait for the next `config` release?

I don't have a strong preference here, happy to go either ways

mx-psi:
> > Should we wait for the next `config` release?
> 
> I don't have a strong preference here, happy to go either ways

They may have to do a release anyway for the gRPC issue, so I would hold on merging this for a bit to see if we get a release for this module as well.

mx-psi (approved):
LGTM, I merged main and bumped to v0.2.0 which has the changes we need for this

service/telemetry.go
```diff
@@ -75,13 +74,20 @@ func (tel *telemetryInitializer) init(res *resource.Resource, settings servicete
 	}
 
 	settings.Logger.Info("Setting up own telemetry...")
+	configuredSDK, err := config.NewSDK(
```
mx-psi, line 77: One question: are we not passing the resource anymore here?

codeboten: This is a good question. I'm working on an issue in the go-contrib repo where setting resource attributes via config is currently restricted to service name, as the go-jsonschema library doesn't have support for additionalProperties on an object that has other properties specified.

Putting this back into draft until that's resolved (https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4828)

codeboten: @mx-psi here's roughly what i think the change will look like with a resource: https://github.com/open-telemetry/opentelemetry-collector/pull/9131/commits/51f6ed6a554182c23febabe2bb80827dfddf55e1

the code isn't in the config package yet though, so it will likely change

mx-psi: :+1: Makes sense, let's wait on that issue to be solved then :)

Issue #5935, opened 2022-08-18
[pdata] Allow strict enforcement of json payloads.


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**
To enable the collector to receive json payloads from new/old otel exports, the collector was changed to skip unknown fields instead of error (#5312, #5931).  This is a proper default behavior, but there may be situations where users rely on the collector strictly enforcing json payload structure.  It would be nice if the collector provided the option to enable strict enforcement of json payloads.

**Describe the solution you'd like**
Maybe a startup option for the collector?  Or something in the collector config's service section.

**Describe alternatives you've considered**
Pushing responsibility to receivers.  This is possible, but would require individual receivers to make changes/choices, versus the collector enforcing the structure for all incoming data.

PR #9139, opened 2023-12-19
[exporter/otlp] add error message when no spans can be read from JSON

### LINKED ISSUE #6320: Proto mismatch errors not obvious when posting traces
**Describe the bug**
When sending test traces via `otlp-http`, if the format of the trace is mismatched the server still returns a 200 with no indication of a problem.

```bash
$ curl -i http://localhost:4318/v1/traces -X POST -H "Content-Type: application/json" -d @trace.json

HTTP/1.1 200 OK
Content-Type: application/json
Date: Sat, 15 Oct 2022 00:30:09 GMT
Content-Length: 2

{}
```

**Steps to reproduce**
config: _BELOW_

docker-compose.yaml:
```yaml
version: "3"
services:
  otel-collector:
    image: otel/opentelemetry-collector:0.60.0
    command: ["--config=/etc/otel-collector-config.yaml"]
    volumes:
      - ./config.yaml:/etc/otel-collector-config.yaml
    ports:
      - "4317:4317"   # OTLP gRPC receiver
      - "4318:4318"   # OTLP http receiver
      - "13133:13133" # Health Check
```

trace.json ([from this example](https://docs.lightstep.com/otel/send-otlp-over-http-to-lightstep))
```
{
  "resourceSpans": [
    {
      "resource": {
        "attributes": [
          {
            "key": "service.name",
            "value": {
              "stringValue": "curl-test-otel-pipeline"
            }
          }
        ]
      },
      "instrumentationLibrarySpans": [
        {
          "spans": [
            {
              "traceId": "71699b6fe85982c7c8995ea3d9c95df2",
              "spanId": "3c191d03fa8be065",
              "name": "test-span",
              "kind": 1,
              "droppedAttributesCount": 0,
              "events": [],
              "droppedEventsCount": 0,
              "status": {
                "code": 1
              }
            }
          ],
          "instrumentationLibrary": {
            "name": "local-curl-example"
          }
        }
      ]
    }
  ]
 }
```

1. `docker compose up`
2. from another terminal `curl -i http://localhost:4318/v1/traces -X POST -H "Content-Type: application/json" -d @trace.json`

A 200 response if the trace data was 👍
Anything else with a message mentioning the proto mistmatch

**What did you see instead?**
```bash
$ curl -i http://localhost:4318/v1/traces -X POST -H "Content-Type: application/json" -d @trace.json

HTTP/1.1 200 OK
Content-Type: application/json
Date: Sat, 15 Oct 2022 00:30:09 GMT
Content-Length: 2

{}
```

**What version did you use?**
Version: `0.60.0`

**What config did you use?**
```yaml
exporters:
  logging:

extensions:
  health_check: {}
  memory_ballast: {}

processors:
  batch: {}
  memory_limiter:
    check_interval: 5s
    limit_mib: 409
    spike_limit_mib: 128

receivers:
  otlp:
    protocols:
      grpc:
      http:

service:
  extensions:
  - health_check
  - memory_ballast
  pipelines:
    traces:
      exporters:
      - logging
      - otlp
      processors:
      - memory_limiter
      - batch
      receivers:
      - otlp
  telemetry:
    logs:
      level: "debug"
```

atoulme:
**Description:**
If no spans can be read from the JSON payload, return an error message.

**Link to tracking Issue:**
Fixes #6320

**Testing:**
Unit test.

### DIFF
--- a/receiver/otlpreceiver/internal/trace/otlp.go
+++ b/receiver/otlpreceiver/internal/trace/otlp.go
@@ -11,7 +11,10 @@ import (
 	"go.opentelemetry.io/collector/receiver/receiverhelper"
 )
 
-const dataFormatProtobuf = "protobuf"
+const (
+	dataFormatProtobuf = "protobuf"
+	noSpans            = "no spans read in payload"
+)
 
 // Receiver is the type used to handle spans from OpenTelemetry exporters.
 type Receiver struct {
@@ -34,7 +37,9 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
 	// We need to ensure that it propagates the receiver name as a tag
 	numSpans := td.SpanCount()
 	if numSpans == 0 {
-		return ptraceotlp.NewExportResponse(), nil
+		resp := ptraceotlp.NewExportResponse()
+		resp.PartialSuccess().SetErrorMessage(noSpans)
+		return resp, nil
 	}
 
 	ctx = r.obsreport.StartTracesOp(ctx)
--- a/receiver/otlpreceiver/otlp_test.go
+++ b/receiver/otlpreceiver/otlp_test.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


receiver/otlpreceiver/internal/trace/otlp.go
```diff
@@ -34,7 +37,9 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
 	// We need to ensure that it propagates the receiver name as a tag
 	numSpans := td.SpanCount()
 	if numSpans == 0 {
-		return ptraceotlp.NewExportResponse(), nil
+		resp := ptraceotlp.NewExportResponse()
+		resp.PartialSuccess().SetErrorMessage(noSpans)
+		return resp, nil
```
TylerHelmuth, line 42: I'm not sure this is correct for the OTLP specification.  The spec says:

`If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.`

- https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#full-success
- https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#full-success-1

atoulme: OK, so this approach is invalid, I'll reframe.

bogdandrutu: Do we still need this PR?

atoulme: I want to reuse the test and fix the impl.

receiver/otlpreceiver/otlp_test.go
```diff
@@ -384,6 +385,87 @@ func TestHandleInvalidRequests(t *testing.T) {
...
+		cfg,
+		consumertest.NewNop())
+	require.NoError(t, err)
+	assert.NotNil(t, tr)
+	require.NoError(t, tr.Start(context.Background(), componenttest.NewNopHost()))
+	t.Cleanup(func() {
+		require.NoError(t, tr.Shutdown(context.Background()))
+	})
+
+	traceJSON := `{
```
bogdandrutu, line 416: What is bad about this request?

receiver/otlpreceiver/internal/trace/otlp.go
bogdandrutu: What exactly are you trying to test?

receiver/otlpreceiver/otlp_test.go
atoulme: the span is invalid as its fields are encoded with camel case instead of snake case, such as `traceId` instead of `trace_id`.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#json-protobuf-encoding

receiver/otlpreceiver/internal/trace/otlp.go
atoulme: The issue associated talks about it a bit and I felt I could probably try and take an approach where the partial success response would return an error message that indicates to the user that no spans were found in payload. The better fix might be to apply to fix https://github.com/open-telemetry/opentelemetry-collector/issues/5935 now, given the feedback.

receiver/otlpreceiver/otlp_test.go
bogdandrutu: This is not an invalid message per the definition of the protocol.

atoulme: ok sorry. https://github.com/open-telemetry/opentelemetry-collector/issues/6320#issuecomment-1279738087 I was looking in the wrong place.

bogdandrutu: That comment says that we cannot do anything unfortunately for that.

bogdandrutu:
Closing since the request is valid in the test. Feel free to open if we need other changes.

PR #9140, opened 2023-12-19
move contract test out of component

### LINKED ISSUE #6454: remove otlphttpreceiver dependency from otlpexporter
Currently the `otlpexporter` has a dependency on `otlphttpreceiver` for testing purposes. This causes any package that includes the `otlpexporter` to also include the `otlphttpreceiver` unnecessarily.

atoulme:
**Description:**
remove dependency of otlphttpreceiver on otlpexporter

**Link to tracking Issue:**
Fixes #6454


internal/test/consume_contract_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package otlpexporter
+package test
```
bogdandrutu, line 4: If we move this we create again the same circular dependency and even more:

otlpexporter -> main otel package (with internal)
otlpreceiver -> main otel package (with internal)

Now after this change `main otel package (with internal)` depends on `otlpexporter` and `otlpreceiver`

atoulme: Can we create a new module for those tests? I was asking @codeboten if there is a package we can use

bogdandrutu: We can definitely have some "e2e" module that we don't "export" (in internal).

atoulme: Thanks! I'll get on that right away.

internal/e2e/go.mod
```diff
@@ -0,0 +1,125 @@
+module go.opentelemetry.io/collector/internal/e2e
```
bogdandrutu, line 1: We should have it in the `versions.yaml` under ignored.

bogdandrutu (approved):

internal/test/consume_contract_test.go
atoulme: ok this is done, the diff is much better

codeboten (approved):
Nice! this is great, thanks @atoulme

codeboten:
> Replace statements are out of date, please run "make crosslink" and commit the changes in this PR.

atoulme:
good to go

PR #9141, opened 2023-12-19
[component] deprecate component.UnmarshalConfig

### LINKED ISSUE #7102: Remove component.UnmarshalConfig

atoulme:
**Description:**
Deprecate component.UnmarshalConfig. Prefer `conf.Unmarshal(cfg)` instead
This PR also makes the `WithErrorUnused()` option enabled by default.

**Link to tracking Issue:**
Fixes #7102

**Documentation:**
Added deprecation information

### DIFF
--- a/component/config.go
+++ b/component/config.go
@@ -27,12 +27,9 @@ var configValidatorType = reflect.TypeOf((*ConfigValidator)(nil)).Elem()
 // UnmarshalConfig helper function to UnmarshalConfig a Config.
 // It checks if the config implements confmap.Unmarshaler and uses that if available,
 // otherwise uses Map.UnmarshalExact, erroring if a field is nonexistent.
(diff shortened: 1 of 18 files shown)


component/config.go
```diff
@@ -27,12 +27,9 @@ var configValidatorType = reflect.TypeOf((*ConfigValidator)(nil)).Elem()
 // UnmarshalConfig helper function to UnmarshalConfig a Config.
 // It checks if the config implements confmap.Unmarshaler and uses that if available,
 // otherwise uses Map.UnmarshalExact, erroring if a field is nonexistent.
+// Deprecated: Use `conf.Unmarshal(cfg)` instead.
```
bogdandrutu, line 30: See the comment, and why this is needed

atoulme:
Indeed, that's not a good solution, closing.

PR #9142, opened 2023-12-19
[configtls] add include_system_ca_certs_pool

### LINKED ISSUE #7774: Allows users to configure to include system certs pool in the request
As a user of `TLSSetting`, I want an option to allow appending ca certs on top of the system certs pool. Today, if the user specify any CA file for `TLSSetting` it will only use that.

The background is the the application our exporter calls can have either cacert that is well known(e.g: Digi cert that is available in os) or private(e.g: Splunk CMP stack). We want the exporter to be able to communicate to both.

Add a new config option in `TLSSetting`
```
type TLSSetting struct {
   UseSystemCACerts bool
}
```

when loading ca cert:
```
func (c TLSSetting) loadCert(caPath string) (*x509.CertPool, error) {
	caPEM, err := os.ReadFile(filepath.Clean(caPath))
	if err != nil {
		return nil, fmt.Errorf("failed to load CA %s: %w", caPath, err)
	}

	var certPool *x509.CertPool
	if c.UseSystemCACerts {
		certPool, _ = x509.SystemCertPool()
	}
	if certPool == nil {
		certPool = x509.NewCertPool()
	}
	if !certPool.AppendCertsFromPEM(caPEM) {
		return nil, fmt.Errorf("failed to parse CA %s", caPath)
	}
	return certPool, nil
}
```

An alternative is to always load system certs in the ca pool. However, this would increase HTTPs traffic payload for everyone

**Additional context**
N/A

atoulme:
**Description:**
Add `include_system_ca_certs_pool` to configtls, allowing to load system certs and additional custom certs.

**Link to tracking Issue:**
Fixes #7774

**Testing:**
UTs

**Documentation:**
Added to README

### DIFF
--- a/config/configtls/configtls.go
+++ b/config/configtls/configtls.go
@@ -29,9 +29,6 @@ var systemCertPool = x509.SystemCertPool
 // Note: Since there isn't anything specific to a server connection. Components
 // with server connections should use TLSSetting.
 type TLSSetting struct {
-	// If true, load system CA certificates pool in addition to the certificates
-	// configured in this struct.
-	IncludeSystemCACertsPool bool `mapstructure:"include_system_ca_certs_pool"`
 	// Path to the CA cert. For a client this verifies the server certificate.
 	// For a server this verifies client certificates. If empty uses system root CA.
 	// (optional)
@@ -39,6 +36,10 @@ type TLSSetting struct {
 
 	// In memory PEM encoded cert. (optional)
 	CAPem configopaque.String `mapstructure:"ca_pem"`
+
+	// If true, load system CA certificates pool in addition to the certificates
+	// configured in this struct.
+	IncludeSystemCACertsPool bool `mapstructure:"include_system_ca_certs_pool"`
 
 	// Path to the TLS cert to use for TLS required connections. (optional)
 	CertFile string `mapstructure:"cert_file"`
--- a/config/configtls/configtls.go
+++ b/config/configtls/configtls.go
@@ -27,9 +27,9 @@ const defaultMaxTLSVersion = 0
 // Note: Since there isn't anything specific to a server connection. Components
 // with server connections should use TLSSetting.
 type TLSSetting struct {
-	// If true, load system certificates in addition to the certificates
+	// If true, load system CA certificates pool in addition to the certificates
 	// configured in this struct.
-	UseSystemCACerts bool `mapstructure:"use_system_ca_certs"`
+	IncludeSystemCACertsPool bool `mapstructure:"include_system_ca_certs_pool"`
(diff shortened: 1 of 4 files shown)


config/configtls/configtls.go
```diff
@@ -26,6 +26,9 @@ const defaultMaxTLSVersion = 0
 // Note: Since there isn't anything specific to a server connection. Components
 // with server connections should use TLSSetting.
 type TLSSetting struct {
+	// If true, load system certificates in addition to the certificates
+	// configured in this struct.
+	UseSystemCACerts bool `mapstructure:"use_system_ca_certs"`
```
TylerHelmuth, line 31: The name implies to me that we are going to use the System CA Certs instead of other options supplied.  How about `IncludeSystemCACerts` or maybe something with `Pool` in the name.

atoulme: done!

config/configtls/configtls.go
```diff
@@ -274,7 +277,14 @@ func (c TLSSetting) loadCert(caPath string) (*x509.CertPool, error) {
 		return nil, fmt.Errorf("failed to load CA %s: %w", caPath, err)
 	}
 
-	certPool := x509.NewCertPool()
+	var certPool *x509.CertPool
+	if c.IncludeSystemCACertsPool {
+		if certPool, err = x509.SystemCertPool(); err != nil {
+			return nil, err
+		}
```
codeboten, line 284: Maybe just being overly paranoid, but is there any code path where there isn't an error and the pool is still uninitialized?

atoulme: I'm so sorry to have to confirm that you do indeed sound strangely unhinged, but I can maybe check if certPool is nil and do something about it?

atoulme: Joke aside, thanks for the feedback, I have changed the code to be a bit more defensive.

config/configtls/configtls.go
```diff
@@ -22,10 +22,15 @@ const defaultMinTLSVersion = tls.VersionTLS12
...
 
+var systemCertPool = x509.SystemCertPool
+
 // TLSSetting exposes the common client and server TLS configurations.
 // Note: Since there isn't anything specific to a server connection. Components
 // with server connections should use TLSSetting.
 type TLSSetting struct {
+	// If true, load system CA certificates pool in addition to the certificates
+	// configured in this struct.
+	IncludeSystemCACertsPool bool `mapstructure:"include_system_ca_certs_pool"`
```
dmitryax, line 33: Let's move it according to the docs after `CAPem `?  Also, please add a line break as other fields have

atoulme: done

dmitryax (approved):

PR #9147, opened 2023-12-19
[chore] [exporterhelper] Add an option for items based queue sizing

dmitryax:
Introduce an option to limit the queue size by the number of items instead of number of requests. This is preliminary step for having the exporter helper v2 with a batcher sender placed after the queue sender. Otherwise, it'll be hard for the users to estimate the queue size based on the number of requests without batch processor in front of it.

This change doesn't effect the existing functionality and the items based queue limiting cannot be utilized yet.

### DIFF
--- a/exporter/exporterhelper/internal/bounded_memory_queue.go
+++ b/exporter/exporterhelper/internal/bounded_memory_queue.go
@@ -16,35 +16,40 @@ import (
 // the producer are dropped.
 type boundedMemoryQueue[T any] struct {
 	component.StartFunc
-	items chan queueRequest[T]
+	QueueCapacityLimiter[T]
+	inCh  chan<- queueRequest[T]
+	outCh <-chan queueRequest[T]
 }
 
 // NewBoundedMemoryQueue constructs the new queue of specified capacity, and with an optional
 // callback for dropped items (e.g. useful to emit metrics).
-func NewBoundedMemoryQueue[T any](capacity int) Queue[T] {
+func NewBoundedMemoryQueue[T any](capacityLimiter QueueCapacityLimiter[T]) Queue[T] {
+	inCh, outCh := newUnboundedChannel[queueRequest[T]]()
 	return &boundedMemoryQueue[T]{
-		items: make(chan queueRequest[T], capacity),
+		QueueCapacityLimiter: capacityLimiter,
+		inCh:                 inCh,
+		outCh:                outCh,
 	}
 }
 
 // Offer is used by the producer to submit new item to the queue. Calling this method on a stopped queue will panic.
 func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
-	select {
-	case q.items <- queueRequest[T]{ctx: ctx, req: req}:
-		return nil
-	default:
+	if !q.QueueCapacityLimiter.claim(req) {
 		return ErrQueueIsFull
 	}
+	q.inCh <- queueRequest[T]{ctx: ctx, req: req}
+	return nil
 }
 
 // Consume applies the provided function on the head of queue.
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
 func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
-	item, ok := <-q.items
+	item, ok := <-q.outCh
 	if !ok {
 		return false
 	}
+	q.QueueCapacityLimiter.release(item.req)
 	// the memory queue doesn't handle consume errors
 	_ = consumeFunc(item.ctx, item.req)
 	return true
@@ -52,17 +57,8 @@ func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) err
 
 // Shutdown closes the queue channel to initiate draining of the queue.
 func (q *boundedMemoryQueue[T]) Shutdown(context.Context) error {
-	close(q.items)
+	close(q.inCh)
 	return nil
-}
-
-// Size returns the current size of the queue
-func (q *boundedMemoryQueue[T]) Size() int {
-	return len(q.items)
-}
-
-func (q *boundedMemoryQueue[T]) Capacity() int {
-	return cap(q.items)
 }
 
 type queueRequest[T any] struct {
--- a/exporter/exporterhelper/internal/bounded_memory_queue_test.go
+++ b/exporter/exporterhelper/internal/bounded_memory_queue_test.go
@@ -8,7 +8,6 @@ package internal
 import (
 	"context"
 	"errors"
-	"fmt"
 	"strconv"
 	"sync"
 	"testing"
@@ -23,7 +22,7 @@ import (
 // We want to test the overflow behavior, so we block the consumer
 // by holding a startLock before submitting items to the queue.
 func TestBoundedQueue(t *testing.T) {
-	q := NewBoundedMemoryQueue[string](1)
+	q := NewBoundedMemoryQueue[string](NewRequestsCapacityLimiter[string](1))
 
 	assert.NoError(t, q.Offer(context.Background(), "a"))
 
@@ -73,7 +72,7 @@ func TestBoundedQueue(t *testing.T) {
 // only after Stop will mean the consumers are still locked while
 // trying to perform the final consumptions.
 func TestShutdownWhileNotEmpty(t *testing.T) {
-	q := NewBoundedMemoryQueue[string](1000)
+	q := NewBoundedMemoryQueue[string](NewRequestsCapacityLimiter[string](1000))
 
 	assert.NoError(t, q.Start(context.Background(), componenttest.NewNopHost()))
 	for i := 0; i < 10; i++ {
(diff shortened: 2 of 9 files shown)


exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -16,19 +16,24 @@ import (
...
 	return &boundedMemoryQueue[T]{
-		items: make(chan queueRequest[T], capacity),
+		QueueCapacityLimiter: capacityLimiter,
+		items:                make(chan queueRequest[T], capacityLimiter.Capacity()),
 	}
 }
 
 // Offer is used by the producer to submit new item to the queue. Calling this method on a stopped queue will panic.
 func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
+	if !q.QueueCapacityLimiter.claim(req) {
```
bogdandrutu, line 34: We should have an unbounded channel then and rely only on the QueueCapacityLimiter for this.

exporter/exporterhelper/internal/bounded_memory_queue_test.go
```diff
@@ -23,7 +23,7 @@ import (
 // We want to test the overflow behavior, so we block the consumer
 // by holding a startLock before submitting items to the queue.
 func TestBoundedQueue(t *testing.T) {
-	q := NewBoundedMemoryQueue[string](1)
+	q := NewBoundedMemoryQueue[string](NewNopCapacityLimiter[string](1))
```
bogdandrutu, line 26: Probably fake is better than NewNopCapacityLimiter? Of NumRequestsCapacityLimiter which adds 1 per every request.

exporter/exporterhelper/internal/persistent_queue.go
```diff
@@ -41,6 +42,8 @@ import (
 //	 index          index   x
 //	                        xxxx deleted
 type persistentQueue[T any] struct {
+	QueueCapacityLimiter[T]
```
bogdandrutu, line 45: Why embedding ?

exporter/exporterhelper/internal/bounded_memory_queue.go
```diff
@@ -16,19 +16,24 @@ import (
 // the producer are dropped.
 type boundedMemoryQueue[T any] struct {
 	component.StartFunc
+	QueueCapacityLimiter[T]
```
bogdandrutu, line 19: Why embedding ?

exporter/exporterhelper/internal/persistent_queue.go
dmitryax: To have the `Capacity()` implemented by the limiter

exporter/exporterhelper/internal/bounded_memory_queue_test.go
dmitryax: Changed to use `NewRequestsCapacityLimiter`

exporter/exporterhelper/internal/bounded_memory_queue.go
dmitryax: I updated it to use an unbounded channel and utilize the request based limiter 👍

exporter/exporterhelper/internal/bounded_memory_queue.go
dmitryax: Same as above. Also, with changing `nopCapacityLimiter` to `requestsCapacityLimiter`, now we can reuse `Size()` as well

exporter/exporterhelper/internal/unbounded_channel.go
```diff
@@ -0,0 +1,34 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package internal // import "go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+
+// newUnboundedChannel creates a pair of buffered channels with no cap.
+// Caller should close the input channel when done. That will trigger draining of the output channel.
+// Closing the output channel by client will panic.
+func newUnboundedChannel[T any]() (chan<- T, <-chan T) {
```
bogdandrutu, line 9: Why do you need two channels and a goroutine here. I am confused why not just a simple unbounded channel.

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+func (icl *itemsCapacityLimiter[T]) release(el T) {
+	icl.baseCapacityLimiter.release(uint64(el.ItemsCount()))
+}
+
+// nolint:unused
+func (icl *itemsCapacityLimiter[T]) sizeOf(el T) uint64 {
+	return uint64(el.ItemsCount())
+}
+
+var _ QueueCapacityLimiter[itemsCounter] = (*itemsCapacityLimiter[itemsCounter])(nil)
```
bogdandrutu, line 86: Don't need this since you have the func at line 67. That confirms that you can return `*itemsCapacityLimiter` as `QueueCapacityLimiter`

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+func (rcl *requestsCapacityLimiter[T]) release(_ T) {
+	rcl.baseCapacityLimiter.release(1)
+}
+
+// nolint:unused
+func (rcl *requestsCapacityLimiter[T]) sizeOf(_ T) uint64 {
+	return 1
+}
+
+var _ QueueCapacityLimiter[any] = (*requestsCapacityLimiter[any])(nil)
```
bogdandrutu, line 112: Same comment. Here is not needed, but in general we should start having this checks in the test files.

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+func (rcl *requestsCapacityLimiter[T]) claim(_ T) bool {
+	return rcl.baseCapacityLimiter.claim(1)
+}
+
+// nolint:unused
+func (rcl *requestsCapacityLimiter[T]) release(_ T) {
+	rcl.baseCapacityLimiter.release(1)
+}
+
+// nolint:unused
```
bogdandrutu, line 107: Why do you need this?

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+	return &requestsCapacityLimiter[T]{baseCapacityLimiter: newBaseCapacityLimiter[T](capacity)}
+}
+
+// nolint:unused
+func (rcl *requestsCapacityLimiter[T]) claim(_ T) bool {
+	return rcl.baseCapacityLimiter.claim(1)
+}
+
+// nolint:unused
+func (rcl *requestsCapacityLimiter[T]) release(_ T) {
```
bogdandrutu, line 103: And everywhere else:

```suggestion
func (rcl *requestsCapacityLimiter[T]) release(T) {
```

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+//nolint:unused
+func (bcl *baseCapacityLimiter[T]) claim(claimed uint64) bool {
+	if bcl.used.Load()+claimed > bcl.cap {
+		return false
+	}
+	bcl.used.Add(claimed)
+	return true
+}
+
+//nolint:unused
```
bogdandrutu, line 57: why?

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+func (bcl *baseCapacityLimiter[T]) release(capacity uint64) {
```
bogdandrutu, line 58: s/capacity/size

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+	return int(bcl.used.Load())
+}
+
+//nolint:unused
+func (bcl *baseCapacityLimiter[T]) claim(claimed uint64) bool {
+	if bcl.used.Load()+claimed > bcl.cap {
+		return false
+	}
+	bcl.used.Add(claimed)
+	return true
```
bogdandrutu, line 54: This can overflow. To correctly do this, use this code:
```suggestion
	if bcl.used.Add(claimed) > bcl.cap {
		bcl.used.Add(-claimed)
		return false
	}
	return true
```

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+func (bcl *baseCapacityLimiter[T]) Capacity() int {
+	return int(bcl.cap)
+}
+
+func (bcl *baseCapacityLimiter[T]) Size() int {
+	return int(bcl.used.Load())
+}
+
+//nolint:unused
+func (bcl *baseCapacityLimiter[T]) claim(claimed uint64) bool {
```
bogdandrutu, line 49: s/claimed/size

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,112 @@
...
+
+//nolint:unused
+func (bcl *baseCapacityLimiter[T]) release(capacity uint64) {
+	bcl.used.Add(^(capacity - 1))
+}
+
+// itemsCapacityLimiter is a capacity limiter that limits the queue based on the number of items (e.g. spans, log records).
+type itemsCapacityLimiter[T itemsCounter] struct {
+	baseCapacityLimiter[T]
+}
```
bogdandrutu, line 65: why not 
```suggestion
type itemsCapacityLimiter[T itemsCounter] baseCapacityLimiter[T]
```

exporter/exporterhelper/internal/unbounded_channel.go
dmitryax: What kind of simple unbounded channel are you referring to?

bogdandrutu: I had something else in my mind which does not work. What if we create a channel with same Capacity as the "capacity" of the limiter? Every element has at least size 1, so we should be good. Does that mean we are allocating too much memory? Can you check how much memory will allocate?

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: We won't get the exported `Size` and `Capacity` methods inherited from `baseCapacityLimiter` then

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: removed 👍

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: updated

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: done

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: Thanks. Updated

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: removed

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: Apparently, the linter doesn't recognize the unexported methods as used if they are referenced through an interface.

exporter/exporterhelper/internal/unbounded_channel.go
dmitryax: I put a draft with the bounded capacity https://github.com/open-telemetry/opentelemetry-collector/pull/9164 with some benchmarks:

Unbounded:
```
Benchmark_QueueUsage_1M_items_10_250k-10     	       9	 118011278 ns/op	 7497951 B/op	  250017 allocs/op
Benchmark_QueueUsage_1M_items_10_1M-10    	       3	 478627375 ns/op	29998074 B/op	 1000004 allocs/op
Benchmark_QueueUsage_100M_items_10_10M-10    	       1	4688934042 ns/op	299973192 B/op	 9999824 allocs/op
```

Bounded:
```
Benchmark_QueueUsage_1M_items_10_25k-10     	      25	  44122055 ns/op	24003696 B/op	      23 allocs/op
Benchmark_QueueUsage_1M_items_10_1M-10    	       7	 176613071 ns/op	24003396 B/op	      20 allocs/op
Benchmark_QueueUsage_100M_items_10_10M-10    	       1	1693857875 ns/op	2400016832 B/op	      62 allocs/op
```

B/op seems like cumulative. Not sure how to check the max RAM usage. But it is still enough to see that the bounded option takes 2Gb of memory for 100M of the item's capacity configuration value.

The unbounded has a lot of churn due to slice recreation on the overflow. That potentially can be fixed. We could handle the overflow manually and preallocate bigger arrays.

Also, I have an idea of how to integrate the capacity limiter right into the current unbounded channel implementation. It will remove the unnecessary decoupling between capacity limiting and the channel enqueuing. It should also make the code a bit simpler and maybe more performant. I can try that out.

exporter/exporterhelper/internal/queue_capacity.go
bogdandrutu: Are you sure?

bogdandrutu: Probably you know :D, I was just surprised...

exporter/exporterhelper/internal/queue_capacity.go
bogdandrutu: Please add a comment because most people will be shocked about this. Also would be nice if we determine the linter that fails and file a bug there.

dmitryax: I founded the reported issue and linked it in the comment https://github.com/dominikh/go-tools/issues/1440

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: That's why we had all the type aliases in `pdata` initially :) 

From https://tip.golang.org/ref/spec#Type_declarations: 
```
A defined type may have [methods](https://tip.golang.org/ref/spec#Method_declarations) associated with it. It does not inherit any methods bound to the given type
```

exporter/exporterhelper/internal/unbounded_channel.go
bogdandrutu: For simplicity I would go for the moment with a buffered channel. It has better performance and also seem to have significant less churn (allocations). Let's go with that for the moment.

dmitryax: Ok, I updated https://github.com/open-telemetry/opentelemetry-collector/pull/9164 and marked it as ready

bogdandrutu:
@dmitryax this is outdated? Can you re-open after https://github.com/open-telemetry/opentelemetry-collector/pull/9164 if still needed?

PR #9151, opened 2023-12-19
[pkg/sharedcomponent] new package pkg/sharedcomponent

### LINKED ISSUE #4266: Make SharedComponents public
It is currently internal and only used by OTLP receiver however there are other receivers that need the same functionality (any receiver that implements more than one data type needs this).

We have a duplicate in the contrib repo which is used by some contrib receivers, which is also internal. Any distro that implements a receiver like this is also forced to reimplement the SharedComponent.

I think we can simply make it public in the core and use everywhere (not sure if there is any other reason to keep it internal).

cc @bogdandrutu

atoulme:
**Description:**
Expose the internal package internal/sharedcomponent as pkg/sharedcomponent

**Link to tracking Issue:**
Fixes #4266

**Testing:**
N/A

**Documentation:**
N/A


bogdandrutu:
Before doing this, I would consider better to polish this code and move it to `component/componentshared`

bogdandrutu:
Another thing to look for is that we start first request and stop first request, maybe we should do ref-count and only stop after the same number of starts.

atoulme:
>Before doing this, I would consider better to polish this code and move it to component/componentshared or component/sharedcomponent.

Would it be its own module or part of the component module?

>But would like to do a pass on the API (consistency with other APIs that we have) and make sure is well tested and explained.

It probably can use some love, I'll take a look as well if that's ok.

> Another thing to look for is that we start first request and stop first request, maybe we should do ref-count and only stop after the same number of starts.

Sounds like a good idea to me!

bogdandrutu:
> Would it be its own module or part of the component module?

No.

> It probably can use some love, I'll take a look as well if that's ok.

Please.

> Sounds like a good idea to me!

Not sure if it is that easy.

atoulme:
> Not sure if it is that easy.

When did that ever stop me :D

atoulme:
Closing - let's take this one piecemeal, I'll file issues to follow up.

atoulme:
I have filed https://github.com/open-telemetry/opentelemetry-collector/issues/9156 and https://github.com/open-telemetry/opentelemetry-collector/issues/9155. We can keep https://github.com/open-telemetry/opentelemetry-collector/issues/4266 to move under component/componentshared.

PR #9157, opened 2023-12-20
[chore] polish sharedcomponent API

### LINKED ISSUE #9156: Review the internal/sharedcomponent API
Review the internal/sharedcomponent API and make sure it is compliant with coding best practices.

atoulme:
Attempt to simplify as much as possible the API exposed by the sharedcomponent package ahead of exposing it as part of a published module.

Relates to #9156 

Changes:
* Remove the map field in the struct, make the struct a map
* Remove the `NewSharedComponents` function, just initialize the map instead.
* Rename GetOrAdd to LoadOrStore

### DIFF
--- a/internal/sharedcomponent/sharedcomponent.go
+++ b/internal/sharedcomponent/sharedcomponent.go
@@ -9,7 +9,6 @@ package sharedcomponent // import "go.opentelemetry.io/collector/internal/shared
 import (
 	"context"
 	"sync"
-	"sync/atomic"
 
 	"go.opentelemetry.io/collector/component"
 )
@@ -51,7 +50,6 @@ func (scs Map[K, V]) LoadOrStore(key K, create func() (V, error), telemetrySetti
 		seenSettings: map[*component.TelemetrySettings]struct{}{
 			telemetrySettings: {},
 		},
-		activeCount: &atomic.Int32{},
 	}
 	scs[key] = newComp
 	return newComp, nil
@@ -62,10 +60,9 @@ func (scs Map[K, V]) LoadOrStore(key K, create func() (V, error), telemetrySetti
 type Component[V component.Component] struct {
 	component V
 
-	activeCount *atomic.Int32 // a counter keeping track of the number of active uses of the component
-	startOnce   sync.Once
-	stopOnce    sync.Once
-	removeFunc  func()
+	startOnce  sync.Once
+	stopOnce   sync.Once
+	removeFunc func()
 
 	telemetry    *component.TelemetrySettings
 	seenSettings map[*component.TelemetrySettings]struct{}
@@ -76,8 +73,7 @@ func (r *Component[V]) Unwrap() V {
 	return r.component
 }
 
-// Start starts the underlying component if it never started before. Each call to Start is counted as an active usage.
-// Shutdown will shut down the underlying component if called as many times as Start is called.
+// Start starts the underlying component if it never started before.
 func (r *Component[V]) Start(ctx context.Context, host component.Host) error {
 	var err error
 	r.startOnce.Do(func() {
@@ -90,17 +86,11 @@ func (r *Component[V]) Start(ctx context.Context, host component.Host) error {
 			_ = r.telemetry.ReportComponentStatus(component.NewPermanentErrorEvent(err))
 		}
 	})
-	r.activeCount.Add(1)
 	return err
 }
 
-// Shutdown shuts down the underlying component if all known usages, measured by the number of times
-// Start was called, are accounted for.
+// Shutdown shuts down the underlying component.
 func (r *Component[V]) Shutdown(ctx context.Context) error {
-	if r.activeCount.Add(-1) > 0 {
-		return nil
-	}
-
 	var err error
 	r.stopOnce.Do(func() {
 		// It's important that status for a shared component is reported through its
--- a/internal/sharedcomponent/sharedcomponent.go
+++ b/internal/sharedcomponent/sharedcomponent.go
@@ -13,24 +13,20 @@ import (
 	"go.opentelemetry.io/collector/component"
 )
 
-// Map keeps reference of all created instances for a given shared key such as a component configuration.
-type Map[K comparable, V component.Component] interface {
-	// LoadOrStore returns the already created instance if exists, otherwise creates a new instance
-	// and adds it to the map of references.
-	LoadOrStore(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*Component[V], error)
-}
-
-func NewMap[K comparable, V component.Component]() Map[K, V] {
-	return &mapImpl[K, V]{
+func NewMap[K comparable, V component.Component]() *Map[K, V] {
+	return &Map[K, V]{
 		components: map[K]*Component[V]{},
 	}
 }
 
-type mapImpl[K comparable, V component.Component] struct {
+// Map keeps reference of all created instances for a given shared key such as a component configuration.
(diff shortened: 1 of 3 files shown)


internal/sharedcomponent/sharedcomponent.go
```diff
@@ -53,25 +45,27 @@ func (scs *SharedComponents[K, V]) GetOrAdd(key K, create func() (V, error), tel
...
 	}
-	scs.comps[key] = newComp
+	scs[key] = newComp
 	return newComp, nil
 }
 
 // SharedComponent ensures that the wrapped component is started and stopped only once.
-// When stopped it is removed from the SharedComponents map.
+// When stopped it is removed from the Map.
 type SharedComponent[V component.Component] struct {
```
bogdandrutu, line 62: Just Component?

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -95,17 +90,23 @@ func (r *SharedComponent[V]) Start(ctx context.Context, host component.Host) err
...
-// Shutdown implements component.Component.
+// Shutdown shuts down the underlying component if all known usages, measured by the number of times
+// Start was called, are accounted for.
 func (r *SharedComponent[V]) Shutdown(ctx context.Context) error {
+	if r.activeCount.Add(-1) > 0 {
+		return nil
+	}
+
 	var err error
 	r.stopOnce.Do(func() {
```
bogdandrutu, line 105: No longer need this since you count the `activeCount`.

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -95,17 +90,23 @@ func (r *SharedComponent[V]) Start(ctx context.Context, host component.Host) err
 			_ = r.telemetry.ReportComponentStatus(component.NewPermanentErrorEvent(err))
 		}
 	})
+	r.activeCount.Add(1)
```
bogdandrutu, line 93: Implementing this is a problem, as I mentioned in the comment in the other PR. The problem is that we have a contract that says after calling shutdown on a component that component cannot emit to the telemetry pipeline anymore. This will not be the case for example for the otel receiver because if I have a Traces and Metrics receiver that will share this when I shutdown the first one it will still emit telemetry.

So I am not sure how we can do this, we may have to stick with the current implementation which is to shutdown on the first request.

internal/sharedcomponent/sharedcomponent.go
atoulme: yes less is more adopting it asap

internal/sharedcomponent/sharedcomponent.go
atoulme: understood ; I can add tests to check this contract is enforced well.

bogdandrutu: We have something https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/receivertest/contract_checker.go but does not check for the scenario I mentioned: Receiver supports multiple signal types, start an instance of multiple types on same port (like otlp receiver) then shutdowns one of them and expects to not add any data to the pipeline for that signal.

internal/sharedcomponent/sharedcomponent.go
atoulme: alright, done

internal/sharedcomponent/sharedcomponent.go
atoulme: I may have a way to go about this, but I'm going to break this changeset in 2, one is going to be just the API changes, and the other the changes in the lifecycle. I have too much going on in this PR.

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -12,23 +13,13 @@ import (
 	"go.opentelemetry.io/collector/component"
 )
 
-// SharedComponents a map that keeps reference of all created instances for a given configuration,
-// and ensures that the shared state is started and stopped only once.
-type SharedComponents[K comparable, V component.Component] struct {
-	comps map[K]*SharedComponent[V]
-}
+// Map keeps reference of all created instances for a given shared key such as a component configuration.
+type Map[K comparable, V component.Component] map[K]*Component[V]
```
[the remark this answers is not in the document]
atoulme: OK I can revert.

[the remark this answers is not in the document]
atoulme: By revert I mean, I'll add back a map field in the struct

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -12,23 +13,25 @@ import (
 	"go.opentelemetry.io/collector/component"
 )
 
-// SharedComponents a map that keeps reference of all created instances for a given configuration,
-// and ensures that the shared state is started and stopped only once.
-type SharedComponents[K comparable, V component.Component] struct {
-	comps map[K]*SharedComponent[V]
+// Map keeps reference of all created instances for a given shared key such as a component configuration.
+type Map[K comparable, V component.Component] interface {
```
bogdandrutu, line 17: Why interface? We don't expect others to implement or have alternative implementations.

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -12,23 +13,25 @@ import (
...
+	return &mapImpl[K, V]{
+		components: map[K]*Component[V]{},
 	}
 }
 
-// GetOrAdd returns the already created instance if exists, otherwise creates a new instance
-// and adds it to the map of references.
-func (scs *SharedComponents[K, V]) GetOrAdd(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*SharedComponent[V], error) {
-	if c, ok := scs.comps[key]; ok {
+type mapImpl[K comparable, V component.Component] struct {
```
bogdandrutu, line 29: Separate PR we may consider to do this map thread safe.

bogdandrutu (approved):

internal/sharedcomponent/sharedcomponent.go
atoulme: Going overboard, thanks for feedback, will remove.

internal/sharedcomponent/sharedcomponent.go
atoulme: Sure thing!

atoulme: #9170 is ready for after this is merged.

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -12,23 +13,13 @@ import (
...
-// GetOrAdd returns the already created instance if exists, otherwise creates a new instance
+// LoadOrStore returns the already created instance if exists, otherwise creates a new instance
 // and adds it to the map of references.
-func (scs *SharedComponents[K, V]) GetOrAdd(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*SharedComponent[V], error) {
-	if c, ok := scs.comps[key]; ok {
+func (m Map[K, V]) LoadOrStore(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*Component[V], error) {
+	if c, ok := m[key]; ok {
 		// If we haven't already seen this telemetry settings, this shared component represents
 		// another instance. Wrap ReportComponentStatus to report for all instances this shared
 		// component represents.
```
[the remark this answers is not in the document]
atoulme: There is a subtle bug here that we need to review. If more than one error is produced by the successive reports, we ignore them.

PR #9164, opened 2023-12-20
[chore] [exporterhelper] Items based queue sizing with bounded channel

dmitryax:
Introduce an option to limit the queue size by the number of items instead of number of requests. This is preliminary step for having the exporter helper v2 with a batcher sender placed after the queue sender. Otherwise, it'll be hard for the users to estimate the queue size based on the number of requests without batch processor in front of it.

This change doesn't effect the existing functionality and the items based queue limiting cannot be utilized yet.

Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/9147

### DIFF
--- a/exporter/exporterhelper/internal/queue_capacity.go
+++ b/exporter/exporterhelper/internal/queue_capacity.go
@@ -1,0 +1,74 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package internal // import "go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+
+import (
+	"sync/atomic"
+)
+
+type itemsCounter interface {
+	ItemsCount() int
+}
+
+// Sizer is an interface that returns the size of the given element.
+type Sizer[T any] interface {
+	SizeOf(T) uint64
+}
+
+// ItemsSizer is a Sizer implementation that returns the size of a queue element as the number of items it contains.
+type ItemsSizer[T itemsCounter] struct{}
+
+func (is *ItemsSizer[T]) SizeOf(el T) uint64 {
+	return uint64(el.ItemsCount())
+}
+
+// RequestSizer is a Sizer implementation that returns the size of a queue element as one request.
+type RequestSizer[T any] struct{}
+
+func (rs *RequestSizer[T]) SizeOf(T) uint64 {
+	return 1
+}
+
+type queueCapacityLimiter[T any] struct {
+	used *atomic.Uint64
+	cap  uint64
+	sz   Sizer[T]
+}
+
+func (bcl queueCapacityLimiter[T]) Capacity() int {
+	return int(bcl.cap)
+}
+
+func (bcl queueCapacityLimiter[T]) Size() int {
+	return int(bcl.used.Load())
+}
+
+func (bcl queueCapacityLimiter[T]) claim(el T) bool {
+	size := bcl.sizeOf(el)
+	if bcl.used.Add(size) > bcl.cap {
+		bcl.releaseSize(size)
+		return false
+	}
+	return true
+}
+
+func (bcl queueCapacityLimiter[T]) release(el T) {
+	bcl.releaseSize(bcl.sizeOf(el))
+}
+
+func (bcl queueCapacityLimiter[T]) releaseSize(size uint64) {
+	bcl.used.Add(^(size - 1))
+}
+
+func (bcl queueCapacityLimiter[T]) sizeOf(el T) uint64 {
+	return bcl.sz.SizeOf(el)
+}
+
+func newQueueCapacityLimiter[T any](sizer Sizer[T], capacity int) *queueCapacityLimiter[T] {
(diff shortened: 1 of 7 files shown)


exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"sync/atomic"
+)
+
+type itemsCounter interface {
+	ItemsCount() int
+}
+
+// QueueCapacityLimiter is an interface to control the capacity of the queue.
+type QueueCapacityLimiter[T any] interface {
```
bogdandrutu, line 15: Instead of this complicated interface, what if we have only a "Sizer":

```golang

// Can even be just a typedef of the func.
type Sizer[T any interface {
	sizeof(el T) uint64
}

type baseCapacityLimiter[T any] struct {
	used *atomic.Uint64
	cap  uint64
	sizer Sizer[T]
}

func newBaseCapacityLimiter[T any](capacity int, sizer Sizer[T]) baseCapacityLimiter[T] {
	return baseCapacityLimiter[T]{
		used: &atomic.Uint64{},
		cap:  uint64(capacity),
		sizer: sizer,
	}
}

func (bcl *baseCapacityLimiter[T]) Capacity() int {
	return int(bcl.cap)
}

func (bcl *baseCapacityLimiter[T]) Size() int {
	return int(bcl.used.Load())
}

// nolint:unused // false positive https://github.com/dominikh/go-tools/issues/1440
func (bcl *baseCapacityLimiter[T]) claim(el T) bool {
	size := sizer.sizeof(el)
	if bcl.used.Add(size) > bcl.cap {
		bcl.used.Add(^(size - 1))
		return false
	}
	return true
}

// nolint:unused // false positive https://github.com/dominikh/go-tools/issues/1440
func (bcl *baseCapacityLimiter[T]) release(el T) {
	size := sizer.sizeof(el)
	bcl.used.Add(^(size - 1))
}

type requestsSizer[T any] struct {}

func NewRequestsCapacityLimiter[T any](capacity int) QueueCapacityLimiter[T] {
	return &requestsSizer[T]{}
}

// nolint:unused // false positive https://github.com/dominikh/go-tools/issues/1440
func (rcl *requestsSizer[T]) sizeOf(T) uint64 {
	return 1
}

```

dmitryax: Thanks for the suggestion. It helped to get rid of the interface and `nolint` comments

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type itemsCounter interface {
+	ItemsCount() int
+}
+
+// sizeOfFunc is a function that returns the size of the given element.
+type sizeOfFunc[T any] func(T) uint64
+
+type QueueCapacityLimiter[T any] struct {
```
bogdandrutu, line 17: Instead of exposing this, can we just pass a capacity and a sizer to the queue constructor?

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (bcl QueueCapacityLimiter[T]) sizeOf(el T) uint64 {
+	if bcl.sz == nil {
+		return 1
+	}
+	return bcl.sz(el)
+}
+
+// NewItemsCapacityLimiter returns a capacity limiter that limits the queue based on the number of items in requests.
+func NewItemsCapacityLimiter[T itemsCounter](capacity int) QueueCapacityLimiter[T] {
```
bogdandrutu, line 59: Return pointer, and make also all funcs work on pointer, no need to copy 24B all the time.

bogdandrutu (approved):

exporter/exporterhelper/internal/queue_capacity.go
```diff
@@ -0,0 +1,73 @@
...
+func (bcl QueueCapacityLimiter[T]) release(el T) {
+	bcl.releaseSize(bcl.sizeOf(el))
+}
+
+func (bcl QueueCapacityLimiter[T]) releaseSize(size uint64) {
+	bcl.used.Add(^(size - 1))
+}
+
+func (bcl QueueCapacityLimiter[T]) sizeOf(el T) uint64 {
+	if bcl.sz == nil {
```
bogdandrutu, line 52: Don't use nil for request base, force to always provide a func.

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: done

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: done

exporter/exporterhelper/internal/queue_capacity.go
dmitryax: done

PR #9168, opened 2023-12-21
[configtls] support setting cipher suites

### LINKED ISSUE #8105: Enable ability to set cipher suite for OTLP receiver when running TLS
I would like the ability to specify which ciphers receivers (OTLP in this instance) can accept for configuring TLS. It's helpful that we can configure TLSv1.2 and TLSv1.3, but would be great if we can then limit what ciphers users can use when connecting to the receiver

Unless I'm mistaken, I don't see in the TLS config code for the OTEL collector the ability to specify TLS ciphers

atoulme:
**Description:**
Add `cipher_suites` to configtls:
Users can specify a list of cipher suites to pick from. If left blank, a safe default list is used.

**Link to tracking Issue:**
Fixes #8105

**Testing:**
Unit tests

**Documentation:**
godoc and README

### DIFF
--- a/config/configtls/configtls.go
+++ b/config/configtls/configtls.go
@@ -54,6 +54,11 @@ type TLSSetting struct {
 	// If not set, refer to crypto/tls for defaults. (optional)
 	MaxVersion string `mapstructure:"max_version"`
 
+	// CipherSuites is a list of TLS cipher suites that the TLS transport can use.
+	// If left blank, a safe default list is used.
+	// See https://go.dev/src/crypto/tls/cipher_suites.go for a list of supported cipher suites.
+	CipherSuites []string `mapstructure:"cipher_suites"`
+
 	// ReloadInterval specifies the duration after which the certificate will be reloaded
 	// If not set, it will never be reloaded (optional)
 	ReloadInterval time.Duration `mapstructure:"reload_interval"`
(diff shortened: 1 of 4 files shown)


dmitryax (approved):

config/configtls/configtls.go
```diff
@@ -175,16 +180,39 @@ func (c TLSSetting) loadTLSConfig() (*tls.Config, error) {
...
+		found := false
+		for _, supported := range tls.CipherSuites() {
+			if suite == supported.Name {
+				result = append(result, supported.ID)
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("invalid TLS cipher suite: %q", suite)
```
codeboten, line 210: could this join errors instead of failing on the first one? this would be helpful in the case where there are multiple invalid cipher suites specified

codeboten (approved):
One non-blocking comment

config/configtls/configtls.go
atoulme: Sounds good, added code towards supporting this.

config/configtls/configtls.go
```diff
@@ -175,16 +181,43 @@ func (c TLSSetting) loadTLSConfig() (*tls.Config, error) {
...
+			}
+		}
+		if !found {
+			errs = append(errs, fmt.Errorf("invalid TLS cipher suite: %q", suite))
+		}
+	}
+	if len(errs) != 0 {
+		return nil, errors.Join(errs...)
+	}
+	return result, nil
```
codeboten, line 218:
```suggestion
	return result, errors.Join(errs...)
```

codeboten (commented):
Thanks @atoulme!

PR #9169, opened 2023-12-21
[confmap] Return error when decoding negative values into uints

### LINKED ISSUE #9060: [processor/memory_limiter] Negative limit values are not rejected
**Describe the bug**
Negative limit values in the configuration are accepted and converted into `0xffffffff - val`

**Steps to reproduce**
Use config 
```
  memory_limiter:
    limit_mib: -1
    spike_limit_mib: -2
    check_interval: 5s
```

Collector refuse to start with an error message.

**What did you see instead?**
Collector is started with huge limits
```
2023-12-10T17:17:16.225-0800	info	service@v0.90.1/telemetry.go:86	Setting up own telemetry...
2023-12-10T17:17:16.225-0800	info	service@v0.90.1/telemetry.go:203	Serving Prometheus metrics	{"address": ":8888", "level": "Basic"}
2023-12-10T17:17:16.225-0800	info	exporter@v0.90.1/exporter.go:275	Development component. May change in the future.	{"kind": "exporter", "data_type": "metrics", "name": "debug"}
2023-12-10T17:17:16.226-0800	info	memorylimiterprocessor@v0.90.1/memorylimiter.go:102	Memory limiter configured	{"kind": "processor", "name": "memory_limiter", "pipeline": "metrics", "limit_mib": 4294967295, "spike_limit_mib": 4294967294, "check_interval": 5}
```

**What version did you use?**
Version: `v0.91.1`

crobert-1:
**Description:**
This adds a decode hook for unmarshalling negative integers into uint types. This will now return an error instead of converting negative values into large uint values.

Fixes #9060 

**Testing:**
Added unit tests for confmap functionality, functional tests in memory limiter processor (the original component this issue was filed against)


confmap/confmap.go
```diff
@@ -368,3 +369,15 @@ func zeroSliceHookFunc() mapstructure.DecodeHookFuncValue {
...
 	}
 }
+
+// This hook is used to solve the issue: https://github.com/open-telemetry/opentelemetry-collector/issues/9060
+// Decoding should fail when converting a negative integer to any type of unsigned integer. This prevents
+// negative values being decoded as large uint values.
+func negativeUintHookFunc() mapstructure.DecodeHookFuncValue {
+	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
+		if from.CanInt() && from.Int() < 0 && to.CanUint() {
+			return nil, fmt.Errorf("cannot convert negative value %v to uint", from.Int())
```
crobert-1, line 379: `uint` here is technically any unsigned integer type, so it could be `uint`, `uint8`, etc. I assume using `uint` as the general term here in an error message is clear enough to users, but we could re-word to something like `an unsigned integer` to be more generic.

confmap/confmap_test.go
```diff
@@ -212,6 +213,58 @@ func TestMapKeyStringToMapKeyTextUnmarshalerHookFunc(t *testing.T) {
...
+
+			assert.NoError(t, err)
+			assert.Equal(t, cfg.UintTest, uint32(testValue))
+		})
+	}
+
+}
+
+func TestUintUnmarshalerFailure(t *testing.T) {
+	testValue := -1000
```
bogdandrutu, line 256: Can you also have a test with `uint64(-1000)` which should correctly parse to uint64

crobert-1: Added, but let me know if I misunderstood your request 👍

bogdandrutu: Looking again at the tests, I feel that they are "unreal". The common case is not to parse it from a `map[string]any{"uint_test": tt.testValue}` but actually from a `yaml` file. The reason I am asking is how do you get here in the first place to have "-1000" for example? Is it parsign `"uint_test": 18446744073709550616` or `"uint_test": -1000` since the int/uint representation will get interesting here (`18446744073709550616` is `uint64(-1000)`)

crobert-1: This specific added test case (`TestUint64Unmarshaler`) is not super valuable, as we're using `go` itself to do the conversion of negative to positive. This test is only testing to ensure a `uint64`. Not super relevant to this PR, but it's at least a valid test and good to have.

I think the other tests (`TestUintUnmarshalerSuccess`, `TestUintUnmarshalerFailure`) are much more valuable, as it goes through much of the same code path as the yaml unmarshal logic. The YAML gets parsed into a golang map [here.](https://github.com/open-telemetry/opentelemetry-collector/blob/bb76d9cf45839b2f34bc88e033e0533e2ff92b29/otelcol/internal/configunmarshaler/configs.go#L26) There is no type asserted in this conversion, the purpose is simply to get it into a workable format. [Later](https://github.com/open-telemetry/opentelemetry-collector/blob/bb76d9cf45839b2f34bc88e033e0533e2ff92b29/otelcol/internal/configunmarshaler/configs.go#L45) in the same method the map is converted to the component's configuration options, where type is enforced. The call stack goes down to the [central `Unmarshal` method](https://github.com/open-telemetry/opentelemetry-collector/blob/bb76d9cf45839b2f34bc88e033e0533e2ff92b29/confmap/confmap.go#L84), which handles the typed conversion to get the component's final configuration.  

This `Unmarshal` method is the same method we're calling in tests, so whether the collector is parsing YAML from a config file, or a unit test defines a configuration and passes in a map, the same code is being executed, with the same input arguments and types. From this I believe there are valid and helpful tests.

crobert-1:
I believe this is ready to be merged.

@bogdandrutu Do you have any concerns here?

confmap/confmap.go
TylerHelmuth: I prefer `an unsigned integer`.

crobert-1: Done 👍

TylerHelmuth (approved):

songy23 (approved):

evan-bradley (approved):

mx-psi (commented):
This comes from #9532. Can we add a note to remove the hook once we disable `WeaklyTypedInput`?

crobert-1:
> This comes from #9532. Can we add a note to remove the hook once we disable `WeaklyTypedInput`?

Done. I've added a TODO referencing the issue. 👍

mx-psi (approved):

PR #9170, opened 2023-12-21
[chore] make sharedcomponent Map threadsafe

atoulme:
**Description:**
Add just enough code to make sharedcomponent Map thread safe.

**Link to tracking Issue:**
Relates to #9156

Follow up to #9157, should be reviewed after it is merged.

### DIFF
--- a/internal/sharedcomponent/sharedcomponent.go
+++ b/internal/sharedcomponent/sharedcomponent.go
@@ -22,11 +22,14 @@ func NewMap[K comparable, V component.Component]() *Map[K, V] {
 // Map keeps reference of all created instances for a given shared key such as a component configuration.
 type Map[K comparable, V component.Component] struct {
 	components map[K]*Component[V]
+	lock       sync.Mutex
 }
 
 // LoadOrStore returns the already created instance if exists, otherwise creates a new instance
 // and adds it to the map of references.
 func (m *Map[K, V]) LoadOrStore(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*Component[V], error) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
 	if c, ok := m.components[key]; ok {
 		// If we haven't already seen this telemetry settings, this shared component represents
 		// another instance. Wrap ReportComponentStatus to report for all instances this shared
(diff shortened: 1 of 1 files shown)


bogdandrutu:
Please rebase

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -22,11 +22,14 @@ func NewMap[K comparable, V component.Component]() *Map[K, V] {
...
 }
 
 // LoadOrStore returns the already created instance if exists, otherwise creates a new instance
 // and adds it to the map of references.
 func (m *Map[K, V]) LoadOrStore(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*Component[V], error) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
 	if c, ok := m.components[key]; ok {
 		// If we haven't already seen this telemetry settings, this shared component represents
 		// another instance. Wrap ReportComponentStatus to report for all instances this shared
```
sfc-gh-bdrutu, line 35: This still makes me worry. I don't understand the need, if we already report in graph. @mwear can you confirm?

mwear: There are two issues with shared component and status reporting. We automate status reporting from graph for start and shutdown; components report runtime status after start and before shutdown. The lines you point out are in reference to runtime status (reported from within the component). The shared component represents multiple component instances and needs to report status for each instance. We wrap ReportComponentStatus on TelemetrySettings so that each call reports the same status for each of the "instances" it represents.

At start time, the issue is that graph sees the shared component as multiple instances and starts components sequentially. When reporting from outside the component, it is not using the wrapped ReportComponentStatus method, so it reports status for each component instance, one at a time. This causes a race for the shared component instances, because a status report can come from the component at any time after start is called. However, the other effective instances have not been started yet, which can result in invalid state transitions, due to the underlying finite state machine. A component must be `Starting` before it can transition to `OK` or `RecoverableError`. So, a shared component and it's instances need to be set to `Starting` at the same time so that subsequent transitions will be valid.

atoulme:
Anything else needed for this enhancement?

dmitryax (approved):

mwear (approved):

bogdandrutu (approved):

PR #9171, opened 2023-12-21
[chore] when adding new module, use the version from main

codeboten:
This will make updating core from contrib much easier, otherwise we need to add replace statements everywhere in contrib.

### DIFF
--- a/cmd/otelcorecol/go.mod
+++ b/cmd/otelcorecol/go.mod
@@ -84,7 +84,7 @@ require (
 	go.opentelemetry.io/collector/config/confighttp v0.91.0 // indirect
 	go.opentelemetry.io/collector/config/confignet v0.91.0 // indirect
 	go.opentelemetry.io/collector/config/configopaque v0.91.0 // indirect
-	go.opentelemetry.io/collector/config/configretry v0.91.0 // indirect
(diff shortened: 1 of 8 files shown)


cmd/otelcorecol/go.mod
```diff
@@ -84,7 +84,7 @@ require (
 	go.opentelemetry.io/collector/config/confighttp v0.91.0 // indirect
 	go.opentelemetry.io/collector/config/confignet v0.91.0 // indirect
 	go.opentelemetry.io/collector/config/configopaque v0.91.0 // indirect
-	go.opentelemetry.io/collector/config/configretry v0.91.0 // indirect
+	go.opentelemetry.io/collector/config/configretry v0.0.0-20231221085427-9027a8d9cc3f // indirect
```
bogdandrutu, line 87: You cannot do this when you add the module. You need a second PR for doing this.

bogdandrutu (approved):

cmd/otelcorecol/go.mod
codeboten: yup... that's right, it needs to be a follow up. i usually leave the module version a v0.0.0 to remember 😆

PR #9172, opened 2023-12-21
[mdatagen] copy files from contrib to core

codeboten:
This is to allow us to use it in core as well


bogdandrutu (approved):

dmitryax (approved):

mx-psi (commented):
Should this be at https://github.com/open-telemetry/opentelemetry-go-build-tools instead?

songy23 (approved):
> Should this be at https://github.com/open-telemetry/opentelemetry-go-build-tools instead?

Seems no because of dependency issue. (mdatagen depends on core while core depends on build-tools)

mx-psi (approved):
Let's leave it here then :) AFAICT the dependency is just on the autogenerated code used for testing, which we could split off into a separate module to avoid a circular module dependency, but no need to do this now

PR #9173, opened 2023-12-21
[chore] Enable goleak for all (passing) tests

crobert-1:
**Description:**
It was suggested that we add the [goleak package](https://pkg.go.dev/go.uber.org/goleak) to every test in core. This change adds `goleak` to every package that is succeeds with goleak. There a number that are not successful, the full list of which I've posted in the bug.

I generated these files using a shell script to copy a template `main_test.go` file into each package, then modified the package name. I'm sure there was a better way to automate this, but it worked well enough at this point. Here's the script:
 ```
cp ./main_test.go $1

PACKAGE_NAME=$(basename $1)

sed -i '' -e "s|package component|package $PACKAGE_NAME|g" $1/main_test.go

pushd . && cd $1 && go mod tidy 
go test -v .
```
Usage example:
```
$ ./add_leak_test reciever/scrapererror
```

**Link to tracking Issue:**
#9165

**Testing:**
All added tests are passing, but there are a number failing. Note that there's no sign of `goleak` running until it fails.

### CODE REVIEW

File: client/main_test.go
Comment (crobert-1) [inline client/main_test.go:1]: I chose to add (or modify in the case of `cmd/builder/internal/builder/main_test.go`) a new file called `main_test.go` to each package as I thought that was the most clear and uniform way to accomplish adding the central `goleak` test method to each package. The alternative is a different file name (happy to change if something else would be more clear), or add the test to existing test. To me, adding to existing test files introduced inconsistencies (which test file in each package should it be added to?) as well as making it harder to find and harder to know whether `goleak` is actually being used for the given package. Feedback is welcome though if others disagree with this approach.

File: client/main_test.go [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
crobert-1, line 14: This is taken from `goleak` [documentation](https://pkg.go.dev/go.uber.org/goleak#readme-quick-start) on how to test for leaks in an entire package, instead of adding a check to every single test. 

This will technically lead to less clear error messages where someone will have to figure out which test is causing the leak in a given package, but I think that's the much better option over adding a check to every single test in the repo.

atoulme: I think you can rename from `main_test.go` to `goleak_test.go` and `TestMain` to `TestGoLeaks` to explain this testing a bit more.

yurishkuro: the function name must be TestMain, is the standard name understood by go test toolchain. Putting VerifyTestMain in there makes it apply to all tests in the package, which won't happen if you name it  TestGoLeaks.

fwit in Jaeger I recommended naming the files `package_test.go`, but only adding them to packages with multiple test files - if there's only one test file then TestMain can go directly into it.

yurishkuro: Another thing we did in Jaeger is add a linter script that flags all packages that are missing goleak. Right now it's info-only, but once we fix all packages (<30 remaining) we will make it fail the CI.

https://github.com/jaegertracing/jaeger/issues/5006 for reference

atoulme: Thanks I learned something!

bogdandrutu (approved):

bogdandrutu:
Please rebase

client/main_test.go
bogdandrutu: I like @yurishkuro suggestion for `package_test.go`. I would still add always this file even only one test, just to remove the possible problem if we add a second test file in the future.

crobert-1: I've renamed to `package_test.go` 👍

PR #9175, opened 2023-12-22
Move error out of `ReportComponentStatus` function signature, use `ReportStatus` instead

### LINKED ISSUE #9148: `TelemetrySettings.ReportComponentStatus(...) error`: unsure how to handle returned error
We are using new component status reporting capabilities that replace `host.ReportFatalError`.

The new method to report component status has a signature that returns an error:
 `TelemetrySettings.ReportComponentStatus(...) error`

How should we handle the error being returned? It is ignored in most places right now.

atoulme:
Fixes #9148

### DIFF
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -37,13 +37,6 @@ type TelemetrySettings struct {
 
 	// ReportComponentStatus allows a component to report runtime changes in status. The service
 	// will automatically report status for a component during startup and shutdown. Components can
-	// use this method to report status after start and before shutdown. ReportComponentStatus
-	// will only return errors if the API used incorrectly. The two scenarios where an error will
-	// be returned are:
-	//
-	//   - An illegal state transition
-	//   - Calling this method before component startup
-	//
-	// If the API is being used properly, these errors are safe to ignore.
-	ReportComponentStatus func(*StatusEvent) error
+	// use this method to report status after start and before shutdown.
+	ReportComponentStatus func(*StatusEvent)
 }
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -38,5 +38,12 @@ type TelemetrySettings struct {
 	// ReportComponentStatus allows a component to report runtime changes in status. The service
 	// will automatically report status for a component during startup and shutdown. Components can
 	// use this method to report status after start and before shutdown.
-	ReportComponentStatus func(*StatusEvent)
+	// Deprecated: [v0.92.0] This function will be removed in a future release.
+	// Use ReportStatus instead.
+	ReportComponentStatus func(*StatusEvent) error
+
+	// ReportStatus allows a component to report runtime changes in status. The service
+	// will automatically report status for a component during startup and shutdown. Components can
+	// use this method to report status after start and before shutdown.
+	ReportStatus func(*StatusEvent)
 }
--- a/service/internal/servicetelemetry/telemetry_settings.go
+++ b/service/internal/servicetelemetry/telemetry_settings.go
@@ -43,7 +43,7 @@ type TelemetrySettings struct {
 // ToComponentTelemetrySettings returns a TelemetrySettings for a specific component derived from
 // this service level Settings object.
 func (s TelemetrySettings) ToComponentTelemetrySettings(id *component.InstanceID) component.TelemetrySettings {
-	status := status.NewComponentStatusFunc(id, s.Status.ReportStatus)
+	statusFunc := status.NewReportStatusFunc(id, s.Status.ReportStatus)
 	return component.TelemetrySettings{
 		Logger:         s.Logger,
 		TracerProvider: s.TracerProvider,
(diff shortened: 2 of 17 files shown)


service/service.go
```diff
@@ -113,7 +113,9 @@ func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
 		MetricsLevel:   cfg.Telemetry.Metrics.Level,
 		// Construct telemetry attributes from build info and config's resource attributes.
 		Resource: pcommonRes,
-		Status:   status.NewReporter(srv.host.notifyComponentStatusChange),
+		Status: status.NewReporter(srv.host.notifyComponentStatusChange, func(err error) {
+			srv.telemetry.Logger().Warn("Invalid transition", zap.Error(err))
```
mwear, line 117: I think `debug` is probably the proper log level. Failed state transitions are expected behavior when using the `ReportComponentStatus` API.

atoulme: Sure thing. Is this approach valid overall?

mwear: It's an approach worth considering. My only reservation is that logging about invalid state transitions might give the user the impression that something is wrong with status reporting, when the invalid transitions are largely expected. I do think we want to give some feedback if the API is used before start however.

atoulme: ok that's doable by using errors.Is, just need to make sure the error is exported. I'll check.

service/service.go
```diff
@@ -113,7 +114,13 @@ func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
 		MetricsLevel:   cfg.Telemetry.Metrics.Level,
 		// Construct telemetry attributes from build info and config's resource attributes.
 		Resource: pcommonRes,
-		Status:   status.NewReporter(srv.host.notifyComponentStatusChange),
+		Status: status.NewReporter(srv.host.notifyComponentStatusChange, func(err error) {
+			if errors.Is(err, status.ErrStatusNotReady) {
+				srv.telemetry.Logger().Warn("Invalid transition", zap.Error(err))
+			} else {
+				srv.telemetry.Logger().Debug("Invalid transition", zap.Error(err))
```
mwear, line 121: I'm wondering if we might omit logging about invalid transitions altogether since they are expected to happen and do not indicate an underlying problem using the API.

atoulme: sure. I can do that.

Now the bigger problem is how to deprecate this method to remove the error returned - I'll work on that next.

bogdandrutu (commented):
This looks good to me

service/internal/servicetelemetry/telemetry_settings.go
```diff
@@ -43,12 +43,17 @@ type TelemetrySettings struct {
 // ToComponentTelemetrySettings returns a TelemetrySettings for a specific component derived from
 // this service level Settings object.
 func (s TelemetrySettings) ToComponentTelemetrySettings(id *component.InstanceID) component.TelemetrySettings {
+	status := status.NewComponentStatusFunc(id, s.Status.ReportStatus)
```
mwear, line 46: Super minor nit, can we call this `statusFunc`?

mwear (approved):
Looks great. I like the more concise method name -- `ReportStatus`. Thanks for cleaning this up!

service/internal/servicetelemetry/telemetry_settings.go
atoulme: Sure!

atoulme:
Thanks I will follow up in a few days!

service/internal/servicetelemetry/telemetry_settings.go
atoulme: done

component/telemetry.go
```diff
@@ -37,13 +37,13 @@ type TelemetrySettings struct {
...
-	// If the API is being used properly, these errors are safe to ignore.
+	// use this method to report status after start and before shutdown.
+	// Deprecated: [v0.92.0] This function will be removed in a future release.
+	// Use ReportStatus instead.
 	ReportComponentStatus func(*StatusEvent) error
+
+	// ReportStatus allows a component to report runtime changes in status. The service
+	// will automatically report status for a component during startup and shutdown. Components can
+	// use this method to report status after start and before shutdown.
+	ReportStatus func(*StatusEvent)
```
bogdandrutu, line 48: What about we use interface instead, something like:

```golang
type StatusReporter interface {
   Report(*StatusEvent)
}

type TelemetrySettings struct {
  ....
  StatusReporter StatusReporter // I am not sure about the name of the variable, I like the interface name
}
```

This allows us to add more funcs to this interface in the future. For that we can even use optional interfaces that the StatusReporters can optionally implement, or we add an un-exported func and we have a "NopStatusReporters" that everyone must embed to implement the interface.

atoulme: trying it

atoulme: a bit more research needed, especially as we can use this interface to resolve this comment:
https://github.com/open-telemetry/opentelemetry-collector/blob/b3d04f58abda0f0550b2bfa547390ee2ce150be8/component/telemetry.go#L17-L19

mwear: I don't anticipate needing additional status reporting methods for components. If we have some examples, I'd be open to considering it. As this is a method on `component.TelemetrySettings`, to be used by components, is there any value in having a optional interface? FWIW, the service does need multiple methods, and when I made that change, I intentionally did not expose the additional method on `component.TelemetrySettings`. The other difference with the `servicetelemetry.TelemetrySettings` version is that it takes the `component.InstanceID` as a parameter, whereas the component version is bound to an `InstanceID` for a specific component via a closure.

atoulme:
Is there anything left to do?

bogdandrutu (approved):

Issue #9165, opened 2023-12-20
Consider enforcing https://github.com/uber-go/goleak in all tests


dmitryax:
This will require fixing all existing leaks first.

Suggested in https://github.com/open-telemetry/opentelemetry-collector/pull/9100#issuecomment-1856422113

crobert-1:
I've submitted a PR that adds `goleak` for all packages in which it's currently passing. Here's the complete list (that I'm aware of) of failing packages:
```
cmd/otelcorecol/
config/configgrpc/
config/configtls/
exporter/debugexporter/
exporter/exporterhelper/
exporter/exportertest/
exporter/internal/common/
exporter/loggingexporter/
exporter/otlpexporter/
exporter/otlphttpexporter/
internal/e2e/
internal/obsreportconfig/
obsreport/obsreporttest/
otelcol/
otelcol/otelcoltest/
processor/batchprocessor/
processor/memorylimiterprocessor/
processor/processorhelper/
receiver/otlpreceiver/internal/logs/
receiver/otlpreceiver/internal/metrics/
receiver/otlpreceiver/internal/trace/
receiver/otlpreceiver/
receiver/receiverhelper/
receiver/scraperhelper/
service/internal/proctelemetry/
service/
```

crobert-1:
It looks like a pretty common cause of `goleak` failures is the `go.opencensus.io/stats/views` package, specifically the `start` [method](https://github.com/census-instrumentation/opencensus-go/blob/01e6da5fc01c42aca1e0ce315f41744876e9fcbb/stats/view/worker.go#L287). This is called by `init()`, so it's running in the background for any collector package that's importing the `go.opencensus.io` package, even indirectly.

```
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 21 in state select, with go.opencensus.io/stats/view.(*worker).start on top of the stack:
go.opencensus.io/stats/view.(*worker).start(0xc0001a7b80)
	/Users/crobert/go/pkg/mod/go.opencensus.io@v0.24.0/stats/view/worker.go:292 +0x9f
created by go.opencensus.io/stats/view.init.0 in goroutine 1
	/Users/crobert/go/pkg/mod/go.opencensus.io@v0.24.0/stats/view/worker.go:34 +0x8d
```

_I'm still investigating to better understand the impact and possible solution for this one cause of failures._

yurishkuro:
[...]

PR #9210, opened 2024-01-03
[chore] Enable goleak for tests failing on opencensus-go

crobert-1:
**Description:**
As described [here](https://github.com/open-telemetry/opentelemetry-collector/issues/9165#issuecomment-1874732291) and [here](https://github.com/open-telemetry/opentelemetry-collector/issues/9165#issuecomment-1874836336), there are quite a few packages that directly or indirectly import the [OpenCensus-go stats package](https://pkg.go.dev/go.opencensus.io/stats). Within the [stats/view package](https://pkg.go.dev/go.opencensus.io@v0.24.0/stats/view) there's an init() function that starts goroutines that aren't cleaned up. This causes goleak to fail. 

Since the failure is caused by OpenCensus always indirectly (from what I can tell), and OpenCensus is now read-only, I believe it's fair to ignore leaks caused by this dependency and specific leaking situation.

#9165

**Testing:**
All added tests are passing


internal/testutil/testutil.go
```diff
@@ -116,3 +118,10 @@ func createExclusionsList(exclusionsText string, t testing.TB) []portpair {
 	}
 	return exclusions
 }
+
+// This prevent catching the leak generated by opencensus defaultWorker.Start which at
+// this time is part of the package's init call.
+// See https://github.com/open-telemetry/opentelemetry-collector/issues/9165#issuecomment-1874836336 for more context.
+func IgnoreOpenCensusWorkerLeak() goleak.Option {
```
bogdandrutu, line 125: Not worth all the complex circular dependencies for a 1 line func. Would suggest to inline the function.

bogdandrutu:
Please rebase

bogdandrutu (approved):

PR #9214, opened 2024-01-03
[configopaque] configopaque.String implements fmt.Stringer

### LINKED ISSUE #9213: `configopaque.String` should implement `fmt.Stringer`
a configopaque.String will by default be replaced with `[Redacted]`, is there a reason to override the behaviour here?

_Originally posted by @codeboten in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/29446#discussion_r1406497947_

atoulme:
**Description:**
configopaque.String implements `fmt.Stringer`, outputting [REDACTED] when used as `fmt.Sprintf("%s", opaquestring)`

**Link to tracking Issue:**
Fixes #9213


config/configopaque/opaque.go
```diff
@@ -18,3 +18,7 @@ var _ encoding.TextMarshaler = String("")
 func (s String) MarshalText() ([]byte, error) {
 	return []byte(maskedString), nil
 }
+
+func (s String) String() string {
```
mx-psi, line 22: nit: Add

```
var _ fmt.Stringer = String("")
```

at the top of the file

mx-psi (commented):
I am slightly worried that we will break something in contrib with this if someone is relying on `fmt.Sprintf` having the current behavior. Can you think of any way to check for this?

atoulme:
> I am slightly worried that we will break something in contrib with this if someone is relying on `fmt.Sprintf` having the current behavior. Can you think of any way to check for this?

Contrib tests pass? :)

I ran:
```
$> git grep fmt.Sprintf | grep -v test.go | grep -i  -E '(token|password)'
```

I received:
| Match | Impact |
| ------------- | ------------- |
| exporter/dynatraceexporter/config/config.go:            c.HTTPClientSettings.Headers["Authorization"] = configopaque.String(fmt.Sprintf("Api-Token %s", c.APIToken)) | `APIToken` is a string. The configopaque.String constructed might be a problem. The exporter is deprecated and should be removed now that we're in 2024. |
| exporter/logzioexporter/factory.go:     defaultURL := fmt.Sprintf("%s/?token=%s", getListenerURL(""), cfg.Token) | `Token` is a `configopaque.String`, should have impact |
| exporter/logzioexporter/factory.go:             return fmt.Sprintf("%s/?token=%s", getListenerURL(cfg.Region), cfg.Token), nil | same |
| extension/bearertokenauthextension/bearertokenauth.go:  token := fmt.Sprintf("%s %s", b.scheme, b.tokenString) | `tokenString` is a string. |
| extension/bearertokenauthextension/bearertokenauth.go:          expect = fmt.Sprintf("%s %s", b.scheme, expect) | false match |
| internal/metadataproviders/openshift/metadata.go:       req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", o.token)) | `token` is a string |
| internal/metadataproviders/openshift/metadata.go:       req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", o.token)) | `token` is a string |
| pkg/stanza/fileconsumer/internal/emittest/sink.go:              assert.Fail(t, fmt.Sprintf("Timed out waiting for token: %s", expected)) | false match |
| pkg/stanza/fileconsumer/internal/emittest/sink.go:              assert.Fail(t, fmt.Sprintf("Timed out waiting for token: %s", expected)) | false match |
| receiver/cloudfoundryreceiver/uaa.go:   logger.Debug(fmt.Sprintf("creating new cloud foundry UAA token client with url %s username %s", config.Endpoint, username)) | false match |
| receiver/cloudfoundryreceiver/uaa.go:           utp.logger.Debug(fmt.Sprintf("received new cloud foundry UAA token which expires in %d seconds", expiresInSeconds)) | false match |
| receiver/elasticsearchreceiver/client.go:               userPass := fmt.Sprintf("%s:%s", c.Username, c.Password) | `Password` is a `configopaque.String` |
| receiver/postgresqlreceiver/client.go:  connStr := fmt.Sprintf("port=%s host=%s user=%s password=%s %s %s", port, host, conf.username, conf.password, dbField, sslConnectionString(conf.tls)) | `password` is a string |
| receiver/saphanareceiver/client.go:     connector, err := sapdriver.NewDSNConnector(fmt.Sprintf("hdb://%s:%s@%s", c.receiverConfig.Username, c.receiverConfig.Password, c.receiverConfig.TCPAddr.Endpoint)) | `Password` is a `configopaque.String`|

The impact therefore is likely:
* [ ] `exporter/dynatraceexporter/config/config.go`
* [ ] `exporter/logzioexporter/factory.go`
* [ ] `receiver/elasticsearchreceiver/client.go`
* [ ] `receiver/saphanareceiver/client.go`

I can follow up with those four use cases, sounds good?

.chloggen/configopaque_stringer.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
mx-psi, line 4: Given the impact you discovered in contrib, I would mark this as a `breaking` just to be safe and make sure more eyes are on this
```suggestion
change_type: breaking
```

bogdandrutu: For my to learn: Confused based on what golang rules this is breaking.

atoulme: I guess Pablo is trying to make sure this is advertised as a change that can break some uses of the configopaque.String as part of its use with `fmt.Stringer`. This is not really a change impacting the go API.

mx-psi: > For my to learn: Confused based on what golang rules this is breaking.

Yes, arguably it's a breakage in behavior, but it's not impacting the API

mx-psi (approved):
LGTM (modulo the discussion on breaking vs enhancement), regardless of this discussion being solved I am going to wait to merge this until after the 0.92.0 release to give us some time to catch any errors caused by this.

mx-psi:
I am going to merge this, we can observe the impact over the next two weeks and think whether it should be breaking based on that

TylerHelmuth:
@mx-psi this ended up being a breaking change for the honeycombmarkerexporter.  We were doing `req.Header.Set(honeycombTeam, fmt.Sprint(e.apiKey))` instead of `req.Header.Set(honeycombTeam, string(e.apiKey))` which resulted in the header being set to `[REDACTED]`.  I've updated the code here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30402.

TylerHelmuth:
Finding any more spots here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30405

mx-psi:
Filed #9267 to make this more prominent in the changelog
Filed #9268 to try and catch more instances of using this

PR #9217, opened 2024-01-04
[chore][config/configgrpc] Enable goleak check

crobert-1:

Enables goleak to run on the configgrpc package. Requires ignoring the opencensus-go leak.

A few tests required an additional client connection close command to exit the client go routine. Additional information can be found [here](https://pkg.go.dev/google.golang.org/grpc#DialContext) in the explanation for the method grpc.DialContext, specifically this comment:
```
Users should call ClientConn.Close to terminate all the pending operations after this function returns.
```
configgrpc's method [`ToClientConn`](https://github.com/open-telemetry/opentelemetry-collector/blob/36730599ae4fb2ffc05d8e6e8a42e0bb0880703f/config/configgrpc/configgrpc.go#L178) is directly calling the `grpc.DialContext` method referenced.

#9165 

Added goleak check is passing


config/configgrpc/configgrpc_test.go
```diff
@@ -886,7 +889,8 @@ func TestClientInfoInterceptors(t *testing.T) {
 					require.NoError(t, tt.Shutdown(context.Background()))
 				}()
 
-				grpcClientConn, errClient := gcs.ToClientConn(context.Background(), componenttest.NewNopHost(), tt.TelemetrySettings)
+				var errClient error
+				grpcClientConn, errClient = gcs.ToClientConn(context.Background(), componenttest.NewNopHost(), tt.TelemetrySettings)
```
bogdandrutu, line 893: you can use `defer func() {assert.NoError(t, grpcClientConn.Close())}()` instead.

config/configgrpc/configgrpc_test.go
```diff
@@ -641,6 +641,7 @@ func TestHttpReception(t *testing.T) {
 				assert.NoError(t, errResp)
 				assert.NotNil(t, resp)
 			}
+			assert.NoError(t, grpcClientConn.Close())
```
bogdandrutu, line 644: Use defer closer to where the connection is created, since you want that to run even if test stops prematurely.

config/configgrpc/configgrpc_test.go
```diff
@@ -685,6 +686,7 @@ func TestReceiveOnUnixDomainSocket(t *testing.T) {
 	resp, errResp := c.Export(ctx, ptraceotlp.NewExportRequest(), grpc.WaitForReady(true))
 	assert.NoError(t, errResp)
 	assert.NotNil(t, resp)
+	assert.NoError(t, grpcClientConn.Close())
```
bogdandrutu, line 689: same comment.

bogdandrutu (approved):

PR #8964, opened 2023-11-20
[extension/memorylimiter] add memory limiter extension, a copy of memorylimiter processor

timannguyen:
**Description:**

following https://github.com/open-telemetry/opentelemetry-collector/issues/8632, this change is to move memory limiter to being an extension. This allows us to place the component to reject incoming connections due to limited memory, providing better protection to the otel collector. 

missing feature: receiver fairness. issue where a receiver hogs all the resource can happen. 

**Link to tracking Issue:** <Issue number if applicable>
 https://github.com/open-telemetry/opentelemetry-collector/issues/8632

**Testing:** 
unit test

**Documentation:**

README.md

PR #9218, opened 2024-01-04
[chore] [extension/memory_ballast] Move ballast size calculation to the init step

dmitryax:
This is needed to unblock moving `memory_limiter` from the processor to the extension.

`memory_limiter` uses the ballast size to accurately calculate the memory usage. Even if the memory_ballast extension is deprecated, the user can still use it. In order to make them work together, we need to make sure that the ballast is calculated before the `memory_limiter` extension is started. This change moves the ballast calculation in the ballast extension from the Start to the Init step.

Alternatively, we could make the new `memory_limiter` extension fail if `memory_ballast` is found. But that would:
1. make the logic of memory_limiter processors and extension split more complicated;
2. force users to handle the ballast deprecation before trying the new `memory_limiter` extension.

### DIFF
--- a/.chloggen/change-memory-balast-init.yaml
+++ b/.chloggen/change-memory-balast-init.yaml
@@ -1,0 +1,13 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: 'extension/memory_ballast'
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Move memory ballast initialization to from the start to the build step"
+
+# One or more tracking issues or pull requests related to the change
+issues: [9218]
--- a/extension/ballastextension/factory.go
+++ b/extension/ballastextension/factory.go
@@ -27,5 +27,5 @@ func createDefaultConfig() component.Config {
 }
 
 func createExtension(_ context.Context, set extension.CreateSettings, cfg component.Config) (extension.Extension, error) {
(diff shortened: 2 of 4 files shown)


bogdandrutu:
Before jumping to this ballast change, I would like to understand:
* We deprecated `memory_ballast`, how does the `memory_[processor|extension]` work with the new golang way of doing ballast?
* If we remove the deprecated `memory_ballast` do we still need this? Why do we not wait to remove the `memory_ballast` extension?

dmitryax:
> We deprecated memory_ballast, how does the memory_[processor|extension] work with the new golang way of doing ballast?

With the deprecation of the `memory_ballast` extension, we recommend using `GOMEMLIMIT` env var instead. `memory_limiter` would work well with it. Actually, one of the features of the `memory_limiter` processor (soft limit forcing GC) becomes useless. We can even change the implementation just to set `GOMEMLIMIT` instead if it's not set. The main feature of the memory_limiter extension rejecting the incoming requests is the most important and doesn't conflict with `GOMEMLIMIT` in any way.

> If we remove the deprecated memory_ballast do we still need this? Why do we not wait to remove the memory_ballast extension?

If we remove it, we don't need this PR. But, as I mentioned in the description, I thought it's better to not require removing the ballast extension before trying the memory_limiter extension. We should probably keep the ballast for another year as deprecated. We could fail to start the memory_limiter extension if ballast is configured, but that also makes the logic of splitting memory_limiter processor/extension a bit more complicated. I'm fine with that approach though.

mx-psi:
> We could fail to start the memory_limiter extension if ballast is configured [...]

I like this approach better; nudging users towards not using the ballast extension seems like a good idea.

dmitryax:
> I like this approach better; nudging users towards not using the ballast extension seems like a good idea.

OK, I'll close this PR and add the ballast extension check after https://github.com/open-telemetry/opentelemetry-collector/pull/8964 is merged

PR #6708, opened 2022-12-08
Reload client CAs in server TLS settings on file change

akorompai:
**Description:** Reload client CAs in server TLS settings when file has been modified

**Link to tracking Issue:** #6524

**Testing:** manual only

**Documentation:** none

PR #9220, opened 2024-01-04
[config/configtls] Enable goleak check

crobert-1:

This change enables goleak to check the config/configtls package for memory and goroutine leaks. There was originally a goroutine leak which is fixed by calling the newly added method `TLSServerSetting.Shutdown`.

The major change of this PR is to make the `*clientCAsFileReloader` object a member of the `TLSServerSetting` struct, as well as adding a public `Shutdown` method to `TLSServerSetting`. This was the cleanest way I could add a way to shut down the reloader, otherwise it will be a leaked goroutine. 

Also, I had to change `LoadTLSConfig()` to be a [pointer type receiver](https://gobyexample.com/methods), so that it could modify the given object by reference instead of only having the value copy of the `TLSServerSetting` object it was called on.

#9165

Goleak is passing successfully, as well as all existing tests.


jpkrohling (approved):

jpkrohling:
Sorry about the go.mod conflict. Once that is fixed, this can be merged. Thank you!

jpkrohling (approved):

config/configtls/configtls.go
```diff
@@ -327,24 +330,25 @@ func (c TLSClientSetting) LoadTLSConfig() (*tls.Config, error) {
...
 	if c.ClientCAFile != "" {
-		reloader, err := newClientCAsReloader(c.ClientCAFile, &c)
+		var err error
+		c.reloader, err = newClientCAsReloader(c.ClientCAFile, c)
 		if err != nil {
 			return nil, err
 		}
 		if c.ReloadClientCAFile {
-			err = reloader.startWatching()
+			err = c.reloader.startWatching()
```
bogdandrutu, line 345: Strange logic that we create a reloader even when not starting, maybe split the logic? Separate PR would be preferred for that.

config/configtls/configtls.go
```diff
@@ -327,24 +330,25 @@ func (c TLSClientSetting) LoadTLSConfig() (*tls.Config, error) {
 }
 
 // LoadTLSConfig loads the TLS configuration.
-func (c TLSServerSetting) LoadTLSConfig() (*tls.Config, error) {
+func (c *TLSServerSetting) LoadTLSConfig() (*tls.Config, error) {
```
bogdandrutu, line 333: Since for the same `TLSServerSetting` we may call this multiple times, it may be better to actually return a `done` callback instead and have the reloader shutdown there.

jpkrohling: Great catch!

config/configtls/configtls.go
crobert-1: To make sure I understand, you're saying it's possible to get here even when the collector is not starting, so we should fix this logic to make sure the reloader only starts watching during start, am I following your thought?

config/configtls/configtls.go
crobert-1: So `done` would be a function that shuts down the reloader, and is called by the user on shutdown or when they no longer need the current TLS config, is that correct?

config/configtls/clientcasfilereloader.go
```diff
@@ -132,7 +132,7 @@ func (r *clientCAsFileReloader) handleWatcherEvents() {
 
 func (r *clientCAsFileReloader) shutdown() error {
 	if r.shutdownCH == nil {
-		return fmt.Errorf("client CAs file watcher is not running")
```
crobert-1, line 135: I don't think we should return an error if the shutdown channel is nil.

[From documentation:](https://pkg.go.dev/go.opentelemetry.io/collector/component#Component)
```
	// This method must be safe to call:
	//   - without Start() having been called
	//   - if the component is in a shutdown state already
```
Maybe I'm misinterpreting what `safe to call` means, but I don't think we should return an error unless there's an issue shutting down. I don't think we should error if shutdown is successful. (Which I believe is the second point from docs here).

Even though this itself isn't specifically a collector component, I believe docs still apply as it's a part of collector shutdown.

.chloggen/goleak_configtls.yaml
```diff
@@ -0,0 +1,29 @@
...
+note: Add shutdown call back to TLSServerSetting to fix memory leaks
+
+# One or more tracking issues or pull requests related to the change
+issues: [9165]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The TLSServerSetting.LoadTLSConfig method signature has been modified to now return an extra
```
crobert-1, line 19: Feedback is welcome here, I want to make sure the action needed by users is as clear as possible.

config/configtls/configtls.go
crobert-1: I believe I've implemented what was requested here 👍

crobert-1: Looking more at consumers of `TLSServerSetting.LoadTLSConfig`, it looks like there isn't a simple way to keep track of the lifecycle of these objects in a uniform way externally. It may make sense to keep track internally and then call all of the shutdown methods on `TLSServerSetting.Shutdown`. This will make it so this is no longer a breaking change, and the only extra requirement on users is calling the shutdown method here.

.chloggen/goleak_configtls.yaml
crobert-1: My new suggested resolution is a `bug_fix` instead of `breaking`, so this is no longer relevant.

config/configtls/configtls.go
crobert-1: I implemented my proposed solution and added tests.

crobert-1:
@jpkrohling and @bogdandrutu: I believe I've addressed the concerns mentioned, another look would be appreciated when you're able!

config/configtls/clientcasfilereloader.go
TylerHelmuth: I think your change makes sense.  I interpret that comment to mean "calling shutdown on something not started should not error".

config/configtls/configtls.go
```diff
@@ -373,13 +378,13 @@ func (c ClientConfig) LoadTLSConfig() (*tls.Config, error) {
 }
 
 // LoadTLSConfig loads the TLS configuration.
-func (c ServerConfig) LoadTLSConfig() (*tls.Config, error) {
+func (c *ServerConfig) LoadTLSConfig() (*tls.Config, error) {
```
TylerHelmuth, line 381: Goland yells at me when I mix receiver types. I believe best practice is not to mix types, so if this needs to be a pointer I think all the methods need updating.

config/configtls/configtls.go
```diff
@@ -389,17 +394,29 @@ func (c ServerConfig) LoadTLSConfig() (*tls.Config, error) {
...
 	return tlsCfg, nil
 }
 
 func (c ServerConfig) loadClientCAFile() (*x509.CertPool, error) {
 	return c.loadCert(c.ClientCAFile)
 }
 
+func (c ServerConfig) Shutdown() error {
+	var err error
+	if c.ReloadClientCAFile {
```
TylerHelmuth, line 412: Since this value is public, I believe it is possible for it to be set to true when calling `LoadTLSConfig`, which will cause some `shutdowns` funcs to be remembered, then for `ReloadClientCAFile` to be set to false, and then `Shutdown` to be called resulting in the `shutdown` funcs being executed.

I think it would be best to always try to loop through the list and invoke anything inside.

config/configtls/configtls.go
crobert-1: Done 👍

config/configtls/configtls.go
crobert-1: Good catch, thanks! I've updated.

receiver/otlpreceiver/otlp.go
```diff
@@ -186,8 +187,18 @@ func (r *otlpReceiver) Start(ctx context.Context, host component.Host) error {
...
+		}
+
+		if r.cfg.GRPC != nil && r.cfg.GRPC.TLSSetting != nil {
+			err = multierr.Append(err, r.cfg.GRPC.TLSSetting.Shutdown())
+		}
+	}
+
 	if r.serverHTTP != nil {
-		err = r.serverHTTP.Shutdown(ctx)
+		err = multierr.Append(err, r.serverHTTP.Shutdown(ctx))
```
TylerHelmuth, line 201: I'd prefer `errors.Join` instead of `multierr`.

crobert-1: Agreed, updated.

TylerHelmuth: If you don't want the resulting error to be nested I believe you need to keep a slice of errors and do the `errors.Join` once at the end.  I'm not sure here if that matters or not.

crobert-1: Thanks for catching that, I've updated.

receiver/otlpreceiver/otlp.go
```diff
@@ -186,8 +187,18 @@ func (r *otlpReceiver) Start(ctx context.Context, host component.Host) error {
...
 
+	if r.cfg != nil {
+		if r.cfg.HTTP != nil && r.cfg.HTTP.TLSSetting != nil {
+			err = r.cfg.HTTP.TLSSetting.Shutdown()
+		}
+
+		if r.cfg.GRPC != nil && r.cfg.GRPC.TLSSetting != nil {
+			err = multierr.Append(err, r.cfg.GRPC.TLSSetting.Shutdown())
+		}
+	}
```
TylerHelmuth, line 198: Looking at this some more I really don't like that any component that uses `configtls` has to add this check.  I am worried about components not realizing that the `configtls.TLSSetting` embedded in `confighttp.ServerConfig` needs a `Shutdown` call.

I think it is out of scope of this PR, which is only trying to solve a leak added by existing functionality`, but I think we should take another look at how `config` modules should handle goroutines and `Shutdown`.

TylerHelmuth (approved):

yurishkuro:
Q: why do we implement reloading of client CA file with a background goroutine instead of using the same approach as `certReloader` which does it on-demand when GetCertificate is called?

https://github.com/open-telemetry/opentelemetry-collector/blob/ede9e304314dddbdeb10b5ee06b1a49ab88e4f69/config/configtls/configtls.go#L140

This would avoid a goroutine leak and the unpleasant side-effects of having to call Shutdown on a "config" object.

mx-psi:
@crobert-1 What's the status of this PR? Does it need more reviews? Do you need to fix something? (No rush, just trying to clean up our backlog)

crobert-1:
> @crobert-1 What's the status of this PR? Does it need more reviews? Do you need to fix something? (No rush, just trying to clean up our backlog)

Just waiting for reviews 👍

mx-psi:
Can you take a look at Yuri's question above?

mx-psi:
@crobert-1 Mind taking a look at https://github.com/open-telemetry/opentelemetry-collector/pull/9220#issuecomment-2119380494 ?

crobert-1:
> @crobert-1 Mind taking a look at [#9220 (comment)](https://github.com/open-telemetry/opentelemetry-collector/pull/9220#issuecomment-2119380494) ?

My apologies for the delay, thanks for the ping! I'm taking a look at this. There's no context from the [original issue](https://github.com/open-telemetry/opentelemetry-collector/issues/6524) or [implementing PR](https://github.com/open-telemetry/opentelemetry-collector/pull/6708) why this design was chosen. However, the issue says it should be done like the `certReloader`, which, as @yurishkuro pointed out, is done in a lazy manner on `Get`.

I'm working on implementing this to simply read the file on `Get` rather than the proactive watching approach that's currently implemented. If it works properly, I'll open a new PR instead of this approach.

_Marking as draft until I get a chance to properly prove the alternative approach will work._

akorompai:
> > @crobert-1 Mind taking a look at [#9220 (comment)](https://github.com/open-telemetry/opentelemetry-collector/pull/9220#issuecomment-2119380494) ?
> 
> My apologies for the delay, thanks for the ping! I'm taking a look at this. There's no context from the [original issue](https://github.com/open-telemetry/opentelemetry-collector/issues/6524) or [implementing PR](https://github.com/open-telemetry/opentelemetry-collector/pull/6708) why this design was chosen. However, the issue says it should be done like the `certReloader`, which, as @yurishkuro pointed out, is done in a lazy manner on `Get`.
> 
> I'm working on implementing this to simply read the file on `Get` rather than the proactive watching approach that's currently implemented. If it works properly, I'll open a new PR instead of this approach.
> 
> _Marking as draft until I get a chance to properly prove the alternative approach will work._

@crobert-1 Sorry for all the troubles my PR is causing, I didn't realize that the shutdown concern wasn't addressed before it got merged  (https://github.com/open-telemetry/opentelemetry-collector/pull/6708#discussion_r1088291202)
Please note that a dedicated client `certReloader` was implemented in a previous PR, but got rejected in favor of a fs watcher solution (https://github.com/open-telemetry/opentelemetry-collector/pull/6525#issuecomment-1322557224)

PR #9225, opened 2024-01-05
[receiver/otlpreceiver] Enable goleak check

crobert-1:

Add `goleak` to detect leaks in tests. Leaking goroutines were detected that were caused by a dependency that we can ignore (`go.opencensus.io/stats/view.(*worker).start`), a `Shutdown` call was also added that was missing.

#9165 

Added check is passing as well as existing tests.


receiver/otlpreceiver/factory_test.go
```diff
@@ -261,6 +261,7 @@ func TestCreateMetricReceiver(t *testing.T) {
 			assert.NoError(t, err)
 			if tt.wantStartErr {
 				assert.Error(t, mr.Start(context.Background(), componenttest.NewNopHost()))
+				assert.NoError(t, mr.Shutdown(context.Background()))
```
bogdandrutu, line 264: Do we need to close shutdown if it returned error? Why?

crobert-1: The test case that's causing a leak (`invalid_http_port`) is successfully starting a GRPC server with the default config, but failing to start a http server because of an invalid port. The HTTP server causes the `mr.Start` call to fail and return an error, but the GRPC server is up and running waiting for requests.

Upon further investigation, this should be handled by the receiver itself, not a caller since the collector fails to start on this error, the `Shutdown` call should always be called on an unsuccessful `Start` call. I've removed this call from tests and moved it into the `Start` function on error. Let me know if there's a better way to handle this.

crobert-1:
Failing test is frequency of #9230

receiver/otlpreceiver/otlp.go
```diff
@@ -168,11 +169,15 @@ func (r *otlpReceiver) startHTTPServer(host component.Host) error {
...
+func (r *otlpReceiver) Start(ctx context.Context, host component.Host) error {
 	if err := r.startGRPCServer(host); err != nil {
 		return err
 	}
 	if err := r.startHTTPServer(host); err != nil {
+		// It's possible that a valid GRPC server configuration was specified,
+		// but an invalid HTTP configuration. If that's the case, the successfully
+		// started GRPC server must be shutdown to ensure no goroutines are leaked.
+		err = multierr.Append(err, r.Shutdown(ctx))
 		return err
```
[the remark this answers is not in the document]
crobert-1: Done

receiver/otlpreceiver/otlp.go
```diff
@@ -168,12 +169,15 @@ func (r *otlpReceiver) startHTTPServer(host component.Host) error {
...
+func (r *otlpReceiver) Start(ctx context.Context, host component.Host) error {
 	if err := r.startGRPCServer(host); err != nil {
 		return err
 	}
 	if err := r.startHTTPServer(host); err != nil {
-		return err
+		// It's possible that a valid GRPC server configuration was specified,
+		// but an invalid HTTP configuration. If that's the case, the successfully
+		// started GRPC server must be shutdown to ensure no goroutines are leaked.
+		return multierr.Append(err, r.Shutdown(ctx))
```
TylerHelmuth, line 180: nit: should we be using `errors.Join` instead of multierr?

crobert-1: `errors.Join` is probably preferred, I've updated.

TylerHelmuth (approved):

bogdandrutu (approved):

Issue #8940, opened 2023-11-16
Recursive unmarshaling of components


djaglowski:
> Technically, we could do steps 1-3 again for only these components but it looks like a substantial overhaul to separate this functionality from the unmarshaler.

That was more or less the approach I had in mind. I agree it would probably require a substantial overhaul, but I thought about it some more after reading your summary and I still think it is possibly worth the effort. I've attempted to draft a solution that would solve all our issues with minimal downsides, but let me know if there's some fatal flaw that I've missed.

I think the problems with implementing templating within our current unmarshaling facilities are that they expect a one-to-one mapping between component configs and factories, and that all component unmarshaling is done within a section of the config instead of in the scope of the whole config.

We would probably still want to take advantage of `configunmarshaler.Configs` to unmarshal individual components, but would want to process the whole config after that to do template expansion. I would see a "config processor" being something similar to a converter that implements the following interface. A list of these would be registered with `otelcol.ConfigProvider` and would be used during config unmarshaling.

```golang
type ConfigProcessor interface {
	// Prefix for configs like "template/"
	Prefix() string

	// Config type for component configs that need
	// processing like `template/my_receiver`
	ComponentConfig() any

	// Key and type for the processor's config (e.g. "templates" and `templatesConfig`)
	ProcessorConfig() (string, any)

	// Handles configs starting with the processor's prefix
	// given the available factories and its config.
	// All instances of the ComponentConfig type should be
	// replaced by component.Config when this function returns.
	Process(pc *PartialConfig, factories Factories, config any) error
}
```

We could then split config unmarshaling into stages:

First, unmarshal the config as much as possible, considering during unmarshaling that there may be some keys that need further processing. So we could unmarshal components, but any config keys prepended with `template/` would need to be handled by a processor. I think this first step would produce a struct like:

```golang
type PartialConfig struct {
	Receivers: map[string]any // Factory or types returned by ConfigProcessor.ComponentConfig()
	[...]
	Service: PartialService // Some type that enables accessing/creating pipelines for the case of templating 
}
```

We would then run the `Process` method of each `ConfigProcessor` on the `PartialConfig`, with the expectation that after all processors have run, we're left with `map[string]component.Config` for each component class. I think the template converter would run in a similar fashion to how it runs now with this: it would remove the component template from the config, then add all components and add them to the pipelines.

[...]

PR #9228, opened 2024-01-05
[otelcol] Allow passing confmap.Providers to otelcol.Collector

evan-bradley:
**Description:**

Demonstrates one way to resolve https://github.com/open-telemetry/opentelemetry-collector/issues/4759. This implements the second approach that I've outlined here: https://github.com/open-telemetry/opentelemetry-collector/issues/4759#issuecomment-1863082504.

I think the main advantage of this approach is that it cleans up the API for the package. `otelcol.ConfigProvider` is a fairly thin wrapper around `confmap.Resolver`, so I think we could ultimately remove the interface, and any custom functionality for config merging or unmarshaling could be exposed to users through settings rather through a custom implementation.

I'm opening up this PR as a draft now as a point of discussion, if we decide we'd like to take another approach I'll close this and change course.

**Link to tracking Issue:**

Works toward  https://github.com/open-telemetry/opentelemetry-collector/issues/4759

**Testing:**

_TODO_

**Documentation:**

_TODO_


otelcol/collector.go
```diff
@@ -66,10 +66,16 @@ type CollectorSettings struct {
...
 
+	// Deprecated: [v0.xx.0] Use ConfigProviderSettings instead.
 	// ConfigProvider provides the service configuration.
 	// If the provider watches for configuration change, collector may reload the new configuration upon changes.
 	ConfigProvider ConfigProvider
 
+	// ConfigProviderSettings allows configuring the way the Collector retrieves its configuration
+	// The Collector will reload based on configuration changes from the ConfigProvider if any
+	// confmap.Providers watch for configuration changes.
+	ConfigProviderSettings ConfigProviderSettings
```
bogdandrutu, line 77: Before/After doing this:
1. You need to add a new `ConfigConverter` (similar to confmap converter) that allows to change the provided `Config` and before using it. There is an use-case that some have about this.
2. After this is merged we should change `ConfigProvider` to be a struct, no need to be an interface.

evan-bradley: 1. I think this should be covered by https://github.com/open-telemetry/opentelemetry-collector/issues/8940. I'd prefer to to do it after since I expect that will need a few rounds of review.
2. Sounds good.

yurishkuro: > There is a use-case that some have about this.

Is there a unit test for that use case?

evan-bradley:
This should be ready for review.

cc @mx-psi

otelcol/collector.go
```diff
@@ -105,8 +112,20 @@ type Collector struct {
...
-		return nil, errors.New("invalid nil config provider")
+	var err error
+	configProvider := set.ConfigProvider
+
+	if configProvider == nil {
+		resolverSet := set.ConfigProviderSettings.ResolverSettings
+		// Provide a default set of providers and converters if none have been specified.
+		if len(resolverSet.Providers) == 0 && len(resolverSet.Converters) == 0 {
+			set.ConfigProviderSettings = newDefaultConfigProviderSettings(resolverSet.URIs)
+		}
```
mx-psi, line 123: Should we return an error here instead and expose `newDefaultConfigProviderSettings` as part of the public API?

evan-bradley: Good callout. I would like to keep `newDefaultConfigProviderSettings` private and the defaults list to the builder once this change is in. Ideally we could remove the default providers and converters from the dependency tree for Collector builds that don't use them.

This statement was to make it so this PR isn't a breaking change, but I think it's better placed in `NewCommand` so we don't have to make a breaking change in `NewCollector`, so I moved it there.

mx-psi: Makes sense, thanks!

otelcol/collector.go
```diff
@@ -105,8 +112,14 @@ type Collector struct {
 
 // NewCollector creates and returns a new instance of Collector.
 func NewCollector(set CollectorSettings) (*Collector, error) {
```
mx-psi, line 114: During the interim period where one can provide both a `ConfigProvider` and a `ConfigProviderSettings`, should we check that if `configProvider` is not nil, then `ConfigProviderSettings` has the zero value for that type and error otherwise?

mx-psi (approved):

otelcol/collector.go
evan-bradley: We should do that, good catch. I added some validation to check this.

mx-psi (approved):

evan-bradley:
@bogdandrutu Could you give this another look?

evan-bradley:
@bogdandrutu @mx-psi Any additional thoughts on this? I think this is good to go, and will pair well with https://github.com/open-telemetry/opentelemetry-collector/pull/9516.

mx-psi:
@evan-bradley can you fix the conflict?

evan-bradley:
Thanks @mx-psi. Should be good now.

djaglowski (approved):

splunkericl:
> otelcol.ConfigProvider is a fairly thin wrapper around confmap.Resolver, so I think we could ultimately remove the interface, and any custom functionality for config merging or unmarshaling could be exposed to users through settings rather through a custom implementation.

Hey @evan-bradley, I just noticed this change today. We have a custom implementation of `ConfigProvider` mostly leveraging `Watch()` method to update our collector configurations dynamically. I don't see how the `Watch` method can be triggered with the new `ResolverSettings`?

mx-psi:
@splunkericl can you file an issue for this?

PR #9232, opened 2024-01-06
[chore] [exporterhelper] Integrate capacity limiting into the communication channel

dmitryax:
Integrate capacity limiting into internal channels used by both memory and persistent queues. Otherwise, with the independent capacity limiter, it's hard to ensure that queue size is always accurate going forward.

Benchmarks before:
```
goos: darwin
goarch: arm64
Benchmark_QueueUsage_1000_requests-10      	    3252	    325010 ns/op	  246059 B/op	      10 allocs/op
Benchmark_QueueUsage_100000_requests-10    	      39	  29811116 ns/op	24002870 B/op	      10 allocs/op
Benchmark_QueueUsage_10000_items-10        	    3404	    349753 ns/op	  246052 B/op	      10 allocs/op
Benchmark_QueueUsage_1M_items-10           	      40	  29415583 ns/op	24002858 B/op	      10 allocs/op
BenchmarkPersistentQueue_TraceSpans
BenchmarkPersistentQueue_TraceSpans/#traces:_1_#spansPerTrace:_1-10         	  338180	      3836 ns/op	    2851 B/op	      78 allocs/op
BenchmarkPersistentQueue_TraceSpans/#traces:_1_#spansPerTrace:_10-10        	   81369	     15822 ns/op	   14598 B/op	     289 allocs/op
BenchmarkPersistentQueue_TraceSpans/#traces:_10_#spansPerTrace:_10-10       	   13066	     90155 ns/op	  130087 B/op	    2417 allocs/op
```

Benchmarks after:
```
Benchmark_QueueUsage_1000_requests-10      	    4210	    278175 ns/op	  246055 B/op	      10 allocs/op
Benchmark_QueueUsage_100000_requests-10    	      42	  25835945 ns/op	24002968 B/op	      10 allocs/op
Benchmark_QueueUsage_10000_items-10        	    4376	    279571 ns/op	  246056 B/op	      10 allocs/op
Benchmark_QueueUsage_1M_items-10           	      42	  26483907 ns/op	24002995 B/op	      10 allocs/op
BenchmarkPersistentQueue_TraceSpans
BenchmarkPersistentQueue_TraceSpans/#traces:_1_#spansPerTrace:_1-10         	  328268	      4251 ns/op	    2854 B/op	      78 allocs/op
BenchmarkPersistentQueue_TraceSpans/#traces:_1_#spansPerTrace:_10-10        	  101683	     12238 ns/op	   14582 B/op	     289 allocs/op
BenchmarkPersistentQueue_TraceSpans/#traces:_10_#spansPerTrace:_10-10       	   13382	     86464 ns/op	  130154 B/op	    2417 allocs/op
```


exporter/exporterhelper/internal/persistent_queue.go
```diff
@@ -43,7 +42,10 @@ import (
 //	 index          index   x
 //	                        xxxx deleted
 type persistentQueue[T any] struct {
-	*queueCapacityLimiter[T]
+	// sizedElementsChannel is used by the persistent queue for two purposes:
+	// 1. a communication channel notifying the consumer that a new item is available.
+	// 2. capacity control based on the size of the items.
+	*sizedElementsChannel[permanentQueueEl]
```
bogdandrutu, line 48: QQ: What is the memory footprint (size) of this channel? I can think of where we allow 1G elements on disk, would this be 1GB?

dmitryax: That's a reasonable concern. I was thinking of addressing this by introducing the unbound channel, as I tried in the previous PR. But I can actually change the implementation a bit and bring back the `struct{}` channel for the persistent queue. Let me try that.

dmitryax: Applied. PTAL

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,101 @@
...
+func (vcq *sizedElementsChannel[T]) shutdown() {
+	close(vcq.ch)
+}
+
+func (vcq *sizedElementsChannel[T]) Size() int {
+	return int(vcq.initTotalSize + vcq.used.Load())
+}
+
+func (vcq *sizedElementsChannel[T]) Capacity() int {
+	return int(vcq.cap)
```
sfc-gh-bdrutu, line 100: return `cap(ch)`?

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,101 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package internal // import "go.opentelemetry.io/collector/exporter/exporterhelper/internal"
+
+import "sync/atomic"
+
+// sizedElementsChannel is a channel wrapper for sized elements with a capacity set to a total size of all the elements.
+// The channel will accept elements until the total size of the elements reaches the capacity.
+type sizedElementsChannel[T any] struct {
+	used *atomic.Uint64
```
sfc-gh-bdrutu, line 11: Consider to use https://pkg.go.dev/sync/atomic#Int64, simplifies math logic.

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	vcq.ch <- el
+	return nil
+}
+
+// initBulkEnqueue puts the elements into the queue with the given size. It's used by the persistent queue to
+// initialize the queue with the elements recovered from the disk.
+func (vcq *sizedElementsChannel[T]) initBulkEnqueue(els []T, totalSize uint64) {
+	vcq.initElementsCount = uint64(len(els))
+	vcq.initTotalSize = totalSize
```
sfc-gh-bdrutu, line 55: Unnecessary, you can put it into `used` and everything will work correct?

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+	vcq.ch <- el
+	return nil
+}
+
+// initBulkEnqueue puts the elements into the queue with the given size. It's used by the persistent queue to
+// initialize the queue with the elements recovered from the disk.
+func (vcq *sizedElementsChannel[T]) initBulkEnqueue(els []T, totalSize uint64) {
+	vcq.initElementsCount = uint64(len(els))
```
sfc-gh-bdrutu, line 54: Why do you need this?

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// newSizedElementsChannel creates a sized elements channel. Each element is assigned a size by the provided sizer.
+// chanCapacity is the capacity of the underlying channel which usually should be equal to the capacity of the queue to
+// avoid blocking the producer.
+func newSizedElementsChannel[T any](capacity int) *sizedElementsChannel[T] {
+	return &sizedElementsChannel[T]{
+		used: &atomic.Uint64{},
+		cap:  uint64(capacity),
+		ch:   make(chan T, capacity),
```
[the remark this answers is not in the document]
dmitryax: It's possible that the capacity of the channel can be higher than the configured capacity. It happens when we restore a persistent queue from a disk that is bigger than the configured capacity. In that case, we want to extend the channel to keep all the restored items but not allow any new items into the channel until the capacity goes down to the configured level.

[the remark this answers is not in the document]
bogdandrutu: In this code I don't see where that happens.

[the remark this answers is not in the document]
bogdandrutu: Please comment around the cap variable.

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: The problem is that the total size of the restored queue can be off. So we need to keep it separate and cler this value once the restored queue is drained

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: It is the same reason as in the previous comment. We need to know when the restored queue is drained

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: Applied

exporter/exporterhelper/internal/sized_elements_channel.go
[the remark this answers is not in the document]
dmitryax: added a comment

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: It can be a different value as I mentioned above

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,101 @@
...
+
+// sizedElementsChannel is a channel wrapper for sized elements with a capacity set to a total size of all the elements.
+// The channel will accept elements until the total size of the elements reaches the capacity.
+type sizedElementsChannel[T any] struct {
+	used *atomic.Uint64
+	cap  uint64
+	ch   chan T
+
+	// The following fields are used by the persistent queue to initialize the queue with the elements recovered from
+	// the disk. Synchronization is not required because the persistent queue doesn't access it concurrently.
```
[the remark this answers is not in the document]
dmitryax: You're right. I added a mutex specifically for mutating these two values cause they have to stay in sync

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,111 @@
...
+		vcq.initElementsCount--
+		if vcq.initElementsCount == 0 || vcq.initTotalSize < size {
+			vcq.initTotalSize = 0
+		} else {
+			vcq.initTotalSize -= size
+		}
+		return el, true
+	}
+
+	vcq.used.Add(^(size - 1))
```
bogdandrutu, line 96: https://pkg.go.dev/sync/atomic#Int64.Add you can do Add(-size) that is the reason I suggested to use `Int64` to simplify this.

exporter/exporterhelper/internal/sized_elements_channel.go
bogdandrutu: Why? I am not sure I understand the logic and what you try to achieve. Can you tell me with some numbers for capacity/initial size what you try to achieve? I am not sure, but I still believe to achieve (what I think you try) can be done without these values.

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,111 @@
...
+		cap:  int64(capacity),
+		ch:   make(chan T, capacity),
+	}
+}
+
+// enqueue puts the element into the queue with the given sized if there is enough capacity.
+// Returns an error if the queue is full. The callback is called before the element is committed to the queue.
+// If the callback returns an error, the element is not put into the queue and the error is returned.
+func (vcq *sizedElementsChannel[T]) enqueue(el T, size int64, callback func() error) error {
+	if vcq.initTotalSize+vcq.used.Add(size) > vcq.cap {
```
bogdandrutu, line 44: Wrong usage of `initTotalSize` since is not protected by the mutex.

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,111 @@
...
+		vcq.used.Add(^(size - 1))
```
bogdandrutu, line 45: Same use `-size`

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,111 @@
...
+	vcq.ch <- el
+	return nil
+}
+
+// initBulkEnqueue puts the elements into the queue with the given size. It's used by the persistent queue to
+// initialize the queue with the elements recovered from the disk. Cannot be called concurrently.
+// Callers must ensure that the queue is empty when calling this function.
+func (vcq *sizedElementsChannel[T]) initBulkEnqueue(els []T, totalSize int64) {
+	vcq.initElementsCount = int64(len(els))
+	vcq.initTotalSize = totalSize
```
bogdandrutu, line 63: Unprotected usage. This is UB.

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,111 @@
...
+// The callback is called before the element is removed from the queue. It must return the size of the element.
+func (vcq *sizedElementsChannel[T]) dequeue(callback func(T) (size int64)) (T, bool) {
+	el, ok := <-vcq.ch
+	if !ok {
+		return el, false
+	}
+
+	size := callback(el)
+
+	if vcq.initElementsCount > 0 {
```
bogdandrutu, line 84: Unprotected usage. This is UB.

exporter/exporterhelper/internal/sized_elements_channel.go
```diff
@@ -0,0 +1,111 @@
...
+		vcq.initSizeMutex.Lock()
+		defer vcq.initSizeMutex.Unlock()
```
bogdandrutu, line 86: Defer here ends when the func ends not when this block, it is a bad pattern.

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: Oh, right. updated

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: fixed

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: Ok. I added RW mutex to handle it

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: fixed

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: fixed

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: This slows down the hot path a bit, but not significantly. I updated the benchmark results

exporter/exporterhelper/internal/sized_elements_channel.go
dmitryax: At the start of the collector, we restore a number of requests and total items from the disk as separate values. It's guaranteed that the number of requests matches the actual number because of the WAL indexes. But there is no such guarantee for the number of items, it can be off in 2 scenarios:
- User switches from request-based to item-based sizing, so restored number of items is 0.
- We backup the total size of items on disk every 5 reads/writes as an optimization. So, if the collector is killed, the restored number of items can be slightly off. Let's say the collector is killed with 12 requests in the queue (10 items each), but the total number of items was flushed on disk when there were 10 requests in the queue. After restart, we restore 12 requests with 100 items in total instead of 120.

So we need to keep track of the restored queue separately until it's drained.

This is how it's currently implemented in the persistent queue. I'm just moving the logic here. Let me know if you have a better idea.

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// newSizedElementsChannel creates a sized elements channel. Each element is assigned a size by the provided sizer.
+// chanCapacity is the capacity of the underlying channel which usually should be equal to the capacity of the queue to
+// avoid blocking the producer.
+func newSizedElementsChannel[T any](capacity int, opts ...sizedElementsChannelOption[T]) *sizedElementsChannel[T] {
+	sech := &sizedElementsChannel[T]{
+		used: &atomic.Int64{},
+		cap:  int64(capacity),
+		ch:   make(chan T, capacity),
```
sfc-gh-bdrutu, line 42: Useless allocation if this gets re-initialized in the options.

dmitryax: Do you think we should avoid that? The override in the option happens under pretty rare conditions. We can avoid that but it'll add a bit of extra code after applying the options
```
	if sech.ch == nil {
		sech.ch = make(chan T, capacity)
	}
```

dmitryax: I applied this

exporter/internal/queue/bounded_memory_queue.go
```diff
@@ -30,41 +30,36 @@ type MemoryQueueSettings[T any] struct {
...
 }
 
 // Offer is used by the producer to submit new item to the queue. Calling this method on a stopped queue will panic.
 func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
-	if !q.queueCapacityLimiter.claim(req) {
-		return ErrQueueIsFull
-	}
-	q.items <- queueRequest[T]{ctx: ctx, req: req}
-	return nil
+	return q.sizedElementsChannel.enqueue(memQueueEl[T]{ctx: ctx, req: req}, q.sizer.Sizeof(req), nil)
```
bogdandrutu, line 40: nit: I like more push/pop than enqueue/dequeue.

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package queue // import "go.opentelemetry.io/collector/exporter/internal/queue"
+
+import "sync/atomic"
+
+// sizedElementsChannel is a channel wrapper for sized elements with a capacity set to a total size of all the elements.
+// The channel will accept elements until the total size of the elements reaches the capacity.
+type sizedElementsChannel[T any] struct {
```
bogdandrutu, line 10: s/sizedElementsChannel/sizedChannel?

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+	// It happens when we restore a persistent queue from a disk that is bigger than the pre-configured capacity.
+	cap int64
+	ch  chan T
+}
+
+type sizedElementsChannelOption[T any] func(*sizedElementsChannel[T])
+
+// withPreloadElements puts the elements into the queue with the given size. It's used by the persistent queue to
+// initialize the queue with the elements recovered from the disk.
+func withPreloadElements[T any](els []T, totalSize int64) sizedElementsChannelOption[T] {
```
bogdandrutu, line 23: To simplify this, probably next PR, we need to have 2 sizes for persistent:
* in memory size which is fixed and smaller than the storage;
* storage size which is how much to store in the storage;

bogdandrutu, line 23: Since no other options for the moment, I would not make this an option but another ctor:

```golang
func newSizedElementsChannelWithPreloadElements[T any](capacity int64, els []T, totalSize int64) *sizedElementsChannel[T] {
```

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+		for _, el := range els {
+			vcq.ch <- el
+		}
+	}
+}
+
+// newSizedElementsChannel creates a sized elements channel. Each element is assigned a size by the provided sizer.
+// chanCapacity is the capacity of the underlying channel which usually should be equal to the capacity of the queue to
+// avoid blocking the producer.
+func newSizedElementsChannel[T any](capacity int, opts ...sizedElementsChannelOption[T]) *sizedElementsChannel[T] {
```
bogdandrutu, line 40: Change capacity to be int64 since "size" you pass as int64.

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+// shutdown closes the queue channel to initiate draining of the queue.
+func (vcq *sizedElementsChannel[T]) shutdown() {
+	close(vcq.ch)
+}
+
+func (vcq *sizedElementsChannel[T]) Size() int {
+	return int(vcq.used.Load())
+}
+
+func (vcq *sizedElementsChannel[T]) Capacity() int {
```
bogdandrutu, line 111: We should change this to be int64 (happy to do it in a different PR). Same for Size since now we want to add bytes.

bogdandrutu (approved):

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+	if vcq.used.Add(-size) < 0 {
+		vcq.used.Store(0)
+	}
+	return el, true
+}
+
+// syncSize updates the used size to 0 if the queue is empty.
+// The caller must ensure that this call is not called concurrently with enqueue.
+// It's used by the persistent queue to ensure the used value correctly reflects the reality which may not be always
+// the case in case if the queue size is restored from the disk after a crash.
```
bogdandrutu, line 95: Can you explain why? Because we don't calculate the size correctly?

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+	if sech.ch == nil {
+		sech.ch = make(chan T, capacity)
+	}
+	return sech
+}
+
+// enqueue puts the element into the queue with the given sized if there is enough capacity.
+// Returns an error if the queue is full. The callback is called before the element is committed to the queue.
+// If the callback returns an error, the element is not put into the queue and the error is returned.
+func (vcq *sizedElementsChannel[T]) enqueue(el T, size int64, callback func() error) error {
```
bogdandrutu, line 57: Add a comment that size MUST be positive. We can even consider to change it to uint64? That my make the -size ugly to use with atomics, but I think it is ok.

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+func (vcq *sizedElementsChannel[T]) dequeue(callback func(T) (size int64)) (T, bool) {
+	el, ok := <-vcq.ch
+	if !ok {
+		return el, false
+	}
+
+	size := callback(el)
+
+	// The used size and the channel size might be not in sync if the queue in case it's restored from the disk.
+	// In that case we need to make sure it doesn't go below 0.
```
bogdandrutu, line 85: Why? You use this argument quite a lot, but never explain it in the code documentation.

exporter/internal/queue/persistent_queue.go
```diff
@@ -148,86 +143,100 @@ func (pq *persistentQueue[T]) initPersistentContiguousStorage(ctx context.Contex
...
+type permanentQueueEl struct{}
+
+// restoreQueueSizeFromStorage restores the queue size from storage.
+func (pq *persistentQueue[T]) restoreQueueSizeFromStorage(ctx context.Context) (uint64, error) {
+	var queueSize uint64
+	val, err := pq.client.Get(ctx, queueSizeKey)
+	if err == nil {
+		queueSize, err = bytesToItemIndex(val)
+	}
+	if err != nil {
```
bogdandrutu, line 178: You return on err, so you can move the `if err == nil {` after the not nil and no need to check the error and say `return bytesToItemIndex(val)`

exporter/internal/queue/sized_elements_channel.go
```diff
@@ -0,0 +1,113 @@
...
+func newSizedElementsChannel[T any](capacity int, opts ...sizedElementsChannelOption[T]) *sizedElementsChannel[T] {
+	sech := &sizedElementsChannel[T]{
+		used: &atomic.Int64{},
+		cap:  int64(capacity),
+	}
+	for _, opt := range opts {
+		opt(sech)
+	}
+	if sech.ch == nil {
+		sech.ch = make(chan T, capacity)
```
bogdandrutu, line 49: This will be a problem if we switch to use bytes, since it will be very large and lots of memory.

exporter/internal/queue/bounded_memory_queue.go
dmitryax: SGTM. Also good for differentiating from the queue

exporter/internal/queue/sized_elements_channel.go
dmitryax: Although the `sizedChannel` name may not be descriptive enough, I'm good with it because it's shorter.

exporter/internal/queue/sized_elements_channel.go
dmitryax: Because we don't flush the current queue size on the disk on every read/write. Should I mention that in the comment?

exporter/internal/queue/sized_elements_channel.go
dmitryax: Right. We can think of a solution outside of this PR, right?

exporter/internal/queue/bounded_memory_queue.go
dmitryax: updated

exporter/internal/queue/sized_elements_channel.go
dmitryax: changed

exporter/internal/queue/sized_elements_channel.go
dmitryax: Makes sense, updated

exporter/internal/queue/sized_elements_channel.go
dmitryax: Switched it to int64

exporter/internal/queue/sized_elements_channel.go
dmitryax: Added the comment. The atomic size field used to be in uint64, and then you asked to change it to int64 to simplify the subtraction. Do you think we should accept uint64 and convert it to int64?

exporter/internal/queue/sized_elements_channel.go
dmitryax: Added a comment

exporter/internal/queue/persistent_queue.go
dmitryax: Thanks! fixed

exporter/internal/queue/sized_elements_channel.go
dmitryax: done

bogdandrutu (approved):
Ship it :)

PR #9091, opened 2023-12-12
Deprecate exporterhelper.RetrySettings in favor of configretry.BackOffConfig

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/9089

PR #9235, opened 2024-01-08
[config/configretry] Allow zero multiplier and arbitrary randomization factor

mx-psi:
**Description:** 

Partial revert of #9091 to unblock open-telemetry/opentelemetry-collector-contrib/pull/30167

This does not mean that usage of a zero multiplier or a randomization factor outside of [0,1] is blessed upon by Collector maintainers, just that we need to unblock the release :)


codeboten (commented):
Although this would unblock the release, this changes the behaviour of Validate, see its previous behaviour: https://github.com/open-telemetry/opentelemetry-collector/pull/9091/files#diff-4dfc275bb0788b5ee1310c9f3d6aed3d5c7e514e28be5b87802907d3ed5e0c4dL50-L55

codeboten:
My bad, after reviewing the history, I realize that this validation was added here: https://github.com/open-telemetry/opentelemetry-collector/pull/9089, in which case I agree that loosening the rules for validation would be ok to unblock the release.

codeboten (approved):

.chloggen/mx-psi_configretry.yaml
```diff
@@ -0,0 +1,26 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: | 
+  These values were previously accepted and lead to unspecified behavior.
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
```
bogdandrutu, line 25: Since this goes into the release, will be confusing for users to read both initial changelog and this. I would rather update the initial one.

sfc-gh-bdrutu (approved):
Only the changelog entry comment.

.chloggen/mx-psi_configretry.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
dmitryax, line 4: Should this change just be mentioned in `.chloggen/addretrysetvalidation.yaml`? It's not a bugfix, right?

codeboten:
@dmitryax @bogdandrutu i moved the randomization_factor note in the subtext of the original changelog entry, PTAL

bogdandrutu (approved):

.chloggen/mx-psi_configretry.yaml
codeboten: Fixed

dmitryax (approved):

Issue #9252, opened 2024-01-09
OTel Collector Builder: Go "monorepo" support


jmacd:
**Is your feature request related to a problem? Please describe.**
The only supported way to build an OTel Collector requires generating a "fresh" go.mod file for the main package. This creates a challenging workflow for users of a "monorepo", which is to describe a single go.mod file used to build a large code base. This can also create a security concern; users who are familiar with Go module tooling may expect to update go.mod files using the Go toolchain. However, when they do, the next invocation of the `builder` will wipe out any changes that were made to the go.mod files.

**Describe the solution you'd like**
The proposed solution is to offer a mode in which a new go.mod file is not created. Instead, when this mode is selected, each of the modules that would have been required will be individually updated. There is a concept of the "enclosing module" which is determined by the Go toolchain. When the builder is invoked in this new mode, it will `go get` each module using the version in the build configuration, _only when the configured module is newer than the one in the enclosing Go module_, with emphasis on use of semantic version comparison to control whether an update is performed.

We can imagine two steps, the first already described: with a `--skip-new-go-module` flag the builder will `go get` each component instead of generating a new go.mod file. The second, nice-to-have but not essential feature called for, for users of a monorepo, is to use the configuration of the enclosing module to update the build configuration by editing the yaml and writing a new configuration. This `builder sync` command, named for its simlarity to the `go work sync` command, would rewrite the build configuration using the go module version that is found in the enclosing module for every component. That would allow the build configuration to be distributed and re-built outside of the monorepo, using the intended "current" versions that were in effect inside the monorepo.

**Describe alternatives you've considered**
In the case where as user has an OTel Collector build and wants to update one of the indirect dependencies, instead of a simple `go get` command, they can continue with the status quo. They will add a `replace` statement to the configuration and re-invoke the builder. This is error-prone for users who are not familiar with the collector build process.

kristinapathak:
https://github.com/open-telemetry/opentelemetry-collector/issues/10975#issuecomment-2390479021

jmacd:
I consider this finished!

PR #9253, opened 2024-01-09
WIP [builder] Support for `--skip-new-go-module` proof of concept

jmacd:
**Description:** Adds a --skip-new-go-module flag to the OTC builder. This enables users working in an existing go module environment (say, a "monorepo") to update the module they have, vs forcing the use of a new module.

With the new support inside an existing Go module, a collector main package can be generated using a `go:generate` directive. For example, in the directory where I want my collector built, the file `generate.go` has this line:

```
//go:generate builder --skip-new-go-module --skip-compilation --strict-versioning --config=./build-config.yaml
```

In the same directory, the `build-config.yaml` describes the collector to build. The builder generates the other files in the same directory. At this point, normal Go workflows can be used to update indirect dependencies. The optional `--strict-versioning` checks that the build configuration matches that resulting from the enclosing Go module.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/9252

**Testing:** WIP: No.

**Documentation:** Yes.


jmacd:
As discussed in today's Collector SIG call this is a proof that a solution to #9252 is a relatively simple change.  I've added one more commit, implementing and documenting a "strict versioning" mode, which applies additional checks following the `GetModules` operation. 

When strict versioning is configured, 

1. The existing INFO message about mismatch between the builder's own version and the configuration's `otelcol_version` field becomes an error.
2. A mismatch between `otelcol_version` and the final go module becomes an error
3. A mismatch between any component and the final go module becomes an error.

Each of these errors could have happened in esoteric and contrived cases without the new `--skip-new-go-module` functionality, for example if a component had a spurious dependency on the core library _and_ the component in the build depends on a newer version of the core library. When `--skip-new-go-module` functionality is configured, these checks become a way for a user to know if the build configuration is up-to-date with the enclosing go module. (In a perfect world, we might want a command that automatically updates the build configuration instead of a strict failure mode, but having a strict failure mode is much less work than auto-updating a build configuration.)

At this point, this PR is exactly where I want it to be for my internal build workflow, which uses a monorepo and builds collectors using internal components. I surely hope this work will be accepted. I would like to hear that there is sponsorship for this idea before I add tests of all the new code. Approvers, please take a look, thanks -- if you express support, I'll add tests, thanks.

codeboten (commented):
Thanks @jmacd, I think this use-case makes sense to support. I'll be glad to review the PR once the tests are added 👍🏻

jmacd:
Have discussed @kristinapathak taking on this work, will close for now.

PR #9259, opened 2024-01-10
[chore] simplify replacements with one mod edit call instead of many

atoulme:
This speeds up replacements and make build logs easier to parse.


codeboten (approved):

codeboten (changes requested):
Looks like something's not working as expected:

```
 Setting contrib at /tmp/opentelemetry-collector-contrib to use this core checkout
/bin/bash: -c: line 1: unexpected EOF while looking for matching `"'
/bin/bash: -c: line 2: syntax error: unexpected end of file
```

atoulme:
tabs and spaces I bet. Looking.

codeboten (approved):

PR #9041, opened 2023-12-05
[consumer] Allow annotating consumer errors with metadata

evan-bradley:
**Description:**

Revival of https://github.com/open-telemetry/opentelemetry-collector/pull/7439

This explores one possible way to allow adding metadata to errors returned from consumers. The goal here is to allow transmitting more data back up the pipeline if there is an error at some stage, with the goal of it being used by an upstream component, e.g. a component that will retry data, or a receiver that will propagate an error code back to the sender.

The current design eliminates the permanent/retryable error types in favor of a single error type that supports adding signal data to be retried. If no data is added to be retried, the error is considered permanent. Currently there is no distinction made between the signals for the sake of simplicity, the caller should know what signal is used when retrieving the retryable items from the error. Any options for retrying the data (e.g. a delay) are offered as options when adding data to retry.

The error type currently supports a few general metadata fields that are copied when a downstream error is wrapped:
* Partial successes can be expressed by setting the number of rejected items.
* gRPC and HTTP status codes can be set and translated between if necessary.

**Link to tracking Issue:**

Resolves #7047

cc @dmitryax

PR #9260, opened 2024-01-10
Treat PartialSuccess as Success

### LINKED ISSUE #9243: OTLP PartialSuccess responses should not be interpreted as errors, items should count as "rejected" in pipeline metrics
**Describe the bug**
In my opinion, the Collector's OTLP exporter handles PartialSuccess incorrectly. The present handling of this field is to return an error to the caller, when in fact the opposite behavior was originally intended in the specification. The meaning of PartialSuccess starts with a successful request. The consumer of the data has acknowledged successful receipt of valid data; the delivery pipeline is operational, but something is maybe ill-formed about the data. There is not an error code. This is a qualified success.

As described in https://github.com/open-telemetry/oteps/pull/238, there is an opportunity for Collectors to count the points rejected, as described in PartialSuccess messages, via a new `otel.outcome` label, which we might call "rejected".  Rejected points are those for which errors must be suppressed because some of the data was correct and all of the data was valid. It would be appropriate to return success and count rejected points, which are those in a category different than dropped which cannot be retried and were part of a successful export request.

It would also be appropriate -- and that the second half of this issue -- for Collector pipelines to report this kind of qualified success in such a way that OTel SDKs receive true information about PartialSuccess when they pass OTLP through a Collector pipeline. In other words, I would be happy if we built support to propagate PartialSuccess all the way (backwards) through the pipeline. Also, components should be able to introduce partial success responses themselves, as discussed in https://github.com/open-telemetry/opentelemetry-collector/issues/7669. 

As an example, consider a batch processor component forked from the core `batchprocessor`, here: https://github.com/open-telemetry/otel-arrow/tree/main/collector/processor/concurrentbatchprocessor

This component has been extended to block the caller until their own timeout takes place, or until the final outcome of the batch is known. After the batch export response is known, there are two cases:

1. All of the items were included in the same export, in which case we have all or none failure.
2. Some of the items were successful (batches were successful), some were not (batches failed)

This is a model case for the use of PartialSuccess -- there was some successful data which we have to acknowledge with a success, and there was some failure which we cannot allow to look like a transient error. It is incorrect to view it as a permanent error. This batch processor could use PartialSuccess to express back to the client exactly how many of its points were accepted.

We must recognize that this information can't always be perfect. If an item is included in a batch and then part of that batch has PartialSuccess w/ `rejected_points` != 0, then it will not be possible to assign an integer count correctly to the corresponding inputs. See https://github.com/open-telemetry/oteps/pull/238 for more details.

**Steps to reproduce**
Return data with PartialSuccess to a collector, for example, where one metric point out of 1000 has a problem with validation. This is the case PartialSuccess was meant to address. The collector returns an error when 999 points were correct.

The response should be a success. The pipeline metrics should count 1 `rejected` items. 

**What did you see instead?**
I see this code trigger:
```
	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedSpans() == 0) {
		return consumererror.NewPermanent(fmt.Errorf("OTLP partial success: \"%s\" (%d rejected)", resp.PartialSuccess().ErrorMessage(), resp.PartialSuccess().RejectedSpans()))
	}
```

with `(1 rejected)`. 

**What version did you use?**
0.91.1

**What config did you use?**
An `otlp/metrics` pipeline with 
```
      otelarrow/metrics:
        endpoint: vendorname.com:443
```
with a vendor that performs metric point validation.

jmacd:
**Description:** Changes the treatment of [PartialSuccess](https://opentelemetry.io/docs/specs/otlp/#partial-success), making them successful and logging a warning instead of returning an error to the caller. These responses are meant to convey successful receipt of valid data which could not be accepted for other reasons, specifically to cover situations where the OpenTelemetry SDK and Collector have done nothing wrong, specifically to avoid retries. While the existing OTLP exporter returns a permanent error (also avoids retries), it makes the situation look like a total failure when in fact it is more nuanced. 

As discussed in the tracking issue, it is a lot of work to propagate these "partial" successes backwards in a pipeline, so the appropriate simple way to handle these items is to return success.

In this PR, we log a warning. In a future PR, (IMO) as discussed in https://github.com/open-telemetry/oteps/pull/238, we should count the spans/metrics/logs that are rejected in this way using a dedicated outcome label.

**Link to tracking Issue:**
Part of #9243

**Testing:** Tests for the "partial success" warning have been added.

**Documentation:** PartialSuccess behavior was not documented. Given the level of detail in the README, it feels appropriate to continue not documenting, otherwise lots of new details should be added.


jmacd:
(Realize I need to fix the otlphttpexporter too!)

exporter/otlpexporter/otlp.go
```diff
@@ -97,7 +98,12 @@ func (e *baseExporter) pushTraces(ctx context.Context, td ptrace.Traces) error {
 	}
 	partialSuccess := resp.PartialSuccess()
 	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedSpans() == 0) {
-		return consumererror.NewPermanent(fmt.Errorf("OTLP partial success: \"%s\" (%d rejected)", resp.PartialSuccess().ErrorMessage(), resp.PartialSuccess().RejectedSpans()))
+		// TODO: These should be counted as "rejected", similar to dropped items.
```
TylerHelmuth, line 101: If there is an issue tracking this can we include it in the todo?

TylerHelmuth (commented):
This aligns with my understanding of https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#partial-success and https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#partial-success-1

exporter/otlpexporter/otlp.go
jmacd: I've updated the title of #9243 to include the additional scope. This PR would address the first half.

The reason this second half is tricky to fix now, is that exporters are constrained by the exporterhelper in terms of metrics they generate. Exporterhelper has no way to distinguish partial success, so without addressing that first, or expanding scope inside this PR, we have the option of counting all-success or all-failed. I believe all-success is the better option until we have a better solution.

I referred to this situation as a "success error", a case where the exporter wants the exporterhelper to see success (i.e., no error) but would like to modify the counts that are applied as it generates metrics. I am not sure how we should approach this from the `exporterhelper` perspective, @dmitryax -- I see in https://github.com/open-telemetry/opentelemetry-collector/pull/9041 there is discussion about adding additional fields to _error responses_, but this is a case calling for additional fields to _success responses_. It would require a new code style if every receiver/processor has to remember that some exporter errors are really successes, meaning to introduce special-case errors with "OK" status to convey partial success.

jmacd:
Reviewers, this is ready for review.

evan-bradley (approved):

TylerHelmuth (approved):

exporter/otlpexporter/otlp.go
```diff
@@ -110,7 +116,12 @@ func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) erro
 	}
 	partialSuccess := resp.PartialSuccess()
 	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedDataPoints() == 0) {
-		return consumererror.NewPermanent(fmt.Errorf("OTLP partial success: \"%s\" (%d rejected)", resp.PartialSuccess().ErrorMessage(), resp.PartialSuccess().RejectedDataPoints()))
+		// TODO: These should be counted as "rejected", similar to dropped items.
+		// https://github.com/open-telemetry/opentelemetry-collector/issues/9243
+		e.settings.Logger.Warn("partial success",
+			zap.String("message", resp.PartialSuccess().ErrorMessage()),
+			zap.Int64("num_rejected", resp.PartialSuccess().RejectedDataPoints()),
```
dmitryax, line 123: I think we should be consistent with other log messages where we use `dropped_data_points`, `dropped_spans` and `dropped_records` as the key. We can use the `rejected` term, but so far, it's been used in logs for retryable errors propagated to the client, and `dropped` is used for permanent errors.

jmacd: I'd like us to find consistent terminology. From my review in https://github.com/open-telemetry/oteps/pull/238, I find that "dropped" most consistently refers to data that was never send to the downstream service. The term "rejected" is already standardized in the PartialSuccess message.

I would not like us to continue using "dropped" to refer to data that was sent and failed. That data "failed" after being sent to a consumer, so it wasn't dropped. There's even a chance the consumer successfully received that data, but that a timeout occurred, leaving the situation unclear, with success on the server but timeout at the client. I would like "dropped" to refer to data that was not transmitted, because it is helpful in constructing pipeline equations for monitoring. 

The logic I want to enforce is that whatever is received is either sent or dropped. If the data was sent, it can fail or not, but it wasn't dropped until after it was sent (and the receiver could have processed it anyway).

codeboten: Since the error before this change was using rejected, and the current change only prefixes this w/ `num_`, could we move the discussion on standardizing terminology to a separate issue? @dmitryax

dmitryax: From https://github.com/open-telemetry/oteps/pull/238, it's still not very clear how to distinguish permanent failures from retriable. We use `dropped` and `rejected` in the collector accordingly, even though it wasn't applied consistently.

In this particular case, the `rejected` term makes sense from the standpoint of a server sending this response, in my opinion. But in the collector, we drop that portion of the data.

Otherwise, the PR looks good to me. I'm not going to block it just because of the zap log key.

jmacd: I've updated this PR to be consistent as requested by @dmitryax. However, I'll note that exporterhelper is not fully consistent with itself. When data drops after retry attempts, it is `dropped_items` independent of the signal type, whereas the other type of failure has three specific keys.

I'll keep working to clarify https://github.com/open-telemetry/oteps/pull/238, thanks for the input. I don't agree with the categorization of "rejected" and "dropped", but we can discuss it there. Is there any more documentation on this choice of words? Even as I read the code for exporterhelper, I can't quite make sense of it. I see one log message from the baseExporter,
```
	if err != nil {
		be.set.Logger.Error("Exporting failed. Rejecting data."+be.exportFailureMessage,
			zap.Error(err), zap.Int("rejected_items", req.ItemsCount()))
	}
```
one for the queueSender:
```
		consumeErrHandler := func(err error, req Request) {
			o.set.Logger.Error("Exporting failed. Dropping data."+o.exportFailureMessage,
				zap.Error(err), zap.Int("dropped_items", req.ItemsCount()))
		}
```
and one for the converter
```
			set.Logger.Error("Failed to convert logs. Dropping data.",
				zap.Int("dropped_log_records", ld.LogRecordCount()),
				zap.Error(err))
```
What do you call it when a sampler drops data? How is that the same / not the same as these cases? I see the sampler-dropping case as similar in outcome to the converter failure -- you couldn't convert the data so you never sent it -- the coinflip was negative so you never sent it. I see the sampler-dropping case as dissimilar to the queueSender and baseExporter outcome, because the data was sent.

My position is that metrics should only count after the final outcome is known, so including attempts to retry, meaning that retries only delay the outcome.  I only count outcomes once per point/span/record. Therefore, retries are not visible in pipeline metrics -- they will be visible in transport-specific tracing and metrics instrumentation. So, e.g., you'll see two RPCs if there is one retry, you'll see twice as many bytes sent if you have to send a request twice due to retries, and so on.

exporter/otlphttpexporter/otlp.go
```diff
@@ -270,12 +270,17 @@ func tracesPartialSuccessHandler(protoBytes []byte, contentType string) error {
 	}
 	partialSuccess := exportResponse.PartialSuccess()
 	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedSpans() == 0) {
-		return consumererror.NewPermanent(fmt.Errorf("OTLP partial success: %s (%d rejected)", partialSuccess.ErrorMessage(), partialSuccess.RejectedSpans()))
+		// TODO: These should be counted as "rejected", similar to dropped items.
+		// https://github.com/open-telemetry/opentelemetry-collector/issues/9243
+		e.logger.Warn("partial success",
+			zap.String("message", exportResponse.PartialSuccess().ErrorMessage()),
+			zap.Int64("num_rejected", exportResponse.PartialSuccess().RejectedSpans()),
```
dmitryax, line 277: These should be updated as well right?

exporter/otlpexporter/otlp.go
dmitryax: "Exporting failed. Rejecting data" is logged with the error returned to the caller. So it can be retried potentially.

"Exporting failed. Dropping data." in the queueSender is logged on the data consumed from the queue asynchronously, so there is no way to retry them.

"Failed to convert logs. Dropping data." is returned with the permanent error, so they should not be retried.

exporter/otlphttpexporter/otlp.go
jmacd: I'm pretty confused.  The commit https://github.com/open-telemetry/opentelemetry-collector/pull/9260/commits/7920e791f80f8b7658cc00b9616e5f791307e021 shows I changed these lines, followed by a clean merge commit, and my local copy was different than what you saw in the review.

I rebased and force-pushed, now the fix I had intended appears.

jmacd: I was confused between the two otlp.go files (the one HTTP and the other gRPC).

jmacd:
Test failure looks unrelated:

```
Error: ./factory_test.go:142:3: unknown field GRPCClientSettings in struct literal of type otlpexporter.Config
Error: ./traces_test.go:39:3: unknown field GRPCClientSettings in struct literal of type otlpexporter.Config
```

dmitryax (approved):

Issue #9195, opened 2023-12-29
Add documentation or logging reminding people to enable configured extensions


swar8080:
**Is your feature request related to a problem? Please describe.**
People (like myself) might forget to enable an extension after configuring it, like below:

```
extensions:
  health_check:

service:
 extensions: [health_check]
```

I don't think it's intuitive why you'd want to configure an extension but not enable it? Also, a lot of extensions' documentation omit the `service.extensions` part of their example yaml config, making people think they just need the `extensions` part.   

**Describe the solution you'd like**
- Update the READMEs to include the `service.extensions` yaml as a reminder
- Maybe add a warning/debug log to the collector if any extension is configured but not enabled. Possibly this is implemented by comparing the configured extensions in `otelcol.Config.Extensions` to the enabled extensions in `otelcol.Config.Service.Extensions` immediately after validating the config

**Additional context**
I can make any of the above changes

atoulme:
You can say the same thing about exporters, receivers, processors or connectors. Just because they’re configured doesn’t guarantee they are going to be used in pipelines.

This is to allow a separation of concerns between configuration and execution I guess.

You could create a separate command to check if all components are used as a dry run validation.

swar8080:
> You can say the same thing about exporters, receivers, processors or connectors. Just because they’re configured doesn’t guarantee they are going to be used in pipelines. This is to allow a separation of concerns between configuration and execution I guess.

I focused on extensions because they're most likely to accidentally configure but then not enable. Pipelines are front and centre in the documentation and most tutorials, so people are more likely to know they have to pick the pipelines for the other components. Since extensions aren't added to a pipeline it's less obvious they need to be enabled separately under `services`

> You could create a separate command to check if all components are used as a dry run validation.

In my case, I was just copy-pasting incomplete yaml from extensions' READMEs, so I likely wouldn't know that this command exists 

So I can go ahead and just update the READMEs for now, but lmk if you think the originally suggested code change makes sense too

PR #9266, opened 2024-01-11
Update extension READMEs to remind people to enable extensions after configuring them

swar8080:
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/9195

### DIFF
--- a/extension/ballastextension/README.md
+++ b/extension/ballastextension/README.md
@@ -53,6 +53,9 @@ Config that uses 64 Mib of memory for the ballast:
 extensions:
   memory_ballast:
     size_mib: 64
+
+service:
+  extensions: [memory_ballast]
 ```
 
 Config that uses 20% of the total memory for the ballast:
--- a/extension/zpagesextension/README.md
+++ b/extension/zpagesextension/README.md
@@ -29,6 +29,9 @@ Example:
 ```yaml
 extensions:
   zpages:
+
+service:
+  extensions: [zpages]
 ```
 
 The full list of settings exposed for this exporter are documented [here](./config.go)


extension/ballastextension/README.md
```diff
@@ -53,6 +53,9 @@ Config that uses 64 Mib of memory for the ballast:
 extensions:
   memory_ballast:
     size_mib: 64
+
+service:
+  extensions: [memory_ballast]
```
mx-psi, line 58: We don't generally have this for any other component, e.g. see https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlpexporter

I would suggest removing this:

```suggestion
```

And add some wording like:

> Note that you need to **add the extension to the `services::extensions` list** for it to be enabled

instead. Do you think that is sufficiently clear?

swar8080: Either approach sounds good, but this is helpful for lazy developers that just want to copy/paste some yaml for the extension to work. They might miss that extra note and then waste time troubleshooting like I did. It's not intuitive that configuring an extension doesn't enable it

mx-psi: > Either approach sounds good, but this is helpful for lazy developers that just want to copy/paste some yaml for the extension to work

I understand, it just feels weird to do this only for extensions. All components in the OpenTelemetry Collector have this property, and we are not documenting this for any other kind of component. Why are extensions special in this sense?

>  They might miss that extra note and then waste time troubleshooting like I did.

I feel like the way to do this is with a warning, strict mode or some other kind of automated validation

> It's not intuitive that configuring an extension doesn't enable it

Changing the default behavior on this would be a big breaking change. If you have suggestions on a backwards-compatible way of doing this, a migration plan or a tool to help with this, I am happy to read a proposal, but in general this is hard to change today.

swar8080: The other triager in https://github.com/open-telemetry/opentelemetry-collector/issues/9195 also didn't think my approach was the right one, so all good to close this suggestion

> I understand, it just feels weird to do this only for extensions. All components in the OpenTelemetry Collector have this property, and we are not documenting this for any other kind of component. Why are extensions special in this sense?

I think it's more intuitive for other components since you need to pick a pipeline, and pipelines are a more well-understood concept than extensions. 

Extensions were originally listed under services to control the order they are started and stopped, according to the README. However, I don't think the order matters for most if not all the extensions, so people are less likely to think about that compared to pipelines

>I feel like the way to do this is with a warning, strict mode or some other kind of automated validation
I think the lazy developer would also miss the documentation for strict mode :)

>Changing the default behavior on this would be a big breaking change. If you have suggestions on a backwards-compatible way of doing this, a migration plan or a tool to help with this, I am happy to read a proposal, but in general this is hard to change today.

Yeah this would probably be overkill to change, but one idea would be to fallback to the order that extensions are configured if you forget to set the order under `services`

mx-psi: > The other triager in https://github.com/open-telemetry/opentelemetry-collector/issues/9195 also didn't think my approach was the right one, so all good to close this suggestion

We can wait to see what others think, but in its current state I think it would leave our documentation in an inconsistent state and I would like to avoid that.

> Extensions were originally listed under services to control the order they are started and stopped, according to the README. However, I don't think the order matters for most if not all the extensions, so people are less likely to think about that compared to pipelines
> 
> > I feel like the way to do this is with a warning, strict mode or some other kind of automated validation
> 
> I think the lazy developer would also miss the documentation for strict mode :). A warning log on startup would be helpful

:+1: I think a warning is likely to be less controversial than this.

> fallback to the order that extensions are configured if you forget to set the order under services

YAML keys [don't have a defined order](https://yaml.org/spec/1.2.2/#3221-mapping-key-order), so I don't think this is a good idea (for example, AIUI intermediate steps from the YAML the user writes until it gets to the Collector are free to change the order of the keys per the YAML spec). Also in practical terms, any Go YAML parser out there (at least that I know of) will map this to a Go map, which also does not preserve the order of the keys, so we would need to write some custom code and custom structure to save the map (which is both a lot of work and would need us to rethink the API we offer users).

PR #9267, opened 2024-01-11
Mark configopaque change as a breaking change

mx-psi:
**Description:** 

Follow up to #9214.

I think marking this as breaking is warranted since we have found 6 instances of breakage in contrib (fixed by open-telemetry/opentelemetry-collector-contrib#30304, open-telemetry/opentelemetry-collector-contrib#30402 and open-telemetry/opentelemetry-collector-contrib#30405).

I am also:

- Adding subtext explicitly stating how to fix this
- Adding this to the user changelog, since it has user-facing implications.

### DIFF
--- a/.chloggen/configopaque_stringer.yaml
+++ b/.chloggen/configopaque_stringer.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: configopaque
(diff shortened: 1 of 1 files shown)


mx-psi:
cc @TylerHelmuth

TylerHelmuth (approved):

atoulme (approved):

.chloggen/configopaque_stringer.yaml
```diff
@@ -15,11 +15,13 @@ issues: [9213]
...
+  Explicitly cast the opaque string to a string before using it in `fmt.Sprintf` to restore the previous behavior.
 
 # Optional: The change log or logs in which this entry should be included.
 # e.g. '[user]' or '[user, api]'
 # Include 'user' if the change is relevant to end users.
 # Include 'api' if there is a change to a library API.
 # Default: '[user]'
-change_logs: [api]
\ No newline at end of file
+change_logs: [user, api]
```
codeboten, line 27: Can you elaborate on what the user impact is? It's not clear to me

mx-psi: I was thinking it could impact logs that the users see or other printing information. It also may just help users see this quickly if they find a bug related to this. I am fine taking this out if you don't think it's a good idea/the logs thing does not really count as 'user-facing'

bogdandrutu (approved):

.chloggen/configopaque_stringer.yaml
codeboten: i think it's fine to leave it in, i was more curious if there was a wider impact to users other than seeing things as redacted in their logs 👍🏻

PR #9268, opened 2024-01-11
[TEST] See if contrib tests break with a panic on configopaque fmt.Stringer impl

mx-psi:
**Description:** A quick way to test implications of #9214

### DIFF
--- a/config/configopaque/opaque.go
+++ b/config/configopaque/opaque.go
@@ -23,5 +23,6 @@ func (s String) MarshalText() ([]byte, error) {
 }
 
 func (s String) String() string {
-	panic("testing if this is called somewhere else")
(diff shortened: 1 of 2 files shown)


mx-psi:
From https://pkg.go.dev/fmt#hdr-Format_errors:

> If an Error or String method triggers a panic when called by a print routine, the fmt package reformats the error message from the panic, decorating it with an indication that it came through the fmt package. For example, if a String method calls panic("bad"), the resulting formatted message will look like

:upside_down_face: I am going to run the panic on a separate goroutine, since this way it can't be caught by `recover`

mx-psi:
Alright, contrib tests are passing so it seems like there are no other usages of `configopaque.String` `String` method anywhere in contrib tests. That makes me a bit more certain that we have caught all usages

Issue #9219, opened 2024-01-04
Terminal errors are not being logged by the expoterhelper in case queued_retry is not used


rapphil:
**Describe the bug**
Terminal errors are not being logged by the expoterhelper in case queued_retry is not used. This was noticed in the prometheusremotewrite exporter starting with collector v0.85.0.

The prometheusremotewrite exporter is a snowflake component because it explicitly disable the queued_retry during component initialization and instead uses its own retry mechanism. This has to be done to avoid out of order samples. However we still want log errors returned by the `ComsumeMetricsFunc` of this component. 

We strongly believe that this is a regression caused by #8369.

**Steps to reproduce**
Use the prometheusremotewrite exporter and try to point to a endpoint that does not exists.

**What did you expect to see?**
I'd expect an error message to be published (example from v0.84.0):
```
2024-01-04T18:02:40.249Z      error   exporterhelper/queued_retry.go:364      Exporting failed. Try enabling retry_on_failure config option to retry on retryable errors      {"kind": "exporter", "data_type": "metrics", "name": "prometheusremotewrite", "error": "Permanent error: Permanent error: Permanent error: remote write returned HTTP status 404 Not Found; err = %!w(<nil>): <html>\r\n<head><title>404 Not Found</title></head>..........<center>openresty</center>\r\n</body>\r\n</html>\r\n"}]}
d101670931-collector-1          | go.opentelemetry.io/collector/exporter/exporterhelper.(*retrySender).send
d101670931-collector-1          |       go.opentelemetry.io/collector/exporter@v0.84.0/exporterhelper/queued_retry.go:364
d101670931-collector-1          | go.opentelemetry.io/collector/exporter/exporterhelper.(*metricsSenderWithObservability).send
d101670931-collector-1          |       go.opentelemetry.io/collector/exporter@v0.84.0/exporterhelper/metrics.go:191
d101670931-collector-1          | go.opentelemetry.io/collector/exporter/exporterhelper.(*queuedRetrySender).start.func1
d101670931-collector-1          |       go.opentelemetry.io/collector/exporter@v0.84.0/exporterhelper/queued_retry.go:202
d101670931-collector-1          | go.opentelemetry.io/collector/exporter/exporterhelper/internal.(*boundedMemoryQueue).StartConsumers.func1
d101670931-collector-1          |       go.opentelemetry.io/collector/exporter@v0.84.0/exporterhelper/internal/bounded_memory_queue.go:47
[...]

PR #9270, opened 2024-01-12
[exporterhelper] Log errors even if retry is not used

bryan-aguilar:
**Description:** Fixes regression of #8792. Test was changed in #8862 to use `WithRetry()` which removed the assertion that an error is still logged. 

**Link to tracking Issue:** #9219


dmitryax:
It'll result in double reporting non-permanent because of [this](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/queue_sender.go#L124). The error reporting is in a messy state right now, I was planning to make it better. Let me think though this a bit

bryan-aguilar:
Closing in favor of #9282

Issue #6945, opened 2023-01-13
Allow otlphttpexporter to export in JSON format


songy23:
**Is your feature request related to a problem? Please describe.**

Currently the otlphttpexporter only supports exporting data in Binary Protobuf Encoding: https://github.com/open-telemetry/opentelemetry-collector/blob/f2a0133bf677b9fe2e18b7c02eb1b6bf5e849999/exporter/otlphttpexporter/otlp.go#L132
While the otlp spec and the otlp http receiver support both binary protobuf encoding and JSON encoding. It will be nice to be able to export telemetry data in JSON using the otlphttpexporter.

**Describe the solution you'd like**

Add a config to otlphttpexporter like:
```
exporters:
  otlphttp:
    ...
    encoding: [binary_protobuf | json]
```
And have the corresponding encoder in otlphttpexporter.

Feel free to mark this as duplicated if there's an existing issue on otlphttpexporter json support.

bogdandrutu:
We want to discourage use of JSON between collectors (unnecessary overhead). We do accept receiving because JS libraries need that.

vainiusd:
This was also mentioned in opentelemetry-collector-contrib [issue 10836](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/10836).

It could be useful when sending logs to collectors that do not support OTLP.

aubm:
Just to mention that I've been trying to send logs collected by the OpenTelemetry Collector to Logstash and it looks like support for Protobuf is limited and I can't get it working properly.

I've just opened an issue on the Github Logstash Protobuf codec plugin repository.
You can find all the details here: https://github.com/logstash-plugins/logstash-codec-protobuf/issues/70

At this point, if I don't find a way to make it work and I can't get otlphttpexporter to offer JSON encoding, which is well supported by Logstash, I'll have to implement my own exporter.
This is a bummer, and probably an obstacle for more people to adopt it as well.

Just give me the go and I'll submit a PR.

atoulme:
You could try to use the splunkhec exporter in raw mode to push data over to logstash, that's not really meant for it but might get you what you need. If you can push a test showing the problems with logstash, I'm sure folks here can help on their end.

aubm:
Hi @tvaintrob, I ended up starting working on a generic HTTP/JSON exporter.
I might submit a PR to contribute it back eventually if anyone is interested.

[...]

PR #9276, opened 2024-01-12
[exporter/otlphttp] added support for configurable telemetry encoding

tvaintrob:
**Description:**
This PR adds support for encoding configuration in the `otlphttp` exporter.

**Link to tracking Issue:** 
#6945

**Testing:** 
Updated existing tests, and added relevant tests 

**Documentation:** 
Updated the `otlphttp` docs to include the new configuration option.


exporter/otlphttpexporter/otlp_test.go
```diff
@@ -600,6 +597,128 @@ func TestPartialSuccess_metrics(t *testing.T) {
...
+				err = exp.Start(context.Background(), componenttest.NewNopHost())
+				require.NoError(t, err)
+				t.Cleanup(func() {
+					require.NoError(t, exp.Shutdown(context.Background()))
+				})
+
+				// generate data
+				traces := ptrace.NewTraces()
+				err = exp.ConsumeTraces(context.Background(), traces)
+				require.NoError(t, err)
```
songy23, line 652: Can you add tests on malformed data? Same for metrics and logs

songy23 (commented):
Thanks @tvaintrob please also add a changelog

bogdandrutu:
What is the use-case for this? I remember that we intentionally did not want to offer this, because it is just a slower version of the proto and you don't have the dependency issue like you have on the client side (web) to  avoid depending on protobuf.

tvaintrob:
@bogdandrutu in my case we have a WAF that we want to send messages through but it needs to do a deep validation of the messages being passed, unfortunately it doesn't support reading binary data so we need to send the messages using JSON

exporter/otlphttpexporter/otlp_test.go
tvaintrob: what do you mean by malformed data? invalid `encoding` field?

tvaintrob:
@songy23 fixed all comments except for the additional tests

exporter/otlphttpexporter/otlp_test.go
songy23: I mean invalid proto or json text data, i.e. when err is not nil.

songy23 (commented):
Please fix the failed tests https://github.com/open-telemetry/opentelemetry-collector/actions/runs/7513270300/job/20574065081?pr=9276

songy23:
We can mention in the documentation that json should only be used in endpoints that can only accept json, but discouraged in other use cases, e.g. collector-to-collector case. Does that sound good? @bogdandrutu

TylerHelmuth:
I agree with supporting `http/json`. Although it isn't performant, it feels correct to implement all parts of the OTLP specification for the exporter.

exporter/otlphttpexporter/otlp_test.go
songy23: Chatted offline, with the current setup it's difficult to generate data that cannot be parsed, so mark this as resolved.

exporter/otlphttpexporter/config_test.go
```diff
@@ -51,6 +51,7 @@ func TestUnmarshalConfig(t *testing.T) {
 				NumConsumers: 2,
 				QueueSize:    10,
 			},
+			Encoding: EncodingType(""),
```
songy23, line 54: Why is this line needed?

exporter/otlphttpexporter/factory_test.go
```diff
@@ -36,6 +36,7 @@ func TestCreateDefaultConfig(t *testing.T) {
 	assert.Equal(t, ocfg.RetryConfig.MaxInterval, 30*time.Second, "default retry MaxInterval")
 	assert.Equal(t, ocfg.QueueConfig.Enabled, true, "default sending queue is enabled")
 	assert.Equal(t, ocfg.Compression, configcompression.Gzip)
+	assert.Equal(t, ocfg.Encoding, EncodingType(""))
```
songy23, line 39: Need to update `factory.CreateDefaultConfig()` to set default encoding to EncodingProto

tvaintrob: This broke the tests from contrib, and because the default handler is already proto it seemed fine to leave empty

exporter/otlphttpexporter/factory.go
```diff
@@ -36,7 +36,6 @@ func createDefaultConfig() component.Config {
 	return &Config{
 		RetryConfig: configretry.NewDefaultBackOffConfig(),
 		QueueConfig: exporterhelper.NewDefaultQueueSettings(),
-		Encoding:    EncodingProto,
```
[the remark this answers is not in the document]
tvaintrob: How do I go about doing it? The contrib repo doesn't have a reference to the updated config struct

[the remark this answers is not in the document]
songy23: OK, let's leave a TODO here and update the contrib tests when the PR is merged and the core dependency is updated in contrib. Default should be made explicit IMO.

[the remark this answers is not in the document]
songy23: Discussed offline, there's a circular dependency between the otlphttpexporter config and a test in f5cloudexporter https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30703. So for now the default cannot be changed.

songy23 (dismissed):
LGTM

cc @open-telemetry/collector-approvers PTAL

exporter/otlphttpexporter/config.go
```diff
@@ -12,6 +12,14 @@ import (
 	"go.opentelemetry.io/collector/exporter/exporterhelper"
 )
 
+// EncodingType defines the type for content encoding
+type EncodingType string
```
TylerHelmuth, line 16: Should we implement `UnmarshalText` to force this type to be either `proto` or `json` when used in a config?

exporter/otlphttpexporter/config.go
```diff
@@ -26,6 +34,9 @@ type Config struct {
 
 	// The URL to send logs to. If omitted the Endpoint + "/v1/logs" will be used.
 	LogsEndpoint string `mapstructure:"logs_endpoint"`
+
+	// The encoding to export telemetry (default: "proto")
+	Encoding EncodingType `mapstructure:"encoding"`
```
TylerHelmuth, line 39: I think we should have some validation, either as part of `Validate` or `UnmarshalText`, that enforces only `proto` or `json`.

exporter/otlphttpexporter/factory_test.go
TylerHelmuth: I agree that `factory.CreateDefaultConfig()` needs updated to set `Encoding: EncodingProto`.  What broke in contrib?

exporter/otlphttpexporter/otlp.go
```diff
@@ -87,7 +88,18 @@ func (e *baseExporter) start(_ context.Context, host component.Host) error {
...
-	request, err := tr.MarshalProto()
+
+	var err error
+	var request []byte
+	switch e.config.Encoding {
+	case EncodingJSON:
+		request, err = tr.MarshalJSON()
+	case EncodingProto:
+		request, err = tr.MarshalProto()
+	default:
```
TylerHelmuth, line 99: Since we have explicit cases for the encodings we handle I think this should error instead of fallback to proto.

exporter/otlphttpexporter/otlp.go
```diff
@@ -97,7 +109,18 @@ func (e *baseExporter) pushTraces(ctx context.Context, td ptrace.Traces) error {
...
```
TylerHelmuth, line 120: Since we have explicit cases for the encodings we handle I think this should error instead of fallback to proto.

exporter/otlphttpexporter/otlp.go
```diff
@@ -106,7 +129,18 @@ func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) erro
...
```
TylerHelmuth, line 140: Since we have explicit cases for the encodings we handle I think this should error instead of fallback to proto.

exporter/otlphttpexporter/otlp.go
```diff
@@ -120,7 +154,16 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
...
 		return consumererror.NewPermanent(err)
 	}
-	req.Header.Set("Content-Type", protobufContentType)
+
+	switch e.config.Encoding {
+	case EncodingJSON:
+		req.Header.Set("Content-Type", jsonContentType)
+	case EncodingProto:
+		req.Header.Set("Content-Type", protobufContentType)
+	default:
```
TylerHelmuth, line 163: Since we have explicit cases for the encodings we handle I think this should error instead of fallback to proto.

exporter/otlphttpexporter/otlp.go
```diff
@@ -259,47 +300,74 @@ func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler par
...
+	if contentType == protobufContentType {
+		err := exportResponse.UnmarshalProto(responseBytes)
+		if err != nil {
+			return fmt.Errorf("error parsing protobuf response: %w", err)
+		}
+	} else if contentType == jsonContentType {
+		err := exportResponse.UnmarshalJSON(responseBytes)
+		if err != nil {
+			return fmt.Errorf("error parsing json response: %w", err)
+		}
```
TylerHelmuth, line 318: I think this could also be a switch statement that errors if one of the expected content types is not received.

exporter/otlphttpexporter/otlp.go
```diff
@@ -259,47 +300,74 @@ func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler par
...
```
TylerHelmuth, line 343: I think this could also be a switch statement that errors if one of the expected content types is not received.

exporter/otlphttpexporter/otlp.go
```diff
@@ -259,47 +300,74 @@ func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler par
...
```
TylerHelmuth, line 368: I think this could also be a switch statement that errors if one of the expected content types is not received.

exporter/otlphttpexporter/factory_test.go
tvaintrob: The `f5cloudexporter` broke because one of the tests explicitly creates an instance of the config as part of a test, the new field caused it to mismatch with the asserted one

See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30703

exporter/otlphttpexporter/otlp.go
tvaintrob: Since there is an issue with adding the default value explicitly (breaking contrib) I cant change this, unless breaking contrib is acceptable (I can create a small fix in the `f5cloudexporter` once this change is merged)

exporter/otlphttpexporter/otlp.go
tvaintrob: Since there is an issue with adding the default value explicitly (breaking contrib) I cant change this, unless breaking contrib is acceptable (I can create a small fix in the `f5cloudexporter` once this change is merged)

exporter/otlphttpexporter/otlp.go
tvaintrob: Since there is an issue with adding the default value explicitly (breaking contrib) I cant change this, unless breaking contrib is acceptable (I can create a small fix in the `f5cloudexporter` once this change is merged)

exporter/otlphttpexporter/otlp.go
tvaintrob: Since there is an issue with adding the default value explicitly (breaking contrib) I cant change this, unless breaking contrib is acceptable (I can create a small fix in the `f5cloudexporter` once this change is merged)

exporter/otlphttpexporter/otlp.go
tvaintrob: i changed to a switch statement but left previous behaviour, before this PR if the content-type was not proto a `nil` was returned

exporter/otlphttpexporter/otlp.go
tvaintrob: i changed to a switch statement but left previous behaviour, before this PR if the content-type was not proto a `nil` was returned

exporter/otlphttpexporter/otlp.go
tvaintrob: i changed to a switch statement but left previous behaviour, before this PR if the content-type was not proto a `nil` was returned

tvaintrob:
@TylerHelmuth updated the PR with the requested changes, not all are possible due to the issues with contrib

Thanks

exporter/otlphttpexporter/otlp.go
TylerHelmuth: Can you share more about what breaks in contrib? Can we go fix it now and then merge this PR?

exporter/otlphttpexporter/otlp.go
```diff
@@ -301,21 +301,20 @@ func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler par
...
 			return fmt.Errorf("error parsing protobuf response: %w", err)
 		}
-	} else if contentType == jsonContentType {
+	case jsonContentType:
 		err := exportResponse.UnmarshalJSON(responseBytes)
 		if err != nil {
 			return fmt.Errorf("error parsing json response: %w", err)
 		}
+	default:
+		return nil
```
TylerHelmuth, line 317: I agree that returning nil was the previous default behavior, but I think that was incorrect.  I am ok taking care of that in a separate PR tho, as long as we create an issue to track it.

exporter/otlphttpexporter/config.go
```diff
@@ -20,6 +22,37 @@ const (
 	EncodingJSON  EncodingType = "json"
 )
 
+var (
+	_ encoding.TextMarshaler   = (*EncodingType)(nil)
+	_ encoding.TextUnmarshaler = (*EncodingType)(nil)
+)
+
+// MarshalText marshals EncodingType to text.
+func (e EncodingType) MarshalText() (text []byte, err error) {
```
TylerHelmuth, line 31: I believe we can get by without adding `MarshalText`, reducing our public API.

TylerHelmuth (commented):
Thanks for opening https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30703.  I'd say we fix that before merging this PR.

songy23:
> Thanks for opening https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30703. I'd say we fix that before merging this PR.

Turns out to be straightforward to fix it, I have https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30759

exporter/otlphttpexporter/factory_test.go
songy23: The test in f5cloudexporter is fixed, you should be able to update `factory.CreateDefaultConfig()` now

tvaintrob:
Thanks for the update @songy23 ill update this one to include the default encoding

exporter/otlphttpexporter/config.go
```diff
@@ -4,14 +4,45 @@
...
+type EncodingType string
+
+const (
+	EncodingProto EncodingType = "proto"
+	EncodingJSON  EncodingType = "json"
+)
+
+var _ encoding.TextUnmarshaler = (*EncodingType)(nil)
+
+// UnmarshalText unmarshalls text to a Level.
```
songy23, line 27:
```suggestion
// UnmarshalText unmarshalls text to an EncodingType.
```

exporter/otlphttpexporter/otlp.go
songy23: I think this should return an error with unsupported content type now?

exporter/otlphttpexporter/otlp.go
```diff
@@ -259,47 +300,71 @@ func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler par
...
+		if err != nil {
+			return fmt.Errorf("error parsing protobuf response: %w", err)
+		}
+	case jsonContentType:
+		err := exportResponse.UnmarshalJSON(responseBytes)
+		if err != nil {
+			return fmt.Errorf("error parsing json response: %w", err)
+		}
+	default:
+		return nil
```
songy23, line 341: Same here

exporter/otlphttpexporter/otlp.go
```diff
@@ -259,47 +300,71 @@ func handlePartialSuccessResponse(resp *http.Response, partialSuccessHandler par
...
```
songy23, line 365: Same here

songy23 (commented):
Minor comments, LGTM otherwise

exporter/otlphttpexporter/otlp.go
TylerHelmuth: I'm definitely ok with updating this to return an error as part of this PR.

TylerHelmuth (commented):
@tvaintrob please add some test cases that hit the new error scenarios and a test for the new UnmarshalText function.

tvaintrob:
@TylerHelmuth @songy23 updated the PR, i rather take care of the extra error cases in a different PR, feels out of scope for this one

songy23 (approved):
LGTM, please file an issue to return error on invalid content-type in requests.

tvaintrob:
> LGTM, please file an issue to return error on invalid content-type in requests.

https://github.com/open-telemetry/opentelemetry-collector/issues/9413 created an issue for that, thanks

tvaintrob:
@TylerHelmuth just waiting on your approval to merge this one

Thanks!

TylerHelmuth (commented):
@tvaintrob could you add a could more test scenarios to cover the error cases `Codecov` is complaining about?

tvaintrob:
> @tvaintrob could you add a could more test scenarios to cover the error cases `Codecov` is complaining about?

I had a talk with @songy23 about it, he said that it's ok to leave it like that, do you have any example on how to generate malformed data?

TylerHelmuth:
I think you'd have to call the private functions directly. I am ok moving forward without these tests since they cover such unlikely, simple scenarios

sarathsp06 (approved):
LGTM

mx-psi:
@tvaintrob please fix the merge conflict and I can merge afterwards :)

tvaintrob:
@mx-psi Hi, resolved the merge conflict, thanks!

mx-psi (approved):

PR #9282, opened 2024-01-13
[exporterhelper] Cleanup logging for export failures

dmitryax:
This change makes the logging of the exported error failures cleaner.
1. Ensure an error message is logged every time and only once when data is dropped/rejected due to export failure.
2. Update the wording. Specifically, don't use "dropped" term when an error is reported back to the pipeline. If there is no queue configured, the exporter doesn't drop data by itself but rather rejects it. Keep the "dropped" wording for failures after the enabled queue.
3. Properly report any error reported by a queue. For example, a persistent storage error must be reported as a storage error, not as "queue overflow".

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9219

### DIFF
--- a/.chloggen/exporter-helper-cleanup-error-logs.yaml
+++ b/.chloggen/exporter-helper-cleanup-error-logs.yaml
@@ -1,0 +1,27 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporters
+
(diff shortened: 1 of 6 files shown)


bryan-aguilar (commented):
Thank you for the fix @dmitryax. Can we add tests to verify that error logs are emitted if no retry option is added to queue? Example in https://github.com/open-telemetry/opentelemetry-collector/pull/9270

bogdandrutu (approved):

dmitryax:
@bryan-aguilar, sure. I added a test `TestQueueFailedRequestDropped`

PR #9295, opened 2024-01-16
[chore] Clarify configopaque expectations

### LINKED ISSUE #9274: Document `configopaque` security expectations
The aim of the `configopaque` package is to have opaque types to prevent leakage of secrets in configuration. We should document under what situations do we aim to redact output, since it's kind of fuzzy right now (it used to be just for marshaling, now we also redact logs). 

It may also be that we miss some cases, would we be justified in making breaking changes if something like this happens again?

mx-psi:
**Description:** 

Documents `configopaque` expectations.

The intention is that it is clear that we can add interface implementations after 1.0 if not doing so would leak opaque values.

**Link to tracking Issue:** Fixes #9274

### DIFF
--- a/config/configopaque/doc.go
+++ b/config/configopaque/doc.go
@@ -17,5 +17,5 @@
 //
 // If new interfaces that would leak opaque values are added to the standard library
 // or become widely used in the Go ecosystem, these will eventually be implemented
-// by configopaque.String as well.
+// by configopaque.String as well. This is not considered a breaking change.
 package configopaque // import "go.opentelemetry.io/collector/config/configopaque"
--- a/config/configopaque/doc.go
+++ b/config/configopaque/doc.go
@@ -1,10 +1,21 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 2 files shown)


mx-psi:
cc @TylerHelmuth

atoulme (approved):

config/configopaque/doc.go
```diff
@@ -1,10 +1,21 @@
...
+// The only way to view the value stored in a configopaque.String is to first convert
+// it to a string by casting with the builtin `string` function.
+//
+// To achieve this, configopaque.String implements standard library interfaces
+// like fmt.Stringer, encoding.TextMarshaler and others to ensure that the
+// underlying value is masked when printed or serialized.
+//
+// If new interfaces that would leak opaque values are added to the standard library
+// or become widely used in the Go ecosystem, these will eventually be implemented
+// by configopaque.String as well.
```
TylerHelmuth, line 20: Should we callout that implementing new interfaces is not considered a breaking change?

mx-psi: Makes sense, done

TylerHelmuth (approved):

codeboten (approved):

PR #9302, opened 2024-01-17
[otelcol] Add unit test for implicit config use case

yurishkuro:
**Description:** 
* Add unit test that illustrates running the command without an external config
* This models how Jaeger all-in-one uses collector because all-in-one can be run without arguments, for maximum user friendliness

### DIFF
--- a/otelcol/command_test.go
+++ b/otelcol/command_test.go
@@ -29,7 +29,7 @@ func TestNewCommandNoConfigURI(t *testing.T) {
 
 // This test emulates usage of Collector in Jaeger all-in-one, which
 // allows running the binary with no explicit configuration.
-func TestNewCommandImplicitConfig(t *testing.T) {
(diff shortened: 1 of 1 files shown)


otelcol/command_test.go
```diff
@@ -26,6 +27,26 @@ func TestNewCommandNoConfigURI(t *testing.T) {
 	require.Error(t, cmd.Execute())
 }
 
+// This test emulates usage of Collector in Jaeger all-in-one, which
+// allows running the binary with no explicit configuration.
+func TestNewCommandImplicitConfig(t *testing.T) {
```
bogdandrutu, line 32: Can we call this "ProgrammaticallyPassConfig"?

bogdandrutu (approved):

bogdandrutu (approved):

Issue #9300, opened 2024-01-17
[receiver/otlp] Unify error handling


mx-psi:
Since this error handling is the same for all signals I do wonder if it could be extracted and re-used, but that is an optimization we could handle later.

_Originally posted by @TylerHelmuth in https://github.com/open-telemetry/opentelemetry-collector/pull/8080#discussion_r1449180186_

PR #9307, opened 2024-01-17
[receiver/otlp] Refactor error handling into helper

TylerHelmuth:
Moves reused code into helper function

**Link to tracking Issue:**
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9300

**Testing:**
Added unit tests

### DIFF
--- a/receiver/otlpreceiver/internal/errors/errors.go
+++ b/receiver/otlpreceiver/internal/errors/errors.go
@@ -10,7 +10,7 @@ import (
 	"go.opentelemetry.io/collector/consumer/consumererror"
 )
 
-func GetStatusFromError(err error) *status.Status {
+func GetStatusFromError(err error) error {
 	s, ok := status.FromError(err)
 	if !ok {
 		// Default to a retryable error
@@ -21,5 +21,5 @@ func GetStatusFromError(err error) *status.Status {
 		}
 		s = status.New(code, err.Error())
 	}
-	return s
+	return s.Err()
 }
--- a/receiver/otlpreceiver/internal/errors/errors.go
+++ b/receiver/otlpreceiver/internal/errors/errors.go
@@ -13,6 +13,8 @@ import (
 func GetStatusFromError(err error) *status.Status {
 	s, ok := status.FromError(err)
 	if !ok {
+		// Default to a retryable error
+		// https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures
 		code := codes.Unavailable
 		if consumererror.IsPermanent(err) {
 			code = codes.InvalidArgument
--- a/receiver/otlpreceiver/internal/errors/errors.go
+++ b/receiver/otlpreceiver/internal/errors/errors.go
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package util // import "go.opentelemetry.io/collector/receiver/otlpreceiver/internal/util"
+package errors // import "go.opentelemetry.io/collector/receiver/otlpreceiver/internal/errors"
 
 import (
 	"google.golang.org/grpc/codes"
--- a/receiver/otlpreceiver/internal/logs/otlp.go
+++ b/receiver/otlpreceiver/internal/logs/otlp.go
@@ -8,7 +8,7 @@ import (
 
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/plog/plogotlp"
-	"go.opentelemetry.io/collector/receiver/otlpreceiver/internal/util"
+	"go.opentelemetry.io/collector/receiver/otlpreceiver/internal/errors"
 	"go.opentelemetry.io/collector/receiver/receiverhelper"
 )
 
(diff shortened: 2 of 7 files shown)


receiver/otlpreceiver/internal/util/errors.go
```diff
@@ -0,0 +1,23 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package util // import "go.opentelemetry.io/collector/receiver/otlpreceiver/internal/util"
```
mx-psi, line 4: nit: maybe rename to `package status`? IMO `util` is an antipattern but I don't have a good alternative in mind

mx-psi (approved):

receiver/otlpreceiver/internal/util/errors.go
TylerHelmuth: if we're ok sharing a name with the existing `status` package we depend on, should I call it `errors`? That was the original name I chose but didn't like that it overlapped with the `errors` package.

mx-psi: `errors` works for me as well, this is internal so we can always change it later as the package evolves

TylerHelmuth: Updated

receiver/otlpreceiver/internal/metrics/otlp.go
```diff
@@ -51,14 +48,7 @@ func (r *Receiver) Export(ctx context.Context, req pmetricotlp.ExportRequest) (p
...
-		s, ok := status.FromError(err)
-		if !ok {
-			code := codes.Unavailable
-			if consumererror.IsPermanent(err) {
-				code = codes.InvalidArgument
-			}
-			s = status.New(code, err.Error())
-		}
+		s := errors.GetStatusFromError(err)
 		return pmetricotlp.NewExportResponse(), s.Err()
```
codeboten, line 52:
```suggestion
		return pmetricotlp.NewExportResponse(), errors.GetStatusFromError(err).Err()
```

receiver/otlpreceiver/internal/trace/otlp.go
```diff
@@ -52,14 +49,7 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
...
-		s, ok := status.FromError(err)
-		if !ok {
-			code := codes.Unavailable
-			if consumererror.IsPermanent(err) {
-				code = codes.InvalidArgument
-			}
-			s = status.New(code, err.Error())
-		}
+		s := errors.GetStatusFromError(err)
 		return ptraceotlp.NewExportResponse(), s.Err()
```
codeboten, line 53:
```suggestion
		return ptraceotlp.NewExportResponse(), errors.GetStatusFromError(err).Err()
```

receiver/otlpreceiver/internal/logs/otlp.go
```diff
@@ -51,14 +48,7 @@ func (r *Receiver) Export(ctx context.Context, req plogotlp.ExportRequest) (plog
...
-		s, ok := status.FromError(err)
-		if !ok {
-			code := codes.Unavailable
-			if consumererror.IsPermanent(err) {
-				code = codes.InvalidArgument
-			}
-			s = status.New(code, err.Error())
-		}
+		s := errors.GetStatusFromError(err)
 		return plogotlp.NewExportResponse(), s.Err()
```
codeboten, line 52:
```suggestion
		return plogotlp.NewExportResponse(), errors.GetStatusFromError(err).Err()
```

codeboten (approved):

receiver/otlpreceiver/internal/logs/otlp.go
bogdandrutu: If this is the case, why not always return the error from GetStatusFromError?

receiver/otlpreceiver/internal/errors/errors.go
```diff
@@ -0,0 +1,23 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"go.opentelemetry.io/collector/consumer/consumererror"
+)
+
+func GetStatusFromError(err error) *status.Status {
+	s, ok := status.FromError(err)
+	if !ok {
+		code := codes.Unavailable
```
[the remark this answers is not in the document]
TylerHelmuth: I believe the original PR used `Unavailable` instead of `Unknown` because the default state was to return a retryable error unless the error was permanent, and `Unknown` is not retryable according to OTLP: https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures

receiver/otlpreceiver/internal/logs/otlp.go
TylerHelmuth: Returning the entire `Status` makes `GetStatusFromError` more flexible for future use, but since there is not ab actual use case yet, and this is an internal package, I am ok with `GetStatusFromError` returning an `error` that made from a `Status`.  

I don't feel strongly about this either way, so let me know what you all prefer.

receiver/otlpreceiver/internal/errors/errors.go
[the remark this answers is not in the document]
bogdandrutu: Then it makes sense to at leas comment this for others like me that will ask for same thing. Thanks for clarifying.

Then should we use "Unknown" for Permanent? Maybe a separate PR though.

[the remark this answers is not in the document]
TylerHelmuth: Added a comment but I agree that changes to the code can be its own PR.

receiver/otlpreceiver/internal/logs/otlp.go
bogdandrutu: Since this is an internal API I prefer to simplify usage, since everywhere we call `.Err()`

TylerHelmuth: Updated

Issue #9281, opened 2024-01-12
`go install` fails for `mdatagen`


atoulme:
Expected:
```
go install github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@latest
```
should install mdatagen under $GOBIN.

Actual:
```
go install github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@latest
go: github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@latest (in github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@v0.92.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

bogdandrutu:
Not sure if this is a real problem, because this seem to work in CI, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/build-and-test.yml#L120 and https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile#L81

atoulme:
It works in this repository, but not under a different folder. This means we can't use it for contrib.
I have amended the steps to reproduce.

cpheps:
We are also experiencing this when trying to install the mdatagen tool via `go install` for our distribution.

khushhboo:
Is there a fix for this yet? Trying to update some generated files in vcenterreceiver.

atoulme:
Weird, please make dure you are on latest main and run make generate.

khushhboo:
Issue persists, tried from inside a folder and from repository as well. Make generate fails.

Error displayed:
go: go.opentelemetry.io/collector/cmd/mdatagen@latest (in go.opentelemetry.io/collector/cmd/mdatagen@v0.97.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.

atoulme:
Please check that GOBIN is in your PATH. This might be the problem.

woody1872:
Also seeing this issue:

```
$ go install github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@latest
go: github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@latest (in github.com/open-telemetry/opentelemetry-collector/cmd/mdatagen@v0.99.0):
	The go.mod file for the module providing named packages contains one or
[...]

PR #9310, opened 2024-01-17
[mdatagen] update crosslink to avoid replace statements

codeboten:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9281

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -448,7 +448,7 @@ checklinks:
 .PHONY: crosslink
 crosslink: $(CROSSLINK)
 	@echo "Executing crosslink"
-	$(CROSSLINK) --root=$(shell pwd) --prune
+	$(CROSSLINK) --root=$(shell pwd) --prune --skip cmd/mdatagen/go.mod
 
 FILENAME?=$(shell git branch --show-current)
 .PHONY: chlog-new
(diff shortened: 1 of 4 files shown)


atoulme (approved):
I am ok with that fix, however, it means mdatagen will continue to import packages from this repository. Will this matter to mdatagen? Will it be updated separately?

codeboten:
> I am ok with that fix, however, it means mdatagen will continue to import packages from this repository. Will this matter to mdatagen? Will it be updated separately?

With this change, mdatagen will only used released versions of the packages from this repo. I don't know what a workaround this is if we want mdatagen to be installed via `go install`

codeboten:
@atoulme i suppose we could still use unreleased versions of the packages as long as they're explicitly stated in the go.mod file, without replace statements 🤷🏻

TylerHelmuth:
If we make a change to mdatagen, like say adding a new field to the readme headers, does this change prevent us from testing that change on one of core's components until after the release?

codeboten:
> If we make a change to mdatagen, like say adding a new field to the readme headers, does this change prevent us from testing that change on one of core's components until after the release?

It does not. Its just that if a change happens in one of the core components, it wont be automatically tested by mdatagen

TylerHelmuth (approved):

bogdandrutu (approved):

PR #9319, opened 2024-01-19
[service] Move TracerProvider initialization to service/telemetry package

mx-psi:
**Description:** 

Redo of #8171

Depends on #9131

**Link to tracking Issue:** Updates #8170

PR #9320, opened 2024-01-19
[service] Refactor telemetry initialization

mx-psi:
**Description:** 

A part of #9319 that can be done without waiting for #9131


codeboten (commented):
Thanks for cleaning this up. WDYT of adding a config_test.go file in the internal resource package?

mx-psi:
> Thanks for cleaning this up. WDYT of adding a config_test.go file in the internal resource package?

Done!

codeboten (approved):

PR #9321, opened 2024-01-20
[component] Move internal/sharedcomponent to component/sharedcomponent

### LINKED ISSUE #4266: Make SharedComponents public

atoulme:
**Description:**
Move internal/sharedcomponent to component/sharedcomponent.

**Link to tracking Issue:**
Fixes #4266


component/sharedcomponent/sharedcomponent.go
```diff
@@ -4,7 +4,7 @@
 // Package sharedcomponent exposes functionality for components
 // to register against a shared key, such as a configuration object, in order to be reused across signal types.
 // This is particularly useful when the component relies on a shared resource such as os.File or http.Server.
-package sharedcomponent // import "go.opentelemetry.io/collector/internal/sharedcomponent"
+package sharedcomponent // import "go.opentelemetry.io/collector/component/sharedcomponent"
```
bogdandrutu, line 7: As mentioned in my previous comment, I am not sure this is right now:
1. The fact that we need to pass TelemetrySettings feels a hack and I want to better look into why we need that. I do understand some use-case but maybe we have a different "SharedTelemetrySettings" for that. Also not sure why start/stop calls the TelemetrySettings that way vs we return the right status all the time for multiple calls.
2. Right now we stop first time, which may not be ideal. Ideally we should call start first time and stop last time, but this is not possible for grpc servers. But for other "shared" components that probably should be the case and an external mechanism block receiving/emitting data to match the Receiver/Processor behaviors.

bogdandrutu (commented):
Please start looking into the items listed before making this public.

component/sharedcomponent/sharedcomponent.go
atoulme: I can file issues for both items.

I'm not clear on 1, but I'll give it my best try.

I assumed for 2. we abandoned because that wasn't easily possible, but we can look again with your suggestion.

mx-psi (commented):
Just a sanity check: why do we need this for 1.0? Can't we postpone it?

component/sharedcomponent/sharedcomponent.go
atoulme: I have opened https://github.com/open-telemetry/opentelemetry-collector/pull/9527 to try and tackle 1.

atoulme:
Closing - I think we can release 1.0 without this package.

PR #9333, opened 2024-01-21
[chore][extension/zpages] move handlers inside zpageextension

atoulme:
**Description:**
Move zpageextension handlers that can execute with public host APIs and CreateSettings information inside the zpageextension code, reducing dependency on service package.

Given that this change doesn't affect any public API, I have declared it under a chore PR.

One function providing pipelines information remains only serviceable with access to internal service APIs, and can be moved subsequently.


extension/zpagesextension/internal/zpages/package_test.go
```diff
@@ -0,0 +1,14 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package zpages
```
bogdandrutu, line 4: `zpages` -> `templates`?

bogdandrutu (changes requested):
I understand that this makes sense, but I want to make sure we are not exposing new public APIs. I would like to better understand why is better to have the public API that exposes the internal graph vs having this logic here.

bogdandrutu:
>Hang on, I didn't touch the graph API in this PR. I just moved the other zpages functions that could be fulfilled from the zpages extension without access to internals.

I don't understand the motivation to have the "Extension" part move. I would understand that we do move the Trace/Metrics part, but I don't understand the Extension, there is no "logic" for me on why the code should be there. So if you convince me why (not just because it can :) ) I would be happy to move forward with this.

atoulme:
> I don't understand the motivation to have the "Extension" part move.

With those moves, I move all logic into the extension. It feels cleaner to me, since it all depends on public APIs from the collector, and it allows a better separation of concerns.
This is going to help stabilize the service package a bit, reduces the burden of maintenance - I can probably add a test or two for this function as part of the zpages extension, bump up coverage.

atoulme:
I don't know if I should divide up this PR in more cases or go for a big change with #9334 to make my intentions more apparent. I'll close this PR and concentrate on #9334.

PR #9334, opened 2024-01-21
Move all code related to zpages to the zpages extension

atoulme:
**Description:**
This PR moves all the code related to zpages logic into the zpages extension itself.
There is no zpage code in the service package.
The zpage code still relies on casting host to get a struct representing the current structure of the pipelines, represented as simple types.
All other reporting logic of the zpagesextension is rendered internally to the extension using public APIs of `Host`.


component/graph.go
```diff
@@ -0,0 +1,17 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package component // import "go.opentelemetry.io/collector/component"
+
+// Graph is a struct representing the pipelines
+// currently registered with the `Host`.
+type Graph struct {
```
bogdandrutu, line 8: Please don't expose any new public API about our internal graph/pipelines.

bogdandrutu (changes requested):
See the only main comment, which is we don't want to expose more public APIs about our internal representation.

component/graph.go
atoulme: OK, I'll expose it internally for now.

atoulme:
> See the only main comment, which is we don't want to expose more public APIs about our internal representation.

I have removed any new APIs showing our internal representation.

PR #9347, opened 2024-01-23
[chore] remove LogStabilityLevel, not used

atoulme:
Remove `LogStabilityLevel` as it is not used anywhere and is an internal struct.


dmitryax:
It's part of the component builder now https://github.com/open-telemetry/opentelemetry-collector/blob/0017761d4f8704a772f212a6ed0abc6e7448c2f4/exporter/exporter.go#L217

Issue #9337, opened 2024-01-22
Make memory_limiter and otlp_receiver return 503 status code instead of 500 on failure


climberfree:
**The problem**
When the collector fails due to memory_limiter, the agent receives a **500 status code**. This status code i**s not retryable**, and the error sometimes resolves within minutes, especially after creating one more collector's instance through autoscaling.

**The solution**
We propose **making the `memory_limiter` and `otlp_receiver` return a 503 status** code instead of 500 in situations where the failure is temporary and expected to be resolved shortly.

**Benefits:**
- makes the situation retryable, allowing the agent to attempt a new connection after a brief delay.
- aligns with common HTTP status code semantics, indicating that the service is temporarily unavailable.

dmitryax:
This should be already resolved and available in 0.93.0 with https://github.com/open-telemetry/opentelemetry-collector/pull/8080 merged

TylerHelmuth:
@dmitryax I think #8080 actually only fixed the problem when using grpc.  If using http, we still have these line in our code:
 
 https://github.com/open-telemetry/opentelemetry-collector/blob/d78b748d7ae84c623c386876945d1efcb87f076c/receiver/otlpreceiver/otlphttp.go#L105-L109
 
 With #8080 the value in `err` is now properly a `Status` with code `Unavailable`, but our `http` logic is still returning `500`.

Issue #8132, opened 2023-07-25
otlphttp exporter drops data because memory limiter (of another instance) is considered as non recoverable error


sumo-drosiek:
**Describe the bug**

Data are being dropped by exporter, because receiver is throttled by memory limiter. This is incorrect in my opinion as receiver may be behind load balancer, so after retry packet may be forwarded to healthy instance. Also memory usage is not constant, and may not be subject of limitation after retry

**Steps to reproduce**

Not sure how to reproduce, we observe it in our environemnt when receiver instance is memory pressured

**What did you expect to see?**

I expect that receiver will try to resend data after some time, or put it back to sending queue

**What did you see instead?**

I see that data is dropped

**What version did you use?**

`v0.81.0`

**What config did you use?**

```
exporters:                                                                                                                                                                                                                                                                                                                                                                                                                            
  otlphttp:                                                                                                                                                                                                                                                                                                                                                                                                                           
    endpoint: ...                                                                                                                                                                                                                                                                                                                                                       
    retry_on_failure:                                                                                                                                                                                                                                                                                                                                                                                                                 
      max_elapsed_time: 600s                                                                                                                                                                                                                                                                                                                                                                                                          
    sending_queue:                                                                                                                                                                                                                                                                                                                                                                                                                    
      queue_size: 1000                                                                                                                                                                                                                                                                                                                                                                                                                
      storage: file_storage
```

**Environment**
N/A

**Additional context**
N/A

neetle:
We're also seeing this issue with exporting traces via the GRPC version of this client when a memorylimiter is in play. 

Looking at the error logs, we're seeing:
```
error	exporterhelper/queued_retry.go:391	Exporting failed. The error is not retryable. Dropping data.	{"kind": "exporter", "data_type": "traces", "name": "otlp", "error": "Permanent error: rpc error: code = Unknown desc = data refused due to high memory usage", "dropped_items": n}
```

From my interpretation of the code & the logs, given an exporter in service A and a receiver & processor in service B:
1) The exporter in service A sends a request to the receiver in service B.
2) The receiver in service B passes the traces to the processor 
3) The processor in service B returns an error. This is propagated back up the pipeline to the receiver.
4) The receiver in service B gets the error. It runs it through https://github.com/open-telemetry/opentelemetry-collector/blob/3089ea878eef022913a053600f3599b3fddffb95/receiver/otlpreceiver/otlphttp.go#L157 or similar in the GRPC version. 
[...]

Issue #6725, opened 2022-12-09
[receiver/otlpreceiver] Support Rate Limiting


blakeroberts-wk:
**Is your feature request related to a problem? Please describe.**

The OpenTelemetry Specification outlines throttling for both [gRPC](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-throttling) and [HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-throttling); however, the OTLP receiver does not currently support this (optional) specification.

Right now, if a processor is under pressure the only option it has is to return an error informing the receiver to tell the client that the request failed and is not retry-able.

**Describe the solution you'd like**

It would be neat if the receiver offered an implementation of `error` that could be returned to it to signal that it should return an appropriately formatted response to the client signaling that the request was rate limited. The format of the response should follow the semantic convention (e.g. the HTTP receiver should return a status code of `429` and set the "Retry-After" header.

For example, the OTLP receiver could export the following `error` implementation:

```go
package errors

import "time"

type ErrorRateLimited struct {
	Backoff time.Duration
}

func (e *ErrorRateLimited) Error() string {
	return "Too Many Requests"
}

var ErrRateLimited error = &ErrorRateLimited{}

func NewErrRateLimited(backoff time.Duration) error {
	return &ErrorRateLimited{
		Backoff: backoff,
	}
}
```

Any processor or exporter in the pipeline could return (optionally wrapping) this error:

```go
import "go.opentelemetry.io/collector/receiver/otlpreceiver"

func (p *processor) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
	return otlpreceiver.NewErrRateLimited(time.Minute)
}
```

Then when handling errors from the pipeline, the receiver could check for this error:

```go
var (
	err error
	w http.ResponseWriter
)

errRateLimited := &ErrorRateLimited{}
if errors.As(err, &errRateLimited) {
	w.Header().Set("Retry-After", strconv.FormatInt(int64(errRateLimited.Backoff)/1e9, 10))
	w.WriteHeader(http.StatusTooManyRequests)
}
```

**Describe alternatives you've considered**

[...]

PR #8080, opened 2023-07-12
[OTLP/GRPC] Ensure OTLP receiver handles consume errors correctly 

VihasMakwana:
**Description:** Follow the receiver contract and return Unavailable for non-permanent and InvalidArgument for permanent errors for OTLP/gRPC receiver.

Leave the "Retry-After" field blank and let the client implement an exponential backoff strategy.

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/4335

**Testing:** Added relevant test cases.

PR #9357, opened 2024-01-23
[receiver/otlp] Return proper http response code based on retryable errors

TylerHelmuth:
**Description:**
Updates the receiver's http response to return a proper http status based on whether or not the pipeline returned a retryable error.  Builds upon the work done in https://github.com/open-telemetry/opentelemetry-collector/pull/8080 and https://github.com/open-telemetry/opentelemetry-collector/pull/9307

**Link to tracking Issue:**

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9337
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/8132
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9636
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/6725

**Testing:**

Updated lots of unit tests


receiver/otlpreceiver/internal/errors/errors.go
```diff
@@ -23,3 +25,22 @@ func GetStatusFromError(err error) error {
...
+	// to see if a code is retryable.
+	// See https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures-1
+	// to see a list of retryable http status codes.
+	switch s.Code() {
+	// Retryable
+	case codes.Canceled, codes.DeadlineExceeded, codes.Aborted, codes.OutOfRange, codes.Unavailable, codes.DataLoss:
+		return http.StatusServiceUnavailable
+	// Not Retryable
+	default:
+		return http.StatusInternalServerError
```
evan-bradley, line 44: Permanent errors will be [converted to `status.InvalidArgument`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/internal/errors/errors.go#L20), which is equivalent to an HTTP 400. We should change one of the two codes so they're semantically equivalent between gRPC and HTTP.

TylerHelmuth: I thought about that too, but since [the spec says](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures-1):

```
This specification does not use Status.code field and the server MAY omit Status.code field. The clients are not expected to alter their behavior based on Status.code field but MAY record it for troubleshooting purposes.
```

I opted for now for 500 since it is what we were returning before.  If we want to change something we should change `GetStatusFromError` to return `codes. INTERNAL` or `codes.UNKNOWN`, but we aren't required to by the spec.

TylerHelmuth: With https://github.com/open-telemetry/opentelemetry-collector/pull/9415 merged the status codes are now aligned.

receiver/otlpreceiver/otlphttp.go
```diff
@@ -42,7 +43,7 @@ func handleTraces(resp http.ResponseWriter, req *http.Request, tracesReceiver *t
 
 	otlpResp, err := tracesReceiver.Export(req.Context(), otlpReq)
 	if err != nil {
-		writeError(resp, enc, err, http.StatusInternalServerError)
+		writeError(resp, enc, err, errors.GetHTTPStatusCodeFromStatus(err))
```
bogdandrutu, line 46: I don't like that now we also have a similar but different logic in `errorMsgToStatus`. Can we consolidate?

TylerHelmuth: Agreed. I took a look at consolidating before opening the PR. I believe it can be refactored, but it caused a lot of changes unrelated to the issue I was trying to solve. To keep the PRs smaller and targeted I'd like to use this PR as a solution to the issue and then a future PR that is only a refactor. I can open an issue to track that refactor if you'd like.

TylerHelmuth: @bogdandrutu are you ok with that approach?

TylerHelmuth:
@open-telemetry/collector-maintainers please review

TylerHelmuth:
@open-telemetry/collector-maintainers please review

mx-psi:
@bogdandrutu PTAL

0x006EA1E5:
Hi, I was just looking at this issue myself locally, and have found this open issue...

I notice this, #9307 and #8080 are somewhat coupled to GRCP status codes.

I also notice we have [`consumer/consumererror`](https://github.com/open-telemetry/opentelemetry-collector/tree/main/consumer/consumererror), which seems to address somewhat similar concerns, i.e. retryable and permanent errors.

Can you clarify if `consumer/consumererror` would be appropriate here, or is it intended for something else?  I see `consumererror.New` is [found in a number of places in the contrib codebase](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-collector-contrib%20%22consumererror.New%22&type=code).

Have you considered using `consumererror.IsPermanent(err)` in the receiver to determine the response status code?

Using `consumererror` would seem to make it a bit more user friendly to actually create the errors in the downstream processor, e.g., something like `return td, consumererror.NewTraces(errDataRefused, td)` in [`memorylimiter`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/memorylimiterprocessor/memorylimiter.go#L62).

TylerHelmuth:
@0x006EA1E5 as you've deduced, `consumer/consumererror` is normally used downstream from receivers to tell receivers what kind of errors they got from `r.nextConsumer.Consume*`.  In this instance we're in a receiver and the error we're propagating back to the caller is subject to the OTLP specification (which is closely tied to grpc objects even for http) so returning `consumererror` would not be appropriate. [The receiver does use `consumererror` when initially handling the error returned from downstream components.](https://github.com/open-telemetry/opentelemetry-collector/blob/47de8647738d93209537f559874be6c303d3666b/receiver/otlpreceiver/internal/errors/errors.go#L19)

TylerHelmuth:
@bogdandrutu @open-telemetry/collector-approvers please take a look. `errors.GetHTTPStatusCodeFromStatus(err)` does resemble some other internal functions for handling errors in other code paths, but I think a separate refactor after this is added will be best to keep this PR small and targeted.

receiver/otlpreceiver/internal/errors/errors.go
```diff
@@ -24,3 +26,22 @@ func GetStatusFromError(err error) error {
...
+	}
+	// See https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures
+	// to see if a code is retryable.
+	// See https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures-1
+	// to see a list of retryable http status codes.
+	switch s.Code() {
+	// Retryable
+	case codes.Canceled, codes.DeadlineExceeded, codes.Aborted, codes.OutOfRange, codes.Unavailable, codes.DataLoss:
+		return http.StatusServiceUnavailable
```
keruitan-wk, line 42: Is it possible to add a case for returning 429 or a custom HTTP status code that would address https://github.com/open-telemetry/opentelemetry-collector/issues/6725?

TylerHelmuth: @keruitan-wk the [otlp htttp throttling specification](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-throttling) states:

> If the server receives more requests than the client is allowed or the server is overloaded, the server SHOULD respond with HTTP 429 Too Many Requests or HTTP 503 Service Unavailable and MAY include [“Retry-After”](https://tools.ietf.org/html/rfc7231#section-7.1.3) header with a recommended time interval in seconds to wait before retrying.

My interpretation of that specification is that `503` is an adequate response code to communicate to clients that the request can be retried.  I don't see the spec making any distinction on how a client should handle a `429` vs a `503` After this PR is merged the `otlpreceiver` will be able to return `503`s which will trigger retry attempts from a client. 

I've updated the description to close https://github.com/open-telemetry/opentelemetry-collector/issues/6725 when this PR is merge.

keruitan-wk: HTTP 4xx status codes indicate that it is a client side error, while 5xx status codes indicate server side error. While both 429 and 503 are retryable, it would be nice to be able to distinguish between the two as 429 is meant to tell the client that they are sending too many requests while the collector is working as expected, vs 503 that something is wrong with the collector and they can try again later.

TylerHelmuth: I agree with you, but the client response to either code is the same according to the OTLP specification: exponential backoff retries.

I agree that getting the exactly proper http status code would be nice.  I could add a check in the switch statement for the grpc status code `RESOURCE_EXHAUSTED` which I think is the code that translates to `429`.

jesusvazquez (approved):
LGTM I've personally tested this change on tempo and has worked as expected, we get proper mappings now. Tempo uses the receiver code for both handling and being up to date with the latest otel changes.

jpkrohling (approved):

songy23 (approved):
LGTM

mx-psi (approved):

receiver/otlpreceiver/otlphttp.go
TylerHelmuth: Created https://github.com/open-telemetry/opentelemetry-collector/issues/9864

PR #9358, opened 2024-01-23
Replace local dependencies for mdatagen, unblocks release action

bogdandrutu:


codeboten (approved):
I'm approving this to unblock the release... I'm not sure what the right way to address this is, but i'll re-open https://github.com/open-telemetry/opentelemetry-collector/issues/9281

mmynk:
@bogdandrutu @codeboten I wanted to understand what the `replace` directives are solving here. This has rendered it impossible to use `go install` for `mdatagen`.
It's been 4 months and there is still no way to install `mdatagen` without cloning this repo.

PR #9361, opened 2024-01-23
[service] fix opencensus bridge in periodic readers

codeboten:
The periodic metric readers were not connecting to the opencensus bridge prior to this change, meaning anyone trying to emit metrics via the console or OTLP from components that used opencensus were failing to do so.

### DIFF
--- a/service/internal/proctelemetry/config.go
+++ b/service/internal/proctelemetry/config.go
@@ -71,7 +71,9 @@ func InitMetricReader(ctx context.Context, reader config.MetricReader, asyncErro
 		return initPullExporter(reader.Pull.Exporter, asyncErrorChannel)
 	}
 	if reader.Periodic != nil {
-		opts := []sdkmetric.PeriodicReaderOption{}
+		opts := []sdkmetric.PeriodicReaderOption{
+			sdkmetric.WithProducer(opencensus.NewMetricProducer()),
+		}
 		if reader.Periodic.Interval != nil {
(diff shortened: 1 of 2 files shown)


codeboten:
I tested this manually to confirm this fixes the bug.

Not sure the best way to test this in a unit test, i would need a destination where I can send periodic reader data configured, which i don't have in a test here... I figured this was a temporary piece of code anyways, until the remaining contrib components are migrated away from opencensus

service/internal/proctelemetry/config.go
```diff
@@ -71,7 +71,9 @@ func InitMetricReader(ctx context.Context, reader config.MetricReader, asyncErro
 		return initPullExporter(reader.Pull.Exporter, asyncErrorChannel)
 	}
 	if reader.Periodic != nil {
-		opts := []sdkmetric.PeriodicReaderOption{}
+		opts := []sdkmetric.PeriodicReaderOption{
+			sdkmetric.WithProducer(opencensus.NewMetricProducer()),
```
TylerHelmuth, line 75: I dont think I understand what `WithProducer` does, how was this doing anything before? Or has it been very broken without this?

bogdandrutu (approved):

service/internal/proctelemetry/config.go
codeboten: it has been very broken for periodic metric readers that cares about components that produce opencensus metrics.

PR #9378, opened 2024-01-24
[chore] Clarify timeline strictness in stabilization issues.

TylerHelmuth:


.github/ISSUE_TEMPLATE/stabilization.md
```diff
@@ -6,7 +6,7 @@ labels: 'stabilization'
 assignees: ''
 ---
 
-Before stabilizing a module, an approver or maintainer must make sure that the following criteria have been met for at least two successive minor version releases:
+Before stabilizing a module, an approver or maintainer must make sure that the following criteria have been met for at least two successive minor version releases (regardless of when this issue was opened):
```
mx-psi, line 9: I guess the thing that blocks us then is not the issue but doing the announcements on Slack/the Collector SIG meetings. Is it worth removing these and making a general announcement?

TylerHelmuth: > Is it worth removing these and making a general announcement?

Does `these` refer to opening these issues?  We could probably get by with only the Milestones, but personally I like these issues as a helpful checklist and place for centralized stabilization discussion.

atoulme: I think @mx-psi is saying that the longest wait is the announcements, because otherwise you'd be able to open the issue, check the checklist and see that for the last 2 releases, everything was done, and quickly move to stable. But you wouldn't really make announcements before opening the issue. So we could make a big announcement of our intentions in general, and dispense of announcing each module one by one.

I think after this wave, we could make a general announcement and remove the requirement, if we feel good about our progress.

atoulme (approved):
Approved, as a step towards a bit less waiting.

.github/ISSUE_TEMPLATE/stabilization.md
mx-psi: Yup, exactly what Antoine said :) Thanks for clarifying my confusing message

mx-psi (approved):
I think this is a change is in the right direction, we can discuss the announcements thing separately

.github/ISSUE_TEMPLATE/stabilization.md
TylerHelmuth: Ok yes that makes sense to me.

Issue #9364, opened 2024-01-24
[confignet] Should the two structs implement `ConfigValidator`?


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**

Since these structs could be used in configuration, should we add `Validate` functions to implement `ConfigValidator`? For `NetAddr`, for example, we could validate that `Transport` is one of `"tcp", "tcp4", "tcp6", "udp", "udp4", "udp6", "ip", "ip4", "ip6", "unix", "unixgram" and "unixpacket"`.

mx-psi:
Should we make `Transport` into a type with `UnmarshalText`?

PR #9385, opened 2024-01-24
[confignet] Change `Transport` from `string` to `TransportType`

TylerHelmuth:
**Description:** 
Changes `Transport` from a `string` to a new `TransportType`.  Implements `UnmarshalText` for `TransportType` to enforce values.

This PR may be too much - it introduces a breaking change a lot of new public APIs that may not be worth it for such a small module. If we don't like the surface area this creates or the breaking change, but we still want to enforce transport type values, I think implementing `Validate` keeps the API footprint smaller and isn't breaking.

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9364

Added godoc comments


config/confignet/confignet.go
```diff
@@ -5,10 +5,53 @@ package confignet // import "go.opentelemetry.io/collector/config/confignet"
...
+	TCP6       TransportType = "tcp6"
+	UDP        TransportType = "udp"
+	UDP4       TransportType = "udp4"
+	UDP6       TransportType = "udp6"
+	IP         TransportType = "ip"
+	IP4        TransportType = "ip4"
+	IP6        TransportType = "ip6"
+	UNIX       TransportType = "unix"
+	UNIXGRAM   TransportType = "unixgram"
+	UNIXPACKET TransportType = "unixpacket"
```
mx-psi, line 28: Wondering if these should be `Unix`, `Unixgram` and `UnixPacket` instead

config/confignet/confignet.go
```diff
@@ -27,7 +70,7 @@ type NetAddr struct {
 
 	// Transport to use. Known protocols are "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only), "udp", "udp4" (IPv4-only),
 	// "udp6" (IPv6-only), "ip", "ip4" (IPv4-only), "ip6" (IPv6-only), "unix", "unixgram" and "unixpacket".
-	Transport string `mapstructure:"transport"`
+	Transport TransportType `mapstructure:"transport"`
```
TylerHelmuth, line 73: This is breaking stuff in Contrib (expected).  Do we need to introduce this change as a new field called `TransportType` and follow the deprecation process to get rid of `Transport` and then rename back to `Transport`?  I'd prefer not to do another 3-release cycle change.

config/confignet/confignet.go
```diff
@@ -5,10 +5,53 @@ package confignet // import "go.opentelemetry.io/collector/config/confignet"
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"time"
 )
 
+// TransportType represents a type of network transport protocol
+type TransportType string
```
TylerHelmuth, line 14: If we go forward with this change, should we call this `TransportType` or `Transport`?  There was a comment [here](https://github.com/open-telemetry/opentelemetry-collector/issues/6490#issuecomment-1307663325) that maybe we shouldn't include `Type` in these names.

My opinion is to keep `Type` in the name.

TylerHelmuth: Related to https://github.com/open-telemetry/opentelemetry-collector/issues/9388

dmitryax: I think we should finalize the discussion in https://github.com/open-telemetry/opentelemetry-collector/issues/9388 before making any changes here

dmitryax: `Transport` sounds good to me in this particular case though

config/confignet/confignet.go
```diff
@@ -5,10 +5,53 @@ package confignet // import "go.opentelemetry.io/collector/config/confignet"
...
+	TCP6       TransportType = "tcp6"
+	UDP        TransportType = "udp"
+	UDP4       TransportType = "udp4"
+	UDP6       TransportType = "udp6"
+	IP         TransportType = "ip"
+	IP4        TransportType = "ip4"
+	IP6        TransportType = "ip6"
+	Unix       TransportType = "unix"
+	Unixgram   TransportType = "unixgram"
+	UnixPacket TransportType = "unixpacket"
```
dmitryax, line 28: The constants should start with the type as we do in other components e.g. `TransportTypeUnixPacket`. It might be worth putting such guidelines in docs.

config/confignet/confignet.go
TylerHelmuth: Wrote up a quick summary of state of names here: https://github.com/open-telemetry/opentelemetry-collector/issues/9388#issuecomment-1915344957.  I think if we standardize on not including `Type` that there are some names in pdata that become "out-of-standard".  Since pdata is 1.0 and sets the precedence, I'm definitely in favor of keeping `Type` in the name.

I also personally find it clearer at a glance.  `confignet.Transport` conveys less information (in my opinion), than `confignet.TransportType`.  I feel like `Type` in the name quickly tells me it is used for enums.

config/confignet/confignet.go
TylerHelmuth: Totally agree.  I can document all the enum naming decisions once we decide something on the `Type` or not `Type` discussion.

config/confignet/confignet.go
mx-psi: I think we should file a separate PR for the general rule and discuss it there (I mentioned what I think we should cover on https://github.com/open-telemetry/opentelemetry-collector/issues/9388#issuecomment-1919075284, if you think there are other items to add please say so!)

TylerHelmuth: Based on the new expectations for enumeration names, the name needs to convey a sense of categorization and I don't feel like the word `Transport` on its own is enough. I believe `TransportType` meets our expectations better.

config/confignet/confignet.go
```diff
@@ -36,20 +79,20 @@ type NetAddr struct {
 // Dial equivalent with net.Dialer's DialContext for this address.
 func (na *NetAddr) Dial(ctx context.Context) (net.Conn, error) {
 	d := net.Dialer{Timeout: na.DialerConfig.Timeout}
-	return d.DialContext(ctx, na.Transport, na.Endpoint)
+	return d.DialContext(ctx, string(na.Transport), na.Endpoint)
```
mx-psi, line 82: nit: I think having an explicit `String` method would be better

TylerHelmuth: @mx-psi `String` won't play well with the consts unless we make the function a value receiver right? But then we'd be mixing value and reference receivers.

config/confignet/confignet.go
TylerHelmuth: Looking at this again and taking the new Enumeration expectations into account, I think it is better if the config field name matches the enum name.  I've pushed a commit that adds a new field, `TransportType`, and deprecates `Transport`.  If we can get this merged before 0.94.0 that means we can remove `Transport` after `0.94.0` and mark as stable for the `0.95.0` release.

config/confignet/confignet.go
TylerHelmuth: Hmm but that breaks user configurations, not just the API.  What do you think?

config/confignet/confignet.go
TylerHelmuth: Hmm but that breaks user configurations (config goes from transport to transport_type), not just the API. What do you think?

mx-psi: @TylerHelmuth we can put `transport` as the `mapstructure` tag for `TransportType` and put some dummy value on the deprecated field (we tipically have used `-` for unexported fields). Since `TransportType` unmarshals in the same way as `Transport`, this should make it non-breaking for user configurations

config/confignet/confignet.go
mx-psi: Agreed, `TransportType` seems to fit the rules better

config/confignet/confignet_test.go
```diff
@@ -15,8 +15,8 @@ import (
 
 func TestNetAddrTimeout(t *testing.T) {
 	nac := &NetAddr{
-		Endpoint:  "localhost:0",
-		Transport: "tcp",
+		Endpoint:      "localhost:0",
+		TransportType: "tcp",
```
mx-psi, line 19:
```suggestion
		TransportType: confignet.TransportTypeTCP,
```

config/confignet/confignet_test.go
```diff
@@ -50,8 +50,8 @@ func TestTCPAddrTimeout(t *testing.T) {
 
 func TestNetAddr(t *testing.T) {
 	nas := &NetAddr{
-		Endpoint:  "localhost:0",
-		Transport: "tcp",
+		Endpoint:      "localhost:0",
+		TransportType: "tcp",
```
mx-psi, line 54:
```suggestion
		TransportType: confignet.TransportTypeTCP,
```

receiver/otlpreceiver/testdata/uds.yaml
```diff
@@ -1,7 +1,7 @@
 # The following entry demonstrates how to specify a Unix Domain Socket for the server.
 protocols:
   grpc:
-    transport: unix
+    transport_type: unix
```
mx-psi, line 4: I think we should revert this back using the technique I mentioned on a comment above

mx-psi (commented):
IMO this PR should not make any changes that affect end users like changing the struct tags related to this. 

We may consider doing the breaking change on a separate PR (personally I don't think it is worth it), but since we can consider the changes separately, I think we should keep each change on a separate PR

config/confignet/confignet.go
TylerHelmuth: Oh clever, I'll make that change

config/confignet/confignet.go
mx-psi: Yup, makes sense, we can resolve this

receiver/otlpreceiver/testdata/uds.yaml
TylerHelmuth: Will do

config/confignet/confignet.go
TylerHelmuth: Done

receiver/otlpreceiver/testdata/uds.yaml
TylerHelmuth: Done

mx-psi (approved):

config/confignet/confignet.go
```diff
@@ -5,10 +5,53 @@ package confignet // import "go.opentelemetry.io/collector/config/confignet"
...
+	TransportTypeUnix       TransportType = "unix"
+	TransportTypeUnixgram   TransportType = "unixgram"
+	TransportTypeUnixPacket TransportType = "unixpacket"
+)
+
+// UnmarshalText unmarshalls text to a TransportType.
+// Valid values are "tcp", "tcp4", "tcp6", "udp", "udp4",
+// "udp6", "ip", "ip4", "ip6", "unix", "unixgram" and "unixpacket"
+func (tt *TransportType) UnmarshalText(in []byte) error {
+	switch typ := TransportType(in); typ {
```
bogdandrutu, line 35: What about empty string, that will error.

config/confignet/confignet.go
```diff
@@ -36,20 +84,32 @@ type NetAddr struct {
...
+	tt := string(na.TransportType)
+	if na.Transport != "" {
+		tt = na.Transport
+	}
+
+	return d.DialContext(ctx, tt, na.Endpoint)
 }
 
 // Listen equivalent with net.ListenConfig's Listen for this address.
 func (na *NetAddr) Listen(ctx context.Context) (net.Listener, error) {
```
bogdandrutu, line 97: Would also add a `Validate` for NetAddr and `NetTCP` since user may manually setup the transport (in code) and may be a wrong invalid value (typo).

config/confignet/confignet.go
```diff
@@ -36,20 +84,32 @@ type NetAddr struct {
 // Dial equivalent with net.Dialer's DialContext for this address.
 func (na *NetAddr) Dial(ctx context.Context) (net.Conn, error) {
 	d := net.Dialer{Timeout: na.DialerConfig.Timeout}
-	return d.DialContext(ctx, na.Transport, na.Endpoint)
+
+	tt := string(na.TransportType)
+	if na.Transport != "" {
+		tt = na.Transport
+	}
```
bogdandrutu, line 91: I think there is a bug here. The reason is that implementations may set the default `Transport` to `"foo"` then user sets it to `"bar"`, based on this logic if I am not mistaken you will end up with `"foo"`.

So in order to make it work (I think we can still do it if you want to not just break the go API) with implementing the `Unmarshaler` interface and before unmarshalling set the `TransportType` to the value of `Transport` then here you can ignore old value. Something like:

```golang
func (na *NetAddr) Unmarshal(confmap.Confmap cm) error {
  na.TrasportType = TransportType(na.Trasport)
  return cm.Unmarshal(na)
}
```

config/confignet/confignet.go
TylerHelmuth: I'll add a `transportTypeEmpty`, like `configcompression` has.

config/confignet/confignet.go
TylerHelmuth: I'll add a `Validate` function for `NetAddr`, but `TCPAddr` doesn't allow users to set a `TransportType` so there isn't anything for us to Validate yet.

config/confignet/confignet.go
TylerHelmuth: What I tried didn't work for cases like these: https://github.com/open-telemetry/opentelemetry-collector/blob/1a57837accdff8bf9abc769a8e49b5207b52fad2/config/configgrpc/configgrpc_test.go#L618-L624 where a user is setting the public members directly, bypassing any unmashaling.

I've reverted the code to do a check for `na.Transport` in `Dial` and `Listen`. Can you explain more how that check introduces a bug for those using `Transport` instead of `TransportType`?

config/confignet/confignet.go
TylerHelmuth: @mx-psi I believe configs won't be broken, but we'll break scenarios [like this](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/7878581831/job/21497142469?pr=9385) where the user is expecting `Transport` to be set instead of `TransportType`.  @bogdandrutu is this the bug you're thinking of?

TylerHelmuth: All of this is just making me want to remove `Transport` and do this as a breaking change since this is an infrequently used package and the fix is easy.

config/confignet/confignet.go
TylerHelmuth: @bogdandrutu can you explain further?

mx-psi:
@bogdandrutu This is waiting on your review

config/confignet/confignet.go
bogdandrutu: User in their "defaultConfig"

```
return { Transport: "unix" } // configure the deprecated value (until I change the config)
```

In the yaml file:

```
  "transport": "tcp",
``` 

Then now will Call `Dial`, which protocol will be used?

TylerHelmuth: Oh I see what you mean.  I agree that is a bug.

Your proposed fix allows the default config being overridden by user config, but since `Transport` and `TransportType` are both public, users could still explicitly set `Transport` somewhere other than in via the default config function, meaning we'd still have to check for `Transport` being set in `Dial` and `Listen`.

Since this would only be a breaking change in the API and not for end user configs, I am now very much in favor of doing this change as a breaking API change.  I've update the PR to reflect that opinion.

config/confignet/confignet.go
TylerHelmuth: Ive update the PR to do this as a breaking change.

TylerHelmuth:
All tests are now passing except Contrib, which I'll fixed after this is merged should we choose to go forward with this breaking change approach.

mx-psi (approved):

Issue #6767, opened 2022-12-13
Clearly separate Config and Settings terms used in struct names throughout the codebase


dmitryax:
Currently, we use the `Config` term for collector configuration as a whole and parts of it defined for receivers/processors/exporters/extensions. Examples:

- `otelcol.Config`
- `component.Config`
- `otlpexporter.Config`

But for unmarshalling some deeper parts of the user config, we use `Settings` structs, for example:

- [x] `exporterhelper.QueueSettings`
- [x] `configgrpc.GRPCClientSettings`
- [x] `configtls.TLSSetting`

At the same time `service/telemetry` package still uses `Config`, e.g. `MetricsConfig`, `LogsSamplingConfig`, and `TracesConfig`.

The suggestion is to use `Config` term for all the structs used to unmarshal any parts of the YAML config provided by users. We will keep `Settings` only for structs that are populated programmatically, like:

- `component.TelemetrySettings`
- `component.ReceiverSettings`

bogdandrutu:
I think we should start working on this, because we cannot delay until the stability.

atoulme:
List of renames:
* [ ] `GRPCClientSettings`
* [ ] `GRPCServerSettings`
* [ ] `HTTPClientSettings`
* [ ] `HTTPServerSettings`
* [ ] `CORSSettings`
* [ ] `QueueSettings`
* [ ] `TimeoutSettings`
* [ ] `ScraperControllerSettings`

TylerHelmuth:
Removing from the confighttp milestone as the public structs now use `Config`.

codeboten:
Closing this in favour of https://github.com/open-telemetry/opentelemetry-collector/issues/9428

codeboten:
exporterhelper still needs updating

jade-guiton-dd:
I'll work on this.

jade-guiton-dd:
If I'm not mistaken, #11264 was the last step for this issue, so it can probably be closed.

mx-psi:
Taking a look at the output of `rg 'type \w*Setting' -tgo -i` it looks like this is indeed the case :rocket:

PR #9391, opened 2024-01-25
[exporterhelper] Deprecate TimeoutSettings, use TimeoutConfig instead

atoulme:
**Description:**
Deprecate TimeoutSettings, use TimeoutConfig instead

**Link to tracking Issue:**
#6767


dmitryax:
Same as with https://github.com/open-telemetry/opentelemetry-collector/pull/9393. I think there is no need to break the existing option for a function that will be replaced with another one

atoulme:
Going to close for now, and we can always reopen if plans change.

PR #9392, opened 2024-01-25
[config/confighttp] Deprecate CORSSettings, use CORSConfig instead

atoulme:
**Description:**
Deprecate CORSSettings, use CORSConfig instead

**Link to tracking Issue:**
#6767

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -398,7 +398,7 @@ func responseHeadersHandler(handler http.Handler, headers map[string]configopaqu
 // CORSSettings configures a receiver for HTTP cross-origin resource sharing (CORS).
 // See the underlying https://github.com/rs/cors package for details.
 // Deprecated: [v0.94.0] Use CORSConfig instead
-type CORSSettings CORSConfig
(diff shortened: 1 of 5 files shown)


TylerHelmuth (approved):
Assuming we're still on board with `Config` over `Settings` this looks right to me. Thanks for starting these deprecations early!

atoulme:
I assumed we were still going with those. They'll make checking APIs much easier.

config/confighttp/confighttp.go
```diff
@@ -397,7 +397,12 @@ func responseHeadersHandler(handler http.Handler, headers map[string]configopaqu
 
 // CORSSettings configures a receiver for HTTP cross-origin resource sharing (CORS).
 // See the underlying https://github.com/rs/cors package for details.
-type CORSSettings struct {
+// Deprecated: [v0.94.0] Use CORSConfig instead
+type CORSSettings CORSConfig
```
dmitryax, line 401: This should be an alias so both types are interoperable 

```suggestion
type CORSSettings = CORSConfig
```

dmitryax (approved):

PR #9398, opened 2024-01-25
[service] Move MeterProvider initialization to service/telemetry

mx-psi:
**Description:**

- [service] Move TracerProvider initialization to service/telemetry package
- Move MeterProvider initialization to service/telemetry

**Link to tracking Issue:** TBF


mx-psi:
To avoid adding public API that we would likely remove this later, I am going to close this and start with #9454 instead

PR #9403, opened 2024-01-26
[configgrpc] Deprecate GRPCServerSettings, use GRPCServerConfig instead

atoulme:
**Description:**
Deprecate GRPCServerSettings, use GRPCServerConfig instead

**Link to tracking Issue:**
#6767


config/configgrpc/configgrpc.go
```diff
@@ -117,7 +117,11 @@ type KeepaliveEnforcementPolicy struct {
 }
 
 // GRPCServerSettings defines common settings for a gRPC server configuration.
-type GRPCServerSettings struct {
+// Deprecated: [v0.94.0] Use GRPCServerConfig instead
+type GRPCServerSettings = GRPCServerConfig
+
+// GRPCServerConfig defines common settings for a gRPC server configuration.
+type GRPCServerConfig struct {
```
bogdandrutu, line 124: qq: Should we also make this `ServerConfig` since it will be used as `configgrpc.ServerConfig`? I personally like the shorter name.

bogdandrutu (approved):

config/configgrpc/configgrpc.go
djaglowski: Even `configgrpc.Server` seems clear enough to me, since config is also redundant.

dmitryax: I like `configgrpc.ServerConfig`. `configgrpc.Server` sounds a bit unclear.

@open-telemetry/collector-approvers WDYT?

We already did this change to `confighttp`. If we think that it's a good option we should drop `HTTP` from `HTTPServerConfig ` and `HTTPServerConfig` before the next release

mx-psi: `ServerConfig` makes more sense to me, it's weird to have a `ToServer` method on a struct called `Server`

TylerHelmuth: I also like `configgrcp.ServerConfig`

atoulme: alright, I see how it is. I'll get it changed.

dmitryax: Another thing to consider is that it's currently used as an embedded field in the exporter configs. 

Changing

```
type Config struct {
	configgrpc.GRPCClientSettings `mapstructure:",squash"`
}
```

to 

```
type Config struct {
	configgrpc.ClientConfig `mapstructure:",squash"`
}
```

maybe not ideal. So maybe we should un-embed it to?

```
type Config struct {
	GRPCClientConfig configgrpc.ClientConfig `mapstructure:",squash"`
}
```

As another option, we can suggest wrapping it into `GRPCConfig` as already done in otlpreceiver https://github.com/open-telemetry/opentelemetry-collector/blob/f37e3766c4af98fe17e8846704f0484d5c0142c4/receiver/otlpreceiver/config.go#L24 and do the same in otlp exporters

dmitryax:
Putting the blocker label to avoid breaking the merged HTTP configs after the discussion is resolved

config/configgrpc/configgrpc.go
mx-psi: > So maybe we should un-embed it to?

Yes, I think this is something that should be fixed by unembedding.

bogdandrutu: I think everyone is ok with `ServerConfig`.

dmitryax: Un-embedding sounds good to me for the otlp exporter. But I think we should wrap https://github.com/open-telemetry/opentelemetry-collector/blob/f37e3766c4af98fe17e8846704f0484d5c0142c4/receiver/otlpreceiver/config.go#L39 in `GRPCConfig` for consistency

bogdandrutu: That an implementation detail for that receiver not for this thread.

dmitryax: Yeah. We just need to do it before the next release because that API has already changed in a breaking way, so we can quickly update it

atoulme: Pushing `GRPCServerConfig -> ServerConfig` now.

dmitryax: Oh, I was wrong this doesn't introduce the breaking chance since `ServerConfig` isn't embedded in this receiver. So no more changes are needed

dmitryax (approved):

config/configgrpc/configgrpc.go
dmitryax: I pushed a PR to update confighttp accordingly https://github.com/open-telemetry/opentelemetry-collector/pull/9453

PR #9404, opened 2024-01-26
[confighttp] Deprecate HTTPClientSettings, use HTTPClientConfig instead

atoulme:
**Description:**
Deprecate HTTPClientSettings, use HTTPClientConfig instead

**Link to tracking Issue:**
#6767

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -30,7 +30,11 @@ import (
 const headerContentEncoding = "Content-Encoding"
 
 // HTTPClientSettings defines settings for creating an HTTP client.
-type HTTPClientSettings struct {
+// Deprecated: [v0.94.0] Use HTTPClientConfig instead
+type HTTPClientSettings = HTTPClientConfig
+
+// HTTPClientConfig defines settings for creating an HTTP client.
+type HTTPClientConfig struct {
 	// The target URL to send data to (e.g.: http://some.url:9411/v1/traces).
 	Endpoint string `mapstructure:"endpoint"`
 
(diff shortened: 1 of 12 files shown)


dmitryax (approved):

config/confighttp/confighttp.go
```diff
@@ -103,19 +107,28 @@ type HTTPClientSettings struct {
...
-	return HTTPClientSettings{
+	return HTTPClientConfig{
 		MaxIdleConns:    &maxIdleConns,
 		IdleConnTimeout: &idleConnTimeout,
 	}
 }
 
 // ToClient creates an HTTP client.
-func (hcs *HTTPClientSettings) ToClient(host component.Host, settings component.TelemetrySettings) (*http.Client, error) {
+func (hcs *HTTPClientConfig) ToClient(host component.Host, settings component.TelemetrySettings) (*http.Client, error) {
```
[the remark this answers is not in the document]
dmitryax: Misread it. nvm

exporter/otlphttpexporter/config.go
```diff
@@ -14,9 +14,9 @@ import (
 
 // Config defines configuration for OTLP/HTTP exporter.
 type Config struct {
-	confighttp.HTTPClientSettings `mapstructure:",squash"`     // squash ensures fields are correctly decoded in embedded struct.
-	QueueConfig                   exporterhelper.QueueSettings `mapstructure:"sending_queue"`
-	RetryConfig                   configretry.BackOffConfig    `mapstructure:"retry_on_failure"`
+	confighttp.HTTPClientConfig `mapstructure:",squash"`     // squash ensures fields are correctly decoded in embedded struct.
```
dmitryax, line 17: This is a breaking change. the old field should be kept as deprecated

dmitryax: Actually. It'll be pretty hard to keep it backward-compatible. I think we should allow breaking change here. It's only applicable to the API, not the user interface tho. Please mark it accordingly in the changelog

atoulme: yes, I was looking for a way to do struct field aliasing or something like that, but no good solution seems to exist. This only breaks for programmatic use so mostly touching on tests in contrib. I will add to changelog, thanks!

PR #9406, opened 2024-01-26
[service/telemetry] Deprecate telemetry.Settings in favor of telemetry.CreateSettings

mx-psi:
**Description:** Deprecate `telemetry.Settings` in favor of `telemetry.CreateSettings`. This is for consistency with how we name similar structs in exporters, receivers, processors, connectors and extensions.

### DIFF
--- a/.chloggen/mx-psi_createsettings.yaml
+++ b/.chloggen/mx-psi_createsettings.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service/telemetry
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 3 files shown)


atoulme (approved):

dmitryax:
The `Create` prefix for the receivers, processors, and exporters comes from the fact that it's passed to functions like `exporter.CreateLogsFunc`. @bogdandrutu made this point recently. Given that it's passed to the telemetry constructor, maybe we can also rename this one. However, I think we need to take into account other `Settings` in the project to decide. It is probably better to start with an issue

mx-psi:
> The `Create` prefix for the receivers, processors, and exporters comes from the fact that it's passed to functions like `exporter.CreateLogsFunc`. @bogdandrutu made this point recently. Given that it's passed to the telemetry constructor, maybe we can also rename this one. However, I think we need to take into account other `Settings` in the project to decide. It is probably better to start with an issue

I want to move towards https://github.com/open-telemetry/opentelemetry-collector/issues/4970#issuecomment-1911978462, which I guess would align this, but I still think this change makes sense in isolation (in that they are settings used to create the telemetry providers)

bogdandrutu (approved):

mx-psi:
Closing until we discuss #9428. Not sure if I captured the questions we want to decide on correctly, PTAL

PR #9410, opened 2024-01-28
[chore](featuregate): use errors.Join instead of multierr.Append

mmorel-35:
**Description:** use errors.Join instead of multierr.Append

**Link to tracking Issue:**  #8210

### DIFF
--- a/featuregate/flag.go
+++ b/featuregate/flag.go
@@ -4,10 +4,9 @@
 package featuregate // import "go.opentelemetry.io/collector/featuregate"
 
 import (
+	"errors"
 	"flag"
 	"strings"
-
-	"go.uber.org/multierr"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


featuregate/flag.go
```diff
@@ -59,7 +58,7 @@ func (f *flagValue) Set(s string) error {
 		case '+':
 			id = id[1:]
 		}
-		errs = multierr.Append(errs, f.reg.Set(id, val))
+		errs = errors.Join(errs, f.reg.Set(id, val))
```
TylerHelmuth, line 61: I can check on Monday, but I believe this will change the output of the error format in this case. I believe to be equivalent using errors.Join you'll need to keep a slice of errors and then join them at the end.

TylerHelmuth: I believe it needs to look like this: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5a3a3eaa01949c870a8c1cd4de7ff8203c0ad9ff/pkg/ottl/parser.go#L110-L112.  You can see the berfore-and-after here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/29332/files#diff-79e887c0c48e52e9341081ec2f052ca7b3ecc9ec98dd657e7189bc559c9a80fb

PR #9411, opened 2024-01-29
Add support of `Host` header

### LINKED ISSUE #9395: Add support of `Host` header in otlphttpexporter
We try to use opentlelemetry collector gateway behind envoy as a gateway and send data there from collector layer in lambda function. 

lambda function otel collector layer => Servicemesh gateway => otel collector gateway

Typically envoy uses authority/host headers for HTTP2/HTTP1 for request when routing
so for demonstrating the request should look like

`curl https://evnoy-service -H 'host: otel-collector-host'`

I plan to add a config for host header into otlphttp exporter 
https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/config.go
and set Request.header in the moment of request 
https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/otlp.go#L124

Similar PR for grpc exporter:
https://github.com/open-telemetry/opentelemetry-collector/pull/8228
I have a similar pull request ongoing in golang client
https://github.com/open-telemetry/opentelemetry-go/pull/4780

**Additional context**
There is a tricky part in implementation of  [net.http.Request](https://pkg.go.dev/net/http#Request) library. This library doesn't override Host header when it is passed as Request.Headers field . They have a special field Request.Host if Host header has to be overrided so configuration option for Host should be different than for generic `Headers`

enuret:
Closes #9395


config/confighttp/confighttp.go
```diff
@@ -236,9 +236,16 @@ type headerRoundTripper struct {
...
 // RoundTrip is a custom RoundTripper that adds headers to the request.
 func (interceptor *headerRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Set Host header if provided
+	hostHeader, found := interceptor.headers["Host"]
+	if found && hostHeader != "" {
+		// `Host` field should be set to override default `Host` header value which is Endpoint
+		req.Host = string(hostHeader)
+	}
 	for k, v := range interceptor.headers {
 		req.Header.Set(k, string(v))
```
jpkrohling, line 246: would Host still be available at this point?

jpkrohling (commented):
I believe this needs some visible user documentation as well (readme, likely).

config/confighttp/confighttp.go
enuret: hi, yes! It works in test that I wrote and test is broken without this change. I also just tested in my environment and it works.

jpkrohling: I took a closer look at Go's documentation, and there's one mismatch here compared to that. Go's stdlib precedence is:

- req.Host
- req.URL.Host (from HTTPServerConfig.Endpoint.Host)
- req.Headers["Host"] (from HTTPServerConfig.Headers["Host"])

Right now, we don't have an equivalent to req.Host in confighttp, so this PR introduces the following precedence order:

- HTTPServerConfig.Headers["Host"]
- HTTPServerConfig.Endpoint.Host

While I don't expect this to be a breaking change affecting any users at all given that this header must be explicitly set as part of the headers field, I would prefer if we'd introduce a `HostHeader` field, to be consistent with Go ("path of least surprise" type of thing).

But I'm not 100% convinced myself. I'd appreciate other approvers to give their opinions.

jpkrohling (commented):
I'm not 100% sure this is the way to move with this PR. Go tries to be smart by promoting the Host header to req.Host field, but at the HTTP level, it's all headers... 

I'm fine with both the approach this PR takes and with adding a new field to the config specifically for the host (like Go does with the Request object).

enuret:
I don't mind adding a separate config, whatever community decides

jpkrohling:
I added a topic to discuss today during the SIG meeting. Feel free to join us!

https://docs.google.com/document/d/1r2JC5MB7GupCE7N32EwGEXs9V_YIsPgoFiLP4VWVMkE/edit

jpkrohling (approved):
As discussed during today's SIG meeting, we have consensus that the approach this PR originally took is the desired one.

config/confighttp/confighttp.go
MadVikingGod: The `http.Request` doesn't use the Headers["Host"].  

> // For client requests, Host optionally overrides the Host
	// header to send. If empty, the Request.Write method uses
	// the value of URL.Host. Host may contain an international
	// domain name.
	Host [string](https://pkg.go.dev/builtin#string)

codeboten (approved):

Issue #9208, opened 2024-01-03
Restrict component identifiers character set


mx-psi:
Similar to what we did for feature gates on #8766, we should have a clear idea on what characters we want to allow on component names and verify it.

At a minimum, we want to disallow `/` characters in the component name and have the component name not be empty. My proposal would be to further restrict this for now to an ASCII alphanumeric + `_` identifier, starting with an ASCII alphabetic character.

This would be a breaking change, so it is needed before GA of the component package. We should do it as part of building a `component.Type` or a `component.ID`.

PR #9414, opened 2024-01-29
[component] Add MustType constructor for component.Type

mx-psi:
**Description:** 

- Adds `component.MustType` to create a type. This function panics if the type has invalid characters.
- Adds `component.Type.String` to recover the string
- Use `component.MustType` and `component.Type.String` everywhere in this codebase. To do this I changed `component.Type` into an opaque struct and checked for compile-time errors.

Some notes:

1. All components currently on core and contrib follow this rule. This is still breaking for other components.
2. A future PR will change this into a struct, to actually validate this (right now you can just do `component.Type("anything")` to bypass validation). I want to do this in two steps to avoid breaking contrib tests: we first introduce this function, and after that we change into a struct.

**Link to tracking Issue:** Updates #9208


mx-psi:
I wonder if I should add a function to do `component.NewID(component.MustType(<string literal>))` in one go, since it is a *very* common pattern in tests and internals.

component/config_test.go
```diff
@@ -194,3 +195,222 @@ func TestValidateConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestNewType(t *testing.T) {
+	tests := []struct {
+		name      string
+		shouldErr bool
+	}{
+		{name: "active_directory_ds"},
```
mx-psi, line 204: These are all the component types we have today, at least for components that have `metadata.yaml`

component/config.go
```diff
@@ -110,6 +112,46 @@ func callValidateIfPossible(v reflect.Value) error {
...
 
+var _ fmt.Stringer = (Type)("")
+
+// String returns the string representation of the type.
+func (t Type) String() string {
+	return string(t)
+}
+
+// typeRegexp is used to validate the type of a component.
+// A type must start with an ASCII alphabetic character and
```
TylerHelmuth, line 123: I dont think there are any examples in Contrib, but is there any fixed reason why someone couldnt start a component type with a number?

component/config.go
```diff
@@ -110,6 +112,46 @@ func callValidateIfPossible(v reflect.Value) error {
...
+// can only contain ASCII alphanumeric characters and '_'.
+var typeRegexp = regexp.MustCompile(`^[a-zA-Z][0-9a-zA-Z_]*$`)
+
+// newType creates a type. It returns an error if the type is invalid.
+// A type must
+// - have at least one character,
+// - start with an ASCII alphabetic character and
+// - can only contain ASCII alphanumeric characters, '-' and '_'.
```
TylerHelmuth, line 131: This comment mentions `-` but I dont see `-` being allowed in `typeRegexp`

component/config.go
```diff
@@ -110,6 +112,46 @@ func callValidateIfPossible(v reflect.Value) error {
...
+	}
+	return Type(ty), nil
+}
+
+// MustType creates a type. It panics if the type is invalid.
+// A type must
+// - have at least one character,
+// - start with an ASCII alphabetic character and
+// - can only contain ASCII alphanumeric characters, '-' and '_'.
+func MustType(strType string) Type {
```
bogdandrutu, line 147: We use "MustNewType" And "NewType" as canonical ways to name constructors.

bogdandrutu (approved):

evan-bradley:
>I wonder if I should add a function to do `component.NewID(component.MustType(<string literal>))` in one go, since it is a very common pattern in tests and internals.

It sounds like we want to also validate the component name according to what you wrote in https://github.com/open-telemetry/opentelemetry-collector/issues/9208, could we transition from `NewID` to `MustNewID` and validate `component.MustNewID("mytype")`? In my opinion we shouldn't require an extra function call to do the validation. This would also line up with `component.MustNewIDWithName("mytype", "myname")`.

component/config.go
mx-psi: One argument in favor of not allowing numbers is that it makes the component types conform to the [Unicode default identifier syntax](https://unicode.org/reports/tr31/#D1). If we ever want to expand from ASCII to the full set of identifier allowed by Unicode, we should restrict it now.

Another argument is just that we should start being conservative about the set of identifiers we allow. Restricting the set of identifiers would be a breaking change, expanding the set of identifiers is not a breaking change. So if in doubt, we should leave it out for now and think about it later.

component/config.go
mx-psi: Fixed!

component/config.go
mx-psi: Changed, thanks

bogdandrutu (approved):
Thanks for this, some feedback:
* I would prefer to also have a `NewType` with error since some may not feel comfortable with panic (not sure why, but good to have it I think).
* Can metadatagen actually check the type and fail to generate code if the type is not ok?

component/config.go
TylerHelmuth: You've convinced me

component/identifiable.go
```diff
@@ -26,11 +26,23 @@ func NewID(typeVal Type) ID {
 	return ID{typeVal: typeVal}
 }
 
+// NewID builds a Type and returns a new ID with the given Type and empty name.
```
evan-bradley, line 29:
```suggestion
// MustNewID builds a Type and returns a new ID with the given Type and empty name.
```

evan-bradley (approved):

component/identifiable.go
mx-psi: Done!

mx-psi:
:thinking: This test is failing consistently, but I am not sure how it is related to the changes on this PR

```
--- FAIL: TestTelemetryEnabled (0.22s)
    awsxray_test.go:87: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/awsxrayexporter/awsxray_test.go:87
        	Error:      	Not equal: 
        	            	expected: int(1)
        	            	actual  : int64(0)
        	Test:       	TestTelemetryEnabled
    awsxray_test.go:88: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/awsxrayexporter/awsxray_test.go:88
        	Error:      	Not equal: 
        	            	expected: int(1)
        	            	actual  : int64(0)
        	Test:       	TestTelemetryEnabled
    awsxray_test.go:89: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/awsxrayexporter/awsxray_test.go:89
        	Error:      	Should be true
        	Test:       	TestTelemetryEnabled
    awsxray_test.go:91: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/awsxrayexporter/awsxray_test.go:91
        	Error:      	Not equal: 
        	            	expected: int(1)
        	            	actual  : int64(0)
        	Test:       	TestTelemetryEnabled
FAIL
FAIL	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awsxrayexporter	0.810s
```

cmd/mdatagen/validate.go
```diff
@@ -29,10 +30,20 @@ func (md *metadata) Validate() error {
 	return errs
 }
 
+// typeRegexp is used to validate the type of a component.
+// A type must start with an ASCII alphabetic character and
+// can only contain ASCII alphanumeric characters and '_'.
+// This must be kept in sync with the regex in component/config.go.
+var typeRegexp = regexp.MustCompile(`^[a-zA-Z][0-9a-zA-Z_]*$`)
+
 func (md *metadata) validateType() error {
```
bogdandrutu, line 39: I would call `NewType` and ignore return value, only return error.

bogdandrutu (approved):
Can merge after fixing or disagreeing with my nit comment.

component/config.go
```diff
@@ -110,6 +112,47 @@ func callValidateIfPossible(v reflect.Value) error {
 // Type is the component type as it is used in the config.
 type Type string
 
+var _ fmt.Stringer = (Type)("")
```
TylerHelmuth, line 115: nit: did @bogdandrutu want to start moving these to the test files? This file already needs to use `fmt` so IDK if there is a benefit to moving it.

TylerHelmuth (approved):

component/config.go
atoulme: See https://github.com/open-telemetry/opentelemetry-collector/issues/9431

cmd/mdatagen/validate.go
mx-psi: Agree that that would be better, but it's not easy to do: it would mean adding a `replace` statement on `cmd/mdatagen/go.mod` for `component`, and that makes `go install` fail. See e.g. open-telemetry/opentelemetry-collector-contrib/issues/27855 for a recent instance of this issue with `telemetrygen`

component/config.go
mx-psi: Moved!

TylerHelmuth (approved):

mx-psi:
I have limited visibility into contrib tests because they are broken because of unrelated reasons, but I am reasonably sure that all the tests related to this have been fixed on contrib, so I will merge this by EOD

PR #9415, opened 2024-01-29
[receiver/otlp] Switch from InvalidArgument to Internal

TylerHelmuth:
**Description:** 
The otlp receiver was recently updated via https://github.com/open-telemetry/opentelemetry-collector/pull/8080 to properly propagate consumer errors back to clients as either permanent or retriable. The code we're using to indicate a non-retriable error is `codes.InvalidArgument`, which is the equivalent of `400` in HTTP.  

While 100% correct according to the [OTLP specification](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures) to indicate a non-retriable error, I think `codes.Internal` (which is equivalent to HTTP `500`), better conveys the actual state of the collector in these situations.

Related to https://github.com/open-telemetry/opentelemetry-collector/pull/9357#discussion_r1467011281

**Testing:** 
Updated tests


bogdandrutu (approved):

TylerHelmuth:
@evan-bradley my reasoning for switching to `Internal` is more based on my knowledge of HTTP than gRPC, so please correct me if I'm wrong.

My thinking was that `InvalidArgument` (or in my mind `400`), is something that a receiver would return to the client if the incoming request wasn't proper.  But once the data has made it to the next consumer, the receiver must have validated and "approved" the request, translating it into pdata, meaning it is no longer possible for the client to have pushed a incorrect payload.  If we see a permanent error in a consumer, my thinking was that it must be because a consumer messed up.

If a consumer is messing up but we return `InvalidArgument` it is possible the grpc client could interpret the code as their fault and they could think they need to change something in there payload.  If we send back a `Internal` it is clearer that it is the Collectors fault.

The spec does say `If more appropriate, another gRPC status code may be used. `, so we are welcome to use a different code if appropriate.

> Most exporters use the exporter helper which is async by default, so generally errors returned from the OTLP receiver don't come from backends and therefore are likely issues with a pipeline.

I'll add that for the case of propagating errors back up the consumer chain, the otlpreceiver is only making these code assumptions for errors that are not already `status.Status` errors. If we propagate the errors along the pipeline as `status.Status` (or something that implements `GRPCStatus() *Status`) then the otlpreceiver will send them along in their original state.

For the sake of argument, I'll provide a counter-argument to this change: it is possible there is a processor that expects specific data to be present in the payload and errors otherwise.  IDT anything like that exists in contrib, but in that theoretical scenario an InvalidArgument would be more appropriate. In my opinion this would be handled by the component returning an appropriate `status.Status` error.

evan-bradley:
>I'll add that for the case of propagating errors back up the consumer chain, the otlpreceiver is only making these code assumptions for errors that are not already `status.Status` errors.

I think that's a good point. For the cases where we can say this is an error with the data sent to the Collector, we can have the component with this context make that decision. Otherwise we can assume it's an issue originating within the Collector.

.chloggen/update-GetStatusFromError-code.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
evan-bradley, line 4:
```suggestion
change_type: breaking
```
Should this be a breaking change since we're changing the status code that's returned?

.chloggen/update-GetStatusFromError-code.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: receiver/otlp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update grcp code from `codes.InvalidArgument` to `codes.Internal` when propagating consumer error
```
evan-bradley, line 10:
```suggestion
note: Update gRPC code from `codes.InvalidArgument` to `codes.Internal` when a permanent error doesn't contain a gRPC status
```

receiver/otlpreceiver/internal/errors/errors.go
```diff
@@ -17,7 +17,7 @@ func GetStatusFromError(err error) error {
 		// https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#failures
 		code := codes.Unavailable
 		if consumererror.IsPermanent(err) {
-			code = codes.InvalidArgument
+			code = codes.Internal
```
evan-bradley, line 20:
```suggestion
			// If an error is permanent but doesn't have an attached gRPC status, assume it is server-side.
			code = codes.Internal
```
Or wording of your choice. I just think we should leave a note on why we chose this code since I don't think the spec link above will adequately explain it.

evan-bradley (approved):

codeboten (approved):

PR #9427, opened 2024-01-30
[configopaque] Mark as stable

### LINKED ISSUE #9167: Stabilize configopaque
Before stabilizing a module, an approver or maintainer must make sure that the following criteria have been met for at least two successive minor version releases:

*Edited by @mx-psi on 2023-01-12 to add ~~three~~ two more issues*

- [x] No open issues or PRs in the module that would require breaking changes\
    - [x] #9272
    - [x] #9274 
- [x] No TODOs in the module code that would require breaking changes
- [x] No deprecated symbols in the module
- [x] No symbols marked as experimental in the module
- [x] The module follows the [Coding guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md)

Please also make sure to publicly announce our intent to stabilize the module on:

- [x] The #otel-collector CNCF Slack Channel
- [x] The #opentelemetry CNCF Slack channel
- [x] A Collector SIG meeting (if unable to attend, just add to the agenda)

To help other people verify the above criteria, please link to the announcement and other links used to complete the above in a comment on this issue.

Once all criteria are met, close this issue by moving this module to the `stable` module set.

mx-psi:
**Description:** Fixes #9167

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -7,6 +7,7 @@ module-sets:
     modules:
       - go.opentelemetry.io/collector/featuregate
       - go.opentelemetry.io/collector/pdata
+      - go.opentelemetry.io/collector/config/configopaque
   beta:
     version: v0.93.0
     modules:
@@ -20,7 +21,6 @@ module-sets:
       - go.opentelemetry.io/collector/config/configgrpc
       - go.opentelemetry.io/collector/config/confighttp
       - go.opentelemetry.io/collector/config/confignet
-      - go.opentelemetry.io/collector/config/configopaque
       - go.opentelemetry.io/collector/config/configretry
       - go.opentelemetry.io/collector/config/configtelemetry
       - go.opentelemetry.io/collector/config/configtls


songy23 (approved):

bogdandrutu (commented):
Feedback that I would like to make sure we address:
* Should we apply the same reasoning as @mx-psi did for `component.Type` and make this a struct that has a for example a `string Raw()` to access the opaque string?
* https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configopaque/opaque_test.go uses yaml/v3 but we don't use that to unmarshal our configs. Should we actually test with `confmap`?
* Not necessary impacting this, but would like all the implementation checks to be moved to tests (see https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configopaque/opaque.go#L17) so that the linker does unnecessary bring that dependency to non-test code if unnecessary. In this case is a system func so not that worry, but want to set a pattern.

mx-psi:
> Should we apply the same reasoning as @mx-psi did for component.Type and make this a struct that has a for example a string Raw() to access the opaque string?

Not all strings are valid `component.Type`s, but all strings are valid `configopaque.String`s, so I don't think we get anything out of making this a struct.

bogdandrutu:
> Not all strings are valid component.Types, but all strings are valid configopaque.Strings, so I don't think we get anything out of making this a struct.

Indeed, but I am worried about usability, I feel that explicitly have to call a func to get access to the underlying string clarifies the usage intent. This is not a blocker, but want to make sure we agree that is not an improvement and doesn't worth the effort.

atoulme:
> * Not necessary impacting this, but would like all the implementation checks to be moved to tests (see https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configopaque/opaque.go#L17) so that the linker does unnecessary bring that dependency to non-test code if unnecessary. In this case is a system func so not that worry, but want to set a pattern.

https://github.com/open-telemetry/opentelemetry-collector/pull/9430/files is open to follow up on this.

mx-psi:
Summary table of pros/cons (happy to edit and add more):

|      | `configopaque.String` is a string alias                                                          | `configopaque.String` is a struct                                                                                       |
|------|--------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------|
| Pros | - You can assign a string literal to `configopaque.String` fields <br><br> - No need for breaking changes | - More explicit way of getting the value of a `configopaque.String`                                                     |
| Cons | - Less explicit way of getting the value out of a `configopaque.String` (type cast)              | - You need to explicitly use the constructor to assign to a `configopaque.String` field <br><br> - We need to break everybody :( |

mx-psi:
> Should we actually test with confmap?

The way I see it we should test this on confmap:
1. `confmap` should work correctly with Go types that conform to standard Go interfaces
2. `configopaque` implements all the standard Go interfaces
3. Therefore --> `confmap` should behave nicely with `configopaque`, and it's on `confmap` to test this

bogdandrutu:
@mx-psi then why do we test it with `yaml/v3`? What is special about it?

bogdandrutu:
> But also, it be a big breaking change - at this point the use is quite widespread - the benefits may not outweigh the costs.

We can do it to not break contrib that bad:
* Add constructor; Move everyone to this
* Add getter func; Move everyone to this

Then change.

mx-psi:
> @mx-psi then why do we test it with `yaml/v3`? What is special about it?

Nothing special, just something that uses `encoding.TextMarshaler`. Happy to change that test something else

TylerHelmuth:
Looking at the pro/con table I'm not convinced that a change to a struct at this time is necessary.  I think enough custom components outside Contrib are currently using this package that the breaking change would be significant.  We've not received any user feedback that casting to a string is tricky and we've identified that expectation in the godoc. My opinion is that we move forward with the type as a string.

mx-psi:
I filed:

- #9442 for moving the tests around
- #9441 for continuing the discussion on string vs type alias. I want to do a poll to see what people think.

bogdandrutu:
@mx-psi should we add also an rc phase for this?

mx-psi:
@bogdandrutu On #8975 we decided not to do more release candidates, and instead be more time in beta. The criteria to be able to pass from beta to stable are the ones listed on #9167 (but of course ultimately we need approval from all maintainers :)).

Is there anything specific that worries you about this package?

atoulme (approved):

mx-psi:
@open-telemetry/collector-approvers as discussed offline, I believe we can merge this, PTAL

djaglowski (approved):

astencel-sumo (approved):
:tada: :rocket:

mx-psi:
Will wait until next Wednesday to merge this to get more feedback

dmitryax (approved):

jpkrohling (approved):

PR #9437, opened 2024-01-31
[mdatagen] Add checks for README.md and metadata.yaml files

atoulme:
**Description:**
Add checks for README.md and metadata.yaml files as part of the generated tests created by mdatagen.

These tests replace the Makefile targets https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/Makefile#L333-L341 executed through github here https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/build-and-test.yml#L187-L190.

Moving to generated tests removes the dependency on cmd/otelcontribcol.

**Testing:**
Tested manually in contrib.

### DIFF
--- a/cmd/mdatagen/templates/component_test.go.tmpl
+++ b/cmd/mdatagen/templates/component_test.go.tmpl
@@ -53,6 +53,13 @@ func (aneh *assertNoErrorHost) ReportFatalError(err error) {
 	assert.NoError(aneh, err)
 }
 
+func TestREADMEmdPresent(t *testing.T) {
+	require.FileExists(t, "README.md")
+}
+
+func TestMetadataYamlPresent(t *testing.T) {
+	require.FileExists(t, "metadata.yaml")
+}
 
 {{ if isExporter }}
 func Test_ComponentLifecycle(t *testing.T) {
--- a/.chloggen/add_generated_tests_mdatagen.yaml
+++ b/.chloggen/add_generated_tests_mdatagen.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: mdatagen
+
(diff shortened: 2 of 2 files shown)


cmd/mdatagen/templates/component_test.go.tmpl
```diff
@@ -53,6 +53,13 @@ func (aneh *assertNoErrorHost) ReportFatalError(err error) {
 	assert.NoError(aneh, err)
 }
 
+func TestREADMEmdPresent(t *testing.T) {
+	require.FileExists(t, "README.md")
+}
+
+func TestMetadataYamlPresent(t *testing.T) {
+	require.FileExists(t, "metadata.yaml")
+}
```
bogdandrutu, line 62: Not sure I understand how this can happen otherwise since it is generated by that file. If anything we can have this as a requirement in our ci/cd.

cmd/mdatagen/templates/component_test.go.tmpl
```diff
@@ -53,6 +53,13 @@ func (aneh *assertNoErrorHost) ReportFatalError(err error) {
 	assert.NoError(aneh, err)
 }
 
+func TestREADMEmdPresent(t *testing.T) {
+	require.FileExists(t, "README.md")
+}
```
bogdandrutu, line 58: Not sure about this requirement though. I think this is a requirement for our ci/cd not for tests.

cmd/mdatagen/templates/component_test.go.tmpl
atoulme: you're right. I'll drop it then.

cmd/mdatagen/templates/component_test.go.tmpl
atoulme: Not sure how to enforce it otherwise. The ci/cd is using the list of components from otelcontribcol right now to check file presence.

bogdandrutu: Why do we care if other components have a README (that are not in our repos)? In our repos we use ci/cd and can use for example the list from the `versions.yaml` and enforce that each module has a README?

atoulme: I will look at using the list from `versions.yaml`, thanks!

PR #9439, opened 2024-01-31
[mdatagen] Remove use of ReportFatalError in generated tests

atoulme:
**Description:**
Remove use of ReportFatalError in generated tests

**Link to tracking Issue:**
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30579

### DIFF
--- a/cmd/mdatagen/templates/component_test.go.tmpl
+++ b/cmd/mdatagen/templates/component_test.go.tmpl
@@ -10,7 +10,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/component/componenttest"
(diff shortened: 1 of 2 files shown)


cmd/mdatagen/templates/component_test.go.tmpl
```diff
@@ -33,27 +32,6 @@ import (
 {{ end }}
 )
 
-// assertNoErrorHost implements a component.Host that asserts that there were no errors.
```
bogdandrutu, line 36: Should we instead generate an assert no error StatusReporter?

atoulme: I'd do it in phases. I am fairly certain that if we generate an assert no error StatusReporter, tests will fail because it's ok for them to report status errors on start if say we have no database to connect to.

I would also create this reporter in a common lib rather than in every generated test.

bogdandrutu: I think we can assert if the status is fatal not on error.

bogdandrutu (approved):

Issue #9162, opened 2023-12-20
Provide warning when `$` is used in config for env var expansion


bryan-aguilar:
From Collector SIG Discussion 12/20/2023

Use of the `$` in collector configurations for environment variable substitution will be deprecated in the future in favor of `${}` or `${env:foo}`.  It would be nice if we started giving users warnings during collector startup if `$` is used for env variable expansion to ease the transition pain.

crobert-1:
+1 on the warning message. It should also include the recommended syntax for expansion so users know exactly how to fix it as soon as they're aware the current method is being deprecated.

tomasmota:
I can make a pr on this

mx-psi:
@tomasmota assigned to you :)

tomasmota:
thanks @mx-psi , maybe you can give me a hint here. It seems like the only point at which we "know" we are dealing with a $ENV is inside this function we pass to os.Expand. The thing is that it can be ran several times and we probably only want to print the warning once, so I suppose the solution would be setting some sort of flag from inside it? Not really sure how to approach it.

mx-psi:
Hey @tomasmota, sorry for the delay. 

> It seems like the only point at which we "know" we are dealing with a $ENV is inside this function we pass to os.Expand

I would do this in the body of `expandEnv` (the function that calls `os.Expand`). Note that we want to log a warning if somebody is using the `$ENV` syntax, but not if they are using `${ENV}`. `os.Expand` doesn't tell you this, so you probably want to use some sort of regex.

> The thing is that it can be ran several times and we probably only want to print the warning once, so I suppose the solution would be setting some sort of flag from inside it? Not really sure how to approach it.

If you want to log once per environment variable, I think your best option is to keep a set of the environment variables for which you have already logged. You can add a new field to the `converter` struct of type `map[string]struct{}` and keep track of which variables you have already logged for.

One more thing here is that you don't really have a logger today. You can start with using a noop logger while I and other work on passing a logger to the expandconverter

mx-psi:
I just merged #9547 that adds the logging, we now need to actually pass a logger to close this!

mx-psi:
@tomasmota There is a draft PR (#9908) by @ankitpatel96 on this, we were discussing how to do this also with @evan-bradley. If you have an opinion on how to approach this, could you leave a review on #9908 or create an alternative PR?

PR #9443, opened 2024-01-31
[confmap] Pass ConverterSettings and ProviderSettings to converters and providers

mx-psi:
**Description:** 

For both #5615 and #9162 we need to be able to log during the confmap resolution.

My proposed solution is to pass a `*zap.Logger` to converters and providers during initialization. These components can then use this to log any warnings they need. This PR does the first step: being able to pass anything to converters and providers during initialization.

The obvious alternative to this is to change the interface of `confmap.Provider` and `confmap.Converter` to pass any warnings in an explicit struct. I think the `*zap.Logger` alternative is more natural for developers of providers and converters: you just use a logger like everywhere else in the Collector.

One problem for the Collector usage of `confmap` is: How does one pass a `*zap.Logger` before knowing how a `*zap.Logger` should be configured? I think we can work around this by:
1. Passing a special 'deferred' Logger that just stores the warnings without actually logging them (we can use something like `zaptest/observer` for this)
2. Resolving configuration
3. Building a `*zap.Logger` with said configuration
4. Logging the entries stored in (1) with the logger from (3) (using `zaptest/observer` we can do that by taking the `zapcore.Core` out of the logger and manually writing)

**We don't actually need ProviderSettings today, just ConverterSettings, but I think it can still be useful.**

**Link to tracking Issue:** Relates to #5615 and #9162

### DIFF
--- a/confmap/provider.go
+++ b/confmap/provider.go
@@ -8,6 +8,10 @@ import (
 	"fmt"
 )
 
+// ProviderSettings are the settings to initialize a Provider.
+// Any Provider should take this as a parameter in its constructor.
+type ProviderSettings struct{}
+
 // Provider is an interface that helps to retrieve a config map and watch for any
 // changes to the config map. Implementations may load the config from a file,
 // a database or any other source.
--- a/confmap/provider/yamlprovider/provider.go
+++ b/confmap/provider/yamlprovider/provider.go
@@ -38,7 +38,7 @@ func NewWithSettings(_ confmap.ProviderSettings) confmap.Provider {
 // Examples:
 // `yaml:processors::batch::timeout: 2s`
 // `yaml:processors::batch/foo::timeout: 3s`
-// Deprecated: Use NewWithSettings instead.
+// Deprecated: [v0.94.0] Use NewWithSettings instead.
 func New() confmap.Provider {
 	return NewWithSettings(confmap.ProviderSettings{})
 }
--- a/confmap/provider/yamlprovider/provider.go
+++ b/confmap/provider/yamlprovider/provider.go
@@ -16,6 +16,19 @@ const schemeName = "yaml"
 
 type provider struct{}
 
+// NewWithSettings returns a new confmap.Provider that allows to provide yaml bytes.
+//
+// This Provider supports "yaml" scheme, and can be called with a "uri" that follows:
+//
+//	bytes-uri = "yaml:" yaml-bytes
+//
+// Examples:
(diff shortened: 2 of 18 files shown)


confmap/provider/yamlprovider/provider.go
```diff
@@ -25,6 +38,7 @@ type provider struct{}
 // Examples:
 // `yaml:processors::batch::timeout: 2s`
 // `yaml:processors::batch/foo::timeout: 3s`
+// Deprecated: Use NewWithSettings instead.
 func New() confmap.Provider {
 	return &provider{}
```
evan-bradley, line 43:
```suggestion
	return NewWithSettings(confmap.ProviderSettings{})
```

confmap/provider.go
```diff
@@ -8,6 +8,10 @@ import (
 	"fmt"
 )
 
+// ProviderSettings are the settings to initialize a Provider.
+// Any Provider should take this as a parameter in its constructor.
+type ProviderSettings struct{}
```
bogdandrutu, line 13: Should we have "global" settings or each provider defines it's own settings? I would say better to have each provider with it's settings. The argument that we can have a property added here and passed everywhere is not strong since we still need to change each provider constructor to use it, so because of that I would simply go with a `Settings` struct in each package and deal with possible duplicate definition of properties.

Pros:
 * Can involve independently;

Cons:
 * Duplicate properties;

bogdandrutu (approved):

confmap/provider.go
mx-psi: Having independent settings would make harder fixing #4759: the builder would need to know what settings to provide for each component. With a single shared settings struct that we control we don't have that problem. In general, we have not allowed settings to vary across components (e.g. for exporters, connectors, extensions, receivers, processors... they each have a different but shared set of settings to initialize components), why should we do it in this case? Do you have specific ideas as to what settings could be independent across providers?

bogdandrutu:
> We don't actually need ProviderSettings today, just ConverterSettings, but I think it can still be useful.

At one point I was very close to remove `EnvConvertor` (for some personal reasons I dropped that effort to not have to deal with more issues). Is that an alternative to both issues?
* We have expanding available (and recursive) in the confmap resolver which supports almost all cases of the `EnvConvertor`.

confmap/provider.go
bogdandrutu: You are correct, I think we should stick with your proposal.

Issue #9388, opened 2024-01-24
[configcompression] Standardize enums


TylerHelmuth:
For `CompressionType` there are some things:
* Capitalization does not make that much sense
* Maybe rename to `Compression`, but anyway we also need to fix the enum value names.

_Originally posted by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/issues/6490#issuecomment-1307663325_

TylerHelmuth:
There is a lot of good history for this package here: https://github.com/open-telemetry/opentelemetry-collector/pull/4651. This is actually the only PR ever made for `configcompression`.

dmitryax:
> The values do no need "Capitalized"

I agree

> The const names are proper

It's inconsistent with other types constants that start with the type. Ideally should be `CompressionTypeGzip` instead of `Gzip` but given that the package is `configconmpression` maybe unnecessary.

> `CompressionType` is a good name

If we leave 2 as is, I'd say we should change `CompressionType` to `Compression`.

We can also have `configconmpression.Type` and `configconmpression.TypeGzip`. That would be consistent with all other constants in the project.

TylerHelmuth:
> It's inconsistent with other types constants that start with the type. Ideally should be CompressionTypeGzip instead of Gzip but given that the package is configconmpression maybe unnecessary.

I believe this discussion came up in the original PR and `Compression` was dropped bc of the package name.

> We can also have configconmpression.Type and configconmpression.TypeGzip. That would be consistent with all other constants in the project.

I see that this would align with `component.Type` but what else?

dmitryax:
> I see that this would align with component.Type but what else?

What do you mean what else? If we have `configconmpression.Type` as type and `configconmpression.TypeGzip` as consts it'll be consistent with all other use cases in Collector, e.g. `component.DataType` and `component.DataTypeTraces`

TylerHelmuth:
Current situations we should finalize for enum naming:

1. A package contains the "object" in the package name and contains enums for that "object"
    - such as `component.Type` and the proposed `configcompression.Type` 
2. A package contains enums but they are not directly related to the package name.
    - such as the [proposed `confignet.TransportType`](https://github.com/open-telemetry/opentelemetry-collector/pull/9385#discussion_r1465587930) and `pcommon.ValueType`.
    
In situation 1 should we standardize on the enum being named `Type`? If we do, does that mean `pmetric.MetricType` is out of compliance with the standard?

[...]

PR #9417, opened 2024-01-29
[configcompression] Rename Enums to match standard

TylerHelmuth:
**Description:** 
If we choose to keep the name

**Link to tracking Issue:** 
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/9388

PR #9447, opened 2024-01-31
Add enumeration expectations

TylerHelmuth:
**Description:** 
Adds expectations for how the repository handle enumerations

Relates to https://github.com/open-telemetry/opentelemetry-collector/issues/9388

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -251,6 +251,20 @@ To keep naming patterns consistent across the project, naming patterns are enfor
   - `func CreateTracesExport(...) {...}`
   - `func CreateTracesToTracesFunc(...) {...}`
 
+### Enumerations
+
+To keep naming patterns consistent across the project, enumeration patterns are enforced to make intent clear:
+
+- Enumerations should be defined using a type alias
+- Enumerations should use either `int` or `string` as the underlying type
+- Enumeration name should be:
+  - Prefixed with a clarifying name *unless* the package name describes the entity of the enumeration.
+  - Suffixed with `Type`
+- Constant values of an enumeration should include the enumeration type name in the name:
+  - `pcommon.ValueTypeStr`
+  - `pmetric.MetricTypeGauge`
+
+
 ### Recommended Libraries / Defaults
 
 In order to simplify developing within the project, library recommendations have been set
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -255,14 +255,14 @@ To keep naming patterns consistent across the project, naming patterns are enfor
 
 To keep naming patterns consistent across the project, enumeration patterns are enforced to make intent clear:
 
-- Enumerations should be defined using a type alias
+- Enumerations should be defined using a type definition, such as `type Level int32`.
 - Enumerations should use either `int` or `string` as the underlying type
 - Enumeration name should be:
   - Prefixed with a clarifying name *unless* the package name describes the entity of the enumeration.
   - Suffixed with `Type`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -251,6 +251,20 @@ To keep naming patterns consistent across the project, naming patterns are enfor
...
 
+### Enumerations
+
+To keep naming patterns consistent across the project, enumeration patterns are enforced to make intent clear:
+
+- Enumerations should be defined using a type alias
+- Enumerations should use either `int` or `string` as the underlying type
+- Enumeration name should be:
+  - Prefixed with a clarifying name *unless* the package name describes the entity of the enumeration.
+  - Suffixed with `Type`
```
TylerHelmuth, line 262: Starting with a strict statement here so we can poke holes at this.

Immediately an enumeration like `component.Kind`, which has constants like `ReceiverKind`, would not match this role.

CONTRIBUTING.md
```diff
@@ -251,6 +251,20 @@ To keep naming patterns consistent across the project, naming patterns are enfor
   - `func CreateTracesExport(...) {...}`
   - `func CreateTracesToTracesFunc(...) {...}`
 
+### Enumerations
+
+To keep naming patterns consistent across the project, enumeration patterns are enforced to make intent clear:
+
+- Enumerations should be defined using a type alias
```
dmitryax, line 258: I don't agree with this. I think it should be type definition as we have already implemented across the project

CONTRIBUTING.md
dmitryax: Why is `Type` as a literal required? Enums should be prefixed with _a type_. E.g. `configtelemetry.LevelNone` when we have `type Level int32`

CONTRIBUTING.md
TylerHelmuth: You're right, I've used the wrong term

CONTRIBUTING.md
TylerHelmuth: > An enum const should be prefixed with its type.

Agreed, I'll update that rule. 

I think the largest question of the PR is "When should an Enum use `Type` in its name or be named `Type`". I've started the PR by claiming "All Enums should include `Type` in the name".

I don't think that is correct, but I am struggling to put into words the situations when `Type` is appropriate.

dmitryax: > I don't think that is correct, but I am struggling to put into words the situations when Type is appropriate.

I don't think we should dictate using `Type` as literal. Sometimes it may be a good fit, sometimes not

TylerHelmuth: @dmitryax Updated to try and put into words what `Type` represents, which I am currently interpreting as a sense of limited categorization.

mx-psi (commented):
To be clear: in the case of `configcompression` this would mean basically doing the changes in #9417, right? @dmitryax do you agree?

bogdandrutu (approved):

bogdandrutu:
> To be clear: in the case of configcompression this would mean basically doing the changes in https://github.com/open-telemetry/opentelemetry-collector/pull/9417, right? @dmitryax do you agree?

I disagree, because I think having "compression" as suffix in the package name + https://github.com/open-telemetry/opentelemetry-collector/pull/9447/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R261 should lead us to rename the enum as `Type` then make values like `TypeGzip`.

mx-psi:
Ah, you are right @bogdandrutu :+1:

mx-psi (approved):

Issue #9315, opened 2024-01-18
prefix should be consistent for internal metrics


codeboten:
Currently, metrics generated by the collector include the prefix `otelcol` as configured here: https://github.com/open-telemetry/opentelemetry-collector/blob/8131d31601b886ad90b2cc8ae2e83ea96a9e2ca6/service/internal/proctelemetry/config.go#L215

This is not the case when configuring metrics to be emitted via OTLP, as the prefix configuration is specific to prometheus. I'd like to propose the prefix be applied to the metric names, so that they remain consistent when users switch from one mechanism to another.

codeboten:
Note that there's also an OTEP that has specific metrics that may require a rename later on https://github.com/open-telemetry/oteps/pull/238, but adding a prefix would still be useful in the short term.

bogdandrutu:
My vote is to not do this, and to if needed use relevant names for the meter that produces these metrics.

codeboten:
> My vote is to NOT do this, and to if needed use relevant names for the meter that produces these metrics.

@bogdandrutu As an end-user, wouldn't this potentially cause a problem? In the sense that this would rely on my backend is to interpret the meter name and apply it as a prefix if I wanted to be able to continue using my dashboards? Or I would end up having to double generate the telemetry? And then cut-over once I'm satisfied w/ the level of historical data about my collector to.

Would an alternative here would be to strip the prefix from the prometheus generated metrics via feature gate move towards using `target_info` to provide a namespace? I'm not sure if this would work, maybe @dashpole has an idea here?

I guess ultimately, what I'm trying to achieve here is a recommended path moving forward for end-users that would like to transition from prom metrics to OTLP metrics (or maybe even use both together if there's a case for this, i don't know)

dashpole:
> Would an alternative here would be to strip the prefix from the prometheus generated metrics via feature gate move towards using target_info to provide a namespace? I'm not sure if this would work, maybe @dashpole has an idea here?

The instrumentation scope is the recommended mechanism to use to "namespace" metrics in prometheus going forward.  It will appear as a label on prometheus metrics, and you will get an "otel_scope_info" metric if you add scope attributes.

But I would encourage us to try to move away from having an additional prefix only for prometheus metrics.  I'm working to allow the same names to be used in otel + prometheus in the near future, at which point we should have the exact same names.

jmacd:
I went to look for supporting evidence in the specification for metrics. 
[...]

PR #9455, opened 2024-02-02
[mdatagen] add CustomMetricName func

codeboten:
This allows components to generate a consistent metric name. Doing this in mdatagen will allow us to remove the need to export methods in helpers, for example `BuildCustomMetricName`.

Eventually we might want to define the custom metrics for each component's internal telemetry via metadata.yaml, at which point we can remove this func. For now it allows us to clean up some more code.

### DIFF
--- a/.chloggen/codeboten_mdatagen-custom-metric.yaml
+++ b/.chloggen/codeboten_mdatagen-custom-metric.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: mdatagen
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: add convenience method to generate consistent custom metric names
+
+# One or more tracking issues or pull requests related to the change
+issues: [9455]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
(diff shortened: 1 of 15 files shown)


dmitryax (approved):

bogdandrutu (commented):
I would maybe consider `connector/forward` as the meter name instead, since that would nicely group all the metrics.

codeboten:
> I would maybe consider connector/forward as the meter name instead, since that would nicely group all the metrics.

@bogdandrutu i think there's a few different ways we're currently customizing telemetry for internal components. In some cases, we're using a specific meter name:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3952e23232bf6edd4f11d024c174d3aaf6b61a82/exporter/awss3exporter/internal/metadata/generated_status.go#L19

In others we're prefixing the component details in the name of the metric:

https://github.com/open-telemetry/opentelemetry-collector/blob/f5a7315cf88e10c0bce0166b35d9227727deaa61/processor/processorhelper/obsreport.go#L36

And in yet another case, we're adding specific component in a separate label:

https://github.com/open-telemetry/opentelemetry-collector/blob/f5a7315cf88e10c0bce0166b35d9227727deaa61/exporter/exporterhelper/obsexporter.go#L64

I think it makes sense to consolidate these into a consistent experience for component owners are well as end users, I would suggest using issue https://github.com/open-telemetry/opentelemetry-collector/issues/9315 to decide on the right path to move forward. This PR allows us to hide the details of getting a custom metric name behind a generated method internal to components rather than an exposed API in the processor helper. WDYT?

bogdandrutu:
> In others we're prefixing the component details in the name of the metric:

We are doing that for legacy reasons, opencensus did not have a namespace/scope.

> And in yet another case, we're adding specific component in a separate label:

I find no problem with this, since we record this from "exporterhelper" it is a fine way of doing it. We could have created different scopes per initialization or using one scope and put that as attribute. If you ask me what I would prefer is to put that in the scope, but again when that was created we had opencensus without namespace/scope so this was not an option.

PR #9467, opened 2024-02-05
[service] introduce otelFormat feature gate

codeboten:
This feature gate will allow end users to migrate away from the telemetry produced to be opencensus compatible and over to the OpenTelemetry formatted metrics.

### DIFF
--- a/.chloggen/codeboten_emit-otel-format-metrics.yaml
+++ b/.chloggen/codeboten_emit-otel-format-metrics.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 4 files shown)


service/telemetry_test.go
```diff
@@ -299,6 +368,13 @@ func TestTelemetryInit(t *testing.T) {
 					},
 				}
 			}
+			if tc.otelFormat {
```
mx-psi, line 371: For the other flags we are explicitly passing them to `newColTelemetry`, should we follow the same pattern here?

.chloggen/codeboten_emit-otel-format-metrics.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "introduce feature gate `telemetry.useOtelFormat` to allow users a chance to migrate away from opencensus-formatted metrics"
```
mx-psi, line 10: I think this warrants an update to `docs/observability.md` explaining what the feature gate does

Issue #9429, opened 2024-01-30
Does DataType belong to component? Can we make it a type itself?


bogdandrutu:

mx-psi:
On April 3rd we discussed creating a new `configsignal` module that would contain this type as an enum. We should make sure that we can use `DataType` inside a `component.ID` or change the service config (we use a `component.ID` on this configuration to support naming pipelines something like `traces/2`)

ankitpatel96:
I'll try to come up with a solution for this one

TylerHelmuth:
With a plausible implementation example via https://github.com/open-telemetry/opentelemetry-collector/pull/10947, I'd like to take a vote on whether @open-telemetry/collector-contrib-approvers @open-telemetry/collector-approvers view this change as valueable.

Benefits of the change:
- Cleaner code to maintain.  Instead of doing `pipelineID component.ID` we'd have `piplelineID component.PipelineID`.  Using `component.Signal` is also a clearer term than `component.DataType` since Signal is an OTel term and DataType is not.
- Ability to validate pipeline names separately from component names.

Downsides of this change:
- It is almost purely an API change - end users (users running a Collector) do not care about this change.
- Breaking changes late in the life of `component`.  List of breaking changes can be found in https://github.com/open-telemetry/opentelemetry-collector/pull/10947. I _think_ it would be possible to follow a 3-release cycle deprecation strategy, but the interface changes make it harder.

If we do not do this refactor before 1.0, I do not believe it could be done as an enhancement in 1.X - it would have to be added via a 2.0 breaking change.

Vote 🚀 if you think we **should** do this refactor.
Vote 🌵 if you think we **should not** do this refactor.

mx-psi:
I would propose we give one week for voting and that we don't do the refactor in case of a tie

evan-bradley:
I'm voting yes on the refactor with the expectation that this is a straightforward change. We'll have to live with this for a while, and I think renames like this can help improve the contributor experience, particularly for new contributors.

codeboten:
The proposal in #10947 doesn't address the initial issue of moving DataType out of component correct?

TylerHelmuth:
[...]

PR #9472, opened 2024-02-06
[component] Change component.Type underlying type to a struct

### LINKED ISSUE #9208: Restrict component identifiers character set

mx-psi:
**Description:** 

Follow up to #9414 and open-telemetry/opentelemetry-collector-contrib/pull/31038.

**Link to tracking Issue:** Fixes #9208.


component/config.go
```diff
@@ -155,14 +157,18 @@ func MustNewType(strType string) Type {
 // collecting metrics, traces and logs, this can expand in the future.
 type DataType = Type
```
mx-psi, line 158: I am keeping this for now, but we can revisit once we tackle #9429

bogdandrutu: It is an alias to Type because we use `component.ID` (which is not generic) for the `pipeline.ID`. Because of that we made this equal with Type.

mx-psi:
I think open-telemetry/opentelemetry-collector-contrib/pull/31263 should fix the remaining issues with contrib tests

jpkrohling (approved):

PR #9513, opened 2024-02-07
[cmd/builder] Allow configuring confmap providers

evan-bradley:
**Description:**

Allow configuring confmap providers in the builder's config. If the field isn't set, the default set of providers is used.

**Link to tracking Issue:**

Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/4759.

**Testing:**

Extended unit tests.

**Documentation:**

Updated the readme to include the new options in the example manifest file.

cc @mx-psi

### DIFF
--- a/.chloggen/builder-confmap-providers.yaml
+++ b/.chloggen/builder-confmap-providers.yaml
@@ -16,7 +16,7 @@ issues: [4759]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  If no providers or converters are specified, the defaults are used.
+  If no providers are specified, the defaults are used.
   The default providers are: env, file, http, https, and yaml.
 
   To configure providers, use the `providers` key in your OCB build
--- a/.chloggen/builder-confmap-providers.yaml
+++ b/.chloggen/builder-confmap-providers.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: cmd/builder
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Allow configuring `confmap.Provider`s in the builder.
+
+# One or more tracking issues or pull requests related to the change
+issues: [4759]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  If no providers or converters are specified, the defaults are used.
+  The default providers are: env, file, http, https, and yaml.
+
+  To configure providers, use the `providers` key in your OCB build
+  manifest with a list of Go modules for your providers.
+  The modules will work the same as other Collector components.
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/cmd/builder/README.md
+++ b/cmd/builder/README.md
@@ -12,20 +12,20 @@ dist:
   description: Local OpenTelemetry Collector binary
   output_path: /tmp/dist
 exporters:
-  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.96.0
-  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.96.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.97.0
+  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.97.0
 
 receivers:
-  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.96.0
+  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.97.0
 
 processors:
-  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.96.0
+  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.97.0
 
 providers:
-  - gomod: go.opentelemetry.io/collector/confmap/provider/fileprovider v0.96.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/fileprovider v0.97.0
 
 converters:
-  - gomod: go.opentelemetry.io/collector/confmap/converter/expandconverter v0.96.0
+  - gomod: go.opentelemetry.io/collector/confmap/converter/expandconverter v0.97.0
 EOF
 $ builder --config=otelcol-builder.yaml
 $ cat > /tmp/otelcol.yaml <<EOF
--- a/cmd/builder/README.md
+++ b/cmd/builder/README.md
@@ -12,20 +12,20 @@ dist:
   description: Local OpenTelemetry Collector binary
   output_path: /tmp/dist
 exporters:
-  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.97.0
-  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.97.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.99.0
+  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.99.0
 
 receivers:
-  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.97.0
+  - gomod: go.opentelemetry.io/collector/receiver/otlpreceiver v0.99.0
 
 processors:
-  - gomod: go.opentelemetry.io/collector/processor/batchprocessor v0.97.0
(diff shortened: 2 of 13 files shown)


cmd/builder/internal/builder/config.go
```diff
@@ -36,6 +36,8 @@ type Config struct {
 	Receivers    []Module     `mapstructure:"receivers"`
 	Processors   []Module     `mapstructure:"processors"`
 	Connectors   []Module     `mapstructure:"connectors"`
+	Providers    *[]Module    `mapstructure:"providers"`
+	Converters   *[]Module    `mapstructure:"converters"`
```
evan-bradley, line 40: Making these optional introduces a lot of code complexity, but is necessary unless we want to take one of two routes:
1. Remove the ability for users to specify none of either and check whether the set of modules is empty. In my opinion, we should let users specify they want nothing included in their distribution in case they have unusual requirements.
2. Remove the set of default converters/providers and require the user to always specify the modules they want included.

I don't necessarily mind option 2, but it would be a fairly significant breaking change and would need a long transition period. If we take option 2, we could move the defaults list to `otelcol` and just always have the modules baked into every distro until we complete the transition.

cmd/builder/internal/builder/templates/main.go.tmpl
```diff
@@ -16,8 +23,27 @@ func main() {
...
+	}
+	set := otelcol.CollectorSettings{
+		BuildInfo: info,
+		Factories: components,
+		ConfigProviderSettings: otelcol.ConfigProviderSettings{
+			ResolverSettings: confmap.ResolverSettings{
+				Providers:  makeMapProvidersMap(providers...),
+				Converters: []confmap.Converter{
+					{{- range .Converters}}
+					{{.Name}}.New(confmap.ConverterSettings{}),
```
evan-bradley, line 39: The need to pass in `confmap.ConverterSettings{}` here indicates to me that we need to make some changes to the `otelcol` API. I will issue another PR to demonstrate what I think may be a good solution.

evan-bradley: I think if we change the APIs a bit, we can just pass the `New` function for each module and have them instantiated later on when we have a logger to pass. I think the approach in this PR will be the cleanest and most straightforward: https://github.com/open-telemetry/opentelemetry-collector/pull/9516.

mx-psi:
Both dependencies have been merged, is this ready for review?

cmd/builder/internal/builder/config.go
mx-psi: An alternative is to have a custom unmarshaling function that differentiates between unset and set but empty. I think this is cleaner and I agree that even if we go with 2 we would want to have a transition period

cmd/builder/internal/builder/templates/main.go.tmpl
mx-psi: Agreed that passing the factories here makes more sense, let's get that PR merged first

cmd/builder/internal/builder/templates/go.mod.tmpl
```diff
@@ -5,6 +5,10 @@ module {{.Distribution.Module}}
 go 1.21
 
 require (
+	go.opentelemetry.io/collector/confmap/converter/expandconverter v{{.Distribution.OtelColVersion}}
+	{{- range .Providers}}
+	{{if .GoMod}}{{.GoMod}}{{end}}
+	{{- end}}
```
mx-psi, line 11: This should be added conditionally on `{{.Distribution.OtelColVersion}}` being new enough to avoid issues like #8692

cmd/builder/internal/builder/templates/main.go.tmpl
```diff
@@ -16,8 +21,25 @@ func main() {
...
+		Factories: components,
+		ConfigProviderSettings: otelcol.ConfigProviderSettings{
+			ResolverSettings: confmap.ResolverSettings{
+				Providers:  makeMapProvidersMap(providers...),
+				Converters: []confmap.Converter{
+					expandconverter.New(confmap.ConverterSettings{}),
+				},
+			},
+		},
+	}
```
mx-psi, line 40: Ditto, this API will only work with new enough versions of the builder, so I would guard this with a check to `otelcol_version`

mx-psi (commented):
We don't have a clear policy regarding how big the difference in versions between the builder and the API used is, but given the precedent with #8692, I think we should strive to keep some compatibility with older versions of the API, at least for some amount of time. To that end we should guard uses of new modules or new API with a check of the `otelcol_version`

cmd/otelcorecol/builder-config.yaml
```diff
@@ -24,6 +24,13 @@ processors:
 connectors:
   - gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.97.0
 
+providers:
+  - gomod: go.opentelemetry.io/collector/confmap/provider/envprovider v0.95.0
```
MovieStoreGuy, line 28: Should this be set at v0.97.0 ?

mx-psi:
Will wait for #9516 before giving another pass to this one

evan-bradley:
@mx-psi I've updated this to use confmap factories and to account for https://github.com/open-telemetry/opentelemetry-collector/pull/9897.

I'm a little unclear on how we want to handle backwards compatibility now. I've updated this to not use the new APIs if the builder is working with versions before v0.99.0, but it would appear that the new strict versioning functionality will force the builder and core API versions to be equal, according to [this line](https://github.com/open-telemetry/opentelemetry-collector/blob/e9b432dcea3699dde1e0118e7a3c6df3dff36137/cmd/builder/internal/builder/main.go#L78). Is it safe to just assume that any future version of the builder will support these APIs?

mx-psi:
@evan-bradley I did not intend for that change to go through just yet, and this is also not documented, so I filed #9999 to document it. We can handle this separately.

IMO we should be a bit more lenient and allow for some version skew between the builder and the Collector being built (at least one minor version of difference should be fine based on our deprecation guidelines, and would be in line with the [Kubernetes version skew policy](https://kubernetes.io/releases/version-skew-policy/))

evan-bradley:
Thanks for the clarification. I agree we should probably allow some minor skew even if it's a bit inconvenient, it would feel pretty demanding to have to update the builder every single release. 

I think this PR should be unaffected by that change and should still be ready to go.

.chloggen/builder-confmap-providers.yaml
```diff
@@ -0,0 +1,31 @@
...
+note: Allow configuring `confmap.Provider`s in the builder.
+
+# One or more tracking issues or pull requests related to the change
+issues: [4759]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  If no providers or converters are specified, the defaults are used.
```
mx-psi, line 19:
```suggestion
  If no providers are specified, the defaults are used.
```

cmd/builder/internal/builder/main.go
```diff
@@ -79,6 +80,9 @@ func Generate(cfg Config) error {
 		}
 		cfg.Logger.Info("You're building a distribution with non-aligned version of the builder. Compilation may fail due to API changes. Please upgrade your builder or API", zap.String("builder-version", defaultOtelColVersion))
 	}
+
+	// confmap factories were introduced in v0.99.0.
+	cfg.SupportsConfmapFactories = semver.Compare("v"+cfg.Distribution.OtelColVersion, "v0.99.0") > -1
```
mx-psi, line 85: For the `RequireOtelColVersion` we do this check on cmd/builder/internal/builder/config.go on a method from config. Should we do the same here?

cmd/builder/internal/builder/main_test.go
```diff
@@ -33,12 +33,22 @@ require (
 )`)
 )
 
+func newInitializedConfig(t *testing.T) Config {
+	cfg := NewDefaultConfig()
+	// Validate and ParseModules will be called before the config is
+	// given to Generate.
+	assert.NoError(t, cfg.Validate())
```
mx-psi, line 40: We need to make this a bit less cumbersome to use, but until then, I think we should call the same functions used here https://github.com/open-telemetry/opentelemetry-collector/blob/4a5809293b286f39d1508e69e760e4b00d5da7eb/cmd/builder/internal/command.go#L61-L75

(I guess `SetGoPath` can be avoided here, but at least the ones that affect generate should be here):
```suggestion
	assert.NoError(t, cfg.Validate())
	assert.NoError(t, cfg.SetRequireOtelColModule())
```

cmd/builder/README.md
```diff
@@ -12,14 +12,20 @@ dist:
   description: Local OpenTelemetry Collector binary
   output_path: /tmp/dist
 exporters:
-  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.86.0
-  - gomod: go.opentelemetry.io/collector/exporter/debugexporter v0.86.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/alibabacloudlogserviceexporter v0.97.0
```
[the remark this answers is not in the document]
evan-bradley: I went ahead and updated these. If possible, it would be really nice to get this into the next release.

cmd/builder/internal/builder/main.go
evan-bradley: Great catch! I missed this, but it would be nice to consolidate where we handle backwards compatibility. I've refactored this so all checks are done in the same place.

cmd/builder/internal/builder/main.go
```diff
@@ -79,6 +79,7 @@ func Generate(cfg Config) error {
 		}
 		cfg.Logger.Info("You're building a distribution with non-aligned version of the builder. Compilation may fail due to API changes. Please upgrade your builder or API", zap.String("builder-version", defaultOtelColVersion))
 	}
+
```
mx-psi, line 82:
```suggestion
```

mx-psi (approved):
Will merge on Monday morning

@MovieStoreGuy block before then if you still have any concerns

PR #9516, opened 2024-02-07
[confmap] Add converter and provider settings to confmap.ResolverSettings 

evan-bradley:
**Description:**

Follows https://github.com/open-telemetry/opentelemetry-collector/pull/9443, relates to https://github.com/open-telemetry/opentelemetry-collector/pull/9513.

This builds on https://github.com/open-telemetry/opentelemetry-collector/pull/9228 to demonstrate the concept.

This shows one way of extending the otelcol APIs to allow passing converters and providers from the builder with the new settings structs for each type.

I think this approach has a few advantages:
1. This follows our pattern of passing in "factory" functions instead of instances to the object that actually uses the instances.
2. Makes the API more declarative: the settings specify which modules to instantiate and which settings to instantiate them with, but don't require the caller to actually do this.
3. Compared to the current state, this allows us to update the config at different layers. A distribution's `main.go` file can specify the providers/converters it wants and leave the settings to be created by `otelcol.Collector`.

The primary drawbacks I see here are:
1. This is a little more opinionated since you don't have access to the converter/provider instances or control how they are instantiated. I think this is acceptable and provides good encapsulation.
2. The scheme->provider map can now only be specified by the providers' schemes, which is how it is currently done by default. I would want to hear what use cases we see for more complex control here that necessitates using schemes not specified by the providers.

cc @mx-psi


mx-psi:
@evan-bradley can you fix the merge conflict? I think that would make it ready for review :)

evan-bradley:
@mx-psi Should be good-to-go now, please take a look.

confmap/resolver.go
```diff
@@ -27,18 +27,35 @@ type Resolver struct {
 	watcher chan error
 }
 
+type ConverterFactory = func(ConverterSettings) Converter
+
+type ProviderFactory = func(ProviderSettings) Provider
```
mx-psi, line 32: I assume we will probably want to make this closer to `exporter.Factory` and the like, but I am fine doing that in a different PR

confmap/resolver.go
```diff
@@ -27,18 +27,35 @@ type Resolver struct {
...
+
 // ResolverSettings are the settings to configure the behavior of the Resolver.
 type ResolverSettings struct {
 	// URIs locations from where the Conf is retrieved, and merged in the given order.
 	// It is required to have at least one location.
 	URIs []string
 
+	// ProviderFactories is a list of Provider creation functions.
+	// It is required to have at least one Provider.
```
mx-psi, line 41: Where do we verify this?

mx-psi (approved):

mx-psi:
@evan-bradley Can you address the linter issue? 
```
Error: otelcoltest/config.go:48:6: func `makeMapProvidersMap` is unused (unused)
```

confmap/resolver.go
evan-bradley: I updated the comment to handle the transition period where we handle both this field and the `Providers` field, but we handle that here: https://github.com/open-telemetry/opentelemetry-collector/pull/9516/files#diff-ced9915c3a382d1bd71cb1e86872a4f1b680424f691b2ec91e9d6982df056621R86

confmap/resolver.go
evan-bradley: That's a good idea, making this an interface would make these modules more flexible going forward. I'm okay if we want to tackle that first, but I'm also okay doing it later since it will let us more effectively use `ConverterSettings` and `ProviderSettings` in the meantime.

evan-bradley:
@mx-psi Thanks for the heads-up, I think everything should be taken care of now.

confmap/resolver.go
mx-psi: Thanks!

confmap/resolver.go
mx-psi: Thinking about this a bit more, while I am not too worried about breaking these APIs, I think we should tackle it first to avoid further complexity on the builder (we would have to check `otelcol_version` and generate different code depending on the version chosen)

evan-bradley: I missed this comment, I'll make this update shortly.

otelcol/collector.go
```diff
@@ -116,6 +116,9 @@ func NewCollector(set CollectorSettings) (*Collector, error) {
 	var err error
 	configProvider := set.ConfigProvider
 
+	set.ConfigProviderSettings.ResolverSettings.ProviderSettings = confmap.ProviderSettings{Logger: zap.NewNop()}
```
evan-bradley, line 119: I wanted to avoid a dependency on `confmaptest` so I just directly instantiated a no-op logger here in preparation for actually instantiating a logger. I'm open to alternate approaches if we want to encapsulate this somehow.

confmap/resolver.go
evan-bradley: Okay, I've updated all of the confmap modules to be created using factories. It's a little heavy, but I think it works. Let me know what you think.

otelcol/collector.go
evan-bradley: I updated `confmap.NewResolver` to use a no-op logger if no logger is provided, so this code can just be updated to pass in the actual logger once that is implemented.

mx-psi (approved):
LGTM! I agree that the factories feel a bit 'heavy', but I think this works

cc @open-telemetry/collector-approvers, will merge this tomorrow unless anyone blocks before then

confmap/converter/expandconverter/expand.go
```diff
@@ -22,6 +22,7 @@ type converter struct {
 }
 
 // New returns a confmap.Converter, that expands all environment variables for a given confmap.Conf.
+// Deprecated [v0.99.0]: Use NewFactory instead.
```
TylerHelmuth, line 25: I believe the deprecation notice needs to be on the last line (or on a solo line) for the linter to pick it up. See https://github.com/open-telemetry/opentelemetry-collector/pull/9947

confmap/resolver.go
```diff
@@ -65,10 +80,42 @@ func NewResolver(set ResolverSettings) (*Resolver, error) {
...
 	}
 
-	if len(set.Providers) == 0 {
+	if len(set.ProviderFactories) == 0 && len(set.Providers) == 0 {
 		return nil, errors.New("invalid map resolver config: no Providers")
 	}
 
+	if set.ProviderSettings.Logger == nil {
+		set.ProviderSettings.Logger = zap.NewNop()
+	}
```
TylerHelmuth, line 89: Should we be returning an error here or allowing future panics instead?  If the user passed in a `ResolverSettings` with no logger and a `ProviderFactories` that expects a logger, is it right to expect `NewResolver` to handle the mistake?

This check is making the code much safer to execute, but will it create confusion when not logs are written and no errors happen in this situation?

evan-bradley: That's a good question. The approach I'd like to take here is to make this an optional setting akin to calling an OTel language API. If you want to opt-in to logs they will be printed by some providers, otherwise it is just a no-op.

evan-bradley: I've updated `ProviderSettings` with a comment explaining this.

TylerHelmuth (approved):

mx-psi:
@evan-bradley can you address the linter on here?

```
Error: loader.go:248:13: SA1019: fileprovider.NewWithSettings is deprecated: [v0.99.0] Use NewFactory instead. (staticcheck)
	cp, err := fileprovider.NewWithSettings(confmaptest.NewNopProviderSettings()).Retrieve(context.Background(), "file:"+filePath, nil)
	           ^
```

evan-bradley:
@mx-psi The build is green! :tada:

PR #9523, opened 2024-02-07
[otlpexporter] Validate the configuration explicitly

atoulme:
**Description:**
Use `Config.Validate` to validate the presence of the gRPC endpoint instead of making an assertion at creation time.

### DIFF
--- a/exporter/otlpexporter/config.go
+++ b/exporter/otlpexporter/config.go
@@ -4,6 +4,8 @@
 package otlpexporter // import "go.opentelemetry.io/collector/exporter/otlpexporter"
 
 import (
+	"errors"
+
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configgrpc"
 	"go.opentelemetry.io/collector/config/configretry"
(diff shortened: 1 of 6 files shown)


atoulme:
Leaning towards no changelog since it feels like this is not a functional change, but please weigh in.

exporter/otlpexporter/testdata/invalid_configs.yaml
```diff
@@ -0,0 +1,55 @@
...
+    enabled: true
+    num_consumers: 2
+    queue_size: 10
+  retry_on_failure:
+    enabled: true
+    initial_interval: 10s
+    randomization_factor: -5
+    multiplier: 1.3
+    max_interval: 60s
+    max_elapsed_time: 10m
```
bogdandrutu, line 55: New line please.

exporter/otlpexporter/config.go
```diff
@@ -19,4 +21,11 @@ type Config struct {
 	configgrpc.ClientConfig `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
 }
 
+func (c *Config) Validate() error {
+	if c.SanitizedEndpoint() == "" {
```
bogdandrutu, line 25: Why sanitize here? The `newExporter` was not using this func.

atoulme: Yes, I believe this was in effect a miss. The old code checks for "", but fails to check for "http://" or "https://", and since this is dropped afterwards, may be sending an empty string to the gRPC lib.

exporter/otlpexporter/testdata/invalid_configs.yaml
atoulme: done

bogdandrutu (approved):

bogdandrutu (approved):

PR #9525, opened 2024-02-08
[chore] add code coverage for SanitizedString

atoulme:


config/configgrpc/configgrpc_test.go
```diff
@@ -1102,3 +1102,37 @@ type mockHost struct {
 func (nh *mockHost) GetExtensions() map[component.ID]component.Component {
 	return nh.ext
 }
+
+func TestSanitizedEndpoint(t *testing.T) {
```
bogdandrutu, line 1106: I kind of don't like that user has to call `SanitizedEndpoint`. I think we should have a custom `Unmarshal` that strips the prefix from the Endpoint and we have an internal `isHttps` param that we use in `toDialOptions`?

atoulme: Going to close this PR in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/9535

PR #9526, opened 2024-02-08
[component] Deprecate ErrNilNextConsumer

atoulme:
**Description:**
Deprecate ErrNilNextConsumer and implement nil checks in the builder structs.

**Testing:**
Unit testing.


bogdandrutu (approved):
I like this.

atoulme:
Opening https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31172 to unblock this PR.

receiver/receiver.go
```diff
@@ -60,23 +65,23 @@ type Factory interface {
...
+	// an error will be returned instead. `nextConsumer` is never nil.
 	CreateMetricsReceiver(ctx context.Context, set CreateSettings, cfg component.Config, nextConsumer consumer.Metrics) (Metrics, error)
 
 	// MetricsReceiverStability gets the stability level of the MetricsReceiver.
 	MetricsReceiverStability() component.StabilityLevel
 
 	// CreateLogsReceiver creates a LogsReceiver based on this config.
 	// If the receiver type does not support the data type or if the config is not valid
-	// an error will be returned instead.
+	// an error will be returned instead. `nextConsumer` is never nil.
```
codeboten, line 84: not sure i understand this comment, are you saying that `nextConsumer` must never be nil? The comment doesn't make that clear

atoulme: In the past implementors would check that nextConsumer may be nil. We are making the strong guarantee to implementors that `nextConsumer` is never nil and there's no need to check and return errors for it.

component/component.go
```diff
@@ -11,6 +11,8 @@ import (
 var (
 	// ErrNilNextConsumer can be returned by receiver, or processor Start factory funcs that create the Component if the
 	// expected next Consumer is nil.
+	// Deprecated: [v0.95.0] The next consumer is now checked as part of the creation of the pipelines.
```
atoulme, line 14:
```suggestion
	// Deprecated: [v0.96.0] The next consumer is now checked as part of the creation of the pipelines.
```

dmitryax (approved):
👍

PR #9527, opened 2024-02-08
Offer a way to deal with status report with sharedcomponent.Map

### LINKED ISSUE #9155: [internal/sharedcomponent] start and stop according to usage
Right now, we start on first start and stop on first stop.

It would be good to record the number of times the component has started and only stop it when stop has been called as many times start was called.

atoulme:
**Description:**
Given that we start and stop components of a shared component map once, but want to have idempotence, this PR changes the way it handles telemetry status report by removing explicit handling of report status entirely.

Rather, we rely on errors returned by Start and Shutdown.

For Start, we cache the error and return it on any subsequent runs.

For Shutdown, we only run the underlying Shutdown on the last call and return the error if any. Any previous calls return nil.

This change in approach allows to be closer to the lifecycle of a normal component and doesn't have the overhead of dealing with status reporting, which is handled like every other component by the service.

**Link to tracking Issue:**
Fixes #9155

**Testing:**
Unit tests

### DIFF
--- a/internal/sharedcomponent/sharedcomponent.go
+++ b/internal/sharedcomponent/sharedcomponent.go
@@ -9,6 +9,7 @@ package sharedcomponent // import "go.opentelemetry.io/collector/internal/shared
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 
 	"go.opentelemetry.io/collector/component"
 )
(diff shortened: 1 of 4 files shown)


internal/sharedcomponent/sharedcomponent.go
```diff
@@ -9,112 +9,78 @@ package sharedcomponent // import "go.opentelemetry.io/collector/internal/shared
...
+	components map[K]*componentEntry[V]
 }
 
-// LoadOrStore returns the already created instance if exists, otherwise creates a new instance
-// and adds it to the map of references.
-func (m *Map[K, V]) LoadOrStore(key K, create func() (V, error), telemetrySettings *component.TelemetrySettings) (*Component[V], error) {
+// LoadOrStore looks up the already created instance if exists, otherwise creates a new instance
+// and adds it to the map of references. It returns the wrapped shared component, the instance of
+// the component and a possible error on creation.
+func (m *Map[K, V]) LoadOrStore(key K, create func() (V, error)) (component.Component, V, error) {
```
bogdandrutu, line 32: Why removing the "Unwrap" pattern in favor of returning multiple values?

bogdandrutu (commented):
I don't understand how this PR handles the status, or why we move to shutdown to the last call (which I remember I explained that is not possible).

atoulme:
> I don't understand how this PR handles the status, or why we move to shutdown to the last call (which I remember I explained that is not possible).

I don't think we need to if we consistently return errors. For shutdown, I lost reference, but meant to ask if we could make it configurable to do first or last. WDYT?

atoulme:
I implemented a boolean parameter to allow to run stop first or last.

PR #9535, opened 2024-02-08
[otlpexporter] change handling of https:// endpoints

atoulme:
**Description:**
Remove special treatment of TLS for https endpoints.

The OTLP exporter README mentions that endpoints using https:// will override TLS credentials:
  > -If a scheme of `https` is used then client transport security is enabled and overrides the `insecure` setting.
  
This change removes this override and instead abides by the TLS configuration.


atoulme:
I reviewed with @dmitryax (thank you!) and this goes against the [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/1729/files). I will rework this PR.

PR #9537, opened 2024-02-09
[configgrpc] Clean up the logic associated with https endpoints

atoulme:
**Description:**
This PR changes the logic associated with https endpoints handling to make it more obvious that it is overriding TLS settings. It keeps all other TLS settings rather than disposing of them. It adds a warning to let the user know that they have a conflicting configuration if they both set the insecure flag to true and use a https endpoint.


config/configgrpc/configgrpc.go
```diff
@@ -192,6 +192,29 @@ func (gcs *ClientConfig) ToClientConn(ctx context.Context, host component.Host,
...
+	tlsSetting := &gcs.TLSSetting
+	if gcs.TLSSetting.Insecure {
+		// Per https://github.com/alanwest/opentelemetry-specification/blob/main/specification/protocol/exporter.md
+		// If the endpoint uses the https scheme, then the TLS insecure setting must be overridden.
+		if strings.HasPrefix(gcs.Endpoint, "https://") {
+			settings.Logger.Warn("This https endpoint overrides the TLS `insecure` setting. Set the `insecure` option to false or drop the `https` scheme to remove this warning.", zap.String("endpoint", gcs.Endpoint))
+			tlsSetting = &configtls.TLSClientSetting{}
+			*tlsSetting = gcs.TLSSetting
+			tlsSetting.Insecure = false
+		}
```
bogdandrutu, line 205: As mentioned in a different PR, we should actually have a custom Unmarshal and override the `Insecure` based on the prefix and remove the prefix from the endpoint.

bogdandrutu (changes requested):

config/configgrpc/configgrpc.go
atoulme: ah, right. Sorry. I'll go do it that way instead.

bogdandrutu: During unmarshalling you don't have a logger, and not sure it brings anything useful to the user. If you really want we can error.

atoulme: alright - how about a config converter? These have a logger.

bogdandrutu: > alright - how about a config converter? These have a logger.

They probably don't as well, since they are applied before we know the final config.

atoulme: I checked - we use log.Println in our distro :( ok then

PR #9538, opened 2024-02-09
[configgrpc] Deprecate configgrpc.SanitizedEndpoint

atoulme:
**Description:**
Deprecate the public function `SanitizedEndpoint`. It will be removed in a future release.


atoulme:
Close in favor of #9512

PR #9544, opened 2024-02-09
[chore] implement shutdown verifier for metrics and logs

atoulme:


processor/processortest/shutdown_verifier.go
```diff
@@ -44,9 +44,61 @@ func verifyTracesDoesNotProduceAfterShutdown(t *testing.T, factory processor.Fac
...
+func verifyLogsDoesNotProduceAfterShutdown(t *testing.T, factory processor.Factory, cfg component.Config) {
+	// Create a proc and output its produce to a sink.
+	nextSink := new(consumertest.LogsSink)
+	proc, err := factory.CreateLogsProcessor(context.Background(), NewNopCreateSettings(), cfg, nextSink)
+	if err != nil {
+		if errors.Is(err, component.ErrDataTypeIsNotSupported) {
+			return
+		}
+		require.NoError(t, err)
+	}
```
bogdandrutu, line 56:
```suggestion
	if errors.Is(err, component.ErrDataTypeIsNotSupported) {
		return
	}
	require.NoError(t, err)
```

processor/processortest/shutdown_verifier.go
```diff
@@ -44,9 +44,61 @@ func verifyTracesDoesNotProduceAfterShutdown(t *testing.T, factory processor.Fac
...
+	// Now shutdown the proc.
+	assert.NoError(t, proc.Shutdown(context.Background()))
+
+	// The Shutdown() is done. It means the proc must have sent everything we
+	// gave it to the next sink.
+	assert.EqualValues(t, generatedCount*2, nextSink.DataPointCount())
+}
+
 // VerifyShutdown verifies the processor doesn't produce telemetry data after shutdown.
 func VerifyShutdown(t *testing.T, factory processor.Factory, cfg component.Config) {
```
bogdandrutu, line 100: You can write a test, that also can be used as an example for the nop processor.

bogdandrutu (approved):

processor/processortest/shutdown_verifier.go
atoulme: sounds good

processor/processortest/shutdown_verifier.go
bogdandrutu: @atoulme when I comment somewhere in your PR about a refactoring like this, I expect you to apply it everywhere else. Please, also you should do it for trace.

atoulme: sure thing!

processor/processortest/shutdown_verifier.go
atoulme: test won't pass as nopprocessor doesn't pass on data - I suppose that's to be the expected behavior of the nopprocessor? Not sure. In the mean time I'll work on a simple processor accepting sinks.

codeboten (approved):

PR #9545, opened 2024-02-09
[service/extensions] Move the extensions package to internal

atoulme:
**Description:**
I am offering to move the extensions package to an internal package as I see it used only by the service package.


bogdandrutu:
>I am offering to move the extensions package to an internal package as I see it used only by the service package.

Please don't the idea is that people can use the extensions and the graph with a common service. See @mx-psi goal of using this as a library.

atoulme:
Thanks, I thought that might be the case but wanted to put a proposal out in earnest. I'm interested to learn more if anyone has info.

PR #9551, opened 2024-02-11
[receivertest] add metrics support to contract_checker

atoulme:
**Description:**
add support for metrics in contract checker

### DIFF
--- a/receiver/receivertest/contract_checker.go
+++ b/receiver/receivertest/contract_checker.go
@@ -507,28 +507,48 @@ func CreateOneLogWithID(id UniqueIDAttrVal) plog.Logs {
 	return data
 }
 
-func CreateEveryMetricTypeWithID(id UniqueIDAttrVal) pmetric.Metrics {
+func CreateGaugeMetricWithID(id UniqueIDAttrVal) pmetric.Metrics {
 	data := pmetric.NewMetrics()
 	gauge := data.ResourceMetrics().AppendEmpty().ScopeMetrics().AppendEmpty().Metrics()
 	gauge.AppendEmpty().SetEmptyGauge().DataPoints().AppendEmpty().Attributes().PutStr(
 		UniqueIDAttrName,
 		string(id),
 	)
+	return data
+}
+
+func CreateSumMetricWithID(id UniqueIDAttrVal) pmetric.Metrics {
+	data := pmetric.NewMetrics()
 	sum := data.ResourceMetrics().AppendEmpty().ScopeMetrics().AppendEmpty().Metrics()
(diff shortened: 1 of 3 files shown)


receiver/receivertest/contract_checker.go
```diff
@@ -390,7 +390,76 @@ func idSetFromLogs(data plog.Logs) (idSet, error) {
...
 }
 
-// TODO: Implement mockConsumer.ConsumeMetrics()
+func (m *mockConsumer) ConsumeMetrics(_ context.Context, data pmetric.Metrics) error {
+	ids, err := idSetFromMetrics(data)
+	require.NoError(m.t, err)
+	return m.consume(ids)
+}
+
+// idSetFromLogs computes an idSet from given plog.Logs. The idSet will contain ids of all log records.
```
dmitryax, line 399:
```suggestion
// idSetFromLogs computes an idSet from given pmetric.Metrics. The idSet will contain ids of all metric data points.
```

receiver/receivertest/contract_checker.go
```diff
@@ -437,3 +506,42 @@ func CreateOneLogWithID(id UniqueIDAttrVal) plog.Logs {
 	)
 	return data
 }
+
+func CreateEveryMetricTypeWithID(id UniqueIDAttrVal) pmetric.Metrics {
```
dmitryax, line 510: I don't think we should use the same id across several metrics. The ID is supposed to be unique per datapoint

atoulme: alright, changed

receiver/receivertest/contract_checker_test.go
```diff
@@ -52,30 +97,108 @@ func (s *exampleReceiver) Receive(data plog.Logs) {
...
+	data := CreateOneSpanWithID(id)
+
+	// Send the generated data to the receiver.
+	g.receiver.ReceiveTraces(data)
+
+	// And return the ids for bookkeeping by the test.
+	return []UniqueIDAttrVal{id}
+}
+
+func (g *exampleLogGenerator) GenerateTraces() []UniqueIDAttrVal {
```
dmitryax, line 155: Why is this method added?

atoulme: removed

receiver/receivertest/contract_checker.go
dmitryax: This isn’t addressed. Why do we need every metric type with the same ID? Can this method take metric type instead?

atoulme: Sorry, bad push. Should be good now.

dmitryax (approved):

Issue #9592, opened 2024-02-15
Go semconv seems not be consistent with otel semconv spec


aygalinc:
**Describe the bug**
Starting with the semconv 1.21.0, info in trace resource seems to not correspond to semantic convention. 
As example all info related to http is missing, is it intended ? 

I see that exporter-contrib are base on the generated semconv, so basically if trace that follow semantic convention in version > 1.20 is ingested some exporter failed to map field to trace store. We use the azuremonitor in our production system and it is not usable anymore.

aygalinc:
Link to : https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31255

codeboten:
thanks for reporting this @aygalinc, as of the [merge of v1.25.0](https://github.com/open-telemetry/opentelemetry-collector/pull/10075), the generated_attribute_group.go includes the http semconv. Please re-open if that doesn't solve the issue

PR #9593, opened 2024-02-15
feat: generate attribute group semconv file

aygalinc:
**Description:** 
Adding a feature: it provide an attribute group file in semconv generation that contains additionnal part of the semantic convention that can be use by various exporter. 

**Link to tracking Issue:** #9592


aygalinc:
I had the test but i cannot generate the semconv package with or without my change. I use the make file to do it but I hit :

aygalinc:
Fix by explicitly set the container version of semconv otel image. I think its a good practice for build reproductibility.

Issue #9594, opened 2024-02-16
[client] `Metadata` is not thread safe


TylerHelmuth:
**Describe the bug**
Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31289

I believe there is a bug in `Metadata.Get` when accessed by multiple threads.  

If the `Metadata` struct is added to a context that is then shared between multiple exporters and those exporters attempt a `Get` at the same time the collector will panic with `fatal error: concurrent map writes`.

We are seeing this happen right now with the headerssetterextension.  The extension is telling the exporters to grab details out of the context, but it is the same context used between all the exporters in the pipeline.

Code that I suspect is not thread safe: https://github.com/open-telemetry/opentelemetry-collector/blob/59b45a575d9094b977f3904e8d1408a47ef1d70f/client/client.go#L150-L171

**Steps to reproduce**

1. Setup a collector to receive OTLP metrics
2. Configure a headerssetter extension to extract a value from the context.
4. Use the headerssetter extension as auth to multiple exporters (the more exporters, the faster you get crashes)
5. Send data through the pipeline for several minutes. 
6. Watch for the collector to crash 

**What did you expect to see?**
No crashes, multiple exporters could extract from the same context

**What did you see instead?**
Panic with `fatal error: concurrent map writes`

**What version did you use?**
Latest

TylerHelmuth:
Is it expected that this feature NOT be thread-safe and that users (like the extension) should implement protection around `Metadata.Get` to make it thread safe?

jpkrohling:
I believe it should be made thread safe. The initial assumption I had back then was that receivers would write and other components would read, meaning that only one part would write at any given time. I think the optimization that causes the write was added later and a lock should have been added there, given the initial assumption changed.

TylerHelmuth:
As @bogdandrutu pointed out, the issue is happening specifically because of `m.data[key] = v`.  If the request headers match the configured value exactly then this issue does not happen.

PR #9595, opened 2024-02-16
[client] Fix race condition in `Metadata.Get`

TylerHelmuth:
**Description:** 
Fixes a bug where `Get` was not thread safe.

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9594

Tested locally


dmitryax:
@TylerHelmuth PTAL at the failing tests

client/client.go
```diff
@@ -143,11 +145,18 @@ func FromContext(ctx context.Context) Info {
...
 }
 
 // Get gets the value of the key from metadata, returning a copy.
 func (m Metadata) Get(key string) []string {
+	if m.data == nil {
+		return make([]string, 0)
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
bogdandrutu, line 159: There are few other issues here:
1. We don't copy the initial map, so even if we protect this that does not mean we are thread-safe.
2. I would rather remove the "optimization" and remove the lock.

TylerHelmuth: @bogdandrutu I believe if we copy the map will will still need the lock.  The new copy of the original map will live in `Metadata.data` which lives in `Info` which lives in the `context.Context`. The same `context.Context` (and therefore `Metadata.data`) is being shared between multiple exporters.  When multiple exporters use the headersetters extension to access values in the context we get the error. Since multiple goroutines are trying to access the same map I believe we need the lock.

I started with a `sync.Map` but it made some unit tests fail in contrib.  If you like that approach more I can switch back to it and deal with contrib.

bogdandrutu: > Since multiple goroutines are trying to access the same map I believe we need the lock.

Not true, this is only because we also write (see the "optimization" in the code). Without that we will not get any error.

TylerHelmuth: Oh I see what you mean. I'll make those changes.

TylerHelmuth: @bogdandrutu let me know what you think about the new implementation

TylerHelmuth:
@jpkrohling please also take a look.

client/client.go
```diff
@@ -141,21 +141,27 @@ func FromContext(ctx context.Context) Info {
 
 // NewMetadata creates a new Metadata object to use in Info. md is used as-is.
 func NewMetadata(md map[string][]string) Metadata {
+	c := make(map[string][]string, len(md))
+	for k, v := range md {
+		c[k] = v
```
bogdandrutu, line 146: Let's use `strings.ToLower(k)` here and in get do a `m.data[strings.ToLower(k)]` then the optimization we removed is no longer needed.

bogdandrutu (approved):

client/client.go
TylerHelmuth: Done

config/configgrpc/configgrpc_test.go
```diff
@@ -796,7 +797,9 @@ func TestContextWithClient(t *testing.T) {
 	for _, tC := range testCases {
 		t.Run(tC.desc, func(t *testing.T) {
 			cl := client.FromContext(contextWithClient(tC.input, tC.doMetadata))
-			assert.Equal(t, tC.expected, cl)
+			assert.Equal(t, tC.expected.Addr, cl.Addr)
+			assert.Equal(t, tC.expected.Auth, cl.Auth)
+			assert.True(t, reflect.DeepEqual(tC.expected.Metadata, cl.Metadata))
```
bogdandrutu, line 802: Does assert.Equal not do enough?

TylerHelmuth: Oh that was a holdover from the sync.Map attemp, good eye

Issue #9505, opened 2024-02-07
[otlpexporter] Validate endpoint has port


atoulme:
```
2024-02-07T02:18:55.694Z	error	exporterhelper/queue_sender.go:123	Exporting failed. No more retries left. Dropping data.	{"kind": "exporter", "data_type": "traces", "name": "otlp/XXX", "error": "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: address example.com: missing port in address\"", "dropped_items": 335}
go.opentelemetry.io/collector/exporter/exporterhelper.(*queueSender).consume
	go.opentelemetry.io/collector/exporter@v0.91.0/exporterhelper/queue_sender.go:123
go.opentelemetry.io/collector/exporter/exporterhelper/internal.(*boundedMemoryQueue[...]).Consume
	go.opentelemetry.io/collector/exporter@v0.91.0/exporterhelper/internal/bounded_memory_queue.go:55
go.opentelemetry.io/collector/exporter/exporterhelper/internal.(*QueueConsumers[...]).Start.func1
	go.opentelemetry.io/collector/exporter@v0.91.0/exporterhelper/internal/consumers.go:43
```

If you omit the port from the endpoint used by the otlp exporter, it starts ok but fails to run with an error. Instead, we should have a default port or validate the endpoint on start.

atmask:
@atoulme I'm just getting familiar with the project and looking for an entry point to get started with. Could I take a look into this if there's not a time constraint (might have to do a bit of environment set up)? Is there a preference between setting the default port or doing the validation?

atmask:
Okay I've found how the otlp exporter makes use of the `configgrpc.Config` in the exporter config. I think that the solution I will add is to also validate that a port is specified as part of the otlp exporter's config validation.

Kimbohlovette:
Hello @Aneurysm9  can I take this issue
I would like to work on this issue

atmask:
@Kimbohlovette There is already a PR to resolve this issue. Please see: https://github.com/open-telemetry/opentelemetry-collector/pull/9632.

Kimbohlovette:
Alright @atmask thanks very much
I will look for another issue

atmask:
@atoulme This has been stalled in PR review for a month. Can we please get a review from a code owner?

atoulme:
Closing as complete.

PR #9632, opened 2024-02-23
[otlp exporter] Validate exporter endpoint has port

atmask:
**Description:** This PR updates the otlp exporter config validation to ensure that the "endpoint" specified for the exporter includes a port. The goal of this is to fail fast if the configuration is invalid instead of waiting for an error to arise. The PR adds a function to the ClientConfig defined in configgrpc that parses the port defined in the endpoint. The otlp exporter uses this port parsing to validate that 

**Link to tracking Issue:** [Issue 9505](https://github.com/open-telemetry/opentelemetry-collector/issues/9505)

**Testing:** 

configgrpc:
- Add unit tests to validate ports are correctly parsed in configrpc and that an error is returned if the port is omitted

otlpexporter:
- Add unit test for conifg validation function to ensure that the validation fails for missing endpoint or missing port and that a well-formed endpoint does not fail

### DIFF
--- a/exporter/otlpexporter/config.go
+++ b/exporter/otlpexporter/config.go
@@ -5,6 +5,9 @@ package otlpexporter // import "go.opentelemetry.io/collector/exporter/otlpexpor
 
 import (
 	"errors"
+	"fmt"
+	"net"
+	"strconv"
 	"strings"
 
 	"go.opentelemetry.io/collector/component"
@@ -23,9 +26,20 @@ type Config struct {
 }
 
 func (c *Config) Validate() error {
-	if c.sanitizedEndpoint() == "" {
+	endpoint := c.sanitizedEndpoint()
+	if endpoint == "" {
 		return errors.New(`requires a non-empty "endpoint"`)
 	}
+
+	// Validate that the port is in the address
+	_, port, err := net.SplitHostPort(endpoint)
+	if err != nil {
+		return err
+	}
+	if _, err := strconv.Atoi(port); err != nil {
+		return fmt.Errorf(`invalid port "%s"`, port)
+	}
+
 	return nil
 }
 
--- a/exporter/otlpexporter/config_test.go
+++ b/exporter/otlpexporter/config_test.go
@@ -111,6 +111,14 @@ func TestUnmarshalInvalidConfig(t *testing.T) {
 			name:     "invalid_tls",
 			errorMsg: `invalid TLS min_version: unsupported TLS version: "asd"`,
 		},
+		{
+			name:     "missing_port",
+			errorMsg: `missing port in address`,
+		},
+		{
+			name:     "invalid_port",
+			errorMsg: `invalid port "port"`,
+		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			cfg := factory.CreateDefaultConfig()
--- a/exporter/otlpexporter/testdata/invalid_configs.yaml
+++ b/exporter/otlpexporter/testdata/invalid_configs.yaml
@@ -71,3 +71,33 @@ invalid_tls:
     multiplier: 1.3
     max_interval: 60s
     max_elapsed_time: 10m
+missing_port:
+  endpoint: example.com
+  timeout: 10s
+  sending_queue:
+    enabled: true
+    num_consumers: 2
+    queue_size: 10
+  retry_on_failure:
+    enabled: true
+    initial_interval: 10s
+    randomization_factor: 0.7
+    multiplier: 1.3
+    max_interval: 60s
+    max_elapsed_time: 10m
+invalid_port:
+  endpoint: example.com:port
+  timeout: 10s
+  sending_queue:
+    enabled: true
+    num_consumers: 2
+    queue_size: 10
+  retry_on_failure:
+    enabled: true
+    initial_interval: 10s
+    randomization_factor: 0.7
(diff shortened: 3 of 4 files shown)


exporter/otlpexporter/config.go
```diff
@@ -22,10 +22,31 @@ type Config struct {
 }
 
 func (c *Config) Validate() error {
+	if err := c.validateEndpointNotEmpty(); err != nil {
+		return err
+	}
+	if err := c.validateEndpointHasPort(); err != nil {
+		return err
+	}
```
TylerHelmuth, line 30: The logic inside these functions is simple enough that we can keep everything inside `Validate`.

config/configgrpc/configgrpc.go
```diff
@@ -172,6 +172,16 @@ func (gcs *ClientConfig) SanitizedEndpoint() string {
 	}
 }
 
+func (gcs *ClientConfig) EndpointPort() (string, error) {
```
TylerHelmuth, line 175: We're working towards GA for `configgrpc` right now and I am hesitant to add new public APIs. For now, let's move this function into the otlpexporter.

exporter/otlpexporter/config.go
atmask: Addressed in ec2b9d32

config/configgrpc/configgrpc.go
atmask: Addressed in ec2b9d3

exporter/otlpexporter/config.go
```diff
@@ -25,7 +27,26 @@ func (c *Config) Validate() error {
 	if c.SanitizedEndpoint() == "" {
 		return errors.New(`requires a non-empty "endpoint"`)
 	}
+
+	// Validate endpoint has a valid port
+	port, err := c.endpointPort()
```
bogdandrutu, line 32: We have a default port, why not using that if not provided?

exporter/otlpexporter/config.go
```diff
@@ -25,7 +27,23 @@ func (c *Config) Validate() error {
...
+	_, err := c.endpointPort()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
+func (c *Config) endpointPort() (string, error) {
+	sanitizedEndpoint := c.SanitizedEndpoint()
+	_, port, err := net.SplitHostPort(sanitizedEndpoint)
```
TylerHelmuth, line 41:
```suggestion
	_, port, err := net.SplitHostPort(c.SanitizedEndpoint())
```

exporter/otlpexporter/config.go
```diff
@@ -25,7 +27,23 @@ func (c *Config) Validate() error {
...
+
+	// Validate endpoint has a valid port
+	_, err := c.endpointPort()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
+func (c *Config) endpointPort() (string, error) {
```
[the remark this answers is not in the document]
TylerHelmuth: @atmask ya that is normally the process - go update contrib and we'll get that merged to unblock this PR.

exporter/otlpexporter/config_test.go
```diff
@@ -114,3 +114,26 @@ func TestUnmarshalInvalidConfig(t *testing.T) {
 	}
 
 }
+
+func TestConfigValidation(t *testing.T) {
```
TylerHelmuth, line 118: Instead of a new function we normally test the config via `TestUnmarshalConfig`/`TestUnmarshalInvalidConfig`.  You can add new test scenarios to `testdata/` and then add scenarios to the appropriate test function.

atmask: I see, that make sense. Moved the new test scenario into the testdata

atmask:
Related PR from otel-collector-contrib to update loadbalancingexporter for the change in otlp config validation: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31425/files

exporter/otlpexporter/config.go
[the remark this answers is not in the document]
atmask: PR for otel-collector-contrib linked below

exporter/otlpexporter/testdata/invalid_configs.yaml
```diff
@@ -53,3 +53,19 @@ invalid_retry:
     multiplier: 1.3
     max_interval: 60s
     max_elapsed_time: 10m
+missing_port:
+  endpoint: example.com
```
iblancasa, line 57: Would it make sense to test an endpoint using a wrong port? Like `example.com:stuff` instead of `example.com:8080`

atmask: Addressed in latest push. Resolving

iblancasa (approved):

TylerHelmuth (approved):

atmask:
Just updated to fix the lint issue for new commits going into main and rebased on main. @TylerHelmuth could you please review this or poke a codeowner to push this PR through so that it can get closed off?

dmitryax (approved):

dmitryax:
I think using the default port is better, but we can add the validation for now since it's not working without specifying the port explicitly anyway. 

I think we don't have clear guidance on how components should mutate config structs passed to them when it's required, like in this case. It'd be great to have an issue for that so we can track it separately.

PR #9637, opened 2024-02-26
Give NoOp create settings a unique name

dashpole:

Long story, but i'm working on updating the prometheus dependency: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30934

As part of that update, we need to adapt to a change that makes the prometheus servers' self-observability metrics independent. See https://github.com/prometheus/prometheus/issues/13507 and https://github.com/prometheus/prometheus/pull/13610

One way to adapt to this change is by adding a label to each receivers' metrics to differentiate one Prometheus receiver from another.  I've tried taking that approach in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30934, but the current issue is that the NoOp components all have the same name, which causes the self-observability metrics to collide.

I can work around this in the prometheus receiver's own tests, but I can't work around the issue in the `generated_component_test.go` tests, since those are generated.

This PR makes the ID returned by `NewNopCreateSettings` unique by giving it a unique name.

**Link to tracking Issue:**

Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30883

cc @Aneurysm9


dashpole:
This breaks a few tests in contrib, but it should be easy to fix those when we bump the dependency.

Aneurysm9 (commented):
This generally looks good to me, but it should have a changelog entry as it may confound test expectations.

dashpole:
Added a changelog.

Aneurysm9 (approved):

codeboten (approved):

dashpole:
I'll need to update contrib first to fix the failing tests.

dashpole:
seems like the changelog check is broken?

```
/usr/local/lib/node_modules/markdown-link-check/markdown-link-check:97
                    if (program.opts().ignore.some((ignorePath) => resolved.includes(ignorePath))) {
```

codeboten:
Looks like it resolved :)

PR #9495, opened 2024-02-06
[configtls] Rename config structs for consistency

arjunmahishi:
**Description:** <Describe what has changed.>

Simply renames a few structs in the `configtls` package for consistence.

`TLSClientSetting` to `ClientConfig`
`TLSServerSetting` to `ServerConfig`
`TLSSetting` to `Config`

**Link to tracking Issue:** <Issue number if applicable>

Fixes #9474

PR #9658, opened 2024-02-28
[configtls] Add NewDefault* funcs

TylerHelmuth:
**Description:** 
Adds new `NewDefault*` funcs for all 3 config structs. 

In anticipation of the name changes from https://github.com/open-telemetry/opentelemetry-collector/pull/9495 I've named the functions using the new, preferred name. 

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9657

Added tests

**Documentation:** 
godoc


config/configtls/configtls.go
```diff
@@ -75,6 +75,11 @@ type Config struct {
 	ReloadInterval time.Duration `mapstructure:"reload_interval"`
 }
 
+// NewDefaultConfig creates a new TLSSetting with any default values set.
+func NewDefaultConfig() Config {
```
bogdandrutu, line 79: Is there a need for this from users perspective? Or we think it is enough only the Client/Server? Happy to keep it if we have few needs.

bogdandrutu (approved):

config/configtls/configtls.go
TylerHelmuth: I dont have a specific user cases, only following the pattern defined in https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#default-configuration. The pattern implies to me that we should have a `NewDefault` for all public config APIs.  It is probably overkill, but it does allow us to add new default values to `Config` in the future if we want.

astencel-sumo: I'd rather have the `NewDefaultConfig` function.

One thing is that we can use it in the `NewDefaultClientConfig` and `NewDefaultServerConfig`, although admittedly it doesn't need to be exported for this.

The other thing is that users can create their own configs that embed the `configtls.Config` struct and then they can use the `configtls.NewDefaultConfig` to create their instances of their configs.

I hope you don't ask me for potential examples of other configs that embed `configtls.Config`, because I don't have any such examples. :smile:

astencel-sumo (approved):

codeboten (approved):

Issue #9475, opened 2024-02-06
[configtls] Validate `MinVersion` and `MaxVersion`


mx-psi:
We effectively restrict the allowed versions to the set defined here:

https://github.com/open-telemetry/opentelemetry-collector/blob/f5a7315cf88e10c0bce0166b35d9227727deaa61/config/configtls/configtls.go#L382

We should validate for it! This can be done after 1.0 per the rules we have here https://github.com/open-telemetry/opentelemetry-collector/blob/f5a7315cf88e10c0bce0166b35d9227727deaa61/VERSIONING.md#L52-L54

Still, it would be nice to have before 1.0 IMO

molejnik88:
Hi,
Isn't `MinVersion` validated here: [configtls.go#L176](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/configtls.go#L176) and `MaxVersion` validated here: [configtls.go#L180](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/configtls.go#L180)?

molejnik88:
I'd like to work on this issue. Can you please assign it to me?

atoulme:
Done, go for it

PR #9664, opened 2024-02-28
[config/configtls] Validate MinVersion and MaxVersion

molejnik88:
Add `Validate()` method to `TLSSetting` and validate tls `min_version` and `max_version`.

#9475

### DIFF
--- a/config/configtls/configtls.go
+++ b/config/configtls/configtls.go
@@ -173,6 +173,24 @@ func (r *certReloader) GetCertificate() (*tls.Certificate, error) {
 	return r.cert, nil
 }
 
+func (c TLSSetting) Validate() error {
+	minTLS, err := convertVersion(c.MinVersion, defaultMinTLSVersion)
+	if err != nil {
+		return fmt.Errorf("invalid TLS min_version: %w", err)
+	}
+
+	maxTLS, err := convertVersion(c.MaxVersion, defaultMaxTLSVersion)
+	if err != nil {
+		return fmt.Errorf("invalid TLS max_version: %w", err)
+	}
+
+	if maxTLS < minTLS && maxTLS != defaultMaxTLSVersion {
+		return errors.New("invalid TLS configuration: min_version cannot be greater than max_version")
+	}
+
+	return nil
+}
+
 // loadTLSConfig loads TLS certificates and returns a tls.Config.
 // This will set the RootCAs and Certificates of a tls.Config.
 func (c Config) loadTLSConfig() (*tls.Config, error) {
--- a/.chloggen/issue-9475-configtls-validate.yaml
+++ b/.chloggen/issue-9475-configtls-validate.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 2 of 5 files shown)


config/configtls/configtls.go
```diff
@@ -153,6 +153,24 @@ func (r *certReloader) GetCertificate() (*tls.Certificate, error) {
...
+	if err != nil {
+		return fmt.Errorf("invalid TLS min_version: %w", err)
+	}
+
+	maxTLS, err := convertVersion(c.MaxVersion, defaultMaxTLSVersion)
+	if err != nil {
+		return fmt.Errorf("invalid TLS max_version: %w", err)
+	}
+
+	if maxTLS < minTLS && maxTLS != defaultMaxTLSVersion {
```
dmitryax, line 167: What if the user sets `max_version` lower than `min_version` but the same as the default value? Looks like it should be checked before falling back to the defaults

molejnik88: Currently `defaultMaxTLSVersion` is set to 0 and user is not able to provide such value. `convertVersion` would return an error, because `"0"` is not present in `tlsVersions` [map](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/configtls.go#L410).
There are cases when `maxTLS` returned by `convertVersion` is lower than `minTLS`, e.g. user provides value for `min_version`, but omits `max_version`. In such case `maxTLS == 0` and `minVersion` is a non-zero value (greater than zero). When `0` is provided to tls.Config, the max supported version is used by `tls` package.

I guess there is a risk when `defaultMaxVersion` is set to a different value than `0`. Maybe it would be a better idea to explicitly compare `maxTLS` with `0`?
```
if maxTLS < minTLS && maxTLS != 0 {`
```

config/configtls/configtls.go
```diff
@@ -153,6 +153,24 @@ func (r *certReloader) GetCertificate() (*tls.Certificate, error) {
 	return r.cert, nil
 }
 
+func (c TLSSetting) Validate() error {
+	minTLS, err := convertVersion(c.MinVersion, defaultMinTLSVersion)
+	if err != nil {
+		return fmt.Errorf("invalid TLS min_version: %w", err)
+	}
+
+	maxTLS, err := convertVersion(c.MaxVersion, defaultMaxTLSVersion)
```
molejnik88, line 162: `convertVersion` is also used in [TLSSetting.loadTLSConfig()](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/configtls.go#L196). Both client and server implementation call it in `LoadTLSConfig()`. Isn't it a problem that the logic that validates TLS version is used in load methods and `Validate()`?

config/configtls/configtls.go
dmitryax: I believe `if maxTLS < minTLS && maxTLS != 0 {`  is cleaner. I would even remove `defaultMaxTLSVersion` completely and use zero value instead, but that's out of the scope probably

config/configtls/configtls.go
dmitryax: What `Validate()` do you mean?

dmitryax: Yes, I think this is fine

dmitryax (approved):

Issue #9666, opened 2024-02-29
[otlphttpexporter] partial success handlers attempt to unmarshal a nil value when response content-length is 0


senojj:
**Describe the bug**
In the otlphttpexporter exporter's otlp.go file, when the export HTTP response body is read using the function `readResponseBody` a value of `nil, nil` will be returned when `resp.ContentLength` is 0.

Later, that `nil` value is passed to one of the partial success handlers (e.g. `tracesPartialSuccessHandler`). No check is made for a `nil` value in the `protoBytes` argument, so an attempt to unmarshal the `nil` value is possible.

Edit: This is only the case when the content type is specified as "application/json". This does not apply when the content type is specified as "application/x-protobuf".

**Steps to reproduce**
Provide an empty in response to an export. Ensure that an expected Content-Type response header field is given, such as "application/json" or "application/x-protobuf".

**What did you expect to see?**
An empty response body should not result in an error. It is perfectly valid for the logic to continue without processing a response body.

**What did you see instead?**
An error is raised and the successful request is potentially retried after a backoff duration.

**What version did you use?**
fb912b9

**What config did you use?**
Any configuration using otlphttpexporter as an exporter.

**Environment**
Any environment

PR #9667, opened 2024-02-29
[otlphttpexporter] return nil from partial success handler when HTTP response body is empty

senojj:
Fixing a bug - When exporting using the otlphttpexporter, after receiving a successful HTTP response, when the response body's content length is 0 and the content type is specified as either "application/json" or "application/x-protobuf", an attempt will be made to unmarshal a nil value within any of the partial success response handler functions. This results in an error, and a potential resend of the original export request.

To fix this scenario, a check was added to the `tracesPartialSuccessHandler`, `metricsPartialSuccessHandler`, and `logsPartialSuccessHandler` functions for a `nil` value in the `protoBytes` argument. When `nil`, the function will return with a `nil` value, indicating the absence of any error.

**Link to tracking Issue:** #9666


TylerHelmuth (commented):
_Technically_ the OTLP protocol specifies the response MUST be the signal-specific object, but I am totally ok with this defensive code. 

Can you add some new tests?

senojj:
> _Technically_ the OTLP protocol specifies the response MUST be the signal-specific object, but I am totally ok with this defensive code.
> 
> Can you add some new tests?

I didn't realize that was specified in the OTLP protocol. Thank you for pointing that out. 

I will be happy to add some test cases to catch this.

exporter/otlphttpexporter/otlp_test.go
```diff
@@ -467,22 +467,58 @@ func TestPartialResponse_missingHeaderButHasBody(t *testing.T) {
...
+	contentTypes := []struct {
+		contentType string
+	}{
+		{contentType: "application/x-protobuf"},
+		{contentType: "application/json"},
+	}
+	for _, tt := range contentTypes {
+		for _, telemetryType := range []string{"logs", "metrics", "traces"} {
+			t.Run("Missing header but has body for content type "+tt.contentType+" "+telemetryType, func(t *testing.T) {
+				response := ptraceotlp.NewExportResponse()
```
TylerHelmuth, line 479: For success they probably all look the same, but technically each signal has their own expected response object.  Can you update the test to use the proper response per signal?

senojj: good catch, will update.

senojj: I _think_ I now have the tests in a good place. I apologize for so much back and forth as I am getting used to the project details.

iblancasa (approved):

exporter/otlphttpexporter/otlp_test.go
```diff
@@ -524,53 +634,145 @@ func TestPartialSuccess_shortContentLengthHeader(t *testing.T) {
...
+				var data []byte
+				var err error
+
+				switch ct.contentType {
+				case jsonContentType:
+					data, err = serializer.MarshalJSON()
+				case protobufContentType:
+					data, err = serializer.MarshalProto()
+				default:
+					panic(ct.contentType)
```
TylerHelmuth, line 680: Lets to `require.Fail` instead of a panic.

TylerHelmuth (approved):

dmitryax:
@senojj thanks for the fix. Can you please add a changelog entry?

senojj:
> @senojj thanks for the fix. Can you please add a changelog entry?

This is done. I apologize for the delay in adding that.

dmitryax (approved):

Issue #9473, opened 2024-02-06
`scraperhelper.NewScraper` is awkward to use


mx-psi:
While working on #9208 I realized that the current way `scraperhelper.NewScraper` works is a bit awkward. The function takes a `name string` and then builds an ID with it:

https://github.com/open-telemetry/opentelemetry-collector/blob/f5a7315cf88e10c0bce0166b35d9227727deaa61/receiver/scraperhelper/scraper.go#L64

It is very common among contrib scrapers to name the scraper with the same name as the component itself, and so we end up building a `component.Type`, converting it to string to pass to `NewScraper` and then building it again. I propose we change the first argument to take a `component.Type` instead.

bogdandrutu:
I think Scraper interface needs a bit more "love" than that. Will give a try to send some PRs/ideas soon

mx-psi:
>  the empty value is still in use

Where do we use the empty value?

mrasu:
I'd like to try this at https://github.com/open-telemetry/opentelemetry-collector/pull/11082

codeboten:
Thanks @mrasu, I'll assign it to you

codeboten:
Re-opening as it wont be completed until the deprecated methods have been removed

PR #9670, opened 2024-03-02
Adds a WithID method to allow setting IDs with names (not just types)

hughesjj:
**Description:**
Optional PR inspired by `-contrib`'s [31457](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31457) I came across [9473](https://github.com/open-telemetry/opentelemetry-collector/issues/9473)

I recently had to update a bunch of code related to the `Type` regex lockdown (good thing imo!  I'm glad we're doing it!) and wanted some functionality like this

1. The validation logic is only "accessible" outside of the `component` package if you call `NewType` (one one of the `Must` functions which calls such), but a lot of legacy code directly instantiates `Type`.  While that can always be refactored, I thought having a `Validate` command would be less invasive and allows us to centralize our validation logic.  Might want to add it to `component.ID` as well.
2. It's a tad annoying that we take a raw string value for the constructor of `scraper` when the underlying member variable is an ID. IMHO an ideal world would have us either change the type of that member to `component.Type`, or permit the constructor to take in any valid `component.ID`.   

Also, as of now we don't length-limit our types, but we may want to start doing that...

### DIFF
--- a/receiver/scraperhelper/scraper_test.go
+++ b/receiver/scraperhelper/scraper_test.go
@@ -1,3 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
 package scraperhelper
 
 import (


component/config.go
```diff
@@ -123,19 +123,22 @@ func (t Type) String() string {
 // This must be kept in sync with the regex in cmd/mdatagen/validate.go.
 var typeRegexp = regexp.MustCompile(`^[a-zA-Z][0-9a-zA-Z_]*$`)
 
+const NoType = Type("")
```
bogdandrutu, line 126: Why do you need this public?

component/config.go
```diff
@@ -151,6 +154,13 @@ func MustNewType(strType string) Type {
 	return ty
 }
 
+func (t Type) Validate() error {
```
bogdandrutu, line 157: @mx-psi do I remember wrongly or you were working to change this to be a struct? So this is not needed since we cannot construct an invalid type.

bogdandrutu (commented):
We need a deeper refactoring/cleaning of the Scraper, not sure if it is ok to add this now.

component/config.go
mx-psi: Yes, see #9472 which is ready to merge. I think that PR would make this one unnecessary since it would be impossible to construct an invalid type

mx-psi: > Well, technically we still [don't allow empty string names](https://github.com/open-telemetry/opentelemetry-collector/blob/94ae7762c9458cfe199cfaa3d16a4ecc1f7330be/component/config.go#L126) for type right? Given the regex does not permit empty strings

I don't think we should allow empty names, at least not through an empty `component.Type`, but I am not sure what you want to accomplish exactly, so it's better if we discuss in an issue

> (also we should probably add a generous length limit imho)

Would you be open to filing an issue about this?

PR #9672, opened 2024-03-02
[confighttp] newDefault for confighttp package structs

### LINKED ISSUE #9655: [confighttp] Add `NewDefault*` functions for all config structs

Sanket-0510:
**Description:**  
NewDefault methods to all the exported structs in confighttp package

**Link to tracking Issue:** 
closes #9655 

**Testing:** 
Added test for the same.

**Documentation:**
 godoc

### DIFF
--- a/.chloggen/confighttp-add-newdefault-func.yaml
+++ b/.chloggen/confighttp-add-newdefault-func.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: breaking
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: confighttp
--- a/.chloggen/confighttp-add-newdefault-func.yaml
+++ b/.chloggen/confighttp-add-newdefault-func.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confighttp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adds `NewDefault*` functions for all the config structs.
+
+# One or more tracking issues or pull requests related to the change
+issues: [9655]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api]
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -105,18 +105,17 @@ type ClientConfig struct {
 // the default values of 'MaxIdleConns' and 'IdleConnTimeout'.
 // Other config options are not added as they are initialized with 'zero value' by GoLang as default.
 // We encourage to use this function to create an object of ClientConfig.
-func (c *ClientConfig) NewDefaultClientConfig() ClientConfig {
+func NewDefaultClientConfig() ClientConfig {
 	// The default values are taken from the values of 'DefaultTransport' of 'http' package.
 
-	//Check Whether the headers are set to nil, if yes then set it to default map.
-	if c.Headers == nil {
-		c.Headers = make(map[string]configopaque.String)
-	}
+	// Configure with the non-nil map value.
+	headers := make(map[string]configopaque.String)
 
 	maxIdleConns := 100
 	idleConnTimeout := 90 * time.Second
 
 	return ClientConfig{
+		Headers:         headers,
 		MaxIdleConns:    &maxIdleConns,
 		IdleConnTimeout: &idleConnTimeout,
 	}
@@ -293,14 +292,13 @@ type ServerConfig struct {
 }
 
 // NewDefaultServerConfig creates new ServerConfig with default values set
-func (c *ServerConfig) NewDefaultServerConfig() ServerConfig {
+func NewDefaultServerConfig() ServerConfig {
+	// Create a new map for ResponseHeaders
+	responseHeaders := make(map[string]configopaque.String)
 
-	//Check Whether the ResponseHeaders are set to nil, if yes then set it to default map.
-	if c.ResponseHeaders == nil {
-		c.ResponseHeaders = make(map[string]configopaque.String)
+	return ServerConfig{
+		ResponseHeaders: responseHeaders,
 	}
-
-	return ServerConfig{}
 }
 
 // Deprecated: [v0.99.0] Use ToListener instead.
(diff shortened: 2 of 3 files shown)


Sanket-0510:
why are contrib-tests failing? : |

.chloggen/confighttp-add-newdefault-func.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
astencel-sumo, line 4: Why is this a breaking change?

Sanket-0510: wait it should be enhancement

config/confighttp/confighttp.go
```diff
@@ -276,6 +276,11 @@ type ServerConfig struct {
 	ResponseHeaders map[string]configopaque.String `mapstructure:"response_headers"`
 }
 
+// NewDefaultServerConfig creates new ServerConfig with default values set
+func NewDefaultServerConfig() ServerConfig {
+	return ServerConfig{}
```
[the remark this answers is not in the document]
Sanket-0510: I have added the check for ClientConfig also, I hope this fixes the panic

config/confighttp/confighttp.go
```diff
@@ -105,10 +105,15 @@ type ClientConfig struct {
...
 	// The default values are taken from the values of 'DefaultTransport' of 'http' package.
+
+	// Configure with the non-nil map value.
+	headers := make(map[string]configopaque.String)
+
 	maxIdleConns := 100
 	idleConnTimeout := 90 * time.Second
 
 	return ClientConfig{
+		Headers:         headers,
```
astencel-sumo, line 116: Why instantiate a variable `headers`?
```suggestion
		Headers:         make(map[string]configopaque.String),
```

config/confighttp/confighttp.go
```diff
@@ -276,6 +281,16 @@ type ServerConfig struct {
 	ResponseHeaders map[string]configopaque.String `mapstructure:"response_headers"`
 }
 
+// NewDefaultServerConfig creates new ServerConfig with default values set
+func NewDefaultServerConfig() ServerConfig {
+	// Create a new map for ResponseHeaders
+	responseHeaders := make(map[string]configopaque.String)
+
+	return ServerConfig{
+		ResponseHeaders: responseHeaders,
```
astencel-sumo, line 290:
```suggestion
		ResponseHeaders: make(map[string]configopaque.String),
```

config/confighttp/confighttp.go
Sanket-0510: Yes it will be just waste of space and memory. P.S - corrected it :sweat_smile:

astencel-sumo (approved):

Sanket-0510:
@astencel-sumo some contrib tests are failing in github actions, but I did ran all tests on local but nothing failed there, I also looked into the failing test it is expecting nil value for our map.

TylerHelmuth (commented):
`ClientConfig` and `ServerConfig` depend on `configtls`.  I think we should wait on this PR until the `NewDefault` functions exist for `configtls`.

astencel-sumo:
> `ClientConfig` and `ServerConfig` depend on `configtls`. I think we should wait on this PR until the `NewDefault` functions exist for `configtls`.

I don't think that's the problem. The reason the contrib tests are failing is that I asked Sanket to make the `NewDefaultClientConfig` and `NewDefaultServerConfig` to initialize the `Headers` and `ResponseHeaders` maps to empty maps. Previously those where `nil`. The tests in contrib expect those properties to be `nil`.

See e.g. here: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/8231961010/job/22508354464?pr=9672#step:4:1109

```console
=== FAIL: . TestLoadConfig/zipkin/2 (0.00s)
    config_test.go:86: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/zipkinexporter/config_test.go:86
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -39,3 +39,4 @@
        	            	   Timeout: (time.Duration) 5000000000,
        	            	-  Headers: (map[string]configopaque.String) <nil>,
        	            	+  Headers: (map[string]configopaque.String) {
        	            	+  },
```
I'm not exactly sure what's the preferred course of action here.

- Allow contrib tests to fail and fix them after updating core in contrib?
- Fix those tests before merging this change so that those tests to not check that `Response` and `ResponseHeaders` are `nil`?
- ...?

What's your opinion @TylerHelmuth ?

TylerHelmuth:
Oh my comment was only a general statement, not a reason for the failing tests.  I dont think we should introduce `NewDefault*` functions here until they can depend on the `NewDefault*` functions from `configtls`.

For the tests themselves I'd say we have 2 options:
1. Not use empty map as the default value and instead use nil
2. Allow contrib tests to fail and fix them after updating core in contrib

astencel-sumo:
> Oh my comment was only a general statement, not a reason for the failing tests. I dont think we should introduce `NewDefault*` functions here until they can depend on the `NewDefault*` functions from `configtls`.

Oh, I see :smile: Makes sense, @Sanket-0510 this means this PR should wait for https://github.com/open-telemetry/opentelemetry-collector/pull/9658. After https://github.com/open-telemetry/opentelemetry-collector/pull/9658 is merged, you can incorporate changes from `main` (by merging or rebasing) and then we can continue with this.

> For the tests themselves I'd say we have 2 options:
> 
>     1. Not use empty map as the default value and instead use nil
> 
>     2. Allow contrib tests to fail and fix them after updating core in contrib

I don't like 1, I believe the functions should not leave the map as `nil` and expose users to potential `panic`. Let's go with 2 if possible.

TylerHelmuth:
@Sanket-0510 this is ready to use the `configtls. NewDefault*` funcs

config/confighttp/confighttp.go
```diff
@@ -107,10 +107,12 @@ type ClientConfig struct {
 // We encourage to use this function to create an object of ClientConfig.
 func NewDefaultClientConfig() ClientConfig {
 	// The default values are taken from the values of 'DefaultTransport' of 'http' package.
+
```
TylerHelmuth, line 110:
```suggestion
```

config/confighttp/confighttp.go
```diff
@@ -286,6 +288,14 @@ type ServerConfig struct {
 	ResponseHeaders map[string]configopaque.String `mapstructure:"response_headers"`
 }
 
+// NewDefaultServerConfig creates new ServerConfig with default values set
+func NewDefaultServerConfig() ServerConfig {
+
```
TylerHelmuth, line 293:
```suggestion
```

config/confighttp/confighttp.go
```diff
@@ -286,6 +288,14 @@ type ServerConfig struct {
...
+	return ServerConfig{
+		ResponseHeaders: make(map[string]configopaque.String),
```
TylerHelmuth, line 295: This should use configtls.NewDefaultServerConfig and configauth.NewDefaultAuthentication and NewDefaultCORSConfig

config/confighttp/confighttp.go
```diff
@@ -107,10 +107,12 @@ type ClientConfig struct {
 // We encourage to use this function to create an object of ClientConfig.
 func NewDefaultClientConfig() ClientConfig {
 	// The default values are taken from the values of 'DefaultTransport' of 'http' package.
+
 	maxIdleConns := 100
 	idleConnTimeout := 90 * time.Second
 
 	return ClientConfig{
+		Headers:         make(map[string]configopaque.String),
```
TylerHelmuth, line 115: We should also use configtls.NewDefaultClientConfig and configauth.NewDefaultAuthentication

PR #9717, opened 2024-03-06
[exporterhelper] Make panics into errors

mx-psi:
**Description:** Turns panics into errors


exporter/exporterhelper/common.go
```diff
@@ -42,55 +43,74 @@ func (b *baseRequestSender) setNextSender(nextSender requestSender) {
...
+	return nil
+}
+
+var _ Option = (optionFunc)(nil)
+
+type optionFunc func(*baseExporter) error
+
+func (o optionFunc) apply(be *baseExporter) error {
+	return o(be)
+}
```
dmitryax, line 65: Why don't just use `type Option func(*baseExporter) error`?

mx-psi: I was being consistent with https://github.com/open-telemetry/opentelemetry-collector/blob/94ae7762c9458cfe199cfaa3d16a4ecc1f7330be/featuregate/registry.go#L45-L48

I have a slight preference for this instead of the `func` because it is more flexible but I am fine changing it if you prefer the `func`.

dmitryax: It just seems like too much extra code. I would prefer a function or at least removing `optionFuncNoErr`.

mx-psi: Oki, changed back to `func` with an `error`, we can always change back if needed

dmitryax (approved):

PR #9723, opened 2024-03-07
[chore] [exporterhelper] Update docs to remove requeuing

carsonip:

- **[chore] [exporterhelper] Update docs to remove requeuing** as it was removed in #9090
- **Fix typo**

There is no functional change and should not require a changelog.


dmitryax (approved):
Thanks @carsonip!

PR #9738, opened 2024-03-11
[exporterhelper] Add default batching for OTLP data type

dmitryax:
Introduce default batching functionality based on the internal data type (pdata). This makes the exporter batching capability available to the regular exporter helpers without using custom requests.

Updates #8122

### DIFF
--- a/exporter/exporterhelper/logs.go
+++ b/exporter/exporterhelper/logs.go
@@ -82,7 +82,10 @@ func NewLogsExporter(
 	if pusher == nil {
 		return nil, errNilPushLogsData
 	}
-	logsOpts := []Option{withMarshaler(logsRequestMarshaler), withUnmarshaler(newLogsRequestUnmarshalerFunc(pusher))}
+	logsOpts := []Option{
+		withMarshaler(logsRequestMarshaler), withUnmarshaler(newLogsRequestUnmarshalerFunc(pusher)),
+		withBatchFuncs(mergeLogs, mergeSplitLogs),
+	}
 	return NewLogsRequestExporter(ctx, set, requestFromLogs(pusher), append(logsOpts, options...)...)
 }
 
--- a/exporter/exporterhelper/logs_batch.go
+++ b/exporter/exporterhelper/logs_batch.go
@@ -1,0 +1,137 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterhelper // import "go.opentelemetry.io/collector/exporter/exporterhelper"
+
+import (
+	"context"
+	"errors"
+
+	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+	"go.opentelemetry.io/collector/pdata/plog"
+)
+
+// mergeLogs merges two logs requests into one.
+func mergeLogs(_ context.Context, r1 Request, r2 Request) (Request, error) {
+	lr1, ok1 := r1.(*logsRequest)
+	lr2, ok2 := r2.(*logsRequest)
+	if !ok1 || !ok2 {
+		return nil, errors.New("invalid input type")
+	}
+	lr2.ld.ResourceLogs().MoveAndAppendTo(lr1.ld.ResourceLogs())
+	return lr1, nil
+}
+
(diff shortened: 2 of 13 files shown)


exporter/exporterhelper/logs_batch.go
```diff
@@ -0,0 +1,136 @@
...
+			if extractedLogs.LogRecordCount() == 0 {
+				break
+			}
+			capacityLeft -= extractedLogs.LogRecordCount()
+			if destReq == nil {
+				destReq = &logsRequest{ld: extractedLogs, pusher: srcReq.pusher}
+			} else {
+				extractedLogs.ResourceLogs().MoveAndAppendTo(destReq.ld.ResourceLogs())
+			}
+			if capacityLeft == 0 {
```
splunkericl, line 61: Do we want to do `capacityLeft <= 0` check? If for whatever reason a regression is introduced in logs extractions prior, at least the data request isn't lost.

exporter/exporterhelper/logs.go
```diff
@@ -82,7 +82,10 @@ func NewLogsExporter(
 	if pusher == nil {
 		return nil, errNilPushLogsData
 	}
-	logsOpts := []Option{withMarshaler(logsRequestMarshaler), withUnmarshaler(newLogsRequestUnmarshalerFunc(pusher))}
+	logsOpts := []Option{
+		withMarshaler(logsRequestMarshaler), withUnmarshaler(newLogsRequestUnmarshalerFunc(pusher)),
+		withBatchFuncs(mergeLogs, mergeSplitLogs),
```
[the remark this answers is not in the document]
dmitryax: It doesn't enable it for everyone. `WithBatcher` is still required on the exporter to provide this option at least. Then users have to enable batching explicitly

exporter/exporterhelper/logs_batch.go
dmitryax: Good point

dmitryax: updated

splunkericl (approved):
Thanks for adding the merge and mergeSplit function. This makes using exporter batching framework much easier.

exporter/exporterhelper/metrics_batch.go
```diff
@@ -0,0 +1,236 @@
...
+		if !ok {
+			return nil, errors.New("invalid input type")
+		}
+		if srcReq.md.DataPointCount() <= capacityLeft {
+			if destReq == nil {
+				destReq = srcReq
+			} else {
+				srcReq.md.ResourceMetrics().MoveAndAppendTo(destReq.md.ResourceMetrics())
+			}
+			capacityLeft = cfg.MaxSizeItems - destReq.md.DataPointCount()
```
mx-psi, line 46: Here and in the other mergesplit funcs I think it would be clearer if capacityLeft  is defined inside the for loop and then here we just do
```suggestion
			capacityLeft -= destReq.md.DataPointCount()
```

exporter/exporterhelper/metrics_batch.go
```diff
@@ -0,0 +1,236 @@
...
+				break
+			}
+			capacityLeft -= extractedMetrics.DataPointCount()
+			if destReq == nil {
+				destReq = &metricsRequest{md: extractedMetrics, pusher: srcReq.pusher}
+			} else {
+				extractedMetrics.ResourceMetrics().MoveAndAppendTo(destReq.md.ResourceMetrics())
+			}
+
+			if capacityLeft <= 0 {
```
mx-psi, line 62: I don't understand when `capacityLeft` would be negative. Could you add a comment?

mx-psi (approved):
LGTM, just two minor nits to clarify the code

exporter/exporterhelper/metrics_batch.go
dmitryax: We cannot set it inside the loop because `capacityLeft` has to be preserved between the iterations. But I can actually apply your suggestion anyway. This is the correct behavior. The current code could be considered as a bug if we had more than 2 iterations

exporter/exporterhelper/metrics_batch.go
dmitryax: It cannot be. I can change it to `==` to make it clear. We can rely on tests instead of adding extra safeguards in the production code

PR #9740, opened 2024-03-11
[exporterhelper] Fix persistent queue size backup on reads

carsonip:

Persistent queue size backup on reads should depend on readIndex, not
writeIndex.


TylerHelmuth (commented):
@carsonip nice find.  Can you add a test for this situation?

dmitryax (approved):

dmitryax:
Thank you! 

@TylerHelmuth a test to catch this might be too complicated. Given that this code path covers a very narrow edge case and is only beneficial for use cases when the collector is killed, I think it's ok to merge as is

PR #9750, opened 2024-03-13
[config] Deprecate component.UnmarshalConfig

### LINKED ISSUE #7101: Investigate alternative for `Unmarshaler` and `Marshaler`
The `Unmarshaler` because of the `Unmarshal` definition, has an issue when recursively unmarshal other structs that implement the `Unmarshaler`, see https://github.com/open-telemetry/opentelemetry-collector/pull/6029 where the "given" destination to Conf.Unmarshal cannot be checked against the Unmarshaler implementation to avoid a deadlock. The Marshaler seem to not have (I did not carefully check) the same problem, but this needs to be confirmed.

Some possible solutions:
* See yaml.v2 interfaces, they seem to work fine. We may want to switch to that.
* Ignore this issue, document properly in the Conf.Unmarshal the caveat.

### LINKED ISSUE #7102: Remove component.UnmarshalConfig

atoulme:
**Description:**
This PR removes the top level if/else in `component.UnmarshalConfig`, handling recursive state in the confmap.Conf object instead.
This PR deprecates `component.UnmarshalConfig` in favor of calling directly `Unmarshal` on the confmap.Conf object.

**Link to tracking Issue:**
Fixes #7102
Fixes #7101

### DIFF
--- a/.chloggen/embedded_struct_unmarshaler.yaml
+++ b/.chloggen/embedded_struct_unmarshaler.yaml
@@ -7,7 +7,7 @@ change_type: deprecation
 component: component
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Deprecate `component.UnmarshalConfig`, use `conf.Unmarshal(&intoCfg)` instead.
+note: Deprecate `component.UnmarshalConfig`, use `(*confmap.Conf).Unmarshal(&intoCfg)` instead.
 
 # One or more tracking issues or pull requests related to the change
 issues: [7102]
--- a/.chloggen/embedded_struct_unmarshaler.yaml
+++ b/.chloggen/embedded_struct_unmarshaler.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: breaking
+change_type: deprecation
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: component
--- a/.chloggen/embedded_struct_unmarshaler.yaml
+++ b/.chloggen/embedded_struct_unmarshaler.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate `component.UnmarshalConfig`, use `conf.Unmarshal(&intoCfg)` instead.
+
+# One or more tracking issues or pull requests related to the change
+issues: [7102]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api]
--- a/component/config.go
+++ b/component/config.go
@@ -27,7 +27,7 @@ type Config any
 var configValidatorType = reflect.TypeOf((*ConfigValidator)(nil)).Elem()
 
 // UnmarshalConfig helper function to UnmarshalConfig a Config.
-// Deprecated: [v0.97.0] Use conf.Unmarshal(&intoCfg)
+// Deprecated: [v0.101.0] Use conf.Unmarshal(&intoCfg)
 func UnmarshalConfig(conf *confmap.Conf, intoCfg Config) error {
 	return conf.Unmarshal(intoCfg)
 }
--- a/component/config.go
+++ b/component/config.go
@@ -27,13 +27,8 @@ type Config any
 var configValidatorType = reflect.TypeOf((*ConfigValidator)(nil)).Elem()
 
 // UnmarshalConfig helper function to UnmarshalConfig a Config.
-// It checks if the config implements confmap.Unmarshaler and uses that if available,
-// otherwise uses Map.UnmarshalExact, erroring if a field is nonexistent.
+// Deprecated: Use conf.Unmarshal(&intoCfg)
 func UnmarshalConfig(conf *confmap.Conf, intoCfg Config) error {
-	if cu, ok := intoCfg.(confmap.Unmarshaler); ok {
-		return cu.Unmarshal(conf)
-	}
-
 	return conf.Unmarshal(intoCfg)
 }
 
--- a/component/config.go
+++ b/component/config.go
@@ -27,7 +27,7 @@ type Config any
 var configValidatorType = reflect.TypeOf((*ConfigValidator)(nil)).Elem()
 
 // UnmarshalConfig helper function to UnmarshalConfig a Config.
-// Deprecated: Use conf.Unmarshal(&intoCfg)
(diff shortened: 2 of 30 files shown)


atoulme:
Contrib tests fix: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31727

cmd/mdatagen/loader.go
```diff
@@ -124,11 +124,7 @@ func (m *metric) Unmarshal(parser *confmap.Conf) error {
 	if !parser.IsSet("enabled") {
 		return errors.New("missing required field: `enabled`")
 	}
-	err := parser.Unmarshal(m)
```
mx-psi, line 127: I think this one I already merged on a different PR, right? Can you rebase your branch?

cmd/mdatagen/metricdata.go
mx-psi, line 1: Can we do all changes in this file in a different PR? AFAICT they can be done independently

component/config_test.go
```diff
@@ -419,28 +417,3 @@ func TestNewType(t *testing.T) {
...
-	embeddedUnmarshallingConfig
-}
-
-type embeddedUnmarshallingConfig struct {
-}
-
-func (euc *embeddedUnmarshallingConfig) Unmarshal(_ *confmap.Conf) error {
-	return nil // do nothing.
-}
-func TestStructWithEmbeddedUnmarshaling(t *testing.T) {
```
mx-psi, line 435: Why are we removing this test? Is it being move somewhere else?

confmap/confmap.go
```diff
@@ -143,31 +147,40 @@ func (l *Conf) ToStringMap() map[string]any {
...
+		mapstructure.StringToSliceHookFunc(","),
+		mapKeyStringToMapKeyTextUnmarshalerHookFunc(),
+		mapstructure.StringToTimeDurationHookFunc(),
+		mapstructure.TextUnmarshallerHookFunc(),
+		unmarshalerHookFunc(result, topLevelUnmarshaling),
+		// after the main unmarshaler hook is called,
+		// we unmarshal the embedded structs if present to merge with the result:
+		unmarshalerEmbeddedStructsHookFunc(),
+		zeroSliceHookFunc(),
+	}
```
mx-psi, line 162: This is also something we can do on a separate PR!

confmap/confmap.go
```diff
@@ -37,7 +40,8 @@ func NewFromStringMap(data map[string]any) *Conf {
 // Conf represents the raw configuration map for the OpenTelemetry Collector.
 // The confmap.Conf can be unmarshalled into the Collector's config using the "service" package.
 type Conf struct {
-	k *koanf.Koanf
+	k    *koanf.Koanf
+	self any
```
mx-psi, line 44: Before I think we could call the same `Conf` to unmarshal it on two different structs, but now it can only be used once, right? Is there any way we can restore that property? If not, is there a way where we can error out if we have already unmarshaled?

component/config_test.go
atoulme: It's moved under confmap.

confmap/confmap.go
atoulme: right, I'll revert that change. It's not helping the diff.

cmd/mdatagen/loader.go
atoulme: Done, thanks

confmap/confmap.go
atoulme: This doesn't break your ability to use the same `Conf` to unmarshal on two different structs. It stops recursive calls as part of unmarshaling from calling `Unmarshal` on the same object twice.

confmap/confmap.go
```diff
@@ -167,7 +171,13 @@ func decodeConfig(m *Conf, result any, errorUnused bool) error {
 	if err != nil {
 		return err
 	}
-	return decoder.Decode(m.ToStringMap())
+	if err = decoder.Decode(m.ToStringMap()); err != nil {
```
atoulme, line 174: Splitting this change into #9763 as well.

component/config_test.go
atoulme: brought it back with a few changes so it passes!

mx-psi:
Can you add 'Fixes #7101' to the PR description?

confmap/confmap.go
```diff
@@ -310,33 +315,47 @@ func unmarshalerEmbeddedStructsHookFunc() mapstructure.DecodeHookFuncValue {
...
-		if _, ok = from.Interface().(map[string]any); !ok {
-			return from.Interface(), nil
+		unmarshalMethod := reflect.ValueOf(toPtr).MethodByName("Unmarshal")
+		// check that the Unmarshal method is not defined on one of the squash embedded structs defined.
+		// This allows us to have composition where the Unmarshaler may embed structs implementing Unmarshaler.
+		if to.Type().Kind() == reflect.Struct {
+			for i := 0; i < to.Type().NumField(); i++ {
+				// check embedded structs:
+				f := to.Type().Field(i)
+				if f.IsExported() && f.Anonymous && slices.Contains(strings.Split(f.Tag.Get("mapstructure"), ","), "squash") {
```
dmitryax, line 341: Why do we need to check for the `mapstructure` tag?

atoulme: This is explained in the comment above: if the embedded struct is using the `,squash` value for the `mapstructure` key, then this struct Unmarshal method should not be used as the main struct unmarshaling.

.chloggen/embedded_struct_unmarshaler.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
andrzej-stencel, line 4: Isn't this a `deprecation`?

component/config.go
```diff
@@ -27,13 +27,8 @@ type Config any
 var configValidatorType = reflect.TypeOf((*ConfigValidator)(nil)).Elem()
 
 // UnmarshalConfig helper function to UnmarshalConfig a Config.
-// It checks if the config implements confmap.Unmarshaler and uses that if available,
-// otherwise uses Map.UnmarshalExact, erroring if a field is nonexistent.
+// Deprecated: [v0.101.0] Use conf.Unmarshal(&intoCfg)
 func UnmarshalConfig(conf *confmap.Conf, intoCfg Config) error {
```
andrzej-stencel, line 31: Do we want to change the way this function works with the deprecation? Is there a reason we cannot leave the function unchanged until it's completely removed?

.chloggen/embedded_struct_unmarshaler.yaml
atoulme:
```suggestion
change_type: deprecation
```

component/config.go
atoulme: It does, it removes the old way to call into unmarshaling by calling the Unmarshaler Unmarshal function.

mx-psi (approved):

component/config.go
mx-psi: AIUI this is a no-op change

.chloggen/embedded_struct_unmarshaler.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate `component.UnmarshalConfig`, use `conf.Unmarshal(&intoCfg)` instead.
```
evan-bradley, line 10:
```suggestion
note: Deprecate `component.UnmarshalConfig`, use `(*confmap.Conf).Unmarshal(&intoCfg)` instead.
```
Up to you if this is more readable, I just want to make it clear we're talking about a method on a `confmap.Conf` object and not a `conf` package, since we use the word `config` in so many different contexts in our APIs.

evan-bradley (approved):

Issue #7621, opened 2023-05-03
Removal date suppose to be the date when feature gate will be removed, but that is irrelevant since once it is marked as Stable cannot be changed anyway. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/featuregate/registry.go#L150


bogdandrutu:

MovieStoreGuy:
Just quickly looking at what you're referring to, do you mean https://github.com/open-telemetry/opentelemetry-collector/blob/main/featuregate/registry.go#L110 by chance?

TylerHelmuth:
I think there are 2 ways to tackle this:

1. Stick with the [k8s process](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-stages) and have `ToVersion` represent the version that the gate is completely removed. With this method users do not have an idea when the gate will be marked `stable`, but do know when the gate was removed.
2. Start requiring `ToVersion` when promoting a feature gate to `beta` and have it represent when a feature will be marked `stable`.  Allow it to be set with `alpha` optionally to represent when a get will be marked `deprecated`.  With this method users will know when a gate will be marked `stable` or `deprecated`, but it locks in beta to a set timeframe when maybe we want more flexibility on how long something stays in beta.

TylerHelmuth:
Based on the discussion in the SIG meeting today we will go with option 1 with the following additions:
- A `stable` gate should be able to be set to true without failing.
- A `deprecated` gate should be able to be set to false without failing.
- When a `stable` or `deprecated` gate is set a warning log should be written explaining that the gate will be removed in `ToVersion` and that continued use of the gate will cause errors after that version.

PR #9752, opened 2024-03-13
[chore] Remove `telemetry.useOtelForInternalMetrics` stable feature gate

atoulme:
**Description:**
Remove the stable feature gate `telemetry.useOtelForInternalMetrics`. It was declared stable in 0.95.0.

### DIFF
--- a/.chloggen/remove_stable_featuregate.yaml
+++ b/.chloggen/remove_stable_featuregate.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: telemetry
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove telemetry.useOtelForInternalMetrics stable feature gate
+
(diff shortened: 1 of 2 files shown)


mx-psi (approved):
cc @codeboten

mx-psi (approved):
cc @codeboten

mx-psi:
Not sure how I approved twice, lol

codeboten (commented):
This should have a changelog with breaking change right? as users will no longer be able to specify the feature gate

atoulme:
I wasn't sure. I looked at previous removal of stable feature gate and it had no changelog entry.

codeboten:
Right, i *think* it should always be in the changelog, otherwise the collector won't start with:

```
Error: invalid argument
```

TylerHelmuth:
Example: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/28659

codeboten:
@TylerHelmuth right, but the collector would continue to start until this change goes out no? isn't that what makes it a breaking change?

TylerHelmuth:
>  but the collector would continue to start until this change goes out

Well I didnt think it did but [I was wrong](https://github.com/open-telemetry/opentelemetry-collector/issues/7621).  So ya, we need a changelog.

bogdandrutu (approved):

atoulme:
I'll add a changelog, one moment.

TylerHelmuth (approved):

codeboten (approved):

PR #9762, opened 2024-03-14
Nicer error message when passing an empty configuration file

Kimbohlovette:
This PR checks if `cfg.Validate()` error is `errMissingReceivers` error then returns a nicely formated error.


otelcol/collector.go
```diff
@@ -179,6 +180,9 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	}
 
 	if err = cfg.Validate(); err != nil {
+		if errors.Is(err, errMissingReceivers) {
+			return fmt.Errorf("no receiver configuration specified in collector configuration file")
+		}
```
TylerHelmuth, line 185: The `errMissingReceivers` is ok.  What we want to change it detecting, during `Validate` that the configuration is completely empty, not just that it is missing receiver configuration.

Kimbohlovette: Alright. Let me check inside the cfg.Validate() and  make the changes in there

otelcol/collector.go
```diff
@@ -179,6 +180,9 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 	}
 
 	if err = cfg.Validate(); err != nil {
+		if errors.Is(err, errInvalidOrMissingConfigFile) {
+			return errInvalidOrMissingConfigFile
+		}
```
TylerHelmuth, line 185:
```suggestion
```

We can return the error directly.

otelcol/config.go
```diff
@@ -6,14 +6,16 @@ package otelcol // import "go.opentelemetry.io/collector/otelcol"
...
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/service"
 )
 
 var (
-	errMissingExporters = errors.New("no exporter configuration specified in config")
-	errMissingReceivers = errors.New("no receiver configuration specified in config")
+	errMissingExporters           = errors.New("no exporter configuration specified in config")
+	errMissingReceivers           = errors.New("no receiver configuration specified in config")
+	errInvalidOrMissingConfigFile = errors.New("missing or invalid configuration file")
```
TylerHelmuth, line 18:
```suggestion
	errEmptyOrMissingConfigFile = errors.New("missing or empty configuration file")
```

otelcol/config.go
```diff
@@ -44,6 +46,11 @@ type Config struct {
 func (cfg *Config) Validate() error {
 	// Currently, there is no default receiver enabled.
 	// The configuration must specify at least one receiver to be valid.
+
+	if cfg == nil || reflect.ValueOf(cfg).Len() < 1 {
+		return errInvalidOrMissingConfigFile
```
TylerHelmuth, line 51:
```suggestion
		return errEmptyOrMissingConfigFile
```

otelcol/config.go
```diff
@@ -44,6 +46,11 @@ type Config struct {
 func (cfg *Config) Validate() error {
 	// Currently, there is no default receiver enabled.
 	// The configuration must specify at least one receiver to be valid.
+
+	if cfg == nil || reflect.ValueOf(cfg).Len() < 1 {
```
TylerHelmuth, line 50: Lets move this above the comments - those should be associate with the existing receiver check.

TylerHelmuth, line 50: `cfg` is not a Array, Chan, Map, Slice, String, or pointer to Array, so this call will panic.  I believe we'll need to check the len of each of `cfg`'s sections and if they are all 0 then return this error.

otelcol/collector.go
Kimbohlovette: Okay

otelcol/config.go
Kimbohlovette: Alright. I'm applying the changes right away

Kimbohlovette: Alright no problem then. The reason why I did not want to check the length of individual properties of the cfg is that that config may grow. That's another property may be added to cfg in the future

Kimbohlovette:
Hello @TylerHelmuth I've made the required changes. Please can you look at it?

TylerHelmuth (commented):
@Kimbohlovette looks good, but I prefer the word `empty` over `invalid`.  There are lots of reasons a config could be invalid, but this check is specifically catching empty configs.

Kimbohlovette:
> @Kimbohlovette looks good, but I prefer the word `empty` over `invalid`. There are lots of reasons a config could be invalid, but this check is specifically catching empty configs.

Okay I'm changing it to invalid

Kimbohlovette:
Hello @TylerHelmuth I'm made that single change and also updated the branch from main

Kimbohlovette:
@TylerHelmuth @jpkrohling. I have some tests failing but  tests ran successfully locally. Please you throw some context as to why these tests may be failing?

TylerHelmuth (approved):

Kimbohlovette:
@dmitryax  please could you review my PR?

dmitryax (approved):

Issue #7890, opened 2023-06-14
Revisit how collector internal metric are distributed across telemetry levels


dmitryax:
There are several verbosity levels that can be used to configure how many metrics the collector exposes:
```
	// Level is the level of telemetry metrics, the possible values are:
	//  - "none" indicates that no telemetry data should be collected;
	//  - "basic" is the recommended and covers the basics of the service telemetry.
	//  - "normal" adds some other indicators on top of basic.
	//  - "detailed" adds dimensions and views to the previous levels.
```

The problem is that they are barely used. Most of the metrics are exposed at the `basic` level. There is only one metric in batch processor exposed at the `detailed` level. The `normal` level is not being used. The suggestion is to revisit all the metrics and further distribute them across the levels.

The default level is `basic` (the lowest), which is the most common and provides most of the metrics. We can move a significant portion of the metrics to the `normal` level, which can become a new default. So default behavior doesn't change for the end user. While `basic` level can be kept to the bare minimum reserved for collector core: 
- process metrics
- accepted and refused data by receivers
- sent and failed to send data by exporters
- sending queue size and enqueue failures

OTel components metrics will use only normal or detailed levels.

Long-term, we can consider providing a user an option to override this level per component.

victoramsantos:
Agree. I was testing the difference between these levels and just found the same, that the detailed level logs has one more metric than others (namely `otelcol_processor_batch_batch_send_size_bytes_bucket`).

dmitryax:
According to the proposed guidelines, I think we can move `Basic` to `Normal` level by default and move the following metric sets from `Basic` to `Normal`:

1. **GRPC/HTTP server/client metrics.** 
```
otelcol_http_client_duration histogram
otelcol_http_client_request_size counter
otelcol_http_client_response_size counter
otelcol_http_server_duration histogram
otelcol_http_server_request_size counter
otelcol_http_server_response_size counter
otelcol_rpc_server_duration histogram
otelcol_rpc_server_request_size histogram
otelcol_rpc_server_requests_per_rpc histogram
otelcol_rpc_server_response_size histogram
otelcol_rpc_server_responses_per_rpc histogram
```
These metrics were not emitted before the transition to OTel instrumentation and can be pretty noisy even with enabled `telemetry.disableHighCardinalityMetrics`. It may even be worth considering moving them (or a portion of them) to the `Detailed` level.

2. Batch processor metrics:
```
otelcol_processor_batch_batch_send_size histogram
[...]

Issue #9510, opened 2024-02-07
[component] TelemetrySettings.MetricsLevel is experimental


mx-psi:
The `MetricsLevel` field is marked as experimental https://pkg.go.dev/go.opentelemetry.io/collector/component#TelemetrySettings. We should decide what to do with it (stabilize it, remove it, change it in some way)

codeboten:
I expected TelemetrySettings to eventually include a `LoggerProvider` once the logging bridge API is implemented in otel-go. This makes me wonder if TelemetrySettings should be moved out of component

mx-psi:
> I expected TelemetrySettings to eventually include a LoggerProvider once the logging bridge API is implemented in otel-go.

I had assumed we would keep on using `*zap.Logger` with a `zapcore.Core` that uses the `LoggerProvider` behind the scenes.

> This makes me wonder if TelemetrySettings should be moved out of component

But I am in favor of this regardless, I am just not sure where it should go (`service/telemetry` implies a dependency of the components on `service`)

bogdandrutu:
> I expected TelemetrySettings to eventually include a LoggerProvider once the logging bridge API is implemented in otel-go. This makes me wonder if TelemetrySettings should be moved out of component

Since we pass the `TelemetrySettings` in the start/creation of the components even if you move it, you depend on it so you have the same problem

bogdandrutu:
>  We should decide what to do with it (stabilize it, remove it, change it in some way)

I want to remove it, since it was used to configure "high cardinality" metrics in the past and we should have views and other configs for that.

mx-psi:
@codeboten Does the `useOtelWithSDKConfigurationForInternalTelemetry` feature gate still allow setting `configtelemetry.Level`? If so, I think we should remove its usage (at least once we add support for views). 

It would also be interesting to tie ignoring this level in components that use it today with the `useOtelWithSDKConfigurationForInternalTelemetry` flag

mx-psi:
@codeboten Could you take a look at https://github.com/open-telemetry/opentelemetry-collector/issues/9510#issuecomment-1934046671 ?

codeboten:
After discussing this at the 3-Apr-2024 SIG call, it was suggested that the use of a level is easier for users so we should keep it and remove the message about it being experimental.

More advanced use-cases will be able to enable all metrics and use views for maximum flexibility/control over what metrics are emitted.

mx-psi:
[...]

PR #9767, opened 2024-03-14
Distribute internal metrics across different levels

dmitryax:
**Description:**

This change distributes the reported internal metrics across available levels and updates the level set by default:

1. The default level is changed from `basic` to `normal`, which can be overridden with `service::telmetry::metrics::level` configuration.

2. The following batch processor metrics are updated to be reported starting from `normal` level instead of `basic` level:
  - `processor_batch_batch_send_size`
  - `processor_batch_metadata_cardinality` 
  - `processor_batch_timeout_trigger_send` 
  - `processor_batch_size_trigger_send` 

3. The following GRPC/HTTP server and client metrics are updated to be reported starting from `detailed` level: 
  - `http.client.*` metrics 
  - `http.server.*` metrics 
  - `rpc.server.*` metrics 
  - `rpc.client.*` metrics

**Link to tracking Issue:** https://github.com/open-telemetry/opentelemetry-collector/issues/7890

### DIFF
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -30,6 +30,7 @@ import (
 	"go.opentelemetry.io/collector/config/configcompression"
 	"go.opentelemetry.io/collector/config/confignet"
 	"go.opentelemetry.io/collector/config/configopaque"
+	"go.opentelemetry.io/collector/config/configtelemetry"
 	"go.opentelemetry.io/collector/config/configtls"
 	"go.opentelemetry.io/collector/config/internal"
 	"go.opentelemetry.io/collector/extension/auth"
@@ -253,8 +254,10 @@ func (gcs *ClientConfig) toDialOptions(host component.Host, settings component.T
 
 	otelOpts := []otelgrpc.Option{
 		otelgrpc.WithTracerProvider(settings.TracerProvider),
-		otelgrpc.WithMeterProvider(settings.MeterProvider),
 		otelgrpc.WithPropagators(otel.GetTextMapPropagator()),
+	}
+	if settings.MetricsLevel >= configtelemetry.LevelDetailed {
+		otelOpts = append(otelOpts, otelgrpc.WithMeterProvider(settings.MeterProvider))
 	}
 
 	// Enable OpenTelemetry observability plugin.
@@ -356,8 +359,10 @@ func (gss *ServerConfig) toServerOption(host component.Host, settings component.
 
 	otelOpts := []otelgrpc.Option{
 		otelgrpc.WithTracerProvider(settings.TracerProvider),
-		otelgrpc.WithMeterProvider(settings.MeterProvider),
 		otelgrpc.WithPropagators(otel.GetTextMapPropagator()),
+	}
+	if settings.MetricsLevel >= configtelemetry.LevelDetailed {
+		otelOpts = append(otelOpts, otelgrpc.WithMeterProvider(settings.MeterProvider))
 	}
 
 	// Enable OpenTelemetry observability plugin.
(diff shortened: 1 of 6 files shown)


mx-psi (commented):
How does this interact with https://github.com/open-telemetry/opentelemetry-collector/issues/9510#issuecomment-1932569156 ? 

I think we should make a decision on whether to remove Metrics Level and if so what's our plan until we remove it

dmitryax:
I'm ok with having other configs for that like views. AFAIU this isn't possible even with `telemetry.useOtelWithSDKConfigurationForInternalTelemetry`. @codeboten, please correct me if I am wrong. If there are are other plans to introduce this functionality, I'd like to help with that

config/confighttp/confighttp.go
```diff
@@ -366,18 +370,20 @@ func (hss *ServerConfig) ToServer(host component.Host, settings component.Teleme
...
 		}),
-	)
+	}
+	if settings.MetricsLevel >= configtelemetry.LevelDetailed {
+		otelOpts = append(otelOpts, otelhttp.WithMeterProvider(settings.MeterProvider))
+	}
+
+	// Enable OpenTelemetry observability plugin.
+	// TODO: Consider to use component ID string as prefix for all the operations.
+	handler = otelhttp.NewHandler(handler, "http", otelOpts...)
```
jaronoff97, line 386: Rather than setting "http" here should we instead use the component ID per the TODO? Or is the TODO no longer relevant?

processor/batchprocessor/metrics.go
```diff
@@ -56,8 +57,18 @@ func newBatchProcessorTelemetry(set processor.CreateSettings, currentMetadataCar
...
 
 func (bpt *batchProcessorTelemetry) createOtelMetrics(set component.TelemetrySettings, currentMetadataCardinality func() int) error {
-	var errors, err error
-	meter := metadata.Meter(set)
+	var (
+		errors, err error
+		meter       metric.Meter
+	)
+
+	// BatchProcessor metrics are not subject to the same level of filtering as other components.
```
codeboten, line 65: Isn't this true for all components? If the default is normal, aren't all metrics (other than instrumentation library metrics) emitted?

codeboten (commented):
This looks ok, just one question

processor/batchprocessor/metrics.go
dmitryax: > Isn't this true for all components?

Yes, it's true for all components. With this comment, I wanted to say that the batch processor is the only component in core emitting metrics (with the normal level). It's probably confusing. Let me update it

> If the default is normal, aren't all metrics (other than instrumentation library metrics) emitted?

Right

config/confighttp/confighttp.go
dmitryax: Good catch. I don't recall why I added "http" here. Probably by mistake. I've reverted it to "". The TODO comment should be addressed separately

config/configgrpc/configgrpc.go
```diff
@@ -253,9 +254,11 @@ func (gcs *ClientConfig) toDialOptions(host component.Host, settings component.T
 
 	otelOpts := []otelgrpc.Option{
 		otelgrpc.WithTracerProvider(settings.TracerProvider),
-		otelgrpc.WithMeterProvider(settings.MeterProvider),
 		otelgrpc.WithPropagators(otel.GetTextMapPropagator()),
 	}
+	if settings.MetricsLevel >= configtelemetry.LevelDetailed {
+		otelOpts = append(otelOpts, otelgrpc.WithMeterProvider(settings.MeterProvider))
```
TylerHelmuth, line 260: Will only adding the `WithMeterProvider` in this condition break users who are emitting metrics with OTel on lower metric levels?

dmitryax: Sorry I'm not sure I understand the question

TylerHelmuth: I misunderstood that `otelgrpc.WithMeterProvider` is whats configuring the grpc instrumentation, you've done this right.

TylerHelmuth (approved):

PR #9771, opened 2024-03-15
Revert "[chore] change the way we unmarshal the config in tests"

atoulme:
Reverts open-telemetry/opentelemetry-collector#9765

We need to revert those changes as contrib has issues with them in isolation from #9750.


TylerHelmuth (approved):

mx-psi (approved):

codeboten (approved):

codeboten:
Should this revert not address these errors:

```
Error: ../../receiver/opencensusreceiver/factory.go:54:37: cannot use rCfg.NetAddr.Transport (variable of type confignet.TransportType) as string value in argument to newOpenCensusReceiver
```

PR #9775, opened 2024-03-15
[telemetry] emit metrics with _ instead of /

### LINKED ISSUE #9774: internal collector metrics should use `_` instead of `/` in names

codeboten:
This is addressing an issue w/ the names of the metrics generated by the Collector for its internal metrics. Note that this change only impacts users that emit telemetry using OTLP, which is currently still in experimental support. The prometheus metrics already replaced `/` with `_`.

Fixes #9774


internal/obsreportconfig/obsmetrics/obsmetrics.go
```diff
@@ -7,6 +7,7 @@
 package obsmetrics // import "go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 
 const (
-	NameSep = "/"
-	Scope   = "go.opentelemetry.io/collector/obsreport"
+	NameSep   = "/"
+	MetricSep = "."
```
dmitryax, line 11: On the SIG call, I suggested using `_` instead of `.` until we stabilize the metrics. The reason is that AFAIK Prometheus is working on supporting dots. So once it's ready, it'll be another breaking change for all the users using Prometheus to scrape the internal metrics. Suddenly all the metrics will be converted from `otelcol_receiver_accepted_metric_points` to `otelcol.receiver.accepted.metric_points` If we adopt `_` for now, I believe we can avoid that. 

I understand that `_` is not semantically correct from OTel's perspective, but we will have to make a big migration later anyway, so it might make sense to reduce the amount of breaking changes.

I'm happy to hear other opinions. And I'm not sure about the timeline for dots support in Prometheus. Maybe it's much further away than new stable Collector metrics.

cc @open-telemetry/collector-approvers  @open-telemetry/collector-contrib-approvers

dmitryax: Otherwise, once the dots are supported in Prometheus, we might need a config option to convert `.`->`_` somewhere to preserve the behavior and avoid breaking the users. It can be even a Prometheus receiver config option. In that case, we should be good.

@dashpole, maybe you have some ideas about it?

dashpole: Note that use of dots vs underscores will be negotiated using headers, so users will be able to control it server-side: https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#query-time.  If it is enabled by default at some point in the future, it might still be a surprise for users.

codeboten: @dmitryax the reason i went with `.` instead of `_` is that other metrics (for example the grpc instrumentation) uses `.` already when emitting OTLP. Having both separators, depending on the source of the metric, seemed confusing.

dmitryax: > Note that use of dots vs underscores will be negotiated using headers, so users will be able to control it server-side: https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#query-time. If it is enabled by default at some point in the future, it might still be a surprise for users.

Thanks for the clarification. Looks like, even if it is enabled by default, it won't happen any time soon. I assume the controlling headers will be exposed as a configuration in the Go OTel exporter as well as somewhere in service::telemetry::metrics` section of the collector config.

dmitryax: > @dmitryax the reason i went with . instead of _ is that other metrics (for example the grpc instrumentation) uses . already when emitting OTLP. Having both separators, depending on the source of the metric, seemed confusing.

Makes sense

dmitryax:
I'm worried that the emitted internal metrics are not covered by tests. It'll be great to see some metrics output changed in tests along with this change. @codeboten, do you know if we have a GH issue to introduce such coverage, by chance?

codeboten:
@dmitryax not that i know of. I can spend some time trying to add tests here, i agree that its not ideal not to have this tested with the specific metric names (hard coded in the tests)

dmitryax:
> I can also open a follow up issue to tackle it separately 🤷🏻

That would be great! Thank you.

internal/obsreportconfig/obsmetrics/obsmetrics.go
```diff
@@ -7,6 +7,7 @@
 package obsmetrics // import "go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 
 const (
-	NameSep = "/"
-	Scope   = "go.opentelemetry.io/collector/obsreport"
+	NameSep   = "/"
```
dmitryax, line 10: I assume it's still being used for spans. Maybe it's worth renaming it to SpanSep for clarity?

dmitryax (approved):

internal/obsreportconfig/obsmetrics/obsmetrics.go
dmitryax: Maybe even `MetricNameSep` and `SpanNameSep`...

codeboten: correct, i can rename it

codeboten: done

dmitryax (approved):

internal/obsreportconfig/obsmetrics/obsmetrics.go
```diff
@@ -7,6 +7,7 @@
 package obsmetrics // import "go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
 
 const (
-	NameSep = "/"
-	Scope   = "go.opentelemetry.io/collector/obsreport"
+	SpanNameSep   = "/"
+	MetricNameSep = "."
```
TylerHelmuth, line 11: This won't be a breaking change for end users who scrape the metrics with prometheus right? But it will be a breaking change for anyone using the otel go sdk to emit metrics right? I am worried about breaking end-user telemetry in a single release, should we add a feature gate?

mx-psi: Isn't OTel Go SDK support under a feature gate already?

jaronoff97: +1 to tyler's concern here. By making the separator here a `.` instead of a `_` we would then create three possibilities for users today:
```
otelcol.process_uptime # the metric for process uptime after this pr merges for users emitting otlp
otelcol_process_uptime # the metric for process uptime after this pr merges for users scraping via prometheus
process_uptime # the metric for process uptime prior to this merges for users emitting otlp
```

Rather than creating a third potential for this metric name, if the separator were `_` we would simply have this state:

```
otelcol_process_uptime # after this change, no matter if a user is sending OTLP or Prometheus
process_uptime # the metric for process uptime prior to this merges for users emitting otlp
```

Creating a third state disincentivizes users from sending OTLP (because it would break their dashboards/alerts) or upgrading their collector further for the same reason.

We already do not follow the conventions mentioned [here](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-naming), see naming for time and utilization. Prometheus currently rules the metric world for metric naming, independently of the recommendations made by our own semantic convention today. Given that, I think for legacy conversions we should always be opting for not breaking users by following the existing convention. We should reconsider this once semantic convention exists for not only metric naming recommendations (when to use a . or a _), but also for pipeline monitoring.

codeboten: Right the problem is see with using `_`, is that today, changing a collector from emitting metrics using prometheus to OTLP has a problem with metrics we don't control.

As an example, the metric for grpc instrumentation changes from 

```
otelcol_rpc_server_requests_per_rpc
```

to:

```
rpc.server.responses_per_rpc
```

This is because the prometheus metric replaces `.` with `_` (as it does for other collector metrics `/`). Replacing the separator with `_` instead of `.` means users will still have to contend with a broken metric, it's just not clear which metrics will be broken to them, as they may not know what constitutes an instrumentation metric vs a collector metric. I opted to go with `.` as it at least aligns with the opentelemetry semantic convention.

TylerHelmuth: > Creating a third state disincentivizes users from sending OTLP (because it would break their dashboards/alerts) or upgrading their collector further for the same reason.

It is true that breaking the name is going to disincentivize users, but I don't think we want to hold ourselves to prometheus naming standards when OTel has its own naming standard.

I can see an argument for waiting until more semantic conventions are stabilized before doing the breaking change to require only 1 breaking change for our end users.

I am ok keeping this change behind the `useOtelWithSDKConfigurationForInternalTelemetry` since it is still alpha.

jaronoff97: yeah my other concern is that by changing the convention for the collector's own metrics, we actually invalidate any documents for stable/beta components that mention existing metrics. [these documents](https://github.com/search?q=repo:open-telemetry/opentelemetry-collector+otelcol_+language:Markdown&type=code&l=Markdown) mention the important metrics for various components, none mentioning that these are "prometheus convention" but rather that they are the important metrics for monitoring your collector. By creating a dual state, all of these documents would need to be edited as part of this change and then edited again upon guidance from the OTEP. 

Similarly, I would argue that by not keeping the metrics consistent we are introducing a bug to the collector – we would not be able to make the `useOtelWithSDKConfigurationForInternalTelemetry` flag beta because then anyone emitting metrics via OTLP would see incorrect metric names for components like the batch processor.

TylerHelmuth: @codeboten is there a world where we handle switch to the OTEL SDK in 2 steps:
1. Get the `useOtelWithSDKConfigurationForInternalTelemetry` to `Stable` without breaking the emitted metric names:
    - this gets us off the prometheus receivers/opencensus dependency and aligned with the project the collector is a part of.
2. Redo the metric names to meet OTel stable metric conventions.

TylerHelmuth: Ya let's discuss some more at the SIG

codeboten (commented):
As discussed at the 3-Apr-2024 SIG call, decided on `_` for the metrics we control in the Collector. Updated this PR to reflect this

@jaronoff97 @TylerHelmuth PTAL

TylerHelmuth (approved):
It hurts, but at least it isn't a breaking change for the collector's own metrics.

dmitryax:
This change removes the `otelcol_` prefix from the http/grp metrics, right?

I believe we also agreed to have https://github.com/open-telemetry/opentelemetry-collector/pull/9767 resolved first to not export the rpc/http metrics by default. I think if we stop emitting them by default in the same release, it'll be a better experience for users (enable -> adopt for the name change) than (adopt for the name change -> enable once they are disabled in the future). The blast radius of the breaking change is higher in the former case.

Issue #9045, opened 2023-12-05
Stabilize module confignet


codeboten:
Before stabilizing a module, an approver or maintainer must make sure that the following criteria have been met for at least two successive minor version releases:

- [x] No open issues or PRs in the module that would require breaking changes
- [x] No TODOs in the module code that would require breaking changes
- [x] No deprecated symbols in the module
- [x] No symbols marked as experimental in the module
- [x] The module follows the [Coding guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md)

Please also make sure to publicly announce our intent to stabilize the module on:

- [x] The #otel-collector CNCF Slack Channel
- [x] The #opentelemetry CNCF Slack channel
- [x] A Collector SIG meeting (if unable to attend, just add to the agenda)

To help other people verify the above criteria, please link to the announcement and other links used to complete the above in a comment on this issue.

Once all criteria are met, close this issue by moving this module to the `stable` module set.

codeboten:
Open issues regarding confignet:
- [ ] https://github.com/open-telemetry/opentelemetry-collector/issues/4331

dmitryax:
I think we should have 1.0 milestones for each module that we want to release as 1.0 so we can include all the issues there

mx-psi:
@codeboten I think it would be helpful to have links as in #8978 so that right before marking as 1.0 one can quickly check that we did not accidentally introduce any new TODOs or experimental symbols

atoulme:
We need to address the new comments in https://github.com/open-telemetry/opentelemetry-collector/pull/7954 as well.

mx-psi:
Filed #9159 to add this to coding guidelines

atoulme:
I see the comments in #7954 have been acted upon with #9105 - can we move to announce stabilization intent, anything else left to consider?

mx-psi:
Should we remove `Context` from the method names? (That is, effectively just add `context.Context` to the original methods?)

codeboten:
See https://github.com/open-telemetry/opentelemetry-collector/pull/9258

codeboten:
Blocked by the stabilization work for confmap

atoulme:
Is this now unblocked since confmap was just declared stable?

PR #9801, opened 2024-03-20
[confignet] Mark as stable

TylerHelmuth:
**Description:**
Mark `confignet` as Stable

**Link to tracking Issue:** 
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9045


atoulme (approved):
LGTM! I have reviewed the code and the test coverage, everything looks great.

dmitryax (approved):
I've been thinking how to reduce code duplication between `TCPAddrConfig`  and `AddrConfig`, but cannot see a clean simple way. Go isn't very flexible sometimes.

versions.yaml
```diff
@@ -9,6 +9,7 @@ module-sets:
       - go.opentelemetry.io/collector/pdata
       - go.opentelemetry.io/collector/config/configopaque
       - go.opentelemetry.io/collector/config/configcompression
+      - go.opentelemetry.io/collector/config/confignet
```
astencel-sumo, line 12: It's funny how this module goes from 0.x not to 1.0 but straight to 1.3 :thinking:

.chloggen/confignet-mark-stable.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
astencel-sumo, line 25:
```suggestion
change_logs: [api]
```

versions.yaml
TylerHelmuth: `1.0`, `1.1`, and `1.2` are overrated lol

.chloggen/confignet-mark-stable.yaml
TylerHelmuth: This is technically an API change, but we've been putting these releases in the user-facing config bc its more visible and an important milestone.

astencel-sumo (approved):

mx-psi (commented):
By stabilizing this, we commit to two things from confmap:

1. Supporting `mapstructure` YAML tags. I am fine with that, I don't think there is much of a point on changing it, but I know there has been some discussion about this (@atoulme may have more details?). Is this resolved?
2. Supporting `UnmarshalText` on config unmarshaling. This is a standard library interface, so I think it is a pretty safe assumption.

We also commit to one thing from `component`:
1. `AddrConfig` implements `component.ConfigValidator`: https://github.com/open-telemetry/opentelemetry-collector/blob/15201f1e5967fd1bc1e51c118178ab109170fe07/config/confignet/confignet.go#L106. We are commiting to keep on supporting that interface.

For the implicit dependency on `component`, I don't really get why we are adding `Validate()`: the `UnmarshalText` method from the transport should already do validation at unmarshaling time. Shouldn't that be enough? If we can remove that method, we eliminate one implicit dependency.

TylerHelmuth:
>  I don't really get why we are adding Validate()

@mx-psi I believe `Validate()` should be included since `AddrConfig.Transport` is a public field.  Users are able to set `AddrConfig.Transport` manually, bypassing `TransportType.UnmarshalText`. See https://github.com/open-telemetry/opentelemetry-collector/pull/9385#discussion_r1477104190.

TylerHelmuth:
@mx-psi based on your comments it sounds like we have sufficient dependencies on `component` and `confmap` to hold off marking this stable until those 2 modules are stable.

mx-psi:
> @mx-psi based on your comments it sounds like we have sufficient dependencies on `component` and `confmap` to hold off marking this stable until those 2 modules are stable.

Or at least I think if we merge this, we should have a way to validate that we are not changing these assumptions

TylerHelmuth:
Will reopen once `component` and `confmap` are stable

atoulme:
There is no dependency to component left and confmap is stable, reopening.

mx-psi (approved):
I think it's pretty clear that `Validate` is here to stay, which was the only dependency left with component. I vote we merge this.

@open-telemetry/collector-approvers I will merge this after v0.109.0 is released unless somebody blocks the PR

songy23 (approved):

MovieStoreGuy (approved):

PR #9802, opened 2024-03-20
[chore] typo: connect -> connection

atoulme:


config/confignet/confignet.go
```diff
@@ -58,7 +58,7 @@ func (tt *TransportType) UnmarshalText(in []byte) error {
 // DialerConfig contains options for connecting to an address.
 type DialerConfig struct {
 	// Timeout is the maximum amount of time a dial will wait for
-	// a connect to complete. The default is no timeout.
+	// a connection to complete. The default is no timeout.
```
TylerHelmuth, line 61: This comment is lifted from https://pkg.go.dev/net#Dialer so I'm not sure we need to change it.

atoulme:
Alright, no need for this change, closing.

PR #9814, opened 2024-03-21
[processor/batch] prevent potential goroutine leak

### LINKED ISSUE #9739: When multiple goroutine calls consume, mb.newShard(md) will be called multiple times.
**Describe the bug**

The problematic code is as shown below:
image

**Steps to reproduce**

Here is a simple code showing this bug:
```go
package main

import (
	"fmt"
	"sync"
	"time"
)

type multiShardBatcher struct {
	batchers sync.Map

	lock sync.Mutex
	size int
}
type shard struct {
}

func (mb *multiShardBatcher) newShard() *shard {
	b := &shard{}
	fmt.Println("new shard")
	time.Sleep(1 * time.Second)
	return b
}

func (mb *multiShardBatcher) consume(key string) *shard {

	b, ok := mb.batchers.Load(key)
	if !ok {
		mb.lock.Lock()
		b, _ = mb.batchers.LoadOrStore(key, mb.newShard())
		mb.lock.Unlock()
	}
	return b.(*shard)
}

func main() {

	mb := &multiShardBatcher{}

	group := sync.WaitGroup{}
	for i := 0; i < 10; i++ {
		group.Add(1)
		go func() {
			mb.consume("key")
			group.Done()
		}()
	}
	group.Wait()

}
```

image

**What did you see instead?**

**What version did you use?**

**What config did you use?**

**Environment**

**Additional context**

astencel-sumo:
Fixes #9739

### DIFF
--- a/processor/batchprocessor/batch_processor.go
+++ b/processor/batchprocessor/batch_processor.go
@@ -307,9 +307,9 @@ func (mb *multiShardBatcher) consume(ctx context.Context, data any) error {
 	}
 	aset := attribute.NewSet(attrs...)
 
+	mb.lock.Lock()
 	b, ok := mb.batchers.Load(aset)
 	if !ok {
-		mb.lock.Lock()
 		if mb.metadataLimit != 0 && mb.size >= mb.metadataLimit {
 			mb.lock.Unlock()
 			return errTooManyBatchers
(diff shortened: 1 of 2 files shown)


processor/batchprocessor/batch_processor.go
```diff
@@ -307,9 +307,9 @@ func (mb *multiShardBatcher) consume(ctx context.Context, data any) error {
 	}
 	aset := attribute.NewSet(attrs...)
 
+	mb.lock.Lock()
```
sfc-gh-bdrutu, line 310: This is a very poor solution to the problem. A way simpler and correct solution would be to just re-check the map and leave fast if already there after capturing the lock.

sfc-gh-bdrutu (changes requested):

Issue #9458, opened 2024-02-05
[configcompression] Use if instead of switch on UnmarshalText


mx-psi:
nit: Sorry, but my mind goes crazy when I see a switch used like an if :)) Can we fix this probably in a separate PR :)

_Originally posted by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/pull/9416#discussion_r1474920110_

mx-psi:
I guess ideally we would have a linter for this

AkhigbeEromo:
Hello @TylerHelmuth @mx-psi , i would like to work on this issue

PR #9826, opened 2024-03-23
Used an if statement instead of switch on unmarshal text in `configcompression`

AkhigbeEromo:
Changed the switch statement in `UnmarshalText` function to an if statement {configcompression}
Link to the issue:  #9458


AkhigbeEromo:
@TylerHelmuth please do you know why its failing?

TylerHelmuth:
@AkhigbeEromo the only reason I can think that the check is failing is bc your branch name is really long.  What happens if you run `make apidiff-compare` locally?

Issue #9812, opened 2024-03-21
[configgrpc] Update public methods to accept context.Context


TylerHelmuth:

AkhigbeEromo:
hello @TylerHelmuth, while working on this issue I realized that all the public functions in `configgrpc.go` already accept context. The only public function that does not accept context is a deprecated function called `SanitizedEndpoint()`

TylerHelmuth:
@AkhigbeEromo correct.  There is 1 more round of deprecation and then we'll be able to call this closed.

Issue #9482, opened 2024-02-06
[configrpc] Remove `SanitizedEndpoint`?


mx-psi:
As far as I can tell, this function is not used by any code, either in core or contrib. Should we remove it?

TylerHelmuth:
I think we need to reconsider removing this feature.  The https://github.com/open-telemetry/opentelemetry-collector/pull/9616 showed the otlpexporter relies on it.

We could reproduce the functionality directly in the otlpexporter, but leaving it in configgrpc seems ok as well.

mx-psi:
> I think we need to reconsider removing this feature. https://github.com/open-telemetry/opentelemetry-collector/pull/9616 showed the otlpexporter relies on it.
>
> We could reproduce the functionality directly in the otlpexporter, but leaving it in configgrpc seems ok as well.

@TylerHelmuth Do we have more than one usage of this function? If not, I would vote to move this to the OTLP exporter

TylerHelmuth:
@mx-psi it has been used in [configgrpc's `ToClientConn` function](https://github.com/open-telemetry/opentelemetry-collector/blob/6affda628707970bfdb875036b72cbf3ff04e9cd/config/configgrpc/configgrpc.go#L183) for a long time.  Looking at usage of `ToClientConn`, it is used in the `otlpexporter`, `coralogixexporter`, `opencensusexporter`, `otelarrowexporter`, `skywalkingexporter`, `jaegerremotesamplingextension`, and `testbed`.

atoulme:
The OTLP exporter has another way to deal with sanitized endpoint by stripping the scheme of the URL at unmarshaling time.

OK, so for this to happen:
1. We need embedded struct unmarshaling support https://github.com/open-telemetry/opentelemetry-collector/pull/9635
2. Then we can implement an unmarshaler for configgrpc. We need one anyway for https://github.com/open-telemetry/opentelemetry-collector/pull/9537

mx-psi:
@TylerHelmuth As you said, I don't think uses on `configrpc` itself count; we can just unexport the function, my concern is about having it as part of the public API.

PR #9836, opened 2024-03-25
[configgrpc] remove deprecated funcs

TylerHelmuth:
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9482
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9812


TylerHelmuth:
Depends on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31991

config/configgrpc/configgrpc.go
```diff
@@ -150,9 +150,7 @@ type ServerConfig struct {
 	IncludeMetadata bool `mapstructure:"include_metadata"`
 }
 
-// SanitizedEndpoint strips the prefix of either http:// or https:// from configgrpc.ClientConfig.Endpoint.
-// Deprecated: [v0.97.0]
```
mx-psi, line 154: Since the function still exists, we can keep the comment (after lowercasing the function name and removing the deprecation), can't we?

mx-psi (approved):
Just one nit about the comment

config/configgrpc/configgrpc.go
TylerHelmuth: Added the comment back

PR #9837, opened 2024-03-25
Log when Environment Provider tries to pull unset or empty env var

ankitpatel96:
**Description:** 
Creates a logger in the confmap.ProviderSettings and uses it to log when there is a missing or blank environment variable referenced in config.  For now the noop logger is used everywhere except tests.

**Link to tracking Issue:** [5615](https://github.com/open-telemetry/opentelemetry-collector/issues/5615)

**Testing:** 
I wrote unit tests that ensured 

1. logging occurred when an environment variable was unset 
2. logging occcured when the env var was empty.  
3. there was no log when an env var was used correctly

I also started the otel collector with the sample config - and added an env var reference in the sample config.  I then inserted a print statement next to each log call to see whether my code paths were hit in the live application.  I then went through the 3 cases mentioned above and ensured that logging behavior was accurate.


confmap/provider/envprovider/provider.go
```diff
@@ -9,28 +9,41 @@ import (
...
 	if !strings.HasPrefix(uri, schemeName+":") {
 		return nil, fmt.Errorf("%q uri is not supported by %q provider", uri, schemeName)
 	}
+	envVarName := uri[len(schemeName)+1:]
+	val, exists := os.LookupEnv(envVarName)
+	if !exists {
+		emp.logger.Warn(fmt.Sprintf("Environment variable %s is used in configuration but is unset", envVarName))
+	} else if len(val) == 0 {
+		emp.logger.Warn(fmt.Sprintf("Environment variable %s is used in configuration but is empty", envVarName))
+	}
```
mx-psi, line 44: The Collector structure logging approach is to have the same message always and keep parts that differ in named fields. 

Also, I am fine logging something when you have an empty variable but I would not make it a warning, maybe an info (or even debug?).

```suggestion
	if !exists {
		emp.logger.Warn("Configuration references unset environment variable", zap.String("name", envVarName))
	} else if len(val) == 0 {
		emp.logger.Info("Configuration references empty environment variable", zap.String("name", envVarName))
	}
```

confmap/provider.go
```diff
@@ -6,11 +6,23 @@ package confmap // import "go.opentelemetry.io/collector/confmap"
...
+	Logger *zap.Logger
+}
+
+func NewProviderSettings(logger *zap.Logger) ProviderSettings {
+	return ProviderSettings{Logger: logger}
+}
+
+func NewProviderSettingsNoopLogger() ProviderSettings {
+	return NewProviderSettings(zap.NewNop())
+}
```
mx-psi, line 25: When we actually have a useful logger, we would likely want to change this. To avoid having to go through breaking changes, I would define this on an `internal` package. We can either define this in a package within `confmap/internal` or, since we use these functions in `otelcol`, just on the top level internal folder

ankitpatel96: This is kinda tricky actually...
If i move just `NewProviderSettings` to confmap/internal then `NewProviderSettingsNoopLogger` has to call NewProviderSettings which imports ProviderSettings which is an import loop.  If I change `NewProviderSettingsNoopLogger` to just initialize the struct itself it defeats the point of having `NewProviderSettings` at all

If I move both functions in there, `NewProviderSettingsNoopLogger` can't be accessed everywhere it's needed.

I'm not sure about the utility about moving these into the top level internal folder either?  If its in the top level folder can't anyone access the functions?  Apologies if this is obvious honestly my Go package knowledge is kinda rusty.

At this point I'm tempted to just get rid of `NewProviderSettings` entirely...

mx-psi: >  If its in the top level folder can't anyone access the functions?

`<prefix>/internal/<suffix>` can only be imported in packages whose path shares `<prefix>`. So moving to a top-level internal folder (`go.opentelemetry.io/collector/internal/<something>`) means that anything within this repository can access it, but nothing outside of it can.

> At this point I'm tempted to just get rid of NewProviderSettings entirely...

I am also fine with that, it leads to some repetition but it should be temporary

mx-psi (commented):
LGTM after rewording the function name. We may undergo a bit of churn on the `confmap` code still, so I think it's okay to leave `NewNopProviderSettings` there until we figure out how the code structure will look like

.chloggen/log-env-empty.yaml
```diff
@@ -0,0 +1,26 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
```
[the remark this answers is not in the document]
ankitpatel96: great point..

mx-psi (approved):

confmap/provider/envprovider/provider.go
```diff
@@ -9,28 +9,41 @@ import (
...
 // NewWithSettings returns a new confmap.Provider that reads the configuration from the given environment variable.
 //
 // This Provider supports "env" scheme, and can be called with a selector:
 // `env:NAME_OF_ENVIRONMENT_VARIABLE`
-func NewWithSettings(confmap.ProviderSettings) confmap.Provider {
-	return &provider{}
+func NewWithSettings(a confmap.ProviderSettings) confmap.Provider {
+	return &provider{
+		logger: a.Logger,
+	}
```
evan-bradley, line 31:
```suggestion
func NewWithSettings(ps confmap.ProviderSettings) confmap.Provider {
	return &provider{
		logger: ps.Logger,
	}
```
Nit: could we make the variable name an abbreviation of the type? It's not immediately clear to me when reading this what `a` stands for. Really minor for how short this function is, but if it were to grow in the future it could be a bit of mental overhead.

evan-bradley (approved):
Looks good to me, thanks for adding this.

confmap/provider/envprovider/provider.go
ankitpatel96: great point!  fixed

Issue #9532, opened 2024-02-08
[confmap] Stricter types on configuration resolution


mx-psi:
We [enable](https://github.com/open-telemetry/opentelemetry-collector/blob/cf51a35663302c133ee6caba1946fcb7c2421ca4/confmap/confmap.go#L151) the [`WeaklyTypedInput` option from mapstructure](https://pkg.go.dev/github.com/mitchellh/mapstructure#DecoderConfig). This option implies the following implicit conversions from the internal `confmap`'s `map[string]any` to our `Config` structs:

1. bools to string (true = "1", false = "0")
2. numbers to string (base 10)
3. bools to int/uint (true = 1, false = 0)
4. strings to int/uint (using [`strconv.ParseInt`](https://pkg.go.dev/strconv#ParseInt))
5. int to bool (true if value != 0)
6. string to bool (accepts: 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. Anything else is an error)
7. empty array = empty map and vice versa
8. negative numbers to overflowed uint values (base 10)
9. slice of maps to a merged map
10. single values are converted to slices if required. Each
  element is weakly decoded. For example: "4" can become []int{4}
  if the target type is an int slice.

We also do implicit casting to string when expanding a `${provider:URI}` variable: https://github.com/open-telemetry/opentelemetry-collector/blob/cf51a35663302c133ee6caba1946fcb7c2421ca4/confmap/expand.go#L124-L139

I think this is too lax and commits us to supporting too many weird edge cases (in particular all but 2 and 4 seem like clearly things we should remove, and I would argue even 2 and 4 should be removed eventually). This conflicts with the changes I proposed on #9515, so I think we should resolve this one first.

djaglowski:
How do we assess the impact of this? Doesn't this effectively mean a very large number of breaking changes for users? 

> commits us to supporting too many weird edge cases

Can you give an example? Is the problem that custom unmarshalers may preempt these input type conversions?

mx-psi:
@djaglowski Here is an example configuration that is accepted by the OpenTelemetry Collector today and has all the edge cases listed above for `WeaklyTypedInput`:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        max_concurrent_streams: -1 # Wraps around to math.MaxUint32 - 1 (8)

exporters:
  otlp:
    endpoint: endpoint.local
    headers:
[...]

Issue #9531, opened 2024-02-08
[confmap] Restrict characters in environment variable URIs even more!


mx-psi:
In #5706, we discussed restricting URIs in characters for all providers. We ended up disallowing `$` in #6268 for the recursive implementation done later.

I think we should restrict the character set even more for the `env` provider, in particular by following the restrictions defined by the OpenTelemetry Configuration Working Group (see [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution)). While, as mentioned in https://github.com/open-telemetry/opentelemetry-collector/issues/5706#issuecomment-1189909801, there are more characters allowed on both Windows and Linux, the set of environment variable names allowed by the spec is reasonable and (pressumably!) what the vast majority of users use.

We should probably restrict it also on the `expandconverter` while it exists

Issue #8565, opened 2023-10-03
[confmap] Strings containing numbers in nonstandard notation are normalized to decimal notation


rassmate:
**Describe the bug**
When using the $env:ENVIRONMENT_VARIABLE_NAME syntax in the collector configuration a string with numbers only starting with zero is interpreted as a number.

**Steps to reproduce**
https://github.com/rassmate/opentelemetry-collector/commit/855679d1b20f46b214f2389d590ee79573617337

**What did you expect to see?**
When working with environment variables I expected to get the exact value that the environment variable holds.

**What did you see instead?**
A string got interpreted into a number very different from the one in the environment variable when the value had a leading zero.

**What version did you use?**
v0.84

**What config did you use?**
```yaml
env: 
  name: TEST
  value: "0123"
...
processors:
  transform:
    error_mode: ignore
    trace_statements:
      - context: span
        statements:
          - set(resource.attributes["TEST"], "${env:TEST}") 
... 
```      

**Environment**
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")
Macbook air m2. Same error in openshift cluster

**Additional context**

mx-psi:
Thanks for reporting this.

This comes from the YAML library we use, a minimal example can be seen here: https://go.dev/play/p/eO1O_q2ITAU. It is documented on the README:

> Octals encode and decode as 0777 per YAML 1.1, rather than 0o777 as specified in YAML 1.2, because most parsers still use the old format

You don't need to be using `env` for this: directly setting an integer-type option to `01234` will cause it to be interpreted as the equivalent octal number (83). 

In your particular case, setting TEST to `"!!str 0123"` (using [YAML explicit tags](https://yaml.org/spec/1.2.2/#24-tags)) or  with explicit quotes as in `"\"0123\""` would be a valid workaround (this workaround is valid so long as the variable is not being used to set a configuration value of `int` type).

I asked upstream for a toggle on the library to avoid this quirk on go-yaml/yaml/issues/996.

rassmate:
Ok, thanks for the clarification.

I think it is a bit strange though that in the envprovider.go even use the yaml-parser to get environment variable values.
I would have guessed on just fetching the clean value from the environment without any parsing.

mx-psi:
Right, the `${env:...}` feature actually supports YAML syntax, e.g. you can do

```yaml
processors:
[...]

PR #9854, opened 2024-03-26
[docs/rfc] RFC about environment variables

### LINKED ISSUE #9515: [confmap] Decide on desired behavior for `${..}` and `${env:..}`
This writeup is intended to solve #8565 and define what to do with configuration resolution mechanisms before we try to remove the expandconverter.

### Current state

Assume you have a Collector instance that has a field with type `string` and another field with type `int`. Let's also assume that there are two environment variables:

- `UNQUOTED_OCTAL` set to the value `0123`
- `QUOTED_OCTAL` set to the value `"0123"`

Lastly assume that:
- the string field is set to `Value is <input>`
- the integer field is set to `<input>`

Where the value of `<input>` varies according to the table value. What should the configuration resolve to in this situation, if anything? The following table summarizes the current state:

<details>
<summary>Current behavior (click to expand)</summary>

|     `<input>`    | Interpolated string field value |       integer field value                                         |
|------------------------|----------------|-------------------------------------------------------|
| 0123 | `Value is 0123` | `83` |
| "0123" | `Value is "0123"` |  `83` |
| ${env:UNQUOTED_OCTAL}  | `Value is 83`        | `83`                                                  |
| ${env:QUOTED_OCTAL}    | `Value is 0123`      | `83`                                                  |
| ${UNQUOTED_OCTAL}      | `Value is 0123`       | `83`                                                  |
| ${QUOTED_OCTAL}        | `Value is "0123"`     | error  |

</details>

### Discussion of current state

I feel like a desirable property of our configuration system would be the following:

> **(Property 1)** Let `str`, `a`, `b` and `c` be strings and let `<s>` be the value of a string `s`. Assume that `str` is the concatenation of `a`, `b`, and `c` (i.e. `<str>` is the same as `<a><b><c>`). Then a configuration with a field set to `<a>${env:B}<c>` and an environment variable `B` set to `<b>` result in the same configuration value as a field set to `<str>`.

A second property that seems desirable to me is the following:

> **(Property 2)** Let  `ENV` be an environment variable name. Assume that both `${env:ENV}` and `${ENV}` both resolve to a scalar configuration value. Then a string field set to `${env:ENV}` and a field set to `${ENV}` should resolve to the same configuration.  

The table above shows that we do not satisfy these naive properties:
1. The difference between `Value is 0123` and `Value is ${env:UNQUOTED_OCTAL}` violates property 1.
2. The difference between `Value is "0123"` and `Value is ${env:QUOTED_OCTAL}` violates property 1.
3. The difference between `${env:QUOTED_OCTAL}` and `${QUOTED_OCTAL}` violates property 2.

I find violation 1 in particular very surprising, so I am going to focus on it. 

### Why (1) happens

In the envprovider we use `yaml.Unmarshal` into a structure . When parsing `${env:UNQUOTED_OCTAL}`, the `0123` value will get parsed into an integer in octal base, and stored as `83` in the `confmap.Retrieved` struct. The fact that this was written in octal is then lost.

When we get to unmarshaling, since we set `WeaklyTypedInput: true` in our mapstructure configuration, [the field will be converted to a string in base 10](https://github.com/mitchellh/mapstructure/blob/8508981c8b6c964e6986dd8aa85490e70ce3c2e2/mapstructure.go#L580).

### If we want to fix this...

I think we can make two changes to the configuration resolution logic that would address these. I am more convinced about change 1 than about change 2

#### Change 1: Defer scalar value resolution until you get to mapstructure

Instead of `yaml.Unmarshal`ing into an `any` struct, we can unmarshal into a custom structure that defers evaluating scalar values and stores them as strings in the `confmap.Retrieved` struct. This seems possible with the current library we use (see a very rough PoC here: https://go.dev/play/p/g8hciqbI921). We leave the actual mapping to int/float/bool to the `mapstructure` library.

This would resolve violation 1, but not violation 2 or violation 3. The table would look like this (changes highlighted with a †):

<details>
<summary>Behavior after change 1 (click to expand) </summary>

|     `<input>`    | Interpolated string field value |       integer field value                                         |
|------------------------|----------------|-------------------------------------------------------|
| 0123 | `Value is 0123` | `83` |
| "0123" | `Value is "0123"` |  `83` |
| ${env:UNQUOTED_OCTAL}  | `Value is 0123`†         | `83`                                                  |
| ${env:QUOTED_OCTAL}    | `Value is 0123`      | `83`                                                  |
| ${UNQUOTED_OCTAL}      | `Value is 0123`       | `83`                                                  |
| ${QUOTED_OCTAL}        | `Value is "0123"`     | error  |

</details>

#### Change 2: Make `${..}` parse YAML syntax, but fail if the result is not a scalar value

We only support `${..}` when resolving to a scalar value. We can make it also parse YAML instead of passing the raw value (possibly with Change 1 included), and fail if the result is not a scalar value. (We can eventually make it work exactly as `${env:..}`, but for now this is a more minimal change).

This would not work the same way as the Configuration WG proposal for the `${..}`, which is the main reason I am not entirely convinced about doing this, but the difference would be in edge cases like `"0123"` (explicit quotes) or `!!str 0123` (in the latter case in particular it feels like it's clearly okay to parse as YAML since the user clearly used YAML specific syntax).

<details>
<summary>Behavior after change 1 + change 2 (click to expand) </summary>

|     `<input>`    | Interpolated string field value |       integer field value                                         |
|------------------------|----------------|-------------------------------------------------------|
| 0123 | `Value is 0123` | `83` |
| "0123" | `Value is "0123"` |  `83` |
| ${env:UNQUOTED_OCTAL}  | `Value is 0123`†         | `83`                                                  |
| ${env:QUOTED_OCTAL}    | `Value is 0123`      | `83`                                                  |
| ${UNQUOTED_OCTAL}      | `Value is 0123`       | `83`                                                  |
| ${QUOTED_OCTAL}        | `Value is 0123`†     | error  |

</details>

### Things to decide before confmap 1.0

1. Are property 1 and property 2 things we want?
2. If the answer to (1) is yes, is adhering to this properties something to do before 1.0?
3. If the answer to (2) is yes, should we do change 1?
4. If the answer to (2) is yes, should we do change 2?

mx-psi:
**Description:**  Adds an RFC about how environment variable resolution should work
**Link to tracking Issue:** Fixes #9515, relates to:
- #8215
- #8565
- #9162
- #9531 
- #9532


docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+1.  ***Expansion should happen only when the user expects it***. We
+    should aim to expand when the user expects it and keep the original
+    value when we don't (e.g. because the syntax is used for something
+    different).
+2.  ***Expansion should have predictable behavior***.
+3.  ***Multiple expansion methods, if present, should have similar behavior.***
+    Switching from `${env:ENV}` to `${ENV}` or vice versa
+    should not lead to (too many?) surprises.
+4.  ***When the syntax overlaps, expansion should be aligned with*** 
+    [***the expansion defined by the Configuration Working Group***](https://github.com/open-telemetry/opentelemetry-specification/blob/032213cedde54a2171dfbd234a371501a3537919/specification/configuration/file-configuration.md#environment-variable-substitution).
```
mx-psi, line 35: I filed open-telemetry/opentelemetry-specification/issues/3963 to ensure this. I don't think this should completely block merging this RFC: I think at this point it is unavoidable that there will be differences with respect to the File Configuration SDK spec, but this should be a design goal.

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+#### We are less restrictive than the Configuration WG
+
+The Configuration WG defines an [*environment variable expansion feature
+for SDK
+configurations*](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution).
+This accepts only non empty alphanumeric + underscore identifiers
+starting with alphabetic or underscore. If the Configuration WG were to
+expand this in the future (e.g. to include other features present in
+Bash-like syntax), we would not be able to expand our braces syntax to
+support new features without breaking users.
```
mx-psi, line 153: As an example, open-telemetry/opentelemetry-specification/pull/3948 proposes to add the Bash syntax for specifying a default

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+
+1.  The *braces syntax*, `${ENV}`.
+2.  The *env provider syntax*, `${env:ENV}`.
+
+These both have **the same character set and behavior**. They both use
+the env provider under the hood.
+
+The naked syntax supported in BASH is not supported in the Collector.
+Escaping is supported by using two dollar signs. Escaping is also
+honored for unsupported identifiers like `${1}`.
```
mx-psi, line 170: This is a bit vague, I don't know how escaping should actually work. There is some discussion about this for the SDK file configuration over at open-telemetry/opentelemetry-specification/issues/3914

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Character set
+
+An environment variable identifier must be a nonempty ASCII alphanumeric
+or underscore starting with an alphabetic or underscore character. Its
+maximum length is 200 characters. Both syntaxes support recursive
+resolution.
+
+When an invalid identifier is found, a warning log is emitted, and the
+string is kept as-is.
```
mx-psi, line 200: I am undecided about this, I feel like we should fail in some cases but I am not sure about which. open-telemetry/opentelemetry-specification/issues/3981 has some more discussion about this

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+
+| Raw value    | Field type | Current behavior, `${ENV}`, single field | Current behavior,  `${env:ENV}` , single field | Desired behavior, entire field | Desired behavior, inline string field |
+|--------------|------------|------------------------------------------|------------------------------------------------|--------------------------------|---------------------------------------|
+| `123`        | integer    | 123                                      | 123                                            | 123                            | n/a                                   |
+| `0123`       | integer    | 83                                       | 83                                             | 83                             | n/a                                   |
+| `0123`       | string     | 0123                                     | 83                                             | 0123                           | 0123                                  |
+| `0xdeadbeef` | string     | 0xdeadbeef                               | 3735928559                                     | 0xdeadbeef                     | 0xdeadbeef                            |
```
mx-psi, line 212: We had some discussion about the desired behavior when setting the entire field at KubeCon EU (cc @evan-bradley). Should this be an error instead? Should we use base 10 notation?

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+| `"0123"`     | string     | "0123"                                   | 0123                                           | 0123                           | 0123                                  |
+| `!!str 0123` | string     | !!str 0123                               | 0123                                           | 0123                           | 0123                                  |
```
mx-psi, line 214: For 'Desired behavior, inline string field' here, the behavior of the File Configuration SDK would be to set `!!str 0123`. I think this is a rare edge case, but as written it would set `0123` since the `AsString` method would not return the original representation

mx-psi (commented):
Self-reviewing to start some threads about topics I want to discuss

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Goals of an expansion system
+
+The following are considered goals of the expansion system:
+
+1.  ***Expansion should happen only when the user expects it***. We
+    should aim to expand when the user expects it and keep the original
+    value when we don't (e.g. because the syntax is used for something
+    different).
+2.  ***Expansion should have predictable behavior***.
```
codeboten, line 30:
```suggestion
2.  ***Expansion must have predictable behavior***.
```

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,216 @@
...
+
+These differ in the character set allowed for environment variable names
+as well as the type of parsing they return. Escaping is supported in all
+syntaxes by using two dollar signs.
+
+### Type casting rules
+
+A provider or converter takes a string and returns some sort of value
+after potentially doing some parsing. This gets stored in a
+confmap.Conf. When unmarshalling, we use mapstructure with
```
codeboten, line 54:
```suggestion
`confmap.Conf`. When unmarshalling, we use [mapstructure](https://github.com/mitchellh/mapstructure) with
```

docs/rfcs/env-vars.md
evan-bradley: I've been mulling over this some more. I think the desired behavior expressed here is the right way to go.

ankitpatel96 (approved):

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,220 @@
...
+for numeric types. The `yaml.v3` parser mostly follows the YAML v1.2
+specification with [*some
+exceptions*](https://github.com/go-yaml/yaml#compatibility). You can see
+how it works for some edge cases in this example:
+[*https://go.dev/play/p/RtPmH8aZA1X*](https://go.dev/play/p/RtPmH8aZA1X).
+
+When unmarshalling, we use mapstructure with WeaklyTypedInput
+**disabled**. We check via a hook an `AsString` method from confmap.Conf
+and use its return value when it is valid and we are mapping to a string
+field. This method has default casting rules for unambiguous scalar
```
songy23, line 189: Curious about the motivation here. My preference would be not doing any casting but simply returning the original representation for fields expecting string type, e.g. `0123` -> `"0123"` rather than `"83"`

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Desired behavior
+
+*This section is written as if the changes were already implemented.*
+
+The Collector supports **two** different syntaxes for environment
+variable resolution:
+
+1.  The *braces syntax*, `${ENV}`.
+2.  The *env provider syntax*, `${env:ENV}`.
```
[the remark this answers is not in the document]
mx-psi: Being compatible with the Configuration from SDKs. The Configuration WG has pushed against this see e.g. the comments on open-telemetry/opentelemetry-specification/pull/3974.

I would prefer supporting a single syntax, but I think consistency with SDKs trumps that

docs/rfcs/env-vars.md
mx-psi: @TylerHelmuth The one edge case here is `${1}` from open-telemetry/opentelemetry-collector-contrib/issues/9984. Should we fail in that case and force uses to use escaping syntax?

docs/rfcs/env-vars.md
mx-psi: This is what I am proposing (see the table at the bottom of the document). Would you word it differently to make it clearer?

songy23: Oops I mis-read the 2nd row in the examples. I think this is fine then, maybe mention the examples below on how this works.

docs/rfcs/env-vars.md
mx-psi: Since that is what we currently do and it seems impossible to detect all kinds of syntaxes past or future that would overlap with this, I have changed this to an error on 3f0546741415843f5dca369b80d263a239fb31f2

docs/rfcs/env-vars.md
mx-psi: I explicitly mention the comparison table in the body of the text now :)

songy23 (approved):
LGTM

docs/rfcs/env-vars.md
mx-psi: I clarified this with a regex, I think we can follow the discussion on the spec issue if anything needs further clarification

docs/rfcs/env-vars.md
TylerHelmuth: @mx-psi I feel like `${1}` corresponds with the regex and should be expanded. If users need the literal `${1}` can they do `$${1}`?

mx-psi: > I feel like since ${1} corresponds with the regex it should be expanded

I think we are on the same page, but to be clear `1` is not a valid identifier per the 'desired behavior' section (valid identifiers need to start with underscore or ASCII alphabetic characters), so it should **not** be expanded, but rather raise an error.

TylerHelmuth: Ah ok, then we are on the same page, I just got the outcome wrong. My opinion is the `${1}` should fail and if users need that literal string they should escape: `$${1}`.

This prevents us from having any edge cases with our unwrap syntax while keeping all functionality available.

docs/rfcs/env-vars.md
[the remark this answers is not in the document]
mx-psi: Added a sentence on c8307a0256c7ef56e4f72eb8c7709071d7f3347c to make this more clear

mx-psi:
### Remaining work

- [ ] Decide on https://github.com/open-telemetry/opentelemetry-collector/pull/9854#discussion_r1561181980 by raising this on the spec and seeing if we can align on that
- [ ] Decide on https://github.com/open-telemetry/opentelemetry-collector/pull/9854#discussion_r1549449041. I am going to try and do a PoC of the confmap behavior to double check what makes more sense

astencel-sumo (approved):

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,222 @@
...
+
+These both have **the same character set and behavior**. They both use
+the env provider under the hood. This means we support the exact same
+syntax as the Configuration WG.
+
+The naked syntax supported in BASH is not supported in the Collector.
+Escaping is supported by using two dollar signs. Escaping is also
+honored for unsupported identifiers like `${1}` (i.e. anything that
```
TylerHelmuth, line 175: Based on our discussion in [this thread](https://github.com/open-telemetry/opentelemetry-collector/pull/9854#discussion_r1549442566) this sentence should be updated to clarify that `${1}` is no supported.

docs/rfcs/env-vars.md
TylerHelmuth: @mx-psi where in the spec do we need to raise this issue?

docs/rfcs/env-vars.md
```diff
@@ -0,0 +1,222 @@
...
+variable resolution:
+
+1.  The *braces syntax*, `${ENV}`.
+2.  The *env provider syntax*, `${env:ENV}`.
+
+These both have **the same character set and behavior**. They both use
+the env provider under the hood. This means we support the exact same
+syntax as the Configuration WG.
+
+The naked syntax supported in BASH is not supported in the Collector.
```
evan-bradley, line 173:
```suggestion
The naked syntax supported in Bash is not supported in the Collector.
```
Nit

docs/rfcs/env-vars.md
evan-bradley: >For 'Desired behavior, inline string field' here, the behavior of the File Configuration SDK would be to set `!!str 0123`.

Can you link to where this is defined? I can't see where this is addressed this outside of declaring a suggested minimum version of the YAML spec.

docs/rfcs/env-vars.md
mx-psi: Tyler and I discussed this offline but for the benefit of anybody else keeping track of this: this is being discussed on open-telemetry/opentelemetry-specification/pull/4002

codeboten:
Let's merge this and change it as needed, the spec change was merged 👍🏻

Issue #9855, opened 2024-03-26
Components subcommand no longer outputs component names


FISHMANPET:
**Describe the bug**
The output of the `components` subcommand no longer contain the names of the component, merely `{}`

**Steps to reproduce**
1. Download the latest collector or build your own.
2. Run `collector components`

**What did you expect to see?**
I expected the output to include the names of components, like this (from the 0.96.0 release)
```
    - name: signalfx
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
```
**What did you see instead?**
```
    - name: {}
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
```

**What version did you use?**
v0.97.0

**What config did you use?**
N/A

**Environment**
MacOS 13.6.6
For the compiled version, built using ocb 0.97.0, running go 1.22.1

**Additional context**
I'm wondering if this is related to the changes described in #9208 ?

TylerHelmuth:
Tested this with `make otelcontribcol` in Contrib and was able to reproduce:

```
buildinfo:
    command: otelcontribcol
    description: Local OpenTelemetry Collector Contrib binary, testing only.
    version: 0.97.0-dev
receivers:
    - name: {}
      stability:
        logs: Undefined
        metrics: Beta
        traces: Undefined
    - name: {}
      stability:
        logs: Development
        metrics: Alpha
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Alpha
        traces: Undefined
    - name: {}
      stability:
        logs: Beta
        metrics: Beta
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Alpha
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Alpha
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Beta
        traces: Undefined
    - name: {}
      stability:
        logs: Alpha
        metrics: Beta
        traces: Alpha
    - name: {}
      stability:
        logs: Undefined
        metrics: Alpha
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Beta
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Beta
        traces: Undefined
    - name: {}
      stability:
        logs: Undefined
        metrics: Undefined
        traces: Beta
    - name: {}
      stability:
        logs: Undefined
        metrics: Beta
[...]

PR #9856, opened 2024-03-26
[component] make Type implement MarshalText

TylerHelmuth:
Adds `MarshalYAML` function so that `Type` can be properly marshaled as yaml.

We could alternatively fix this issue by updating the `components` command to explicitly call `Type.String` and use that value instead of `component.Type` in its struct.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9855

Tested locally using `otelcorecol`.  The unit test `TestNewBuildSubCommand` didnt catch the issue because it was also using `yaml.Marshal` like the command, so the expected value also had `{}` as the name.


component/config.go
```diff
@@ -119,6 +119,11 @@ func (t Type) String() string {
 	return t.name
 }
 
+// MarshalYAML marshals returns the Type name.
+func (t Type) MarshalYAML() (interface{}, error) {
```
dmitryax, line 123: I'm surprised we use yaml in that command. I think `MarshalText` should work as well. That would be consistent with the rest of the codebase

dmitryax:
Also, the test for this functionality is pretty useless. It'd be great to update it to compare the expected output with an actual yaml string instead of doing the same marshaling. Could you please update it?

component/config.go
TylerHelmuth: I just tried with `MarshalText` and it didn't work.  I can include both functions tho.

TylerHelmuth: NVM, it works if the implementation of `MarshalText` actually conforms to the interface lol

TylerHelmuth:
I believe the windows test failure is unrelated

mx-psi (approved):

dmitryax (approved):
Thanks, @TylerHelmuth!

Issue #9809, opened 2024-03-21
[configretry] Update public methods to accept context.Context


TylerHelmuth:

AkhigbeEromo:
Hello @TylerHelmuth, Just submitted a PR concerning this issue

TylerHelmuth:
Reviewing the module there are only 2 public functions, `Validate` and `NewDefaultBackOffConfig`. Neither of these functions qualify for adding context, so this module is covered.

AkhigbeEromo:
Okay, understood. Will close the PR now

PR #9857, opened 2024-03-26
Updated public methods to accept context.context in `configretry`

AkhigbeEromo:
Deprecate `NewDefaultBackOffConfig` and use `NewDefaultBackOffConfigContext` instead.

Link: https://github.com/open-telemetry/opentelemetry-collector/issues/9809


dmitryax:
We don't need add context to the `NewDefault...Config` methods

TylerHelmuth:
https://github.com/open-telemetry/opentelemetry-collector/issues/9809 is closed so we won't need this PR.

PR #9861, opened 2024-03-27
[confmap] support unmarshaling for embedded structs with and without squashing

atoulme:
This is taking a small slice of #9750 to document the behavior of confmap and make sure we can unmarshal embedded structs.

PR #9862, opened 2024-03-27
[confmap] Remove top level condition when considering struct as Unmarshalers.

atoulme:
This is a slice of #9750 focusing on removing the top level condition on unmarshaling structs.


confmap/confmap.go
```diff
@@ -158,7 +159,7 @@ func decodeConfig(m *Conf, result any, errorUnused bool) error {
 			mapKeyStringToMapKeyTextUnmarshalerHookFunc(),
 			mapstructure.StringToTimeDurationHookFunc(),
 			mapstructure.TextUnmarshallerHookFunc(),
-			unmarshalerHookFunc(result),
+			unmarshalerHookFunc(result, topLevelUnmarshaling),
 			// after the main unmarshaler hook is called,
 			// we unmarshal the embedded structs if present to merge with the result:
 			unmarshalerEmbeddedStructsHookFunc(),
```
mx-psi, line 165: If I understand correctly, there is no need to have `self` living at the `confmap.Conf` level, this is just used by `decodeConfig`. Could we have some sort of helper struct that holds `self` and has the unmarshaling hooks as methods? I think that would be cleaner (keeps the logic contained within `decodeConfig` instead of spilling into `confmap.Conf` itself) and possibly simpler to understand

confmap/confmap_test.go
```diff
@@ -454,6 +454,7 @@ func TestEmbeddedUnmarshalerError(t *testing.T) {
 }
 
 func TestEmbeddedMarshalerError(t *testing.T) {
+	t.Skip("This test fails because the main struct calls the embedded struct Unmarshal method, and doesn't execute the embedded struct hook.")
```
mx-psi, line 457: What change/PR/issue to be addressed will make us stop skipping this test?

atoulme:
Blocked on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31802

confmap/confmap_test.go
atoulme: You will need this one: https://github.com/open-telemetry/opentelemetry-collector/pull/9861

confmap/confmap_test.go
```diff
@@ -494,7 +495,7 @@ func TestDirectUnmarshaler(t *testing.T) {
 		private: "keep already configured members",
 	}}
 	assert.NoError(t, tc.Unmarshal(cfgMap))
-	assert.Equal(t, "make sure this is only called directly", tc.Another)
+	assert.Equal(t, "make sure this is only called directly is only called directly", tc.Another)
```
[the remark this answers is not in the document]
atoulme: not a typo. This is in an awkward state because of the reduction of the PR down to this change.

confmap/confmap.go
atoulme: I don't think that works. This change addresses the successive calls on `confmap.Conf`, being called recursively by `Unmarshaler`. The `confmap.Conf` object has to be stateful to keep track of its top level object.

mx-psi: I am at a loss here, sorry. If `confmap.Conf` has to be stateful, then how does https://github.com/open-telemetry/opentelemetry-collector/pull/9750#discussion_r1525365294 also hold? Wouldn't we have to reset the state each time we call `Unmarshal`?

atoulme: It resets the state before calling `Unmarshal` on the `Unmarshaler` interface, yes.

confmap/confmap.go
```diff
@@ -287,7 +288,9 @@ func unmarshalerEmbeddedStructsHookFunc() mapstructure.DecodeHookFuncValue {
 			// embedded structs passed in via `squash` cannot be pointers. We just check if they are structs:
 			if to.Type().Field(i).IsExported() && to.Type().Field(i).Anonymous {
 				if unmarshaler, ok := to.Field(i).Addr().Interface().(Unmarshaler); ok {
-					if err := unmarshaler.Unmarshal(NewFromStringMap(fromAsMap)); err != nil {
+					c := NewFromStringMap(fromAsMap)
+					c.self = unmarshaler
```
mx-psi, line 292: Would storing the `skipTopLevelUnmarshaler` bool instead make a difference? If so, we should add a test where it makes a difference.

confmap/confmap_test.go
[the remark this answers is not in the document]
mx-psi: Okay, I ran this locally and I get this now. I should have done that before, sorry.

I don't think it makes sense to both skip the test and change it on the same PR. I guess we can just merge #9861 before this one and that deals with this test.

confmap/confmap.go
```diff
@@ -39,7 +39,8 @@ func NewFromStringMap(data map[string]any) *Conf {
 // Conf represents the raw configuration map for the OpenTelemetry Collector.
 // The confmap.Conf can be unmarshalled into the Collector's config using the "service" package.
 type Conf struct {
-	k *koanf.Koanf
+	k    *koanf.Koanf
+	self any
```
mx-psi, line 43: Let's add documentation for this. This is my suggestion, which should also serve to see if I finally understood the code

```suggestion

        // self stores the struct where we are unmarshaling Conf.
        // Conf structs built outside of the confmap package will have this set to nil.
        // self it not nil on auxiliary Conf structs built in the mapstructure hooks that check for the Unmarshaler interface.
	self any
```

mx-psi (commented):
I think I finally understood this! I filed #9879 separately that adds two confmap tests. I would want to merge that one and #9861 first if that makes sense to you.

confmap/confmap_test.go
mx-psi: Thanks! Let's continue discussing this bit on this thread https://github.com/open-telemetry/opentelemetry-collector/pull/9862#discussion_r1546384483

mx-psi:
It would also be interesting to try and reproduce the failing contrib test on a unit test here, so we can work on it. The offending `Unmarshal` function  is here https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e4c5b51ec0e24ca163bcb64d944455577f90a09f/pkg/stanza/operator/helper/time.go#L49

confmap/confmap_test.go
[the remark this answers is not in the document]
atoulme: Will rebase once it's merged.

atoulme:
> It would also be interesting to try and reproduce the failing contrib test on a unit test here, so we can work on it. The offending `Unmarshal` function is here https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e4c5b51ec0e24ca163bcb64d944455577f90a09f/pkg/stanza/operator/helper/time.go#L49

added a skipped test here: TestRecursiveUnmarshaling

atoulme:
If you want a foolproof solution, we might have to do something a little more drastic. We might need to keep a hashset of all pointers of unmarshaler methods ever called, to make sure we don't run into cycles.

mx-psi:
> If you want a foolproof solution, we might have to do something a little more drastic. We might need to keep a hashset of all pointers of unmarshaler methods ever called, to make sure we don't run into cycles.

I guess we can forbid what the stanza package does instead. Do you think there is a benefit to being able to do what stanza does? I can't see any advantage right now

atoulme:
No, I don't see any advantage, except maybe if you're trying to unmarshal and set all values on the struct atomically. Even then, this is a corner case. I think we're ok.

confmap/confmap_test.go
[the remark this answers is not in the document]
atoulme: meh, that's still happening. This test is no longer valid. I added a comment explaining what is happening.

confmap/confmap.go
atoulme: OK, that worked. Until such time there is a reason to revisit, let's keep it simple with your approach.

mx-psi (commented):
I am okay with the current approach, I guess we should document the limitation about nested unmarshalling before merging, but I think we can go with this.

cc @open-telemetry/collector-approvers, I would appreciate a second pair of eyes since this is somewhat subtle. Even if this is ready before, I will wait at least until Monday to merge to give others some time to review.

atoulme:
One more PR in contrib needed: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32577

confmap/confmap.go
```diff
@@ -40,7 +40,8 @@ func NewFromStringMap(data map[string]any) *Conf {
 // Conf represents the raw configuration map for the OpenTelemetry Collector.
 // The confmap.Conf can be unmarshalled into the Collector's config using the "service" package.
 type Conf struct {
-	k *koanf.Koanf
+	k                       *koanf.Koanf
+	skipTopLevelUnmarshaler bool
```
evan-bradley, line 44: At the risk of being overly verbose, could you describe why we need this field? I'm not fully up-to-speed on the change that this PR fits into, but just looking at the diff here, it's not immediately clear to me what the overall goal is. Given that this code is fairly nuanced and somewhat complex, I think having a few sentences for any future code spelunkers would be helpful.

We could also add it somewhere else if this field isn't the right place, this just seems like something I can fairly easy hit "see references" in my editor to follow.

atoulme: added more comments. Not sure that is enough though.

atoulme:
The last contrib issue is solved by https://github.com/open-telemetry/opentelemetry-collector/pull/9862. Unfortunately, I couldn't find a good way to fix this - in effect, a breaking change.

confmap/confmap.go
evan-bradley: Thank you for the detailed explanation! I think the comments help, though I think in the comment on the field, it would be nice if you explain what the "circular dependency" is. I think describing how we get ourselves into an infinite recursion by showing the call chain would help.

I have another question on the overall picture here. With what you've described, I think the approach in this PR makes sense. That said, while looking through this, I've noticed it seems a majority of what most components do in the `Unmarshal` methods on their configs actually has to do with validation. For example, in the logging exporter snippet you referenced, I think this could be moved to a `Validate` method that occurs after unmarshaling has finished. The checks themselves could be done on fields on the unmarshaled `Config` struct and I don't think should require access to the confmap object.

```golang
if conf.IsSet("loglevel") && conf.IsSet("verbosity") { 
 	return fmt.Errorf("'loglevel' and 'verbosity' are incompatible. Use only 'verbosity' instead") 
 } 
```

I would assume most components should only implement an `Unmarshal` method to fix e.g. parsing of ULID/UUIDv8 IDs when unmarshaling. If we moved most of the validation logic to `Validate`, I think we could even eliminate a lot of `Unmarshal` implementations. Do you think this is viable, and if so, could we use that to reduce scope on what `Unmarshal` needs to do and reduce complexity here?

atoulme: You are right to point out that we can move code to `Validate`. That makes sense and would be a better place in the lifecycle. If you can get away with it, that's the better route.

`Unmarshal` however is the only place where we can see the raw yaml, and make actions based of this. For example, when we have a default configuration that assumes a configuration to be present, we want to see if the specific key is not set in the yaml to delete it.

You can see it in action here:
https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.go#L68-L70

We also want to avoid breaking changes in configuration. For example, we will handle the incoming yaml and remap it to manage deprecated attributes.
Going to your example, if we don't do this in the `Unmarshal` method, we end up adding fields to the `Config` object just to perform validation later:
https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/loggingexporter/config.go#L63-L85
Here is a more complex example:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/config.go#L585-L630

This is also in play with our discussions around configgrpc, where we see a need to set the configuration right at unmarshaling time rather than making all sorts of introspection later:
https://github.com/open-telemetry/opentelemetry-collector/pull/9537#discussion_r1484541860

evan-bradley: >For example, when we have a default configuration that assumes a configuration to be present, we want to see if the specific key is not set in the yaml to delete it.

This is a good callout, thanks. I noticed those statements when looking through some config structs, but missed the reason why that was happening. To solve this case, could we add a struct tag that indicates a pointer field should be cleared if it hasn't been set by the user? The caveat here is that we couldn't set this on fields that are defined outside of the component's package, but I think it would solve most of our use cases.

>We also want to avoid breaking changes in configuration. For example, we will handle the incoming yaml and remap it to manage deprecated attributes.
Going to your example, if we don't do this in the `Unmarshal` method, we end up adding fields to the `Config` object just to perform validation later:

Personally, I would prefer to handle both of these cases by adding pointer fields to the `Config` object that are checked during `Validate`. I think it would be best if the `Config` struct acts as the source of truth for renames/optional fields/etc. and components don't mix which fields are recognized (deprecated or not) between the struct and methods on the struct.

>This is also in play with our discussions around configgrpc, where we see a need to set the configuration right at unmarshaling time rather than making all sorts of introspection later

It doesn't look like this reads from the `confmap.Conf` object inside `Unmarshal`. Could we also move this logic to `Validate`?

atoulme: > This is a good callout, thanks. I noticed those statements when looking through some config structs, but missed the reason why that was happening. To solve this case, could we add a struct tag that indicates a pointer field should be cleared if it hasn't been set by the user? The caveat here is that we couldn't set this on fields that are defined outside of the component's package, but I think it would solve most of our use cases.

We can have a decoder hook that does that. That sounds like a RFE for quality of life to cut down on `Unmarshal` functions. Maybe worth its own issue.

> Personally, I would prefer to handle both of these cases by adding pointer fields to the Config object that are checked during Validate. I think it would be best if the Config struct acts as the source of truth for renames/optional fields/etc. and components don't mix which fields are recognized (deprecated or not) between the struct and methods on the struct.

Sure, but that's not what we have been doing. I don't have a take on this ; just trying to make this work. Another approach you could consider is to use a `scratch` type field that is a `map[string]string` and can hold whatever was left after unmarshaling and didn't get mapped to a known field, so you can do this later.

> It doesn't look like this reads from the confmap.Conf object inside Unmarshal. Could we also move this logic to Validate?

In that particular case, we have to handle a delicate set of overlapping concerns with how the TLS settings are set based on the scheme of the endpoint. My particular concern was that it was surprising to see the config not declare TLS settings and yet infer them via the scheme endpoint. I'd much rather we set the config properly so when we report it to opAmp or whatever system, we're closer to what is effectively being configured.

evan-bradley: Got it, thank you for all the insight. Sounds like there may be some follow up work, but I don't see a reason any of that should block this.

confmap/confmap.go
```diff
@@ -41,6 +41,10 @@ func NewFromStringMap(data map[string]any) *Conf {
 // The confmap.Conf can be unmarshalled into the Collector's config using the "service" package.
 type Conf struct {
 	k *koanf.Koanf
+	// If true, upon unmarshaling do not call the Unmarshal function on the struct
+	// if it implements Unmarshaler and is the top-level struct.
+	// This avoids running into a circular dependency.
```
evan-bradley, line 46: We should define what the circular dependency is here. I think calling out the potential for an infinite recursion caused by `Unmarshaler.Unmarshal` and `Conf.Unmarshal` calling each other would probably be a good mention.

evan-bradley (approved):
Looks good to me, I think this is probably the cleanest solution given the current state of things.

mx-psi (approved):
LGTM after adressing Evan's comment. This will need an update on contrib since it's breaking the tests there

confmap/confmap_test.go
codeboten: I'm not sure I understand, with #9861 merged, is this skip still needed?

.chloggen/remove_top_level_block.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove top level lock when considering struct as Unmarshalers
```
codeboten, line 10: This says a top level lock is removed, but I don't see that in the change. Can you point me to where this lock is removed?

atoulme: here: https://github.com/open-telemetry/opentelemetry-collector/pull/9862/files#diff-3217721f23d028519f5993ab4e56eaff85988734dab4068a7eb411b838cd2786R331

confmap/confmap.go
atoulme: thanks, addressed.

.chloggen/remove_top_level_block.yaml
TylerHelmuth: @atoulme I don't see a lock in that snippet.

Or are you saying by adding `skipTopLevelUnmarshaler` to the condition you now allow invokers to choose to unmarshal to level struct?

atoulme: I have renamed the PR from using the term "lock" to "condition". Hopefully that clarifies a bit.

codeboten: Thanks @atoulme!

codeboten (approved):

codeboten:
@atoulme are these expected:

```
    config_warnings_test.go:92: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/datadogexporter/config_warnings_test.go:92
        	Error:      	elements differ
        	            	
        	            	extra elements in list B:
        	            	([]interface {}) (len=1) {
        	            	 (string) (len=125) "\"metrics::histograms::send_count_sum_metrics\" has been deprecated in favor of \"metrics::histograms::send_aggregation_metrics\""
        	            	}
        	            	
        	            	
        	            	listA:
        	            	([]string) (len=1) {
        	            	 (string) (len=125) "\"metrics::histograms::send_count_sum_metrics\" has been deprecated in favor of \"metrics::histograms::send_aggregation_metrics\""
        	            	}
        	            	
        	            	
        	            	listB:
        	            	([]string) (len=2) {
        	            	 (string) (len=125) "\"metrics::histograms::send_count_sum_metrics\" has been deprecated in favor of \"metrics::histograms::send_aggregation_metrics\"",
        	            	 (string) (len=125) "\"metrics::histograms::send_count_sum_metrics\" has been deprecated in favor of \"metrics::histograms::send_aggregation_metrics\""
        	            	}
        	Test:       	TestSendAggregations/metrics::histograms::send_count_sum_metrics_set_to_false
```

mx-psi:
@atoulme So that we can merge this soon and address @codeboten's comment, would you be able to do the following?

1. File a draft PR on opentelemetry-collector-contrib that pulls in this version of opentelemetry-collector + adds the necessary fixes in contrib. This is to 'prove' that we can satisfactorily fix the issue in contrib 
2. Post the PR from (1) here. Once this is done, we can merge this PR.
3. Update contrib and mark the PR from (1) as ready to review with only the fixes

atoulme:
Here is the PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32660

It is past midnight, I will follow up tomorrow on it.

atoulme:
The contrib PR shows the changes combined with test changes to contrib will work. I think we are at 2. I have reached out to Alex to make sure he's ok with this.

atoulme:
@codeboten are you ok with this approach and the steps I have followed, as well as the checks they provide? Can we merge this PR?

atoulme:
Can we agree to merge this by end of next week unless someone objects?

mx-psi:
I'll put a reminder to merge in a wekk, @codeboten feel free to merge before then if you are okay with it :)

PR #9885, opened 2024-04-03
move internal/testdata to pdata/testdata

codeboten:
This reduces dependencies from the consumer package while making testdata available across repos. It will allow us to remove duplicated code and its a fairly small surface area.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9886


atoulme (approved):
LGTM, I expect this package to become eventually stable though. There is no particular reason to not have it stable that I can think of.

codeboten:
@mx-psi it could definitely be marked as stable, i just didn't know if it's something we would care to do. There isn't a large surface area, but since its mostly used for tests I wasn't sure if we want to make guarantees about its stability

mx-psi:
> @mx-psi it could definitely be marked as stable, i just didn't know if it's something we would care to do. There isn't a large surface area, but since its mostly used for tests I wasn't sure if we want to make guarantees about its stability

I wouldn't mind getting this to 1.x, it wouldn't be too bad to get a 2.x of this module if we ever need to, given there will be no non-test dependencies

mx-psi (approved):

PR #9891, opened 2024-04-04
[exporterhelper] Make concurrency limit apply per-batch

### LINKED ISSUE #9952: [exporterhelper] Batch sender produces inconsistent and suboptimal batch sizes
**Describe the bug**

From https://github.com/open-telemetry/opentelemetry-collector/issues/8122#issuecomment-2035883137

In my experiments, the batch sender is producing inconsistent batch sizes which could be lower than desired due to goroutine scheduling even after https://github.com/open-telemetry/opentelemetry-collector/pull/9761 . The scenario I usually run into is that given queue sender concurrency = batch sender concurrency limit = N, and they are all blocked on send, when the send eventually returns, activeRequest will first be set to N-1, then a new consumer goroutine comes in and increments the active request, then realize N-1+1 == concurrencyLimit, and will send off the request right away, causing an undesirably small request to be exported without batching.

I tried to fix it by resetting bs.activeRequests to 0 next to close(b.done). While it fixes for sendMergeBatch, it will not work for sendMergeSplitBatch since it may be exporting something outside of activeBatch. I assume there might be a way around this for merge split batch, but I'm not sure if it is worth the trouble and complexity to workaround unfavorable goroutine scheduling.

**Steps to reproduce**

Use batch sender with sending_queue num_consumers=100, send events to it.

Batch sender consistently produces batch of size 100

**What did you see instead?**

Batch sender produces first batch of size 100, then most of the time size is 1.

**What version did you use?**

v0.97.0

**What config did you use?**

```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: localhost:4317
      http:
        endpoint: localhost:4318

exporters:
  elasticsearch:
    endpoints: [ "http://localhost:9200" ]
    logs_index: foo
    user: ****
    password: ****
    sending_queue:
      enabled: true
      storage: file_storage/elasticsearchexporter
      num_consumers: 100
      queue_size: 10000000
    retry:
      enabled: true
      max_requests: 10000

extensions:
  file_storage/elasticsearchexporter:
    directory: /tmp/otelcol/file_storage/elasticsearchexporter

service:
  extensions: [file_storage/elasticsearchexporter]
  pipelines:
    logs:
      receivers: [otlp]
      processors: []
      exporters: [elasticsearch]
```

**Environment**

Linux Mint 21.3

**Additional context**

carsonip:

The current issue with concurrency limit in batch sender is that given unfavorable goroutine scheduling, the batch size will be much smaller than MinSizeItems due to the concurrency limit check. This may overwhelm the backend with a lot of small requests, which defeats the purpose of batching.

With this PR, concurrency limit will only apply per-batch, meaning that active batch will be flushed due to concurrency check only if all the goroutines are waiting for the same active batch to be flush. This aims to avoid all the goroutines meaninglessly waiting for new requests to come in, while there will be no spare consumers from the queue to do so.

Compared to the existing implementation, for example, concurrencyLimit = 10, and a group of 5 goroutines are waiting for an export to finish, and 4 goroutines are pending active batch to be ready. If the 10th consumer goroutine calls batch sender send, in the past, it should flush active batch immediately, but with this change it will not. Instead, it will wait for up to FlushTimeout for the other group to return and potentially produce a bigger batch.

This change will have implications on throughput, but I reckon that exporting small requests may also be bad for throughput.

**Link to tracking Issue:** Fixes #9952 

Batch sender concurrency limit test is extended to test both merge and merge split. The new tests will most likely fail before this PR.


exporter/exporterhelper/batch_sender.go
```diff
@@ -136,7 +137,7 @@ func (bs *batchSender) resetTimer() {
 // Caller must hold the lock.
 func (bs *batchSender) isActiveBatchReady() bool {
 	return bs.activeBatch.request.ItemsCount() >= bs.cfg.MinSizeItems ||
-		(bs.concurrencyLimit > 0 && bs.activeRequests.Load() >= bs.concurrencyLimit)
+		(bs.concurrencyLimit > 0 && bs.activeBatch.goroutines >= bs.concurrencyLimit)
```
dmitryax, line 140: concurrencyLimit is the limit of incoming requests = number of queue workers. The number of spawn goroutines is unbound. The reason is to avoid blocking the exporter pipeline where queue and batching both enabled until flush timeout

dmitryax:
This approach sounds good to me.

Another option is to find a way to synchronize https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/batch_sender.go#L197 to drop the number of active requests by the number of released goroutines. That would be significantly more complicated, though. If you can try this out, that would be great to compare. Let me know

carsonip:
>  drop the number of active requests by the number of released goroutines. That would be significantly more complicated, though. If you can try this out, that would be great to compare. 

Implemented in alternative PR #10337

Issue #9896, opened 2024-04-05
OTel Collector Builder: strict versioning support


kristinapathak:
**Is your feature request related to a problem? Please describe.**
When building a collector, it's possible to have a mismatch in the core library version and component versions.  This can cause difficult to understand build failures, such as in https://github.com/open-telemetry/opentelemetry-collector/issues/8692.

**Describe the solution you'd like**
When strict versioning is enabled, the builder could check the configured version for the core collector against any included component versions to ensure that they match, as well as ensure that the builder version and collector version match.

**Describe alternatives you've considered**
People modify their builder config until it works 🙁

**Additional context**
@mx-psi [mentioned](https://github.com/open-telemetry/opentelemetry-collector/pull/9631#discussion_r1551557048) making this the default behavior instead of a flag.  Is that something that others also want?

PR #9631, opened 2024-02-23
[builder] Support for --skip-new-go-module

kristinapathak:
A continuation of https://github.com/open-telemetry/opentelemetry-collector/pull/9253

Description: Adds a --skip-new-go-module flag to the OTC builder. This enables users working in an existing go module environment (say, a "monorepo") to update the module they have, vs forcing the use of a new module.

With the new support inside an existing Go module, a collector main package can be generated using a go:generate directive. For example, in the directory where I want my collector built, the file generate.go has this line:

//go:generate builder --skip-new-go-module --skip-compilation --strict-versioning --config=./build-config.yaml
In the same directory, the build-config.yaml describes the collector to build. The builder generates the other files in the same directory. At this point, normal Go workflows can be used to update indirect dependencies. The optional --strict-versioning checks that the build configuration matches that resulting from the enclosing Go module.

Link to tracking Issue: https://github.com/open-telemetry/opentelemetry-collector/issues/9252

Testing: Unit tests added.

Documentation: Yes.

PR #9897, opened 2024-04-05
[builder] Add  strict versioning

kristinapathak:
Adds strict version checking in the builder. This enables users to ensure that the versions specified in the builder config are the versions used in the go.mod when building the collector binary.  This can be disabled with --skip-strict-versioning.

**Link to tracking Issue:** #9896

**Testing:** Added unit tests

**Documentation:** Added to builder README


mx-psi (commented):
Thanks for filing this separately. I think this should be the default behavior, we have had people confused about this in the past (see #8692) and I think it is better to spit out an error than to silently do the wrong thing.

I would therefore prefer to make this PR into a breaking change and add a temporary flag (`--skip-strict-versioning`?) to temporarily allow users to revert this check and allow them to migrate. We can mention this flag in the error message and remove it after a few versions.

Would you mind switching from `--strict-versioning` to `--skip-strict-versioning` and updating the release note?

Thanks!

cc @open-telemetry/collector-approvers for visibility

cmd/builder/internal/builder/main.go
```diff
@@ -139,21 +179,37 @@ func GetModules(cfg Config) error {
...
+	// Perform strict version checking.  For each component listed and the
+	// otelcol core dependency, check that the enclosing go module matches.
+	mf, mvm, err := cfg.readGoModFile()
+	if err != nil {
+		return err
+	}
+
+	coremod, corever := cfg.coreModuleAndVersion()
+	if mvm[coremod] != corever {
+		return fmt.Errorf("core collector version calculated by component dependencies %q does not match configured version %q: %w", mvm[coremod], corever, ErrStrictMode)
```
mx-psi, line 195:
```suggestion
		return fmt.Errorf("core collector version calculated by component dependencies %q does not match configured version %q: %w. Use --skip-strict-versioning to temporarily disable this check. This flag will be removed in a future minor version", mvm[coremod], corever, ErrStrictMode)
```

cmd/builder/internal/builder/main.go
```diff
@@ -139,21 +179,37 @@ func GetModules(cfg Config) error {
...
+		if module == mf.Module.Path {
+			// Main module is not checked, by definition.  This happens
+			// with --skip-new-go-module where the enclosing module
+			// contains the components used in the build.
 			continue
 		}
-		return nil
+
+		if mvm[module] != version {
+			return fmt.Errorf("component %q version calculated by dependencies %q does not match configured version %q: %w", module, mvm[module], version, ErrStrictMode)
```
mx-psi, line 208:
```suggestion
			return fmt.Errorf("component %q version calculated by dependencies %q does not match configured version %q: %w. Use --skip-strict-versioning to temporarily disable this check. This flag will be removed in a future minor version", module, mvm[module], version, ErrStrictMode)
```

.chloggen/builder-strict-versioning.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add strict version checking when using the builder and the flag --skip-strict-versioning for skipping.
+
+# One or more tracking issues or pull requests related to the change
+issues: [9896]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:  Strict version checking will error on mismatches between the otelcol_version configured and the builder version or versions in the go.mod.
```
mx-psi, line 18: Let's add some more info about the flag and the fact that it will be removed

```suggestion
subtext:  Strict version checking will error on mismatches between the `otelcol_version` configured and the builder version or versions in the go.mod. This check can be temporarily disabled by using the `--skip-strict-versioning` flag. This flag will be removed in a future minor version.
```

cmd/builder/README.md
```diff
@@ -145,3 +145,20 @@ then commit the code in a git repo. A CI can sync the code and execute
 ocb --skip-generate --skip-get-modules --config=config.yaml
 ```
 to only execute the compilation step.
+
+### Strict versioning checks
+
+Strict version checking is enabled by default and checks the relevant `go.mod` 
+file for the following things after `go get`ing all components and calling 
+`go mod tidy`:
```
mx-psi, line 153: Since this is here to stay, I would reword this a bit
```suggestion
The builder checks the relevant `go.mod`
file for the following things after `go get`ing all components and calling 
`go mod tidy`:
```

cmd/builder/README.md
```diff
@@ -145,3 +145,20 @@ then commit the code in a git repo. A CI can sync the code and execute
...
+1. The `dist::otelcol_version` field in the build configuration must
+   match the core library version calculated by the Go toolchain,
+   considering all components.  A mismatch could happen, for example,
+   when one of the components depends on a newer release of the core
+   collector library.
+2. For each component in the build configuration, the version included
+   in the `gomod` module specifier must match the one calculated by
+   the Go toolchain, considering all components.  A mismatch could
+   happen, for example, when the enclosing Go module uses a newer
+   release of the core collector library.
```
mx-psi, line 164:
```suggestion
   release of the core collector library.
   
The `--skip-strict-versioning` flag disables these versioning checks. 
This flag is available temporarily and 
**will be removed in a future minor version**.
```

.chloggen/builder-strict-versioning.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: builder
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add strict version checking when using the builder and the flag --skip-strict-versioning for skipping.
```
mx-psi, line 10: I want to make clear the flag is temporary in the release note as well

```suggestion
note: Add strict version checking when using the builder. Add the temporary flag `--skip-strict-versioning `for skipping this check.
```

cmd/builder/internal/builder/main.go
```diff
@@ -4,28 +4,65 @@
...
 	"go.uber.org/zap/zapio"
+
+	"go.opentelemetry.io/collector/cmd/builder/internal/builder/modfile"
 )
 
 var (
 	// ErrGoNotFound is returned when a Go binary hasn't been found
-	ErrGoNotFound = errors.New("go binary not found")
+	ErrGoNotFound       = errors.New("go binary not found")
+	ErrStrictMode       = errors.New("failing due to strict mode")
```
mx-psi, line 27: I would at least be consistent with the docs and release note naming:
```suggestion
	ErrStrictMode       = errors.New("failing due to strict versioning check")
```

but it would be better to have something that does not require you to read the documentation, maybe:

```suggestion
	ErrStrictMode       = errors.New("mismatch in go.mod and builder configuration versions")
```

cmd/builder/test/core.builder.yaml
```diff
@@ -1,20 +1,20 @@
 dist:
   module: go.opentelemetry.io/collector/builder/test/core
-  otelcol_version: 0.94.0
+  otelcol_version: 0.97.0
```
mx-psi, line 3: No need to update this on this PR, we can file a separate PR for updating these versions

examples/k8s/otel-config.yaml
```diff
@@ -65,7 +65,7 @@ spec:
       - command:
           - "/otelcol"
           - "--config=/conf/otel-agent-config.yaml"
-        image: otel/opentelemetry-collector:0.94.0
```
mx-psi, line 68: This and other changes in this file should also be on its own separate PR

cmd/builder/internal/builder/modfile/modfile.go
```diff
@@ -0,0 +1,45 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package modfile // import "go.opentelemetry.io/collector/cmd/builder/internal/builder/modfile"
```
mx-psi, line 4: Why not use https://pkg.go.dev/golang.org/x/mod/modfile ?

cmd/builder/internal/builder/config.go
```diff
@@ -38,6 +44,8 @@ type Config struct {
 	Connectors   []Module     `mapstructure:"connectors"`
 	Replaces     []string     `mapstructure:"replaces"`
 	Excludes     []string     `mapstructure:"excludes"`
+
+	downloadModules retry `mapstructure:"-"`
```
mx-psi, line 48: I am not against this refactor, but we can do it on a separate PR, we don't need to do it here

cmd/builder/internal/builder/main.go
```diff
@@ -4,28 +4,65 @@
...
+func makeGoCommand(cfg Config, args []string) *exec.Cmd {
 	// #nosec G204 -- cfg.Distribution.Go is trusted to be a safe path and the caller is assumed to have carried out necessary input validation
 	cmd := exec.Command(cfg.Distribution.Go, args...)
 	cmd.Dir = cfg.Distribution.OutputPath
+	return cmd
+}
+
+// runGoCommandStdout is similar to `runGoCommand` but returns the
+// standard output.  The subcommand is only printed with Verbose=true,
+// since it is invoked frequently when --skip-new-go-module is set.
```
mx-psi, line 42: I think this comment is not relevant until #9631. I would just change `runGoCommand` for now and make it return the standard output if needed

cmd/builder/internal/builder/main.go
```diff
@@ -139,21 +179,37 @@ func GetModules(cfg Config) error {
...
-	retries := 3
-	failReason := "unknown"
-	for i := 1; i <= retries; i++ {
-		if err := runGoCommand(cfg, "mod", "download"); err != nil {
-			failReason = err.Error()
-			cfg.Logger.Info("Failed modules download", zap.String("retry", fmt.Sprintf("%d/%d", i, retries)))
-			time.Sleep(5 * time.Second)
+	if cfg.SkipStrictVersioning {
+		return downloadModules(cfg)
+	}
```
mx-psi, line 184: I am confused about this, shouldn't this be the opposite, i.e.:

```suggestion
	if !cfg.SkipStrictVersioning {
		return downloadModules(cfg)
	}
```

?

cmd/builder/internal/builder/main.go
```diff
@@ -139,21 +179,37 @@ func GetModules(cfg Config) error {
...
+
+	for _, mod := range cfg.allComponents() {
+		module, version, _ := strings.Cut(mod.GoMod, " ")
+		if module == mf.Module.Path {
+			// Main module is not checked, by definition.  This happens
+			// with --skip-new-go-module where the enclosing module
+			// contains the components used in the build.
 			continue
 		}
```
mx-psi, line 205: Is this only relevant with #9631?

mx-psi (changes requested):
Thanks for the change! I left a few comments :)

cmd/builder/test/core.builder.yaml
kristinapathak: These tests fail without this update due to the strict versioning checks.

cmd/builder/internal/builder/main.go
kristinapathak: the logic in `downloadModules()` was already a part of the `GetModules()` function.  Since we do strict checking before this logic, I put it into its own function that can be called while returning.

cmd/builder/test/core.builder.yaml
mx-psi: Ah, fair enough then :)

cmd/builder/internal/builder/main.go
mx-psi: Ah, right, sorry I got confused but the code is alright :)

cmd/builder/internal/builder/main.go
kristinapathak: I updated the comment here (and variable names).  It shouldn't be possible to hit this if statement without `--skip-new-go-module`, but I would like to keep this if statement here if that's okay. 🙂 If not, I can remove it.

kristinapathak:
@mx-psi, thanks for the review! I think I addressed everything. 🙂 Let me know what you think.

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +140,66 @@ func GetModules(cfg Config) error {
...
+
+	// Perform strict version checking.  For each component listed and the
+	// otelcol core dependency, check that the enclosing go module matches.
+	modulePath, dependencyVersions, err := cfg.readGoModFile()
+	if err != nil {
+		return err
+	}
+
+	corePath, coreVersion := cfg.coreModuleAndVersion()
+	if dependencyVersions[corePath] != coreVersion {
```
mx-psi, line 163: The error here and below would be a bit confusing if `corePath` is not on `dependencyVersions`. While this would be a weird situation that I don't think we can encounter now, IMO for future-proofing we should do the full check  (`corePathVersion, ok := dependencyVersions[corePath]`) and provide a more appropriate error message if `ok` is false

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +140,66 @@ func GetModules(cfg Config) error {
...
+
+	for _, mod := range cfg.allComponents() {
+		module, version, _ := strings.Cut(mod.GoMod, " ")
+		if module == modulePath {
+			// No need to check the version of components that are part of the
+			// module we're building from.
+			continue
+		}
+
+		if dependencyVersions[module] != version {
```
mx-psi, line 177: Ditto here, IMO we should do the explicit check for presence in `dependencyVersions`

cmd/builder/internal/builder/main.go
```diff
@@ -165,3 +211,42 @@ func processAndWrite(cfg Config, tmpl *template.Template, outFile string, tmplPa
...
+func (c *Config) coreModuleAndVersion() (string, string) {
+	module := "go.opentelemetry.io/collector"
+	if c.Distribution.RequireOtelColModule {
+		module += "/otelcol"
+	}
+	return module, "v" + c.Distribution.OtelColVersion
+}
+
+func (c *Config) allComponents() []Module {
+	return append(c.Exporters,
```
mx-psi, line 224:
```suggestion
	// TODO: Use slices.Concat when we drop support for Go 1.21
	return append(c.Exporters,
```

mx-psi (commented):
Code LGTM, I left a couple small suggestions to improve the error messages. Would love to increase the code coverage a bit. If that's too hard on this PR, just let me know, we can postpone to your second PR :)

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +144,74 @@ func GetModules(cfg Config) error {
...
+			// No need to check the version of components that are part of the
+			// module we're building from.
+			continue
+		}
+
+		moduleDepVersion, ok := dependencyVersions[module]
+		if !ok {
+			return fmt.Errorf("component %w: '%s'. %s", ErrDepNotFound, module, skipStrictMsg)
+		}
+		if moduleDepVersion != version {
```
kristinapathak, line 189: It'll be easier to add a test for this in https://github.com/open-telemetry/opentelemetry-collector/pull/9631, so I'd prefer to wait and add it then.

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +144,74 @@ func GetModules(cfg Config) error {
...
+	if err != nil {
+		return err
+	}
+
+	corePath, coreVersion := cfg.coreModuleAndVersion()
+	coreDepVersion, ok := dependencyVersions[corePath]
+	if !ok {
+		return fmt.Errorf("core collector %w: '%s'. %s", ErrDepNotFound, corePath, skipStrictMsg)
+	}
+	if coreDepVersion != coreVersion {
```
kristinapathak, line 171: It'll be easier to add a test for this in https://github.com/open-telemetry/opentelemetry-collector/pull/9631, so I'd prefer to wait and add it then.

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +144,74 @@ func GetModules(cfg Config) error {
...
 
 	// ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules
-	if err := runGoCommand(cfg, "get", "cloud.google.com/go"); err != nil {
-		return fmt.Errorf("failed to go get: %w", err)
+	if _, err := runGoCommand(cfg, "get", "cloud.google.com/go"); err != nil {
+		return fmt.Errorf("%w: %v", errGoGetFailed, err)
 	}
 
-	if err := runGoCommand(cfg, "mod", "tidy", "-compat=1.21"); err != nil {
+	if _, err := runGoCommand(cfg, "mod", "tidy", "-compat=1.21"); err != nil {
```
kristinapathak, line 151: I haven't been able to add coverage for this because if I would expect this to fail, the previous `go get` [command](https://github.com/open-telemetry/opentelemetry-collector/pull/9897/files#diff-117127e4cf9f6587102dc92a7240d1e2c00944ee47ce23e02731f3a41466bdb1R147) fails first.

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +144,74 @@ func GetModules(cfg Config) error {
...
 	}
 
+	if cfg.SkipStrictVersioning {
+		return downloadModules(cfg)
+	}
+
+	// Perform strict version checking.  For each component listed and the
+	// otelcol core dependency, check that the enclosing go module matches.
+	modulePath, dependencyVersions, err := cfg.readGoModFile()
+	if err != nil {
```
kristinapathak, line 162: I haven't been able to add coverage for this because if I would expect this to fail, the previous `go get` [command](https://github.com/open-telemetry/opentelemetry-collector/pull/9897/files#diff-117127e4cf9f6587102dc92a7240d1e2c00944ee47ce23e02731f3a41466bdb1R147) fails first.

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +144,74 @@ func GetModules(cfg Config) error {
...
+
+func downloadModules(cfg Config) error {
 	cfg.Logger.Info("Getting go modules")
 	// basic retry if error from go mod command (in case of transient network error). This could be improved
 	// retry 3 times with 5 second spacing interval
 	retries := 3
 	failReason := "unknown"
 	for i := 1; i <= retries; i++ {
-		if err := runGoCommand(cfg, "mod", "download"); err != nil {
+		if _, err := runGoCommand(cfg, "mod", "download"); err != nil {
```
kristinapathak, line 206: I haven't been able to add coverage for this because if I would expect this to fail, the previous `go get` [command](https://github.com/open-telemetry/opentelemetry-collector/pull/9897/files#diff-117127e4cf9f6587102dc92a7240d1e2c00944ee47ce23e02731f3a41466bdb1R147) fails first.

cmd/builder/internal/builder/main.go
```diff
@@ -165,3 +223,43 @@ func processAndWrite(cfg Config, tmpl *template.Template, outFile string, tmplPa
...
+func (c *Config) allComponents() []Module {
+	// TODO: Use slices.Concat when we drop support for Go 1.21
+	return append(c.Exporters,
+		append(c.Receivers,
+			append(c.Processors,
+				append(c.Extensions,
+					c.Connectors...)...)...)...)
+}
+
+func (c *Config) readGoModFile() (string, map[string]string, error) {
```
kristinapathak, line 244: I haven't been able to add coverage for this function because in the code path, if I would expect this to fail, the previous `go get` [command](https://github.com/open-telemetry/opentelemetry-collector/pull/9897/files#diff-117127e4cf9f6587102dc92a7240d1e2c00944ee47ce23e02731f3a41466bdb1R147) fails first.

I can add a separate test that targets this function directly, otherwise not sure how to test this.

cmd/builder/internal/builder/main.go
```diff
@@ -131,29 +144,74 @@ func GetModules(cfg Config) error {
...
+	for _, mod := range cfg.allComponents() {
+		module, version, _ := strings.Cut(mod.GoMod, " ")
+		if module == modulePath {
+			// No need to check the version of components that are part of the
+			// module we're building from.
+			continue
+		}
+
+		moduleDepVersion, ok := dependencyVersions[module]
+		if !ok {
```
kristinapathak, line 186: It'll be easier to add a test for this in #9631, so I'd prefer to wait and add it then.

kristinapathak:
@mx-psi, let me know your thoughts on my unit test comments. 🙂

mx-psi (approved):
Thanks for taking a thorough look at coverage. We can address the remaining easy cases in the next PR.

This LGTM, but in order to merge, I need you to fix the linter error:
```
 internal/builder/main.go:131:49: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
		return fmt.Errorf("%w: %v", errCompileFailed, err)
```

Issue #9872, opened 2024-04-01
[component] Reject `component.ID`s that are too long


mx-psi:
Follow up to #9208, proposed by @hughesjj at https://github.com/open-telemetry/opentelemetry-collector/pull/9670#discussion_r1514964449.

While this is a breaking change I think we can make it in one go since it's likely to affect nobody.

atoulme:
@hughesjj what would be the max length allowed?

TylerHelmuth:
k8s uses 63, thats good enough for me

ankitpatel96:
I can work on this!

PR #9901, opened 2024-04-05
Add length limit to component

ankitpatel96:

Changes component.Type validation regex to only allow a max of 63 characters in a type name.

**Link to tracking Issue:** [9872](https://github.com/open-telemetry/opentelemetry-collector/issues/9872)

**Testing:** Unit tests


.chloggen/component-id-length-limit.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Changes component.Type validation regex to only allow a max of 63 characters in a type name.
```
mx-psi, line 10: The fact that we use a regex is an implementation detail. I would phrase it in a way that is more meaningful to Go API end-users:
```suggestion
note: Restricts maximum length for `component.Type` to 63 characters.
```

mx-psi (approved):

codeboten (approved):

Issue #9892, opened 2024-04-04
otlp http exporter doesn't handle 429s properly


jaronoff97:
**Describe the bug**
Currently, when the OTLP HTTP exporter's request is responded to with any HTTP error code (429 for example), the status code is dropped which makes back propagation impossible. This is because we create a [new formatted error](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/otlp.go#L186-L196) and return that to the exporter's caller.

**Steps to reproduce**
Setup a collector with any receiver, no processors, and an OTLP http exporter. Have the exporter send to a server that always responds with a 429. Send data to the receiver, the status code will be a 500.

**What did you expect to see?**
I expected the receiver to observe a 429.

**What did you see instead?**
The receiver observed a 500.

**What version did you use?**
v0.97.0

**What config did you use?**
```
exporters:
  otlphttp:
    logs_endpoint: "destination"
    sending_queue:
      enabled: false
    retry_on_failure:
      max_elapsed_time: 20s
  logging:
    verbosity: detailed
receivers:
  webhookevent:
    endpoint: "0.0.0.0:8080"
    include_metadata: true
processors:
service:
  pipelines:
    logs:
      exporters:
      - otlphttp
      processors: []
      receivers:
      - webhookevent
  telemetry:
    metrics:
      level: detailed
    resource:
      service.name: test
```

**Environment**
"Ubuntu 20.04"

**Additional context**

TylerHelmuth:
I believe the correct behavior is for the otlphttpexporter to return a grpc `Status` error that represents what happened.  If the response from the server includes a grpc `Status` error in the body (which it must according to OTLP protocol), that exact `Status` could be used to make the error.

TylerHelmuth:
@jaronoff97 looking at the exporter again, it will only return a permanent error (500) if the response code from the server is not retryable.  If it is retryable it returns a `throttleRetry` error and the retry sender will try again. Eventually the retry sender will give up and return a [basic error](https://github.com/open-telemetry/opentelemetry-collector/blob/2ff9795371c0972cf1c7ffc3b753de07cbfd00db/exporter/exporterhelper/retry_sender.go#L103) that will get treated by the otlpreceiver as a 500.

What happens if you configure your otlp exporter to not retry?

TylerHelmuth:
@dmitryax @evan-bradley what are your thoughts?

jaronoff97:
[...]

PR #9905, opened 2024-04-05
OTLP HTTP Exporter: Propagate HTTP 429s 

jaronoff97:
**Description:** 
Changes otlphttp status code handling to propagate the error code as a grpc status code. This follows the logic that was implemented for the http receiver [here](https://github.com/open-telemetry/opentelemetry-collector/pull/9893/files). 

**Link to tracking Issue:** #9892

**Testing:** local tests were updated, going to test a local build.

**Documentation:** none yet...


exporter/otlphttpexporter/otlp.go
```diff
@@ -389,3 +393,27 @@ func (e *baseExporter) logsPartialSuccessHandler(protoBytes []byte, contentType
 	}
 	return nil
 }
+
+// This is a copy from the OTLP HTTP Receiver, i'm not sure the right place to share this logic...
+func newStatusFromMsgAndHTTPCode(errMsg string, statusCode int) *status.Status {
```
jaronoff97, line 398: not sure where this should go, it's the same function as the otlp receiver but I don't think it makes sense to import the receiver from the exporter... Should this go in consumer errors?

exporter/otlphttpexporter/otlp.go
```diff
@@ -16,13 +16,14 @@ import (
 	"time"
 
 	"go.uber.org/zap"
-	"google.golang.org/genproto/googleapis/rpc/status"
+	spb "google.golang.org/genproto/googleapis/rpc/status"
```
evan-bradley, line 19: Why rename the import here?

evan-bradley (approved):
Overall looks good to me.

exporter/otlphttpexporter/otlp.go
jaronoff97: this follows what the OTLP HTTP Receiver does, I thought it would be good to keep it consistent :)

jaronoff97:
NOTE! before merging this, I'm going to run some tests to be sure this does exactly what we expect! I'll report back

exporter/otlphttpexporter/otlp.go
TylerHelmuth: For now lets undo this change since it is unrelated to the functionality we're trying to change. We can do this in a different PR if necessary.

exporter/otlphttpexporter/otlp.go
jaronoff97: does what I did look good?

exporter/otlphttpexporter/otlp.go
```diff
@@ -184,16 +186,18 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
...
+		errString = fmt.Sprintf(
 			"error exporting items, request to %s responded with HTTP Status Code %d, Message=%s, Details=%v",
 			url, resp.StatusCode, respStatus.Message, respStatus.Details)
 	} else {
-		formattedErr = fmt.Errorf(
+		errString = fmt.Sprintf(
 			"error exporting items, request to %s responded with HTTP Status Code %d",
 			url, resp.StatusCode)
 	}
+	formattedErr = newStatusFromMsgAndHTTPCode(errString, resp.StatusCode).Err()
```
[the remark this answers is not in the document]
jaronoff97: that is what I'm doing here essentially – this .Err comes from the status returned from that method

exporter/otlphttpexporter/otlp.go
TylerHelmuth: Yup

jaronoff97:
Screenshot 2024-04-12 at 4 38 09 PM
Okay I ran this in my dev environment and confirmed this indeed works!

TylerHelmuth (approved):
This looks good to me as an intermediary step towards https://github.com/open-telemetry/opentelemetry-collector/pull/9041

mwear (approved):

exporter/otlphttpexporter/otlp_test.go
```diff
@@ -128,12 +128,12 @@ func TestErrorResponses(t *testing.T) {
 			isPermErr:      true,
 		},
 		{
-			name:           "419",
```
songy23, line 131: Can we keep the test case on 419 and add 429 as a new case?

internal/httphelper/helper.go
```diff
@@ -0,0 +1,36 @@
...
+		c = codes.Unauthenticated
+	case http.StatusForbidden:
+		c = codes.PermissionDenied
+	case http.StatusNotFound:
+		c = codes.Unimplemented
+	case http.StatusTooManyRequests:
+		c = codes.ResourceExhausted
+	case http.StatusBadGateway, http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+		c = codes.Unavailable
+	default:
```
songy23, line 32: I would add 413 -> codes.FailedPrecondition, 408 -> codes.DeadlineExceeded etc. Not suggesting to do it in this PR though.

exporter/otlphttpexporter/otlp_test.go
jaronoff97: I think this was a typo in the previous test, there is no [http status code 419](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors)

internal/httphelper/helper.go
jaronoff97: yeah that should be done in a followup, I want this to be focused on the 429 bug

songy23 (approved):
LGTM

codeboten (approved):

Issue #9921, opened 2024-04-09
[exporter/debug] Debug exporter has sampling enabled by default


andrzej-stencel:
**Describe the bug**

The Debug exporter has sampling enabled by default. I don't think this behavior matches the intended usage of the component.

Whenever I use the Debug exporter, it's to see **all** the information that goes into it. The debug exporter sampling things out is misleading to me.

**What version did you use?**

v0.97.0

**Steps to reproduce**

1. Prepare empty input file `input.log`

```shell
$ rm input.log; touch input.log
```

2. Start the collector (in the background, to reuse the console)

```shell
$ ./otelcol-contrib-0.97.0-linux_amd64 --config ./config.yaml &
2024-04-09T09:49:11.057+0200    info    service@v0.97.0/telemetry.go:97 Serving metrics {"address": ":8888", "level": "Basic"}
2024-04-09T09:49:11.057+0200    info    exporter@v0.97.0/exporter.go:275        Development component. May change in the future.        {"kind": "exporter", "data_type": "logs", "name": "debug"}
2024-04-09T09:49:11.059+0200    info    service@v0.97.0/service.go:143  Starting otelcol-contrib...     {"Version": "0.97.0", "NumCPU": 16}
2024-04-09T09:49:11.059+0200    info    extensions/extensions.go:34     Starting extensions...
2024-04-09T09:49:11.059+0200    info    adapter/receiver.go:45  Starting stanza receiver        {"kind": "receiver", "name": "filelog", "data_type": "logs"}
2024-04-09T09:49:11.059+0200    info    service@v0.97.0/service.go:169  Everything is ready. Begin running and processing data.
2024-04-09T09:49:11.059+0200    warn    localhostgate/featuregate.go:63 The default endpoints for all servers in components will change to use localhost instead of 0.0.0.0 in a future version. Use the feature gate to preview the new default.       {"feature gate ID": "component.UseLocalHostAsDefaultHost"}

$
```

3. Add 30 lines to the `input.log` file, 1 line every 0.1 seconds. This results in 30 lines written to the `input.log` file across 3 seconds, 10 lines every second.

```shell
$ for i in $(seq 1 30); do echo "log" >> input.log; sleep 0.1; done
2024-04-09T09:49:15.939+0200    info    fileconsumer/file.go:260        Started watching file   {"kind": "receiver", "name": "filelog", "data_type": "logs", "component": "fileconsumer", "path": "0408.log"}
[...]

Issue #9372, opened 2024-01-24
[exporter/debug] Make output configurable


astencel-sumo:
It would be awesome if I could (...) choose which metadata I want to see.
Eg. for logs sometimes I might be interested only in logs, sometimes in logs and some metadata and sometimes in metadata only.

_Originally posted by @perk-sumo in https://github.com/open-telemetry/opentelemetry-collector/issues/7806#issuecomment-1571929331_

Things I believe it's worth to configure:

- `output.attributes` whether or not to include records' attributes. They're probably essential for metrics, but for logs - it depends on the nature of the logs. For some users, attributes may be essential; for others, they only obscure the output.
- Whether or not the exporter uses the collector's logger to output data

Other possible configuration options:

- `output.trace_id`, `output.span_id` whether or not to include trace ID, span ID - they are probably essential for traces, but not necessarily for logs.
- `output.timestamp` whether or not to include timestamp - again some users might find them useful, others might want to remove them for clarity

Other configurability option suggestions are welcome.

Sanket-0510:
Hey this is sanket, I would like to express my views over this issue and this CNCF project. 

1. The Issue is the follow up for the issue #7806. Which was regarding the behavior of basic, normal and detailed verbosity levels for the telemetry data such as matrics and logs. The `basic` and the `normal` verbosity level gives same output but the detailed one gives too much verbose output we need too figure out a middle ground for this. 
One way as per proposed by @astencel-sumo  is to make normal level to output in the single like that is to suppress the detailed output. The format for the same is still in the discussion from the design perspective. 

2. while addressing the issue of level and verbosity @perher came with a feature from the user perspective such that it shall be upto user to see what kindda output the person wants to see in form the metrics or the logs data. whether someone wants to see only logs or only metadata, the person shall be able to describe template for the same and shall get the output in that format only. This definitely solves big headache from the user side.


As per suggested by @astencel-sumo  yes we can configure the output for the 
```
1. output.attributes
2. output.trace_id
3. output.span_id
4. output.timestamp
```
we can also configure for the

```
5. output.ip_address
6. output.duration
7. output.message
```


#### This is the what I have understood so far regarding this CNCF mentorship project and the I am looking forward to propose well optimised proposal for the same.

ChristianCiach:
[...]

PR #9922, opened 2024-04-09
[chore][exporter/debug] refactor code to make independent from Logging exporter

astencel-sumo:
This PR is the first part of this draft PR: https://github.com/open-telemetry/opentelemetry-collector/pull/9298.

This refactoring is needed to implement [[exporter/debug] change behavior of "normal" verbosity to be different from "basic" #7806](https://github.com/open-telemetry/opentelemetry-collector/issues/7806). I want to change the behavior of the Debug exporter, but leave the behavior of the Logging exporter unchanged.

**Link to tracking Issue:**

- https://github.com/open-telemetry/opentelemetry-collector/issues/7806


mx-psi (commented):
No strong opinion here, but an alternative solution could be to have a custom unmarshal function/initialization logic for the logging exporter that overrides verbosity from `normal` to set to `basic`. That way we would avoid duplicating code. Any thoughts on which one is better?

astencel-sumo:
> No strong opinion here, but an alternative solution could be to have a custom unmarshal function/initialization logic for the logging exporter that overrides verbosity from `normal` to set to `basic`. That way we would avoid duplicating code. Any thoughts on which one is better?

I feel we might have more changes coming to the Debug exporter that I don't think we want to introduce to the Logging exporter - see e.g. https://github.com/open-telemetry/opentelemetry-collector/issues/9372, https://github.com/open-telemetry/opentelemetry-collector/issues/9921. I would leave the Logging exporter's code unchanged, waiting to be removed in September 2024, and only work with the Debug exporter's code.

mx-psi:
Okay, I am fine with this approach, let me ping @open-telemetry/collector-approvers just in case someone has a different opinion and we can wait a couple days to give people time to voice their opinion

exporter/debugexporter/exporter.go
```diff
@@ -0,0 +1,83 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
mx-psi, line 3: Can we add a comment here and in the original copy of this code that states:
- where the other copy of this codes live
- that, for security issues, we should update both

?

astencel-sumo: I'm not sure it's that important, as these files don't have much logic to begin with, but I've added the following: https://github.com/open-telemetry/opentelemetry-collector/pull/9922/commits/8dce26b26d3d5d8fd5514552747d0c084b5fa504.

mx-psi (approved):
Will merge tomorrow unless somebody else objects

PR #9940, opened 2024-04-10
[cmd/mdatagen] Move component config test from cmd/builder

dmitryax:
The tests generated by cmd/builder are skipped in contrib because they cause the CI timeouts. We moved the lifecycle tests to the tests generated by mdatagen, but the config/factory smoke tests are still part of the files generated by cmd/builder.

Recently, the loadbalancing exporter got an invalid camelCase config field because of this coverage gap.

This change moves the config/factory tests from cmd/builder to cmd/mdatagen. So, they are always generated for every component, even if not used in any collector bundle.


dmitryax:
cc @swiatekm-sumo

codeboten (approved):
Nice 👍🏻

PR #9950, opened 2024-04-12
[cmd/mdatagen] [chore] Small fixes for new resource attributes filter

dmitryax:
- Make sure we always pass a string to the filter.Match even if the attribute value has a different type. Otherwise, it panics.
- Make sure we show the if_configured warning if the user sets include/exclude without enabled.
- Simplify generated tests

Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/9660


dmitryax:
The coverage hasn't changed. I just touched a part that is not covered and probably don't need to be

codeboten (approved):

PR #9951, opened 2024-04-12
make sure ci job uses always the latest go patch available

cpanato:
**Description:** 

- make sure ci job uses always the latest go patch available
- update deprecated goreleaser flag

### DIFF
--- a/.github/workflows/api-compatibility.yml
+++ b/.github/workflows/api-compatibility.yml
@@ -35,7 +35,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
         with:
-          go-version: ~1.21.5
+          go-version: '1.21'
+          check-latest: true
 
       # Generate apidiff states of Main
       - name: Generate-States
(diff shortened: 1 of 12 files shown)


cpanato:
cc @jpkrohling

TylerHelmuth:
If I recall correctly I purposefully switch from using the latest patch version to specifying a specific patch version so that we had complete control over the go version we built/tested with

cpanato:
> purposefully switch from using the latest patch version to spe

I see, but then should we pin that to the patch version and not use the `~`?

cpanato:
@jpkrohling @TylerHelmuth is this something you guys want to have? if yes can you reopen and lets merge it?

cpanato:
i will rebase and fix the conflicts

TylerHelmuth:
Agreed

cpanato:
@jpkrohling @TylerHelmuth done, ptal

thanks!

jpkrohling:
I asked the Collector leads if it's something we want, and if so, I'll fix the conflicts.

cpanato:
rebased

codeboten:
I don't see why we shouldn't do this, although one thing to keep in mind is that CI will be broken when a new version of go is released (because of the helpful toolchain directive)

jpkrohling:
Can't renovate take care of bumping it as well?

cpanato:
renovate or dependabot can bump

mx-psi:
> renovate or dependabot can bump

Is this something we have an issue/PR for?

PR #9959, opened 2024-04-12
[mdatagen] generate goleak package test

codeboten:
This automates the generation of package_test.go for any component that uses mdatagen.

The following configuration can be used to skip or ignore certain funcs:

```yaml
tests:
  goleak:
    skip: true

tests:
  goleak:
    ignore:
      top:
        - "go.opencensus.io/stats/view.(*worker).start"
```


crobert-1:
There's an [open issue in the contrib repository](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30483) for this with some discussion. Might be good to add a reference in the PR description, or copy/move it to core?

cmd/mdatagen/loader.go
```diff
@@ -201,11 +201,17 @@ func (a attribute) TestValue() string {
 	return ""
 }
 
+type goLeak struct {
+	Skip   bool     `mapstructure:"skip"`
+	Ignore []string `mapstructure:"ignore"`
```
crobert-1, line 206: There are a few places in `contrib` that require using `goleak.IgnoreAnyFunction` instead of `goleak.IgnoreTopFunction`. We may need to create some kind of `Ignore` that has the string that needs ignored, and then type of ignore to differentiate between top/any, or other possible future required options.

cmd/mdatagen/main.go
```diff
@@ -66,6 +66,10 @@ func run(ymlPath string) error {
 				filepath.Join(ymlDir, "generated_component_test.go"), md, packageName); err != nil {
 				return err
 			}
+			if err = generateFile(filepath.Join(tmplDir, "package_test.go.tmpl"),
```
crobert-1, line 69: I'm not very familiar with the inner-workings of `mdatagen`, but would it be very hard to modify so that sub-packages also have `goleak` generated?

For example:
```
.//consumer/consumertest/package_test.go
.//consumer/consumererror/package_test.go
.//consumer/package_test.go
```
_It's likely out of the scope of this PR, just raising as something that could be done here, or will need to be done later._

crobert-1 (commented):
Thanks for doing this, automatically generating the tests will be really helpful 👍

dmitryax (approved):

songy23 (approved):

songy23:
Need a rebase

cmd/mdatagen/loader.go
codeboten: I've added an option for the following:

```

tests:
  goleak:
    ignore:
      top:
        - "go.opencensus.io/stats/view.(*worker).start"
      any:
        - "go.opencensus.io/stats/view.(*worker).somethingelse"
```

PR #9960, opened 2024-04-12
[mdatagen] Rename include/exclude config options

dmitryax:
The `include` and `exclude ` options in the resource attributes group sound confusing. It's easy to assume that matching filters will include or exclude resource attributes themselves while they control emitted resource metrics.

The proposal is to change the include/exclude options to `metrics_include`/`metrics_exclude` with detailed comments. These names make it cleaner that matching rules limit the emitted metrics, not resource attributes.

Updates https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/25134

### DIFF
--- a/cmd/mdatagen/internal/samplereceiver/internal/metadata/generated_config.go
+++ b/cmd/mdatagen/internal/samplereceiver/internal/metadata/generated_config.go
@@ -57,9 +57,14 @@ func DefaultMetricsConfig() MetricsConfig {
 
 // ResourceAttributeConfig provides common config for a particular resource attribute.
 type ResourceAttributeConfig struct {
-	Enabled bool            `mapstructure:"enabled"`
-	Include []filter.Config `mapstructure:"include"`
-	Exclude []filter.Config `mapstructure:"exclude"`
+	Enabled bool `mapstructure:"enabled"`
+	// Experimental: MetricsInclude defines a list of filters for attribute values.
+	// If the list is not empty, only metrics with matching resource attribute values will be emitted.
+	MetricsInclude []filter.Config `mapstructure:"metrics_include"`
+	// Experimental: MetricsExclude defines a list of filters for attribute values.
+	// If the list is not empty, metrics with matching resource attribute values will not be emitted.
+	// MetricsInclude has higher priority than MetricsExclude.
+	MetricsExclude []filter.Config `mapstructure:"metrics_exclude"`
 
 	enabledSetByUser bool
 }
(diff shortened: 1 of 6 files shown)


dmitryax:
No changelog entry is needed since this functionality is not released yet

bryan-aguilar (approved):

astencel-sumo:
I'm not sure if this change will make it easier to understand this configuration.

I had trouble understanding the original PR https://github.com/open-telemetry/opentelemetry-collector/pull/9660 and I think it's because the configuration of filtering out metrics via the `resource_attributes` property is confusing.

Perhaps it's good to mark this newly introduced property as experimental to allow us making changes to it after we release it, when/if we find a more idiomatic way of filtering metrics by resource attributes?

dmitryax:
Added the experimental notice. Any suggestions are welcome

astencel-sumo:
Thinking about this again, why not use OTTL to filter (and/or transform) a component's emitted telemetry instead of trying to come up (again?) with a declarative configuration language? Seems that we already had this problem, and we had solved it with OTTL. :slightly_smiling_face: 

Simple things like emitting or not a metric are fine by me to configure with `metrics.<metric-name>.enabled`, but the use case of "discarding all metrics that have a resource attribute that matches an expression" seem to be much better served by a full-fledged expression language. If I'm not mistaken, OTTL was invented exactly for these scenarios.

Similar to how users can attach Stanza operators to e.g. Filelog receiver to postprocess the telemetry before it leaves the component, we could provide a standard, cross-signal way to postprocess telemetry that could be attached to any Otelcol component (definitely to receivers, maybe connectors?).

Of course the same can be done later in the pipeline with the `transform` or `filter` processor. If we still want to allow users to filter/process the telemetry before it leaves the component, perhaps we should let the users use the standard, already familiar (hopefully!) way.

astencel-sumo (approved):

TylerHelmuth (approved):

astencel-sumo:
> @astencel-sumo, it's not a new language. This is a result of [open-telemetry/opentelemetry-collector-contrib#25161](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/25161). OTTL is only applicable to OTel data, not for filtering random slices and maps.

I think we may be mixing two different things here. The one you're talking about is how to expose some (arbitrary) filtering settings in the configuration. I have no problem with this.

The other thing is that in this specific case - filtering out metrics based on their resource attributes - the filtering configuration is used to actually filter telemetry - specifically the metrics that the component emits.

Let me comment on the original issue for this: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/25134#issuecomment-2060919769.

PR #9968, opened 2024-04-15
[service] fix bug in sampler

codeboten:
The configuration for the recordSampler has multiple configurations for the RemoteParentSampler which doesn't appear to make any sense. I suspect the original intent was to configure both local and remote samplers with sampled and not sampled.


mx-psi (approved):
The hypothesis sounds reasonable to me. The original author was @bogdandrutu, but we don't need to wait for them to confirm

PR #9982, opened 2024-04-17
Memoize the result of Resolve to avoid duplicate read operations to the config

ms-hujia:

For the implementation of `ConfigProvider`, the result of `Resolve` is memoized to avoid duplicate read operations to the config. The benefits include:
1. It could optimize the initialization of OTEL Collector. For `Collector`, the existing code will perform another extra read operation to the config right after the first one.
1. It could avoid missing change notification from `Watch()` in some rare corner case. This could happen after the watch is tear down and before new watch is setup in the second invocation to `Resolve()`. 

N/A.

No change to the interface and behavior, so all existing tests should be enough.

N/A.

### DIFF
--- a/otelcol/configprovider.go
+++ b/otelcol/configprovider.go
@@ -65,6 +65,10 @@ type ConfmapProvider interface {
 
 type configProvider struct {
 	mapResolver *confmap.Resolver
+	resolveRet  *struct {
+		conf *confmap.Conf
+		err  error
+	}
 }
 
 var _ ConfigProvider = &configProvider{}
@@ -94,7 +98,7 @@ func NewConfigProvider(set ConfigProviderSettings) (ConfigProvider, error) {
 }
 
 func (cm *configProvider) Get(ctx context.Context, factories Factories) (*Config, error) {
-	conf, err := cm.mapResolver.Resolve(ctx)
+	conf, err := cm.resolve(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("cannot resolve the configuration: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


otelcol/configprovider.go
```diff
@@ -124,14 +128,29 @@ func (cm *configProvider) Shutdown(ctx context.Context) error {
...
 
+func (cm *configProvider) resolve(ctx context.Context) (*confmap.Conf, error) {
+	if cm.resolveRet == nil {
+		conf, err := cm.mapResolver.Resolve(ctx)
+		cm.resolveRet = &struct {
+			conf *confmap.Conf
+			err  error
+		}{
+			conf: conf,
+			err:  err,
```
TylerHelmuth, line 147: Remembering a previous error could have some edge cases when the configuration is provided by a url or env var or config file.  It may have had a problem during the first resolve but not the second.

TylerHelmuth (commented):
I am curious if any performance gains are worth this extra complexity.  Could you provide some benchmarks?

ms-hujia:
> I am curious if any performance gains are worth this extra complexity. Could you provide some benchmarks?

As I tested, the performance difference is very small (<0.05s) in benchmark. Unless I use a giant config file or emulate by a web server with delay, the improvement would be great.

As I explained in description, the point here is mainly avoiding a duplicate read operation in the same function of `collector`.

otelcol/configprovider.go
ms-hujia: Agree, from the perspective of `ConfigProvider`. Also, I noticed that this breaks the logic of reload config in `Collector`.

From the perspective of `ConfigProvider`, error of the first call will fail fast and there won't a second call as there's no retry logic in `Collector`.

So I've two proposals as below:

* Keep memoization. Introduce a flag such as `reload` to the methods of `ConfigProvider` to allow override the memoized results.
* Remove memoization. Combine `GetConfmap` and `Get` as one method of `ConfigProvider`.

 Which change looks better to you?

ms-hujia: After investigation, I found that the duplicate read behavior is introduced by #6833. So, I think changing the behavior of `GetConfmap()` to reuse the result of `Get()` should be the best solution.

Changing the interface of `ConfigProvider` is not a good idea.

ms-hujia:
Ping for review... @TylerHelmuth @mx-psi @atoulme

PR #9983, opened 2024-04-17
[featuregate] add Gate.operatorType for operator register

li-zeyuan:
*Description:** <Describe what has changed.>
bug:  collector launch fail with '--feature-gates=logs.jsonParserArray'

how to fix: 
1. `func init()`always register operator to `operator.DefaultRegistry`.
2. use `Gate.operatorType` double check operator state in `featuregate.globalRegistry` when unmarshal configuration. Because id and operatorType are not the same string. (id: "logs.jsonParserArray", operatorType: "json_array_parser")

relevancy: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32478

**Link to tracking Issue:** 
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32313

### DIFF
--- a/.chloggen/fix-featuregate-for-operator.yaml
+++ b/.chloggen/fix-featuregate-for-operator.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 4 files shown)


djaglowski (changes requested):
The concept of an operator type is specific to contrib's pkg/stanza and not something which should be represented in the featuregate package in any way.

I don't think we need any changes to the core collector in order to solve the bug in contrib.

djaglowski:
Closing this as I believe it is a nonstarter

PR #9985, opened 2024-04-17
Update go.opentelemetry.io/proto/otlp to v1.2.0

dashpole:
**Description:**

Update the OTLP proto dependency and generated code.  I plan to use the metric.metadata field to support additional prometheus types per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-metadata


dashpole:
(this probably doesn't need a changelog entry)

codeboten (approved):

PR #9987, opened 2024-04-17
Remove `GetExporters` from component.Host

TylerHelmuth:
**Description:** 
Remove the deprecated `GetExporters` function from `component.Host`

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/7370


TylerHelmuth:
Depends on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32498

mx-psi (commented):
I think we should
1. Assert somewhere that the service implementation of `component.Host` still has this method.
2. File an issue on the service 1.0 milestone tracker to remove this from the service implementation

9429

TylerHelmuth:
> File an issue on the service 1.0 milestone tracker to remove this from the service implementation

@mx-psi to avoid losing any context, I was planning to move https://github.com/open-telemetry/opentelemetry-collector/issues/7370 from the component milestone to the service milestone once this is merged. I will add a not in that issue for what extra stuff needs done for `service` as part of delaying the work until then. Is that sufficient?

service/host.go
```diff
@@ -14,6 +14,11 @@ import (
 	"go.opentelemetry.io/collector/service/internal/graph"
 )
 
+type getExporters interface {
```
mx-psi, line 17: nit: Link to related issue above?

mx-psi (approved):
Using the same issue works for me :)

evan-bradley (approved):

service/host.go
TylerHelmuth: Added a TODO

Issue #9818, opened 2024-03-22
Please consider a big refactor (including renaming probably) for this package before marking stable.


atoulme:
Please consider a big refactor (including renaming probably) for this package before marking stable.

_Originally posted by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/issues/9805#issuecomment-2015423773_

TylerHelmuth:
If we move this to `consumer` we'd be adding a direct dependency between `confighttp`/`configgrpc` and `consumer`, but I think thats ok.  Considering the functionality within `client` exists to help share context between components I think this makes sense.

When we do the move we can expand some of the naming to include the keyword `Connection`.  So `Info` could become `ConnectionInfo`, `Addr` becomes `ConnectionAddr`, `Auth` becomes `ConnectionAuth`, and `Metadata` becomes `ConnectionMetadata`.  It is kinda verbose, but directly conveys what the struct represents. `NewContext` can  potentially become `NewContextWithConnectionInfo`.

I would prefer not to restrict the `Info` struct to any specific component type.  We don't restrict how components make connections, so I think it is valid for any component to be allowed to add its connection to the context and pass it own down the component pipeline. It should also be valid for any component to ask if there is an `Info` struct in the current `ctx`.

TylerHelmuth:
We took a shot at this via https://github.com/open-telemetry/opentelemetry-collector/pull/9996, but ultimately could not get consensus from approvers and maintainers on accepting the change.  Ultimately, I think @jpkrohling summed up the feelings [here](https://github.com/open-telemetry/opentelemetry-collector/pull/9996#pullrequestreview-2059894167):

> I'm still not convinced we need this: while this change isn't backwards incompatible and there's a clear migration path, I don't think it's worth the hassle. Every authenticator out there (ours and external) will need to eventually migrate to the new package name. Even if the new name is really great, I don't think we should pursue this or similar changes before v1.
>
> About the Metadata vs. AuthData, we could certainly have a separate discussion about unifying them, but I still see them as serving different purposes: metadata is typically just the headers from HTTP or gRPC, while AuthData doesn't necessarily has to be backed by a map of a slice of strings.
>
> As someone else mentioned on yesterday's call: this package has been stable and serving it's purpose for a long time now, we shouldn't really be changing it in the name of stability for v1.

[...]

PR #9996, opened 2024-04-18
[client, consumer] Move client features to consumerconnection

TylerHelmuth:
#### Description

This PR moves `client` features to `consumerconnection`. This attempts to remove ambiguity from the name `client` and continues to centralizes consumer as the location for features that pass data between components.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/9818

#### Testing
Duplicated the unit tests

#### Documentation
Updated the package comments and added deprecated warnings to the old `client` package.

### DIFF
--- a/.chloggen/client-move-to-consumer.yaml
+++ b/.chloggen/client-move-to-consumer.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: collector/client
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecates collector/client.  Use `consumerconnection` instead.
+
+# One or more tracking issues or pull requests related to the change
+issues: [9996]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api]
--- a/client/client.go
+++ b/client/client.go
@@ -87,7 +87,7 @@ type ctxKey struct{}
 
 // Info contains data related to the clients connecting to receivers.
 //
-// Deprecated: [v0.99.0] Use consumerconnection.Info instead
+// Deprecated: [v0.99.0] Use consumerconn.Info instead
 type Info struct {
 	// Addr for the client connecting to this collector. Available in a
 	// best-effort basis, and generally reliable for receivers making use of
@@ -106,7 +106,7 @@ type Info struct {
 // AuthData represents the authentication data as seen by authenticators tied to
 // the receivers.
 //
-// Deprecated: [v0.99.0] Use consumerconnection.AuthData instead
+// Deprecated: [v0.99.0] Use consumerconn.AuthData instead
 type AuthData interface {
 	// GetAttribute returns the value for the given attribute. Authenticator
 	// implementations might define different data types for different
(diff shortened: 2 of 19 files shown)


consumer/consumerconnection/client.go
```diff
@@ -0,0 +1,166 @@
...
+//	exporters:
+//	  debug:
+//	service:
+//	  extensions: [oidc]
+//	  pipelines:
+//	    traces:
+//	      receivers: [otlp]
+//	      processors: [authprinter]
+//	      exporters: [debug]
+package consumerconnection // import "go.opentelemetry.io/collector/consumer/consumerconnection"
```
bogdandrutu, line 77: `consumerconnection` go likes shorter names, `conn` is a good replacement for `connection`. 

Not sure if is not cleaner to have `consumer.ConnInfo` instead? OR `consumerinfo.Conn` and `consumerinfo.AuthData`?

TylerHelmuth: Updated to `consumerconn` for now.  That gives us `consumerconn.Info` and `consumerconn.AuthData` which I like since sharing connection data is the overall theme of the package and `Info` and `AuthData` are pieces of connection data the package supports sharing.

.chloggen/client-move-to-consumer.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: collector/client
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecates collector/client.  Use `consumerconnection` instead.
```
atoulme, line 10:
```suggestion
note: Deprecates collector/client.  Use `consumerconn` instead.
```

consumer/consumerconn/client.go
```diff
@@ -0,0 +1,166 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package consumerconnection contains generic representations of clients connecting to
```
atoulme, line 4:
```suggestion
// Package consumerconn contains generic representations of clients connecting to
```

consumer/consumerconn/client.go
```diff
@@ -0,0 +1,166 @@
...
+//	exporters:
+//	  debug:
+//	service:
+//	  extensions: [oidc]
+//	  pipelines:
+//	    traces:
+//	      receivers: [otlp]
+//	      processors: [authprinter]
+//	      exporters: [debug]
+package consumerconn // import "go.opentelemetry.io/collector/consumer/consumerconnection"
```
atoulme, line 77:
```suggestion
package consumerconn // import "go.opentelemetry.io/collector/consumer/consumerconn"
```

consumer/consumerconn/client_test.go
```diff
@@ -0,0 +1,97 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package client contains generic representations of clients connecting to
```
atoulme, line 4:
```suggestion
// Package consumerconn contains generic representations of clients connecting to
```

client/client.go
```diff
@@ -114,16 +118,21 @@ type AuthData interface {
 	GetAttributeNames() []string
 }
 
+// Deprecated: [v0.99.0] Use consumerconn.MetadataHostName instead
 const MetadataHostName = "Host"
 
 // NewContext takes an existing context and derives a new context with the
 // client.Info value stored on it.
+//
+// Deprecated: [v0.99.0] Use consumerconnection.NewContextWithConnection instead
```
codeboten, line 127:
```suggestion
// Deprecated: [v0.99.0] Use consumerconn.NewContextWithConnection instead
```

jpkrohling (commented):
I'm still not convinced we need this: while this change isn't backwards incompatible and there's a clear migration path, I don't think it's worth the hassle. Every authenticator out there (ours and external) will need to eventually migrate to the new package name. Even if the new name is really great, I don't think we should pursue this or similar changes before v1.

About the Metadata vs. AuthData, we could certainly have a separate discussion about unifying them, but I still see them as serving different purposes: metadata is typically just the headers from HTTP or gRPC, while AuthData doesn't necessarily has to be backed by a map of a slice of strings.

As someone else mentioned on yesterday's call: this package has been stable and serving it's purpose for a long time now, we shouldn't really be changing it in the name of stability for v1.

PR #9997, opened 2024-04-18
[builder] only compare major and minor versions from gomod

kristinapathak:
#### Description
When building from a commit hash, I got this error:
```
Error: mismatch in go.mod and builder configuration versions: core collector version calculated by component dependencies "v0.98.1-0.20240416174005-d0f15e2463f8" does not match configured version "v0.98.0". Use --skip-strict-versioning to temporarily disable this check. This flag will be removed in a future minor version
```
It may be more useful to only compare the major and minor versions

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/issues/9896
Found in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32544

#### Documentation
updated readme

### DIFF
--- a/cmd/builder/internal/builder/main.go
+++ b/cmd/builder/internal/builder/main.go
@@ -16,6 +16,7 @@ import (
 
 	"go.uber.org/zap"
 	"golang.org/x/mod/modfile"
+	"golang.org/x/mod/semver"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/builder/internal/builder/main.go
```diff
@@ -167,7 +168,7 @@ func GetModules(cfg Config) error {
 	if !ok {
 		return fmt.Errorf("core collector %w: '%s'. %s", ErrDepNotFound, corePath, skipStrictMsg)
 	}
-	if coreDepVersion != coreVersion {
+	if semver.MajorMinor(coreDepVersion) != semver.MajorMinor(coreVersion) {
```
evan-bradley, line 171: I think this should be alright, but could we use `semver.Compare` here?

kristinapathak: `semver.Compare` includes the patch version in its comparison, but my goal is to allow patch versions to be different.

kristinapathak:
Talked to @codeboten and I think we still need this 😅

codeboten (approved):

codeboten:
Merging this will unblock updating contrib to the latest core deps.

PR #9999, opened 2024-04-19
[cmd/builder] Remove undocumented strictness check

mx-psi:
#### Description

Partially reverts #9897. This was not documented on the original PR and is IMO too strict. 
We likely want to allow for some skew between versions.

Mentioned on https://github.com/open-telemetry/opentelemetry-collector/pull/9513#issuecomment-2065586307

### DIFF
--- a/Makefile.Common
+++ b/Makefile.Common
@@ -4,7 +4,7 @@ ALL_PKGS := $(sort $(shell go list ./...))
 # COVER_PKGS is the list of packages to include in the coverage
 COVER_PKGS := $(shell go list ./... | tr "\n" ",")
 
-GOTEST_TIMEOUT?=120s
+GOTEST_TIMEOUT?=240s
 GOTEST_OPT?= -race -timeout $(GOTEST_TIMEOUT)
 GOCMD?= go
 GOTEST=$(GOCMD) test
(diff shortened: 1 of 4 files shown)


Makefile.Common
```diff
@@ -4,7 +4,7 @@ ALL_PKGS := $(sort $(shell go list ./...))
 # COVER_PKGS is the list of packages to include in the coverage
 COVER_PKGS := $(shell go list ./... | tr "\n" ",")
 
-GOTEST_TIMEOUT?=120s
+GOTEST_TIMEOUT?=240s
```
evan-bradley, line 7: If we're seeing timeouts in all our tests, we can probably remove the manual increase in the Windows unit tests workflow, which I just bumped to 240s as well.

evan-bradley (approved):

Makefile.Common
mx-psi: Yeah I was testing if the timeout would help (it did), so I made your change from #9992 apply to every test suite

kristinapathak (approved):

codeboten (commented):
Thanks @mx-psi, please resolve the conflict

Makefile.Common
codeboten: It's pretty small, but it would be better if this change was not in this PR :)

codeboten (approved):

PR #10000, opened 2024-04-19
[cmd/builder] Improve TestVersioning

mx-psi:
#### Description

Improves `TestVersioning` added on #9897.

The configurations were not what one would usually find, since when running the builder we do some validation and setup.
This makes it closer to the real case.

In the process I removed the 'invalid' cases, since these error out today already

### DIFF
--- a/cmd/builder/internal/builder/main_test.go
+++ b/cmd/builder/internal/builder/main_test.go
@@ -106,50 +106,18 @@ func TestVersioning(t *testing.T) {
 			expectedErr: nil,
 		},
 		{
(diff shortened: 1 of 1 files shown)


songy23 (approved):

evan-bradley (approved):
Looks good to me, I made some similar changes in #9513. Congrats on claiming #10000.

mx-psi:
> Looks good to me, I made some similar changes in #9513. Congrats on claiming #10000.

Thank you! :smile: :rocket: :tada:

PR #10001, opened 2024-04-19
[service/telemetry] Switch to a factory pattern

mx-psi:
#### Description

Switches `service/telemetry` to a factory pattern. To avoid adding a lot of public API in one go:
1. the actual factory builder is in an internal package
2. I have not added the `CreateMeterProvider` method yet

There are two goals with this: one is to make progress on #4970, the other is to allow initializing telemetry sooner: 

#### Link to tracking issue
Updates #4970. 

#### Testing

Updates existing tests to use `NewFactory`


atoulme (approved):
LGTM. Feels like a nice clean up.

otelcol/unmarshaler.go
```diff
@@ -32,6 +27,10 @@ type configSettings struct {
 // unmarshal the configSettings from a confmap.Conf.
 // After the config is unmarshalled, `Validate()` must be called to validate.
 func unmarshal(v *confmap.Conf, factories Factories) (*configSettings, error) {
+
+	telFactory := telemetry.NewFactory()
```
TylerHelmuth, line 31: I recognize this works, but it feels kinda arbitrary to create a new factory here.  Would it be better to create a new telemetry factory at the same level we create the `Factories` and then pass it around?

service/service.go
```diff
@@ -89,16 +89,27 @@ func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to get logger: %w", err)
-	}
 
 	// Fetch data for internal telemetry like instance id and sdk version to provide for internal telemetry.
 	res := resource.New(set.BuildInfo, cfg.Telemetry.Resource)
 	pcommonRes := pdataFromSdk(res)
 
-	logger := tel.Logger()
+	telFactory := telemetry.NewFactory()
```
TylerHelmuth, line 97: To highlight the point above, I don't like making 2 calls to `NewFactory`

otelcol/unmarshaler.go
mx-psi: I want to do do this indeed, but I want to keep public API changes in service and friends for another PR. Would that be okay with you? Or do you want to do it all in one go?

TylerHelmuth: Ya thats ok

TylerHelmuth (approved):

evan-bradley (approved):
I like this approach. It addresses https://github.com/open-telemetry/opentelemetry-collector/issues/4970 in a way that feels consistent with the rest of our APIs.

service/telemetry/internal/factory.go
```diff
@@ -0,0 +1,116 @@
...
+	AsyncErrorChannel chan error
+	ZapOptions        []zap.Option
+}
+
+// Factory is factory interface for telemetry.
+// This interface cannot be directly implemented. Implementations must
+// use the NewFactory to implement it.
+type Factory interface {
+	// CreateDefaultConfig creates the default configuration for the telemetry.
+	// TODO: Should we just inherit from component.Factory?
```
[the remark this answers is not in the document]
TylerHelmuth: I think I'd prefer not to depend on anything from `component` in here, since this struct this factory builds is not a component.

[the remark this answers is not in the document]
evan-bradley: Agreed on not having a dependency on `component`, I think we'd want to consider what commonalities the two have and move all of those things into a shared package.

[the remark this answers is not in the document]
mx-psi: > Will we want to utilize multiple telemetry factories within a service?

I feel like the answer to that is no (but API users are still free to do that by building their own factory that glues the two together). The type could be useful for having a consistent way to refer to what telemetry factory you used in logs/telemetry.

> Do we think we may want to offer alternate factories and let users choose between them in the service config?

Probably also no? You should be able to eventually configure this through the builder IMO, but after that the `service::telemetry` config section should only support the config from your telemetry factory.

mx-psi:
I believe all remaining discussions are things I can solve while I continue working on #4970, so I will merge this right after we release v0.101.0

PR #7578, opened 2023-04-26
Add batchprocessor support for client metadata

jmacd:
**Description:** 

Re-do of https://github.com/open-telemetry/opentelemetry-collector/pull/7325

**Link to tracking Issue:** #4544 

**Testing:** One new test.

**Documentation:** README updated.

PR #10002, opened 2024-04-19
feat(processor): preserve auth in batch processor

grandwizard28:
#### Description
This PR preserves the authData in the client context after the batch processor. The idea is that the auth data can be used after the batch processor in subsequent processors and exporters.

#### Testing

- Current unit tests modified to include the context
- Need help with adding unit tests which choose auth is preserved.

### DIFF
--- a/processor/batchprocessor/batch_processor.go
+++ b/processor/batchprocessor/batch_processor.go
@@ -110,7 +110,7 @@ var _ consumer.Metrics = (*batchProcessor)(nil)
 var _ consumer.Logs = (*batchProcessor)(nil)
 
 // newBatchProcessor returns a new batch processor component.
-func newBatchProcessor(set processor.CreateSettings, cfg *Config, batchFunc func() batch) (*batchProcessor, error) {
+func newBatchProcessor(ctx context.Context, set processor.CreateSettings, cfg *Config, batchFunc func() batch) (*batchProcessor, error) {
 	// use lower-case, to be consistent with http/2 headers.
 	mks := make([]string, len(cfg.MetadataKeys))
 	for i, k := range cfg.MetadataKeys {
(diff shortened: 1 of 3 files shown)


mx-psi:
@grandwizard28 Thanks for the PR! Are there any related issues to this?

grandwizard28:
Hey @jpkrohling @mx-psi,
Any thoughts here?

processor/batchprocessor/batch_processor.go
```diff
@@ -129,7 +129,9 @@ func newBatchProcessor(set processor.CreateSettings, cfg *Config, batchFunc func
 		metadataLimit:    int(cfg.MetadataCardinalityLimit),
 	}
 	if len(bp.metadataKeys) == 0 {
-		s := bp.newShard(nil)
+		// Retrieve info from the context and preserve Auth
+		info := client.FromContext(ctx)
```
jpkrohling, line 133: The context at this level doesn't have auth data, as `newBatchProcessor` isn't called in response to incoming data, but during the startup of the collector.

jpkrohling (changes requested):
I'm not sure this is accurate. I believe this feature exists already, and was implemented by @jmacd on #7578. If that's not working for some reason, I'd be interested in hearing about that in an issue.

processor/batchprocessor/batch_processor.go
grandwizard28: Understood! The idea is to propogate the existing `context` in all shards created.

The `newShard` function is the same regardless of whether we want to start at startup with a single shard or start multiple shards as per the metadata. I figured we just need to amend the signature of `newShard` to give it an existing (if present) AuthData.

Is my understanding fair?

grandwizard28:
If you look at the `exportCtx` @jpkrohling, it looks like the following:
```
exportCtx := client.NewContext(context.Background(), client.Info{
		Metadata: client.NewMetadata(md),
	})
```

There is no AuthData preservation happening here at all. We are creating a new context altogether.

Do you want me to still create an issue?

jpkrohling:
No, the code you are looking for is here: 

https://github.com/open-telemetry/opentelemetry-collector/blob/7855bf2ababecdddfea82e4b1a15fd81be2d2d66/processor/batchprocessor/batch_processor.go#L326

This will create a new shard based on the context for that is available in the consume function, which does have client metadata.

grandwizard28:
Hey @jpkrohling,
Created an issue for this: https://github.com/open-telemetry/opentelemetry-collector/issues/10093

PR #10007, opened 2024-04-19
[otelcol] Pass a real logger to ProviderSettings

TylerHelmuth:
#### Description

This PR creates a un-configurable logger to be used in `confmap.ProviderSettings` so that logging can occur while configuration is resolved. It configures the zap logger using reasonable default values since this logger will only be used while resolving config.

Example of the logger in action

```
receivers:
  nop:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:4317
    headers:
      # Not set
      x-test: ${env:TEMP3}
  debug:
    # set to "detailed"
    verbosity: $TEMP

service:
  pipelines:
    traces:
      receivers:
        - nop
      exporters:
        - debug
```

image

#### Link to tracking issue

#### Testing

### DIFF
--- a/otelcol/collector.go
+++ b/otelcol/collector.go
@@ -15,6 +15,7 @@ import (
 
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/confmap"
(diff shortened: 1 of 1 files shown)


evan-bradley:
@TylerHelmuth What do you think about the approach here: https://github.com/open-telemetry/opentelemetry-collector/pull/9908?

I believe this would allow us to configure the logger by storing the logs until they can be printed by an instantiated logger.

TylerHelmuth:
Maybe thats worth it 🤷 I figured this is such a niche need that the extra complexity to pass the logs around wasn't worth it.

If being able to configure the logger that prints confmap logs is important, I can look into the technique used in https://github.com/open-telemetry/opentelemetry-collector/pull/9908, although I'd prefer not to expose the logs in the API.

TylerHelmuth:
Example of the "wait for the configured logger" approach: https://github.com/open-telemetry/opentelemetry-collector/pull/10007

PR #10029, opened 2024-04-24
Documentation improvements - Comments in key functions

ankitpatel96:
#### Documentation
I wrote comments on a bunch of important functions that helped me understand how the collector works.
I also created some other documentation in https://github.com/open-telemetry/opentelemetry-collector/pull/10068 - but split it up from this PR.


docs/startup.md
```diff
@@ -0,0 +1,32 @@
...
+flowchart TD
+    A("`**command.NewCommand**`") -->|1| B("`**UpdateSettingsUsingFlags**`")
+    A --> |2| C("`**NewCollector**
+    Creates and returns a new instance of Collector`")
+    A --> |3| D("`**Collector.Run**
+    Starts the collector and waits for its completion.  Also includes the control logic for config reloading and shutdown`")
+    D --> E("`**SetupConfigurationComponents**`")
+    E -->  |1| F("`**getConfMap**`")
+    E ---> |2| G("`**Service.New**
+     Initializes telemetry and logging, then initializes the pipelines`")
```
mx-psi, line 12: Logging is part of telemetry :)
```suggestion
     Initializes telemetry, then initializes the pipelines`")
```

docs/startup.md
```diff
@@ -0,0 +1,32 @@
+todo: build out config parts (getconfmap).  document how settings and config get rendered
+```mermaid
+flowchart TD
+    A("`**command.NewCommand**`") -->|1| B("`**UpdateSettingsUsingFlags**`")
+    A --> |2| C("`**NewCollector**
+    Creates and returns a new instance of Collector`")
+    A --> |3| D("`**Collector.Run**
+    Starts the collector and waits for its completion.  Also includes the control logic for config reloading and shutdown`")
```
mx-psi, line 8: I find 'waits for its completion' a bit hard to understand. I would also remove the config reloading bit (I don't find it particularly important to call out when compared to any other part of the Collector's behavior). Maybe just this?
```suggestion
    Starts the collector and blocks until it shuts down`")
```

docs/startup.md
```diff
@@ -0,0 +1,32 @@
...
+    E ---> |2| G("`**Service.New**
+     Initializes telemetry and logging, then initializes the pipelines`")
+    E --> |3| Q("`**Service.Start**
+    1. Start all extensions.
+    2. Notify extensions about Collector configuration
+    3. Start all pipelines.
+    4. Notify extensions that the pipeline is ready.
+    `")
+    Q --> R("`**Graph.StartAll**
+    calls Start on each component in reverse topological order`")
```
mx-psi, line 20:
```suggestion
    Calls Start on each component in reverse topological order`")
```

service/service.go
```diff
@@ -72,6 +72,10 @@ type Service struct {
 	collectorConf     *confmap.Conf
 }
 
+// New has a few responsibilities:
+// 1.  initializes the Service function given a complete set of Settings
+// 2.  sets up internal facing telemetry
+// 3.  builds the graph from the settings
```
mx-psi, line 78: This will be part of the public documentation, so we need to be careful about what we commit to here. I would just say something like this:

```suggestion
// New creates a new Service, its telemetry and component graph.
```

In particular, I don't think we should talk about 'the graph': maybe in the future we have multiple graphs or no graph at all

mx-psi (changes requested):
Thanks! I think all the information here is super helpful for newcomers. I left a ton of comments, let's go through them first and we can discuss if we split this PR in one for comments and another one for docs later

docs/startup.md
```diff
@@ -0,0 +1,32 @@
+todo: build out config parts (getconfmap).  document how settings and config get rendered
+```mermaid
```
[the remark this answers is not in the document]
ankitpatel96: I definitely think we need to still include them in some form.  I totally agree with your concerns about maintainability - but these top level private API bits are very important for how the program actually works.  Without the names the diagram can't serve as a guide to reading the source code at all - I guess its the difference between documentation designed for a user of the Collector API and someone who wants to start contributing.

docs/startup.md
ankitpatel96: I'm ok cutting it down to this - but I'll add the detail of the control logic to the Collector.Run function comment.

service/service.go
ankitpatel96: Hmmm I agree that as a function level comment perhaps this is an implementation detail.  I'm ok removing it because there are comments within the functions / inside the functions that get called that explain some of these details.

service/internal/graph/graph.go
```diff
@@ -73,18 +75,22 @@ func Build(ctx context.Context, set Settings) (*Graph, error) {
...
+	// Keep track of connectors and where they are used. (map[connectorID][]pipelineID).
 	connectorsAsExporter := make(map[component.ID][]component.ID)
 	connectorsAsReceiver := make(map[component.ID][]component.ID)
 
+	// Build each pipelineNodes struct for each pipeline by parsing the pipelineCfg.
+	// Also populates the connectors, connectorsAsExporter and connectorsAsReceiver maps.
 	for pipelineID, pipelineCfg := range set.PipelineConfigs {
 		pipe := g.pipelines[pipelineID]
 		for _, recvID := range pipelineCfg.Receivers {
+			// checks if this receiver is a connector or a regular receiver
```
ankitpatel96, line 93: capitalize

docs/key-files.md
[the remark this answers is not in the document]
mx-psi: I guess we should continue discussing over at #10068, but I will reply here to this comment just so that the convo is in one place.

----


> What are the other examples of docs like this? I saw confmap had an excellent README - are there others?

Some other README examples:
- [cmd/builder](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/builder/README.md)
- [configgrpc](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/README.md)
- [httpprovider](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider/httpprovider/README.md)
- [connector](https://github.com/open-telemetry/opentelemetry-collector/blob/main/connector/README.md)
- [featuregate](https://github.com/open-telemetry/opentelemetry-collector/blob/main/featuregate/README.md)

> I also don't know if I agree with moving it to the otelcol/ directory - I think we should keep this as a separate file that we link to to from CONTRIBUTING.md. Some of the open source projects that I've taken some inspiration from in this section are Redis and sqllite - both of which are known as well documented code bases. Those two projects have a similar section in their main README - I'm not sure if we would want to do that but I do think it should be more prominently located.

I am fine with `internal-architecture.md`, let's go with that! We used to have `architecture.md` for https://opentelemetry.io/docs/collector/architecture/, this one can be developer-focused and have implementation details.

> I would also love some input as to what other files you think fit into this theme - I'm pretty sure this isnt a great list and I'd love to add a few more sections (or someone could of course contribute those sections)

I think your list is a good start when it comes to the general structure of the Collector.

bogdandrutu (approved):

codeboten (approved):

Issue #10014, opened 2024-04-22
builder test failure is blocking release


codeboten:
See https://github.com/open-telemetry/opentelemetry-collector/actions/runs/8786975189/job/24111641909?pr=10012 for full log:

```
--- FAIL: TestVersioning (21.43s)
    --- FAIL: TestVersioning/defaults (0.89s)
        main_test.go:166: 
            	Error Trace:	/home/runner/actions-runner/_work/opentelemetry-collector/opentelemetry-collector/cmd/builder/internal/builder/main_test.go:166
            	Error:      	Target error should be in err chain:
            	            	expected: ""
            	            	in chain: "failed to go get: go subcommand failed with args '[get cloud.google.com/go]': exit status 1, error message: go: go.opentelemetry.io/collector/confmap/converter/expandconverter@v0.99.0: reading go.opentelemetry.io/collector/confmap/converter/expandconverter/go.mod at revision confmap/converter/expandconverter/v0.99.0: unknown revision confmap/converter/expandconverter/v0.99.0\n"
            	            		"failed to go get"
            	Test:       	TestVersioning/defaults
    --- FAIL: TestVersioning/require_otelcol (0.57s)
        main_test.go:166: 
            	Error Trace:	/home/runner/actions-runner/_work/opentelemetry-collector/opentelemetry-collector/cmd/builder/internal/builder/main_test.go:166
            	Error:      	Target error should be in err chain:
            	            	expected: ""
            	            	in chain: "failed to go get: go subcommand failed with args '[get cloud.google.com/go]': exit status 1, error message: go: go.opentelemetry.io/collector/confmap/converter/expandconverter@v0.99.0: reading go.opentelemetry.io/collector/confmap/converter/expandconverter/go.mod at revision confmap/converter/expandconverter/v0.99.0: unknown revision confmap/converter/expandconverter/v0.99.0\n"
            	            		"failed to go get"
            	Test:       	TestVersioning/require_otelcol
    --- FAIL: TestVersioning/old_component_version (0.28s)
        main_test.go:166: 
            	Error Trace:	/home/runner/actions-runner/_work/opentelemetry-collector/opentelemetry-collector/cmd/builder/internal/builder/main_test.go:166
            	Error:      	Target error should be in err chain:
            	            	expected: ""
[...]

PR #10030, opened 2024-04-24
[chore][cmd/builder] Test for unreleased versions

evan-bradley:
#### Description

Add a test to help prevent our replace statements from going out of date and causing situations like https://github.com/open-telemetry/opentelemetry-collector/issues/10014. This is also probably just a decent test case to have since it's a syntactically valid but nonexistent version.


jpkrohling (commented):
I like this PR, and I think it's a valid test, but can you describe a situation where this would fail? I saw the linked PRs, and it's still not clear to me what are the error conditions. Would this test fail in a certain revision of this repo?

evan-bradley:
>Would this test fail in a certain revision of this repo?

This test should only fail if our replace statements in the tests go out of date. It protects against failures in other tests that would occur from bumping the revision at a particular point in the release process with certain changes.

Here's the scenario:
1. We are on version X, and make changes to the builder to include new modules that only goes into effect once version X+1 is released. This is necessary for when we make Collector API changes and want to maintain backwards compatibility with versions <X+1. We gate the changes by checking the version of the Collector APIs used within the builder code.
2. The release process kicks off, and everything is bumped to version X+1, so the builder code now includes the new modules. However, these modules are not publicly available, so attempts to grab them with `go get` in the builder tests will fail unless we have the necessary `replace` statements in place.

The goal of this test is to ensure that the replace statements are up-to-date by forcing this failure scenario. If any modules are missing, the test will fail because Go will attempt to grab the nonexistent versions. The statements can then be updated during the PR for the builder changes instead of in the middle of a release.

jpkrohling:
> The goal of this test is to ensure that the replace statements are up-to-date by forcing this failure scenario

Sounds good, but let me phrase the previous question differently: are we confident that this test will indeed fail on that scenario?

evan-bradley:
Sorry, I had meant to rewrite the test yesterday but was not able to get around to it. The prior version of the test would catch most cases, but not all, so thanks for keeping us on track. I've updated the test to directly check the go.mod file for modules that are missing from our replace statements, so I feel confident now that we don't have any gaps so long as the whole module chain is represented within the go.mod file.

Unfortunately it is still possible to run into errors if the replace statements aren't used in a test, but this at least ensures the list is up to date.

cmd/builder/internal/builder/main_test.go
```diff
@@ -272,6 +276,101 @@ func TestGenerateAndCompile(t *testing.T) {
...
+		}
+
+		_, ok := replaceMods[req.Mod.Path]
+		require.True(t, ok, fmt.Sprintf("Module missing from replace statements list: %s", req.Mod.Path))
+
+		replaceMods[req.Mod.Path] = true
+	}
+
+	for k, v := range replaceMods {
+		require.True(t, v, fmt.Sprintf("Module not used: %s", k))
```
jpkrohling, line 370: This should technically be assert instead of require, but it's OK to leave it like this. Also, you can use Truef, which accepts message and args like this: 

```suggestion
		require.Truef(t, v, "Module not used: %s", k)
```

jpkrohling (approved):
I was confused about this test, until I realized that this is testing the release process for Core/Contrib, while the builder is used beyond those two distributions. I see that we have a list of replaces statements here already, so, the tests are already "biased" towards those two use cases. I'm fine with this change as it is, but I wonder if this an indication that we need better testing for the releases repo itself, so that downstream distributions can mirror what we'd be doing there?

cmd/builder/internal/builder/main_test.go
evan-bradley: Thanks for the tip on `Truef`, I knew there was something that did that but couldn't remember how it worked.

>This should technically be assert instead of require, but it's OK to leave it like this.
I would actually agree with this for most of our test case assertions, but in this case I think `require` makes more sense. Not stopping test execution after the assertion fails would notify us of all extra modules instead of just the first one that's missing. What do you think?

evan-bradley:
>I was confused about this test, until I realized that this is testing the release process for Core/Contrib, while the builder is used beyond those two distributions.

This is specifically to aid in releasing the builder. It's only required because the builder is released at the same time all Core modules are released, if we released the builder after Core/Contrib we wouldn't need this.

We need to check that the `go.mod` files that other tests in `main_test.go` may generate don't use any modules that don't have associated replace statements, to prevent the tests from failing because they want a version that isn't yet available. We will usually only hit these failures during a release process, and that would be tedious to manually test for, so this test should automate that by checking we didn't miss anything.

I'll revise the comment on the test to be a little more concise. Unfortunately this is a strange edge case that we hopefully won't hit often, so it's difficult to quickly explain.

jpkrohling (approved):

cmd/builder/internal/builder/main_test.go
```diff
@@ -289,52 +395,62 @@ func makeModule(dir string, fileContents []byte) error {
 	return nil
 }
 
+func replaceModules() []string {
```
bogdandrutu, line 398: Why func vs a regular variable/constant?

cmd/builder/internal/builder/main_test.go
```diff
@@ -31,6 +34,7 @@ require (
 	go.opentelemetry.io/collector/receiver v0.94.1
 	go.opentelemetry.io/collector v0.96.0
 )`)
+	modulePrefix = "go.opentelemetry.io/collector"
```
bogdandrutu, line 37: const?

bogdandrutu (approved):

cmd/builder/internal/builder/main_test.go
evan-bradley: String arrays can't be constants in Go, so I thought making it a function would prevent mutation. A variable is also fine though since this is just a test, I've updated it accordingly.

cmd/builder/internal/builder/main_test.go
evan-bradley: Good call, updated.

mx-psi (approved):

PR #10037, opened 2024-04-26
[mdatagen] move telemetry into its own file

codeboten:
This is in preparation for using mdatagen for component telemetry.


TylerHelmuth (approved):

songy23 (approved):

evan-bradley (approved):
Thanks for adding the generated tests. Those have been a bit of a sore point when adding new components.

PR #10040, opened 2024-04-29
[cmd/builder] Disable cgo by default

### LINKED ISSUE #10028: [cmd/builder] Disable CGO by default
When using a binary output by the builder with no options, I can't use it by default in a scratch image unless it is compiled with `CGO_ENALBED=0`.

Turn off CGO by default in the builder. Allow users to enable it through setting the `CGO_ENABLED` environment variable.

**Additional context**
Our contributing guidelines [forbid using CGO](https://github.com/open-telemetry/opentelemetry-collector/blob/326ef7c28a1e3e500676c23448900cecfd84e6f6/CONTRIBUTING.md#using-cgo), and no upstream components I'm aware of use it. We also have `CGO_ENABLED=0` set anywhere we build Collector binaries in upstream repositories.

Users have also run into this before: https://github.com/open-telemetry/opentelemetry-collector/issues/6373

evan-bradley:
#### Description

Disables cgo by default in the builder. I'm proposing this as a breaking change because I doubt it will affect many users, most of whom I expect to be using `CGO_ENABLED=0` already, or at a minimum prefer cgo to be disabled. If we would like to have a transition period for this, I can adjust the PR and plan for that.

#### Link to tracking issue
Fixes #10028 

#### Testing

I can't find a good way to directly test this that doesn't involve either setting up a lot of scaffolding to either create a mock `go` cli or injects the necessary code into a top-level function. To keep things simple I have just relied on the existing test suite. I had thought of trying to test the output binary, but Go disables cgo if no compiler is detected on the system, so the test wouldn't be reliable across developer machines.

In the future, if we emit a Dockerfile from the builder, we can test that leaving the option as a default runs inside a `scratch` image.

#### Documentation

Updated the builder readme.

### DIFF
--- a/cmd/builder/internal/builder/main.go
+++ b/cmd/builder/internal/builder/main.go
@@ -29,7 +29,26 @@ var (
 	skipStrictMsg      = "Use --skip-strict-versioning to temporarily disable this check. This flag will be removed in a future minor version"
 )
 
-func runGoCommand(cfg Config, args ...string) ([]byte, error) {
+func getGoEnv(cfg Config) ([]string, error) {
+	// #nosec G204 -- cfg.Distribution.Go is trusted to be a safe path and the caller is assumed to have carried out necessary input validation
+	cmd := exec.Command(cfg.Distribution.Go, "env")
+	cmd.Dir = cfg.Distribution.OutputPath
+
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("go env failed: %w, error message: %s", err, stderr.String())
+	}
+
+	str := stdout.String()
+	str = strings.ReplaceAll(str, "'", "")
+
+	return strings.Split(str, "\n"), nil
+}
+
+func runGoCommand(cfg Config, env []string, args ...string) ([]byte, error) {
 	if cfg.Verbose {
 		cfg.Logger.Info("Running go subcommand.", zap.Any("arguments", args))
 	}
@@ -37,6 +56,7 @@ func runGoCommand(cfg Config, args ...string) ([]byte, error) {
 	// #nosec G204 -- cfg.Distribution.Go is trusted to be a safe path and the caller is assumed to have carried out necessary input validation
 	cmd := exec.Command(cfg.Distribution.Go, args...)
 	cmd.Dir = cfg.Distribution.OutputPath
+	cmd.Env = append(cmd.Env, env...)
 
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
@@ -110,6 +130,16 @@ func Compile(cfg Config) error {
 
 	cfg.Logger.Info("Compiling")
 
+	env := []string{}
+	if !cfg.Distribution.CgoEnabled {
+		initialEnv, err := getGoEnv(cfg)
+		if err != nil {
+			return err
+		}
+		env = append(env, initialEnv...)
+		env = append(env, "CGO_ENABLED=0")
+	}
+
 	var ldflags = "-s -w"
 
 	args := []string{"build", "-trimpath", "-o", cfg.Distribution.Name}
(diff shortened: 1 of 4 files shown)


cmd/builder/internal/builder/main.go
```diff
@@ -110,6 +130,16 @@ func Compile(cfg Config) error {
 
 	cfg.Logger.Info("Compiling")
 
+	env := []string{}
+	if !cfg.Distribution.CgoEnabled {
+		initialEnv, err := getGoEnv(cfg)
+		if err != nil {
+			return err
+		}
+		env = append(env, initialEnv...)
```
evan-bradley, line 139: I'm not a huge fan of having this here, but when setting `CGO_ENABLED=0` without it, I get this locally:
```
exit status 1, error message: warning: GOPATH set to GOROOT () has no effect
go: module cache not found: neither GOMODCACHE nor GOPATH is set
```

Looking at the output of `go env` when the env var is set vs. when it isn't, the output is different. So I just got the output when the env var is unset and copy it into the variables for when we set it. This feels off, but at the same time if the `go` cli is changing other variables in unexpected ways, perhaps it's safer for us to explicitly set these.

cmd/builder/internal/builder/main.go
```diff
@@ -29,14 +29,34 @@ var (
...
+		return nil, fmt.Errorf("go env failed: %w, error message: %s", err, stderr.String())
+	}
+
+	str := stdout.String()
+	str = strings.ReplaceAll(str, "'", "")
+
+	return strings.Split(str, "\n"), nil
+}
+
+func runGoCommand(cfg Config, env []string, args ...string) ([]byte, error) {
```
evan-bradley, line 51: I'm not a huge fan of the new function signature considering `env` won't often be used. I think if we add an option to pass env vars to the go build command, we may be able to just read from `cfg` and eliminate this.

evan-bradley:
>```
failed to compile the OpenTelemetry Collector distribution: go subcommand failed with args '[build -trimpath -o  -ldflags=-s -w]': exit status 1, error message: warning: GOPATH set to GOROOT () has no effect
            	            	go: module cache not found: neither GOMODCACHE nor GOPATH is set
            	            	go: module cache not found: neither GOMODCACHE nor GOPATH is set
```

It looks like my workaround didn't work on Windows. If anyone has any ideas, they would be welcome, otherwise I'll spend some time later to dig into this.

File: cmd/builder/internal/builder/main.go [hunk 3]
```diff
@@ -139,7 +169,7 @@ func GetModules(cfg Config) error {
 		return nil
 	}
 
-	if _, err := runGoCommand(cfg, "mod", "tidy", "-compat=1.21"); err != nil {
+	if _, err := runGoCommand(cfg, []string{}, "mod", "tidy", "-compat=1.21"); err != nil {
```
odeke-em, line 172: You could pass in `nil` instead of `[]string{}`

cmd/builder/internal/builder/main.go
```diff
@@ -194,7 +224,7 @@ func downloadModules(cfg Config) error {
 	retries := 3
 	failReason := "unknown"
 	for i := 1; i <= retries; i++ {
-		if _, err := runGoCommand(cfg, "mod", "download"); err != nil {
+		if _, err := runGoCommand(cfg, []string{}, "mod", "download"); err != nil {
```
odeke-em, line 227: Ditto about using nil

cmd/builder/internal/builder/main.go
```diff
@@ -235,7 +265,7 @@ func (c *Config) allComponents() []Module {
 
 func (c *Config) readGoModFile() (string, map[string]string, error) {
 	var modPath string
-	stdout, err := runGoCommand(*c, "mod", "edit", "-print")
+	stdout, err := runGoCommand(*c, []string{}, "mod", "edit", "-print")
```
odeke-em, line 268: Ditto about using nil

TylerHelmuth:
What happens to ocb users who depended on CGO being enabled? Will their builds work but have issues at runtime or will they just fail to build?

cmd/builder/internal/builder/main.go
evan-bradley: Good catch! I'll update this.

evan-bradley:
>What happens to ocb users who depended on CGO being enabled? Will their builds work but have issues at runtime or will they just fail to build?

There's a `cgo_enabled` option that can be set to true to allow enabling cgo. If cgo is left disabled at build time and a component that requires cgo is used, [our contributing guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#using-cgo) suggest printing a warning and defaulting to a no-op.

PR #10041, opened 2024-04-29
[chore] Allow sometimes skipping deprecation process when adding variadic arguments

evan-bradley:
#### Description

Call out that unnamed types, e.g. the function signature of an exported function, should not be relied upon by API consumers. In particular, updating a function to be variadic will break users who were depending on that function's signature.

#### Link to tracking issue
Helps https://github.com/open-telemetry/opentelemetry-collector/pull/9041


mx-psi (commented):
I think this should still be considered a breaking change, it is not a common exception, and while it's not super frequent to pass functions as arguments in Go it can happen.

I would be in favor of adding an exception for making the two step deprecation process described on https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes for changes like this, while acknowledging this is still breaking

evan-bradley:
Got it, I misunderstood the intent of this section. I've added a section to our contributing document instead.

mx-psi (approved):

codeboten (approved):
Would be good to discuss this at the SIG this week to ensure everyone is in agreement on the exception

atoulme (approved):

codeboten (approved):
Thanks @evan-bradley

PR #10054, opened 2024-04-30
[mdatagen] use mdatagen to produce component internal telemetry

codeboten:
#### Description

This updates mdatagen to generate internal telemetry for components based on metadata.yaml configuration.

#### Testing

Added tests to mdatagen and updated the batch processor to use this as well for synchronous counters and histogram

### DIFF
--- a/cmd/mdatagen/loader.go
+++ b/cmd/mdatagen/loader.go
@@ -221,6 +221,71 @@ type tests struct {
 	ExpectConsumerError bool   `mapstructure:"expect_consumer_error"`
 }
 
+type counter struct {
+	ValueType   string `mapstructure:"value_type"`
+	Monotonic   bool   `mapstructure:"monotonic"`
+	Synchronous bool   `mapstructure:"synchronous"`
+}
+
+type histogram struct {
+	ValueType string `mapstructure:"value_type"`
+}
+
+func (m *internalMetric) Unmarshal(parser *confmap.Conf) error {
+	if !parser.IsSet("enabled") {
+		return errors.New("missing required field: `enabled`")
+	}
+	return parser.Unmarshal(m)
+}
+
+func (m internalMetric) Type() string {
+	typeStr := ""
+	if m.Counter != nil {
+		switch m.Counter.ValueType {
+		case "int":
+			typeStr += "Int64"
+		case "double":
+			typeStr += "Float64"
+		}
+
+		if !m.Counter.Monotonic {
+			typeStr += "UpDown"
+		}
+		typeStr += "Counter"
+		return typeStr
+	}
+	if m.Histogram != nil {
+		switch m.Histogram.ValueType {
+		case "int":
+			return "Int64Histogram"
+		case "double":
+			return "Float64Histogram"
+		}
+	}
+	return typeStr
+}
+
+type internalMetric struct {
+	// Enabled defines whether the metric is enabled by default.
+	Enabled bool `mapstructure:"enabled"`
+
+	// Description of the metric.
+	Description string `mapstructure:"description"`
+
+	// Unit of the metric.
+	Unit string `mapstructure:"unit"`
+
+	// Counter stores metadata for counter metric type
+	Counter *counter `mapstructure:"counter,omitempty"`
+
+	// Histogram stores metadata for histogram metric type
+	Histogram *histogram `mapstructure:"histogram,omitempty"`
+}
+
+type telemetry struct {
+	Metrics map[metricName]internalMetric `mapstructure:"metrics"`
+}
+
 type metadata struct {
 	// Type of the component.
 	Type string `mapstructure:"type"`
@@ -228,6 +293,8 @@ type metadata struct {
 	Parent string `mapstructure:"parent"`
 	// Status information for the component.
 	Status *Status `mapstructure:"status"`
+	// Telemetry information for the component.
+	Telemetry telemetry `mapstructure:"telemetry"`
 	// SemConvVersion is a version number of OpenTelemetry semantic conventions applied to the scraped metrics.
 	SemConvVersion string `mapstructure:"sem_conv_version"`
 	// ResourceAttributes that can be emitted by the component.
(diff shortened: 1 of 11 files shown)


songy23 (commented):
Tests failed https://github.com/open-telemetry/opentelemetry-collector/actions/runs/8898569599/job/24435992495?pr=10054

codeboten:
Thanks @songy23, will wait to get feedback on the rest of the PR before fixing to ensure contributors are ok with this approach. One of the annoying things is that most of the internal telemetry is defined in helper packages, which don't currently have metadata.yaml configuration since they're not components. This is a solvable problem, but I do want to raise this as something that will have to change.

Ideally I would like to see internal telemetry generated by `*helper` packages listed in the documentation for components that use them. For example, i would like to see `exporter_sent_log_records` listed as a metric that is emitted by every exporter in their documentation. I think it would make it easier for end users to understand the telemetry their pipelines emit, than to have to search for the helper's documentation. Alternatively a link from one to the other works too :)

cmd/mdatagen/loader.go
```diff
@@ -221,13 +221,80 @@ type tests struct {
...
+		case "int":
+			return "Int64Histogram"
+		case "double":
+			return "Float64Histogram"
+		}
+	}
+	return typeStr
+}
+
+type internalMetric struct {
```
dmitryax, line 268: Can we reuse the existing `metric` struct instead? It'd be nice to provide a consistent interface for metadata.yaml, and potentially config. We would need to add `histogram` there. I don't see a reason to use different naming, e.g., counter vs. sum, given that the end-user outcome is the same.

codeboten: @dmitryax i was trying to tie this to the type of instrument that would be generated as i thought it would be easier to generate the otel instrument, and leave it up to components to use the otel API directly for recording measurements. i can change it though

codeboten: updated the code to remove the need for new internalMetric type. PTAL

cmd/mdatagen/templates/telemetry.go.tmpl
```diff
@@ -6,6 +6,9 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
+    {{- if .Telemetry.Metrics }}
+    "go.uber.org/multierr"
```
TylerHelmuth, line 10: A lot of contrib components have been following https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/25121 to stop using multierr and this may undo that progress.

renovate.json
```diff
@@ -35,8 +35,10 @@
       },
       {
         "matchManagers": ["gomod"],
-        "matchSourceUrlPrefixes": ["https://go.opentelemetry.io/contrib"],
-        "groupName": "All go.opentelemetry.io/contrib packages"
+        "matchSourceUrls": [
+          "https://github.com/open-telemetry/opentelemetry-go-contrib"
+        ],
+        "groupName": "All opentelemetry-go-contrib packages"
```
TylerHelmuth, line 41: Unrelated I'm guessing

cmd/mdatagen/metricdata.go
```diff
@@ -149,9 +158,9 @@ type sum struct {
 
 // Unmarshal is a custom unmarshaler for sum. Needed mostly to avoid MetricValueType.Unmarshal inheritance.
 func (d *sum) Unmarshal(parser *confmap.Conf) error {
-	if !parser.IsSet("aggregation_temporality") {
-		return errors.New("missing required field: `aggregation_temporality`")
-	}
+	// if !parser.IsSet("aggregation_temporality") {
+	// 	return errors.New("missing required field: `aggregation_temporality`")
+	// }
```
TylerHelmuth, line 163: Can this be deleted?

cmd/mdatagen/metadata-schema.yaml
```diff
@@ -125,3 +125,42 @@ tests:
...
+    unit:
+    # Required: metric type with its settings.
+    <sum|gauge|histogram>:
+      # Required for sum and gauge metrics: type of number data point values.
+      value_type: <int|double>
+      # Required for sum metric: whether the metric is monotonic (no negative delta values).
+      monotonic: bool
+      # Required for sum metric: whether reported values incorporate previous measurements
+      # (cumulative) or not (delta).
+      aggregation_temporality: <delta|cumulative>
```
TylerHelmuth, line 161: I see this option, but I don't see it used in the template.  Is it hooked up?

codeboten: this is not needed for the instrument, i can remove it

cmd/mdatagen/metricdata.go
codeboten: It can, i was leaving it in here to remind myself to talk about it w/ @dmitryax :D

renovate.json
codeboten: yup :(

cmd/mdatagen/templates/telemetry.go.tmpl
codeboten: mdatagen uses multierr, i can make a note to remove its usage separately

cmd/mdatagen/metricdata.go
codeboten: Deleted

cmd/mdatagen/metadata-schema.yaml
codeboten: removed

cmd/mdatagen/templates/telemetry.go.tmpl
codeboten: See https://github.com/open-telemetry/opentelemetry-collector/pull/10080

TylerHelmuth (approved):

dmitryax (approved):

cmd/mdatagen/loader.go
dmitryax: Thanks. LGTM

cmd/mdatagen/metricdata.go
dmitryax: We can probably move this validation level app and apply it differently to `metrics` (require) and `internal_metrics` (require no be not set).

codeboten: I tried to do this for a bit, but the problem is that since `metrics` are being unmarshaled as a map, there isn't a good place to inject the logic to return an error only when parsing `metrics` and not `telemetry:metrics`

Will open an issue to ensure this functionality can be brought back later on

PR #10056, opened 2024-04-30
[otelcol] Add a custom zapcore.Core for confmap logging

TylerHelmuth:
#### Description

Provides a logger to confmap that buffers logs in memory until the primary logger can be used. Once the primary logger exists, places that used the original logger are given the updated Core.

If an error occurs that would shut down the collector before the primary logger could be created, the logs are written to stdout/err using a fallback logger.

Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/10008

I've pushed the testing I did to show how the logger successfully updates.  Before config resolution the debug log in confmap is not printed, but afterwards it is.

test config:
```yaml
receivers:
  nop:

exporters:
  otlphttp:
    endpoint: http://0.0.0.0:4317
    headers:
      # Not set
      x-test: ${env:TEMP3}
  debug:
    # set to "detailed"
    verbosity: $TEMP

service:
  telemetry:
    logs:
      level: debug
  pipelines:
    traces:
      receivers:
        - nop
      exporters:
        - debug
```

image

#### Link to tracking issue

#### Testing

### DIFF
--- a/otelcol/buffered_core.go
+++ b/otelcol/buffered_core.go
@@ -1,0 +1,70 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// This logger implements zapcore.Core and is based on zaptest/observer.
+package otelcol // import "go.opentelemetry.io/collector/otelcol"
+
+import (
+	"sync"
+
+	"go.uber.org/zap/zapcore"
+)
+
+type loggedEntry struct {
+	zapcore.Entry
+	Context []zapcore.Field
+}
+
+func newBufferedCore(enab zapcore.LevelEnabler) *bufferedCore {
+	return &bufferedCore{LevelEnabler: enab}
+}
+
+var _ zapcore.Core = (*bufferedCore)(nil)
+
+type bufferedCore struct {
+	zapcore.LevelEnabler
+	mu      sync.RWMutex
+	logs    []loggedEntry
+	context []zapcore.Field
+}
+
+func (bc *bufferedCore) Level() zapcore.Level {
+	return zapcore.LevelOf(bc.LevelEnabler)
+}
+
+func (bc *bufferedCore) Check(ent zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	if bc.Enabled(ent.Level) {
+		return ce.AddCore(ent, bc)
+	}
+	return ce
+}
+
+func (bc *bufferedCore) With(fields []zapcore.Field) zapcore.Core {
+	return &bufferedCore{
+		LevelEnabler: bc.LevelEnabler,
+		logs:         bc.logs,
+		context:      append(bc.context[:len(bc.context):len(bc.context)], fields...),
+	}
+}
+
+func (bc *bufferedCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
+	all := make([]zapcore.Field, 0, len(fields)+len(bc.context))
+	all = append(all, bc.context...)
+	all = append(all, fields...)
+	bc.mu.Lock()
+	bc.logs = append(bc.logs, loggedEntry{ent, all})
+	bc.mu.Unlock()
+	return nil
+}
+
+func (bc *bufferedCore) Sync() error {
+	return nil
+}
+
+func (bc *bufferedCore) TakeLogs() []loggedEntry {
+	bc.mu.Lock()
+	logs := bc.logs
+	bc.logs = nil
+	bc.mu.Unlock()
+	return logs
+}
--- a/otelcol/buffered_core.go
+++ b/otelcol/buffered_core.go
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // This logger implements zapcore.Core and is based on zaptest/observer.
+
 package otelcol // import "go.opentelemetry.io/collector/otelcol"
 
 import (
@@ -43,7 +44,7 @@ func (bc *bufferedCore) With(fields []zapcore.Field) zapcore.Core {
 	return &bufferedCore{
 		LevelEnabler: bc.LevelEnabler,
 		logs:         bc.logs,
-		context:      append(bc.context[:len(bc.context):len(bc.context)], fields...),
+		context:      append(bc.context, fields...),
 	}
 }
 
--- a/otelcol/buffered_core.go
+++ b/otelcol/buffered_core.go
@@ -6,6 +6,7 @@
 package otelcol // import "go.opentelemetry.io/collector/otelcol"
 
 import (
+	"fmt"
 	"sync"
 
 	"go.uber.org/zap/zapcore"
@@ -24,9 +25,10 @@ var _ zapcore.Core = (*bufferedCore)(nil)
 
 type bufferedCore struct {
 	zapcore.LevelEnabler
-	mu      sync.RWMutex
-	logs    []loggedEntry
-	context []zapcore.Field
+	mu        sync.RWMutex
+	logs      []loggedEntry
+	context   []zapcore.Field
+	logsTaken bool
 }
 
 func (bc *bufferedCore) Level() zapcore.Level {
@@ -44,17 +46,21 @@ func (bc *bufferedCore) With(fields []zapcore.Field) zapcore.Core {
 	return &bufferedCore{
 		LevelEnabler: bc.LevelEnabler,
 		logs:         bc.logs,
+		logsTaken:    bc.logsTaken,
 		context:      append(bc.context, fields...),
 	}
 }
 
 func (bc *bufferedCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
+	bc.mu.Lock()
+	defer bc.mu.Unlock()
+	if bc.logsTaken {
+		return fmt.Errorf("the buffered logs have already been taken so writing is no longer supported")
+	}
 	all := make([]zapcore.Field, 0, len(fields)+len(bc.context))
 	all = append(all, bc.context...)
 	all = append(all, fields...)
-	bc.mu.Lock()
 	bc.logs = append(bc.logs, loggedEntry{ent, all})
-	bc.mu.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


otelcol/collector.go
```diff
@@ -109,14 +110,62 @@ type Collector struct {
...
+		Encoding:         "console",
+		EncoderConfig:    ec,
+		OutputPaths:      []string{"stderr"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+	logger, err := zapCfg.Build(options...)
+	if err != nil {
+		return nil, err
+	}
+	return &collectorCore{core: logger.Core()}, nil
```
TylerHelmuth, line 156: This was my lazy way of getting a useable core to wrap, there is probably a better way

mx-psi: This seems to work, in the actual implementation I guess `newCollectorCore` would start with an observer and also return some sort of handler to update the core once the logger is ready. We would need to handle concurrent access to the `core` field inseide `collectorCore` as well

TylerHelmuth: Ya I could combine this with the `observer` stuff.  I'll touch it up.

TylerHelmuth: Updated.  Here is what the Warning log looks like now:

image

TylerHelmuth: Personally I still like the idea of passing a basic logger to confmap to handle logging before the config has been rresolved, and then updating the confmap's logger to the configured logger once we're able.

codeboten: It feels like this would solve most of the problems. I guess the problem is that someone's ever going to use a configuration to emit logs via OTLP directly, they may never see logs for a configuration failure. I don't know if there's a sane way to handle this other than via documentation though, since without config you can't configure the logger you wanted to emit logs with.

TylerHelmuth: I am going to write up a rfc doc for how the collector should handle logging before it has a logger instantiated, that should help us clarify the behavior we want to implement in this PR.

djaglowski:
Generally looks good to me

otelcol/buffered_core.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (bc *bufferedCore) Sync() error {
+	return nil
+}
+
+func (bc *bufferedCore) TakeLogs() []loggedEntry {
+	bc.mu.Lock()
+	logs := bc.logs
+	bc.logs = nil
```
mx-psi, line 68: Should we make it so that core is unusable after calling `TakeLogs` once? Right now, if misused we can lose some logs (e.g. if you were to first call `TakeLogs` and then replace the core, calls to `Write` in this time slice would be lost forever).

If we make it so that `Write` returns an error after the first call to `TakeLogs` we can at least see that something went wrong (or, alternatively, make it so that the API makes it impossible for you to call things in the wrong order, although that would mean some coupling between `bufferedCore` and `collectorCore`)

otelcol/collector.go
```diff
@@ -201,6 +209,16 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
...
 		return err
 	}
+	if col.updateConfigProviderLogger != nil {
+		col.updateConfigProviderLogger(col.service.Logger().Core())
+	}
+	if col.bc != nil {
+		x := col.bc.TakeLogs()
+		for _, log := range x {
+			ce := col.service.Logger().Core().Check(log.Entry, nil)
+			ce.Write(log.Context...)
```
mx-psi, line 219: We should check `ce != nil` as done here: https://github.com/uber-go/zap/blob/fcf8ee58669e358bbd6460bef5c2ee7a53c0803a/logger.go#L246-L248

otelcol/collector.go
```diff
@@ -246,11 +264,36 @@ func (col *Collector) DryRun(ctx context.Context) error {
...
 func (col *Collector) Run(ctx context.Context) error {
 	if err := col.setupConfigurationComponents(ctx); err != nil {
 		col.setCollectorState(StateClosed)
+		logger, loggerErr := newFallbackLogger(col.set.LoggingOptions)
+		if loggerErr != nil {
+			fmt.Printf("unable to create fallback logger, %v", err)
+		} else {
+			if col.bc != nil {
+				for _, log := range col.bc.TakeLogs() {
+					logger.Log(log.Level, log.Message, log.Context...)
```
mx-psi, line 291: Why not do the same as above? (`Check` + `Write`)

otelcol/collector_core.go
```diff
@@ -0,0 +1,58 @@
...
+func (c *collectorCore) Sync() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.core.Sync()
+}
+
+func (c *collectorCore) SetCore(core zapcore.Core) {
+	c.mu.Lock()
+	c.core = core
+	c.mu.Unlock()
```
mx-psi, line 57: Use defer for consistency?

otelcol/collector_core.go
```diff
@@ -0,0 +1,58 @@
...
+	"sync"
+
+	"go.uber.org/zap/zapcore"
+)
+
+var _ zapcore.Core = (*collectorCore)(nil)
+
+type collectorCore struct {
+	core zapcore.Core
+	mu   sync.Mutex
```
mx-psi, line 16: Would a `RWMutex` make more sense here? (I am actually not sure). 

My reasoning: `core` must already guarantee that you can call its methods concurrently, the only thing we have to guard against is updating the field itself. It should be safe to `Lock` in `SetCore`, but `RLock` in any other method.

otelcol/buffered_core.go
TylerHelmuth: I think this is a good restriction to start with.  If we need to later we could open it up, but based on our current use case this enforcement seems reasonable.

otelcol/collector.go
TylerHelmuth: That shouldn't be necessary, all the `CheckedEntry` methods are safe to call on nil: https://github.com/uber-go/zap/blob/fcf8ee58669e358bbd6460bef5c2ee7a53c0803a/zapcore/entry.go#L234-L236

otelcol/collector.go
TylerHelmuth: I forgot to update this instance, but you're right, it should use the same logic.

otelcol/collector_core.go
TylerHelmuth: I think you're right

otelcol/collector.go
mx-psi: Hm, I doubt this will change but I am a little scared since this behavior is not guaranteed by the method's docstring

otelcol/buffered_core.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func (bc *bufferedCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
+	all := make([]zapcore.Field, 0, len(fields)+len(bc.context))
+	all = append(all, bc.context...)
+	all = append(all, fields...)
+	bc.mu.Lock()
+	bc.logs = append(bc.logs, loggedEntry{ent, all})
+	bc.mu.Unlock()
```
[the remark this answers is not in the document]
mx-psi: :shrug: Up to you, I don't have a strong opinion here, I feel like it's slightly safer when it comes to refactoring but not by much

TylerHelmuth:
@mx-psi I believe I've address all feedback

otelcol/collector.go
```diff
@@ -246,11 +266,40 @@ func (col *Collector) DryRun(ctx context.Context) error {
...
+}
+
 // Run starts the collector according to the given configuration, and waits for it to complete.
 // Consecutive calls to Run are not allowed, Run shouldn't be called once a collector is shut down.
 func (col *Collector) Run(ctx context.Context) error {
 	if err := col.setupConfigurationComponents(ctx); err != nil {
 		col.setCollectorState(StateClosed)
+		logger, loggerErr := newFallbackLogger(col.set.LoggingOptions)
+		if loggerErr != nil {
+			fmt.Printf("unable to create fallback logger, %v", err)
```
mx-psi, line 289: Should we just return the errors here?
```suggestion
			return errors.Join(err, fmt.Errorf("unable to create fallback logger: %w", loggerErr))
```

mx-psi:
@TylerHelmuth can you address the lint error?

```
Error: collector.go:289:10: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)
```

I can merge after that

codeboten (approved):

evan-bradley (approved):

mx-psi:
The Windows test failure seems to be legit:
```
    collector_windows_service_test.go:146: 
        	Error Trace:	D:/a/opentelemetry-collector/opentelemetry-collector/otelcol/collector_windows_service_test.go:146
        	Error:      	Should be true
        	Test:       	TestCollectorAsService/ConfigFileNotFound
```

It's failing here

https://github.com/open-telemetry/opentelemetry-collector/blob/7b63bfc9a46b692451c838bb7acd2e3692580bf4/otelcol/collector_windows_service_test.go#L146

with this test 

https://github.com/open-telemetry/opentelemetry-collector/blob/7b63bfc9a46b692451c838bb7acd2e3692580bf4/otelcol/collector_windows_service_test.go#L54-L58

The Windows service uses a special zap core https://github.com/open-telemetry/opentelemetry-collector/blob/7b63bfc9a46b692451c838bb7acd2e3692580bf4/otelcol/collector_windows.go#L203

I am wondering if the fallback logger is somehow messing with this

TylerHelmuth:
@pjanotti I need some help here, why can the windows service not handle instantiating a fallback logger before returning the error on start? If I use `zap.NewNOP` it works, but something about the builder causes problems.

otelcol/collector_windows.go
```diff
@@ -202,7 +202,7 @@ func (w windowsEventLogCore) Sync() error {
 
 func withWindowsCore(elog *eventlog.Log, serviceConfig **service.Config) func(zapcore.Core) zapcore.Core {
 	return func(core zapcore.Core) zapcore.Core {
-		if serviceConfig != nil {
+		if serviceConfig != nil && *serviceConfig != nil {
```
TylerHelmuth, line 205: @mx-psi fixed the windows test.  The issue was that when the config could not resolve the underlying serviceConfig was nil, but this check was only checking the pointer to the pointer not being nil.  When we passed the options to the fallback logger we ended up in this method and tried to do `(*serviceConfig).Telemetry.Logs.OutputPaths` on `nil`.  Added an extra nil check to make sure the underlying `serviceConfig` is also not nil.

codeboten: Thanks for getting to the bottom of this Windows test, can we make you the official Windows reviewer? 😄

TylerHelmuth: No thank you

pjanotti (approved):

PR #10065, opened 2024-05-01
mdatagen: add wildcard name matching for configs

braydonk:
#### Description
This PR adds support for wildcard name matching.

Generated MetricsConfig and ResourceAttributesConfig will now support providing names not just as full names, but also using `*` wildcards and multimatching with `{x,y,etc}`. This allows you to apply configs to groups of metrics and resource attributes, simplifying configs.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10074

#### Testing
I `go install`ed mdatagen from this PR and ran it on a receiver in `opentelemetry-collector-contrib` and was able to use wildcard matching to enable/disable groups of metrics as expected.

#### Documentation

I have written a specification in a gist: https://gist.github.com/braydonk/ccb6775331fdd5dca91a650330b9839f

I am not sure where this documentation should live.


braydonk:
I think that code coverage change is fine here because none of the `return err` cases for template file generation are covered anyway.

braydonk:
The changelog issue can only be resolved once the issue is moved from `contrib` to here.

cmd/mdatagen/templates/match.go.tmpl
```diff
@@ -0,0 +1,234 @@
...
+package metadata
+
+import (
+	"errors"
+	"io"
+	"sort"
+	"strings"
+)
+
+var ErrNotMultimatch = errors.New("this index doesn't represent a valid multimatch pattern")
```
dmitryax, line 12: I don't see a reason to export any API in this file. We should only export config field types and `Unmarshal` methods

cmd/mdatagen/templates/match.go.tmpl
```diff
@@ -0,0 +1,234 @@
...
+	for pattern, assign := range patternMap {
+		AddPattern(&patterns, NewPattern(pattern, assign))
+	}
+
+	for _, pattern := range patterns {
+		matched := []string{}
+		for _, name := range matchNames {
+			if pattern.Match(name) {
+				matched = append(matched, name)
+			}
```
dmitryax, line 226: We should fail unmarshalling if we run into a pattern that doesn't match any metric names. It'll be aligned with the strict matching behavior

cmd/mdatagen/templates/match.go.tmpl
```diff
@@ -0,0 +1,234 @@
...
+			if patternScan.isLast() {
+				return true
+			}
+			patternNext, err := patternScan.peek()
+			if err == nil && strChar == patternNext {
+				lastWildcard = patternScan.idx
+				patternScan.idx += 2
+			}
+
+		case '{':
```
dmitryax, line 65: Can you please point to some specifications for this sort of pattern matching? Does it usually go with wildcard matching?

dmitryax:
@braydonk, are you still working on this? Just checking. Let me know if you have any concerns with my comments

braydonk:
Hi @dmitryax yes I am. I realized that I forgot to mention that I am writing a full spec to explain how it works, and while I was doing that found a number of edge cases that prompted me to rewrite the parser. Taking longer than expected. I am still working on this though and will keep it unstale.

cmd/mdatagen/templates/match.go.tmpl
braydonk: Done in [d82df38](https://github.com/open-telemetry/opentelemetry-collector/pull/10065/commits/d82df38d422314a454be87805bd2f69bde76b472)

braydonk:
What the actual errors look like from collector output is not super obvious from the PR, so here are some examples. I used the hostmetrics process scraper in my example.

**Not matching any names with the patterns**
```yaml
hostmetrics:
  collection_interval: 10s
  scrapers:
    process:
      metrics:
        "nothing.*":
          enabled: false
```
```bash
$ ./otelcontribcol --config config.yaml 
Error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'receivers': error reading configuration for "hostmetrics": error reading settings for scraper type "process": 1 error(s) decoding:

* error decoding 'metrics': specified patterns don't match any names
2024/05/28 16:06:44 collector server run finished with error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'receivers': error reading configuration for "hostmetrics": error reading settings for scraper type "process": 1 error(s) decoding:

* error decoding 'metrics': specified patterns don't match any names
```

**Parsing errors in a pattern**
```yaml
hostmetrics:
  collection_interval: 10s
  scrapers:
    process:
      metrics:
        "a_.*":
          enabled: false
```
```bash
$ ./otelcontribcol --config config.yaml
Error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'receivers': error reading configuration for "hostmetrics": error reading settings for scraper type "process": 1 error(s) decoding:

* error decoding 'metrics': pattern parsing error:
a_.*
  ^
invalid pattern: unexpected token
2024/05/28 16:06:52 collector server run finished with error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'receivers': error reading configuration for "hostmetrics": error reading settings for scraper type "process": 1 error(s) decoding:

* error decoding 'metrics': pattern parsing error:
a_.*
  ^
invalid pattern: unexpected token
```

braydonk:
I'm gonna wait to address the conflicts, cause this is just going to keep coming up continuously throughout the review.

braydonk:
Not stale

cmd/mdatagen/templates/match.go.tmpl
```diff
@@ -0,0 +1,492 @@
...
+				pt.groups,
+				group{typ: groupIdentifier, values: []string{identifier}},
+			)
+		}
+
+		// After a valid group, only a '.' and another group
+		// or end of input are valid.
+		if !p.isFinished() {
+			curr, _ := p.current()
+			if curr != '.' {
```
dmitryax, line 116: If we allow a group as a whole to be either an identifier, a group, or a wildcard, can we use `strings.Split(pattern, '.')` and just read each part instead of having this loop? I think it should simplify the code.

I like the approach you took. However, even if we do the parsing ourselves, I'd like to keep the code simpler if possible.

cmd/mdatagen/main.go
```diff
@@ -124,6 +124,23 @@ func run(ymlPath string) error {
 		toGenerate[filepath.Join(tmplDir, "resource.go.tmpl")] = filepath.Join(codeDir, "generated_resource.go")
 		toGenerate[filepath.Join(tmplDir, "resource_test.go.tmpl")] = filepath.Join(codeDir, "generated_resource_test.go")
 	}
+	if err = generateFile(filepath.Join(tmplDir, "match.go.tmpl"),
```
dmitryax, line 127: This will be generated for every component in contrib. We need to generate these files only for scrapers `len(md.Metrics) > 0`.

cmd/mdatagen/templates/config.go.tmpl
```diff
@@ -10,6 +10,12 @@ import (
 )
 {{ if .Metrics -}}
 
+var MetricNames = []string{
```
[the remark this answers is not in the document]
braydonk: Good catch thanks, this was an accident

cmd/mdatagen/templates/match.go.tmpl
braydonk: I'll give that a shot. When I first tried that I was having trouble making nice errors messages, but I think I can come up with a way after all. Will give it a try soon

cmd/mdatagen/main.go
braydonk: Done in [f416166](https://github.com/open-telemetry/opentelemetry-collector/pull/10065/commits/f4161668f1a6afea7486679a3d2d3f35495c5410)

cmd/mdatagen/templates/config.go.tmpl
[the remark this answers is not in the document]
braydonk: Done in [a758241](https://github.com/open-telemetry/opentelemetry-collector/pull/10065/commits/a75824167ff8bb9038e84d5e7e9f39b08d8e1746)

I couldn't render directly in `expandPatternMap` because there could be either metrics or resource attributes passed in. I render both arrays more locally to their usage though instead of as globals.

cmd/mdatagen/templates/match.go.tmpl
braydonk: I tried this but I found it was becoming equally as complex. The complexity saved by this approach is:
* Shared scanner object, so I'm not needing to pass around indexes for context to different spots
* I can share identifier parsing code between an identifier as a root group, and an identifier as part of a multimatch

How strongly do you feel on this? I can try to make it work if you think it's the way to go.

cmd/mdatagen/main.go
braydonk: Correction in [a522523](https://github.com/open-telemetry/opentelemetry-collector/pull/10065/commits/a52252349053ff3d76313b6e476f75fa3d5bd423)

PR #10066, opened 2024-05-01
[otelcol] rfc for how to log during startup

TylerHelmuth:
#### Description

This is an RFC to help us decide how we want `otelcol` to provide a logger before the primary logger is created. As we discuss I will update the doc.  Before this is merged we should have decided on a solution and the Accepted Solution section must be updated.

Related to https://github.com/open-telemetry/opentelemetry-collector/pull/10056

#### Link to tracking issue
This unblocks:
- https://github.com/open-telemetry/opentelemetry-collector/issues/9162
- https://github.com/open-telemetry/opentelemetry-collector/issues/5615

### DIFF
--- a/docs/rfcs/logging-before-config-resolution.md
+++ b/docs/rfcs/logging-before-config-resolution.md
@@ -26,7 +26,7 @@ be shared with users such as:
 
 1. Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.
 2. Log timestamps must be accurate to when the log was written, regardless of when the log is written to the user-specified location(s)
-3. If an error occurs before the primary logger is created any previously written logs MUST be written to the user-specified location(s) if they have not already been written there.
+3. If an error occurs before the primary logger is created any previously written logs MUST be written to the either stout or stderr if they have not already been written there.
 
 ## Current behavior
 
--- a/docs/rfcs/logging-before-config-resolution.md
+++ b/docs/rfcs/logging-before-config-resolution.md
@@ -62,6 +62,6 @@ Downsides:
 
 ## Accepted Solution
 
-[Buffer Logs in Memory and Write Them Once the Primary Logger Exists](#Buffer-Logs-in-Memory-and-Write-Them-Once-the-Primary-Logger-Exists)
+[Buffer Logs in Memory and Write Them Once the Primary Logger Exists](#buffer-logs-in-memory-and-write-them-once-the-primary-logger-exists)
 
 This solution, while more complex, allows the collector to write out the logs in the user-specified format whenever possible.  A fallback logger must be used in situations where the primary logger could not be created and the collector is shutting down, such as when encountering an error during configuration resolution, but otherwise the primary logger will be used to write logs that occurred before the primary logger existed.
--- a/docs/rfcs/logging-before-config-resolution.md
+++ b/docs/rfcs/logging-before-config-resolution.md
@@ -26,7 +26,8 @@ be shared with users such as:
 
 1. Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.
 2. Log timestamps must be accurate to when the log was written, regardless of when the log is written to the user-specified location(s)
-3. If an error causes the collector to gracefully terminate before the primary logger is created any previously written logs MUST be written to the either stout or stderr if they have not already been written there.
+3. The EntryCaller (what line number the log originates from) must be accurate to where the log was written.
+4. If an error causes the collector to gracefully terminate before the primary logger is created any previously written logs MUST be written to the either stout or stderr if they have not already been written there.
 
 ## Current behavior
 
@@ -61,4 +62,6 @@ Downsides:
 
 ## Accepted Solution
 
-I'll fill this out once we decide.
+[Buffer Logs in Memory and Write Them Once the Primary Logger Exists](#Buffer-Logs-in-Memory-and-Write-Them-Once-the-Primary-Logger-Exists)
+
+This solution, while more complex, allows the collector to write out the logs in the user-specified format whenever possible.  A fallback logger must be used in situations where the primary logger could not be created and the collector is shutting down, such as when encountering an error during configuration resolution, but otherwise the primary logger will be used to write logs that occurred before the primary logger existed.
--- a/docs/rfcs/logging-before-config-resolution.md
+++ b/docs/rfcs/logging-before-config-resolution.md
@@ -26,7 +26,7 @@ be shared with users such as:
 
 1. Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.
 2. Log timestamps must be accurate to when the log was written, regardless of when the log is written to the user-specified location(s)
-3. If an error occurs before the primary logger is created any previously written logs MUST be written to the either stout or stderr if they have not already been written there.
+3. If an error causes the collector to gracefully terminate before the primary logger is created any previously written logs MUST be written to the either stout or stderr if they have not already been written there.
 
 ## Current behavior
 
--- a/docs/rfcs/logging-before-config-resolution.md
+++ b/docs/rfcs/logging-before-config-resolution.md
@@ -1,0 +1,62 @@
+# How To Log Before Config Resolution
+
+## Overview
+
+The OpenTelemetry Collector supports configuring a primary logger that the collector and its components use to write logs.
+This logger cannot be created until the user's configuration has been completely resolved.
+There is a need to write logs during the collector start-up, before the primary logger is instantiated, such as during
+configuration resolution or config validation. This document describes
+
+- why providing logging capabilities during startup is important
+- the current (as of v0.99.0) behavior of the Collector
+- different solutions to the problem
+- the accepted solution
+
+## Why Logging During Startup is Important
+
+When the collector is starting it tries to resolve user configuration as quickly as possible.
+But the Collector's configuration resolution strategy is not trivial - it allows for complex interactions between
+multiple, different config sources that must all resolve without error. During this process important information could
+be shared with users such as:
+- [Warnings about deprecated syntax](https://github.com/open-telemetry/opentelemetry-collector/issues/9162)
+- [Warnings about undesired, but handled, situations](https://github.com/open-telemetry/opentelemetry-collector/issues/5615)
+- Debug information
+
+## Requirements for any solution
+
+1. Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.
+
+## Current behavior
+
+As of v0.99.0, the collector does not provide a way to log before the primary logger is instantiated.
+
+## Solutions
+
+### Buffer Logs in Memory and Write Them Once the Primary Logger Exists
+
+The Collector could provide a temporary logger that, when written to, keeps the logs in memory. These logs could then
+be passed to the primary logger to be written out in the properly configured format/level.
+
+Benefits:
+- Logs are written in the user-specified format/level
+
+Downsides:
+- If an error occurs before the primary logger is created the buffered logs will not be written out
+- If the primary logger is used to write any logs before the buffered logs are passed, logs will be out of order
+
+### Create a Logger Using the Primary Logger's Defaults
+
+When the user provides no primary logger configuration the Collector creates a Logger using a set of default values.
+The Collector could, very early in startup, create a logger using these exact defaults and use it until the primary
+logger is instantiated.
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,62 @@
...
+
+The Collector could provide a temporary logger that, when written to, keeps the logs in memory. These logs could then
+be passed to the primary logger to be written out in the properly configured format/level.
+
+Benefits:
+- Logs are written in the user-specified format/level
+
+Downsides:
+- If an error occurs before the primary logger is created the buffered logs will not be written out
+- If the primary logger is used to write any logs before the buffered logs are passed, logs will be out of order
```
evan-bradley, line 45: I think we can fix this if we can find a good solution. This is an implementation detail of the Collector<->service interface.

docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,62 @@
...
+### Buffer Logs in Memory and Write Them Once the Primary Logger Exists
+
+The Collector could provide a temporary logger that, when written to, keeps the logs in memory. These logs could then
+be passed to the primary logger to be written out in the properly configured format/level.
+
+Benefits:
+- Logs are written in the user-specified format/level
+
+Downsides:
+- If an error occurs before the primary logger is created the buffered logs will not be written out
```
evan-bradley, line 44: I would argue for using a fallback zap logger that has a high log level (debug/trace) in this case.

evan-bradley (commented):
Thanks for writing this up.

docs/rfcs/logging-before-config-resolution.md
TylerHelmuth: I agree that we can prevent this downside from happening with good coding standards, but ultimately it is a possibility.

docs/rfcs/logging-before-config-resolution.md
TylerHelmuth: Are you suggesting a second solution requirement like:

`2. If an error occurs before the primary logger is create any previously written logs MUST be written to either stout/stderr if they have not already`

docs/rfcs/logging-before-config-resolution.md
codeboten: This could be mitigate at initialization of the logger no? Once the config has been resolved, it should be possible to output the buffered logs before making the logger available to components, or am i missing something here

docs/rfcs/logging-before-config-resolution.md
evan-bradley: Yeah, I think that would be good.

docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,62 @@
...
+When the user provides no primary logger configuration the Collector creates a Logger using a set of default values.
+The Collector could, very early in startup, create a logger using these exact defaults and use it until the primary
+logger is instantiated.
+
+Benefits:
+- Logs order is preserved
+- Logs are still written when an error occurs before the primary logger can be instantiated
+
+Downsides:
+- Logs may be written in a format/level that differs from the format/level of the primary logger
```
codeboten, line 58: i think this downside alone makes this solution not viable. It would require users to always configure multiple sources for their logs if they want their entire logs, unless their configured logging option is the same at the default, in which case i would wonder why make it configurable at all :)

codeboten (commented):
I'm in favour of the first option, buffer the logs until we either get a configured logger, or emit them over stderr on the case of the configuration failing to resolve.

One question here is what happens in the case of the configuration taking a long time to resolve (if someone is using a remote, ie s3, provider)?

docs/rfcs/logging-before-config-resolution.md
evan-bradley: Okay, I see more of what you mean now. Yes, I suppose that we can't guarantee logs will have completely correct ordering. I do think we could fairly easily achieve it so long as we find a way to have the service itself print out the confmap logs. The only places we have a zap logger will be in the confmap components and in the service. I think it would be fairly straightforward to unmarshal the config, create the logger, then log the confmap logs before anything else is done.

I think we should change the statement to say that it opens up the opportunity for them to be printed out of order, not that they necessarily will be printed out of order.

djaglowski: As long as logs are timestamped (which they always should be), this isn't much of an issue. Ultimately the timestamps define the order, rather than their order of appearance in a medium.

evan-bradley:
>One question here is what happens in the case of the configuration taking a long time to resolve (if someone is using a remote, ie s3, provider)?

That's a good question. I think unfortunately users will just see a delay before the logs are printed on Collector startup. I think this is an acceptable tradeoff considering the Collector is usually started without a human looking at the console.

docs/rfcs/logging-before-config-resolution.md
TylerHelmuth: > As long as logs are timestamped (which they always should be), this isn't much of an issue. Ultimately the timestamps define the order, rather than their order of appearance in a medium.

I'm ok with this, I'll add to the requirements: `log timestamps must be preserved no matter when they end up being written to stdout/stderr

docs/rfcs/logging-before-config-resolution.md
TylerHelmuth: What about the situation when the collector errors during config resolution? The returned errors are not currently logged using the user-configured format/level. Also any errors written by a fallback logger ([proposed here](https://github.com/open-telemetry/opentelemetry-collector/pull/10066#discussion_r1586778011) would not be guaranteed to match the configured format/level either.

docs/rfcs/logging-before-config-resolution.md
TylerHelmuth: For some additional context, Refinery (a Honeycomb sampling application) also let's users configure how Refinery exports logs. The config resolution is much simpler than the Collectors, but when it experiences an error that prevents it from being able to create a logger it throws up its hands and says ["here's the error via fmt.Printf, good luck"](https://github.com/honeycombio/refinery/blob/36b6fc77acfb7e5a8fd144652c9a0768fe06c6d9/config/file_config.go#L401).

I think following similar fallback logic, but with an actual zap logger, is totally valid.

TylerHelmuth:
I've updated https://github.com/open-telemetry/opentelemetry-collector/pull/10056 to implement solution 1 and all solution requirements as an example implementation

docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,64 @@
...
+be shared with users such as:
+- [Warnings about deprecated syntax](https://github.com/open-telemetry/opentelemetry-collector/issues/9162)
+- [Warnings about undesired, but handled, situations](https://github.com/open-telemetry/opentelemetry-collector/issues/5615)
+- Debug information
+
+## Requirements for any solution
+
+1. Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.
+2. Log timestamps must be accurate to when the log was written, regardless of when the log is written to stdout/sterr
+3. If an error occurs before the primary logger is created any previously written logs MUST be written to either stout or stderr if they have not already been written there.
```
mx-psi, line 29: The log may not be written to stdout/stderr at all: we allow specifying one or multiple files as output

```suggestion
2. Log timestamps must be accurate to when the log was written, regardless of when the log is written to the user-specified location(s)
3. If an error occurs before the primary logger is created any previously written logs MUST be written to the user-specified location(s) if they have not already been written there.
```

docs/rfcs/logging-before-config-resolution.md
mx-psi: > What about the situation when the collector errors during config resolution? 

If the configuration is invalid, we can't trust it, so we need to fall back. Since we can't do better, using our default/what most applications do makes sense

mx-psi (commented):
From https://github.com/open-telemetry/opentelemetry-collector/pull/10008#discussion_r1582802541, can we list whether preserving the following is required?

1. The timestamp
2. The [EntryCaller](https://pkg.go.dev/go.uber.org/zap@v1.27.0/zapcore#EntryCaller) (what line number the log originates from)
3. The ['terminal behavior'](https://github.com/uber-go/zap/blob/fcf8ee58669e358bbd6460bef5c2ee7a53c0803a/logger.go#L346-L356) (whether to panic at panic level)

I think there is consensus about preserving (1), I think (2) would be nice, and I personally don't care much about (3) (we don't really use that at all, and we tell people not to panic at runtime)

TylerHelmuth:
@mx-psi preserving 1 is definitely required (and called out in the rfc).  I think 2 should be required as well, if we all agree I can add that. 3 I don't think needs to be preserved as panicing goes against our development principals.

docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,64 @@
...
+be shared with users such as:
+- [Warnings about deprecated syntax](https://github.com/open-telemetry/opentelemetry-collector/issues/9162)
+- [Warnings about undesired, but handled, situations](https://github.com/open-telemetry/opentelemetry-collector/issues/5615)
+- Debug information
+
+## Requirements for any solution
+
+1. Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.
+2. Log timestamps must be accurate to when the log was written, regardless of when the log is written to the user-specified location(s)
+3. If an error occurs before the primary logger is created any previously written logs MUST be written to the either stout or stderr if they have not already been written there.
```
TylerHelmuth: @mx-psi I added back the `stdout`/`stderr` requirement since it is not possible to write to a user-specific location without the primary logger, and in this situation the collector is terminating and primary logger could not be created.

mx-psi: Right, makes sense, the change was only needed in (2)

mx-psi:
> @mx-psi preserving 1 is definitely required (and called out in the rfc). I think 2 should be required as well, if we all agree I can add that. 3 I don't think needs to be preserved as panicing goes against our development principals.

Oki, that works for me. A solution with a changeable `zapcore.Core` should work here then, and we won't need to copy any internals if we don't require (3)

TylerHelmuth:
@codeboten @mx-psi @djaglowski @evan-bradley I've updated the RFC with an accepted solution based on the PRs discussion.

djaglowski (approved):

codeboten (approved):

docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,67 @@
...
+
+The Collector could provide a temporary logger that, when written to, keeps the logs in memory. These logs could then
+be passed to the primary logger to be written out in the properly configured format/level.
+
+Benefits:
+- Logs are written in the user-specified format/level
+
+Downsides:
+- If an error occurs before the primary logger is created the buffered logs will not be written out
+- If the primary logger is used to write any logs before the buffered logs are passed, logs will be out of order
```
evan-bradley, line 48:
```suggestion
- If the primary logger is used to write any logs before the buffered logs are passed, logs may be out of order. There are no guarantees that logs will be written in order, so the log timestamps should be taken as the source of truth for ordering.
```

docs/rfcs/logging-before-config-resolution.md
```diff
@@ -0,0 +1,67 @@
...
+## Current behavior
+
+As of v0.99.0, the collector does not provide a way to log before the primary logger is instantiated.
+
+## Solutions
+
+### Buffer Logs in Memory and Write Them Once the Primary Logger Exists
+
+The Collector could provide a temporary logger that, when written to, keeps the logs in memory. These logs could then
+be passed to the primary logger to be written out in the properly configured format/level.
```
mx-psi, line 41: Should we explicitly mention what happens once we have a primary logger with components (e.g. confmap providers) created before it was available? I think we expect to switch to the primary logger once available, right?

TylerHelmuth: I call this requirement out above:

> Once the primary logger is instantiated, it should be usable anywhere in the Collector that does logging.

Is that what you mean?

mx-psi: Ah, yes! Missed that :)

mx-psi (approved):

evan-bradley (approved):

PR #9257, opened 2024-01-10
[service] Validate pipeline type against component types

ycombinator:
**Description:** <Describe what has changed.>

This change adds another layer of validation to pipelines. It validates that all the components in a pipeline are of the same type as the pipeline.

For example, if a `metrics` pipeline contains a `traces`-only receiver, the `otelcol validate -config ...` command will fail.

**Link to tracking Issue:** 
Fixes #8007.

**Testing:** 
Added unit test + existing tests are passing.

**Documentation:**
godoc.

PR #10078, opened 2024-05-03
Revert pipeline type validation

### LINKED ISSUE #10031: otelcol.Collector.DryRun() now instantiates all components
**Describe the bug**
This behavior _might_ be intended. However, if you write your own entrypoint into the collector, as I have, and you call `otelcol.Collector.DryRun()` with a goal of determining whether the configuration "looks valid" before calling `Run()`, then `service.New()` is now called as part of the dry run, and its return value is discarded. This behavior is a consequence of [this commit](https://github.com/open-telemetry/opentelemetry-collector/commit/de3ef01bff4f1e4db0eee94abce862c66582f208).

Due to the above behavior, calling `DryRun()` now, for example, calls `open()` and `bind()` for the self-monitoring telemetry-service port and the health-check extension port, among other ports. When `Run()` is subsequently called, the second instance of each server component fails with a `bind()` error because its port has already been opened and bound.

This _might be a misuse_ of the `DryRun()` function. Let me know, and thanks in advance.

**Steps to reproduce**
Write the above code.

The old behavior.

**What did you see instead?**
```
Error: failed to start extensions: failed to bind to address 0.0.0.0:13133: listen tcp 0.0.0.0:13133: bind: address already in use; failed to shutdown pipelines: no existing monitoring routine is running
2024/04/24 22:03:22 collector server run finished with error: failed to start extensions: failed to bind to address 0.0.0.0:13133: listen tcp 0.0.0.0:13133: bind: address already in use; failed to shutdown pipelines: no existing monitoring routine is running
```

**What version did you use?**
v0.98.0

**Environment**
go version go1.22.0 linux/amd64

ycombinator:
#### Description

This PR reverts the change made in https://github.com/open-telemetry/opentelemetry-collector/pull/9257 due to problems reported in https://github.com/open-telemetry/opentelemetry-collector/issues/10031.

#### Link to tracking issue
Fixes #10031.

### DIFF
--- a/.chloggen/revert-validate-pipeline-types.yaml
+++ b/.chloggen/revert-validate-pipeline-types.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'breaking'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "The `validate` sub-command no longer validates that each pipeline's type is the same as its component types"
+
+# One or more tracking issues or pull requests related to the change
+issues: [10031]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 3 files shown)


TylerHelmuth (approved):

TylerHelmuth:
@ycombinator please add a changelog entry

mx-psi (approved):

PR #10094, opened 2024-05-06
[chore] use mdatagen for exporterhelper metrics

codeboten:
Uses the new mdatagen capabilities to generate internal telemetry details for exporterhelper.


cmd/mdatagen/main.go
```diff
@@ -62,22 +62,22 @@ func run(ymlPath string) error {
 				filepath.Join(codeDir, "generated_status.go"), md, "metadata"); err != nil {
 				return err
 			}
-			if err = generateFile(filepath.Join(tmplDir, "telemetry.go.tmpl"),
```
songy23, line 65: Why move these few lines?

songy23 (approved):
LGMT except minor nits

cmd/mdatagen/main.go
codeboten: Not super clear from the description, but the commit for the change describes it:

> ensure package test and telemetry is generated for pkg type

PR #10109, opened 2024-05-07
[DO NOT MERGE] Showcase PR: Add profiling pdata

dmathieu:
#### Description

This adds the profiling proto schema to pdata, so the receiver/consumer interfaces can then start accepting that data.

Proto files:
https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto/profiles/v1experimental

#### Testing

This is being tested by auto-generated tests.

cc @open-telemetry/profiling-approvers @open-telemetry/profiling-maintainers @open-telemetry/profiling-triagers


dmathieu:
Sure.
Should I put it into `experimental/pdata/pprofile` (in which case pdatagen can't be used, has to be refactored or something else, as it only handles putting generated code into the `pdata` package), or `pdata/pprofileexperimental`?

bogdandrutu:
>Sure.
Should I put it into experimental/pdata/pprofile (in which case pdatagen can't be used, has to be refactored or something else, as it only handles putting generated code into the pdata package), or pdata/pprofileexperimental?

Just have a different module for `pdata/pprofile`

dmathieu:
I have moved it into `pdata/experimental`.
I couldn't move it to `experimental/pdata`, as we need access to the `pdata/internal` package.

mx-psi (changes requested):
This doesn't need to be in an `experimental` folder, we just need it to be in a different Go module versioned `0.x`. That way when we go from `0.x` to `1.x`, we won't need to change the imports for this package

dmathieu:
This is very unclear.
To try to clarify, you want it in `pdata/pprofile`, but with its own `go.mod`?

mx-psi:
> This is very unclear. To try to clarify, you want it in `pdata/pprofile`, but with its own `go.mod`?

Sorry, yes, that's the preferred way of going about this. There is some prior discussion on #6508 and #4705, in particular see this summary https://github.com/open-telemetry/opentelemetry-collector/issues/6508#issuecomment-1400023599

The idea is to minimize changes in import paths while honoring semver (since profiles are still experimental, we need its related packages to be 0.x)

dmathieu:
Thank you very much for your input 😃 
I've moved it back into `pdata/pprofile`, with its own `go.mod`.

mx-psi (commented):
How are the pcommon slice changes related to the profile? Could we get those in first in a separate PR?

dmathieu:
It's required because the primitive slices struct type only allowed numbers, not strings.
Sure, I'm happy to split this into a set of smaller PRs.

mx-psi:
> Sure, I'm happy to split this into a set of smaller PRs.

:ok_hand:  Thanks, since pcommon is part of a 1.0 module I would like us to consider this changes separately and carefully

petethepig (approved):

mx-psi (commented):
Note: we need to update the README per https://github.com/open-telemetry/opentelemetry-collector/blob/27f59cbdbed9eea9a6b440474de3b78f4b23cd8a/CONTRIBUTING.md?plain=1#L690 on this PR

mx-psi:
#10155 caused a conflict

dmathieu:
Yes, but the commit history isn't the same between my extracted PRs and this one. So I will be opening a new one with the actual pdata changes.

dmathieu:
Closing this in favor of the last step PR, #10195.

PR #10121, opened 2024-05-08
[chore][docs/rfc] Add RFC on configuring confmap Providers

evan-bradley:
#### Description

Documents how we want to configure confmap Providers, with the default decision being to use URI fragments.

Per the SIG discussion on 2024-05-08, I have included as many alternatives as I could think of. I think a few of them have merit, but I don't think any of the options are without downsides.

### DIFF
--- a/docs/rfcs/confmap-provider-configuration.md
+++ b/docs/rfcs/confmap-provider-configuration.md
@@ -49,20 +49,28 @@ requirements:
 ## Solutions
 
 Providers are invoked through passing `--config` flags to the Collector binary
-with a scheme and URI to obtain. A single instance of each Provider is then
-tasked with retrieving from config from all URIs passed to the Collector for the
-scheme the provider is registered to handle.
+or by using the braces syntax inside a Collector config file (`${scheme:uri}`).
+Each invocation contains a scheme specifying how to obtain config and URI
+specifying the config to be obtained. A single instance of a Provider is created
+for each scheme and is tasked with retrieving config for its scheme for each
+corresponding URI passed to the Collector.
 
-We have the following places where it may make sense to hook into options for
-Providers:
+With the above in mind, we have the following places where it may make sense to
+support specifying options for Providers:
 
 1. Parts of the URI we are requesting.
-2. Part of the string passed to `--config`.
+2. Part of the string passed to `--config` or inside braces in a config file.
 3. Separate flags to configure Providers per config URI.
 4. Use a separate config file that specifies config sources inside a map
    structure.
 5. Extend the Collector's config schema to support specifying additional places
-   to obtain configuration
+   to obtain configuration.
+
+All of the above options are targeted toward configuring how specific URIs are
+resolved into config. To configure how a Provider resolves every URI it
+receives, we should consider how to extend the above options to be specified
+without a URI and to ensure the options are always applied to all URI
+resolutions.
 
 ### Configure options inside the URI
 
@@ -98,7 +106,7 @@ Advantages:
 
 - Not likely to be used by config backends for any of our supported protocols,
   so has a low chance of conflict.
-- Fits into existing config URIs.
+- Fits into existing config URIs for URL-based Providers.
 
 Disadvantages:
 
@@ -106,6 +114,8 @@ Disadvantages:
   their use in upstream Providers.
 - Doesn't conform to the spirit of how fragments should be used according to RFC
   3986.
+- Will not work with Providers like the env Provider. This is likely minor, as
+  the env Provider can be expected to work differently than URL-based Providers.
 - Only allows easily specifying key-value pairs.
 
 We could likely partially circumvent the key-value pair limitation by
@@ -175,6 +185,14 @@ Disadvantages:
 - Using the existing bracket syntax inside config files does not support
   options, and neither does specifying them through the command line.
 - Complicates the config schema and the config resolution process.
+
+## Exceptions
+
+None of these approaches are proposed to be the only way Providers can be
+configured, but as a suggested unified approach for Providers that handle URLs
+with hierarchical structure. The notable exception within the current set of
+default Providers is the env Provider, which users will expect to behave with
+shell semantics instead of parsing environment variables as URIs.
 
 ## Resolution
 
--- a/docs/rfcs/confmap-provider-configuration.md
+++ b/docs/rfcs/confmap-provider-configuration.md
@@ -45,6 +45,8 @@ requirements:
 2. Can be used as a suggestion for configuring custom providers.
 3. Is consistent between Providers.
 4. Is configurable per config URI.
+5. All valid URIs are still expressible, even if an escaping mechanism is
+   required.
 
 ## Solutions
 
@@ -116,6 +118,7 @@ Advantages:
 
 - Not likely to be used by config backends for any of our supported protocols,
   so has a low chance of conflict.
+- Could be escaped using an invalid fragment character according to RFC 3986.
 - Fits into existing config URIs for URL-based Providers.
 
 Disadvantages:
--- a/docs/rfcs/confmap-provider-configuration.md
+++ b/docs/rfcs/confmap-provider-configuration.md
@@ -81,6 +81,11 @@ fragments.
 
 #### Queries
 
+Breaking changes:
+
+- confmap Providers would have breaking changes since they would now consume URI
+  queries. There would be no breaking changes to the confmap API.
+
 Advantages:
 
 - Explicitly intended to specify non-hierarchical data in a URI.
@@ -101,6 +106,11 @@ Disadvantages:
 
 We could specify options in a query parameter-encoded string placed into the URI
 fragment.
+
+Breaking changes:
+
+- confmap Providers would have breaking changes since they would now consume
+  fragments. There would be no breaking changes to the confmap API.
 
 Advantages:
 
@@ -139,11 +149,17 @@ would require we adopt an unconventional format.
 
 ### Separate flags to configure Providers per config URI
 
+Breaking changes:
+
+- Will need factory options if we provide config through a mechanism similar to
+  `component.Factory`, along with making a Provider instance per URI.
+- Otherwise will need to break `confmap.Provider` interface to support providing
+  options in `Retrieve`.
+
 Advantages:
 
 - Allows us to keep config URIs opaque.
 - Options live right next to config URIs on the command line.
-- Does not require breaking changes.
 
 Disadvantages:
 
@@ -158,12 +174,18 @@ Disadvantages:
 We could allow specifying a config file that contains all URIs to obtain the
 Collector's configuration from.
 
+Breaking changes:
+
+- Will need factory options if we provide config through a mechanism similar to
+  `component.Factory`, along with making a Provider instance per URI.
+- Otherwise will need to break `confmap.Provider` interface to support providing
+  options in `Retrieve`.
+
 Advantages:
 
 - Allows us to keep URIs opaque.
 - Map structures are easier to work with than command-line arguments for complex
   config.
-- Shouldn't require any breaking changes.
 
 Disadvantages:
 
@@ -176,6 +198,13 @@ This is a variant of providing a separate config source-only configuration file
 that instead puts those URIs and their options inside the main configuration
 file.
 
+Breaking changes:
+
+- Will need factory options if we provide config through a mechanism similar to
+  `component.Factory`, along with making a Provider instance per URI.
+- Otherwise will need to break `confmap.Provider` interface to support providing
+  options in `Retrieve`.
+
 Advantages:
 
 - Compared to the other option, keeps a single configuration file.
@@ -187,6 +216,19 @@ Disadvantages:
 - Using the existing bracket syntax inside config files does not support
   options, and neither does specifying them through the command line.
 - Complicates the config schema and the config resolution process.
+
+## Per-Provider options
+
+Configuring options for all URI resolutions done by a Provider may be useful for
+cases like providing HTTP(s) options or default behavior for env var resolution.
+To configure Providers for all URIs, we would need to take one of the following
+approaches:
+
+- Config flags that are namespaced per-provider.
+- Put options into a config file.
+
+Both approaches share the breaking changes, advantages, and disadvantages with
+the corresponding approaches for per-URI configuration described above.
 
 ## Exceptions
 
--- a/docs/rfcs/confmap-provider-configuration.md
+++ b/docs/rfcs/confmap-provider-configuration.md
@@ -12,11 +12,18 @@ the behavior of how the source is obtained.
 For example, consider the case where the Collector obtains configuration over
 HTTP from an HTTP endpoint. A user may want to:
 
-1. Poll the HTTP endpoint for configuration and reload the Collector service if
-   the configuration changes.
-2. Change the polling interval if polling has been enabled.
-3. Authenticate the request to get configuration by including a header in the
+1. Poll the HTTP endpoint for configuration at a configurable interval and
+   reload the Collector service if the configuration changes.
+2. Authenticate the request to get configuration by including a header in the
    request.
+
+This would produce a set of options like the following:
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Desired state
+
+We would like a mechanism to configure Providers that meets the following
+requirements:
+
+1. Broadly applicable to all current upstream providers.
+2. Can be used as a suggestion for configuring custom providers.
+3. Is consistent between Providers.
+4. Is configurable per config URI.
```
mx-psi, line 47: Are there things we want to configure per-provider? i.e. is there a need for a 'configure this for all invocations of a given provider' vs a 'configure this for this specific usage of the provider'? I am thinking about something like the TLS certificate bundle you want to use for the HTTPS provider. To be clear, for this particular case, we have ways to 'configure' that today (install your extra certificate at the system level, handle HTTPS outside of the Collector...), but I am wondering if this pattern exists outside of it

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Solutions
+
+Providers are invoked through passing `--config` flags to the Collector binary
+with a scheme and URI to obtain. A single instance of each Provider is then
+tasked with retrieving from config from all URIs passed to the Collector for the
+scheme the provider is registered to handle.
```
mx-psi, line 54: We can also use providers within a configuration file, right? We need to consider both cases

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3), which
+specifies the format of a URI, specifies the different parts of a URI and
+suggests two places where we could pass options to Providers: queries and
+fragments.
+
+#### Queries
+
+Advantages:
+
+- Explicitly intended to  specify non-hierarchical data in a URI.
```
mx-psi, line 78:
```suggestion
- Explicitly intended to specify non-hierarchical data in a URI.
```

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+- Often used for this purpose.
+- Fits into existing config URIs.
```
mx-psi, line 80: Not sure how well it fits with the `env` provider, I would expect people to want to use Bash-like syntax or Powershell-like syntax there

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+  3986.
+- Only allows easily specifying key-value pairs.
+
+We could likely partially circumvent the key-value pair limitation by
+recursively calling confmap Providers to resolve files, env vars, HTTP URLs,
+etc. For example:
+
+```text
+https://config.com/config#refresh-interval=env:REFRESH_INTERVAL&headers=file:headers.yaml
+```
```
mx-psi, line 117: You can do this since #7504 with the following syntax:
```text
https://config.com/config#refresh-interval=${env:REFRESH_INTERVAL}&headers=${file:headers.yaml}
```

At least, within a configuration file, I am not sure about CLI

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Add to the string passed to the config flag
+
+Instead of configuring Providers inside their URIs, we could add an extra
+section to the string passed to `--config` and configure how a URI is retrieved
+through that config. This approach is undesirable because the config flag
+currently only takes valid URIs, and updating the format to accommodate this
+would require we adopt an unconventional format.
+
+### Separate flags to configure Providers per config URI
```
mx-psi, line 130: I am not entirely opposed to allow configuring things via flags, but if we ever do this I would do that with a `flag` provider that  you can call like `${flag:my-flag-name}` and it would give you the (YAML-parsed?) value of `--my-flag-name=val` (or possibly `--some-namespace-here-my-flag-name`).

This would have to piggyback onto some other solution that allows passing this `${flag:my-flag-name}` value though

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,184 @@
...
+Disadvantages:
+
+- There are now two ways to include config inside a Collector config file.
+- Using the existing bracket syntax inside config files does not support
+  options, and neither does specifying them through the command line.
+- Complicates the config schema and the config resolution process.
+
+## Resolution
+
+We will configure providers through URI fragments. These are seldom used in
```
mx-psi, line 181: We may want to explicitly exclude the `env` provider from here

docs/rfcs/confmap-provider-configuration.md
evan-bradley: This is probably something we should at least consider. I think it would be reasonable to allow users to specify options that will apply to call URI resolutions instead of having to configure every resolution individually. In addition to the case you called out, I can think of a couple cases where we may want this:

* HTTP(S) provider: Configure any client settings: timeouts, certificates, headers, etc.
* env provider: Behavior for when env vars aren't set

docs/rfcs/confmap-provider-configuration.md
evan-bradley: Good point, I added a note to make this clear.

docs/rfcs/confmap-provider-configuration.md
evan-bradley: I think the braces syntax only applies to URIs in files and doesn't work in the CLI. The reason I simplified the syntax a little was to avoid requiring users to escape characters that a shell may consume like `$`. On the other hand, we need some obvious way to indicate that these are URIs for extra config and not values in themselves. I'm not sure the way I have it written right now will suffice for that.

We also need to consider the way the substitution works for these options. If we have the confmap Resolver do resolution for embedded config URIs, `headers.yaml` would need to be marshaled into a single line, such that this:

```text
https://config.com/config#refresh-interval=${env:REFRESH_INTERVAL}&headers=${file:headers.yaml}
```

would resolve to this:

```text
https://config.com/config#refresh-interval=10s&headers=['Api-Token':'0xdeadbeef' 'Accept-Encoding':'text/yaml']
```

The Resolver would need to know when to "stringify" files like this vs. when to substitute the direct contents of a file. Or we just require users to write the stringified version, which simplifies things for us at the cost of worse UX since in my opinion this representation would be unpleasant to work with.

My assumption was that we implement custom unmarshaling logic for the URI fragment options that takes all of this into account. So `${file:headers.yaml}` is unmarshaled directly into a map that is assigned to a `Headers` field on a Provider's config struct.

docs/rfcs/confmap-provider-configuration.md
evan-bradley: If we were to take this approach, how could we determine which flags are valid? It sounds like flag names could be arbitrary, so would passing `--my-unsupported-flag` cause an error, or just store a value in the flag provider that goes unused? Or am I entirely misunderstanding your suggestion?

docs/rfcs/confmap-provider-configuration.md
evan-bradley: I called this out in an "Exceptions" section above, let me know if that covers it.

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,204 @@
...
+
+Disadvantages:
+
+- The flags would need to be placed in a certain position in the arguments list
+  to specify which URI they apply to.
+- Configuring URIs present in files requires users to look in two places for
+  each URI and is suboptimal UX.
+- Complicating the flags like this would be suboptimal UX.
+
+### Separate config file
```
TylerHelmuth, line 156: I haven't gotten to read this whole RFC yet, but I very much want to avoid this option as I think it will be pretty confusing.

TylerHelmuth (commented):
Can the RFC include any prior art on this type of configuration? My guess is that it'll be cli flags that apply to all Providers.

docs/rfcs/confmap-provider-configuration.md
evan-bradley: I agree, a config file to configure a config file isn't ideal.

evan-bradley:
>Can the RFC include any prior art on this type of configuration?

I spent some time thinking about this, but I'm not aware of anything that matches what we're doing closely enough to make sense. The best I can think of would be infrastructure as code tools like Terraform, Pulumi, or CloudFormation, but I think what they are doing is sufficiently different that it's hard to use them as a basis. If there are other tools or approaches you think would be worth investigating, let me know and I can take a look.

I can also do a writeup on how our configuration fits into the larger landscape of infrastructure configuration approaches if you think it would be beneficial to take a broader look at our whole approach, but didn't want to detract from the more practical discussion at hand of how to do things like authenticate requests in our HTTP providers.

docs/rfcs/confmap-provider-configuration.md
evan-bradley: I've added a section on this.

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Desired state
+
+We would like a mechanism to configure Providers that meets the following
+requirements:
+
+1. Broadly applicable to all current upstream providers.
+2. Can be used as a suggestion for configuring custom providers.
+3. Is consistent between Providers.
+4. Is configurable per config URI.
```
mx-psi, line 47: I feel like a requirement should be that any valid URI today is expressible after this change (possibly with some escaping mechanism)

docs/rfcs/confmap-provider-configuration.md
mx-psi: > The reason I simplified the syntax a little was to avoid requiring users to escape characters that a shell may consume like $.

I am a bit hesitant about dropping `$` both for consistency and because `#refresh-interval={env:REFRESH_INTERVAL}` is (I think?) a valid part of an URI, so I wouldn't want to prevent users from writing that.

> The Resolver would need to know when to "stringify" files like this vs. when to substitute the direct contents of a file. 

My expectation is that this would work the same way it does if you use an `${...}` syntax in the middle of a string in the actual config.

docs/rfcs/confmap-provider-configuration.md
mx-psi: > If we were to take this approach, how could we determine which flags are valid? It sounds like flag names could be arbitrary, so would passing --my-unsupported-flag cause an error, or just store a value in the flag provider that goes unused? Or am I entirely misunderstanding your suggestion?

I was thinking about requiring the flags to be under some namespace (e.g. `--flag-provider-my-flag-name`). I am not sure what the behavior should be for unused flags under that namespace (ideally an error at config resolution time if they go unused).

docs/rfcs/confmap-provider-configuration.md
```diff
@@ -0,0 +1,246 @@
...
+Using this strategy would also allow us to more easily get env vars and to get
+values from files for things like API tokens.
+
+### Add to the string passed to the config flag
+
+Instead of configuring Providers inside their URIs, we could add an extra
+section to the string passed to `--config` and configure how a URI is retrieved
+through that config. This approach is undesirable because the config flag
+currently only takes valid URIs, and updating the format to accommodate this
+would require we adopt an unconventional format.
```
mx-psi, line 148: The more I read, the more I would prefer something like this: no conflict, it's clear what it means, you can configure per URI... This seems easy enough in the config itself (e.g. we could have `${http://example.com}[refresh-interval=10s]` or something like this), but I don't see now how to put that into practice for the CLI right now :thinking:

docs/rfcs/confmap-provider-configuration.md
evan-bradley: >I am a bit hesitant about dropping $ both for consistency and because #refresh-interval={env:REFRESH_INTERVAL} is (I think?) a valid part of an URI, so I wouldn't want to prevent users from writing that.

It would be a valid part of the URI, but the assumption here is that we reserve the fragment portion of URIs for our own use in upstream confmap components. Custom components would be free to use this portion of the URI as they wish, including forwarding it to a backend. I'm fine either way with whether recursive resolution requires `$`.

>My expectation is that this would work the same way it does if you use an ${...} syntax in the middle of a string in the actual config.

I'm alright with that, that's probably the least surprising behavior here.

docs/rfcs/confmap-provider-configuration.md
evan-bradley: Thanks for explaining.

The way I would see this working right now is to make config flags order-sensitive, like so:

```
--config=http://myurl.com
--flag-provider-refresh-interval=10s
--flag-provider-auth-header=${env:MYPASS}
--config=/path/to/file.yaml
[...]
```

We would then need to give the Provider's `Retrieve` method some kind of handle to obtain these flag values from. Do you think this would be viable, or do you have another solution you think may be better here?

docs/rfcs/confmap-provider-configuration.md
evan-bradley: Agreed. It looks like something like `##` wouldn't be a valid normal URI fragment [according to RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#appendix-A), so that may be a good default for at least the URI option. I think the other options won't need to mess with the URI.

docs/rfcs/confmap-provider-configuration.md
evan-bradley: I would see it work like this for each format:

File:

```
val: ${http://example.com}[refresh-interval=10s]
```

CLI:

```
--config=http://example.com
--resolution-options=refresh-interval=10s&auth-header=Api-Token
```

If we don't do a config flag, we'll have to find some character that's not supported [here](https://www.rfc-editor.org/rfc/rfc3986#appendix-A) and use it as a delimiter between the URI and options. Looking at the `gen-delims` and `sub-delims` character classes, `|`, `\`, `{`, `}`, `^`, and `` ` `` could be options.

mx-psi (commented):
If we were to design this from scratch, something like `${<provider>:URI}[params]` would be my desired syntax, but I am worried that this would be too much of a breakage to do today (the recent confmap changes show that people do actively use a lot of edge cases).

I think the next best solution is URI fragments. My main hesitation there is that, while this works for HTTP(s) providers (it's a client-side only feature), this may not work for other URIs (e.g. it's a valid character on file names, other systems may use fragments for something meaningful). However, I feel like this breakage is much smaller for the currently existing providers (`#` is not a common character in filenames). I would like to propose the following plan:

1. We rework this RFC to propose fragments as the solution, with no mention about what the specific syntax is (e.g. we table the adding `$` or not adding it). We add an explicit escaping mechanism for `#`.
2. For all providers that currently allow `#` in the URIs (possibly others in contrib), we implement `#` escaping and start logging a warning if an unescaped `#` is present.
3. After a few releases we make having a `#` in a URI a hard error.

If we agree on this plan and the PR is merged with these changes, this unblocks `confmap` 1.0 (we would have to wait to make the specific providers 1.0 until the cycle above is finished).

Whenever we want to add the first parameter we will have to restart some of the discussions here, but this can be done after 1.0 is completed.

TylerHelmuth:
Is there an option to support a `${<provider>:URI}[params]` style if we introduce a new specialized flag? Like instead of `--config` its `--fancy-config`? That way we get the best syntax without breaking users?

evan-bradley:
I think at that point we should just have an alpha-stability feature gate that we use to determine whether params are supported. I see the benefit there, but I think it would also be confusing to have to opt-in per-file. In cases where I forget to use `--fancy-config`, I would get weird syntax errors (or worse, no error at all) without a clear indicator of why.

I'm also a bit nervous about committing long-term to maintaining two ways to parse the config.

evan-bradley:
This is replaced by https://github.com/open-telemetry/opentelemetry-collector/pull/10776.

PR #10127, opened 2024-05-09
ExporterHelper tests - switch to DataType instead of Type

ankitpatel96:
#### Description
`newBaseExporter` needs a `DataType` - up until now we've been passing in a `Type`.  In preparation of splitting `DataType` and `Type`, pass in a real `DataType`.

#### Link to tracking issue
related to https://github.com/open-telemetry/opentelemetry-collector/issues/9429

In preparation for https://github.com/open-telemetry/opentelemetry-collector/pull/10069

PR #10128, opened 2024-05-09
MemoryLimiterProcessor - switch to  MustNewID instead of zero value ID

ankitpatel96:
#### Description
In an upcoming PR, I change Type to be an interface.  This means the zero value of Type will be nil - which will cause this test to fail.  Initializing ID instead of relying on the zero value fixes this

#### Link to tracking issue
related to https://github.com/open-telemetry/opentelemetry-collector/issues/9429

In preparation for https://github.com/open-telemetry/opentelemetry-collector/pull/10069


mx-psi (approved):

mx-psi:
Same comment applies here! https://github.com/open-telemetry/opentelemetry-collector/pull/10127#issuecomment-2102951647

PR #10136, opened 2024-05-10
[confighttp] Don't marshal CustomRoundTripper

evan-bradley:
#### Description

While trying to marshal a `confmap.Conf` created using `Conf.Marshal(*otelcol.Config)`, I noticed that calling `yaml.Marshal` on the `confmap.Conf` throws an error because it tries to marshal the `CustomRoundTripper` field. I'm not sure how we enforce this, but I think we should ensure all our configs can be re-marshaled.

#### Testing

I added a test that shows the issue. I would have preferred to require that configs can be unmarshaled into something more generic than YAML, but I think YAML is a good default.

I can issue follow-up PRs for other `config` packages if this one looks good. I don't know how we enforce this more broadly within our components, but I would like it for our unmarshaling and marshaling to be largely lossless operations after things like `Config.Unmarshal` and `Config.Validate` are called the first time. That is, we should be able to go YAML [user] -> otelcol.Config -> YAML [0] -> otelcol.Config -> YAML [1] and have YAML 0 and YAML 1 be identical.

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -57,7 +57,7 @@ type ClientConfig struct {
 	Headers map[string]configopaque.String `mapstructure:"headers"`
 
 	// Custom Round Tripper to allow for individual components to intercept HTTP requests
-	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
+	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error) `mapstructure:"-"`
 
 	// Auth configuration for outgoing HTTP calls.
 	Auth *configauth.Authentication `mapstructure:"auth"`
--- a/config/confighttp/go.mod
+++ b/config/confighttp/go.mod
@@ -15,12 +15,14 @@ require (
 	go.opentelemetry.io/collector/config/configtelemetry v0.100.0
 	go.opentelemetry.io/collector/config/configtls v0.100.0
 	go.opentelemetry.io/collector/config/internal v0.100.0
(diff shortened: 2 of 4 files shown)


config/confighttp/go.mod
evan-bradley, line 1: I don't like the direct dependencies on confmap or yaml here, but these were already in the dependency tree, and we do essentially depend on both (confmap through the `mapstructure` tag and yaml because we should be able to marshal into YAML).

BinaryFissionGames (approved):

config/confighttp/confighttp.go
```diff
@@ -57,7 +57,7 @@ type ClientConfig struct {
 	Headers map[string]configopaque.String `mapstructure:"headers"`
 
 	// Custom Round Tripper to allow for individual components to intercept HTTP requests
-	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error)
+	CustomRoundTripper func(next http.RoundTripper) (http.RoundTripper, error) `mapstructure:"-"`
```
evan-bradley, line 60: Directly calling `yaml.Marshal` on `ClientConfig` will still fail, this only helps when we go through confmap.

mx-psi:
Related issue: #8627

evan-bradley:
Closing this per the 2024-05-15 SIG discussion:
1. I will update `ClientConfig` to remove this field.
2. We will add tests to mdatagen to enforce this for all components going forward. I'll see how we could extend this requirement to our config packages.

PR #10139, opened 2024-05-10
[otelcol] Obtain the Collector's effective configuration from `otelcol.Config`

evan-bradley:
#### Description

The `otelcol.ConfmapProvider` interface accurately reports the config provided to the Collector by the user, but fails to effectively report the Collector's effective configuration. In particular, it misses:
* Default values for fields in Config structs.
* Transformations done to Config structs by their `Unmarshal` or `Validate` methods.
* Custom marshaling of types after we know the type of the config. This is most obvious with `configopaque.String`, where we want these values to always be redacted when sent out of the Collector.

I think we should attempt to get the Collector's effective configuration from `otelcol.Config` instead of using the map compiled by the confmap.Resolver. This also allows us to get rid of `otelcol.ConfmapProvider`.

I have manually tested this using an OpAMP Supervisor setup, but wasn't able to add unit tests due to the way `otelcol.Collector` is written.


evan-bradley:
To demonstrate the issue I'm seeing, this will fail:

```golang
cfg, _ := col.configProvider.Get(ctx, factories)
conf := confmap.New()
_ = conf.Marshal(cfg)
// This fails
bytes, err := yaml.Marshal(conf.ToStringMap())
```

This will succeed:

```golang
cfg, _ := col.configProvider.Get(ctx, factories)
// This succeeds
bytes, err := yaml.Marshal(cfg)
```

I'm fine just letting `ConfigWatcher.NotifyConfig` take an opaque `any`-typed parameter and telling the user to just marshal it with whatever format they want, but I don't fully understand why the first one fails. The second one also appears to make us more dependent on `yaml` since we don't have mapstructure as a buffer (e.g. we need to add `yaml:"-"` to `confighttp.ClientConfig` for the second snippet to work).

extension/extension.go
```diff
@@ -48,7 +47,7 @@ type ConfigWatcher interface {
 	// The extension owns the `confmap.Conf`. Callers must ensure that it's safe for
 	// extensions to store the `conf` pointer and use it concurrently with any other
 	// instances of `conf`.
-	NotifyConfig(ctx context.Context, conf *confmap.Conf) error
+	NotifyConfig(ctx context.Context, data any) error
```
sfc-gh-bdrutu, line 50: Please avoid "any" in public APIs as much as possible. This is usually a sign of a poorly designed API.

sfc-gh-bdrutu:
>As a result, I think we should attempt to get the Collector's effective configuration from otelcol.Config instead of using the map compiled by the confmap.Resolver. I initially intended to generate a confmap.Conf from otelcol.Config and call yaml.Marshal on that, but this encounters errors such as being unable to [marshal Prometheus config](https://github.com/prometheus/common/blob/main/config/http_config.go#L85) which has invalid zero values. These errors don't occur when calling yaml.Marshal on otelcol.Config directly. I've updated and tested updating the ConfigWatcher interface to just accept an opaque any-typed data object and marshal that, and the end-to-end flow works. I haven't dug into this far enough to fully understand the differences between unmarshaling each type.

I think you should look into fixing marshal. /cc @atoulme

extension/extension.go
evan-bradley: Agreed, this was mostly just done as a PoC. We agreed to use `confmap.Conf` during this week's SIG meeting (which I'd prefer to do anyway), so I'll be reverting this.

evan-bradley:
>I think you should look into fixing marshal.

Thanks for the tip, I think that's also the best path forward. I found that the issue is that we don't do anything with `yaml` tags, which is what most third-party structs annotate on their fields. I will issue a PR to find a solution to that problem then update this PR afterward.

djaglowski:
> > I think you should look into fixing marshal.
> 
> Thanks for the tip, I think that's also the best path forward. I found that the issue is that we don't do anything with `yaml` tags, which is what most third-party structs annotate on their fields. I will issue a PR to find a solution to that problem then update this PR afterward.

I have a possible solution in https://github.com/open-telemetry/opentelemetry-collector/pull/10282.

Basically it adds a hook which looks for structs that have `yaml` tags but no `mapstructure` tags. For such structs, it uses the yaml package to "natively" convert the struct into `map[string]any`, which can then be marshaled by mapstructure in a generic way.

evan-bradley:
@mackjmr You're right, this will work now; thanks for keeping an eye on it. I haven't been able to get back to this, but I have more time now and will push an update this week.

djaglowski (approved):

otelcol/collector.go
```diff
@@ -189,6 +178,12 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 
 	col.serviceConfig = &cfg.Service
 
+	conf := confmap.New()
+
+	if err = conf.Marshal(cfg); err != nil {
+		return fmt.Errorf("could not marshal configuration: %w", err)
```
evan-bradley, line 184: Another option could be to make this only a warning log, if we're comfortable that all extensions asking for the effective configuration will get an empty map.

otelcol/configprovider_test.go
```diff
@@ -108,106 +106,3 @@ func TestConfigProviderFile(t *testing.T) {
...
-	cmp, ok := cp.(ConfmapProvider)
-	require.True(t, ok)
-
-	cmap, err := cmp.GetConfmap(context.Background())
-	require.NoError(t, err)
-
-	assert.EqualValues(t, yamlMap, cmap.ToStringMap())
-}
-
-func TestStrictlyTypedCoda(t *testing.T) {
```
mx-psi, line 143: I don't think this test should be deleted

otelcol/configprovider.go
```diff
@@ -142,12 +144,3 @@ func (cm *configProvider) Watch() <-chan error {
 func (cm *configProvider) Shutdown(ctx context.Context) error {
 	return cm.mapResolver.Shutdown(ctx)
 }
-
-func (cm *configProvider) GetConfmap(ctx context.Context) (*confmap.Conf, error) {
```
mx-psi, line 146: Should we keep a dummy implementation of this method for one or more releases that returns an error saying that this functionality is deprecated?

otelcol/configprovider_test.go
evan-bradley: Thanks for catching that, I must have messed up the merge. It's back now.

otelcol/configprovider.go
evan-bradley: I removed this because I don't expect anyone else is using it, but thinking on it, this should go through the normal deprecation process. I've added a deprecation notice to this method and we can just remove the implementation when we remove the interface. I would feel comfortable removing both one release after deprecating them.

mx-psi (approved):

yurishkuro:
is this purely internal change, or can the end user print the config now?

evan-bradley:
This only affects the Collector APIs, but it is a step towards allowing the user to print the config.

Issue #8215, opened 2023-08-09
Password containing `$` doesn't work when passed via `${env:DB_PASSWORD}`


wand2016:
### Component(s)

receiver/postgresql

### What happened?

Password containing `$` doesn't work when passed via `${env:DB_PASSWORD}`.
When passed directly, it works.

## Steps to Reproduce

1. Prepare PostgreSQL DB:
   - username: `user`
   - password: `p@s$w0rd`
2. Configure postgresql receiver:

```yaml
  postgresql:
    ...
    username: "${env:DB_USERNAME}"
    password: "${env:DB_PASSWORD}"
```

3. Pass environment variables to OpenTelemetry CR:

```yaml
apiVersion: opentelemetry.io/v1alpha1
kind: OpenTelemetryCollector
...
spec:
  config: <configuration mentioned above>
  env:
    - name: DB_USERNAME
      value: user
    - name: DB_PASSWORD
      value: p@s$w0rd
```

## Expected Result

Authentication passes.

## Actual Result

Authentication fails.

Given password directly, authentication passes.

```yaml
  postgresql:
    ...
    username: "${env:DB_USERNAME}"
    password: "p@s$$w0rd"
```

### Collector version

otel/opentelemetry-collector-contrib:0.81.0

## Environment

contrib Docker Image

mx-psi:
I moved this to opentelemetry-collector since this is not specific to the postgresql receiver, but rather applies to configuration more generally. 

I think this would not happen if the [expandconverter](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/converter/expandconverter) was not enabled. Since this is a legacy component that we want to remove in favor of `${env:...}` syntax, we could add a feature gate to disable the expand converter and keep it at alpha for now.

@open-telemetry/collector-maintainers what do you think? Would you be open to such feature gate?

atoulme:
Discussed in SIG 12/20, we will work towards deprecating the naked `$` notation and support `${...}` instead, and therefore avoid this type of bug moving forward.

See https://github.com/open-telemetry/opentelemetry-specification/pull/3744 for the spec change.

As part of this, we will probably deprecate the `$` expand, as noted by Pablo, we can't really log a warning right now when this happens because the expansion happens before logging is set up. We can look into that.

TylerHelmuth:
We've made progress toward this issue.  The collector now:

1. supports logging during confmap resolution
2. prints a log if `$ENV` syntax is used.

[...]

Issue #7111, opened 2023-02-03
Remove expandconverter


bogdandrutu:
Need to investigate if the current expand providers logic in the [`confmap.Resolver`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/resolver.go#L168) has all the equivalent functionality. Then we also need to propose some kind of deprecation process: e.g. disable by default via a featuregate; or other alternatives.

One caveat that I know and need to document is that because of this we have a workaround in prometheus receiver see the note [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#getting-started). Once this is disable that is no longer necessary which is a good thing, but may break some users so we need to find a way to handle/communicate that.

mx-psi:
> but may break some users so we need to find a way to handle/communicate that.

We may want to have some sort of subcommand/flag/separate tool to migrate a config from the old syntax to the new one. This could be useful for other future changes in the configuration (whether breaking or just recommendations). If that sounds like a good idea, should we create a separate issue to discuss how to go about it?

bogdandrutu:
Let's start that issue. We definitely need it. I think the best we can do is to have it as a subcommand but this is just me prematurely commenting 👯

mx-psi:
See #7631 :)

mx-psi:
Per https://github.com/open-telemetry/opentelemetry-collector/issues/8215#issuecomment-1865028717 before doing this we need to add support for `${...}` to the provider, to align with https://github.com/open-telemetry/opentelemetry-specification/pull/3744

PR #10144, opened 2024-05-13
[confmap] Start deprecation processor for env var expand syntax ($ENV)

TylerHelmuth:
#### Description
Starts the deprecation for support of env var expand syntax ($ENV) by adding a feature gate and a deprecation warning

#### Link to tracking issue

#### Testing
Local testing

### DIFF
--- a/otelcol/configprovider.go
+++ b/otelcol/configprovider.go
@@ -14,6 +14,16 @@ import (
 	"go.opentelemetry.io/collector/confmap/provider/httpprovider"
 	"go.opentelemetry.io/collector/confmap/provider/httpsprovider"
 	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
+	"go.opentelemetry.io/collector/featuregate"
+)
+
+// AllowEnvVarExpansionFeatureGate is the feature gate that controls whether the collector
+// supports configuring the OpenTelemetry SDK via configuration
+var AllowEnvVarExpansionFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"confmap.allowEnvVarExpansion",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("controls whether the collector supports expanding $ENV in configuration"),
+	featuregate.WithRegisterFromVersion("v0.101.0"),
 )
 
 // ConfigProvider provides the service configuration.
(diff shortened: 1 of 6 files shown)


otelcol/configprovider.go
```diff
@@ -14,6 +14,16 @@ import (
 	"go.opentelemetry.io/collector/confmap/provider/httpprovider"
 	"go.opentelemetry.io/collector/confmap/provider/httpsprovider"
 	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
+	"go.opentelemetry.io/collector/featuregate"
+)
+
+// PreventEnvVarExpansionFeatureGate is the feature gate that controls whether the collector
+// supports configuring the OpenTelemetry SDK via configuration
+var PreventEnvVarExpansionFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"confmap.preventEnvVarExpansion",
```
evan-bradley, line 23:
```suggestion
	"confmap.NoBracelessEnvVars",
```
I think we should refine the name a bit, since we will still allow env var expansion even after the deprecation process. We could also say `NoNakedEnvVars`, but I think "braceless" is a bit clearer to someone who isn't familiar with our terminology.

otelcol/configprovider.go
```diff
@@ -14,6 +14,16 @@ import (
 	"go.opentelemetry.io/collector/confmap/provider/httpprovider"
 	"go.opentelemetry.io/collector/confmap/provider/httpsprovider"
 	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
+	"go.opentelemetry.io/collector/featuregate"
+)
+
+// PreventEnvVarExpansionFeatureGate is the feature gate that controls whether the collector
+// supports configuring the OpenTelemetry SDK via configuration
```
evan-bradley, line 21:
```suggestion
// PreventEnvVarExpansionFeatureGate disables the ability to use environment variables
// without wrapping the variable in braces.
```
I don't think we need to explicitly call out SDK configuration here; I also don't think this has to do with the Collector's SDK instance.

TylerHelmuth: Looks like I had a bad copy-paste

otelcol/configprovider.go
mx-psi: Should we have a single feature gate for all changes related to the environment variable revamp? Or should we have one for each change? I am inclined to say we should have one (easier to figure out how to revert back to the old behavior without having to understand its intricacies)

TylerHelmuth: What is another change? I thought `$ENV` was the only thing we were deprecating

TylerHelmuth:
I opened this before seeing https://github.com/open-telemetry/opentelemetry-collector/issues/7111, looks like we need to do some work to allow `${}` without of `expandconverter`.  Closing this for now, I'll work on https://github.com/open-telemetry/opentelemetry-collector/issues/7111 instead.

otelcol/configprovider.go
mx-psi: Making types stricter and unified between `${ENV}` and `${env:ENV}` is the other part

PR #10148, opened 2024-05-14
Add stringSlice and int64Slice to pcommon, and let tests handle non-number slices

dmathieu:
#### Description

This adds a string slice to pcommon, that will be used by the profiles pdata.

#### Link to tracking issue

Part of #10109.

#### Testing

This is unit-tested.

cc @mx-psi


mx-psi (approved):
I think this should not be a controversial addition, but since this is adding new API on a 1.0 module I would like other @open-telemetry/collector-approvers to take a look.

AIUI the motivation is to support fields like https://github.com/open-telemetry/opentelemetry-proto/blob/a05597bff803d3d9405fcdd1e1fb1f42bed4eb7a/opentelemetry/proto/profiles/v1experimental/pprofextended.proto#L97, so the risk is that we may not use this after all (sounds very unlikely though)

mx-psi:
@dmathieu To confirm: the intent is to use these types for fields like [`string_table`](https://github.com/open-telemetry/opentelemetry-proto/blob/a05597bff803d3d9405fcdd1e1fb1f42bed4eb7a/opentelemetry/proto/profiles/v1experimental/pprofextended.proto#L97) or [`location_indices`](https://github.com/open-telemetry/opentelemetry-proto/blob/a05597bff803d3d9405fcdd1e1fb1f42bed4eb7a/opentelemetry/proto/profiles/v1experimental/pprofextended.proto#L86), right?

dmathieu:
Yes, it is. My pdata PR shows that use as well.

mx-psi:
Thanks for confirming. I will wait until Friday before merging this to give time to others to review

TylerHelmuth (commented):
Looking at https://github.com/open-telemetry/opentelemetry-collector/pull/10109 I only see the new types being used in `pprofile`.  Is it required that the new types be in `pcommon`?

mx-psi:
> Looking at #10109 I only see the new types being used in `pprofile`. Is it required that the new types be in `pcommon`?

IMO it is required after profiling is declared stable. It's not strictly required now, but it's better in that we don't need people to update their import paths for this after profiling goes to 1.0

TylerHelmuth:
> It's not strictly required now, but it's better in that we don't need people to update their import paths for this after profiling goes to 1.0

I guess the risk is that this goes into 1.x pcommon and then needs removed/modified.  Unlikely, but how would we handle that? Deprecate it and remove in 2.0?

mx-psi:
> Deprecate it and remove in 2.0?

Yup, it would become dead code

TylerHelmuth:
What would we do if it needs to be modified instead of removed?

mx-psi:
> What would we do if it needs to be modified instead of removed?

We would have to deprecate it and come up with a different name. I don't think that's going to happen here though, what a `StringSlice` type should look like is kind of already defined by the other `<Type here>Slice` types we already support

TylerHelmuth (approved):

mx-psi:
Waiting for v0.101.0 to be released, will merge after that

Issue #9209, opened 2024-01-03
Add policy for experimental Go API


mx-psi:
When working on #8935, one of the discussion points was about experimental symbols https://github.com/open-telemetry/opentelemetry-collector/pull/8935#discussion_r1396237074, which we decided to postpone. My proposal is to rely on separate Go modules as much as we can, and fall back to Go build tags whenever that is not feasible.

### Prior work

For previous issues that touch on this topic see #4705 and #4832

- **General discussion**. golang/go/issues/34409 discusses adding an `// Unstable`, `// Experimental` or a similar prefix, and what alternatives are available. After skimming, I think https://github.com/golang/go/issues/34409#issuecomment-553546180 and https://github.com/golang/go/issues/34409#issuecomment-767069905 are particularly interesting comments.
- **opentelemetry-go**. Experimental or unstable APIs have been added 
    - as part of [separate modules with a separate v0.0.x versioning schema]( https://github.com/open-telemetry/opentelemetry-go/blob/e6c50b5924a889a391dacacf47eb37227092005b/versions.yaml#L47-L50)
    - directly to stable modules, with a comment indicating they may change. This is limited to interfaces that may change per the spec and helpers are available to deal with this. See open-telemetry/opentelemetry-go/issues/3277 for some discussion on this.
- **grpc-go**. gRPC Go has been brought up in past discussion about this. Its versioning policy has [experimental symbols](https://github.com/grpc/grpc-go/blob/71cc0f167537667a878ec5995cf8167c69c362df/Documentation/versioning.md#versioning-policy). This is a common source of complaints for users, see https://github.com/open-telemetry/opentelemetry-collector/issues/4832#issuecomment-1039394229 for a compilation of issues.

### Proposed solution

1. Create a new module set[^1] called `experimental` following the versioning schema `v0.0.x`. Parts of the codebase we deem very unstable will be on separate modules under this module set. Symbols in these modules may not follow our usual deprecation guidelines (e.g. API may be removed without notice and we may remove the module after deprecating).
2. For symbols that need to be added to an existing stable or beta module, add them under new build tags. These build tags:
    - Should have a common namespace (e.g. all should start with `otel_collector_`)
    - Should be independent (enabling any subset of available build tags should produce code that can be built and works)
    - Code using experimental symbols under this build tag should produce a working build when the build tag is not present (with possible missing functionality) or produce a meaningful error message (e.g. `panic` with a meaningful message if the build tag is not present). 

[...]

PR #10154, opened 2024-05-15
Adds Profiles Signal

petethepig:
### Description

This is a PR that adds profiling support to collector. Relevant links:
* [OTEP 239 — Introduces Profiling Data Model v2](https://github.com/open-telemetry/oteps/pull/239)

### State of this PR

This is currently work in progress. The code compiles and tests pass, however I'm planning to split this one into smaller chunks to simplify the review process.

TODOs:
* rebase off of https://github.com/open-telemetry/opentelemetry-collector/pull/10109
* double check that I added profiling support to all of the components
* add instructions on how to test this end-to-end
* make sure I properly marked profiles as experimental everywhere

### Testing

Workflow I use to test this right now:

* Run `make otelcorecol` to verify that collector compiles
* Run `./bin/otelcorecol_darwin_arm64 --config=file:tmp/config.yaml` (config is [here](https://gist.github.com/petethepig/418a19b9a5b6f574598a23e93a2303be))
* Run `curl -i -X POST http://localhost:4318/v1experimental/profiles -H 'Content-Type: application/json' -d @profile.json` (profile.json is [here](https://gist.github.com/petethepig/418a19b9a5b6f574598a23e93a2303be))
* ^ This prints the profile in text form

---

My plan is to have an end-to-end pipeline where you can run collector with an example client and example backend to test the whole pipeline. I did this with an earlier version of this PR, but haven't updated it yet. Will do this soon and update PR descriptions with instructions on how to run it.


mx-psi:
It's great to have this PR to have a full picture of what support for profiles entails, thanks for moving this along! 

As we discussed on the last Collector SIG, when we work on adding profiles support, we need to add this in a way that is mindful of the Collector 1.0 effort and can separate experimental APIs related to profiling from the ones we intend to stabilize now. How to do this is an open question (see #9209 for related discussion that we may need to address when doing this), and I am happy to help here to move this along. 

In general, the approach I want to explore first is to put profiling related interfaces, structs and options into separate modules. If something needs to be on the same module as 1.0 stuff (I currently don't see any blockers like this), then we'll need to see if we can refactor things to make it go away or deal with it in the way proposed in #9209.

PR #10158, opened 2024-05-15
Add reason dimension to exporter and receiver failure metrics

### LINKED ISSUE #10157: Add 'reason' attribute to `otelcol_exporter_send_failed_*` metrics
I am interested in monitoring data loss which occurs when exporting data from one instance of the Collector to another, specifically using the [loadbalancingexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter).

At the moment I just see a course grained metric which counts the export failures, but gives me no data on the cause. Was it a permanent or retryable error? Was it a badly configured endpoint, or did the downstream receiver actively reject the data?

I can look into the logs to see info on specific failures, but this is tedious and less easy to understand.

I propose that we add a `reason` dimension to the  `otelcol_exporter_send_failed_*` metrics. This `reason` could be the GRPC status of the response (I understand that GRPC status is uses as the internal representation of these kind of problems).

It is possible to try to correlate export failure metrics with downstream receiver error metrics.  We can also try to correlate with "know failure causes", such as `memorylimiterprocessor` errors, which could mean the upstream export failed.

We can also check the logs, and even - depending on the system - extra metrics from theses logs.

However, this is all much harder work

**Additional context**

We could also consider adding a similar attribute to the `otelcol_receiver_refused_spans` metric.

I have had a look at the code, and it seems like a fairly small change in / around `exporter/exporterhelper/obsexporter.go`

0x006EA1E5:
#### Description
Adds `reason` attribute to `otelcol_exporter_send_failed_*` and `otelcol_receiver_refused_*` metrics

#### Link to tracking issue
Fixes #10157

#### Testing
TODO

#### Documentation
TODO

### DIFF
--- a/exporter/exporterhelper/obsexporter.go
+++ b/exporter/exporterhelper/obsexporter.go
@@ -11,6 +11,7 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/status"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configtelemetry"
@@ -70,7 +71,7 @@ func (or *ObsReport) StartTracesOp(ctx context.Context) context.Context {
 // EndTracesOp completes the export operation that was started with StartTracesOp.
 func (or *ObsReport) EndTracesOp(ctx context.Context, numSpans int, err error) {
 	numSent, numFailedToSend := toNumItems(numSpans, err)
-	or.recordMetrics(noCancellationContext{Context: ctx}, component.DataTypeTraces, numSent, numFailedToSend)
+	or.recordMetrics(noCancellationContext{Context: ctx}, component.DataTypeTraces, numSent, numFailedToSend, err)
 	endSpan(ctx, err, numSent, numFailedToSend, obsmetrics.SentSpansKey, obsmetrics.FailedToSendSpansKey)
 }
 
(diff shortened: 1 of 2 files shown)


bogdandrutu:
In general things that are high cardinality like generic "errors" are not best suited for metrics, and usually they should just be recorded like logs or span attributes.

0x006EA1E5:
> In general things that are high cardinality like generic "errors" are not best suited for metrics, and usually they should just be recorded like logs or span attributes.

The suggestion is to use the GRPC status code, not the actual error text, i.e., https://github.com/grpc/grpc-go/blob/master/codes/codes.go#L37

So the cardinality it ≈17.

Status code is commonly used as a metric dimension, for example for [http metrics](https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/).

And typically, (in my experience of the collector), the actual number of statuses seen in responses will be much lower, so time series will not be generated for most of the possible values.  Actually, in my experience, the error will normally be `UNAVAILABLE`, with a much lower number of `UNKNOWN`, `DEADLINE_EXCEEDED`, and `RESOURCE_EXHAUSTED`, so I wouldn't expect cardinality to increase by so much overall.

I'm also suggesting that we only add this dimension when the telemetry is configured as `LevelDetailed`, so users worried about cardinality have some control here.

bogdandrutu:
> The suggestion is to use the GRPC status code, not the actual error text, i.e., https://github.com/grpc/grpc-go/blob/master/codes/codes.go#L37

Why not accept the code then?

0x006EA1E5:
> Why not accept the code then?

I don't understand. Do you mean use the numeric status code instead of the status code text?

0x006EA1E5:
I don't mind using either the numeric code, or the equivalent name, although I would think the name would be a bit more informative / easier to read.

And as the exporter could be using either GRPC or HTTP (or potentially another protocol), then the GRPC status code number may be a bit confusing.

0x006EA1E5:
Is there anything I can do to progress this?

atoulme:
You can join a SIG meeting, we have one for the collector in 10 minutes. It runs weekly.

codeboten:
There's an otep that discusses additional details around monitoring a telemetry pipeline https://github.com/open-telemetry/oteps/pull/259, might be worth taking a look there as well

Issue #10093, opened 2024-05-06
ctx propagation in otlp/http vs otlp/grpc


grandwizard28:
**Context**
I have written a custom auth extension which derives a key called consumer. I want to batch requests based on consumer by using the batch processor and use the same in a custom exporter.
Issue 1: Batch processor uses the incoming metadata keys to batch. I can set the client.Metadata in my custom auth extension but the otlp/grpc receiver overrides client.Metadata.
Issue 2: For the exporter, I tried setting client.Auth as well in my custom auth extension but the batch processor does not respect it. It creates a new context with only metadata, discarding the Auth (if set) by any authenticators.

**Describe the bug**
I have created a custom server authenticator extension. At the end I am calling the following: 
```
return client.NewContext(ctx, client.Info{
		Metadata: client.NewMetadata(map[string][]string{
		      MetadataKeyId:      {key},
	        }),
	}), nil
```
Basically, I am trying to set some information in the metadata.

Now in one of my subsequent processors, I am trying to access this key. It works fine for http requests but the key is not present for grpc requests.

**Steps to reproduce**

**What did you expect to see?**
Context should be preserved for both otlp/http and otlp/grpc

**What did you see instead?**

**What version did you use?**
v0.99.0

**What config did you use?**
```
extensions:
  customauthextension:
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
        include_metadata: true
        max_recv_msg_size_mib: 16
        auth:
          authenticator: customauthextension
      http:
        cors:
          allowed_origins:
            - '*'
        endpoint: 0.0.0.0:4318
        include_metadata: true
        auth:
          authenticator: customauthextension
exporters:
  logging:
processors:
  customprocessor:
service:
  extensions:
    - customauthextension
  pipelines:
    logs:
      receivers: [otlp]
      processors: [customprocessor, batch]
      exporters: [logging]
```

*** Additional Context ***
I have been trying for a way to propagate context data from my custom authenticator to all processors and exporters for some time now. Here is a related PR regarding the same: https://github.com/open-telemetry/opentelemetry-collector/pull/10002

grandwizard28:
Update:
If I set a custom context key in my authenticator like :
```
return context.WithValue(ctx, mycustomKey, key), nil
```

and then access it with
```
ctx.Value(mycustomKey)
```
in my processor, it worksss!!

grandwizard28:
[...]

PR #10162, opened 2024-05-15
test(configgrpc|confighttp): sample test showcasing issue 10093

grandwizard28:
#### Description
This is a sample test showcasing issue https://github.com/open-telemetry/opentelemetry-collector/issues/10093

### DIFF
--- a/config/configgrpc/configgrpc_test.go
+++ b/config/configgrpc/configgrpc_test.go
@@ -963,6 +963,118 @@ func TestClientInfoInterceptors(t *testing.T) {
 	}
 }
 
+func TestClientMetadataWithAuthInterceptorsAndIncludeMetadata(t *testing.T) {
+	testCases := []struct {
+		desc            string
+		tester          func(context.Context, ptraceotlp.GRPCClient)
+		includeMetadata bool
+	}{
+		{
+			desc: "overwritten",
+			tester: func(ctx context.Context, cl ptraceotlp.GRPCClient) {
+				resp, errResp := cl.Export(ctx, ptraceotlp.NewExportRequest())
+				require.NoError(t, errResp)
+				require.NotNil(t, resp)
+			},
+			includeMetadata: true,
+		},
+		{
+			desc: "found",
+			tester: func(ctx context.Context, cl ptraceotlp.GRPCClient) {
+				resp, errResp := cl.Export(ctx, ptraceotlp.NewExportRequest())
+				require.NoError(t, errResp)
+				require.NotNil(t, resp)
+			},
+			includeMetadata: false,
+		},
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			mock := &grpcTraceServer{}
+			var l net.Listener
+
+			// prepare the server
+			{
+				gss := &ServerConfig{
+					NetAddr: confignet.AddrConfig{
+						Endpoint:  "localhost:0",
+						Transport: confignet.TransportTypeTCP,
+					},
+					IncludeMetadata: tC.includeMetadata,
+				}
+				gss.Auth = &configauth.Authentication{
+					AuthenticatorID: mockID,
+				}
+				host := &mockHost{
+					ext: map[component.ID]component.Component{
+						mockID: auth.NewServer(auth.WithServerAuthenticate(func(ctx context.Context, headers map[string][]string) (context.Context, error) {
+							return client.NewContext(ctx, client.Info{
+								Metadata: client.NewMetadata(metadata.Pairs("some-key-set-in-auth", "some-value-set-in-auth")),
+							}), nil
+						})),
+					},
+				}
+				srv, err := gss.ToServer(context.Background(), host, componenttest.NewNopTelemetrySettings())
+				require.NoError(t, err)
+				ptraceotlp.RegisterGRPCServer(srv, mock)
+
+				defer srv.Stop()
+
+				l, err = gss.NetAddr.Listen(context.Background())
+				require.NoError(t, err)
+
+				go func() {
+					_ = srv.Serve(l)
+				}()
+			}
+
+			// prepare the client and execute a RPC
+			{
+				gcs := &ClientConfig{
+					Endpoint: l.Addr().String(),
+					TLSSetting: configtls.ClientConfig{
+						Insecure: true,
+					},
+					Headers: map[string]configopaque.String{
+						"header1": "value1",
+					},
+				}
+
+				tt, err := componenttest.SetupTelemetry(componentID)
+				require.NoError(t, err)
+				defer func() {
+					require.NoError(t, tt.Shutdown(context.Background()))
+				}()
+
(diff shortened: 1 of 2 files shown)


grandwizard28:
@jpkrohling same test showcasing issue https://github.com/open-telemetry/opentelemetry-collector/issues/10093

config/configgrpc/configgrpc_test.go
```diff
@@ -963,6 +963,118 @@ func TestClientInfoInterceptors(t *testing.T) {
...
+					IncludeMetadata: tC.includeMetadata,
+				}
+				gss.Auth = &configauth.Authentication{
+					AuthenticatorID: mockID,
+				}
+				host := &mockHost{
+					ext: map[component.ID]component.Component{
+						mockID: auth.NewServer(auth.WithServerAuthenticate(func(ctx context.Context, headers map[string][]string) (context.Context, error) {
+							return client.NewContext(ctx, client.Info{
+								Metadata: client.NewMetadata(metadata.Pairs("some-key-set-in-auth", "some-value-set-in-auth")),
```
jpkrohling, line 1012: I see what you mean now. Authenticators are supposed to only provide data via the AuthData, not Metadata. Metadata is only supposed to be set by the parts that deal with the connection with the client, such as gRPC and HTTP servers (like those in receivers). Your downstream consumer of this information would have to call AuthData.GetAttribute(...). 

https://github.com/open-telemetry/opentelemetry-collector/blob/3d0189df8a26a5d654b51080dcd07ae57d394091/client/client.go#L26-L30

Here's an implementation example:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5cc9e03d054728b8f97d9214aa58910aa493c502/extension/basicauthextension/extension.go#L105-L107

grandwizard28: Yeah, but consider the following:

- This doesnt happen for oltp/http
- If I try putting in the AuthData, the batch processor will drop it. See this: https://github.com/open-telemetry/opentelemetry-collector/pull/10002

I feel the behaviours should be similar for oltp/http and oltp/grpc. Wouldn't you agree?

jpkrohling: Can you add a test case here for otlp/http? There should be no difference between the receivers, and if the authenticator is overriding the metadata, none should see the original headers after the authenticator runs.

grandwizard28: Table depicting what happens to metadata set in Auth:

| Protocol/Config | Include Metadata = True | Include Metadata = False |
| ------ | ------ | ----- |
| GRPC | overriden | not overriden |
| HTTP | not overriden | not overriden |

grandwizard28: @jpkrohling does the above make sense?

config/configgrpc/configgrpc_test.go
```diff
@@ -963,6 +963,118 @@ func TestClientInfoInterceptors(t *testing.T) {
...
+					},
+					IncludeMetadata: tC.includeMetadata,
+				}
+				gss.Auth = &configauth.Authentication{
+					AuthenticatorID: mockID,
+				}
+				host := &mockHost{
+					ext: map[component.ID]component.Component{
+						mockID: auth.NewServer(auth.WithServerAuthenticate(func(ctx context.Context, headers map[string][]string) (context.Context, error) {
+							return client.NewContext(ctx, client.Info{
```
jpkrohling, line 1011: The right thing here would be to obtain the client from the context, set _only_ the Auth attributes, and wrap it in the context that is being returned, like this:

```go
mockID: auth.NewServer(auth.WithServerAuthenticate(func(ctx context.Context, headers map[string][]string) (context.Context, error) {
	cl := client.FromContext(ctx)
	cl.Auth = &mockAuthData{
		Attributes: map[string]string{"some-key-set-in-auth": "some-value-set-in-auth"},
	}

	return client.NewContext(ctx, cl), nil
})),
```

config/configgrpc/configgrpc_test.go
```diff
@@ -963,6 +963,118 @@ func TestClientInfoInterceptors(t *testing.T) {
...
+				grpcClientConn, errClient := gcs.ToClientConn(context.Background(), componenttest.NewNopHost(), tt.TelemetrySettings())
+				require.NoError(t, errClient)
+				defer func() { assert.NoError(t, grpcClientConn.Close()) }()
+
+				cl := ptraceotlp.NewGRPCClient(grpcClientConn)
+				ctx, cancelFunc := context.WithTimeout(context.Background(), 2*time.Second)
+				defer cancelFunc()
+
+				// test
+				tC.tester(ctx, cl)
```
jpkrohling, line 1058: This is counter-intuitive, as we set the headers on the ClientSettings, but the client making use of the construct is responsible for using the headers. So, the headers you are setting at ClientConfig aren't being sent to the actual service. You'll need something like this:

```go
// this is what we expect clients to do before making a gRPC call
headers := map[string]string{}
for k, v := range gcs.Headers {
	headers[k] = string(v)
}
md := metadata.New(headers)
ctx = metadata.NewOutgoingContext(ctx, md)
```

config/confighttp/confighttp_test.go
```diff
@@ -1207,6 +1207,83 @@ func TestServerAuth(t *testing.T) {
...
+			desc:            "metadata:false",
+			includeMetadata: false,
+		},
+	}
+
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			host := &mockHost{
+				ext: map[component.ID]component.Component{
+					mockID: auth.NewServer(auth.WithServerAuthenticate(func(ctx context.Context, headers map[string][]string) (context.Context, error) {
```
jpkrohling, line 1229: Same comments as gRPC.

jpkrohling (commented):
I realize that perhaps the API isn't very intuitive at the moment, and we either need better documentation or to prevent the API from being misused. 

I found a couple of inconsistencies with the test, and I'm sending a new commit to this PR to show how we expect the APIs to be used.

grandwizard28:
Right! The intended usage is to set only the AuthData in custom authenticators and not touch metadata at all.

If someone tries to set the metadata in a custom authenticator (as I did), http would allow it to propagate but grpc wont. (With IncludeMetadata: true) Even though this is not the intended usage, the two protocols behave differently in this instance.

My Reason for doing so:
The whole reason I tried to set Metadata in the Authenticator was because the batch processor was dropping the AuthData altogether.

All I want is to be able to propagate Authdata across the collector.

jpkrohling:
> All I want is to be able to propagate Authdata across the collector

You can do that today already, except if you use the batching processor: that one will group data from potentially different connections into one single batch. Those connections came with potentially different auth data.

The proper solution would be to enhance the batch processor, so that it can also group data by auth data in addition to metadata.

jpkrohling:
@grandwizard28 , do you think this should be closed?

PR #10165, opened 2024-05-15
[chore] follow codeql recommendation

codeboten:
CodeQL is currently reporting "Invalid Go toolchain version: As of Go 1.21, toolchain versions must use the 1.N.P syntax.". This PR attempts to fix this.

### DIFF
--- a/Makefile.Common
+++ b/Makefile.Common
@@ -67,7 +67,7 @@ fmt: $(GOIMPORTS)
 .PHONY: tidy
 tidy:
 	rm -fr go.sum
-	$(GOCMD) mod tidy -compat=1.21
+	$(GOCMD) mod tidy -compat=1.21.0
 
 .PHONY: lint
 lint: $(LINT)
--- a/cmd/builder/go.mod
+++ b/cmd/builder/go.mod
@@ -3,7 +3,7 @@
 
 module go.opentelemetry.io/collector/cmd/builder
 
-go 1.21
+go 1.21.0
 
 require (
 	github.com/hashicorp/go-version v1.6.0
--- a/cmd/mdatagen/go.mod
+++ b/cmd/mdatagen/go.mod
@@ -1,6 +1,6 @@
 module go.opentelemetry.io/collector/cmd/mdatagen
 
-go 1.21
+go 1.21.0
 
 require (
 	github.com/google/go-cmp v0.6.0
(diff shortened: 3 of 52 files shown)


mx-psi (changes requested):
This breaks our compatibility promise: we are dropping support for Go 1.21.x for x less than 10. There were also recently some fixes upstream (see golang/go@27ed85d4d1702e868730ab6ea2ad6326988c615c) that I think should make this message go away.

codeboten:
> This breaks our compatibility promise: we are dropping support for Go 1.21.x for x less than 10. There were also recently some fixes upstream (see [golang/go@27ed85d](https://github.com/golang/go/commit/27ed85d4d1702e868730ab6ea2ad6326988c615c)) that I think should make this message go away.

Happy to close this, however this PR does bring up a good question of whether we should support versions of go 1.21.x with known CVEs.

mx-psi:
> > This breaks our compatibility promise: we are dropping support for Go 1.21.x for x less than 10. There were also recently some fixes upstream (see [golang/go@27ed85d](https://github.com/golang/go/commit/27ed85d4d1702e868730ab6ea2ad6326988c615c)) that I think should make this message go away.
> 
> Happy to close this, however this PR does bring up a good question of whether we should support versions of go 1.21.x with known CVEs.

We should definitely encourage using the latest version. We could add a `toolchain` directive on the builder to force the latest version to be used.

In terms of support however, I don't think we lose much by continuing to support those versions, but we can discuss it!

codeboten:
> In terms of support however, I don't think we lose much by continuing to support those versions, but we can discuss it!

Right, I'm thinking that if we were to only support versions greater than any with known CVEs it would act as incentive to any consumers of the collector libraries to also update to newer versions. I can open a separate issue to discuss and close this PR after I do so.

bogdandrutu:
100% we should enforce latest (hopefully no CVEs) version when we build the final binary that we release as a binary, but for libraries like artifacts we cannot enforce this.

codeboten:
@mx-psi @bogdandrutu to appease codeql and ensure we continue to support all version 1.21, i updated the version in go.mod to 1.21.0

mx-psi (approved):
LGTM. I'd like us to revisit `1.21` vs `1.21.0` once the patch I mentioned above in golang/go is released, but I am okay with doing this for now

PR #10173, opened 2024-05-17
[confmap] Remove deprecated ResolverSettings fields

evan-bradley:
#### Description

These fields were deprecated in v0.99.0 and aren't used in any upstream repositories.

These appear to still be used in downstream distributions. If we want to lengthen the deprecation period for these fields, I'll open a PR to instead set a timeline for their removal.

### DIFF
--- a/.chloggen/remove-confmap-instances.yaml
+++ b/.chloggen/remove-confmap-instances.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove deprecated `Providers` and `Converters` from `confmap.ResolverSettings`
+
+# One or more tracking issues or pull requests related to the change
+issues: [10173]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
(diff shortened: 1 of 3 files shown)


songy23 (approved):

confmap/resolver.go
```diff
@@ -98,32 +87,15 @@ func NewResolver(set ResolverSettings) (*Resolver, error) {
...
-		}
-		converters = set.Converters
-	} else {
-		converters = make([]Converter, len(set.ConverterFactories))
-		for i, factory := range set.ConverterFactories {
-			converters[i] = factory.Create(set.ConverterSettings)
-		}
+	converters := make([]Converter, len(set.ConverterFactories))
+	for i, factory := range set.ConverterFactories {
+		converters[i] = factory.Create(set.ConverterSettings)
```
bogdandrutu, line 98: Separate PR/issue. It would be great to have all public APIs like `Create` always accept `context.Context` as the first argument.

bogdandrutu (approved):

mx-psi (approved):

.chloggen/remove-confmap-instances.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate `Providers` and `Converters` in `confmap.ResolverSettings`
```
andrzej-stencel, line 10: This is actually a removal, right? Deprecation happened in https://github.com/open-telemetry/opentelemetry-collector/pull/9516 (which was marked as an `enhancement` :thinking: )
```suggestion
note: Remove deprecated `Providers` and `Converters` from `confmap.ResolverSettings`
```

PR #10176, opened 2024-05-18
[confighttp] Add support for cookies

### LINKED ISSUE #10175: [confighttp] Support cookie sessions for sticky session cookies
Some load balancing solutions rely on cookies for sticky sessions and enforcing a max session age.

The curent confighttp does not expose a way to set cookie usage on the http client.
We would like to add this support.

atoulme:
#### Description
Add support for cookies in HTTP clients with `cookies::enabled`.

#### Link to tracking issue
Fixes #10175

#### Testing
Unit test

#### Documentation
Added to README

### DIFF
--- a/.chloggen/cookie_support.yaml
+++ b/.chloggen/cookie_support.yaml
@@ -7,7 +7,7 @@ change_type: enhancement
 component: confighttp
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add support for cookies in HTTP clients with `cookies_enabled`.
+note: Add support for cookies in HTTP clients with `cookies::enabled`.
 
 # One or more tracking issues or pull requests related to the change
 issues: [10175]
--- a/.chloggen/cookie_support.yaml
+++ b/.chloggen/cookie_support.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confighttp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for cookies in HTTP clients with `cookies_enabled`.
+
+# One or more tracking issues or pull requests related to the change
+issues: [10175]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: The method `confighttp.ToClient` will return a client with a `cookiejar.Jar` which will reuse cookies from server responses in subsequent requests.
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/config/confighttp/README.md
+++ b/config/confighttp/README.md
@@ -36,9 +36,6 @@ README](../configtls/README.md).
 - [`http2_ping_timeout`](https://pkg.go.dev/golang.org/x/net/http2#Transport)
 - [`cookies`](https://pkg.go.dev/net/http#CookieJar)
   - [`enabled`] if enabled, the client will store cookies from server responses and reuse them in subsequent requests.
-  - [`insecure`] if true, the client accepts setting cookies for any domain. This is useful for testing but is insecure:
-    it means that the HTTP server for foo.co.uk can set a cookie for bar.co.uk.
-    If false, the client will allow setting cookies based on the list provided by https://publicsuffix.org/
 
 Example:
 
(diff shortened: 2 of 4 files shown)


config/confighttp/confighttp.go
```diff
@@ -227,9 +232,18 @@ func (hcs *ClientConfig) ToClient(ctx context.Context, host component.Host, sett
 		}
 	}
 
+	var jar http.CookieJar
+	if hcs.CookiesEnabled {
+		jar, err = cookiejar.New(nil)
+		if err != nil {
+			return nil, err
```
atoulme, line 239: This line cannot be tested, the New method returns nil always for now. I am just checking the error returned in case this changes in the future.

.chloggen/cookie_support.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: The method `confighttp.ToClient` will return a client with a `cookiejar.Jar` which will reuse cookies from server responses in subsequent requests.
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
mx-psi, line 25: This PR introduces an API only change, right?
```suggestion
change_logs: [api]
```

config/confighttp/confighttp.go
```diff
@@ -99,6 +100,10 @@ type ClientConfig struct {
 	// HTTP2PingTimeout if there's no response to the ping within the configured value, the connection will be closed.
 	// If not set or set to 0, it defaults to 15s.
 	HTTP2PingTimeout time.Duration `mapstructure:"http2_ping_timeout"`
+
+	// CookiesEnabled enables reusing cookies between requests.
+	// This is useful to manage sticky session cookies.
+	CookiesEnabled bool `mapstructure:"cookies_enabled"`
```
mx-psi, line 106: Would it make sense to have a cookies section instead?

.chloggen/cookie_support.yaml
atoulme: No, this is a new config setting

config/confighttp/confighttp.go
atoulme: What would that look like?

mx-psi:
```
cookies:
    enabled: true
```

atoulme: oh. Not sure. I guess maybe cookies might have more options. I'll do it.

.chloggen/cookie_support.yaml
atoulme: I was just trying to excuse the coverage build failure. Sorry I wasn't more clear.

atoulme:
@mx-psi are we good to go?

mx-psi:
> @mx-psi are we good to go?

I'd like to have more context on what the use cases for this are. I think no further code changes are necessary, but I also want to know *why* we are doing this.

config/confighttp/confighttp.go
```diff
@@ -227,9 +236,18 @@ func (hcs *ClientConfig) ToClient(ctx context.Context, host component.Host, sett
 		}
 	}
 
+	var jar http.CookieJar
+	if hcs.Cookies != nil && hcs.Cookies.Enabled {
+		jar, err = cookiejar.New(nil)
```
mx-psi, line 241: Based on the docs of `cookiejar.Options`:

```
// PublicSuffixList is the public suffix list that determines whether
// an HTTP server can set a cookie for a domain.
//
// A nil value is valid and may be useful for testing but it is not
// secure: it means that the HTTP server for foo.co.uk can set a cookie
// for bar.co.uk.
```

and then it says

> A public suffix list implementation is in the package golang.org/x/net/publicsuffix.

Should we look into using https://pkg.go.dev/golang.org/x/net/publicsuffix ?

atoulme: ok I'll add that, good find thank you!

config/confighttp/README.md
```diff
@@ -34,6 +34,11 @@ README](../configtls/README.md).
 - [`disable_keep_alives`](https://golang.org/pkg/net/http/#Transport)
 - [`http2_read_idle_timeout`](https://pkg.go.dev/golang.org/x/net/http2#Transport)
 - [`http2_ping_timeout`](https://pkg.go.dev/golang.org/x/net/http2#Transport)
+- [`cookies`](https://pkg.go.dev/net/http#CookieJar)
+  - [`enabled`] if enabled, the client will store cookies from server responses and reuse them in subsequent requests.
+  - [`insecure`] if true, the client accepts setting cookies for any domain. This is useful for testing but is insecure:
+    it means that the HTTP server for foo.co.uk can set a cookie for bar.co.uk.
+    If false, the client will allow setting cookies based on the list provided by https://publicsuffix.org/
```
mx-psi, line 41: Do we have an use case for this? If the only use case is for testing on the module itself, I would make the field private

atoulme: yes? I can see how this might be useful when troubleshooting.

mx-psi: Unless you think it's essential, I would rather leave it out until someone asks for it explicitly so that we commit to a smaller public API

mx-psi (approved):

PR #10178, opened 2024-05-20
[chore] Remove unused package test file from otelcorecol

bogdandrutu:


cmd/otelcorecol/package_test.go
```diff
@@ -1,14 +0,0 @@
...
-package main
-
-import (
-	"testing"
-
-	"go.uber.org/goleak"
-)
-
-func TestMain(m *testing.M) {
-	goleak.VerifyTestMain(m)
```
songy23, line 13: I think this is used to catch goleak regression? https://github.com/open-telemetry/opentelemetry-collector/pull/9210 

cc @crobert-1

crobert-1: There are no more tests in this package, so it's no longer needed. 👍

crobert-1 (approved):

Issue #10161, opened 2024-05-15
[confmap] Unify behavior of `${env:ENV}` and `${ENV}`


mx-psi:
As decided on #9854, we want to unify the behavior of `${env:ENV}` and `${ENV}`. Currently, each has a different behavior: the `${ENV}` syntax does not do any YAML parsing, while the `${env:ENV}` syntax does parse the contents as YAML. We want to unify both and make them have the same behavior, described in RFC #9854.

One approach for this is to have some 'default' confmap provider used by the `${..}` syntax, and control what provider is used (a `expandconverter` provider or the `env` provider) through a feature gate.

cc @TylerHelmuth

TylerHelmuth:
@mx-psi I see a couple approaches.

1. Allow `envprovider` to expand `${env:}` and `${}`.  This is conceptually simple, but I think breaks some logic in `resolver.expandValue` that requires providers to have an schema so it can identify which provider to use.
2. Allow `resolver.expandValue` to expand `${}` as an env var when it doesn't contain a uri (basically do env var expansion [here](https://github.com/open-telemetry/opentelemetry-collector/blob/e55d22aec61d7b9787b88448ec4a2ae8fb5eefd6/confmap/expand.go#L103-L105).  This is doable, but would mean duplicating the yaml parsing logic and logging logic (which means adding a Logger to the ResolverSettings) from the `envprovide`.
3. Leave expandconverter in place, but restrict it to only allowing `${}` and add yaml parsing.  
4. Create the concept of a "default" provider that is used by the Resolver when no schema is set. This would be a new filed on the ResolverSettings. For our collector instances we'd set the default to be the envprovider. The envprovider would need to be updated to not error when no schema is provided.

I've experimented with 2 and think it would be workable.  It would take care of this issue and https://github.com/open-telemetry/opentelemetry-collector/issues/8215 at once.  It did feel a little messy but there are probably ways to share code between the 2 places.

I would also be ok with option 4, as it provides a lot of flexibility.

evan-bradley:
Option 4 feels the cleanest to me. I think we should do it without the awareness of any of the Providers by prepending the default schema to the URI when the URI is given to the Provider, like the [Resolver currently does with files](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/resolver.go#L136) passed to `config` arguments. If we provide this option, we may also want to consider whether to make it configurable for the command line flags

TylerHelmuth:
[...]

PR #10182, opened 2024-05-20
[confmap] Add ability to set default provider

TylerHelmuth:
#### Description
This PR adds support for expanding `${}` syntax when no schema is provided by allowing Resolver to use a default provider.

In a followup PR I'll update otelcol with a feature gate that allow switching between a default envProvider and the expandconverter.

#### Link to tracking issue

#### Testing
Added unit tests

#### Documentation
Added godoc strings


confmap/expand.go
```diff
@@ -98,10 +98,17 @@ func findURI(input string) string {
 
 // findAndExpandURI attempts to find and expand the first occurrence of an expandable URI in input. If an expandable URI is found it
 // returns the input with the URI expanded, true and nil. Otherwise, it returns the unchanged input, false and the expanding error.
+// This method expects input to start with ${ and end with }
 func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (any, bool, error) {
 	uri := findURI(input)
 	if uri == "" {
-		// No URI found, return.
+		if mr.defaultProvider != nil {
```
evan-bradley, line 105: Could we move this code into `findURI` (which will probably require making it a method on `*Resolver`) and just use the returned `uri` in our regular flow? Adding an exception like this instead of baking the concept of a default scheme into our resolution process makes me worried we are opening ourselves up to missing cases in the future.

In particular, it looks like `findURI` checks whether a scheme is provided, and I think we should have logic that says "use the provided scheme, or a default scheme if one has been defined".

.chloggen/confmap-default-provider.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
evan-bradley, line 25:
```suggestion
change_logs: [api]
```

confmap/expand.go
TylerHelmuth: I'll look at that option again, but I believe it will mean returning a string by combining the default provider's scheme and the current input.  So if the input is `${PORT}`, `findURI` needs to return `${env:PORT}` and I was trying to avoid messing with the input string too much.

mx-psi: > I was trying to avoid messing with the input string too much.

Are there any specific problems you are trying to avoid? IMO just adding the prefix should be fine and would make the code cleaner

TylerHelmuth: We're also need to be careful (I think) about using the default provider on nested URIs

TylerHelmuth: Also that will cause a problem lower in `findAndExpandURI` when the identified uri is used as the exact replace string from the input

mx-psi: Should we return some structured data? Some struct that contains the raw URI and has methods to get the full URI (i.e. adding the default if needed), and the provider?

confmap/resolver.go
```diff
@@ -34,11 +35,16 @@ type ResolverSettings struct {
...
-	// if a Provider is not given.
+	// ProviderFactories is a slice of Provider creation functions.
+	// It is required to have at least one provider.
 	ProviderFactories []ProviderFactory
 
+	// DefaultProvider is the provider that is used if ${} syntax is used but no schema is provided.
+	// If no DefaultProvider is set, ${} with no schema will not be expanded.
+	// It is strongly recommended to set an EnvProvider as the default to align with the
+	// OpenTelemetry Configuration Specification
+	DefaultProvider ProviderFactory
```
TylerHelmuth, line 46: Should we require the `DefaultProvider`  be in the `ProviderFactories` list? Should we automatically add `DefaultProvider` to `ProviderFactories` in `NewResolver`?

If we want to make the use of `DefaultProvider` as close to the current workflow as possible we need it to be in the providers map.

confmap/expand.go
TylerHelmuth: @mx-psi @evan-bradley I've pushed an experiment with findURI that allows expanding stuff like `${env:CUSTOM${NAME}}`.  If we like the approach/extra functionality I will add a lot more unit tests.  For now I've updated the existing to match the new behavior when a DefaultProvider is set.

confmap/resolver.go
TylerHelmuth: But I also implemented a way that allows `DefaultProvider` to not be in the providers list.

evan-bradley: >But I also implemented a way that allows `DefaultProvider` to not be in the providers list.

I think that's alright, throwing an error is probably non-ideal if it's not in there. Could we add it to the `ProviderFactories` list in this case? I think it's weird to have a "default" that is not in the list of what you could use if you actually provided a scheme.

confmap/expand.go
TylerHelmuth: > keep all the defaulting logic contained to a single function that's separated from the actual resolution logic

This ends up being difficult bc of the support for nested URIs and the default provider not being in the `providers` list and not requiring a `location` since we know the uri.  I can keep experimenting, but unless we add the DefaultProvider to the providers map we wont be able to use 

```go
                lURI, locErr := newLocation(uri)
		if locErr != nil {
			return nil, false, locErr
		}
		if strings.Contains(lURI.opaqueValue, "$") {
			return nil, false, fmt.Errorf("the uri %q contains unsupported characters ('$')", lURI.asString())
		}
		ret, err = mr.retrieveValue(ctx, **lURI)**
```

exactly bc `mr.retrieveValue` expects the sheme to be in the providers map

confmap/resolver.go
TylerHelmuth: If we agree that adding the defaultprovider to the providers map is ok, that will simplify other aspects of the feature

confmap/expand.go
TylerHelmuth: I've pushed another change that utilizes DefaultProvider being in the provider map and modifies the uri returned from `findURI` to include the default provider scheme.  This reduces the number of changes needed in other methods in the resolution logic, but feels a little magical to me.  Maybe a struct returned from `findURI` instead of 2 strings would be better?

confmap/expand.go
```diff
@@ -76,30 +76,42 @@ func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, erro
...
+		return "", ""
+	}
+	return mr.findURI(remaining)
+}
+
 // findURI attempts to find the first expandable URI in input. It returns an expandable
 // URI, or an empty string if none are found.
 // Note: findURI is only called when input contains a closing bracket.
-func findURI(input string) string {
+func (mr *Resolver) findURI(input string) (string, string) {
```
bogdandrutu, line 90: Please explain the returned values, hard to understand.

confmap/resolver.go
```diff
@@ -20,9 +20,10 @@ var driverLetterRegexp = regexp.MustCompile("^[A-z]:")
 
 // Resolver resolves a configuration as a Conf.
 type Resolver struct {
-	uris       []location
-	providers  map[string]Provider
-	converters []Converter
+	uris            []location
+	providers       map[string]Provider
+	defaultProvider Provider
```
bogdandrutu, line 25: s/defaultProvider/defualtSchema?

confmap/resolver.go
bogdandrutu: Would probably accept a "string" and call this "DefualtSchema".

confmap/expand.go
bogdandrutu, line 90: Instead of changing the `findURI` logic that much, I would rather change it to return all "expandable" even if no schema, and in the expandUri would add the logic if no schema and a defaultSchema present do that, otherwise do nothing and continue to expand.

TylerHelmuth: I agree with reverting back to my previous implementation that does not do any magic scheme appending in `findURI` and doing more default logic in other functions.  I'll work on this.

TylerHelmuth: @bogdandrutu I believe I've addressed all your feedback, ready for another review.

confmap/expand.go
```diff
@@ -76,30 +76,31 @@ func (mr *Resolver) expandValue(ctx context.Context, value any) (any, bool, erro
 	return value, false, nil
 }
 
-// findURI attempts to find the first expandable URI in input. It returns an expandable
+// findURI attempts to find the first potentially expandable URI in input. It returns a potentially expandable
 // URI, or an empty string if none are found.
 // Note: findURI is only called when input contains a closing bracket.
-func findURI(input string) string {
+func (mr *Resolver) findURI(input string) string {
```
bogdandrutu, line 82: not using anything from `mr`, no need to be a function on Resolver struct.

```suggestion
func findURI(input string) string {
```

confmap/expand.go
```diff
@@ -113,6 +114,9 @@ func (mr *Resolver) findAndExpandURI(ctx context.Context, input string) (any, bo
 	if err != nil {
 		return input, false, err
 	}
+	if !changed {
+		return input, false, nil
+	}
 	repl, err := toString(uri, expanded)
```
bogdandrutu, line 120: Got confused why `uri` is needed, so created a small cleanup [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/10244)

confmap/expand.go
```diff
@@ -138,11 +142,26 @@ func toString(strURI string, input any) (string, error) {
...
+	var lURI location
+	var err error
+	// If no ':' then we have no scheme, check if there is a default scheme configured
+	if !strings.Contains(uri, ":") {
+		// If we have no default scheme there is nothing that can be expanded, return the original input unchanged
+		if mr.defaultScheme == "" {
+			return input, false, nil
+		}
+		lURI = location{scheme: mr.defaultScheme, opaqueValue: uri}
+	} else {
```
bogdandrutu, line 158: There is a behavior change here, before having `${TEST_ENV}${file:foo.txt}` would expend the file always, with the current change you will find the first "uri" and will return that nothing changed and will not call again to find `${file:foo.txt}`, can you add a test for this?

I think in findUri you need to pass one option to say "skipNoSchemaUri" if  `mr.defaultScheme == ""`. Then you change it to only check for `:` if `skipNoSchemaUri == true`

TylerHelmuth: I think you're right good catch: https://github.com/open-telemetry/opentelemetry-collector/pull/10246

confmap/expand.go
```diff
@@ -138,11 +142,26 @@ func toString(input any) (string, error) {
...
+	// strip ${ and }
+	uri := input[2 : len(input)-1]
+
+	var lURI location
+	var err error
+	// If no ':' then we have no scheme, check if there is a default scheme configured
+	if !strings.Contains(uri, ":") {
+		// If we have no default scheme there is nothing that can be expanded, return the original input unchanged
+		if mr.defaultScheme == "" {
+			return input, false, nil
```
bogdandrutu, line 155: Cannot happen anymore.

confmap/expand.go
```diff
@@ -138,11 +142,26 @@ func toString(input any) (string, error) {
...
+		}
+		lURI = location{scheme: mr.defaultScheme, opaqueValue: uri}
```
bogdandrutu, line 157: I would rather make `uri = mr.defaultScheme + ":" + uri` then apply same logic of using newLocation. Ensures we have consistent checks.

bogdandrutu (approved):

confmap/expand_test.go
```diff
@@ -317,6 +407,9 @@ func TestResolverExpandStringValues(t *testing.T) {
 
 func newEnvProvider() ProviderFactory {
 	return newFakeProvider("env", func(_ context.Context, uri string, _ WatcherFunc) (*Retrieved, error) {
+		if uri[0:4] != "env:" {
+			uri = "env:" + uri
+		}
```
evan-bradley, line 412: What's the purpose of this? Shouldn't this be handled by the Resolver and not by a provider? It would probably be good to leave a comment to explain it.

confmap/resolver.go
```diff
@@ -34,11 +35,16 @@ type ResolverSettings struct {
 	// It is required to have at least one location.
 	URIs []string
 
-	// ProviderFactories is a list of Provider creation functions.
-	// It is required to have at least one ProviderFactory
-	// if a Provider is not given.
+	// ProviderFactories is a slice of Provider creation functions.
+	// It is required to have at least one provider.
```
evan-bradley, line 39:
```suggestion
	// ProviderFactories is a slice of Provider factories.
	// It is required to have at least one factory.
```

confmap/resolver.go
```diff
@@ -34,11 +35,16 @@ type ResolverSettings struct {
...
 	ProviderFactories []ProviderFactory
 
+	// DefaultScheme is the scheme that is used if ${} syntax is used but no schema is provided.
+	// If no DefaultScheme is set, ${} with no schema will not be expanded.
+	// It is strongly recommended to set an EnvProvider as the default scheme to align with the
```
evan-bradley, line 44:
```suggestion
	// It is strongly recommended to set "env" as the default scheme to align with the
```

evan-bradley (approved):
Looks good to me, I like the implementation approach we ended up on.

confmap/expand_test.go
TylerHelmuth: Comment added

codeboten (approved):

PR #10183, opened 2024-05-21
Restrict characters in environment variable names

ankitpatel96:
#### Description
Restricts Environment Variable names according to the [restrictions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md#environment-variable-substitution) published by the OpenTelemetry Configuration Working Group.  Changes both the expand converter and the env provider.  Defines one regex to be used for both of these.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9531

#### Testing
Unit tests


TylerHelmuth (approved):
Excellent!

.chloggen/env-var-name-verify.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: envprovider
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Restricts Environment Variable names.  Must start with a letter or and underscore, and can only contain underscores, letters, or numbers.
```
mx-psi, line 10: We are now restricting to ASCII as well!
```suggestion
note: Restricts Environment Variable names.  Environment variable names must now be ASCII only and start with a letter or an underscore, and can only contain underscores, letters, or numbers.
```

confmap/confmap.go
```diff
@@ -21,9 +22,12 @@ import (
 
 const (
 	// KeyDelimiter is used as the default key delimiter in the default koanf instance.
-	KeyDelimiter = "::"
+	KeyDelimiter      = "::"
+	EnvVarNamePattern = `^[a-zA-Z_][a-zA-Z0-9_]*$`
 )
 
+var EnvVarNameRegexp = regexp.MustCompile(EnvVarNamePattern)
```
mx-psi, line 29: I'd rather move this to a `confmap/internal` package, because
- upstream may change the way this is expressed (see open-telemetry/opentelemetry-specification/issues/4027)
- morally, this feels like something that should be exposed by the `go.opentelemetry.io/contrib/config` module instead of us

confmap/converter/expandconverter/expand_test.go
```diff
@@ -253,6 +275,38 @@ func TestDeprecatedWarning(t *testing.T) {
 	}
 }
 
+func TestNewExpandConverterWithErrors(t *testing.T) {
+	var testCases = []struct {
```
mx-psi, line 279: Can we add a test case that starts with a number? e.g. `${1ENV}`

mx-psi (approved):

PR #10195, opened 2024-05-21
Generate profiling pdata

dmathieu:
#### Description

This generates the pdata for profiles.

#### Link to tracking issue
Last step of #10109.

#### Testing

The generated code is unit-tested by generated tests.

cc @mx-psi


petethepig (approved):

versions.yaml
```diff
@@ -56,6 +56,10 @@ module-sets:
       - go.opentelemetry.io/collector/semconv
       - go.opentelemetry.io/collector/service
       - go.opentelemetry.io/collector/filter
+  experimental:
+    version: v0.0.1
+    modules:
+      - go.opentelemetry.io/collector/pdata/pprofile
```
mx-psi, line 63: I think we can put this in the beta module set, unless we expect very frequent breaking changes to the data model (I assume there will be some, but since the Profiling SIG has been working on this for some time, I wouldn't expect a complete redesign, right?).

I realize now that my mention of #9209 may have been confusing, but I was thinking about how to add API in other Go modules if they are 1.0 (i.e. the part where I talk about if we need to use build tags/other approaches). 

If you still think it's better to have a different module set for this, then we would need to
- add support for this at the release automation level
- add instructions for release managers to follow when releasing that
- Define when breaking changes are allowed in this module set, when will we do releases, how will the release version increase

pdata/pprofile/go.mod
```diff
@@ -0,0 +1,25 @@
+module go.opentelemetry.io/collector/pdata/pprofile
+
+go 1.21
+
+require (
+	github.com/stretchr/testify v1.9.0
+	go.opentelemetry.io/collector/pdata v0.0.0-00010101000000-000000000000
```
mx-psi, line 7: You can just put the latest version here
```suggestion
	go.opentelemetry.io/collector/pdata v1.8.0
```

versions.yaml
dmathieu: Following a discussion with @mx-psi, I have moved the package to beta, as beta packages are allowed breaking changes.

mx-psi (approved):
LGTM. This has to wait until contrib-tests work again (should happen soon)

mx-psi:
cc @open-telemetry/collector-approvers last call to review this one

PR #10200, opened 2024-05-21
[confmap] Read YAML struct tags

evan-bradley:
#### Description

Third party configuration structs, such as those used in the Prometheus receiver, [use YAML struct tags](https://github.com/prometheus/common/blob/main/config/http_config.go#L313) that are necessary for `confmap` to understand if we want to be able to marshal YAML. In particular, in `(*Conf).ToStringMap`, we convert structs to a Map object, where ignoring flags like `omitempty` or `squash` causes the map to have a different structure than it would if we called `yaml.Marshal` on the same struct.

This PR allows `ToStringMap` to read `omitempty` and `inline` when struct fields have been annotated with the `yaml` tag. These are 2/3 of the flags the `yaml` tag supports: https://pkg.go.dev/gopkg.in/yaml.v3#Marshal.

Caveats:
* Things like `yaml:",squash"` and `mapstructure:",inline"` will be possible now. I figure that allowing any valid flag regardless of the tag that it applies to is okay (we also support `remain` as an alias) and won't incur any issues, but if we are concerned about conflicts or stability guarantees, I can enforce a stricter schema for which tags support which flags.
* I'm leaving out support for the `json` tag for now since we don't support JSON the way we support YAML, but I can add support for it here if we'd like.

Required for https://github.com/open-telemetry/opentelemetry-collector/pull/10139.

### DIFF
--- a/.chloggen/confmap-yaml-tag.yaml
+++ b/.chloggen/confmap-yaml-tag.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Support marshaling using the `omitempty` and `inline` flags with the `yaml` tag
+
+# One or more tracking issues or pull requests related to the change
+issues: [10200]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This is only for compatibility with third-party configuration structs, first-party structs
+  should still use the `mapstructure` tag and its preferred flags.
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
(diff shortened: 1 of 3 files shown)


bogdandrutu (commented):
Few things I want to understand before reviewing/approving this:
1. Is the plan to support both long term?
2. What about yaml unmarshallers like the https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler? Another thing is for example [encoding.TextUnmarshal](https://pkg.go.dev/encoding#TextUnmarshaler) which is currently supported by us and yaml, if behavior is different which one we use?
3. Can we just move to `yaml`? We have validate Config test in the mdatagen tool which can help with these checks.

evan-bradley:
@bogdandrutu I'm closing this in favor of #10282, but I think your questions are still valid. To give a few answers:

> Is the plan to support both long term?

The approach in #10282 is to defer to `yaml.Marshal` if no mapstructure tags are found. I think this is okay to support in the long term since we are deferring to a third party 

> What about yaml unmarshallers like the [pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler](https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshaler)?

We use `yaml.Unmarshaler` in #10282 now.

>Another thing is for example [encoding.TextUnmarshal](https://pkg.go.dev/encoding#TextUnmarshaler) which is currently supported by us and yaml, if behavior is different which one we use?

#10282 uses a list of hooks to implement this. The hook order will dictate whether `TextMarshaler` is used or if `yaml.Marshal` is called on the struct. 

>Can we just move to yaml? We have validate Config test in the mdatagen tool which can help with these checks.

I think this could be feasible, but I will need to look into it. I think we can fall back to `yaml` tags as an intermediary step before deciding.

PR #10203, opened 2024-05-21
Update validate command to use the new configuration options

dmitryax:
After https://github.com/open-telemetry/opentelemetry-collector/pull/10173, the `validate` sum-command works differently from the root command. 

If a downstream distribution provides `URIs` as part of the new `ConfigProviderSettings` without `--config` flag, the `validate` command fails with 
```
2024/05/21 15:53:25 main.go:85: application run finished with error: at least one config flag must be provided
```


samiura (approved):

bogdandrutu (approved):

songy23 (approved):

evan-bradley (approved):
Thanks for catching this.

PR #10205, opened 2024-05-22
Add support for TimeEncoding for Zap logger

### LINKED ISSUE #10204: Allow configuration of the service Zap logger Encodetime
It doesn't seem possible to set the base telemetry logger's `EncoderConfig.EncodeTime` explicitly.  Rather, it's hardcoded to `ISO8601TimeEncoder` (for console format), or epoch (for JSON format).

Add a new field to `LogsConfig` to allow users to explicitly choose the time encoding independent of the log format (console vs JSON)

- Using reflection to modify the logger instance that the service creates 👎
- Changing my logging pipeline to handle the epoch format

**Additional context**
N/A

jeff303:
#### Description
- Add `TimeEncoding` field under `LogsConfig` to control Zap logger time encoding explicitly
- Update `newLogger` to read and interpret the selected option

#### Link to tracking issue
Fixes #10204

#### Testing

None yet (wasn't able to find existing tests for `newLogger`)

#### Documentation

None yet (unsure where the docs for this live)

### DIFF
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -92,6 +92,19 @@ type LogsConfig struct {
 	//
 	// By default, there is no initial field.
 	InitialFields map[string]any `mapstructure:"initial_fields"`
+
+	// TimeEncoding controls the zap logger's time encoding
+	// Permitted values are:
+	//  - "Epoch"
+	//  - "EpochMillis"
+	//  - "EpochNanos"
+	//  - "ISO8601"
+	//  - "RFC3339"
+	//  - "RFC3339Nano"
+	// Leaving the value blank results in the default time encoding, which
+	// depends on the message encoding (for JSON, it will be EpochNanos,
+	// and for console, it will be ISO8601
+	TimeEncoding string `mapstructure:"time_encoding"`
 }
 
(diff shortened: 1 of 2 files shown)


service/telemetry/config.go
```diff
@@ -92,6 +92,19 @@ type LogsConfig struct {
...
+	//  - "Epoch"
+	//  - "EpochMillis"
+	//  - "EpochNanos"
+	//  - "ISO8601"
+	//  - "RFC3339"
+	//  - "RFC3339Nano"
+	// Leaving the value blank results in the default time encoding, which
+	// depends on the message encoding (for JSON, it will be EpochNanos,
+	// and for console, it will be ISO8601
+	TimeEncoding string `mapstructure:"time_encoding"`
```
bogdandrutu, line 107: We would prefer not to derive from the zap.Config. Is this there already? if yes, can you add a link to confirm the name?

jeff303: Apologies, but I'm not following this.  I wasn't able to find any mechanism by which I can set the Zap logger's time encoding through the existing telemetry config, hence the new field.  One can only indirectly influence it (ex, by choosing console as the format), but there is no way to use, say, RFC3339 time encoding along with JSON formatted messages.

PR #10208, opened 2024-05-22
[exporterhelper] Adding shutdown & startup debug logs for retry, queue, and batch sender

### LINKED ISSUE #10201: There should be more observability around exporter helper start up and shutdown

As a user of otel collector, I find it hard to see which exporter helper component is stuck starting up or shutting down. Some of the examples in the past:
- persistent queue helper start up can be stuck: https://github.com/open-telemetry/opentelemetry-collector/issues/9451
- batch sender shut down can be stuck due to config: https://github.com/open-telemetry/opentelemetry-collector/issues/10061

When debugging these problems, it isn't clear that the components are stuck.

A debug message is added at the beginning of startup/shutdown and at the end of the startup/shutdown. Components we should add:
- retry sender
- batch sender
- queue sender

Add it as INFO logs but it can be too noisy

**Additional context**

timannguyen:
#### Description

Adding debug logs to starting and shutdown of batch, retry, and queue sender in exporterhelper to understand which component is stuck shutting down or starting up. This is added in order to improve debugging when one of these components are stuck shutting down or starting up.

#### Link to tracking issue
Fixes #10201

#### Testing

N/A

#### Documentation

N/A

### DIFF
--- a/exporter/exporterhelper/batch_sender.go
+++ b/exporter/exporterhelper/batch_sender.go
@@ -14,6 +14,8 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+	"go.opentelemetry.io/collector/internal/obsreportconfig/obsmetrics"
+	"go.opentelemetry.io/otel/attribute"
 )
 
 // batchSender is a component that places requests into batches before passing them to the downstream senders.
@@ -38,7 +40,8 @@ type batchSender struct {
 	mu          sync.Mutex
 	activeBatch *batch
 
-	logger *zap.Logger
+	traceAttribute attribute.KeyValue
+	logger         *zap.Logger
 
 	shutdownCh chan struct{}
 	stopped    *atomic.Bool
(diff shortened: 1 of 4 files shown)


splunkericl (approved):

exporter/exporterhelper/queue_sender.go
```diff
@@ -137,15 +139,20 @@ func (qs *queueSender) Start(ctx context.Context, host component.Host) error {
 			return nil
 		}))
 
+	qs.logger.Debug("Started Queue Sender", zap.String(obsmetrics.ExporterKey, qs.fullName), zap.Error(err))
 	errs = multierr.Append(errs, err)
```
atoulme, line 143:
```suggestion
	errs = multierr.Append(errs, err)
	qs.logger.Debug("Started Queue Sender", zap.String(obsmetrics.ExporterKey, qs.fullName), zap.Error(errs))
```

exporter/exporterhelper/batch_sender.go
```diff
@@ -91,6 +94,7 @@ func (bs *batchSender) Start(_ context.Context, _ component.Host) error {
 		}
 	}()
 
+	bs.logger.Debug("Started Batch Sender", zap.String(obsmetrics.ExporterKey, bs.fullName))
```
bogdandrutu, line 97: I think the `obsmetrics.ExporterKey` entry is not necessary, the logger will have a "name" entry (see https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/components/loggers.go#L38) with this.

bogdandrutu (changes requested):

exporter/exporterhelper/batch_sender.go
timannguyen: removed the exporter key from all the logs.

exporter/exporterhelper/queue_sender.go
timannguyen: moved

timannguyen:
@bogdandrutu is there anything else i need to address?

PR #10209, opened 2024-05-22
[chore] update lint configuration

codeboten:
This addresses some of the following warnings:

```
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`.
WARN [config_reader] The configuration option `run.skip-dirs-use-default` is deprecated, please use `issues.exclude-dirs-use-default`.
WARN [config_reader] The configuration option `output.format` is deprecated, please use `output.formats`
WARN [config_reader] The configuration option `linters.govet.check-shadowing` is deprecated. Please enable `shadow` instead, if you are not using `enable-all`.
WARN [config_reader] The configuration option `linters.errcheck.ignore` is deprecated, please use `linters.errcheck.exclude-functions`.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -57,9 +57,6 @@ output:
 # all available settings of specific linters
 linters-settings:
   govet:
-    # report about shadowed variables
-    check-shadowing: true
-
     # settings per analyzer
     settings:
       printf: # analyzer name, run `go tool vet help` to see all analyzers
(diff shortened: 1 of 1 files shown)


codeboten:
Note that shadowing configuration is no longer needed as it's enabled via `enable-all`

songy23 (approved):

TylerHelmuth (approved):

PR #10215, opened 2024-05-23
[chore] Remove unused ocRegistry, no longer used

bogdandrutu:


songy23 (approved):

codeboten (commented):
I thought the ocregistry was still used by some of the contrib components that use opencensus, is that no longer the case?

bogdandrutu:
The stats module in opencensus it is it own registry, this registry was only for the process metrics :)

codeboten (approved):
My bad, I was confusing this producer registration for the one in the exporter configuration https://github.com/open-telemetry/opentelemetry-collector/blob/570a425961a1a9472326565d2f0ad82d66e3136c/service/internal/proctelemetry/config.go#L168

Issue #10207, opened 2024-05-22
Proposal: split signal-aware packages into per-signal subpackages


dmathieu:
**Is your feature request related to a problem? Please describe.**

As we're planning the addition of profiles into the collector, we will need to add new interfaces into several packages (`consumer`, `receiver`, `processor`, `connector`).
Because the profiles signals is experimental, we would like to avoid making changes to those APIs that would be making them unstable.

**Describe the solution you'd like**

Split each of the packages that have interfaces specific to signals into their own module/package.
We would start with `consumer`, as that API is the closest to 1.0. But each package that has a signal-specific implementation would be impacted.

We could do this two ways:

* A top-level package per signal, with subpackages per feature:`traces/consumer`, `traces/receiver`.
* A top-level package per feature, with subpackages per signal: `consumer/traces`, `receiver/traces`.

With this change, any new signal added in the future could much more easily implement its own interfaces even though the main packages are stable.

**Describe alternatives you've considered**

The alternatives could be to:

* Put the new profiles interfaces directly in the existing packages, which may cause API stability issues.
* Create new packages for the experimental profiles packages, and then move them into the main one when they become stable, which would force folks to update their import paths.

dmathieu:
cc @mx-psi

dmathieu:
If we do this, there will be the question of deprecating/removing the current interfaces.
Since those packages haven't reached 1.0, it should be doable.

djaglowski:
How would this work for connectors, which depend on pairs of signal types?

dmathieu:
It looks like connectors have a specific stability per signal/signal (https://github.com/open-telemetry/opentelemetry-collector/blob/main/connector/connector.go#L108-L118).
So maybe that module could remain as it is. I don't know how it would work once it reaches 1.0 though.

mx-psi:
For consumer, I think it should be easy to do the split and that we should go ahead with it. That way we avoid changing import paths once we move profiles to stable. I think it is also more pressing to decide this for consumer, since it's the one that is closest to 1.0.

For the component types, I think we have more time to think about those, so it may be good if the profiling folks can do an RFC for this part or something similar. There also seem to be several independent pieces that we want to answer here:
- Should we split the `<componentkind>.Factory`? (leaning towards no here since it's too much work for only a small benefit, but we need to figure out the way of adding profiles support, maybe with an optional interface?)
[...]

PR #10220, opened 2024-05-24
Switch consumer to per-signal package

dmathieu:
#### Description

This is a proposal showing the changes in the `consumer` package, as described in #10207.
With this change, the consumer package (and subpackages) can become stable, and new signals (such as profiles) can be added as new modules that aren't stable.

#### Testing

This is unit-tested.


bogdandrutu (commented):
Feedback about implementation (does not mean I agree with the split decision):
1. `clog` - is a bad name since we may have to use something like this for component as well and may confuse between consumer vs component, maybe `conslog`?
2. Do we also need to split `consumertest` package?  Do we need to rename it to be consistent to something like `ctest` (or `constest`)?
3. What about `consumererror`? Same question about naming.

dmathieu:
I have renamed `clog` to `conslog` (did you want the same rename for `ctrace` and `cmetric`?)
I have also split `consumertest` into per-signal packages. I'm looking into `consumererror`.

PR #10222, opened 2024-05-25
Make DataType a separate type within Component

ankitpatel96:
#### Description
Followup from https://github.com/open-telemetry/opentelemetry-collector/pull/10069

The end goal of this PR is to separate DataType from Type.  DataType becomes an enum for Traces, Logs, and Metrics.  

Context (to quote @mx-psi):
> A component.ID currently is a pair of component.Type and a (possibly empty) name for the component (e.g. otlp/name). We use the fact that component.DataType is also a component.Type to also use it for pipeline names (e.g. traces/sampled).

To solve this, I introduce a `component.PipelineID` as suggested by @TylerHelmuth.  This `PipelineID` is very similar to `component.ID` - it just only lets you use a `DataType` as the `Type`.

The PR is large but hundreds of these changes are just changing test files to use a `DataType` instead of a `Type` - I swear its not as bad as it looks.  Most of the actual changes are in `component`, `connector` and `service/internal/graph/`

Followup Work:
I wanted to move `PipelineID` to service/pipelines - but that would create a circular dependency with `InstanceID`, which uses `PipelineID` but lives in `component`.  I want to fix that.

I also want to rename `DataType` to `Signal` or something like that, and move it out to its own package.

I didn't do either of those things in this PR since it's already huge.  

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9429

#### Testing
Unit tests


ankitpatel96:
This is changing an API - and it seems like I will have to change contrib in lockstep with this change (bunch of tests are failing).  Any advice on how to do that would be welcome!

component/config.go
```diff
@@ -160,20 +163,40 @@ func MustNewType(strType string) Type {
...
+	case "metrics":
+		return DataTypeMetrics, nil
+	case "logs":
+		return DataTypeLogs, nil
+	case "traces":
+		return DataTypeTraces, nil
+	default:
+		return "", fmt.Errorf("invalid data type %s", ty)
+	}
+}
```
mx-psi, line 202: This smells like an [`encoding.TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler) implementation

component/pipeline.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+type PipelineID struct {
+	typeVal DataType `mapstructure:"-"`
+	nameVal string   `mapstructure:"-"`
```
mx-psi, line 14: This should be unnecessary, private fields are not modified by mapstructure (reflection in Go does not allow that):

```suggestion
	typeVal DataType
	nameVal string
```

component/pipeline.go
```diff
@@ -0,0 +1,85 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package component // import "go.opentelemetry.io/collector/component"
```
djaglowski, line 4: The concept of a pipeline seems a bit out of place in the component package. Have we considered what it would look like to create a new `pipeline` package?

ankitpatel96: I totally agree - and I want to move this out.  Like I mentioned above:
> Followup Work: I wanted to move PipelineID to service/pipelines - but that would create a circular dependency with InstanceID, which uses PipelineID but lives in component.

I think service/pipelines should contain this and should depend on component, but I just didn't want to make those changes in this PR because its already long enough.

TylerHelmuth: Unfortunately if we put this in `component` and it gets released before the followup PR we'll have breaking changes when we remove it.

Since this PR already updates service to use this concept of pipeline, let's add PipelineID to service now.

component/component.go
```diff
@@ -189,10 +189,3 @@ type CreateDefaultConfigFunc func() Config
 func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
 	return f()
 }
-
-// InstanceID uniquely identifies a component instance
-type InstanceID struct {
```
mx-psi, line 194: Can we add a 

```go
// Deprecated use [<ref to new type>]
type InstanceID = <ref to new type>
```

here?

mx-psi (commented):
Looking at the dependency graph, I think `InstanceID` should not be part of the `service` module. Either we make `service/pipelines` into a separate module or we move this to a separate `pipelines` module altogether.

component/component.go
ankitpatel96: Can I?  That would make component dependent on pipelines... which is most definitely circular

ankitpatel96:
I moved InstanceID and PipelineID into a new pipeline module!

component/component.go
mx-psi: We can use an internal module and make this and the actual type aliases of a symbol in that internal module until we remove it from `component`:

```go
// package internal

type InstanceID struct { /* actual definition goes here *? } 
```

```go
// package component

// Deprecated: [v0.103.0] use pipelines.InstanceID
type InstanceID = internal.InstanceID
```

```go
// package pipelines

type InstanceID = internal.InstanceID
```

ankitpatel96: I don't think this is going to work - 

Pipeline depends on internal, which depends on pipeline since InstanceID depends on pipeline.ID

If I move pipeline.ID into this internal package, we solve that problem but then the component.InstanceID shim doesn't work - that means component imports internal and then internal imports component (InstanceID contains component.ID and component.Kind).  

Open to any ideas to fixing this..

Side note - this internal package has to be collector/internal/<blah> since it needs to be accessible by collector/component and collector/pipeline.

mx-psi: A solution here would be to also move `component.ID` and `component.Kind` to internal and make shims also for these two on `component`. It sounds like a pain, so let's first discuss the rest of the PR

ankitpatel96: Right that would probably work... but I do question whether its worth it for stability concerns.  Moving component.ID to internal/ would make component dependent on the root collector module, thus adding that dependency to almost every other module.

I don't think the blast radius of this change is especially large for contrib and I think, in this case, it makes sense to keep this change as is and not try to provide backwards compatible import shims.

ankitpatel96:
Bump - would really appreciate a review here before I do another merge to fix all these conflicts

component/config.go
```diff
@@ -147,20 +150,44 @@ func MustNewType(strType string) Type {
...
+		*dt = DataTypeMetrics
+		return nil
+	case "logs":
+		*dt = DataTypeLogs
+		return nil
+	case "traces":
+		*dt = DataTypeTraces
+		return nil
+	default:
+		return fmt.Errorf("invalid data type %s", strText)
```
mx-psi, line 191:
```suggestion
		return fmt.Errorf("invalid data type %q", strText)
```

component/identifiable.go
```diff
@@ -9,12 +9,13 @@ import (
 	"strings"
 )
 
-// typeAndNameSeparator is the separator that is used between type and name in type/name composite keys.
-const typeAndNameSeparator = "/"
+// TypeAndNameSeparator is the separator that is used between type and name in type/name composite keys.
+const TypeAndNameSeparator = "/"
```
mx-psi, line 13: Maybe this should be renamed since we also use it for separating `DataType` and pipeline name. Alternatively, move it to internal and make it private

connector/logs_router.go
```diff
@@ -8,16 +8,16 @@ import (
...
+	"go.opentelemetry.io/collector/pipeline"
 )
 
 // LogsRouterAndConsumer feeds the first consumer.Logs in each of the specified pipelines.
 type LogsRouterAndConsumer interface {
 	consumer.Logs
-	Consumer(...component.ID) (consumer.Logs, error)
-	PipelineIDs() []component.ID
+	Consumer(...pipeline.ID) (consumer.Logs, error)
+	PipelineIDs() []pipeline.ID
```
mx-psi, line 20: This (and the ones on metrics and traces router) are breaking changes, it seems hard to work around them so we can just add an API changelog entry

extension/extensiontest/statuswatcher_extension.go
```diff
@@ -21,7 +22,7 @@ func NewStatusWatcherExtensionCreateSettings() extension.Settings {
 
 // NewStatusWatcherExtensionFactory returns a component.ExtensionFactory to construct a status watcher extension.
 func NewStatusWatcherExtensionFactory(
-	onStatusChanged func(source *component.InstanceID, event *component.StatusEvent),
+	onStatusChanged func(source *pipeline.InstanceID, event *component.StatusEvent),
```
mx-psi, line 25: This is also a breaking change, but if we end up not having a type alias for `component.InstanceID`, users of this API are going to have a build time error regardless so they will be 'notified' that way. We can list it in the API changelog as a breaking change as well.

pipeline/pipeline.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// MarshalText implements the encoding.TextMarshaler interface.
+// This marshals the type and name as one string in the config.
+func (id ID) MarshalText() (text []byte, err error) {
+	return []byte(id.String()), nil
+}
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface.
+func (id *ID) UnmarshalText(text []byte) error {
```
mx-psi, line 46: We could have an internal package that has this implementation and allows it to be reused by both `component.ID` and `pipeline.ID`. One way of doing this would be having a function that returns two strings and we can reuse the parsing logic.

service/extensions/extensions.go
```diff
@@ -118,7 +119,7 @@ func (bes *Extensions) NotifyConfig(ctx context.Context, conf *confmap.Conf) err
 	return errs
 }
 
-func (bes *Extensions) NotifyComponentStatusChange(source *component.InstanceID, event *component.StatusEvent) {
+func (bes *Extensions) NotifyComponentStatusChange(source *pipeline.InstanceID, event *component.StatusEvent) {
```
mx-psi, line 122: Add to changelog as breaking change as well

mx-psi (commented):
I think this PR makes sense to merge with this public API. There are a bunch of breaking changes which are hard to work around in some instances. IMO we should
1. document the list of breaking changes on the changelog 
2. evaluate the impact of breaking changes looking at the contrib tests and think about adding two-step deprecation process for the more impactful ones

pipeline/pipeline.go
ankitpatel96: I don't know if this is actually desirable - the root module (which internal is a part of) already imports component, so this would create a dependency loop.  It shouldn't create an import loop but I'm still not sure if this is desirable.

TylerHelmuth (commented):
Finally got a chance to take a look at this, and now I can see that my original suggestion did not factor in `InstanceID`, which even further complicates this change.

Being able to see this change now in its entirety makes me rethink if this change is valuable.  While conceptually I think it is nice, to an end-user this changes nothing. I am concerned about the amount of time we'll spend dealing with this breaking change for what essentially ends up being a rename (a nice rename, but a rename non-the-less.

If, after 1.0, we decide this is valuable, it could be something that is considered for 2.0, but at this point we've been using `component.InstanceID`, `component.Type`, and `component.DataType` successfully without end-users (or component authors) complaining. 

At this point I suggest we abandon this change and call https://github.com/open-telemetry/opentelemetry-collector/issues/9429 completed as "we decided this wasn't worth it".  @ankitpatel96 thank you for your effort on this PR, seeing it all laid out really helped.

PR #10224, opened 2024-05-27
[chore] fix schema for telemetry/metrics/<metric-name>

jpkrohling:
Apparently there's an error on the schema for the telemetry node: looking at a working example, I see this:

```
telemetry:
  metrics:
    processor_tail_sampling_sampling_decision_latency:
      description: Latency (in microseconds) of a given sampling policy
      unit: µs
      enabled: true
      histogram:
        value_type: int
```

However, the current schema seems to indicate that the `metrics` node shouldn't be there. This PR brings the schema to be in-line with the code.

### DIFF
--- a/cmd/mdatagen/metadata-schema.yaml
+++ b/cmd/mdatagen/metadata-schema.yaml
@@ -130,31 +130,32 @@ tests:
 # Optional: map of metric names with the key being the metric name and value
 # being described below.
 telemetry:
-  <metric.name>:
-    # Required: whether the metric is collected by default.
-    enabled: bool
-    # Required: metric description.
-    description:
-    # Optional: extended documentation of the metric.
-    extended_documentation:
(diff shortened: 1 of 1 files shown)


cmd/mdatagen/metadata-schema.yaml
```diff
@@ -130,31 +130,32 @@ tests:
...
-    unit:
-    # Required: metric type with its settings.
-    <sum|gauge|histogram>:
-      # Required: type of number data point values.
-      value_type: <int|double>
-      # Required for sum metric: whether the metric is monotonic (no negative delta values).
-      monotonic: bool
-    # Optional: array of attributes that were defined in the attributes section that are emitted by this metric.
-    attributes: [string]
+  metrics:
```
jpkrohling, line 133: This is really the only difference here. Everything else was just indented under `metrics`.

codeboten (approved):

Issue #10226, opened 2024-05-27
[exporter/debug] Add option to not use collector's internal logger for output


andrzej-stencel:
**Is your feature request related to a problem? Please describe.**

Currently (v0.101.0), the [Debug exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/debugexporter/README.md) uses the collector's [internal logger](https://opentelemetry.io/docs/collector/internal-telemetry/) for output.
This comes with several consequences:

1. The exporter's output is prefixed and suffixed by the internal logger's data.
   - The prefix is the current timestamp and the `info` level marker, similar to `2023-11-10T22:49:03.510-0600    info`
   - The suffix is the exporter's metadata in JSON format, similar to `{"kind": "exporter", "data_type": "traces", "name": "debug"}`
1. The exporter's output is sent to the same destination as the collector's logs ([stderr by default](https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-logs)). It is not possible to send the Debug exporter's output to a different stream than the collector's internal logs.
1. The output from the Debug exporter is sent to the internal logger with the log level `INFO`. This means that if user configures the `service.telemetry.logs.level` to `WARN` or `ERROR`, they will not see the output from the Debug exporter. As a result, it is not possible to configure the collector to see the Debug exporter's output and mute collector's INFO logs at the same time.
1. The exporter's output is subject to the internal logger's sampling, which is separate from the sampling defined in the exporter's configuration. See https://github.com/open-telemetry/opentelemetry-collector/issues/9921 and the note in the "Workaround" section of the issue.

**Describe the solution you'd like**

Add configuration option `use_internal_logger` that would make it possible to opt out of using the internal logger and instead use a logger configured separately from the collector's internal logger.

**Describe alternatives you've considered**

The only alternative I see is not use the Debug exporter and instead build a separate exporter for this, or use existing exporter like the [File exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/fileexporter/README.md).

mx-psi:
Apart from adding this option, I think we should vote on whether we want to change the default behavior. On open-telemetry/opentelemetry-collector/issues/5652 I proposed this change, but since:
- we changed the name from `logging` to `debug` which makes it less salient that this is logging
- on #8134 we enabled sampling generally, which caused #9921

[...]

PR #10227, opened 2024-05-27
[exporter/debug] add option `use_internal_logger`

andrzej-stencel:
#### Description

Adds a new configuration option `use_internal_logger` to the Debug exporter, that makes it possible to prevent unwanted side effects of the exporter using the collector's internal logger.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10226

#### Testing

Updated unit tests to cover the new config option.

#### Documentation

Added documentation for the feature.


exporter/debugexporter/README.md
```diff
@@ -32,6 +32,7 @@ The following settings are optional:
   messages are logged (every Mth message is logged). Refer to [Zap
   docs](https://godoc.org/go.uber.org/zap/zapcore#NewSampler) for more details.
   on how sampling parameters impact number of messages.
+- `use_internal_logger` (default = `true`): if set, uses the collector's internal logger for output. See [below](#using-the-collectors-internal-logger) for description.
```
songy23, line 35:
```suggestion
- `use_internal_logger` (default = `true`): uses the collector's internal logger for output. See [below](#using-the-collectors-internal-logger) for description.
```

Default is true so I don't it needs to set explicitly

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+When `use_internal_logger` is set to `true` (the default), the exporter uses the collector's [internal logger][internal_telemetry] for output.
+This comes with several consequences:
+
+1. The exporter's output is prefixed and suffixed by the internal logger's data.
+   - The prefix is the current timestamp and the `info` level marker, similar to `2023-11-10T22:49:03.510-0600    info`
+   - The suffix is the exporter's metadata in JSON format, similar to `{"kind": "exporter", "data_type": "traces", "name": "debug"}`
+1. The exporter's output is sent to the same destination as the collector's logs. By default, it is the [standard error][stderr] stream,
+   configurable via the [service.telemetry.logs.output_paths][configure_internal_logs] option.
+   It is not possible to send the Debug exporter's output to a different stream than the collector's internal logs.
+1. The output from the Debug exporter is sent to the internal logger with the log level `INFO`.
```
songy23, line 122: Can the log level be made configurable?

exporter/debugexporter/config.go
```diff
@@ -30,6 +30,9 @@ type Config struct {
 
 	// SamplingThereafter defines the sampling rate after the initial samples are logged.
 	SamplingThereafter int `mapstructure:"sampling_thereafter"`
+
+	// UseInternalLogger defines whether he exporter send the output to the collector's internal logger.
```
songy23, line 34:
```suggestion
	// UseInternalLogger defines whether the exporter sends the output to the collector's internal logger.
```

songy23 (commented):
cc @mx-psi

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+## Using the collector's internal logger
+
+When `use_internal_logger` is set to `true` (the default), the exporter uses the collector's [internal logger][internal_telemetry] for output.
+This comes with several consequences:
+
+1. The exporter's output is prefixed and suffixed by the internal logger's data.
+   - The prefix is the current timestamp and the `info` level marker, similar to `2023-11-10T22:49:03.510-0600    info`
+   - The suffix is the exporter's metadata in JSON format, similar to `{"kind": "exporter", "data_type": "traces", "name": "debug"}`
+1. The exporter's output is sent to the same destination as the collector's logs. By default, it is the [standard error][stderr] stream,
+   configurable via the [service.telemetry.logs.output_paths][configure_internal_logs] option.
```
tiffany76, line 120: Just an FYI: we recently had a conversation about how best to notate nested keys in YAML paths. The ultimate decision was to use double colons (::) to avoid any confusion with namespaces using dot notation. 

See [this issue comment](https://github.com/open-telemetry/opentelemetry.io/issues/4484#issuecomment-2117608625) for details and [this tip](https://opentelemetry.io/docs/collector/configuration/#location) that was just added to the docs.

Thanks!

exporter/debugexporter/README.md
andrzej-stencel: You mean when using the internal logger, the user might want to "log" the exporter's output with a higher log level, to prevent its output to be hidden when the user specified a higher log level on the collector? I suppose it might be a workaround for this issue.

I would not use this workaround, as I don't think the exporter should be using the internal logger.

exporter/debugexporter/README.md
andrzej-stencel: Oh, right, makes total sense. Will update the docs. Thank you for pointing this out!

andrzej-stencel: Fixed in https://github.com/open-telemetry/opentelemetry-collector/pull/10227/commits/4f8f733841687a7e97917213df9ebdaa49b64684.

exporter/debugexporter/README.md
mx-psi, line 35: If we want this component to have a separate logger, do we expect people to want to configure this separate logger? If so, should we have a `logger` section instead to group features related to configuring the logger?

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+   configurable via the [service::telemetry::logs::output_paths][configure_internal_logs] option.
+   It is not possible to send the Debug exporter's output to a different stream than the collector's internal logs.
+1. The output from the Debug exporter is sent to the internal logger with the log level `INFO`.
+   This means that if user configures the `service::telemetry::logs::level` to `WARN` or `ERROR`,
+   they will not see the output from the Debug exporter.
+   As a result, it is not possible to configure the collector to see the Debug exporter's output
+   and mute collector's INFO logs at the same time.
+1. The exporter's output is subject to the internal logger's sampling,
+   which is separate from the sampling defined in the exporter's configuration.
+   See https://github.com/open-telemetry/opentelemetry-collector/issues/9921 and the note in the "Workaround" section of the issue.
```
mx-psi, line 129: I am not super at ease with this section. Some of the things mentioned depend on the logger encoding. Some other things may change in the future, since there no guarantee that we will keep the format of logs in the non-machine readable format.

Can we instead just mention that this will use the Collector's internal logger and honor the `service::telemetry::logs` configuration? We can clarify how the Collector internal logger works in a separate place if needed, but I don't think this is the place to do so.

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+
+When `use_internal_logger` is set to `false`, the exporter uses its own logger, independent from the collector's internal logger. This results in the following behavior:
+
+1. The prefix and suffix described above is not present in the exporter's output.
```
mx-psi, line 133:
```suggestion
```

We make no guarantees as to the output of the debug exporter, so we can't guarantee that this will be the case in the future. I would remove this.

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+1. The exporter's output is sent to [standard output][stdout] stream. This will be made configurable in the future. Changing the value of `service::telemetry::logs::output_paths` has no effect on the exporter's output.
```
mx-psi, line 134: I would at least remove the forward-looking statement. We can add information about this once and if we add the feature.
```suggestion
1. The exporter's output is sent to [standard output][stdout] stream. Changing the value of `service::telemetry::logs::output_paths` has no effect on the exporter's output.
```

I would even go as far as just saying:
```suggestion
1. The exporter's output is sent to [standard output][stdout] stream.
```

And just mention that we ignore `service::telemetry::logs` here

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+1. Changing the value of `service::telemetry::logs::level` has no effect on the exporter's output.
```
mx-psi, line 135: Same here, I don't think we should mention this specifically, we should instead mention that we ignore the `service::telemetry::logs` section and all its flags

```suggestion
```

exporter/debugexporter/README.md
```diff
@@ -107,6 +108,38 @@ Attributes:
...
+1. The exporter's sampling is defined by the exporter's configuration options `sampling_initial` and `sampling_thereafter` and nothing else. The configuration in `service::telemetry::logs::sampling` has no effect on the exporter's output.
```
mx-psi, line 136: Same here, 
```suggestion
1. The exporter's sampling is defined by the exporter's configuration options `sampling_initial` and `sampling_thereafter` and nothing else.
```

mx-psi (changes requested):

exporter/debugexporter/README.md
andrzej-stencel: Yeah that works, I'll change the description.

exporter/debugexporter/README.md
andrzej-stencel: I actually think that what the exporter uses under the hood to output data should be hidden from the user as an implementation detail. The user should be able to configure the output of the exporter without thinking if a logging library is being used under the hood or something else. This frees us from being tied to a specific implementation of the output.

exporter/debugexporter/README.md
andrzej-stencel: Updated the documentation in https://github.com/open-telemetry/opentelemetry-collector/pull/10227/commits/02e356d7d6f2f791eff8435b91661cd002c5d3dc, please let me know your thoughts @mx-psi.

exporter/debugexporter/README.md
andrzej-stencel: Removed in https://github.com/open-telemetry/opentelemetry-collector/pull/10227/commits/02e356d7d6f2f791eff8435b91661cd002c5d3dc.

exporter/debugexporter/README.md
andrzej-stencel: Updated in https://github.com/open-telemetry/opentelemetry-collector/pull/10227/commits/02e356d7d6f2f791eff8435b91661cd002c5d3dc.

exporter/debugexporter/README.md
andrzej-stencel: Removed in https://github.com/open-telemetry/opentelemetry-collector/pull/10227/commits/02e356d7d6f2f791eff8435b91661cd002c5d3dc.

exporter/debugexporter/README.md
andrzej-stencel: Removed in https://github.com/open-telemetry/opentelemetry-collector/pull/10227/commits/02e356d7d6f2f791eff8435b91661cd002c5d3dc.

exporter/debugexporter/README.md
andrzej-stencel: I would envision maybe an `output` section that would configure whether you want the resource or scope to be output as separate lines. Potentially also to configure the output more, like define whether timestamps should be output, log severity, etc.

mx-psi (approved):

songy23 (approved):

exporter/debugexporter/factory.go
```diff
@@ -83,12 +84,38 @@ func createLogsExporter(ctx context.Context, set exporter.Settings, config compo
...
+	zapConfig := zap.Config{
+		Level:         zap.NewAtomicLevelAt(zap.InfoLevel),
+		DisableCaller: true,
+		Sampling: &zap.SamplingConfig{
+			Initial:    exporterConfig.SamplingInitial,
+			Thereafter: exporterConfig.SamplingThereafter,
+		},
+		Encoding:      "console",
+		EncoderConfig: encoderConfig,
+		// Send exporter's output to stdout. This should be made configurable.
```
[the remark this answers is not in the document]
jedr: Created https://github.com/open-telemetry/opentelemetry-collector/issues/10472.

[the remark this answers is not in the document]
andrzej-stencel: Haha this was "the other me" that created the issue and the comment. 😅

tiffany76 (approved):

Issue #10218, opened 2024-05-23
[mdatagen] support configuring distributions for histograms


codeboten:
mdatagen supports configuring histograms for internal metrics, it doesn't support setting bucket boundaries for those histograms. This will be needed for some components.

PR #10229, opened 2024-05-27
[mdatagen] add support for bucket boundaries

codeboten:
This allows setting of explicit boundaries for histograms.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10218


cmd/mdatagen/metadata-schema.yaml
```diff
@@ -157,5 +157,7 @@ telemetry:
         value_type: <int|double>
         # Required for sum metric: whether the metric is monotonic (no negative delta values).
         monotonic: bool
+        # Bucket boundaries are only avaiable to set for histogram metrics.
```
codeboten, line 160:
```suggestion
        # Bucket boundaries are only available to set for histogram metrics.
```

PR #10241, opened 2024-05-28
[chore]fix semconv script

codeboten:
The automation script that runs on a cron to auto-generate PRs for new semconv updates has been failing since it was created. This PR fixes a couple of issues w/ the workflow that were preventing it from working.

### DIFF
--- a/.github/workflows/generate-semantic-conventions-pr.yaml
+++ b/.github/workflows/generate-semantic-conventions-pr.yaml
@@ -61,6 +61,7 @@ jobs:
         with:
           repository: open-telemetry/semantic-conventions
           path: tmp-semantic-conventions
+          fetch-tags: true
 
       - name: Update version
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-semantic-conventions-pr.yaml
```diff
@@ -21,15 +21,14 @@ jobs:
...
                                --repo open-telemetry/semantic-conventions \
                                --json tagName \
                                --jq .tagName \
                              | sed 's/^v//')
 
-          rc=$(find semconv -name $latest_version | grep .)
-          if $rc == 0; then
+          found=$(find semconv -name $latest_version)
+          if [[ $found =~ $latest_version ]]; then
             already_added=true
```
jpkrohling, line 32: Does it matter that this is using underscore and the output is defined with dash?

codeboten: It doesn't, you can see where the output is set on line 46:

```
echo "already-added=$already_added" >> $GITHUB_OUTPUT
```

bogdandrutu (approved):

PR #10251, opened 2024-05-29
[PoC] Splitting consumer by signal

mx-psi:
#### Link to tracking issue
Fixes #


codeboten (commented):
Looking at this, I wonder if it would be easier for us to have a `consumerexperimental` package that includes anything that isn't stabilized, and once profiles is stable, we just move it to `consumer`.

Issue #10255, opened 2024-05-29
[exporterhelper/batchsender] batchsender deadlock preventing shutdown


timannguyen:
**Describe the bug**

deadlock during batchsender shutdown

1. main goroutine shutdowns
2. main goroutine close(shutdownch) https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/batch_sender.go#L217
3. ticker goroutine tries to obtain lock https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/batch_sender.go#L69
4. sending goroutine already have the lock and is trying to flush https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/batch_sender.go#L197
5. Deadlock because ticker goroutine will never be able to obtain the lock and to receive resetTimerCh sending goroutine is waiting to push

**Steps to reproduce**

I have a unit test to reproduce the issue https://github.com/timannguyen/opentelemetry-collector/commit/adf2fb9a5c80e2ec29b231c9e0f92f4a7dfb26e3

the mergeFunc would need to take a bit of time in the sendMergeBatch during shutdown to cause this deadlock

**What did you expect to see?**

to shutdown without deadlock

**What did you see instead?**

deadlock when a sending goroutine holding to the lock while the ticket goroutine is trying to get the lock. This prevents shutdown

**What version did you use?**

pdata 1.7.0
otel 0.100.0

**What config did you use?**

**Environment**

MACOS 

ubuntu 

**Additional context**

PR #10258, opened 2024-05-29
[exporterhelper] Fix potential deadlocks in BatcherSender shutdown

dmitryax:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10255

### DIFF
--- a/exporter/exporterhelper/batch_sender_test.go
+++ b/exporter/exporterhelper/batch_sender_test.go
@@ -439,6 +439,63 @@ func TestBatchSender_WithBatcherOption(t *testing.T) {
 	}
 }
 
+// TestBatchSender_ShutdownDeadlock tests that the exporter does not deadlock when shutting down while a batch is being
+// merged.
+func TestBatchSender_ShutdownDeadlock(t *testing.T) {
+	blockMerge := make(chan struct{})
+	waitMerge := make(chan struct{}, 10)
+
+	// blockedBatchMergeFunc blocks until the blockMerge channel is closed
+	blockedBatchMergeFunc := func(_ context.Context, r1 Request, _ Request) (Request, error) {
+		waitMerge <- struct{}{}
+		<-blockMerge
+		return r1, nil
+	}
+
+	bCfg := exporterbatcher.NewDefaultConfig()
+	bCfg.FlushTimeout = 10 * time.Minute // high timeout to avoid the timeout to trigger
+	be, err := newBaseExporter(defaultSettings, defaultDataType, newNoopObsrepSender,
+		WithBatcher(bCfg, WithRequestBatchFuncs(blockedBatchMergeFunc, fakeBatchMergeSplitFunc)))
+	require.NoError(t, err)
+	require.NoError(t, be.Start(context.Background(), componenttest.NewNopHost()))
+
+	sink := newFakeRequestSink()
+
+	// Send 10 concurrent requests and wait for them to start
+	startWG := sync.WaitGroup{}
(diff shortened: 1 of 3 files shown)


TylerHelmuth (approved):

TylerHelmuth:
I'm guessing I missed something since `TestBatchSender_MergeOrSplit` isn't passing anymore

dmitryax:
`TestBatchSender_MergeOrSplit` is good. It's just an accidental debug output, removed https://github.com/open-telemetry/opentelemetry-collector/pull/10261

The actual regression test was still failing. I updated the code, should be good now.

TylerHelmuth (approved):

exporter/exporterhelper/batch_sender_test.go
```diff
@@ -439,6 +439,63 @@ func TestBatchSender_WithBatcherOption(t *testing.T) {
...
+	require.NoError(t, be.Start(context.Background(), componenttest.NewNopHost()))
+
+	sink := newFakeRequestSink()
+
+	// Send 10 concurrent requests and wait for them to start
+	startWG := sync.WaitGroup{}
+	for i := 0; i < 10; i++ {
+		startWG.Add(1)
+		go func() {
+			startWG.Done()
```
crobert-1, line 469:
```suggestion
			defer startWG.Done()
```
This should be after the `be.send` call, right? Maybe use defer instead? Or are you purposefully wanting to continue in the main goroutine before send is complete?

exporter/exporterhelper/batch_sender_test.go
```diff
@@ -439,6 +439,63 @@ func TestBatchSender_WithBatcherOption(t *testing.T) {
...
+		require.Nil(t, be.Shutdown(context.Background()))
+		close(doneShutdown)
+	}()
+	<-startShutdown
+	close(blockMerge)
+	<-doneShutdown
+
+	// The exporter should have sent only one "merged" batch, in some cases it might send two if the shutdown
+	// happens before the batch is fully merged.
+	assert.LessOrEqual(t, uint64(1), sink.requestsCount.Load())
```
crobert-1, line 493: Is it valid to send two batches in some cases, as your comment is referencing? If so, should the assert here be `2` instead of `1`?

exporter/exporterhelper/batch_sender_test.go
dmitryax: No, it should be before send. We need to ensure all the goroutines are started. The send is blocking

exporter/exporterhelper/batch_sender_test.go
dmitryax: It's usually 1 or more in rare cases. LessOrEqual checks that the first argument is less or equal to the second one

exporter/exporterhelper/batch_sender_test.go
crobert-1: > We need to ensure all the goroutines are started. The send is blocking

The goroutines will all be started regardless of where the `startWG.Done()` call is made, if I'm following correctly. The only impact is when the main goroutine will continue after the `startWG.Wait()` call. I'm wondering if it should wait until all sends are complete or not?

dmitryax: the send operations cannot complete until they are unblocked, which we do after the shutdown - that's the purpose of the test case

crobert-1: Okay, thanks for clarifying 👍

PR #10259, opened 2024-05-29
[confmap] Add featuregate to use stable expansion rules

TylerHelmuth:
#### Description

Add a feature gate that provides access to stable expansion rules.  When enabled:

1. `env` is used as the default scheme
2. expandconverter is not used, which means `$ENV` syntax is no longer expanded

#### Link to tracking issue

#### Testing
Tested manually locally


cmd/builder/internal/builder/templates/main.go.tmpl
```diff
@@ -35,14 +35,21 @@ func main() {
...
 
+    {{ if .Distribution.SupportsConfmapFactories }}
+	if confmap.UseStableExpansionRules.IsEnabled() {
+		set.ConfigProviderSettings.ResolverSettings.DefaultScheme = "env"
+	} else {
+		set.ConfigProviderSettings.ResolverSettings.ConverterFactories = []confmap.ConverterFactory{
+			expandconverter.NewFactory(),
+		}
+	}
+    {{- end}}
```
TylerHelmuth, line 51: @mx-psi @evan-bradley this solution has the downside that it will only apply to people that use ocb.  Is that ok?

Does it need to live in `otelcol`, like in `NewCommand`, and be allowed to mess with the incoming settings, ignoring any configured expandconverter?

codeboten (commented):
I'd suggest moving the feature gate to a different package to prevent all the imports this is causing

codeboten (approved):

codeboten (commented):
Lint fails:

```
/home/runner/work/opentelemetry-collector/opentelemetry-collector/.tools/golangci-lint run
Error: configprovider.go:145:51: SA1019: expandconverter.NewFactory is deprecated: [v0.102.0] Expanding $ENV is no longer supported. To expand ${} using env vars use DefaultScheme="env" or the otelcol.useStableExpansionRules feature gate. (staticcheck)
			ConverterFactories: []confmap.ConverterFactory{expandconverter.NewFactory()},
			                                               ^
Error: command_test.go:57:52: SA1019: expandconverter.NewFactory is deprecated: [v0.102.0] Expanding $ENV is no longer supported. To expand ${} using env vars use DefaultScheme="env" or the otelcol.useStableExpansionRules feature gate. (staticcheck)
				ConverterFactories: []confmap.ConverterFactory{expandconverter.NewFactory()},
				                                               ^
Error: command_test.go:91:52: SA1019: expandconverter.NewFactory is deprecated: [v0.102.0] Expanding $ENV is no longer supported. To expand ${} using env vars use DefaultScheme="env" or the otelcol.useStableExpansionRules feature gate. (staticcheck)
```

TylerHelmuth:
I believe the actual deprecation depends on https://github.com/open-telemetry/opentelemetry-collector/pull/10281

cmd/builder/internal/builder/templates/main.go.tmpl
evan-bradley: If possible, this is the process I think would work best:

1. Implement the feature gate in `NewCommand`/`NewCollector`/etc. with what you have here.
2. Turn the feature gate on by default.
3. Allow configuring the default scheme in an ocb manifest.
4. Remove the feature gate and the expand converter from ocb in one go, since there's no way to configure converters in the ocb manifest right now.

evan-bradley: If the two behaviors can coexist, I don't think we should remove a configured expand converter.

TylerHelmuth: Thinking about this again I propose a alternative:

Since this is a change to how `CollectorSettings` is configured, we handle the transition for only our own usage (ocb) - anyone using `NewCommand` is responsible for updating `CollectorSettings` on their own, following the deprecation of the `expandconverter`.

This approach would mean moving the featuregate into the `main.go` generated by ocb, deprecating `expandconverter` and then as part of that deprecation following a standard feature gate process.

This approach saves us the trouble of guessing how other users of `NewCommand` want to handle the deprecation of the `expandconverter` while at the same time enforcing our own Provider expectations established in the rfc.

TylerHelmuth: We would first have to handle [this TODO](https://github.com/open-telemetry/opentelemetry-collector/blob/ed767dc21968b2b2a2ba2ac95f40fe510de9de79/otelcol/command.go#L56) first, where we start requiring incoming `CollectorSettings` to have the providers/converts supplied, making `NewCommand` more dependent on a full set of settings when invoked.

.chloggen/confmap-add-expand-featuregate-2.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: expandconverter
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Expanding `$ENV` is no longer supported.  Use `DefaultScheme="env"` or the `otelcol.useStableExpansionRules` feature gate to expand `${}`.
```
mx-psi, line 10: I find the

>  Expanding `$ENV` is no longer supported.

confusing. If you use this syntax after this change, it will still work, right? It's only if you enable `otelcol.useStableExpansionRules` that this will stop working.

.chloggen/confmap-add-expand-featuregate.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `otelcol.useStableExpansionRules` feature gate to allow enabling stable expansion rules.
```
mx-psi, line 10: I am not sure what the end result of the discussion about one vs several feature gates for this is. If we are going to have several feature gates, then I would change the name of this one, 'stable expansion rules' is a concept that needs some explaining

mx-psi (commented):
I think it would be good to add some config provider tests that show what the change in behavior is before and after enabling the feature gate

cmd/builder/internal/builder/templates/main.go.tmpl
TylerHelmuth: @evan-bradley @mx-psi @codeboten I am still working towards this PR but trying to unblock/simplify it by working on https://github.com/open-telemetry/opentelemetry-collector/issues/10290 first.

TylerHelmuth: Thinking about this some more, I believe by moving the feature gate to builder and only controlling our own use of `expandconverter` allows this PR and https://github.com/open-telemetry/opentelemetry-collector/pull/10359 to happen in parallel.

.chloggen/confmap-add-expand-featuregate.yaml
TylerHelmuth: I'd like to handle all the syntax changes for our own collectors in a single PR (the fact that we provide defaults for others will be handled via https://github.com/open-telemetry/opentelemetry-collector/pull/10359).  What should we name the feature gate?

.chloggen/confmap-add-expand-featuregate-2.yaml
TylerHelmuth: That is correct, I'll update this note.

TylerHelmuth:
This PR is currently under the assumption that we'll stop providing default providers and converters in NewCommand. If we decide to keep doing that, I'll open a new PR that keeps the feature gate in otelcol and does the same logic as the builder when setting defaults.

TylerHelmuth:
Created https://github.com/open-telemetry/opentelemetry-collector/pull/10361 as an alternative

.chloggen/confmap-add-expand-featuregate.yaml
mx-psi: In terms of behavior, the end result for this is:
- The `${ENV}` syntax starts working the same as `${env:ENV}`
- The `$ENV` syntax is no longer supported

For an end-user, I think it would be more meaningful to talk about "unifying syntax" maybe? There are some semantic differences implied by this, which should for the most part be invisible for end-users

mx-psi: > I believe ${ENV} worked the same as ${env:ENV} already,

I don't think this is the case, see https://github.com/open-telemetry/opentelemetry-collector/issues/8565#issuecomment-1921361955 for an actual difference. This relates to `WeaklyTypedInput`

TylerHelmuth: @mx-psi that'd be a tough distinction to describe in the feature gate description, do you think the link I added to the RFC is enough?

mx-psi: Yeah that works for me, I think it's much clearer now thanks!

.chloggen/confmap-add-expand-featuregate.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `otelcol.unifyEnvVarExpansion` feature gate to allow enabling Collector/Configuration SIG environment variable expansion rules.
```
mx-psi, line 10: No strong opinion here (and I get why you are using `otelcol`), but it sounds more intuitive to me to use the `confmap` namespace here (since this is a behavior change in confmap)
```suggestion
note: Add `confmap.unifyEnvVarExpansion` feature gate to allow enabling Collector/Configuration SIG environment variable expansion rules.
```

cmd/builder/internal/builder/templates/main.go.tmpl
```diff
@@ -10,13 +10,21 @@ import (
...
+	"go.opentelemetry.io/collector/featuregate"
 	{{- range .Providers}}
 	{{.Name}} "{{.Import}}"
 	{{- end}}
 	{{- end}}
 	"go.opentelemetry.io/collector/otelcol"
 )
 
+{{ if .Distribution.SupportsConfmapFactories }}
+var useUnifiedEnvVarExpansionRules = featuregate.GlobalRegistry().MustRegister("otelcol.unifyEnvVarExpansion",
```
mx-psi, line 22: (If you accept my other suggestion)
```suggestion
var useUnifiedEnvVarExpansionRules = featuregate.GlobalRegistry().MustRegister("confmap.unifyEnvVarExpansion",
```

mx-psi (approved):

cmd/builder/internal/builder/templates/main.go.tmpl
```diff
@@ -38,14 +46,21 @@ func main() {
...
-				},
 			},
 		},
 		{{- end}}
 	}
 
+    {{ if .Distribution.SupportsConfmapFactories }}
+    if useUnifiedEnvVarExpansionRules.IsEnabled() && set.ConfigProviderSettings.ResolverSettings.DefaultScheme == "" {
+        set.ConfigProviderSettings.ResolverSettings.DefaultScheme = "env"
+    } else {
```
evan-bradley, line 57:
```suggestion
    if useUnifiedEnvVarExpansionRules.IsEnabled() {
        if set.ConfigProviderSettings.ResolverSettings.DefaultScheme == "" {
            set.ConfigProviderSettings.ResolverSettings.DefaultScheme = "env"
        }
    } else {
```
Could we separate these conditions? I'd prefer if setting a default scheme didn't automatically add back the expand converter even if the flag is enabled.

TylerHelmuth:
So it turns out featuregates don't work in `main.go` because it happens before flags are read in.  @codeboten and I found this while writing an integration test.

So I need to rework this PR to do the manipulation inside otelcol like I originally had it 😢

TylerHelmuth:
This solution doesn't work, need https://github.com/open-telemetry/opentelemetry-collector/pull/10391 instead

PR #10260, opened 2024-05-30
WIP do not merge - optional config fields

yurishkuro:
#### Description

Experiment with support optional config structs

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/issues/10266


yurishkuro:
@evan-bradley tests wouldn't be a problem, but I ran into a roadblock that I described in https://github.com/open-telemetry/opentelemetry-collector/issues/10266#issuecomment-2139700225

yurishkuro:
`WithDefault[T]()` was a good idea from @mx-psi, but it didn't work either. When an item is defined in YAML but without any value, i.e. `item:` with no content, then `mapstructure` treats the value as nil, and does not perform any processing. As a result, the custom `Unmarshal()` is still not being called.

See test https://github.com/open-telemetry/opentelemetry-collector/pull/10260/files#diff-24c5961a7871021bd9801008aa5f1fad72efd43f8783a00c7d09dc52c4c64e0dR81

I opened a ticket upstream https://github.com/go-viper/mapstructure/issues/37

yurishkuro:
keep

mx-psi:
I'll have time to look into this next week, apologies for the delay

mx-psi:
Waiting on go-viper/mapstructure/pull/42

yurishkuro:
https://github.com/go-viper/mapstructure/pull/42 has been merged, will come out in v2.2.0, then we I can pick this back up.

yurishkuro:
Pending another upstream fix https://github.com/go-viper/mapstructure/pull/45

yurishkuro:
@mahadzaryab1 fyi this is the main PR I was working on, which needed changes to mapstructure lib. The fun continues, now with contrib packages failing.

mackjmr:
@yurishkuro I tested this PR using type `optional.Optional[int]` for [MaxIdleConns](https://github.com/open-telemetry/opentelemetry-collector/blob/d6f568dd5b6ddfd87a7c6350545cb818f902a06f/config/confighttp/confighttp.go#L74) in confighttp but get the following error:
```
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):
error decoding 'exporters': error reading configuration for "otlphttp": decoding failed due to the following error(s):
'max_idle_conns' expected a map, got 'int'
```

It seems like Optionals [Unmarshal](https://github.com/open-telemetry/opentelemetry-collector/blob/7157a80f52e96fc4364572698ea4cc6ca2699873/confmap/optional/optional.go#L52) is never called due to the following logic in the [Hookfunc](https://github.com/open-telemetry/opentelemetry-collector/blob/7157a80f52e96fc4364572698ea4cc6ca2699873/confmap/confmap.go#L456-L458).

I updated ClientConfig as so:
```
// ClientConfig defines settings for creating an HTTP client.
type ClientConfig struct {
...
	// MaxIdleConns is used to set a limit to the maximum idle HTTP connections the client can keep open.
	// By default, it is set to 100.
	MaxIdleConns optional.Optional[int] `mapstructure:"max_idle_conns"`
...
}
```

and updated `otlphttpexporter`s createDefaultConfig (which I am testing with) as so:
```
func createDefaultConfig() component.Config {
	return &Config{
		RetryConfig: configretry.NewDefaultBackOffConfig(),
		QueueConfig: exporterhelper.NewDefaultQueueSettings(),
		Encoding:    EncodingProto,
		ClientConfig: confighttp.ClientConfig{
			Endpoint: "",
			Timeout:  30 * time.Second,
			Headers:  map[string]configopaque.String{},
			// Default to gzip compression
			Compression: configcompression.TypeGzip,
			// We almost read 0 bytes, so no need to tune ReadBufferSize.
			WriteBufferSize: 512 * 1024,
			MaxIdleConns: optional.WithDefault(
				func() int {
					return 555
				},
			),
		},
	}
}
```

Is there something I am doing wrong here ?

evan-bradley:
@yurishkuro Just checking in, any updates on this? I have some time to dig into it this week, let me know if there's anything you'd like me to investigate.

mahadzaryab1:
I've got a PR for the upstream patch open at https://github.com/go-viper/mapstructure/pull/52. To test this patch out, I believe we'll need to re-write the existing hooks in `confmap` to return `reflect.Value` instead of the actual values to be able to decode nil values.

PR #10263, opened 2024-05-30
Split receiver into per-functionality files

dmathieu:
#### Description

This does no behavior change. It splits the receiver package into smaller, more readable files that are constrained to a single functionality, like we already do in other packages such as `consumer`.

#### Testing

This is unit-tested.


dmathieu:
This shouldn't need a changelog entry.

dmathieu:
Discarded in https://github.com/open-telemetry/opentelemetry-collector/issues/10207

PR #10269, opened 2024-05-30
[chore] Remove redundant wait group from a test

dmitryax:
https://github.com/open-telemetry/opentelemetry-collector/pull/10258/files#r1621017272 made me wonder if the wait group is really necessary. We have another synchronization that waits for at least two requests to enter the merge function, which should be enough. So, we don't necessarily need this wait group.


bogdandrutu:
> We have another synchronization that waits for at least two requests to enter the merge function, which should be enough. So, we don't necessarily need this wait group.

This may fail goleak, because some of the started go routines may not finish until the end of the test.

dmitryax:
> This may fail goleak, because some of the started go routines may not finish until the end of the test.

Makes sense. I simplified the test to keep only two goroutines that must finish once the batcher is shut down. This also makes the test deterministic

codeboten (approved):

PR #10273, opened 2024-05-30
[service] use mdatagen for service metrics

codeboten:
This reverts the reverts https://github.com/open-telemetry/opentelemetry-collector/pull/10271 and adds a mechanism to skip adding a create settings method for the service package component test. Will need to figure out if servicetelemetry.TelemetrySettings should be renamed to fit w/ the other CreateSettings structs before removing this check.


service/metadata.yaml
```diff
@@ -0,0 +1,68 @@
...
+  stability:
+    development: [traces, metrics, logs]
+  distributions: [core, contrib]
+
+tests:
+  goleak:
+    ignore:
+      top:
+        # See https://github.com/census-instrumentation/opencensus-go/issues/1191 for more information.
+        - "go.opencensus.io/stats/view.(*worker).start"
```
jpkrohling, line 14: This is because of ./internal/proctelemetry/config.go only, right?

jpkrohling (approved):

service/metadata.yaml
codeboten: correct

PR #10275, opened 2024-05-31
[confighttp] add confighttp.NewDefaultServerConfig()

atoulme:
#### Description
add a new default method to instantiate the HTTP server config.

#### Link to tracking issue
#9655


songy23 (approved):

config/confighttp/confighttp.go
```diff
@@ -281,6 +281,13 @@ type ServerConfig struct {
 	ResponseHeaders map[string]configopaque.String `mapstructure:"response_headers"`
 }
 
+// NewDefaultServerConfig returns ServerConfig type object with default values.
+// Currently, config options are not added as they are initialized with 'zero value' by GoLang as default.
```
atoulme, line 285:
```suggestion
// Currently, config options are all initialized as zero values.
```

config/confighttp/confighttp.go
```diff
@@ -281,6 +281,13 @@ type ServerConfig struct {
 	ResponseHeaders map[string]configopaque.String `mapstructure:"response_headers"`
 }
 
+// NewDefaultServerConfig returns ServerConfig type object with default values.
+// Currently, config options are all initialized as zero values.
+// We encourage to use this function to create an object of ServerConfig.
+func NewDefaultServerConfig() ServerConfig {
+	return ServerConfig{}
```
jpkrohling, line 288: Can I suggest setting the write timeout here already on this first implementation? This way, once we set that as the default, consumers of this new function won't be affected.

atoulme: Sure thing. What value?

andrzej-stencel, line 288: We should initialize the `ResponseHeaders` to prevent users of the API from panics when trying to write to an uninitialized map ([previous comment](https://github.com/open-telemetry/opentelemetry-collector/pull/9672#discussion_r1518358560)).

We should use configtls.NewDefaultServerConfig and configauth.NewDefaultAuthentication and NewDefaultCORSConfig ([previous comment](https://github.com/open-telemetry/opentelemetry-collector/pull/9672/files#r1583251060))

atoulme: nice! OK will add

atoulme: Let's pick a number: 30s

config/confighttp/confighttp.go
```diff
@@ -306,6 +306,51 @@ type ServerConfig struct {
...
+}
+
+// NewDefaultServerConfig returns ServerConfig type object with default values.
+// We encourage to use this function to create an object of ServerConfig.
+func NewDefaultServerConfig() ServerConfig {
+	tlsDefaultServerConfig := configtls.NewDefaultServerConfig()
+	return ServerConfig{
+		ResponseHeaders:   map[string]configopaque.String{},
+		TLSSetting:        &tlsDefaultServerConfig,
+		CORS:              &CORSConfig{},
```
andrzej-stencel, line 349: Still missing NewDefaultCORSConfig, but it doesn't exist yet, so probably can be added in a separate change.

andrzej-stencel (approved):

jpkrohling:
I discussed this in person with @atoulme yesterday, and I missed the fact that we don't currently expose the timeouts to our end-users. I'm OK with this PR as is, but would prefer to have that option exposed first, and a default being set here, so that we have fewer breaking changes.

atoulme:
If we add the timeout fields, they take over/conflict with the fields declared for one receiver in contrib. I think we could skip the test in contrib, merge this, fix in contrib. WDYT?

mx-psi:
What's the status of this one?

config/confighttp/confighttp.go
atoulme: https://github.com/open-telemetry/opentelemetry-collector/pull/10676 is open to follow up

atoulme:
> What's the status of this one?

It's ready to merge, if we are ok with breaking the build with contrib, or we can merge https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34181 to get a passing build.

jpkrohling:
I merged the contrib PR.

atoulme:
@jpkrohling thanks the tests now pass.

PR #10276, opened 2024-05-31
[extension/zpages] zpagesextension using confighttp.ServerConfig

### LINKED ISSUE #9368: [extension/zpages] zpages extension should use `confighttp.HTTPServerSettings`
The zpages extension declares a http.Server without much option besides setting its endpoint.

It would be good to instead use `confighttp.HTTPServerSettings` to set up the extension server.

atoulme:
#### Description
use confighttp.ServerConfig as part of the configuration of the zpagesextension

#### Link to tracking issue
Fixes #9368


.chloggen/zpagesextension_confighttp.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
codeboten, line 25: can you confirm this is only breaking for the api?

atoulme: This also changes the options for users to enter the endpoint and additional options, right?

atoulme: Maybe I can have 2 changelog entries:
-breaking the api
-enhancement for user

codeboten: Right but this would only be breaking for users that were configuring options in TCPAddr that aren't in ServerConfig which I think is just the dial options correct? If i was only using endpoint, nothing should change since the struct is squashed?

atoulme: yes same field. Users are untouched. API is changed.

codeboten (approved):

atoulme:
Ready to merge?

PR #10225, opened 2024-05-27
[exporter/debug] format log records as one-liners in `normal` verbosity

andrzej-stencel:
#### Description

Changes the behavior of `normal` verbosity of the Debug exporter for logs to display each log record in one line of text.

Note that if the body of the log record contains newlines, the output will be displayed in multiple lines.

This pull request is part of https://github.com/open-telemetry/opentelemetry-collector/issues/7806; it implements the change for logs. The changes for metrics and [traces](https://github.com/open-telemetry/opentelemetry-collector/pull/10280) will be proposed in separate pull requests.

The implementation in this pull request does not display any details on the resource or the scope of the logs. I would like to propose displaying the resource and the scope as separate lines in a separate pull request.

This change applies to the Debug exporter only. The behavior of the Logging exporter remains unchanged. To use this behavior, switch from the deprecated Logging exporter to Debug exporter.

#### Link to tracking issue

- https://github.com/open-telemetry/opentelemetry-collector/issues/7806

#### Testing

Added unit tests for the formatter.

#### Documentation

Described the formatting in the Debug exporter's README.

PR #10280, opened 2024-05-31
[exporter/debug] format spans as one-liners in `normal` verbosity

andrzej-stencel:
#### Description

This is an initial barebones implementation that only outputs the span's name, trace ID and span ID. Other useful fields like duration etc. can be added in follow-up enhancements.

This pull request is part of https://github.com/open-telemetry/opentelemetry-collector/issues/7806; it implements the change for traces. The changes for [logs](https://github.com/open-telemetry/opentelemetry-collector/pull/10225) and metrics will be proposed in separate pull requests.

This change applies to the Debug exporter only. The behavior of the Logging exporter remains unchanged. To use this behavior, switch from the deprecated Logging exporter to Debug exporter.

#### Link to tracking issue

- https://github.com/open-telemetry/opentelemetry-collector/issues/7806

#### Testing

Added unit tests for the formatter.

#### Documentation

Described the formatting in the Debug exporter's README.


.chloggen/debug-exporter-normal-verbosity-traces.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
mx-psi, line 4:
```suggestion
change_type: enhancement
```

Based on https://github.com/open-telemetry/opentelemetry-collector/blob/486b0645aac1eb03407139ae7840a9523f6ae8a2/exporter/debugexporter/README.md?plain=1#L125, this is not a breaking change. I am not comfortable setting the expectation that people should rely on the output being stable. If people want a machine readable output, they should use the file exporter

exporter/debugexporter/internal/normal/traces.go
```diff
@@ -0,0 +1,38 @@
...
+func (normalTracesMarshaler) MarshalTraces(md ptrace.Traces) ([]byte, error) {
+	var buffer bytes.Buffer
+	for i := 0; i < md.ResourceSpans().Len(); i++ {
+		resourceTraces := md.ResourceSpans().At(i)
+		for j := 0; j < resourceTraces.ScopeSpans().Len(); j++ {
+			scopeTraces := resourceTraces.ScopeSpans().At(j)
+			for k := 0; k < scopeTraces.Spans().Len(); k++ {
+				span := scopeTraces.Spans().At(k)
+				spanString := fmt.Sprintf("%s %s %s", span.Name(), span.TraceID().String(), span.SpanID().String())
+				buffer.WriteString(spanString)
```
mx-psi, line 32: Should we split the `Sprintf` into several `WriteString`s here?

.chloggen/debug-exporter-normal-verbosity-traces.yaml
andrzej-stencel: It actually says in the [README](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.102.0/exporter/debugexporter/README.md#warnings) that changes in output format are not considered breaking. :sweat_smile:

exporter/debugexporter/internal/normal/traces.go
andrzej-stencel: I don't see why not. What is the benefit? Is it performance?

mx-psi: Yeah, not sure how much we care about it but using both `fmt.Sprintf` and the bytes buffer seems weird

andrzej-stencel: Changed in https://github.com/open-telemetry/opentelemetry-collector/pull/10280/commits/9317cd4feebb8e1e6150b39e1af9a4b8bbbc4c81.

mx-psi (approved):
LGTM, needs `make goporto` and it should be good to go

mx-psi:
Looks like you need to do `make goporto`

exporter/debugexporter/internal/normal/common.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+// writeAttributes returns a slice of strings in the form "attrKey=attrValue"
+func writeAttributes(attributes pcommon.Map) (attributeStrings []string) {
+	attributes.Range(func(k string, v pcommon.Value) bool {
+		attribute := fmt.Sprintf("%s=%s", k, v.AsString())
```
andrzej-stencel, line 15: I think I could rewrite `writeAttributes` to take `buffer` and use `buffer.WriteString` instead of `fmt.Sprintf` if that helps? Perhaps as an follow-up improvement PR?

mx-psi: Yeah that makes sense, fine for me to do this on a follow up PR

PR #10281, opened 2024-05-31
[otelcol] Removed deprecated `ConfigProvider` field from settings

TylerHelmuth:
#### Description
Removes the deprecated `ConfigProvider` field from `CollectorSettings`.

### DIFF
--- a/otelcol/collector_test.go
+++ b/otelcol/collector_test.go
@@ -82,16 +82,17 @@ func (p mockCfgProvider) Watch() <-chan error {
 }
 
 func TestCollectorStateAfterConfigChange(t *testing.T) {
-	provider, err := NewConfigProvider(newDefaultConfigProviderSettings([]string{filepath.Join("testdata", "otelcol-nop.yaml")}))
-	require.NoError(t, err)
-
 	watcher := make(chan error, 1)
 	col, err := NewCollector(CollectorSettings{
-		BuildInfo:      component.NewDefaultBuildInfo(),
-		Factories:      nopFactories,
-		ConfigProvider: &mockCfgProvider{ConfigProvider: provider, watcher: watcher},
+		BuildInfo: component.NewDefaultBuildInfo(),
+		Factories: nopFactories,
+		// this will be overwritten, but we need something to get past validation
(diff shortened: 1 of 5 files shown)


otelcol/collector_test.go
```diff
@@ -82,16 +82,17 @@ func (p mockCfgProvider) Watch() <-chan error {
...
-		ConfigProvider: &mockCfgProvider{ConfigProvider: provider, watcher: watcher},
+		BuildInfo: component.NewDefaultBuildInfo(),
+		Factories: nopFactories,
+		// this will be overwritten, but we need something to get past validation
+		ConfigProviderSettings: newDefaultConfigProviderSettings([]string{filepath.Join("testdata", "otelcol-nop.yaml")}),
 	})
 	require.NoError(t, err)
+	provider, err := NewConfigProvider(newDefaultConfigProviderSettings([]string{filepath.Join("testdata", "otelcol-nop.yaml")}))
+	require.NoError(t, err)
+	col.configProvider = &mockCfgProvider{ConfigProvider: provider, watcher: watcher}
```
TylerHelmuth, line 95: @evan-bradley this tests highlights what I believe is a loss of functionality with the switch to factories.

Before it was possible to supply your own implementation of the `ConfigProvider` interface - that is no longer possible.  Users of `otelcol` _must_ use the `ConfigProvider` implementation defined in this package.

Is that expected?

evan-bradley: Yes, this is expected. Our implementation of the `otelcol.ConfigProvider` interface is really just tasked with running a `confmap.Resolver` and unmarshaling the config, there's not a whole lot of room for meaningful customization. If we want to provide extensibility, the goal is to do it through modules that users can provide to our `ConfigProvider` through the settings.

TylerHelmuth:
Depends on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33323

otelcol/collector_test.go
TylerHelmuth: Ok cool then this is ready to go

atoulme (approved):

codeboten (approved):

PR #10313, opened 2024-06-04
[exporterhelper] Fix shutdown hang on unstarted exporter

### LINKED ISSUE #10306: [exporterhelper] Batchsender hangs generated component shutdown test
**Describe the bug**

Any exporter using exporterhelper batchsender will cause its generated_component_test.go `-shutdown` test to hang as Shutdown will never return if it is not started.

Control will be stuck at https://github.com/open-telemetry/opentelemetry-collector/blob/release/v0.102.x/exporter/exporterhelper/batch_sender.go#L231

@dmitryax I wonder if the generated shutdown test needs to start the exporter, or batchsender needs to handle unstarted batcher.

**Steps to reproduce**

Use batchsender with `WithBatcher`. Run `generated_component_test.go`

Test passes

**What did you see instead?**

Test hangs

**What version did you use?**

v0.102.0

**What config did you use?**

**Environment**

**Additional context**

carsonip:
#### Description

Fix a bug where shutdown hangs if batch_sender exporter is not started. The bug causes generated component tests to fail as well.

#### Link to tracking issue
Fixes #10306

### DIFF
--- a/.chloggen/fix_batch_sender_shutdown.yaml
+++ b/.chloggen/fix_batch_sender_shutdown.yaml
@@ -1,0 +1,20 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterhelper
+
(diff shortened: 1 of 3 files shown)


dmitryax (approved):

dmitryax:
Thanks @carsonip!

PR #10316, opened 2024-06-04
[mdatagen] add support for optional internal metrics

codeboten:
This allows metrics to be initialized at a later time as is the case for the queue metrics in exporterhelper which are only initialized if a queue is configured.

### DIFF
--- a/cmd/mdatagen/internal/samplereceiver/metadata.yaml
+++ b/cmd/mdatagen/internal/samplereceiver/metadata.yaml
@@ -174,3 +174,12 @@ telemetry:
         async: true
         value_type: int
         monotonic: true
+    optional_metric:
+      enabled: true
+      description: This metric is optional and therefore not initialized in NewTelemetryBuilder.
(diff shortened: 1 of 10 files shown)


cmd/mdatagen/internal/samplereceiver/metadata.yaml
```diff
@@ -174,3 +174,12 @@ telemetry:
         async: true
         value_type: int
         monotonic: true
+    optional_metric:
```
jpkrohling, line 177: Can you use a different name here? It wasn't very apparent to me first whether "optional_metric" is a reserved word for this feature or the metric name. It became clear when I saw "optional: true" below, but perhaps you can use a real metric name instead?

codeboten: Thanks for the suggestion, i changed it to queue_length, which is sorta what this PR is all about 😆

jpkrohling (approved):

PR #10319, opened 2024-06-04
Change default otlp exporter GRPC load balancer to round robin

### LINKED ISSUE #10298: Change default otlp exporter GRPC load balancer to round robin

Using the pick_first load balancer will default to sending data to the same backend. This can potentially cause problems in resource allocation. Having one connection to the same IP address or backend would cause throttling. For scalable services, if this same backend is overloaded, it cannot accept anymore data sent and data can then be dropped instead of sending to another backend, especially if all are being used at max limit. Also, pick_first does no actual load balancing ([link](https://grpc.io/docs/guides/custom-load-balancing/#implementing-your-own-policy)), and instead just tries each address from the name resolver and connects to the one that works.

Updating the load balancer to use the round_robin policy will allow data to be sent to different backends, and therefore more evenly allocate resources. Round robin only picks ready connections, and so is better in a typical cloud-compute setup, where clients are sending data into a scalable service with multiple workers. This is because it allows for connections to alternate between backends, and therefore when resources on one backend are being used, the data is then moved to another available backend based on availability. In round_robin, users that want to send data to only one address can create connections against the address more than once to ensure that multiple connections can be made ([link](https://grpc.io/blog/grpc-on-http2/)). This would not let the connections be throttled. 
In this case, pick_first would resolve to that one address and can only accept data when that connection is available, therefore causing throttling. When having more than one connection to that one address, like what round_robin is able to do, throttling is less of an issue since data can be sent to multiple locations that all send to that one address.

The alternative is to leave pick_first the default and recommend users make a choice. This is not an adequate solution because users expect reliable delivery by default and round_robin is substantially more reliable for minimal additional cost. We have restricted our choices to round_robin and pick_first because these two are registered by default, other custom load balancers would have to be registered by the user. 

**Additional context**

I tested these two different load balancers in the Lightstep/SNCO dashboard. Envoy-edge, a proxy that load balances data, being sent to our service spaningest, which as its name implies, ingests OTLP spans using the arrow format.
In the first image, we see that when sending data from our service, envoy to spaningest arrow, at first, looks pretty even. This is currently using round-robin load balancing, as it distributes traffic in rotation, to different k8s pods, and therefore the resource allocation is even.
image
After the time the change to pick_first is made, we see that each pod changes how much data(spans) is sent drastically, as some pods get more spans than others.
image
Comparing the two, we see that the differences using round robin and pick first load balancers are pretty apparent. Since pick first sends to the first pod that is available, data is sent to that pod and all the other pods are left using no resources.
image

taniyourstruly:
#### Description
Updates the default `pick_first` load balancer to `round_robin`, which allows for better resource allocation and less chances of throttling data being sent to addresses. 

#### Link to tracking issue
Fixes #10298 (has full context of this PR)

#### Testing
Edited tests to allow round_robin load balancing.


jpkrohling (changes requested):
As much as I agree with having round_robin as the default, we try to stick with the default values from the underlying libraries.

jaronoff97:
@jpkrohling can you explain the reasoning behind that for this specific scenario? Per [Tani's issue](https://github.com/open-telemetry/opentelemetry-collector/issues/10298): pick_first is never the right choice:
> Also, pick_first does no actual load balancing ([link](https://grpc.io/docs/guides/custom-load-balancing/#implementing-your-own-policy)), and instead just tries each address from the name resolver and connects to the one that works.

gRPC can't change their default because of legacy clients they need to support / the edge cases are much larger. Given that the collector serves to send data to an arbitrary backend, and most destinations being sent to would benefit from more evenly balanced load, I'm not sure why we would stick with a choice that balances load poorly. I would argue that today we are not setting a reasonable default for users.

jpkrohling:
The base for the reasoning is that we try to keep our config with the value that would cause the least surprise to our users, and that's typically the default we have from the underlying library. Another example that I prefer a different default is for the Min TLS version, which is (IIRC) 1.2 from Go, but we could have it set to 1.3 by default.

That said, I'm open to changing the default value for this, deviating from the underlying library, if we have OKs from maintainers (cc @open-telemetry/collector-maintainers ).

codeboten:
> As much as I agree with having round_robin as the default, we try to stick with the default values from the underlying libraries.

I would be ok with changing the default and documenting the reasoning for it in the components affected by this change. I think it's desirable for end users that the collector provides defaults that make their experience better. As per the research done in https://github.com/open-telemetry/opentelemetry-collector/issues/10298, I think this qualifies.

jpkrohling (commented):
LGTM, but I wanted to settle which strategy we want to make the default: @tsloughter suggested random instead. Once we settle that, we are ready to change the default value.

codeboten (approved):
Thanks @taniyourstruly

jpkrohling (approved):
Thank you for this change, and for persisting in pursuing what you considered the right thing despite my initial reaction.

Issue #9428, opened 2024-01-30
Decide on naming rules for `Settings` structs


mx-psi:
Per https://github.com/open-telemetry/opentelemetry-collector/pull/9406#issuecomment-1912513732 we want to have a set of naming rules for structs ending in `Settings`. Current (as of 1ed45ec12569feb7863637a83b5ceefa70da44bf) structs that have this are the following:

<details>

<summary> Output for <code>rg 'type ([A-Z].*Settings|Settings) struct' -tgo --glob '!internal'</code> (click to expand)</summary>

```
otelcol/collector.go
56:type CollectorSettings struct {

otelcol/configprovider.go
74:type ConfigProviderSettings struct {

extension/extension.go
66:type CreateSettings struct {

receiver/scraperhelper/obsreport.go
43:type ObsReportSettings struct {

receiver/receiverhelper/obsreport.go
47:type ObsReportSettings struct {

receiver/scraperhelper/settings.go
23:type ScraperControllerSettings struct {

connector/connector.go
65:type CreateSettings struct {

confmap/resolver.go
31:type ResolverSettings struct {

receiver/receiver.go
44:type CreateSettings struct {

processor/processor.go
35:type CreateSettings struct {

config/configgrpc/configgrpc.go
51:type GRPCClientSettings struct {
120:type GRPCServerSettings struct {

processor/processorhelper/obsreport.go
59:type ObsReportSettings struct {

service/extensions/extensions.go
165:type Settings struct {

component/telemetry.go
20:type TelemetrySettings struct {

service/service.go
37:type Settings struct {

exporter/exporter.go
35:type CreateSettings struct {

service/telemetry/telemetry.go
37:type Settings struct {

exporter/exporterhelper/queue_sender.go
31:type QueueSettings struct {

exporter/exporterhelper/timeout_sender.go
13:type TimeoutSettings struct {

exporter/exporterhelper/obsexporter.go
46:type ObsReportSettings struct {
```

</details>

Questions that arise to me are:

- [x] Can we remove the `Create` prefix from `CreateSettings`?
- [x] Can we remove other prefixes (e.g. `GRPC` from structs within `configgrpc`, `Collector` from `otelcol`...)?

cc @dmitryax  to fill in other rules.

Adding this to configgrpc milestone per the question above.

TylerHelmuth:
[...]

PR #10320, opened 2024-06-04
deprecate `connector.CreateSettings` in favour of `component.Settings`

codeboten:
This deprecates `connector.CreateSettings` and creates a `Settings` struct in component since the settings struct is common across all components.

Part of #9428


mx-psi (commented):
This would prevent us from sending component-specific settings to a connector vs a receiver. Do we care?

codeboten:
@mx-psi that's a good question i dont have a good sense of whether this is something we'll want in the future, for now all the structs are the same and i thought it might be better to have them all use the same `component.Settings`

codeboten:
Closing as discussed in 5-June-2024 SIG call, the preference is to keep these settings structs separate for now.

PR #10315, opened 2024-06-04
[exporterhelper] Fix potential deadlock in the batch sender

dmitryax:
Concurrent handling of the flush timeouts can run into a deadlock when a batch is simultaneously sent by reaching the minimum size and flush timeout. The deadlock can happen on the following lines:
- https://github.com/open-telemetry/opentelemetry-collector/blob/115bc8e28e009ca93565dc4deb4cf6608fa63622/exporter/exporterhelper/batch_sender.go#L131
- https://github.com/open-telemetry/opentelemetry-collector/blob/115bc8e28e009ca93565dc4deb4cf6608fa63622/exporter/exporterhelper/batch_sender.go#L87

PR #10337, opened 2024-06-05
[exporterhelper] Fix small batch due to scheduling in batch sender

### LINKED ISSUE #9952: [exporterhelper] Batch sender produces inconsistent and suboptimal batch sizes

carsonip:
#### Description

- Correctly keep track of bs.activeRequests, which denotes the number of send waiting for next sender in chain to return. This is already done before this PR, but in a way that is vulnerable to unfavorable goroutine scheduling
- Decrease bs.activeRequests by the number of requests blocked by an activeBatch at once, so that it workarounds the "bug" mentioned in (1).

#### Link to tracking issue
Fixes #9952

#### Testing
Added unit test


carsonip:
Blocked by #10315, otherwise test will be flaky.

exporter/exporterhelper/batch_sender.go
```diff
@@ -149,27 +155,39 @@ func (bs *batchSender) send(ctx context.Context, req Request) error {
...
-	bs.activeRequests.Add(1)
-	defer bs.activeRequests.Add(^uint64(0))
 
 	reqs, err := bs.mergeSplitFunc(ctx, bs.cfg.MaxSizeConfig, bs.activeBatch.request, req)
 	if err != nil || len(reqs) == 0 {
 		bs.mu.Unlock()
 		return err
 	}
+
+	bs.activeRequests.Add(1)
```
dmitryax, line 165: I think this can be pushed under the following condition and we won't need the leftover

exporter/exporterhelper/batch_sender.go
```diff
@@ -149,27 +155,39 @@ func (bs *batchSender) send(ctx context.Context, req Request) error {
...
 	if len(reqs) == 1 || bs.activeBatch.request != nil {
 		bs.updateActiveBatch(ctx, reqs[0])
-		batch := bs.activeBatch
-		if bs.isActiveBatchReady() || len(reqs) > 1 {
-			bs.exportActiveBatch()
+		b := bs.activeBatch
+		b.requests++
+		leftover := len(reqs) > 1
+		if bs.isActiveBatchReady() || leftover {
+			bs.exportActiveBatch(func(b *batch) {
```
dmitryax, line 172: Then the callback will always be the same `bs.activeRequests.Add(-b.requests)` for both cases and can be moved to the exportActiveBatch goroutine

exporter/exporterhelper/batch_sender.go
carsonip: It cannot be pushed down, because we still need the correct counts for `bs.activeRequests.Load() >= bs.concurrencyLimit`. However, I've refactored it to make it cleaner.

exporter/exporterhelper/batch_sender.go
carsonip: Refactored to remove the ugly callback in exportActiveBatch

exporter/exporterhelper/batch_sender.go
carsonip: My understanding of the concurrency control in the batch sender is to work with queue sender num_consumers and avoid meaningless waits. This means that if num_consumers=10 in queue sender, there will be at most 10 concurrent `send` calls issued to batch sender.

> If len(reqs) > 1 && bs.activeBatch.request == nil, we don't need to change requestsBlocked and flip bs.activeRequests back and forth.

I agree that requestsBlocked does not need to be changed as it bypasses activeBatch.

However, `bs.activeRequests` still needs to be flipped back and forth even in the case of `len(reqs) > 1 && bs.activeBatch.request == nil` because it blocks 1 consumer goroutine from queue sender.

Let's imagine a scenario where queue sender num_consumers=2.
- The first consumer C1 calls batch sender `send`, and hits the `len(reqs) > 1 && bs.activeBatch.request == nil` branch, and sends the requests (let's call them "R1") right away to the backend without touching activeBatch.
- If `bs.activeRequests` is NOT incremented while R1 are being sent, and a new consumer C2 calls batch sender `send` with `len(reqs) == 1`, when C2 hits `bs.isActiveBatchReady()`, the concurrency check `bs.activeRequests.Load() >= bs.concurrencyLimit` will be `1 >= 2` (where the 1 is from C2, because C1 is not accounted), then C2 will be subject to a meaningless wait for C1 to finish sending R1.
- If `bs.activeRequests` IS incremented while R1 are being sent, `bs.isActiveBatchReady()` will be true because of the concurrency check, and C2 will be sent immediately and will avoid the meaningless wait.

carsonip: The goal of this PR is to
1. correctly keep track of `bs.activeRequests`, which denotes the number of `send` waiting for next sender in chain to return. This is already done before this PR, but in a way that is vulnerable to unfavorable goroutine scheduling
2. decrease `bs.activeRequests` by the number of requests blocked by an activeBatch at once, so that it workarounds the "bug" mentioned in (1).

> If len(reqs) > 1 && bs.activeBatch.request == nil, we don't need to ... flip bs.activeRequests back and forth

If we do this, `bs.activeRequests` will be incorrect. It will be an underestimate of actual ongoing `send`.

dmitryax: Ok, I get it now. Thanks!

exporter/exporterhelper/batch_sender.go
```diff
@@ -149,18 +154,24 @@ func (bs *batchSender) send(ctx context.Context, req Request) error {
...
 		bs.updateActiveBatch(ctx, reqs[0])
 		batch := bs.activeBatch
+		batch.requestsBlocked++
 		if bs.isActiveBatchReady() || len(reqs) > 1 {
+			if len(reqs) > 1 {
+				// there will be leftover requests and this request will be active until the function returns.
+				batch.requestsBlocked--
+				defer bs.activeRequests.Add(-1)
+			}
 			bs.exportActiveBatch()
```
dmitryax, line 175: I think the logic is correct here. It can probably be simplified a bit more. Let me know WDYT.

```suggestion
	if len(reqs) == 1 {
		batch.requestsBlocked++
	} else {
		// if there was a split, we want to make sure that bs.activeRequests is released once all of the parts are sent instead of using batch.requestsBlocked
		defer bs.activeRequests.Add(-1)
	}
	if len(reqs) == 1 || bs.activeBatch.request != nil {
		bs.updateActiveBatch(ctx, reqs[0])
		batch := bs.activeBatch
		if bs.isActiveBatchReady() || len(reqs) > 1 {
			bs.exportActiveBatch()
```

exporter/exporterhelper/batch_sender.go
```diff
@@ -170,6 +181,7 @@ func (bs *batchSender) sendMergeSplitBatch(ctx context.Context, req Request) err
 		}
 		reqs = reqs[1:]
 	} else {
+		defer bs.activeRequests.Add(-1)
```
dmitryax, line 184: Then this can be removed

carsonip: I don't think so. The increment on line 164 will be missing a corresponding decrement if we remove this line. This applies to case `len(reqs) > 1 && bs.activeBatch.request == nil`, i.e. split requests that don't touch activeBatch. The correct handling should be incrementing activeRequests (line 164) before sending the split requests, and decrement activeRequests after it is done, which is what this line is responsible for.

carsonip: Ignore my previous comment. I looked at the indentation wrong.

dmitryax (approved):

PR #10338, opened 2024-06-05
[connector] deprecate CreateSettings -> Settings

codeboten:
This deprecates CreateSettings in favour of Settings. NewNopCreateSettings is also being deprecated in favour of NewNopSettings


TylerHelmuth (approved):

jmacd:
This sort of refactoring creates a ton of work for anyone with in-flight changes!!! Really not convinced the benefit outweighs the cost.

PR #10340, opened 2024-06-05
[component] Remove deprecated `component.UnmarshalConfig`

### LINKED ISSUE #7102: Remove component.UnmarshalConfig

atoulme:
#### Description
Remove deprecated `component.UnmarshalConfig`

#### Link to tracking issue
Fixes #7102


atoulme:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33404 is open to fix contrib.

TylerHelmuth (approved):

TylerHelmuth:
@atoulme looks like there are a couple more places in Contrib that need updating

atoulme:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33407 to follow up, sorry.

codeboten (approved):

PR #10348, opened 2024-06-06
[filter] unexport CombinedFilter

atoulme:
#### Description
Unexport CombinedFilter, it should not be exported as it implements Filter.

### DIFF
--- a/filter/config.go
+++ b/filter/config.go
@@ -37,6 +37,9 @@ type combinedFilter struct {
 	regexes []*regexp.Regexp
 }
 
+// Deprecated: [v0.103.0] This type will be removed in the future.
+type CombinedFilter combinedFilter
+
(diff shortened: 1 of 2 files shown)


djaglowski (approved):

filter/config.go
```diff
@@ -32,14 +32,14 @@ func (c Config) Validate() error {
 	return nil
 }
 
-type CombinedFilter struct {
+type combinedFilter struct {
```
codeboten, line 35: Can this be marked as deprecated for one release and then made private? I know it's only used in mdatagen internally, but we can't be sure others out there aren't using this package that's 2 months old at this point

atoulme: I didn't pick up it was used anywhere in contrib. I'll follow your direction, sure.

codeboten (approved):
Thanks @atoulme

PR #10352, opened 2024-06-06
[receiver/otlp] Promote `component.UseLocalHostAsDefaultHost` to beta

mx-psi:
#### Description

Promotes `component.UseLocalHostAsDefaultHost` feature gate to beta.

#### Link to tracking issue

Updates #8510


yurishkuro:
+1

atoulme (approved):
+1

.chloggen/mx-psi_enable-component-localhost-fgh.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: otlpreceiver
```
TylerHelmuth, line 7: This affects more than the otlp receiver right? Any receivers using `ToServer` is affected

mx-psi: I'll add an extra API release note on here, and list the components explicitly on the contrib changelog. Does that make sense?

otelcol/testdata/otel-log-to-file.yaml
```diff
@@ -14,7 +14,7 @@ exporters:
 service:
   telemetry:
     logs:
-      level: warn
+      level: info
```
mx-psi, line 17: This is to make the Windows service test pass. Before this change, we had the warning from confighttp and configgrpc, but after it there are no logs at warn level with this config, which makes the assertion that the logs were stored in a file fail

mx-psi:
Please also take a look at the contrib counterpart, open-telemetry/opentelemetry-collector-contrib/pull/33658, cc @TylerHelmuth

.chloggen/mx-psi_enable-component-localhost-fgh.yaml
mx-psi: Actually, I don't think an API release note is needed here, this does not change anything in confighttp/configrpc, only in the specific components

songy23 (approved):

djaglowski (approved):

TylerHelmuth (approved):

evan-bradley (approved):

Issue #10290, opened 2024-06-03
[otelcol] Determine if `NewCommand` should supply any default providers/converters


TylerHelmuth:
There is currently a TODO in `otelcol` [here](https://github.com/open-telemetry/opentelemetry-collector/blob/ed767dc21968b2b2a2ba2ac95f40fe510de9de79/otelcol/command.go#L56) that states:

```go
// Puts command line flags from flags into the CollectorSettings, to be used during config resolution.
func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet) error {
	resolverSet := &set.ConfigProviderSettings.ResolverSettings
	configFlags := getConfigFlag(flags)

	if len(configFlags) > 0 {
		resolverSet.URIs = configFlags
	}
	if len(resolverSet.URIs) == 0 {
		return errors.New("at least one config flag must be provided")
	}
	// Provide a default set of providers and converters if none have been specified.
	// TODO: Remove this after CollectorSettings.ConfigProvider is removed and instead
	// do it in the builder.
	if len(resolverSet.ProviderFactories) == 0 && len(resolverSet.ConverterFactories) == 0 {
		set.ConfigProviderSettings = newDefaultConfigProviderSettings(resolverSet.URIs)
	}
	return nil
}
```

We have removed `ConfigProvider` via https://github.com/open-telemetry/opentelemetry-collector/pull/10281 so we could start working on this todo, except that it would be a breaking change to stop providing default providers/converters.

We have an explicit unit test via https://github.com/open-telemetry/opentelemetry-collector/pull/9302 that tests that we supply default providers/converters for jaeger all-in-one.

Should we move forward with the breaking change (which we know would break jaeger)? What problem are we solving by not providing default providers/converters?

TylerHelmuth:
Setting of default converters is part of the conversation of https://github.com/open-telemetry/opentelemetry-collector/pull/10259, where we are deprecating `expandconverter` in favor of a `DefaultScheme: "env"`.

If we continue to supply defaults, instead of setting the `expandconverter` the default would be to set `DefaultScheme: "env"`.

evan-bradley:
I'd like to remove all defaults and have these supplied by distro authors through the APIs or through the builder.

TylerHelmuth:
@yurishkuro how is Jaeger all-in-one depending on this feature?  Is it using the otelcol API or a community distribution like Core or Contrib?

TylerHelmuth:
[...]

PR #10357, opened 2024-06-06
[otelcol] Don't provide any defaults via `NewCommand`

TylerHelmuth:
#### Description

Updates `otelcol.NewCommand` to error if `otelcol.CollectorSettings` does not have a least 1 provider or converter set.  The previous behavior was to set default values.

Also removed all references to confmap providers or converters. 

#### Link to tracking issue
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/10290

#### Testing
Updated unit tests

### DIFF
--- a/otelcol/command_test.go
+++ b/otelcol/command_test.go
@@ -9,11 +9,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/confmap"
-	"go.opentelemetry.io/collector/confmap/converter/expandconverter"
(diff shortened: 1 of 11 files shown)


otelcol/command_test.go
```diff
@@ -28,33 +24,12 @@ func TestNewCommandNoConfigURI(t *testing.T) {
 	require.Error(t, cmd.Execute())
 }
 
-// This test emulates usage of Collector in Jaeger all-in-one, which
-// allows running the binary with no explicit configuration.
-func TestNewCommandProgrammaticallyPassedConfig(t *testing.T) {
-	cmd := NewCommand(CollectorSettings{Factories: nopFactories})
```
yurishkuro, line 34: can we keep this test, with whatever changes needed to CollectorSettings?

PR #10359, opened 2024-06-06
[otelcol] Add new command to start erroring on empty providers

TylerHelmuth:
#### Description

Add a featuregate to cause an error when no providers or converters are set in `CollectorSettings` that are passed to `NewCommand`.

Also deprecate the functions in `otelcoltest` with versions that require `ConfigProviderSettings` so that all providers and converters deps can be removed when the feature gate is stable.

#### Link to tracking issue

Works towards https://github.com/open-telemetry/opentelemetry-collector/issues/10290.

End state goal here: https://github.com/open-telemetry/opentelemetry-collector/pull/10357

#### Testing
Added tests


mx-psi (commented):
Why do we need a feature gate for this? Isn't it enough to have the new symbol and deprecate the old one? End-users of the binary are not going to be affected by this, right?

TylerHelmuth:
> Why do we need a feature gate for this? Isn't it enough to have the new symbol and deprecate the old one? End-users of the binary are not going to be affected by this, right?

The new symbols are for `otelcoltest` only.  I could move those into a separate PR if you'd like, but'll we'll need those deprecations to work towards removing provider imports.

The feature gate is only for `NewCommand`'s behavior. The feature gate changes `NewCommand` to error instead of providing defaults, which will not affect end users of the ocb.

An alternative to the feature gate is to make a new entry point, like `NewCommandWithSettings` and deprecate `NewCommand`.  The new function would error if there are not providers/converters instead of setting defaults.  Then we could deprecate `NewCommandWithSettings` and rename it back to `NewCommand`.

mx-psi (approved):
This looks good to me. I would be okay with making the breaking change in one for the `otelcoltest` symbols (pressumably those are only used in tests so the breakage would be smaller), but this is also fine

otelcol/command.go
```diff
@@ -16,14 +16,29 @@ import (
...
+// Deprecated: [v0.103.0] use NewCommandMustSetProvider instead
 func NewCommand(set CollectorSettings) *cobra.Command {
+	return commandHelper(set, false)
+}
+
+// NewCommandMustSetProvider constructs a new cobra.Command using the given CollectorSettings.
+// Any URIs specified in CollectorSettings.ConfigProviderSettings.ResolverSettings.URIs
+// are considered defaults and will be overwritten by config flags passed as
+// command-line arguments to the executable.
+// At least one provider or converter must be supplied via CollectorSettings.ConfigProviderSettings.ResolverSettings.
```
evan-bradley, line 29:
```suggestion
// At least one Provider must be supplied via CollectorSettings.ConfigProviderSettings.ResolverSettings.ProviderFactories.
```
Supplying 0 Providers isn't valid, so I think we should call that out here.

otelcol/command.go
```diff
@@ -52,10 +67,11 @@ func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet) error
 	if len(resolverSet.URIs) == 0 {
 		return errors.New("at least one config flag must be provided")
 	}
-	// Provide a default set of providers and converters if none have been specified.
-	// TODO: Remove this after CollectorSettings.ConfigProvider is removed and instead
-	// do it in the builder.
+
 	if len(resolverSet.ProviderFactories) == 0 && len(resolverSet.ConverterFactories) == 0 {
+		if enforceProviders {
+			return errors.New("at least one provider or converter must be provided")
```
evan-bradley, line 73:
```suggestion
			return errors.New("at least one Provider must be supplied")
```
I know this wording is weird since we're checking for both providers and converters, but we should avoid implying that only supplying a Converter is a valid state. If a user providers only a Converter, they'll get a different error (from the Resolver), but it will at least say the same thing.

otelcol/command.go
```diff
@@ -52,10 +67,11 @@ func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet) error
 	if len(resolverSet.URIs) == 0 {
 		return errors.New("at least one config flag must be provided")
 	}
-	// Provide a default set of providers and converters if none have been specified.
-	// TODO: Remove this after CollectorSettings.ConfigProvider is removed and instead
```
evan-bradley, line 56: It's okay if we know this will be tracked, but I think we should probably leave this todo until we fully remove this check.

.chloggen/otelcol-default-providers-featuregate-2.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
evan-bradley, line 4:
```suggestion
change_type: deprecation
```

otelcol/otelcoltest/config.go
```diff
@@ -30,15 +41,33 @@ func LoadConfig(fileName string, factories otelcol.Factories) (*otelcol.Config,
 			ConverterFactories: []confmap.ConverterFactory{expandconverter.NewFactory()},
 		},
 	})
+}
+
+// LoadConfigWithSettingsAndValidate loads a config from the provider settings, and validates the configuration.
+func LoadConfigWithSettingsAndValidate(factories otelcol.Factories, set otelcol.ConfigProviderSettings) (*otelcol.Config, error) {
```
codeboten, line 47: Nit: usually the "WithSettings" would be a suffix to the method. This makes swapping the function in code a bit easier (if you're not using just straight up search and replace)

codeboten (approved):
Thanks @TylerHelmuth, please address @evan-bradley's comments and we can get this merged

otelcol/otelcoltest/config.go
TylerHelmuth: Updated

otelcol/command.go
TylerHelmuth: Added back

otelcol/command.go
```diff
@@ -56,6 +71,9 @@ func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet) error
 	// TODO: Remove this after CollectorSettings.ConfigProvider is removed and instead
 	// do it in the builder.
 	if len(resolverSet.ProviderFactories) == 0 && len(resolverSet.ConverterFactories) == 0 {
+		if enforceProviders {
+			return errors.New("at least one Provider must be supplied")
+		}
 		set.ConfigProviderSettings = newDefaultConfigProviderSettings(resolverSet.URIs)
 	}
```
evan-bradley, line 78:
```suggestion
	if !enforceProviders && len(resolverSet.ProviderFactories) == 0 && len(resolverSet.ConverterFactories) == 0 {
		set.ConfigProviderSettings = newDefaultConfigProviderSettings(resolverSet.URIs)
	}
```
I just realized that we can essentially make this if statement into a no-op and let the downstream packages handle it. Not a huge deal since hitting this error means that a distro was built incorrectly and the error won't be of much help to end users.

evan-bradley (approved):

otelcol/command.go
TylerHelmuth: I'd rather return the error, since we know here something is not right.  I see it as `otelcol.NewCommand` enforcing behavior it wants instead of depending on other packages, which could technically change, enforcing it.

evan-bradley: That's okay with me, it's functionally equivalent either way and we'll remove this soon. Let's leave it as-is.

PR #10323, opened 2024-06-05
[configgrpc] Use own compressors for zstd

jpkrohling:
Uses our own version of the zstd compressor for gRPC servers. The code for it is based on the gzip compressor that comes built-in with gRPC. 

Benchmarks before this PR:
```
Running tool: /usr/bin/go test -benchmem -run=^$ -bench ^BenchmarkCompressors$ go.opentelemetry.io/collector/config/configgrpc

sm_log_requestgoos: linux
goarch: amd64
pkg: go.opentelemetry.io/collector/config/configgrpc
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
BenchmarkCompressors/sm_log_request/raw_bytes_160/compressed_bytes_162/compressor_gzip-16         	   71594	     19066 ns/op	     615 B/op	       4 allocs/op
sm_log_requestBenchmarkCompressors/sm_log_request/raw_bytes_160/compressed_bytes_159/compressor_zstd-16         	  151503	      8544 ns/op	     640 B/op	       6 allocs/op
sm_log_requestBenchmarkCompressors/sm_log_request/raw_bytes_160/compressed_bytes_178/compressor_snappy-16       	 3632570	       303.8 ns/op	     304 B/op	       3 allocs/op
md_log_requestBenchmarkCompressors/md_log_request/raw_bytes_242/compressed_bytes_219/compressor_gzip-16         	   68114	     16938 ns/op	     748 B/op	       4 allocs/op
md_log_requestBenchmarkCompressors/md_log_request/raw_bytes_242/compressed_bytes_209/compressor_zstd-16         	  138091	      8047 ns/op	     896 B/op	       6 allocs/op
md_log_requestBenchmarkCompressors/md_log_request/raw_bytes_242/compressed_bytes_260/compressor_snappy-16       	 3081198	       402.5 ns/op	     400 B/op	       3 allocs/op
lg_log_requestBenchmarkCompressors/lg_log_request/raw_bytes_4850/compressed_bytes_253/compressor_gzip-16        	   43414	     27174 ns/op	     386 B/op	       3 allocs/op
lg_log_requestBenchmarkCompressors/lg_log_request/raw_bytes_4850/compressed_bytes_216/compressor_zstd-16        	  117534	      9903 ns/op	   10112 B/op	       6 allocs/op
[...]

PR #10360, opened 2024-06-06
[configgrpc] create zstd.Decoder's using WithDecoderConcurrency(1)

mostynb:
#### Description

The zstd.Decoder documentation states:
> It is important to use the "Close" function when you no longer need
> the Reader to stop running goroutines, when running with default
> settings. Goroutines will exit once an error has been returned,
> including io.EOF at the end of a stream.
>
> Streams are decoded concurrently in 4 asynchronous stages to give
> the best possible throughput. However, if you prefer synchronous
> decompression, use WithDecoderConcurrency(1) which will decompress
> data as it is being requested only.

Since we store zstd.Decoder's in a sync.Pool, we should probably use WithDecoderConcurrency(1), so that we don't need to call Close() when they are dropped from the pool.

#### Link to tracking issue

Relates to #10323

#### Testing

TODO

### DIFF
--- a/config/configgrpc/internal/zstd.go
+++ b/config/configgrpc/internal/zstd.go
@@ -52,7 +52,7 @@ type reader struct {
 func (c *compressor) Decompress(r io.Reader) (io.Reader, error) {
 	z, inPool := c.poolDecompressor.Get().(*reader)
 	if !inPool {
-		newZ, err := zstd.NewReader(r)
+		newZ, err := zstd.NewReader(r, zstd.WithDecoderConcurrency(1))
 		if err != nil {
 			return nil, err
 		}


songy23 (changes requested):
Looks like https://github.com/open-telemetry/opentelemetry-collector/pull/10394 supersedes this PR since the internal fork of zstd is removed.

cc @jpkrohling

jpkrohling:
Despite this being closed as not merged, I wanted to thank you for your time to open this PR here, @mostynb, and we appreciate the work you put also on the library we are using for gRPC payload compression.

mostynb:
No worries- you're welcome.

PR #10365, opened 2024-06-07
[configtelemetry] add guidelines for each level of config telemetry

### LINKED ISSUE #10286: Describe configtelemetry levels
Prior to this package's stabilization, I would love it if there were more guidance on what each of the levels are for. Right now the documentation is very vague and I think the differences between normal & detailed are not specific enough. This is almost a flavor of #7543 where that metric could have been enabled only in the "detailed" configuration, however, reserving "detailed" that way wasn't an option because we put many metrics behind "detailed" that I would argue belong more in the "normal" or "basic" levels ([ex](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/metrics.go#L65-L67) of a detailed metric that IMO should be a "normal" metric).

_Originally posted by @jaronoff97 in https://github.com/open-telemetry/opentelemetry-collector/issues/10277#issuecomment-2142496499_

atoulme:
#### Description
Add to the package comment a set of guidelines for configtelemetry levels.

#### Link to tracking issue
Fixes #10286


atoulme:
@jaronoff97 ptal

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+//
+// No telemetry data should be collected.
+//
+// 2. configtelemetry.Basic
+//
+// Signals associated with this level cover the essential coverage of the component telemetry.
+//
+// This is the default level recommended when running the collector.
+//
+// The signals associated with this level must have low cardinality.
```
jaronoff97, line 18: Is there a maximum cardinality #  we should recommend here?

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+//
+// The signals associated with this level must represent a small data volume:
+//   - No more than 5 metrics reported.
+//   - At most 1 span actively recording simultaneously, covering the critical path.
+//   - At most 5 log records produced every 30s.
```
jaronoff97, line 23: How were these determined?

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+//
+// Not all signals defined in the component telemetry are active.
+//
+// 3. configtelemetry.Normal
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
+//
+// The signals associated with this level must control cardinality.
```
jaronoff97, line 31: what does control mean in this context?

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+//
+// The signals associated with this level must represent a controlled data volume:
+//   - No more than 20 metrics reported.
```
jaronoff97, line 34: is this 20 individual meters or 20 metric points?

jaronoff97 (commented):
A few thoughts, i like where this is heading.

jaronoff97:
I do think in the future it would be neat if we had wrapper methods around otel APIs for recording telemetry for each level so authors don't need to keep track of these things.

config/configtelemetry/doc.go
atoulme: I made them up!

config/configtelemetry/doc.go
atoulme: 20 different metric time series, as in 20 combination of 20 metric names and dimension names values.

config/configtelemetry/doc.go
atoulme: There is a rationale for the cardinality and the author of the signal can explain the cardinality based on factors.
For example:
* The cardinality is linearly increasing with the number of Kafka topics processed
* The cardinality is correlated to the number of active clients

While this may incur a high cardinality, users can then filter based on it in some way or accumulate the signal by relying on this cardinality.

config/configtelemetry/doc.go
atoulme: sure?

config/configtelemetry/doc.go
jaronoff97: ah I see, thanks for adding that context. I think it would be useful to explain that a bit more in this comment – i.e. "it is acceptable at this level for cardinality to scale linearly with the monitored resources"

config/configtelemetry/doc.go
jaronoff97: Doesn't that mean a single metric that records the http status code would not fall into the normal level because the cardinality is >70?

jaronoff97: [http.server.request.duration](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration)

atoulme: Checking, I think indeed that metric would be too verbose. You can create quite a bit of cardinality if you are running a server and any URI or method adds to cardinality. But say for the httpcheckreceiver, that might be fine because you use a http client to call just one endpoint, so cardinality is fixed.

mx-psi:
@jaronoff97 Can you take another look?

config/configtelemetry/doc.go
jaronoff97: IMO given the metric is labeled "required" this should be covered by the normal telemetry settings.

jaronoff97:
@mx-psi added another comment, i think other collector maintainers/developers should also look at this as well given it would potentially change what the collector is recommended to emit.

config/configtelemetry/doc.go
atoulme: ok, so let's relax this. Can you offer an amendment? Maybe we can set a max cardinality of 500 for now.
```suggestion
//   - A max cardinality (total possible combinations of dimension values) of 500.
```

config/configtelemetry/doc.go
atoulme:
```suggestion
// The signals associated with this level must control cardinality. It is acceptable at this level for cardinality to scale linearly with the monitored resources.
```

config/configtelemetry/doc.go
atoulme:
```suggestion
// The signals associated with this level must show low cardinality, the number of combinations of dimension values. The maximum cardinality allowed is set to 50.
```

config/configtelemetry/doc.go
jaronoff97: I think that's alright, I'd be interested in what enforcement looks like in the future given there's support for cardinality limits on instruments. https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/metric/internal/x/README.md#L11

atoulme: you probably get to write that part ;)

mwear (commented):
I'm curious how configtelemetry level is meant to be used in the collector. Is the idea that it will be used to conditionally record telemetry throughout the codebase (are there any examples)? Have you considered using existing SDK mechanisms to limit telemetry such as samplers and views?

atoulme:
> I'm curious how configtelemetry level is meant to be used in the collector. Is the idea that it will be used to conditionally record telemetry throughout the codebase (are there any examples)? Have you considered using existing SDK mechanisms to limit telemetry such as samplers and views?

you can look at obsreport to give you an idea of some of the bundled telemetry. We use the Go SDK for metrics and traces.

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+//
+// 1. configtelemetry.None
+//
+// No telemetry data should be collected.
+//
+// 2. configtelemetry.Basic
+//
+// Signals associated with this level cover the essential coverage of the component telemetry.
+//
+// This is the default level recommended when running the collector.
```
mx-psi, line 16:
```suggestion
// This is the default level recommended when running the Collector.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+//
+// The signals associated with this level must show low cardinality, the number of combinations of dimension values. The maximum cardinality allowed is set to 50.
```
mx-psi, line 18: How are we going to enforce all the numbers in this doc?

atoulme: The issue says to offer guidance, not enforcement.

Before we go there, is this guidance any good? WDYT?

mx-psi: We don't need to implement enforcement on this PR or have a detailed plan of how to do it, but we need to know that these are reasonably easy to enforce in practice, otherwise the guidance is not actionable

atoulme: It looks reasonably easy to enforce in practice to me. The only tricky thing I guess is to calculate the cardinality of the metric, by computing the total set of combinations that can be formed based on the dimensions of the metric. Is that what needs to be detailed in there?

mx-psi: We should be able to measure the numbers easily, yes, my concern is more what are we going to do if we start deviating from these numbers. Do we get to a point where we just can't add more metrics at a certain level (since removing metrics is presumably a breaking change)? If a library that offers built-in instrumentation and starts producing many more metrics, what are we going to do?

To some extent this is a problem that we are going to have to solve regardless of what we put in here, but I think putting stringent requirements like these without having a good idea of how we are going we make our lives unnecessarily harder. I am tempted to just remove the numbers or make the numbers a rough guidance/approximation instead of a must-follow rule.

atoulme: > We should be able to measure the numbers easily, yes, my concern is more what are we going to do if we start deviating from these numbers. Do we get to a point where we just can't add more metrics at a certain level (since removing metrics is presumably a breaking change)? If a library that offers built-in instrumentation and starts producing many more metrics, what are we going to do?

We will need to reject the library in any case.

If we use say the Go SDK, and all of a sudden, it emits by default all sorts of metrics related to Go GC, this is in itself a change that can be considered a breaking change because it may break folks backends with high cardinality metrics.
> 
> To some extent this is a problem that we are going to have to solve regardless of what we put in here, but I think putting stringent requirements like these without having a good idea of how we are going we make our lives unnecessarily harder. I am tempted to just remove the numbers or make the numbers a rough guidance/approximation instead of a must-follow rule.

Yes, this whole doc is a guidance, I didn't mean for those numbers to feel such a hard rule. The numbers are what Pavol wanted to see, I am resistant also to provide absolute guidelines but offering numbers seemed to help enshrine some of the thinking.

mx-psi: > Yes, this whole doc is a guidance, I didn't mean for those numbers to feel such a hard rule

Can we made this more explicit in the text? (e.g. saying the limits are 'best effort' or adding a 'roughly' here or there to make it clear that this is a guideline but not a hard rule).

> offering numbers seemed to help enshrine some of the thinking.

Yup, I agree it makes the distinction between the levels clearer

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,46 @@
...
+// Signals associated with this level cover the essential coverage of the component telemetry.
+//
+// This is the default level recommended when running the Collector.
+//
+// The signals associated with this level must show low cardinality, the number of combinations of dimension values. The maximum cardinality allowed is set to 50.
+//
+// The signals associated with this level must represent a small data volume:
+//   - No more than 5 metrics reported.
+//   - At most 1 span actively recording simultaneously, covering the critical path.
+//   - At most 5 log records produced every 30s.
```
djaglowski, line 23: I question whether logs should be included in this document at all. If they are, I would think guidance is given in terms of severity level, rather than telemetry level.

Log severity and sampling are well established ways to manage log volume and we already have configuration for them in the collector. I think that trying to bring in the concept of telemetry level in any way is going to be confusing at best.

config/configtelemetry/doc.go
djaglowski: Could we offer the numbers only as an example? e.g. "A typical component might have the following signals at this level."

IMO we're better off focusing the language on helping component developers make decisions about whether a metric or span should be included at a certain level, similar to how we might guide developers how to choose an appropriate log severity. (Also see my other comment about logs not fitting into this guidance.)

config/configtelemetry/doc.go
atoulme: I will remove them

config/configtelemetry/doc.go
atoulme: I have revised the text to take into account your considerations, please review.

atoulme:
> focusing the language on helping component developers make decisions about whether a metric or span should be included at a certain level
That's a good insight and I have added a bit on the intent of the doc at the top. Please help make this better.

mx-psi (approved):

djaglowski (approved):
Thanks @atoulme. This looks good to me.

atoulme:
@dmitryax reviewed with me and had feedback, we should wait before merging this. He explained that the "Basic" telemetry level is meant also to only be for core level libraries, which makes sense to me. It was discussed on an issue.

dmitryax:
> @dmitryax reviewed with me and had feedback, we should wait before merging this. He explained that the "Basic" telemetry level is meant also to only be for core level libraries, which makes sense to me. It was discussed on an issue.

Here is the issue I was referencing https://github.com/open-telemetry/opentelemetry-collector/issues/7890

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,51 @@
...
+// 1. configtelemetry.None
+//
+// No telemetry data should be collected.
+//
+// 2. configtelemetry.Basic
+//
+// Signals associated with this level cover the essential coverage of the collector telemetry.
+// It should only be used for internal collector telemetry.
+//
+// This is the default level recommended when running the Collector.
```
dmitryax, line 21: Currently, Normal is the default. Are we going to change that? I'd suggest not changing it but keeping this level lean instead

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,51 @@
...
+// When adopting a telemetry level, the developer is expected to rely on this guidance to
+// justify their choice of telemetry level.
+//
+// 1. configtelemetry.None
+//
+// No telemetry data should be collected.
+//
+// 2. configtelemetry.Basic
+//
+// Signals associated with this level cover the essential coverage of the collector telemetry.
```
dmitryax, line 18:
```suggestion
// Signals associated with this level provide the essential coverage of the collector telemetry.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,51 @@
...
+// It should only be used for internal collector telemetry.
```
dmitryax, line 19:
```suggestion
// It should only be used for internal collector telemetry generated by the collector core API. Components SHOULD NOT use this level to report additional metrics.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,51 @@
...
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
+// It should be the default for component authors.
+//
+// Signals using this telemetry level can use this guidance:
+//   - The signals associated with this level must control cardinality.
+//     It is acceptable at this level for cardinality to scale linearly with the monitored resources.
+//   - The signals associated with this level must represent a controlled data volume. Examples follow:
+//   - A max cardinality (total possible combinations of dimension values) of 500.
+//   - At most 5 spans actively recording simultaneously.
```
dmitryax, line 41: How is this feasible? Collector is expected to support processing thousands of concurrent requests

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,51 @@
...
+// 3. configtelemetry.Normal
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
+// It should be the default for component authors.
+//
+// Signals using this telemetry level can use this guidance:
+//   - The signals associated with this level must control cardinality.
+//     It is acceptable at this level for cardinality to scale linearly with the monitored resources.
+//   - The signals associated with this level must represent a controlled data volume. Examples follow:
+//   - A max cardinality (total possible combinations of dimension values) of 500.
```
dmitryax, line 40: Is it per monitored resource? If so it seems too much.

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,51 @@
...
+// Signals associated with this level cover the essential coverage of the collector telemetry.
+// It should only be used for internal collector telemetry.
+//
+// This is the default level recommended when running the Collector.
+//
+// Signals using this telemetry level can use this guidance:
+// * The signals associated with this level must show low cardinality, the number of combinations of dimension values.
+// * The signals associated with this level must represent a small data volume. Examples follow:
+//   - A max cardinality (total possible combinations of dimension values) of 50.
+//   - At most a span actively recording simultaneously, covering the critical path.
```
dmitryax, line 27: I don't think we should add this section, given that we don't expect components to emit at this level.

config/configtelemetry/doc.go
atoulme: fixed

config/configtelemetry/doc.go
atoulme: sounds good, removing

config/configtelemetry/doc.go
atoulme: true, does this clarify?
```suggestion
//   - At most 5 spans actively recording simultaneously per request
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+//
+// This is the default level recommended when running the Collector.
+//
+// * All signals defined in the component telemetry are active.
+//
+// 4. configtelemetry.Detailed
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
+//
+// The signals associated with this level may exhibit high cardinality.
```
TylerHelmuth, line 41:
```suggestion
// The signals associated with this level may exhibit high cardinality and/or high dimensionality.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+// 3. configtelemetry.Normal
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
+// It should be the default for component authors.
+//
+// Signals using this telemetry level can use this guidance:
+//   - The signals associated with this level must control cardinality.
+//     It is acceptable at this level for cardinality to scale linearly with the monitored resources.
+//   - The signals associated with this level must represent a controlled data volume. Examples follow:
+//   - A max cardinality (total possible combinations of dimension values) of 500.
```
TylerHelmuth, line 30:
```suggestion
//     - A max cardinality (total possible combinations of dimension values) of 500.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+//   - At most 5 spans actively recording simultaneously.
```
TylerHelmuth, line 31:
```suggestion
//     - At most 5 spans actively recording simultaneously.
```

TylerHelmuth, line 31:
```suggestion
//     - At most 5 spans actively recording simultaneously.
```

TylerHelmuth (approved):

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
 // Package configtelemetry defines various telemetry level for configuration.
 // It enables every component to have access to telemetry level
 // to enable metrics only when necessary.
+//
+// This document provides guidance on which telemetry level to adopt for Collector metrics.
+// When adopting a telemetry level, the developer is expected to rely on this guidance to
```
codeboten, line 9:
```suggestion
// When adopting a telemetry level, component authors are expected to rely on this guidance to
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+// justify their choice of telemetry level.
+//
+// 1. configtelemetry.None
+//
+// No telemetry data should be collected.
```
codeboten, line 14:
```suggestion
// No telemetry data is recorded.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+//
+// 2. configtelemetry.Basic
+//
+// Signals associated with this level provide the essential coverage of the collector telemetry.
```
codeboten, line 18:
```suggestion
// Telemetry associated with this level provides essential coverage of the collector telemetry.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+// It should only be used for internal collector telemetry generated by the collector core API. Components SHOULD NOT use this level to report additional metrics.
```
codeboten, line 19:
```suggestion
// It should only be used for internal collector telemetry generated by the collector core API. Components outside of 
// the core API MUST NOT record additional telemetry at this level.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+//
+// 3. configtelemetry.Normal
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
```
codeboten, line 23:
```suggestion
// Telemetry associated with this level provides complete coverage of the collector telemetry.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+// It should be the default for component authors.
+//
+// Signals using this telemetry level can use this guidance:
```
codeboten, line 26:
```suggestion
// Component authors using this telemetry level can use this guidance:
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+//     - A max cardinality (total possible combinations of dimension values) of 500.
+//     - At most 5 spans actively recording simultaneously.
+//
+// This is the default level recommended when running the Collector.
+//
+// * All signals defined in the component telemetry are active.
+//
+// 4. configtelemetry.Detailed
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
```
codeboten, line 39:
```suggestion
// Telemetry associated with this level provides complete coverage of the collector telemetry.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+// Signals using this telemetry level can use this guidance:
+//   - The signals associated with this level must control cardinality.
+//     It is acceptable at this level for cardinality to scale linearly with the monitored resources.
+//   - The signals associated with this level must represent a controlled data volume. Examples follow:
+//     - A max cardinality (total possible combinations of dimension values) of 500.
+//     - At most 5 spans actively recording simultaneously.
+//
+// This is the default level recommended when running the Collector.
+//
+// * All signals defined in the component telemetry are active.
```
codeboten, line 35:
```suggestion
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,43 @@
...
+//
+// 4. configtelemetry.Detailed
+//
+// Signals associated with this level cover the complete coverage of the component telemetry.
+//
+// The signals associated with this level may exhibit high cardinality and/or high dimensionality.
+//
+// There is no limit on data volume.
+//
+// All signals defined in the component telemetry are active.
```
codeboten, line 45:
```suggestion
```

codeboten (commented):
Some suggestions

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,40 @@
...
+// 3. configtelemetry.Normal
+//
+// Telemetry associated with this level provides complete coverage of the collector telemetry.
+// It should be the default for component authors.
+//
+// Component authors using this telemetry level can use this guidance:
+//   - The signals associated with this level must control cardinality.
+//     It is acceptable at this level for cardinality to scale linearly with the monitored resources.
+//   - The signals associated with this level must represent a controlled data volume. Examples follow:
+//     - A max cardinality (total possible combinations of dimension values) of 500.
```
atoulme, line 31:
```suggestion
//     - A max cardinality (total possible combinations of dimension values) for a given metric of at most 100.
```

config/configtelemetry/doc.go
```diff
@@ -4,4 +4,40 @@
...
+//     - At most 5 spans actively recording simultaneously.
```
atoulme, line 32:
```suggestion
//     - At most 5 spans actively recording simultaneously per active request.
```

config/configtelemetry/doc.go
atoulme: following up on @dmitryax comment above, lowering from 500 to 100.

config/configtelemetry/doc.go
atoulme: addressed below

jaronoff97 (approved):
Looks great, I appreciate that these are now recommendations. I think having basic remain the core telemetry also makes sense to me. Thank you for this.

codeboten (approved):

PR #10372, opened 2024-06-10
Add "inserted" metrics to processors

### LINKED ISSUE #10353: Add "inserted" metrics to processorhelper
The current set of autogenerated metrics in processorhelper do not represent telemetry which was inserted by a processor. This is not a terribly common use case but is technically possible and can be important to understand.

A couple examples:
- [metrics generation processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricsgenerationprocessor)
- An aggregation processor may aggregate 10 items into 1. When doing so, I think it would be reasonable to describe this as 10 dropped, 1 inserted.
- A delta-to-cumulative processor should arguably report N dropped, N inserted.

 I'd like to add one additional metric for each data type, called `processor_inserted_*`.

djaglowski:
#### Link to tracking issue

Resolves #10353

#### Testing

Added equivalent testing to other processor metrics (accepted, refused, dropped).

#### Documentation

Metric documentation is autogenerated.

#### Open Question

My initial implementation includes a breaking change to `componenttest.TestTelemetry` which is public facing API. If we want to avoid an immediate breaking change in this test package, I would propose the following, which I can submit in a prerequisite PR:
1. Deprecate all `TestTelemetry.Check*` methods.
2. Replace with more granular `TestTelemetry.CheckOneSpecificMetric` methods.

### DIFF
--- a/.chloggen/processorhelper-inserted-api.yaml
+++ b/.chloggen/processorhelper-inserted-api.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component/componenttest
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added additional "inserted" count to `TestTelemetry.CheckProcessor*` methods.
+
+# One or more tracking issues or pull requests related to the change
+issues: [10353]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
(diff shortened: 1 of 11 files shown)


mx-psi (commented):
I am okay with breaking API in one go, but the breaking change should appear in the API changelog. 

I am not sure if there is any overlap with open-telemetry/oteps/pull/259 (I guess not in terms of the metric names, but possibly on the namespaces?). Is there a chance we would have to change these after that OTEP is finalized?

djaglowski:
> I am okay with breaking API in one go, but the breaking change should appear in the API changelog.

Thanks, I'll add a changelog - just wanted to agree on the direction before going further with it.
 
> I am not sure if there is any overlap with [open-telemetry/oteps/pull/259](https://github.com/open-telemetry/oteps/pull/259) (I guess not in terms of the metric names, but possibly on the namespaces?). Is there a chance we would have to change these after that OTEP is finalized?

I think there is overlap and if the otep is accepted we will need to make quite a few changes. However, my understanding of the proposal is that all of the counts we currently have for processors, as well as these "inserted" counts would be relevant, though they would be grouped with others into new metrics. I think regardless of this PR we may end up deprecating the current metrics in favor of a new set, but both could coexist for a while.

mx-psi (approved):
LGTM, will wait a few days in case others want to chime in

PR #10307, opened 2024-06-04
POC: Add support for profiles within experimental packages

dmathieu:
#### Description

This is a proof of concept adding support for profiles as experimental packages.

#### Link to tracking issue
Related: https://github.com/open-telemetry/opentelemetry-collector/issues/10207

cc @mx-psi

PR #10253, opened 2024-05-29
[PoC] How to separate profiles in receivers

mx-psi:
- **[chore] Move parts of receiver to internal package and reexport**
- **[PoC] Example on how to separate profiles in receivers**

#### Link to tracking issue
Fixes #

PR #10375, opened 2024-06-10
RFC: Experimental profiling

dmathieu:
#### Description

This is an RFC discussing the proposed solution for adding the profiling signal.
Follows https://github.com/open-telemetry/opentelemetry-collector/issues/10207

PoC: https://github.com/open-telemetry/opentelemetry-collector/pull/10307

cc @open-telemetry/profiling-triagers @open-telemetry/profiling-approvers @open-telemetry/profiling-maintainers 


docs/rfcs/experimental-profiling.md
```diff
@@ -0,0 +1,86 @@
...
+has been discarded.
+It aimed to refactor the current packages with per-signal subpackages, so each
+subpackage could have its own stability level, like pdata does.
+
+This has been discarded, as the Collector SIG does not want the profiling
+signal to impact the road to the collector reaching 1.0.
+
+## Proposed approach
+
+The proposed approach will consist in two main phases:
```
christos68k, line 28:
```suggestion
The proposed approach will consist of two main phases:
```

christos68k (commented):
LGTM

docs/rfcs/experimental-profiling.md
```diff
@@ -0,0 +1,86 @@
...
+a new `XXexperimental` package (eg: `consumerexperimental`) which will be able
+to handle the profiling signal.
+
+The stable related stable packages MUST not be aware of their experimental
+counterpart.
+The experimental package MAY call its stable implementation.
+
+### Mark specific APIs as `experimental`
+
+In order to boot a functional collector with profiles support, some stable
```
mx-psi, line 47: Have you explored any of the alternatives listed on https://github.com/open-telemetry/opentelemetry-collector/issues/10207#issuecomment-2135051693 ? Do you have an idea of what it would entail to make service signal-agnostic?

Not saying this is not the right solution, just that I would like to see a more thorough justification of why the alternatives are not preferred

dmathieu: I've been trying not to make any changes to stable packages, to go with the requirement to be able to move as fast as possible to v1.
So the proposals `Having two Service constructors` and `Have an unstable Setting for enabling profiles` seems hard to do, since they would require some changes to the public API.

As for `Have all functionality related to profiles under a build tag`, the problem with build tags is that they can't be specified for library imports.
So that would make it very hard to start building receivers and connectors.

mx-psi: `Having two Service constructors` would not require changes to the public API, only to the internals. 

I can't speak for other maintainers (I guess to get a clear answer we would need to have a clear picture of the changes in question), but I personally am okay with public API changes to Service if they are a clear win, since `service` as a Go module is still far from being marked as 1.0.

Agreed with the rest (build tags lead to a bad UX for that very reason), but I think we should include those details in the text of the proposal itself.

docs/rfcs/experimental-profiling.md
```diff
@@ -0,0 +1,99 @@
...
+packages need to be aware of the experimental ones.
+
+Currently, the only package this applies to is `service`.
```
mx-psi, line 63: Do we know what the impact would be if we were to change these experimental symbols? It would also be interesting to have the list of experimental symbols available

docs/rfcs/experimental-profiling.md
```diff
@@ -0,0 +1,99 @@
...
+// Notice: This method is EXPERIMENTAL and may be changed or removed in a
+// later release.
+```
+
+As documented, APIs marked as experimental may changed or removed across
+releases, without it being considered as a breaking change.
+
+#### User specified configuration
+
+The user-specified configuration will let users specify a `profiles` pipeline:
```
codeboten, line 80: It would be great if we could display a log warning for end users that have configured experimental (or as per the spec change development) signals.

This would align well with how the collector today warns users if they've configured development components

codeboten (commented):
Thanks @dmathieu, this looks great. Just one comment

mx-psi:
@dmathieu I think #10253 is a possible way to address this, since it addresses this:

> But this can't work, as the WithProfiles method needs to manipulate private values, and can't do that from an outside module.

Could you take a look and let me know if that addresses the issue?

docs/rfcs/experimental-profiling.md
dmathieu: I have changed the RFC to follow https://github.com/open-telemetry/opentelemetry-collector/pull/10253 instead.

docs/rfcs/experimental-profiling.md
mx-psi: What I would like to have is a list of the methods, fields and symbols in the `service` package that we would need to mark as experimental after we implement this proposal (either in text or with a link to the PoC).

From a quick look, I think the only one would be a `profiles` constant for defining profiles pipelines, I don't see any other places (provided that we handle the `Builder` part manually).

mx-psi: Oki, then I think we can frame it as "there are no symbols that would need to be marked as experimental today, if there ever are then implementers may add an experimental comment to them". What do you think?

mx-psi (approved):
I think we can merge this as is and as noted on the Collector SIG I believe we can merge this as the general plan on Friday. 

One remaining step is what to do with service and `<receiver|exporter|connector|processor>.Builder`. The problem is that we don't pass the configuration directly to the service and instead pass `Builders`. These builders encapsulate the factories and config, thus we lose access to the config.

We will need to solve this problem during implementation, some options are:
1. Add a `Config` method to builders similar to the `Factory` method
2. Make builders internal to the service and pass the config and factories explicitly (honestly at this point my preferred option, there is no real abstraction/usefulness to have these exposed)
3. Make the builders into sealed interfaces like with the factory interfaces? (not sure if there's much of a point for this)

dmathieu:
Following a discussion with @mx-psi this morning, we have decided to follow option 2 mentioned above for the builders.

dmathieu:
Sure, I am moving towards keeping the current builders and having it deprecated. The receiver PR is done (the only one which is not a draft, and has open questions).

PR #10391, opened 2024-06-12
[confmap] Add featuregate to use stable expansion rules

TylerHelmuth:
#### Description
This PR adds a feature gate that will handle transitioning users away from expandconverter, specifically expanding `$VAR` syntax.

The wholistic strategy is:
  1. create a new feature gate, `confmap.unifyEnvVarExpansion`, that will be the single feature gate to manage unifying collector configuraiton resolution.
  2. Update expandconverter to return an error if the feature gate is enabled and it is about to expand `$VAR` syntax.
  3. Update `otelcol.NewCommand` to set a `DefaultScheme="env"` when the feature gate is enabled and no `DefaultScheme` is set, this handles `${VAR}` syntax.
  4. Separately, deprecate `expandconverter`.
  5. Follow a normal feature gate cycle.
  6. Removed the `confmap.unifyEnvVarExpansion`  feature gates and `expandconverter` at the same time

Supersedes https://github.com/open-telemetry/opentelemetry-collector/pull/10259

#### Link to tracking issue

#### Testing
Unit tests


confmap/converter/expandconverter/expand.go
```diff
@@ -80,21 +80,27 @@ func (c converter) expandStringValues(value any) (any, error) {
 func (c converter) expandEnv(s string) (string, error) {
 	var err error
 	res := os.Expand(s, func(str string) string {
+		// This allows escaping environment variable substitution via $$, e.g.
+		// - $FOO will be substituted with env var FOO
+		// - $$FOO will be replaced with $FOO
+		// - $$$FOO will be replaced with $ + substituted env var FOO
+		if str == "$" {
+			return "$"
+		}
```
TylerHelmuth, line 89: Can handle this via https://github.com/open-telemetry/opentelemetry-collector/pull/10392 if you want

confmap/resolver.go
```diff
@@ -108,7 +115,8 @@ func NewResolver(set ResolverSettings) (*Resolver, error) {
 
 	converters := make([]Converter, len(set.ConverterFactories))
 	for i, factory := range set.ConverterFactories {
-		converters[i] = factory.Create(set.ConverterSettings)
+		f := factory.Create(set.ConverterSettings)
+		converters[i] = f
```
mx-psi, line 119: Is this change needed?

confmap/converter/expandconverter/expand.go
mx-psi: Merged!

confmap/resolver.go
TylerHelmuth: No, that was from a previous attempt at something, I will revert

otelcol/command.go
```diff
@@ -52,6 +53,11 @@ func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet) error
 	if len(resolverSet.URIs) == 0 {
 		return errors.New("at least one config flag must be provided")
 	}
+
+	if confmap.UseUnifiedEnvVarExpansionRules.IsEnabled() && set.ConfigProviderSettings.ResolverSettings.DefaultScheme == "" {
+		set.ConfigProviderSettings.ResolverSettings.DefaultScheme = "env"
```
TylerHelmuth, line 58: This solution has a downside: it expects that the `ProviderFactories` list includes a provider with the `env` scheme. 

This will be true for all our own usage of `NewCommand`, but may not be true for custom usages.

confmap/resolver.go
```diff
@@ -12,8 +12,15 @@ import (
 
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/featuregate"
 )
 
+var UseUnifiedEnvVarExpansionRules = featuregate.GlobalRegistry().MustRegister("confmap.unifyEnvVarExpansion",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterFromVersion("v0.102.0"),
```
codeboten, line 21:
```suggestion
	featuregate.WithRegisterFromVersion("v0.102.1"),
```

otelcol/command.go
```diff
@@ -52,6 +53,11 @@ func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet) error
 	if len(resolverSet.URIs) == 0 {
 		return errors.New("at least one config flag must be provided")
 	}
+
+	if confmap.UseUnifiedEnvVarExpansionRules.IsEnabled() && set.ConfigProviderSettings.ResolverSettings.DefaultScheme == "" {
```
codeboten, line 57: Can this flag be tested?

otelcol/command.go
TylerHelmuth: Since this is only affects users who enable the feature gate, and the feature gate is only useful if you're using the env provider, I believe we're safe.

confmap/resolver.go
mx-psi: I guess v0.103.0 actually, no?

TylerHelmuth: Updated to 0.103.0

otelcol/command.go
TylerHelmuth: Added

codeboten (approved):

confmap/resolver.go
```diff
@@ -12,8 +12,15 @@ import (
 
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/featuregate"
 )
 
+var UseUnifiedEnvVarExpansionRules = featuregate.GlobalRegistry().MustRegister("confmap.unifyEnvVarExpansion",
```
mx-psi, line 19: Can we make it so that this is not a public symbol? (e.g. move to `confmap/internal`)

mx-psi (approved):
One small change and it's good to go from my side

evan-bradley (approved):

confmap/resolver.go
TylerHelmuth: @mx-psi we cannot, since `otelcol.NewCommand` depends on it.

codeboten (commented):
Needs rebasing

confmap/resolver.go
TylerHelmuth: Can do

PR #10394, opened 2024-06-12
[configgrpc] Revert to third-party compressor for zstd

jpkrohling:
We switched back to our compression logic for zstd when a CVE was found on the third-party library we were using. Now that the third-party library has been fixed, we can revert to that one. For end-users, this has no practical effect. The reproducers for the CVE were tested against this patch, confirming we are not reintroducing the bugs.


TylerHelmuth (approved):

songy23 (commented):
Needs `make gotidy`

codeboten (approved):

PR #10398, opened 2024-06-13
[chore] link release managers

codefromthecrypt:
#### Description

In normal github issues etc, if you `@user`, it makes a link for you. So, it feels weird to see `@user` without a link. Feel free to reject this if it is a cure worse than the disease.


mx-psi (approved):
Thanks!

PR #10401, opened 2024-06-13
Add pprofile wrapper and testdata

dmathieu:
#### Description

This adds pprofile wrapper (the non-generated bit of pprofile), as well as test data.

cc @mx-psi


dmathieu:
This windows failure is surprising. `pprofile` isn't in the `testdata` `go.mod` because it's being replaced with the local version.

cmd/builder/internal/builder/main_test.go
```diff
@@ -78,6 +78,7 @@ var (
 		"/otelcol",
 		"/pdata",
 		"/pdata/testdata",
+		"/pdata/pprofile",
```
mx-psi, line 81: What is this needed for?

pdata/testdata/profile.go
```diff
@@ -0,0 +1,51 @@
...
+	profileStartTimestamp = pcommon.NewTimestampFromTime(time.Date(2020, 2, 11, 20, 26, 12, 321, time.UTC))
+	profileEndTimestamp   = pcommon.NewTimestampFromTime(time.Date(2020, 2, 11, 20, 26, 13, 789, time.UTC))
+)
+
+// GenerateProfiles generates dummy profiling data for tests
+//
+// # Experimental
+//
+// Notice: This method is EXPERIMENTAL and may be changed or removed in a
+// later release.
```
mx-psi, line 23: `pdata/testdata` is not stable https://github.com/open-telemetry/opentelemetry-collector/blob/05a2844bdb4f4483efed9eca9c3d6e54c859ebaa/versions.yaml#L50 so IMO we don't need this. We are also (@open-telemetry/collector-approvers to correct me here if I am wrong) targeting this one for 1.0, so I think adding the function here is fine.

cmd/builder/internal/builder/main_test.go
dmathieu: Without it, the test fails.
https://github.com/open-telemetry/opentelemetry-collector/actions/runs/9597639894/job/26467167308?pr=10401

pdata/testdata/go.mod
```diff
@@ -2,7 +2,10 @@ module go.opentelemetry.io/collector/pdata/testdata
 
 go 1.21.0
 
-require go.opentelemetry.io/collector/pdata v1.10.0
+require (
+	go.opentelemetry.io/collector/pdata v1.10.0
+	go.opentelemetry.io/collector/pdata/pprofile v0.102.1
```
mx-psi, line 7:
```suggestion
	go.opentelemetry.io/collector/pdata/pprofile v0.103.0
```

mx-psi (approved):

PR #10406, opened 2024-06-13
remove opencensus bridge support

### LINKED ISSUE #8945: Collector shutdown leaks a goroutine
**Describe the bug**
Otel Collector shutdown leaves a leaking goroutine
reproduced here: https://github.com/jskiba/opentelemetry-collector/blob/main/cmd/otelcorecol/main_test.go
```
=== RUN   TestCollector_Shutdown
    main_test.go:23: found unexpected goroutines:
        [Goroutine 19 in state select, with go.opencensus.io/stats/view.(*worker).start on top of the stack:
        go.opencensus.io/stats/view.(*worker).start(0xc0001d3400)
        	/Users/jakub.skiba/go/pkg/mod/go.opencensus.io@v0.24.0/stats/view/worker.go:292 +0x9f
        created by go.opencensus.io/stats/view.init.0 in goroutine 1
        	/Users/jakub.skiba/go/pkg/mod/go.opencensus.io@v0.24.0/stats/view/worker.go:34 +0x8d
        ]
--- FAIL: TestCollector_Shutdown (0.44s)
```
I haven't investigated that but opencensus seems to be archived so if it is not a bug in otel collector then it might be problematic.

**Steps to reproduce**
https://github.com/jskiba/opentelemetry-collector/blob/main/cmd/otelcorecol/main_test.go

No leaking goroutines

**What did you see instead?**
Leaking goroutine

**What version did you use?**
Version: [abdeacb](https://github.com/open-telemetry/opentelemetry-collector/commit/abdeacbfc465a8b5e0ab4dd4a9f215cbf623ccbe)

**What config did you use?**
Config: nono
It reproduces every time. The config does not matter here.

**Environment**
OS: macOS Sonoma 14.1.1
Compiler: go 1.21.1

### LINKED ISSUE #10414: Remove opencensus bridge
This will enable us to close the loop on many goleak exceptions that are in place today. Additionally it will allow us to move the meter provider configuration code to use the otel-go config package.

Removing the bridge requires the completion of the migration away from opencensus in the contrib repo.

codeboten:
This removes the configuration of the OpenCensus bridge from the Collector. This means that any metric still relying on it will no longer be emitting metrics.

Closes #10414
Closes #8945

This is blocked until https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29867 is completed.


jpkrohling (approved):
Do we still have components using opencensus? I see two of them are mine, but I can commit to getting them migrated before the new release in two weeks.

codeboten:
> Do we still have components using opencensus? I see two of them are mine, but I can commit to getting them migrated before the new release in two weeks.

There are 4 as listed in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29867

codeboten:
After discussing this further, i'll add a feature gate before this change.

codeboten:
This will stay in draft until https://github.com/open-telemetry/opentelemetry-collector/pull/10542 has been merged and released.

TylerHelmuth (approved):
🎉

Issue #10407, opened 2024-06-14
Update the image version of the sample


chenzhiguo:
**Is your feature request related to a problem? Please describe.**
During the installation of otlp in recent days, it was found that the installed version is somewhat outdated, which is not conducive to users trying out the latest features. It's also unnecessary to modify the installation yaml file every time a new version is released. 

**Describe the solution you'd like**
Would it be more appropriate to change it to "latest"?

chenzhiguo:
Close by #10408

PR #10408, opened 2024-06-14
Update image version to latest #10407

chenzhiguo:
During the installation of otlp in recent days, it was found that the installed version is somewhat outdated, which is not conducive to users trying out the latest features. It's also unnecessary to modify the installation yaml file every time a new version is released. Would it be more appropriate to change it to "latest"?


songy23 (approved):

chenzhiguo:
@songy23 Hi, what can i do for check task `contrib_tests Expected — Waiting for status to be reported`?

songy23:
This PR doesn't affect collector-contrib so you don't need to worry about that @chenzhiguo

Issue #10058, opened 2024-04-30
[component] Status Aggregation in Core


mwear:
[Functions to handle status aggregation](https://github.com/open-telemetry/opentelemetry-collector/blob/fe582df553666510372ada384b79d6f0a2253e0f/component/status.go#L101-L190) were added to core when status reporting was originally implemented. These functions were initially adequate while I was working on an alternative health check extension, but as more requirements surfaced, I had to reimplement alternative logic to handle all of the use cases. [This PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32695) specifically introduces the alternative aggregation logic and [this PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30673) shows how everything works together.

The goal of status aggregation is to take a collection of status events and combine them into a single event that represents the most relevant status based on the collection as a whole. One assumption we made in the original aggregation logic, was that `PermanentError`s should take precedence over `RecoverableError`s. While this still the default aggregation used in the health check extension, users have the option to include or ignore permanent and / or recoverable errors independently. If a user ignores `PermanentError`s, but includes `RecoverableError`s, we want `RecoverableError`s to take priority over `PermanentError`s during aggregation. In order to handle this is this in the extension, I implemented a function that takes an `ErrorPriority` and returns a suitable aggregation function.

We can divide component statuses into two categories. Lifecycle statuses (eg `Starting`, `Stopping`, `Stopped`) and runtime statuses (eg `RecoverableError`, `PermanentError`, `FatalError`). In the original aggregation logic, error statuses take precedence over lifecycle statuses. In the health check extension I had to change the aggregation logic to give precedence to lifecycle events in order to distinguish between a collector that is starting up or shutting down (a 503 status) or a collector in an error state (a 500).

There is a tangentially related issue with `PermanentError`s and the underlying finite state machine that governs transitions between statuses. Currently, a `PermanentError` is a final state. That is, once a component enters this state, no further transitions are allowed. In light of the work I did on the alternative health check extension, I believe we should allow a transition from `PermanentError` to `Stopping` to consistently prioritize lifecycle events for components. This transition also make sense from a practical perspective. A component in a `PermanentError` state is one that has been started and is running, although in a likely degraded state. The collector will call shutdown on the component (when the collector is shutting down) and we should allow the status to reflect that.

[...]

Issue #9957, opened 2024-04-12
[component] Runtime status reporting


mwear:
## Runtime Status Reporting

As part of the component 1.0 milestone we should implement runtime status reporting for core components and come up with guidelines and best practices for incremental adoption by other components. This issue gives some background information on component status reporting, and outlines how it should work for different component types.

### Component Status Events

<img width="664" alt="Screenshot 2024-04-12 at 9 52 14 AM" src="https://github.com/open-telemetry/opentelemetry-collector/assets/2513372/69dbdd73-b294-4bed-93df-67525d1dbb49">

Component status events can be broken down into two categories: lifecycle events (denoted by blue in the diagram) and runtime events (green and red). Lifecycle events e.g. `StatusStarting`, `StatusStopping` are reported automatically by the collector and it is the responsibility of components to report their runtime status. During runtime it will be common for a component to transition between `StatusOK` and `StatusRecoverableError`. In some situations, a component may detect an unrecoverable state, and transition into `StatusPermanentError`. This is a final state that cannot be transitioned out of and indicates a human will have to intervene to fix it. 

The state transitions are governed by a finite state machine and the intention is that components should not have to keep track of their internal state when reporting status. Components can report `StatusOK` when an operation succeeds and `StatusRecoverableError` when an operation fails (in a recoverable way). Status events will only be emitted when a component's state changes. So repeat reports of the same status will have no effect. Likewise, if a component has transitioned into a final state (e.g. `StatusPermanentError`), subsequent attempts to report status will no-op.

#### Consumers of Status Events

There is a [PR for a new version of the health check extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30673) that is based on component status reporting. It uses lifecycle events to determine if the collector is ready and running and allows users to opt-in to having recoverable or permanent errors factored in to collector health. The OpAMP extension will make use of these events for component health at some point in the future. Any extension that implements the option `StatusWatcher` interface can be a consumer of component status events.

### Adoption and Best Practices

[...]

Issue #9823, opened 2024-03-22
Confused about the difference between PermanentError vs FatalError


bogdandrutu:
Per definitions from https://github.com/open-telemetry/opentelemetry-collector/issues/9324

Status | Meaning
-- | --
PermanentError | The component has detected a condition at runtime that will need human intervention to fix. The collector will continue to run in a degraded mode.
FatalError | The collector has experienced a fatal runtime error and will shutdown.

When do I use `PermanentError` vs `FatalError`? If a component is permanently failing does that not mean fatal? Why? If my whole collector doesn't accept any data because that was my only receiver then I think is better to fail than transition in permanent error.

mx-psi:
cc @mwear

mwear:
The intent of having a permanent error state is to allow the collector to run in a degraded mode, and communicate issues to operators through the health check extension or something similar. In your example where the collector has a single non-functional receiver, this would not be very useful, but in a collector with multiple pipelines, it is possible for some pipelines to be running normally, and others to have issues. In a scenario where you are exporting to multiple backends, one exporter could be working fine, while another could be failing due to misconfiguration. The misconfiguration isn't apparent until trying to send data.

I think we need to establish some guidelines, but I do think allowing a collector to run in a degraded fashion, rather than terminating altogether is a useful thing to support.

mwear:
I would suggest that we consider using the following (updated) definitions:

| Status            | Meaning                                                                |
|-------------------|------------------------------------------------------------------------|
| PermanentError    | The component has detected a condition after start that will need human intervention to fix. The collector will continue to run in a degraded mode. |
| FatalError        | The collector has experienced an unrecoverable error during startup and will shutdown. |

A `FatalError` can be used to fail fast during startup when detecting an unrecoverable error condition. An example would be an receiver that is unable to bind to its configured port.

A `PermanentError` would be an unrecoverable error that happens after start. For example, an exporter attempts to send data and receives a http.StatusUnauthorized (401) or codes.Unauthenticated (16).

If you are interested, I attempted to enumerate permanent errors for exporters in #8684. I suspect some of the choices are controversial, and would like to find a more universal implementation, but the description captures my general thoughts.

bogdandrutu:
> A FatalError can be used to fail fast during startup when detecting an unrecoverable error condition. An example would be a receiver that is unable to bind to its configured port. 
> The collector has experienced an unrecoverable error during startup and will shutdown.

[...]

Issue #9324, opened 2024-01-20
[component] component.go - when is component.Start returning errors?


atoulme:
The `component.Start` function returns an error.

When is an error expected to be returned upon starting? Configuration validation happens outside of the lifecycle of the component.

The error is handled as reporting a permanent error:
https://github.com/open-telemetry/opentelemetry-collector/blob/c5a2c78d614321dcece254f7b79281b84127b5fc/service/internal/graph/graph.go#L396-L402

A few examples from contrib from working with lifecycle tests recently:
* A cassandra exporter starts, but cannot connect to Cassandra right away. It returns an error.
* An aws component starts, but cannot create an aws client, because environment variables are missing. It returns an error.

Typically, we see those when the components cannot pass lifecycle tests:
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues?q=is%3Aissue+is%3Aopen+skip+lifecycle

Given that we have component status reporting now, would it make sense to ask component implementers to handle start failures via component status instead of returning an error?

mwear:
Thanks for opening this issue @atoulme. I don't think that we have well defined rules for how components should handle errors during start, but now that we have component status reporting, and an in progress health check extension that makes use of that data, it makes sense for us to come up with guidelines this should work together.

This table has the definitions that I have in mind for each status:

| Status            | Meaning                                                                |
|-------------------|------------------------------------------------------------------------|
| Starting          | The component is starting.                                             |
| OK                | The component is running without issue.                                |
| RecoverableError  | The component has experienced a transient error and may recover.       |
| PermanentError    | The component has detected a condition at runtime that will need human intervention to fix. The collector will continue to run in a degraded mode. |
| FatalError        | The collector has experienced a fatal runtime error and will shutdown. |
| Stopping          | The component is in the process of shutting down.                      |
| Stopped           | The component has completed shutdown.                                  |

Now that we have the permanent error state, I think it makes sense to ask whether or not we need fatal error any longer. The reason it was included with status reporting is that it was a preexisting error state. The intent of the permanent error state is to allow the collector to continue running in a degraded mode and alert users that there was a condition detected at runtime that will require human intervention to fix.

I would propose the following guidelines for start:

- If a component experiences a recoverable error, it should report it, but not return an error from start
- If an error is not detected by the end of start, but becomes apparent later, it should be reported as a permanent error. The same should apply to async errors that could arise from start.
[...]

PR #10413, opened 2024-06-14
[component] Component status reporting rfc

TylerHelmuth:
#### Description
Adds an RFC for component status reporting.  The main goal is to define what component status reporting is, our current, implementation, and how such a system interacts with a 1.0 component. When merged, the following issues will be unblocked:
- https://github.com/open-telemetry/opentelemetry-collector/issues/9823
- https://github.com/open-telemetry/opentelemetry-collector/issues/10058
- https://github.com/open-telemetry/opentelemetry-collector/issues/9957
- https://github.com/open-telemetry/opentelemetry-collector/issues/9324
- https://github.com/open-telemetry/opentelemetry-collector/issues/6506


docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,223 @@
...
+  - Open: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/8816
+- Status Reporting PRs
+  - Closed
+    - https://github.com/open-telemetry/opentelemetry-collector/pull/5304
+    - https://github.com/open-telemetry/opentelemetry-collector/pull/6550
+    - https://github.com/open-telemetry/opentelemetry-collector/pull/6560
+  - Merged
+    - https://github.com/open-telemetry/opentelemetry-collector/pull/8169
+
+Matt: Little did I know this comment would change my life forever: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/8816#issuecomment-1083380460
```
mwear, line 220: LOL, while this is true, very true, I'm not sure it needs to be part of the RFC.

TylerHelmuth:
```suggestion
```

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,223 @@
...
+
+4. Panicking. During runtime, if the collector experienced an unhandled error, it crashes.
+
+These are all the way the components in a collector could report that they were unhealthy.
+
+There are several major gaps in the Collector’s historic reporting of component health. First, many components return recoverable errors from Start, causing the collector to shutdown, while it could recover if the collector was allowed to run. Second, when a component experienced a transient error, such as an endpoint suddenly not working, the component would simply log the error and return it up the pipeline. There was no mechanism for the component to tell the Host or anything else that something was going wrong. Last, when a component experienced an issue it would never be able to recover from, such as receiving a 404 response from an endpoint, the component would log the error and return it up the pipeline.  This situation was handle the same as the transient error, which means the component could not tell the Host or anything else that something was wrong, but worse is that the issue would never get better.
+
+## Current State of Component Health Reporting
```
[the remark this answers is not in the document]
TylerHelmuth: I like that idea

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Open Questions and Proposed Changes
```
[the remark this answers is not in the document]
TylerHelmuth: Done

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
codeboten: +1 to this idea

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Shutdown
+
+The collector will report a Stopping event when shutting down a component. If Shutdown returns an error, the collector will report a PermanentError event. If Shutdown completes without an error, the collector will report a Stopped event.
+Best Practices
+
+#### Start
+
+Under most circumstances, a component does not need to report explicit status during component.Start. An exception to this rule is components that start async work (e.g. spawn a go routine). This is because async work may or may not complete before start returns and timing can vary between executions. A component can halt startup by returning an error from start. If start returns an error, automated status reporting will report a PermanentError on behalf of the component. If a component wishes to halt startup from work spawned in a go routine, it can report a FatalError. If start returns without an error automated status reporting will report OK, as long as the component hasn't already reported for itself.
```
atoulme, line 95: Is this paragraph describing a deviation?

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Existing deviations from those goals
+
+### Fatal Error Reporting
+
+Before the current implementation of component status reporting, a component could stop the collector by using `Host.ReportFatalError`. Now, a component MUST use component status reporting and emit a `FatalError`. This fact is in conflict with Goal 1, which states component health reporting must be opt-in for components.
+
+A couple solutions:
+1. Accept this reality as an exception to Goal 1.
+2. Add back `Host.ReportFatalError`.
```
atoulme, line 146: Let's not do that please.

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,196 @@
...
+3. Remove the ability for components to stop the collector be removing `FatalError`.
```
atoulme, line 147: Sounds like the right future proof approach. We can and should have the collector decide of its behavior when an error is reported.

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
mwear: Here is a draft of what I had in mind: https://github.com/open-telemetry/opentelemetry-collector/pull/10422.

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,196 @@
...
+
+4. Panicking. During runtime, if the collector experienced an unhandled error, it crashes.
+
+These are all the way the components in a collector could report that they were unhealthy.
+
+There are several major gaps in the Collector’s historic reporting of component health. First, many components return recoverable errors from Start, causing the collector to shutdown, while it could recover if the collector was allowed to run. Second, when a component experienced a transient error, such as an endpoint suddenly not working, the component would simply log the error and return it up the pipeline. There was no mechanism for the component to tell the Host or anything else that something was going wrong. Last, when a component experienced an issue it would never be able to recover from, such as receiving a 404 response from an endpoint, the component would log the error and return it up the pipeline.  This situation was handle the same as the transient error, which means the component could not tell the Host or anything else that something was wrong, but worse is that the issue would never get better.
```
mx-psi, line 35:
```suggestion
There are several major gaps in the Collector’s historic reporting of component health. First, many components return recoverable errors from Start, causing the collector to shutdown, while it could recover if the collector was allowed to run. Second, when a component experienced a transient error, such as an endpoint suddenly not working, the component would simply log the error and return it up the pipeline. There was no mechanism for the component to tell the Host or anything else that something was going wrong. Last, when a component experienced an issue it would never be able to recover from, such as receiving a 404 response from an endpoint, the component would log the error and return it up the pipeline.  This situation was handled in the same way as the transient error, which means the component could not tell the Host or anything else that something was wrong, but worse is that the issue would never get better.
```

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Component TelemetrySettings Requirements
+
+The current implementation of component status reporting added a new field to `component.TelemetrySettings`, `ReportStatus`.  This field is technically optional, but would be marked as stable with component 1.0. Are we ok with 1 of the following?
+
+1. Including a component status reporting feature, `component.TelemetrySettings.ReportStatus`, in the 1.0 version of `component.TelemetrySettings`?
+2. Marking `component.TelemetrySettings.ReportStatus` as experimentatal via godoc comments in the 1.0 version of `component.TelemetrySettings`?
+
+Or should we refactor `component` somehow to remove `ReportStatus` from `component.TelemetrySettings`?
+
```
mx-psi, line 171: My gut feeling is that `ReportStatus` should be either part of `component.Host` or an optional interface to be implemented by `component.Host`s

docs/rfcs/component-status-reporting.md
TylerHelmuth: @atoulme is there a specific objection to `Host.ReportFatalError`?

docs/rfcs/component-status-reporting.md
TylerHelmuth: I like the idea of an optional interface that `component.Host`s can implement.

docs/rfcs/component-status-reporting.md
TylerHelmuth: No it is describing the current way components report status during startup and how the collector service handles automatic status reporting.  I'll remove section once the standalone doc is merged.

docs/rfcs/component-status-reporting.md
TylerHelmuth: If we remove this feature, we'd need to decide how components that are currently doing

```go
if errGrpc := jr.grpc.Serve(ln); !errors.Is(errGrpc, grpc.ErrServerStopped) && errGrpc != nil {
    jr.settings.ReportStatus(component.NewFatalErrorEvent(errGrpc))
}
```

Would handle the error returned from `.Serve`.

One option is for the component to return a `NewPermanent` error, but according to goal 1, component status reporting must be optional.  So for components that are not opt-in to component status reporting, and no longer have `Host.ReportFatalError`, what would they do?

docs/rfcs/component-status-reporting.md
mwear: If we keep the behavior of having the collector shutdown when it experiences a FatalError, I think we will need to reinstate `Host.ReportFatalError` to comply with goal 5:

> 5. Component health reporting must only be a mechanism for reporting health - it should have no mechanisms for taking actions on the health it reports.  How consumers of the health reporting system respond to component updates is not a concern of the health reporting system.

I would not be opposed to bringing back `Host.ReportFatalError`. If we did we can update the automation to report a FatalError for the component that reported it. It keeps component status reporting out of the FatalError  API, but still allows us to generate a fatal error event for Watchers.

docs/rfcs/component-status-reporting.md
mwear: The larger goal for components to report runtime status, see #9957. One issue with implementing `ReportStatus` on host, is that it's passed into start, and for a component to have access to `ReportStatus` outside of `Start` it needs to save a handle, either to the function, or to the host. Few do this, but most components save a handle to TelemetrySettings, which provides similar mechanisms for the component to report data about itself. It's not impossible for it be implemented on Host, but its an inconvenient location.

mx-psi: >  Few do this

I think this is something we can remediate with good examples, documentation and possibly some autogenerated test where we assert common types of status transitions (e.g. failed to create a server).

The only difference I see is that you can't report status during component creation, but I see that as a feature, not a bug (there's going to be no status transition before a component has been initialized).

mwear: I am more interested in the ergonomics for component authors. I think its natural for the API to report status to be in the same place as I find a meter, or logger for a component. The extra step of explicitly having to save a separate handle to host, or its ReportStatus method to use outside of Start makes the API more cumbersome to use.

mx-psi: The advantages I see if we put this in `component.Host` are:

1. Allowing component status to evolve independently from the rest of the core APIs and decoupling this effort from Collector 1.0 (the most important benefit in my mind)
2. Making it impossible for components to misuse `ReportStatus` during creation time
3. Potentially removing the awkwardness in having both a `component.TelemetrySettings` and `servicetelemetry.TelemetrySettings`, and simplifying the implementation of #4970

I feel like these outweigh `component.TelemetrySettings` benefits, but it would be great to have a similar list so we can consider pros/cons

mwear: On point 3, we will actually introduce the same awkwardness in a different spot. We'll end up with a `servicehost.Host` and a `component.Host`. This transition from host to telemetry settings was made once before during the original implementation. See this commit for how things looked when ReportStatus was on Host: https://github.com/open-telemetry/opentelemetry-collector/pull/8169/commits/b45540e44bd7ad7d114baa983241f7995d72ca5d.

mx-psi: Thanks for the context, I think I wasn't super clear on (3). We have several vendors interested in providing their own meter provider, tracer provider, logger... (see #4970) so moving this awkwardness to `component.Host` would potentially simplify adding support for this: all other fields in `component.TelemetrySettings` are governed by the `service::telemetry` config section and we could more easily use a factory pattern if we were left with the other fields. That said, I don't think (3) should be the deciding factor, I see (1) as the most important one.

mwear: We would also need to confirm that we can still support sharedcomponent with moving the API to host. The sharedcomponent requirements were not fully understood until sometime after https://github.com/open-telemetry/opentelemetry-collector/commit/b45540e44bd7ad7d114baa983241f7995d72ca5d. The requirements I'm referring to are discussed here: https://github.com/open-telemetry/opentelemetry-collector/pull/10059#issuecomment-2091896948. They are also outlined in the new document: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md#implementation-details.

mwear: My first impression is that supporting sharedcomponent will be a challenge if we move ReportStatus from TelemetrySettings to Host. We were able to solve this problem from within sharedcomponent by chaining together the ReportStatus functions for each instance the component represents. This is done through the TelemetrySettings instances passed to LoadOrStore which is called during create. It's important that we have a ReportStatus function that can report for all instances of a sharecomponent before start-time, because the component needs to report the starting events for its instances at the same time. Trying to discover the instances from within the sharedcomponent at start-time will not work because its already too late. We would need a way to know what instances a sharedcomponent represents from "outside" the sharedcomponent to make this work. That said, the current sharedcomponent situation is not ideal and perhaps needs improvement anyways.

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,112 @@
...
+## Current State of Component Health Reporting
+
+See [Component Status Reporting](../component-status.md)
+
+## The Goals the Component Health Reporting Should Achieve
+
+The following are the goals, as of June 2024 and with Collector 1.0 looming, for a component health reporting system.
+
+1. Runtime component health reporting, which cannot be automatically reported, must be opt-in for components.  Collector components must not be required to use the component health reporting system. This keeps component as compatible as possible with the Collector’s framework as we approach 1.0
+    - The consumers of the health reporting system must be able to identify which components are and are not using the health reporting system.
```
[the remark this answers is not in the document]
TylerHelmuth: This particular goal is supposed to be scoped just to components. The intention is to allow components to be usable in a collector without utilizing any health reporting status framework.

Allowing the Collector Host to report the basic lifecycle checks it can handle is handle via Goal 4.

Should this first goal be reworded to make it clearer that this is only about components?

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,112 @@
...
+2. Component health reporting must be opt-in for collector users.  While the underlying components are always allowed to report their health via the system, the Collector Host or any other listener may only take action when the user has configured the collector accordingly.
+   - As one example of compliance, the current health reporting system is dependent on the user configuring an extension that can watch for status updates.
+3. Component health must be representable as a finite state machine with clear transitions between states.
+4. The Collector Host may report statuses Starting, Ok,  Stopping and PermanentError on behalf of components as documented in the automation section.
```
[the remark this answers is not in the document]
TylerHelmuth: In this case it is the Collector Service (I think technically it is `serviceHost` on `serivce.Service`, which is implementing `component.Host`.

docs/rfcs/component-status-reporting.md
TylerHelmuth: > If we want these to be decoupled, then isn't the solution to just allow components to report fatal/permanent errors w/o any expectations for how that will be handled? Then a separate layer can decide what to do about it. e.g. shut down the collector, attempt to restart components, etc.

If we want to achieve Goal 1 (components are not required to emit status from within their own code), then we'll need some mechanism for a component to tell the Host to shutdown (assuming we choose to keep that feature).

`Host.ReportFatalError` alone isn't in violation of Goal 5 if it is added back exactly as we had it before. The contract would be "A component can call this function to stop the Collector".  The Host would not be taking any action on a Status reported from the Health Reporting System, it would be taking action based on the Host interfaces contract.

Adding back `Host.ReportFatalError` before 1.0 Component is essentially saying we agree that the `component.Host` interface is a stable way to interact with the Host, but the Health Reporting System is not stable yet.

docs/rfcs/component-status-reporting.md
TylerHelmuth: > Why isn't it acceptable to expect them report a permanent error?

At the moment I am working under the assumption that our current Status Reporting System is too new to be marked stable in component 1.0. Under that assumption, it needs to be ok for a component to NOT be using `ReportStatus` in order to interact with the Collector Host, which, until Status Reporting was added, was always done via `component.Host`.

If we want all the custom components out in the wild to be as compatible with component 1.0 as possible, then relying more on `component.Host` to interact with the Collector Host is better than status reporting.

Maybe there is a future where `component.Host` isn't needed anymore and components only use status reporting, but I don't think we're there yet. At the moment I view status reporting as a great feature that components SHOULD take advantage of, but are not required to.

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,112 @@
...
+
+Before the current implementation of component status reporting, a component could stop the collector by using `Host.ReportFatalError`. Now, a component MUST use component status reporting and emit a `FatalError`. This fact is in conflict with Goal 1, which states component health reporting must be opt-in for components.
+
+A couple solutions:
+1. Accept this reality as an exception to Goal 1.
+2. Add back `Host.ReportFatalError`.
+3. Remove the ability for components to stop the collector be removing `FatalError`.
+
+### No way to identify components that are not reporting status
+Goal 1 states that consumers of component status reporting must be able to identify components in use that have not opted in to component status reporting. Our current implementation does not have this feature.
```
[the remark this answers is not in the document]
TylerHelmuth: > All components should have their basic lifecycle reported automatically

Agreed, thats handled by Goal 4.  Goal 1 is attempting to claim that a component should not be required to call `settings.ReportStatus` in order to be called a component and to function successfully in a collector build.

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
djaglowski: Is there a more technical way we could word this, maybe just say `component.Host`?

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
djaglowski: Yeah I think it could be clearer. I think the question of whether or not we will force implementations of `component.Host` to implement the basic lifecycle should probably be answered first, since the language in this goal depends on whether or not that can be assumed.

docs/rfcs/component-status-reporting.md
djaglowski: I would be in favor of not allowing components to directly stop the collector. Instead, I think there should only be able to report a permanent/fatal error (i.e. whatever is the maximum severity error) and then it is always up to the collector as a whole to decide whether that warrants shutting down.

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
mx-psi: Maybe "service's `component.Host` implementation"?

docs/rfcs/component-status-reporting.md
djaglowski: When you say "interact with the Collector Host", are we only talking about components being able to force the collector to shut down?

docs/rfcs/component-status-reporting.md
TylerHelmuth: >  I think there should only be able to report a permanent/fatal error (i.e. whatever is the maximum severity error) and then it is always up to the collector as a whole to decide whether that warrants shutting down.

This is what we have today, but it effectively is a direct way to stop the collector. So the feature "I want to stop the collector" is locked behind the Status Reporting system.

docs/rfcs/component-status-reporting.md
TylerHelmuth: Yes, but the other ways we interact with the host are accessing the available extensions and accessing available factories.

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
TylerHelmuth: I attempted to clarify the term on line 13:

```
For the sake of simplicity, when `Host` is referenced in this document you can think of the Collector service, which manages all the collector pipelines and the components within.
```

But we can use `service.Service` instead.

docs/rfcs/component-status-reporting.md
djaglowski: It sounds like some notion of reporting a fatal error at any time is the only status-related interaction that we need to pin down for a 1.0 of component. I propose that we commit to such a mechanism for 1.0, but we _do not_ commit to a specific behavior in reaction to the event. 

This leaves us room to offer multiple strategies for reacting to a such an event, including:
- Immediate shutdown of the collector.
- Some automatic attempt to restart the component and/or component graph.
- Custom reactions directed via opamp.
- A full blown component status system which includes the notion of a fatal/permanent error.

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
TylerHelmuth: Updated

docs/rfcs/component-status-reporting.md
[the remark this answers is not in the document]
TylerHelmuth: Updated

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,118 @@
...
+Or should we refactor `component` somehow to remove `ReportStatus` from `component.TelemetrySettings`?
+
+## Desired Behavior for 1.0
+
+For each listed deviation, the solution for unblocking component 1.0 is:
+
+- `Fatal Error Reporting` :x:: we do not yet have a solution.
+- `No way to identify components that are not reporting status` :white_check_mark:: This can be implemented as a feature addition to component status reporting without blocking component 1.0
+- `Should component health reporting be an opt-in for component.Host implementations?` :x:: we do not yet have a solution.
+- `Component TelemetrySettings Requirements` :white_check_mark:: We have other structs, such as in pdata, that mark fields/methods as experimental. We can do the same for `component.TelemetrySettings.ReportStatus`.
```
mx-psi, line 95: We don't have any other fields/methods marked as experimental currently, we have committed to add that on `service` if we need to for profiling support (I don't think we will, but we'll see!).

My position is that we should not do this on Go modules used for component development if we have other approaches that we can take, and we seem to have those.

TylerHelmuth: I've switch this back to not resolved.

docs/rfcs/component-status-reporting.md
TylerHelmuth: It also looks like the status reporting parts of shared component only made it into Core's internal SharedComponent, not contrib, so only the otlpreceiver is impacted by SharedComponent.

TylerHelmuth: I've made more progress on the sharedcomponent problem.  It isn't perfect, but I think it is acceptable.  I hope to have a PR up by the end of the week for us to discuss, and then if it is acceptable I believe there are some incremental PRs possible to actually introduce the changes.

TylerHelmuth: Here is an attempt: https://github.com/open-telemetry/opentelemetry-collector/pull/10725.

TylerHelmuth: The initial feedback in https://github.com/open-telemetry/opentelemetry-collector/pull/10725 is promising, so I'm going to call this conversation resolved

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,118 @@
...
+Or should we refactor `component` somehow to remove `ReportStatus` from `component.TelemetrySettings`?
+
+## Desired Behavior for 1.0
+
+For each listed deviation, the solution for unblocking component 1.0 is:
+
+- `Fatal Error Reporting` :x:: we do not yet have a solution.
+- `No way to identify components that are not reporting status` :white_check_mark:: This can be implemented as a feature addition to component status reporting without blocking component 1.0
+- `Should component health reporting be an opt-in for component.Host implementations?` :x:: we do not yet have a solution.
+- `Component TelemetrySettings Requirements` :x:: We want to find a way to not include experimental features on stable structs. This means we still need to figure out how to handle `component.TelemetrySettings.ReportStatus`.
```
TylerHelmuth, line 95: The work in https://github.com/open-telemetry/opentelemetry-collector/pull/10725 will address these 2 points:

- `component.Host` may opt-in to component status reporting via the `consumerstatus.Reporter` interface. If a `component.Host` does not implement `consumerstatus.Reporter`, then calls within a component to report a status will end up being no-op.
- The  `component.TelemetrySettings.ReportStatus` will be removed, resulting in a `component` package with no public APIs relying on component status reporting.

docs/rfcs/component-status-reporting.md
TylerHelmuth: Thinking about this some more while refactoring component status features, I think the solution that provides the best separation of concerns is for the `component` package to provide no way for a component to stop the collector once the component has started. This is what the component's public API will represent once https://github.com/open-telemetry/opentelemetry-collector/pull/10725 is complete as we've already removed `Host.ReportFatalError`. 

With no public API in `component` to stop the collector after `Start`, the stance we'll be taking is: `Components cannot stop a running collector unless the `component.Host` implementation chooses to provide such a feature via a different Interface.`

This will end up being a breaking change for components that want to stop a running collector but do not want to use an unstable package like `componentstatus`.  Fortunately for us this breaking change was already handled when `Host.ReportFatalError` was removed, so any component that falls into this category is already in it.

Other components may choose to use `componentstatus.Reporter` and `componentstatus.NewFatalError`.  As part of the stabilization of `component` we do not need to commit to these interactions.

mx-psi (approved):
Looks good to me

djaglowski (approved):

codeboten (approved):
Thanks @TylerHelmuth

atoulme (approved):

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,119 @@
...
+
+Since the OpenTelemetry Collector is made up of pipelines with components, it needs a way for the components within those pipelines to emit information about their health. This information allows the collector service, or other interested software or people, to make decisions about how to proceed when something goes wrong. This document describes:
+
+1. The historical state of how components reported health
+2. The current state of how components report health
+3. The goals component health reporting should achieve
+4. Existing deviations from those goals
+5. Desired behavior for 1.0
+
+For context throughout this document, component defines a `component.Host` interface, which components may use to interact with the struct that is managing all the collector pipelines and the components. In this repository, our implementation of `component.Host` can be found on `service.Service`.
```
mwear, line 13:
```suggestion
For context throughout this document, component defines a `component.Host` interface, which components may use to interact with the struct that is managing all the collector pipelines and the components. In this repository, our implementation of `component.Host` can be found in `graph.Host`.
```

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Current State of Component Health Reporting
+
+See [Component Status Reporting](../component-status.md)
+
+## The Goals the Component Health Reporting Should Achieve
+
+The following are the goals, as of June 2024 and with Collector 1.0 looming, for a component health reporting system.
+
+1. A `component.Host` implementation, such as the `service.Service`, may report statuses Starting, Ok, Stopping and PermanentError on behalf of components.
```
mwear, line 45:
```suggestion
1. A `component.Host` implementation, such as `graph.Host`, may report statuses Starting, Ok, Stopping and PermanentError on behalf of components.
```

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,119 @@
...
+    - Additional status may be reported in the future
+2. Components may opt-in to reporting health status at runtime. Components must not be required to report health statuses themselves.
+    - The consumers of the health reporting system must be able to identify which components are and are not opting to report their own statuses.
+3. Component health reporting must be opt-in for collector users.  While the underlying components are always allowed to report their health via the system, the `component.Host` implementation, such as the `service.Service`, or any other listener may only take action when the user has configured the collector accordingly.
```
mwear, line 49:
```suggestion
3. Component health reporting must be opt-in for collector users.  While the underlying components are always allowed to report their health via the system, the `component.Host` implementation, such as `graph.Host`, or any other listener may only take action when the user has configured the collector accordingly.
```

docs/rfcs/component-status-reporting.md
```diff
@@ -0,0 +1,119 @@
...
+1. Accept this reality as an exception to Goal 2.
+2. Add back `component.Host.ReportFatalError`.
+3. Remove the ability for components to stop the collector be removing `FatalError`.
+
+### No way to identify components that are not reporting status
+Goal 2 states that consumers of component status reporting must be able to identify components in use that have not opted in to component status reporting. Our current implementation does not have this feature.
+
+### Should component health reporting be an opt-in for `component.Host` implementations?
+
+The current implementation of component status reporting does not add anything to `component.Host` to force a `component.Host` implementation, such as `service.Service`, to be compatible with component status reporting.  Instead, it adds `ReportStatus func(*StatusEvent)` to `component.TelemetrySettings` and things that instantiate components, such as `serivce.Service`, should, but are not required, to pass in a value for `ReportStatus`.
```
mwear, line 70:
```suggestion
The current implementation of component status reporting does not add anything to `component.Host` to force a `component.Host` implementation, such as `graph.Host`, to be compatible with component status reporting.  Instead, it adds `ReportStatus func(*StatusEvent)` to `component.TelemetrySettings` and things that instantiate components, such as `graph.Host`, should, but are not required, to pass in a value for `ReportStatus`.
```

mwear (approved):
LGTM. After the refactor, I believe the references to `service.Service` as a `component.Host` implementation are no longer accurate, and should be replaced by `graph.Host`. I've made some suggestions to that effect.

codeboten:
Reviewed by @mwear, merging now

PR #10417, opened 2024-06-17
[otelcoltest] Move otelcoltest to its own module

TylerHelmuth:
#### Description
As part of https://github.com/open-telemetry/opentelemetry-collector/issues/10290, move `otelcoltest` to its own module.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/10290


.chloggen/move-otelcoltest-2.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
mx-psi, line 25: API only change?
```suggestion
change_logs: [api]
```

mx-psi (commented):
I think there was a reason for this, but I don't remember why we need to change the path and we can't just leave it in its current folder. Could you explain the motivation for that?

TylerHelmuth:
@mx-psi Turns out that LoadConfig was used a lot in Contrib and it would be nice to keep using it.

We could create an internal package in Contrib and duplicate this logic, but this way felt better. 

If we want to do the Contrib solution, I think we should remove LoadConfig and LoadConfigAndValidate entirely.

mx-psi:
> @mx-psi Turns out that LoadConfig was used a lot in Contrib and it would be nice to keep using the version that depends on the providers/converters. We are trying to remove them from otelcol dependencies, but in this case the function is useful for our Contrib tests.
> 
> We could create an internal package in Contrib and duplicate this logic, but this way felt better.
> 
> If we want to do the Contrib solution, I think we should remove LoadConfig and LoadConfigAndValidate entirely.

@TylerHelmuth What I was asking is: why don't we add a `go.mod` and `go.sum` files to the current `otelcol/otelcoltest` folder instead of moving things to a top-level `otelcoltest`? Is there anything that prevents that?

TylerHelmuth:
@mx-psi `crosslink` doesn't like having the nested structure

mx-psi:
@TylerHelmuth We do have that structure on `pdata` and `pdata/pprofile`/`pdata/testdata` though :thinking:

TylerHelmuth:
@mx-psi you're right, I needed to get the module name nested correctly.

otelcol/otelcoltest/nop_factories.go
```diff
@@ -18,6 +18,8 @@ import (
 )
 
 // NopFactories returns a otelcol.Factories with all nop factories.
+//
+// Deprecated: [v0.103.0] Use go.opentelemetry.io/collector/otelcoltest instead
```
mx-psi, line 22: No longer needed?

TylerHelmuth: Done

.chloggen/move-otelcoltest-2.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: otelcoltest
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: move go.opentelemetry.io/collector/otelcol/otelcoltest to its own package, go.opentelemetry.io/collector/otelcoltest
```
mx-psi, line 10:
```suggestion
note: Split off go.opentelemetry.io/collector/otelcol/otelcoltest into its own module
```

.chloggen/move-otelcoltest.yaml
mx-psi, line 1: No longer needed?

mx-psi (commented):
Almost there!

mx-psi (approved):

Makefile
```diff
@@ -281,6 +281,7 @@ check-contrib:
 		-replace go.opentelemetry.io/collector/extension/zpagesextension=$(CURDIR)/extension/zpagesextension  \
 		-replace go.opentelemetry.io/collector/featuregate=$(CURDIR)/featuregate  \
 		-replace go.opentelemetry.io/collector/otelcol=$(CURDIR)/otelcol  \
+		-replace go.opentelemetry.io/collector/otelcol/otelcoltest=$(CURDIR)/otelcol/otelcoltest  \
```
codeboten, line 284: needs a line corresponding to this module at 340

otelcol/otelcoltest/config.go
```diff
@@ -26,8 +26,6 @@ func LoadConfigWithSettings(factories otelcol.Factories, set otelcol.ConfigProvi
 }
 
 // LoadConfig loads a config.Config  from file, and does NOT validate the configuration.
-//
-// Deprecated: [v0.103.0] use LoadConfigWithSettings instead
```
codeboten, line 30: is LoadConfigWithSettings deprecated in v0.104?

otelcol/otelcoltest/go.mod
```diff
@@ -0,0 +1,163 @@
+module go.opentelemetry.io/collector/otelcol/otelcoltest
+
+go 1.21.0
+
+toolchain go1.21.11
```
codeboten, line 5: we don't include the toolchain in any other modules than the generated code

codeboten, line 5:
```suggestion
```

codeboten (approved):

PR #10428, opened 2024-06-18
Add goreleaser check and attempt update

djaglowski:
https://github.com/open-telemetry/opentelemetry-collector/pull/10384 updated `goreleaser` to a new major version and this appears to have [broken the config](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/9565775767/job/26369463094). This problem only became apparent during the release process since the config is not used otherwise.

This PR adds a github action to run `goreleaser check` on all PRs, since changes to the config may not be caught until release is attempted.

I've also included what may be the only necessary change to the config, by [renaming `changelog.skip` to `changelog.disable`](https://github.com/goreleaser/goreleaser/commit/29f30b623ef8f0a19607afab22b3f3a2f9f68172). This passes `goreleaser check` when run locally.

The `monorepo` section of the config fails locally, but the feature appears to require the Pro version so I'm unclear if it will pass with the OSS version.

### DIFF
--- a/.github/workflows/check-goreleaser.yaml
+++ b/.github/workflows/check-goreleaser.yaml
@@ -1,0 +1,29 @@
+name: Check GoReleaser Config
+on:
+  push:
+    branches: [main]
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+*"
+  pull_request:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+        with:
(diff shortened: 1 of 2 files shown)


djaglowski:
The new check passed, [reporting a valid config here](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/9567048220/job/26373795326?pr=10428#step:4:32).

djaglowski:
I reverted the config to demonstrate that the [new action fails when the config is invalid](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/9567116522/job/26374030601?pr=10428#step:4:22), and that changing `skip` to `disable` in the config is correct.

mx-psi (approved):

PR #10436, opened 2024-06-18
[otelcol] Remove all default providers/converters

TylerHelmuth:
#### Description
Removed deprecated `NewCommand`.  Deprecates `NewCommandMustSetProvider`.  Adds `NewCommand` that requires at least one provider be set.

Removes usage of imported providers/converters from otelcol (but they still live in otelcoltest until https://github.com/open-telemetry/opentelemetry-collector/pull/10417 is merged.

#### Link to tracking issue
closes https://github.com/open-telemetry/opentelemetry-collector/issues/10290.

#### Testing
Unit tests


mx-psi (commented):
Can we file an issue on the Jaeger project and ping Yuri there so that they are aware of this change?

TylerHelmuth:
I'm planning to submit another PR to jaeger to handle the deprecation process, was waiting for the 0.103.0 release to finish in Contrib.

mx-psi (approved):
Looks like there are some tests to fix, but otherwise this LGTM. I would like to merge this after Jaeger is fixed

TylerHelmuth:
Jaeger is updated: https://github.com/jaegertracing/jaeger/pull/5657

cmd/builder/internal/builder/main_test.go
```diff
@@ -258,6 +258,8 @@ func TestGenerateAndCompile(t *testing.T) {
 			testCase: "Default Configuration Compilation",
 			cfgBuilder: func(t *testing.T) Config {
 				cfg := newTestConfig()
+				err := cfg.SetBackwardsCompatibility()
```
TylerHelmuth, line 261: These tests all pass in a list of providers, and therefore depend on those providers in the go mod list.  The builder templates only add them when `.Distribution.SupportsConfmapFactories` is true.  That value was not being set so all these tests should have been failing.  They were saved instead by the fact that otelcol depended on the providers.  The changes in this PR remove those dependencies when `otelcoltest` is not being used, so these tests started failing.  The fix was to make sure `.Distribution.SupportsConfmapFactories` is true, which is included in `cfg.SetBackwardsCompatibility()`.

evan-bradley: Nice catch, thanks for the explanation.

otelcol/command.go
```diff
@@ -52,13 +38,24 @@ func commandHelper(set CollectorSettings, enforceProviders bool) *cobra.Command
...
 	return rootCmd
 }
 
+// NewCommandMustSetProvider constructs a new cobra.Command using the given CollectorSettings.
+// Any URIs specified in CollectorSettings.ConfigProviderSettings.ResolverSettings.URIs
+// are considered defaults and will be overwritten by config flags passed as
+// command-line arguments to the executable.
+// At least one Provider must be supplied via CollectorSettings.ConfigProviderSettings.ResolverSettings.ProviderFactories.
+//
+// Deprecated: [v0.104.0] use NewCommand instead
```
evan-bradley, line 52: We should probably have a deprecation changelog entry for this, as inconsequential as it likely is.

otelcol/command.go
```diff
@@ -73,14 +70,8 @@ func updateSettingsUsingFlags(set *CollectorSettings, flags *flag.FlagSet, enfor
 		set.ConfigProviderSettings.ResolverSettings.DefaultScheme = "env"
 	}
 
-	// Provide a default set of providers and converters if none have been specified.
-	// TODO: Remove this after CollectorSettings.ConfigProvider is removed and instead
-	// do it in the builder.
 	if len(resolverSet.ProviderFactories) == 0 && len(resolverSet.ConverterFactories) == 0 {
```
evan-bradley, line 73:
```suggestion
	if len(resolverSet.ProviderFactories) == 0 {
```
I still think we should remove this check and have it handled by the resolver, but if we intend to keep it here, we should probably remove the converter check now that we don't need to look for a "zero-valued" factory config.

mx-psi:
@TylerHelmuth can you resolve the conflict?

TylerHelmuth:
@mx-psi conflict handled.

mx-psi:
@evan-bradley will merge this by Friday EU morning unless you comment by then!

evan-bradley:
Sorry, I was out last week, thanks for not holding this up on me. This looks good to me, the negative diff in the go.mod file is a win in my eyes.

mx-psi:
Agreed, I am pretty happy about the `go.mod` update :)

PR #10439, opened 2024-06-19
Docs - Update contrib release step 1

djaglowski:
When I performed this step yesterday, I found that it was necessary to update the builder configs first, otherwise `make update-otel` would fail. This PR just updates the order of operations to match what worked for me.


mx-psi (approved):
Thanks!

PR #10441, opened 2024-06-19
[chore] Fix comments and clarify docs on `config/configtls`

arminru:
#### Description
Point out that `config/configtls` also applies to HTTPS

#### Link to tracking issue
n/a

#### Testing
n/a

#### Documentation
n/a


config/configtls/README.md
```diff
@@ -11,8 +11,9 @@ Note that mutual TLS (mTLS) is also supported.
 By default, TLS is enabled:
 
 - `insecure` (default = false): whether to enable client transport security for
-  the exporter's gRPC connection. See
-  [grpc.WithInsecure()](https://godoc.org/google.golang.org/grpc#WithInsecure).
+  the exporter's HTTPs or gRPC connection. See
+  [grpc.WithInsecure()](https://godoc.org/google.golang.org/grpc#WithInsecure)
+  and [crypto/tls' InsecureSkipVerify config](https://godoc.org/crypto/tls#Config).
```
codeboten, line 16: InsecureSkipVerify is configurable via the option `insecure_skip_verify` listed below

codeboten (commented):
Thanks @arminru, just one comment

config/configtls/README.md
arminru: This part makes it sound as if `insecure` for HTTP would not disable TLS overall but just cert chain verification, i.e., have the same effect as `insecure_skip_verify`:
https://github.com/open-telemetry/opentelemetry-collector/blob/149ded2ad33df6341a00ae8bd107fcff00bff276/config/configtls/configtls.go#L91C1-L94
Would that need to be corrected?

arminru: From https://github.com/open-telemetry/opentelemetry-collector/blob/426e660499d7bd331f66bb3336daf1932b8a8e4a/config/configtls/configtls.go#L381 it looks like it would indeed disable TLS overall also for HTTP. I addressed this in aeda806.

For the readme here we could go with
```suggestion
  for gRPC.
```
and leave out a reference for HTTP, or we could also not mention the protocol at all in `configtls`.

codeboten (approved):

PR #10445, opened 2024-06-20
[confighttp] Add option to include query params in auth context

### LINKED ISSUE #4806: Provision to include request context(query params, headers) in AuthenticatorFunc
A facility to access request params such as query string in addition to already existing headers would be useful in the `AuthenticatorFunc`

Pass the queryParams along with headers in the map passed to the above func. Probably not the best approach.

**Additional context**
Authenticators such as apiToken validators would need to pick the token from either the header or query string. In such case, the current implementation wont be sufficient. 
We need a more generic way of handling this.

@jpkrohling

jpkrohling:
This PR adds a new option to the ServerConfig's Auth option, allowing users to specify a list of query parameters to add to the sources of auth data, in addition to HTTP headers. Instead of simply adding all parameters, which might be numeruous, we require users to specify which ones to include.

Auth extensions don't need to be changed, but should document which attributes they expect to find in the context and how to configure confighttp to accomplish that.

Fixes #4806

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -291,7 +291,7 @@ type ServerConfig struct {
 	CORS *CORSConfig `mapstructure:"cors"`
 
 	// Auth for this receiver
-	Auth *configauth.Authentication `mapstructure:"auth"`
+	Auth *AuthConfig `mapstructure:"auth"`
 
 	// MaxRequestBodySize sets the maximum request body size in bytes. Default: 20MiB.
 	MaxRequestBodySize int64 `mapstructure:"max_request_body_size"`
@@ -306,6 +306,14 @@ type ServerConfig struct {
 
 	// CompressionAlgorithms configures the list of compression algorithms the server can accept. Default: ["", "gzip", "zstd", "zlib", "snappy", "deflate"]
 	CompressionAlgorithms []string `mapstructure:"compression_algorithms"`
+}
+
+type AuthConfig struct {
+	// Auth for this receiver.
+	*configauth.Authentication `mapstructure:"-"`
+
+	// RequestParameters is a list of parameters that should be extracted from the request and added to the context.
+	RequestParameters []string `mapstructure:"request_params"`
 }
 
 // ToListener creates a net.Listener.
@@ -387,7 +395,7 @@ func (hss *ServerConfig) ToServer(_ context.Context, host component.Host, settin
 			return nil, err
 		}
 
-		handler = authInterceptor(handler, server)
+		handler = authInterceptor(handler, server, hss.Auth.RequestParameters)
 	}
 
 	if hss.CORS != nil && len(hss.CORS.AllowedOrigins) > 0 {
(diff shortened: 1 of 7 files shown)


config/confighttp/confighttp.go
```diff
@@ -289,6 +289,14 @@ type ServerConfig struct {
 	CompressionAlgorithms []string `mapstructure:"compression_algorithms"`
 }
 
+type AuthConfig struct {
+	// Auth for this receiver.
+	*configauth.Authentication `mapstructure:"-"`
+
+	// RequestParameters is a list of parameters that should be extracted from the request and added to the context.
+	RequestParameters []string `mapstructure:"request_params"`
```
TylerHelmuth, line 297: I may be late to the discussion, but should this be part of `configauth.Authentication` instead? Would `configgrpc` need something similar?

config/confighttp/confighttp.go
```diff
@@ -272,7 +272,7 @@ type ServerConfig struct {
 	CORS *CORSConfig `mapstructure:"cors"`
 
 	// Auth for this receiver
-	Auth *configauth.Authentication `mapstructure:"auth"`
+	Auth *AuthConfig `mapstructure:"auth"`
```
TylerHelmuth, line 275: This is a breaking change for the API but not the end user config right? We should add an API changelog.

config/confighttp/confighttp.go
jpkrohling: I think this is particular to HTTP: in gRPC, things are either metadata, or part of the message. The authenticator won't be able to (easily) get a parameter from the message without knowing the model used for that.

config/confighttp/confighttp.go
jpkrohling: New changelog added for the API.

config/confighttp/confighttp.go
```diff
@@ -448,9 +456,16 @@ type CORSConfig struct {
...
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx, err := server.Authenticate(r.Context(), r.Header)
+		sources := r.Header
+		query := r.URL.Query()
+		for _, param := range requestParams {
+			if val, ok := query[param]; ok {
+				sources[param] = val
+			}
+		}
+		ctx, err := server.Authenticate(r.Context(), sources)
```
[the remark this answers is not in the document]
jpkrohling: Great idea, done!

config/confighttp/confighttp.go
```diff
@@ -467,9 +475,16 @@ type CORSConfig struct {
...
 
-func authInterceptor(next http.Handler, server auth.Server) http.Handler {
+func authInterceptor(next http.Handler, server auth.Server, requestParams []string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx, err := server.Authenticate(r.Context(), r.Header)
+		sources := r.Header
+		query := r.URL.Query()
+		for _, param := range requestParams {
+			if val, ok := query[param]; ok {
+				sources[param] = val
```
axw, line 484: Perhaps the docs for RequestParameters should state that request parameters take precedence over HTTP headers with the same name?

axw (approved):
Nice! LGTM with one suggestion. This should be useful for RUM/browser instrumentation, to avoid CORS preflight requests.

mx-psi (approved):

TylerHelmuth (approved):

PR #10450, opened 2024-06-20
[otlpexporter] support validation for dns:// and dns:///

### LINKED ISSUE #10449: [configgrpc] endpoint validation causes failures in dns resolution
The validation that was added for `dns:///` makes it impossible for end users to use dns resolution as documented here: https://github.com/grpc/grpc/blob/master/doc/naming.md#detailed-design

The documentation for NewClient suggests that `dns:///` is correct, but that causes the `authority` to be empty, making it impossible for anyone using a custom authority to configure a custom authority

codeboten:
This allows users to continue using the recommended dns://authority/host:port notation when needing to specify a custom authority.
    
Fixes #10449

### DIFF
--- a/exporter/otlpexporter/config_test.go
+++ b/exporter/otlpexporter/config_test.go
@@ -134,6 +134,17 @@ func TestUnmarshalInvalidConfig(t *testing.T) {
 func TestValidDNSEndpoint(t *testing.T) {
 	factory := NewFactory()
 	cfg := factory.CreateDefaultConfig().(*Config)
-	cfg.Endpoint = "dns:///backend.example.com:4317"
+	cfg.Endpoint = "dns://backend.example.com:4317"
 	assert.NoError(t, cfg.Validate())
+}
+
+func TestSanitizeEndpoint(t *testing.T) {
+	factory := NewFactory()
+	cfg := factory.CreateDefaultConfig().(*Config)
+	cfg.Endpoint = "dns://backend.example.com:4317"
+	assert.Equal(t, "backend.example.com:4317", cfg.sanitizedEndpoint())
+	cfg.Endpoint = "dns:///backend.example.com:4317"
+	assert.Equal(t, "backend.example.com:4317", cfg.sanitizedEndpoint())
+	cfg.Endpoint = "dns:////backend.example.com:4317"
+	assert.Equal(t, "/backend.example.com:4317", cfg.sanitizedEndpoint())
 }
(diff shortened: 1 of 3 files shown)


exporter/otlpexporter/config_test.go
```diff
@@ -134,6 +134,17 @@ func TestUnmarshalInvalidConfig(t *testing.T) {
 func TestValidDNSEndpoint(t *testing.T) {
 	factory := NewFactory()
 	cfg := factory.CreateDefaultConfig().(*Config)
-	cfg.Endpoint = "dns:///backend.example.com:4317"
+	cfg.Endpoint = "dns://backend.example.com:4317"
```
TylerHelmuth, line 137: Is this a valid dns endpoint? I though it would need either `///` in a row or another `/` somewhere to separate the authority and the path

exporter/otlpexporter/config_test.go
```diff
@@ -134,6 +134,17 @@ func TestUnmarshalInvalidConfig(t *testing.T) {
...
 	assert.NoError(t, cfg.Validate())
 }
+
+func TestSanitizeEndpoint(t *testing.T) {
```
TylerHelmuth, line 141: Can you add a test case for `/` somewhere else in the host?

codeboten: done

exporter/otlpexporter/config_test.go
codeboten: it's a good point, according to the grpc docs, the following is valid:

`dns:[//authority/]host[:port] -- DNS (default)`

I don't know how much more validation we should be introducing here

TylerHelmuth: The new string looks invalid to me, I'm surprised an error is not returned. I'm interpreting the regex as "if `//` is present, there must be a third `/`". What am I missing

codeboten: let me add a test for this case. i agree that it should return an error

TylerHelmuth: Oh I see how this test works now.  Currently the validation only checks that there is a port and that it is a valid integer, nothing else.  We should add full dns endpoint validation, but since the current functionality breaks users ability to use authorities in their dns endpoint I think we should add that later and merge this now.

For the sake of having a proper endpoint:

```suggestion
	cfg.Endpoint = "dns://authority/backend.example.com:4317"
```

TylerHelmuth: created https://github.com/open-telemetry/opentelemetry-collector/issues/10488

mx-psi (approved):

evan-bradley (approved):

jpkrohling (approved):

PR #10477, opened 2024-06-28
[mdatagen] fix generated comp test for extensions and unused imports in templates

adrielp:
#### Description

When upgrading beyond `v0.101.0` of `mdatagen` I ran into issues upgrading when running `golanglint-ci` because of unused imports. Namely, in `component_tests` the `component` was removed from [this commit](https://github.com/open-telemetry/opentelemetry-collector/compare/v0.101.0...v0.102.0?diff=split&w=#diff-abf7cf477183f2c7e2f9425655bea3a2e2060eaba8fb6aa9c9a9558017c21f1e) which was not caught by tests (it appears we only test receiver components & not others). The logic that generates tests for extensions, where I caught this error, ends up with a `generated_component_test` that does not use `component` in the package at all. I would've thought the contrib repo would've detected this issue since they also use `mdatagen` and there are extensions there + golangci-lint. I haven't dug into the why it wasn't caught there yet. 

> This is just a quick fix for extensions. I suspect this issue may pop up in other components as time goes on. Long term is more tests for other components.

#### Testing
I ran this locally against my own repo where the issue originally showed & ensured the value would be templated out appropriately.

image

### DIFF
--- a/.chloggen/fix-component-tests.yaml
+++ b/.chloggen/fix-component-tests.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: mdatagen
+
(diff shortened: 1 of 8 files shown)


codeboten (approved):
Thanks for the fix @adrielp!

Issue #10368, opened 2024-06-07
[exporterhelper] Awkwardness due to API between queue sender and batch sender


carsonip:
**Is your feature request related to a problem? Please describe.**

Related to batch sender in #8122 

Currently, queue sender and batch sender work together fine, but the config `sending_queue.num_consumers` and batch sender `concurrencyLimit` make it a little awkward to use. The awkwardness comes from the fact that it is impossible for batch sender to get more requests than queue sender's num_consumers, and may be forced to export a smaller batch due to it.

I am also not sure if this presents a performance problem at scale, which means a lot of goroutines will be involved.

The batch sender concurrency check has also been prone to unfavorable goroutine scheduling, as reported in #9952 .

**Describe the solution you'd like**

Ideally, queue sender can keep sending to batch sender without an artificial limit, such that batching is only triggered by min_size_items and flush_timeout.

I don't have an idea how to implement this yet. Also, this may require a change in contract between exporterhelper parts.

jmacd:
@moh-osman3 and I studied the same problem -- we want to avoid use of a queue_sender but we want concurrency in the batcher. I reason that another user will come along asking not to use the queue sender, not to use the batch sender, and still want concurrency. Therefore, I propose we try to (and @moh-osman3 has been investigating how to) add a new sender stage purely for concurrency control.

In such a future, the existing queue_sender num_consumers I think should dictate how many workers perform the cpu and I/O intensive work of gathering items from the queue, not dictate how many export threads are used.

moh-osman3:
@dmitryax @carsonip This PR https://github.com/open-telemetry/opentelemetry-collector/pull/10478 shows an implementation of the concurrency_sender which will limit the number of goroutines used for export. This removes the concurrenyLimit from the batch_sender so that batch exports are only triggered by size/timeout and the concurrency is limited by the new sender independent of queue_sender.num_consumers. Any thoughts on this approach?

dmitryax:
@jmacd just wanted to clarify that the existing implementation is concurrent. This issue is about limiting concurrency

dmitryax:
I believe the solution should to change API interface between queue and batch from pushing queue->batcher to pulling queue<-batcher. So if queue and batcher are enabled, we treat batcher as queue consumers.  The number of concurrent batchers should be configurable, similar to queue consumers it'll represent the maximum amount of outgoing requests.

The workflow would be the following:
1. Pick a batcher/consumer from the pool, make it an active batcher
[...]

PR #10478, opened 2024-06-28
(exporterhelper):  Add concurrency sender to send batches concurrently

moh-osman3:
#### Description
This PR adds in a new component to the exporterhelper called the concurrency sender. This component is meant to limit the concurrency of RPC's independent from the queue sender's `num_consumers`. This functionality is desirable because the queue_sender does not apply backpressure, so the concurrency_sender can limit requests based on a configured concurrency limit. Additionally, the batch_sender can generate multiple batches from a request that are exported sequentially, so the concurrency_sender provides a way to export these batches in parallel.

This PR also changes the `requestSender` interface

```
type requestSender interface {
	component.Component
	send(context.Context, ...Request) error
	setNextSender(nextSender requestSender)
}
```

This allows us to optionally send multiple requests, which will allow batch producers to receive any errors associated with any of their exported batches.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/10368

#### Testing
Added unit tests and performed local testing to ensure concurrency works and errors are propagated back to batch producers


exporter/exporterhelper/common.go
```diff
@@ -86,6 +86,19 @@ func WithRetry(config configretry.BackOffConfig) Option {
...
 }
 
+// WithRetry overrides the default configretry.BackOffConfig for an exporter.
+// The default configretry.BackOffConfig is to disable retries.
+func WithConcurrency(config ConcurrencySettings) Option {
+	return func(o *baseExporter) error {
+		if !config.Enabled {
+			o.exportFailureMessage += " Try enabling retry_on_failure config option to retry on retryable errors."
+			return nil
+		}
```
jmacd, line 96: I think you can remove this, or you could add "Try enabling the concurrency config option to send more requests in parallel." but generally I dislike these very-wordy messages and think they tend to over-simplify the response a user will take.

jmacd (commented):
This approach looks :100: to me, thanks @moh-osman3.

exporter/exporterhelper/common.go
moh-osman3: Thanks for the review! Hmm changed the message, but I think I will leave it in for now to remain consistent with other exporterhelper options

exporter/exporterhelper/common.go
```diff
@@ -21,7 +21,7 @@ import (
 // requestSender is an abstraction of a sender for a request independent of the type of the data (traces, metrics, logs).
 type requestSender interface {
 	component.Component
-	send(context.Context, Request) error
+	send(context.Context, ...Request) error
```
dmitryax, line 24: Can you please elaborate on why this change is required? I cannot find a reason for it to support concurrency limit.

exporter/exporterhelper/batch_sender.go
```diff
@@ -176,35 +183,34 @@ func (bs *batchSender) sendMergeSplitBatch(ctx context.Context, req Request) err
...
 // sendMergeBatch sends the request to the batch and waits for the batch to be exported.
-func (bs *batchSender) sendMergeBatch(ctx context.Context, req Request) error {
+func (bs *batchSender) sendMergeBatch(ctx context.Context, reqs ...Request) error {
 	bs.mu.Lock()
 
+	var req Request
 	if bs.activeBatch.request != nil {
 		var err error
-		req, err = bs.mergeFunc(ctx, bs.activeBatch.request, req)
+		req, err = bs.mergeFunc(ctx, bs.activeBatch.request, reqs[0])
```
dmitryax, line 207: Why do we ignore other requests?

exporter/exporterhelper/common.go
```diff
@@ -86,6 +86,19 @@ func WithRetry(config configretry.BackOffConfig) Option {
 	}
 }
 
+// WithConcurrency overrides the default ConcurrencySettings for the exporter.
+// The default ConcurrencySettings is to enable concurrency with a limit of 10 goroutines.
+func WithConcurrency(config ConcurrencySettings) Option {
+	return func(o *baseExporter) error {
+		if !config.Enabled {
+			o.exportFailureMessage += " Try enabling the concurrency config option to send more requests in parallel."
```
dmitryax, line 94: This seems misleading. Not sure how enabling concurrency can help with export errors. There is no concurrency limit by default

exporter/exporterhelper/concurrency_sender.go
```diff
@@ -0,0 +1,89 @@
...
+	"golang.org/x/sync/semaphore"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter"
+)
+
+const defaultNumSenders = 10
+
+type ConcurrencySettings struct {
+	Enabled    bool `mapstructure:"enabled"`
```
dmitryax, line 22: This option seems excessive. It's the same as if concurrency is set to any number other than `0`.

We probably don't even need a configuration grouping here. Just one field in the exporter config similar to timeout, e.g.:

```
otlp:
  timeout: 10s
  concurrency: 100
  sending_queue:
   enabled: true
  retry_on_failure:
    enabled: true
```

exporter/exporterhelper/common.go
moh-osman3: Thanks for the review! The main issue comes from returning multiple errors back to a single batch producer. e.g. let's say one request enters the batch_sender and is split into 50 batches. We can't return from sendMergeSplitBatch until we collect a response from all 50 batch exports. In the current implementation we [sequentially](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/batch_sender.go#L190) export these batches waiting for the previous batch to return -- ideally these exports should be concurrent subject to a concurrency limit. To make this concurrent we want to avoid creating too many goroutines that exceeds the concurrency limit. This is reason we need to pass the full slice of requests to the concurrency_sender so it can limit the number of goroutines used for export

exporter/exporterhelper/batch_sender.go
moh-osman3: The input slice of Requests will have length 1 otherwise we need to use sendMergeSplitBatch, so it makes more sense to change this function signature back to `sendMergeBatch(ctx context.Context, req Request) error` and just pass the single request directly

exporter/exporterhelper/common.go
moh-osman3: Simplified the config so this error is now removed

exporter/exporterhelper/concurrency_sender.go
moh-osman3: Thanks, simplified my config as you suggested

exporter/exporterhelper/common.go
moh-osman3: Actually removed this error now since I removed config `Enabled` option

dmitryax:
The biggest problem with this approach is that we ignore the number of queue workers. It means that, typically, every batch would have to be stuck for the flush_timeout period before it can be sent further. That's why we had the concurrency limit built tied to it, so every time all the workers are stuck in one batch, we release them right away. And that behavior is gone with this PR.

exporter/exporterhelper/common.go
```diff
@@ -277,11 +288,7 @@ func newBaseExporter(set exporter.Settings, signal component.DataType, osf obsre
 
 	be.connectSenders()
 
-	if bs, ok := be.batchSender.(*batchSender); ok {
-		// If queue sender is enabled assign to the batch sender the same number of workers.
-		if qs, ok := be.queueSender.(*queueSender); ok {
-			bs.concurrencyLimit = int64(qs.numConsumers)
```
dmitryax, line 283: The biggest problem with this approach is that we ignore the number of queue workers. It means that, typically, every batch would have to be stuck for the flush_timeout period before it can be sent further. That's why we had the concurrency limit built tied to it, so every time all the workers are stuck in one batch, we release them right away. And that behavior is gone with this PR.

dmitryax: So, it doesn't resolve the problem stated in https://github.com/open-telemetry/opentelemetry-collector/issues/10368. Just adding an independent concurrency limit doesn't help for that

moh-osman3: @dmitryax Thanks for the feedback I've spent a bit of time thinking about options here so that the queue_sender consumers don't block when there are more items we can batch. I’m wondering if we can swap the order of the batch sender and the queue sender so that we are actually queueing batches instead of queuing the initial request that comes in? Or otherwise it seems that the queue sender could do some prebatching in the queue_sender (i.e. keep popping items from the queue until we have roughly a batch size to send to the batch_sender).

moh-osman3: Another possibility I explored was unblocking consumers by calling [consumeFunc](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/internal/queue/bounded_memory_queue.go#L52) in a goroutine so that the batch_sender isn't blocked for the flush timeout and more requests can be consumed. However this might not work for the persistent queue which needs to know when `consumeFunc` returns so it can mark items for deletion from the persistent queue.

moh-osman3: Any thoughts on these options?

dmitryax: A few problems with putting the queue before the batcher:
- The queue is what adds asynchronous behavior to exporters. If it's not enabled, the batcher is expected to concurrently block the incoming requests.  
- Persistent queue provides durability. A crashed collector is expected to not lose any data. If we put another memory queue in front of batching, it breaks that promise.

axw: > Another possibility I explored was unblocking consumers by calling [consumeFunc](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/internal/queue/bounded_memory_queue.go#L52) in a goroutine so that the batch_sender isn't blocked for the flush timeout and more requests can be consumed. However this might not work for the persistent queue which needs to know when consumeFunc returns so it can mark items for deletion from the persistent queue.

I've been coming to this conclusion. I'm imagining it would be something like the following:
 - batcher reads, but does not remove, records from the queue until it has a batch of min size
 - batcher sends the batch to the next consumer in the chain -- this could be the concurrency sender
 - on success, batcher removes the records from the queue

Perhaps there could be an option on the batcher to remove records from the queue optimistically, for users that prefer at most once vs. at least once delivery (default being at least once, for durability).

This is not trivial of course - it would mean a substantial change to the queue API.

PR #10495, opened 2024-07-01
[component] Make InstanceID immutable

### LINKED ISSUE #10494: [component] Make InstanceID immutable
While we are in the process of stabilizing the component module, we should make [InstanceID](https://github.com/open-telemetry/opentelemetry-collector/blob/a082f2e439e8f77a9a9503d54d8afea576f2d08c/component/component.go#L193-L198) immutable. Currently it is a struct with all fields exported. Here is a draft PR showing what the changes would look like: https://github.com/open-telemetry/opentelemetry-collector/pull/10495.

mwear:
#### Description
This PR makes component.InstanceID immutable. Previously it was a struct with all fields exported. Technically this is a breaking change, but the only thing using the InstanceID is the in-progress healthcheckv2extension.

#### Link to tracking issue
Fixes #10494

#### Testing
units

#### Documentation
code comments


component/component.go
```diff
@@ -190,9 +190,54 @@ func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
 	return f()
 }
 
-// InstanceID uniquely identifies a component instance
+// InstanceID uniquely identifies a component instance.
 type InstanceID struct {
-	ID          ID
-	Kind        Kind
-	PipelineIDs map[ID]struct{}
```
mx-psi, line 197: Potentially these are being used by other components that are not the healthcheck v2 extension (outside of contrib). Since it's quite easy, would you mind keeping these fields but deprecate them for one version?

component/component.go
```diff
@@ -190,9 +190,54 @@ func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
...
+	}
+	for _, pid := range pipelineIDs {
+		instanceID.pipelineIDs[pid] = struct{}{}
+	}
+	return &instanceID
+}
+
+// InstanceIDWithPipelines derives a new InstanceID from id with additional
+// pipelineIDs added to it.
+func InstanceIDWithPipelines(id *InstanceID, pipelineIDs ...ID) *InstanceID {
```
mx-psi, line 230: What do you think of having a `WithPipelines` method on `InstanceID` instead of this? Similar to the pattern used by `zap.Logger` in the [`With` method](https://pkg.go.dev/go.uber.org/zap#Logger.With)

component/component.go
mx-psi: Ah, right :( 

Doing a quick Github search, we do have a bunch of external users of this https://github.com/search?q=%22component.InstanceID%22+language%3Ago+NOT+is%3Afork+NOT+repo%3Aopen-telemetry%2Fopentelemetry-collector+NOT+repo%3Aopen-telemetry%2Fopentelemetry-collector-contrib&type=code

I would be fine with this if we notify grafana/agent and grafana/alloy of this, they seem to be the biggest users

mx-psi (approved):
This LGTM. I am happy to merge this if we have a link to an issue on grafana/alloy and grafana/agent about this change

mwear:
I'll get on those issues and bring this out of draft and rebase.

mwear:
Issues have been opened in the respective grafana repos:
- https://github.com/grafana/agent/issues/6984
- https://github.com/grafana/alloy/issues/1323

component/component.go
```diff
@@ -190,9 +190,54 @@ func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
...
+
+// Kind returns the component Kind associated with this instance.
+func (id *InstanceID) Kind() Kind {
+	return id.kind
+}
+
+// PipelineIDs returns a set of PipelineIDs associated with this instance.
+func (id *InstanceID) PipelineIDs() map[ID]struct{} {
+	return id.pipelineIDs
+}
```
axw, line 226: Maybe it's over the top, but did you consider returning a copy of the map, to ensure the ID is truly immutable? Otherwise a caller can still modify the object.

I also wonder if it would be sensible to store the pipeline IDs in a way that makes the InstanceID type comparable. Again it doesn't seem to matter at the moment, since InstanceID is passed around by reference everywhere. Just thinking it could be good for future proofing.

axw (commented):
Just a couple of drive-by comments, feel free to ignore :)

component/component.go
mwear: I had considered returning a copy of the map, but I wasn't sure it was worth the performance hit. These will be read often. This is an easy change to make though if anyone has strong opinions. Do you have a suggestion for how we might store PipelineIDs in a way that will make InstanceID comparable? If so, I'd be open to considering it. Bonus if it solves the true immutability problem as well.

axw: > I had considered returning a copy of the map, but I wasn't sure it was worth the performance hit. These will be read often. This is an easy change to make though if anyone has strong opinions.

Fair enough. I don't have a very strong opinion on this, just stumbled on the PR and thought it would be worth checking.

> Do you have a suggestion for how we might store PipelineIDs in a way that will make InstanceID comparable? If so, I'd be open to considering it. Bonus if it solves the true immutability problem as well.

One option would be to:
 - marshal the IDs, with some separator (space?) that's not permitted in type or name, and store as a string
 - change `PipelineIDs` to an iterator-type function like `pcommon.Map.Range` which incrementally unmarshals the IDs
 
The range function could be written in a way that works with Go 1.23's upcoming iterator functions: https://pkg.go.dev/iter#hdr-Iterators

mx-psi: Both of these properties seem to me like things that could be added in the future in a backwards compatible way. If anything, we can document now that this is not comparable and that the map should not be modified, and leave adding this extra guarantees for a future PR.

axw: Seems fine to me. Having `map[ID]struct{}` in the exposed API doesn't seem ideal, and making it an iterator would be breaking again, but I don't think it's worth blocking over.

mx-psi: > making it an iterator would be breaking again, but I don't think it's worth blocking over.

I don't feel like it's incompatible to add a new method for the iterator if we ensure the data is presented in the same way. I'll wait for Matt to give his opinion as well :)

mwear: @axw, I refactored the InstanceID to be comparable per your suggestions. Take a look and see if this is what you had in mind. Let me know if it needs any further work. I chose the second (or third) best name for the iterator to leave the door open to rename it (to `AllPipelineIDs` or just `PipelineIDs`) when we can use the iterator functions coming in Go 1.23.

component/component.go
```diff
@@ -190,9 +193,77 @@ func (f CreateDefaultConfigFunc) CreateDefaultConfig() Config {
 	return f()
 }
 
-// InstanceID uniquely identifies a component instance
+// pipelineDelim is the delimeter for internal representation of pipeline
+// component IDs.
+const pipelineDelim = byte(0x20)
```
axw, line 198: Unfortunately it turns out that IDs can contain a space :(

I mean I can define a receiver like this:

```yaml
receivers:
  "otlp/abc def":
    protocols:
      grpc:
        endpoint: localhost:4317
      http:
        endpoint: localhost:4318
```

Even a null byte in the name is accepted.

Seems a little lax to me. Perhaps revert to a map for now, but keep the method change, and we could follow up on making the type comparable later after finding a way to safely encode the IDs? Could either use size prefixing (non-breaking), or add some restrictions to IDs (breaking, but maybe a good idea anyway).

axw (commented):
@mwear looks great, and exactly what I meant, but I just tested locally and found that IDs can have spaces in them. I'd be inclined to revert to using a map internally, and follow up on making it comparable later as @mx-psi suggested. Sorry for the diversion.

component/component.go
mx-psi: We already introduced restrictions on the part before the `/` and had no user reports of it breaking them, so I would be in favor of doing the same for the name part (after the `/`)

mwear: I can take a look at adding the same restrictions to the name as we currently have on the type. If it's not terrible, maybe we can make both changes? If it turns out to be more work than expected, we can keep the iterator, and use a map internally, then make InstanceID comparable when we have restrictions on the name.

mx-psi: Yeah that makes sense to me

mwear: I opened #10674 to handle restrictions to the name portion of a component.ID. The needs were slightly different so I introduced a different type for the name field.

mx-psi:
Can this move forward now?

mwear:
> Can this move forward now that #10674 is merged?

This can move forward, but now that #10730 has merged, I'm wondering if should update componentstatus.InstanceID instead and what the order of operations should be given #10777.

mx-psi:
Good point, we can wait for #10777

mx-psi:
This is now unblocked since I merged #10777

component/componentstatus/instance.go
```diff
@@ -3,14 +3,89 @@
...
+	"strings"
+
+	"go.opentelemetry.io/collector/component"
+)
 
 // InstanceID uniquely identifies a component instance
 //
 // TODO: consider moving this struct to a new package/module like `extension/statuswatcher`
 // https://github.com/open-telemetry/opentelemetry-collector/issues/10764
+// pipelineDelim is the delimeter for internal representation of pipeline
```
atoulme, line 18:
```suggestion
// pipelineDelim is the delimiter for internal representation of pipeline
```

atoulme (approved):
Changes look good to me, one nit. I didn't check why CI is failing on contrib.

mwear:
> Changes look good to me, one nit. I didn't check why CI is failing on contrib.

This is expected to temporarily break contrib. It's a breaking change to the newly introduced componentstatus.InstanceID. This PR originally targeted component.InstanceID, put was put on hold to accommodate the component status refactor. I'll fix contrib when this merges.

.chloggen/immutable-instance-id.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
TylerHelmuth, line 4:
```suggestion
change_type: 'breaking'
```

component/componentstatus/instance.go
```diff
@@ -3,14 +3,89 @@
...
+	"slices"
+	"sort"
+	"strings"
+
+	"go.opentelemetry.io/collector/component"
+)
 
 // InstanceID uniquely identifies a component instance
 //
 // TODO: consider moving this struct to a new package/module like `extension/statuswatcher`
```
TylerHelmuth, line 16: This TODO needs moved back down to InstanceID (or removed, I think it might be a bad copy and paste from `StatusWatcher`.

.chloggen/immutable-instance-id.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
TylerHelmuth, line 25:
```suggestion
change_logs: [api]
```

TylerHelmuth (approved):

TylerHelmuth:
@mwear please take a look at the failing lint

mx-psi (approved):

songy23 (approved):

PR #10508, opened 2024-07-02
[confmap] Set confmap.unifyEnvVarExpansion feature gate to stable

TylerHelmuth:
#### Description

This PR promotes the `confmap.unifyEnvVarExpansion` feature gate to stable and sets a `ToVersion` of `v0.106.0`, anticipating that the gate be completely removed in that version.

We should weigh if switching the Stable should be done in `v0.105.0` or if it needs more time in `Beta` to give users more time to switch.  Delaying promotion to `Stable` delays confmap 1.0.

If we merge this we need to commit to merging https://github.com/open-telemetry/opentelemetry-collector/pull/10510 in the same release.

#### Link to tracking issue


confmap/converter/expandconverter/expand_test.go
TylerHelmuth, line 1: Make the feature gate stable negates these tests, which set the gate to false.  Since the module is being deprecated, and will be removed shortly after, removing the tests doesn't feel like a big deal.

TylerHelmuth:
Let me know if you'd rather this PR and https://github.com/open-telemetry/opentelemetry-collector/pull/10510 be merged into 1.  If we needed to roll back for some reason 1 PR would be easier.

mx-psi:
I think we can wait at least until #10554 has been out for one release

ben-childs-docusign:
Hi Folks - I just logged this issue it seems like there is still an issue with configurations that use "$$SomeString" syntax to represent the string literal "$SomeString" with this feature flag enabled it is still blocked on startup with version 105.

https://github.com/open-telemetry/opentelemetry-collector/issues/10713

mx-psi:
@ben-childs-docusign Thanks for the report. Have you tried v0.106.1? The fix for that issue was released on v0.106.0

ben-childs-docusign:
Hi @mx-psi yes we deployed the patch and it is working. thanks for the fix!

confmap/resolver.go
```diff
@@ -171,7 +171,11 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 		if err != nil {
 			return nil, err
 		}
-		cfgMap[k] = val
+		if v, ok := val.(string); ok {
```
dmitryax, line 174: This can also be `expandedValue`, `[]any`, and `map[string]any`. We need to handle it a  function calling itself for each of the cases

TylerHelmuth: Updated to handle new types and added more test cases to the e2e expansion tests.

confmap/resolver.go
```diff
@@ -195,6 +195,31 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 	return retMap, nil
 }
 
+func escapeDollarSign(val any) any {
+	switch v := val.(type) {
+	case string:
+		return strings.ReplaceAll(v, "$$", "$")
+	case expandedValue:
+		v.Original = strings.ReplaceAll(v.Original, "$$", "$")
```
dmitryax, line 203: @mx-psi do we need this escaping here? Isn't the Original field supposed to keep the string as is?

mx-psi (approved):

confmap/resolver.go
mx-psi: @dmitryax Original is meant to keep the string representation. I think there are situations where an user would expect expansion to happen and situations where they wouldn't. 

#### Situation where an user expects expansion (and therefore escaping) to happen:

This is a bit of a contrived example but consider the following env vars and files:
- `URL_WITHOUT_PROTOCOL` set to `//example.com/api/v1/traces`
- `HTTP_URL` set to `http:${env:URL_WITHOUT_PROTOCOL}`
- a config file that sets the `otlp::endpoint` to `HTTP_URL`

In this situation, I would expect the OTLP endpoint to get `http://example.com/api/v1/traces`. 

Before #10618, this would fail with the `env` provider, since the contents of `HTTP_URL` would be parsed as a `map[string]any` with a single `http` key.

#### Situation where they do not expect expansion (or escaping) to happen

A password, that could have arbitrary contents

----

I don't know what to make out of this, we could have a third path so if a field is a `configopaque.String` we use the raw literal string (that's definitely doable with a bit of extra work and plumbing). I don't know if my imagination is good enough to cover all cases

TylerHelmuth: The tests still pass without the `OriginalValue` being escaped. Is there a scenario I could add to the e2e tests that would fail if `OriginalValue` isn't escape (or an inverse test if it should definitely not be escaped)

confmap/internal/e2e/expand_test.go
```diff
@@ -5,90 +5,99 @@ package e2etest
...
-		{
-			name:   "env",
-			scheme: "env",
-		},
-	}
-
+func Test_EscapedEnvVars_NoDefaultScheme(t *testing.T) {
 	const expandedValue = "some expanded value"
 	t.Setenv("ENV_VALUE", expandedValue)
+	t.Setenv("ENV_LIST", "['$$ESCAPE_ME','$$ESCAPE_ME']")
```
dmitryax, line 23: Given that $$ENV isn't supported anymore, it's better to add `$${}` here I think?
```suggestion
	t.Setenv("ENV_LIST", "['$$ESCAPE_ME','$${ESCAPE_ME}','${ENV_VALUE}']")
```

dmitryax (approved):

confmap/internal/e2e/expand_test.go
TylerHelmuth: Good idea, added

.chloggen/unify-env-var-gate-stable.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Set the `confmap.unifyEnvVarExpansion` feature gate to Stable. Expansion of `$FOO` env vars is no longer supporter.  Use `${FOO}` or `${env:FOO}` instead.
```
evan-bradley, line 10:
```suggestion
note: Set the `confmap.unifyEnvVarExpansion` feature gate to Stable. Expansion of `$FOO` env vars is no longer supported.  Use `${FOO}` or `${env:FOO}` instead.
```

confmap/converter/expandconverter/go.mod
```diff
@@ -3,24 +3,21 @@ module go.opentelemetry.io/collector/confmap/converter/expandconverter
 go 1.21.0
 
 require (
-	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/collector/confmap v0.106.1
-	go.opentelemetry.io/collector/featuregate v1.12.0
+	go.opentelemetry.io/collector/confmap v0.104.0
```
evan-bradley, line 6:
```suggestion
	go.opentelemetry.io/collector/confmap v0.106.1
```

confmap/internal/e2e/expand_test.go
```diff
@@ -5,90 +5,99 @@ package e2etest
...
+	m := cfgMap.ToStringMap()
+	assert.Equal(t, expectedMap, m)
+}
 
-			// Test that expanded configs are the same with the simple config with no env vars.
-			cfgMap, err := resolver.Resolve(context.Background())
-			require.NoError(t, err)
-			m := cfgMap.ToStringMap()
-			assert.Equal(t, expectedMap, m)
+// Test_EscapedEnvVars tests that the resolver supports escaped env vars working together with expand converter.
```
evan-bradley, line 62:
```suggestion
```
I think the function name should cover this, especially now that the expand converter is no longer involved.

confmap/internal/e2e/expand_test.go
```diff
@@ -5,90 +5,99 @@ package e2etest
...
 	"github.com/stretchr/testify/require"
 
 	"go.opentelemetry.io/collector/confmap"
-	"go.opentelemetry.io/collector/confmap/converter/expandconverter"
 	"go.opentelemetry.io/collector/confmap/provider/envprovider"
 	"go.opentelemetry.io/collector/confmap/provider/fileprovider"
-	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
 )
 
 // Test_EscapedEnvVars tests that the resolver supports escaped env vars working together with expand converter.
```
evan-bradley, line 19:
```suggestion
```

evan-bradley (approved):

mx-psi (approved):

mx-psi:
I am going to wait to merge this until #10510 is approved

internal/globalgates/globalgates.go
```diff
@@ -6,8 +6,9 @@ package globalgates // import "go.opentelemetry.io/collector/internal/globalgate
 import "go.opentelemetry.io/collector/featuregate"
 
 var UseUnifiedEnvVarExpansionRules = featuregate.GlobalRegistry().MustRegister("confmap.unifyEnvVarExpansion",
-	featuregate.StageBeta,
+	featuregate.StageStable,
 	featuregate.WithRegisterFromVersion("v0.103.0"),
+	featuregate.WithRegisterToVersion("v0.106.0"),
```
TylerHelmuth, line 11:
```suggestion
	featuregate.WithRegisterToVersion("v0.109.0"),
```

PR #10510, opened 2024-07-02
[confmap] Deprecate expandconverter

TylerHelmuth:
#### Description
This PR deprecates `expandconverter` and removes its use from `otelcoltest.LoadConfig` and OCB.  

This cannot be merged until the `confmap.unifyEnvVarExpansion` [feature gate is made stable](https://github.com/open-telemetry/opentelemetry-collector/pull/10508).

#### Link to tracking issue
closes https://github.com/open-telemetry/opentelemetry-collector/issues/10161
closes https://github.com/open-telemetry/opentelemetry-collector/issues/7111
closes https://github.com/open-telemetry/opentelemetry-collector/issues/8215


atoulme (approved):

dmitryax:
The same comment here as in https://github.com/open-telemetry/opentelemetry-collector/pull/10723. We need to move the `$$` escaping from the converter to the resolver and ensure ${FOO} expansion without the converter.

TylerHelmuth:
@dmitryax I'll handle that work as part of https://github.com/open-telemetry/opentelemetry-collector/pull/10508

mx-psi (approved):

songy23 (approved):

confmap/converter/expandconverter/expand.go
```diff
@@ -25,6 +25,8 @@ type converter struct {
 
 // NewFactory returns a factory for a  confmap.Converter,
 // which expands all environment variables for a given confmap.Conf.
+//
+// Deprecated: [v0.105.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```
codeboten, line 29:
```suggestion
// Deprecated: [v0.107.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```

confmap/converter/expandconverter/expand.go
```diff
@@ -36,6 +38,7 @@ func newConverter(set confmap.ConverterSettings) confmap.Converter {
 	}
 }
 
+// Deprecated: [v0.105.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```
codeboten, line 41:
```suggestion
// Deprecated: [v0.107.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```

confmap/converter/expandconverter/go.mod
```diff
@@ -1,26 +1,24 @@
+// Deprecated: [v0.105.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```
codeboten, line 1:
```suggestion
// Deprecated: [v0.107.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```

confmap/converter/expandconverter/expand_test.go
dmitryax, line 1: Why tests have to be removed now? Let's keep the code covered even if it's deprecated

TylerHelmuth: Added the tests back.  This required updating the expand converter code to no longer warn based on the env var (which will soon no longer be set to false).  This essentially leaves the expand converter back how it was before we started this transition, but now marked deprecated.  During this deprecation period someone could use the converter to do bash style syntax expansion if they desired.

dmitryax: SGTM

confmap/converter/expandconverter/expand.go
```diff
@@ -25,6 +23,8 @@ type converter struct {
 
 // NewFactory returns a factory for a  confmap.Converter,
 // which expands all environment variables for a given confmap.Conf.
+//
+// Deprecated: [v0.107.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```
dmitryax, line 27: I think we should also say that `envprovider` should be used instead.

confmap/converter/expandconverter/expand.go
```diff
@@ -90,19 +91,6 @@ func (c converter) expandEnv(s string) (string, error) {
 			return "$"
```
dmitryax, line 91: Also, we need to mention that using `expanconverter` will cause double escaping `$$$$`->`$` or we can change this line to return `$$`. This comment should probably be addressed in https://github.com/open-telemetry/opentelemetry-collector/pull/10508

TylerHelmuth: Good point. I'll add it here because it is a result of using the now deprecated expandconverter.

TylerHelmuth:
The [failed test](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10289487381/job/28477383807?pr=10510) will be handled by https://github.com/open-telemetry/opentelemetry-collector/pull/10508.

confmap/converter/expandconverter/expand.go
TylerHelmuth: Updated

confmap/converter/expandconverter/expand.go
TylerHelmuth: Updated

confmap/converter/expandconverter/expand.go
codeboten: Maybe i'm missing something, but i dont see `envprovider` anywhere

dmitryax: Neither do I. @TylerHelmuth did you push your changes?

TylerHelmuth: It does in fact help when the changes are pushed

confmap/converter/expandconverter/go.mod
```diff
@@ -1,12 +1,12 @@
+// Deprecated: [v0.107.0] BASH-style env var expansion is deprecated. Use the `envprovider` instead to expand `${FOO}` and `${env:FOO}`.
+// Using the expandconverter will cause double escaping, so `$$$$` -> `$` instead of `$$`.                                                                                                           //// Using the expandconverter will cause double escaping, so `$$$$` -> `$` instead of `$$`.
```
dmitryax, line 2:
```suggestion
// Using the expandconverter will cause double escaping, so `$$$$` -> `$` instead of `$$`.
```

confmap/converter/expandconverter/expand.go
dmitryax: Thinking more about it, maybe we should change it to `return "$$"`. I don't see any problems with that, and we keep the converter functional as before. WDYT?

codeboten (approved):

confmap/converter/expandconverter/expand.go
TylerHelmuth: That assumes you're using the converter as part of our `confmap.Resolver`.  I'll update the comment to clarify the double escaping only happens when used with `confmap.Resolver`.

dmitryax: Is there a way to use it without the resolver?

dmitryax (approved):

dmitryax:
Merged https://github.com/open-telemetry/opentelemetry-collector/pull/10508. Please rebase

confmap/converter/expandconverter/expand.go
TylerHelmuth: It is an independent package so technically anyone could pull it in and do something with it. That is, granted, an unlikely scenario.

dmitryax: Ok, let's keep it as is 👍

evan-bradley (approved):

.chloggen/deprecate-expandconverter-3.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api]
```
evan-bradley, line 25:
```suggestion
change_logs: [user]
```
I'm tempted to say that we should include this in the user-facing changelog since we encourage users to build their own Collector binaries.

TylerHelmuth: I disagree.  The previous PR handled the user-facing [changelog](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.chloggen/unify-env-var-gate-stable.yaml).  This PR (which could have be in https://github.com/open-telemetry/opentelemetry-collector/pull/10508 and was only separated for size), removes the expandconverter from the ocb's generated main.go, but end users don't need to worry about that.

dmitryax: We need to define the end user here :) I'm ok with any. We struggle on the same Q for madatagen "users".

dmitryax: Potentially, we can have 2 move changelogs: `ocb` and `mdatagen`

TylerHelmuth: I believe `user` denotes those who run the collector and `api` denotes those who import libraries into their code.

dmitryax: I'm merging as is for now. Feel free to continue the discussion and we can change this line if needed

Erog38:
Hey, bumping this because removing this expansion seems to have broken the capability for YAML to have predefined chunks which can be re-used across multi-organization support teams. I'll make an issue and try to repro, but the tl;dr is that we used to do something like this:


```yaml
...
filelog:
  <<: ${internal:filelog-defaults}
  include:
  - some additional include
```

This use to work as a way to allow users to merge predefined configuration with their own overrides and no longer functions. I'm fairly confident it's due to this removal. Is there any hope to address this issue or is it just an unsupported feature?

mx-psi:
@Erog38 Please file an issue with a repro and we can take a look :)

PR #10521, opened 2024-07-02
[chore] bump go 1.21.12

codeboten:
Removed the toolchain directive in pprof/go.mod as it's unnecessary

### DIFF
--- a/pdata/pprofile/go.mod
+++ b/pdata/pprofile/go.mod
@@ -2,8 +2,6 @@ module go.opentelemetry.io/collector/pdata/pprofile
 
 go 1.21.0
 
-toolchain go1.21.11
-
 require (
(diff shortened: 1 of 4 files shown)


pdata/pprofile/go.mod
```diff
@@ -2,8 +2,6 @@ module go.opentelemetry.io/collector/pdata/pprofile
 
 go 1.21.0
 
-toolchain go1.21.11
```
songy23, line 5: Shouldn't this be `toolchain go1.21.12` rather than remove?

codeboten: we don't have toolchain in any modules other than the ones generated by the builder

songy23 (approved):

PR #10524, opened 2024-07-03
Move connector into internal package, in preparation for profiles

dmathieu:
#### Description

This splits the connector package, so the APIs are in an internal package, and redefined publicly for logs/metrics/traces.
In preparation for adding profiles to the package.

#### Link to tracking issue
See https://github.com/open-telemetry/opentelemetry-collector/pull/10375G


dmathieu:
This technically doesn't change any public API. So I suppose it doesn't need a changelog entry.

evan-bradley (approved):
Tests all pass and the organization looks good to me.

connector/connector.go
```diff
@@ -61,383 +52,100 @@ type Metrics interface {
...
-	MetricsToMetricsStability() component.StabilityLevel
-	MetricsToLogsStability() component.StabilityLevel
-
-	LogsToTracesStability() component.StabilityLevel
-	LogsToMetricsStability() component.StabilityLevel
-	LogsToLogsStability() component.StabilityLevel
-
-	unexportedFactoryFunc()
-}
+type Factory = internal.Internal
```
codeboten, line 69: Not blocking as this is internal but why is this not called `internal.Factory`?

codeboten (approved):

connector/builder.go
```diff
@@ -0,0 +1,216 @@
...
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+)
+
+// Builder processor is a helper struct that given a set of Configs and Factories helps with creating processors.
```
codeboten, line 16:
```suggestion
// Builder is a helper struct that given a set of Configs and Factories helps with creating connectors.
```

connector/connector.go
dmathieu: I followed what @mx-psi did in this PR: https://github.com/open-telemetry/opentelemetry-collector/pull/10253/files#diff-56869497b872826bc1bd89731571d212be40aa7fe7252864a80317e6413234e3R126

Since there is a private `factory` struct, I guess it makes sense to use an entirely different name to avoid mixing them.

mx-psi: I am fine with `internal.Factory`, there was no real reason for using `internal.Internal`

dmathieu: I have renamed `internal.Internal` to `internal.Factory`.

codeboten (approved):

PR #10525, opened 2024-07-03
Add componentprofiles package

dmathieu:
#### Description

This adds the componentprofiles package, for handline the profiles data type.

#### Link to tracking issue
See https://github.com/open-telemetry/opentelemetry-collector/pull/10375

cc @mx-psi


component/componentprofiles/config.go
```diff
@@ -0,0 +1,15 @@
+// Copyright The OpenTelemetry Authors
```
mx-psi, line 1: I suppose you will add the interface in a separate PR?

mx-psi (approved):

component/componentprofiles/config.go
dmathieu: We discussed this privately. This was a red herring, as the component package has no interface to implement.

PR #10529, opened 2024-07-03
[internal/localhostgate] Correctly log message about local host gate

mx-psi:
#### Description

Fixes log intended to be logged when the feature gate is enabled, not disabled.

#### Link to tracking issue

Relates to #8510, updates #10352

### DIFF
--- a/internal/localhostgate/featuregate.go
+++ b/internal/localhostgate/featuregate.go
@@ -61,7 +61,7 @@ func EndpointForPort(port int) string {
 func LogAboutUseLocalHostAsDefault(logger *zap.Logger) {
 	if UseLocalHostAsDefaultHostfeatureGate.IsEnabled() {
 		logger.Info(
-			"The default endpoints for all servers in components have changed to use localhost instead of 0.0.0.0. Use the feature gate to temporarily revert to the previous default.",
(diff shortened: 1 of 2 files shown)


internal/localhostgate/featuregate.go
```diff
@@ -59,7 +59,7 @@ func EndpointForPort(port int) string {
 
 // LogAboutUseLocalHostAsDefault logs about the upcoming change from 0.0.0.0 to localhost on server-like components.
 func LogAboutUseLocalHostAsDefault(logger *zap.Logger) {
-	if !UseLocalHostAsDefaultHostfeatureGate.IsEnabled() {
+	if UseLocalHostAsDefaultHostfeatureGate.IsEnabled() {
 		logger.Info(
 			"The default endpoints for all servers in components have changed to use localhost instead of 0.0.0.0. Use the feature gate to temporarily revert to the previous default.",
```
songy23, line 64:
```suggestion
			"The default endpoints for all servers in components have changed to use localhost instead of 0.0.0.0. Disable the feature gate to temporarily revert to the previous default.",
```

if I read it correctly

songy23 (approved):

codeboten (approved):

PR #10530, opened 2024-07-03
Move receiver into an internal package, in preparation for profiles

dmathieu:
#### Description

This splits the receiver package, so the APIs are in an internal package, and redefined publicly for logs/metrics/traces.
In preparation for adding profiles to the package.

#### Link to tracking issue
See https://github.com/open-telemetry/opentelemetry-collector/pull/10375

cc @mx-psi


receiver/receiver.go
```diff
@@ -23,201 +20,66 @@ var (
...
-	// If the receiver type does not support the data type or if the config is not valid
-	// an error will be returned instead. `nextConsumer` is never nil.
-	CreateLogsReceiver(ctx context.Context, set Settings, cfg component.Config, nextConsumer consumer.Logs) (Logs, error)
-
-	// LogsReceiverStability gets the stability level of the LogsReceiver.
-	LogsReceiverStability() component.StabilityLevel
-
-	unexportedFactoryFunc()
-}
+type Factory = internal.Internal
```
dmitryax, line 51: Why not `internal.Factory`?

receiver/receiver.go
```diff
@@ -23,201 +20,66 @@ var (
...
 func WithLogs(createLogsReceiver CreateLogsFunc, sl component.StabilityLevel) FactoryOption {
-	return factoryOptionFunc(func(o *factory) {
-		o.logsStabilityLevel = sl
-		o.CreateLogsFunc = createLogsReceiver
-	})
+	return internal.WithLogs(createLogsReceiver, sl)
 }
 
 // NewFactory returns a Factory.
 func NewFactory(cfgType component.Type, createDefaultConfig component.CreateDefaultConfigFunc, options ...FactoryOption) Factory {
```
dmitryax, line 81: Can be 
```suggestion
var NewFactory = internal.NewFactory
```

receiver/builder.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+
+	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+)
+
+// Builder receiver is a helper struct that given a set of Configs and Factories helps with creating receivers.
+type Builder struct {
```
dmitryax, line 17: Curious why aren't we doing it for the `Builder`?

receiver/receiver.go
dmathieu: I followed what @mx-psi did in this PR: https://github.com/open-telemetry/opentelemetry-collector/pull/10253/files#diff-56869497b872826bc1bd89731571d212be40aa7fe7252864a80317e6413234e3R126

Since there is a private `factory` struct, I guess it makes sense to use an entirely different name to avoid mixing them.

receiver/receiver.go
dmathieu: It actually can't. Using a var gives the following error:

```
# go.opentelemetry.io/collector/receiver
./receiver.go:81:18: internal.Internal (type) is not an expression
```

Using a type also doesn't work because the type conversion between the local types and the internal ones doesn't work automagically.

receiver/builder.go
dmathieu: Because it's a struct, not an interface. So we don't have a choice, but to make its methods available.

receiver/receiver.go
mx-psi: No strong opinion on the name, if `Factory` is preferred I am fine with that

dmathieu: I have renamed `internal.Internal` to `internal.Factory`.

receiver/builder.go
dmitryax: I see. I don't really like that we are doing it to only partially hide the new experimental API while the other parts are being added as is. Seems like a questionable effort to me. I would leave it as is and add profiling parts where needed, including the interface. I doubt they are going to be changed in any way. However, I'm not going to argue about it if this approach is something that was decided already

dmitryax (approved):

receiver/builder.go
mx-psi: > I see. I don't really like that we are doing it to only partially hide the new experimental API while the other parts are being added as is

We don't have to and are not going to partially hide the API. Adding a method to get the config from a receiver would be sufficient to add support for profiles without exposing experimental methods. Getting rid of the builders (making them an internal impl detail in `service`) is also an option.

mx-psi: See #10621 and https://github.com/open-telemetry/opentelemetry-collector/pull/10375#pullrequestreview-2144929463

PR #10540, opened 2024-07-04
[chore] use macos-14 for arm testing

codeboten:
Similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33921

### DIFF
--- a/.github/workflows/build-and-test-arm.yml
+++ b/.github/workflows/build-and-test-arm.yml
@@ -1,9 +1,9 @@
 name: build-and-test-arm
 on:
   push:
-    branches: [ main ]
+    branches: [main]
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+*'
+      - "v[0-9]+.[0-9]+.[0-9]+*"
   merge_group:
   pull_request:
 env:
@@ -11,7 +11,6 @@ env:
   # Make sure to exit early if cache segment download times out after 2 minutes.
   # We limit cache download as a whole to 5 minutes.
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
-  GOPROXY: https://goproxy1.cncf.selfactuated.dev,direct
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


songy23 (approved):

.github/workflows/build-and-test-arm.yml
```diff
@@ -50,7 +49,7 @@ jobs:
         run: make -j4 gotest
   arm-unittest:
     if: ${{ github.actor != 'dependabot[bot]' && (contains(github.event.pull_request.labels.*.name, 'Run ARM') || github.event_name == 'push' || github.event_name == 'merge_group') }}
-    runs-on: actuated-arm64-4cpu-4gb
```
jpkrohling, line 53: what was this one? are we skipping this in favor of macos-14, or are they equivalent?

mx-psi (approved):

.github/workflows/build-and-test-arm.yml
codeboten: the [actuated runners](https://github.com/self-actuated/actuated) were used to test arm only. the macos-14 provides the same coverage AFAIK

mx-psi:
I think after this change what we claim on https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/platform-support.md#tier-2--secondary-support is no longer correct. But realistically, the current arm runners do not provide good test coverage, they are too flaky, so maybe we can just downgrade our support for Linux ARM for now

codeboten:
> I think after this change what we claim on https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/platform-support.md#tier-2--secondary-support is no longer correct. But realistically, the current arm runners do not provide good test coverage, they are too flaky, so maybe we can just downgrade our support for Linux ARM for now

Pinging @atoulme for review on this as well since he's the owner of linux/arm

TylerHelmuth (approved):

codeboten:
@mx-psi as per your suggestion, i've updated the supported platforms doc to reflect this change

atoulme:
I am acknowledging this change. I didn't get a chance to review and weigh in. I will take further actions to bring back support of linux/arm64 back to tier 2. Thanks.

PR #10542, opened 2024-07-04
[service] add service.disableOpenCensusBridge gate

codeboten:
This feature gate allows end users to re-enable the opencensus bridge if there's a need for it. This preceeds #10406 and will be taken out of draft once https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29867 is completed.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/10414


mx-psi (approved):
Would it make sense to have an info level log about this on initialization?

codeboten:
> Would it make sense to have an info level log about this on initialization?

@mx-psi you mean a warning about opencensus telemetry being disabled?

mx-psi:
> > Would it make sense to have an info level log about this on initialization?
> 
> @mx-psi you mean a warning about opencensus telemetry being disabled?

Yep. Some people don't like having a warning that is not actionable, so that's why I was proposing an info-level log

codeboten:
> > > Would it make sense to have an info level log about this on initialization?
> > 
> > 
> > @mx-psi you mean a warning about opencensus telemetry being disabled?
> 
> Yep. Some people don't like having a warning that is not actionable, so that's why I was proposing an info-level log

Right, sorry i meant a "log" not a "warning" about opencensus telemetry being disabled :) sure i can add that

PR #10543, opened 2024-07-04
[otelcol] Tell all loggers to encode timestamps in ISO8601.

igorpeshansky:
#### Description
OTel collector log messages in Windows are currently displayed in scientific notation, e.g.:
```
1.7194132367192364e+09  error   scraperhelper/scrapercontroller.go:197  Error scraping metrics  …
```
This fix will cause the log messages to use ISO8601 timestamps instead.

#### Testing
Manual testing.

#### Documentation
The previous behavior was undocumented, so no documentation changes needed.


djaglowski (approved):
LGTM. Please [add a changelog entry](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#adding-a-changelog-entry). IMO this is best documented as a breaking change since it may affect users who are consuming the collector's logs.

.chloggen/igorpeshansky-windows-logger-timestamps.yaml
```diff
@@ -0,0 +1,28 @@
...
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: otelcol
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change all logged timestamps to ISO8601.
+
+# One or more tracking issues or pull requests related to the change
+issues: []
```
djaglowski, line 13:
```suggestion
issues: [10543]
```

service/telemetry/logger.go
```diff
@@ -10,11 +10,13 @@ import (
...
 func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
 	// Copied from NewProductionConfig.
+	ec := zap.NewProductionEncoderConfig()
+	ec.EncodeTime = zapcore.ISO8601TimeEncoder
 	zapCfg := &zap.Config{
 		Level:             zap.NewAtomicLevelAt(cfg.Level),
 		Development:       cfg.Development,
 		Encoding:          cfg.Encoding,
-		EncoderConfig:     zap.NewProductionEncoderConfig(),
+		EncoderConfig:     ec,
```
TylerHelmuth, line 19: I believe only the change in `collector_windows.go` is necessary.  Are you sure this needs changed also?

igorpeshansky: Wouldn't we have the same issue with the collector's stderr on Linux without this change?

TylerHelmuth: I'll check again next week, but I believe it is getting set below. I've only ever seen the collector write out ISO timestamps, so I don't believe Linux is affected.

igorpeshansky: @TylerHelmuth Any progress on checking this? Thanks.

TylerHelmuth: @igorpeshansky a couple lines below there is:

```go
if zapCfg.Encoding == "console" {
    // Human-readable timestamps for console format of logs.
    zapCfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
}
```

My guess is that for linux this condition is getting met which is why the timestamps when the collector is run in linux are `ISO8601`.  I think if we set the value here, we can remove that conditional.

igorpeshansky:
Sorry, still planning to work on this.

service/telemetry/logger.go
andrzej-stencel: This code means that when `encoding` is not `console` (bur e.g. `json`), timestamps are not ISO. I can confirm this is the behavior:

```
$ cat config.yaml
exporters:
  nop:
receivers:
  nop:
service:
  pipelines:
    logs:
      exporters:
        - nop
      receivers:
        - nop
  telemetry:
    logs:
      encoding: json
$ otelcol-contrib-0.112.0-linux_amd64 --config 1029-telemetry-json-timestamp.yaml
{"level":"info","ts":1730191656.2376983,"caller":"service@v0.112.0/service.go:135","msg":"Setting up own telemetry..."}
{"level":"info","ts":1730191656.23779,"caller":"telemetry/metrics.go:70","msg":"Serving metrics","address":"localhost:8888","metrics level":"Normal"}
{"level":"info","ts":1730191656.238629,"caller":"service@v0.112.0/service.go:207","msg":"Starting otelcol-contrib...","Version":"0.112.0","NumCPU":20}
{"level":"info","ts":1730191656.2386441,"caller":"extensions/extensions.go:39","msg":"Starting extensions..."}
{"level":"info","ts":1730191656.238668,"caller":"service@v0.112.0/service.go:230","msg":"Everything is ready. Begin running and processing data."}
^C{"level":"info","ts":1730191658.5479696,"caller":"otelcol@v0.112.0/collector.go:328","msg":"Received signal from OS","signal":"interrupt"}
{"level":"info","ts":1730191658.5480957,"caller":"service@v0.112.0/service.go:266","msg":"Starting shutdown..."}
{"level":"info","ts":1730191658.5481503,"caller":"extensions/extensions.go:66","msg":"Stopping extensions..."}
{"level":"info","ts":1730191658.548173,"caller":"service@v0.112.0/service.go:280","msg":"Shutdown complete."}
```

TylerHelmuth: Thanks for confirming that behavior

TylerHelmuth (approved):

djaglowski (approved):

atoulme (approved):

codeboten (approved):

PR #10544, opened 2024-07-04
[service] add logger provider configuration support

codeboten:
This allows us to use the otel-go/config package to support configuring external destinations for logs. I'm putting this in draft to gather community feedback on whether this is a desirable feature for the collector.

I used the following configuration with this PR to send data to an OTLP backend:

```yaml
telemetry:
   logs:
     processors:
       - batch:
           exporter:
             otlp:
               protocol: http/protobuf
               endpoint: https://api.honeycomb.io:443
               headers:
                 "x-honeycomb-team": "${env:HONEYCOMB_API_KEY}"
```

This allowed me to see logs in my backend:

Screenshot 2024-07-04 at 1 49 04 PM


frzifus:
While I think the ability to directly configure the destination of collector logs in the telemetry section might be cool for small and simple configurations in certain environments where you can not simply pick up your looks using `system` or `file` logs.
But I feel it would be more intuitive if we could specify a specific pipeline for exporting the data.

Something like this:
```yaml
exporter:
 otlp/honeycomb:
   protocol: http/protobuf
   endpoint: https://api.honeycomb.io:443
     headers:
      "x-honeycomb-team": "${env:HONEYCOMB_API_KEY}"
pipeline:
   telemetry:
     metrics:
       verbosity: detailed
       pipelines: [logs/collector]
     logs:
       verbosity: detailed
       pipelines: [logs/collector]
   logs/collector:
       receivers:  [telemetry/logs]
       processors: [batch]
       exporters:  [otlp/honeycomb]
   metrics/collector:
       receivers:  [telemetry/metrics]
       processors: [batch]
       exporters:  [otlp/honeycomb]
   traces/collector:
      ...
```
Compared to:
```yaml
telemetry:
   logs:
     processors:
       - batch:
           exporter:
             otlp:
               protocol: http/protobuf
               endpoint: https://api.honeycomb.io:443
               headers:
                 "x-honeycomb-team": "${env:HONEYCOMB_API_KEY}"
   metrics:
     processors:
       - batch:
           exporter:
             otlp:
               protocol: http/protobuf
               endpoint: https://api.honeycomb.io:443
               headers:
                 "x-honeycomb-team": "${env:HONEYCOMB_API_KEY}"
   traces:
     ...
```

wdyt?

codeboten:
@frzifus there was a decision made some time ago (i'm struggling to find a link with this info) to not pipe the collector's own telemetry through itself, to avoid the scenario of the collector's own telemetry being interrupted by an overloaded collector. This is why I'm proposing supporting the logging provider directly. 

Are you thinking of the scenario where you'd like to be able to apply some transformations to the logs?

codeboten:
@open-telemetry/collector-approvers this PR should be ready for review

service/telemetry/logger.go
```diff
@@ -28,9 +32,31 @@ func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
...
+					},
+				},
+			),
+		)
+
+		if err != nil {
+			return nil, err
+		}
+		logger = logger.WithOptions(zap.WrapCore(func(_ zapcore.Core) zapcore.Core {
+			return otelzap.NewCore("go.opentelemetry.io/collector/service/telemetry", otelzap.WithLoggerProvider(sdk.LoggerProvider()))
```
mx-psi, line 56: If you reach this point, I believe we are ignoring many of the existing options in the `service::telemetry::logs` section (e.g. output paths for sure, I wonder about . I think we should fail if said options are set and force the user to pick whether they want to use the `processors` section or the other options.

djaglowski (approved):

service/telemetry/logger.go
codeboten: I've updated the code to use zap's `NewTee` function to preserve the existing logs, wdyt?

service/telemetry/logger.go
```diff
@@ -28,9 +32,40 @@ func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
...
+	if len(cfg.Processors) > 0 {
+		sdk, err := config.NewSDK(
+			config.WithContext(ctx),
+			config.WithOpenTelemetryConfiguration(
+				config.OpenTelemetryConfiguration{
+					LoggerProvider: &config.LoggerProvider{
+						Processors: cfg.Processors,
+					},
+					// TODO: add resource
+					// Resource: &config.Resource{},
```
mx-psi, line 49: File issue about this?

mx-psi (approved):
Teeing sounds reasonable, it makes sense to have a fallback logging to be able to see errors related to the logs exporting. 

Not necessarily in this PR but I think it would be good to have docs about this

service/telemetry/logger.go
```diff
@@ -28,9 +32,40 @@ func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
...
+			),
+		)
+
+		if err != nil {
+			return nil, err
+		}
+
+		logger = logger.WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+			return zapcore.NewTee(
+				logger.Core(),
```
mx-psi, line 60: The linting error made me think you can just do this
```suggestion
				c,
```

service/telemetry/logger.go
```diff
@@ -28,9 +32,40 @@ func newLogger(cfg LogsConfig, options []zap.Option) (*zap.Logger, error) {
...
+				},
+			),
+		)
+
+		if err != nil {
+			return nil, err
+		}
+
+		logger = logger.WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+			return zapcore.NewTee(
```
TylerHelmuth, line 59: Using `NewTee` here means that logs will be written to both cores right? So logs would be written to stdout and exported? Are we sure we want to do that initially or is configuring a processor acknowledgement that logs won't go to stdout? For users that want the collector to stop writing logs to stdout and only use the bridge to export their logs they won't have a solution.

service/telemetry/logger.go
codeboten: I updated the code to add the resource

service/telemetry/logger.go
codeboten: i like the idea of having local logs as a fallback, which makes tee'ing nice in that sense. A user could configure the console log exporter if they wanted to have both local logs and data exported to another destination, but the output from the console will be significantly different.

Output from existing logger:
```
2024-10-22T15:14:38.081-0700    info    service@v0.111.0/service.go:280 Shutdown complete.
```

Output from console exporter:
```
{
        "Timestamp": "2024-10-22T15:14:38.081624-07:00",
        "ObservedTimestamp": "2024-10-22T15:14:38.08164-07:00",
        "Severity": 9,
        "SeverityText": "info",
        "Body": {
                "Type": "String",
                "Value": "Shutdown complete."
        },
        "Attributes": [],
        "TraceID": "00000000000000000000000000000000",
        "SpanID": "0000000000000000",
        "TraceFlags": "00",
        "Resource": [
                {
                        "Key": "service.name",
                        "Value": {
                                "Type": "STRING",
                                "Value": "otelcorecol"
                        }
                },
                {
                        "Key": "service.version",
                        "Value": {
                                "Type": "STRING",
                                "Value": "0.111.0-dev"
                        }
                },
                {
                        "Key": "telemetry.sdk.language",
                        "Value": {
                                "Type": "STRING",
                                "Value": "go"
                        }
                },
                {
                        "Key": "telemetry.sdk.name",
                        "Value": {
                                "Type": "STRING",
                                "Value": "opentelemetry"
                        }
                },
                {
                        "Key": "telemetry.sdk.version",
                        "Value": {
                                "Type": "STRING",
                                "Value": "1.31.0"
                        }
                }
        ],
        "Scope": {
                "Name": "go.opentelemetry.io/collector/service/telemetry",
                "Version": "",
                "SchemaURL": ""
        },
        "DroppedAttributes": 0
}
```

service/telemetry/logger.go
codeboten: fixed

service/telemetry/logger.go
mx-psi: One case where I think the fallback is interesting is if the OTLP logs exporting fails and we have some log from e.g. grpc-go that would be helpful in debugging

codeboten: This PR gives users the option to start producing logs over OTLP, without breaking existing functionality. I would suggest we follow up this PR with a feature gate that disables the tee'ing of the logs.

This would give users that don't want to double emit the logs to turn off the local console logging if a provider is configured. This could become the default in the future if users support this idea.

One thing to note is that configuration of the logging today supports sampling logs, which the otel logging provider has no concept of at this time

TylerHelmuth: That works for me. As long as we have some plan for how to give users the option to disable the console logging if they want I'm happy.

TylerHelmuth (approved):

mowies:
this feature is not yet added to the [docs](https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-logs) it seems. is there already a docs issue created for it?

mx-psi:
@mowies I don't think there is one, feel free to create one :)

PR #10555, opened 2024-07-08
[chore] Use string for unit type

mx-psi:
#### Description

Uses string type for `unit` field in mdatagen related files.

#### Link to tracking issue

Needed for #10554


codeboten (approved):
We should probably also update these units to follow semconv, can be done in a separate PR... not sure if that's a breaking change or not

codeboten:
Opened https://github.com/open-telemetry/opentelemetry-collector/issues/10556 to follow up

PR #10435, opened 2024-06-18
[confmap] Move confmap.unifyEnvVarExpansion to beta

TylerHelmuth:
#### Description
Moves confmap.unifyEnvVarExpansion to beta.  This means the collector will, by default, use the env var provider to expand `${FOO}` synatx and will error if the expandconverter is used to expand `$FOO` syntax.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/10161
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/8215
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/7111

PR #10560, opened 2024-07-08
[confmap] Fix bug where expand didn't honor escaping

TylerHelmuth:
#### Description

When we promoted `confmap.unifyEnvVarExpansion` to beta, we found that the new expansion logic in `confmap` wasn't handling escaping of `$$` like it is supposed to.  This PR fixes that bug, but adding escaping logic for `$$`.

@azunna1 this fixes the bug you mentioned in https://github.com/open-telemetry/opentelemetry-collector/pull/10435 around the metricstransformprocessor:

```yaml
  metricstransform:
    transforms:
      - include: '^k8s\.(.*)\.(.*)$$'
        match_type: regexp
        action: update
        new_name: 'kubernetes.$${1}.$${2}'
      - include: '^container_([0-9A-Za-z]+)_([0-9A-Za-z]+)_.*'
        match_type: regexp
        action: update
        new_name: 'container.$${1}.$${2}'
```

#### Testing
Added new unit tests explicitly for escaping logic

### DIFF
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -10,9 +10,10 @@ import (
 	"regexp"
 	"strings"
 
-	"go.opentelemetry.io/collector/internal/featuregates"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/internal/featuregates"
 )
 
 // follows drive-letter specification:
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"strings"
 
+	"go.opentelemetry.io/collector/internal/featuregates"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
@@ -171,7 +172,13 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 		if err != nil {
 			return nil, err
 		}
-		cfgMap[k] = val
+
+		if v, ok := val.(string); ok && featuregates.UseUnifiedEnvVarExpansionRules.IsEnabled() {
+			cfgMap[k] = strings.ReplaceAll(v, "$$", "$")
+		} else {
+			cfgMap[k] = val
+		}
+
 	}
 	retMap = NewFromStringMap(cfgMap)
 
--- a/confmap/testdata/expand-escaped-env.yaml
+++ b/confmap/testdata/expand-escaped-env.yaml
@@ -1,0 +1,29 @@
+test_map:
+  # $$ -> escaped $
+  recv.1: "$$MAP_VALUE_1"
+  # $$$ -> escaped $ + $MAP_VALUE_2
+  recv.2: "$$$MAP_VALUE_2"
+  # $$$$ -> two escaped $
+  recv.3: "$$$$MAP_VALUE_3"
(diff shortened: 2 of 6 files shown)


confmap/resolver.go
```diff
@@ -171,7 +172,13 @@ func (mr *Resolver) Resolve(ctx context.Context) (*Conf, error) {
 		if err != nil {
 			return nil, err
 		}
-		cfgMap[k] = val
+
+		if v, ok := val.(string); ok && featuregates.UseUnifiedEnvVarExpansionRules.IsEnabled() {
```
evan-bradley, line 176: I think `Test_EscapedEnvVars` only validates output with the feature gate enabled, what's the output if it is disabled?

TylerHelmuth: I'll add a test for that as well.

TylerHelmuth: Actually it should be handled by this existing test: https://github.com/open-telemetry/opentelemetry-collector/blob/8de544b176179a21637baf094b833969786f4479/confmap/converter/expandconverter/expand_test.go#L116C29-L116C35.  In order to write a test that mimics the functionality with the feature gate disabled I'd have to reproduced the expand converters logic.

evan-bradley: >In order to write a test that mimics the functionality with the feature gate disabled I'd have to reproduced the expand converters logic in confmap tests.

Thanks for digging into it. That assumes that the expand converter is used, though, right? I guess without this, it should just be that all `$$` remain as `$$` instead of being replaced.

TylerHelmuth: Ya without the expand converter in place, there was no escaping.

confmap/testdata/expand-escaped-env.yaml
```diff
@@ -0,0 +1,29 @@
+test_map:
+  # $$ -> escaped $
+  recv.1: "$$MAP_VALUE_1"
+  # $$$ -> escaped $ + $MAP_VALUE_2
+  recv.2: "$$$MAP_VALUE_2"
+  # $$$$ -> two escaped $
+  recv.3: "$$$$MAP_VALUE_3"
+  # $$$ -> escaped $ + substituted env var
+  recv.4: "$$${MAP_VALUE_2}"
```
mx-psi, line 9: I'd like to see one like this but using the `${env:..}` syntax

TylerHelmuth: Added

evan-bradley (approved):

codeboten (approved):

danelson:
@TylerHelmuth would you happen to know how I can identify if I am affected by this? I assumed something like the following would be problematic but it doesn't appear to be

```yaml
transform/redact_sensitive_information:
  error_mode: ignore
  log_statements:
    - context: log
      statements:
        - replace_all_patterns(attributes, "value", "Bearer ey.*", "$$1<redacted>")
        - replace_pattern(body, "Bearer ey.*", "$$1<redacted>")
```

I have quite a few more uses of `$$` being used in the transform processor and I was hoping to prove I wasn't affected. Worse case I guess I can just toggle off `confmap.unifyEnvVarExpansion`.

dmitryax:
@danelson you likely will be affected. See https://github.com/open-telemetry/opentelemetry-collector/issues/10713

Issue #10538, opened 2024-07-04
wrong status code set in tracing spans


rafaeljesus:
**Describe the bug**

I have a `logs.exporters.otlphttp` that is not properly converting the status code being returned from the logging backend, the problem is that the logging backend (in this case logmixer-ingest) is returning 400 status code but the collector server span (in this case logingest) is setting 500 instead, this is causing my SLOs to incorrectly burn

**Steps to reproduce**

**What did you expect to see?**

I expect the server span status code attribute to correctly match what is being returned by the logging service, which in this case, the logging service returns 400 but the server span sets 500 instead

**What did you see instead?**

here we see that the http client is correctly setting 400 to the span

<img width="1610" alt="Screenshot 2024-07-04 at 14 53 16" src="https://github.com/open-telemetry/opentelemetry-collector/assets/1216425/5be19769-c203-422b-92d6-420f0124a453">

in the span above, we see that the permanent error message contains the 400

<img width="1673" alt="Screenshot 2024-07-04 at 14 48 34" src="https://github.com/open-telemetry/opentelemetry-collector/assets/1216425/15a6cfc0-fb07-4d07-8a80-90e79ab97d43">

but in the server span we see that it is 500 instead

<img width="1606" alt="Screenshot 2024-07-04 at 14 50 10" src="https://github.com/open-telemetry/opentelemetry-collector/assets/1216425/d090dc57-9217-4996-b55f-f3e70b6d0aa9">

**What version did you use?**

v0.95.0

**What config did you use?**

this is the configMap managed by https://github.com/open-telemetry/opentelemetry-operator

```yaml
apiVersion: v1
data:
  collector.yaml: |
    receivers:
      otelarrow:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
            include_metadata: true
            keepalive:
              server_parameters:
                max_connection_age: 180s
                max_connection_age_grace: 240s
            max_recv_msg_size_mib: 20
      otlp:
        protocols:
          http:
            endpoint: 0.0.0.0:4318
            include_metadata: true
    exporters:
      logging:
        verbosity: detailed
      otlphttp:
        auth:
          authenticator: headers_setter
        logs_endpoint: http://logmixer-ingest:9201/v1/otlp
        retry_on_failure:
          enabled: false
        sending_queue:
          enabled: false
        timeout: 5s
    extensions:
      headers_setter:
        headers:
          - from_context: lightstep-org-id
            key: lightstep-org-id
[...]

PR #10574, opened 2024-07-09
Properly handle error code translation

TylerHelmuth:
#### Description

Fixes a bug where the otlpreceiver was not translating grpc error codes to http status codes the same way that the otlp exporter was translating http status codes to grpc error codes.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10538

#### Testing
Added unit tests

### DIFF
--- a/receiver/otlpreceiver/internal/errors/errors.go
+++ b/receiver/otlpreceiver/internal/errors/errors.go
@@ -40,6 +40,18 @@ func GetHTTPStatusCodeFromStatus(s *status.Status) int {
 	case codes.ResourceExhausted:
 		return http.StatusTooManyRequests
 	// Not Retryable
+	case codes.InvalidArgument:
+		return http.StatusBadRequest
+	// Not Retryable
+	case codes.Unauthenticated:
+		return http.StatusUnauthorized
+	// Not Retryable
+	case codes.PermissionDenied:
+		return http.StatusForbidden
+	// Not Retryable
+	case codes.Unimplemented:
+		return http.StatusNotFound
+	// Not Retryable
 	default:
 		return http.StatusInternalServerError
 	}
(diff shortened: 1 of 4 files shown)


evan-bradley (approved):
Verified with [`NewStatusFromMsgAndHTTPCode`](https://github.com/open-telemetry/opentelemetry-collector/blob/af33ac5acc08dbeb43426f86b526d2cfd29318ab/internal/httphelper/helper.go#L15C6-L15C33) and the two conversions should match now.

jaronoff97 (approved):

receiver/otlpreceiver/internal/errors/errors.go
```diff
@@ -40,6 +40,18 @@ func GetHTTPStatusCodeFromStatus(s *status.Status) int {
 	case codes.ResourceExhausted:
 		return http.StatusTooManyRequests
 	// Not Retryable
+	case codes.InvalidArgument:
```
jaronoff97, line 43: @TylerHelmuth maybe we should this function and its tests to the helper i made last time? https://github.com/open-telemetry/opentelemetry-collector/blob/af33ac5acc08dbeb43426f86b526d2cfd29318ab/internal/httphelper/helper.go#L15C6-L15C33

That way the conversions are in the same place

TylerHelmuth: @jaronoff97 it is all moving (and properly solved by remembering the actual http status code) via https://github.com/open-telemetry/opentelemetry-collector/pull/9041, so we should stick with that PR so the permanent solution to this issue. This PR fixes the bug for now.

jaronoff97: cool! I have no problem with that :D

codeboten (approved):

PR #9584, opened 2024-02-14
[chore] arm64 build

atoulme:
This PR adds a linux/arm64 build to the build of the collector, so it may support the goal of #9731

PR #10590, opened 2024-07-10
[chore] remove arm stuff from `build-and-test.yaml`

TylerHelmuth:
These things were added for the [arm build](https://github.com/open-telemetry/opentelemetry-collector/pull/9584) which has been moved to its own job

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -145,13 +145,6 @@ jobs:
     runs-on: ${{ matrix.runner }}
     needs: [setup-environment]
     steps:
-      - name: Set up arkade
-        uses: alexellis/setup-arkade@b1816384b2260cfd2c023c6798d26075786cfc7f # v3
-      - name: Install vmmeter
-        run: |
-          sudo -E arkade oci install ghcr.io/openfaasltd/vmmeter:latest --path /usr/local/bin/
-      - name: Run vmmeter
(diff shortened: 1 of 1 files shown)


songy23 (approved):

atoulme:
This is not just arm stuff, this provides utilization report of the runner.

TylerHelmuth:
@atoulme where is that report going? Are we using it?

adrielp:
@TylerHelmuth   - utilization report looks to be just text in the [post step of the run](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/9877591113/job/27279705543) which is all the [action code exports as](https://github.com/self-actuated/vmmeter-action/blob/master/dist/post/index.js#L33264-L33270). 

@atoulme - do we know the dev's over at self-actuated? I don't mean this in a negative way (seems like a cool set of projects), but it seems like a small team with low-bandwidth and a lot going on. Might be good to contribute the fix, fork, or maybe use something else?

TylerHelmuth:
@adrielp Thanks for pointing that out where the report lives. 

My opinion is that the extra complexity of these steps is not worth the report it is generating. If we want more insight into our CI I'd rather trace our CI or advance the GitHub action receiver that @krzko proposed.

atoulme:
Oh I see. Let's remove it now and report to the vmeter maintainer.

atoulme (approved):

crobert-1 (approved):

codeboten (approved):

Issue #9943, opened 2024-04-11
[exporter/exporterhelper] queue_size and queue_capacity internal metrics are missing the data type attribute


dloucasfx:
**Describe the bug**
When the same exporter definition is used in different data type pipelines; example: `otlp` exporter used in `metrics` and `logs` pipeline, the exact same metric will be initiated twice https://github.com/dloucasfx/opentelemetry-collector/blob/main/exporter/exporterhelper/queue_sender.go#L118-L139 which makes it impossible to know which queue type the metric is measuring.
We need to add the queue data type as an attribute, which requires the `queue_sender` struct to be expanded and exposes the datatype field. 

**Steps to reproduce**
1- Define an exporter, example: otlp
2- Use this exporter in the logs and metrics pipeline
3- monitor the internal metric `otelcol_exporter_queue_size` and notice that there is a single MTS and you can't tell which queue data type it's measuring
````
Metric #0
Descriptor:
     -> Name: otelcol_exporter_queue_size
     -> Description: Current size of the retry queue (in batches)
     -> Unit:
     -> DataType: Gauge
NumberDataPoints #0
Data point attributes:
     -> exporter: Str(otlp)
     -> service_instance_id: Str(0fafd546-8c21-4bd4-a8c7-0faeec4482df)
     -> service_name: Str(otelcontribcol)
     -> service_version: Str(0.97.0-dev)
StartTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 2024-04-11 16:09:44.873 +0000 UTC
Value: 0.000000
````

**What did you expect to see?**
2 MTSes for `otelcol_exporter_queue_size`  one with `log` data type attribute and one with `metric` data type attribute

**What did you see instead?**
a single `otelcol_exporter_queue_size` MTS and you can't tell which queue data type it's measuring

**What version did you use?**
`v0.98.0`

**What config did you use?**
```
receivers:
  hostmetrics:
    collection_interval: 10s
    scrapers:
      filesystem:
      memory:
  tcplog:
    listen_address: "0.0.0.0:54525"
  prometheus/internal:
    config:
      scrape_configs:
        - job_name: 'otel-collector'
          scrape_interval: 10s
          static_configs:
            - targets: [ "127.0.0.1:8899" ]
          metric_relabel_configs:
            - source_labels: [ __name__ ]
              regex: 'otelcol_rpc_.*'
              action: drop
            - source_labels: [ __name__ ]
[...]

PR #10593, opened 2024-07-11
[exporterhelper] Add data_type attribute to internal queue metrics

dmitryax:
Add `data_type` attribute to the internal otelcol_exporter_queue_size metric to report the type of data being processed.

All other metrics have the data type reported as part of their names. We could've done the same for queue metrics, but that would introduce a significant breaking change. We want to avoid that until we have all the metrics standardized with OpenTelemetry semantic conventions.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9943

### DIFF
--- a/component/componenttest/obsreporttest.go
+++ b/component/componenttest/obsreporttest.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"go.opentelemetry.io/otel/attribute"
 	otelprom "go.opentelemetry.io/otel/exporters/prometheus"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
@@ -72,8 +73,10 @@ func (tts *TestTelemetry) CheckExporterLogs(sentLogRecords, sendFailedLogRecords
 	return tts.prometheusChecker.checkExporterLogs(tts.id, sentLogRecords, sendFailedLogRecords)
 }
 
-func (tts *TestTelemetry) CheckExporterMetricGauge(metric string, val int64) error {
-	return tts.prometheusChecker.checkExporterMetricGauge(tts.id, metric, val)
+func (tts *TestTelemetry) CheckExporterMetricGauge(metric string, val int64, extraAttrs ...attribute.KeyValue) error {
+	attrs := attributesForExporterMetrics(tts.id)
+	attrs = append(attrs, extraAttrs...)
+	return tts.prometheusChecker.checkGauge(metric, val, attrs)
 }
 
 // CheckProcessorTraces checks that for the current exported values for trace exporter metrics match given values.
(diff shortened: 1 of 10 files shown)


songy23 (approved):

component/componenttest/obsreporttest.go
```diff
@@ -72,8 +73,10 @@ func (tts *TestTelemetry) CheckExporterLogs(sentLogRecords, sendFailedLogRecords
 	return tts.prometheusChecker.checkExporterLogs(tts.id, sentLogRecords, sendFailedLogRecords)
 }
 
-func (tts *TestTelemetry) CheckExporterMetricGauge(metric string, val int64) error {
-	return tts.prometheusChecker.checkExporterMetricGauge(tts.id, metric, val)
+func (tts *TestTelemetry) CheckExporterMetricGauge(metric string, val int64, extraAttrs ...attribute.KeyValue) error {
```
dmitryax, line 76: @open-telemetry/collector-approvers do you think this change needs a breaking change item in API changelog?

mx-psi: Yes, IMO we don't need a deprecation period but we should add a changelog entry

mx-psi (approved):

internal/obsreportconfig/obsmetrics/obs_exporter.go
```diff
@@ -7,6 +7,9 @@ const (
 	// ExporterKey used to identify exporters in metrics and traces.
 	ExporterKey = "exporter"
 
+	// DataTypeKey used to identify the data type in the queue size metric.
+	DataTypeKey = "data_type"
```
codeboten, line 11: should this be what the [otep](https://github.com/open-telemetry/oteps/pull/259/files#diff-8ab369bae50cfda348c0bd20882db8a1bdbbffb979abc70bff298192a948588bR218) recommends `otel.signal`?

dmitryax: I'm not sure about it. We use `data_type` as a field for logging. Once the OTEP is ready and we adopt it, the metric is going to change anyway. Keeping an attribute name doesn't help with the migration. Also, `otel.signal` could still change before the OTEP is merged?

component/componenttest/obsreporttest.go
dmitryax: added the entry

internal/obsreportconfig/obsmetrics/obs_exporter.go
dmitryax: @codeboten I'm open to any attribute name. Let me know what do you think considering my previous comment.

codeboten: @dmitryax probably ok to go with an attribute consistent w/ logging for now.

codeboten (approved):
I searched through semantic conventions and couldn't find guidance for attributes that specify signal types. Should an issue be opened there?

dmitryax:
> I searched through semantic conventions and couldn't find guidance for attributes that specify signal types. Should an issue be opened there?

Yeah, that's something that can be done before the OTEP

PR #10598, opened 2024-07-11
Add mechanism to report go module for each factory

djaglowski:
#### Description

Adds each component's module to the `components` subcommand output.

```
➜  ./bin/otelcorecol_darwin_arm64 components                                                                                    ✭
...
buildinfo:
    command: otelcorecol
    description: Local OpenTelemetry Collector binary, testing only.
    version: 0.104.0-dev
receivers:
    - name: nop
      module: go.opentelemetry.io/collector/receiver/nopreceiver@v0.104.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
    - name: otlp
      module: go.opentelemetry.io/collector/receiver/otlpreceiver@v0.104.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
processors:
    - name: batch
      module: go.opentelemetry.io/collector/processor/batchprocessor@v0.104.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
    - name: memory_limiter
      module: go.opentelemetry.io/collector/processor/memorylimiterprocessor@v0.104.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
exporters:
    - name: debug
      module: go.opentelemetry.io/collector/exporter/debugexporter@v0.104.0
      stability:
        logs: Development
        metrics: Development
        traces: Development
    - name: logging
      module: go.opentelemetry.io/collector/exporter/loggingexporter@v0.104.0
      stability:
        logs: Deprecated
        metrics: Deprecated
        traces: Deprecated
    - name: nop
      module: go.opentelemetry.io/collector/exporter/nopexporter@v0.104.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
    - name: otlp
      module: go.opentelemetry.io/collector/exporter/otlpexporter@v0.104.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
    - name: otlphttp
      module: go.opentelemetry.io/collector/exporter/otlphttpexporter@v0.104.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
connectors:
    - name: forward
      module: go.opentelemetry.io/collector/connector/forwardconnector@v0.104.0
      stability:
        logs-to-logs: Beta
        logs-to-metrics: Undefined
        logs-to-traces: Undefined
        metrics-to-logs: Undefined
        metrics-to-metrics: Beta
        metrics-to-traces: Undefined
        traces-to-logs: Undefined
[...]

PR #10599, opened 2024-07-11
Enhance builder to capture go module for each component type

### LINKED ISSUE #10570: Components subcommand - Add module@version for each component
The `components` subcommand is useful for understanding which types are available, but it would be helpful in some cases to know the exact module and version which defines each component.

Enhance the `components` subcommand to include the go module and version.

Current Output:
```yaml
receivers:
    - name: otlp
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
```

Proposed Output:
```yaml
receivers:
    - name: otlp
      module: go.opentelemetry.io/collector/receiver/otlpreceiver@v0.104.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
```

The best solution I have found involves using `runtime.Caller`. Basically, within each `<kind>.NewFactory` function we can call `runtime.Caller` to get the module (including version), then save the result onto the corresponding `<kind>.factory` struct and return it via a new method on the corresponding `<kind>.Factory` interface, e.g. `GoModule() string`. [Here is a working prototype.](https://github.com/open-telemetry/opentelemetry-collector/compare/main...djaglowski:opentelemetry-collector:module-info)

I explored whether reflection could be used on the factories directly but could not identify a mechanism to get the corresponding module. The only way I can find to get any information about modules is using [`runtime.debug.ReadBuildInfo`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/runtime/debug/mod.go;l=20), but this provides an comprehensive list of module dependencies for the collector binary, without any reliable way to match them to the factories.

I also explored the possibility of using the OCB manifest in some way, but I have not identified a straightforward way to match modules with corresponding component types. e.g. How do we know for certain that `go.opentelemetry.io/collector/receiver/otlpreceiver` contains a component type called `otlp`? Presumably we could enhance the `FactoryMap` structs to pair the module string with the factory, but this seems overly complicated and won't apply to collectors built outside of OCB.

**Additional context**

This same information would be extremely useful for OpAMP, since a server would be able to confidently understand the exact components available in a collector, including the precise corresponding configuration. Adding it to the `components` subcommand seems like a natural first step. Later, we should be able to make the information available to the opampextension in some way.

djaglowski:
Resolves #10570

Alternative to #10598

This implementation is more reliable that #10598 because it gets information directly from the builder manifest. It relies on additional structure in the `component.go` file, ultimately encoded in `otelcol.Factories`.

An alternative would be to push the enhancements deeper, into the `<kind>.Factories` implementations, so that the module information is available directly alongside the Factory.

### DIFF
--- a/.chloggen/module-info-builder.yaml
+++ b/.chloggen/module-info-builder.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: cmd/builder
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add go module info the builder generated code.
+
+# One or more tracking issues or pull requests related to the change
+issues: [10570]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/cmd/builder/internal/builder/templates/components.go.tmpl
+++ b/cmd/builder/internal/builder/templates/components.go.tmpl
@@ -3,6 +3,9 @@
 package main
 
 import (
+	{{- if .Distribution.SupportsComponentModules}}
+	"go.opentelemetry.io/collector/component"
+	{{- end}}
 	"go.opentelemetry.io/collector/connector"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/extension"
@@ -38,6 +41,12 @@ func components() (otelcol.Factories, error) {
 	if err != nil {
 		return otelcol.Factories{}, err
 	}
+	{{- if .Distribution.SupportsComponentModules}}
+	factories.ExtensionModules = make(map[component.Type]string, len(factories.Extensions))
+	{{- range .Extensions}}
+	factories.ExtensionModules[{{.Name}}.NewFactory().Type()] = "{{.GoMod}}"
+	{{- end}}
+	{{- end}}
 
 	factories.Receivers, err = receiver.MakeFactoryMap(
 		{{- range .Receivers}}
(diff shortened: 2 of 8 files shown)


djaglowski:
Hmm, I thought I had tests passing. I'll move this back to draft status until I get CI resolved.

.chloggen/module-info-builder.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: cmd
```
jpkrohling, line 7:
```suggestion
component: cmd/builder
```

jpkrohling (approved):

jpkrohling (approved):

.chloggen/module-info-builder.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: cmd/builder
```
evan-bradley, line 7:
```suggestion
component: cmd/builder, otelcol
```
Maybe this should just be a separate changelog entry, but we have two separate changes in here:
1. Adding support in the builder for baking this information into the binary
2. Adding support in the `components` subcommand for printing this information.

cmd/builder/internal/builder/templates/components.go.tmpl
evan-bradley, line 1: Just asking because the version requirement and testing setup makes it difficult to verify this from the PR, were you able to test the output of this template?

evan-bradley (approved):
Overall looks good to me.

evan-bradley:
>An alternative would be to push the enhancements deeper, into the `<kind>.Factories` implementations, so that the module information is available directly alongside the Factory.

I think the current placement is good for use in the `components` subcommand, but if we had a use for the modules at runtime, such as for observability, I could see that being a better approach.

cmd/builder/internal/builder/templates/components.go.tmpl
djaglowski: Yeah, but I could only do so manually for now but modifying the condition:

`constraint, err = version.NewConstraint(">= 0.105.0")` => `constraint, err = version.NewConstraint(">= 0.104.0")`

Not sure if there's a better way but it's easy enough to validate manually and going forward will be covered automatically.

.chloggen/module-info-builder.yaml
djaglowski: Yeah, good call. I'll push a separate changelog entry

cmd/builder/internal/builder/templates/components.go.tmpl
evan-bradley: That would be my approach as well, unfortunately the setup is a little tricky. Just wanted to do my due diligence and double-check that we've verified it.

atoulme (approved):

djaglowski:
> I think the current placement is good for use in the components subcommand, but if we had a use for the modules at runtime, such as for observability, I could see that being a better approach.

I'm planning to investigate how to get the module info to the opamp extension, so likely will be looking into this more.

evan-bradley:
>I'm planning to investigate how to get the module info to the opamp extension, so likely will be looking into this more.

If we're comfortable relying on the output of this command, we could use it in the Supervisor to relay it to the OpAMP server. That would still be outside OpAMP extension, but at a minimum this will work for the current upstream Collector+OpAMP design.

PR #10601, opened 2024-07-11
[wip] use otel-go config package to configure MeterProvider

codeboten:
This removes all the code that was mostly a duplication of the code that exists in otel-go contrib's config package.

Follow https://github.com/open-telemetry/opentelemetry-collector/pull/10406


codeboten:
Replaced by https://github.com/open-telemetry/opentelemetry-collector/pull/11079

Issue #10405, opened 2024-06-13
Error loading TLS certificates from ENV variable


epchris:
**Describe the bug**
When attempting to configure TLS using environment variables for the contents of the certs there is an error parsing the value of the certificates.

**Steps to reproduce**
Use the following configuration file:

```
---
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4318
processors:
  batch:

exporters:
  kafka:
    auth:
      tls:
        # DOES NOT WORK: Contents of the cert in the env variable
        ca_pem: ${env:CA_PEM}

        # WORKS
        # ca_pem: |
        #   -----BEGIN CERTIFICATE-----
        #   MIICUTCCAfugAwIBAgIBADANBgkqhkiG9w0BAQQFADBXMQswCQYDVQQGEwJDTjEL
        #   MAkGA1UECBMCUE4xCzAJBgNVBAcTAkNOMQswCQYDVQQKEwJPTjELMAkGA1UECxMC
        #   VU4xFDASBgNVBAMTC0hlcm9uZyBZYW5nMB4XDTA1MDcxNTIxMTk0N1oXDTA1MDgx
        #   NDIxMTk0N1owVzELMAkGA1UEBhMCQ04xCzAJBgNVBAgTAlBOMQswCQYDVQQHEwJD
        #   TjELMAkGA1UEChMCT04xCzAJBgNVBAsTAlVOMRQwEgYDVQQDEwtIZXJvbmcgWWFu
        #   ZzBcMA0GCSqGSIb3DQEBAQUAA0sAMEgCQQCp5hnG7ogBhtlynpOS21cBewKE/B7j
        #   V14qeyslnr26xZUsSVko36ZnhiaO/zbMOoRcKK9vEcgMtcLFuQTWDl3RAgMBAAGj
        #   gbEwga4wHQYDVR0OBBYEFFXI70krXeQDxZgbaCQoR4jUDncEMH8GA1UdIwR4MHaA
        #   FFXI70krXeQDxZgbaCQoR4jUDncEoVukWTBXMQswCQYDVQQGEwJDTjELMAkGA1UE
        #   CBMCUE4xCzAJBgNVBAcTAkNOMQswCQYDVQQKEwJPTjELMAkGA1UECxMCVU4xFDAS
[...]

PR #10603, opened 2024-07-12
[confmap] Use original string value on expansion

mx-psi:
#### Description

Originally on the [env var RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/env-vars.md) I proposed using the YAML parsed value when parsing a string, which has been the default behavior until now of `${env:ENV}` and friends.
An alternative is to use the raw string value, which is what `${ENV}` used to do. The difference (after enabling strict types) does not matter much in practice save for a few edge cases (since strings can be represented in multiple ways), namely on the YAML parsing scenario:
- Quotes get removed (`"opentelemetry"` -> `opentelemetry`)
- YAML type tags get removed (`!!str opentelemetry` -> `opentelemetry`)
- Literal newlines get removed (although this may be a bug in our parser, see go-yaml/yaml/issues/963)
- Escaping sequences are mapped to their own character

It's a matter of opinion which behavior is more intuitive to end users. In this PR I change the behavior to use the raw string value. I think the arguments in favor are:
- It's arguably less surprising (you get what you wrote)
- It solves a real issue faced by an user (see below)

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10405#issuecomment-2225292473

### DIFF
--- a/confmap/internal/e2e/types_test.go
+++ b/confmap/internal/e2e/types_test.go
@@ -228,27 +228,27 @@ func TestStrictTypeCasting(t *testing.T) {
 		{
 			value:       "\"0123\"",
 			targetField: TargetFieldString,
-			expected:    "0123",
+			expected:    "\"0123\"",
 		},
 		{
 			value:        "\"0123\"",
 			targetField:  TargetFieldInt,
-			unmarshalErr: "'field' expected type 'int', got unconvertible type 'string', value: '0123'",
+			unmarshalErr: "'field' expected type 'int', got unconvertible type 'string', value: '\"0123\"'",
 		},
 		{
 			value:       "\"0123\"",
 			targetField: TargetFieldInlineString,
-			expected:    "inline field with 0123 expansion",
+			expected:    "inline field with \"0123\" expansion",
 		},
 		{
 			value:       "!!str 0123",
 			targetField: TargetFieldString,
-			expected:    "0123",
+			expected:    "!!str 0123",
 		},
 		{
 			value:       "!!str 0123",
 			targetField: TargetFieldInlineString,
-			expected:    "inline field with 0123 expansion",
+			expected:    "inline field with !!str 0123 expansion",
 		},
 		{
 			value:        "t",
--- a/confmap/internal/e2e/types_test.go
+++ b/confmap/internal/e2e/types_test.go
@@ -98,27 +98,22 @@ func TestTypeCasting(t *testing.T) {
 		{
 			value:       "\"0123\"",
 			targetField: TargetFieldString,
-			expected:    "0123",
-		},
-		{
-			value:       "\"0123\"",
-			targetField: TargetFieldInt,
-			expected:    83,
+			expected:    "\"0123\"",
 		},
 		{
 			value:       "\"0123\"",
 			targetField: TargetFieldInlineString,
-			expected:    "inline field with 0123 expansion",
+			expected:    "inline field with \"0123\" expansion",
 		},
 		{
 			value:       "!!str 0123",
 			targetField: TargetFieldString,
-			expected:    "0123",
+			expected:    "!!str 0123",
 		},
 		{
 			value:       "!!str 0123",
 			targetField: TargetFieldInlineString,
-			expected:    "inline field with 0123 expansion",
+			expected:    "inline field with !!str 0123 expansion",
 		},
 		{
 			value:       "t",
--- a/confmap/provider.go
+++ b/confmap/provider.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"go.opentelemetry.io/collector/internal/featuregates"
 	"go.uber.org/zap"
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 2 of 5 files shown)


confmap/provider.go
```diff
@@ -138,9 +139,14 @@ func NewRetrievedFromYAML(yamlBytes []byte, opts ...RetrievedOption) (*Retrieved
...
-	switch v := rawConf.(type) {
+	switch rawConf.(type) {
 	case string:
-		opts = append(opts, withStringRepresentation(v))
+		if featuregates.UseUnifiedEnvVarExpansionRules.IsEnabled() {
+			// If rawConf is a string, we override it with the raw yamlBytes.
+			// This is the original ${ENV} expansion behavior.
+			rawConf = string(yamlBytes)
+		}
+		opts = append(opts, withStringRepresentation(string(yamlBytes)))
```
mx-psi, line 149: This change is effectively under the strict types gate

confmap/internal/e2e/types_test.go
```diff
@@ -98,27 +98,22 @@ func TestTypeCasting(t *testing.T) {
 		{
 			value:       "\"0123\"",
 			targetField: TargetFieldString,
-			expected:    "0123",
-		},
-		{
-			value:       "\"0123\"",
-			targetField: TargetFieldInt,
-			expected:    83,
+			expected:    "\"0123\"",
```
TylerHelmuth, line 101: This breaks users right?
If I entered `endpoint: "http://localhost:4317"` instead of my exporter using `http://localhost:4317` would it attempt to use a value that contains double-quote literals?

mx-psi: > If I entered endpoint: "http://localhost:4317" instead of my exporter using http://localhost:4317 would it attempt to use a value that contains double-quote literals?

That would happen if you did 

```
endpoint: ${env:ENDPOINT}
```

with `ENDPOINT` set to `"http://localhost:4317"` (quotes included). The changes only affects how the values passed via a provider are parsed

mx-psi:
If we were to do this we would have to sync with the OpenTelemetry Config WG to see how this lines up with their behavior. I am still not sure if this is a good idea or not

confmap/internal/e2e/types_test.go
```diff
@@ -98,27 +98,22 @@ func TestTypeCasting(t *testing.T) {
...
 			value:       "\"0123\"",
 			targetField: TargetFieldInlineString,
-			expected:    "inline field with 0123 expansion",
+			expected:    "inline field with \"0123\" expansion",
 		},
 		{
 			value:       "!!str 0123",
 			targetField: TargetFieldString,
-			expected:    "0123",
+			expected:    "!!str 0123",
```
codeboten, line 111: This only applies on provider expanded strings correct? QBQ it isn't breaking all yaml rules for everything :D

TylerHelmuth: Correct, only when expanding using the env provider, which happens with `${env:}` and `${}`

confmap/provider.go
```diff
@@ -138,9 +140,14 @@ func NewRetrievedFromYAML(yamlBytes []byte, opts ...RetrievedOption) (*Retrieved
...
```
TylerHelmuth, line 150: This changes the behavior when the feature gate is disabled.  Before it would set the string representation to be the yaml parsed version of the string (so `\n` is removed.

docs/rfcs/env-vars.md
```diff
@@ -216,7 +216,7 @@ loading a field with the braces syntax, `env` syntax.
 | `0123`       | integer    | 83                                       | 83                                             | 83                             | n/a                                   |
 | `0123`       | string     | 0123                                     | 83                                             | 0123                           | 0123                                  |
 | `0xdeadbeef` | string     | 0xdeadbeef                               | 3735928559                                     | 0xdeadbeef                     | 0xdeadbeef                            |
-| `"0123"`     | string     | "0123"                                   | 0123                                           | 0123                           | 0123                                  |
-| `!!str 0123` | string     | !!str 0123                               | 0123                                           | 0123                           | 0123                                  |
+| `"0123"`     | string     | "0123"                                   | 0123                                           | "0123"                         | "0123"                                |
```
TylerHelmuth, line 219: Let's add a row for `foo\nbar`

TylerHelmuth: Add `foo::`

confmap/provider.go
mx-psi: Yup, but the string representation is only used under the strict types feature gate, so it's also under a feature gate

mx-psi:
I have been mulling over this over the weekend and I don't think this is a good idea, since it completely prevents users from escaping content to say "this is a string". A better solution would be to preserve the original string value and use it while unmarshaling with a decoding hook if the target field is a string. I have tried and failed to implement this (not saying it can't be done, but it's definitely messy to do)

mx-psi:
#10618 is a more complex but more sensible version of this

mx-psi:
Closing this in favor of #10618

PR #10604, opened 2024-07-12
[chore] Use 8 jobs by default on top-level make

mx-psi:
#### Description

Uses 8 jobs by default which parallelizes some tasks like tidying, formatting...


mx-psi:
This works, but we would need to find a way to have color in the output

PR #10609, opened 2024-07-14
[chore] rely on strict input parsing to error out on invalid unsigned  int values

atoulme:
This removes a hook for confmap that was used to handle invalid uint values, now that https://github.com/open-telemetry/opentelemetry-collector/issues/9532 is fixed.


atoulme:
True, maybe need to wait for the feature gate to be stable. This TODO is going to block confmap 1.0.

mx-psi:
Blocked by #10793

mx-psi (approved):

PR #10632, opened 2024-07-17
[chore] Add readonly matrix test. Remove invalid mutability assertions

djaglowski:
The primary objective here was to add new test cases for the graph. However, I found that mutability assertions added in #8634 appear to be nondeterministic. Therefore, important test cases cannot be covered with them in place. This effectively removes the assertions about mutability for now.

@dmitryax, I'm curious if you have better ideas here. I am thinking that perhaps it would be better to have an entirely separate set of test cases which are focused on mutability expectations.


mx-psi (approved):

djaglowski:
Can we merge this?

mx-psi:
I was waiting on @dmitryax, I will merge since this is test only and we can change it easily

PR #10633, opened 2024-07-17
[consumererror]  Sketch solution for handling retries inside receivers 

evan-bradley:
This PR isn't intended to be merged, but just to serve as a forum for ongoing discussions in https://github.com/open-telemetry/opentelemetry-collector/pull/9041. In particular, I think it's important to see how the changes to consumererror will be used in our components.

Please note this doesn't compile. The changes necessary would be far-reaching and could result in a lot of churn if we change direction.

This PR is the product of some offline discussions @dmitryax and I had around the requirements for consumer errors. Previously, I was assuming that errors would be handled close to the source of the error (e.g. a processor or exporter), but was informed that we want to have receivers handle errors, which imposes additional requirements that aren't supported in #9041.

I've built this PR off #9041, but please take a look at this commit for how these changes would actually be used: https://github.com/open-telemetry/opentelemetry-collector/pull/10633/commits/07a31a9764602553c9b9002c507b2df07f4dab75.


receiver/otlpreceiver/internal/trace/otlp.go
```diff
@@ -49,7 +51,30 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
 	// NonPermanent errors will be converted to codes.Unavailable (equivalent to HTTP 503)
 	// Permanent errors will be converted to codes.InvalidArgument (equivalent to HTTP 400)
 	if err != nil {
-		return ptraceotlp.NewExportResponse(), errors.GetStatusFromError(err)
+		// See if an error is retryable.
+		cet := consumererror.Traces{}
+		if errors.As(err, &cet) {
```
evan-bradley, line 56: This only covers a single error right now. Navigating the error tree ourselves would be difficult, I think this shows the need for a single error type that contains all the information we want.

receiver/otlpreceiver/internal/trace/otlp.go
```diff
@@ -49,7 +51,30 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
...
+			retries := 3
+
+			// Retry three times; the real logic would be more complex akin to consumerretry.
+			for i := 0; i < retries; i++ {
```
evan-bradley, line 60: This is an extremely rudimentary reproduction of the consumerretry package, where I think this error parsing would ultimately live: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/consumerretry/logs.go#L61.

receiver/otlpreceiver/internal/trace/otlp.go
```diff
@@ -49,7 +51,30 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
...
+				// Re-use the data that was given to be retried.
+				td = cet.Data()
```
evan-bradley, line 62: We had discussed this before and determined that we want to take data from downstream since the data contained upstream is essentially just a pointer. In order for this to work possibly several components upstream with data that may have been mutated, we will need to find an unchanged copy for this to work.

receiver/otlpreceiver/internal/trace/otlp.go
```diff
@@ -49,7 +51,30 @@ func (r *Receiver) Export(ctx context.Context, req ptraceotlp.ExportRequest) (pt
...
+
+				// Update the context to include a key that indicates the component to be retried.
+				retryCtx := context.WithValue(ctx, consumererror.ComponentIDKey, cet.ID())
```
evan-bradley, line 65: I feel like the only reason we would want to use context here is because we don't want to change the existing APIs. I think an options pattern would make more sense going forward to apply component retry filters like this.

internal/fanoutconsumer/traces.go
```diff
@@ -8,34 +8,38 @@ import (
...
 	"go.opentelemetry.io/collector/pdata/ptrace"
 )
 
 // NewTraces wraps multiple trace consumers in a single one.
 // It fanouts the incoming data to all the consumers, and does smart routing:
 //   - Clones only to the consumer that needs to mutate the data.
 //   - If all consumers needs to mutate the data one will get the original mutable data.
-func NewTraces(tcs []consumer.Traces) consumer.Traces {
+func NewTraces(tcs map[component.ID]consumer.Traces) consumer.Traces {
 	// Don't wrap if there is only one non-mutating consumer.
```
evan-bradley, line 22: I put this in the fanoutconsumer because it's frequently used in the graph, but ideally we would have some universal wrapper that handles this for most components.

evan-bradley:
We discussed this during the 2024-07-17 stability WG meeting and come to a common agreement on next steps, so this PR has fulfilled it's purpose and can be closed. In short:
* We likely want to move to a single error type to propagate information upstream.
* We will cut down #9041 to a minimal scope while we investigate how to achieve more complex functionality such as retries.

Issue #9375, opened 2024-01-24
OTLP exporter/receiver v1 distribution


codeboten:
This issue is to capture all the steps needed to release a distribution of the OpenTelemetry Collector that includes an OTLP exporter and an OTLP receiver. This will include all the packages that are required to support this effort:

- [ ] exporter/otlp
- [ ] exporter/otlphttp
- [ ] receiver/otlp
- [ ] otel.io documentation
- [ ] blog post

PR #10643, opened 2024-07-17
[chore] Explicitly note that service and otelcol are not part of 1.0

mx-psi:
#### Description

Explicitly excludes service and otelcol from 1.0 efforts.

These modules are useful for distribution maintainers but not end-users, for which the 1.0 effort is targeted.

### DIFF
--- a/docs/ga-roadmap.md
+++ b/docs/ga-roadmap.md
@@ -21,6 +21,7 @@ All stabilized modules will conform to the API expectations outlined in the [VER
 Explicitly, the following are not in the scope of v1 for the purposes of this document:
 
 * stabilization of additional components/APIs needed by distribution maintainers. Vendors are not the audience
+  * This explicitly excludes the `service` and `otelcol` modules.
 * Collector Builder
 * telemetrygen
 * mdatagen
--- a/docs/ga-roadmap.md
+++ b/docs/ga-roadmap.md
@@ -21,7 +21,7 @@ All stabilized modules will conform to the API expectations outlined in the [VER
 Explicitly, the following are not in the scope of v1 for the purposes of this document:
 
 * stabilization of additional components/APIs needed by distribution maintainers. Vendors are not the audience
-  * This explicitly excludes the `service` and `otelcol` modules, for which we only aim to guarantee behavior stability.
+  * This explicitly excludes the `service` and `otelcol` modules, for which we will only guarantee that there are no breaking changes impacting end-users of the binary after 1.0, while Go API only changes will continue to be admissible until these modules are tagged as 1.0.
 * Collector Builder
 * telemetrygen
 * mdatagen
--- a/docs/ga-roadmap.md
+++ b/docs/ga-roadmap.md
@@ -21,7 +21,7 @@ All stabilized modules will conform to the API expectations outlined in the [VER
 Explicitly, the following are not in the scope of v1 for the purposes of this document:
 
 * stabilization of additional components/APIs needed by distribution maintainers. Vendors are not the audience
-  * This explicitly excludes the `service` and `otelcol` modules.
+  * This explicitly excludes the `service` and `otelcol` modules, for which we only aim to guarantee behavior stability.
 * Collector Builder
 * telemetrygen
 * mdatagen


mx-psi:
#9375 and the Collector v1 board would need an update if we merge this

mx-psi:
To add more context to this: the Collector 1.0 effort is focused on providing stability to end users. On the 2024-07-17 Collector Stability WG meeting we discussed the following that feels (to me at least) as a compelling argument for removing this from Collector 1.0 efforts:
- A Go module marked as 1.0 has two kinds of stability: symbol/Go API stability (not changing/removing public symbols that are part of the public API), and behavior stability. End-users of the binary care only about behavior stability.
- There is a link between Go API stability and Go module versioning in that the components of the Collector such as the OTLP receiver are published as a Go module and their Go module version matches their component version. Therefore we need to mark OTLP receiver and friends as 1.0 and provide both kinds of stability for these modules. Because of our VERSIONING.md requirements, we also need to mark their dependencies as 1.0.
- `service` and `otelcol` are not a dependency of any module. Therefore we don't necessarily need to provide Go API stability for these in order to call a Collector distribution 1.0. End users only care about behavior stability from these, which we can provide.

A question was raised about this in how do we ensure users understand the difference. The main goal and wording we can use for Collector 1.0 stability is to say that we guarantee behavior stability of the binary, but not Go API stability. This is what other projects such as Kubernetes have successfully communicated (`k8s.io` Go modules are 0.x but the binary is 1.x).

songy23 (commented):
> These modules are useful for distribution maintainers but not end-users, for which the 1.0 effort is targeted.

Agree with `otelcol`, but for `service` shouldn't the configs at least be stable for end users?

mx-psi:
> > These modules are useful for distribution maintainers but not end-users, for which the 1.0 effort is targeted.
> 
> Agree with `otelcol`, but for `service` shouldn't the configs at least be stable for end users?

I reworded to clarify that we will guarantee behavior stability

docs/ga-roadmap.md
```diff
@@ -21,7 +21,7 @@ All stabilized modules will conform to the API expectations outlined in the [VER
 Explicitly, the following are not in the scope of v1 for the purposes of this document:
 
 * stabilization of additional components/APIs needed by distribution maintainers. Vendors are not the audience
-  * This explicitly excludes the `service` and `otelcol` modules.
+  * This explicitly excludes the `service` and `otelcol` modules, for which we only aim to guarantee behavior stability.
```
songy23, line 24: Is there a reference that explains "behavior stability" further?

As I understand, a change in `service` config is considered a behavior change, while a change in Go API is not a behavior change, but somewhere in between things can become vague. E.g. what if a Go API change in `service` leads to change in how fatal error is reported etc.

mx-psi: What about the following phrasing?
> We will only guarantee that there are no breaking changes impacting end-users of the binary after 1.0, while Go API only changes will continue to be admissible until the modules are tagged as 1.0

Does this make more sense to you?

songy23: Thx, that makes more sense 👍

songy23 (approved):
LGTM

docs/ga-roadmap.md
```diff
@@ -21,6 +21,7 @@ All stabilized modules will conform to the API expectations outlined in the [VER
 Explicitly, the following are not in the scope of v1 for the purposes of this document:
 
 * stabilization of additional components/APIs needed by distribution maintainers. Vendors are not the audience
+  * This explicitly excludes the `service` and `otelcol` modules, for which we will only guarantee that there are no breaking changes impacting end-users of the binary after 1.0, while Go API only changes will continue to be admissible until these modules are tagged as 1.0.
```
TylerHelmuth, line 24: If we're trying to cut Go API stability from binary stability, are there other modules we could consider?

atoulme:
What is a good estimate of the work to get `service` and `otelcol` stable? If we don't do this for the Collector 1.0, when will it be done? What expectations should we set and should we care?

I want to understand the range of options we can consider here.

* What is the expectation of API to be offered by those modules, given that they participate in the binary behavior?
* Can we move most of the API of those modules to an internal module and expose virtually no API for this release?

docs/ga-roadmap.md
mx-psi: I don't think so because:
1. It would be confusing to release Collector 1.0 without marking the components in that distro as 1.0: end-users identify the version of the Go module with the version of the component.
2. Because of semver, we can't make a 1.0 module A depend on a 0.x module B (if B's types are part of A's public API). So we need to make (effectively) all dependencies of e.g. OTLP receiver 1.0 as well

This only leaves anything 'above' OTLP receiver and friends in the dependency tree as something we can consider out of scope IMO, which boils down to `otelcol` and `service`.

mx-psi:
> What is a good estimate of the work to get service and otelcol stable?

I think the main problem is that we don't have an estimate, because this part of the API has much fewer users than the others and as such we have not really built a backlog of issues to consider here.

>  If we don't do this for the Collector 1.0, when will it be done? 

I expect we will continue to work on stabilizing other modules after 1.0, but I don't think this would fall in the scope of the GA roadmap. As with anything in open source, it's hard to tell :smile: 

> What expectations should we set and should we care?

I don't understand this, you mean what expectations should we set regarding when `otelcol` and `service` will reach 1.0?

> What is the expectation of API to be offered by those modules, given that they participate in the binary behavior?

The functionality provided at the point where we announce Collector 1.0 will continue to be supported with the same behavior until a 2.0 Collector release is done (modulo fixing bugs, which needs a judgement call)

> Can we move most of the API of those modules to an internal module and expose virtually no API for this release?

It's hard to tell, but I think probably the answer is 'no' or at least 'not for most of it'. We have a bunch of users (vendor-backed distros, Jaeger, Agents that include OTel related functionality) of these APIs

docs/ga-roadmap.md
TylerHelmuth: > It would be confusing to release Collector 1.0 without marking the components in that distro as 1.0: end-users identify the version of the Go module with the version of the component.

I think we are going to run into the problem of 1.0 binaries/images with 0.x components regardless of any `service`/`otelcol` decision we make. We're covered by https://github.com/open-telemetry/oteps/blob/main/text/0232-maturity-of-otel.md#maturity-levels and the Java Agent already sets precedence for stable binaries/images that include 0.x instrumentation.

TylerHelmuth: I mean that we'll run into it for `Core` `Contrib` and `Kubernetes` 1.0 distributions.

TylerHelmuth (approved):

mx-psi:
cc @open-telemetry/collector-approvers  I will merge this on Wednesday EU morning unless anybody blocks by then

evan-bradley (approved):

atoulme (approved):
Thanks for answering my questions. I am ok with this approach, let’s deliver everything early and come back and deliver this next.

Kingcb1:
> 
> #### Description
> 
> 
> 
> Explicitly excludes service and otelcol from 1.0 efforts.
> 
> These modules are useful for distribution maintainers but not end-users, for which the 1.0 effort is targeted.
>

PR #10647, opened 2024-07-17
[processorhelper] update units for metrics

codeboten:
This updates units for processhelper's internal telemetry.

For this PR i updated `metadata.yaml` and ran `make gogenerate`, then `make -C processor test`


crobert-1 (commented):
Out of ignorance, what kind of change do we consider changing metric units? Do we need a changelog entry for this?

codeboten:
> Out of ignorance, what kind of change do we consider changing metric units? Do we need a changelog entry for this?

I believe we should have a changelog for this, will add one shortly

codeboten:
Changelog added, PTAL @crobert-1

crobert-1 (approved):

TylerHelmuth (approved):

PR #10658, opened 2024-07-17
[batchprocessor] Update metric units

crobert-1:
#### Description
`1` isn't an informative metric unit. I've made a best-guess attempt at updating metric units, f**eedback is very necessary here**, the units are not very clear here. I attempted to match the unit with metric descriptions.

[Semantic conventions for reference](https://opentelemetry.io/docs/specs/semconv/general/metrics/).

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10652


codeboten:
> Is there any way we can ensure consistency in the units used for all components? Maybe having a common list or guidelines about this?

I suspect we'll have better guidelines in the near future with https://github.com/open-telemetry/oteps/pull/259, but maybe we can start a doc in this repo until the otep lands?

codeboten:
Opened https://github.com/open-telemetry/opentelemetry-collector/pull/10695 to provide guidelines around units

codeboten (approved):

PR #10645, opened 2024-07-17
[telemetry] Use Command/Version to populate service name/version attributes

yurishkuro:
#### Description
Per title

#### Link to tracking issue
Fixes #10644

PR #10490, opened 2024-07-01
[telemetry] Add ability to set service.name for spans emitted by the Collector

jpkrohling:
#### Description
This PR bridges the config that is exposed to Collector users to the internal format expected by the config helpers around the tracer provider.

#### Link to tracking issue
Fixes #10489

#### Testing
Manual testing performed.

Config:
```yaml
receivers:
  otlp:
    protocols:
      http:
      grpc:

exporters:
  debug:

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug]
    metrics:
      receivers: [otlp]
      exporters: [debug]
    logs:
      receivers: [otlp]
      exporters: [debug]
  telemetry:
    traces:
      processors:
        - batch:
            schedule_delay: 1000
            exporter:
              otlp:
                endpoint: https://otlp-gateway-prod-eu-west-2.grafana.net/otlp/v1/traces
                protocol: http/protobuf
                headers:
                  Authorization: "Basic ..."
    metrics:
      level: detailed
      readers:
        - periodic:
            exporter:
              otlp:
                endpoint: https://otlp-gateway-prod-eu-west-2.grafana.net/otlp/v1/metrics
                protocol: http/protobuf
                headers:
                  Authorization: "Basic ..."
    resource:
      "service.name": "otelcol-own-telemetry"
```

Sent this telemetry to the collector: 
```console
telemetrygen traces --traces 1 --otlp-insecure --otlp-attributes='cookbook="own-telemetry"'
```

Resulting in this:

image

#### Documentation
None.

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

PR #10666, opened 2024-07-18
[chore] adding minimal test for attributes

codeboten:
This isn't as good as emitting telemetry w/ the service but at least it ensures resource attributes are set.

Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/10490 and https://github.com/open-telemetry/opentelemetry-collector/pull/10645


atoulme (approved):
I like tests!

PR #10670, opened 2024-07-19
[configgrpc] Send UNAUTHENTICATED on auth failure

### LINKED ISSUE #7646: `otlpreceiver` OTLP receiver is not throwing a valid `grpc-status` for authentication-related errors or in successful scenario

jpkrohling:
Fixes #7646

### DIFF
--- a/.chloggen/jpkroehling-grpc-statuscode.yaml
+++ b/.chloggen/jpkroehling-grpc-statuscode.yaml
@@ -1,0 +1,12 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: configgrpc
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: gRPC auth errors now return gRPC status code UNAUTHENTICATED (16)
+
+# One or more tracking issues or pull requests related to the change
+issues: [7646]
+
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -17,12 +17,14 @@ import (
 	"go.opentelemetry.io/otel"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/balancer"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
 	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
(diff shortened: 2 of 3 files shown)


crobert-1 (approved):

atoulme (approved):

mx-psi (approved):

evan-bradley (approved):
I wonder if we should return something like `INTERNAL` when we get a `404` or `5xx` status code from the auth server in cases where the issue is on the Collector's side and not anything related to the caller. But I don't see any good way to do that right now, and this aligns with how the HTTP interceptor works, so LGTM.

PR #10671, opened 2024-07-19
[extension/ballast] Remove the deprecated memory ballast extension

TylerHelmuth:
#### Description
This PR removes the deprecated memory ballast extension and all the logic in place in memorylimiter and service that was using it.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/8343. I don't want to close it until the [helm chart](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1268) is updated.

#### Testing
Unit tests


.chloggen/remove-ballast-extension.yaml
```diff
@@ -0,0 +1,25 @@
...
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: ballastextension
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Removes the deprecated ballastextension
+
+# One or more tracking issues or pull requests related to the change
+issues: [10671]
```
crobert-1, line 13:
```suggestion
issues: [8343]
```
Nit: I assume we should reference the original issue here?

cmd/otelcorecol/go.mod
```diff
@@ -4,7 +4,7 @@ module go.opentelemetry.io/collector/cmd/otelcorecol
 
 go 1.21.0
 
-toolchain go1.21.12
+toolchain go1.21.11
```
crobert-1, line 7:
```suggestion
toolchain go1.21.12
```
I assume this is unrelated to this PR and we should remove this change?

atoulme: yes that's a good assumption

.chloggen/remove-ballast-extension.yaml
TylerHelmuth: Our practice is to use the PR so that the changelog links to the exact commit that did the thing.  he description links to the issue so you can quickly get to it.

mx-psi (commented):
Can we possibly split this process in more steps?

For example, a first step before outright removal can be:

- Removing it from the Helm chart
- Removing support from it in the service (that is, fixing #8342) and documenting this
- Removing any mention of it on docs and examples on the opentelemetry org

Then a second step can be removing it from the official binaries while keeping a source-only release, and a final step would be removing the source code entirely.

I think this would solve our immediate issue and make the deprecation process more manageable (not all users would get this removed at the same time, so support requests are more easy to manage).

I know this is more annoying but 
- this extension has been here since forever
- we don't know which users have been listening to our deprecations

so I think we need to be gentle and divide this into steps

atoulme (approved):

songy23 (approved):

mx-psi (commented):
Have we seen any users complaining after open-telemetry/opentelemetry-collector-releases/pull/607? I see #10995, have we had any other reports?

If #10995 is the only one, I am comfortable merging this

mx-psi (approved):
Oki, then I would propose we wait until, say, Wednesday next week and merge if we don't get more reports

codeboten (approved):

PR #10674, opened 2024-07-20
[component] Restrict character set for component.ID name

### LINKED ISSUE #10673: [component] Restrict allowed characters for component.ID name
#9208 added restrictions to the allowed characters for component.ID `type`. While working on #10495 we discovered that there are not any restrictions on the `name` field. This issue proposes that we add similar restrictions to `name`.

mwear:
#### Description
While working on https://github.com/open-telemetry/opentelemetry-collector/pull/10495 we discovered that there are not any restrictions on the `name` field of a `component.ID`. There are restrictions on the `type` field introduced in #9208. This PR adds similar restrictions to `name`. 

A type must
- have at least one character,
- start with an ASCII alphabetic character and
- can only contain ASCII alphanumeric characters and '_'.

I found that we need a slightly different set of rules for name as some tests use a digit and others use a uuid as a name.  A name is still optional, but if it's provided it must:
- have at least one character,
- start with an ASCII alphanumeric character and
- can only contain ASCII alphanumeric characters, '_', and '-'.

I'd be willing to adjust these restrictions if anyone has any opinions on what should or should not be allowed.

#### Link to tracking issue
Fixes #10673

#### Testing
Unit tests

#### Documentation
Code comments


component/config.go
```diff
@@ -164,3 +164,51 @@ var (
...
+
+// MarshalText marshals the Name value.
+func (n Name) MarshalText() ([]byte, error) {
+	return []byte(n.value), nil
+}
+
+// nameRegexp is used to validate the name of a component.
+// A name must start with an ASCII alphanumeric character and
+// can only contain ASCII alphanumeric characters, '_', and '-'.
+var nameRegexp = regexp.MustCompile(`^[0-9a-zA-Z][0-9a-zA-Z_-]{0,62}$`)
```
axw, line 186: What do you think about opening this up to all Unicode letters and numbers? i.e. swapping `[0-9]` for `\pN` (numbers) and `[a-zA-Z]` for `\pL` (letters). Perhaps also replacing the hyphen and dash with `\pP` (punctuation).

I think ASCII is reasonable for type, since there's a specific set of them. On the other hand, the name component is arbitrary and user-defined. So, for example, allowing Chinese characters may be more inclusive of native Chinese speakers/readers/writers.

axw (commented):
Looks great, thank you! Just one thought about the allowed characters.

mx-psi (changes requested):
Can we start with introducing this restriction only on [`UnmarshalText`](https://github.com/open-telemetry/opentelemetry-collector/blob/a211cc8eccf94014752608fcad67a62d5ac1d479/component/identifiable.go#L63)?

We can later think about adding `Name` or not, but that is the part that impacts end-users of the binary and I think we can start there

mwear:
Updated with both suggestions @mx-psi and @axw.

axw (approved):
Thank you for the changes, looks great!

component/config.go
```diff
@@ -164,3 +164,15 @@ var (
 	// DataTypeLogs is the data type tag for logs.
 	DataTypeLogs = mustNewDataType("logs")
 )
+
+// nameRegexp is used to validate the name of a component.
+// A name must start with a Unicode alphanumeric character and
+// can only contain Unicode alphanumeric characters and punctuation.
+var nameRegexp = regexp.MustCompile(`^[\pN\pL][\pN\pL\pP]{0,62}$`)
```
mx-psi, line 171: Should we just follow https://www.unicode.org/reports/tr31/#Default_Identifier_Syntax?

mwear: I interpreted this as use the long form for category name. Let me know if that was what you had in mind or if it was something else.

axw: I took @mx-psi's comment to mean following the syntax `<Identifier> := <Start> <Continue>* (<Medial> <Continue>+)*`, where `<Start>`, `<Continue>`, and `<Medial>` are defined in that page in terms of various Unicode categories.

I like the idea of using a standard, but OTOH this one seems quite complicated. There's also the relatively simple https://www.unicode.org/reports/tr31/#Immutable_Identifier_Syntax. Essentially: disallow spaces, syntax, control characters. That might be good enough here?

If it's not straightforward, maybe we could start with something something strict (e.g. a subset of both of the above) and open an issue to expand the allowed names?

mx-psi: > I took @mx-psi's comment to mean following the syntax <Identifier> := <Start> <Continue>* (<Medial> <Continue>+)*, where <Start>, <Continue>, and <Medial> are defined in that page in terms of various Unicode categories.

Yes, that was it

>  Essentially: disallow spaces, syntax, control characters. That might be good enough here?

This sounds reasonable as a start

> a subset of both of the above

That also sounds good. On `component.ID` we did discuss Unicode support  and ended up restricting to ASCII-only for now. I think here there are other considerations here since the user should have full control over the name so
- it's more likely we break existing users if we restrict this to ASCII only
- users can stop using non-ASCII characters if needed (e.g. if sending telemetry related to the Collector itself to a backend that does not have full Unicode support)

mwear: I took another look at the default identifier. In addition to its complexity, it would prohibit a name from starting with a digit. This is something we have in our test cases (e.g. `traces/1`). I'm not sure how much it would be used in the wild. I took the least restrictive approach, which is to restrict the name to 1 to 63 Unicode characters excluding whitespace, control characters, and symbols.

mx-psi (approved):

component/config.go
mx-psi: Makes sense, thanks Matt!

TylerHelmuth (approved):

PR #10678, opened 2024-07-20
[chore] avoid flaky test by using port 0

### LINKED ISSUE #9772: [receiver/otlp] Flaky test: `TestCreateMetricReceiver/default`
--- FAIL: TestCreateMetricReceiver/default (0.00s)
        factory_test.go:234: 
            	Error Trace:	/home/runner/work/opentelemetry-collector/opentelemetry-collector/receiver/otlpreceiver/factory_test.go:234
            	Error:      	Received unexpected error:
            	            	listen tcp 127.0.0.1:42251: bind: address already in use
            	Test:       	TestCreateMetricReceiver/default
            	
https://github.com/open-telemetry/opentelemetry-collector/actions/runs/8299206581/job/22714400794?pr=9771

atoulme:
#### Description
Use port zero since we just start and stop the server.

#### Link to tracking issue
Fixes #9772


songy23 (approved):

receiver/otlpreceiver/factory_test.go
```diff
@@ -148,13 +148,13 @@ func TestCreateMetricReceiver(t *testing.T) {
 	factory := NewFactory()
 	defaultGRPCSettings := &configgrpc.ServerConfig{
 		NetAddr: confignet.AddrConfig{
-			Endpoint:  testutil.GetAvailableLocalAddress(t),
+			Endpoint:  "127.0.0.1:0",
```
codeboten, line 151: I believe this is already what `GetAvailableLocalAddress` no? https://github.com/open-telemetry/opentelemetry-collector/blob/9ef6356835b5dfb784f47f4455854136af3bac85/internal/testutil/testutil.go#L115

atoulme: this is safe once, but the same config is reused throughout, and there is a slight chance the port might be taken.

atoulme:
@codeboten please review?

codeboten (approved):
@atoulme are there other places in the code where this should be changed then?

atoulme:
> @atoulme are there other places in the code where this should be changed then?

yes, there is at least another flaky test that leverages and reuses the same port in between runs, but it specifically tests telemetry and makes calls to the server, which makes things more difficult to move over to this model.
See https://github.com/open-telemetry/opentelemetry-collector/blob/ee4eb85f58e9ed6765eb7a5296308408318af671/service/service_test.go#L308 specifically

PR #10684, opened 2024-07-21
[exporter/processor/receiver] Document factories

### LINKED ISSUE #9323: [component] component.go - Document what errors factoryFuncs can return
For component implementers, it would be good to document what and when a factoryFunc may return an error.

It seems like that `ErrDataTypeIsNotSupported` is explicitly called out as a potential error that can be reported. Instead of returning this error, can factoryFuncs return nil instead?

From lifecycle tests, we know ideally factoryFuncs should stay away from running validation of config or environment. Can we document this explicitly?

atoulme:
#### Description
Document factory functions to mention what error can be reported if the signal is not supported.

#### Link to tracking issue
Fixes #9323

### DIFF
--- a/exporter/internal/factory.go
+++ b/exporter/internal/factory.go
@@ -7,31 +7,33 @@ import (
 	"context"
 
 	"go.opentelemetry.io/collector/component"
-) // Factory is a factory interface for exporters.
+)
+
+// Factory is a factory interface for exporters.
 // This interface cannot be directly implemented. Implementations must
 // use the NewFactory to implement it.
 type Factory interface {
 	component.Factory
 
 	// CreateTracesExporter creates a TracesExporter based on this config.
-	// If the exporter type does not support tracing or if the config is not valid,
-	// an error will be returned instead.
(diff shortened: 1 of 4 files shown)


exporter/internal/factory.go
```diff
@@ -7,31 +7,33 @@ import (
...
 // This interface cannot be directly implemented. Implementations must
 // use the NewFactory to implement it.
 type Factory interface {
 	component.Factory
 
 	// CreateTracesExporter creates a TracesExporter based on this config.
-	// If the exporter type does not support tracing or if the config is not valid,
-	// an error will be returned instead.
+	// If the exporter type does not support tracing,
+	// this function returns the error component.ErrDataTypeIsNotSupported.
```
mx-psi, line 20: Add a doc link? (same for other mentions of this symbol)

atoulme: sorry, not familiar. What does that look like? My IDE resolves the reference so didn't know I had to do more

mx-psi (approved):

Issue #9478, opened 2024-02-06
[confighttp] Review usage of 'pointer to type'


mx-psi:
There are a few fields in `confighttp` structs that use a pointer to int instead of an int. This is an unusual pattern in configuration from components, so I think we should review why we use it and whether it is justified.

To fix this issue we need to:

- Define when we can use 'pointer to type' in configuration vs just type and relying on the zero value
- Review the fields on `confighttp` based on this.

bogdandrutu:
> There are a few fields in confighttp structs that use e.g. a pointer to int instead of an int. This is an unusual pattern in configuration from components, so I think we should review why we use it and whether it is justified.

Some background: Lots of these were added after the initial release of the confighttp, and authors/reviewers decide to use pointers to signal the presence of the field.

I do agree that this is not ideal, but would like to have a concrete option/design on how we would support adding new members to the config when we need to know the presence. I believe we can do some tricks with `Unmarshal`, but would like others to think about :)

bogdandrutu:
This reminds me on something. We should add a `NewDefaultFoo` for all configs, so that we can easily add new fields and set the right default value. 

So we can say that in order to guarantee "backwards compatible behavior" user should always initialize the config with default and overwrite anything as needed instead of creating a struct with "initial value" and leave others unset.

mx-psi:
+1 to using something like Optional, we can probably use `encoding.TextUnmarshaler` to make it work nicely with the configuration without resorting to custom unmarshaling.

@bogdandrutu can you open a separate issue for https://github.com/open-telemetry/opentelemetry-collector/issues/9478#issuecomment-1930475996 ?

atoulme:
I recommend we go at this in multiple PRs:
* [ ] add the defaults to the pointer types in the NewDefaultClientConfig function, so these defaults are always set. This changes none of the logic.
* [ ] Make everyone start to use `NewDefaultClientConfig` instead of creating their own struct to reduce the blast radius
* [ ] Break the API by moving away from pointer types.

mx-psi:
Thanks @mackjmr! Based on this and the discussions we have had offline, I also agree that using a pointer seems like the best option. I think to continue the discussion, we can open a PR on the [Coding guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#coding-guidelines) to settle that this is the way to represent optional stuff.

[...]

PR #10688, opened 2024-07-22
[chore] set default transport defaults on NewDefaultClientConfig

atoulme:
#### Description
Set the defaults on `NewDefaultClientConfig`. This change is a no-op since those defaults are set if the value is not set.

Add documentation to all fields as well.

#### Link to tracking issue
#9478


config/confighttp/confighttp.go
```diff
@@ -73,20 +76,20 @@ type ClientConfig struct {
...
 
 	// MaxConnsPerHost limits the total number of connections per host, including connections in the dialing,
 	// active, and idle states.
-	// There's an already set value, and we want to override it only if an explicit value provided
+	// By default, it is set to http.DefaultTransport.MaxConnsPerHost (0, unlimited).
 	MaxConnsPerHost *int `mapstructure:"max_conns_per_host"`
 
 	// IdleConnTimeout is the maximum amount of time a connection will remain open before closing itself.
-	// There's an already set value, and we want to override it only if an explicit value provided
+	// By default, it is set to http.DefaultTransport.IdleConnTimeout (90s)
```
mx-psi, line 92: I would remove the value in braces, we can add a [Docs link](https://go.dev/doc/comment#doclinks) instead so we don't have to keep the value updated

mx-psi (approved):

mx-psi:
There's some errors in the `httpforwarderextension`, could you take a look at those and see if we can fix them before merging this PR?

TylerHelmuth (approved):

atoulme:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34477 is open to unblock.

mx-psi:
```
=== FAIL: . TestConfig/elasticsearch/metric (re-run 1) (0.01s)
    config_test.go:319: 
        	Error Trace:	/tmp/opentelemetry-collector-contrib/exporter/elasticsearchexporter/config_test.go:319
```

Issue #9511, opened 2024-02-07
[component] Review `component.Host.GetFactory`


mx-psi:
`component.Host.GetFactory` is used by a single component: the receivercreator. While it may not be possible it would be great to consider if we can remove this or maybe make it optional (by having an optional interface that the service can implement). This would simplify the `Host` interface

mx-psi:
One example where we do the optional interface https://github.com/open-telemetry/opentelemetry-collector/blob/91f13c309d00eef5b997a2e810effb2a4ccd95d4/extension/zpagesextension/zpagesextension.go#L58-L66

TylerHelmuth:
Removing it from the `component.Host` interface and leaving it implemented on our implementation of `component.Host` seems fine to me.

On the other hand, the functionality is interesting and simple for any `component.Host` to implement, since it should already have all the component factories it is managing. Unlike https://github.com/open-telemetry/opentelemetry-collector/issues/7370 where `connectors` were the clear solution to components needing to call `GetExporters`, this method is the only way for 1 component to create another.

Since we only have 1 use case that utilizes this feature, I am in favor of the optional interface approach, but would be ok if we choose to keep it on the interface.

atoulme:
I think we could change the behavior so that the `GetFactory` method is available as a func as a receiver factory option. I'll try and draft something for review.

atoulme:
Here is the draft PR with a design experiment: https://github.com/open-telemetry/opentelemetry-collector/pull/10679

I changed the design a bit so you can only get access to receiver factories, hopefully that's fine given the only use case we have is about creating receivers.

mx-psi:
I think #10679 is interesting, but it also seems like something we can do after 1.0. If we are okay with removing this from `component.Host` and keeping on `service`, I think that's the bare minimum we can do now and we can look into something like #10679 afterwards

PR #10709, opened 2024-07-23
Deprecate component.Host.GetFactory

TylerHelmuth:
#### Description
This PR deprecates the `component.Host.GetFactory` interface. This has the benefit of keeping the `component.Host` interface as simple as possible.  Components that were relying on this method can instead check if the underlying `component.Host` implementation supports the interface.  An example of this pattern can be found here: https://github.com/open-telemetry/opentelemetry-collector/blob/91f13c309d00eef5b997a2e810effb2a4ccd95d4/extension/zpagesextension/zpagesextension.go#L58-L66

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/9511

### DIFF
--- a/.chloggen/component-deprecate-GetFactory.yaml
+++ b/.chloggen/component-deprecate-GetFactory.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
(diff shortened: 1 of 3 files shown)


mx-psi (approved):
Could we get a contrib PR doing explicit checking on `component.Host` to see if it implements this method?

mx-psi:
Ah, that would be open-telemetry/opentelemetry-collector-contrib/pull/34234

codeboten (approved):

Issue #5712, opened 2022-07-20
Add ocb Docker image


mx-psi:
**Is your feature request related to a problem? Please describe.**

Per open-telemetry/opentelemetry-collector-releases/pull/164, we want to promote usage of the OpenTelemetry Collector Builder so that people build their own distribution with only the components they use.

We currently only provide the binaries on each release ([example](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.55.0)), but we don't have a Docker image for this, which would ease the use of the builder.

**Describe the solution you'd like**

Provider a Docker image for the builder.

**Describe alternatives you've considered**

Do nothing, if we consider this is not going to get usage.

sakshi1215:
@dmitryax @jpkrohling As far as I could understand, we want to create a Dockerfile & release docker images for the collector builder, can I work on this if this is still relevant?

jpkrohling:
Sure, I think it would be worth the exercise. I'm not convinced this would work seamlessly for some Linux distributions (like Fedora), as selinux would prevent containers from writing to the host's file system, but I might be wrong.

frzifus:
@jpkrohling that should work using the `z` or `Z` mount option like here:
https://github.com/open-telemetry/opentelemetry-go-build-tools/issues/222

jackgopack4:
I see that we have docker images for the official releases, but I'm thinking it might still be useful to have docker images for custom builds; it seems like there's no way to build a custom collector using the [releases distributions](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions) Dockerfiles besides the default contrib, k8s, or core configuration

jackgopack4:
Hi there, i'd like to create another repository under the "otel" organization on DockerHub, I believe it could be called "otel/opentelemetry-collector-builder" or "otel/ocb"; who has the ability to create this? thanks.

mx-psi:
@jackgopack4 I'd guess this would be done through an issue in the [community repository](https://github.com/open-telemetry/community) but we should discuss what name we want first.

jackgopack4:
PRs are up:
https://github.com/open-telemetry/opentelemetry-collector-releases/pull/671
https://github.com/open-telemetry/opentelemetry-collector/pull/11234
Will appreciate reviews on either or both! I'm still working on a blurb for opentelemetry.io as well under "building a custom collector" but will likely get to it Monday

jackgopack4:
[...]

PR #10710, opened 2024-07-23
Migrate ocb binary release to opentelemetry-collector-releases (see corresponding PR#608 in opentelemetry-collector-releases repo)

### LINKED ISSUE #10662: Migrate release of OpenTelemetry Collector Builder (ocb) to the opentelemetry-collector-releases repository
Currently, the ocb binary is released in the opentelemetry-collector repository, as opposed to the rest of the binaries related to the collector, including core, contrib, and k8s distributions in the opentelemetry-collector-releases repository. Since many OpenTelemetry Collector users actually build their own custom opentelemetry collector using ocb, it may be confusing to find the binary in a different repository 

I would like to migrate the release workflows from the opentelemetry-collector repository to the opentelemetry-collector-releases repository. 

Alternatively, the workflows could remain inside the opentelemetry-collector repo but be modified to publish the binaries to the opentelemetry-collector-releases repository instead of the collector repository when the release train is run. I am not opposed to either solution, but will have to get more familiar with the actions and scripts involved with the build and release process (would love to dive into any documentation we have about this process)!

I intend to add a standard dockerfile for the ocb binary, in reference to [Issue 5712](https://github.com/open-telemetry/opentelemetry-collector/issues/5712), but in that ticket and the [subsequent PR](https://github.com/open-telemetry/opentelemetry-collector/pull/6451), it was generally agreed that the ocb binary should be released to the opentelemetry-collector-releases repository prior to adding dockerfile support.

Anyone available to point me in the right direction for the build/release process for the collector? Thanks!

jackgopack4:
**EDIT**: After discussion with the collector SIG call this week, July 31st, I have made small tweaks to the artifact release that will not require an "extra step" of pushing tags. If these changes are accepted, the distributions will be released under tag "vX.XXX.X" and the ocb binaries will be released under "cmd/builder/vX.XXX.X" but both under this repo, opentelemetry-collector-releases. The workflow `builder-release.yaml` in releases repo will push the additional tags necessary for the builder release.

Refactoring the release process as mentioned in past issues/PRs. ocb binary is currently released in a separate release tag under this repository, which makes it harder to find and inconsistent with the rest of the binaries and images releasing in the opentelemetry-collector-releases repository.

Releasing the ocb binaries in opentelemetry-collector-releases makes this process more standardized and also opens the opportunity to release more tools alongside ocb binary down the road, like corresponding docker images (as discussed in #5712).

#### Link to tracking issue
Fixes #10662
Mentioned in [releases issue # 295](https://github.com/open-telemetry/opentelemetry-collector-releases/issues/295)
[Corresponding PR in opentelemetry-collector-releases](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/608)

#### Testing
Sample binary release in [forked jackgopack4/opentelemetry-collector-releases repo](https://github.com/jackgopack4/opentelemetry-collector-releases/releases/tag/cmd%2Fbuilder%2Fv0.106.1)

#### Documentation
Updated corresponding documentation to reflect new release process


.github/workflows/sourcecode-release.yaml
jackgopack4, line 1: the Builder release action in opentelemetry-collector used to release both the ocb binaries and the otel-collector-core source code. Now it only releases the source code, hence the file name change

cmd/builder/.goreleaser.yml
jackgopack4, line 1: necessary to change the release repo for goreleaser

jackgopack4, line 1: moved to the opentelemetry-collector-releases repository

docs/release.md
```diff
@@ -73,19 +71,20 @@ It is possible that a core approver isn't a contrib approver. In that case, the
...
 
-1. Update the `./distribution/**/manifest.yaml` files to include the new release version.
+1. Update the `./distributions/**/manifest.yaml` files to include the new release version.
 
 2. Update the builder version in `OTELCOL_BUILDER_VERSION` to the new release in the `Makefile`. While this might not be strictly necessary for every release, this is a good practice.
 
-3. Create a pull request with the change and ensure the build completes successfully. See [example](https://github.com/open-telemetry/opentelemetry-collector-releases/pull/71).
+3. Create and push a new branch with this change, and apply the tag `cmd/builder/v0.xxx.0` to match the version number used above by running the command `make push-tags TAG=cmd/builder/v0.xxx.0`. This will build and release the ocb binaries under the opentelemetry-collector-releases distribution and is required to be completed **before** creating the pull request to merge the branch in the next step below. Building the distributions relies on using the latest ocb binary and if this workflow does not complete successfully, the rest of the build/release actions may not complete successfully.
```
evan-bradley, line 80: Is there any way we could specify the dependency in the workflows instead of requiring an additional tag? The release process is already fairly complex, it would be nice not to add any new steps.

evan-bradley (commented):
Thank you for taking this on.

docs/release.md
jackgopack4: I am open to suggestions and wanted to make this happen, but after looking at the release workflow (build on PR, build again on merge, then release on tag), I couldn't figure out a good way to separate the dependencies. Especially since the makefile simply downloads the prebuilt ocb release to build the different distributions. 
The other way to approach this might be using a token with cross-repo access to build the ocb release in the opentelemetry-collector repo but push it to the releases repo, but I wasn't certain if that was best practice security-wise and might make it more opaque which release is happening where.

evan-bradley: >Especially since the makefile simply downloads the prebuilt ocb release to build the different distributions.

This is a "dumb" approach, but here's how I would do it:
* Run something like `go build  go.opentelemetry.io/collector/cmd/builder@v0.106.0 -o ./bin/ocb` to build the binary.
* Use the `cache` action to cache the binary using the tag version (and maybe OS+arch) as the cache key.
* Release OCB using the binaries from the cache (or just pull them from the build results).
* Use the cached binary to build the Collector.

We would just need to link the workflows to ensure that OCB is built before the Collector. If necessary, I'd be okay migrating our separate workflows to jobs in a "Release" workflow to allow us to just link the jobs.

>Another option (not without its downsides) is to decouple the OCB release from the Collector distros release, and use OCB v.(N-1) to build Collector v.N.

I'd be okay with this as a backup if we can't do something like the above. It would be nice to continue to build with the latest OCB to let us take advantage of Collector API updates without waiting for an additional release if we can make it work.

jackgopack4: And running `go build` would only rely on the source code release that was generated in the `opentelemetry-collector` repository?

.github/workflows/check-goreleaser.yaml
jackgopack4, line 1: goreleaser file is now located in opentelemetry-collector-releases and the check functionality is migrated to workflows there (specifically builder-test.yaml)

docs/release.md
```diff
@@ -28,27 +28,25 @@ It is possible that a core approver isn't a contrib approver. In that case, the
 1. Update Contrib to use the latest in development version of Core by running `make update-otel` in Contrib root directory. This is to ensure that the latest core does not break contrib in any way. If it results in any changes, submit a PR to Contrib.
    -  🛑 **Do not move forward until this PR is merged.**
 
-2. Determine the version number that will be assigned to the release. Usually, we increment the minor version number and set the patch number to 0. In this document, we are using `v0.85.0` as the version to be released, following `v0.84.0`.
+1. Determine the version number that will be assigned to the release. Usually, we increment the minor version number and set the patch number to 0. In this document, we are using `v0.85.0` as the version to be released, following `v0.84.0`.
```
evan-bradley, line 31: I don't mind moving to numbering like this since it's easier to edit, but could we make this change in a separate PR? Including it in this one makes it harder to review.

docs/release.md
evan-bradley: >And running `go build` would only rely on the source code release that was generated in the `opentelemetry-collector` repository

It should, yeah. So long as you point it at the module, it will only use whatever we publish for that module.

>I wasn't certain if this was something that was handled by goreleaser, which in that case, would need to be run prior to building the collector distributions.

I'm still not sure how to point Goreleaser at a remote module, but at a minimum we should be able to do something similar to the approach you took in the -releases PR to just build it twice.

docs/release.md
jackgopack4: sure, I can do that.

jackgopack4: fixed

docs/release.md
jackgopack4: we publish the source code and corresponding go module here, and that gets pulled by the build script in releases. was able to publish a release in both forked repositories under jackgopack4/opentelemetry-collector and jackgopack4/opentelemetry-collector-releases

jackgopack4: @evan-bradley any further thoughts/concerns on this PR or the releases PR? Thanks.

.chloggen/ocb-migration.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: migrate build and release of ocb binaries to opentelemetry-collector-releases repository
+
+# One or more tracking issues or pull requests related to the change
+issues: [10710]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: ocb will be released under open-telemetry/opentelemetry-collector-releases tagged as "cmd/builder/vX.XXX.X"
```
mx-psi, line 18:
```suggestion
subtext: ocb binaries will now be released under open-telemetry/opentelemetry-collector-releases tagged as "cmd/builder/vX.XXX.X"
```

mx-psi (approved):

docs/release.md
```diff
@@ -86,12 +83,14 @@ The last step of the release process creates artifacts for the new version of th
...
 4. Check out the commit created by merging the PR and tag with the new release version by running the `make push-tags TAG=v0.85.0` command. If you set your remote using `https` you need to include `REMOTE=https://github.com/open-telemetry/opentelemetry-collector-releases.git` in each command. Wait for the new tag build to pass successfully.
 
-5. Ensure the "Release Core", "Release Contrib" and "Release k8s" actions pass, this will
+5. Ensure the "Release Core", "Release Contrib", "Release k8s", and "Builder - Release" actions pass, this will
 
     1. push new container images to `https://hub.docker.com/repository/docker/otel/opentelemetry-collector`, `https://hub.docker.com/repository/docker/otel/opentelemetry-collector-contrib` and `https://hub.docker.com/repository/docker/otel/opentelemetry-collector-k8s`
 
     2. create a Github release for the tag and push all the build artifacts to the Github release. See [example](https://github.com/open-telemetry/opentelemetry-collector-releases/actions/workflows/release-core.yaml).
 
+    3. build and release ocb binaries under a separate tagged Github release, e.g. `cmd/builder/v0.85.0`
```
evan-bradley, line 92: I'm open to other opinions, but I think I would like to keep everything under a single tag if possible. Right now we have all builder release notes grouped in with our main tag release, so users would grab the binary from one release but read the changelog in another, and that feels odd to me. I also feel like the advantages of having a single tag vs. separate tags are 50/50, so keeping a single tag would be simpler. 

The advantages I see are:
* Separate tags makes the ocb binary easier to find amongst the currently ~383 assets we publish. It also provides some logical separation between distro releases and other binary releases.
* A single tag means the release itself is easier to find (in my opinion) since the GH releases page feels geared toward a single stream of releases rather than concurrent releases of similar things. Also means the binaries and changelogs are together, and is a simpler setup.

evan-bradley (commented):
Thanks for your continued work @jackgopack4. I'm overall good with this, though I did leave a comment about how we should approach tagging.

I think we should aim to merge the -releases PR before this one just to be safe. If we don't have both make it into the release, I think it would be better to double publish the release than to not publish one.

docs/release.md
evan-bradley: Sorry for the delay, I was digging into it to see if there's a better way to approach building the source, since I'm also looking to add a binary to be released from -releases. I think that part is good to go, but I left a comment on the tagging I just want to make sure we're being deliberate about. I'll resolve this thread since I think the approach to building the binary is as good as we can do.

docs/release.md
mx-psi: Can we solve discoverability with some links pointing users to where the artifacts are located? that way we have both the separation (both logical and in terms of the artifacts) as well as discoverability, at the cost of one more click for some people

evan-bradley: I don't think the separation is super important since we essentially release everything as a single unit anyway, but that seems like a reasonable compromise.

evan-bradley (approved):

docs/release.md
jackgopack4: I agree that having everything released under one tag would be ideal; I didn't dig deep enough into how the partial goreleaser releases worked for the distributions to implement it for builder, and plus wanted to get others' opinions on if keeping the separate tags was smart, especially if we are adding other things to this release down the line like telemetrygen, etc.

evan-bradley: I'm still okay with a single tag. Everything is pretty tightly coupled, and I don't think the separate tags/releases will really be something we can leverage in the near future. Plus, adding ~8 binaries on top of the existing 383 doesn't really make a big difference, especially when a new Collector distro will presumably add 50+ new binaries anyway.

That said, I don't really feel strongly about this if we can link to it from the main release, and we can always change it later. I'll bring this up at tomorrow's SIG in case anyone has strong opinions.

songy23 (approved):
LGTM

cc release manager for the next release @dmitryax

dmitryax (approved):

PR #10715, opened 2024-07-24
[chore] delete code to set a version and date, as it it not used

atoulme:


cmd/builder/internal/command.go
```diff
@@ -123,7 +123,7 @@ configuration is provided, ocb will generate a default Collector.
 
 func initConfig(flags *flag.FlagSet) error {
 	cfg.Logger.Info("OpenTelemetry Collector Builder",
-		zap.String("version", version), zap.String("date", date))
```
atoulme, line 126: date is never set.

bogdandrutu (approved):

cmd/builder/internal/command.go
atoulme: yes, it wasn't used, so it had to go. If it's in usage, let's make sure it has tests and is exercised in this repository, and then it will be maintained and we can use it in the -releases repository. This might deserve its own issue in this repository.

Issue #10713, opened 2024-07-23
Unable to kafka receiver username config value to "$ConnectionString" with confmap.unifyEnvVarExpansion feature gate enabled


ben-childs-docusign:
**Describe the bug**

We have some config to connect to azure event hubs using the kafka receiver this requires us to set the user name to "$ConnectionString" per azure documentation:
https://learn.microsoft.com/en-us/azure/event-hubs/azure-event-hubs-kafka-overview#shared-access-signature-sas

Prior to this feature gate enabled we were able to do this by configuring the receiver as follows:

```
  config:
    receivers:
      kafka:
        auth:
          sasl:
            mechanism: PLAIN
            username: "$$ConnectionString"
          tls:
            insecure: false
        protocol_version: 1.0.0
        topic: metricstopic
```

With version 105 of the collector we get this error:
```
Error: failed to get config: cannot resolve the configuration: cannot convert the confmap.Conf: variable substitution using $VAR has been deprecated in favor of ${VAR} and ${env:VAR} - please update $ConnectionString or temporarily disable the confmap.unifyEnvVarExpansion feature gate
2024/07/23 21:58:26 collector server run finished with error: failed to get config: cannot resolve the configuration: cannot convert the confmap.Conf: variable substitution using $VAR has been deprecated in favor of ${VAR} and ${env:VAR} - please update $ConnectionString or temporarily disable the confmap.unifyEnvVarExpansion feature gate
```

**Steps to reproduce**

Configure kafka receiver as shown above (no need to even add it to a pipeline)

**What did you expect to see?**

The config is loaded successfully

**What did you see instead?**

The collector crashes on boot

**What version did you use?**

v0.105.0

**What config did you use?**

```
  config:
    receivers:
      kafka:
        auth:
          sasl:
            mechanism: PLAIN
            username: "$$ConnectionString"
          tls:
            insecure: false
        protocol_version: 1.0.0
        topic: metricstopic
```

**Environment**

Ubuntu 20.04 - go 1.21

**Additional context**

Does this regex need to be updated to allow for "$$" prefixed strings to be excluded? Or is there some other fix missing?
https://github.com/open-telemetry/opentelemetry-collector/blob/c239e73bbb83aec650866e15dcae3792d9b7279d/confmap/converter/expandconverter/expand.go#L94

Note that this PR proposes to fix this problem but I should have that fix since I'm using version 105:
https://github.com/open-telemetry/opentelemetry-collector/pull/10560

[...]

PR #10716, opened 2024-07-24
[confmap] Fix expansion of escaped environment variables

dmitryax:
This change fixes the issue where environment variables escaped with $$ are expanded. The collector now converts `$${ENV_VAR}` to `${ENV_VAR}` and `$$ENV_VAR` to `$ENV_VAR` without further expansion (or the expansion failure in case of `$ENV_VAR`).

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10713


atoulme (approved):

jvoravong (approved):

confmap/expand_test.go
```diff
@@ -583,21 +583,21 @@ func Test_EscapedEnvVars(t *testing.T) {
...
-			"recv.7":  "text$",
-			"recv.8":  "$",
-			"recv.9":  "${1}${env:2}",
-			"recv.10": "some${env:MAP_VALUE_4}text",
-			"recv.11": "${env:" + mapValue2 + "}",
-			"recv.12": "${env:${MAP_VALUE_2}}",
-			"recv.13": "env:MAP_VALUE_2}${MAP_VALUE_2}{",
-			"recv.14": "${env:MAP_VALUE_2${MAP_VALUE_2}",
-			"recv.15": "$" + mapValue2,
+			"recv.1":  "$$MAP_VALUE_1",
```
TylerHelmuth, line 586: Can you add tests in `confmap/internal/e2e` that tests all of these scenarios being passed to the expandconverter?

dmitryax: Added a test for cases where we don't expect failures. Do we need to keep your test Test_EscapedEnvVars around?

TylerHelmuth: I think we can remove this version of the test.  The new test will fail with the expand converter is removed, prompting us to add the escape logic back to Resolve.

TylerHelmuth (approved):

PR #10723, opened 2024-07-24
Deprecate expand converter

TylerHelmuth:
#### Link to tracking issue
Fixes #

### DIFF
--- a/confmap/converter/expandconverter/expand.go
+++ b/confmap/converter/expandconverter/expand.go
@@ -25,6 +25,8 @@ type converter struct {
 
 // NewFactory returns a factory for a  confmap.Converter,
 // which expands all environment variables for a given confmap.Conf.
+//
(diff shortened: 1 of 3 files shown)


confmap/converter/expandconverter/expand.go
```diff
@@ -36,6 +38,7 @@ func newConverter(set confmap.ConverterSettings) confmap.Converter {
 	}
 }
 
+// Deprecated: [v0.105.0] BASH-style env var expansion is deprecated. Use `${FOO}` or `${env:FOO}` instead.
```
dmitryax, line 41: Let's first move the `$$` escaping and ensure `${FOO}` expansion without it?

TylerHelmuth: @dmitryax I was testing something, but I believe my [original PR](https://github.com/open-telemetry/opentelemetry-collector/pull/10510) will be the one to go with

PR #10725, opened 2024-07-25
[component] Move component status features to its own module

TylerHelmuth:
#### Description

This PR is an attempt are removing component status reporting features from the component module. The technique was to make a new `componentstatus.StatusReporter` interface that `component.Host` could optionally implement. This has impacts:

1. We no longer need a `servicetelemetry.TelemetrySettings` struct.
2. Components no longer get access to a `ReportStatus` function until the `component.Start` method provides the host.
3. Components that use sharedcomponent will now only report a status for instances that have been started when an error in start occurs.  As an example if an otlpreceiver is used in a traces, metrics, and logs pipeline, when the host make the initial call to `Start`, say for the traces pipeline, sharedcomponent will only report the error returned from `Start` for the traces instance, and then continue on to shutdown.

If this solution's general idea is tenable, then I believe it could be broken down into slightly smaller PRs, such as:
1. Remove the need for `servicetelemetry.TelemetrySettings`: https://github.com/open-telemetry/opentelemetry-collector/pull/10728
2. Move status stuff into its own module: https://github.com/open-telemetry/opentelemetry-collector/pull/10730
3. Remove ReportStatus from TelemetrySettings and implement the `StatusReporter` interface: https://github.com/open-telemetry/opentelemetry-collector/pull/10777

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/pull/10413


internal/sharedcomponent/sharedcomponent.go
```diff
@@ -85,20 +74,72 @@ func (c *Component[V]) Unwrap() V {
...
+			host:           host,
+			sources:        make([]componentstatus.StatusReporter, 0),
+			previousEvents: make([]*componentstatus.StatusEvent, 0),
+		}
+	}
+
+	statusReporter, isStatusReporter := host.(componentstatus.StatusReporter)
+	if isStatusReporter {
+		c.hostWrapper.addSource(statusReporter)
+	}
```
TylerHelmuth, line 88: The new technique for sharedcomponent is to build the chain or reporting sources as Start is getting called.  Previously reported events are remembered so that if Start is called multiple times, each subsequent `StatusReporter` and get the previously emitted `StatusEvents`.

extension/extension.go
```diff
@@ -59,7 +60,7 @@ type StatusWatcher interface {
 	// Extensions that implement this interface must be ready that the ComponentStatusChanged
 	// may be called before, after or concurrently with calls to Component.Start() and Component.Shutdown().
 	// The function may be called concurrently with itself.
-	ComponentStatusChanged(source *component.InstanceID, event *component.StatusEvent)
+	ComponentStatusChanged(source *component.InstanceID, event *componentstatus.StatusEvent)
 }
```
mx-psi, line 64: Should this interface also be in componentstatus?

receiver/otlpreceiver/otlp.go
```diff
@@ -110,7 +111,10 @@ func (r *otlpReceiver) startGRPCServer(host component.Host) error {
 		defer r.shutdownWG.Done()
 
 		if errGrpc := r.serverGRPC.Serve(gln); errGrpc != nil && !errors.Is(errGrpc, grpc.ErrServerStopped) {
-			r.settings.ReportStatus(component.NewFatalErrorEvent(errGrpc))
+			statusReporter, ok := host.(componentstatus.StatusReporter)
+			if ok {
+				statusReporter.ReportStatus(componentstatus.NewFatalErrorEvent(errGrpc))
+			}
```
mx-psi, line 117: We could get some sort of wrapper like 

```go
func Report(host component.Host, e *componentstatus.StatusEvent) {
  statusReporter, ok := host.(componentstatus.StatusReporter)
  if ok {
    statusReporter.ReportStatus(e)
  }
}
```

to make this a bit more user-friendly (I think this was one of Matt's concerns)

mx-psi (commented):
This approach seems reasonable to me, I think the ergonomics are acceptable and it allows us to separate the status reporting mechanism from the rest of the `component` module

internal/sharedcomponent/sharedcomponent.go
djaglowski, line 88: Could all this be moved into `startOnce`? If so, I think it can be:

```suggestion
	if reporter, ok := host.(componentstatus.Reporter); ok {
		c.hostWrapper = &hostWrapper{
			host:           host,
			sources:        []componentstatus.Reporter{ reporter },
			previousEvents: make([]*componentstatus.Event, 0),
		}
	}
```

internal/sharedcomponent/sharedcomponent.go
```diff
@@ -85,20 +74,72 @@ func (c *Component[V]) Unwrap() V {
...
 	var err error
 	c.startOnce.Do(func() {
 		// It's important that status for a shared component is reported through its
 		// telemetry settings to keep status in sync and avoid race conditions. This logic duplicates
 		// and takes priority over the automated status reporting that happens in graph, making the
 		// status reporting in graph a no-op.
-		c.telemetry.ReportStatus(component.NewStatusEvent(component.StatusStarting))
-		if err = c.component.Start(ctx, host); err != nil {
-			c.telemetry.ReportStatus(component.NewPermanentErrorEvent(err))
+		if c.hostWrapper != nil {
```
djaglowski, line 96: It might be simpler to just always create a hostWrapper and then just let it's `ReportStatus` method do nothing as appropriate.

djaglowski (commented):
The approach looks good to me overall.

extension/extension.go
TylerHelmuth: I think that would make sense.  Since that would be a breaking change, I think it makes sense to add it in the PR that moves stuff to `componentstatus`, and deprecate the one in `service/extensions`.

djaglowski:
One other note - `component.InstanceID` was added for the sake of component status (though in theory it may be useful otherwise). I think we may want to move it out of `component` as well.

receiver/otlpreceiver/otlp.go
TylerHelmuth: Great idea, added.

component/componentstatus/status.go
```diff
@@ -1,12 +1,20 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
-package component // import "go.opentelemetry.io/collector/component"
+package componentstatus // import "go.opentelemetry.io/collector/component/componentstatus"
 
 import (
 	"time"
 )
 
+type StatusReporter interface {
```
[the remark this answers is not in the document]
TylerHelmuth: Our guidelines state:

```
Constant values of an enumeration should be prefixed with the enumeration type name in the name
```

So `componentstatus .StatusStarting` and `componentstatus .StatusOk` are correct since the type is named `Status`.  Should we rename the type to something else?

internal/sharedcomponent/sharedcomponent.go
TylerHelmuth: We can't move `c.hostWrapper.addSource(statusReporter)` into `startOnce` since that is the function that is building our chain of sources. For simplicity when calling `c.component.Start` I believe we always want to use `hostWrapper`, even when `component.Host` does not implement `componentstatus.Reporter`, so we shouldn't move the construction of `hostWrapper` behind a condition that requires `componentstatus.Reporter`.

internal/sharedcomponent/sharedcomponent.go
TylerHelmuth: We're actually doing that now, so it is safe to remove these extra checks

internal/sharedcomponent/sharedcomponent.go
djaglowski: Removing the condition makes sense if we're always instantiating the `hostWrapper`, but I'm not sure I understand why it would ever need to be instantiated more or less than once.

TylerHelmuth: I've tried reorganizing a little, does it seem better?

TylerHelmuth:
@djaglowski I've moved `InstanceID` to `componentstatus` (which in the actually implementation I'll handle by adding it when the content is moved to its own package and deprecating the one in `component`).

@mx-psi the fact that this move is possible makes me think _maybe_ https://github.com/open-telemetry/opentelemetry-collector/pull/10222 is back on the table, but I need to think about it more.

service/internal/graph/graph.go
```diff
@@ -578,3 +580,161 @@ func connectorStability(f connector.Factory, expType, recType component.Type) co
...
+
+// TODO: remove as part of https://github.com/open-telemetry/opentelemetry-collector/issues/7370 for service 1.0
+type getExporters interface {
+	GetExporters() map[component.DataType]map[component.ID]component.Component
+}
+
+var _ getExporters = (*Host)(nil)
+var _ component.Host = (*Host)(nil)
+
+type Host struct {
```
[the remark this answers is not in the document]
TylerHelmuth: I moved `host.go` from the root of the component dir to here bc I needed `graph` to be specific to our host instead of using `component.Host`.

I wanted access to our specific host so that I could access its `service/internal/status.Reporter` instance, which has some additional methods, like `ReportOKIfStarting` that the graph needs access to that `componentstatus.Reporter` does not have.

I could experiment with checking if the `component.Host` implements `service/internal/status.Reporter` instead of passing in our host directly.  That will reduce the number of changes, but require some extra interface checks in the graph's `StartAll`.

internal/sharedcomponent/sharedcomponent.go
TylerHelmuth: You're correct, but I included the nil check to act as a way to not call `c.hostWrapper.addSource(statusReporter)` a second time, since now it needs to be called within `startOnce`

service/internal/graph/graph.go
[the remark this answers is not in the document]
djaglowski: It seems odd to define it here long term, but I don't think it necessarily needs to be reworked immediately.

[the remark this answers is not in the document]
TylerHelmuth: I agree that it is odd.  Luckily it is all unexported/internal stuff so I agree that it could be reworked another time.

mwear (commented):
The sharedcomponent changes worry me a little bit. I'd like to see if I can test drive it with the health check extension to make sure it is still working as intended. Otherwise, this generally looks good to me.

TylerHelmuth:
@mwear the key change happens when a `component.Start` returns a synchronous error, which gets back to the host and ends up terminating the startup process.  The other signals for the sharedcomponent won't receive the `NewPermanent` error, but also they never receiver a Starting status either.  They get not statuses reported for them, like another component further down the line that also did not get started.

mwear:
What about the recoverable error case? When reporting status from within the component (I guess this is via host now), it should report status for all of the components it represents. All components represented by the shared component must be in the Starting state before the API can be from called from "within" the shared component. If you report RecoverableError during start, but some instances are not Starting, you will get an invalid state transition for the unstarted instances, and those instances will be stuck reporting an invalid status. I'm not sure if this makes sense, but this is my concern with this change. It's a little hard to test this end to end with the health check extension to verify because there are so many changes. In any case, if these words make any sense, and are accurate, we may have to handle this another way.

mwear:
It's possible that the way you are handling previous events solves my concerns. I'd still like to find a way to verify this end to end.

TylerHelmuth:
@mwear I'll see if I can add a test case in this PR with that example.

TylerHelmuth:
@mwear I've added a very rough e2e test in `internal/e2e/status_test.go` that tests that a component that is using sharedcomponent does end up eventually reporting all statuses for each instance and that an extension watching for statuses does receive all the events.

This gives me confidence that the statuses do end up in the extensions correctly. How the extension handles the statuses is up to the extension right?

One thing of note is that the existing (in main) graph/sharedinstance implementation runs into an invalid state in the graph's fsm because the graph reports `StatusStarting` for the current pipeline and the `sharedinstance` implementation also reports `StatusStarting` for all instances, one of which was already reported as `StatusStarting`.

mwear:
Thanks for the test @TylerHelmuth, this gives me more confidence. I'm aware of the repeat status reporting that was part of the original implementation. The finite state machine is meant to protect and enforce the lifecycle of a component, and invalid state transitions are not necessarily errors, in many cases should be considered a no-op. I know there is some logging around this that I would be in favor or removing because it gives the impression something unexpected happened, when that's not exactly the case. This is discussed in the [status reporting documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md) under the runtime section.

> During runtime a component should not have to keep track of its state. A component should report status as operations succeed or fail and the finite state machine will handle the rest. Changes in status will result in new status events being emitted. Repeat reports of the same status will no-op. Similarly, attempts to make an invalid state transition, such as PermanentError to OK, will have no effect.

TylerHelmuth:
@mwear do you feel comfortable enough to move forward with this approach via https://github.com/open-telemetry/opentelemetry-collector/pull/10730?

mwear:
My biggest reservation is that we need the collector to be observable for it to be 1.0. Part of this means having a functioning health check extension. The current replacement requires component status, so I don't think we can call the collector 1.0 without having component status enabled by default. If we are moving this code for organization purposes, that's fine, but I don't think we can skip this work for 1.0 altogether.

mx-psi:
> Part of this means having a functioning health check extension

The healthcheck extension is [not in our GA roadmap](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/ga-roadmap.md#proposal). I think it's important that we keep working on it, but our current plans don't include this for 1.0 as defined in the GA roadmap

TylerHelmuth:
Ive created https://github.com/open-telemetry/opentelemetry-collector/pull/10777 to complete the transition in core.

PR #10730, opened 2024-07-26
[component] Move component status reporting public API to new componentstatus module

TylerHelmuth:
#### Description

Duplicates component status reporting features from `component` into a separate module, `componentstatus`.  In a future PR, when `component.TelemetrySettings.ReportStatus` is removed, I'll update Core to depend on `componentstatus`.

This work isolates component status public API from `component` and `extensions`, which will allow us to move forward with their 1.0 work while component status reporting matures.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/pull/10725


atoulme (approved):
LGTM, just a bit dubitative, because this adds a new module to stabilize.

TylerHelmuth:
@atoulme the current plan is to not stabilize consumerstatus as part of 1.0 work. Component status reporting currently has not direct user configuration and no public api in service. Once ReportStatus and ComponentStatus Watcher are removed from component and extension respectively there won't be any public status reporting APIs in the modules we marked in scope.

atoulme:
I see, that’s an elegant way to narrow down what we need to stabilize. I appreciate the rationale. Looks good to me.

mx-psi (approved):
LGTM, but we can wait until we have solved the main questions on your draft PR

component/componentstatus/status_test.go
```diff
@@ -0,0 +1,330 @@
...
+				{}: NewEvent(StatusOK),
+			},
+			expectedStatus: StatusOK,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expectedStatus, AggregateStatus(tc.statusMap))
+		})
+	}
+}
```
mwear, line 138:
```suggestion
```

If we delete `AggregateStatus` we can also delete this test.

component/componentstatus/status_test.go
```diff
@@ -0,0 +1,330 @@
...
+				status:    StatusOK,
+				timestamp: maxTime,
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expectedStatus, AggregateEvent(tc.statusMap))
+		})
+	}
+}
```
mwear, line 330:
```suggestion
```
If we delete `AggregateEvent`, we can delete this test.

mwear:
> @atoulme the current plan is to not stabilize consumerstatus as part of 1.0 work. Component status reporting currently has not direct user configuration and no public api in service. Once ReportStatus and ComponentStatus Watcher are removed from component and extension respectively there won't be any public status reporting APIs in the modules we marked in scope.

As mentioned [here](https://github.com/open-telemetry/opentelemetry-collector/pull/10725#issuecomment-2257132196), I have serious reservations about this plan.

mx-psi:
I think we should move forward with this. The healtcheck extension is not in our GA roadmap, and in any case doing this unblocks progress in many other modules needed for 1.0.

component/componentstatus/status.go
```diff
@@ -0,0 +1,211 @@
...
+
+// Timestamp returns the timestamp associated with the Event
+func (ev *Event) Timestamp() time.Time {
+	return ev.timestamp
+}
+
+// NewEvent creates and returns a Event with the specified status and sets the timestamp
+// time.Now(). To set an error on the event for an error status use one of the dedicated
+// constructors (e.g. NewRecoverableErrorEvent, NewPermanentErrorEvent, NewFatalErrorEvent)
+func NewEvent(status Status) *Event {
```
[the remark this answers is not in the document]
TylerHelmuth: Ya lets do it as a follow up.

component/componentstatus/instance.go
```diff
@@ -0,0 +1,13 @@
...
+package componentstatus // import "go.opentelemetry.io/collector/component/componentstatus"
+
+import "go.opentelemetry.io/collector/component"
+
+// InstanceID uniquely identifies a component instance
+type InstanceID struct {
+	ID          component.ID
+	Kind        component.Kind
+	PipelineIDs map[component.ID]struct{}
+}
```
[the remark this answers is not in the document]
TylerHelmuth: Ya lets do it as a follow up.

component/componentstatus/status.go
```diff
@@ -0,0 +1,128 @@
...
+	return ev.status
+}
+
+// Err returns the error associated with the Event.
+func (ev *Event) Err() error {
+	return ev.err
+}
+
+// Timestamp returns the timestamp associated with the Event
+func (ev *Event) Timestamp() time.Time {
```
bogdandrutu, line 82: Timestamp seem to be use-less, because it is always "now", which means consumers (watchers) can also read if they need it.

component/componentstatus/status.go
```diff
@@ -0,0 +1,128 @@
...
+package componentstatus // import "go.opentelemetry.io/collector/component/componentstatus"
+
+import (
+	"time"
+)
+
+// Watcher is an extra interface for Extension hosted by the OpenTelemetry
+// Collector that is to be implemented by extensions interested in changes to component
+// status.
+type Watcher interface {
```
bogdandrutu, line 13: I would move this + the `InstanceId` to the "extension/statuswatcher` or something like that to be consistent with storage and other type of extensions.

component/componentstatus/status.go
mwear: This is important for the healthcheck extension, which also needs to synthesize events. When synthesizing events the timestamp needs to be settable. We used to do this in core, but have removed it, at least temporarily. I was proposing we add an option to set the timestamp in a followup: https://github.com/open-telemetry/opentelemetry-collector/pull/10730#discussion_r1696051237.

bogdandrutu: What about, if this is needed for that extension, we can have `StatusWithTime struct {Status, Time}` in the healthcheck and not here. No point in having it here.

bogdandrutu: Also, until we discuss the next PR, I would rather remove this for the moment.

TylerHelmuth: I prefer to keep this PR a 1-to-1 move of stuff related to component status reporting into `componentstatus`.  In my opinion we can have API discussions about `componentstatus` in future issues/PRs. I can update the godocs to state that everything in `componentstatus` is experimental and currently exempt from our deprecation process if you'd like.

component/componentstatus/status.go
TylerHelmuth: I think you're probably correct, but I'd prefer to encompass all public component status reporting features behind `componentstatus` for now while it is experimental and subject to lots of breaking API discussions.  Once we feel better about its API we could move it into `extension/storage` as a final location. Likely other aspects of `componentstatus` will move around in the future as well.

bogdandrutu: Not suggesting to `extension/storage` but to `extension/statuswatcher` :)

bogdandrutu: I am confused about why we do this then. Is this to just remove unstable code (that is ok)? But still you should at least document all these things as todos (file issues) and then we can do it in a separate PR.

component/componentstatus/status.go
bogdandrutu: > I prefer to keep this PR a 1-to-1 move of stuff related to component status reporting into componentstatus. In my opinion we can have API discussions about componentstatus in future issues/PRs. I can update the godocs to state that everything in componentstatus is experimental and currently exempt from our deprecation process if you'd like.

Then update the title to "move code" instead of adding a new module which to me means we can do a better job.

mwear: I think a timestamp will be useful for other status watchers and worth considering for the generated status event. Otherwise, if there are multiple watchers that care about the timestamp, each one will have to wrap the event in a custom struct and generate timestamps independently.

bogdandrutu: > I think a timestamp will be useful for other status watchers and worth considering for the generated status event. Otherwise, if there are multiple watchers that care about the timestamp, each one will have to wrap the event in a custom struct and generate timestamps independently.

When in doubt leave it out. I am not convinced yet, and until we have 2-3 use-cases we should not add it, since adding later is backwards compatible, but currently we pay the cost of reading the time as well as the maintenance cost.

component/componentstatus/status.go
TylerHelmuth: Removing unstable code from `component` and `extension` is the goal.  Issue created: https://github.com/open-telemetry/opentelemetry-collector/issues/10764

component/componentstatus/status.go
TylerHelmuth: Followup issue created: https://github.com/open-telemetry/opentelemetry-collector/issues/10763

mwear: I will still suggest the timestamp has value. I wrote the following on the issue:

> I believe that we should keep the timestamp as the observed time is not the same time as the observation. Since status watchers are dispatched in sequence, any status watcher that does a non-trivial amount of work, or something that could block (like writing to an event to a channel) could delay the observed time for subsequent watchers. This would be especially problematic if the delay was variable between invocations. Currently the timestamp is used by the health check extension for time calculations.

mx-psi:
AIUI all remaining issues are about the content of the module and not about the split itself so IMO we can go ahead with this. @bogdandrutu @mwear could you have a final look before I merge?

mwear (approved):
LGTM

bogdandrutu (approved):

PR #10733, opened 2024-07-29
Add a timeout alignment property to batch config

hcngo:
#### Description
Add a boolean timeout alignment property to batch processor config and batch sender config. This is to indicate if a batch submission should be aligned at the timeout boundary.

Dropbox is adopting Open Telemetry for their internal metrics system. The system uses 10s sampling rate and expects data submission to be more or less at the sampling rate boundary. This is to ensure metric data is ingested and streamed in time to various backend processes.

Without the timeout alignment feature, metric data can be submitted quite late. Consider the case when timeout is 10s and the batch processor happens to start at 00:00:07 mark. A request{ts=00:00:09} is sent to the batch processor. This request is associated with ts.truncateUp(samplingRate)=00:00:10 system timestamp. However, it is not until 00:00:07 + 10s = 00:00:17 that a batch is submitted. This data submission is quite late.

#### Link to tracking issue
No tracking issue #

#### Testing
Unit tests added to verify the batch processor and batch sender behavior when timeout alignment is used.

#### Documentation
Inline comment to indicate the usage.

### DIFF
--- a/exporter/exporterbatcher/config.go
+++ b/exporter/exporterbatcher/config.go
@@ -19,6 +19,9 @@ type Config struct {
 	// FlushTimeout sets the time after which a batch will be sent regardless of its size.
 	FlushTimeout time.Duration `mapstructure:"flush_timeout"`
 
+	// UseFlushTimeoutAlignment indicates if the flush timeout should be aligned at the timeout boundary.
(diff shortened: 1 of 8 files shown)


hcngo:
@evan-bradley Hi Evan, can I get a review on this PR please?

PR #10734, opened 2024-07-29
[chore] skip stable module check if unset

songy23:
From https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#releasing-opentelemetry-collector:

> If not intending to release stable modules, do not specify a version for Release candidate version stable.

`validate-versions` workflow fails on empty version right now, it should instead skip if stable module candidate version is empty.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -35,9 +35,6 @@ jobs:
           }
 
           validate_stable_version() {
-            if [[ -z "$1" ]]; then
-              exit 0
-            fi
             local regex_pattern_stable='^[1-9][0-9]*\.[0-9]+\.[0-9]+$'
             if [[ ! "$1" =~ $regex_pattern_stable ]]; then
               echo "Invalid stable version format for $2. Major version must be greater than 1."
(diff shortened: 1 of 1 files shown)


jackgopack4 (approved):
tested locally and this works. Ran into this issue recently testing changes to the build/release process for ocb, as well.

.github/workflows/prepare-release.yml
```diff
@@ -35,6 +35,9 @@ jobs:
           }
 
           validate_stable_version() {
+            if [[ -z "$1" ]]; then
```
codeboten, line 38: i would move this check to line 50 to make it easier to find that we're not doing anything with stable if the candidate-stable value is left blank

codeboten (commented):
This makes me wonder if we need the same check for beta values as well... i can imagine we may want to release stable and beta separately in the future

.github/workflows/prepare-release.yml
songy23: makes sense

codeboten (approved):

songy23:
I filed a tracking issue to further fix this script as there are different scenarios to consider

PR #10752, opened 2024-07-30
[configauth] Use correct mapstructure tag in confighttp Auth section

mx-psi:
#### Description

Adds fix to #10749


codeboten (approved):
Thanks @mx-psi

songy23 (approved):

PR #10781, opened 2024-08-01
Move receiver builder into internal service

dmathieu:
#### Description

This moves the receiver builder out of the `receiver` package, and into `service/internal/builders`.
There's no real reason for this struct to be public (folks shouldn't call it), and making it private will allow us to add profiling support to it.

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/pull/10375#pullrequestreview-2144929463


service/service.go
```diff
@@ -44,7 +45,8 @@ type Settings struct {
 	CollectorConf *confmap.Conf
 
 	// Receivers builder for receivers.
-	Receivers *receiver.Builder
+	ReceiversConfigs   map[component.ID]component.Config
+	ReceiversFactories map[component.Type]receiver.Factory
```
dmathieu, line 49: We can't provide a `Builders` option anymore here, since that's private. I've gone with the "easy" way of providing the attributes directly, but we could do things differently?

service/internal/builders/receiver.go
```diff
@@ -1,31 +1,41 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package receiver // import "go.opentelemetry.io/collector/receiver"
+package builders // import "go.opentelemetry.io/collector/service/internal/builders"
```
dmathieu, line 4: I've gone with a single package for all builders, but we could also have one package for each of them?

receiver/receivertest/nop_receiver.go
```diff
@@ -75,11 +75,3 @@ type nopReceiver struct {
 	component.StartFunc
 	component.ShutdownFunc
 }
-
-// NewNopBuilder returns a receiver.Builder that constructs nop receivers.
-func NewNopBuilder() *receiver.Builder {
```
dmitryax, line 80: This has to stay as deprecated before we can remove it. PLease see https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes

service/service.go
```diff
@@ -44,7 +45,14 @@ type Settings struct {
 	CollectorConf *confmap.Conf
 
 	// Receivers builder for receivers.
-	Receivers *receiver.Builder
+	//
+	// Deprecated: use the [ReceiversConfigs] and [ReceiversFactories] options
+	// instead.
+	Receivers builders.Receiver
```
dmitryax, line 51: This is a breaking change since it's a different type now

service/service.go
dmitryax: This new interface doesn't seem better than the builder fields.

Why can't we just add `CreateProfiles` to the existing builders instead?

cc @mx-psi

service/service.go
dmitryax: I believe we could make it not a breaking change if we make `builders.Receiver` an alias to `receiver.Builder`

service/service.go
dmitryax: Ok, after thinking more about it, it's probably good to remove the builders API as redundant anyway. I'm not sure that passing `ReceiversConfigs` and `ReceiversFactories` is the best option, but I don't have any other suggestions at this point

service/service.go
dmathieu: How is that a breaking change?
We're indeed changing it from a struct to an interface, but they match each other, so passing the struct will still work properly.

receiver/builder.go
```diff
@@ -13,13 +14,22 @@ import (
 	"go.opentelemetry.io/collector/consumer"
 )
 
-// Builder receiver is a helper struct that given a set of Configs and Factories helps with creating receivers.
+var errNilNextConsumer = errors.New("nil next Consumer")
+
+// Builder receiver is a helper struct that given a set of Configs and
+// Factories helps with creating receivers.
+//
+// Deprecated: this builder is being internalized within the service module,
```
mx-psi, line 22: Can you add the version this is being deprecated on like we did here: https://github.com/open-telemetry/opentelemetry-collector/commit/2b9697fe09616e1175adcea0d1362272af9b4aed#diff-628140520add1c6bc1e1a54def7b2255f03b89938fd72373b8ba848bbc28fbebR27 ?

receiver/builder.go
```diff
@@ -13,13 +14,22 @@ import (
...
+// and will be removed soon.
 type Builder struct {
 	cfgs      map[component.ID]component.Config
 	factories map[component.Type]Factory
 }
 
 // NewBuilder creates a new receiver.Builder to help with creating components form a set of configs and factories.
+//
+// Deprecated: this builder is being internalized within the service module,
```
mx-psi, line 31: ditto

service/service.go
mx-psi: I believe this is not a breaking change, @dmitryax could you show an example where this would break?

service/service.go
mx-psi: An alternative would be to add `ReceiversConfigs` to the `service.Config` struct. @dmitryax what do you think about that option?

service/internal/builders/receiver.go
mx-psi: Single package is fine for me

service/service.go
mx-psi: As pointed out by @dmathieu, since 	`service.Config` is used by `otelcol.Config` we can't add the receivers config to `service.Config`.

mx-psi: I think keeping in `Settings` is fine for now, we may want to refactor later but since `service` and `otelcol` Go APIs are out of scope for 1.0 this is not a pressing concern

mx-psi (approved):

service/service.go
dmitryax: Imagine if someone extracts the field value and expects a specific type there. That code will break. But that's fine I think. I missed that the new interface matches the old struct.

dmitryax (approved):

PR #10785, opened 2024-08-01
Move extension builder into internal service

dmathieu:
#### Description

This moves the connector builder out of the `connector` package, and into `service/internal/builders`.
There's no real reason for this struct to be public (folks shouldn't call it), and making it private will allow us to add profiling support to it.

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/pull/10375#pullrequestreview-2144929463

While this is not technically required for the profiles work (there is no notion of signals in extensions), this PR is here to keep things consistent.


atoulme:
Before we break the API, we typically deprecate it for at least one release. More here: https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes

mx-psi (approved):

PR #10786, opened 2024-08-01
[confmap] Validate providers scheme when building a Resolver

mx-psi:
#### Description

Validate providers schemes when building a Resolver.

I don't consider this a breaking change since the providers would be useless if they don't follow this pattern.

### DIFF
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -96,7 +96,17 @@ func NewResolver(set ResolverSettings) (*Resolver, error) {
 	providers := make(map[string]Provider, len(set.ProviderFactories))
 	for _, factory := range set.ProviderFactories {
 		provider := factory.Create(set.ProviderSettings)
-		providers[provider.Scheme()] = provider
+		scheme := provider.Scheme()
+		// Check that the scheme follows the pattern.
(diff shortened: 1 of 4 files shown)


confmap/resolver.go
```diff
@@ -96,7 +96,17 @@ func NewResolver(set ResolverSettings) (*Resolver, error) {
...
 		provider := factory.Create(set.ProviderSettings)
-		providers[provider.Scheme()] = provider
+		scheme := provider.Scheme()
+		// Check that the scheme follows the pattern.
+		if !regexp.MustCompile(schemePattern).MatchString(scheme) {
+			return nil, fmt.Errorf("invalid 'confmap.Provider' scheme %q", scheme)
+		}
+		// Check that the scheme is unique.
+		if _, ok := providers[scheme]; ok {
+			return nil, fmt.Errorf("duplicate 'confmap.Provider' scheme %q", scheme)
```
dmitryax, line 106: Should we cover this line?

dmitryax (approved):

confmap/resolver.go
mx-psi: Done

Issue #10787, opened 2024-08-01
Variables in comments in config files that are included are interpolated


antonio:
**Describe the bug**
String interpolation is being applied to comments in config files that are included from another config file, which can cause errors and prevent the collector from starting.

**Steps to reproduce**
Start the collector with the following config:

***config.yaml***
```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
processors:
  batch:

exporters:
  ${file:logging.yaml}

service:
  telemetry:
    metrics:
      level: detailed
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]
```

***logging.yaml***
```yaml
# ${hello.world}
logging:
  verbosity: detailed
```

**What did you expect to see?**
The collector should ignore the comment and carry on.

**What did you see instead?**
The collector tried to replace the variable in the comment with an environment variable that does not exist and exited with an error.

```sh
❯ ./otelcol validate --config=config.yaml
Error: failed to get config: cannot resolve the configuration: environment variable "hello.world" has invalid name: must match regex ^[a-zA-Z_][a-zA-Z0-9_]*$
2024/08/01 20:06:33 collector server run finished with error: failed to get config: cannot resolve the configuration: environment variable "hello.world" has invalid name: must match regex ^[a-zA-Z_][a-zA-Z0-9_]*$
```

**What version did you use?**
v0.106.1

**What config did you use?**
See above

**Environment**
I've manually replicated this on MacOS, go 1.22.5, but I've also seen it fail in CI (Ubuntu, different go versions) 

**Additional context**
Versions prior to v0.106.0 did not have this issue.

```sh
~/Code/opentelemetry-collector-contrib
❯ ./otelcol validate --config=config.yaml

~/Code/opentelemetry-collector-contrib
❯ echo $?
0

~/Code/opentelemetry-collector-contrib
❯ ./otelcol --version
otelcol version 0.105.0
```

mx-psi:
I believe this was caused by #10618 since we now try to parse the raw string as well in case we want to use the value as a string. The fix is to delay erroring out until we know that we have to use it as a string. I will mark this as release:blocker and fix it for v0.107.0

PR #10794, opened 2024-08-02
[confmap] Use values as string if YAML is invalid

### LINKED ISSUE #10759: Use Docker Secrets for Configuration Values
### Discussed in https://github.com/open-telemetry/opentelemetry-collector/discussions/10706

<div type='discussions-op-text'>

<sup>Originally posted by **JaronrH** July 23, 2024</sup>
I currently use Docker Swarm with Docker Secrets to store configuration information.  I would like to load information from Secrets into my configuration.  For example:

```
receivers:
  rabbitmq:
    endpoint: http://rabbitmq_rabbitmq:15672
    username: ${file:/run/secrets/RABBITMQ_USER}
    password: ${file:/run/secrets/RABBITMQ_PASSWORD}
    collection_interval: 10s
```

Docker Secrets are exposed as files in /run/secrets.  The problem (as far as I can tell - I'm not a Go dev) is that the ${file:} processor is trying to load those secrets as yaml files instead of values!  So, I'm getting errors like this as a result:

```
Error: failed to get config: cannot resolve the configuration: yaml: did not find expected node content
2024/07/23 16:08:53 collector server run finished with error: failed to get config: cannot resolve the configuration: yaml: did not find expected node content
```

Is there a better way to do this? Maybe some way to tell opentelemetry-collector-contrib to load everything in /run/secrets into env variables so I can use ${env:*}?

Thanks!</div>

mx-psi:
#### Description

If YAML parsing fails, assume the user wanted to pass the value as a string.

This has the downside that the error messages are less informative: it will tell you it expected something other than a string instead of the YAML parser error.

A future improvement could be to pass these errors down as extra metadata up until the unmarshaling stage.

#### Link to tracking issue

Fixes #10759

#### Testing

Added test case for this.

### DIFF
--- a/.chloggen/mx-psi_fix-invalid-yaml-string.yaml
+++ b/.chloggen/mx-psi_fix-invalid-yaml-string.yaml
@@ -1,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: If loading an invalid YAML string through a provider, use it verbatim instead of erroring out.
+
+# One or more tracking issues or pull requests related to the change
+issues: [10759]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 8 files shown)


TylerHelmuth (commented):
Can we add a test to replicate the issue in https://github.com/open-telemetry/opentelemetry-collector/issues/10787

TylerHelmuth:
NVM, I see the second PR

TylerHelmuth (approved):

dmitryax (approved):

PR #10803, opened 2024-08-06
Add `TrustedKeys` to TLS Config

### LINKED ISSUE #10523: Support trusting only a set of certificate hashes
When using publicly signed client certificates, we need some way to limit the keys that we trust (otherwise any publicly signed cert would work). This is generally achieved by specifying a set of SHA(1|256) fingerprints to trust (e.g. https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.16/administration-guide/56#tls-options-trusted-keys ).

I'd like to introduce a new option into configtls, `TrustedKeys []string`, which would introduce a `VerifyPeerCertificate` callback to the TLS config that rejects the TLS connection if the key hash is not in the list of trusted keys

We could use privately signed certs, but publicly signed client certs is a relatively common occurance

sinkingpoint:
#### Description

When using client certificates, it's sometimes useful to use publicly signed certificates to avoid
having to run your own CA infrastructure. This leaves you open though, because if you trust a public cert then _any_ public cert will work.

This alleviates this by allowing specifying a list of "Trusted Keys". This is an array of SHA256
fingerprints of certs that are trusted, allowing
us to reject any certs that are not the exact one we expect.

#### Link to tracking issue
Fixes #10523

#### Testing

Added two tests where we create a server that trusts a specific key, and then asserts that that key passes verification whereas the other cert doesn't.

#### Documentation

Added new option to README, along with an example

### DIFF
--- a/config/configtls/README.md
+++ b/config/configtls/README.md
@@ -72,6 +72,14 @@ Additionally certificates may be reloaded by setting the below configuration.
    Accepts a [duration string](https://pkg.go.dev/time#ParseDuration),
    valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 
+- `trusted_keys` (optional) : TrustedKeys allows specifying a list of SHA256 hashes to trust when verifying client certificates. This allows only trusting a specific set of keys, rather than all the keys signed by a given CA. Note: The client certificate still has to pass standard TLS verification, so you must still set the required CA values.
+
+Example:
+```
+  trusted_keys:
+    - "29:9D:39:8E:26:E7:72:88:D7:5D:2D:34:FE:8C:41:FA:9F:F1:C3:D7"
+```
+
 How TLS/mTLS is configured depends on whether configuring the client or server.
 See below for examples.
 
--- a/.chloggen/sinkingpoint_trusted-keys.yaml
+++ b/.chloggen/sinkingpoint_trusted-keys.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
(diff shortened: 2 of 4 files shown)


sinkingpoint:
@mx-psi Could I get your eyes on this when you get a sec?

config/configtls/README.md
```diff
@@ -72,6 +72,14 @@ Additionally certificates may be reloaded by setting the below configuration.
    Accepts a [duration string](https://pkg.go.dev/time#ParseDuration),
    valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 
+- `trusted_keys` (optional) : TrustedKeys allows specifying a list of SHA256 hashes to trust when verifying client certificates. This allows only trusting a specific set of keys, rather than all the keys signed by a given CA. Note: The client certificate still has to pass standard TLS verification, so you must still set the required CA values.
```
mx-psi, line 75: We don't necessarily need to implement this now, but if we want to change the hashing algorithm used here, how do you see that transition happening? I guess we could have something like this:
```yaml
tls:
  trusted_keys:
    hashing_function: sha256
    hashes:
      - "29:9D:39:8E:26:E7:72:88:D7:5D:2D:34:FE:8C:41:FA:9F:F1:C3:D7"
```

or instead each element of `tls::trusted_keys` could be something like this:

```yaml
tls:
  trusted_keys:
    - sha256: "29:9D:39:8E:26:E7:72:88:D7:5D:2D:34:FE:8C:41:FA:9F:F1:C3:D7"
```

What do you think is the best approach? I am leaning towards the latter option (it's a bit more flexible)

mx-psi (commented):
Hey, sorry for the delay.  I am interested in discussing two points:

1. Are you taking inspiration from a different application that does this? If so, how do they do it?
2. How are we going to support a different hashing algorithm in the future?

mx-psi:
Hey @namco1992 , feel free to open a PR. I still would want to explore if other applications do this and what's their UX like to make sure we are aligned with them if it makes sense

PR #10818, opened 2024-08-06
[component] Bump maximum length for component names to 1024 characters

### LINKED ISSUE #10816: Length restriction on component names seems overly restrictive
**Describe the bug**
We've recently update to v0.106.1 in our distro of the collector, and we've found that some of the names for components end up being too long for the new validation logic (#10674).

For example, we have this as the component name: `Linux-Messages-File_01J49HCH3SWFXRVASWFZFRT3J2__processor0__logs`

This give this error message:
```
cannot unmarshal the configuration: decoding failed due to the following error(s): error decoding 'processors': in "Linux-Messages-File_01J49HCH3SWFXRVASWFZFRT3J2__processor0__logs" id: invalid character(s) in name "Linux-Messages-File_01J49HCH3SWFXRVASWFZFRT3J2__processor0__logs" error decoding 'service.pipelines[logs/Linux-Messages-File_01J49HCH3SWFXRVASWFZFRT3J2__QRadar-HTTP-1].processors[3]': in "Linux-Messages-File_01J49HCH3SWFXRVASWFZFRT3J2__processor0__logs" id: invalid character(s) in name "Linux-Messages-File_01J49HCH3SWFXRVASWFZFRT3J2__processor0__logs"
```

For context; In BindPlane, we generate the names for components, which makes them verbose like this, and helps us to understand where the component is generated from (in terms of our own schema). So we'd really like to keep the longer component names, but it looks like the new validation has limited it to 63 characters.

I'm looking to understand why we need a character limit on the component name, and if it's possible to raise or remove this limit.

Also, related, the error message is a bit confusing. It claims `invalid character(s) in name` - but actually every character in the name is valid, it's just the length that's the problem; I think the length check should be separate from the regex, so we can give a proper error message here.

**What version did you use?**
v0.106.1

mx-psi:
#### Description

Bumps component name length limit to 1024 characters.

#### Link to tracking issue

Fixes #10816

#### Testing

Added some unit tests.


component/config.go
```diff
@@ -166,9 +166,9 @@ var (
 )
 
 // nameRegexp is used to validate the name of a component. A name can consist of
-// 1 to 63 unicode characters excluding whitespace, control characters, and
+// 1 to 127 unicode characters excluding whitespace, control characters, and
 // symbols.
-var nameRegexp = regexp.MustCompile(`^[^\pZ\pC\pS]{1,63}$`)
+var nameRegexp = regexp.MustCompile(`^[^\pZ\pC\pS]{1,127}$`)
```
TylerHelmuth, line 171: Since this is happening during startup, should the length check be broken our from the character check to improve the error message?

djaglowski:
I appreciate the quick action on this but I'm not sure this is meaningfully better.

As far as I am aware, [63 was suggested as the limit based on k8s](https://github.com/open-telemetry/opentelemetry-collector/issues/9872#issuecomment-2033399248). Is there a specific reason why we would settle on that or 127?

mx-psi:
>  Is there a specific reason why we would settle on that or 127?

Having *some* limit is useful so that we can know under what contexts we can use a `component.ID` (since there are systems that impose some length limit). Raising it to 127 means we drop compatibility with k8s, but compatibility with existing users seems more important.

I don't have a good justification for 127 other than keeping this number as small as possible makes our life easier in the future. We can raise the limit further if we get more user reports of things breaking with that limit, but lowering the limit will be much harder after 1.0 so I would like to keep a reasonable limit.

I don't think we (or at least, I) have a good criteria as to what the limit should be, but I think there is an argument for having some limit, and this seems like something that would work for the users that have reported issues with this so far.

mwear:
The main goal of this change was to prevent whitespace in the name. We didn't have a length requirement previously, it crept in by looking at what we were doing for the type. We can remove it entirely and add it later when and if we have a need for it.

andykellr:
I would suggest removing the limit until it produces a clear error message indicating that the length is the problem.

Once there is a clear error message, a limit could be added. I would support something higher than 128. 1024 sounds reasonable.

mx-psi:
@djaglowski 

> As a representative of the vendor affected by https://github.com/open-telemetry/opentelemetry-collector/issues/10816, I don't believe this is accurate. The autogenerated names are unbounded because there was previously no limit. Any limit is a technically a breaking change, though in practice a large limit should not be a problem.

Ah, I hadn't understood that the length was unbouded in your use case, sorry.

> Can we be less restrictive for now, say 1024

Done!

> and consider a stricter limit prior to 1.0 if someone articulates a use case?

This is okay to me. It's likely that if we get a use case it will be far in the future when this is much harder to change, but this sounds like the best we can do for now.

-----

@mwear 

> We can remove it entirely and add it later when and if we have a need for it.

After 1.0, we are likely to have a much harder time reducing this limit (this is a breaking change after all), so I prefer to start with a large limit if that is acceptable to everyone.

----

@andykellr 

> I would suggest removing the limit until it produces a clear error message indicating that the length is the problem.

I separated the two error messages. I think this should solve your concern.

BinaryFissionGames (approved):
Thanks for jumping on this!

djaglowski (approved):
Thanks @mx-psi!

codeboten (approved):

PR #10845, opened 2024-08-08
[mdatagen] export ScopeName in internal/metadata package

codeboten:
This can be used by components that need to set their scope name manually. Will save component owners from having to store a variable, which may diverge from the scope name used by the component for emitting its own telemetry. There are about a dozen components in contrib that could use this today.

Alternatively, I could just update the scope name strings in the components manually.

### DIFF
--- a/.chloggen/codeboten_mdatagen-expose-scope-name.yaml
+++ b/.chloggen/codeboten_mdatagen-expose-scope-name.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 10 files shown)


.chloggen/codeboten_mdatagen-expose-scope-name.yaml
```diff
@@ -0,0 +1,25 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: mdatagen
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: export ScopeName in internal/metadata package
+
+# One or more tracking issues or pull requests related to the change
+issues: []
```
codeboten, line 13:
```suggestion
issues: [10845]
```

evan-bradley (approved):

PR #10854, opened 2024-08-12
[chore] Refactor `service/host` into `service/internal/graph`

TylerHelmuth:
#### Description
This is a prep PR to reduce the size of https://github.com/open-telemetry/opentelemetry-collector/pull/10777.  As part of the work to make our `component.Host` implementation implement `componentstatus.Reporter` (see https://github.com/open-telemetry/opentelemetry-collector/pull/10852), the `host` struct and `graph` logic need to be closer together.  This is because, as part of https://github.com/open-telemetry/opentelemetry-collector/pull/10777 `StartAll` is changed to depend on our specific `Host` type instead of a `component.Host`.  Our host already has a dependency on `graph`, so it can't be moved into its own module.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/pull/10777
Related to https://github.com/open-telemetry/opentelemetry-collector/pull/10413


service/internal/graph/host.go
```diff
@@ -0,0 +1,164 @@
...
+
+// TODO: remove as part of https://github.com/open-telemetry/opentelemetry-collector/issues/7370 for service 1.0
+type getExporters interface {
+	GetExporters() map[component.DataType]map[component.ID]component.Component
+}
+
+var _ getExporters = (*Host)(nil)
+var _ component.Host = (*Host)(nil)
+
+type Host struct {
```
djaglowski, line 31: Rather than bring all these unrelated implementation details into an already very complicated package, could we move `Host` into a new package (e.g. `service/internal/host`) which both `graph` and `service` depend on?

TylerHelmuth: `Host` depends on `graph.Graph` so I don't think so.

TylerHelmuth: This PR would allow it, but then it would need moved again when `StartAll` is changed to take `Host` instead of `component.Host`.

TylerHelmuth: There is likely refactoring work that could be done to reorganize Host, Graph, and some shared stuff so that there isn't circular dependencies.  In an effort to keep these PRs smaller to make progress on https://github.com/open-telemetry/opentelemetry-collector/pull/10777 I'd prefer to look into this after https://github.com/open-telemetry/opentelemetry-collector/pull/10777 is merged.  Since these are internal packages we should be able to do that reorganization without breaking changes.

TylerHelmuth: Looking into it, I believe the solution would be to move `StartAll` and `ShutdownAll` off of `Graph` and instead make them take a `Graph` and a `Host`.  This mucks with a lot of tests, which would require some internal packages to exporter their stuff, but that would be safe and non-breaking.  Experimenting with it locally it is a lot of extra changes and I'm not actually sure it organizes things any better than leaving `Host` in `service/internal/graph`.

djaglowski: Thanks for looking into it. This sounds more like the direction I hope we can go eventually. I think the graph package can probably have a simpler interface so it doesn't have so much business logic. e.g. pass in a function that should execute against each node according to a topological sort. That said, I agree it's too much at the moment.

mx-psi (approved):

mwear (approved):

songy23 (approved):

PR #10859, opened 2024-08-12
[service] Allow users to disable the tracer provider via the feature gate `service.noopTracerProvider`

### LINKED ISSUE #10858: Increased request latency in 0.106.1
**Describe the bug**

_This issue was originally reported in CNCF slack by [Srinivas Bevara](https://cloud-native.slack.com/archives/C01N6P7KR6W/p1722488402533149) and confirmed by @codeboten who asked me to open this issue_

After Upgrade to v0.106.1 from v0.104.0 we have run into issues where we see an increase in request latency being reported by both the collector's internal telemetry and our ingress proxies. 

Memory remains unaffected however pprof is reporting garbage collection seems to have increased due internal telemetry tracing related allocations:

Image

Image

Image

Image

This is unexpected since tracing telemetry is disabled in our deployments. 

**Steps to reproduce**

Run a load test comparing v0.104.0 and v0.106.1 

- p99 Request latency and throughput should be comparable between both versions ~74ms
- Internal telemetry must not generate recording spans when it is disabled

**What did you see instead?**

- An increase in p99 latency to > 1s
- Profiling reveals increased objected allocations related to internal telemetry span data 
- Profiling reveals an increase in time spent in GC

**What version did you use?**

v0.106.1

**What config did you use?**

Our configuration has many transformations and connector pipelines. The bare minimum you may likely need is the one you use for benchmarking. 

**Environment**

Ubuntu 20.04.1

**Additional context**

Screen shots of increased p99 latency and CPU and steady memory 

Image

Image

Image

codeboten:
Previously the service was returning an instance of a SDK tracer provider regardless of whether there were any processors configured causing resources to be consumed unnecessarily.

Fixes #10858

### DIFF
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -130,6 +130,13 @@ type MetricsConfig struct {
 // TracesConfig exposes the common Telemetry configuration for collector's internal spans.
 // Experimental: *NOTE* this structure is subject to change or removal in the future.
 type TracesConfig struct {
+	// Level determines what level of tracing is enabled across the collector, the
+	// possible values are:
+	//  - "none" indicates that no trace data should be collected;
+	//  - "basic" is the recommended and covers the basics of the service telemetry.
+	//  - "normal" adds some additional spans top of basic.
+	//  - "detailed" adds all available spans to the previous levels.
+	Level configtelemetry.Level `mapstructure:"level"`
 	// Propagators is a list of TextMapPropagators from the supported propagators list. Currently,
 	// tracecontext and  b3 are supported. By default, the value is set to empty list and
 	// context propagation is disabled.
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -130,13 +130,6 @@ type MetricsConfig struct {
 // TracesConfig exposes the common Telemetry configuration for collector's internal spans.
 // Experimental: *NOTE* this structure is subject to change or removal in the future.
 type TracesConfig struct {
-	// Level determines what level of tracing is enabled across the collector, the
-	// possible values are:
-	//  - "none" indicates that no trace data should be collected;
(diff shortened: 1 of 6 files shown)


dmitryax (approved):

codeboten (commented):
The previous approach of using the configured `service::telemetry::trace::providers` wouldn't work because it would break the default behaviour of the zpages extension.

My suggestion in this PR is to move to a similar model as we will be moving to with metrics, where different meters can be configured for the different levels.

Alternatively, we could break the behaviour of the zpages extension, but that doesn't sound great either

service/telemetry/config.go
```diff
@@ -130,6 +130,13 @@ type MetricsConfig struct {
 // TracesConfig exposes the common Telemetry configuration for collector's internal spans.
 // Experimental: *NOTE* this structure is subject to change or removal in the future.
 type TracesConfig struct {
+	// Level determines what level of tracing is enabled across the collector, the
+	// possible values are:
+	//  - "none" indicates that no trace data should be collected;
+	//  - "basic" is the recommended and covers the basics of the service telemetry.
+	//  - "normal" adds some additional spans top of basic.
+	//  - "detailed" adds all available spans to the previous levels.
```
dmitryax, line 138: Are any of them utilized now?

Also, `configtelemetry.Level` defaults to Basic. Should we have None by default here?

codeboten:
@dmitryax I'm going to update this PR to use a feature gate as a short term solution for the original issue

dmitryax:
> @dmitryax I'm going to update this PR to use a feature gate as a short term solution for the original issue

Sounds good to me. Maybe we can figure out to make it work with the pages extension in as a follow up

TylerHelmuth (approved):

dmitryax (approved):

PR #10867, opened 2024-08-13
chore(builder): Update readme

mabdinur:
#### Description

Following the tldr section in the builder [README](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/builder#tldr) raises the following errors:

With v0.99.0:

```
Error: mismatch in go.mod and builder configuration versions: core collector version calculated by component dependencies "v0.107.0" does not match configured version "v0.106.1". Use --skip-strict-versioning to temporarily disable this check. This flag will be removed in a future minor version
```

With latest:
```
Error: invalid 'confmap.ResolverSettings' configuration: DefaultScheme not found in providers list
2024/08/13 06:55:50 collector server run finished with error: invalid 'confmap.ResolverSettings' configuration: DefaultScheme not found in providers list
```

Removing the `fileprovider` resolves the confmap failure on latest. This PR removes the `fileprovider` from the `otelcol-custom` binary. An alternatively we could update the `otelcol.yaml` to include the expected scheme. 

#### Link to tracking issue
Fixes #


ancostas (changes requested):
Pinning latest isn't great practice. Can we pin a version that works instead? Automatically updating this readme would also perhaps make sense if it needs to be updated in conjunction with other changes, and this avoids simply pinning latest.

mabdinur:
> Pinning latest isn't great practice. Can we pin a version that works instead? Automatically updating this readme would also perhaps make sense if it needs to be updated in conjunction with other changes, and this avoids simply pinning latest.


OpenTelemetry Collector binary is built using the latest changes on the [main branch](https://github.com/open-telemetry/opentelemetry-collector/tree/main). Pinning to latest will keep the core collector in-sync with other core components (ex: [debugexporter](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/debugexporter)) and reduce the likelihood of breaking changes. 

However you're right there is no need to set third-party components to latest. To fix the OCB docs I will pin all components to v0.107.0. We can revisit this discussion in the future.

cmd/builder/README.md
```diff
@@ -5,24 +5,21 @@ This program generates a custom OpenTelemetry Collector binary based on a given
 ## TL;DR
 
 ```console
-$ go install go.opentelemetry.io/collector/cmd/builder@latest
+$ go install go.opentelemetry.io/collector/cmd/builder@v0.107.0
```
mabdinur, line 8: @ancostas I will also pin the builder to `v0.107.0`.

.chloggen/munir_update-collector-builder-readme.yaml
```diff
@@ -0,0 +1,7 @@
+change_type: enhancement
```
songy23, line 1: I don't really think this needs a changelog since this is only doc change. You can remove this.

songy23 (approved):
LGTM

codeboten (approved):
@ancostas please review to ensure that your comments have been addressed

ancostas (approved):
LGTM, thanks Munir!

Issue #10870, opened 2024-08-13
[confighttp] Add max_redirects configuration option


rogercoll:
**Is your feature request related to a problem? Please describe.**

Components that use the `confighttp` to initialize an HTTP Client are not able to define the maximum number of redirection the client should perform. The provided client uses the default `CheckRedirect` http configuration: 

> 	// If CheckRedirect is nil, the Client uses its default policy,
	// which is to stop after 10 consecutive requests.

Use case: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34623

**Describe the solution you'd like**

Configuration option to limit the number of redirections, if non provided, the client should continue to work as it is (default redirections)

```yaml
endpoint: otelcol2:55690
    auth:
      authenticator: some-authenticator-extension
...
max_redirects: 0
```

rogercoll:
PoC: https://github.com/open-telemetry/opentelemetry-collector/pull/10877

PR #10877, opened 2024-08-13
feat(confighttp): add max_redirects configuration option

rogercoll:
#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/issues/10870

#### Testing
Unit testing

#### Documentation 
README.md


config/confighttp/confighttp.go
```diff
@@ -31,8 +31,11 @@ import (
 	"go.opentelemetry.io/collector/extension/auth"
 )
 
-const headerContentEncoding = "Content-Encoding"
-const defaultMaxRequestBodySize = 20 * 1024 * 1024 // 20MiB
+const (
```
atoulme, line 34: should this be a separate PR?

config/confighttp/confighttp_test.go
```diff
@@ -560,7 +649,6 @@ func TestHTTPServerWarning(t *testing.T) {
 			require.Len(t, observed.FilterLevelExact(zap.WarnLevel).All(), test.len)
 		})
 	}
-
```
atoulme, line 563: maybe minimize the diff

config/confighttp/confighttp_test.go
```diff
@@ -1307,7 +1395,6 @@ func TestServerWithDecoder(t *testing.T) {
 	srv.Handler.ServeHTTP(response, req)
 	// verify
 	assert.Equal(t, response.Result().StatusCode, http.StatusOK)
-
```
atoulme, line 1310: maybe minimize the diff

config/confighttp/confighttp_test.go
```diff
@@ -430,7 +518,8 @@ func TestHTTPClientSettingWithAuthConfig(t *testing.T) {
 			host: &mockHost{
 				ext: map[component.ID]component.Component{
 					mockID: &authtest.MockClient{
-						ResultRoundTripper: &customRoundTripper{}, MustError: true},
+						ResultRoundTripper: &customRoundTripper{}, MustError: true,
+					},
```
atoulme, line 522: maybe minimize the diff

config/confighttp/confighttp_test.go
```diff
@@ -219,7 +218,6 @@ func TestPartialHTTPClientSettings(t *testing.T) {
 			assert.EqualValues(t, 0, transport.MaxConnsPerHost)
 			assert.EqualValues(t, 90*time.Second, transport.IdleConnTimeout)
 			assert.EqualValues(t, false, transport.DisableKeepAlives)
-
```
atoulme, line 222: maybe minimize the diff

config/confighttp/confighttp_test.go
```diff
@@ -37,8 +37,7 @@ import (
 	"go.opentelemetry.io/collector/internal/localhostgate"
 )
 
-type customRoundTripper struct {
-}
+type customRoundTripper struct{}
```
atoulme, line 40: maybe minimize the diff

config/confighttp/confighttp_test.go
```diff
@@ -335,6 +333,96 @@ func TestHTTPClientSettingsError(t *testing.T) {
...
+			expectedRequests: 1,
+		},
+		{
+			name:             "Defined max redirects",
+			settings:         ClientConfig{MaxRedirects: toIntPtr(5)},
+			expectedRequests: 5,
+		},
+	}
+
+	host := &mockHost{
```
atoulme, line 365: maybe use componenttest.NewNopHost() ?

atoulme:
Changes look great, if you would please reduce the diff to avoid having git history be impacted too much.

config/confighttp/confighttp.go
rogercoll: sgtm 

fixed in https://github.com/open-telemetry/opentelemetry-collector/pull/10877/commits/26bd5939717bc483f02229319cdf1158863fd524

rogercoll:
@atoulme The lint CI is failing after removing the auto formatting changes:

```
confighttp_test.go:40: File is not `gofmt`-ed with `-s` (gofmt)
type customRoundTripper struct{
```

Should we fix them here? Or in a new PR?

rogercoll:
> The lint CI is failing after removing the auto formatting changes

My bad, it was because of a removed space in the diffs. Moving the PR ready for review

rogercoll:
@atoulme Any update on this PR? Any additional feedback would be really appreciated. Thanks for taking a look.

atoulme (approved):
LGTM, thanks for the edits

config/confighttp/confighttp.go
```diff
@@ -242,12 +245,31 @@ func (hcs *ClientConfig) ToClient(ctx context.Context, host component.Host, sett
...
+		Timeout:       hcs.Timeout,
+		Jar:           jar,
 	}, nil
 }
 
+// makeCheckRedirect checks if max redirects are exceeded
+func makeCheckRedirect(max *int) func(*http.Request, []*http.Request) error {
+	if max == nil {
+		return nil
+	} else if *max == 0 {
```
codeboten, line 259: Is this case needed here? if max is set to 0, will the check at line 266 return on the first call the same as is this block here?

codeboten (commented):
Thanks @rogercoll, just one question

config/confighttp/confighttp.go
rogercoll: Good point, the test case is needed because the length of the `via []*http.Request` variable on line 265 will be at least 1. This is because the `CheckRedirect` function of the standard htp client is always called after getting a redirect http response, so at least one request has been made:

```
	// If CheckRedirect is not nil, the client calls it before
	// following an HTTP redirect. The arguments req and via are
	// the upcoming request and the requests made already, oldest
	// first.
``` 

If you remove this conditional branch, the "Zero redirects" test case will fail (never end) as the length check will never match (number of already done request will be always > 0).

rogercoll: Great catch! Actually, this branch is not needed and resulted in the same output when `max_redirects` was set to either 0 or 1. The max redirects function is always called when a redirect response is returned, meaning that at least one request has already been made. The variable `via []*http.Request` on line 265 will always be > 0.

If `max_redirects` is set to 0, the http client should not follow any redirect response (one request). If `max_redirects` is 1, only one redirect response should be followed (2 total requests). `max_redirects = total_requests - 1`

Fixed in https://github.com/open-telemetry/opentelemetry-collector/pull/10877/commits/c98fe0a5fcffcc1994e58db23c5077e02cb2997c

Thanks!

config/confighttp/confighttp.go
```diff
@@ -242,12 +245,27 @@ func (hcs *ClientConfig) ToClient(ctx context.Context, host component.Host, sett
...
 }
 
+// makeCheckRedirect checks if max redirects are exceeded
+func makeCheckRedirect(max *int) func(*http.Request, []*http.Request) error {
+	if max == nil {
+		return nil
+	}
+
+	return func(_ *http.Request, via []*http.Request) error {
+		if *max == (len(via) - 1) {
```
dmitryax, line 262: Maybe a bit more intuitive. Let me know if you agree
```suggestion
		if len(via) > *max {
```

config/confighttp/confighttp.go
```diff
@@ -105,6 +105,9 @@ type ClientConfig struct {
 	HTTP2PingTimeout time.Duration `mapstructure:"http2_ping_timeout"`
 	// Cookies configures the cookie management of the HTTP client.
 	Cookies *CookiesConfig `mapstructure:"cookies"`
+
+	// Maximum number of redirections to follow, if not defined, the Client uses its default policy, which is to stop after 10 consecutive requests.
+	MaxRedirects *int `mapstructure:"max_redirects"`
```
dmitryax, line 110: I think we can even keep it `int` with `10` by default. That should be cleaner. The only problem is it'll be a significant breaking change (no redirects) for a component that doesn't use `NewDefaultClientConfig`, but I don't think we have any of those in contrib. Maybe we can check... All components should use that anyway. WDYT?

rogercoll: I am totally open to adding this, but I am concert about receivers initiating the config without using the default method (max_redirects will be set to 0). Also, the only difference is that maybe we should align with the Go library error message after 10 redirects:

```go
func defaultCheckRedirect(req *Request, via []*Request) error {
	if len(via) >= 10 {
		return errors.New("stopped after 10 redirects")
	}
	return nil
}
```

There is this ongoing effort to use the `NewDefaultClientConfig`: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35457

dmitryax: Different message isn't a problem I thought. I would like it to be merged without a pointer after all exporters use the default constructor. This is the pattern we use in other configs I believe.

rogercoll: Sounds reasonable, let's keep it pointer free :) https://github.com/open-telemetry/opentelemetry-collector/pull/10877/commits/e6b62bb678915f56f5b8c9ee1512c3d08757808e

Should we wait until https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35457 is done?

rogercoll: contrib tests failing due https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36681

jpkrohling (commented):
While I'm OK with the code, I believe this is the type of work that should be done _after_ v1.

rogercoll:
> While I'm OK with the code, I believe this is the type of work that should be done _after_ v1.

Is there any reference document/label that tracks features we want to add after v1?

PR #10888, opened 2024-08-14
[extension] Add ModuleInfo to extension.Settings

### LINKED ISSUE #10876: Make module info available to extensions
As described in https://github.com/open-telemetry/opentelemetry-collector/issues/10570, extensions should be able to make use of the same module information as is available via the `components` subcommand. This will be particularly useful for OpAMP.

dpaasman00:
#### Description
Paired with @djaglowski on adding a new struct, `ModuleInfo`, to `extension.Settings`. The goal is to make the collector's component go modules available to extensions that may want to use that information (in particular the OpAMP extension).

We didn't want to modify the `extension` package but adding this information to the Settings struct seems to be the most painless way to make this information available. No function signatures need to be updated and existing extensions can just ignore the new field on the Settings struct unless they want the information.

#### Link to tracking issue
Fixes #10876 

#### Testing
Updated existing tests.

### DIFF
--- a/extension/extension.go
+++ b/extension/extension.go
@@ -51,6 +51,15 @@ type ConfigWatcher interface {
 	NotifyConfig(ctx context.Context, conf *confmap.Conf) error
 }
 
+// ModuleInfo describes the go module for each component.
+type ModuleInfo struct {
+	Receiver  map[component.Type]string
+	Processor map[component.Type]string
+	Exporter  map[component.Type]string
+	Extension map[component.Type]string
+	Connector map[component.Type]string
+}
+
 // Settings is passed to Factory.Create(...) function.
 type Settings struct {
 	// ID returns the ID of the component that will be created.
@@ -60,6 +69,8 @@ type Settings struct {
 
 	// BuildInfo can be used by components for informational purposes
 	BuildInfo component.BuildInfo
+
+	ModuleInfo ModuleInfo
 }
 
 // CreateFunc is the equivalent of Factory.Create(...) function.
--- a/extension/extension.go
+++ b/extension/extension.go
@@ -70,6 +70,7 @@ type Settings struct {
 	// BuildInfo can be used by components for informational purposes
 	BuildInfo component.BuildInfo
 
+	// ModuleInfo describes the go module for each component.
 	ModuleInfo ModuleInfo
 }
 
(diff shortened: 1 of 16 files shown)


extension/extension_test.go
```diff
@@ -132,5 +132,6 @@ func createSettings(id component.ID) Settings {
 		ID:                id,
 		TelemetrySettings: componenttest.NewNopTelemetrySettings(),
 		BuildInfo:         component.NewDefaultBuildInfo(),
+		ModuleInfo:        ModuleInfo{},
```
djaglowski, line 135: It's not necessary to instantiate this since it's just the zero value.

djaglowski (commented):
Are there any test cases we can add? If nothing else maybe a test that demonstrates how it would be used in extension creation?

djaglowski:
cc: @open-telemetry/collector-maintainers

extension/extension.go
```diff
@@ -51,6 +51,15 @@ type ConfigWatcher interface {
 	NotifyConfig(ctx context.Context, conf *confmap.Conf) error
 }
 
+// ModuleInfo describes the go module for each component.
+type ModuleInfo struct {
+	Receiver  map[component.Type]string
```
mx-psi, line 56: Should this have more structure? For example, should we use https://pkg.go.dev/golang.org/x/mod@v0.20.0/module#Version (or something similar) instead? 

I am a bit hesitant about using `module.Version` directly since that module has a 0.x version, but something similar may make sense

djaglowski: I'm not opposed to structuring it more but it wasn't clear to me either what that structure should be. I think it's worth noting that the module string is something of a standard in its own right, since its interpretable by go tooling.

mx-psi: > I'm not opposed to structuring it more but it wasn't clear to me either what that structure should be

Hm, I guess a light wrapper around the `module.Version` I linked above makes sense. 

But thinking about it more I remembered open-telemetry/opentelemetry-collector-contrib/issues/31136, where we put `int` as the type for a stepping number, that is commonly represented as a string, and then found out that there are edge cases where it's not representable as an int. 

I guess a wrapper would make this situation manageable, but the lesson I learned there is that using a string may be better when you don't control the evolution of what you are representing.

This is all to say: if you feel like a string works, let's go with a string :)

mx-psi (approved):

PR #10889, opened 2024-08-14
chore (deps): migrate rand to rand/v2

### LINKED ISSUE #10885: Migrate to math/rand/v2
Go 1.22 introduced [math/rand/v2](https://pkg.go.dev/math/rand/v2) which replaces the `math/rand` stdlib package. We should migrate to it!

Once we have migrated, we can forbid its usage via [depguard](https://github.com/open-telemetry/opentelemetry-collector/blob/d5d1f82bc728d31d4e757f1c1573e754c335ebd1/.golangci.yml#L103).

Components using `math/rand` v1:

- [x] exporter/exportertest
- [x] receiver/receivertest

(generated using `rg math/rand -l -tgo | cut -d'/' -f1-2 | sort | uniq | sed 's/^/- [ ] /'`)

Counterpart to open-telemetry/opentelemetry-collector-contrib/issues/34676.

webstradev:
#### Description
Migrated rand dependency to rand/v2. We only use rand in tests and mocks and the only two functions we use are float32 and float64 which still exist in the new v2 API so nothing very exciting going on here.

#### Link to tracking issue
Fixes #10885

#### Testing
Ran `make test` successfully

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -110,6 +110,8 @@ linters-settings:
             desc: "Use 'errors' or 'fmt' instead of github.com/pkg/errors"
           - pkg: github.com/hashicorp/go-multierror
             desc: "Use go.uber.org/multierr instead of github.com/hashicorp/go-multierror"
+          - pkg: "math/rand$"
+            desc: "Use the newer 'math/rand/v2' instead of math/rand"
         # Add a different guard rule so that we can ignore tests.
       ignore-in-test:
           deny:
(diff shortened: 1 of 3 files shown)


mx-psi (approved):
Thank you!

mx-psi:
@webstradev would you be willing to also update the depguard rules? Can be done in this PR if you want to :)

webstradev:
> @webstradev would you be willing to also update the depguard rules? Can be done in this PR if you want to :)

Yep will do!

webstradev:
Making a small fix for the depguard rule, needs a $ for the specific version.

webstradev:
> Making a small fix for the depguard rule, needs a $ for the specific version.

@mx-psi Should be good to go now

mx-psi:
Thanks!

PR #10892, opened 2024-08-14
[service] add level for trace configuration

### LINKED ISSUE #10890: [telemetry] add `service::telemetry::trace::level` to allow enabling/disabling of tracer provider
In https://github.com/open-telemetry/opentelemetry-collector/pull/10859, a feature gate was added to allow users to use a noop tracer provider. This is not a great user experience and making the noop tracer provider is not an option if we want to prevent breaking backwards compatibility with the zpages extension.

The proposed solution is to allow the configuration of traces in the same way as we do for metrics, via a `level` option. The only values supported by this level is `none` or `basic` until there is a need for additional levels.

```
service:
  telemetry:
    traces:
      level: none
```

codeboten:
This allows users to configure a Noop TracerProvider if needed. Note that this configuration option breaks the zpages extension, which relies on a SDK implementation.

Fixes #10890


.chloggen/codeboten_trace-level.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type:
```
codeboten, line 4:
```suggestion
change_type: enhancement
```

.chloggen/codeboten_trace-level.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component:
```
codeboten, line 7:
```suggestion
component: service
```

iblancasa:
Please, can you elaborate a little bit more on this?
> Note that this configuration option breaks the zpages extension, which relies on a SDK implementation.

.chloggen/codeboten_trace-level.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Adds `level` configuration option to `service::telemetry::trace` to allow users to disable the default TracerProvider"
+
+# One or more tracking issues or pull requests related to the change
+issues: [10892]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: "This replaces the feature gate `service.noopTracerProvider` introduced in v0.107.0"
```
iblancasa, line 18: Should the "breaking change" with `zpages` be mentioned here?

codeboten: This will not be a breaking change by default, as the default value will continue to generate a TracerProvider that doesn't break zpages.

codeboten:
> Please, can you elaborate a little bit more on this?
> 
> > Note that this configuration option breaks the zpages extension, which relies on a SDK implementation.

The zPages extension registers itself as a span processor to be able to record telemetry about all the spans emitted by the collector. This relies on the code here: https://github.com/open-telemetry/opentelemetry-collector/blob/d2ed276a9215053790ddead2d6479a6beacd28a7/extension/zpagesextension/zpagesextension.go#L52-L58

That code has always worked with the default configuration because there was no way to disable the tracer provider until the introduction of the feature gate this change is replacing. With a no-op tracer provider, the zpages will be unable to register itself and log the following message: `zPages span processor registration is not available`

iblancasa:
Thanks a lot!

TylerHelmuth (approved):
For the sake of consistency with `MetricsConfig` I understand why we're using `Level` here, but my mind just wants a boolean `enabled` option instead. I don't feel strongly enough to block this work.

service/telemetry/tracer.go
```diff
@@ -48,9 +49,10 @@ func attributes(set Settings, cfg Config) map[string]interface{} {
 
 // New creates a new Telemetry from Config.
 func newTracerProvider(ctx context.Context, set Settings, cfg Config) (trace.TracerProvider, error) {
-	if globalgates.NoopTracerProvider.IsEnabled() {
+	if globalgates.NoopTracerProvider.IsEnabled() || cfg.Traces.Level == configtelemetry.LevelNone {
```
songy23, line 52: Should it be documented somewhere that level:none breaks zpage?

.chloggen/codeboten_trace-level.yaml
songy23, line 18: nit: deprecation of this feature gate should be a breaking change

songy23 (approved):
Some nits, LGTM otherwise

.chloggen/codeboten_trace-level.yaml
codeboten: Right, this isn't deprecating the feature gate yet :) i will do that in the following release

service/telemetry/tracer.go
codeboten: I can add a note in the zpages documentation

iblancasa (approved):

PR #10896, opened 2024-08-16
fix replace words in doc

akshaychopra5207:
corrects the description by replacing words in readme


songy23 (approved):
This makes more sense to me

PR #10907, opened 2024-08-17
Support default values and erorrs for env var expansion

### LINKED ISSUE #5228: Support default values for configuration environment variables

When using [Configuration Environment Variables](https://opentelemetry.io/docs/collector/configuration/#configuration-environment-variables), 

**if a variable is NOT SET on otelcol start, the value will be empty and probably break configuration.**

Please provide a way to set *default values* in case ENV variables are not available.

Allow definition of a default value if an ENV variable is not set. Via `${VAR:default_value}` , example:

```yaml
processors:
  attributes/example:
    actions:
      - key: "${DB_KEY:CustomerDefaultKey}"
        action: "${OPERATION:CustomerDefaultOperation}"
```

This mimics the syntax of [grafana promtail](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#use-environment-variables-in-the-configuration)

Another example is `docker-compose` , where *"Default values can be defined inline using typical shell syntax"*
https://github.com/docker/docker.github.io/blob/db650b4772f605683ba4143d0b0de44d0a187db4/compose/compose-file/index.md#interpolation

- `${VARIABLE:-default}` evaluates to `default` if `VARIABLE` is unset or
  empty in the environment.
- `${VARIABLE-default}` evaluates to `default` only if `VARIABLE` is unset
  in the environment.

and more, which is more complex however.

For example, grafana allows to [Override configuration with environment variables](https://grafana.com/docs/grafana/latest/administration/configuration/#override-configuration-with-environment-variables)

For OTEL Collector, this could look like this. Given this configuration file (section):

```yaml
exporters:
  jaeger:
    endpoint: my_default_server:14250
```

Providing an environment variable crafted like so:

```
export OTELCOL_EXPORTERS_JAEGER_ENDPOINT=some_other_server:23456
```

Could overwrite the value from default configuration or configuration files.

(Grafana is supporting [*even more variable substituation ways*](https://grafana.com/docs/grafana/latest/administration/configuration/#variable-expansion), but this is certainly overkill for OTEL Collector.)

At the moment, the configuration file (including ENV vars within the file) is the only way to configure OTEL Collector (?), which may be a deliberate design choice. It is not possible to overwrite specific configuration keys by using:

- CLI Flags
- ENV variables directly addressing those keys (so described approach above)

Which is not bad by all means, since **mixing all those possiblities creates complex and confusing configuration**.

However ENV variables and CLI Flags are easier to use when running OtelCol as a container, in comparison to a config file. When not running on kubernetes [opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator) is no alternative.

yurishkuro:
#### Description
Support shell-style default value and error message for env var expansion.
```
// A default value for unset variable can be provided after :- suffix, for example:
// `env:NAME_OF_ENVIRONMENT_VARIABLE:-default_value`
```

#### Link to tracking issue
Fixes #5228

#### Testing
Unit tests

#### Documentation
* [x] Provider Go docs 
* [ ] Probably needs some other docs changes?


TylerHelmuth:
I'd like to make sure we're in alignment with the configuration sig on the syntax before merging this

yurishkuro:
@TylerHelmuth I though that was settled, per https://github.com/open-telemetry/opentelemetry-collector/issues/5228#issuecomment-2233745849

> The OpenTelemetry Configuration WG now admits the shell syntax for the SDK configuration files. Since we aim to be aligned with them, it makes sense in my opinion to add support in the Collector for this as well.

yurishkuro:
@mx-psi is Configuration WG decision on shell syntax documented / implemented somewhere?

mx-psi:
> @mx-psi is Configuration WG decision on shell syntax documented / implemented somewhere?

Yup, see https://opentelemetry.io/docs/specs/otel/configuration/file-configuration/#environment-variable-substitution

confmap/provider/envprovider/provider_test.go
```diff
@@ -154,6 +154,68 @@ func TestEmptyEnvWithLoggerWarn(t *testing.T) {
...
+		name     string
+		unset    bool
+		varValue string
+		uri      string
+		expected string
+	}{
+		{name: "unset", unset: true, uri: "env:MY_VAR:-default % value", expected: "default % value"},
+		{name: "unset2", unset: true, uri: "env:MY_VAR:-", expected: ""}, // empty default still applies
+		{name: "empty", varValue: "", uri: "env:MY_VAR:-foo", expected: ""},
+		{name: "not empty", varValue: "value", uri: "env:MY_VAR:-", expected: "value"},
```
TylerHelmuth, line 169: Can you add some test cases for:

- `"env:MY_VAR:-test:-test`
- `"env:MY_VAR:?test:?test`
- `"env:-MY_VAR`
- `"env:?MY_VAR`

mx-psi:
Note that the `:?error` syntax is not supported by the Configuration WG. I am in favor of it, but it would be good to at least check that the Configuration WG is not vehemently opposed to eventually adding that

confmap/provider/envprovider/provider.go
```diff
@@ -27,6 +27,14 @@ type provider struct {
 //
 // This Provider supports "env" scheme, and can be called with a selector:
 // `env:NAME_OF_ENVIRONMENT_VARIABLE`
+//
+// A default value for unset variable can be provided after :- suffix, for example:
+// `env:NAME_OF_ENVIRONMENT_VARIABLE:-default_value`
```
codeboten, line 32: Default values are supported by the configuration working group as [defined in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/data-model.md#environment-variable-substitution)

confmap/provider/envprovider/provider.go
```diff
@@ -27,6 +27,14 @@ type provider struct {
...
+//
+// An error message for unset variable can be provided after :? suffix, for example:
+// `env:NAME_OF_ENVIRONMENT_VARIABLE:?error_message`
```
codeboten, line 35: I'm not aware of a definition for setting an error message, is there an open issue somewhere?

yurishkuro: I think we should still have it. I don't understand the approach "let's be like Shell but just different enough so that common things don't work".

codeboten: not disagreeing, this was me asking if this is already a discussion about this somewhere I haven't found yet

mx-psi: > I don't understand the approach "let's be like Shell but just different enough so that common things don't work".

This doesn't mean that we need to add full support for Bash/POSIX-like syntax today. Your PR does not add full support for [all the features that Bash supports](https://tiswww.case.edu/php/chet/bash/bashref.html#Shell-Expansions) and that's okay. As I said above, I am not opposed to adding this, but since alignment with the Configuration WG is an explicit goal we should bring it up on their repo first.

yurishkuro:
I removed error syntax -- 3a3f939.

mx-psi (approved):

Issue #10801, opened 2024-08-05
Move attributes definition for internal metrics to metadata.yaml


dmitryax:
Define all the attributes for internal metrics in the metadata.yaml similar to what we do for the scraper metrics. 

Those attribute definitions should be used to generate documentation and helpers that replace the directly exposed counters with `Record...` functions.

For example:

Instead of 

```golang
type TelemetryBuilder struct {
	ExporterSentSpans metric.Int64Counter
}
```

we would generate

```golang
type TelemetryBuilder struct {
	exporterSentSpans   metric.Int64Counter
}

// RecordExporterSentSpans adds a value to the ExporterSentSpans metric.
func (builder *TelemetryBuilder) RecordExporterSentSpans(ctx context.Context, value int64, exporter string) {
	builder.exporterSentSpans.Add(context.Background(), value, metric.WithAttributeSet(attribute.NewSet(attribute.String("exporter", exporter))))
}
```

dmitryax:
@codeboten, please let me know if it makes sense

codeboten:
100% i would like to move in that direction for all internal telemetry as well

crobert-1:
_Just to confirm with the assignment, I'll work on this 👍_

crobert-1:
Started working on this, just want to bring up a limitation of the upstream `go.opentelemetry.io/otel/attribute` package. As described in this issue's description, attributes are added to a telemetry metric using the `go.opentelemetry.io/otel/attribute` package's functionality of [`attribute.NewSet`](https://pkg.go.dev/go.opentelemetry.io/otel/attribute#NewSet), passing in the correct [KeyValue](https://pkg.go.dev/go.opentelemetry.io/otel/attribute#KeyValue) information.

However, `KeyValue` only supports a subset of valid attribute types that our `metadaya.yaml` supports.

From mdatagen's metadata schema: https://github.com/open-telemetry/opentelemetry-collector/blob/505819e63ae2b615d7c57c9cfe2b4325ae82dc4c/cmd/mdatagen/metadata-schema.yaml#L72

Types supported by `KeyValue` can be seen [here](https://pkg.go.dev/go.opentelemetry.io/otel/attribute#KeyValue).

The missing types are as follows: `bytes`, `slice`, and `map`. I believe this is an existing limitation, unrelated to this change. However, it may be good to document this limitation in the metadata schema file.

crobert-1:
[...]

PR #10911, opened 2024-08-19
[mdatagen] Generate wrapper for OTel calls as a first step towards configurable attributes

codeboten:
Part of #10801


crobert-1 (commented):
Should we add tests as a part of this change as well?

codeboten:
@crobert-1 added test in mdatagen to validate the generated code

cmd/mdatagen/main_test.go
```diff
@@ -620,6 +621,132 @@ func LeveledMeter(settings component.TelemetrySettings, level configtelemetry.Le
...
+					Class:         "receiver",
+				},
+				Telemetry: telemetry{
+					Metrics: map[metricName]metric{
+						"metric1": {
+							Enabled: true,
+							Sum: &sum{
+								Async:           false,
+								Mono:            Mono{true},
+								MetricValueType: MetricValueType{ValueType: pmetric.NumberDataPointValueTypeInt},
```
crobert-1, line 644: Could we add another case where the type is `Histogram`, since that's the other special case in the generated code?

cmd/mdatagen/templates/telemetry.go.tmpl
```diff
@@ -133,4 +133,14 @@ func NewTelemetryBuilder(settings component.TelemetrySettings, options ...teleme
     return &builder, errs
 }
 
+{{- range $name, $metric := .Telemetry.Metrics }}
+{{- if not $metric.Data.Async }}
+
+func (b *TelemetryBuilder) Record{{ $name.Render }}(ctx context.Context, val {{ $metric.Data.BasicType }}, opts ...metric.{{- if eq $metric.Data.Type "Histogram" -}}Record{{- else -}}Add{{- end -}}Option) {
+    b.{{ $name.Render }}.{{- if eq $metric.Data.Type "Histogram" -}}Record{{- else -}}Add{{- end -}}(ctx, val, opts...)
```
crobert-1, line 140: I believe we also need `Record` for gauge metrics.

I found that the [Int64Gauge](https://pkg.go.dev/go.opentelemetry.io/otel/metric#Int64Gauge) metric type has `Record`, not `Add` as rendered here. (Same goes for Float64Gauge) This is hit if you do:

```
telemetry:
  metrics:
    fake_metric:
      enabled: true
      description: Description of metric
      unit: "{x}"
      gauge:
        value_type: int
```

PR #10912, opened 2024-08-19
[service] deprecate TelemetrySettings.MeterProvider

codeboten:
This is replaced by a LeveledMeterProvider method on the struct instead. This
reduces the complexity from the view of component authors, in that there's no need
to check the level before invoking the meter to record a metric.

Closes https://github.com/open-telemetry/opentelemetry-collector/issues/9510

### DIFF
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -22,6 +22,8 @@ type TelemetrySettings struct {
 	TracerProvider trace.TracerProvider
 
 	// MeterProvider that the factory can pass to other instrumented third-party libraries.
+	//
+	// Deprecated [v0.108.0]: use LeveledMeterProvider instead.
 	MeterProvider metric.MeterProvider
 
 	// LeveledMeterProvider returns a MeterProvider for a Level that the factory can
@@ -30,6 +32,8 @@ type TelemetrySettings struct {
 
 	// MetricsLevel controls the level of detail for metrics emitted by the collector.
 	// Experimental: *NOTE* this field is experimental and may be changed or removed.
+	//
+	// Deprecated [v0.108.0]: use LeveledMeterProvider instead.
 	MetricsLevel configtelemetry.Level
 
 	// Resource contains the resource attributes for the collector's telemetry.
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -30,10 +30,9 @@ type TelemetrySettings struct {
 	// pass to other instrumented third-party libraries.
 	LeveledMeterProvider func(level configtelemetry.Level) metric.MeterProvider
 
-	// MetricsLevel controls the level of detail for metrics emitted by the collector.
-	// Experimental: *NOTE* this field is experimental and may be changed or removed.
-	//
-	// Deprecated [v0.108.0]: use LeveledMeterProvider instead.
+	// MetricsLevel represents the configuration value set when the collector
+	// is configured. Components may use this level to decide whether it is
+	// appropriate to avoid computationally expensive calculations.
 	MetricsLevel configtelemetry.Level
 
 	// Resource contains the resource attributes for the collector's telemetry.
--- a/component/telemetry.go
+++ b/component/telemetry.go
@@ -23,7 +23,7 @@ type TelemetrySettings struct {
 
 	// MeterProvider that the factory can pass to other instrumented third-party libraries.
 	//
-	// Deprecated [v0.108.0]: use LeveledMeterProvider instead.
+	// Deprecated [v0.109.0]: use LeveledMeterProvider instead.
 	MeterProvider metric.MeterProvider
 
 	// LeveledMeterProvider returns a MeterProvider for a Level that the factory can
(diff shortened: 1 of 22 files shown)


codeboten:
Mainly what I'm looking for feedback on is https://github.com/open-telemetry/opentelemetry-collector/pull/10912/files#diff-8992ea9c665081035bd866a386bc9510ee77af8cbb7137182e0fbcf38b92d64a and the changes in service.go

The generated code needs some more fixing. Pinging @mx-psi as this is following the proposed approach in #9510

mx-psi (commented):
The approach looks sensible to me. We give up on being able to have different metrics on the same 3rd party library being spread across different levels, but I think that's okay (we are not in the business of deciding 3rd party libraries telemetry for them)

component/telemetry.go
```diff
@@ -22,10 +22,18 @@ type TelemetrySettings struct {
 	TracerProvider trace.TracerProvider
 
 	// MeterProvider that the factory can pass to other instrumented third-party libraries.
+	//
+	// Deprecated [v0.108.0]: use LeveledMeterProvider instead.
 	MeterProvider metric.MeterProvider
 
+	// LeveledMeterProvider returns a MeterProvider for a Level that the factory can
+	// pass to other instrumented third-party libraries.
+	LeveledMeterProvider func(level configtelemetry.Level) metric.MeterProvider
```
TylerHelmuth, line 31: Does this dependency mean we can't mark `component` stable until `configtelemetry` is stable?

mx-psi: Yes, but this dependency was already there, right?

TylerHelmuth: You're right

codeboten:
> We give up on being able to have different metrics on the same 3rd party library being spread across different levels, but I think that's okay (we are not in the business of deciding 3rd party libraries telemetry for them)

Is this something we had the ability to do previous? We don't have a mechanism for using the levels at the instrumentation library level, users can always configure views if they need additional granularity

codeboten:
After spending a long time discussing various options, I think the best way to move forward is to replace `MeterProvider` with `LeveledMeterProvider`, and keep `MetricsLevel`.

Adding `LeveledMeterProvider` gives slightly better ergonomics for component authors, as they may need to access a meter provider directly (for components that don't use mdatagen). Doing this prior to this change, required a check for a level where now just passing the level to the LeveledMeterProvider method is enough.

I tried a few different ways to remove the MetricsLevel, however there are a few components that rely on this information to decide when to make expensive computations (prior to recording). I considered replacing this with a callback that could be set a component instantiation, but that would still require the component (or whatever is doing the recording, telemetryBuilder or whatever) to be aware of the MetricsLevel. As the components themselves instantiate a telemetry builder, they would still need access to the telemetry level, which doesn't save anything.

If others have ideas of how to make this easier/better, I'm all ears, pinging @open-telemetry/collector-approvers for reviews/input

TylerHelmuth:
I spent time thinking about this as well and haven't come up with a better solution that a field on the struct for how to give component access to the actual level.  We have use cases today that require knowing the actual configured telemetry level, and introducing more structs/interfaces/methods all felt more complex and unnecessary than leaving `MetricsLevel`.

mx-psi:
>  there are a few components that rely on this information to decide when to make expensive computations (prior to recording).

Do you have a link for those? I am guessing these need to be recorded by sync instruments, right?

TylerHelmuth:
@mx-psi the [batchprocessor](https://github.com/open-telemetry/opentelemetry-collector/blob/18d5c02ade6f41f2cf8d01d3fa7a0fb8f97534a8/processor/batchprocessor/metrics.go#L51) uses it to check [later](https://github.com/open-telemetry/opentelemetry-collector/blob/18d5c02ade6f41f2cf8d01d3fa7a0fb8f97534a8/processor/batchprocessor/batch_processor.go#L411-L413) if it should calculate a value.

The other is otelarrow, doing similar checks to see if if should do some math. The situation is more complex, but [this calculation](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/cfdf245fb917b31f2b722f2562467a9c00a87071/internal/otelarrow/netstats/netstats.go#L202-L230) only ends up happening based on the metrics level.

codeboten:
> Do you have a link for those? I am guessing these need to be recorded by sync instruments, right?

Right in both cases they're synchronous instruments. I thought of allowing callbacks to be passed in as a workaround but I'm not sure if it's worth the effort.

There are two cases one in core ([batch processor](https://github.com/open-telemetry/opentelemetry-collector/blob/18d5c02ade6f41f2cf8d01d3fa7a0fb8f97534a8/processor/batchprocessor/batch_processor.go#L411)) and one in contrib ([otel arrow netstats package](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/dd5400a3da3d64c4a10e64f936e2de2eeff8490b/internal/otelarrow/netstats/netstats.go#L157))

mx-psi (approved):
> Is this something we had the ability to do previous? We don't have a mechanism for using the levels at the instrumentation library level, users can always configure views if they need additional granularity

You are right, I think this would have been equally hard before this change

> After spending a long time discussing various options, I think the best way to move forward is to replace MeterProvider with LeveledMeterProvider, and keep MetricsLevel.

Thanks for the links! After taking a look I also feel like this is the best solution I can think of

mx-psi:
I think we are ready to merge this after the conflicts are fixed

TylerHelmuth (approved):

Issue #10927, opened 2024-08-20
[Flaky test] Test_ComponentStatusReporting_SharedInstance flaky


songy23:
https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10473477699/job/29005426543?pr=10915

```
--- FAIL: Test_ComponentStatusReporting_SharedInstance (15.01s)
    status_test.go:117: 
        	Error Trace:	D:/a/opentelemetry-collector/opentelemetry-collector/internal/e2e/status_test.go:117
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 6
        	Test:       	Test_ComponentStatusReporting_SharedInstance
    status_test.go:120: 
        	Error Trace:	D:/a/opentelemetry-collector/opentelemetry-collector/internal/e2e/status_test.go:120
        	Error:      	Not equal: 
        	            	expected: 6
        	            	actual  : 7
        	Test:       	Test_ComponentStatusReporting_SharedInstance
FAIL
FAIL	go.opentelemetry.io/collector/internal/e2e	18.766s
FAIL
```

songy23:
This happens on Ubuntu as well https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10473495585/job/29005525483?pr=10918

songy23:
Unfortunately this is still happening, e.g. 3hr ago https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10599797655/job/29375720642#step:6:527

```
--- FAIL: Test_ComponentStatusReporting_SharedInstance (15.02s)
    status_test.go:130: checking errors for metrics, - Receiver - test
    status_test.go:154: events received: StatusStarting,StatusRecoverableError,StatusOK,StatusStopping,StatusStopped,
    status_test.go:130: checking errors for traces, - Receiver - test
    status_test.go:142: 
        	Error Trace:	/home/runner/work/opentelemetry-collector/opentelemetry-collector/internal/e2e/status_test.go:142
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 6
        	Test:       	Test_ComponentStatusReporting_SharedInstance
    status_test.go:145: 
        	Error Trace:	/home/runner/work/opentelemetry-collector/opentelemetry-collector/internal/e2e/status_test.go:145
        	Error:      	Not equal: 
        	            	expected: 6
        	            	actual  : 7
        	Test:       	Test_ComponentStatusReporting_SharedInstance
[...]

PR #10929, opened 2024-08-20
[chore] Try fixing flaky SharedInstance e2e test

TylerHelmuth:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10927.

` go test status_test.go --count 100` passed.

### DIFF
--- a/internal/e2e/status_test.go
+++ b/internal/e2e/status_test.go
@@ -6,6 +6,8 @@ package e2e
 import (
 	"context"
 	"errors"
+	"fmt"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/e2e/status_test.go
```diff
@@ -102,31 +104,34 @@ func Test_ComponentStatusReporting_SharedInstance(t *testing.T) {
 	err = s.Shutdown(context.Background())
 	require.NoError(t, err)
 
-	assert.Equal(t, 5, len(eventsReceived))
+	require.Equal(t, 2, len(eventsReceived))
```
codeboten, line 107: why is the value changed?

TylerHelmuth: To simplify the test I added a check in the test extension's `ComponentStatusChanged` function to only record events for the component the test cares about.  This changed the number or recorded sources from 5 to 2.

internal/e2e/status_test.go
```diff
@@ -101,31 +103,40 @@ func Test_ComponentStatusReporting_SharedInstance(t *testing.T) {
...
+				assert.Equal(t, componentstatus.StatusStopping, e.Status())
+			}
+			if i == 4 {
+				assert.Equal(t, componentstatus.StatusStopped, e.Status())
+			}
+			if i >= 5 {
+				assert.Fail(t, "received too many events")
 			}
 		}
+		t.Logf("events received: %v", eventStr)
```
codeboten, line 139: Just confirming that the two log statements were left intentionally

codeboten (approved):

internal/e2e/status_test.go
TylerHelmuth: Ya I added these so that in the future, if a flaky test comes back, it will be a little easier to debug.

PR #10938, opened 2024-08-21
[confmap] Use stringy representation on stringy types

### LINKED ISSUE #10937: Failing expansion of map string values that look like another type
**Describe the bug**
The issue is similar to https://github.com/open-telemetry/opentelemetry-collector/issues/10552#issuecomment-2293986519 still can be hit if an expansion is part of a map value.

**Steps to reproduce**
Run collector with a config like:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
      http:

exporters:
  otlphttp:
    traces_endpoint: "example.com:4317"
    headers:
      "AUTH-TOKEN": "${AUTH_TOKEN}"

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [otlphttp]
```

and `AUTH_TOKEN` env var set to `123`

Collector to start.

**What did you see instead?**
Collector fails with 

```
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

error decoding 'exporters': error reading configuration for "otlphttp": decoding failed due to the following error(s):

'headers[AUTH-TOKEN]' expected type 'configopaque.String', got unconvertible type 'int', value: '123'
2024/08/20 21:23:23 collector server run finished with error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

error decoding 'exporters': error reading configuration for "otlphttp": decoding failed due to the following error(s):

'headers[AUTH-TOKEN]' expected type 'configopaque.String', got unconvertible type 'int', value: '123'
```

mx-psi:
#### Description

Use string value for stringy types

#### Link to tracking issue

Fixes #10937

#### Testing

Adds tests for stringy type recognition and adds end to end tests


TylerHelmuth (approved):

iblancasa (approved):

confmap/confmap.go
```diff
@@ -244,6 +259,21 @@ func castTo(exp expandedValue, useOriginal bool) (any, error) {
...
 }
 
+// Check if a reflect.Type is of the form T:
+// T = string | map[string]T | []T | [n]T
+func isStringyStructure(t reflect.Type) bool {
+	if t.Kind() == reflect.String {
+		return true
+	}
+	if t.Kind() == reflect.Map {
+		return t.Key().Kind() == reflect.String && isStringyStructure(t.Elem())
```
dmitryax, line 269: Why do we need to require keys to be strings?

dmitryax (approved):
Thank you, @mx-psi. Just on question

confmap/confmap.go
mx-psi: It's not really needed, dropped that requirement

PR #10941, opened 2024-08-21
[service/internal] Remove Ready from status.Reporter

### LINKED ISSUE #10792: [componentstatus] I think we can remove `Ready()` and the ready related code in internal/status. When `ReportStatus` was on `TelemetrySettings` it was possible to call it before start. Now that it's been moved to `Host` this should no longer be possible. We can create an issue and address this in a followup. I'd be willing to take that task if we agree.
I think we can remove `Ready()` and the ready related code in internal/status. When `ReportStatus` was on `TelemetrySettings` it was possible to call it before start. Now that it's been moved to `Host` this should no longer be possible. We can create an issue and address this in a followup. I'd be willing to take that task if we agree.

_Originally posted by @mwear in https://github.com/open-telemetry/opentelemetry-collector/pull/10777#discussion_r1700900597_

mwear:
#### Description
This PR removes the `Ready` method from `status.Reporter`. This check was necessary when `ReportStatus` was on `TelemetrySettings` as it should not be called until component start-time, but was available to a component at create-time. This method now resides on `host` which is not available until component start, making this check unnecessary. This is an internal refactor that should not affect any other users. I'm not sure if this needs a changelog, but I generated one anyways and set the type to breaking, as it removes a method from an interface.

#### Link to tracking issue
Fixes #10792

#### Testing
Units


.chloggen/remove-component-status-ready.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
TylerHelmuth, line 4: Do we need a changelog for this? It looks like `status.Reporter` is internal so that API was not public

mwear: I'll remove it. It's a completely internal refactor.

TylerHelmuth (approved):

mx-psi (approved):

mx-psi:
@mwear can you fix the merge conflict? I can merge after that

codeboten (approved):

PR #10946, opened 2024-08-22
[chore] Add profiles to fanoutconsumer

dmathieu:
#### Description

This adds profiles support in fanoutconsumer. That is required to be able to handle profiles in the service graph.
As this is only changing internal modules, I am marking it as chore (and skipping changelog entry).

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/pull/10375

cc @mx-psi


mx-psi (approved):
Thanks!

PR #10948, opened 2024-08-22
[configgrpc] fix integer overflow in configuration

codeboten:
Change the value of max_recv_msg_size_mib from uint64 to int to avoid a case where misconfiguration caused an integer overflow. Added a Validate function to configgrpc as part of this to validate that users didn't set the value to something that would overflow when converted from megabytes to bytes.

### DIFF
--- a/.chloggen/codeboten_int64-int.yaml
+++ b/.chloggen/codeboten_int64-int.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: configgrpc
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change the value of max_recv_msg_size_mib from uint64 to int to avoid a case where misconfiguration caused an integer overflow.
+
+# One or more tracking issues or pull requests related to the change
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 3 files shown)


.chloggen/codeboten_int64-int.yaml
```diff
@@ -0,0 +1,25 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: configgrpc
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change the value of max_recv_msg_size_mib from uint64 to int to avoid a case where misconfiguration caused an integer overflow.
+
+# One or more tracking issues or pull requests related to the change
+issues: []
```
codeboten, line 13:
```suggestion
issues: [10948]
```

codeboten:
This bug was uncovered by the linter https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10480357311/job/29027844869?pr=10935

config/configgrpc/configgrpc.go
```diff
@@ -320,6 +320,22 @@ func validateBalancerName(balancerName string) bool {
 	return balancer.Get(balancerName) != nil
 }
 
+func (gss *ServerConfig) Validate() error {
+	if gss.MaxRecvMsgSizeMiB*1024*1024 < 0 {
+		return fmt.Errorf("invalid max_recv_msg_size_mib value: %d", gss.MaxRecvMsgSizeMiB)
```
TylerHelmuth, line 325: Cant he error messages include the max valid value?

codeboten: Done

mx-psi (approved):

Issue #10949, opened 2024-08-22
Collector panics on startup for non-nilable config options with unset environment variable


crobert-1:
**Describe the bug**

When the collector is started with a configuration that has a non-nilable config option (e.g. `int` or `uint64`) set to an environment variable that hasn't been set, the collector panics without any information other than the stack.

**Steps to reproduce**

Config:
```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:8518

exporters:
  debug:
    verbosity: detailed
    sampling_initial: ${UNSET_VARIABLE}

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: []
      exporters: [debug]
```

Start the collector with an `int` configuration option, in this case the debug exporter's [`sampling_initial` option](https://github.com/open-telemetry/opentelemetry-collector/blob/18d5c02ade6f41f2cf8d01d3fa7a0fb8f97534a8/exporter/debugexporter/config.go#L29), set to the environment variable `UNSET_VARIABLE`, which isn't set.

**What did you expect to see?**

No panic. I'm not sure if intended behavior here is to error and force the user to set the value of the variable, or to start with a default value, such as `0` for number types.

**What did you see instead?**

```
$ ./bin/otelcorecol_darwin_amd64 --config config.yaml 
panic: reflect: call of reflect.Value.Interface on zero Value

goroutine 1 [running]:
reflect.valueInterface({0x0?, 0x0?, 0xc00041c528?}, 0x85?)
	reflect/value.go:1501 +0xfb
reflect.Value.Interface({0x0?, 0x0?, 0xc0003f5400?})
	reflect/value.go:1496 +0x25
go.opentelemetry.io/collector/confmap.decodeConfig.expandNilStructPointersHookFunc.func2({0x0?, 0x0?, 0xa0b8020?}, {0xa0b8020?, 0xc0003b32c8?, 0xc0003f5480?})
	go.opentelemetry.io/collector/confmap@v0.107.0/confmap.go:306 +0x131
github.com/go-viper/mapstructure/v2.cachedDecodeHook.func3({0x0?, 0x0?, 0x5?}, {0xa0b8020?, 0xc0003b32c8?, 0xc00041c650?})
	github.com/go-viper/mapstructure/v2@v2.1.0/decode_hooks.go:54 +0x2c
[...]

PR #10950, opened 2024-08-22
[confmap] Fix for unset env var setting non-string field.

TylerHelmuth:
#### Description
Still working on making a new test for this scenario, but all existing tests pass.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10949

#### Testing
Manual testing showed the bug resolved, but I'd still like to get an e2e test to check it.

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -253,7 +253,17 @@ func useExpandValue() mapstructure.DecodeHookFuncType {
 		to reflect.Type,
 		data any) (any, error) {
 		if exp, ok := data.(expandedValue); ok {
-			return castTo(exp, to.Kind() == reflect.String)
+			v, err := castTo(exp, to.Kind() == reflect.String)
+			if err != nil {
+				return v, err
+			}
+			// See https://github.com/open-telemetry/opentelemetry-collector/issues/10949
+			// If the `to.Kind` is not a string, then expandValue's original value is useless and
+			// the casted-to value will be nil. In that scenario, we need to use the default value of `to`'s kind.
+			if v == nil {
+				return reflect.Zero(to).Interface(), nil
+			}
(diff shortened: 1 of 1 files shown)


TylerHelmuth:
This doesn't work, the reflected default value ends up being used instead of the config's default value.  I think we'd need `sampling_initial` to be unset if the env var is not set in order to properly merge in the default value.

TylerHelmuth:
Maybe that is ok?

In the string situation the value would be getting set to "" which would also override any default string value.  So 

```yaml
exporters:
  debug:
    verbosity: ${env:UNSET_VAR}
```

would be seen as 

```yaml
exporters:
  debug:
    verbosity: ""
```

Not 

```yaml
exporters:
  debug:
    verbosity: "basic"
```

TylerHelmuth:
I've confirmed that prior to this regression the unset env var would result in the field being set with the type's default value, not the default value from `CreateDefaultConfig`.  So this PR should be a return to that behavior.

mx-psi (approved):
LGTM, is there any way we can reproduce the panic in a unit test?

mx-psi:
This test in `confmap/internal/e2e/types_test.go` panics in main but passes on this PR:

```go
func TestIssue10949_UnsetVar(t *testing.T) {
	resolver := NewResolver(t, "types_expand.yaml")
	conf, err := resolver.Resolve(context.Background())
	require.NoError(t, err)

	var cfg TargetConfig[int]
	err = conf.Unmarshal(&cfg)
	require.NoError(t, err)
	require.Equal(t, 0, cfg.Field)
}
```

Could you add this/something like this as a test?

TylerHelmuth:
@mx-psi added, thank you!

PR #10958, opened 2024-08-23
[service/internal] Allow components to transition from PermanentError to Stopping

### LINKED ISSUE #10058: [component] Status Aggregation in Core

mwear:
#### Description
In #10058 I mentioned:

> There is a tangentially related issue with PermanentErrors and the underlying finite state machine that governs transitions between statuses. Currently, a PermanentError is a final state. That is, once a component enters this state, no further transitions are allowed. In light of the work I did on the alternative health check extension, I believe we should allow a transition from PermanentError to Stopping to consistently prioritize lifecycle events for components. This transition also make sense from a practical perspective. A component in a PermanentError state is one that has been started and is running, although in a likely degraded state. The collector will call shutdown on the component (when the collector is shutting down) and we should allow the status to reflect that.

This PR makes the suggested change and updates the documentation to reflect that. As this is an internal change, I have not included a changelog. Also note, we can close #10058 after this as we've already removed status aggregation from core during the recent component status refactor.

#### Link to tracking issue
Fixes #10058

#### Testing
units

#### Documentation
Updated docs/component-status.md and associated diagram.


atoulme (approved):

mwear:
Pinging since the stalebot is engaging.

TylerHelmuth (approved):

mwear:
Is there any chance this can be reopened? It was approved, but never merged.

atoulme (approved):

mx-psi (approved):
Sorry that this didn't get attention. Will merge this on Monday unless there are comments before then

PR #10993, opened 2024-08-28
[chore]: enable bool-compare rule from testifylint

mmorel-35:
#### Description

Testifylint is a linter that provides best practices with the use of testify.

This PR enables [bool-compare](https://github.com/Antonboom/testifylint?tab=readme-ov-file#bool-compare) rule from [testifylint](https://github.com/Antonboom/testifylint)

It's linter provided by golangci-lint.

Here all available rules are activated except those who require to be fixed. This PR only fixes bool-compare so the quantity of changes stays reasonnable for reviewers.


iblancasa (approved):

.golangci.yml
```diff
@@ -121,6 +121,23 @@ linters-settings:
           files:
             - "!**/*_test.go"
 
+  testifylint:
+    // TODO: enable all rules
```
songy23, line 125:
```suggestion
    # TODO: enable all rules
```

mx-psi (approved):

PR #10996, opened 2024-08-28
[otlphttpreceiver] Allow H2C connections to otlp http receiver

### LINKED ISSUE #10952: Allow for http2 connection to otlphttpreceiver for insecure connections
**Describe the bug**
Right now, the OTLP HTTP receiver only allows for http2 upgrades if you specify a TLS configuration. This is an issue for environments that terminate TLS before the collector.

**Steps to reproduce**
1. Run a collector withthe config provided
2. Run the curl locally:
```bash
curl -v --http2-prior-knowledge http://localhost:4318/test
```

that http2 succeeds

**What did you see instead?**
A failure
```
* Host localhost:4318 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:4318...
* Immediate connect fail for ::1: Cannot assign requested address
*   Trying 127.0.0.1:4318...
* Connected to localhost (127.0.0.1) port 4318
* [HTTP/2] [1] OPENED stream for http://localhost:4318/test
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: http]
* [HTTP/2] [1] [:authority: localhost:4318]
* [HTTP/2] [1] [:path: /test]
* [HTTP/2] [1] [user-agent: curl/8.5.0]
* [HTTP/2] [1] [accept: */*]
> GET /test HTTP/2
> Host: localhost:4318
> User-Agent: curl/8.5.0
> Accept: */*
> 
* Remote peer returned unexpected data while we expected SETTINGS frame.  Perhaps, peer does not support HTTP/2 properly.
* Closing connection
curl: (56) Remote peer returned unexpected data while we expected SETTINGS frame.  Perhaps, peer does not support HTTP/2 properly.
```

**What version did you use?**
v0.107.0

**What config did you use?**
```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
processors:
  batch:

exporters:
  otlp:
    endpoint: otelcol:4317

extensions:
  health_check:
  pprof:
  zpages:

service:
  extensions: [health_check, pprof, zpages]
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
    metrics:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
```

**Environment**

Mac OS

**Additional context**
[This](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/confighttp.go#L365-L373) is the only place where the http2 protocol is allowed right now. Related [issue](https://github.com/golang/go/issues/46602). We could manually create an http2 handler following something like this:
```go
	if serverOptions.AllowH2CUpgrade {
		h2s := &http2.Server{IdleTimeout: 0}

		// handler.ServeHTTP blocks until the connection is closed. It creates a new server which managed the ]
		// upgraded connection and uses handler to serve requests to it.
		handler = h2c.NewHandler(handler, h2s)
	}
```

iblancasa:
#### Description
Allow H2C connections to otlp http receiver

Fixes #10952

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -21,6 +21,7 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/metric/noop"
 	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
 	"golang.org/x/net/publicsuffix"
 
 	"go.opentelemetry.io/collector/component"
@@ -110,6 +111,10 @@ type ClientConfig struct {
 	HTTP2PingTimeout time.Duration `mapstructure:"http2_ping_timeout"`
 	// Cookies configures the cookie management of the HTTP client.
 	Cookies *CookiesConfig `mapstructure:"cookies"`
+
+	// AllowH2CUpgrade enables clients to upgrade an existing
+	// HTTP/1.1 connection to an HTTP/2 connection without using TLS.
+	AllowH2CUpgrade bool `mapstructure:"allow_h2c_upgrade"`
 }
 
 // CookiesConfig defines the configuration of the HTTP client regarding cookies served by the server.
@@ -332,6 +337,10 @@ type ServerConfig struct {
 	// is zero, the value of ReadTimeout is used. If both are
 	// zero, there is no timeout.
 	IdleTimeout time.Duration `mapstructure:"idle_timeout"`
+
+	// AllowH2CUpgrade enables clients to upgrade an existing
+	// HTTP/1.1 connection to an HTTP/2 connection without using TLS.
+	AllowH2CUpgrade bool `mapstructure:"allow_h2c_upgrade"`
 }
 
 // NewDefaultServerConfig returns ServerConfig type object with default values.
@@ -345,6 +354,7 @@ func NewDefaultServerConfig() ServerConfig {
 		WriteTimeout:      30 * time.Second,
 		ReadHeaderTimeout: 1 * time.Minute,
 		IdleTimeout:       1 * time.Minute,
+		AllowH2CUpgrade:   false,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


iblancasa:
@djaglowski could you take a look?

config/confighttp/confighttp.go
```diff
@@ -419,6 +429,11 @@ func (hss *ServerConfig) ToServer(_ context.Context, host component.Host, settin
 		o.apply(serverOpts)
 	}
 
+	if hss.AllowH2CUpgrade {
+		h2s := &http2.Server{IdleTimeout: 0}
```
jpkrohling, line 433: I feel like this could surprise users out there. HTTP/2 requires TLS, and without an explicit request from admins to disable TLS, we shouldn't be just accepting insecure connections.

jpkrohling (changes requested):
I believe this should be opt-in, and have a log warning stating that this is a "weird" configuration. Only very specific setups have a valid reason to have this enabled, such as sidecars terminating TLS right before handing over to the main container.

config/confighttp/confighttp.go
iblancasa: Ok so, what do you think about setting the option fo `false` by default and adding a warning message here?

iblancasa:
Is there anything else pending from this PR?

iblancasa:
Can somebody from @open-telemetry/collector-approvers take a look? Thank you

jpkrohling (approved):
LGTM, however, I feel bad for introducing new features when we are aiming for v1.

If the Collector approvers decide that this is something we should NOT have for v1, I'd be on their side.

abernix:
Heya! Curious if there was there any further alignment with the Collector team on landing this prior to getting it auto-closed. Thanks!

iblancasa:
> Heya! Curious if there was there any further alignment with the Collector team on landing this prior to getting it auto-closed. Thanks!

I closed the PR since it seems there was not real interest on the feature.

ziqin:
Could this PR be reopened?

H2C is beneficial to collecting from local services for whom don't want to introduce gRPC dependencies but expecting an improved performance.

PR #11000, opened 2024-08-29
[extension] Move optional interfaces to extensioncapabilities

mx-psi:
#### Description

Moves extension optional interfaces to a new module, `extensioncapabilities`.

The intent is to be able to evolve these interfaces independently from the core interfaces on core.

These are used by at most one component/project, so I think they are not mature yet for 1.0


.chloggen/mx-psi_extensioncapabilities-depr.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: extension
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate `extension.ConfigWatcher`, `extension.PipelineWatcher` and `extension.Dependent` in favor of `extensioncapabilities` module.
```
evan-bradley, line 10:
```suggestion
note: Deprecate `extension.ConfigWatcher`, `extension.PipelineWatcher` and `extension.Dependent` in favor of equivalents in the `extensioncapabilities` module.
```

service/extensions/extensions.go
```diff
@@ -92,7 +93,7 @@ func (bes *Extensions) Shutdown(ctx context.Context) error {
 func (bes *Extensions) NotifyPipelineReady() error {
 	for _, extID := range bes.extensionIDs {
 		ext := bes.extMap[extID]
-		if pw, ok := ext.(extension.PipelineWatcher); ok {
+		if pw, ok := ext.(extensioncapabilities.PipelineWatcher); ok {
```
evan-bradley, line 96: We will need to continue to check the deprecated interface here during the deprecation process.

extension/extensioncapabilities/interfaces.go
```diff
@@ -0,0 +1,25 @@
...
+package extensioncapabilities // import "go.opentelemetry.io/collector/extension/extensioncapabilities"
+
+import (
+	"go.opentelemetry.io/collector/extension/internal"
+)
+
+// Dependent is an optional interface that can be implemented by extensions
+// that depend on other extensions and must be started only after their dependencies.
+// See https://github.com/open-telemetry/opentelemetry-collector/pull/8768 for examples.
+type Dependent = internal.Dependent
```
evan-bradley, line 15: Could you explain this pattern of re-exporting the types? I know I've seen it in some of the profiling PRs that have been coming out, but I would like to understand if we're using this as a convention going forward, or if it should be applied in certain scenarios.

mx-psi: The first thing that comes to my mind to do this move is to move the definition of `extension.Dependent` to `extensioncapabilities.Dependent`, and then do

```
// extension.go
type Dependent = extensioncapabilities.Dependent
```

However, `extension.Dependent` definition depends on `extension.Extension`, so doing this would create an import cycle and it would not compile. To avoid that we use an internal module to define the symbols, and expose them twice: one time deprecated, another time not deprecated.

Once we remove the deprecated symbols, we can get rid of `extension/internal` and just define them as normal

service/extensions/extensions.go
mx-psi: Since the interface is exactly the same, I don't think that's needed: if something fulfills the old interface, it should also fulfill the new interface. I wrote an example here: https://go.dev/play/p/cAdIQq2r7Vl

evan-bradley: Right, of course, I got a bit mixed up there. This is completely okay then, thanks for going so far as to write an example.

extension/extensioncapabilities/interfaces.go
evan-bradley: Got it, this is just for the deprecation process. That all makes sense then, thank you for the explanation.

evan-bradley (approved):

extension/extensioncapabilities/go.mod
```diff
@@ -0,0 +1,43 @@
+module go.opentelemetry.io/collector/extension/extensioncapabilities
```
TylerHelmuth, line 1: Would love a shorter name if we can think of one. I can't, but maybe someone else can

TylerHelmuth (approved):

extension/extensioncapabilities/go.mod
mx-psi: Other than just "capabilities" I am out of ideas

axw: Could be shortened to "extensioncaps" a la the Linux capabilities family of commands, e.g. https://man7.org/linux/man-pages/man8/getpcaps.8.html. Seems OK as it is though.

mx-psi: Yeah, I feel like this is okay, we can rename later if we want to since this is not going 1.0 now. It's definitely on the longer side for a package name but it's not the longest if you take contrib component names into account

PR #11001, opened 2024-08-29
[chore][VERSIONING.md] Add long term support policy

### LINKED ISSUE #10004: Define minimum supported period for 1.0
On the [GA roadmap] we stated as a requirement for 1.0 that

> A minimum support period for 1.0 is documented, similarly to [API and SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-support) stability guarantees.

We likely want to differentiate between Go API guarantees vs behavior guarantees and have the guarantees be per component

mx-psi:
#### Description

Adds support policy for Collector SIG software artifacts with the following principles:
- Support means that we will, at a minimum, fix critical bugs and security issues. We MAY fix other bugs and/or add features if needed (e.g. because a feature eases transition to a v2), but this is not required
- We support artifacts for end-users for longer
- Artifacts may contain sub-components that are not stable. The important bit is that users can easily tell when something is unstable.

#### Link to tracking issue
Fixes #10004


mx-psi:
cc @open-telemetry/collector-approvers @open-telemetry/collector-contrib-approvers

VERSIONING.md
```diff
@@ -1,13 +1,41 @@
...
+    * The Collector MUST provide a configuration or CLI flag to be able to opt out of using unstable components or features.
+    * The Collector's telemetry (e.g. Collector logs) MUST provide the ability to identify usage of unstable components or features.
+
+### Long term support
+
+The OpenTelemetry Collector SIG provides long term support for binary distributions of the OpenTelemetry Collector and its components. The following policies apply to long term support for any major version starting on v1:
+
+* A binary distribution of the OpenTelemetry Collector MUST be supported for a minimum of **one year** after the release of the next major version of said distribution.
+* Components MUST be supported for a minimum of **6 months** after the release of the next major version of said component or after the component has been marked as deprecated.
+* Components MAY be removed from a binary distribution of the OpenTelemetry Collector after the component has been marked as deprecated for a minimum of **6 months**.
```
evan-bradley, line 32: Just to be clear, this is elaborating on the previous point and describes a stable version of a component being removed from a distribution with a beta version?

VERSIONING.md
```diff
@@ -1,13 +1,41 @@
...
+- Binary distributions of the OpenTelemetry Collector.
+- Collector components.
+
+These artifacts are versioned according to the [semantic versioning v2.0.0](https://semver.org/) specification.
+
+### General considerations
+
+Binary distributions produced by the Collector SIG contain components and features with varying [levels of stability](README.md#stability-levels). We abide by the following principles to relate the Collector's version to the stability of its components and features:
+
+* The Collector's core framework MUST be stable in order for a Collector distribution to be v1.0.0 or higher.
```
evan-bradley, line 21: Do we have the scope of the core framework defined anywhere? Is this just describing the core framework packages used by the components within the distribution?

VERSIONING.md
mx-psi: I merged it into a single point. The intent is to say that we can remove components from v1 distros, provided we have deprecated the components before

VERSIONING.md
mx-psi: We have not previously defined it and I am not sure how to best write it. By core framework I mean anything that is not a component that you can plug (so e.g. configuration resolution, service behavior, CLI flags...). Put it another way, the common parts of all OCB builds

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
+### General considerations
+
+Binary distributions produced by the Collector SIG contain components and features with varying [levels of stability](README.md#stability-levels). We abide by the following principles to relate the Collector's version to the stability of its components and features:
+
+* The Collector's core framework behavior MUST be stable in order for a Collector distribution to be v1.0.0 or higher.
+* Users can easily understand when they are opting in to use a component or feature that is not stable.
+    * The Collector MUST provide a configuration or CLI flag to be able to opt out of using unstable components or features.
+    * The Collector's telemetry (e.g. Collector logs) MUST provide the ability to identify usage of unstable components or features.
+
+### Long term support
```
TylerHelmuth, line 26: How will we handle supported Go versions after 1.0? When 1.24 comes out do we drop support for 1.22 in the 1.0 modules?

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
+
+These artifacts are versioned according to the [semantic versioning v2.0.0](https://semver.org/) specification.
+
+### General considerations
+
+Binary distributions produced by the Collector SIG contain components and features with varying [levels of stability](README.md#stability-levels). We abide by the following principles to relate the Collector's version to the stability of its components and features:
+
+* The Collector's core framework behavior MUST be stable in order for a Collector distribution to be v1.0.0 or higher.
+* Users can easily understand when they are opting in to use a component or feature that is not stable.
+    * The Collector MUST provide a configuration or CLI flag to be able to opt out of using unstable components or features.
```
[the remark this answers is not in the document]
mx-psi: I reworded this to

> The Collector MUST be configurable in a way that does not use unstable components or features.

My point is that we can have, say, alpha components, but one should be able to not use them. This is already the case today: you can not add them to the pipeline, or you can disable a feature gate that you don't want. It was not meant to say anything about what's in the binary

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
+    * The Collector's telemetry (e.g. Collector logs) MUST provide the ability to identify usage of unstable components or features.
+
+### Long term support
+
+The OpenTelemetry Collector SIG provides long term support for binary distributions of the OpenTelemetry Collector and its components. The following policies apply to long term support for any major version starting on v1:
+
+* A binary distribution of the OpenTelemetry Collector MUST be supported for a minimum of **one year** after the release of the next major version of said distribution.
+* Components MUST be supported for a minimum of **6 months** after the release of the next major version of said component or after the component has been marked as deprecated. If a component has been deprecated for 6 months it MAY be removed from a binary distribution of the OpenTelemetry Collector.
```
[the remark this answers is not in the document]
mx-psi: Called out on 6f3dda2

VERSIONING.md
```diff
@@ -135,3 +162,10 @@ on its intended usage (e.g. when calling a method or when passed to any method o
     * Contrib modules will be kept up to date with this project's releases.
 * GitHub releases will be made for all releases.
 * Go modules will be made available at Go package mirrors.
+
+### Long term support
+
+The OpenTelemetry Collector SIG provides long term support for Go modules. Support for modules depend on the module's [target audiences](CONTRIBUTING.md#target-audiences):
+
+- Modules intended for **component developers** MUST be supported for a minimum of **1 year** after the release of the next major version of said module or after the module has been marked as deprecated.
+- Modules intended for **Collector library users** MUST be supported for a minimum of **6 months** after the release of the next major version of said module or after the module has been marked as deprecated.
```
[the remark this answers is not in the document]
mx-psi: I feel the other way :smile: at least for modules intended for component developers. We can't have an ecosystem of components outside of contrib if we don't guarantee that the APIs used to build components are stable and available for a long time. Originally I was going to write 2-3 years for component developers, I went with this to start with something a bit more conservative

For the ones meant for library users, I guess 3 months may be fine now because every major user in that audience (vendors and Jaeger) is involved with the Collector

VERSIONING.md
```diff
@@ -135,3 +162,10 @@ on its intended usage (e.g. when calling a method or when passed to any method o
     * Contrib modules will be kept up to date with this project's releases.
 * GitHub releases will be made for all releases.
 * Go modules will be made available at Go package mirrors.
+
+### Long term support
```
jpkrohling, line 166: I feel like this needs a "after v1" remark.

mx-psi: Clarified, thanks!

jpkrohling (approved):

VERSIONING.md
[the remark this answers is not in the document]
TylerHelmuth: Well I definitely don't want to support the APIs for 3 years, so I can live with 1.

TylerHelmuth (approved):

evan-bradley (approved):

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
-**Users are provided a codebase of value that is stable and secure.**
+**Users are provided software artifacts of value that are stable and secure.**
 
-## Public API expectations
+The policies are divided depending on the artifact's target audience. While an artifact is supported, [critical bugs](docs/release.md#bugfix-release-criteria) and security vulnerabilities MUST be addressed. The main criteria for the length of support for an artifact is how easy it is for an artifact's target audience to adapt to disruptive changes.
 
-The following public API expectations apply to all modules in opentelemetry-collector and opentelemetry-collector-contrib.
+## End-user focused software artifacts
+
+End-user focused software artifacts are those that are intended to be used by [end-users](CONTRIBUTING.md#end-users) of the OpenTelemetry Collector. These artifacts include
```
codeboten, line 11:
```suggestion
End-user focused software artifacts are those that are intended for [end-users](CONTRIBUTING.md#end-users) of the OpenTelemetry Collector. These artifacts include 
```

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
 
-## Public API expectations
+The policies are divided depending on the artifact's target audience. While an artifact is supported, [critical bugs](docs/release.md#bugfix-release-criteria) and security vulnerabilities MUST be addressed. The main criteria for the length of support for an artifact is how easy it is for an artifact's target audience to adapt to disruptive changes.
 
-The following public API expectations apply to all modules in opentelemetry-collector and opentelemetry-collector-contrib.
+## End-user focused software artifacts
+
+End-user focused software artifacts are those that are intended to be used by [end-users](CONTRIBUTING.md#end-users) of the OpenTelemetry Collector. These artifacts include 
+- Binary distributions of the OpenTelemetry Collector.
+- Collector components, such as receivers, processors, and exporters.
```
codeboten, line 13: Should this clarify that what is meant by components is the go module that encompass these components?

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
+
+These artifacts are versioned according to the [semantic versioning v2.0.0](https://semver.org/) specification.
+
+### General considerations
+
+Binary distributions produced by the Collector SIG contain components and features with varying [levels of stability](README.md#stability-levels). We abide by the following principles to relate the Collector's version to the stability of its components and features:
+
+* The Collector's core framework behavior MUST be stable in order for a Collector distribution to be v1.0.0 or higher.
+* Users can easily understand when they are opting in to use a component or feature that is not stable.
+    * The Collector MUST be configurable in a way that does not use unstable components or features.
```
codeboten, line 23: I'm not super clear on what this sentence means, are we talking about users having the ability to build a collector with only stable components?

VERSIONING.md
```diff
@@ -1,13 +1,40 @@
...
+    * The Collector's telemetry (e.g. Collector logs) MUST provide the ability to identify usage of unstable components or features.
+
+### Long term support after v1
+
+The OpenTelemetry Collector SIG provides long term support for stable binary distributions of the OpenTelemetry Collector and its components. The following policies apply to long term support for any major version starting on v1:
+
+* A binary distribution of the OpenTelemetry Collector MUST be supported for a minimum of **one year** after the release of the next major version of said distribution.
```
codeboten, line 30: It would be good to have a section in this doc specifically defining what "supported" means

codeboten: When does support for a binary end?

VERSIONING.md
mx-psi: Clarified in 78a6621

VERSIONING.md
mx-psi: You should be able to not use any components that are unstable (by not including them in your pipelines) and you should be able to disable feature gates that are unstable. Anything that you can't disable/opt out/not use should be stable. I am not sure how to phrase it better, I appreciate any suggestions

VERSIONING.md
mx-psi: > It would be good to have a section in this doc specifically defining what "supported" means

The intention was that this sentence above:

> [critical bugs](docs/release.md#bugfix-release-criteria) and security vulnerabilities MUST be addressed

was the definition of supported (or, at least, a minimum criteria for what supported means). We may do more than this, but that's the minimum.

> When does support for a binary end?

This was also meant to be already covered in the text by this:

> A binary distribution of the OpenTelemetry Collector MUST be supported for a minimum of **one year** after the release of the next major version of said distribution.

So, basically, we support the distro indefinitely until we release the next major version, after which we support it for one more year.

atoulme (approved):

VERSIONING.md
```diff
@@ -135,3 +162,10 @@ on its intended usage (e.g. when calling a method or when passed to any method o
     * Contrib modules will be kept up to date with this project's releases.
 * GitHub releases will be made for all releases.
 * Go modules will be made available at Go package mirrors.
+
+### Long term support after v1
```
tiffany76, line 166:
```suggestion
### Long-term support after v1
```

VERSIONING.md
```diff
@@ -135,3 +162,10 @@ on its intended usage (e.g. when calling a method or when passed to any method o
...
+
+The OpenTelemetry Collector SIG provides long term support for stable Go modules. Support for modules depend on the module's [target audiences](CONTRIBUTING.md#target-audiences). The following policies apply to long term support for any major version starting on v1:
```
tiffany76, line 168:
```suggestion
The OpenTelemetry Collector SIG provides long-term support for stable Go modules. Support for modules depend on the module's [target audiences](CONTRIBUTING.md#target-audiences). The following policies apply to long-term support for any major version starting on v1:
```

tiffany76 (commented):
Just a few copy-edit suggestions.

VERSIONING.md
tiffany76: On second thought, “required” doesn’t sound right. So how about

The Collector MUST be configurable so that unstable components or features can be excluded, ensuring that a fully stable configuration is possible.

mx-psi:
> I would add to this document a paragraph that explains that this policy is the current consensus, but it can change in the future, based on project direction and SIG discussions.

Done!

djaglowski (approved):

mx-psi:
I'll merge this by end of week unless there are further comments, we can iterate on the wording and clarify support more but this feels like it is good enough for a first version given the number of approvals

VERSIONING.md
```diff
@@ -1,13 +1,42 @@
...
+The policies are divided depending on the artifact's target audience. While an artifact is supported, [critical bugs](docs/release.md#bugfix-release-criteria) and security vulnerabilities MUST be addressed. The main criteria for the length of support for an artifact is how easy it is for an artifact's target audience to adapt to disruptive changes.
 
-The following public API expectations apply to all modules in opentelemetry-collector and opentelemetry-collector-contrib.
+These policies reflect the current consensus of the OpenTelemetry Collector SIG. They are subject to change as the project evolves.
+
+## Software artifacts for end users
+
+Software artifacts intended for [end users](CONTRIBUTING.md#end-users) of the OpenTelemetry Collector include 
+- Binary distributions of the OpenTelemetry Collector.
+- Go modules that expose Collector components, such as receivers, processors, connectors, extensions and exporters.
```
andrzej-stencel, line 15: This mentions Go modules for components. What about all other Go modules published by the project, like e.g. https://pkg.go.dev/go.opentelemetry.io/collector/otelcol or https://pkg.go.dev/go.opentelemetry.io/collector/config/confighttp? Do they get a different treatment than the Go modules for components?

mx-psi: Yes, they have a smaller support window https://github.com/open-telemetry/opentelemetry-collector/blob/8205489835974f8599e3553e03819ad8a0b6a5db/VERSIONING.md?plain=1#L173 

The rationale for this is that they are equipped to deal with more disruptive changes https://github.com/open-telemetry/opentelemetry-collector/blob/012cf862283a1af411c5b1ace844a6c1d7a82766/CONTRIBUTING.md?plain=1#L45-L46

VERSIONING.md
mx-psi: @codeboten if you think we should file an issue to continue discussing this please do so!

Issue #6506, opened 2022-11-08
Review `component` public API


codeboten:
Before considering `component` 1.0, we need to review the public API to ensure what is exposed makes sense https://pkg.go.dev/go.opentelemetry.io/collector/component

codeboten:
Some thoughts:

- should `BuildInfo` be in this package? since its more directly tied to the service running, it might make more sense to move it to that package
- `TelemetrySettings` should be moved into the `telemetry` package

bogdandrutu:
> should BuildInfo be in this package? since its more directly tied to the service running, it might make more sense to move it to that package

I think cannot be moved to the "service" since you need to pass this to components, so will create a circular dependency.

> TelemetrySettings should be moved into the telemetry package

Same issue about dependency.

bogdandrutu:
> Before considering component 1.0, we need to review the public API to ensure what is exposed makes sense https://pkg.go.dev/go.opentelemetry.io/collector/component

@codeboten we also need to make sure that "consumer" package is 1.0 before :)

mx-psi:
IMO `BuildInfo` does not make sense in `service`: the components should not know anything about the service other than through the `component.Host` interface; ideally we should be able to completely replace service with a new implementation without changing anything about the components.

For `TelemetrySettings`, I think it makes sense to move it to `telemetry`, but my proposal would be to make `telemetry` into a separate module outside of `service`. I want to work on this at some point, but haven't gotten around to it lately.

mx-psi:
I am wondering if we should split `component` into further packages. For example, we could have `componentstatus` and `componentfactory` as separate packages. It currently feels like a mish-mash of important but unrelated things

TylerHelmuth:
Did a review of Component's public API and here are my thoughts:

1. We could rename/refactor `DataType` (see https://github.com/open-telemetry/opentelemetry-collector/issues/9429).  Let's keep discussion on that specific field in that issue.

2. We could move `component.Telemetry` to its own package as `component.Telemetry` is not used anywhere without `component`. But I struggle to find end-user value in such a change at this point.


Component API notes

`component.Factory` has a comment that claims:

```
// This interface cannot be directly implemented. Implementations must
// use the factory helpers for the appropriate component type.
```

but the interface itself doesn't include anything that prevents implementations

Could move component.Telemetry to its own package?

[...]

PR #4338, opened 2021-11-02
Make `component.Factory` inherit `internalinterface.InternalInterface`

mx-psi:
**Description:** 

Make `component.Factory` embed `internalinterface.InternalInterface`, so that the component factories must be implemented using the helpers.

This also includes extensions (which are not part of the focus of #3921, since extensions don't have per-signal methods), but I think for consistency it's better to add them (moving the internalinterface dependency to individual interfaces wouldn't be a breaking change if we find this is too restrictive).

This is a follow up to #4175.

**Link to tracking Issue:** fixes #3921

PR #11017, opened 2024-08-29
[component] Remove incorrect comment on Factory interface

TylerHelmuth:
#### Description
Removes an incorrect comment from the `component.Factory interface`.

This comment was added via https://github.com/open-telemetry/opentelemetry-collector/pull/4338 when the extension/receiver/processor/exporter factory types all still [lived in](https://github.com/mx-psi/opentelemetry-collector/blob/b4be13e74923ebefb521f69a97288ce8d6f51ea9/component/exporter.go#L61) `component`. Since the factories have been moved to other modules, `component.Factory` needs to be implementable.

The specific extension/receiver/processor/exporter factory types are unimplementable since they are defined in internal packages. 

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/6506

### DIFF
--- a/component/component.go
+++ b/component/component.go
@@ -165,9 +165,6 @@ func (sl StabilityLevel) LogMessage() string {
 }
 
 // Factory is implemented by all Component factories.
-//
-// This interface cannot be directly implemented. Implementations must
-// use the factory helpers for the appropriate component type.
 type Factory interface {
 	// Type gets the type of the component created by this factory.
 	Type() Type


mx-psi (approved):
I am in favor of this, while I believe it is technically feasible to make the interface sealed, I feel like the change could be a ~biggish refactor with no clear benefit

codeboten (approved):

PR #11020, opened 2024-08-30
[extension] Split extension/experimental/storage into a separate module

### LINKED ISSUE #11022: Move `extension/experimental/storage` to a separate module

mx-psi:
#### Description

Fixes #11022


mx-psi:
@codeboten Should we rename to `extension/storage`? I remember with the `internal/featuregates` separation into a module using the same name was a concern, so I wonder if this is also something we should address here

TylerHelmuth (approved):

codeboten (approved):
> @codeboten Should we rename to extension/storage? I remember with the internal/featuregates separation into a module using the same name was a concern, so I wonder if this is also something we should address here

@mx-psi i seem to recall this was only a problem if a package the collector contrib repo depends on also depends on the internal package being created. If not it's not an issue i *think*, we can always prove this by merging the change as is, and updating contrib to see if it fails w/ the errors we were seeing w/ featuregates

mx-psi:
Alright, I am going to try merging this then and if the `make update-otel` goes well

mx-psi:
Need #11029 next

PR #11026, opened 2024-08-30
[service] implement a noop tracer provider

codeboten:
This is to address some of the memory concerns around the SDK's noop tracer provider. Instead of using the published noop tracer provider, we implement our own to avoid the cost of context propagation, which has been proven to be high enough to cause issues for end users.

Part of #10858

### DIFF
--- a/.chloggen/codeboten_noop-providers.yaml
+++ b/.chloggen/codeboten_noop-providers.yaml
@@ -1,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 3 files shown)


mx-psi (commented):
Is there any upstream issue about the
>  memory concerns around the SDK's noop tracer provider 

?

codeboten:
> Is there any upstream issue about the

I attended the go SIG on 28-Aug-2024, and the end result of the discussion was that because the no-op tracer still has to propagate context, the proposed solution if we need an alternative was to implement our own no-op tracer provider.

mx-psi (approved):

TylerHelmuth (approved):

PR #11042, opened 2024-09-04
[exporter] [chore] move exporter.exporterhelper.request to internal

sfc-gh-sili:
#### Description

This PR moves `exporter/exporterhelper/request.go` to `exporter/internal/request.go` to avoid circular dependency.

Context: As part of the effort to move from a queue->batch pushing model to a queue->batch pulling model in exporter, we will be depending on `Request` from `exporter/exporterbatcher`. However, `exporter/exporterhelper` already depends on `exporter/exporterbatcher`, so we need to move `Request` out of `exporter/exporterhelper`

#### Link to tracking issue

#### Testing

Ran `opentelemetry-collector$ make` to make sure all tests still pass.


exporter/exporter.go
```diff
@@ -22,6 +22,8 @@ type Logs = internal.Logs
 // Settings configures exporter creators.
 type Settings = internal.Settings
 
+type Request = internal.Request
```
mx-psi, line 25: Request is a concept exclusive to the exporterhelper, right? I don't think it should be exported in `exporter`. Why not just do this re-exporting in exporterhelper?

mx-psi (changes requested):

exporter/exporter.go
sfc-gh-sili: That makes sense. Thanks for the feedback!
I was trying to follow the pattern of the re-exporting things in `exporter` but I agree having it in `exporterhelper` makes much sense in this case.

exporter/exporterhelper/exporterhelper.go
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterhelper // import "go.opentelemetry.io/collector/exporter/exporterhelper"
+import "go.opentelemetry.io/collector/exporter/internal"
+
+// Request represents a single request that can be sent to an external endpoint.
+// Experimental: All the following APIs are at the early stage of development and may change without backward compatibility
+// until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
+type Request = internal.Request
+type RequestErrorHandler = internal.RequestErrorHandler
```
mx-psi, line 11: Can we keep the original docstring?  This will not render correctly, it will just add docs to `Request` but not to `RequestErrorHandler`

sfc-gh-sili: Makes sense. Thank you, Pablo.

mx-psi (approved):

Issue #10370, opened 2024-06-10
[ConfMap] - Providing custom merge method for koanf's Map.Merge


jagan2221:
**Is your feature request related to a problem? Please describe.**
go.opentelemetry.io/collector/confmap exposes Merge method currently. I am writing a custom provider which uses confMap, I am trying to implement custom merge logic instead of default merging in underlying koanf map. Seems this is possible in koanf's library, by providing custom merge logic to koanf.Merge method, But confmap exposes normal merge currently and not custom merge functionality.

Ref-> https://github.com/knadh/koanf/blob/ef5fbbc9b413b38887e95d5600a330da5a842aa3/koanf.go#L405

**Describe the solution you'd like**
Expose an additional Merge method in confMap which takes in custom merge function as argument and pass it to koanf's merge.
confMap.Merge(in *Conf, opts *options)

VihasMakwana:
@jagan2221 are you actively working on this?

Issue #8754, opened 2023-10-25
Append components flag


djaglowski:
**Is your feature request related to a problem? Please describe.**
When merging configurations, it would be helpful to be able to do the following:
- Add a receiver, processor, or exporter to an existing pipeline
- Add an extension to the service

For example, consider the following configs.

```yaml
# main.yaml
receivers:
  otlp/in:
processors:
  batch:
exporters:
  otlp/out:
extensions:
  file_storage:

service:
  pipelines:
    traces:
      receivers: [ otlp/in ]
      processors: [ batch ]
      exporters: [ otlp/out ]
  extensions: [ file_storage ]
```

```yaml
# extra_receiver.yaml
receivers:
  foo:

service:
  pipelines:
    traces:
      receivers: [ foo ]
```

```yaml
# extra_extension.yaml
extension:
  bar:

service:
  extensions: [ bar ]
```

If a user provides the above configs in the same order, it will clobber `service::pipelines::traces::receivers` and `service::extensions`, effectively removing the `otlp/in` receiver and `file_storage` extension from the configuration.

**Describe the solution you'd like**
I propose that these lists are components are a special case and that we should provide the _option_ to append to them, rather than clobber them.

Specifically I propose that we add a flag to the collector, e.g. `--append-components` or `--merge-components-mode=append`, which will enable this behavior when merging configs.

**Describe alternatives you've considered**
The same idea could be applied at a more granular level. A simple option would be to have a similar series of flags, e.g. `--append-receivers`, ...

**Additional context**
Several recent SIG meetings have included some discussion of idea along these lines. Additionally, some issues may be related.
- https://github.com/open-telemetry/opentelemetry-collector/issues/8394
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18509#issuecomment-1728506617

pavolloffay:
+1 on the proposal. 

Are there any updates on this?

tigrannajaryan:
> Specifically I propose that we add a flag to the collector, e.g. `--append-components` or `--merge-components-mode=append`, which will enable this behavior when merging configs.

Why does this need to be an option? Is there a case where appending the pipeline lists is undesirable?

Also, the list-appending logic should be name-aware so that if the same-named component is in both input lists it only appears once in the destination list.

djaglowski:
> Why does this need to be an option? Is there a case where appending the pipeline lists is undesirable?

[...]

Issue #8394, opened 2023-09-12
Allow instantiating extensions in separate configs


evan-bradley:
**Is your feature request related to a problem? Please describe.**

There is currently no way to easily instantiate extensions across multiple configuration sources. The default behavior in koanf is to [overwrite slices](https://github.com/knadh/koanf#custom-merge-strategies) when merging multiple maps. This results in behavior where the `service.extensions` key will be overridden and take the value of the final source if it is set multiple times.

Take the following two configuration files:

```yaml
extensions:
  health_check:
    endpoint: localhost:61234

service:
  extensions: [health_check]
```

```yaml
extensions:
  zpages:

service:
  extensions: [zpages]
```

If I start the Collector with the following, only the zpages extension will be started.

```
otelcol --config=health_check.yaml --config=zpages.yaml
```

The effective config will look like:

```yaml
extensions:
  health_check:
    endpoint: localhost:61234
  zpages:

service:
  extensions: [zpages]
```

**Describe the solution you'd like**
At a minimum, provide a naming scheme similar to what we provide for pipelines to allow config sources to instantiate extensions they define. This could look like:

```yaml

service:
  extensions/health: [health_check]
  extensions/zpages: [zpages]
```

**Describe alternatives you've considered**

A more comprehensive solution could be to override the default koanf merging strategy to support merging slices. I would propose the following rules for merging slices:

1. Slices values are not deduplicated
2. Slices are merged by appending slices to previous values.

The caveat with changing this behavior is that it would be more difficult to set slice values to empty lists, which would now require setting to `nil` (or another primitive value) before setting them to an empty list.
 
We could also simply recommend that extensions are all instantiated in a single file, but in my view this is not a good solution.

**Additional context**

It's possible I've missed an existing mechanism for this, but I've looked through the code and docs and don't see an any way to accomplish this. If I've missed something we should make it more visible.

PR #11046, opened 2024-09-04
[confmap] - New merging mode to combine components

VihasMakwana:
Koanf's default merging strategy currently overrides static values such as slices, numbers, and strings. However, lists of components should be treated as a special case. This pull request introduces a new feature gate to allow for merging lists instead of discarding the existing ones.

With this new merging strategy:
- Lists under the `service` configuration are merged rather than replaced.
- The merging logic is name-aware, meaning that if components with the same name appear in both lists, they will only appear once in the final merged list.
- This strategy only affects configurations under `service::*`.

#### Link to tracking issue
Related issues:
- https://github.com/open-telemetry/opentelemetry-collector/issues/8754
- https://github.com/open-telemetry/opentelemetry-collector/issues/8394
- https://github.com/open-telemetry/opentelemetry-collector/issues/10370

#### Testing
- Added

#### Documentation
- _To be added_

Note: I’d appreciate your feedback on this 🙏


VihasMakwana:
@codeboten @open-telemetry/collector-approvers  I would appreciate any feedback you guys might have!

mx-psi:
I think we need to discuss more on #8754 before moving forward with any concrete solution

VihasMakwana:
> I think we need to discuss more on #8754 before moving forward with any concrete solution

Sure. Let's discuss more on #8754.

I'll move this PR to draft then.

PR #11048, opened 2024-09-04
Pprofile marshal/unmarshalling and gRPC implementations

dmathieu:
#### Description

This adds pprofile marshaling/unmarshaling, as well as the gRPC implementation, so we can start setup the OTLP receivers and exporters.


.chloggen/pprofile-profile-id-type-change.yaml
mx-psi, line 1: Since I can't see any users of this API according to this search https://github.com/search?q=-repo%3Aopen-telemetry%2Fopentelemetry-collector++go.opentelemetry.io%2Fcollector%2Fpdata%2Fpprofile+language%3AGo+-is%3Afork&type=code I think we can do this change in one step.

mx-psi (approved):

PR #11049, opened 2024-09-04
Experiment with new consumererror API

evan-bradley:
This shows how the proposed changes in #9041 may look.


evan-bradley:
This was discussed at an offline meeting on 2024-09-04 and has served its purpose, so I'm closing it.

PR #11051, opened 2024-09-04
otelcol: fix windows event log core placement

braydonk:
#### Description
In the CollectorSettings LoggingOptions, when the options are `zap.WrapCore`, the final core wrapped will end up running first. For this reason, the windowsEventLogCore must be the first option in the list of user supplied options, to ensure user supplied options are run before the core that writes to the Windows Event Log.

Prior art: I fixed this before in #5298.

#### Link to tracking issue
#5297 

#### Testing
Testing was manual by running the Collector as a service after this change. I tried to add a unit test but could not find an effective way to do that due to all the type indirection done by the `zap` package. I was hoping to find an easy way to tell that the `windowsEventLogCore` is the first core wrapped in the options but I could not find a way to pull that off. Open to ideas if anyone can find a way.


otelcol/collector_windows.go
```diff
@@ -141,6 +138,19 @@ func openEventLog(serviceName string) (*eventlog.Log, error) {
 	return elog, nil
 }
 
+func loggingOptionsWithEventLogCore(
```
braydonk, line 141: I extracted this to a function for 2 reasons:
1. I was trying to write a unit test for it (failed to find a way).
2. I wanted the comment explaining why the prepend is necessary to really stand out for any future contributors looking at this.

braydonk:
I don't yet understand why this breaks the `LogToFile` test. It seems like this test is painful to run locally so might take a bit for me to get it set up and figure out what's going on.

braydonk:
This test is broken for me on main, doesn't seem related to my changes. I think I'll need help to figure out why, I am super lost at this point. @pjanotti any chance you can help me figure out why this test appears to be busted?

pjanotti:
Thanks for spotting this @braydonk

> the windowsEventLogCore must be the first option in the list of user supplied options, to ensure user supplied options are run before the core that writes to the Windows Event Log.

So there were some options in telemetry, e.g.: log level, being ignored?

otelcol/collector_windows.go
```diff
@@ -141,6 +138,19 @@ func openEventLog(serviceName string) (*eventlog.Log, error) {
...
+	elog *eventlog.Log,
+	serviceConfig **service.Config,
+	userOptions []zap.Option,
+) []zap.Option {
+	return append(
+		// The event log core must run last, so it needs to be the first option
```
pjanotti, line 147: Given that I already changed that order in the past (and can't recall if by accident or intentionally) explicit wording seems advisable: 
```suggestion
		// The order below must be preserved - see PR #11051
		// The event log core must run last, so it needs to be the first option
```

.chloggen/fix_windows_service_event_log_core.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
pjanotti, line 25: Isn't the result of the change visible to `user`?

otelcol/collector_test.go
```diff
@@ -564,6 +564,8 @@ func newEnvProvider() confmap.ProviderFactory {
 			return confmap.NewRetrieved(float64(6.4))
 		case "env:BOOL":
 			return confmap.NewRetrieved(true)
+		case "env:ProgramData":
```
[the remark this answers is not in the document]
braydonk: Left in by accident from manual tests. Removed in [d8e89c2](https://github.com/open-telemetry/opentelemetry-collector/pull/11051/commits/d8e89c2daaef6860fbb23191572b7b535ca97edb)

pjanotti (approved):
LGTM

mx-psi (approved):

PR #11054, opened 2024-09-04
Avoid public APIs with internal params in custom Options

bogdandrutu:


confmap/provider.go
```diff
@@ -112,21 +112,29 @@ type retrievedSettings struct {
 }
 
 // RetrievedOption options to customize Retrieved values.
-type RetrievedOption func(*retrievedSettings)
+type RetrievedOption interface {
```
bogdandrutu, line 115: Not a breaking change because users could not create `func(*retrievedSettings)` types anyway because `retrievedSettings` was private.

atoulme:
I think ideally those tests would be tested with a test using a different package name to make sure we can validate the APIs work outside of the module.

atoulme (approved):

PR #11002, opened 2024-08-29
[configtelemetry] Mark module as stable

mx-psi:
#### Description

Fixes #10277

Depends on #10912

Issue #11027, opened 2024-09-02
[configtelemetry] Remove it?


mx-psi:
Personal opinion, I would try first to completely remove this module instead of making it stable. Why?
* Otel doesn't have a concept of telemetry level and us in otel-collector inventing this is not the right thing.
* This concept was added to control "expensive" telemetry (more exactly only metrics), but that can be now controlled with views.

_Originally posted by @bogdandrutu in https://github.com/open-telemetry/opentelemetry-collector/issues/11002#issuecomment-2322504322_

mx-psi:
Putting this into a release:blocker issue to avoid releasing the module if we don't want to

TylerHelmuth:
Both points are valid, but based on the discussions we had previously around this feature the benefits outweighed the downside of us define the concept. They were:

1. Views are hard for end-users to comprehend and implement. Additionally, introducing the concept of levels for metrics does not prevent the end user from using views if they want to.
2. We've already had metric level in place for a long time and users are using it.

We've also updated mdatagen so that all components within our control document their own telemetry and under which metric level the metrics is emitted: https://github.com/open-telemetry/opentelemetry-collector/blob/15542e67e1346a87273a8c83c7f44582eab3874e/processor/batchprocessor/metadata.yaml#L12-L50.

I feel this is a time that it is beneficial to deviate from the spec to help our end users.

PR #11055, opened 2024-09-04
Revert "[configtelemetry] Mark module as stable"

bogdandrutu:
Reverts open-telemetry/opentelemetry-collector#11002

Per discussion today 09/04 we will postpone the stabilization of this module until we have a good understanding of where this will exactly be used. The decision was that level will not be used in the component, so stabilization of this model no longer blocks progress on component module stabilization.


mx-psi:
Since we are going to use this for sure in the debug exporter, I mentioned on https://github.com/open-telemetry/opentelemetry-collector/issues/11027#issuecomment-2329480381 that we should just move forward with this. Is that not the case?

bogdandrutu:
@mx-psi if we end up using only there, then I would move it to that module.

mx-psi (approved):

codeboten (approved):
It's ok to revert, this is currently used in the debug exporter, and in service config for both metrics and traces.

PR #11058, opened 2024-09-05
PAUSED - [exporter] Split an API of exporter/queue: from pop(updateSizeCallback) to pop() and updateSize(deltaSize)

sfc-gh-sili:
#### Description

This PR splits an API of `exporter/internal/sized_channel`:
`pop(updateSizeCallback func(int))` => `pop()` and `updateSize(deltaSize)`

Why?
As part of the effort to move from a queue->batch pushing model to a pulling model, we would like to parallelize reading especially for persistent queues. "popping an item from the sized channel" in this context, is more of a signal that the queue is non-empty and ready to be read. We would like to enable the caller to complete the following READ process asynchronously and then report back.

POC: https://github.com/sfc-gh-sili/opentelemetry-collector/pull/2

This change does not affect any user because queue is an internal package. Nor does this PR change any behavior.

#### Link to tracking issue

#### Testing
Ran `opentelemetry-collector$ make` to make sure all tests still pass.

### DIFF
--- a/exporter/internal/queue/sized_channel.go
+++ b/exporter/internal/queue/sized_channel.go
@@ -62,22 +62,19 @@ func (vcq *sizedChannel[T]) push(el T, size int64, callback func() error) error 
 // pop removes the element from the queue and returns it.
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
-// The callback is called before the element is removed from the queue. It must return the size of the element.
-func (vcq *sizedChannel[T]) pop(callback func(T) (size int64)) (T, bool) {
+func (vcq *sizedChannel[T]) pop() (T, bool) {
 	el, ok := <-vcq.ch
-	if !ok {
-		return el, false
-	}
+	return el, ok
+}
 
-	size := callback(el)
-
+// updateSize updates size. It should be called when an item is removed from the queue.
+func (vcq *sizedChannel[T]) updateSize(deltaSize int64) {
 	// The used size and the channel size might be not in sync with the queue in case it's restored from the disk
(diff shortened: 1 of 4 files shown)


exporter/internal/queue/sized_channel.go
```diff
@@ -62,22 +62,19 @@ func (vcq *sizedChannel[T]) push(el T, size int64, callback func() error) error
...
-	if !ok {
-		return el, false
-	}
-
-	size := callback(el)
+	return el, ok
+}
 
+// updateSize updates size. It should be called when an item is removed from the queue.
+func (vcq *sizedChannel[T]) updateSize(deltaSize int64) {
```
dmitryax, line 71: I like how this PR simplifies the code a bit. However, we increase coupling here: clients are now responsible for managing the queue size after popping an item. Is there another way we can keep the queue size management inside `sizedChannel` and get the new pulling behavior we need?

sfc-gh-sili: The awkwardness comes from that we only know the size of the item after reading is complete, and reading is slow so we want to parallelize.

In the new model, we want to send a cue to the next worker to standby once the current worker is able to acquire the item (but before actually starting the read). Putting that to pseudocode

```
_, item <- sizedChannel.ch // 1
sendCueToTheNextWorker() // 2
request = read() // 3
updateSize(size(request)) // 4
```

Now let's assume we are putting all above steps inside `sizedChannel`, then step 3 and 4 need to be callbacks passed from `queue` as a callback; step 2 need to be a callback passed from `queueBatcher` to `queue` then to `sizedChannel`. I find that too many layers intertwined together.

Let me know what you think - I am open to ideas

Issue #11015, opened 2024-08-29
Exporter/Internal/queue size_channel.pop() is blocked even when queue is full


timannguyen:
**Describe the bug**

[size_channel.pop()](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/internal/queue/sized_channel.go#L67) is blocking when queue is full. This looks to happen when there are high number of connections and high throughput.

**Steps to reproduce**

TBA. will be writing a test to replicate

**What did you expect to see?**

size_channel.pop() should receive data and returns when queue is not empty.

**What did you see instead?**

size_channel.pop() is blocked indefinitely even if queue is full.

**What version did you use?**

v0.103.0

**What config did you use?**

```
settings:
  connectors:
    routing/system:
      error_mode: ignore
      table:
        - pipelines:
            - logs/system_telemetry.routing/system
          statement: route() where attributes["Into"] == ".splunk.edge.system.telemetry_stream_dataset"
        - pipelines:
            - logs/dd_3cfb1516-3b4d-472c-b065-7c739f5f8144.routing/system
          statement: route() where attributes["Into"] == ".splunk.edge.system.unhandled"
  exporters:
    S2S/shared.pipelines.default_splunk_cloud_destination:
      ackEnabled: false
      caCertificate: '[redacted]'
      clientCertificate: '[redacted]'
      clientPrivateKey: '[redacted]'
      compressionLevel: 0
      connectTimeoutMillis: 5000
      connection: scpbridge
      datasetName: shared.pipelines.default_splunk_cloud_destination
      disablePersistentQueue: false
      endpoints: '[redacted]'
      handshakeMaxRetries: 3
      handshakeTimeoutMillis: 1000
      heartbeatIntervalMillis: 30000
      internalName: shared.pipelines.default_splunk_cloud_destination
      retryInitialIntervalMillis: 5000
      retryMaxElapsedTimeMillis: 0
      retryMaxIntervalMillis: 30000
      s2sConnectionTtlDuration: 30s
      s2sConnectionsPerTarget: 0
      s2sFreezeDuration: 10m0s
      sendQueueNumConsumers: 10
      sendQueueSize: 10000
      sendTimeoutMillis: 5000
      serviceInstanceID: a7bc7b90-609e-11ef-aac8-002248496f87
    S3/acies_logs:
      awsRegion: us-east-1
      dataFormat: hecJson
      datasetName: product-telemetry
      env: production
      flush_timeout: 1h
[...]

PR #11063, opened 2024-09-05
return queue is full error if sized_channel is full

splunkericl:
#### Description
This change fixes a potential deadlock bug for persistent queue.

There is a race condition in persistent queue that caused `used` in `sizedChannel` to become out of sync with `ch` len. This causes `Offer` to be deadlocked in specific race condition. For example:
1. Multiple consumers are calling Consume
2. Multiple producers are calling Offer to insert into the queue
  a. All elements are taken from consumers. ch is empty
3. One consumer completes consume, calls onProcessingFinished
  a. Inside sizedChannel, syncSize is invoked, used is reset to 0 when other consumers are still waiting for lock to consume
4. More Offer is called inserting elements -> used and ch len should equal
5. As step 3a consumers completes, used is decreased  -> used is lower than ch len
  a. More Offer is called inserting since used is below capacity. however, ch is full.
  b. goroutine calling offer is holding the mutex but can’t release it as ch is full.
  c. no consumer can acquire mutex to complete previous onProcessingFinished

This change returns an error if channel is full instead of waiting for it to unblock.

#### Link to tracking issue
Fixes # https://github.com/open-telemetry/opentelemetry-collector/issues/11015

#### Testing
- Added concurrent test in persistent queue that can reproduce the problem(note: need to re-run it 100 times as the race condition is not consistent).
- Added unit test for sizedChannel

#### Documentation
Added comment in the block explaining it

### DIFF
--- a/exporter/internal/queue/persistent_queue_test.go
+++ b/exporter/internal/queue/persistent_queue_test.go
@@ -593,7 +593,6 @@ func TestPersistentQueueStartWithNonDispatchedConcurrent(t *testing.T) {
 	case <-doneCtx.Done():
 		assert.Fail(t, "timed out waiting for producers to complete")
 	}
-	
 	assert.Zero(t, pq.sizedChannel.Size())
 }
 
--- a/exporter/internal/queue/persistent_queue_test.go
+++ b/exporter/internal/queue/persistent_queue_test.go
@@ -539,12 +539,12 @@ func TestPersistentQueueStartWithNonDispatchedConcurrent(t *testing.T) {
 	pq := createTestPersistentQueueWithItemsCapacity(t, ext, 250)
 
 	proWg := sync.WaitGroup{}
-	for j := 0; j < 50; j++ {
+	for j := 0; j < 2; j++ {
 		proWg.Add(1)
 		go func() {
 			defer proWg.Done()
 			// Put in items up to capacity
-			for i := 0; i < 10; i++ {
+			for i := 0; i < 250; i++ {
 				for {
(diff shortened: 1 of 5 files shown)


splunkericl:
Fixed all the checks. @dmitryax @mx-psi can you guys take a look at this PR? The change is small and mostly just testing.

splunkericl:
cc @sfc-gh-sili as well as snowflake partner is helping out

exporter/internal/queue/persistent_queue_test.go
```diff
@@ -531,6 +532,69 @@ func TestPersistentQueueStartWithNonDispatched(t *testing.T) {
...
+	case <-conDone:
+	case <-doneCtx.Done():
+		assert.Fail(t, "timed out waiting for consumers to complete")
+	}
+
+	select {
+	case <-proDone:
+	case <-doneCtx.Done():
+		assert.Fail(t, "timed out waiting for producers to complete")
+	}
```
dmitryax, line 595: Can you please add this check as well?

```suggestion
	}
	
	assert.Zero(t, pq.sizedChannel.Size())
```

dmitryax (approved):
LGTM

exporter/internal/queue/persistent_queue_test.go
```diff
@@ -531,6 +532,71 @@ func TestPersistentQueueStartWithNonDispatched(t *testing.T) {
...
+
```
dmitryax, line 596:
```suggestion

```

PR #11064, opened 2024-09-05
Migrate collector semconv codegen to weaver

jsuereth:
NOTE: This PR is (currently) an example for discussion on the evolution of semconv codegen in the collector.

I regenerated v1.27.0 semconv package so you can get a feel for the changes - I can revert that if we go this route.

#### Description
Migrate semconv codegen to weaver.

Discovered a few things with codegen in the collector that may be problematic going forward, but did the bare minimum here:

- Migrate from semconvgen docker image (transitive dependency from opentelemetry-go-build-tools).  This image is deprecated, will break in the future
- Split out codegen for event, resource + trace.  These were not producing attributes today, and going forward I think some time/thought should be made into what this looks like.  Semantic conventions has shifted direction overtime, and we have recommendations for codegen that we should co-design a new solution w/ Collector maintainers.
- Update docker calls to run as local user rather than root (meaning generated files are no longer owned by root).
- Remove requirement level from attribute registry (this was never an accurate description and any reliance on it was fundamentally broken).
- Update documentation to leverage automatic git resolution of semconv tags built into weaver.

#### Link to tracking issue
Tracked at: https://github.com/open-telemetry/weaver/issues/227

#### Testing
I manually ran the codegen against latest semconv and compared agianst current versions. I can run this codegen in existing directories for clarity.

#### Documentation
Documentation remains unchanged for codegen.  I can add links to weaver documentation if needed.

#### Open Questions / Concerns

- The Collector, unlike other systems, does NOT fragment stable vs. unstable portions of semantic conventions in codegen, instead having docs outline the current stability of constants.  This might be ok given the versioned-packages, want to ensure that's true.
- The existing collector semconvgen template *appears* to filter on stable attributes, but generated code does not, making me think perhaps the filter wasn't doing what was expected or was broken over time.
- The codegen around events, traces and resource appears to not be producing anything.  I think we can improve that, but I'm curious what collector authors/maintainers need here.  For now I set up the ability to define a template and preserve existing behavior, but the previous template would not have produced anything useful for these, so I did not duplicate its logic.


semconv/weaver/registry/collector/weaver.yaml
```diff
@@ -0,0 +1,136 @@
...
+      .groups
+      | map(select(.type == "event"))
+      | map(. + {root_namespace: (if .id | index(".") then .id | split(".") | .[1] else "other" end)})
+      | sort_by(.root_namespace)
+    application_mode: single
+    file_name: generated_event.go
+  - pattern: trace.go.j2
+    filter: >
+      .groups
+      | map(select(.type == "trace"))
```
lmolkova, line 21: should it be span?

semconv/weaver/registry/collector/attribute_group.go.j2
```diff
@@ -0,0 +1,75 @@
...
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Code generated from semantic convention specification. DO NOT EDIT.
+
+package semconv
+
+{% for group in ctx %}
+{% if group.attributes | length > 0 %}
+{# TODO - we're grouping by registry namespace, not attribute gorup, so we lose group docs #}
```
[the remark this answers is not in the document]
jsuereth: fixed

semconv/weaver/registry/collector/weaver.yaml
jsuereth: Yes, fixed

semconv/v1.27.0/generated_attribute_group.go
```diff
@@ -5,165 +5,217 @@
...
 const (
-	// The provenance filename of the built attestation which directly relates to the
-	// build artifact filename. This filename SHOULD accompany the artifact at publish
-	// time. See the SLSA Relationship specification for more information.
+
+	// The provenance filename of the built attestation which directly relates to the build artifact filename. This filename SHOULD accompany the artifact at publish time. See the [SLSA Relationship] specification for more information.
 	//
+	// Stability: Experimental
 	// Type: string
-	// Requirement Level: Optional
```
lmolkova, line 31: requirement level disappeared

lmolkova (approved):
Weaver/template changes LGTM!

semconv/v1.27.0/generated_attribute_group.go
jsuereth: Yes - There is not such things are requirement level on the attribute registry, that's signal specific.

I'm not even sure where it was pulling requirement level from before, but I call that out in the PR description.

PR #11071, opened 2024-09-06
Profiles support in OTLP receiver

dmathieu:
#### Description

This adds profiles support to the OTLP receiver.

cc @mx-psi


receiver/otlpreceiver/config.go
```diff
@@ -32,6 +32,9 @@ type HTTPConfig struct {
 
 	// The URL path to receive logs on. If omitted "/v1/logs" will be used.
 	LogsURLPath string `mapstructure:"logs_url_path,omitempty"`
+
+	// The URL path to receive profiles on. If omitted "/v1/profiles" will be used.
+	ProfilesURLPath string `mapstructure:"profiles_url_path,omitempty"`
```
mx-psi, line 37: I can't find the `/v1/profiles` part on the spec anywhere. I would prefer to remove this option for now to not have anything profiling specific in the configuration until profiling is more stable

dmathieu: Sure, I can remove the option. But how would the HTTP handler decide which endpoint it listens to?

dmathieu: Also, the only place where signal paths are specificed is for SDK exporters, which doesn't make much sense to add profiles to.

mx-psi: > But how would the HTTP handler decide which endpoint it listens to?

We can hardcode `/v1/profiles` for now if that's what the profiling SIG believes will end up being used?

> Also, the only place where signal paths are specificed is for SDK exporters, which doesn't make much sense to add profiles to.

The OTLP specification mentions it for all other signals https://opentelemetry.io/docs/specs/otlp/#otlphttp-request

receiver/receiverhelper/obsreport.go
```diff
@@ -121,6 +122,24 @@ func (rec *ObsReport) EndMetricsOp(
 	rec.endOp(receiverCtx, format, numReceivedPoints, err, component.DataTypeMetrics)
 }
 
+// StartProfilesOp is called when a request is received from a client.
```
mx-psi, line 125: I would also prefer to leave telemetry out at least until we figure out open-telemetry/opentelemetry.io/pull/5092 and we can mark this telemetry as unstable properly

receiver/otlpreceiver/config.go
dmathieu: Ah, I was looking in the spec, and the OTLP spec is in the proto repository 🤦 
I have made the config static.

mx-psi:
I think we should wait until https://github.com/open-telemetry/opentelemetry-proto/pull/582 is merged (I don't mind if it's merged but not released, but I would like some sort of consensus)

mx-psi:
@dmathieu Can you fix the merge conflicts? I can merge after that

dmathieu:
Done.

mx-psi (approved):

bhaskarbanerjee:
@dmathieu any movement on this PR. Reading 'While the signal is still in development and isn’t yet recommended for production use, the Profiling SIG has made substantial progress on many fronts.' on the [blog](https://opentelemetry.io/blog/2024/state-profiling/) does it imply that another PR made it through? I'm little confused with the state of the PR and verbiage on the blog. Seeking your thoughts

dmathieu:
This PR was merged last september. It has been released multiple times since.

PR #11073, opened 2024-09-06
[chore][component/componenttest] Codify comments' requirements

crobert-1:
#### Description
When using the `TestTelemetry` struct, functionality requires that it's setup before checking metric values. This requirement was stated in method comments for each getter method. The suggestion was made [here](https://github.com/open-telemetry/opentelemetry-collector/pull/11068#pullrequestreview-2285280440) that rather than just having a comment, we should panic if `SetupTelemetry` isn't called, to enforce this requirement and avoid incorrect usages.

#### Link to tracking issue
Related to #11068

#### Testing
This is a test only change.

### DIFF
--- a/component/componenttest/obsreporttest.go
+++ b/component/componenttest/obsreporttest.go
@@ -36,6 +36,7 @@ const (
 )
 
 type TestTelemetry struct {
+	isSetup      bool
 	ts           component.TelemetrySettings
 	id           component.ID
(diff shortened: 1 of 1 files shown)


crobert-1:
_Failing test is frequency of https://github.com/open-telemetry/opentelemetry-collector/issues/10927, I've added frequency there._

PR #11079, opened 2024-09-06
use MeterProvider via the config package

codeboten:
This updates the Collector to configure the MeterProvider via the config package instead of its own configuration. This allows us to remove a bunch of code that is duplicated, and to add support for views pretty quickly as a result, with very little code.

Leaving this in draft, until the config package can support an exclude filter for configuring stream attribute keys as described in https://github.com/open-telemetry/opentelemetry-specification/pull/4188


codeboten:
This is waiting on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6126

codeboten:
the work in this PR will be done in https://github.com/open-telemetry/opentelemetry-collector/pull/11458

PR #11082, opened 2024-09-08
Refine scraperhelper.NewScraper to take component.Type directly

### LINKED ISSUE #9473: `scraperhelper.NewScraper` is awkward to use

mrasu:
#### Description

Create new function to  enable to pass `component.Type` to `scraperhelper.NewScraper`

Because many contrib scrapers name the scraper with the same name as the component itself, create a function to take `component.Type`.
For example, https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0e2bea5ac2763421e1a3943093f67184e109265f/receiver/activedirectorydsreceiver/factory_windows.go#L35 passes `metadata.Type.String()` as a name of scraper. With this change, you can pass `metadata.Type` instead 

#### Link to tracking issue
Fixes #9473

### DIFF
--- a/.chloggen/refine-new-scraper.yaml
+++ b/.chloggen/refine-new-scraper.yaml
@@ -10,7 +10,7 @@ component: scraperhelper
 note: deprecate NewScraper, should use NewScraperWithComponentType
 
 # One or more tracking issues or pull requests related to the change
-issues: []
+issues: [11082]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/refine-new-scraper.yaml
+++ b/.chloggen/refine-new-scraper.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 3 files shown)


mx-psi:
@mrasu Can you open a PR on https://github.com/open-telemetry/opentelemetry-collector-contrib to make all components use the new API?

mrasu:
@mx-psi ok, I will

mrasu:
Thanks to @djaglowski, optenetelemetry-collector-contrib use the new function now :smiley_cat: 
[open-telemetry/opentelemetry-collector-contrib@`924040b` (#35073)](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35073/commits/924040ba1d2de382740b6299a3aba96593165b57)

PR #10910, opened 2024-08-19
Add ingoing and outgoing counts to processorhelper

djaglowski:
#### Description

Implements ingoing and outgoing counts as described in https://github.com/open-telemetry/opentelemetry-collector/issues/10708.

Issue #10708, opened 2024-07-23
Simple processor metrics


djaglowski:
**Is your feature request related to a problem? Please describe.**

We currently [define a set of metrics for processors](https://github.com/open-telemetry/opentelemetry-collector/blob/21c3f3643dc895d6b7a721fb2a86d41eedf53562/processor/processorhelper/obsreport.go#L63-L121) which cannot be automatically recorded by the collector framework, and as far as I can tell are not _actually_ recorded by any processors in core or contrib. In short, the metrics describe the number of items accepted, dropped, inserted, and refused. (I believe these metrics are loosely inspired by proposals such as https://github.com/open-telemetry/oteps/pull/259, but no such design has actually landed.)

We also have proposals such as https://github.com/open-telemetry/opentelemetry-collector/issues/10355 which would measure size of data according to the same categories, but again, it is not clear that there is any path to automatically recording these values. In the case of size metrics, it would additionally be very difficult for processor developers to accurately capture such measurements because it's unclear how to attribute inserted, dropped, or refused _sizes_ as a portion of an overall payload.

**Describe the solution you'd like**

Looking at this from a collector implementation perspective, what we clearly _can_ observe automatically, is the data being passed in and out of each processor. It would be difficult to determine precisely which parts of that data were inserted or deleted, but we can certainly count the number of items and measure the size of the payload (according to a standard such as [`ProtoMarshaler.Sizer()`](https://github.com/open-telemetry/opentelemetry-collector/blob/9907ba50df0d5853c34d2962cf21da42e15a560d/pdata/ptrace/pb.go#L11)).

Therefore, I propose that we should define new metrics in `processorhelper` which are _automatically_ recorded. If another more nuanced proposal eventually lands, we can add those later.

Importantly, in order to handle error propagation consistently, these metrics would be recorded _after the processor successfully passes data to the next consumer_. This can be accomplished within `processorhelper` by looking at data before and after the processing function, and taking into consideration the result of passing the data along to the next consumer.

Specifically, I propose the following metrics per data type. I'm showing only logs but assume there would be equivalents for metrics and traces.

Incoming and outgoing counts. 
```yaml
    otelcol_processor_incoming_log_records: # or processor_consumed_log_records
      enabled: true
      description: Number of log records passed to the processor.
      unit: "{records}"
      sum:
        value_type: int
        monotonic: true
    otelcol_processor_outgoing_log_records: # or processor_emitted_log_records
      enabled: true
[...]

PR #11083, opened 2024-09-09
[processorhelper] deprecating unused methods

codeboten:
`[Logs|Traces|Metrics]Inserted` was not used in the core/contrib repos, marking them as deprecated.

### DIFF
--- a/processor/processorhelper/obsreport.go
+++ b/processor/processorhelper/obsreport.go
@@ -120,6 +120,8 @@ func (or *ObsReport) TracesDropped(ctx context.Context, numSpans int) {
 }
 
 // TracesInserted reports that the trace data was inserted.
+//
+// Deprecated: [v0.109.0] This method was not used in core/contrib and it's unclear when it should have been used.
 func (or *ObsReport) TracesInserted(ctx context.Context, numSpans int) {
 	or.recordData(ctx, component.DataTypeTraces, int64(0), int64(0), int64(0), int64(numSpans))
 }
@@ -140,6 +142,8 @@ func (or *ObsReport) MetricsDropped(ctx context.Context, numPoints int) {
 }
 
 // MetricsInserted reports that the metrics were inserted.
+//
+// Deprecated: [v0.109.0] This method was not used in core/contrib and it's unclear when it should have been used.
 func (or *ObsReport) MetricsInserted(ctx context.Context, numPoints int) {
 	or.recordData(ctx, component.DataTypeMetrics, int64(0), int64(0), int64(0), int64(numPoints))
 }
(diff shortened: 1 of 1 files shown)


bogdandrutu (approved):

djaglowski:
The rationale for these metrics was described in https://github.com/open-telemetry/opentelemetry-collector/issues/10353.

They are complimentary to the `Dropped` metrics in that they can account for items which were added rather than removed by a processor. For example, a processor which computes a metric based on two others would record that it has inserted data points. Arguably, aggregation should be described as both dropping and inserting data. e.g. 10 log records replaced by 1.

In any case, `Accepted`, `Refused`, and `Dropped` are used only where component authors have manually recorded them so in my opinion we should keep or deprecate all of them together.

codeboten:
> In any case, Accepted, Refused, and Dropped are used only where component authors have manually recorded them so in my opinion we should keep or deprecate all of them together.

I agree. Looking through the other metrics, the only processor that makes use of them is the memory limiter processor. I'm not against the idea that processors should report this information. I thought it would make sense to report them via the incoming/outgoing metrics that were added in https://github.com/open-telemetry/opentelemetry-collector/pull/10910 with the `otel.outcome` attribute for being the differentiator as suggested in https://github.com/open-telemetry/oteps/pull/259

djaglowski:
> I thought it would make sense to report them via the incoming/outgoing metrics that were added in #10910 with the `otel.outcome` attribute for being the differentiator as suggested in [open-telemetry/oteps#259](https://github.com/open-telemetry/oteps/pull/259)

I'm certainly in favor of this - just suggesting that we should be keeping or deprecating these metrics as wholistic sets, whichever we choose.

codeboten:
@djaglowski do you want to open an issue to migrate the existing processors to using incoming/outgoing w/ `outcome` attributes and we can reference that issue as we make changes to the telemetry

djaglowski:
I have to walk back my statement as meaning to apply only to the incoming/outgoing metrics. I'm not opposed to the `outcome` idea but I don't understand how we could automatically record this attribute.

djaglowski:
> it wouldn't need to be automatic....

For context, the incoming/outgoing metrics were added based on https://github.com/open-telemetry/opentelemetry-collector/issues/10708 which specifically calls out automatically recorded metrics. 

IMO this is critical. There's going to be _very_ little value in any metrics that must be recorded manually by each and every processor, especially when there's so much nuance to their meaning that the OTEP has gone stale before leaving draft state.

I do think there is a simple version of the outcome dimension that can be automatically recorded based purely on whether the next consumer returned an error, but this is quite different than trying to match the OTEP.

Issue #10467, opened 2024-06-26
[confighttp] Make the compression level configurable


rnishtala-sumo:
Current compression algorithms seem to support multiple compression levels. These levels determine the speed and efficiency of the compression algorithm.

Lower compression levels provide faster speed but larger file sizes. Compression levels could be 3, 5, 7 or 11. Level 3 seems to be the default across multiple compression algorithms. It could benefit a user to give them the option to save on AWS egress costs by configuring a higher compression level with a tradeoff on processing speed.

Example configuration could looks like below:

```
exporter:
  otlphttp:
    endpoint: otelcol2:55690
    auth:
      authenticator: some-authenticator-extension
    tls:
      ca_file: ca.pem
      cert_file: cert.pem
      key_file: key.pem
    headers:
      test1: "value1"
      "test 2": "value 2"
    compression
       type: zstd
       level: 7
    cookies:
      enabled: true
```

rnishtala-sumo:
I like the idea of configuring this as a single string

```
exporter:
  otlphttp:
    endpoint: otelcol2:55690
    auth:
      authenticator: some-authenticator-extension
    tls:
      ca_file: ca.pem
      cert_file: cert.pem
      key_file: key.pem
    headers:
      test1: "value1"
      "test 2": "value 2"
    compression: zstd/7
    cookies:
      enabled: true
```

wasim-nihal:
Hi @mx-psi /@rnishtala-sumo , is this open for contribution?

rnishtala-sumo:
@wasim-nihal yes, please proceed if this issue is of interest to you. I can assist with the PR review.

rnishtala-sumo:
@mx-psi if there's no one working on this, could this issue be assigned to me?

mx-psi:
Assigned!

rnishtala-sumo:
@dmitryax @mx-psi I've raised a PR for this issue. Would appreciate your review.

rnishtala-sumo:
The latest approach to configuring this is as follows

```
exporter:
  otlphttp:
    endpoint: otelcol2:55690
    auth:
      authenticator: some-authenticator-extension
    tls:
      ca_file: ca.pem
      cert_file: cert.pem
      key_file: key.pem
    headers:
      test1: "value1"
      "test 2": "value 2"
    compression
       type: zstd
    compression_params:
       level: 10
    cookies:
      enabled: true
```

PR #11084, opened 2024-09-09
feat(compression): Add the option of configuring compression levels

rnishtala-sumo:
#### Description
Add the option of configuring compression levels. This is not intended to be a breaking change for the user.

#### Link to tracking issue
Issue - https://github.com/open-telemetry/opentelemetry-collector/issues/10467

#### Testing
Unit tests were added/changed

#### Documentation
https://github.com/rnishtala-sumo/opentelemetry-collector/blob/configure-compression-levels/config/confighttp/README.md#client-configuration

### DIFF
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -5,6 +5,8 @@ package configcompression // import "go.opentelemetry.io/collector/config/config
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/klauspost/compress/zlib"
 )
@@ -14,8 +16,8 @@ type Type string
 type Level int
 
 type TypeWithLevel struct {
-	Type  Type  `mapstructure:"type"`
-	Level Level `mapstructure:"level"`
+	Type  Type
+	Level Level
 }
 
 const (
@@ -35,20 +37,39 @@ func (ct *Type) IsCompressed() bool {
 	return *ct != typeEmpty && *ct != typeNone
 }
 
-func (ct *TypeWithLevel) UnmarshalText() (TypeWithLevel, error) {
-	typ := ct.Type
-	if (typ == TypeGzip && isValidLevel(int(ct.Level))) ||
-		(typ == TypeZlib && isValidLevel(int(ct.Level))) ||
-		(typ == TypeDeflate && isValidLevel(int(ct.Level))) ||
-		typ == TypeSnappy ||
-		typ == TypeZstd ||
-		typ == TypeLz4 ||
-		typ == typeNone ||
-		typ == typeEmpty {
-		return TypeWithLevel{Type: typ, Level: ct.Level}, nil
+func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
+	var compressionTyp Type
+	var level int
+	var err error
+	parts := strings.Split(string(in), "/")
+	compressionTyp = Type(parts[0])
+	level = zlib.DefaultCompression
+	if len(parts) > 1 {
+		level, err = strconv.Atoi(parts[1])
+		if err != nil {
+			return fmt.Errorf("invalid compression level: %q", parts[1])
+		}
+		if compressionTyp == TypeSnappy ||
+			compressionTyp == typeNone ||
+			compressionTyp == typeEmpty {
+			return fmt.Errorf("compression level is not supported for %q", compressionTyp)
+		}
+	}
+	ct.Level = Level(level)
+	if (compressionTyp == TypeGzip && isValidLevel(level)) ||
+		(compressionTyp == TypeZlib && isValidLevel(level)) ||
+		(compressionTyp == TypeDeflate && isValidLevel(level)) ||
+		compressionTyp == TypeSnappy ||
+		compressionTyp == TypeZstd ||
+		compressionTyp == typeNone ||
+		compressionTyp == typeEmpty {
+		ct.Level = Level(level)
+		ct.Type = compressionTyp
+		return nil
 	}
 
-	return TypeWithLevel{Type: typ, Level: ct.Level}, fmt.Errorf("unsupported compression type/level %q/%q", typ, ct.Level)
+	return fmt.Errorf("unsupported compression type/level %s/%d", compressionTyp, ct.Level)
+
 }
 
 func isValidLevel(level int) bool {
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -68,7 +68,7 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 		return nil
 	}
 
-	return fmt.Errorf("unsupported compression type/level %s/%d", compressionTyp, ct.Level)
+	return fmt.Errorf("unsupported compression type and level(default if not specified) %s - %d", compressionTyp, ct.Level)
 
 }
 
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -44,7 +44,7 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 	parts := strings.Split(string(in), "/")
 	compressionTyp = Type(parts[0])
 	level = zlib.DefaultCompression
-	if len(parts) > 1 {
+	if len(parts) == 2 {
 		level, err = strconv.Atoi(parts[1])
 		if err != nil {
 			return fmt.Errorf("invalid compression level: %q", parts[1])
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -38,12 +38,10 @@ func (ct *Type) IsCompressed() bool {
 }
 
 func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
-	var compressionTyp Type
-	var level int
 	var err error
 	parts := strings.Split(string(in), "/")
-	compressionTyp = Type(parts[0])
-	level = zlib.DefaultCompression
+	compressionTyp := Type(parts[0])
+	level := zlib.DefaultCompression
 	if len(parts) == 2 {
 		level, err = strconv.Atoi(parts[1])
 		if err != nil {
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -16,8 +16,8 @@ type Type string
 type Level int
 
 type TypeWithLevel struct {
-	Type  Type
-	Level Level
+	Type  Type  `mapstructure:"type"`
+	Level Level `mapstructure:"level"`
 }
 
 const (
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -48,6 +48,7 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 			return fmt.Errorf("invalid compression level: %q", parts[1])
 		}
 		if compressionTyp == TypeSnappy ||
+			compressionTyp == TypeLz4 ||
 			compressionTyp == typeNone ||
 			compressionTyp == typeEmpty {
 			return fmt.Errorf("compression level is not supported for %q", compressionTyp)
@@ -58,6 +59,7 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 		(compressionTyp == TypeZlib && isValidLevel(level)) ||
 		(compressionTyp == TypeDeflate && isValidLevel(level)) ||
 		compressionTyp == TypeSnappy ||
+		compressionTyp == TypeLz4 ||
 		compressionTyp == TypeZstd ||
 		compressionTyp == typeNone ||
 		compressionTyp == typeEmpty {
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -3,7 +3,13 @@
 
 package configcompression // import "go.opentelemetry.io/collector/config/configcompression"
 
-import "fmt"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/klauspost/compress/zlib"
+)
 
 // Type represents a compression method
 type Type string
@@ -39,4 +45,54 @@ func (ct *Type) UnmarshalText(in []byte) error {
 		return nil
 	}
 	return fmt.Errorf("unsupported compression type %q", typ)
+}
+
+// IsZstd returns true if the compression type is zstd.
+func (ct *Type) IsZstd() bool {
+	parts := strings.Split(string(*ct), "/")
+	return parts[0] == string(TypeZstd)
+}
+
+// IsGzip returns true if the compression type is gzip and the specified compression level is valid.
+func (ct *Type) IsGzip() bool {
+	parts := strings.Split(string(*ct), "/")
+	if parts[0] == string(TypeGzip) {
+		if len(parts) > 1 {
+			levelStr, err := strconv.Atoi(parts[1])
+			if err != nil {
+				return false
+			}
+			if levelStr == zlib.BestSpeed ||
+				levelStr == zlib.BestCompression ||
+				levelStr == zlib.DefaultCompression ||
+				levelStr == zlib.HuffmanOnly ||
+				levelStr == zlib.NoCompression {
+				return true
+			}
+		}
+		return true
+	}
+	return false
+}
+
+// IsZlib returns true if the compression type is zlib and the specified compression level is valid.
+func (ct *Type) IsZlib() bool {
+	parts := strings.Split(string(*ct), "/")
+	if parts[0] == string(TypeZlib) || parts[0] == string(TypeDeflate) {
+		if len(parts) > 1 {
+			levelStr, err := strconv.Atoi(parts[1])
+			if err != nil {
+				return false
+			}
+			if levelStr == zlib.BestSpeed ||
+				levelStr == zlib.BestCompression ||
+				levelStr == zlib.DefaultCompression ||
+				levelStr == zlib.HuffmanOnly ||
+				levelStr == zlib.NoCompression {
+				return true
+			}
+		}
+		return true
+	}
+	return false
 }
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -5,14 +5,18 @@ package configcompression // import "go.opentelemetry.io/collector/config/config
 
 import (
 	"fmt"
-	"strconv"
-	"strings"
 
 	"github.com/klauspost/compress/zlib"
 )
 
 // Type represents a compression method
 type Type string
+type Level int
+
+type TypeWithLevel struct {
+	Type  Type  `mapstructure:"type"`
+	Level Level `mapstructure:"level"`
+}
 
 const (
 	TypeGzip    Type = "gzip"
@@ -31,82 +35,26 @@ func (ct *Type) IsCompressed() bool {
 	return *ct != typeEmpty && *ct != typeNone
 }
 
-func (ct *Type) UnmarshalText(in []byte) error {
-	typ := Type(in)
-	if typ == TypeGzip ||
-		typ == TypeZlib ||
-		typ == TypeDeflate ||
+func (ct *TypeWithLevel) UnmarshalText() (TypeWithLevel, error) {
+	typ := ct.Type
+	if (typ == TypeGzip && isValidLevel(int(ct.Level))) ||
+		(typ == TypeZlib && isValidLevel(int(ct.Level))) ||
+		(typ == TypeDeflate && isValidLevel(int(ct.Level))) ||
 		typ == TypeSnappy ||
 		typ == TypeZstd ||
 		typ == TypeLz4 ||
 		typ == typeNone ||
 		typ == typeEmpty {
-		*ct = typ
-		return nil
+		return TypeWithLevel{Type: typ, Level: ct.Level}, nil
 	}
-	return fmt.Errorf("unsupported compression type %q", typ)
+
+	return TypeWithLevel{Type: typ, Level: ct.Level}, fmt.Errorf("unsupported compression type/level %q/%q", typ, ct.Level)
 }
 
-// IsZstd returns true if the compression type is zstd.
-// The specified compression level is not validated.
-// Because zstd supports returning an encoder level that closest matches the compression ratio of a specific zstd compression level.
-// Many input values will provide the same compression level.
-func (ct *Type) IsZstd() bool {
-	parts := strings.Split(string(*ct), "/")
-	return parts[0] == string(TypeZstd)
-}
-
-// Compression level isn't supported for snappy.
-func (ct *Type) IsSnappy() bool {
-	parts := strings.Split(string(*ct), "/")
-	if len(parts) > 1 {
-		return false
-	}
-	return parts[0] == string(TypeSnappy)
-}
-
-// IsGzip returns true if the compression type is gzip and the specified compression level is valid.
-func (ct *Type) IsGzip() bool {
-	parts := strings.Split(string(*ct), "/")
-	if parts[0] == string(TypeGzip) {
-		if len(parts) > 1 {
-			levelStr, err := strconv.Atoi(parts[1])
-			if err != nil {
-				return false
-			}
-			if levelStr == zlib.BestSpeed ||
-				levelStr == zlib.BestCompression ||
-				levelStr == zlib.DefaultCompression ||
-				levelStr == zlib.HuffmanOnly ||
-				levelStr == zlib.NoCompression {
-				return true
-			}
-			return false
-		}
-		return true
-	}
-	return false
-}
-
-// IsZlib returns true if the compression type is zlib and the specified compression level is valid.
-func (ct *Type) IsZlib() bool {
-	parts := strings.Split(string(*ct), "/")
-	if parts[0] == string(TypeZlib) || parts[0] == string(TypeDeflate) {
-		if len(parts) > 1 {
-			levelStr, err := strconv.Atoi(parts[1])
-			if err != nil {
-				return false
-			}
-			if levelStr == zlib.BestSpeed ||
-				levelStr == zlib.BestCompression ||
-				levelStr == zlib.DefaultCompression ||
-				levelStr == zlib.HuffmanOnly ||
-				levelStr == zlib.NoCompression {
-				return true
-			}
-			return false
-		}
-		return true
-	}
-	return false
+func isValidLevel(level int) bool {
+	return level == zlib.BestSpeed ||
+		level == zlib.BestCompression ||
+		level == zlib.DefaultCompression ||
+		level == zlib.HuffmanOnly ||
+		level == zlib.NoCompression
 }
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -74,9 +74,9 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 
 // Checks the validity of zlib/gzip/flate compression levels
 func isValidLevel(level int) bool {
-	return level == zlib.BestSpeed ||
-		level == zlib.BestCompression ||
-		level == zlib.DefaultCompression ||
+	return level == zlib.DefaultCompression ||
 		level == zlib.HuffmanOnly ||
-		level == zlib.NoCompression
+		level == zlib.NoCompression ||
+		level == zlib.BestSpeed ||
+		(level >= zlib.BestSpeed && level <= zlib.BestCompression)
 }
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -69,7 +69,6 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 	}
 
 	return fmt.Errorf("unsupported compression type and level(default if not specified) %s - %d", compressionTyp, ct.Level)
-
 }
 
 // Checks the validity of zlib/gzip/flate compression levels
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -48,9 +48,21 @@ func (ct *Type) UnmarshalText(in []byte) error {
 }
 
 // IsZstd returns true if the compression type is zstd.
+// The specified compression level is not validated.
+// Because zstd supports returning an encoder level that closest matches the compression ratio of a specific zstd compression level.
+// Many input values will provide the same compression level.
 func (ct *Type) IsZstd() bool {
 	parts := strings.Split(string(*ct), "/")
 	return parts[0] == string(TypeZstd)
+}
+
+// Compression level isn't supported for snappy.
+func (ct *Type) IsSnappy() bool {
+	parts := strings.Split(string(*ct), "/")
+	if len(parts) > 1 {
+		return false
+	}
+	return parts[0] == string(TypeSnappy)
 }
 
 // IsGzip returns true if the compression type is gzip and the specified compression level is valid.
@@ -69,6 +81,7 @@ func (ct *Type) IsGzip() bool {
 				levelStr == zlib.NoCompression {
 				return true
 			}
+			return false
 		}
 		return true
 	}
@@ -91,6 +104,7 @@ func (ct *Type) IsZlib() bool {
 				levelStr == zlib.NoCompression {
 				return true
 			}
+			return false
 		}
 		return true
 	}
--- a/config/configcompression/compressiontype.go
+++ b/config/configcompression/compressiontype.go
@@ -72,6 +72,7 @@ func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
 
 }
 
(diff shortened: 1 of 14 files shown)


rnishtala-sumo:
@Dmitrii Anoshin @Alex Boten @mx-psi would appreciate some attention to this PR. It doesn't have any breaking changes.

config/confighttp/compressor.go
```diff
@@ -22,36 +24,47 @@ type writeCloserReset interface {
...
 
+// Gets the compression type and level from the configuration.
+func getCompression(compressionField configcompression.Type) (compressionType configcompression.Type, compressionLevel int) {
+	parts := strings.Split(string(compressionField), "/")
+
+	compressionLevel = zlib.DefaultCompression
+	compressionType = configcompression.Type(parts[0])
+	if len(parts) > 1 {
+		levelStr := parts[1]
+		if level, err := strconv.Atoi(levelStr); err == nil {
```
dmitryax, line 39: We should have return the error here instead of falling back to default. This should never happen though if we apply validation initially

config/confighttp/README.md
```diff
@@ -26,6 +26,27 @@ README](../configtls/README.md).
 - `compression`: Compression type to use among `gzip`, `zstd`, `snappy`, `zlib`, and `deflate`.
   - look at the documentation for the server-side of the communication.
   - `none` will be treated as uncompressed, and any other inputs will cause an error.
+  - Compression levels can now be configured as part of the compression type like below. Not specifying any compression level will result in the default.
```
dmitryax, line 29: `snappy` should be mentioned as well, without levels (we should fail for snappy/3)

config/configcompression/compressiontype.go
```diff
@@ -3,7 +3,13 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/klauspost/compress/zlib"
+)
 
 // Type represents a compression method
 type Type string
```
dmitryax, line 15: I don't think we can reuse this type and keep the level encoded in the string. API users may rely on `configcompression.Type == configcompression.TypeGzip` which won't work anymore...

I think we need new additional types like these

```
type Config struct {
	Type Type
	Level Level
}

type Level int
```

`Config` may be also called `TypeWithLevel`.

Given that this module is 1.x, it should be okay to add. However, we would need to make a breaking change to the Co API (not an end-user breaking change) in config HTTP by changing the `confighttp.ClientConfig.Compression` field from `compressionconfig.Type` to `compressionconfig.Config`, which should be okay given that the confighttp module is still 0.x.

Then we can have all validation in `Config.UnmarshalText` and assume that Config is always valid

rnishtala-sumo: ok this is kind of what I had before in an earlier commit - https://github.com/open-telemetry/opentelemetry-collector/pull/11084/commits/74b1d10754ea0d5f128241058b847c1842ae9ab3#diff-a559328153338200c250ddd678e69c4d0576cad113fa1feaed06b65548e33752R28

It sounds like I should go back to that approach?

config/confighttp/compressor.go
rnishtala-sumo: Falling back to the default compression level here, since that's whats used when there isn't any level specified.

rnishtala-sumo: So in this case we'd be treating an invalid level as no level specified. Could this work as an alternative to rejecting compression config entirely? I chose this approach because level is intended to be optional.

config/configcompression/compressiontype.go
rnishtala-sumo: > confighttp.ClientConfig.Compression field from compressionconfig.Type to compressionconfig.Config
won't this be a breaking change for the user as well?, i.e with the above suggested change we would need two fields to specify compression like below

```
compression:
  type: "gzip"
  level: 1
```

rnishtala-sumo: @dmitryax I've made changes based on the above comment. Please let me know if this is what you had in mind.

rnishtala-sumo: looks like the tests are failing because of errors like below, since this is now a breaking change

`Error: config/config.go:147:24: invalid operation: cfg.Compression != "" (mismatched types configcompression.TypeWithLevel and untyped string)`

config/confighttp/confighttp.go
```diff
@@ -67,7 +70,7 @@ type ClientConfig struct {
 	Auth *configauth.Authentication `mapstructure:"auth"`
 
 	// The compression key for supported compression types within collector.
-	Compression configcompression.Type `mapstructure:"compression"`
+	Compression configcompression.TypeWithLevel `mapstructure:"compression"`
```
rnishtala-sumo, line 73: This breaking change is causing multiple tests to fail

config/configcompression/compressiontype.go
dmitryax: > won't this be a breaking change for the user as well?, 

No, we should keep the string encoding `<type>/<level>` where the `/<level>` part is optional

config/configcompression/compressiontype.go
```diff
@@ -24,18 +34,25 @@ func (ct *Type) IsCompressed() bool {
 	return *ct != typeEmpty && *ct != typeNone
 }
 
-func (ct *Type) UnmarshalText(in []byte) error {
-	typ := Type(in)
-	if typ == TypeGzip ||
-		typ == TypeZlib ||
-		typ == TypeDeflate ||
+func (ct *TypeWithLevel) UnmarshalText() (TypeWithLevel, error) {
```
dmitryax, line 37: The function signature should not be changed. We need to keep `encoding.TextUnmarshaler` interface implementation

config/configcompression/compressiontype.go
rnishtala-sumo: ok, I guess what I don't understand is how this is possible if we change the type from `compressionconfig.Type` to `compressionconfig.Config`, the latter being a struct. Doesn't this imply that we're moving away from `<type>/<level>`?

config/configcompression/compressiontype.go
rnishtala-sumo: Yes made the change.

config/configcompression/compressiontype.go
rnishtala-sumo: I may have misunderstood the ask here. The following has been implemented

- Keep the compressionType
- Define new TypeWithLevel
- Unmarshal the `<type>/<level>` string encoding to set type and level in `TypeWithLevel`. Also add validation logic in here

rnishtala-sumo: > No, we should keep the string encoding <type>/<level> where the /<level> part is optional

The latest commit ensures that the `/level` part is optional

config/configcompression/compressiontype.go
```diff
@@ -24,18 +36,45 @@ func (ct *Type) IsCompressed() bool {
...
-		typ == typeEmpty {
-		*ct = typ
+func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
+	var compressionTyp Type
+	var level int
+	var err error
+	parts := strings.Split(string(in), "/")
+	compressionTyp = Type(parts[0])
+	level = zlib.DefaultCompression
+	if len(parts) > 1 {
```
dmitryax, line 46: Let's fail if `len(parts) > 2`
```suggestion
	if len(parts) == 1 {
```

config/configcompression/compressiontype.go
```diff
@@ -24,18 +36,45 @@ func (ct *Type) IsCompressed() bool {
...
+		}
+	}
+	ct.Level = Level(level)
+	if (compressionTyp == TypeGzip && isValidLevel(level)) ||
+		(compressionTyp == TypeZlib && isValidLevel(level)) ||
+		(compressionTyp == TypeDeflate && isValidLevel(level)) ||
+		compressionTyp == TypeSnappy ||
+		compressionTyp == TypeZstd ||
+		compressionTyp == typeNone ||
+		compressionTyp == typeEmpty {
```
dmitryax, line 64: Let's fail if the level is provided for types that don't support it

config/confighttp/compressor.go
```diff
@@ -22,36 +22,32 @@ type writeCloserReset interface {
...
-func newCompressor(compressionType configcompression.Type) (*compressor, error) {
-	switch compressionType {
+func newCompressor(compressionType configcompression.TypeWithLevel) (*compressor, error) {
+	// compressionType, compressionLevel := getCompression(compressionType)
+
+	switch compressionType.Type {
 	case configcompression.TypeGzip:
-		return gZipPool, nil
+		var _ writeCloserReset = (*gzip.Writer)(nil)
+		return &compressor{pool: sync.Pool{New: func() any { w, _ := gzip.NewWriterLevel(nil, int(compressionType.Level)); return w }}}, nil
```
dmitryax, line 37: The pools have to be kept reusable. We should not create them on every request

config/configcompression/compressiontype.go
rnishtala-sumo: I think you mean `len(parts) == 2`? because this condition is for when `/level` is specified.

config/confighttp/confighttp.go
```diff
@@ -216,8 +216,13 @@ func (hcs *ClientConfig) ToClient(ctx context.Context, host component.Host, sett
 
 	// Compress the body using specified compression methods if non-empty string is provided.
 	// Supporting gzip, zlib, deflate, snappy, and zstd; none is treated as uncompressed.
+	var compressionTypeWithLevel configcompression.TypeWithLevel
+	err = compressionTypeWithLevel.UnmarshalText([]byte(hcs.Compression))
```
dmitryax, line 220: We don't need to call UnmarshalText explicitly. Changing `ClientConfig.Compression` field from `configcompression.Type` to `configcompression.TypeWithLevel` should be enough

config/configcompression/compressiontype.go
rnishtala-sumo: yes this is handled in the prior condition - https://github.com/open-telemetry/opentelemetry-collector/pull/11084/files#diff-ae510a297dd561956c65c1e375f12870c19e281c486b51017e3280e0d3de08f2R51

rnishtala-sumo: Here's the test for it - https://github.com/open-telemetry/opentelemetry-collector/pull/11084/files#diff-fc75b0e352923bac06c0d1b09268c82efe3ad8523c0107926c5e919d4ab9dceeR78

config/confighttp/confighttp.go
rnishtala-sumo: Doing that seemed to be a breaking change. There are errors like this seen in multiple contrib components

```Error: config/config.go:147:24: invalid operation: cfg.Compression != "" (mismatched types configcompression.TypeWithLevel and untyped string)```

Comment (dmitryax) [reply config/confighttp/confighttp.go:220]: Yes, it is a Go API breaking change, but not an end-user breaking change. I think this is the cleanest solution. @open-telemetry/collector-approvers let me know if you have other ideas

File: config/configcompression/compressiontype.go (hunk 6 above)
Comment (dmitryax) [reply config/configcompression/compressiontype.go:46]: Yes, I meant `len(parts) == 2` for this condition, sorry

File: config/confighttp/confighttp.go (hunk 9 above)
Comment (rnishtala-sumo) [reply config/confighttp/confighttp.go:220]: if we make the API breaking change, are going to mute all the failing tests (in this repo) and have the contrib component owners fix that later. Sorry, I'm not clear on this process.

File: config/confighttp/compressor.go (hunk 8 above)
Comment (rnishtala-sumo) [reply config/confighttp/compressor.go:37]: yeah this one seems tricky, I moved the pools inside the function, because we now need to pass in the level to the writer. Going back this [comment](https://github.com/open-telemetry/opentelemetry-collector/issues/8216#issuecomment-1735802401). Do you think we need to pool the encoders in `confighttp`?

Comment (rnishtala-sumo) [reply config/confighttp/compressor.go]: @dmitryax I have made changes to keep the pools reusable as suggested. Please let me know if you see issues with it.

File: config/confighttp/confighttp.go (hunk 9 above)
Comment (dmitryax) [reply config/confighttp/confighttp.go:220]: Right. If we go this route, we would need to quickly submit a follow-up fix in the contrib repo

File: config/confighttp/README.md [hunk 10]
```diff
@@ -26,6 +26,29 @@ README](../configtls/README.md).
...
+    - `gzip`
+      - NoCompression: `gzip/0`
+      - BestSpeed: `gzip/1`
+      - BestCompression: `gzip/9`
+      - DefaultCompression: `gzip/-1`
```
dmitryax, line 34: Let's keep `-1` for internal representation
```suggestion
      - DefaultCompression: `gzip`
```

config/confighttp/README.md
```diff
@@ -26,6 +26,29 @@ README](../configtls/README.md).
...
+    - `zlib`
+      - NoCompression: `zlib/0`
+      - BestSpeed: `zlib/1`
+      - BestCompression: `zlib/9`
+      - DefaultCompression: `zlib/-1`
```
dmitryax, line 39: Same for others
```suggestion
      - DefaultCompression: `zlib`
```

config/configcompression/compressiontype.go
```diff
@@ -24,18 +36,45 @@ func (ct *Type) IsCompressed() bool {
...
-		typ == TypeDeflate ||
-		typ == TypeSnappy ||
-		typ == TypeZstd ||
-		typ == typeNone ||
-		typ == typeEmpty {
-		*ct = typ
+func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
+	var compressionTyp Type
+	var level int
+	var err error
```
dmitryax, line 42: I don't think we need all these declarations separately.

config/confighttp/confighttp.go
rnishtala-sumo: How about keeping this as is (non-breaking) for this PR? and handle all breaking changes (core and contrib) in  2 other PRs. The only reason is, they seem to affect multiple contrib components (almost all)

rnishtala-sumo: @dmitryax I have pushed the API breaking changes and the contrib tests(`contrib-tests / contrib-tests-matrix`) seem to be failing.

dmitryax: If you have an idea how to make it in several steps without breaking contrib and get this result, that would be great. I'm not aware of any

config/configcompression/compressiontype.go
```diff
@@ -25,19 +37,45 @@ func (ct *Type) IsCompressed() bool {
...
+		compressionTyp == TypeZstd ||
+		compressionTyp == typeNone ||
+		compressionTyp == typeEmpty {
+		ct.Level = Level(level)
+		ct.Type = compressionTyp
 		return nil
 	}
-	return fmt.Errorf("unsupported compression type %q", typ)
 
+	return fmt.Errorf("unsupported compression type/level %s/%d", compressionTyp, ct.Level)
```
dmitryax, line 71: This would be confusing output if user didn't provide level

config/confighttp/compressor.go
```diff
@@ -23,45 +23,61 @@ type writeCloserReset interface {
...
 // writerFactory defines writer field in CompressRoundTripper.
 // The validity of input is already checked when NewCompressRoundTripper was called in confighttp,
-func newCompressor(compressionType configcompression.Type) (*compressor, error) {
-	switch compressionType {
+func newCompressor(compressionType configcompression.TypeWithLevel) (*compressor, error) {
+	switch compressionType.Type {
 	case configcompression.TypeGzip:
-		return gZipPool, nil
+		if gZipCompressor.pool.Get() == nil {
+			gZipCompressor.pool = sync.Pool{New: func() any { w, _ := gzip.NewWriterLevel(nil, int(compressionType.Level)); return w }}
```
dmitryax, line 49: This seems incorrect. We are taking compressor with _any_ level from the pool. If user picks `zlib/1` first and it goes back to the pool, the second call with any other level like `zlib/2` would still give the same `zlib/1` instance from the pool. We cannot reuse them compressors with different levels.

We might need a map of pools built dynamically here

config/configcompression/compressiontype.go
rnishtala-sumo: this should print the default level, however I can remove it since this line is only reached when the compression level isn't specified.

config/confighttp/compressor.go
rnishtala-sumo: I created a map of compressor pools for compressors that support levels, all the tests pass with this change
https://github.com/open-telemetry/opentelemetry-collector/pull/11084/commits/c7f36ad44861a58491a09f537207a20decb3c186

config/configcompression/compressiontype.go
rnishtala-sumo: I made a change to the error message, please let me know if you think this is clearer

https://github.com/open-telemetry/opentelemetry-collector/pull/11084/commits/c7f36ad44861a58491a09f537207a20decb3c186

config/confighttp/confighttp.go
rnishtala-sumo: I think we can stick to the current plan. These are the only changes that need to go into the contrib repository

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35843/files

config/configcompression/compressiontype.go
```diff
@@ -25,19 +37,45 @@ func (ct *Type) IsCompressed() bool {
...
+	return fmt.Errorf("unsupported compression type and level(default if not specified) %s - %d", compressionTyp, ct.Level)
+
+}
+
+func isValidLevel(level int) bool {
+	return level == zlib.BestSpeed ||
+		level == zlib.BestCompression ||
+		level == zlib.DefaultCompression ||
+		level == zlib.HuffmanOnly ||
+		level == zlib.NoCompression
```
axw, line 80:
```suggestion
	// https://pkg.go.dev/compress/gzip#NewWriterLevel says:
	//     "The compression level can be DefaultCompression, NoCompression, HuffmanOnly
	//     or any integer value between BestSpeed and BestCompression inclusive."
	return level == zlib.DefaultCompression ||
			level == zlib.NoCompression ||
			level == zlib.HuffmanOnly ||
			(level >= zlib.BestSpeed && level <= zlib.BestCompression)
```

config/configcompression/compressiontype.go
```diff
@@ -25,19 +37,45 @@ func (ct *Type) IsCompressed() bool {
...
+		ct.Type = compressionTyp
 		return nil
 	}
-	return fmt.Errorf("unsupported compression type %q", typ)
 
+	return fmt.Errorf("unsupported compression type and level(default if not specified) %s - %d", compressionTyp, ct.Level)
+
+}
+
+func isValidLevel(level int) bool {
```
axw, line 75: Should we have zlib/gzip/flate somewhere in the function name?

axw (commented):
I wonder if the string format is a bit too limiting, and if we would be better off with supporting either a string (compression algorithm name only) or a struct (algorithm name + algorithm-specific options, such as level)? That would for example allow one to configure the window size for zstd, or a dictionary (I'm not sure that would be a good idea though tbh.)

That could still be done later though.

config/configcompression/compressiontype.go
rnishtala-sumo: The compression levels seem to be discrete values, which is why I didn't use a range.

config/configcompression/compressiontype.go
rnishtala-sumo: I could add a comment for this function, instead of changing the name.

rnishtala-sumo:
> I wonder if the string format is a bit too limiting

@axw we chose to stick with a string format because we didn't want this PR to be a breaking change for the user.

config/configcompression/compressiontype.go
axw: @rnishtala-sumo please see the comment I included above, which is taken directly from the compress/gzip package docs. `BestSpeed-BestCompression (1-9)` is a range, while `DefaultCompression` and `NoCompression` have special meaning.

rnishtala-sumo: the collector uses this library where they are discrete - https://pkg.go.dev/github.com/klauspost/compress/gzip#pkg-constants

axw:
>> I wonder if the string format is a bit too limiting
>
> @axw we chose to stick with a string format because we didn't want this PR to be a breaking change for the user.

What I meant was, could we support both, so it's not breaking. Actually it turns out this works with your changes, but perhaps by accident. e.g. I just tested your branch locally and you can configure like this:

```yaml
exporters:
  otlphttp:
    endpoint: ...
    compression:
      type: gzip
      level: 9
```

However, the type & level are not validated in this case, as the validation is currently being done in the `UnmarshalText` method which does not get invoked in this case. I would suggest introducing a `Validate` method on `TypeWithLevel`, and doing the type/level validation there. This method will be automatically invoked after config is parsed, by config structs implementing [component.ConfigValidator](https://github.com/open-telemetry/opentelemetry-collector/blob/403c782d50c630ad945feebe8b26e1f656ccd3c1/component/config.go#L27-L31).

config/configcompression/compressiontype.go
axw: The same comment exists on https://pkg.go.dev/github.com/klauspost/compress/gzip#NewWriterLevel. It's a standard part of zlib.

config/configcompression/compressiontype.go
```diff
@@ -25,19 +37,46 @@ func (ct *Type) IsCompressed() bool {
...
-	if typ == TypeGzip ||
-		typ == TypeZlib ||
-		typ == TypeDeflate ||
-		typ == TypeSnappy ||
-		typ == TypeZstd ||
-		typ == TypeLz4 ||
-		typ == typeNone ||
-		typ == typeEmpty {
-		*ct = typ
+func (ct *TypeWithLevel) UnmarshalText(in []byte) error {
```
axw, line 40: As mentioned in https://github.com/open-telemetry/opentelemetry-collector/pull/11084#issuecomment-2428122545, I would recommend splitting all validation out into a separate `func (*TypeWithLevel) Validate() error` method.

config/configcompression/compressiontype.go
```diff
@@ -25,19 +37,46 @@ func (ct *Type) IsCompressed() bool {
...
+			return fmt.Errorf("compression level is not supported for %q", compressionTyp)
+		}
+	}
+	ct.Level = Level(level)
+	if (compressionTyp == TypeGzip && isValidLevel(level)) ||
+		(compressionTyp == TypeZlib && isValidLevel(level)) ||
+		(compressionTyp == TypeDeflate && isValidLevel(level)) ||
+		compressionTyp == TypeSnappy ||
+		compressionTyp == TypeLz4 ||
+		compressionTyp == TypeZstd ||
```
axw, line 63: Seems that we're supporting specifying a level for zstd -- should there be validation for the level here?

config/configcompression/compressiontype.go
axw: Could do. IMO it would be clearer to include it in the name, but that's just my opinion.

Having said that, I think the function could be inlined with a bit of restructuring in the caller. Something like this:

```go
type TypeWithLevel struct {
        Type  Type   `mapstructure:"type"`

        // Level holds the compression level for algorithms that support one.
        //
        // If Level is nil, then the algorithm's default compression level will be used.
        Level *int `mapstructure:"level"`
}

func (ct *TypeWithLevel) Validate() error {
        switch ct.Type {
        case TypeGzip, TypeZlib, TypeDeflate, TypeSnappy, TypeZstd, TypeLz4, typeNone, typeEmpty:
        default:
                return fmt.Errorf("invalid compression type %q", ct.Type)
        }
        if ct.Level != nil {
                switch ct.Type {
                case TypeGzip, TypeZlib, TypeDeflate:
                        if level < -1 || level > 9 {
                                return fmt.Errorf("invalid compression level %d for %q", *ct.Level, ct.Type)
                        }
                case TypeZstd:
                        // TODO check zstd level
                default:
                        return fmt.Errorf("compression level is not supported for %q", ct.Type)
                }
        }
        return nil
}
```

config/configcompression/compressiontype.go
rnishtala-sumo: I made the change.

config/configcompression/compressiontype.go
rnishtala-sumo: I took the approach of having all the validation in `UnmarshalText` based on this previous review comment by @dmitryax  - https://github.com/open-telemetry/opentelemetry-collector/pull/11084#discussion_r1776033836

config/configcompression/compressiontype.go
rnishtala-sumo: zstd has this function which returns an encoder level that closest matches the compression ratio of a specific zstd compression level. So I didn't see a need to specify a range of valid levels for it.

This function (its used here) - https://pkg.go.dev/github.com/klauspost/compress/zstd#EncoderLevelFromZstd

axw: OK. It's somewhat tying to the implementation details of klauspost/compress/zstd, but I suppose it's fine.

config/configcompression/compressiontype.go
axw: I have an alternative proposal - what do you think of the below @dmitryax?

---

Users who want to specify a level should use structured config like:

```yaml
compression:
  type: gzip
  level: 9
```

For backwards compatibility support parsing the algorithm name ("type"), but don't invent a new format for including the level in there. i.e. you could specify `compression: gzip` (implies using default compression level), but not `compression: gzip/9` (if you need that, use the structured approach).

To support the above, move validation to its own Validate method, so it applies to both the backwards compatible string parsing case & structured config case.

config/configcompression/compressiontype.go
axw: Continuing discussion on that thread.

andrzej-stencel (approved):
I like this approach. It's additive and doesn't break stable components.

config/confighttp/compression_test.go
```diff
@@ -35,6 +36,14 @@ func TestHTTPClientCompression(t *testing.T) {
 	compressedZstdBody := compressZstd(t, testBody)
 	compressedLz4Body := compressLz4(t, testBody)
 
+	const (
+		GzipLevel    configcompression.Type = "gzip/1"
+		ZlibLevel    configcompression.Type = "zlib/1"
+		DeflateLevel configcompression.Type = "deflate/1"
+		ZstdLevel    configcompression.Type = "zstd/11"
```
dmitryax, line 43: Please make them unexported even if it's test file

config/confighttp/compression_test.go
```diff
@@ -104,19 +113,27 @@ func TestHTTPClientCompression(t *testing.T) {
 
 			req, err := http.NewRequest(http.MethodGet, srv.URL, reqBody)
 			require.NoError(t, err, "failed to create request to test handler")
-
+			parts := strings.Split(string(tt.encoding), "/")
+			compressionLevel := configcompression.Level(gzip.BestSpeed)
+			compressionType := configcompression.Type(parts[0])
+			if len(parts) == 2 {
```
dmitryax, line 119: You can use unmarshalText instead of replicating the code in the test

dmitryax (commented):
Nits

config/confighttp/compressor.go
```diff
@@ -23,45 +24,76 @@ type writeCloserReset interface {
...
+	pools map[string]*compressor
+}
 
 type compressor struct {
 	pool sync.Pool
 }
 
+var (
+	compressorPools                   = &compressorMap{pools: make(map[string]*compressor)}
+	gZipCompressor                    = &compressor{}
```
dmitryax, line 37: Don't need to pre define them anymore

rnishtala-sumo: removed the declarations for gzip, zlib and zstd. The other compression types need a declaration.

config/confighttp/compression_test.go
rnishtala-sumo: Made this change.

config/configcompression/compressiontype.go
dmitryax: @axw, are you suggesting keeping the mixed type value: string for plain type and struct for type+level? I'm not sure I like this approach. We don't have anything like this in the collector config interface. I think we should keep it strictly typed. I'm actually surprised that it currently accepts both types, that was not desirable.

What if we just add another extra optional field called `compression_params` or `compression_config`  _in addition_ to `compression`? It'll be a fully backward-compatible additive change. It'll also allow us to add any other compression related configs in the future. To define compression level the users would need to provide:
```
compression: gzip
compression_params:
  level: 9
```

@rnishtala-sumo @axw @andrzej-stencel @codeboten WDYT?

axw: > @axw, are you suggesting keeping the mixed type value: string for plain type and struct for type+level?

Yes, but only for backwards compatibility. I would suggest phasing out the plain string format, leaving only the struct format.

> What if we just add another extra optional field called compression_params or compression_config in addition to compression? It'll be a fully backward-compatible additive change. It'll also allow us to add any other compression related configs in the future. To define compression level the users would need to provide:

I prefer this over adding the level to the string format, but it means the configuration is no longer a single object/type. So this may lead to more cumbersome code.

dmitryax: This is an interesting approach to handle backward compatibility for such change. I'm open to this. Interested to hear other opinions from other @open-telemetry/collector-approvers

rnishtala-sumo: Just as a reminder, we started of including the level as part of the string, to keep this a non-breaking change for the user. I'd personally like to keep it this way, since changing this from a `string` to a `struct` is a major breaking change to a commonly used feature.

Maybe we keep `compression` a string (with the optional level) and introduce a new field in future (`compression_config`) that is a struct from the beginning which could allow for more compression config params. 

Example:

```
compression_config:
  type: gzip
  level: 1
```

This approach allows us to phase out (deprecate and remove) `compression` after introducing `compression_config`. I think this is less confusing for a user, since they'd be clear about the type of the `compression` field.

dmitryax: @rnishtala-sumo, @axw suggests an approach on how to convert the field from string to struct in a backward compatible way.

rnishtala-sumo: > To support the above, move validation to its own Validate method, so it applies to both the backwards compatible string parsing case & structured config case

right, I got that. Just thought using the same field as a string and as a struct might be confusing to the user, until we phase out the string type. Did we follow this pattern elsewhere?

dmitryax: > Did we follow this pattern elsewhere?

This is a new pattern. We haven't considered it yet. But moving from string to struct is something that we potentially can run into with other config fields. So if there is a solution for that, it would be interesting to try.

rnishtala-sumo: ok, I think all that's needed to do this is to keep unmarshal as is and move the validation to a new method.

andrzej-stencel: Configuring compression type is probably a (reasonably) widely used feature. Configuring compression details on the other hand will probably only be used by a small number of users. I'm not sure if we want to introduce a breaking change (even if with a long deprecation time) for a big number of users to satisfy needs of a small number of users?

On the other hand, after giving it a bit of thought, the "stringly typed" approach is also not ideal. I would lean towards the approach with leaving `compression` a string as it is now and adding `compression_config` or similar field for additional settings. Unless we feel it's too awkward and the breaking change is justified.

rnishtala-sumo: I have pushed the changes suggested by @axw in this [commit](https://github.com/open-telemetry/opentelemetry-collector/pull/11084/commits/976f44c646fe92c5909f9780df3a43a3565aefd2). This does the following:
- Moves the level validation outside Unmarshal
- Allows the compression to be specified as a string/struct

rnishtala-sumo: JFYI, I tested that with the above change both of the following configuration are valid

compression struct
```
exporters:
  otlphttp:
    endpoint: otelcol:4317
    compression:
      type: gzip
      level: 1
```

compression string
```
exporters:
  otlphttp:
    endpoint: otelcol:4317
    compression: gzip
```

axw: Thanks @rnishtala-sumo. To be explicit, I (obviously) think this approach looks good :)

I suppose we'll need to wait for a couple more opinions from approvers to make a final decision

rnishtala-sumo: @dmitryax @andrzej-stencel do you think we could proceed with @axw's approach?

dmitryax: Update: We discussed this during the Collector SIG call on Dec 4. Most of the Collector leads present there (@mx-psi and @jpkrohling) voted to introduce another field (`compression_params`) to avoid breaking the existing user interface even in a backward compatible way. @rnishtala-sumo is going to submit another PR to implement that approach.

rnishtala-sumo: Here's the new PR with option2 - https://github.com/open-telemetry/opentelemetry-collector/pull/11805

Issue #10919, opened 2024-08-20
[telemetry] Simplify disabling of prometheus endpoint for internal telemetry metrics


pirgeo:
As it is now possible to export internal telemetry metrics via OTLP, scraping the Prometheus metrics from the collector is no longer the only way to get that data. Since users will likely export via both export paths, they might want to disable the Prometheus endpoint.

Have a look at the [instantiation code today](https://github.com/open-telemetry/opentelemetry-collector/blob/c1c8fcd6ccde9e469b5e2e0569dd897c7e6b32a9/service/telemetry.go#L44-L48). You have two ways of disabling the Prometheus endpoint: 
1. Set `service: telemetry: metrics: level: none`. However, this will immediately return a `noop` MeterProvider and thus disable all exporting (including OTLP-based exporting).
2. Set `service: telemetry: metrics: address: ""` (to an empty string). The code checks whether the string in the address is longer than 0 characters. However, having to know that only the empty string disables the additional Prometheus exporter feels unintuitive to me and makes the collector config confusing to read if you don't know exactly why you need to specify an empty string. 

**Describe the solution you'd like**
If another exporter is explicitly defined, the Prometheus exporter is automatically disabled. However, it can be explicitly added alongside the OTLP exporter if required.

**Describe alternatives you've considered**
I am opening this issue to discuss possible solutions. Here are a few I came up with

- Setting the address to `null` explicitly should disable the Prom exporter
- Similarly, setting the address to `"none"` or `"disabled"` should turn off the exporter. I don't like this option, though, as it uses magic strings and is not intuitive.
- Turning the Prometheus exporter off by default if another exporter is explicitly configured
- Deprecating the `service: telemetry: metrics: address: ` configuration option and requiring Prometheus exporters to be explicitly set up so they follow the same structure as the OTLP exporters. This could break backward compatibility, though, so we might need to introduce a feature gate.

pirgeo:
Probably similar to the discussion in https://github.com/open-telemetry/opentelemetry-collector/issues/10890, might also help with disabling metrics, for example here https://github.com/open-telemetry/opentelemetry-collector/issues/10780

pirgeo:
The plot thickens. It seems `service: telemetry: metrics: address` is already deprecated: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/service.go#L175-L177. However, I don't see that warning message in my collector startup logs. 

[...]

PR #11093, opened 2024-09-09
[service] fix proctelemetry registration bug

codeboten:
This bug caused proctelemetry metrics to not be registered if a user configured the Collector's internal telemetry via `readers` only and disabled `address`. The check in the `if` statement is no longer needed since a no-op meter provider will be configured unless the telemetry level is set.

Mentioned in #10919


dmitryax (approved):

internal/testutil/testutil.go
```diff
@@ -86,6 +86,22 @@ func GetAvailableLocalIPv6Address(t testing.TB) string {
...
+	}
+	portInt, err := strconv.Atoi(port)
+	if err != nil {
+		return nil
+	}
+	return &config.Prometheus{
+		Host: &host,
+		Port: &portInt,
+	}
+}
```
bogdandrutu, line 103: Not sure this needs to be here, can be a private func in service_test

bogdandrutu (approved):

internal/testutil/testutil.go
codeboten: sure I can move it

codeboten: done. moved `GetAvailableLocalAddressPrometheus` at the same time as its only used in `service` and it's an "internal" package

TylerHelmuth (approved):

PR #11114, opened 2024-09-10
Revert "[processorhelper] deprecating unused methods"

djaglowski:
Reverts open-telemetry/opentelemetry-collector#11083

I'd like to propose that these metrics not be deprecated unless `Accepted`, `Refused`, and `Dropped` are also deprecated. These four metrics describe a complete picture of telemetry flowing through processors.

Personally I'd like to deprecate all four in favor of https://github.com/open-telemetry/opentelemetry-collector/issues/10708 because the latter can be captured automatically.


djaglowski:
After discussing with @codeboten and @mx-psi, we agreed that the associated "accepted", "refused", and "dropped" metrics will also be deprecated.

PR #11123, opened 2024-09-10
[builder]: Fix version for providers in the default config

bogdandrutu:


TylerHelmuth (approved):

cwegener:
I assume there will be a 0.109.1 for this fix shortly? Or is there a workaround for this issue?

cwegener:
> ~I assume there will be a 0.109.1 for this fix shortly? Or is there a workaround for this issue?~
> 
> Nevermind. I think I misread this change.

Ah. This change actually does fix the builder errors. I was running into caching problems during my testing. Applying this change explicitly to my builder config file does indeed fix the errors.

PR #11126, opened 2024-09-10
[builder] Remove builder default confmap providers

bogdandrutu:
#### Description

The main problem with this, is that the providers will start to have a different version than the builder and they will be independently mark as stable which will make it very hard to track all combination of builder versions with the "associated default" providers for that version.

This also fixes an issue, that for builder 0.109.0 the default providers will not work since the env and file are no longer using 0.109.0 version.

### DIFF
--- a/.chloggen/fixbuilderproviders.yaml
+++ b/.chloggen/fixbuilderproviders.yaml
@@ -1,0 +1,20 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'breaking'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: 'builder'
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove default config providers. This will require users to always specify the providers.
+
+# One or more tracking issues or pull requests related to the change
+issues: [11126]
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -39,7 +39,7 @@ type Config struct {
 	Receivers    []Module     `mapstructure:"receivers"`
 	Processors   []Module     `mapstructure:"processors"`
 	Connectors   []Module     `mapstructure:"connectors"`
-	Providers    *[]Module    `mapstructure:"providers"`
+	Providers    []Module     `mapstructure:"providers"`
 	Replaces     []string     `mapstructure:"replaces"`
 	Excludes     []string     `mapstructure:"excludes"`
 
@@ -79,6 +79,13 @@ type Module struct {
 	Path   string `mapstructure:"path"`   // an optional path to the local version of this module
 }
 
+func (mod *Module) Validate() error {
+	if mod.GoMod == "" {
+		return ErrMissingGoMod
+	}
+	return nil
+}
+
 type retry struct {
 	numRetries int
 	wait       time.Duration
@@ -114,17 +121,13 @@ func NewDefaultConfig() Config {
 
 // Validate checks whether the current configuration is valid
 func (c *Config) Validate() error {
-	var providersError error
-	if c.Providers != nil {
-		providersError = validateModules("provider", *c.Providers)
-	}
 	return multierr.Combine(
 		validateModules("extension", c.Extensions),
 		validateModules("receiver", c.Receivers),
 		validateModules("exporter", c.Exporters),
 		validateModules("processor", c.Processors),
 		validateModules("connector", c.Connectors),
-		providersError,
+		validateProviders(c.Providers),
 	)
 }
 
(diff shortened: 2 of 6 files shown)


cmd/builder/internal/builder/config.go
```diff
@@ -207,34 +203,9 @@ func (c *Config) ParseModules() error {
 		return err
 	}
 
-	if c.Providers != nil {
-		providers, err := parseModules(*c.Providers)
-		if err != nil {
-			return err
-		}
-		c.Providers = &providers
-	} else {
```
atoulme, line 216: is it a problem if no providers are defined, should the builder error out in that case?

cmd/builder/internal/builder/config.go
```diff
@@ -207,34 +203,9 @@ func (c *Config) ParseModules() error {
...
-			},
-			{
-				GoMod: "go.opentelemetry.io/collector/confmap/provider/yamlprovider v" + c.Distribution.OtelColVersion,
-			},
-		})
-		if err != nil {
-			return err
-		}
-		c.Providers = &providers
+	c.Providers, err = parseModules(c.Providers)
```
TylerHelmuth, line 206: I fear this will break most ocb users. If we go this route, is there a gentler approach we could take with feature gates or timelines?

bogdandrutu: >The main problem with this, is that the providers will start to have a different version than the builder and they will be independently mark as stable which will make it very hard to track all combination of builder versions with the "associated default" providers for that version.

> This also fixes an issue, that for builder 0.109.0 the default providers will not work since the env and file are no longer using 0.109.0 version.

bogdandrutu: I would have suggested to actually wait for all providers to be marked stabled in one version, that would have simplified a lot. Otherwise, I am not sure.

bogdandrutu: >I fear this will break most ocb users. If we go this route, is there a gentler approach we could take with feature gates or timelines? 

Also, right now they are broke anyway. Because if they don't specify any provider, it will try to use env/file at 109 and will fail.

mx-psi: I don't think there's much that prevents us from marking the other providers as 1.0, I will open the PR for those

mx-psi (dismissed):
I don't think we should do this in one step, and definitely not without updating all our examples and our own manifests in -releases.

bogdandrutu:
@mx-psi please propose a solution when blocking. We have a bug right now and we need a solution, this is my proposal to that, do you have an alternative?

cmd/builder/internal/builder/config.go
bogdandrutu: I don’t think we should rush the stability.

mx-psi:
> @mx-psi please propose a solution when blocking. We have a bug right now and we need a solution, this is my proposal to that, do you have an alternative?

For the immediate term, I can think of three options:
1. Publishing a `v0.109.0` for these two providers with the same contents as `v1.15.0`
2. Hardcoding the versions of these two providers on the builder
3. Fetching the `versions.yaml` file from Github and checking what's the `v1.x` version associated with a particular `v0.x`

I think doing (1) is harmless, fixes the issue for everybody and gives us more time

bogdandrutu:
I like 1 the most. I don't think 3 is feasible, I don't really like 2 because will grow a lot. 
> Publishing a v0.109.0 for these two providers with the same contents as v1.15.0

mx-psi:
After #11193 and #11194 all provider modules we include in core are now stable, so we can revisit this.

To be honest, this has been broken for a few releases and we haven't gotten a lot of complaints, so maybe it's fine to just go with this change if we give clear migration instructions. I am going to remove my 'request changes'.

cmd/builder/internal/builder/config.go
mx-psi: @TylerHelmuth revisiting this, if my memory serves we have had two issues about this in the last three weeks, so I am tempted to say we should just do this breaking change since it hasn't been a huge deal for people. Do you think we should wait more/do it differently?

bogdandrutu:
@mx-psi @codeboten @dmitryax I would propose to move forward with this and remove the problems in the future.

dmitryax:
I agree. This PR LGTM

.chloggen/fixbuilderproviders.yaml
```diff
@@ -0,0 +1,20 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'breaking'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: 'builder'
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove default config providers. This will require users to always specify the providers.
```
mx-psi, line 10: Can we add subtext with what users have to specify to keep the previous behavior?

mx-psi (approved):

mx-psi:
Let's give @TylerHelmuth one week to reply since he raised concerns about this

cmd/builder/internal/builder/config.go
TylerHelmuth: Will doing this mean anyone using ocb who doesn't include providers in the manifests.yaml will build a collector with 0 providers? I am pretty sure a collector setup like that will compile but not run, as we require at least 1 provider.

If that is true, then users who don't specify providers (most I'm guessing) will start getting runtime errors with their collectors. I much rather we do this gracefully instead of one breaking change.

If the assumptions I've made are wrong then the risk goes down.

TylerHelmuth:
> To be honest, this has been broken for a few releases and we haven't gotten a lot of complaints, so maybe it's fine to just go with this change if we give clear migration instructions. I am going to remove my 'request changes'.

@mx-psi no one else complained bc we released both stable and beta tags right? So the current default providers worked.

cmd/builder/internal/builder/config.go
TylerHelmuth: I'd like to see a confirmation on what happens with this change if a manifest is fed to ocb with no providers configured.

bogdandrutu:
> @mx-psi no one else complained bc we released both stable and beta tags right? So the current default providers worked.

They were broken for a long time.

mx-psi:
> > To be honest, this has been broken for a few releases and we haven't gotten a lot of complaints, so maybe it's fine to just go with this change if we give clear migration instructions. I am going to remove my 'request changes'.
> 
> @mx-psi no one else complained bc we released both stable and beta tags right? So the current default providers worked.

My understanding is that we did not fix it for v0.110.0, e.g. see https://github.com/open-telemetry/opentelemetry-collector/releases/tag/confmap%2Fprovider%2Fenvprovider%2Fv0.110.0 (not found).

cmd/builder/internal/builder/config.go
bogdandrutu: I made it clear that ocb command will fail with a clear message. So customer will not get into production bad collectors that are unable to start.

TylerHelmuth: I'd personally rather see this as a incremental change behind a feature gate instead of a breaking change.

We do incremental changes behind feature gates for breaking changes that only affect component developers and library authors, I think the same pattern should apply to end-user breaking changes.

mx-psi: I am fine with that, although we need to add support for feature gates on ocb first then :smile:

TylerHelmuth: 😿

TylerHelmuth: If feature gates support in ocb isn't digestible, then we should tell users who experience the breaking change exactly what to put in their manifest to get back to the previous behavior

mx-psi: I am fine with adding support, also fine with just a temporary CLI flag for this that we eventually remove

bogdandrutu: For this fix, having a flag or changing minimal the config I am confused where the benefit is. Also to remember this was and is broken.

bogdandrutu: I am 100% to add support in the future for featureflags.

mx-psi: For this particular change, I am fine with any of
- an error message that tells you what do very explicitly
- a feature gate
- a dedicated config flag

The "at least one provider is required" may not be super clear to people (they may not be aware of what providers they used to put in there).

.chloggen/fixbuilderproviders.yaml
TylerHelmuth: Agreed. This is going to break a lot of OCB users, I'd really like to make it as easy as possible to fix which means supplying as much help as possible.

cmd/builder/internal/builder/config.go
TylerHelmuth: Agreed.  Since this breaking change will affect end users of OCB who were not already setting providers, they'll be the most confused about what a provider is and what they need to set.  We need to make it obvious what to do from the error message and the changelog entry (and I believe the OCB readme as well), so that they can learn how to fix their manifest.

.chloggen/fixbuilderproviders.yaml
bogdandrutu: >Agreed. This is going to break a lot of OCB users, I'd really like to make it as easy as possible to fix which means supplying as much help as possible.

I don't believe it, recent data show that is not as big of a deal.

andrzej-stencel (approved):

bogdandrutu:
Closing this in favor of https://github.com/open-telemetry/opentelemetry-collector/pull/11566

Issue #11130, opened 2024-09-11
[configgrpc] enable xds support


rcrowe:
**Is your feature request related to a problem? Please describe.**

Our gRPC load-balancing infrastructure relies on in-process gRPC xDS, rather than a sidecar or more traditional load-balancer.
We're seeing stale endpoints being dialled as things scale up/down due to the configuration window on the round-robin service not being quick enough to pickup the changes.

**Describe the solution you'd like**

[configgrpc](https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/configgrpc) to import `_ "google.golang.org/grpc/xds"` enabling `endpoint` to be set with `xds:///{endpoint}` and the `balancer` config to be left empty.

PR #11137, opened 2024-09-11
[configgrpc] Register xDS balancer

rcrowe:
#### Description

In order for an endpoint prefixed with `xds://...` to enable xDS support the balancer needs to be registered

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/11130


atoulme:
Please add a changelog.

PR #11145, opened 2024-09-11
[processorhelper] deprecate additional obsreport funcs

codeboten:
These functions were only used in one processor in both core and contrib. Moving the metrics directly to that processor and deprecating the processorhelper funcs for now.


codeboten:
Alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/11114

djaglowski (approved):

PR #11158, opened 2024-09-12
[chore]: Add a helper function to send error responses for network receivers.

VihasMakwana:
This PR adds a helper function designed for network receivers to facilitate proper error handling.

All the receivers should adhere to receiver contract defined https://github.com/open-telemetry/opentelemetry-collector/blob/1c47d89fc38df12133adc78c4ea7bc4d3374e6b6/receiver/doc.go#L23-L40
#### Testing

Added

#### Usage example for a network receiver

```go

err := nextConsumer.ConsumeLogs(...)
if err != nil {
       confighttp.Error(responseWriter, err) // this will update the status code depending upon the error
       // log the error
}

```

## Note:
Currently, many network based receivers don't follow the contract and they return `Service Unavailable` (without checking for the error type). 
Implementing this helper would enhance both the readability and reusability of the code.


mx-psi (commented):
Thanks for your PR! This feels like it has some overlap with #9041, for which there is active discussion still. I would suggest we wait until that PR gets merged, check if your use case is addressed, and then retake this

VihasMakwana:
> Thanks for your PR! This feels like it has some overlap with #9041, for which there is active discussion still. I would suggest we wait until that PR gets merged, check if your use case is addressed, and then retake this

Thank you for sharing the PR link; I wasn’t aware of it before.

I’d like to point out that this PR adds a helper function to allow network receivers to send errors back to the sender more easily. Currently, receivers would need to handle this manually with code like:

```go
if consumererror.IsPermanent(err) {
     response.WriteHeader(http.StatusBadRequests)
} else {
    response.WriteHeader(http.StatusInternalServerError)
}
```

Implementing this manually in every receiver could limit reusability and make the code less maintainable.

----

The PR you linked is about implementing a new method for consumers to return errors, whereas my PR is specifically designed for use by network receivers. Let me know what do you think?

mx-psi:
> The PR you linked is about implementing a new method for consumers to return errors, whereas my PR is specifically designed for use by network receivers. Let me know what do you think?

I didn't mean to say that your PR was unnecessary if we merge #9041, only that consumers report errors may affect the design of the API on this PR. I may be wrong, do you think these can be dealt with independently?

VihasMakwana:
@mx-psi thanks for your response. This PR currently uses the existing consumererror API, which differentiates between permanent and non-permanent errors. Referring to the code snippet [here](https://github.com/open-telemetry/opentelemetry-collector/pull/9041#issuecomment-2313397907), we can easily update this changes of this PR to adopt the new API once it is implemented in the contrib repository. I expect this might take some time. 

Therefore, I believe we can handle these PRs independently. After this PR is merged, we can update the receivers to align with the contract. 

Let me know your thoughts on this. I appreciate you taking the time to review it! 😄

VihasMakwana:
@mx-psi If you have any feedback or if there's anything else you need from me to move forward, please let me know.

mx-psi:
I won't have time to review this soon, sorry. And if it has any relation to #9041 I would like #9041 to be merged first in any case, it has been blocked for a long time and I wouldn't want to add extra burden

VihasMakwana:
> I won't have time to review this soon, sorry. And if it has any relation to #9041 I would like #9041 to be merged first in any case, it has been blocked for a long time and I wouldn't want to add extra burden

That makes sense. I'll move these changes to contrib repo then and begin working on stabilising the network receivers. Let me know how it sounds.

mx-psi:
Sounds reasonable!

VihasMakwana:
Let's revisit this later when https://github.com/open-telemetry/opentelemetry-collector/pull/9041 gets merged

PR #11160, opened 2024-09-12
[mdatagen] add stability detail for metrics

codeboten:
This allows component authors to specify the stability level of telemetry emitted by components. This provides visibility to end users as to what they can expect from the telemetry.


cmd/mdatagen/metadata-schema.yaml
```diff
@@ -138,6 +138,8 @@ telemetry:
       enabled: bool
       # Required: metric description.
       description:
+      # Optional: the stability level of the metric. Set to alpha by default.
+      stability: [alpha|stable|deprecated]
```
mx-psi, line 142: Should we have also a 'since' field?

```yaml
stability:
  level: [alpha|stable|deprecated]
  since: # version
```

atoulme (approved):

mx-psi (approved):

Issue #11155, opened 2024-09-12
[exporter/debug] Add profiling support


mx-psi:
The debug exporter would be really useful for initial testing and validation of the profiling signal.

This PR adds support for the OTLP exporter and can be used as inspiration: #11131

PR #11131, opened 2024-09-11
Profiles support in OTLP Exporter

dmathieu:
#### Description

This adds profiles support to the OTLP exporters.
Note: This includes some commits from #11071 that were required in here too.

PR #11161, opened 2024-09-12
[chore]: Add Profiles Marshaler to otlptext

julianocosta89:
#### Description
This is the initial work towards #11155.
This PR adds profiles marshalers to `otlptext`, which will be used in the `debugexporter` later on.

To continue the work on the `debugexporter` I'd wait on the PR #11131, as it adds some shared components.

#### Testing
Added `TestProfilesText`, testing the following scenarios: 
- `empty_profiles`
- `two_profiles`


.chloggen/julianocosta89_debugexporter-profiles.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: debugexporter
```
dmathieu, line 7: This is a chore PR, and only modifies internal packages. Shouldn't wee keep the changelog entry for when the marshaler is used?

exporter/internal/otlptext/profiles.go
```diff
@@ -0,0 +1,43 @@
...
+			buf.logEntry("ScopeProfiles SchemaURL: %s", ilp.SchemaUrl())
+			buf.logInstrumentationScope(ilp.Scope())
+			profiles := ilp.Profiles()
+			for k := 0; k < profiles.Len(); k++ {
+				buf.logEntry("Profile #%d", k)
+				profile := profiles.At(k)
+				buf.logAttr("Profile ID", profile.ProfileID())
+				buf.logAttr("Start time", profile.StartTime().String())
+				buf.logAttr("End time", profile.EndTime().String())
+				buf.logAttributes("Attributes", profile.Attributes())
```
dmathieu, line 37: Should we log the samples?

.chloggen/julianocosta89_debugexporter-profiles.yaml
mx-psi: Agreed, I think we can remove the changelog entry

exporter/internal/otlptext/profiles.go
julianocosta89: @dmathieu that sounds good!
Should I loop through all samples from each profile though?
That would add an extra loop in a already 3 loops chain.

I was thinking about something like this:

```go
if ps := profile.Profile().Sample(); ps.Len() != 0 {
	for l := 0; l < ps.Len(); l++ {
		buf.logAttr("Sample at %d", l)
		buf.logAttr("Value", ps.At(l).Value())
	}
}
```

WDYT?

dmathieu: I'm not sure how this is different from looping through each profile. Isn't that what it does?
Is it the amount of data possibly being logged that concerns you?

julianocosta89: Yep

julianocosta89: Do we usually get a lot of samples on each profile?

dmathieu: The number of samples is `sampling frequency * number of cores * report interval`. So there can be many, yes.
But this is a debug exporter, and it will only be enabled if the required telemetry level is "detailed". So I'd be tempted to display their value anyway.

mx-psi: I think for detailed level it makes sense to print the samples. In the normal level output (which would live here https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/debugexporter/internal/normal ) we can abbreviate it more

julianocosta89: @dmathieu would you have some time to take a look?
I've changed a couple of stuff to have all values printed (or most of it).

pdata/testdata/profile.go
```diff
@@ -38,6 +38,8 @@ func fillProfileOne(profile pprofile.ProfileContainer) {
 	profile.SetEndTime(profileEndTimestamp)
 	profile.SetDroppedAttributesCount(1)
 
+	profile.Profile().LocationIndices().FromRaw([]int64{1})
```
dmathieu, line 41: Changing the profiles testdata is probably what is making all the other tests fail.

dmathieu: For other profiles, the testdata remains fairly simple. And tests that need more create their own data (possibly with a testdata start).
Doing that would avoid having to make changes everywhere to let the tests pass.

julianocosta89: Thank you! Moved it to its own method.

dmathieu: I'll let @mx-psi chime in, but I think it'd be better if this method were in the otlptext package, as pdata is public and therefore has much stricter stability requirements.

mx-psi: > I think it'd be better if this method were in the otlptext package

Agreed, better to make it internal 

>  pdata is public and therefore has much stricter stability requirements

In this case the `pdata/testdata` folder is its own module so it wouldn't be as bad, but still if the only user is `otlptext` then it should be in `otlptext`

julianocosta89: moved to `otlptext`, thank you @dmathieu and @mx-psi ⭐

julianocosta89:
I'm looking for some guidance here.
This test is flaky: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10957119763/job/30424592592

This is because sometimes we get the returned value as:

```
-> value: Map({"intValue":"42"})
-> key: Str(answer)
```

and sometimes like this:
```
-> key: Str(answer)
-> value: Map({"intValue":"42"})
```

Any suggestions?

dmathieu:
Your failing issue is because maps are not ordered in Go. So when the attribute table is being built with [FromRaw](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pcommon/map.go#L275-L292), the generated slice doesn't guarantee order.

We could possibly ensure order by sorting the keys before injecting in FromRaw. But I'm not sure that's relevant. Maybe figuring out a way to not rely on order of the output would be better?

julianocosta89:
> Your failing issue is because maps are not ordered in Go. So when the attribute table is being built with [FromRaw](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pcommon/map.go#L275-L292), the generated slice doesn't guarantee order.
> 
> We could possibly ensure order by sorting the keys before injecting in FromRaw. But I'm not sure that's relevant. Maybe figuring out a way to not rely on order of the output would be better?

The problem is that the test is a simple comparison `created_prof` == `two_profiles.out`.

I've simplified the test to validate the `AttributeTable` with just one element to avoid the flakiness.

dmathieu (approved):

mx-psi (approved):
Since we don't give any guarantees regarding the debug exporter output, I don't think we need to get it perfect, and this looks good enough to me. Merging!

PR #11166, opened 2024-09-13
[Chore] Migration error and helper into scraper model

### LINKED ISSUE #11003: [receiver] Move scraper packages into scraper module
We have two packages, [`scrapererror`](https://pkg.go.dev/go.opentelemetry.io/collector/receiver@v0.108.1/scrapererror) and [`scraperhelper`](https://pkg.go.dev/go.opentelemetry.io/collector/receiver@v0.108.1/scraperhelper) that are part of the receiver module, but are not necessary for Collector v1 (we only intend to stabilize the OTLP receiver at first).

We can make a new `receiver/scraper` module that has these two as packages.  This would reduce the API surface that we have to stabilize.

zzzk1:
#### Description
Move helper and error packages into scraper module.

#### Link to tracking issue
Fixes #11003


receiver/scraper/helper/internal/metadata/generated_telemetry.go
```diff
@@ -6,6 +6,7 @@ import (
 	"errors"
 
 	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
```
zzzk1, line 9: I'm confused why the newly generated code has the line.

receiver/scraper/error/doc.go
```diff
@@ -2,4 +2,4 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Package scrapererror provides custom error types for scrapers.
-package scrapererror // import "go.opentelemetry.io/collector/receiver/scrapererror"
+package error // import "go.opentelemetry.io/collector/receiver/scraper/error"
```
mx-psi, line 5: Can we keep this as scrapererror? (Here and in other files?) And also name the folder `scrapererror`?

`error` is (I think) already a common package name so to avoid collisions and to help existing users to migrate more easily, I would rather keep the `scrapererror` name

receiver/scraper/helper/config.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package scraperhelper // import "go.opentelemetry.io/collector/receiver/scraperhelper"
```
mx-psi, line 4: Same here, I would keep this as `scraperhelper`, the repetition with the parent folder is fine

mx-psi (commented):
Thanks for the PR! This is a good start, I have two comments:

1. Can we keep the original names (`scraperhelper`, `scrapererror`)? When importing, Go will only use the package name, so we tend to repeat the parent package name if the name is not specific
2. To help in the transition, we should keep the original `scraper` packages for one release as documented here: https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#breaking-changes. To do this, you can have a type alias for each symbol like this:

```go
type ScrapeFunc = scraperhelper.ScrapeFunc
```

We then can deprecate the whole package as mentioned here https://go.dev/wiki/Deprecated#examples

receiver/scraper/helper/internal/metadata/generated_telemetry.go
mx-psi: > I'm confused why the newly generated code has the line.

:thinking: I am not sure myself, it's probably not related to the changes on this PR, so for now I will take a look but for now you can keep on working on the stuff I mentioned below

mx-psi:
Looks like this needs a `make gotidy`

mx-psi (approved):

receiver/scraper/Makefile
```diff
@@ -0,0 +1 @@
+include ../../Makefile.Common
```
bogdandrutu, line 1:
```suggestion
include ../../Makefile.Common

```

bogdandrutu (commented):
Can you try to make this to preserve history. in the old packages (to simplify migration) would be easier to make the old types/funcs aliases of the new one.

PR #11169, opened 2024-09-13
[mdatagen] Fix metric tests on boolean attributes

jade-guiton-dd:
#### Description

Since commit 8f3ca8a, `mdatagen` improperly generates metric tests: boolean attributes are always expected to be `true` in the output, no matter what the injected test value was. For example, [this line in collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/19b01db307d984e62f81f1880b63327ba092cd81/receiver/apachesparkreceiver/internal/metadata/generated_metrics_test.go#L1250) is incorrectly changed to expect `true` instead of `false` after updating mdatagen and running it again.

The reason is [this line](https://github.com/open-telemetry/opentelemetry-collector/blob/8f3ca8aaf58539e5afab54d365f97624efe7cbc3/cmd/mdatagen/templates/metrics_test.go.tmpl#L182), which attempts to generate `assert.True` or `assert.False` based on the expected test value. However, `(attributeInfo $attr).TestValue` does not return a boolean but the *string* `true` or `false`, both of which are considered "truthy" by the templating engine.

#### Testing
I manually rebuilt mdatagen, ran `make generate` in collector-contrib, and ran the tests, to confirm that the test related to the previously mentioned change was no longer failing.

### DIFF
--- a/.chloggen/fix-mdatagen-metrics-bool-attr.yaml
+++ b/.chloggen/fix-mdatagen-metrics-bool-attr.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: mdatagen
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix incorrect generation of metric tests with boolean attributes.
+
+# One or more tracking issues or pull requests related to the change
+issues: [11169]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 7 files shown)


codeboten (commented):
Thanks for the fix @jade-guiton-dd, could you add a test with the value shown in the link you posted: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/19b01db307d984e62f81f1880b63327ba092cd81/receiver/apachesparkreceiver/internal/metadata/generated_metrics_test.go#L1250

mx-psi (approved):

jade-guiton-dd:
> Thanks for the fix @jade-guiton-dd, could you add a test with the value shown in the link you posted

I added an attribute in the samplereceiver metrics which should trigger a test failure if this bug shows up again (as long as the computation of TestValues isn't changed). Is this what you had in mind @codeboten?

codeboten (approved):
Yup, thanks @jade-guiton-dd, looks great

PR #11144, opened 2024-09-11
[processorhelper] report signal as attribute on incoming/outgoing items

codeboten:
This updating the existing metric points that were recently added to use signal as an attribute instead of separating the metric name. It follows the suggestions in [otep 259](https://github.com/open-telemetry/oteps/pull/259) for the metric and attribute names.

Putting this in draft to get some feedback from @djaglowski before moving forward with this change

PR #11171, opened 2024-09-13
Add 'pipeline' attribute to processor metrics

### LINKED ISSUE #10908: Processor telemetry should include Pipeline ID as attribute

The `processorhelper` package defines several metrics but provides only a "Processor ID" attribute. This is insufficiently observable because a processor ID may apply to multiple instances of a processor. In effect, all instances of the processor _configuration_ are reporting metrics as if they are the same _instance_. 

Users should be able to understand the behavior of a processor within the context of each pipeline because the usage may be very different per pipeline.

As an example, suppose a processor used in two pipelines. In one pipeline, it reports 1000/1000 log records accepted, 0 for other metrics. In the other pipeline, it reports 10/10 log records rejected, 0 for all other metrics. Currently, the metric describes this as if 99% of data is being accepted by the processor. The reality that one pipeline is very healthy while the other is very unhealthy can easily be overlooked because of the lack of a pipeline ID attribute.

Add a "Pipeline ID" attribute to all processor helper metrics.

This is proposed as an alternative to https://github.com/open-telemetry/opentelemetry-collector/pull/10664, which would eliminate the need for the Pipeline ID attribute, but would introduce breaking changes elsewhere.

**Additional Context**

https://github.com/open-telemetry/opentelemetry-collector/issues/10708#issuecomment-2246340782

djaglowski:
Resolves https://github.com/open-telemetry/opentelemetry-collector/issues/10908

This ends up being a pretty straightforward change in actual code, but a tedious change in tests. It's also obviously a breaking change for anyone depending on these metrics.

### DIFF
--- a/processor/processortest/nop_processor.go
+++ b/processor/processortest/nop_processor.go
@@ -13,6 +13,7 @@ import (
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/consumer/consumerprofiles"
 	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/pipeline"
 	"go.opentelemetry.io/collector/processor"
 	"go.opentelemetry.io/collector/processor/processorprofiles"
 )
@@ -20,11 +21,12 @@ import (
 var nopType = component.MustNewType("nop")
 
 // NewNopSettings returns a new nop settings for Create*Processor functions.
-func NewNopSettings() processor.Settings {
+func NewNopSettings(dt pipeline.Signal) processor.Settings {
 	return processor.Settings{
 		ID:                component.NewIDWithName(nopType, uuid.NewString()),
 		TelemetrySettings: componenttest.NewNopTelemetrySettings(),
 		BuildInfo:         component.NewDefaultBuildInfo(),
+		PipelineID:        pipeline.NewID(dt),
 	}
 }
 
--- a/.chloggen/processor-pipeline-id-attribute-2.yaml
+++ b/.chloggen/processor-pipeline-id-attribute-2.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: processorhelper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Processor metrics now include the `pipeline` attribute.
+
+# One or more tracking issues or pull requests related to the change
+issues: [11171]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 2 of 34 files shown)


djaglowski:
One notable detail is that the `otel.signal` attribute added in https://github.com/open-telemetry/opentelemetry-collector/pull/11144 is arguably redundant, since pipeline ID's contain the data type. However, we likely will want `otel.signal` for receiver and exporter metrics as well, so perhaps it makes sense to leave it. (Connector metrics would need to differentiate between incoming and outgoing data type though.)

mx-psi (commented):
The approach looks reasonable to me. There are some breaking changes on the test modules, I feel like those are okay (and that maybe we should split that part of the modules off to make our lives easier, but that's for another PR)

djaglowski:
> The approach looks reasonable to me. There are some breaking changes on the test modules, I feel like those are okay (and that maybe we should split that part of the modules off to make our lives easier, but that's for another PR)

I added a changelog to note to breaking change in `processortest.NewNopSettings`

djaglowski:
I opened https://github.com/open-telemetry/opentelemetry-collector/issues/11179 to capture thoughts on the broader question of which attributes to apply to each kind of component. This PR aligns with what I believe is the best overall schema, but please take a look.

processor/processortest/nop_processor.go
```diff
@@ -20,11 +20,12 @@ import (
 var nopType = component.MustNewType("nop")
 
 // NewNopSettings returns a new nop settings for Create*Processor functions.
-func NewNopSettings() processor.Settings {
+func NewNopSettings(dt component.DataType) processor.Settings {
```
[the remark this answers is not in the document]
djaglowski: It seems likely we'll want to include the data type for other component kinds. I suppose we could pass the pipeline ID here for processors, but not sure there's any case to do it for other kinds. This is probably something worth trying to iron out as part of https://github.com/open-telemetry/opentelemetry-collector/issues/11179.

processor/internal/processor.go
```diff
@@ -14,4 +17,7 @@ type Settings struct {
 
 	// BuildInfo can be used by components for informational purposes
 	BuildInfo component.BuildInfo
+
+	// PipelineID indicates which pipeline contains this processor.
+	PipelineID pipeline.ID
```
bogdandrutu, line 22: Settings are passed to the factory method correct? Which already signals which "signal" is this for. I am confused why adding it here?

bogdandrutu (commented):
Is this already done?

processor/internal/processor.go
djaglowski: This indicates the pipeline ID, not the signal. The rationale for adding pipeline ID is explained in https://github.com/open-telemetry/opentelemetry-collector/issues/10908

mx-psi (approved):

bogdandrutu:
Please rebase, looks like everything changed since this change was made.

djaglowski:
I'm closing this in favor of the graph based approach, demonstrated in https://github.com/open-telemetry/opentelemetry-collector/pull/11311#pullrequestreview-2345261928 and more robustly proposed in https://github.com/open-telemetry/opentelemetry-collector/issues/11343.

PR #11172, opened 2024-09-13
[chore] Remove prometheus need to check test metrics in componenttest

bogdandrutu:


codeboten (commented):
i love this idea, will need to do a full review but i think it makes sense to remove. are there still e2e tests that validate the output from the service?

bogdandrutu:
> i love this idea, will need to do a full review but i think it makes sense to remove. are there still e2e tests that validate the output from the service?
Yes in https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/service_test.go

bogdandrutu:
@codeboten ping :)

codeboten (approved):

codeboten:
```
/home/runner/work/opentelemetry-collector/opentelemetry-collector/.tools/golangci-lint run
Error: componenttest/otelchecker.go:131:49: `unknwon` is a misspelling of `unknown` (misspell)
		return metricdata.DataPoint[N]{}, fmt.Errorf("unknwon metric type: %T", a)
```

PR #11175, opened 2024-09-14
[pdata]: Add support to MoveTo for Map, allow avoiding copies

bogdandrutu:
#### Description

Can be used to remove unnecessary copies from places like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c52a9a7ae1563127254530d416a8c569c54750bd/pkg/ottl/ottlfuncs/func_replace_all_patterns.go#L100


TylerHelmuth (approved):

dmitryax (approved):
LGTM

PR #11182, opened 2024-09-16
[chore] Remove use of alias types in service/telemetry

bogdandrutu:
This PR moves internal definitions that were aliased publicly to the public package, and makes sure that hides any other type so that the API surface is the same.


djaglowski (approved):

codeboten (approved):

codeboten:
I was looking at the original PR, @mx-psi was there a reason to have the internal aliases?

mx-psi (approved):
I believe the internal aliasing stuff may have been needed as part of the refactor but not now. In any case I believe this leaves the API surface the same so if we were to need this we can always revisit this

codeboten:
Thanks for confirming!

PR #11185, opened 2024-09-17
[chore] Move metrics initialization in service/telemetry

bogdandrutu:


service/telemetry/factory.go
```diff
@@ -7,15 +7,26 @@ import (
...
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configtelemetry"
+	"go.opentelemetry.io/collector/featuregate"
+	"go.opentelemetry.io/collector/service/internal/resource"
 	"go.opentelemetry.io/collector/service/telemetry/internal"
 )
 
+// disableHighCardinalityMetricsfeatureGate is the feature gate that controls whether the collector should enable
+// potentially high cardinality metrics. The gate will be removed when the collector allows for view configuration.
+var disableHighCardinalityMetricsfeatureGate = featuregate.GlobalRegistry().MustRegister(
```
codeboten, line 24: Should this be moved to globalgates instead?

codeboten (approved):
this looks mostly ok, a big chunk of this code (all the stuff in otelinit) will be removed once https://github.com/open-telemetry/opentelemetry-collector/pull/11079 lands

service/telemetry/factory.go
bogdandrutu: Why? Where else is used? My understanding is that in globalgates we put things shared between components/modules.

codeboten: ah, i just thought of it as a convenient location for all feature gates 🤣

PR #11198, opened 2024-09-17
Exporterhelper: address timeout misalignment w/ context deadline

### LINKED ISSUE #11183: Exporterhelper does not respect context deadlines
**Describe the bug**

In a gateway configuration of the collector, it is likely for an OTLP receiver and exporter to be configured. Any gRPC exporter/receiver pair that both propagate timeout information are likely to fall into this problem scenario.  (For example, the OTel-Arrow components will behave the same after https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34742 merges.)

In a synchronous pipeline, it means the caller's context is likely to have a deadline before it reaches an exporter, typically added by the pipeline receiver or propagated from an earlier pipeline segment's exporter.

Note that the timeout sender does recognize the incoming deadline, but not intentionally. [The Golang context package will never increase a deadline, but it will shorten a deadline](https://pkg.go.dev/context#WithDeadline):

> WithDeadline returns a copy of the parent context with the deadline adjusted to be no later than d. If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent.

The result is that exporterhelper will pass through a timeout shorter than its configured timeout. This may be working as intended for some, and a bug for others.  When I set a 30s timeout on the exporter and my receiver observes a 5s timeout instead, it's likely to cause confusion.

The behavior of the retry sender is even more problematic.  The retry sender is likely to create the problem described above by ignoring the arriving context timeout.  When the retry sender's max_elapsed setting is greater than the arriving context timeout, we are virutally assured of issuing one short-timeout request at the end of a series of retries.

Say you've configured your exporterhelper like this:

```
retry_sender:
  enabled: true
timeout: 30s
```

If there is an incoming context deadline that is shorter than the configured max_elapsed setting, retry sending will continue past the original deadline.  Since the arriving context deadline is preserved through all of this, the final export request is likely to have a timeout shorter than 30s.  This leads to confusion.

**Steps to reproduce**
To reproduce this I created a client application that would send as fast as it can through a collector pipeline w/ both timeout and retry sender configured.

I was expecting to see what I saw, because I've seen this in production and wanted to reproduce it.

**What did you wish to see instead?**
I have several wishes here.

First, the retry sender should consider the context deadline. When `backoffDelay` is computed it should be compared against the context deadline (if set).  If there is a deadline that is less than the backoff delay, we should not fall into the selct statement and wait for cancelation, we should fail fast at that point w/ an gRPC Aborted code, message "insufficient context deadline for retry".

Second, the exporter timeout sender would ideally have a way to impose a minimum timeout.  When the timeout sender is configured with a 5s timeout and the arriving context has a 1s deadline, I think I'd like two optional behaviors:

- Ignore the arriving timeout, let the configured timeout override. With this option, I expect the caller's export will end in as deadline-exceeded while the pipeline continues to export the data w/ an independent deadline.
- Fail fast.  When the arriving context timeout is less than the configured timeout, fail immediately w/ an Aborted code.

Third, I would like for the OTLP receiver to support the same feature, a configuration that allows ignoring Receiver to ignore arriving timeouts when they are unreasonably short.  This will have to be done on a per-receiver basis, but possibly a `receiverhelper` library could help receivers have uniform behavior.

**What version did you use?**
v0.108.x

**What config did you use?**
as shown above.

**Environment**

**Additional context**

jmacd:
#### Description

Retry sender: if the context deadline will expire before a configured attempt to retry, abort the request.  Otherwise, the retry sender will delay until the context timeout happens, which is a predictable waste of resources. Better to return immediately. As for the aborted status code, I could be convinced instead to stop retrying and return the final-retry-attempt's error. 

Timeout sender: if the context deadline will expire before the full configured timeout, present two optional new behaviors in addition to the current behavior. The three are:

- sustain: existing behavior, let a short timeout pass through to the exporter
- abort: new behavior, do not send a request if the timeout expires after the context deadline
- ignore: new behavior, send a request with the configured timeout, ignore the context deadline. 

#### Link to tracking issue
Fixes #11183.

#### Testing

TODO: This is draft for discussion purposes. I have not added or modified existing tests, and I'm aware of stylistic requirements that are not met (e.g., the use of bare string values, vs. use of a constrained value such as `configcompression.Type`).

#### Documentation

WIP: DRAFT.

### DIFF
--- a/exporter/exporterhelper/retry_sender.go
+++ b/exporter/exporterhelper/retry_sender.go
@@ -111,7 +111,7 @@ func (rs *retrySender) send(ctx context.Context, req Request) error {
 		if deadline, has := ctx.Deadline(); has && time.Until(deadline) < backoffDelay {
 			// The delay is longer than the deadline.  There is no point in
 			// waiting for cancelation.
-			return fmt.Errorf("Exporting aborted. Deadline is shorter than retry interval.")
+			return fmt.Errorf("request will be cancelled before next retry: %w", err)
 		}
 
 		backoffDelayStr := backoffDelay.String()
--- a/exporter/exporterhelper/retry_sender.go
+++ b/exporter/exporterhelper/retry_sender.go
@@ -108,6 +108,12 @@ func (rs *retrySender) send(ctx context.Context, req Request) error {
 			backoffDelay = max(backoffDelay, throttleErr.delay)
 		}
 
+		if deadline, has := ctx.Deadline(); has && time.Until(deadline) < backoffDelay {
+			// The delay is longer than the deadline.  There is no point in
+			// waiting for cancelation.
+			return fmt.Errorf("Exporting aborted. Deadline is shorter than retry interval.")
+		}
+
 		backoffDelayStr := backoffDelay.String()
 		span.AddEvent(
 			"Exporting failed. Will retry the request after interval.",
--- a/exporter/exporterhelper/timeout_sender.go
+++ b/exporter/exporterhelper/timeout_sender.go
@@ -17,6 +17,15 @@ type TimeoutConfig struct {
 	// Timeout is the timeout for every attempt to send data to the backend.
 	// A zero timeout means no timeout.
 	Timeout time.Duration `mapstructure:"timeout"`
+
+	// OnShortTimeout is the exporter's disposition toward short timeouts,
+	// any time the incoming context's deadline is shorter than the
+	// configured Timeout.  The choices are:
+	//
+	// - sustain: use the shorter-than-configured timeout
+	// - abort: abort the request for having insufficient deadline on arrival
+	// - ignore: ignore the previous deadline and proceed with the configured timeout
+	OnShortTimeout string `mapstructure:"on_short_timeout"`
 }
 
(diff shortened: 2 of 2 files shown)


exporter/exporterhelper/retry_sender.go
```diff
@@ -108,6 +108,12 @@ func (rs *retrySender) send(ctx context.Context, req Request) error {
 			backoffDelay = max(backoffDelay, throttleErr.delay)
 		}
 
+		if deadline, has := ctx.Deadline(); has && time.Until(deadline) < backoffDelay {
+			// The delay is longer than the deadline.  There is no point in
+			// waiting for cancelation.
+			return fmt.Errorf("Exporting aborted. Deadline is shorter than retry interval.")
```
bogdandrutu, line 114:
```suggestion
			return fmt.Errorf("request will be cancelled before next retry: %w", err)
```

bogdandrutu (approved):

jmacd:
@bogdandrutu Thank you. I will work the rest of this PR and add tests.

PR #11201, opened 2024-09-17
[processorhelper] deprecated accepted/refused/dropped metrics

codeboten:
These were only used by the memory limiter processor and were never automatically calculated by the processorhelper. It's better to move them to processor specific metrics that can be managed within the component itself.


codeboten:
This follows the decision that was mentioned in https://github.com/open-telemetry/opentelemetry-collector/pull/11114

mx-psi (approved):

djaglowski (approved):

Issue #11167, opened 2024-09-13
[CI] Contrib tests should also test updated mdatagen


mx-psi:
When we run `make update-otel` we update the `mdatagen` version as well. To avoid issues in contrib, we have the contrib tests but they are not testing this.

I believe in order to do this it would be as simple as adding a `-replace` option on check-contrib here https://github.com/open-telemetry/opentelemetry-collector/blob/5fc39ba6334f396c3b5497bf6d29db12823cb50c/Makefile#L260C1-L260C14 to also replace the version of mdatagen

jade-guiton-dd:
For fully testing the implications of updating `mdatagen`, I think it will also be necessary to forcefully rebuild `.tools/mdatagen` and run `make generate` in `contrib`. Adding a matching `-dropreplace` in the `restore-contrib` target will also be necessary.

PrateekKumar1709:
Hi, I would like to work on this issue. Can you please assign this to me?

mx-psi:
#11519 added an additional check on mdatagen: https://github.com/open-telemetry/opentelemetry-collector/blob/e76145a63a597531dcc7fd109f8cf8c620135cc9/cmd/mdatagen/internal/metadata.go#L55 which did not make CI fail. Reopening this.

jade-guiton-dd:
I'll take a loot at this.

mx-psi:
The errors I am seeing on CI in #12305 make me think this is broken again :(

jade-guiton-dd:
@mx-psi That's because my fix involving `generate-contrib` was disabled in #11795, as it sometimes failed because of non-convergent `go mod tidy`. I'll un-revert it when #37418 gets merged. I'll reopen this issue and set it as blocked for now.

PR #11204, opened 2024-09-17
[component] Refactor to use pipeline.ID and pipeline.Signal

TylerHelmuth:
#### Description
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/11209

This PR is a non-breaking implementation of https://github.com/open-telemetry/opentelemetry-collector/pull/10947. It adds a new module, `pipeline`, which houses a `pipeline.ID` and `pipeline.Signal`. `pipeline.ID` is used to identify a pipeline within the service. `pipeline.Signal` is uses to identify the signal associated to a pipeline.

I do this work begrudgingly. As the PR shows, this is a huge refactor when done in a non-breaking way, will require 3 full releases, and doesn't benefit our [End Users or, in my opinion, our Component Developers or Collector Library Users](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md#target-audiences). I view this refactor as a Nice-To-Have, not a requirement for Component 1.0.

#### Link to tracking issue
Works towards https://github.com/open-telemetry/opentelemetry-collector/issues/9429


cmd/mdatagen/internal/templates/component_test.go.tmpl
TylerHelmuth, line 1: This causes contrib tests to fail, related to https://github.com/open-telemetry/opentelemetry-collector/issues/11167, but if I don't update it yet then linting fails.  Should I add a `// nolint`?

.chloggen/component-refactor-datatype-2.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: componentprofiles
```
mx-psi, line 7: I feel like we can do this one in one go since this is quite experimental yet

component/componentprofiles/config.go
```diff
@@ -3,13 +3,23 @@
 
 package componentprofiles // import "go.opentelemetry.io/collector/component/componentprofiles"
 
-import "go.opentelemetry.io/collector/component"
+import (
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/pipeline"
+)
 
+// nolint
```
mx-psi, line 11: Why does this need a nolint?

component/componentstatus/instance.go
```diff
@@ -66,9 +81,44 @@ func (id *InstanceID) AllPipelineIDs(f func(component.ID) bool) {
 	}
 }
 
+// AllPipelineIDsWithPipelineIDs calls f for each pipeline this instance is associated with. If
```
mx-psi, line 84: I can't think of a better name but this made me laugh :smile:

.chloggen/component-refactor-datatype-6.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: connector
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecates `TracesRouterAndConsumer`, `NewTracesRouter`, `MetricsRouterAndConsumer`, `NewMetricsRouter`, `LogsRouterAndConsumer`, and `NewLogsRouter`. Use `TracesRouterAndConsumerWithPipelineIDs`, `NewTracesRouterWithPipelineIDs`, `MetricsRouterAndConsumerWithPipelineIDs`, `NewMetricsRouterWithPipelineIDs`, `LogsRouterAndConsumerWithPipelineIDs`, and `NewLogsRouterWithPipelineIDs` instead.
```
mx-psi, line 10: I would be fine with breaking this as well in 1-step

cmd/mdatagen/internal/templates/component_test.go.tmpl
mx-psi: It's fine to me to make the change in one step for the connector interfaces. If others disagree, then I would add `// nolint` here

exporter/exporterqueue/queue.go
```diff
@@ -24,7 +25,12 @@ type Queue[T any] queue.Queue[T]
 
 // Settings defines settings for creating a queue.
 type Settings struct {
-	DataType         component.DataType
+
+	// Deprecated: [v0.110.0] Use Signal instead
+	// nolint
```
mx-psi, line 30: Why do we need nolint here?

.chloggen/component-refactor-datatype-2.yaml
TylerHelmuth: We end up needing `DataTypeProfiles` to support the deprecation process in `service`.  The deprecated `Config.Validate` for example needs it.

We could update that method to convert from `SignalProfiles` to a `DataType`, but I'm not sure thats easier.

component/componentprofiles/config.go
TylerHelmuth: The function below this returns a `component.DataType` which is deprecated.  The linter complains we are using it, but we need to use it.

component/componentstatus/instance.go
TylerHelmuth: Yes, the connector interfaces are the same.

.chloggen/component-refactor-datatype-6.yaml
TylerHelmuth: I will update the connector stuff to be breaking.

exporter/exporterqueue/queue.go
TylerHelmuth: Even tho I marked this field as deprecated itself, the linter still complained that it was using the deprecated `component.DataType`.

.chloggen/component-refactor-datatype-6.yaml
TylerHelmuth: Updated

.chloggen/component-refactor-datatype-2.yaml
mx-psi: Fair

component/componentprofiles/go.mod
```diff
@@ -2,7 +2,10 @@ module go.opentelemetry.io/collector/component/componentprofiles
 
 go 1.22.0
 
-require go.opentelemetry.io/collector/component v0.109.0
+require (
+	go.opentelemetry.io/collector/component v0.0.0-00010101000000-000000000000
```
mx-psi, line 6:
```suggestion
	go.opentelemetry.io/collector/component v0.109.0

**Review by mx-psi — APPROVED** [review]
Other than fixing the merge conflicts, this looks good to me

File: exporter/exporterqueue/queue.go [hunk 7]
```diff
@@ -24,7 +26,12 @@ type Queue[T any] queue.Queue[T]
 
 // Settings defines settings for creating a queue.
 type Settings struct {
-	DataType         component.DataType
+
+	// Deprecated: [v0.110.0] Use Signal instead
```
codeboten, line 30: Does this deprecated note work as expected if there's a nolint below it? i think alternatively nolint can be added to the end of line 32

exporter/exportertest/contract_checker.go
```diff
@@ -33,7 +34,12 @@ type CheckConsumeContractParams struct {
 	T                    *testing.T
 	NumberOfTestElements int
 	// DataType to test for.
+	//
+	// Deprecated: [v0.110.0] Use Signal instead
+	// nolint
```
codeboten, line 39: same question here

otelcol/config.go
```diff
@@ -153,5 +154,41 @@ func (cfg *Config) Validate() error {
 			return fmt.Errorf("service::pipelines::%s: references exporter %q which is not configured", pipelineID, ref)
 		}
 	}
+
+	// Check that all pipelines reference only configured components.
+	for pipelineID, pipeline := range cfg.Service.PipelinesWithPipelineID {
```
codeboten, line 159: do we need a second for loop here?

exporter/exporterqueue/queue.go
TylerHelmuth: I'll move it to be safe

exporter/exportertest/contract_checker.go
TylerHelmuth: I'll move it to be safe

otelcol/config.go
TylerHelmuth: This is a duplicate of the `for pipelineID, pipeline := range cfg.Service.Pipelines {` loop above it.

I didn't want to handle the merging of `cfg.Service.Pipelines` into `cfg.Service.PipelinesWithPipelineID` via the Validate function (it is instead handled in `service.initGraph`) so I instead opted to validate both `cfg.Service.Pipelines` and `cfg.Service.PipelinesWithPipelineID`.

The line above, `if err := cfg.Service.Validate(); err != nil`, will validate that only one of the 2 is set, so in reality only 1 for the 2 for loops is ever entered.

codeboten (approved):
Thanks for addressing my comments

PR #11205, opened 2024-09-17
Deprecate service::telemetry::metrics::address

bogdandrutu:
Existing configs like:

```yaml
service:
  telemetry:
    metrics:
      level: "basic"
      address: "localhost:8888"
```

Should change to:

```yaml
service:
  telemetry:
    metrics:
      level: "basic"
      readers:
        - pull:
            exporter:
              prometheus:
                host: "localhost"
                port: 8888
```


service/service_test.go
```diff
@@ -711,8 +638,13 @@ func newNopConfigPipelineConfigs(pipelineCfgs pipelines.Config) Config {
 				InitialFields:     map[string]any(nil),
 			},
 			Metrics: telemetry.MetricsConfig{
-				Level:   configtelemetry.LevelBasic,
-				Address: "localhost:8888",
+				Level: configtelemetry.LevelBasic,
+				Readers: []config.MetricReader{{
+					Pull: &config.PullMetricReader{Exporter: config.MetricExporter{Prometheus: &config.Prometheus{
+						Host: newPtr("localhost"),
+						Port: newPtr(8080),
```
codeboten, line 645:
```suggestion
						Port: newPtr(8888),
```

service/telemetry/config.go
```diff
@@ -5,14 +5,27 @@ package telemetry // import "go.opentelemetry.io/collector/service/telemetry"
...
+	"go.opentelemetry.io/collector/featuregate"
 )
 
+var _ confmap.Unmarshaler = (*Config)(nil)
+
+var disableAddressFieldForInternalTelemetryFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"telemetry.disableAddressFieldForInternalTelemetry",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterFromVersion("v0.110.0"),
+	featuregate.WithRegisterToVersion("v0.110.0"),
```
codeboten, line 26: probably need a version much further into the future

service/telemetry/config.go
```diff
@@ -143,11 +156,47 @@ type TracesConfig struct {
 	Processors []config.SpanProcessor `mapstructure:"processors"`
 }
 
+func (c *Config) Unmarshal(conf *confmap.Conf) error {
+	if err := conf.Unmarshal(c); err != nil {
+		return err
+	}
+	if disableAddressFieldForInternalTelemetryFeatureGate.IsEnabled() {
```
codeboten, line 163: missing a `!` here?

service/telemetry/config.go
```diff
@@ -5,14 +5,27 @@ package telemetry // import "go.opentelemetry.io/collector/service/telemetry"
...
 	"go.uber.org/zap/zapcore"
 
 	"go.opentelemetry.io/collector/config/configtelemetry"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/featuregate"
 )
 
+var _ confmap.Unmarshaler = (*Config)(nil)
+
+var disableAddressFieldForInternalTelemetryFeatureGate = featuregate.GlobalRegistry().MustRegister(
```
codeboten, line 22: would be good to add a test to validate this gate's effects

service/telemetry/config.go
bogdandrutu: Done.

service/service_test.go
bogdandrutu: Done.

service/telemetry/config.go
bogdandrutu: Done.

service/telemetry/config.go
bogdandrutu: yeah.

bogdandrutu:
This is a much more verbose solution. Users would be wondering why are we doing it. Do we have an issue where we track the transition with the reasoning behind it?

https://github.com/open-telemetry/opentelemetry-collector/issues/7641

codeboten:
> This is a much more verbose solution. Users would be wondering why are we doing it. Do we have an issue where we track the transition with the reasoning behind it?

@dmitryax this was done in part to support the same configuration as is proposed by the opentelemetry-configuration schema, but also to support additional exporters, and more functionality like views in the future

service/telemetry/config.go
```diff
@@ -119,7 +132,7 @@ type MetricsConfig struct {
 	//  - "detailed" adds dimensions and views to the previous levels.
 	Level configtelemetry.Level `mapstructure:"level"`
 
-	// Address is the [address]:port that metrics exposition should be bound to.
+	// Deprecated: [v0.110.0] use readers configuration.
```
codeboten, line 135:
```suggestion
	// Deprecated: [v0.111.0] use readers configuration.
```

codeboten (approved):

Issue #10766, opened 2024-07-31
OTE-01-001 WP1: Usage of Multiple Vulnerable Dependencies 


mx-psi:
From the security audit we have a vulnerable dependency in `internal/tools`. This is very low priority to fix since it's only used in a dev environment, but I am noting it here for completeness:

> Improper neutralization of special elements used in an SQL Command (“SQL Injection”)
> Vulnerable Dependency:  github.com/jackc/pgx/v5@v5.4.3 => [update to v5.6.0]
> Introduced by: github.com/ryanrolds/sqlclosecheck@v0.5.1 (go1.20)
> Affected by:
> CVE-2024-27289

> [!NOTE]  
> [2024 OpenTelemetry security audit](https://opentelemetry.io/blog/2024/security-audit-results/) finding reference:
> **OTE-01-001 WP1**: Usage of Multiple Vulnerable Dependencies

mx-psi:
**Status update**: This has so far not been prioritized since it only affects the tooling used in CI and thus the potential impact for end-users is much lower.

PR #11214, opened 2024-09-19
[chore] update version to fix vulnerability

mowies:
#### Description
Fixes a minor vulnerability with a `replace` statement

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10766

### DIFF
--- a/internal/tools/go.mod
+++ b/internal/tools/go.mod
@@ -227,3 +227,5 @@ retract (
 	v0.57.1 // Release failed, use v0.57.2
 	v0.57.0 // Release failed, use v0.57.2
 )
+
+replace github.com/jackc/pgx/v5 v5.4.3 => github.com/jackc/pgx/v5 v5.6.0


internal/tools/go.mod
```diff
@@ -227,3 +227,5 @@ retract (
 	v0.57.1 // Release failed, use v0.57.2
 	v0.57.0 // Release failed, use v0.57.2
 )
+
+replace github.com/jackc/pgx/v5 v5.4.3 => github.com/jackc/pgx/v5 v5.6.0
```
mx-psi, line 231: Can we try with an indirect requirement?

mowies:
Ok i tried different things here, but the vulnerable dependency is not in the `go.sum` file to begin with. So there's no way of replacing it here.
@mx-psi I also tried with an indirect statement but `go mod tidy` will remove it again.

My conclusion is that this needs to be fixed upstream.

PR #11215, opened 2024-09-19
[chore] split exporterhelper so a separate profiles module can use it

dmathieu:
#### Description

This is part of #11131, splitting the exporterhelper module with a new `internal` submodule that holds everything shared by the signals.
This is so a new `exporterhelperprofiles` module can make use of the shared structs.


mx-psi (approved):

mx-psi:
This doesn't imply any changes in the public API so I am merging

PR #11217, opened 2024-09-19
[extension/zpages] Add an option to enable expvar

### LINKED ISSUE #11081: Add expvar handler to zpages extension

Go stdlib's expvar provides some useful information about Go runtime, and sometimes it is used by various libraries that could be imported by OTEL components to report their own internal state. But the HTTP handler is not exposed by default so there is no way to introspect that state of the libraries in the collector.

Since zpage extension is the agreed mechanism for introspection of the running collector, and already provides a webserver, we can register expvar handler on that server (also mentioned in https://github.com/open-telemetry/opentelemetry-collector/issues/2155#issuecomment-728167103).

The zpage extension config can have a flag that enables expvar, which could be off by default.

There could be a separate expvar extension just to create a new webserver and open a port, but it seems like an overkill and not aligned with zpages as the main introspection facility.

HongChenTW:
#### Description
Add an option to enable expvar, default is `disable`.

#### Link to tracking issue
Resolve #11081

#### Testing
Add a test case to verify the newly added option works.


HongChenTW:
Hi team, seems I mis-merged the others' commits and accidentally add many participants, should I open another PR to fix this issue?

johnmantios:
hey @HongChenTW, since no one is answering, you can try to move your branch's HEAD to the last commit you worked on and take it again from there

https://stackoverflow.com/a/36170904

HongChenTW:
Thanks @johnmantios 👍, I'll take a look into the post, actually you also remind me that I had a PR here.

extension/zpagesextension/config_test.go
```diff
@@ -37,5 +37,6 @@ func TestUnmarshalConfig(t *testing.T) {
 			ServerConfig: confighttp.ServerConfig{
 				Endpoint: "localhost:56888",
 			},
+			EnableExpvar: false,
```
yurishkuro, line 40: ineffectual, false is the default value fr `bool`s

extension/zpagesextension/README.md
```diff
@@ -33,7 +33,7 @@ extensions:
   zpages:
 ```
 
-The full list of settings exposed for this exporter are documented [here](./config.go)
+The full list of settings exposed for this extension are documented [here](./config.go)
```
yurishkuro, line 36: mention new endpoint in the README?

extension/zpagesextension/config_test.go
HongChenTW: Fixed in another PR https://github.com/open-telemetry/opentelemetry-collector/pull/11964

Issue #11221, opened 2024-09-19
Unit Test service/internal/proctelemetry/TestProcessTelemetryWithHostProc failing


jackgopack4:
**Describe the bug**
unit test TestProcess TelemetryWithHostProc is failing for metric name `otelcol_process_cpu_seconds`

**Steps to reproduce**
using linux host, run `go test` inside `service/internal/proctelemetry` folder

**What did you expect to see?**
expect tests to pass

**What did you see instead?**

```
--- FAIL: TestProcessTelemetryWithHostProc (0.21s)
    process_telemetry_linux_test.go:49: 
                Error Trace:    /home/vscode/otel/opentelemetry-collector/service/internal/proctelemetry/process_telemetry_linux_test.go:49
                Error:          "0" is not greater than "0"
                Test:           TestProcessTelemetryWithHostProc
                Messages:       otelcol_process_cpu_seconds
```

**What version did you use?**
upstream main repo

**What config did you use?**
n/a

**Environment**
microsoft go 1.23.1 devcontainer in vs code

**Additional context**
failing in pipeline here: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/10945728350/job/30390637590?pr=11208#step:6:634

It appears that the test is looking for the prefix "process_cpu_seconds" here: https://github.com/open-telemetry/opentelemetry-collector/blob/1edd31145fb83d321312cc12cb8a5bf33e329bde/service/internal/proctelemetry/process_telemetry_linux_test.go#L42
all the metrics that it is testing start with otelcol. Will submit a PR to fix this test.

PR #11208, opened 2024-09-18
Re-add version code to cmd/builder

jackgopack4:
Revert "[chore] delete code to set a version and date, as it it not used (#10715)"
Remove date string but re-add functions that check for version number
This (mostly) reverts commit b53f57d.

#### Description
This re-enables the functionality to set a version number for ocb binary (cmd/builder). It was erroneously removed. It also adds unit tests for this functionality, as a warning is given on CI with no code coverage.

#### Link to tracking issue
Relates to https://github.com/open-telemetry/opentelemetry-collector-releases/issues/664.
Closes #11220 along with https://github.com/open-telemetry/opentelemetry-collector-releases/pull/665

#### Testing
ran build and release cycles inside personal fork repositories (jackgopack4/opentelemetry-collector and jackgopack4/opentelemetry-collector-releases)

#### Documentation
.chloggen file

PR #11222, opened 2024-09-19
[chore] fix TestProcessTelemetryWithHostProc test

jackgopack4:
#### Description
Fixes failing unit test TestProcessTelemetryWithHostProc under service/internal/proctelemetry/process_telemetry_linux_test.go due to incorrect metric name 

#### Link to tracking issue
Fixes [#11221](https://github.com/open-telemetry/opentelemetry-collector/issues/11221)

#### Testing
unit tests

#### Documentation
none

### DIFF
--- a/service/internal/proctelemetry/process_telemetry_linux_test.go
+++ b/service/internal/proctelemetry/process_telemetry_linux_test.go
@@ -39,7 +39,7 @@ func TestProcessTelemetryWithHostProc(t *testing.T) {
 		} else {
 			metricValue = metric.Metric[0].GetGauge().GetValue()
 		}
-		if strings.HasPrefix(metricName, "process_uptime") || strings.HasPrefix(metricName, "process_cpu_seconds") {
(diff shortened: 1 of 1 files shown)


codeboten:
I'm curious why this test isn't failing here: https://github.com/open-telemetry/opentelemetry-collector/pull/11208

jackgopack4:
very odd. It seems from looking at it that if you want this condition to catch those specific parameters, we should add this change.
Is it possible that this test is flaky in its current state, because the cpu seconds may sometimes be more than zero for these metrics, but sometimes not?
[These](https://github.com/open-telemetry/opentelemetry-collector/blob/07c3e17b2045bd6ffc791578f4c2951335d6c585/service/internal/proctelemetry/process_telemetry_test.go#L34) are the parameters this test is testing, no?

jackgopack4:
seems like this test just missed getting updated when the metrics were changed to add an "otelcol_" prefix:

https://github.com/open-telemetry/opentelemetry-collector/commit/5753a58f712bf6f20f9daf2c660da1060cc1bebf

yurishkuro (approved):

codeboten (approved):

PR #11223, opened 2024-09-19
Add mdatagen replace option to check-contrib target

### LINKED ISSUE #11167: [CI] Contrib tests should also test updated mdatagen

PrateekKumar1709:
#### Description
This PR addresses the issue of contrib tests not testing the updated mdatagen version. It modifies the `check-contrib` and `restore-contrib` targets in the Makefile to ensure that the contrib tests use the local version of the `mdatagen` package.

#### Link to tracking issue
Fixes #11167 

#### Testing
All existing tests pass with `make test`


codeboten (approved):
Thanks @PrateekKumar1709

bogdandrutu (approved):

codeboten (approved):

Issue #9382, opened 2024-01-24
[config/confighttp] Resolve the TODO on confighttp.go related to prefixing operations with component id.


atoulme:
From [confighttp.go](https://github.com/open-telemetry/opentelemetry-collector/blob/d4bddc86da1bab117e6dce2dd76d3607fdc63513/config/confighttp/confighttp.go#L362-L373):
```
	// Enable OpenTelemetry observability plugin.
	// TODO: Consider to use component ID string as prefix for all the operations.
	handler = otelhttp.NewHandler(
		handler,
		"",
		otelhttp.WithTracerProvider(settings.TracerProvider),
		otelhttp.WithMeterProvider(settings.MeterProvider),
		otelhttp.WithPropagators(otel.GetTextMapPropagator()),
		otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
			return r.URL.Path
		}),
	)
```

I have tried to resolve this TODO once before with https://github.com/open-telemetry/opentelemetry-collector/pull/9123

enesonus:
Hi @atoulme I would like to work on this issue. Is there any recommended readings or some files to look at?

atoulme:
@enesonus go for it, not sure how to do it

VihasMakwana:
@atoulme can you assign this to me, I'll take it up.

mx-psi:
I think this can be removed from Collector v1, since we don't provide any stability guarantees for traces

jade-guiton-dd:
What is the status of this issue @atoulme @VihasMakwana? I believe the scope is a bit unclear. Do we want to:
1. make confighttp set the component ID as the span name prefix automatically; or,
2. let users of confighttp set an arbitrary prefix for their component?

The original TODO seems to suggest 1, but this seems to have been abandoned in favor of 2 because the component ID is not available in the confighttp code.

If 2 is fine, then I believe this issue was already resolved by #11769, which allows users to use `otelhttp.WIthSpanNameFormatter` themselves, in a way that will overwrite the default one we currently set, albeit at the cost of relying on a possibly unstable API.

atoulme:
I opened this issue to remove this TODO and get the component to stable state. The TODO predates my time with the project. It's been thorny to resolve as everybody ends up guessing what the comment means in the context of the time it was written at. I believe you neatly address it in your comment. I also believe 2. addresses the need.

Can we please remove this TODO and close this issue? Would you like to open a PR?

[...]

PR #11230, opened 2024-09-20
[confighttp]: add an option to add span formatter

VihasMakwana:
#### Description

This PR adds a new server option to modify span formatter. 
it also updates `otlpreceiver` to use the option.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/9382

#### Testing
Added.

---

`component.ID` is not available in this section of code. Since `ToServer` is intended to be general and can be used by `extensions`, `receivers`, and `exporters`, modifying it to accept `internal.Settings` would introduce a breaking change, which I'm not comfortable with. Therefore, I believe it would be better to provide an option to change the prefix instead.


VihasMakwana:
@bogdandrutu @open-telemetry/collector-approvers I'd appreciate your review on this!

config/confighttp/confighttp.go
```diff
@@ -462,15 +472,14 @@ func (hss *ServerConfig) ToServer(_ context.Context, host component.Host, settin
 	otelOpts := []otelhttp.Option{
 		otelhttp.WithTracerProvider(settings.TracerProvider),
 		otelhttp.WithPropagators(otel.GetTextMapPropagator()),
-		otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
-			return r.URL.Path
+		otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
+			return strings.TrimPrefix(fmt.Sprintf("%s:%s", operation, r.URL.Path), ":")
```
andrzej-stencel, line 476: If I understand correctly, this is a shortcut to remove the leading colon in case the prefix is not set. It has the side effect though that if the prefix starts with a colon, it will be erroneously removed. This is probably not very plausible, but I'd rather not have unwanted side effects, even if we think those are theoretical only.

```suggestion
			if len(operation) > 0 {
				return fmt.Sprintf("%s:%s", operation, r.URL.Path)
			}
			return r.URL.Path
```

andrzej-stencel (approved):
Other than the one nit, this looks sensible to me. Still definitely needs a look from @codeboten or @bogdandrutu who have more insight into this functionality.

config/confighttp/confighttp.go
VihasMakwana: This looks much neater! Thanks!

VihasMakwana:
@bogdandrutu @codeboten The CI looks good now. PTAL!

dmitryax (approved):

mx-psi (approved):
LGTM. @bogdandrutu since this is related to a TODO you added, would you mind double checking that this looks good?

VihasMakwana:
@bogdandrutu @open-telemetry/collector-maintainers can someone take a look here?

config/confighttp/confighttp.go
```diff
@@ -462,15 +471,17 @@ func (hss *ServerConfig) ToServer(_ context.Context, host component.Host, settin
...
+			}
 			return r.URL.Path
 		}),
 		otelhttp.WithMeterProvider(settings.LeveledMeterProvider(configtelemetry.LevelDetailed)),
 	}
 
 	// Enable OpenTelemetry observability plugin.
-	// TODO: Consider to use component ID string as prefix for all the operations.
-	handler = otelhttp.NewHandler(handler, "", otelOpts...)
+	handler = otelhttp.NewHandler(handler, serverOpts.spanPrefix, otelOpts...)
```
dmathieu, line 484: This parameter is not a span prefix. It's the operation name.
See https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#NewHandler

While for the default span name formatter, the operation is used as span prefix, it could be something different if someone uses the `WithSpanNameFormatter` option.

VihasMakwana: @dmathieu I understand your perspective, but we use `spanPrefix` specifically as a prefix for an operation in this context, using `otelhttp.WithSpanNameFormatter`
https://github.com/open-telemetry/opentelemetry-collector/blob/dffbad7c5ce1101a09c7bc5d6f4632bc53a8e176/config/confighttp/confighttp.go#L474-L479
That's why I named it `WithSpanPrefix`.

--- 
Using `WithOperationName` might suggest that the option is meant for naming the operation itself, which isn’t accurate. Instead, it serves as a prefix.


For eg.
For example, if you specify `WithSpanPrefix("foo")` and make an HTTP request to `http://host:port/path/bar`, the tracer would be named `foo:/path/bar`.

Does that help clarify?

config/confighttp/confighttp.go
```diff
@@ -410,6 +411,14 @@ func WithDecoder(key string, dec func(body io.ReadCloser) (io.ReadCloser, error)
 	})
 }
 
+// WithSpanPrefix creates a span prefixed with the specified name.
+// Ideally, this prefix should be the component's ID.
+func WithSpanPrefix(spanPrefix string) ToServerOption {
+	return toServerOptionFunc(func(opts *toServerOptions) {
+		opts.spanPrefix = spanPrefix
+	})
+}
```
sfc-gh-bdrutu, line 420: We should make it a bit more generic and accept `otelhttp.Option` or at least accept exactly same formatter? You may now want this, but later people may want other things, etc.

sfc-gh-bdrutu: The problem though, is that otelhttp is not stable, and we are looking to mark this stable soon. So I think we may not be able to accept this until `otelhttp` is marked stable.

mx-psi: @dmathieu since you are around, any idea of whether `otelhttp` will be marked stable any time soon?

VihasMakwana: @sfc-gh-bdrutu @bogdandrutu can you take a look now? I've made it more generic.

dmathieu: > any idea of whether otelhttp will be marked stable any time soon?

That is not on our radar at the moment. However, we are committed to keeping the API stable, or providing deprecation warnings for at least two versions.

VihasMakwana: @sfc-gh-bdrutu @bogdandrutu can you take a look?

.chloggen/add-new-server-option.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
sfc-gh-bdrutu, line 25: api

config/confighttp/confighttp.go
```diff
@@ -556,6 +564,15 @@ func maxRequestBodySizeInterceptor(next http.Handler, maxRecvSize int64) http.Ha
 	})
 }
 
+func PrefixFormatter(prefix string) func(string, *http.Request) string {
```
sfc-gh-bdrutu, line 567: why public?

sfc-gh-bdrutu (changes requested):
I have the concern that we really on the "formater" definition from the httptrace which is not stable. The reason this is a concern compare to using that package, is because be are exposing this in our public API.

Also: The API surface (especially because things are stable now) should be minimal.

receiver/otlpreceiver/otlp.go
```diff
@@ -162,7 +162,10 @@ func (r *otlpReceiver) startHTTPServer(ctx context.Context, host component.Host)
 	}
 
 	var err error
-	if r.serverHTTP, err = r.cfg.HTTP.ToServer(ctx, host, r.settings.TelemetrySettings, httpMux, confighttp.WithErrorHandler(errorHandler)); err != nil {
+	if r.serverHTTP, err = r.cfg.HTTP.ToServer(ctx, host, r.settings.TelemetrySettings, httpMux,
+		confighttp.WithErrorHandler(errorHandler),
+		confighttp.WithSpanFormatter(confighttp.PrefixFormatter(r.settings.ID.String())),
```
TylerHelmuth, line 167: Is this a breaking change? If this is merged will the spans that the otlpreceiver creates about itself have new names?

config/confighttp/confighttp.go
mx-psi: A possible way around this is to move the option to a separate module. We are discussing this in #11769

config/confighttp/confighttp.go
```diff
@@ -412,11 +413,21 @@ func WithDecoder(key string, dec func(body io.ReadCloser) (io.ReadCloser, error)
 	})
 }
 
+// WithSpanFormatter specifies which formater to use for span.
+// Ideally, this prefix in span name should be the component's ID.
+func WithSpanFormatter(formater func(string, *http.Request) string) ToServerOption {
+	return toServerOptionFunc(func(opts *toServerOptions) {
+		opts.formater = formater
```
flc1125, line 420:
```go
if formater != nil {
	// ....
}
```

receiver/otlpreceiver/otlp.go
mx-psi, line 167: If we want to do this, we can use #11769 now :)

mx-psi: @TylerHelmuth Is your concern that we should not do this if it is a breaking change?

PR #11236, opened 2024-09-20
[chore]: remove testifylint from Makefile

mmorel-35:
#### Description

All Testifylint rules are now fixed and ensured to be applied with golangci-lint. 
This removes the helper configuration in Makefile and tools module.


codeboten (approved):
🎉 thanks @mmorel-35!

bogdandrutu (approved):

PR #11237, opened 2024-09-20
[chore] Remove usage of experimental golang packages

bogdandrutu:


bogdandrutu:
Need to wait for golang 1.24 so we can drop support for 1.22

Issue #10825, opened 2024-08-07
[exporterhelper] Add an ability to batch by a key from pdata or context


dmitryax:
Introduce a new `exporterhelper.BatcherOption` with the following signature:

```golang
func WithBatchIdentifier(f IdentifyBatchFunc, batchIdentifiersLimit int) BatcherOption`
```

where 

```golang
// IdentifyBatchFunc returns an identifier for a batch. This function can be used to separate particular
// Requests into different batches. Batches with different identifiers will not be merged together.
// Provided context can be used to extract information from the context and use it as a part of the identifier as well.
type IdentifyBatchFunc func(ctx context.Context, r Request) string
```

`batchIdentifiersLimit` defines a maximum number of allowed identifiers. Requests producing batch identifiers that exceed this limit will be dropped. If this value is zero, then there is no limit on the number of identifiers.

The naming can be changed. The idea is that we need to provide exporters with a way to produce different batches based on some custom string key taken from the `Request` of context. This API would allow us to replicate the `metadata_keys` option in the batch processor, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md#batching-and-client-metadata.

sfc-gh-sili:
I can take a look at this issue. @dmitryax Would you mind assigning to me?

jmacd:
@bogdandrutu FYI

jmacd:
As the author of the batchprocessor metadata-keys feature, I would report that we eventually found it necessary to add metadata-key support in the OTel-Arrow streaming exporter itself. This is because if you want to use metadata as authorization at the stream level (e.g., so a load balancer can see it), then individual export streams have to be pinned to the metadata values. I am not sure if this will be possible in the exporterhelper, but thought I should mention it.

See this: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/otelarrowexporter/README.md#matching-metadata-per-stream

sfc-gh-sili:
Checked in with @jmacd offline to have a better understanding of their use case:

OpenTelemetry Arrow exporter and receiver supports transport using the Apache Arrow data format. It is ideal for case where network is expensive. Otel-Arrow exporter converts OTLP data into Arrow format and send them as multiple batches in a **stream**, whereas Otel-Arrow receiver contains logic to reverse the process. See the graph below:

[...]

Issue #9591, opened 2024-02-15
Applying memory_limiter extension


dmitryax:
### Problem

As part of https://github.com/open-telemetry/opentelemetry-collector/issues/8632, we want receivers to be able to reject incoming requests if the collector is running out of memory limits instead of accepting the data transforming to OTLP and get it rejected by the `memory_limiter` processor. Scraping receivers would skip the scraping in that situation.

We introduced a memory_limiter extension in https://github.com/open-telemetry/opentelemetry-collector/pull/8964, but is not been applied yet.

https://github.com/open-telemetry/opentelemetry-collector/pull/9397 proposes a way to configure a `memory_limiter` extension explicitly by receivers using httpconfig. Following this approach would complicate the user configuration interface as they would have to explicitly connect every receiver with a memoery_limiter extension to ensure that it's applied to all of them.

### Proposed solution

We can introduce an option to configure the `memory_limiter` extension in a way that it's applied to every receiver. Particular receivers can override the default memory limiter and connect to a different one.

Configuration example:

```yaml
extensions:
  memory_limiter:
    limit_mib: 800
    apply_to_all_receivers: true
  memory_limiter/internal_metrics:
    limit_mib: 1000

receivers:
  otlp:
    protocols:
      grpc:
        endpoint: localhost:4317
      http:
        endpoint: localhost:4318
  prometheus:
    memory_limiter: memory_limiter/internal_metrics
    config:
      scrape_configs:
        - job_name: 'otel-collector'
          scrape_interval: 10s
          static_configs:
            - targets: ["localhost:8888"]

processors:
  batch:

exporters:
  debug:
    verbosity: detailed

service:
  extensions: [memory_limiter, memory_limiter/internal_metrics]
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [debug]
    metrics:
      receivers: [otlp, prometheus]
      processors: [batch]
      exporters: [debug]
```

In this configuration, the default `memory_limiter` extension is implicitly applied to `otlp` receiver, and the additional `memory_limiter/internal_metrics` extension is applied to the prometheus receiver.

The user cannot define more than one `memory_limiter` extension with `apply_to_all_receivers: true`; the config validation will fail.

If `memory_limiter` extension with `apply_to_all_receivers: true` is defined, all the receivers used in the pipelines must support the `memory_limiter` extension. Otherwise, the collector won't start.

[...]

PR #11248, opened 2024-09-23
Concurrent batch processor features → batch processor

### LINKED ISSUE #8245: exporterhelper support for back-pressure

For an exporter that uses exporterhelper, I have the following requirements.

- When a data item is consumed by the exporter, the exporter should have an option to block until the exporter succeeds 
- When a data item is consumed by the exporter and the request fails, the exporter should return the error to the caller

These two requirements need to be met whether or not there are retries enabled, and whether or not there there is a queue, meaning whether or not `QueueSettings.Enabled` is true. The problem is that when `QueueSettings.Enabled` is true, which enables parallelism (i.e., NumConsumers > 1), then the caller will not get the error from the downstream receiver and the caller has no way to block until the receiver returns.

 Here is where today's exporterhelper w/ `QueueSettings.Enabled==false` will block until the call returns and return the caller's error: https://github.com/open-telemetry/opentelemetry-collector/blob/ebed8dd0943a98447d7f0e265cf58ec197bff93a/exporter/exporterhelper/queued_retry.go#L291

Here is where, when `QueueSettings.Enabled==true` and the queue is full, the code will not block and returns a generic error immediately, instead. https://github.com/open-telemetry/opentelemetry-collector/blob/ebed8dd0943a98447d7f0e265cf58ec197bff93a/exporter/exporterhelper/queued_retry.go#L312

Here is where, when `QueueSettings.Enabled==true` and the queue is not full, the code returns a `nil` without waiting to hear from the downstream receiver. https://github.com/open-telemetry/opentelemetry-collector/blob/ebed8dd0943a98447d7f0e265cf58ec197bff93a/exporter/exporterhelper/queued_retry.go#L316

As it stands, the only way for an exporter to block on the downstream service and receive its responses synchronously is to disable queueing. When Queueing is enabled, there are two problems (a) no option to block on queue-full, (b) no option to wait on response.

Optional behavior that enables blocking on a queue full and waiting for a response.

No alternatives come to mind.

**Additional context**
The [OTel-Arrow project](https://github.com/open-telemetry/otel-arrow) has a developed a bridging-mode for OpenTelemetry Collectors that enables high-compression on the network link between two collectors. For this setup to work reliably and under high load, we need the `otelarrow` exporter to be configured with a blocking pipeline. We would recommend both of the optional behaviors described in this issue to `otelarrow` users.

### LINKED ISSUE #11213: The reported data does not contain links
**Describe the bug**
When using OpenTelemetry-javaagent0.9.7 for trace propagation across applications, and you expect the reported data from the second application to contain links but find that they are missing, despite confirming that the parentSpanId of the first span in the second application matches the spanId of the last span in the first application, indicating that context propagation is working correctly, there could be several reasons why links are not present

**Steps to reproduce**
I have two applications.
1. I installed the OpenTelemetry Java Agent 0.97 on both of them and specified the service names. 
2. After restarting the applications, I deployed and started the Collector 0.101 to send data to the server. 
3. Then, I made a request to an interface that spans across both applications to enable cross-application tracing.

I expect the data reported by the agent on the second application to include links.

**What did you see instead?**
I can see that the first span reported by the second application has a parentSpanId that matches the last spanId of the first application, indicating that the context propagation is working correctly. However, I do not see any relevant links content in the reported data.

**What version did you use?**
opentelemetry-javaagent 0.97
collector 0.101：

**What config did you use?**
For OpenTelemetry Java Agent, I have configured the serviceName
for collector：
  receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
processors:
  batch:
    send_batch_size: 50
    send_batch_max_size: 300
    timeout: 5s     
exporters:
  otlp:
      endpoint: 10.70.50.36:12800
      tls:
        insecure: true
service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
    metrics:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlp]
  telemetry:
    logs:
      level: DEBUG 

**Environment**

**Additional context**

jmacd:
#### Description

Contributes downstream functionality back to the core component. 

- Enable unlimited concurrency
- Wait for context timeout or batch response from pipeline
- Trace the batch processor.
- Add `EarlyReturn` feature: enable legacy behavior (true)
- Add `MaxConcurrency` feature: enable legacy behavior (1)

TODO: Add feature gate. 
- The `EarlyReturn=true` legacy behavior should transition to false.
- The `MaxConcurrency=1` legacy behavior should transition to unlimited.

**Caveats** Despite a year+ long effort to add equivalent and satisfactory batching support in the `exporterhelper` subcomponent, we still today lack support for back-pressure with batching and reliable error transmission. I believe it is time to say "Yes/And" here. I support the efforts to improve `exporterhelper` and will contribute to that project myself, however I think repairing the legacy batch processor is also a good idea, given how long this has taken.

The changes here were developed downstream, see https://github.com/open-telemetry/otel-arrow/blob/main/collector/processor/concurrentbatchprocessor

#### Link to tracking issue

I'm listing a number of issues that are connected with this, both issues pointing to unmet needs in the exporterhelper batcher and missing features in the legacy batcher. Accepting these changes will allow significantly improved batching support in the interim period until the new batching support is complete.

Part of #10825 -- until this features is complete, users who depend on `metadata_keys` in the batch processor will not be able to upgrade
Part of #10368 -- I see this as the root cause, we haven't been able to introduce concurrency to the exporterhelper w/o also introducing a queue, which interferes with error transmission
Fixes #8245  -- my original report about the problem solved here -- we add concurrency with batching and error transmission and do not depend on a queue (persistent or in-memory)
Part of #9591 -- users must use one of the available memory limiting mechanisms in conjunction with the batch processor
Part of #8122 -- until this is finished, users depend on the original batch processor
Part of #7460 -- another statement of #8245; the batch processor does not propagate errors today, and this fixes the batch processor's contribution to the problem.

#### Testing

New tests are included.

#### Documentation

TODO/WIP


bogdandrutu:
Confused how you fix https://github.com/open-telemetry/opentelemetry-collector/issues/11213?

jmacd:
@bogdandrutu that was a mistake.

bogdandrutu (commented):
Hi Josh, this change is a bit too large and complicated to be reviewed in one PR. Can we split the 2 functionalities in 2 different PRs? 
1. The PR for multiple consumers.
2. The PR for sync.

PR #11254, opened 2024-09-23
[chore] Move back receiver definitions, make profile embed receiver

bogdandrutu:


bogdandrutu:
@dmathieu the current implementation is very bad because the public "soon to be stable" `receiver.Factory` has an unstable API, so current solution definitely is not working as expected. See https://pkg.go.dev/go.opentelemetry.io/collector/receiver#Factory the press the "internal.Factory" to read the definition.

dmathieu:
> Why?

In the service, the receiver builder holds a list of receivers:
https://github.com/open-telemetry/opentelemetry-collector/blob/4709346e0f34a84227c4648d65a02e61d3461262/service/internal/builders/receiver.go#L20-L23

But it also needs to be able to treat it as a profiles receiver, if we wish to create one:
https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/builders/receiver.go#L104

Hence the current implementation in main. The full (with profiles) implementation is hidden in internal. The one in the `receiver` packages only includes the stable methods.
Anything unstable is in `receiverprofiles`, but it's still the same implementation, so the service can use it.

Similarly, we need it in the OTLP receiver, to be able to create the receiver and have it handle profiles:
https://github.com/open-telemetry/opentelemetry-collector/blob/2be0dfb8e48776bafdfe93700a910b7b94c75b76/receiver/otlpreceiver/factory.go#L29-L38

Reverting this would have a much bigger impact that this PR, and would likely require reverting and rethinking most of the profiling work I've been doing those past 3 months.

This approach was decided in https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/experimental-profiling.md (https://github.com/open-telemetry/opentelemetry-collector/pull/10375)

bogdandrutu:
@dmathieu look at the PR and you see that it works. I am not sure where all these are coming from. You can see that the service can still create Profiles.

bogdandrutu:
> The full (with profiles) implementation is hidden in internal. The one in the receiver packages only includes the stable methods.

This is not true, go ahead and check that `f := receiver.NewFactory(...); f.CreateProfilesReceiver(...)` will compile

bogdandrutu:
> Reverting this would have a much bigger impact that this PR, and would likely require reverting and rethinking most of the profiling work I've been doing those past 3 months.

All you need to do is to create the factory using `receiverprofiles.NewFactory` when the component supports profiles, that is the only change required.

dmathieu:
You're right, sorry, it does work.
IMHO, there are still a couple problems with your approach. The main one being that folks now have to change the module they use for the receiver even for traces and metrics, as soon as they want to support traces, which isn't a great approach.

It's unlikely to happen, but with this approach, we also can't have 2 unstable signals at the same time (and use them together).
I've taken a stab at an alternative proposal, which I believe solves the stability problem, and is cleaner: #11270.

bogdandrutu:
> I've taken a stab at an alternative proposal, which I believe solves the stability problem, and is cleaner: https://github.com/open-telemetry/opentelemetry-collector/pull/11270.

I have 2 problems I am looking to solve:
* Stability, which indeed is solved by your proposal;
* GoDoc, which does not work nice when using type aliasing with internal. See https://pkg.go.dev/go.opentelemetry.io/collector/receiver#pkg-types, which causes lots of issues for the developers to read our documentation.

> It's unlikely to happen, but with this approach, we also can't have 2 unstable signals at the same time (and use them together).

We can, we just need to call that package "receiverexperimental" or something like that instead of the "receiverprofiles".

> The main one being that folks now have to change the module they use for the receiver even for traces and metrics, as soon as they want to support traces, which isn't a great approach.

I think this is a minimal problem in my opinion (we are just looking for things that are not ideal here), since anyway they have to use the "experimental" module anyway. I care way more about our "stable" developer experience than the "experimental" developer experience, if you need experimental it is ok to do more work.

bogdandrutu:
> receiver.NewFactory needs to be able to handle profiles.

This is another bad design, see https://www.hyrumslaw.com

bogdandrutu:
@open-telemetry/collector-approvers please review this, since @dmathieu work is not affected and this is a pure collector design I am sure we can move forward with this solution that has clear advantages vs existing or proposed alternatives:
* Does not expose unstable API;
* Does clean the godoc
* Does not have ensure profiles are always present see https://www.hyrumslaw.com/ for why that can be considered public API.

dmitryax (approved):
This solution looks good to me

PR #11268, opened 2024-09-24
[DRAFT] [NOT TO MERGE] add benchmark tests for traceprocessor

asreehari-splunk:
#### Link to tracking issue
Fixes #

### DIFF
--- a/processor/batchprocessor/batch_processor_test.go
+++ b/processor/batchprocessor/batch_processor_test.go
@@ -80,6 +80,91 @@ func TestProcessorLifecycle(t *testing.T) {
 	}
 }
 
+func BenchmarkBatchTraceProcessor(b *testing.B) {
+	b.StopTimer()
+	cfg := Config{
+		Timeout:                  100 * time.Millisecond,
+		SendBatchSize:            1000,
+		MetadataCardinalityLimit: defaultMetadataCardinalityLimit,
+	}
+	runTraceProcessorWoMetadataBenchmark(b, cfg)
+}
+
+func runTraceProcessorWoMetadataBenchmark(b *testing.B, cfg Config) {
+	ctx := context.Background()
+
+	sink := new(consumertest.TracesSink)
+	cfg.SendBatchSize = 128
+	creationSet := processortest.NewNopSettings()
+	creationSet.MetricsLevel = configtelemetry.LevelDetailed
+	batcher, err := newBatchTracesProcessor(creationSet, sink, &cfg)
+	require.NoError(b, err)
+	require.NoError(b, batcher.Start(context.Background(), componenttest.NewNopHost()))
+
+	b.StartTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			requestCount := 100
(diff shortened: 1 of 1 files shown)


processor/batchprocessor/batch_processor_test.go
```diff
@@ -80,6 +80,91 @@ func TestProcessorLifecycle(t *testing.T) {
...
 }
 
+func BenchmarkBatchTraceProcessor(b *testing.B) {
+	b.StopTimer()
+	cfg := Config{
+		Timeout:                  100 * time.Millisecond,
+		SendBatchSize:            1000,
+		MetadataCardinalityLimit: defaultMetadataCardinalityLimit,
+	}
+	runTraceProcessorWoMetadataBenchmark(b, cfg)
```
dmitryax, line 90: why do we call different functions in the benchmarks? the idea is to have only `MetadataKeys:             []string{"X-Sf-Token"},` line different between benchamarks

processor/batchprocessor/batch_processor_test.go
```diff
@@ -80,6 +80,91 @@ func TestProcessorLifecycle(t *testing.T) {
...
+	require.NoError(b, batcher.Start(context.Background(), componenttest.NewNopHost()))
+
+	b.StartTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			requestCount := 100
+			spansPerRequest := 10
+			sentResourceSpans := ptrace.NewTraces().ResourceSpans()
+			for requestNum := 0; requestNum < requestCount; requestNum++ {
+				td := testdata.GenerateTraces(spansPerRequest)
```
dmitryax, line 111: Please move `td` building outside. We don't need to benchmark it. We need to benchmark `batcher.ConsumeTraces(context.Background(), td))` only

dmitryax:
This is not intended for merging

processor/batchprocessor/batch_processor_test.go
asreehari-splunk: in the second case, I am putting the metadata in the context. i did not want to add a if else, so i separated it

dmitryax: We don't need to put metadata in the context. We are benchmarking if it's safe to add `MetadataKeys: "X-Sf-Token"` by default for non-passthrough (typical) workloads

processor/batchprocessor/batch_processor_test.go
```diff
@@ -80,6 +80,94 @@ func TestProcessorLifecycle(t *testing.T) {
...
+		MetadataCardinalityLimit: defaultMetadataCardinalityLimit,
+	}
+	runTraceProcessorWoMetadataBenchmark(b, cfg)
+}
+
+func runTraceProcessorWoMetadataBenchmark(b *testing.B, cfg Config) {
+	ctx := context.Background()
+
+	sink := new(consumertest.TracesSink)
+	cfg.SendBatchSize = 128
```
dmitryax, line 97: Please remove this line

PR #11270, opened 2024-09-25
Split receiver factory so the stable one isn't aware of profiles

dmathieu:
#### Description

This is an alternative solution for #11254.


bogdandrutu:
Please consider to mark this as draft, since I want to keep the discussion in the other PR. See https://github.com/open-telemetry/opentelemetry-collector/pull/11254#issuecomment-2374573369

PR #11273, opened 2024-09-25
[exporter/otlphttpexporter] Use NewDefaultClientConfig instead of manually creating struct

### LINKED ISSUE #9478: [confighttp] Review usage of 'pointer to type'

mackjmr:
#### Description
This PR makes usage of `NewDefaultClientConfig` instead of manually creating the `confighttp.ClientConfig` struct. The updates to defaults are maintained (Timeout, Compression, WriteBufferSize) whereas the fields that match the default are not manually set anymore (Endpoint, Headers).

It also sets to nil the fields that are set in default but weren't set previously (MaxIdleConns, MaxIdleConnsPerHost, MaxConnsPerHost, IdleConnTimeout) to retain the same behaviour. I am on the fence about this one, should we switch to using the defaults ?

#### Link to tracking issue
#11274

### DIFF
--- a/exporter/otlphttpexporter/factory.go
+++ b/exporter/otlphttpexporter/factory.go
@@ -38,6 +38,10 @@ func createDefaultConfig() component.Config {
 	clientConfig.Compression = configcompression.TypeGzip
 	// We almost read 0 bytes, so no need to tune ReadBufferSize.
 	clientConfig.WriteBufferSize = 512 * 1024
+	clientConfig.MaxIdleConns = nil
+	clientConfig.MaxIdleConnsPerHost = nil
+	clientConfig.MaxConnsPerHost = nil
+	clientConfig.IdleConnTimeout = nil
 
 	return &Config{
 		RetryConfig:  configretry.NewDefaultBackOffConfig(),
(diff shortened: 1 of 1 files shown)


songy23 (approved):

mackjmr:
@bogdandrutu This relates to the second point in: https://github.com/open-telemetry/opentelemetry-collector/issues/9478#issuecomment-2241891068. We are going to change the pointer fields in the ClientConfig (potentially through approach in #10260). 

The reasoning is that it will then be easier to update these fields once all instances are using NewDefaultClientConfig instead of creating the struct manually. I listed all places where the struct is created manually here: #11274.

That said, if we go with the optional.Optional approach, we will still need to update the fields that I unset manually [here](https://github.com/open-telemetry/opentelemetry-collector/blob/b032a6adc2e67feea802e41b7e121929ee4cd314/exporter/otlphttpexporter/factory.go#L41-L43). I added a question at the end of #11274 on how we want to treat fields that are set in the default but currently unset in components.

bogdandrutu (approved):

PR #11275, opened 2024-09-25
Unmarshal span and link flags from JSON

MrAlias:
For both the `Span` and `SpanLink`, the current unmarshalJsoniter methods do not unmarshal the `flags` field. This updates the functionality to correctly decode these values into the destination.

Fix #11267

Signed-off-by: Tyler Yahn <codingalias@gmail.com>

PR #11279, opened 2024-09-25
[chore] Fix unreleased bug that JSON integers should support string or int

bogdandrutu:
Bug added in https://github.com/open-telemetry/opentelemetry-collector/pull/11275

### DIFF
--- a/pdata/ptrace/json.go
+++ b/pdata/ptrace/json.go
@@ -113,7 +113,7 @@ func (dest Span) unmarshalJsoniter(iter *jsoniter.Iterator) {
 				iter.ReportError("readSpan.parentSpanId", fmt.Sprintf("parse parent_span_id:%v", err))
 			}
 		case "flags":
-			dest.orig.Flags = iter.ReadUint32()
+			dest.orig.Flags = json.ReadUint32(iter)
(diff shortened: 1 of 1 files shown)


pdata/ptrace/json.go
```diff
@@ -113,7 +113,7 @@ func (dest Span) unmarshalJsoniter(iter *jsoniter.Iterator) {
 				iter.ReportError("readSpan.parentSpanId", fmt.Sprintf("parse parent_span_id:%v", err))
 			}
 		case "flags":
-			dest.orig.Flags = iter.ReadUint32()
+			dest.orig.Flags = json.ReadUint32(iter)
```
songy23, line 116: Can we add a test for this?

bogdandrutu: We have unit-tests for the `json.ReadUint32`. Or for what?

bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/json/number_test.go#L67

songy23: OK that's good

Issue #11143, opened 2024-09-11
[exporterhelper] Move batching API to a separate module


mx-psi:
We can look into separating the batching API into a different module. This helps in the separation of concerns and potentially allows us to stabilize `exporterhelper` without stabilizing the batching API.

mx-psi:
The symbols that would be moved to a different module would be those that are marked with Experimental/early stage of development, that is:
- `BatcherOption`
- `WithRequestBatchFuncs`
- `WithBatcher`
- `WithRequestQueue`
- `Request`
- `RequestErrorHandler`
- `RequestFromLogsFunc`, `RequestFromMetricsFunc`, `RequestFromTracesFunc`

mx-psi:
From conversation in Collector stability meeting on 2025/09/08 we can retake this (the module seems stable enough to be able to do this)

Issue #11142, opened 2024-09-11
[exporterhelper] Review usage of API and internalize anything not currently used


mx-psi:
There are parts of `exporterhelper`, `exporterqueue` that are not currently used by any component. We can make these parts internal for now to reduce the amount of API that we stabilize with exporter 1.0. One such example is `exporterqueue.Config`.

mx-psi:
As an example here is a Github search that shows that [`exporterhelper.RequestErrorHandler`](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/exporterhelper#RequestErrorHandler) is not used anywhere: https://github.com/search?q=language%3AGo+-is%3Afork+%2Fexporterhelper%5C.RequestErrorHandler%2F&type=code

To complete this issue we need to:
1. Search on Github for the symbols to see if they are used or not
2. If they are not used, open a first PR to deprecate the symbol(s)
4. After the symbol(s) have been deprecated for one release, remove them

jade-guiton-dd:
After searching for all uses of public symbols from `exporterhelper` and `exporterqueue` outside of themselves on Github (including some cases where the package was renamed on import), it seems that the unused symbols are those introduced by #8122 and marked as "experimental".

This concerns all of `exporterqueue`, and the following from `exporterhelper`:
- Functions `New(Logs|Metrics|Traces)RequestExporter`, `WithRequestBatchFuncs`, `WithRequestQueue`
- Types `BatcherOption`, `Request`, `RequestErrorHandler`, `RequestFrom(Logs|Metrics|Traces)Func`

The only experimental symbol which *is* used is `exporterhelper.WithBatcher`. Note that it takes a vararg argument of type `BatcherOption`, but I found no use of it.

Before deprecating and removing these symbols, I think a discussion with whoever is working on 8122 would be necessary.

jade-guiton-dd:
No clear decision has been taken for what to do with the batching API, closing this for now.

PR #11285, opened 2024-09-27
[exporterhelper, exporterqueue] Deprecate unused public symbols

jade-guiton-dd:
#### Description

To help stabilize the `exporter` module faster, this PR deprecates some public symbols from `exporterhelper` and the entirety of `exporterqueue`, which don't seem to be used by anyone on Github, so that they might be fully internalized in a future release.

(Another option would be removing them entirely, if some aren't used internally. Yet another would be moving them to a new module as suggested by #11143, if it turns out some *are* currently in use in the ecosystem.)

All the symbols involved were introduced by #8122; of the symbols introduced by that PR, only `WithBatcher` is used in the ecosystem, and thus isn't deprecated by this PR. It has a variable argument of type `BatcherOption`, which *is* deprecated as it isn't used. A future release should remove this variable argument.

To avoid the linter complaining about internal use of deprecated symbols, `exporterqueue` functionality was split into an internal package with most of the functionality, and deprecated aliases in the old exporterqueue package. The new internal package is `exporter/internal/exporterqueue` instead of `exporter/exporterqueue/internal`, to allow its internal use by `exporter/exporterhelper`.

An issue arose with the aliasing process: `exporterqueue` exports multiple generic types, but generic aliases are only supported starting from Go 1.23. So, unless we wish to update the Go version soon, the only option was to declare a different defined type, and convert at runtime between the two in the alias function shims. There is some surprising behavior in exactly which equivalent types Go allows to convert between, which leads to some strange code in said shims.

#### Link to tracking issue
Updates #11142


mx-psi (commented):
Looks good, just one comment about deprecating the whole of `exporterqueue`

bogdandrutu:
I don't think we should "delete" everything that is not used. Please at least consider to wait until @dmitryax reviews and approves this.

jade-guiton-dd:
@dmitryax do you have an opinion on how we should proceed regarding the batching API?

PR #11286, opened 2024-09-27
[chore] Move back processor definitions, make profile embed processor

bogdandrutu:
Same as https://github.com/open-telemetry/opentelemetry-collector/pull/11254 but for processor


service/internal/builders/processor_test.go
```diff
@@ -24,12 +24,12 @@ func TestProcessorBuilder(t *testing.T) {
...
-		processor.NewFactory(
+		processorprofiles.NewFactory(
 			component.MustNewType("all"),
 			func() component.Config { return &defaultCfg },
-			processor.WithTraces(createProcessorTraces, component.StabilityLevelDevelopment),
-			processor.WithMetrics(createProcessorMetrics, component.StabilityLevelAlpha),
-			processor.WithLogs(createProcessorLogs, component.StabilityLevelDeprecated),
+			processorprofiles.WithTraces(createProcessorTraces, component.StabilityLevelDevelopment),
+			processorprofiles.WithMetrics(createProcessorMetrics, component.StabilityLevelAlpha),
+			processorprofiles.WithLogs(createProcessorLogs, component.StabilityLevelDeprecated),
```
dmitryax, line 32: I think we should rename `processorprofiles` to something like `expprocessor` without waiting for another new signal

dmitryax (approved):

service/internal/builders/processor_test.go
bogdandrutu: Can do that later, since is experimental.

PR #11287, opened 2024-09-27
Deprecate funcs that repeate receiver in name

bogdandrutu:
This change makes the names to be more "Go" friendly. Also it will not become an issue that types will try to implement different Factory types (receiver,processor) because that is forbidden because of our design. This also makes the names consistent with connector.Factory.


songy23 (approved):

djaglowski (approved):
Since we're reconsidering the names, I'm curious if `Create` is meaningfully different than `New` or if e.g. `factory.NewLogs` would be more idiomatic.

bogdandrutu:
> Since we're reconsidering the names, I'm curious if Create is meaningfully different than New or if e.g. factory.NewLogs would be more idiomatic.

That would mean I have to change connectors as well :(((... Happy to do it if others feel the same.

dmitryax:
`New` may be a better name for functions, but:
1. We would also need to rename all the `CreateLogsFunc` functions to `NewLogsFunc`, which is not that appealing to me...
2. AFAIK `New` is typically used as a standalone function, not as a factory method. This is what we've been doing as well.

So I would lean towards keeping `Create`.

dmitryax (approved):

PR #11289, opened 2024-09-27
Remove deprecated funcs form processorhelper.ObsReport

bogdandrutu:


codeboten (approved):
Thanks for taking this on, can you add a note about the removal of the `dropped` data point? it was always recorded as a 0 since nothing was increasing that counter, but it would be good to have changelog for it being removed.

PR #11292, opened 2024-09-27
[chore][confmap]: print a warning if invalid character is encountered in yaml

VihasMakwana:
#### Description

If you use the following (invalid) configuration and start the collector:
```
receivers:
  receiver:
	  field: val
```
you'll see an error like this:
```
Error: failed to get config: cannot resolve the configuration: retrieved value (type=string) cannot be used as a Conf
2024/09/28 03:11:06 collector server run finished with error: failed to get config: cannot resolve the configuration: retrieved value (type=string) cannot be used as a Conf
```

However, the error message isn’t helpful. The issue arises from an invalid character (a horizontal tab - ASCII code 9) in the YAML, which may not be visible in text editors. It would be beneficial to print a more informative error message when such problems are detected.

This is a side-effect of https://github.com/open-telemetry/opentelemetry-collector/pull/10794.

### DIFF
--- a/confmap/provider.go
+++ b/confmap/provider.go
@@ -143,6 +143,7 @@ func withStringRepresentation(stringRepresentation string) RetrievedOption {
 func NewRetrievedFromYAML(yamlBytes []byte, opts ...RetrievedOption) (*Retrieved, error) {
 	var rawConf any
 	if err := yaml.Unmarshal(yamlBytes, &rawConf); err != nil {
+		fmt.Printf("unmarshaling from yaml failed: %v\nAttempting to use it as a string verbatim. You can ignore this warning if the collector starts successfully.\n\n", err)
 		// If the string is not valid YAML, we try to use it verbatim as a string.
 		strRep := string(yamlBytes)
 		return NewRetrieved(strRep, append(opts, withStringRepresentation(strRep))...)


VihasMakwana:
After this change, the error message would look like:

```
unmarshaling from yaml failed: yaml: line 3: found character that cannot start any token
Attempting to use it as a string verbatim. You can ignore this warning if the collector starts successfully.

Error: failed to get config: cannot resolve the configuration: retrieved value (type=string) cannot be used as a Conf
2024/09/28 03:18:00 collector server run finished with error: failed to get config: cannot resolve the configuration: retrieved value (type=string) cannot be used as a Conf
```

confmap/provider.go
```diff
@@ -143,6 +143,7 @@ func withStringRepresentation(stringRepresentation string) RetrievedOption {
 func NewRetrievedFromYAML(yamlBytes []byte, opts ...RetrievedOption) (*Retrieved, error) {
 	var rawConf any
 	if err := yaml.Unmarshal(yamlBytes, &rawConf); err != nil {
+		fmt.Printf("unmarshaling from yaml failed: %v\nAttempting to use it as a string verbatim. You can ignore this warning if the collector starts successfully.\n\n", err)
```
dmitryax, line 146: I shouldn't use `fmt.Printf`. We need to print everything with configured loggers.

dmitryax:
The current message isn't ideal, but I don't think the additional helps significantly

confmap/provider.go
VihasMakwana: Hmm. I was confused about the same.
I followed the same step as `config.Validate(..)`.

It prints a message if `service::telemetry::metrics` validation fails.

Needs some additional work to involve loggers in this section but I get your point.

VihasMakwana:
> The current message isn't ideal, but I don't think the additional helps significantly

I agree to this one. I think in an ideal scenario, we should only have one error message out of two. Let me know what do you think.

VihasMakwana:
@dmitryax I think this whole thing boils down to using config as string.

If we encounter errors while decoding the otel yaml config itself, we should just throw an error straight away (dont use it verbatim as a string because there's no point of doing that)

If we successfully pass decoding step, then we should us it verbatim as a string while expanding.

What are your thoughts?

mx-psi:
I think the right approach here is to have some sort of wrapper that adds additional error information to the `Conf` object and to show this error information if parsing fails for some reason.

VihasMakwana:
I'll open a new PR with another approach

PR #11299, opened 2024-09-30
[exporter/otlphttpexporter] Remove unnecessary nil assignment in default client config

mackjmr:
#### Description
This is a follow up to #11273 in which I had set fields `MaxIdleConns`, `MaxIdleConnsPerHost`, `MaxConnsPerHost`, `IdleConnTimeout` to nil manually to keep backwards compatibility. 

However, in the call to [ToClient](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/confighttp/confighttp.go#L141-L166) the `http.Transport` defaults are used. The call to `NewDefaultClientConfig` also uses the `http.Transport` defaults.

Thus, not setting to nil will maintain the same behaviour/ is unnecessary.

#### Link to tracking issue
Fixes #

### DIFF
--- a/exporter/otlphttpexporter/factory.go
+++ b/exporter/otlphttpexporter/factory.go
@@ -38,10 +38,6 @@ func createDefaultConfig() component.Config {
 	clientConfig.Compression = configcompression.TypeGzip
 	// We almost read 0 bytes, so no need to tune ReadBufferSize.
 	clientConfig.WriteBufferSize = 512 * 1024
-	clientConfig.MaxIdleConns = nil
-	clientConfig.MaxIdleConnsPerHost = nil
-	clientConfig.MaxConnsPerHost = nil
-	clientConfig.IdleConnTimeout = nil
 
 	return &Config{
 		RetryConfig:  configretry.NewDefaultBackOffConfig(),
--- a/exporter/otlphttpexporter/factory_test.go
+++ b/exporter/otlphttpexporter/factory_test.go
@@ -50,7 +50,21 @@ func TestCreateMetricsExporter(t *testing.T) {
 	require.NotNil(t, oexp)
 }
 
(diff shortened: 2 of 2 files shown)


songy23 (approved):

exporter/otlphttpexporter/config_test.go
```diff
@@ -30,6 +30,11 @@ func TestUnmarshalDefaultConfig(t *testing.T) {
 	assert.Error(t, component.ValidateConfig(cfg))
 }
 
+var defaultMaxIdleConns = 100
+var defaultMaxIdleConnsPerHost = 0
+var defaultMaxConnsPerHost = 0
+var defaultIdleConnTimeout = 90 * time.Second
```
bogdandrutu, line 36: use defaults as they are. `http.DefaultTransport. IdleConnTimeout` and everywhere else

mackjmr: Updated 👍

exporter/otlphttpexporter/config_test.go
```diff
@@ -67,10 +69,14 @@ func TestUnmarshalConfig(t *testing.T) {
...
-				Timeout:         time.Second * 10,
-				Compression:     "gzip",
+				ReadBufferSize:      123,
+				WriteBufferSize:     345,
+				Timeout:             time.Second * 10,
+				Compression:         "gzip",
+				MaxIdleConns:        &defaultTransport.MaxIdleConns,
+				MaxIdleConnsPerHost: &defaultTransport.MaxIdleConnsPerHost,
+				MaxConnsPerHost:     &defaultTransport.MaxConnsPerHost,
+				IdleConnTimeout:     &defaultTransport.IdleConnTimeout,
```
bogdandrutu, line 79: This is not ok, because you may change the defaults now correct? Cannot use pointer to that, we should have `var defaultIdleConnTimeout = defaultTransport.IdleConnTimeout`.

mackjmr: youre right, updated in last commit 👍

PR #11301, opened 2024-09-30
[chore] Remove unnecessary views, already using the hint API for buckets

bogdandrutu:


service/telemetry/internal/otelinit/config.go
```diff
@@ -117,39 +116,21 @@ func InitPrometheusServer(registry *prometheus.Registry, address string, asyncEr
 }
 
 func batchViews(disableHighCardinality bool) []sdkmetric.View {
```
codeboten, line 118: nit: this func could be renamed now that it's not doing anything w/ batch views

codeboten (approved):
thanks for cleaning this up, just one non-blocking comment

Issue #11308, opened 2024-09-30
Batch processor concurrency and error transmission


jmacd:
**Description**
The OTel-Arrow project forked the batch processor component and extended it with support for concurrency and error transmission. We have been satisfied with this approach and wish to contribute the whole set of features back to the core repository and use feature-flags to transition away from the legacy behaviors.

The functional changes include:

- Tracing support: new root spans are linked with incoming requests; incoming requests are linked without outgoing exports
- Concurrent export behavior
- Synchronous error transmission

The existing component differs from our desired behavior in the second two of these points--it is limited to 1 concurrent export and does not transmit errors back. In this issue, we propose the use of feature-flags to transition from both of these legacy behaviors, which run counter to our performance and reliability goals.

The proposed configuration changes:
- early_return (current default: true; eventual default: false)
- max_concurrency (current default: 1; eventual default: unlimited)

The changes here were developed downstream, see https://github.com/open-telemetry/otel-arrow/blob/main/collector/processor/concurrentbatchprocessor

Link to tracking issue
Here are a number of issues that are connected with this one, both issues pointing to unmet needs in the exporterhelper batcher and missing features in the legacy batcher. Accepting these changes will allow significantly improved batching support in the interim period until the new batching support is complete.

Part of https://github.com/open-telemetry/opentelemetry-collector/issues/10825 -- until this features is complete, users who depend on metadata_keys in the batch processor will not be able to upgrade
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/10368 -- I see this as the root cause, we haven't been able to introduce concurrency to the exporterhelper w/o also introducing a queue, which interferes with error transmission
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/8245 -- my original report about the problem solved here -- we add concurrency with batching and error transmission and do not depend on a queue (persistent or in-memory)
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/9591 -- users must use one of the available memory limiting mechanisms in conjunction with the batch processor
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/8122 -- until this is finished, users depend on the original batch processor
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/7460 -- another statement of https://github.com/open-telemetry/opentelemetry-collector/issues/8245; the batch processor does not propagate errors today, and this fixes the batch processor's contribution to the problem.

bogdandrutu:
[...]

PR #11309, opened 2024-09-30
Refactor batch processor export method

jmacd:
#### Description

Split the internal `batcher.export()` interface into three methods. This is a refactoring that was applied in https://github.com/open-telemetry/otel-arrow/tree/main/collector/processor/concurrentbatchprocessor and is being back-ported as part of #11308. The reason this refactoring is needed is that the parent context of the export() request will be manipulated in common code (vs signal-specific code) for tracing support.

#### Link to tracking issue
Part of #11308

#### Testing
Existing tests cover this.

### DIFF
--- a/processor/batchprocessor/batch_processor.go
+++ b/processor/batchprocessor/batch_processor.go
@@ -96,13 +96,19 @@ type shard struct {
 // batch is an interface generalizing the individual signal types.
 type batch interface {
 	// export the current batch
-	export(ctx context.Context, sendBatchMaxSize int, returnBytes bool) (sentBatchSize int, sentBatchBytes int, err error)
+	export(ctx context.Context, req any) error
+
+	// splitBatch returns a full request built from pending items.
+	splitBatch(ctx context.Context, sendBatchMaxSize int) (sentBatchSize int, req any)
 
 	// itemCount returns the size of the current batch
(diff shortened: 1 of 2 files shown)


processor/batchprocessor/batch_processor.go
```diff
@@ -258,10 +264,16 @@ func (b *shard) resetTimer() {
...
+	err := b.batch.export(b.exportCtx, req)
 	if err != nil {
 		b.processor.logger.Warn("Sender failed", zap.Error(err))
 	} else {
+		var bytes int
+		if b.processor.telemetry.detailed {
+			bytes = b.batch.sizeBytes(req)
+		}
 		b.processor.telemetry.record(trigger, int64(sent), int64(bytes))
 	}
```
bogdandrutu, line 278:
```suggestion
	if err != nil {
		b.processor.logger.Warn("Sender failed", zap.Error(err))
		return
	}
	var bytes int
	if b.processor.telemetry.detailed {
		bytes = b.batch.sizeBytes(req)
	}
	b.processor.telemetry.record(trigger, int64(sent), int64(bytes))
```

bogdandrutu (approved):

bogdandrutu (approved):

Issue #11179, opened 2024-09-16
Attributes for component instancing


djaglowski:
**Is your feature request related to a problem? Please describe.**

[Component instancing is complicated.](https://github.com/open-telemetry/opentelemetry-collector/issues/10534) Since it seems unlikely that we will change this prior to 1.0, I want to raise the broader discussion of how the collector's own telemetry should ascribe telemetry to component instances.

First, why should we ascribe telemetry to component instances rather than to component configurations? As an observability tool it is critical that we are a good exemplar of an observable system. If we cannot accurately represent the internal state of the collector, I believe we will have failed to accomplish this goal.

In practice, what is most important is that instances derived from the same configuration may have very different states.

**Describe the solution you'd like**

Many components utilize global state to override the default instancing behavior (i.e. sharedcomponent or similar), but to keep things simple, I will first propose an attribution schema which ignores this problem.

### Ignoring shared instances

The following schema provides sufficient information to ascribe telemetry to specific instances. Since each kind of pipeline component (receivers, processors, exporters, connectors) has its own instancing rules, there are separate sets of attributes proposed for each kind.

#### Receivers

- `receiver`: The component ID
- `otel.signal`: `logs`, `metrics`, or `traces`

#### Processors

- `processor`: The component ID
- `otel.signal`: `logs`, `metrics`, or `traces`
- `pipeline`: The pipeline ID

Note: `otel.signal` is redundant because `pipeline` contains the same info. Is it worth keeping for consistency with receivers and exporters, even if it is not consistent with connectors?

#### Exporters

- `exporter`: The component ID
- `otel.signal`: `logs`, `metrics`, or `traces`

#### Connectors

Note: `otel.signal` as used in the other component kinds is insufficient even for metrics which describe ingoing or outgoing telemetry. This is because multiple instances of a connector may have the same type of incoming or outgoing telemetry. Moreover, the set of attributes should be sufficient for all telemetry describing the instance. This leaves two options:

1. Separate incoming and outgoing attributes:
- `connector`: The component ID
- `otel.incoming_signal`:  `logs`, `metrics`, or `traces`
- `otel.outgoing_signal`:  `logs`, `metrics`, or `traces` 

2. `otel.signal` with special values:
- `connector`: The component ID
- `otel.signal`: `logs->logs`, `logs->metrics`, `logs->traces`, `metrics->logs`, `metrics->metrics`, etc

### Acknowledging shared instances

[...]

PR #11147, opened 2024-09-11
[processorhelper] Add throughput metrics

djaglowski:
This follows up https://github.com/open-telemetry/opentelemetry-collector/pull/10910.

I expect to revamp this if https://github.com/open-telemetry/opentelemetry-collector/pull/11144 is merged, but I wanted to put it out in order to get feedback (cc @codeboten) on whether this is currently the way to implement detailed metrics.

It also still requires tests, but I'm unclear if the generated test code is intended to allow testing of detailed metrics yet.

PR #11311, opened 2024-09-30
WIP - Demonstrate graph-based metrics

djaglowski:
This is very incomplete but I want to show the basic idea of adding incoming/outgoing metrics at the graph level.

Obviously this code is pretty gnarly to begin with, and this only makes it worse, but this is superior to other approaches because:
1. All attributes described in #11179 are already available exactly where they are needed.
2. Unlike #11171, changes are not needed elsewhere.
3. Unlike #10910, these metrics will apply to all processors, not just those built with processorhelper. This benefit translates to all component kinds too though.
4. Additional metrics, such as proposed in #11147 can be added quite easily.
5. For processors and connectors, it's likely we could move span capture to this level.
6. Future functionality such as proposed in #10963 may be inserted in these consumer wrappers as well.


mx-psi (commented):
I think this is the right approach to do this; it gives us more flexibility in terms of the API and we have the visibility we need at this level. I support this.

cc @open-telemetry/collector-maintainers PTAL so we can discuss the high-level approach before moving forward

PR #11321, opened 2024-10-01
[chore][graph] Separate node types

djaglowski:
Having spent some time on #11311, I think it may be time to start refactoring this codebase into a more maintainable state. This PR just moves the various types of nodes into separate files, which I think is a bit more readable when considering changes.


bogdandrutu:
@djaglowski 100% agree. Do I remember asking you do to this?

bogdandrutu (approved):

PR #11326, opened 2024-10-01
Fix false fuzzer crash

AdamKorcz:
#### Description

The fuzzer report a null-pointer dereference which is not true.  This fixes that.

#### Link to tracking issue
Fixes https://issues.oss-fuzz.com/issues/42538213

### DIFF
--- a/receiver/otlpreceiver/fuzz_test.go
+++ b/receiver/otlpreceiver/fuzz_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/consumer/consumertest"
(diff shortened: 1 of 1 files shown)


songy23 (approved):

codeboten (commented):
I can't see any details of the report, can you add information here?
Screenshot 2024-10-01 at 11 29 25 AM

AdamKorcz:
> I can't see any details of the report, can you add information here? Screenshot 2024-10-01 at 11 29 25 AM

The fuzzer runs into  nil-dereference panic here because `r.nextConsumer` is `nil`:

https://github.com/open-telemetry/opentelemetry-collector/blob/e6ef444d9774a400fbb01c8ed29bb9aa7f2ef9f9/receiver/otlpreceiver/internal/metrics/otlp.go#L41

codeboten (approved):

PR #11331, opened 2024-10-01
(exporterhelper) Retry sender: fail request if context timeout < next retry

jmacd:
#### Description
The retry sender will delay until the context is canceled, where instead it could fail fast with a transient error and a clear message that no more retries are possible given the configuration.

#### Link to tracking issue
Part of #11183 

#### Testing
One new test.


bogdandrutu (approved):

newly12:
I think apart from comparing deadline from context with backoff delay, we should also compare max elapsed time with it? and one problem with context deadline is that it is not set properly in the collector, f.e. timeout sender is the next sender of retry sender, but it sets its own context with deadline, not the same context retry sending uses.

PR #11333, opened 2024-10-02
[chore][graph] Remove connectorNode's separate baseConsumer

djaglowski:
Follows #11330

Currently, `connectorNode` contains separate `component.Component` and `baseConsumer` fields. These fields are essentially two representations of the same component, but `baseConsumer` may be wrapped in another consumer that inherits capabilities. Rather than maintain two separate handles, this PR switches to a unified field. I believe this change helps normalize the connector node with other types of consumer nodes and will enable further refactoring opportunities.

### DIFF
--- a/service/internal/graph/connector.go
+++ b/service/internal/graph/connector.go
@@ -30,7 +30,6 @@ type connectorNode struct {
 	exprPipelineType pipeline.Signal
 	rcvrPipelineType pipeline.Signal
 	component.Component
-	baseConsumer
 }
 
 func newConnectorNode(exprPipelineType, rcvrPipelineType pipeline.Signal, connID component.ID) *connectorNode {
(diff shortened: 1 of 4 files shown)


service/internal/graph/connector.go
```diff
@@ -80,40 +79,28 @@ func (n *connectorNode) buildTraces(
 	}
 	next := connector.NewTracesRouter(consumers)
 
-	switch n.exprPipelineType {
-	case pipeline.SignalTraces:
+	if n.exprPipelineType == pipeline.SignalTraces {
```
bogdandrutu, line 82: Can we include this in the switch bellow, will simplify reading.

Same everywhere.

bogdandrutu (approved):

service/internal/graph/connector.go
djaglowski: Done

bogdandrutu (approved):

PR #11334, opened 2024-10-02
Go1.23 timers

### LINKED ISSUE #11332: Batch processor can deadlock/freeze using Go 1.23 timers
**Describe the bug**

When compiled with Go 1.23, the batch processor is able to get into the following stuck state, waiting on receiving a stopped timer chan that never sends.

This is the goroutine that is forever waiting for the `timerCh` to send after it's stopped.
```
goroutine 65 [chan receive, 19 minutes]:
    go.opentelemetry.io/collector/processor/batchprocessor.(*shard).stopTimer(0x40008c4900)
    go.opentelemetry.io/collector/processor/batchprocessor@v0.109.0/batch_processor.go:250 +0x4c
go.opentelemetry.io/collector/processor/batchprocessor.(*shard).processItem(0x40008c4900, {0x45919a0?, 0x43956371c0?})
    go.opentelemetry.io/collector/processor/batchprocessor@v0.109.0/batch_processor.go:239 +0xcc
go.opentelemetry.io/collector/processor/batchprocessor.(*shard).startLoop(0x40008c4900)
    go.opentelemetry.io/collector/processor/batchprocessor@v0.109.0/batch_processor.go:220 +0x138
created by go.opentelemetry.io/collector/processor/batchprocessor.(*shard).start in goroutine 1
    go.opentelemetry.io/collector/processor/batchprocessor@v0.109.0/batch_processor.go:184 +0x74
```
Given that our batch processor is synchronised on a single shard/goroutine, this causes the entire collector pipeline to deadlock.

We end up with thousands of goroutines that are forever waiting on a chan receive from the shard:
```
goroutine 169 [chan send, 19 minutes]:
go.opentelemetry.io/collector/processor/batchprocessor.(*singleShardBatcher).consume(0x45919a0?, {0x41810edb60?, 0x41cad43b00?}, {0x45919a0, 0x4569c487d0})
    go.opentelemetry.io/collector/processor/batchprocessor@v0.109.0/batch_processor.go:276 +0x30
go.opentelemetry.io/collector/processor/batchprocessor.(*batchProcessor).ConsumeTraces(0x4000e12580, {0x50165e8, 0x788df80}, {0x460a8569f0, 0x41d82d52a0})
    go.opentelemetry.io/collector/processor/batchprocessor@v0.109.0/batch_processor.go:346 +0x6c
<rest of stack omitted>
```

### Temporary fix

**We were able to fix it by disabling Go 1.23 timers**. Looking into the nature of this bug, and also seeing that Go 1.23 had changes to timer channels, I grew suspicious that it was something related to Go 1.23 timers. Since setting `GODEBUG="asynctimerchan=1"`, the issue has not come back in several days (it was happening several times a day before). That leads me to believe that it has fixed the issue, but I have not determined the exact cause.

I also noticed that collector upstream is still compiled on Go 1.22, so this issue wouldn't be coming up yet, and it would only come up for people compiling their own collectors. So, this ticket can serve as a future warning, particularly as it's unlikely to be caught in any tests.

**Steps to reproduce**

This is very difficult to reproduce, we have hundreds of pods running the collector and this only happens to one of the pods every so often.

Collector continuing to run normally.

**What did you see instead?**

Collector deadlocks, it accepts new requests but fails to process them. This results in a large allocation of stack memory because many goroutines are created by the OTLP receiver, as well as heap growing until the process get OOM killed.

**What version did you use?**
OpenTelemetry Collector v0.109.0, compiled manually.
I believe this was also happening in 0.110.0, but the stacks I have are 0.109.0. 

**What config did you use?**

```yaml
  batch:
    send_batch_size: 4096
    send_batch_max_size: 10_000
```

**Environment**
Manually compiled with our own internal distribution of the collector. 
Golang 1.23.0.
arm64
running on a `ubuntu:jammy` container.

shivanthzen:
#### Link to tracking issue
Fixes #11332


processor/batchprocessor/batch_processor_1_23.go
```diff
@@ -0,0 +1,12 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build 1.23
+
+package batchprocessor // import "go.opentelemetry.io/collector/processor/batchprocessor"
+
+func (b *shard) stopTimer() {
+	if b.hasTimer() {
```
codeboten, line 9: Is it possible to add a test to capture the regression this is fixing?

codeboten (commented):
Thanks @shivanthzen, one question

processor/batchprocessor/batch_processor.go
```diff
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+
```
codeboten, line 4:
```suggestion
```

processor/batchprocessor/batch_processor_1_22.go
```diff
@@ -0,0 +1,12 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !1.23
```
mx-psi, line 4: Shouldn't this be `go1.23`? At least on https://stackoverflow.com/a/38439941 that is the format used. Has the format changed?

shivanthzen: For builds using  go version 1.23 and above , file tagged with `//go:build 1.23` will be used, for go version below 1.23 , file tagged with `//go:build !1.23` will be used.

mx-psi: Right, what I meant is that I expected this to be
```suggestion
//go:build !go1.23
```

or at least this is how it used to work for Go in older versions, not `1.23`

shivanthzen:
I suggest this MR closed without merging in favour of https://github.com/golang/go/issues/69312.

codeboten:
It looks like the fix went into 1.23.2 https://github.com/golang/go/issues?q=milestone%3AGo1.23.2+label%3ACherryPickApproved, https://github.com/golang/go/issues/69333

Is the fix for us to ensure we use 1.23.2 when building the next release?

mx-psi:
Double checked by re-running a random test in the -releases repository for the last commit: 
https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/11152590145/job/31035238607

Looks like we are all set:
```
Setup go version spec ~1.23
Attempting to download ~1.23...
matching ~1.23...
Acquiring 1.23.2 from https://github.com/actions/go-versions/releases/download/1.23.2-11145[9](https://github.com/open-telemetry/opentelemetry-collector-releases/actions/runs/11152590145/job/31035238607#step:4:10)22912/go-1.23.2-linux-x64.tar.gz
```

codeboten:
Might be worth at least capturing the fix in a changelog if no code changes are required

shivanthzen:
Parent issue has been closed. Fix was implemented in golang

PR #11335, opened 2024-10-03
Fix an issue when the error is nil but is different than nil

bogdandrutu:


andrzej-stencel (approved):

receiver/receiverhelper/obsreport_test.go
```diff
@@ -296,10 +296,30 @@ func TestCheckReceiverLogsViews(t *testing.T) {
...
+	}
+	return p // Will always return a non-nil error.
+}
+
+func TestIsErrorNil(t *testing.T) {
+	assert.True(t, isErrorNil(nil))
+	assert.False(t, isErrorNil(returnsError(true)))
+	assert.False(t, returnsError(true) == nil)
+	assert.True(t, isErrorNil(returnsError(false)))
+	assert.False(t, returnsError(false) == nil)
```
mx-psi, line 316: Linter is complaining about these two lines, I guess the best way of putting it is this?
```suggestion
	assert.NotNil(t, returnsError(true))
	assert.True(t, isErrorNil(returnsError(false)))
	assert.NotNil(t, returnsError(false))
```

songy23 (commented):
Need to fix lint https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11157335012/job/31011433393?pr=11335

receiver/receiverhelper/obsreport_test.go
bogdandrutu: Unfortunately `assert.NotNil` (`assert.Nil`) does the right checks, so will not work

```golang
// isNil checks if a specified object is nil or not, without Failing.
func isNil(object interface{}) bool {
	if object == nil {
		return true
	}

	value := reflect.ValueOf(object)
	switch value.Kind() {
	case
		reflect.Chan, reflect.Func,
		reflect.Interface, reflect.Map,
		reflect.Ptr, reflect.Slice, reflect.UnsafePointer:

		return value.IsNil()
	}

	return false
}
```

mx-psi (approved):

songy23 (approved):

codeboten (commented):
Is this needed everywhere `.Error()` is called?

bogdandrutu:
I don't think this is scalable. I think we should probably do:
1. Add a check for concurrency to all components.
2. Productionize something like https://github.com/open-telemetry/opentelemetry-collector/pull/11346 and test at runtime that each component does not return bogus error.

jmacd (commented):
Looks good. It would be nice to know which piece of code is responsible for the nil-error object and try to fix it.

bogdandrutu:
>Looks good. It would be nice to know which piece of code is responsible for the nil-error object and try to fix it.

It was the race condition we fixed last night. https://github.com/open-telemetry/opentelemetry-collector/pull/11349

codeboten:
Is this PR still needed now that the race condition has been fixed?

bogdandrutu:
No. I don't think the PR is needed, maybe if we can get https://github.com/open-telemetry/opentelemetry-collector/pull/11335#issuecomment-2392846815 into an issue and decide what guard we want in the future.

codeboten:
Opened https://github.com/open-telemetry/opentelemetry-collector/issues/11407 to track adding concurrency tests

PR #11339, opened 2024-10-03
Use slices of pointers in pprofile

### LINKED ISSUE #11281: Inefficient use of sliceOfValues in profiles data
We should never use sliceOfValues for large objects (more than ~32B). See https://github.com/open-telemetry/opentelemetry-collector/pull/10195 where everything is marked as "nullable" which will make re-allocation which happens quite often during unmarshaling very expensive because copies lots of data.

One concrete example `MappingSlice` or `LocationSlice` should not be a value slice. But someone needs to re-evaluate all of them.

dmathieu:
#### Description

All these slices of values have a potential to store large amounts of data, and should therefore be slices of pointers.

#### Link to tracking issue
Fixes #11281

cc @mx-psi @bogdandrutu


dmathieu:
I ended up turning all of them from values to pointers, as each of them looks like it could be big (link may be the only grey one).
I'm happy to revert any of them if you think they aren't appropriate.

bogdandrutu (approved):

Issue #11343, opened 2024-10-03
Auto-instrumented pipeline components


djaglowski:
**Is your feature request related to a problem? Please describe.**

The collector's components should be observable. Intuitively, to some extent there must be some instrumentation which can be applied to components in a universal and uniform way.

Goals of this proposal:
1. Articulate a mechanism which enables us to _automatically_ capture signals from _all pipeline components_.
2. Define attributes that are (a) specific enough to describe individual component [_instances_](https://github.com/open-telemetry/opentelemetry-collector/issues/10534) and (b) consistent enough for correlation across signals.
3. Define specific metrics for each kind of pipeline component.
4. Define specific logs for each kind of pipeline component.
5. Define specific spans for processors and connectors.

**Describe the solution you'd like**

### Mechanism

The mechanism of telemetry capture should be _external_ to components. Specifically, we should inject instrumentation at each point where a component passes data to another component, and, at each point where a component consumes data from another component. In terms of the component graph, this means that every _edge_ in the graph will have two layers of instrumentation - one for the producing component and one for the consuming component. Generally these layers will act independently (e.g. record a metric about the data coming out of a receiver), but in the case of processors and connectors, their adjacent layers can act in concert (e.g. record the start and end of a span).

###

Attributes are discussed in detail in https://github.com/open-telemetry/opentelemetry-collector/issues/11179, though only in the context of metrics. To restate the consensus from that issue, all signals should use the following attributes:

#### Receivers

- `receiver`: The component ID
- `otel.signal`: `logs`, `metrics`, `traces`, **OR `ALL`**

#### Processors

- `processor`: The component ID
- `otel.signal`: `logs`, `metrics`, `traces`, **OR `ALL`**
- `pipeline`: The pipeline ID, **OR `ALL`**

#### Exporters

- `exporter`: The component ID
- `otel.signal`: `logs`, `metrics` `traces`, **OR `ALL`**

#### Connectors

- `connector`: The component ID
- `otel.signal`: `logs->logs`, `logs->metrics`, `logs->traces`, `metrics->logs`, `metrics->metrics`, etc, **OR `ALL`**

Notes: The use of `ALL` is based on the assumption that components are instanced either in the default way, or as a single instance per configuration (e.g. otlp receiver).

### Metrics

There are two straightforward measurements that can be made on any pdata:
1. A count of "items" (spans, data points, or log records). These are low cost but broadly useful, so they should be enabled by default.
[...]

PR #11344, opened 2024-10-03
[chore][graph] Remodel node id as attribute sets

djaglowski:
This PR formalizes the notion that attributes are the defining identifiers for component instances in the graph package. It moves us towards the proposal described in #11343.


service/internal/graph/attribute/attribute.go
```diff
@@ -0,0 +1,99 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package attribute // import "go.opentelemetry.io/collector/service/internal/graph/attribute"
+
+import (
+	"fmt"
+	"hash/fnv"
+
+	"go.opentelemetry.io/otel/attribute"
+
```
djaglowski, line 11:
```suggestion
```

service/internal/graph/graph_test.go
```diff
@@ -2068,11 +2068,11 @@ func TestGraphBuildErrors(t *testing.T) {
 				},
 			},
 			expected: `cycle detected: ` +
-				`connector "nop/conn1" (traces to traces) -> ` +
-				`processor "nop" in pipeline "traces/2" -> ` +
 				`connector "nop/conn" (traces to traces) -> ` +
 				`processor "nop" in pipeline "traces/1" -> ` +
-				`connector "nop/conn1" (traces to traces)`,
+				`connector "nop/conn1" (traces to traces) -> ` +
```
codeboten, line 2073: Why the change in test here?

codeboten (commented):
The change generally looks ok to me and aligns with what we've previously discussed around attributes. Probably needs to wait until the RFC for telemetry is merged before moving forward.

service/internal/graph/graph_test.go
djaglowski: It's really just a change in where the start of the cycle is reported, but it still reports the same cycle deterministically.

What it comes down to is that this PR changes the way node IDs are generated. Previously, we built a particualr string for each kind of component. Now we build the ID from the attribute set. Both methods are deterministic, but the graph presumably contains some logic where the node ids are considered in an order that is sensitive to their values (e.g. sort the ids, then for each one, follow the edges, etc). The logic graph that is produced is the same, but when an error is encountered, it is sensitive to the state it was in during the build process.

djaglowski:
Now that #11406 is merged, this should be ready for another review. Important note - this doesn't actually change anything meaningful for users yet. It's just a way of managing the nodes in the graph internally that is closer to what the RFC states should be presented to users. Further work will be needed before users see any standardization of attributes.

djaglowski:
Closing in favor of #12057

PR #11358, opened 2024-10-04
Fix for #11351

### LINKED ISSUE #11351: Processor In/Out metrics not recorded if process func fails (even for skipping next component)
See https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/processorhelper/metrics.go#L56

shivanthzen:
#### Link to tracking issue
Fixes #11351

### DIFF
--- a/processor/processorhelper/documentation.md
+++ b/processor/processorhelper/documentation.md
@@ -6,6 +6,14 @@
 
 The following telemetry is emitted by this component.
 
+### otelcol_processor_errors
+
+Number of errors emitted from the processor [alpha]
+
+| Unit | Metric Type | Value Type | Monotonic |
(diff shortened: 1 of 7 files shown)


djaglowski:
Thanks for working on this @shivanthzen. While I agree this fixes the issue, we need to be very deliberate about the metrics we add.

In this case, we have previously discussed adding an `outcome` attribute to the outgoing metric, which would serve nearly the same purpose as the new metric you've suggested.

However, I think we should make a quick fix where the outgoing value is reported as 0 if there is an error.

djaglowski:
I've opened https://github.com/open-telemetry/opentelemetry-collector/pull/11360 with the solution I described

shivanthzen:
Makes sense, thanks 👍

PR #11360, opened 2024-10-04
[processorhelper] Fix bug where record in/out metrics were skipped

### LINKED ISSUE #11351: Processor In/Out metrics not recorded if process func fails (even for skipping next component)

djaglowski:
Resolves #11351

### DIFF
--- a/.chloggen/processorhelper-in-out-fix.yaml
+++ b/.chloggen/processorhelper-in-out-fix.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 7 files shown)


djaglowski:
The contrib test failure appears to be expected with this change. The filter processor returns `ErrSkipProcessingData` if it filtered all data points, which happens in the test. Previously this caused a single data point to _not_ be counted as incoming.

bogdandrutu (approved):

bogdandrutu:
There are lint issues.

dmitryax:
@djaglowski do you have a fix for the tests on contib side by chance?

djaglowski:
I'll update core and make the fix.

PR #11366, opened 2024-10-04
[exporter/otlp] Report runtime status

mwear:
#### Description
This PR adds runtime status reporting for the otlp and otlphttp exporters. It's an updated version of #8788 which was one of several approaches experimented with to add this functionality. This work was paused to allow the [consumererror](https://github.com/open-telemetry/opentelemetry-collector/pull/9041) work to evolve, as it appeared as though there might be a way to uniformly apply the same logic to all exporters (via the exporterhelper), but that work has taken a different direction, and it looks like a uniform approach will not be possible.

This PR implements runtime status reporting as discussed in #9957. The choices for which statuses represent permanent errors are up for debate, but the key point is that a permanent error is an error discovered at runtime that will require user intervention to fix.

This implementation makes use of the finite state machine that underlies the status reporting system. The finite state machine will ensure that:
- Only changes in status are reported. Repeat reports of the same status will no-op.
- If a component transitions into a PermanentError all further status reports will no-op.

This means the exporter does not have to reason about current or previous statuses. It can report status based on its current view of the world and the status reporting system will handle the rest. Flapping between recoverable and ok is meant to be handled by watchers consuming status events. The healthcheckv2extension handles this by using a time based approach (e.g. recovery interval). Other watchers may choose to handle this situation differently.

For more information on status reporting, the state machine, etc see: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md

While all components report status during start and shutdown (via automation), this is the first component to report runtime status. This will allow the [healthcheckv2extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/healthcheckv2extension) to replace the currently non-functioning `check_collector_pipeline` capability of the original [healthcheckextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/healthcheckextension) and should serve as an example for other components that wish to report runtime status moving forward.

#### Link to tracking issue
Implements #9957 for the otlp and otlphttp exporters

#### Testing
units/manual

#### Documentation
code comments

### DIFF
--- a/exporter/otlpexporter/otlp.go
+++ b/exporter/otlpexporter/otlp.go
@@ -92,14 +92,11 @@ func (e *baseExporter) shutdown(context.Context) error {
 	return nil
 }
 
-func (e *baseExporter) pushTraces(ctx context.Context, td ptrace.Traces) (err error) {
-	defer func() {
-		e.reportStatusFromError(err)
-	}()
+func (e *baseExporter) pushTraces(ctx context.Context, td ptrace.Traces) error {
 	req := ptraceotlp.NewExportRequestFromTraces(td)
 	resp, respErr := e.traceExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
-	if err = processError(respErr); err != nil {
-		return
+	if err := processError(respErr); err != nil {
+		return err
 	}
 	partialSuccess := resp.PartialSuccess()
 	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedSpans() == 0) {
@@ -108,17 +105,23 @@ func (e *baseExporter) pushTraces(ctx context.Context, td ptrace.Traces) (err er
 			zap.Int64("dropped_spans", resp.PartialSuccess().RejectedSpans()),
 		)
 	}
-	return
+	return nil
 }
 
-func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) (err error) {
-	defer func() {
-		e.reportStatusFromError(err)
-	}()
+func (e *baseExporter) pushTracesWithStatus(ctx context.Context, td ptrace.Traces) error {
+	if err := e.pushTraces(ctx, td); err != nil {
+		componentstatus.ReportStatus(e.host, componentstatus.NewRecoverableErrorEvent(err))
+		return err
+	}
+	componentstatus.ReportStatus(e.host, componentstatus.NewEvent(componentstatus.StatusOK))
+	return nil
+}
+
+func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) error {
 	req := pmetricotlp.NewExportRequestFromMetrics(md)
 	resp, respErr := e.metricExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
-	if err = processError(respErr); err != nil {
-		return
+	if err := processError(respErr); err != nil {
+		return err
 	}
 	partialSuccess := resp.PartialSuccess()
 	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedDataPoints() == 0) {
@@ -127,17 +130,23 @@ func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) (err
 			zap.Int64("dropped_data_points", resp.PartialSuccess().RejectedDataPoints()),
 		)
 	}
-	return
+	return nil
 }
 
-func (e *baseExporter) pushLogs(ctx context.Context, ld plog.Logs) (err error) {
-	defer func() {
-		e.reportStatusFromError(err)
-	}()
+func (e *baseExporter) pushMetricsWithStatus(ctx context.Context, md pmetric.Metrics) error {
+	if err := e.pushMetrics(ctx, md); err != nil {
+		componentstatus.ReportStatus(e.host, componentstatus.NewRecoverableErrorEvent(err))
+		return err
+	}
+	componentstatus.ReportStatus(e.host, componentstatus.NewEvent(componentstatus.StatusOK))
+	return nil
+}
+
+func (e *baseExporter) pushLogs(ctx context.Context, ld plog.Logs) error {
 	req := plogotlp.NewExportRequestFromLogs(ld)
 	resp, respErr := e.logExporter.Export(e.enhanceContext(ctx), req, e.callOptions...)
-	if err = processError(respErr); err != nil {
-		return
+	if err := processError(respErr); err != nil {
+		return err
 	}
 	partialSuccess := resp.PartialSuccess()
 	if !(partialSuccess.ErrorMessage() == "" && partialSuccess.RejectedLogRecords() == 0) {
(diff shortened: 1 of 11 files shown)


exporter/otlpexporter/otlp.go
```diff
@@ -209,3 +233,21 @@ func getThrottleDuration(t *errdetails.RetryInfo) time.Duration {
...
+// A component status of PermanentError indicates the component is in a state that will require user
+// intervention to fix. Typically this is a misconfiguration detected at runtime. A component
+// PermanentError has different semantics than a consumererror. For more information, see:
+// https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md
+func isComponentPermanentError(st *status.Status) bool {
+	switch st.Code() {
+	case codes.NotFound:
+		return true
+	case codes.PermissionDenied:
+		return true
```
bogdandrutu, line 246: Why is this permanent? If auth token is passed with the request this does not mean exporter is not able to handle other requests.

exporter/otlpexporter/otlp.go
```diff
@@ -209,3 +233,21 @@ func getThrottleDuration(t *errdetails.RetryInfo) time.Duration {
...
 }
+
+// A component status of PermanentError indicates the component is in a state that will require user
+// intervention to fix. Typically this is a misconfiguration detected at runtime. A component
+// PermanentError has different semantics than a consumererror. For more information, see:
+// https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md
+func isComponentPermanentError(st *status.Status) bool {
+	switch st.Code() {
+	case codes.NotFound:
+		return true
```
bogdandrutu, line 244: This can easily be resolved correct? If the server is temporary down this should not mark the component as permanent error.

bogdandrutu (commented):
Before implementation, can we document somewhere (ideal in RFC) or in https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md when an exporter is consider permanent error? I feel like we have different understandings of this and I don't want you to make lots of code changes until we agree on the basic definition.

mwear:
> Before implementation, can we document somewhere (ideal in RFC) or in https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md when an exporter is consider permanent error?

I have these documented here: https://github.com/open-telemetry/opentelemetry-collector/issues/9957. Is it ok to discuss on that issue, or would you prefer it to be an RFC?

mwear:
If you're suggesting that the errors identified as Permanent in this PR (and #9957) are in many cases recoverable, we can identify all errors as recoverable and skip permanent for now.

bogdandrutu:
I am not the biggest expert, but I feel at least some are recoverable, and some are contextual (example auth when passing auth tokens from the client).

mwear:
I think going with recoverable for all errors, at least for now, makes sense. We can discuss what errors should be permanent on #9957, but I suspect we'll be able to come up with counter arguments for most of the cases.

mwear:
I updated this PR to handle all errors as recoverable. If we are happy with this approach, we could consider making this an option for all exporters via exporter helper similar to: https://github.com/open-telemetry/opentelemetry-collector/pull/8684 likely as a followup.

exporter/otlpexporter/otlp.go
```diff
@@ -102,14 +108,17 @@ func (e *baseExporter) pushTraces(ctx context.Context, td ptrace.Traces) error {
...
 	}
-	return nil
+	return
 }
 
-func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) error {
+func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) (err error) {
+	defer func() {
+		e.reportStatusFromError(err)
+	}()
```
bogdandrutu, line 117: I would do this way, because it is much easier to read:

```suggestion
func (e *baseExporter) pushMetricsWithStatus(ctx context.Context, md pmetric.Metrics) err {
	err := e.pushMetrics(ctx, md)
	if err != nil {
		componentstatus.ReportStatus(e.host, componentstatus.NewRecoverableErrorEvent(err))
	}
	return err
}

func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) error {
```

exporter/otlpexporter/otlp.go
```diff
@@ -192,6 +203,14 @@ func processError(err error) error {
 	return err
 }
 
+func (e *baseExporter) reportStatusFromError(err error) {
+	if err != nil {
+		componentstatus.ReportStatus(e.host, componentstatus.NewRecoverableErrorEvent(err))
+		return
+	}
+	componentstatus.ReportStatus(e.host, componentstatus.NewEvent(componentstatus.StatusOK))
```
bogdandrutu, line 211: Are we not concerned for every request to report the status? Should we do it periodically, or only when changes?

mwear: The [finite state machine](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md#transitioning-between-statuses) handles this for us. It will report status only if there has been a change, otherwise it will no-op.

mwear:
Would anyone be interested in reviewing this, or at least removing the stale label?

exporter/otlphttpexporter/otlp.go
```diff
@@ -131,6 +143,15 @@ func (e *baseExporter) pushMetrics(ctx context.Context, md pmetric.Metrics) erro
 	return e.export(ctx, e.metricsURL, request, e.metricsPartialSuccessHandler)
 }
 
+func (e *baseExporter) pushMetricsWithStatus(ctx context.Context, md pmetric.Metrics) error {
+	if err := e.pushMetrics(ctx, md); err != nil {
+		componentstatus.ReportStatus(e.host, componentstatus.NewRecoverableErrorEvent(err))
+		return err
+	}
+	componentstatus.ReportStatus(e.host, componentstatus.NewEvent(componentstatus.StatusOK))
```
atoulme, line 151: this might create a lot of events reporting OK, is that really what you want? Is it there to recover from the error status?

exporter/otlpexporter/otlp.go
josepdaniel: Hey @mwear, I'm not sure this is entirely true. I've built a simple statuswatcher extension that simply logs status changes and I can see my extension getting notified on same-state transitions for the same component:

```
got a status change for filelog/0, StatusStarting
got a status change for filelog/0, StatusOK
got a status change for otlp_http/0, StatusOK
got a status change for otlp_http/0, StatusOK
got a status change for otlp_http/0, StatusOK
```

(BTW this is built on some hacked-together wrapper around the otlp_http exporter with status reporting, until something like this PR can get merged). 

I can see that this might be an issue for extensions that do something expensive on each status report. If we get the FSM to recognize same-state-transitions and no-op them do you think we'd be unblocked to move ahead with this PR? 

Getting status reports on the core exporters is something we are invested in, so I'd be happy to contribute with a PR if you believe this will unblock us.

PR #11386, opened 2024-10-07
[cmd/builder] add ocb feature to install go binary to temp dir

### LINKED ISSUE #11382: remove local golang dependency for ocb binary
`ocb` binaries, even as they are compiled and built for specific architectures, still require a golang environment to run. This seems counterintuitive; I might expect `go install` command or compiling from scratch to require golang compiler, but I wouldn't expect a compiled binary to still require golang in the environment

I would like to remove the golang dependency for the ocb binary, allowing users to run ocb in a minimal (e.g. scratch, alpine, etc.) environment

Alternative is adding some explicit mention of golang requirement for `ocb` in documentation, both on github and opentelemetry.io

**Additional context**
I have been working on and submitted https://github.com/open-telemetry/opentelemetry-collector-releases/pull/671 which has a "thicker" image than required due the golang dependency for ocb binary. Making this change would allow a smaller image, such as from `scratch` that the collector distros run on.

jackgopack4:
#### Description
Removes requirement for ocb to have go binary installed in local environment. If go binary is not found in local path, downloads go binary for appropriate platform and architecture, based on the version that ocb was built with, installs to temporary directory, and deletes that directory when complete.

#### Link to tracking issue
Fixes #11382

#### Testing
Unit tests for each function added, as well as an additional test for `SetGoPath()` where `$PATH` is mocked to not have an existing go environment

#### Documentation
.chloggen file


andrzej-stencel:
This is an interesting solution to a make it easier for users to use the Builder without thinking about other dependencies - like to make sure Go is installed.

Still, I'm not convinced that downloading Go should be in scope of the Builder's responsibilities. This increases the complexity of the Builder's code. I'd rather keep the Builder's code as small as possible and have users use other tools to download Go. Happy to hear the maintainers' opinion on this.

jackgopack4:
@andrzej-stencel now that we have the docker images published I feel like that helps abstract away the requirement of having a local golang environment; there's of course also more security concerns with having ocb act as a "trojan horse," even if it only downloads golang to a temporary directory and removing afterwards. I don't feel strongly about this PR and I think we can probably let it expire in accordance with your comments about the issue https://github.com/open-telemetry/opentelemetry-collector/issues/11382

jpkrohling:
> I'm not convinced that downloading Go should be in scope of the Builder's responsibilities

I would vote to NOT have this feature. Setting up Go should not ocb's job.

PR #11388, opened 2024-10-07
fix version init

luccabb:
#### Description

```
debug.ReadBuildInfo()
```

returns an error for some environments, which then panics on `info.Main.Version`:

```
panic: runtime error: invalid memory address or nil pointer dereference
```

this fixes the issue by only updating the version if `debug.ReadBuildInfo()` succeeds


cmd/builder/internal/version.go
```diff
@@ -17,8 +17,10 @@ var (
 func init() {
 	// the second returned value is a boolean, which is true if the binaries are built with module support.
 	if version == "" {
```
bogdandrutu, line 19: Can you please for me write this `if version != "" { return }`?

bogdandrutu (approved):

codeboten (approved):

bogdandrutu (approved):

luccabb:
@bogdandrutu @codeboten @songy23 

any chance of a `v0.111.1` release?

PR #11396, opened 2024-10-08
[exporter] exporter queue Read()

sfc-gh-sili:
#### Description

This PR adds a public function `GetNextItem` to queue (both persistent queue and bounded memory queue)

Why this change?
Instead of blocking until consumption of the item is done, we would like to separate the API for reading and committing consumption.

Before:
`Consume(consumeFunc)`

After:
`idx, item = Read()`
`OnProcessingFinished(idx)`

#### Link to tracking issue

### DIFF
--- a/exporter/internal/queue/bounded_memory_queue.go
+++ b/exporter/internal/queue/bounded_memory_queue.go
@@ -44,13 +44,18 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
 func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
-	item, ok := q.sizedChannel.pop(func(el memQueueEl[T]) int64 { return q.sizer.Sizeof(el.req) })
+	_, item, ok := q.GetNextItem()
 	if !ok {
 		return false
 	}
 	// the memory queue doesn't handle consume errors
-	_ = consumeFunc(item.ctx, item.req)
+	_ = consumeFunc(context.Background(), item)
 	return true
+}
+
+func (q *boundedMemoryQueue[T]) GetNextItem() (uint64, T, bool) {
+	item, ok := q.sizedChannel.pop(func(el memQueueEl[T]) int64 { return q.sizer.Sizeof(el.req) })
+	return 0, item.req, ok
 }
 
 // Should be called to remove the item of the given index from the queue once processing is finished.
--- a/exporter/internal/queue/bounded_memory_queue.go
+++ b/exporter/internal/queue/bounded_memory_queue.go
@@ -44,7 +44,7 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
 func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
-	_, item, ok := q.GetNextItem()
+	_, item, ok := q.Read(context.Background())
 	if !ok {
 		return false
 	}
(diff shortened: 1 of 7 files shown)


exporter/internal/queue/bounded_memory_queue.go
```diff
@@ -44,15 +44,20 @@ func (q *boundedMemoryQueue[T]) Offer(ctx context.Context, req T) error {
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped and emptied.
 func (q *boundedMemoryQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
```
bogdandrutu, line 46: Should we remove this?

exporter/internal/queue/queue.go
```diff
@@ -33,6 +33,10 @@ type Queue[T any] interface {
 	Size() int
 	// Capacity returns the capacity of the queue.
 	Capacity() int
+	// GetNextItem pulls the next available item from the queue along with its index. Once processing is
+	// finished, the index should be called with OnProcessingFinished to clean up the storage. If no new
+	// item is available, returns false.
+	GetNextItem() (uint64, T, bool)
```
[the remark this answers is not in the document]
bogdandrutu: >If no new item is available, returns false.

Does it mean that it does return immediately? I think this is not the current implementation. See the other function comment "The function returns true when an item is consumed or false if the queue is stopped.". I think we need to make this blocking as well.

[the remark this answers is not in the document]
sfc-gh-sili: Instead of exposing `getNextItem()` directly, I changed the PR to expose `Read()` instead. For persistent queue, `Read()` will be block until there is a new item in the queue or until the queue is stopped, while `getNextItem()` is only called when we are sure (from the channel) there is a new item in the queue, and thus is non-blocking.
Comment explaining `Read()` is in `queue.go`. What do you think?

bogdandrutu (commented):
Overall LGTM on the new API

exporter/internal/queue/queue_test.go
```diff
@@ -0,0 +1,18 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package queue
+
+import (
+	"context"
+)
+
+func consume[T any](q Queue[T], consumeFunc func(context.Context, T) error) bool {
```
dmitryax, line 10: Is this used in tests only now? Will it be cleaner to update tests instead?

sfc-gh-sili: I agree. This function is called 21 times in total, and I thought it's easier to update those separately.

exporter/internal/queue/persistent_queue.go
```diff
@@ -193,31 +193,14 @@ func (pq *persistentQueue[T]) restoreQueueSizeFromStorage(ctx context.Context) (
 // The call blocks until there is an item available or the queue is stopped.
 // The function returns true when an item is consumed or false if the queue is stopped.
 func (pq *persistentQueue[T]) Consume(consumeFunc func(context.Context, T) error) bool {
```
bogdandrutu, line 195: Is this still needed?

bogdandrutu (approved):

exporter/internal/queue/persistent_queue.go
sfc-gh-sili: Thanks! I thought I cleaned that up.. Let me remove this

dmitryax (approved):

exporter/internal/queue/consumers.go
```diff
@@ -40,9 +40,12 @@ func (qc *Consumers[T]) Start(ctx context.Context, host component.Host) error {
 			startWG.Done()
 			defer qc.stopWG.Done()
 			for {
-				if !qc.queue.Consume(qc.consumeFunc) {
+				index, ctx, req, ok := qc.queue.Read(context.Background())
+				if !ok {
 					return
 				}
+				consumeErr := qc.consumeFunc(ctx, req)
+				qc.queue.OnProcessingFinished(index, consumeErr)
```
bogdandrutu, line 48:
```suggestion
				qc.queue.OnProcessingFinished(index, qc.consumeFunc(ctx, req))
```

bogdandrutu (approved):

Issue #10769, opened 2024-07-31
Own-telemetry: Ability to enabled/disable individual metrics


djaglowski:
**Is your feature request related to a problem? Please describe.**

Our notion of basic/normal/detailed telemetry levels is a quick and convenient way for users to tune the volume of the collector's own telemetry, but it is quite crude compared to what we provide to control telemetry not generated by the collector. For example, out `mdatagen` library provides the ability to enable or disable any individual metric. _Not_ having this ability for the collector's own telemetry means that users may not be able to get the metrics they need without enabling all metrics.

**Describe the solution you'd like**

In the telemetry configuration, we should offer the ability to enable or disable individual metrics, regardless of telemetry level.

The telemetry level should be used as the starting point for which metrics are enabled or disabled, but then individual metric settings can override.

codeboten:
Thanks for opening the issue. One question I have is how this works w/ views, because configuration of views would allow this same functionality. I'd rather not have multiple ways of doing the same thing if possible

djaglowski:
I'm not familiar enough with views to give a good answer here. Are they exposed to our users via the collector config? Or are they an implementation detail? For that matter, is there a reason we don't use views instead of in-code checks against metric level?

codeboten:
@djaglowski views will be configurable in the same way as they are for SDK via config, see https://github.com/open-telemetry/opentelemetry-configuration/blob/cc7cd375d893e440afdc294a00f98038ebaa4eab/examples/kitchen-sink.yaml#L221 as an example

The main reason I can see wanting to keep a telemetry level configuration option is for convenience to end users. It would be easier to configure one setting than to have to configure views for all metrics. But then, if we have multiple ways to configured things, we get into the issue of precedence.

djaglowski:
> The main reason I can see wanting to keep a telemetry level configuration option is for convenience to end users.

I agree but couldn't these be implemented as views? (Then for example component developers don't have to check `if telemetrySettings.Level == Detailed`, etc.)

If we did that, does that normalize configuration in a way where users could add or subtract from our predefined views by adding their own?

codeboten:
> I agree but couldn't these be implemented as views? (Then for example component developers don't have to check if telemetrySettings.Level == Detailed, etc.)

[...]

PR #11406, opened 2024-10-09
RFC - Pipeline Component Telemetry

djaglowski:
This PR adds a RFC for normalized telemetry across all pipeline components. See https://github.com/open-telemetry/opentelemetry-collector/issues/11343

edit by @mx-psi:
- Announced on #otel-collector-dev on 2024-10-23: https://cloud-native.slack.com/archives/C07CCCMRXBK/p1729705290741179
- Announced on the Collector SIG meeting from 2024-10-30


docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,119 @@
...
+      description: Size of items emitted from the component.
+      unit: "By"
+      sum:
+        value_type: int
+        monotonic: true
+```
+
+### Spans
+
+A span should be recorded for each execution of a processor or connector. The instrumentation layers adjacent to these components can start and end the span as appropriate.
```
codeboten, line 98: This needs additional details. Why is this limited to processors/connectors? Should there be any span events recorded?

codeboten (commented):
Thanks for opening this as a RFC @djaglowski!

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Goals
+
+1. Articulate a mechanism which enables us to _automatically_ capture telemetry from _all pipeline components_.
+2. Define attributes that are (A) specific enough to describe individual component [_instances_](https://github.com/open-telemetry/opentelemetry-collector/issues/10534) and (B) consistent enough for correlation across signals.
+3. Define specific metrics for each kind of pipeline component.
+4. Define specific spans for processors and connectors.
```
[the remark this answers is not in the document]
djaglowski: Good question. It's because the proposal is entirely based on the mechanism described, which is to insert instrumentation _in between_ pipeline components. In other words, there is a clear way to interact with data flowing both in and out of processors and connectors, but receivers and exporters we only have a generic way to interact with the data at one end.

All that said, I realized there is a flaw in my thinking about processors and connectors, because we cannot assume a direct relationship between data passed in and emitted from these components. We may be starting a lot of spans and not ending some, or ending the wrong ones. I will remove spans from this proposal for now.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,119 @@
...
+
+- `otel.component.kind`: `exporter`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`, **OR `ALL`**
+
+#### Connectors
+
+- `otel.component.kind`: `connector`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs->logs`, `logs->metrics`, `logs->traces`, `metrics->logs`, `metrics->metrics`, etc, **OR `ALL`**
```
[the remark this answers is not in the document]
djaglowski: If we annotate the incoming and outgoing counts with separate attributes, then connectors such as `count` or `otlpjson` will report a single data stream that aggregates across multiple instances. I believe this could lead to problems very similar to the one described [here](https://github.com/open-telemetry/opentelemetry-collector/issues/10908), where users need to be able to differentiate between instances in order to understand a problem. I think this is fundamentally a question of whether or not we should represent component _instances_ to users, or whether we're trying to aggregate and present them in some unified way. My opinion is that we have tried to gloss over this distinction for a long time but the nuances matter in some cases and we're not doing users a favor by hiding the distinction.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,119 @@
...
+      sum:
+        value_type: int
+        monotonic: true
+```
+
+### Spans
+
+A span should be recorded for each execution of a processor or connector. The instrumentation layers adjacent to these components can start and end the span as appropriate.
+
+### Logs
```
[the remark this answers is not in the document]
djaglowski: I'm not sure the mechanism I've described is able to do this. Also, since component status is still unstable, I'd prefer not to take that as a dependency here. My goal with this RFC is not to describe all standard telemetry for pipeline components, but to describe what can be done with the proposed mechanism.

docs/rfcs/component-universal-telemetry.md
djaglowski: I'm going to remove spans from this proposal based on the note [here](https://github.com/open-telemetry/opentelemetry-collector/pull/11406#discussion_r1795776208).

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+
+- `otel.component.kind`: `exporter`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`, **OR `ANY`**
+
+#### Connectors
+
+- `otel.component.kind`: `connector`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs->logs`, `logs->metrics`, `logs->traces`, `metrics->logs`, `metrics->metrics`, etc, **OR `ANY`**
```
bogdandrutu, line 45: I think these are very hard to use. We should probably split this into 2 labels to make them easier to search. I would use "otel.signal" for the incoming, and "otel.output.signal" for the outgoing (not sure about the second name)

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+
+Notes: The use of `ANY` indicates that values are not associated with a particular signal or pipeline. This is used when a component enforces non-standard instancing patterns. For example, the `otlp` receiver isa singleton, so the values are aggregated across signals. Similarly, the `memory_limiter` processor is a singleton, so the values are aggregated across pipelines.
```
bogdandrutu, line 47: I would not have it instead. The presence of the label with a value any just adds overhead in my opinion.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+1. Articulate a mechanism which enables us to _automatically_ capture telemetry from _all pipeline components_.
+2. Define attributes that are (A) specific enough to describe individual component [_instances_](https://github.com/open-telemetry/opentelemetry-collector/issues/10534) and (B) consistent enough for correlation across signals.
+3. Define specific metrics for each kind of pipeline component.
+4. Define specific logs for all kinds of pipeline component.
+
+### Mechanism
+
+The mechanism of telemetry capture should be _external_ to components. Specifically, we should observe telemetry at each point where a component passes data to another component, and, at each point where a component consumes data from another component. In terms of the component graph, every _edge_ in the graph will have two layers of instrumentation - one for the producing component and one for the consuming component. Importantly, each layer generates telemetry ascribed to a single component instance, so by having two layers per edge we can describe both sides of each handoff independently.
+
+### Attributes
```
bogdandrutu, line 18: Talk about the "scope" as well.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+- `otel.signal`: `logs->logs`, `logs->metrics`, `logs->traces`, `metrics->logs`, `metrics->metrics`, etc, **OR `ANY`**
+
+Notes: The use of `ANY` indicates that values are not associated with a particular signal or pipeline. This is used when a component enforces non-standard instancing patterns. For example, the `otlp` receiver isa singleton, so the values are aggregated across signals. Similarly, the `memory_limiter` processor is a singleton, so the values are aggregated across pipelines.
+
+### Metrics
+
+There are two straightforward measurements that can be made on any pdata:
+
+1. A count of "items" (spans, data points, or log records). These are low cost but broadly useful, so they should be enabled by default.
+2. A measure of size, based on [ProtoMarshaler.Sizer()](https://github.com/open-telemetry/opentelemetry-collector/blob/9907ba50df0d5853c34d2962cf21da42e15a560d/pdata/ptrace/pb.go#L11). These are high cost to compute, so by default they should be disabled (and not calculated).
```
bogdandrutu, line 54: +1

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+
+The location of these measurements can be described in terms of whether the data is "incoming" or "outgoing", from the perspective of the component to which the telemetry is ascribed.
+
+1. Incoming measurements are attributed to the component which is _consuming_ the data.
+2. Outgoing measurements are attributed to the component which is _producing_ the data.
```
bogdandrutu, line 59: Challenge for this: Do we really need both? If ever component records this, the outgoing of component X will be equal with next component Y metrics.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+
+For both metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to whether or not the function call returned an error. Outgoing measurements will be recorded with `outcome` as `failure` when the next consumer returns an error, and `success` otherwise. Likewise, incoming measurements will be recorded with `outcome` as `failure` when the component itself returns an error, and `success` otherwise.
```
bogdandrutu, line 61: Which function call?

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
...
+      sum:
+        value_type: int
+        monotonic: true
+    otelcol_component_outgoing_size:
+      enabled: false
+      description: Size of items emitted from the component.
+      unit: "By"
+      sum:
+        value_type: int
+        monotonic: true
```
bogdandrutu, line 92: This is something that triggers another question to me:

* To control if a metric is enabled and recorded is in the creation of the SDK. We will need a way for components (for this metrics as well) to communicate this with the service telemetry.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,118 @@
+# Auto-Instrumented Component Telemetry
```
bogdandrutu, line 1: Not sure if this rfcs should do that or not, but would be good to have a "generic" section for components telemetry where we say what are the "recommended" attributes that everyone should use for all the signals. I know we enforce that for logger, but we don't do that for trcaes/metrics.

docs/rfcs/component-universal-telemetry.md
andrzej-stencel: I was also considering this. Are there situations where `outgoing` metric of component A is not equal to `incoming` metric of component B? I think this is the same information. Does splitting this into two different metrics help in usage?

I was considering a "one metric per graph edge" approach, as described in the below examples.

The advantage of the "one metric per edge" approach is that with sufficiently [elaborate](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.10.0/deploy/helm/sumologic/conf/logs/otelcol/config.yaml#L562-L591) pipelines, the amount of metrics is cut in half (asymptotically).

Is there information that cannot be expressed with this "one metric per edge" approach? Are there other disadvantages, e.g. usability?

<details><summary>Example 1: Simple pipeline with Filter processor</summary>

```yaml
exporters:
  debug:
processors:
  filter:
    logs:
      log_record:
        - 'severity_number < SEVERITY_NUMBER_WARN'
receivers:
  filelog:
    include:
      - logs.json
service:
  pipelines:
    logs:
      exporters:
        - debug
      processors:
        - filter
      receivers:
        - filelog
```

Metrics according to this proposal:

```console
otelcol_component_incoming_items{otel.component.kind="receiver",otel.component.id="filelog",otel.signal="logs"} 2
otelcol_component_outgoing_items{otel.component.kind="receiver",otel.component.id="filelog",otel.signal="logs"} 2
otelcol_component_incoming_items{otel.component.kind="processor",otel.component.id="filter",otel.signal="logs"} 2
otelcol_component_outgoing_items{otel.component.kind="processor",otel.component.id="filter",otel.signal="logs"} 1
otelcol_component_incoming_items{otel.component.kind="exporter",otel.component.id="exporter",otel.signal="logs"} 1
otelcol_component_outgoing_items{otel.component.kind="exporter",otel.component.id="exporter",otel.signal="logs"} 1
```

Here's another proposal, with one metric per graph edge:

```console
otelcol_component[_incoming]_items{otel[.destination].component.kind="receiver",otel[.destination].component.id="filelog",otel.signal="logs"} 2
otelcol_component_items{otel.source.component.kind="receiver",otel.source.component.id="filelog",otel.destination.component.kind="processor",otel.destination.component.id="filter",otel.signal="logs"} 2
otelcol_component_items{otel.source.component.kind="processor",otel.source.component.id="filter",otel.destination.component.kind="exporter",otel.destination.component.id="debug",otel.signal="logs"} 1
otelcol_component[_outgoing]_items{otel.component.kind="exporter",otel.component.id="exporter",otel.signal="logs"} 1
```
</details>

<details><summary>Example 2: Pipeline with connector</summary>

```yaml
connectors:
  otlpjson:
exporters:
  debug:
receivers:
  filelog:
    include:
      - otlp-log.json
      - otlp-metric.json
      - otlp-span.json
service:
  pipelines:
    logs/input:
      exporters: [otlpjson]
      receivers: [filelog]
    logs/otlp:
      exporters: [debug]
      receivers: [otlpjson]
    metrics/otlp:
      exporters: [debug]
      receivers: [otlpjson]
    traces/otlp:
      exporters: [debug]
      receivers: [otlpjson]
```

This proposal:

```console
otelcol_component_incoming_items{otel.component.kind="receiver",otel.component.id="filelog",otel.signal="logs"} 3
otelcol_component_outgoing_items{otel.component.kind="receiver",otel.component.id="filelog",otel.signal="logs"} 3

otelcol_component_incoming_items{otel.component.kind="connector",otel.component.id="otlpjson",otel.signal="logs->logs"} 3
otelcol_component_incoming_items{otel.component.kind="connector",otel.component.id="otlpjson",otel.signal="logs->metrics"} 3
otelcol_component_incoming_items{otel.component.kind="connector",otel.component.id="otlpjson",otel.signal="logs->traces"} 3
otelcol_component_outgoing_items{otel.component.kind="connector",otel.component.id="otlpjson",otel.signal="logs->logs"} 1
otelcol_component_outgoing_items{otel.component.kind="connector",otel.component.id="otlpjson",otel.signal="logs->metrics"} 1
otelcol_component_outgoing_items{otel.component.kind="connector",otel.component.id="otlpjson",otel.signal="logs->traces"} 1

otelcol_component_incoming_items{otel.component.kind="exporter",otel.component.id="debug",otel.signal="logs"} 1
otelcol_component_outgoing_items{otel.component.kind="exporter",otel.component.id="debug",otel.signal="logs"} 1
otelcol_component_incoming_items{otel.component.kind="exporter",otel.component.id="debug",otel.signal="metrics"} 1
otelcol_component_outgoing_items{otel.component.kind="exporter",otel.component.id="debug",otel.signal="metrics"} 1
otelcol_component_incoming_items{otel.component.kind="exporter",otel.component.id="debug",otel.signal="traces"} 1
otelcol_component_outgoing_items{otel.component.kind="exporter",otel.component.id="debug",otel.signal="traces"} 1
```

One metric per edge:

```console
otelcol_component[_incoming]_items{otel[.destination].component.kind="receiver",otel[.destination].component.id="filelog",otel.signal="logs"} 3

otelcol_component_items{otel.source.component.kind="receiver",otel.source.component.id="filelog",otel.destination.component.kind="connector",otel.destination.component.id="otlpjson",otel.signal="logs"} 3
otelcol_component_items{otel.source.component.kind="receiver",otel.source.component.id="filelog",otel.destination.component.kind="connector",otel.destination.component.id="otlpjson",otel.signal="metrics"} 3
otelcol_component_items{otel.source.component.kind="receiver",otel.source.component.id="filelog",otel.destination.component.kind="connector",otel.destination.component.id="otlpjson",otel.signal="traces"} 3

otelcol_component_items{otel.source.component.kind="connector",otel.source.component.id="otlpjson",otel.destination.component.kind="exporter",otel.destination.component.id="debug",otel.signal="logs"} 1
otelcol_component_items{otel.source.component.kind="connector",otel.source.component.id="otlpjson",otel.destination.component.kind="exporter",otel.destination.component.id="debug",otel.signal="metrics"} 1
otelcol_component_items{otel.source.component.kind="connector",otel.source.component.id="otlpjson",otel.destination.component.kind="exporter",otel.destination.component.id="debug",otel.signal="traces"} 1

otelcol_component[_outgoing]_items{otel[.source].component.kind="exporter",otel[.source].component.id="exporter",otel.signal="logs"} 1
otelcol_component[_outgoing]_items{otel[.source].component.kind="exporter",otel[.source].component.id="exporter",otel.signal="metrics"} 1
otelcol_component[_outgoing]_items{otel[.source].component.kind="exporter",otel[.source].component.id="exporter",otel.signal="traces"} 1
```

</details>

djaglowski: > Do we really need both? If ever component records this, the outgoing of component X will be equal with next component Y metrics.

This is only true for linear pipelines. Any fanout or merge of data streams breaks this. e.g. Receiver A emits 10 items, receiver B emits 20, processor C consumes 30.

> The advantage of the "one metric per edge" approach is that with sufficiently [elaborate](https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/v4.10.0/deploy/helm/sumologic/conf/logs/otelcol/config.yaml#L562-L591) pipelines, the amount of metrics is cut in half (asymptotically).

Philosophically, I think users should reasonably expect that they can observe an object in isolation to get a relatively complete picture of how it is behaving. Offloading half of the picture for the sake of reducing data volume is a premature optimization in my opinion. In any case, if we support in [some way the ability to disable individual metrics](https://github.com/open-telemetry/opentelemetry-collector/issues/10769), then users can just disable one or the other to opt into exactly this same optimization.

More concretely, say we measure only incoming values. Some things that immediately become difficult for users:
1. How much data is my filter processor discarding? I can find the incoming value easily, but in order to answer the question I also have to figure out which component(s) it is sending data to, query for _their_ incoming values, aggregate those values, then compare to the incoming. I'm sure this is easier in some backends than others but in I don't think users will often write queries for this - they will just accept a lack of visibility. The query is even more difficult if you want it to be resilient to changes in the configuration.
3. How much data is my receiver emitting? Same convoluted answer as above.
4. In a linear pipeline, what is the net effect of all my processors? Same convoluted answer as above (vs just comparing outgoing from the receiver to incoming of the exporter).

The same problems exists if you only measure outgoing. The problem are even worse if we actually pin metrics to edges by including attributes from both components.

docs/rfcs/component-universal-telemetry.md
djaglowski: That makes sense to me. It's effectively the same as if the metric was aggregated over the dimension, which is what we're trying to communicate anyways.

Curious what @codeboten thinks too since he commented on this already.

docs/rfcs/component-universal-telemetry.md
djaglowski: I am not up to speed on the latest design but wouldn't it be intuitive that `component.TelemetrySettings` is able to answer the question of whether or not a metric is enabled?

docs/rfcs/component-universal-telemetry.md
djaglowski: I was thinking about this too. The attributes are almost a separate proposal because they apply more broadly. Really all telemetry emitted by a component should have these attributes, but this proposal only describes telemetry that can be captured automatically.

I think https://github.com/open-telemetry/opentelemetry-collector/pull/11344 demonstrates how these exact attributes can be available when every component is created. Maybe we can also pack them into the `TelemetrySettings` in a way where any signals automatically include them?

In any case, do you think I should peel off a separate RFC for attributes? (We do already have #11179 also.)

docs/rfcs/component-universal-telemetry.md
djaglowski: I think you're probably right. Will give some more thought to how to name them.

djaglowski: I split these into two with your suggested names.

docs/rfcs/component-universal-telemetry.md
codeboten: > To control if a metric is enabled and recorded is in the creation of the SDK. We will need a way for components (for this metrics as well) to communicate this with the service telemetry.

This makes me think that the components will need to be able to return views (we've talked about this a few times).

The alternative here would be to only support enabling/disabling of metrics via view configuration.

docs/rfcs/component-universal-telemetry.md
codeboten: I think it's ok to remove it... if we find a clear use for it later we can add it back in

docs/rfcs/component-universal-telemetry.md
bogdandrutu: >This makes me think that the components will need to be able to return views (we've talked about this a few times).

Yes, that would be great.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,119 @@
...
+
+The location of these measurements can be described in terms of whether the data is "incoming" or "outgoing", from the perspective of the component to which the telemetry is ascribed.
+
+1. Incoming measurements are attributed to the component which is _consuming_ the data.
+2. Outgoing measurements are attributed to the component which is _producing_ the data.
+
+For both metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to whether or not the corresponding function call returned an error. Specifically, incoming measurements will be recorded with `outcome` as `failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, outgoing measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and `success` otherwise.
+
+```yaml
+    otelcol_component_incoming_items:
```
bogdandrutu, line 65: There is an advantage to have the `otel.component.kind` in the name of the metric vs as a label:
* The rule of thumb for having a label vs being in the name is if the sum across that label makes sense or not. I think actually it doesn't make that much sense.
* The mechanism of disabling (you can still disable the calculation, but not the recording since you don't know which labels will be recorded) and avoiding computation is at "metric" level. 

I think it may not be bad to have per component kind metrics. I would not necessary go to have per component kind and type level, but that can be another option.

docs/rfcs/component-universal-telemetry.md
djaglowski: I've added clarification in the doc.

docs/rfcs/component-universal-telemetry.md
djaglowski: I've added language about the instrumentation scope.

djaglowski:
Based on some offline feedback, I've broadened the scope of the RFC, while simultaneously clarifying that it is intended to evolve as we identify additional standards.

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: `otelcol_component_consumed_items` and `otelcol_component_produced_items`?

djaglowski: > The rule of thumb for having a label vs being in the name is if the sum across that label makes sense or not. I think actually it doesn't make that much sense.

I am on the fence about this. I agree the case is weak but aggregation along this dimension might at least give a sense of scale.

What I mean is that if you consider each component as doing some non-trivial amount of work for a amount of data, you can aggregate to understand relatively how much work the collector is doing internally. e.g. Say you are considering some changes to a configuration and compare v(n) to v(n+1). If you aggregate across all dimensions of "incoming" you may see that even with the same inputs the collector is having 10x items passed between components.

Either way, I'm not necessarily against having metrics per kind. Some consequences though:
- To have consistent attributes with other signals, we should still apply the `otel.signal` attribute, even though this is redundant with the metric name.
- It's more difficult for users to answer questions like, "which of the components is producing the most items" because they have to join across 3 metrics before sorting?

> The mechanism of disabling (you can still disable the calculation, but not the recording since you don't know which labels will be recorded) and avoiding computation is at "metric" level.

I'm not sure I understand. Are you pointing out that users would be forced to enable/disable the metric for all kinds, but maybe they should be able to control it for each kind independently? e.g. Only record items produced by receivers and consumed by exporters?

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
mx-psi: This is no longer in the RFC text, marking the comment as resolved

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
mx-psi: Agreed that this can be out of scope for now

docs/rfcs/component-universal-telemetry.md
mx-psi: I think we resolved this on yesterday's stability meeting, can this be marked as solved (and we can come back to this in the next revision of the RFC), or is there something left to discuss on this thread?

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
mx-psi: > `otelcol_component_consumed_items` and `otelcol_component_produced_items`?

I like these names

mx-psi: Criteria discussedfrom 2024-10-21 meeting for continuing the conversation
- Looking at what we do in semantic conventions for similar cases (databases?)
- Thinking about discoverability, specifically in the context of additional, component kind specific metrics

docs/rfcs/component-universal-telemetry.md
djaglowski: > It sounds to me like these issues with having one metric per edge are caused by the fan-in and fan-out parts of the pipeline being "implicit".

Edges are implicit in our configuration, but components are explicit. Similarly, even in the code edges are barely more than a function call between components, whereas components are actual instances with state and attributes. I think this reflects the configuration, documentation, etc, and therefore it makes sense to ascribe signals to components rather than edges.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,158 @@
...
+
+1. A count of "items" (spans, data points, or log records). These are low cost but broadly useful, so they should be enabled by default.
+2. A measure of size, based on [ProtoMarshaler.Sizer()](https://github.com/open-telemetry/opentelemetry-collector/blob/9907ba50df0d5853c34d2962cf21da42e15a560d/pdata/ptrace/pb.go#L11).
+  These are high cost to compute, so by default they should be disabled (and not calculated).
+
+The location of these measurements can be described in terms of whether the data is "consumed" or "produced", from the perspective of the
+component to which the telemetry is ascribed.
+
+1. Incoming measurements are attributed to the component which is _consuming_ the data.
+2. Outgoing measurements are attributed to the component which is _producing_ the data.
```
jaronoff97, line 87: Will scrape based receivers (like prometheus) increment the consumed or produced measurement? In my mind, prometheus _produces_ data and has no _consumption_ therefore we would see 

```
otelcol_component_consumed_items{"otel.component.kind": "receiver", "otel.component.id": "prometheus", "otel.signal": "metrics"} 0
otelcol_component_produced_items{"otel.component.kind": "receiver", "otel.component.id": "prometheus", "otel.signal": "metrics"} 100
```

Is my understanding correct? And if so, can we document as such?

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,158 @@
...
+edge we can describe both sides of each handoff independently.
+
+Telemetry captured by this mechanism should be associated with an instrumentation scope corresponding to the package which implements
+the mechanism. Currently, that package is `service/internal/graph`, but this may change in the future. Notably, this telemetry is not
+ascribed to individual component packages, both because the instrumentation scope is intended to describe the origin of the telemetry,
+and because no mechanism is presently identified which would allow us to determine the characteristics of a component-specific scope.
+
+### Auto-Instrumented Metrics
+
+There are two straightforward measurements that can be made on any pdata:
```
jaronoff97, line 77: Should we be making use of the defined config telemetry levels [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtelemetry/doc.go) as well? Do these metrics fall within the "Basic" level or "Normal" level? Would it be helpful to add additional attributes or even additional metrics if this is the case?

jaronoff97 (commented):
a few questions, I really like this proposal overall :)

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,158 @@
...
+2. Outgoing measurements are attributed to the component which is _producing_ the data.
+
+For both metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to
+whether or not the corresponding function call returned an error. Specifically, consumed measurements will be recorded with `outcome` as
+`failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, produced
+measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and
+`success` otherwise.
+
+```yaml
+    otelcol_component_consumed_items:
```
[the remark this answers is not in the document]
djaglowski: Minor correction - receivers would not report a "consumed" metric based on my proposal, since this cannot be automatically captured (as the mechanism of consuming varies for each receiver).

If I'm understanding your point though, I think it is essentially the same as made by Bogdan [here](https://github.com/open-telemetry/opentelemetry-collector/pull/11406#discussion_r1803413871).

> I think it would essentially be useful to document some of the equations put forth in [this OTEP](https://github.com/open-telemetry/oteps/pull/259/files#diff-8ab369bae50cfda348c0bd20882db8a1bdbbffb979abc70bff298192a948588bR245).

If you have specific suggestions please share them.

docs/rfcs/component-universal-telemetry.md
djaglowski: Receivers will not have a "consumed" metric. (Likewise, exporters will not have a "produced" metric.)

This proposal is focused on telemetry which can be _automatically_ captured. The opportunity we have to do that is in when data is passed between components.

docs/rfcs/component-universal-telemetry.md
djaglowski: There is open debate about whether those levels are going to be removed so I don't want this proposal to hinge on them. I don't think it's necessary for this proposal to worry about the mechanisms for enabling and disabling metrics, so long as there is a way to define metrics as enabled or disabled by default, and another mechanism for users to override the default.

docs/rfcs/component-universal-telemetry.md
djaglowski: > Looking at what we do in semantic conventions for similar cases (databases?)

It's not clear to me that any existing semantic conventions are very relevant to this situation.

> Thinking about discoverability, specifically in the context of additional, component kind specific metrics

I think this was a good point. To elaborate on it, we may eventually define for example an additional metric that only applies to receivers. If so, the metric name will likely have a prefix such as `otelcol.receiver`, so it would be co-discoverable with `otelcol.receiver.produced.items` but not `otelcol.component.produced.items`.

jaronoff97: > To have consistent attributes with other signals, we should still apply the otel.signal attribute, even though this is redundant with the metric name.
I believe what bogdan is suggesting is putting the component kind in the metric name (i.e. `otelcol_receiver_produced_items`) which would avoid the redundancy here.

> It's more difficult for users to answer questions like, "which of the components is producing the most items" because they have to join across 3 metrics before sorting?
Maybe this is naive of me... but are there processors that increase the data produced? I guess this theoretically is possible, but I'm not aware of any processors that do this today. As a result, i think you would max across the receiver produced items.

I think my only qualm with this naming (including component kind in the measure name) comes down to user experience. If a user only wants minimal telemetry out of the collector, giving them as few signals (keeping `otelcol_component_produced_items`) as possible while still containing useful data would be advantageous.

docs/rfcs/component-universal-telemetry.md
jaronoff97: ah okay, so a receiver regardless of its scraping or receiving data both "produce" data! I like it :bow:

docs/rfcs/component-universal-telemetry.md
jaronoff97: very fair. I guess it is an implementation detail.

docs/rfcs/component-universal-telemetry.md
djaglowski: Based on the points made in this thread, I have updated the proposed metrics to include component kinds in their names.

docs/rfcs/component-universal-telemetry.md
djaglowski: Is there anything actionable here for this RFC? If not, can we resolve this comment?

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,158 @@
...
+      enabled: true
+      description: Number of items passed to the component.
+      unit: "{items}"
+      sum:
+        value_type: int
+        monotonic: true
+    otelcol_component_produced_items:
```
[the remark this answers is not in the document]
jaronoff97: ah i see. So a downstream user will need some amount of configuration viewing to properly reassemble the metrics into the right places. That's alright with me, but it may be worth calling this out as a design decision / limitation.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Attributes
+
+All signals should use the following attributes:
+
+### Receivers
+
+- `otel.component.kind`: `receiver`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics`, `traces`
```
dmathieu, line 36:
```suggestion
- `otel.signal`: `logs`, `metrics`, `traces`, `profiles`
```

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Processors
+
+- `otel.component.kind`: `processor`
+- `otel.component.id`: The component ID
+- `otel.pipeline.id`: The pipeline ID
+- `otel.signal`: `logs`, `metrics`, `traces`
```
dmathieu, line 43:
```suggestion
- `otel.signal`: `logs`, `metrics`, `traces`, `profiles`
```

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Exporters
+
+- `otel.component.kind`: `exporter`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`
```
dmathieu, line 49:
```suggestion
- `otel.signal`: `logs`, `metrics` `traces`, `profiles`
```

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Connectors
+
+- `otel.component.kind`: `connector`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`
+- `otel.output.signal`: `logs`, `metrics` `traces`
```
dmathieu, line 56:
```suggestion
- `otel.output.signal`: `logs`, `metrics` `traces`, `profiles`
```

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+3. Define specific metrics for each kind of pipeline component.
+4. Define specific logs for all kinds of pipeline component.
+
+## Attributes
+
+All signals should use the following attributes:
+
+### Receivers
+
+- `otel.component.kind`: `receiver`
```
jaronoff97, line 34: given the kind is now in the measurement name, do we still need this attribute?

djaglowski: Yes. Only metrics have names, so we need this in order to be able to correlate with other signals. Even the metrics should still have the attribute so that they can be correlated.

docs/rfcs/component-universal-telemetry.md
djaglowski: Marking this as resolved as there doesn't appear to be anything actionable.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+### Auto-Instrumented Metrics
+
+There are two straightforward measurements that can be made on any pdata:
+
+1. A count of "items" (spans, data points, or log records). These are low cost but broadly useful, so they should be enabled by default.
+2. A measure of size, based on [ProtoMarshaler.Sizer()](https://github.com/open-telemetry/opentelemetry-collector/blob/9907ba50df0d5853c34d2962cf21da42e15a560d/pdata/ptrace/pb.go#L11).
+  These are high cost to compute, so by default they should be disabled (and not calculated).
+
+The location of these measurements can be described in terms of whether the data is "consumed" or "produced", from the perspective of the
+component to which the telemetry is attributed. Metrics which contain the term "procuded" describe data which is emitted from the component,
```
wildum, line 84:
```suggestion
component to which the telemetry is attributed. Metrics which contain the term "produced" describe data which is emitted from the component,
```

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+are unified by the component implementation. For example, the `otlp` receiver is a singleton, so its telemetry is not specific to a signal.
+Similarly, the `memory_limiter` processor is a singleton, so its telemetry is not specific to a pipeline.
+
+## Auto-Instrumentation Mechanism
+
+The mechanism of telemetry capture should be _external_ to components. Specifically, we should observe telemetry at each point where a
+component passes data to another component, and, at each point where a component consumes data from another component. In terms of the
+component graph, every _edge_ in the graph will have two layers of instrumentation - one for the producing component and one for the
+consuming component. Importantly, each layer generates telemetry ascribed to a single component instance, so by having two layers per
+edge we can describe both sides of each handoff independently.
```
wildum, line 68: +1 for this mechanism. It should enable very useful debugging capabilities that are described in this [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34096) and were discussed in this [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/10963)

wildum (approved):

jaronoff97 (approved):

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: I'm resolving this comment, but please reopen if you think the justification is not sufficient.

docs/rfcs/component-universal-telemetry.md
djaglowski: I'm resolving this comment, but please reopen if you think the justifications are not sufficient.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+Metrics provide most of the observability we need but there are some gaps which logs can fill. Although metrics would describe the overall
+item counts, it is helpful in some cases to record more granular events. e.g. If a produced batch of 10,000 spans results in an error, but
+100 batches of 100 spans succeed, this may be a matter of batch size that can be detected by analyzing logs, while the corresponding metric
+reports only that a 50% success rate is observed.
+
+For security and performance reasons, it would not be appropriate to log the contents of telemetry.
+
+It's very easy for logs to become too noisy. Even if errors are occurring frequently in the data pipeline, they may only be of interest to
+many users if they are not handled automatically.
```
evan-bradley, line 191:
```suggestion
It's very easy for logs to become too noisy. Even if errors are occurring frequently in the data pipeline, only the errors that are not
handled automatically will be of interest to most users.
```
Just a wording suggestion, it took me a few times to parse the sentence as-written.

evan-bradley (approved):

jpkrohling:
Some of my comments might have been discussed before, in which case, feel free to ignore me and just mark the items as resolved.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+- `otel.component.kind`: `exporter`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`, `profiles`
+
+### Connectors
+
+- `otel.component.kind`: `connector`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`
+- `otel.output.signal`: `logs`, `metrics` `traces`, `profiles`
```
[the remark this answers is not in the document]
djaglowski: I think that makes sense. I will update it.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+      sum:
+        value_type: int
+        monotonic: true
+```
+
+### Auto-Instrumented Logs
+
+Metrics provide most of the observability we need but there are some gaps which logs can fill. Although metrics would describe the overall
+item counts, it is helpful in some cases to record more granular events. e.g. If a produced batch of 10,000 spans results in an error, but
+100 batches of 100 spans succeed, this may be a matter of batch size that can be detected by analyzing logs, while the corresponding metric
```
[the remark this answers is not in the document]
djaglowski: Makes sense if we agree that we should capture a span for the consume function.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+ascribed to individual component packages, both because the instrumentation scope is intended to describe the origin of the telemetry,
+and because no mechanism is presently identified which would allow us to determine the characteristics of a component-specific scope.
+
+### Auto-Instrumented Metrics
+
+There are two straightforward measurements that can be made on any pdata:
+
+1. A count of "items" (spans, data points, or log records). These are low cost but broadly useful, so they should be enabled by default.
+2. A measure of size, based on [ProtoMarshaler.Sizer()](https://github.com/open-telemetry/opentelemetry-collector/blob/9907ba50df0d5853c34d2962cf21da42e15a560d/pdata/ptrace/pb.go#L11).
+  These are high cost to compute, so by default they should be disabled (and not calculated).
```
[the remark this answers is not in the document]
djaglowski:
```suggestion
  These may be high cost to compute, so by default they should be disabled (and not calculated). This default setting may change in the future if it is demonstrated that the cost is generally acceptable.
```

How's this wording?

docs/rfcs/component-universal-telemetry.md
dmitryax, line 34: OTel SemConv guidelines require metric attributes to be namespaces so they can be placed in a global registry along with resource attributes. I don't think we can use `otel` for collector. It probably should be `otelcol`. Or `otel.collector` instead. However, `otel.` seems like [kind of a restricted namespace](https://github.com/open-telemetry/semantic-conventions/blob/de8f94ed5c8601f6ef8d596906d4eb5c2ff895ec/docs/general/attribute-naming.md#otel-namespace). Not sure if we can use it or not. I think we need to run this by the otel semantic conventions WG group before merging the RFC.

mx-psi: @dmitryax if you think contacting the semantic convention SIG is a requirement for this RFC to be merged, please request changes on the PR so that we follow [the process](https://github.com/open-telemetry/opentelemetry-collector/tree/main/docs/rfcs#merge-rules)

djaglowski: I'm fine with changing the attribute names but delaying this functionality further for an entirely separate naming process seems unnecessary. Can we really not choose our own attributes names?

mx-psi: This sentence from the doc linked by @dmitryax makes me hesitant about keeping the current approach:

> Any additions to the otel.* namespace MUST be approved as part of OpenTelemetry specification.

djaglowski: > This sentence from the doc linked by @dmitryax makes me hesitant about keeping the current approach:
> 
> > Any additions to the otel.* namespace MUST be approved as part of OpenTelemetry specification.

I agree we can't just use the `otel` namespace. Can't we choose another on our own though? I like the `otelcol` suggestion.

mx-psi: `otelcol` works for me, it's also the one we use for metrics so it makes sense to be consistent here

tigrannajaryan: @mx-psi asked me to chime in to help interpret the spec. 

You are reading it correctly. If you want to use `otel.` prefix it needs to go through semconv group.

What can you use instead? If you read the [preceding paragraph](https://github.com/open-telemetry/semantic-conventions/blob/de8f94ed5c8601f6ef8d596906d4eb5c2ff895ec/docs/general/attribute-naming.md#recommendations-for-application-developers) the closest applicable rules seems to be this:

>The name is specific to your company and may be possibly used outside the company as well. To avoid clashes with names introduced by other companies (in a distributed system that uses applications from multiple vendors) it is recommended to prefix the new name by your company's reverse domain name, e.g. com.acme.shopname.

If we consider "OpenTelemetry" the "company" then we should use "io.opentelemetry." as the org-specific prefix, e.g. "io.opentelemetry.collector." for the Collector. 

However, I am not sure I like such a long prefix. "otelcol" or "otel.collector" looks much nicer to me.

I cannot find spec recommendations about when a new top-level namespace (like "otelcol") is allowed to be used. I would argue the spec should have a rule like this and the rule should be that it needs to go through semconv, just like for anything under "otel." namespace.

My recommendation is this:
- Decide which prefix you want to use: "otelcol." or "otel.collector." appear to be your top choices.
- Submit a request to get this added to semconv. We should be able to convince semconv group to accept it and I do not anticipate difficulties here.

You don't need to be blocked or wait until acceptance by semconv group, there is a ton of work that needs to happen anyway and replacing metric names should be trivial anytime before you declare them stable.

One additional thing that I think is worth thinking about: is this Collector-only telemetry or it can be applicable to SDK processors/exporters? Collector's telemetry is going to be much richer and needs to be more expressive, so I am not sure if it is beneficial to model Collector and SDK the same way. However, if we end up deciding that indeed we want to model them the same way, with SDK's use cases being virtually a subset of Collector's use cases, then "otel." prefix seems to be the most natural choice.

djaglowski: Thanks @tigrannajaryan. I'll open an issue to add `otelcol` as a top level namespace in the semantic conventions.

> One additional thing that I think is worth thinking about: is this Collector-only telemetry or it can be applicable to SDK processors/exporters? Collector's telemetry is going to be much richer and needs to be more expressive, so I am not sure if it is beneficial to model Collector and SDK the same way. However, if we end up deciding that indeed we want to model them the same way, with SDK's use cases being virtually a subset of Collector's use cases, then "otel." prefix seems to be the most natural choice.

This is intended to be collector-only telemetry. It is attributed to the collector's _instances_ of components, which are governed by a model that is not intended to be aligned with other parts of the project.

docs/rfcs/component-universal-telemetry.md
andrzej-stencel: I get the **philosophical** argument that components are easier to notice or to reason about. I generally tend to favor solutions that are easy to reason about even if not completely optimized.

In this case, I'm on the fence. I'd argue that **conceptually** the information we want to express - "how much telemetry has been sent from component A to B" -  is not a property of either of the components (either A or B), but rather it's a property of the relation between those components.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+whether or not the corresponding function call returned an error. Specifically, consumed measurements will be recorded with `outcome` as
+`failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, produced
+measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and
+`success` otherwise.
+
+```yaml
+    otelcol.receiver.produced.items:
+      enabled: true
+      description: Number of items emitted from the receiver.
+      unit: "{items}"
```
dmitryax, line 97: Unit should not be plural according to https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units.

Also, after reading the RFC, it's still unclear to me why signal goes to an attribute and unit is an item now. According to https://opentelemetry.io/docs/specs/semconv/general/metrics/#general-guidelines aggregation over all signals should give a meaningful value. I guess potentially it can be meaningful but unlikely useful. Is there any other reason to not keep separate metric names for spans, logs records and metric points other than reaggregation capability?

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+        value_type: int
+        monotonic: true
+    otelcol.processor.consumed.items:
+      enabled: true
+      description: Number of items passed to the processor.
+      unit: "{items}"
+      sum:
+        value_type: int
+        monotonic: true
+    otelcol.processor.produced.items:
```
bogdandrutu, line 108: Should this remove replace https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/processorhelper/metadata.yaml#L11?

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+while metrics which contain the term "consumed" describe data which is received by the component.
+
+For both metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to
+whether or not the corresponding function call returned an error. Specifically, consumed measurements will be recorded with `outcome` as
+`failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, produced
+measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and
+`success` otherwise.
+
+```yaml
+    otelcol.receiver.produced.items:
```
bogdandrutu, line 94: Can we remove the "otelcol." prefix and use a meaningful scope instead?

bogdandrutu (approved):
Approved with comments.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+   and (B) consistent enough for correlation across signals.
+2. Articulate a mechanism which enables us to _automatically_ capture telemetry from _all pipeline components_.
+3. Define specific metrics for each kind of pipeline component.
+4. Define specific logs for all kinds of pipeline component.
+
+## Attributes
+
+All signals should use the following attributes:
+
+### Receivers
```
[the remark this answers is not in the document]
codeboten: I agree w/ @djaglowski that it would be better to capture the scope of extensions separately. Opened https://github.com/open-telemetry/opentelemetry-collector/issues/11698 to track this work

docs/rfcs/component-universal-telemetry.md
codeboten: This would replace `processor_outgoing_items`

bogdandrutu: Sure, both metrics there will be replaced correct?

docs/rfcs/component-universal-telemetry.md
dmitryax: I don't think we can remove the prefix as the Otel semconv requires metric names to be defined in the global registry and do not conflict with each other

mx-psi: > PS: I meant [scope](https://github.com/open-telemetry/opentelemetry-specification/blob/ca2a0d40865c400361e6d443d040e836c0a54b81/specification/glossary.md?plain=1#L166) not prefix.

Ah, my bad. Agree with Dmitrii that some prefix should stay. The scope name should also be set (but I think we can just follow the OTel Go convention and we don't need to cover it on this RFC)

docs/rfcs/component-universal-telemetry.md
codeboten: Correct

docs/rfcs/component-universal-telemetry.md
codeboten: Having the `otelcol` prefix would make it easier for users transitioning from the current metrics to the new ones. 

Additionally, if there is a time where SDKs emit metrics about their telemetry pipelines, we could avoid accidentally aggregating metrics for collector pipelines with SDK pipelines (assuming they would be using the same metric names)

For these reasons i'm in favour of keeping the prefix and also including the scope. I guess if users feel strongly enough, views could always be configured to rename metrics

djaglowski: I will update the RFC to state that signals should include a scope name as well.

docs/rfcs/component-universal-telemetry.md
djaglowski: I think this comment thread is sufficient to document this decision and that it doesn't need to be called out explicitly in the RFC, but unresolve if you disagree.

docs/rfcs/component-universal-telemetry.md
djaglowski: I will update the units.

> it's still unclear to me why signal goes to an attribute and unit is an item now.

Let's separate this into two questions:

### 1. Should there be a `signal` attribute?

Yes, because these attributes are not just for metrics. With metrics if we put the signal in the metric name that may be enough, but you cannot do the same for e.g. a log. If we want to be able to correlate all telemetry from each instance, then we should have the exact same attributes across all signals.

### 2. Should signal type be in metric names? (in which case we use units like "data points" instead of "items")

We could do this, but if we agree with the rationale for question 1, then it would be redundant information. If only looking at this point, maybe it could go either way, but I think the question of aggregation is more important.

> I guess potentially it can be meaningful but unlikely useful.

I think the total size of data is very meaningful if aggregated across signals, as this tends to correlate to a large extent with volume of data flowing over the network and cost to customers. I agree item count is less obvious though. However, if the size case makes sense, then I think it is helpful to keep the structure of the two metrics the same.

--- 

Therefore, I think the best approach is as proposed in the RFC. This allows us correlate across all data types while being specific enough to ascribe telemetry to individual component instances. It also allows us to aggregate across telemetry types. And although not terribly important, it does not contain redundant information.

djaglowski: I've updated the units to `item` instead of `items`

docs/rfcs/component-universal-telemetry.md
djaglowski: Question about scope name:

If the telemetry is captured by an internal package, should the scope name be that of the module? e.g. `go.opentelemetry.io/service` instead of `go.opentelemetry.io/service/internal/graph`

I ask because as an internal module, it seems likely that the package name could change even if the `service` module is stable, which would break consistency of telemetry between versions. I think pinning to the module name is only slightly less helpful but more stable.

mx-psi: I agree that using the module name is preferable

djaglowski: I've added a section calling out the requirement for an instrumentation scope.

docs/rfcs/component-universal-telemetry.md
mx-psi: A couple of examples that use the approach on this RFC: Vector [uses `component_received_events_counts`](https://vector.dev/docs/administration/monitoring/) and Fluentbit [uses `fluentbit_input_records_total`](https://docs.fluentbit.io/manual/administration/monitoring#telemetry-pipeline). I think we should go with the approach currently in the RFC

dmitryax: Thanks for the detailed response, @djaglowski. That makes sense to me. Let's keep the current proposal. Given that, I think we should better document what an `item` is somewhere in the docs maybe.

dmitryax (approved):
LGTM

mx-psi:
This has enough approvals and has entered the 'final comment period'. I will merge this on 2024-11-27 if nobody blocks before.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Connectors
+
+- `otel.component.kind`: `connector`
+- `otel.component.id`: The component ID
+- `otel.signal`: `logs`, `metrics` `traces`
+- `otel.output.signal`: `logs`, `metrics` `traces`, `profiles`
+
+Note: The `otel.signal`, `otel.output.signal`, or `otel.pipeline.id` attributes may be omitted if the corresponding component instances
+are unified by the component implementation. For example, the `otlp` receiver is a singleton, so its telemetry is not specific to a signal.
```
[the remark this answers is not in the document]
jpkrohling: I would probably prefer "10 logs" and "20 metrics". I think we can refine this text over time. The main idea is to convey that there are things that are signal-specific, and things that are valid for the entire receiver, and this should reflect in the set of attributes being used.

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
jade-guiton-dd: It would be nice to have a clear consensus on this issue of logs vs. spans for information not easily captured by metrics, so I can try to align with that for #11581.

docs/rfcs/component-universal-telemetry.md
```diff
@@ -0,0 +1,212 @@
...
+With the above considerations, this proposal includes only that we add a DEBUG log for each individual outcome. This should be sufficient for
+detailed troubleshooting but does not impact users otherwise.
+
+In the future, it may be helpful to define triggers for reporting repeated failures at a higher severity level. e.g. N number of failures in
+a row, or a moving average success %. For now, the criteria and necessary configurability is unclear so this is mentioned only as an example
+of future possibilities.
+
+### Auto-Instrumented Spans
+
+It is not clear that any spans can be captured automatically with the proposed mechanism. We have the ability to insert instrumentation both
```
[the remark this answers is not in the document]
jpkrohling: > Do we accept that the span is just measuring a quick handoff to the internal state of the component? 

Yes: two incoming spans from different connections would belong to two different traces. They would be assembled into one batch and exporter as part of a single outgoing call by the exporter. This to me means that we have three traces: two representing the original requests, and one starting with the batch processor and ending with the exporter call. The entire flow can be understood by looking at the span links.

> Is this going to be misleading to users?

Good question: the scenario I described above is what I would expect as a user, but not sure everybody sees it the same way.

Do you think we can refine this later, or is it important to answer as part of this PR?

jpkrohling (approved):
There are a couple of things to iron out, but I'm giving my approval already, as those are details that could be part of a follow-up PR. I don't want to block progress on dependent tasks because of those two rather small points.

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: I generally agree with what you're saying but I agree it would be best handled later, and perhaps authored by someone other than myself who has a more complete understanding of tracing best practices. I'll leave this in place for now but create a tracking issue to reconsider this section.

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: See https://github.com/open-telemetry/opentelemetry-collector/pull/11406#discussion_r1856690368

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: I've opened https://github.com/open-telemetry/opentelemetry-collector/issues/11743 to track this

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: I've opened https://github.com/open-telemetry/opentelemetry-collector/issues/11743 to track this discussion

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
djaglowski: >The main idea is to convey that there are things that are signal-specific, and things that are valid for the entire receiver, and this should reflect in the set of attributes being used.

I agree with this in principle. The current text leaves room for this, but additional language can be added later to better represent the intention here.

codeboten (approved):
Thanks @djaglowski

jpkrohling (approved):
I approved this before, but I'll approve again, to make it explicit that I'm OK with the latest state of this PR.

mx-psi:
Per https://github.com/open-telemetry/opentelemetry-collector/pull/11406#issuecomment-2493239777 I am merging this :tada:

jade-guiton-dd:
To make sure everyone involved is aware: I filed a PR (#11956) to amend this RFC. I am proposing adding a third `outcome` attribute value to make tracing the source of errors easier.

PR #11411, opened 2024-10-10
[chore] Move go templates from pdatagen to separate files

bogdandrutu:


dmitryax (approved):
This is nice! Thank you! I actually wanted to do the same

PR #11414, opened 2024-10-10
[exporter] moving queue sender from pushing model to pulling model

sfc-gh-sili:
#### Description

This PR implements a new component that pulls from queueSender to solve the issue mentioned in https://github.com/open-telemetry/opentelemetry-collector/issues/10368.

The idea is that instead of allocating a group of reading goroutine and block them until the corresponding batch gets flushed, we now use a goroutine to read and then use the same goroutine go flush while allocating new goroutine to read.

Design doc:

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10368

### DIFF
--- a/exporter/exporterhelper/internal/base_exporter.go
+++ b/exporter/exporterhelper/internal/base_exporter.go
@@ -106,6 +106,10 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
 
 	if be.BatcherCfg.Enabled {
 		bs := NewBatchSender(be.BatcherCfg, be.Set, be.BatchMergeFunc, be.BatchMergeSplitfunc)
+		if bs.mergeFunc == nil || bs.mergeSplitFunc == nil {
+			err = multierr.Append(err, fmt.Errorf("Batch merge function has to be provided"))
+		}
+		be.BatchSender = bs
 		if !be.queueCfg.Enabled {
 			be.BatchSender = bs
 		}
--- a/exporter/exporterhelper/internal/base_exporter.go
+++ b/exporter/exporterhelper/internal/base_exporter.go
@@ -97,7 +97,8 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
 				ExporterSettings: be.Set,
 			},
 			be.queueCfg)
-		be.QueueSender = NewQueueSender(q, be.Set, be.queueCfg.NumConsumers, be.ExportFailureMessage, be.Obsrep)
+		be.QueueSender = NewQueueSender(q, be.Set, be.queueCfg.NumConsumers,
+			be.ExportFailureMessage, be.Obsrep, be.BatcherCfg, be.BatchMergeFunc, be.BatchMergeSplitfunc)
 		for _, op := range options {
 			err = multierr.Append(err, op(be))
 		}
@@ -105,10 +106,9 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
 
 	if be.BatcherCfg.Enabled {
 		bs := NewBatchSender(be.BatcherCfg, be.Set, be.BatchMergeFunc, be.BatchMergeSplitfunc)
-		if bs.mergeFunc == nil || bs.mergeSplitFunc == nil {
-			err = multierr.Append(err, fmt.Errorf("WithRequestBatchFuncs must be provided for the batcher applied to the request-based exporters"))
+		if !be.queueCfg.Enabled {
+			be.BatchSender = bs
 		}
-		be.BatchSender = bs
 	}
 
 	if err != nil {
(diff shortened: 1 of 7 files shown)


exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -93,6 +93,25 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
...
+			err = multierr.Append(err, opt(bs))
+		}
+
+		if bs.mergeFunc == nil || bs.mergeSplitFunc == nil {
+			err = multierr.Append(err, fmt.Errorf("WithRequestBatchFuncs must be provided for the batcher applied to the request-based exporters"))
+		}
+
+		// Setting these callbacks to base exporter, because if they come in from options, these are only set to batch sender.
+		be.BatchMergeFunc = bs.mergeFunc
+		be.BatchMergeSplitfunc = bs.mergeSplitFunc
```
bogdandrutu, line 108: Confused about this, can you point to the code please?

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -93,6 +93,25 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
...
+
+		if !be.queueCfg.Enabled {
+			be.BatchSender = bs
```
bogdandrutu, line 111: Do you need to create `bs` at all if not used for `!be.queueCfg.Enabled`?

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,259 @@
...
+	"go.opentelemetry.io/collector/exporter/internal"
+)
+
+type batch struct {
+	ctx     context.Context
+	req     internal.Request
+	idxList []uint64
+}
+
+type Batcher struct {
```
bogdandrutu, line 24: Do we plan to soon delete BatchSender. I expect that to be the case after we have a "blocking 1 element queue"?

bogdandrutu: Do you see that we may need multiple implementations of the Batcher? If yes, we can probably use an interface?

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,259 @@
...
+	shutdownCh       chan bool
+	allocateWorkerCh chan bool
+
+	stopWG sync.WaitGroup
+}
+
+func NewBatcher(cfg exporterbatcher.Config, queue Queue[internal.Request],
+	maxWorkers int, exportFunc func(context.Context, internal.Request) error,
+	mergeFunc exporterbatcher.BatchMergeFunc[internal.Request],
+	mergeSplitFunc exporterbatcher.BatchMergeSplitFunc[internal.Request]) *Batcher {
```
bogdandrutu, line 47: @dmitryax do you remember why the merge and split are not part of the Request?

bogdandrutu (dismissed):

bogdandrutu (commented):
These are some high-level comments, also if you fix an issue use "Fixes #Number" and github will automatically close the issue.

exporter/internal/queue/batcher.go
dmitryax: Do you mean having them as methods? I don't remember precisely. Probably because it involves many `Request` objects.

exporter/exporterhelper/internal/base_exporter.go
sfc-gh-sili: This is an example: https://github.com/open-telemetry/opentelemetry-collector/blob/a7d019fa71199ede23838ba12ee8a29d5c7c27c3/exporter/exporterhelper/internal/batch_sender_test.go#L37

`WithRequestBatchFunc` is set to batcher but not to `base_exporter`.

exporter/exporterhelper/internal/base_exporter.go
sfc-gh-sili: The only reason this exist is to set up `batchMergeFunc`. `batchMergeFunc` comes in as a batcher option 

https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/base_exporter.go#L33

which has to be set on a batchSender. Maybe we should deprecate this interface now that a `batchsender` is not necessarily created. Let me work on a separate PR for this

exporter/internal/queue/batcher.go
sfc-gh-sili: I am leaning toward getting rid of `batch_sender` and implementing the blocking logic in as a blocking queue ([like this](https://github.com/open-telemetry/opentelemetry-collector/pull/11447/files#diff-cc61518e1acc5a429a7fc835d303a127d1f3f5160e2df552dd38766e6e1a9fc9)). That way `queue_sender` (or maybe we should call it `queue_batch_sender`) would handle all cases so we don't maintain two separate senders. I do find it slightly awkward that the blocking queue is holding `map[int]chan` in order to inform blocking senders about completion though. WDYT?

exporter/exporterhelper/internal/base_exporter.go
sfc-gh-sili: Let me clean up this API so we don't need this work around: https://github.com/open-telemetry/opentelemetry-collector/pull/11448

PR #11416, opened 2024-10-11
[docs/rfc] Define RFC merge process

mx-psi:
#### Description

Define RFC merge process in the `docs/rfc/README.md`. This is meant to be a tool for making decisions about hard stuff :)


mx-psi:

djaglowski (approved):

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+## Approval process
+
+An RFC is just like any other PR in this repository, with the exception of more stringent criteria
+for merging it.
+
+### Stakeholders
+
+To define merge criteria and voting, each RFC has a set of 'stakeholders'. All
+opentelemetry-collector approvers are considered stakeholders. Additional stakeholders (e.g.
+maintainers of related SIGs or experts) may be explicitly noted in the RFC.
```
songy23, line 40:
```suggestion
maintainers of related SIGs or code owners of the components) may be explicitly noted in the RFC.
```

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+1. *Quorum*: For an RFC to be mergeable, it needs to have at least **two approvals** from the
+   approvers set as well as approvals from any additional stakeholders.
+2. *Waiting period*: Maintainers need to announce their intent to merge the RFC with a Github
+   comment, and wait for at least **4 business days** after making the announcement to merge the
+   RFC.
+3. *Objections*: Objections should be communicated as a 'request changes' review. All objections
+   must be addressed or voted before merging an RFC. If unable to resolve an objection, any
+   maintainer may call for a vote to be made on the objection (see below). The result is binding.
+4. *Modifications*: Non-trivial modifications to an RFC reset the waiting period. RFC authors are
+   encouraged to re-request reviews if the RFC has been modified significantly.
```
songy23, line 55: I think in this case RFC authors are obligated, rather than encouraged, to re-request reviews. WDYT

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+We are purposefully light on the structure of RFCs, with the focus being the decision process and
+not the document itself. When in doubt, the [OTEP template][2] can be a good starting point.
+Regardless of the structure, the RFC should make clear what commitments are being made by the
+Collector SIG when merging it.
+
+## Announcement
+
+RFCs should be announced in a Collector SIG meeting and on the #otel-collector-dev Slack channel.
+
+## Approval process
```
TylerHelmuth, line 31: Do we need an amendment process as well? We have made minor tweaks to merged RFC documents in the past.

evan-bradley: There's a **Modifications to this document** section below that applies to this document and seems suitable, maybe that could apply to RFCs generally as well?

docs/rfcs/README.md
djaglowski: I think experts is a better standard. This allows code owners but also could be anyone else whose experience and knowledge are very relevant. For example, a RFC pertaining to semantic conventions may need buy in from contributors in a different part of the project.

songy23: Sounds reasonable to me, and looks like this just lists examples of stakeholders so lgtm as-is

dmitryax (approved):

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+
+### Merge rules
+
+We use a Lazy Consensus method with the following rules:
+
+1. *Quorum*: For an RFC to be mergeable, it needs to have at least **two approvals** from the
+   approvers set as well as approvals from any additional stakeholders.
+2. *Waiting period*: Maintainers need to announce their intent to merge the RFC with a Github
+   comment, and wait for at least **4 business days** after making the announcement to merge the
+   RFC.
```
bogdandrutu, line 50: Should we do this by applying a label? Easier for people to filter for this.

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+3. *Objections*: Objections should be communicated as a 'request changes' review. All objections
+   must be addressed or voted before merging an RFC. If unable to resolve an objection, any
+   maintainer may call for a vote to be made on the objection (see below). The result is binding.
```
bogdandrutu, line 53: Document that: Maintainers will be able to drop any `request changes` based on the vote results or consensus.

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+4. *Modifications*: Non-trivial modifications to an RFC reset the waiting period. RFC authors are
+   encouraged to re-request reviews if the RFC has been modified significantly.
+
+### Voting
+
+If there is no consensus on a particular issue, a vote may be called by any of the maintainers. The
+vote will be open to all stakeholders for at least **5 business days** or until at least one third
+of the stakeholders have voted, whichever comes last. The vote will be decided by simple majority,
```
bogdandrutu, line 61: Same, encourage to apply a label "voting-needed" to every PR/issue that needs this.

bogdandrutu (approved):

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+on the OpenTelemetry Collector and not on other parts of OpenTelemetry; if you have a cross-cutting
+proposal, file an [OTEP][1] instead.
 
-Accepted RFCs:
+# RFC Process
 
-- [OpenTelemetry Collector Processor Exploration](./processing.md)
+## Scope
+
+The RFC process is intended to be used for significant changes to the Collector. Major design
```
codeboten, line 11:
```suggestion
The Request For Comments (RFC) process is intended to be used for significant changes to the Collector. Major design
```

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+decisions, specially those that imply a change that is hard to reverse, should probably be
```
codeboten, line 12:
```suggestion
decisions, specially those that imply a change that is hard to reverse, should generally be
```

docs/rfcs/README.md
```diff
@@ -1,10 +1,73 @@
...
+opentelemetry-collector approvers are considered stakeholders. Additional stakeholders (e.g.
+maintainers of related SIGs or experts) may be explicitly noted in the RFC.
+
+### Merge rules
+
+We use a Lazy Consensus method with the following rules:
+
+1. *Quorum*: For an RFC to be mergeable, it needs to have at least **two approvals** from the
+   approvers set as well as approvals from any additional stakeholders.
+2. *Waiting period*: Maintainers need to announce their intent to merge the RFC with a Github
```
codeboten, line 48:
```suggestion
2. *Waiting period*: Maintainers need to announce their intent to merge the RFC with a GitHub
```

codeboten (approved):
Non-blocking suggestions

docs/rfcs/README.md
mx-psi: Changed "are encouraged to" by "must"

docs/rfcs/README.md
mx-psi: I modified the last section so that it also apples to "modifications to existing RFCs"

docs/rfcs/README.md
mx-psi: Added an `rfc:final-comment-period` label

docs/rfcs/README.md
mx-psi: Done!

docs/rfcs/README.md
mx-psi: Added `rfc:vote-needed` label

docs/rfcs/README.md
```diff
@@ -1,10 +1,75 @@
...
+
+### Stakeholders
+
+To define merge criteria and voting, each RFC has a set of 'stakeholders'. All
+opentelemetry-collector approvers are considered stakeholders. Additional stakeholders (e.g.
+maintainers of related SIGs or experts) may be explicitly noted in the RFC.
+
+### Merge rules
+
+We use a Lazy Consensus method with the following rules:
```
evan-bradley, line 44: Should we link to a definition of Lazy Consensus? From what I can tell, this term is taken from the Apache Foundation and is defined by them here: https://www.apache.org/foundation/glossary.html#LazyConsensus

It's not difficult to find the definition, and I think our rules stand on their own, but a link could provide some context.

docs/rfcs/README.md
```diff
@@ -1,10 +1,75 @@
...
+
+1. *Quorum*: For an RFC to be mergeable, it needs to have at least **two approvals** from the
+   approvers set as well as approvals from any additional stakeholders.
+2. *Waiting period*: Maintainers need to announce their intent to merge the RFC with a GitHub
+   comment. They will need to add a `rfc:final-comment-period` label to the PR and wait for at least
+   **4 business days** after making the announcement to merge the RFC.
+3. *Objections*: Objections should be communicated as a 'request changes' review. All objections
+   must be addressed or voted before merging an RFC. If unable to resolve an objection, any
```
evan-bradley, line 52:
```suggestion
   must be addressed before merging an RFC. If addressing an objection does not appear feasible, any
```
I think this makes the distinction between resolving an objection through voting vs. addressing it through other means a little clearer.

docs/rfcs/README.md
```diff
@@ -1,10 +1,75 @@
...
+   maintainer may call for a vote to be made on the objection (see below). This will be signified by
+   adding a `rfc:vote-needed` label to the PR. The voting result is binding and a maintainer can
+   drop any 'request changes' based on the vote results or consensus.
```
evan-bradley, line 55:
```suggestion
   drop any 'request changes' reviews based on the vote results or consensus.
```

docs/rfcs/README.md
```diff
@@ -1,10 +1,75 @@
...
+4. *Modifications*: Non-trivial modifications to an RFC reset the waiting period. RFC authors must
+   re-request reviews if the RFC has been modified significantly.
```
evan-bradley, line 57:
```suggestion
   re-request any approving, comment, or 'request changes' reviews if the RFC has been modified significantly.
```
My understanding is that we don't want to have approving reviews applied to an outdated version of the RFC, so I think we should be explicit on that here. For the sake of completeness, we can probably just enumerate the other types of reviews GitHub supports re-requesting a review on as well.

songy23 (approved):

docs/rfcs/README.md
mx-psi:
```suggestion
We use a [Lazy Consensus](https://www.apache.org/foundation/glossary.html#LazyConsensus) method with the following rules:
```

mx-psi: Sounds like a good link to put there, I actually took this [from the TC charter](https://github.com/open-telemetry/community/blob/5a8dca71fbdaf97a2c36f7ac4002f2b158456cc5/tech-committee-charter.md?plain=1#L96)

docs/rfcs/README.md
mx-psi: Makes sense, added!

PR #11448, opened 2024-10-15
[exporter] Disable the API to pass in configurations using a callback that operates on batch sender

sfc-gh-sili:
#### Description

As part of the effort to solve https://github.com/open-telemetry/opentelemetry-collector/issues/10368, we no longer guarantee to initialize a `batchSender` when `batcher` is enabled. Therefore, we would like to remove the interface to set `mergeFunc` and `mergeSplitFunc` as a callback that operates on `batchSender`. Instead, users should use the alternative `WithBatchFuncs` that is a callback that operates `baseExporter`.

Context: https://github.com/open-telemetry/opentelemetry-collector/pull/11414

#### Link to tracking issue


dmitryax:
Please add a `breaking` API changelog item

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -275,30 +275,14 @@ func WithCapabilities(capabilities consumer.Capabilities) Option {
 	}
 }
 
-// WithRequestBatchFuncs sets the functions for merging and splitting batches for an exporter built for custom request types.
-func WithRequestBatchFuncs(mf exporterbatcher.BatchMergeFunc[internal.Request], msf exporterbatcher.BatchMergeSplitFunc[internal.Request]) BatcherOption {
```
dmitryax, line 279: This is a significant cut of the functionality. There is no longer any way to use batching for custom requests. Are we planning to bring it back later? Can we make `WithBatchFuncs` exported now as an alternative?

bogdandrutu: > Are we planning to bring it back later? 

I think we will do by adding it to the `internal.Request`.

dmitryax: By adding required methods or as an optional interface?

sfc-gh-sili: > By adding required methods or as an optional interface?
It should be optional because batching is not necessarily enabled.

.chloggen/disable-batch-option.yaml
```diff
@@ -0,0 +1,25 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api]
```
bogdandrutu, line 25:
```suggestion
change_logs: [api]

```

exporter/exporterhelper/internal/base_exporter.go
bogdandrutu: They will be optional for the final user to implement. To instantiate the Batcher they will be required.

dmitryax: Ok, but for now, can we just export `WithBatchFuncs` to not remove the capability? Something
```
func WithBatchFuncs(mf exporterbatcher.BatchMergeFunc[Request], msf exporterbatcher.BatchMergeSplitFunc[Request]) Option {
	return internal.WithBatchFuncs(mf, msf)
}
```

bogdandrutu: I am confused, isn't the case that nobody should use this because `Request` is still internal?

bogdandrutu: Ignore, is exported in the helper.

bogdandrutu (approved):

exporter/exporterhelper/internal/base_exporter.go
bogdandrutu, line 279: Delete also `BatcherOption`

.chloggen/disable-batch-option.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Disables setting batch option to batch sender directly.
+
+# One or more tracking issues or pull requests related to the change
+issues: [10368]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
dmitryax, line 18: Please mention that `WithRequestBatchFuncs` BatcherOption is removed in favor of `WithBatchFuncs` Option

PR #11455, opened 2024-10-15
[chore] Remove unused/unnecessary queue.Sizer types

bogdandrutu:


exporter/exporterqueue/queue.go
```diff
@@ -70,14 +70,14 @@ type PersistentQueueSettings[T any] struct {
 // If cfg.StorageID is nil then it falls back to memory queue.
 // Experimental: This API is at the early stage of development and may change without backward compatibility
 // until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
-func NewPersistentQueueFactory[T itemsCounter](storageID *component.ID, factorySettings PersistentQueueSettings[T]) Factory[T] {
+func NewPersistentQueueFactory[T any](storageID *component.ID, factorySettings PersistentQueueSettings[T]) Factory[T] {
```
dmitryax, line 73: We still want to introduce other queue measurements going forward, right? So this would need to be brought back to an interface implementing `ItemsCount() int` which will be a breaking change technically...

exporter/internal/queue/queue.go
```diff
@@ -38,22 +38,11 @@ type Queue[T any] interface {
...
 	Sizeof(T) int64
 }
 
-// ItemsSizer is a Sizer implementation that returns the size of a queue element as the number of items it contains.
-type ItemsSizer[T itemsCounter] struct{}
-
-func (is *ItemsSizer[T]) Sizeof(el T) int64 {
-	return int64(el.ItemsCount())
-}
-
```
dmitryax, line 56: I understand it is not being used, but it may be seen as a good reference explaining why we need the `RequestSizer` at all.

exporter/exporterqueue/queue.go
bogdandrutu: We will accept a `Sizer` instead correct? We will pass the Request, so no need to enforce the request implements this API, the implementation can do what they want.

exporter/internal/queue/queue.go
bogdandrutu: It is still in tests... What is the point to be internal public, serves same reasons.

exporter/exporterqueue/queue.go
dmitryax: Ok. Sounds good

dmitryax (approved):

PR #11468, opened 2024-10-16
mdatagen: get package name from metadata

### LINKED ISSUE #11231: Allow `mdatagen` to generate packages with names other than `metadata`
In contrib, receivers will soon need to make shifts in the metrics they produce to align with Semantic Conventions. Depending on what exact changes are necessary, this is really hard to do with only one `metadata` package, especially if you want to not break existing usage.

I would like `mdatagen` to be able to generate packages with names other than `metadata`. This would allow receivers making the semconv transition to have an entirely separate yaml file and generated package with the semconv version of the produced metrics, and for example control the produced set of metrics via featuregate.

I already tried to do this all in one `metadata.yaml` and `metadata` package, but it was messy and in some cases had conflicts that were impossible to reconcile.

braydonk:
#### Description
I noticed a mistake in my previous PR open-telemetry/opentelemetry-collector-contrib#11232; some function calls did not pass the correct package name in (passing in "metadata" instead of the intended generated package name). This PR attempts to address the potential for this mistake to even occur by providing a wrapper `generateFile` function that automatically uses the generated package name from the metadata. The original version of the function that accepts a package name is intact for the templates that are going in the base package instead of the generated one.

#### Link to tracking issue
Bug originally from open-telemetry/opentelemetry-collector-contrib#11231  
Fixes open-telemetry/opentelemetry-collector-contrib#11469 

#### Testing
`make mdatagen-test`

Also added a custom generated package to the sample components, which would have caught the original bug.


braydonk:
I believe this change should not need a changelog. It is a fix for a bug that never made it into a release, and otherwise provides no user-visible changes.

braydonk:
Unstale, looks like this fell through the cracks before the original feature got merged. I'll introduce a changelog since this now fixes a released feature.

cmd/mdatagen/internal/sampleprocessor/doc.go
```diff
@@ -3,6 +3,7 @@
 
 // Generate a test metrics builder from a sample metrics set covering all configuration options.
 //go:generate mdatagen metadata.yaml
+//go:generate mdatagen metadata_custom.yaml
```
dmitryax, line 6: Let's avoid the duplication. Maybe having a custom package in the processor and leaving the receiver as is.

braydonk: I opted to remove duplication slightly differently in [f106762](https://github.com/open-telemetry/opentelemetry-collector/pull/11468/commits/f106762ff9919469a61dd49341e6c04c8957d63b).

I left the custom package in `samplereceiver`. This still ensures the extent of coverage:

* Generates the representative set of templates for a metadata package
* Demonstrates the capability of generating two separate packages with different names

The custom package in `sampleprocessor` was not adding anything to the above coverage by just doing it in `samplereceiver`, so I removed that. I think the current duplication is acceptable since it still provides the necessary coverage, but if you disagree I can change it.

cmd/mdatagen/internal/samplereceiver/metadata_custom.yaml
```diff
@@ -0,0 +1,201 @@
+# Sample metadata file with all available configurations for a receiver.
+
+type: sample
+scope_name: go.opentelemetry.io/collector/internal/receiver/samplereceiver
+github_project: open-telemetry/opentelemetry-collector
+
+generated_package_name: custom
```
dmitryax, line 7: `internal/metadata` package is abandoned now and should be removed, right?

cmd/mdatagen/internal/sampleprocessor/doc.go
dmitryax: I still don't understand why we need two packages for the same receiver. This will be confusing to maintain going forward. Also, we need to show how mdatagen is supposed to be used. Having two yaml files and two packages for the same module is not something we should recommend. Can we just add `generated_package_name` to the existing yaml for the receiver and move the package to a custom path?

cmd/mdatagen/internal/samplereceiver/metadata_custom.yaml
dmitryax: _Commented too early under assumption that https://github.com/open-telemetry/opentelemetry-collector/pull/11468/files#diff-7d3195204f99934e80e03730709378a3054a1c056d953091368ad2232d282442R5 line is removed_

cmd/mdatagen/internal/sampleprocessor/doc.go
braydonk: Having two yaml files and two packages is the reason the custom package name feature was added in the first place. It was the only reasonable way I could see for a receiver to have a featuregated semantic convention transition, as doing it within one metadata package seems infeasible based on my attempts. We wouldn't want two packages to exist forever, but for the lifetime of the featuregate (which for a semconv transition will likely be a longer horizon) two packages would need to coexist.

braydonk:
Not stale, still relevant for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35325

I'll wait until knowing the change isn't rejected to fix the conflict

braydonk:
Same still applies, restating from above.

Not stale, still relevant for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35325

I'll wait until knowing the change isn't rejected to fix the conflict.

braydonk:
Still relevant for now. We may decide to adopt https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/15995, in which case I will close this and make a PR that rips out the custom name functionality, but otherwise this PR will be necessary to fix it as the feature is technically broken at the moment.

PR #11477, opened 2024-10-17
Hide profiles behind a feature gate

dmathieu:
#### Description

This hides profiles support behind a feature gate, so folks have to enable them explicitly to be able to use them.

### DIFF
--- a/.chloggen/profiles-featuregate.yaml
+++ b/.chloggen/profiles-featuregate.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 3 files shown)


mx-psi (approved):

bogdandrutu (approved):

atoulme:
Is it time to move this feature gate to beta stage? Can it be tracked with an issue otherwise?

dmathieu:
That's dependent on the status of the signal on OTLP, which is not stable and still has upcoming breaking changes.
Moving to beta with breaking changes seems wrong.

mx-psi:
> Is it time to move this feature gate to beta stage? Can it be tracked with an issue otherwise?

I think having a tracking issue makes sense but I agree with Damien that it is too early

PR #11482, opened 2024-10-18
[chore][VERSIONING.md] Document approach to optional experimental interfaces

mx-psi:
#### Description

Documents the approach we have taken for adding experimental functionality to interfaces. We take this approach with:

- `receiver.Factory`, may implement the [`receiverprofiles.Factory`](https://pkg.go.dev/go.opentelemetry.io/collector/receiver/receiverprofiles#Factory) optional interface (similar for other pipeline components)
- `component.Host` may implement the [`componentstatus.Reporter`](https://pkg.go.dev/go.opentelemetry.io/collector/component/componentstatus#Reporter) optional interface.
- `extension.Extension` may implement the optional interfaces in [`extensioncapabilities`](https://pkg.go.dev/go.opentelemetry.io/collector/extension/extensioncapabilities).

In these cases we reserve the right to change the optional, experimental parts of the underlying struct we return. You can also check whether you are using experimental functionality by checking your `go.mod` and seeing if you use one of these experimental modules.


VERSIONING.md
```diff
@@ -62,6 +62,9 @@ Unless otherwise specified in the documentation, the following may change in any
 * **Protocol compatibility**. Changing the default minimum version of a supported protocol (e.g. TLS) or dropping support for protocols when there are security concerns is not considered a breaking change.
 * **Dependency updates**. Updating dependencies is not considered a breaking change except when their types are part of the
 public API or the update may change the behavior of applications in an incompatible way.
+* **Underlying type for interfaces**. If a struct exported as an interface has an experimental
+method, this method may change or be removed in a minor version. The method will be published in an
+optional interface under an experimental module to signal it is experimental.
```
dmathieu, line 67: Should the method also be documented as experimental?

mx-psi: If the method is in an experimental module (one released under v0.x), do you think this is needed?

dmathieu: Ah, sorry, I thought you were allowing struct methods to be defined in stable modules, as long as they were defined in experimental interfaces.

mx-psi: Sorry, still confused :sweat_smile: To make it more clear: let's talk about the case of the `otlpreceiver.NewFactory` function. 

The `NewFactory` function returns `receiver.Factory`. The underlying struct that is returned by this function also implements the `receiverprofiles.Factory` interface. I want to make sure we write down that it is fine to change the extra methods in `receiverprofiles.Factory` (i.e. `CreateProfiles`) in a minor version. How can we best write this?

Right now, since `receiverprofiles.Factory` is in a different module it can remain `0.x` while the `receiver` module goes 1.x, and this rule allows changes in `CreateProfiles`. Is this okay? I am not sure in this question:

> Should the method also be documented as experimental?

where that method would be documented as experimental

dmathieu: Ah my bad. Yes, your comment resolves my question. Thanks.

dmathieu (approved):

Issue #7051, opened 2023-01-27
Investigate if golang generics help with interface duplications in the consumer package


bogdandrutu:

evan-bradley:
I've looked into this, and generics could be used to reduce a lot of duplication, but there would be some tradeoffs.

Pretty much all of the code could be deduplicated if we created a `consumer.Consumer[S any]` interface with a type parameter for a specific signal, and a generic `Consume` method to consume that signal. I think it would then make sense to instantiate the type for each signal to make it easy to use. For traces, this would be:

```golang
type Traces = Consumer[ptrace.Traces]

type ConsumeTracesFunc = ConsumeFunc[ptrace.Traces]

var NewTraces = NewConsumer[ptrace.Traces] 
```

The downside of this is that each signal's consumer interface becomes mutually exclusive for a struct to implement, since they all implement `Consume` and Go doesn't support function overloading. So helpers like `consumertest.Consumer` would need to be specific for a signal instead of supporting all signals. This would also expose the generic interfaces from the `consumer` package, which causes the package to have a larger API surface.

I'm unable to find any other way to use generics that meaningfully cuts down on duplication, in part because each consumer interface is only ~20 SLOC.

codeboten:
Can this issue be closed then? Or is there a follow up to document this anywhere?

evan-bradley:
Sorry, I completely missed the notification for this.

I'm fine closing this issue. I can't think of any good place to document this, but if there's somewhere you think would make sense, I'd be happy to do it.

codeboten:
Based on @evan-bradley's findings here, I think we're good to close the issue.

PR #11490, opened 2024-10-18
Implement Backoff in a generic way

bogdandrutu:


consumer/consumer.go
```diff
@@ -4,11 +4,14 @@
 package consumer // import "go.opentelemetry.io/collector/consumer"
 
 import (
+	"context"
 	"errors"
 
 	"go.opentelemetry.io/collector/consumer/internal"
 )
 
+type ConsumeFunc[T any] func(context.Context, T) error
```
TylerHelmuth, line 13: Are concerns from https://github.com/open-telemetry/opentelemetry-collector/issues/7051#issuecomment-1420818111 still relevant with this design?

mx-psi (commented):
Should this block #11492?

sfc-gh-bdrutu:
> Should this block https://github.com/open-telemetry/opentelemetry-collector/pull/11492?

I think my change is not a breaking change, unless I am missing something. Would be good to understand if you would call this a breaking change, and how would I validate it in the future.

consumer/consumer.go
sfc-gh-bdrutu: No, because we don't change the interface (you can see that I did not change any consumer). Also, if I understand correctly my change is backwards compatible (not breaking change) and equivalent with the previous code, except I extract the common ConsumeFunc. This only hints users that they can implement this instead of every function if their logic is signal agnostic.

sfc-gh-bdrutu:
@mx-psi I don't mind waiting for this small change (adding the generic helper func) before stabilizing consumer to be 100% safe. I want first feedback if this improves our story.

PR #11494, opened 2024-10-21
[debugexporter] Add support for profiles signal

### LINKED ISSUE #11155: [exporter/debug] Add profiling support

julianocosta89:
#### Description

This PR adds Profiles support to the `debugexporter` component.

#### Link to tracking issue
Fixes #11155.

#### Testing
Added tests to all modified modules:
- `exporter/debugexporter/exporter_test.go`
- `exporter/debugexporter/factory_test.go`
- `exporter/debugexporter/internal/normal/profiles_test.go`


exporter/debugexporter/factory.go
```diff
@@ -38,6 +40,14 @@ func NewFactory() exporter.Factory {
 	)
 }
 
+func NewFactoryProfiles() exporterprofiles.Factory {
```
bogdandrutu, line 43: We should not put this contract in public.

/cc @mx-psi @dmathieu

dmathieu: Is this method even necessary? In the OTLP exporter, we enable profiles within `NewFactory`.

https://github.com/open-telemetry/opentelemetry-collector/blob/403c782d50c630ad945feebe8b26e1f656ccd3c1/exporter/otlpexporter/factory.go#L23-L33

mx-psi: Agreed, we can do the same as in the OTLP receiver

cmd/otelcorecol/go.mod
```diff
@@ -95,7 +95,7 @@ require (
 	go.opentelemetry.io/collector/connector/connectorprofiles v0.111.0 // indirect
 	go.opentelemetry.io/collector/connector/connectortest v0.0.0-20241021181817-007f06b7c4a8 // indirect
 	go.opentelemetry.io/collector/consumer v0.111.1-0.20241021181817-007f06b7c4a8 // indirect
-	go.opentelemetry.io/collector/consumer/consumererror v0.0.0-20241021181817-007f06b7c4a8 // indirect
+	go.opentelemetry.io/collector/consumer/consumererror v0.111.0 // indirect
```
julianocosta89, line 98: I don't think I should touch this file, but `make gotidy` was modifying it and all tests were failing without it.

exporter/debugexporter/factory.go
```diff
@@ -28,13 +30,14 @@ const (
 )
 
 // NewFactory creates a factory for Debug exporter
-func NewFactory() exporter.Factory {
-	return exporter.NewFactory(
+func NewFactory() exporterprofiles.Factory {
```
bogdandrutu, line 33: Same comment you MUST return `exporter.Factory` here.

julianocosta89: Thanks again @bogdandrutu! Fixed ✅

dmathieu (approved):

exporter/debugexporter/exporter.go
```diff
@@ -95,3 +101,20 @@ func (s *debugExporter) pushLogs(_ context.Context, ld plog.Logs) error {
...
+
+	if s.verbosity == configtelemetry.LevelBasic {
+		return nil
+	}
+
+	buf, err := s.profilesMarshaler.MarshalProfiles(pd)
+	if err != nil {
+		return err
+	}
+	s.logger.Info(string(buf))
```
mx-psi, line 118: Add test for this?

exporter/debugexporter/go.mod
```diff
@@ -9,6 +9,8 @@ require (
 	go.opentelemetry.io/collector/confmap v1.18.0
 	go.opentelemetry.io/collector/consumer v0.112.0
 	go.opentelemetry.io/collector/exporter v0.112.0
+	go.opentelemetry.io/collector/exporter/exporterhelper/exporterhelperprofiles v0.0.0-00010101000000-000000000000
```
mx-psi, line 12: I think we can use v0.112.0 here now

exporter/debugexporter/exporter.go
julianocosta89: Hey @mx-psi the test actually follows the same structure for the other signals.
I've added 2 tests in the `exporter_test`:
- `TestErrors`
- `TestProfilesNoErrors`

Those are exactly the same tests all other signals have.

Let me know if you'd like me to add anything different for Profiles.

mx-psi: Makes sense, thanks. I think we should increase coverage (and test that we are actually logging something), but we don't need to do it now, we can do it together with all signals

mx-psi (approved):
LGTM. I know there is an open question regarding the format, I think we can go with this and iterate since we can change the output format freely

mx-psi:
@bogdandrutu Anything left from your side? I intend to merge this by Friday EU morning

PR #11503, opened 2024-10-21
[chore] Update additional modules to psuedo versions

evan-bradley:
#### Description

Convert additional modules to use pseudo-version numbers. In cases where the module already has a released version, upgrade to pseudo versions to prevent ambiguous import errors (e.g. processor and processortest).

### DIFF
--- a/client/go.mod
+++ b/client/go.mod
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	github.com/stretchr/testify v1.9.0
-	go.opentelemetry.io/collector/consumer v0.111.0
+	go.opentelemetry.io/collector/consumer v0.0.0-20241021181817-007f06b7c4a8
 	go.opentelemetry.io/collector/pdata v1.17.0
 	go.uber.org/goleak v1.3.0
 )
--- a/client/go.mod
+++ b/client/go.mod
@@ -4,7 +4,7 @@ go 1.22.0
 
 require (
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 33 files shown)


bogdandrutu:
> Running make update-otel in contrib produces errors like unknown revision pipeline/pipelineprofiles/v0.111.0. I think the larger-than-usual number of modules we're creating this release is making this issue stand out.

After we do the release this is not a problem correct? So then why not doing the release?

evan-bradley:
The [first step](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#releasing-opentelemetry-collector) in the releasing document says to run this to check for errors. You're right that this shouldn't be a problem after the release, but this is my first release, so I'm looking to play it safe and double-check before releasing core.

bogdandrutu:
I understand, but I am confused about changes to `go.opentelemetry.io/collector/consumer`

evan-bradley:
`go.opentelemetry.io/collector/consumer/consumererror` was made its own module, but `go.opentelemetry.io/collector/consumer@v0.111.0` includes `go.opentelemetry.io/collector/consumer/consumererror` as a package. `go.opentelemetry.io/collector/consumer/consumererror` then becomes an ambiguous import because it can come from the package inside of consumer or from its own module. Therefore I had to update consumer to a version that doesn't include the `go.opentelemetry.io/collector/consumer/consumererror` package.

sfc-gh-bdrutu (approved):

bogdandrutu (approved):

PR #11519, opened 2024-10-23
[chore] Split mdatagen configs in different files

bogdandrutu:


evan-bradley (approved):

cmd/mdatagen/internal/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+	var errs error
+	if err := md.validateType(); err != nil {
+		errs = errors.Join(errs, err)
+	}
+
+	if md.Parent != "" {
+		if md.Status != nil {
+			// status is not required for subcomponents.
+			errs = errors.Join(errs, errors.New("status must be empty for subcomponents"))
+		}
```
mx-psi, line 55: It looks like this error condition was not present on the original file, and it is making the contrib release fail. Part of the solution is #11167, but could we also have these kinds of changes listed in the PR description (and maybe changelog as well)? It was a bit hard for me to debug this (since the PR title talks about just moving files around, I discarded this PR as the cause of the error initially)

atoulme: I vote to reset back to the original logic, and have an issue if we want to evolve the logic.

PR #11532, opened 2024-10-23
[exporterqueue] Bare minimum frame of queue batcher + unit test.

sfc-gh-sili:
#### Description

This PR is a bare minimum implementation of a component called queue batcher. On completion, this component will replace `consumers` in `queue_sender`, and thus moving queue-batch from a pulling model instead of pushing model.

Limitations of the current code
* This implements only the case where batching is disabled, which means no merge of splitting of requests + no timeout flushing.
* This implementation does not enforce an upper bound on concurrency

All these code paths are marked as panic currently, and they will be replaced with actual implementation in coming PRs. This PR is split from https://github.com/open-telemetry/opentelemetry-collector/pull/11507 for easier review.

Design doc:

#### Link to tracking issue

### DIFF
--- a/exporter/internal/queue/batcher.go
+++ b/exporter/internal/queue/batcher.go
@@ -73,12 +73,6 @@ func (qb *BaseBatcher) flushAsync(batchToFlush batch) {
 
 // Shutdown ensures that queue and all Batcher are stopped.
 func (qb *BaseBatcher) Shutdown(ctx context.Context) error {
-	// TODO: queue shutdown is done here to keep the behavior similar to queue consumer.
-	// However batcher should not be responsible for shutting down the queue. Move this up to
-	// queue sender once queue consumer is cleaned up.
-	if err := qb.queue.Shutdown(ctx); err != nil {
-		return err
-	}
 	qb.stopWG.Wait()
 	return nil
 }
--- a/exporter/internal/queue/batcher.go
+++ b/exporter/internal/queue/batcher.go
@@ -1,0 +1,141 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package queue // import "go.opentelemetry.io/collector/exporter/internal/queue"
+
+import (
+	"context"
+	"math"
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+	"go.opentelemetry.io/collector/exporter/internal"
+)
+
+type batch struct {
+	ctx     context.Context
+	req     internal.Request
+	idxList []uint64
+}
+
+type Batcher struct {
+	batchCfg exporterbatcher.Config
+
+	queue      Queue[internal.Request]
+	maxWorkers int
+
+	exportFunc func(context.Context, internal.Request) error
+
+	readingBatch *batch
+	timer        *time.Timer
+	shutdownCh   chan bool
+
+	stopWG sync.WaitGroup
+}
+
+func NewBatcher(batchCfg exporterbatcher.Config, queue Queue[internal.Request],
+	maxWorkers int, exportFunc func(context.Context, internal.Request) error) *Batcher {
+	return &Batcher{
+		batchCfg:   batchCfg,
+		queue:      queue,
+		maxWorkers: maxWorkers,
+		exportFunc: exportFunc,
+		stopWG:     sync.WaitGroup{},
+		shutdownCh: make(chan bool, 1),
+	}
+}
+
+// If preconditions pass, flush() take an item from the head of batch list and exports it.
+func (qb *Batcher) flush(batchToFlush batch) {
+	err := qb.exportFunc(batchToFlush.ctx, batchToFlush.req)
+	for _, idx := range batchToFlush.idxList {
+		qb.queue.OnProcessingFinished(idx, err)
+	}
+}
+
+// allocateFlusher() starts a goroutine that calls flushIfNecessary(). It blocks until a worker is available.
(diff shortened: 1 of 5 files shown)


exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,141 @@
...
+
+type Batcher struct {
+	batchCfg exporterbatcher.Config
+
+	queue      Queue[internal.Request]
+	maxWorkers int
+
+	exportFunc func(context.Context, internal.Request) error
+
+	readingBatch *batch
```
dmitryax, line 31: is this a batch that is currently being built? Not sure why it's `reading`... a bit confusing. `currentBatch` maybe?

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,141 @@
...
+				}
+
+				return
+			}
+			if !qb.batchCfg.Enabled {
+				qb.readingBatch = &batch{
+					req:     req,
+					ctx:     context.Background(),
+					idxList: []uint64{idx}}
+				qb.allocateFlusher(*qb.readingBatch)
```
dmitryax, line 110: `allocateFlusher` sounds like we just allocate but don't flush. Should it be `flush` or `flushAsync`?

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,141 @@
...
+	qb.stopWG.Add(1)
+	go func() {
+		defer qb.stopWG.Done()
+		for {
+			idx, _, req, ok := qb.queue.Read(context.Background())
+
+			if !ok {
+				qb.shutdownCh <- true
+				if qb.readingBatch != nil {
+					panic("batching is supported yet so reading batch should always be nil")
```
dmitryax, line 100:
```suggestion
					panic("batching isn't supported yet so reading batch should always be nil")
```

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,141 @@
...
+				qb.allocateFlusher(*qb.readingBatch)
+				qb.readingBatch = nil
+			} else {
+				panic("not implemented")
+			}
+		}
+	}()
+
+	qb.stopWG.Add(1)
+	go func() {
```
dmitryax, line 119: These two main go routines should probably be better documented. Maybe moving them to functions or having comments before each of them explaining what they are responsible for

sfc-gh-sili: That's a good point. Thanks for the advice!

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,145 @@
...
+	panic("not implemented")
+}
+
+// Start ensures that queue and all consumers are started.
+func (qb *Batcher) Start(ctx context.Context, host component.Host) error {
+	if err := qb.queue.Start(ctx, host); err != nil {
+		return err
+	}
+
+	if qb.batchCfg.Enabled {
```
bogdandrutu, line 78: I think would be simpler if Batcher is an interface, and the "not enabled" and "enabled" are 2 different implementations.

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,145 @@
...
+	idxList []uint64
+}
+
+type Batcher struct {
+	batchCfg exporterbatcher.Config
+
+	queue      Queue[internal.Request]
+	maxWorkers int
+
+	exportFunc func(context.Context, internal.Request) error
```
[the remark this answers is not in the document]
sfc-gh-sili: You are right.
This was modeled after `consumeFunc` used in `consumer` but it seems having `req.Export()` here is enough for the use case.

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,145 @@
...
+
+// flushAsync() starts a goroutine that calls flushIfNecessary(). It blocks until a worker is available.
+func (qb *Batcher) flushAsync(batchToFlush batch) {
+	// maxWorker = 0 means we don't limit the number of flushers.
+	if qb.maxWorkers == 0 {
+		qb.stopWG.Add(1)
+		go func() {
+			qb.flush(batchToFlush)
+			qb.stopWG.Done()
+		}()
```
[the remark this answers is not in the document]
sfc-gh-sili: Thanks for the advice!

exporter/internal/queue/batcher.go
```diff
@@ -0,0 +1,145 @@
...
+		return
+	}
+	panic("not implemented")
+}
+
+// Start ensures that queue and all consumers are started.
+func (qb *Batcher) Start(ctx context.Context, host component.Host) error {
+	if err := qb.queue.Start(ctx, host); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
sfc-gh-sili: That's a good point. Let me keep them here for now because this is consistent with `queue.consumer`, but I'll leave a note to move this up to queue sender once `queue.consumer` is cleaned up.

exporter/internal/queue/batcher.go
sfc-gh-sili: That's a good idea.
Implemented a special-case batcher called `DisabledBatcher` to handle the "not enabled" case. Let me know what you think of it.

exporter/internal/queue/batcher.go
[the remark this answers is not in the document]
sfc-gh-sili: https://github.com/open-telemetry/opentelemetry-collector/pull/11534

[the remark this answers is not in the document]
dmitryax: PR is merged. We can remove this I guess

dmitryax (approved):

PR #11558, opened 2024-10-29
Fix calling count methods on nil pdata and pprofile structs

dmathieu:
#### Description

I am working on a profiles receiver, and while writing tests for it, I noticed I must set the signal struct to something not nil, I'd get panics due to `getOrig()` being nil.

This change allows calling count methods for nil signal structs, so we can look for the count of data being sent as a nil-comparison.

### DIFF
--- a/.chloggen/pdata-nil-count.yaml
+++ b/.chloggen/pdata-nil-count.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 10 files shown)


bogdandrutu (commented):
https://github.com/open-telemetry/opentelemetry-collector/pull/11558/files#diff-6b83ec9e9c41d5d8ccb8bb01350fdf83040233b9e0431352365823364e012a16R12

dmathieu:
This use case is not for runtime, and it's not due to not using `NewXX`, but in tests and due to a test where `ConsumeProfiles` is never being called. So the variable is never initialized.
With this change, the structs also don't become usable at runtime. They would still panic if the public method relying on `getOrig` is called. Only the counts would stop triggering a panic.

I'll try to go around this, or remove that test on our end.

PR #11559, opened 2024-10-29
[chore] Add issue template for RFC votes

### LINKED ISSUE #11552: Add issue template for voting
We want to vote on a couple of things, we now have a [voting process for design documents](https://github.com/open-telemetry/opentelemetry-collector/tree/main/docs/rfcs#voting), to make the process easier I want to create a template with instructions for voting that can be used for public voting.

mx-psi:
#### Description

Adds issue template for making votes.

#### Link to tracking issue
Fixes #11552

### DIFF
--- a/.github/ISSUE_TEMPLATE/vote.md
+++ b/.github/ISSUE_TEMPLATE/vote.md
@@ -30,7 +30,7 @@ For any stakeholder team, we consider the people that were part of the team at t
 
 ### Result
 
-The vote is in progress.
+The vote is in progress. A minimum of X votes is required to select an option.
 
 <!--Uncomment the sentence below when the vote has been closed and the result is known.-->
 <!--The vote has been closed. X stakeholders voted out of a total of Y. **Option N** has been selected with Z votes from the stakeholders.-->
--- a/.github/ISSUE_TEMPLATE/vote.md
+++ b/.github/ISSUE_TEMPLATE/vote.md
@@ -10,13 +10,15 @@ A [vote](https://github.com/open-telemetry/opentelemetry-collector/blob/main/doc
 
 ### Stakeholders
 
-Any person in the community may vote. Votes of the stakeholders are binding. The stakeholders as defined in the RFC are:
+Any person in the community may vote. Votes of the stakeholders are binding. As defined in the RFC, there are N stakeholders for this RFC divided as follows:
 
-| Stakeholders |
-|-------------|
-| @open-telemetry/collector-approvers |
+| Stakeholders                        | As defined on date | Number of people |
+|-------------------------------------|--------------------|------------------|
+| @open-telemetry/collector-approvers | xxxx-xx-xx         | N                |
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/vote.md
```diff
@@ -0,0 +1,57 @@
...
+
+A [vote](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/README.md#voting) has been called for RFC #XXXX following the RFC process.
+
+### Stakeholders
+
+Any person in the community may vote. Votes of the stakeholders are binding. The stakeholders as defined in the RFC are:
+
+| Stakeholders |
+|-------------|
+| @open-telemetry/collector-approvers |
```
djaglowski, line 17: It may be helpful for onlookers if we explicitly state the current number of stakeholders and the number of votes required to pass.

Related: There's an edge case where membership in the approvers group changes while the vote is open. Should we lock these numbers down when the vote is opened, or allow them to change until the vote is closed?

djaglowski (approved):

.github/ISSUE_TEMPLATE/vote.md
mx-psi: > Should we lock these numbers down when the vote is opened, or allow them to change until the vote is closed?

This is interesting. I added a sentence noting that we lock these numbers down when the vote starts, since I feel like this is the more practical option, but if you feel differently please let me know, I am undecided as to what option is best

mx-psi:
@open-telemetry/collector-approvers I will merge this on Monday. This is just a way to make the RFC decision process easier to follow with a template, so I don't think there would be anything controversial.

Note that after discussing with Dan I added the following clarification:
> For any stakeholder team, we consider the people that were part of the team at the time the vote is called

.github/ISSUE_TEMPLATE/vote.md
```diff
@@ -0,0 +1,59 @@
...
+
+- [ ] Announce the vote in the #otel-collector-dev CNCF Slack channel.
+- [ ] Add an entry to announce the vote in the next Collector SIG meeting.
+
+For closing the vote:
+
+- [ ] At least five business days have passed since the vote was announced.
+- [ ] At least one third of the stakeholders have voted.
+- [ ] The "Voting options" and "Result" sections have been updated reflecting the votes casted.
+- [ ] The RFC has been updated reflecting the votes casted.
```
TylerHelmuth, line 59: Since we are posting in `#otel-collector-dev` and welcoming votes from non-stakeholders I think we should define something about how we consider those votes.  

While ultimately the stakeholder votes are what determine a "accept" or not I think we should call out that stakeholders should consider community votes when casting their vote.

mx-psi: I added the folllowing:
> Stakeholders are encouraged to consider the views from the wider community when casting their vote

TylerHelmuth (approved):

bogdandrutu (approved):

PR #11575, opened 2024-10-30
updated grpc config to use DialContext

### LINKED ISSUE #11537: OTLP Exporter will not respect domain based NO_PROXY setting due to grpc bug
**Describe the bug**

There is a bug in go-grpc that prevents the NO_PROXY setting from being properly applied based on domain name values in the NO_PROXY string. See issue [here](https://github.com/grpc/grpc-go/issues/7779). At this point it seems that they're aware of the issue and working on a fix but just wanted to socialize this to see if we might need a temp patch. Note: this bug does not apply to localhost based NO_PROXY settings

**Proposed Patch:**
We should likely use DialContext instead of NewClient in the configgrpc package until this is resolved. 

**Steps to reproduce**

Set an HTTP/HTTPS proxy and use the otlp exporter with a domain name that matches an entry in your NO_PROXY

Proxy setting to remain unset
**What did you see instead?**

HTTPS_PROXY applied
**What version did you use?**

v0.112.0

akats7:
#### Description

This is related #11537 and [this grpc issue](https://github.com/grpc/grpc-go/issues/7779),  to reiterate there is a bug in the grpc-go NewClient that makes the way the hostname is resolved incompatible with the way proxy setting are applied. Due to this domain based NO_PROXY setting do not work for grpc connections. They are working on a fix in the grpc library but until then it seems like it might be a good idea to revert to using DialContext instead of NewClient. 

If there's a workaround for this that anyone is aware of that could be suitable as well, the most clear one of using passthrough doesn't work since we sanitize the endpoint.

#### Link to tracking issue
Fixes #11537 

#### Testing

I added some logging where the proxy setting are evaluated to show the behavior.

**DialContext**
______
InMatch
host :  otel*******<DOMAIN-NAME>
m.host :  <DOMAIN-NAME> 
hasSuffix :  true
m.matchHost :  true
______

**NewClient**
______
InMatch
host :  10.*.*.*.*
m.host :  <DOMAIN-NAME> 
hasSuffix :  false
m.matchHost :  false
______


akats7:
> Does this need a changelog? (seems yes)

yep added

config/configgrpc/configgrpc.go
```diff
@@ -261,7 +261,7 @@ func (gcs *ClientConfig) ToClientConn(
 	if err != nil {
 		return nil, err
 	}
-	return grpc.NewClient(gcs.sanitizedEndpoint(), grpcOpts...)
+	return grpc.DialContext(ctx, gcs.sanitizedEndpoint(), grpcOpts...)
```
songy23, line 264: Need to suppress the linter warning https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11598212701/job/32293604408?pr=11575

akats7: @songy23 do you know where that should be configured

songy23, line 264:
```suggestion
	//lint:ignore SA1019 see https://github.com/open-telemetry/opentelemetry-collector/pull/11575
	return grpc.DialContext(ctx, gcs.sanitizedEndpoint(), grpcOpts...)
```

songy23 (approved):

akats7:
@songy23 anything else needed to get this merged, just want to ensure it’s included in the next release

songy23:
@akats7 I don't have the access to merge PRs unfortunately

songy23:
@open-telemetry/collector-maintainers Are we good with merging this? It's a simple 1-line bug fix

bogdandrutu:
> It's a simple 1-line bug fix 

Famous last words :)) How confident are you in the fix? I have not had time to read about.

songy23:
Context are in https://github.com/grpc/grpc-go/issues/7556 & https://github.com/grpc/grpc-go/issues/7779. gRPC maintainers acked there's an issue with proxy in `grpc.NewClient` which makes it incompatible with `grpc.DialContext`. I did not have time to reproduce it myself but @akats7 has tried and left the findings in the PR description.

PR #11578, opened 2024-10-30
[service] fix bug with inconsistent resources

codeboten:
The tracer and logger provider were instantiating different resources objects that didn't have a `service.instance.id` attribute. This change fixes that by instantiating the SDK in the service and passing it to the factory via the telemetry.Settings struct.

This also removes the duplicate code to instantiate the SDK multiple times, which will be useful when we move to instantiating MeterProvider via the config SDK. This bug is blocking the change to bump up the dependency on the config package.

There are a few alternatives that were considered:
1. could set the resource's service.instance.id on the config object instead. this seemed messy as it would be editing the config struct and the instantiation of the SDK would remain duplicated.
2. update the factory to pass in a resource object option, i didn't want to update the NewFactory func.
3. update the CreateLogger, CreateTracerProvider to receive either a resource or sdk parameter, both of those seemed incorrect as well.


service/telemetry/factory.go
```diff
@@ -68,11 +69,11 @@ func NewFactory() Factory {
 	return newFactory(createDefaultConfig,
 		withLogger(func(ctx context.Context, set Settings, cfg component.Config) (*zap.Logger, log.LoggerProvider, error) {
 			c := *cfg.(*Config)
-			return newLogger(ctx, set, c)
+			return newLogger(set, c)
 		}),
-		withTracerProvider(func(ctx context.Context, set Settings, cfg component.Config) (trace.TracerProvider, error) {
+		withTracerProvider(func(_ context.Context, set Settings, cfg component.Config) (trace.TracerProvider, error) {
```
TylerHelmuth, line 74: can the `ctx` in the `withLogger` function also be `_`?

TylerHelmuth (approved):

service/telemetry/factory.go
codeboten: yup, this has been fixed

songy23 (approved):

PR #11580, opened 2024-10-31
[chore] [exporterqueue] Add splitting to the experimental pull-based batcher

sfc-gh-sili:
#### Description

This PR follows https://github.com/open-telemetry/opentelemetry-collector/pull/11546 and add support for splitting (i.e. support setting a maximum request size)

Design doc:

#### Link to tracking issue

### DIFF
--- a/exporter/internal/queue/default_batcher.go
+++ b/exporter/internal/queue/default_batcher.go
@@ -59,8 +59,8 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
 					continue
 				}
 
-				if reqList[0].ItemsCount() >= qb.batchCfg.MinSizeItems {
-					// If the incoming request is large enough that we decide to flush immediately, we flush all parts.
+				// If there was a split, we flush everything immediately.
+				if reqList[0].ItemsCount() >= qb.batchCfg.MinSizeItems || len(reqList) > 1 {
 					qb.currentBatch = nil
 					qb.currentBatchMu.Unlock()
 					for i := 0; i < len(reqList); i++ {
--- a/exporter/internal/queue/default_batcher.go
+++ b/exporter/internal/queue/default_batcher.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"go.opentelemetry.io/collector/component"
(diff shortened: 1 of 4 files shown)


exporter/internal/queue/default_batcher.go
```diff
@@ -42,35 +43,73 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
...
 
-			if qb.currentBatch.req.ItemsCount() > qb.batchCfg.MinSizeItems {
-				batchToFlush := *qb.currentBatch
-				qb.currentBatch = nil
-				qb.currentBatchMu.Unlock()
-
-				// flushAsync() blocks until successfully started a goroutine for flushing.
-				qb.flushAsync(batchToFlush)
-				qb.resetTimer()
+				if reqList[0].ItemsCount() >= qb.batchCfg.MinSizeItems {
```
dmitryax, line 63: Should we also do it if `len(reqList) > 1`?

exporter/internal/queue/default_batcher.go
```diff
@@ -42,35 +43,73 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
...
+						qb.flushAsync(batch{
+							req:     reqList[i],
+							ctx:     ctx,
+							idxList: []uint64{idx}})
+						// TODO: handle partial failure
+					}
+					qb.resetTimer()
+				} else {
+					qb.currentBatch = &batch{
+						req:     reqList[0],
```
dmitryax, line 77: It looks like we should. Otherwise, we just ignore other items potentially. This situation shouldn't be allowed typically. However, I think we should either make a best effort here or throw a warning.

exporter/internal/queue/default_batcher.go
sfc-gh-sili: Thanks! Do you mean that we should check if other items in the req list are large enough?
I was thinking we should flush everything out immediately if a request was split, so that the all the pieces arrive closer to each other.

sfc-gh-sili: That makes sense. Let me add that to the if clause too.

dmitryax (approved):

Issue #11405, opened 2024-10-09
Remove builder support to build old version


bogdandrutu:
One of the reason the mess with default providers (see https://github.com/open-telemetry/opentelemetry-collector/issues/11403 or https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.109.0 or the need to disable strict check, see https://github.com/open-telemetry/opentelemetry-collector/pull/11168) is the support of building older version of the collector with newer builder, which also is very fragile because it makes the assumption that the generated files (templates) will know how to talk to multiple versions. For this, because I believe this is not a very useful feature, but brings us a lot of troubles, already caused issues because we could not fix things simpler (for example to hardcode the versions for default confmap providers, etc.).

For this, we need to remove the config `otelcol_version` and always use the hardcoded "default_version".

We already print an warning https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/builder/internal/builder/main.go#L77

bogdandrutu:
@mx-psi here is a followup on how we can remove a lot of complications, and I believe is reasonable to have this and we can have much better checks, etc.

bogdandrutu:
Looks like I am consistent :))))

PR #11588, opened 2024-11-01
Remove builder support to build old version, and the otelcol_version config

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/11405


songy23 (approved):
LGTM

songy23:
cc @mx-psi

.chloggen/otelcol_version.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove builder support to build old version, and the otelcol_version config
+
+# One or more tracking issues or pull requests related to the change
+issues: [11405]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: User should remove this property from their config, to build older versions use older builders.
```
mx-psi, line 18: We should have an error that specifically tells you this for a few releases instead of "unknown configuration setting"

.chloggen/otelcol_version.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api, user]
```
mx-psi, line 25: I don't think this is an API change, the builder is not usable as a library, is it?

cmd/builder/internal/builder/config.go
```diff
@@ -16,7 +16,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const defaultOtelColVersion = "0.112.0"
+const defaultOtelColVersion = "v0.112.0"
```
mx-psi, line 19: Does the release automation that bumps this version work well with adding the `v` here?

cmd/builder/internal/builder/main_test.go
```diff
@@ -188,7 +188,7 @@ func TestVersioning(t *testing.T) {
 				cfg.SkipStrictVersioning = true
 				cfg.Exporters = []Module{
 					{
-						GoMod: "go.opentelemetry.io/collector/exporter/otlpexporter v0.97.0",
+						GoMod: "go.opentelemetry.io/collector/exporter/otlpexporter v0.112.0",
```
mx-psi, line 191: Do we need to bump this and the one above on every release? If so, can we automate it?

.chloggen/otelcol_version.yaml
bogdandrutu: Done

.chloggen/otelcol_version.yaml
bogdandrutu: Done.

cmd/builder/internal/builder/config.go
bogdandrutu: Tested, it works, see where the versions don't have the `v` prefix but it doesn't matter:

https://github.com/open-telemetry/opentelemetry-collector/blob/main/Makefile#L498

cmd/builder/internal/builder/config.go
```diff
@@ -110,6 +112,9 @@ func NewDefaultConfig() Config {
 
 // Validate checks whether the current configuration is valid
 func (c *Config) Validate() error {
+	if c.Distribution.OtelColVersion != "" {
+		return errors.New("the OtelColVersion config is deprecated, please remove usage of it")
```
mx-psi, line 116:
```suggestion
		return errors.New("`otelcol_version` has been removed. To build with an older Collector API, use an older (aligned) builder version instead")
```

mx-psi (approved):

cmd/builder/internal/builder/config.go
bogdandrutu: Thank you, applied.

PR #11594, opened 2024-11-04
(batchprocessor): Move single shard batcher creation to the constructor

wildum:
#### Description

The single-shard batcher is currently created and started in the start function. If the consume function is called before the start function, sb.single will be nil and it will panic. I'm not sure if that can happen in the Otel collector but it might happen in distributions that have a different runtime (such as Alloy - Grafana's distribution).

I think that conceptually, it's better to create the shard in the constructor and start it in the start function. 

PS: I did not create an entry in the changelog because the initial [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/11314) for this change didn't but let me know if I should add one.

Credits also to @thampiotr for this fix

### DIFF
--- a/.chloggen/batch-processor-init-single-shard.yaml
+++ b/.chloggen/batch-processor-init-single-shard.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: batchprocessor
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Move single shard batcher creation to the constructor
+
+# One or more tracking issues or pull requests related to the change
+issues: [11594]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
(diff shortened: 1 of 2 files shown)


bogdandrutu:
> I'm not sure if that can happen in the Otel collector but it might happen in distributions that have a different runtime (such as Alloy - Grafana's distribution).

In the collector it cannot and Grafana uses the components wrong /cc @jpkrohling. The contract says [here]( https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L14).

Another problem is that Alloy is not a otel distribution if they don't respect collectors contracts. cc @jpkrohling  

FYI: I am ok with this change, but if in the future we break others because they do not respect the contract that is not considered a breaking change and fixes may not be approved/accepted.

bogdandrutu (approved):

bogdandrutu:
@wildum please add a changelog.

thampiotr:
> In the collector it cannot and Grafana uses the components wrong /cc @jpkrohling. The contract says [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L14).

You're right, we shouldn't depend on the implementation detail like this. We're working [on a fix](https://github.com/grafana/alloy/pull/2027) for this in Alloy and should be able to handle OTel component's lifecycle correctly. cc @jpkrohling

wildum:
> @wildum please add a changelog.

Done. I put 'enhancement' because it's not a bug for the collector. Let me know if you prefer 'bug_fix'.

> In the collector it cannot and Grafana uses the components wrong /cc @jpkrohling. The contract says [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L14).
Another problem is that Alloy is not a otel distribution if they don't respect collectors contracts. cc @jpkrohling
FYI: I am ok with this change, but if in the future we break others because they do not respect the contract that is not considered a breaking change and fixes may not be approved/accepted.

Thanks for pointing this out. We noticed this today while investigating the problem we had with the batch processor. We will work towards improving Alloy to respect the documented contracts.

PR #11595, opened 2024-11-04
Create receiver/scraper module and migrate helper/error modules

### LINKED ISSUE #11003: [receiver] Move scraper packages into scraper module

jackgopack4:
#### Description
moves scrapererror and scraperhelper functions to new `receiver/scraper` module in order to assist stabilizing receiver for collector 1.0

#### Link to tracking issue
Fixes #11003 

#### Testing
existing tests are migrated; no new code added and therefore no new tests needed

#### Documentation
.chloggen files


jackgopack4:
only failing test is a metrics test that is passing on local VMs, not sure what the issue is there.

jackgopack4:
I added aliases like @bogdandrutu asked for in https://github.com/open-telemetry/opentelemetry-collector/pull/11166#pullrequestreview-2367004440 but I am wondering how long these aliases will need to stay? If they are offered in the stable version of receiver, but point to a beta component, does that meet collector 1.0 requirements?

Issue #10808, opened 2024-08-06
Define plan for `service::telemetry` configuration section evolution


mx-psi:
We intend to use the configuration schema defined by the OpenTelemetry Configuration WG to be used under the `service::telemetry` section. This schema may not be ready or stable when we intend to release the Collector v1 distro. To prepare for this, we need to decide how we are going to handle the transition. Do we want to support the old and the new configuration schema? Do we want to adopt the Configuration WG one and do some backwards-compatible shim if any breaking changes occur?

iblancasa:
@mx-psi can you link to the OpenTelemetry Configuration WG schema?

codeboten:
Here's the repo for the schema https://github.com/open-telemetry/opentelemetry-configuration

mx-psi:
Filed open-telemetry/opentelemetry.io/pull/6219 to resolve this

codeboten:
That PR has been merged, closing this issue

PR #11458, opened 2024-10-15
wip: update config dependency

codeboten:
This brings in some breaking changes from the config package, mainly the headers on OTLP exporters are no longer a list of key value pairs but a list of name/value pairs.

PR #11611, opened 2024-11-05
update config dependency

codeboten:
This PR does a couple of things that I couldn't quite split up so I put together a PR w/ individual commits to help reviewers get through it. This PR does the following:

1. update `go.opentelemetry.io/contrib/config` package to latest. this brings in breaking changes. in order to prevent those breaking changes from impacting end users, i've also added a layer of config unmarshaling
2. updates the collector to instantiate the meter provider (and exporters) via the config package. this allows us to remove all the code in `otelinit`. the reason for including this change was that unmarshaling the config was causing circular dependencies i didn't want to address by moving code that could be deleted around.

Replacement for https://github.com/open-telemetry/opentelemetry-collector/pull/11458.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/12021


mjnowen:
Looking forward to seeing this get merged!

mjnowen:
Not stale?

codeboten:
Could use some feedback from @open-telemetry/collector-approvers  on this latest commit to the PR to update the config support from v0.2.0 to v0.3.0.

### tl;dr there's a backwards incompatible change in how headers are passed in and I wanted to support users transitioning

i wanted to support both the 0.3.0 and the 0.2.0 schema, which resulted in some migration code. I still have a few tests to add, and this change will need a PR in go-contrib to be merged before it can be completely done. Anyways, i could use your input: https://github.com/open-telemetry/opentelemetry-collector/pull/11611/commits/df832652441ac889705c3e77e04426a4e01b2122


This is waiting on https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6412

dmitryax:
The approach to handle the backward incompatible change looks good to me 👍

service/telemetry/internal/migration/v0.3.0.go
```diff
@@ -0,0 +1,174 @@
...
+func (c *LogsConfigV030) Unmarshal(conf *confmap.Conf) error {
+	if err := conf.Unmarshal(c); err != nil {
+		var v02 logsConfigV020
+		// try unmarshaling using v0.2.0 struct
+		if e := conf.Unmarshal(&v02); e != nil {
+			// error could not be resolved through backwards
+			// compatibility attempts
+			return err
+		}
+		// TODO: add a warning log to tell users to migrate their config
```
dmitryax, line 169: I guess it might be complicated to pass a logger here, but it's pretty important. I think we should come up with a solution by the time we release this change

service/service.go
```diff
@@ -363,3 +385,43 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
 	}
 	return pcommonRes
 }
+
+func disableHighCardinalityMetrics() []config.View {
```
jpkrohling, line 389: looks OK for now, but if we expect this list to grow before we remove the gate, it might make sense to externalize this somehow?

service/service.go
```diff
@@ -363,3 +385,43 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
...
+						semconv118.AttributeNetHostPort,
+					},
+				},
+			},
+		},
+	}
+}
+
+func ptr[T any](v T) *T {
+	return &v
```
jpkrohling, line 426: why can't go just provide this natively :-(

jpkrohling (approved):
I appreciate your efforts to not break our current users.

service/telemetry/internal/migration/v0.3.0.go
mx-psi: The way we do this on the Datadog exporter is to have [a private field](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adef54ee022ddd503bf78afdfb2a01b599853579/pkg/datadog/config/config.go#L104) that is [written to during unmarshaling](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adef54ee022ddd503bf78afdfb2a01b599853579/pkg/datadog/config/config.go#L267) and then [a method to log warnings](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adef54ee022ddd503bf78afdfb2a01b599853579/pkg/datadog/config/config.go#L107-L112).

I don't love it, but I haven't been able to come up with anything better

service/service.go
mx-psi: This is the purpose of #11754

mx-psi (commented):
I think this is a good opportunity to resolve #10808 by writing down what our plans are regarding this.

service/service.go
codeboten: is this a question or a comment about go?

atoulme: yes

jaronoff97 (approved):
This will close #12021 , thanks alex :bow:

service/service.go
```diff
@@ -362,3 +384,136 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
...
+
+	// Batch processor metrics
+	if level < configtelemetry.LevelDetailed {
+		scope := ptr("go.opentelemetry.io/collector/processor/batchprocessor")
+		views = append(views, dropViewOption(&config.ViewSelector{
+			MeterName:      scope,
+			InstrumentName: ptr("otelcol_processor_batch_batch_send_size_bytes"),
+		}))
+	}
+	return views
```
codeboten, line 514: @mx-psi ended up moving the logic to append views in here, as the views are needed to instantiate the SDK, PTAL

codeboten:
Attempting to fix the integration test that's failing in contrib: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37425

Unclear to me whether this change is making that test flakier or not as this issue exists: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34836

service/service.go
```diff
@@ -362,3 +383,136 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
...
+}
+
+func configureViews(level configtelemetry.Level) []config.View {
+	views := []config.View{}
+
+	if disableHighCardinalityMetricsFeatureGate.IsEnabled() {
+		views = append(views, []config.View{
+			{
+				Selector: &config.ViewSelector{
+					MeterName: ptr("go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"),
```
mx-psi, line 405: Why not use `otelgrpc.ScopeName`, in case it changes?

service/service.go
```diff
@@ -362,3 +383,136 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
...
+							semconv118.AttributeNetSockPeerAddr,
+							semconv118.AttributeNetSockPeerPort,
+							semconv118.AttributeNetSockPeerName,
+						},
+					},
+				},
+			},
+			{
+				Selector: &config.ViewSelector{
+					MeterName: ptr("go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"),
```
mx-psi, line 419: ditto

service/service.go
```diff
@@ -362,3 +383,136 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
...
+				},
+			},
+		}...)
+	}
+
+	if level < configtelemetry.LevelDetailed {
+		// Drop all otelhttp and otelgrpc metrics if the level is not detailed.
+		views = append(views,
+			dropViewOption(&config.ViewSelector{
+				MeterName: ptr("go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"),
```
mx-psi, line 437: ditto

service/service.go
```diff
@@ -362,3 +383,136 @@ func pdataFromSdk(res *sdkresource.Resource) pcommon.Resource {
...
+			}),
+			dropViewOption(&config.ViewSelector{
+				MeterName: ptr("go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"),
```
mx-psi, line 440: ditto

mx-psi (commented):
The new place for views looks reasonable

service/service.go
codeboten: i didn't want to pull in an additional dependency for a string

mx-psi (approved):
I am fine with copy-pasting the scope name, it seems unlikely to change

codeboten:
The exporter test is still consistently failing https://github.com/open-telemetry/opentelemetry-collector/actions/runs/12932562549/job/36070135348?pr=11611, @songy23 do you have any thoughts on what's causing it to fail? i can't repro locally

codeboten:
>  i can't repro locally
was finally able to repro locally... will continue debugging the issue

songy23:
> do you have any thoughts on what's causing it to fail?

Datadog relies on the internal meter to report some component specific metrics. Those metrics are then exposed with the internal Prometheus server. If anything changes the metric names / attributes from internal meter and/or Prometheus, it will break the Datadog integrations.

codeboten:
Thanks for your help @songy23, bug has been fixed

songy23 (approved):

andrzej-stencel (approved):

evan-bradley (approved):

service/service_test.go
```diff
@@ -556,12 +556,16 @@ func assertMetrics(t *testing.T, metricsAddr string, expectedLabels map[string]l
 		"otelcol_process_runtime_heap_alloc_bytes":       false,
 		"otelcol_process_runtime_total_alloc_bytes":      false,
 		"otelcol_process_uptime":                         false,
+		"promhttp_metric_handler_errors_total":           false,
```
dmitryax, line 559: @codeboten, is this metric expected to be added? I'm not sure if it provides any value to the collector internal metrics...

codeboten: it's added by the prometheus server that is started via the SDK. we might be able to drop it via a view

dmitryax: ok, that sounds good. Let me create an issue

PR #11613, opened 2024-11-05
[chore] Small refactoring in builder config validation/parsing

bogdandrutu:


cmd/builder/internal/builder/config.go
```diff
@@ -207,29 +171,26 @@ func validateModules(name string, mods []Module) error {
 	return nil
 }
 
-func parseModules(mods []Module) ([]Module, error) {
-	var parsedModules []Module
-	for _, mod := range mods {
-		if mod.Import == "" {
-			mod.Import = strings.Split(mod.GoMod, " ")[0]
+func parseModules(mods []Module) error {
```
codeboten, line 174: This is changing the behaviour or parseModules to not return an updated version of the modules, is that wanted? Or should this be receiving `[]*Module` if we want to change the modules in place?

bogdandrutu: > Or should this be receiving []*Module if we want to change the modules in place?

Why? The slice is implemented as a pointer to memory and size/cap. The memory can be changed in place, we have tests for that.

bogdandrutu: See https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/builder/internal/builder/config_test.go#L21

PR #11632, opened 2024-11-08
[receiver/scraperhelper] exit before scraping if the receiver was shutdown

atoulme:
#### Description
exit before scraping if the receiver was shutdown, do not wait the initial delay.


pjanotti (approved):

bogdandrutu (approved):

Makeuptehrani:
> 
> #### Description
> exit before scraping if the receiver was shutdown, do not wait the initial delay.
> 
>

PR #11633, opened 2024-11-09
system/telemetry metrics OTLP GRPC Exporter: add CA certificate support

ms-jcorley:
#### Description
Adds support for CA certificates to the OTLP gRPC exporter for metrics in system/telemetry.
Currently if a "Certificate" is specified in the config, it is silently ignored.

#### Testing
Both a positive and negative unit test was added.
The good test CA certificate was copied from another unit test.
I'm going to file an issue for updating all test certificates as all the ones I checked seem to expire in 2032 which is only 8 years away.

#### Documentation
None.


service/telemetry/internal/otelinit/config.go
```diff
@@ -224,6 +225,14 @@ func initOTLPgRPCExporter(ctx context.Context, otlpConfig *config.OTLPMetric) (s
 		opts = append(opts, otlpmetricgrpc.WithEndpoint(u.Host))
 		if u.Scheme == "http" {
 			opts = append(opts, otlpmetricgrpc.WithInsecure())
+		} else {
+			if otlpConfig.Certificate != nil {
```
dmathieu, line 229: Nit, but do we shouldn't need that extra indentation here

```suggestion
		} else if otlpConfig.Certificate != nil {
```

dmathieu (commented):
This will need a changelog entry.

service/telemetry/internal/otelinit/config.go
ms-jcorley: Ah yes, I was thinking about future code there and also didn't realize I should run make lint (I'm new here)... fixed.

ms-jcorley:
> This will need a changelog entry.

@dmathieu, changelog was added and I fixed the lint issue.  Waiting for a review/approval.

ms-jcorley:
> Can you run `make gotidy`?

@dmathieu, Done.

dmathieu (approved):

ms-jcorley:
Hi @codeboten, it looks like this is blocked on your final review.  This is quite a small change, could you take a look when you get a chance?
Thank you!

codeboten (approved):
Thanks @ms-jcorley

PR #11637, opened 2024-11-09
[exporter] Flip on queue batcher

sfc-gh-sili:
#### Description

This PR solves https://github.com/open-telemetry/opentelemetry-collector/issues/10368.

Previously we use a pushing model between the queue and the batch, resulting the batch size to be constrained by the `sending_queue.num_consumers`, because the batch cannot accumulate more than `sending_queue.num_consumers` blocked goroutines provide.

This PR changes it to a pulling model. We read from the queue until threshold is met or timeout, then allocate a worker to asynchronously send out the request.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/10368

#### Testing

This PR swaps out `batch_sender` directly and still passes all the existing tests.

### DIFF
--- a/.chloggen/11637-exporter-queue-batcher.yaml
+++ b/.chloggen/11637-exporter-queue-batcher.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: breaking
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: exporterqueue
--- a/.chloggen/11637-exporter-queue-batcher.yaml
+++ b/.chloggen/11637-exporter-queue-batcher.yaml
@@ -1,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterqueue
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change exporter queue batching to use a pulling model.
+
+# One or more tracking issues or pull requests related to the change
+issues: [8122, 10368]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  If both queuing and batching is enabled for exporter, we now use a pulling model instead of a
+  pushing model. num_consumer in queue configuration is now used to specify the maximum number of
+  concurrent workers that are sending out the request. 
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/11637-exporter-queue-batcher.yaml
+++ b/.chloggen/11637-exporter-queue-batcher.yaml
@@ -7,7 +7,7 @@ change_type: breaking
 component: exporterqueue
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Change exporter queue batching to use a pulling model.
+note: Introduce a feature gate exporter.UsePullingBasedExporterQueueBatcher to use the new pulling model in exporter queue batching.
 
 # One or more tracking issues or pull requests related to the change
 issues: [8122, 10368]
--- a/exporter/exporterhelper/internal/base_exporter.go
+++ b/exporter/exporterhelper/internal/base_exporter.go
@@ -26,7 +26,7 @@ import (
 )
 
 var usePullingBasedExporterQueueBatcher = featuregate.GlobalRegistry().MustRegister(
-	"telemetry.UsePullingBasedExporterQueueBatcher",
+	"exporter.UsePullingBasedExporterQueueBatcher",
 	featuregate.StageAlpha,
 	featuregate.WithRegisterFromVersion("v0.114.0"),
 	featuregate.WithRegisterDescription("if set to true, turns on the pulling-based exporter queue bathcer"),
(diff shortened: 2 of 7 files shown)


exporter/internal/queue/batcher.go
```diff
@@ -94,9 +94,3 @@ func (qb *BaseBatcher) flushAsync(batchToFlush batch) {
 		qb.workerPool <- true
 	}()
 }
-
-// Shutdown ensures that queue and all Batcher are stopped.
-func (qb *BaseBatcher) Shutdown(_ context.Context) error {
-	qb.stopWG.Wait()
-	return nil
-}
```
bogdandrutu, line 102: Why this change?

exporter/internal/queue/default_batcher.go
```diff
@@ -124,25 +124,19 @@ func (qb *DefaultBatcher) startTimeBasedFlushingGoroutine() {
...
-					qb.currentBatchMu.Unlock()
-					continue
-				}
-				batchToFlush := *qb.currentBatch
-				qb.currentBatch = nil
-				qb.currentBatchMu.Unlock()
-
-				// flushAsync() blocks until successfully started a goroutine for flushing.
-				qb.flushAsync(batchToFlush)
-				qb.resetTimer()
```
bogdandrutu, line 138: Not sure I understand this, can we do this in a separate PR?

exporter/internal/queue/batcher.go
sfc-gh-sili: See the other comment.

exporter/internal/queue/default_batcher.go
sfc-gh-sili: Thanks! Here it is: https://github.com/open-telemetry/opentelemetry-collector/pull/11666
`batch_sender_test` helped me detect that the original implementation is missing a flush on shutdown.

sfc-gh-sili:
@dmitryax  Hi Dimitrii, I wonder if you know what would be a better way to make sure existing tests pass with both feature gate on and off. Manually enabling then disabling in every single exporter test could work but I wonder if there's other option

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -21,9 +21,17 @@ import (
 	"go.opentelemetry.io/collector/exporter/exporterbatcher"
 	"go.opentelemetry.io/collector/exporter/exporterqueue" // BaseExporter contains common fields between different exporter types.
 	"go.opentelemetry.io/collector/exporter/internal"
+	"go.opentelemetry.io/collector/featuregate"
 	"go.opentelemetry.io/collector/pipeline"
 )
 
+var usePullingBasedExporterQueueBatcher = featuregate.GlobalRegistry().MustRegister(
+	"telemetry.UsePullingBasedExporterQueueBatcher",
+	featuregate.StageBeta,
```
dmitryax, line 30: Why starting with Beta? That sounds too aggressive. Let's start with Alpha

.chloggen/11637-exporter-queue-batcher.yaml
```diff
@@ -0,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
dmitryax, line 4:
```suggestion
change_type: enhancement
```

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -27,7 +27,7 @@ import (
 
 var usePullingBasedExporterQueueBatcher = featuregate.GlobalRegistry().MustRegister(
 	"telemetry.UsePullingBasedExporterQueueBatcher",
```
dmitryax, line 29:
```suggestion
	"exporter.UsePullingBasedExporterQueueBatcher",
```

.chloggen/11637-exporter-queue-batcher.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterqueue
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change exporter queue batching to use a pulling model.
```
dmitryax, line 10:
```suggestion
note: Introduce a feature gate exporter.UsePullingBasedExporterQueueBatcher to use the new pulling model in exporter queue batching.
```

exporter/exporterhelper/internal/queue_sender.go
```diff
@@ -71,30 +72,49 @@ type QueueSender struct {
...
 		numConsumers:   numConsumers,
 		traceAttribute: attribute.String(ExporterKey, set.ID.String()),
 		obsrep:         obsrep,
 		exporterID:     set.ID,
 	}
-	consumeFunc := func(ctx context.Context, req internal.Request) error {
-		err := qs.NextSender.Send(ctx, req)
-		if err != nil {
-			set.Logger.Error("Exporting failed. Dropping data."+exportFailureMessage,
-				zap.Error(err), zap.Int("dropped_items", req.ItemsCount()))
```
dmitryax, line 93: Doesn't the function definition itself stay the same so we don't need to split it?

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -27,7 +27,7 @@ import (
...
-	featuregate.StageBeta,
+	featuregate.StageAlpha,
 	featuregate.WithRegisterFromVersion("v0.114.0"),
```
dmitryax, line 31:
```suggestion
	featuregate.WithRegisterFromVersion("v0.115.0"),
```

exporter/exporterhelper/internal/queue_sender.go
sfc-gh-sili: Good point. Thanks!

exporter/exporterhelper/internal/batch_sender_test.go
```diff
@@ -158,7 +158,6 @@ func TestBatchSender_BatchExportError(t *testing.T) {
 			assert.Eventually(t, func() bool {
 				return sink.requestsCount.Load() == tt.expectedRequests &&
 					sink.itemsCount.Load() == tt.expectedItems &&
-					be.BatchSender.(*BatchSender).activeRequests.Load() == 0 &&
```
dmitryax, line 161: Why is this line removed? We should wait until there are no in-flight requests

dmitryax (approved):
One question about tests. Otherwise LGTM

exporter/exporterhelper/internal/batch_sender_test.go
sfc-gh-sili: This is because if the new feature is turned on, we use `queueBatcher` instead of `BatchSender` so `BatchSender` is a nil. In the previous PR, I set up the test to run with both the feature on and off so this line has to be removed.

dmitryax: Ok sounds good 👍

PR #11653, opened 2024-11-12
Add local module replacement to ocb template

### LINKED ISSUE #11649: ocb should support replacing providers and converters with local codepaths for testing
ocb `builder-config.yaml` manifest files allow replacing components such as connectors, extensions, receivers, exporters, and processors by specifying the `path` value below the name/version of the module. This doesn't work for providers or converters.
example:
```
extensions:
  - gomod: github.com/repo/module-name v0.12.0
    path: /my/local/path/to/repo/module-name
```

I plan to add some replace logic in the go.mod templates for ocb for providers and converters similar to the logic in https://github.com/open-telemetry/opentelemetry-collector/blob/9c4c007a12b48e794e8946c725e4dbdbb7d9d996/cmd/builder/internal/builder/templates/go.mod.tmpl#L37-L52

if this feature shouldn't be supported, we should document which components are allowed to be replaced with local modules in ocb and which ones are not.

**Additional context**
`confmap.Converter` only was added as of https://github.com/open-telemetry/opentelemetry-collector/pull/11584 so builder must be ran or built from source in order to properly test this feature

jackgopack4:
#### Description
Adds ability to replace Providers and Converters with local paths when building custom collector in ocb

#### Link to tracking issue
Closes #11649

#### Testing
local build of ocb with locally downloaded converter and provider (can add this type of test to github actions if others think it is necessary)

#### Documentation
.chloggen file

### DIFF
--- a/cmd/builder/internal/builder/templates/go.mod.tmpl
+++ b/cmd/builder/internal/builder/templates/go.mod.tmpl
@@ -34,10 +34,10 @@ require (
 	github.com/knadh/koanf/providers/confmap v0.1.0 // indirect
 )
 
-{{- range .Providers}}
+{{- range .Converters}}
 {{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
 {{- end}}
-{{- range .Converters}}
+{{- range .Providers}}
 {{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
 {{- end}}
 {{- range .Connectors}}
--- a/cmd/builder/internal/builder/templates/go.mod.tmpl
+++ b/cmd/builder/internal/builder/templates/go.mod.tmpl
@@ -34,6 +34,12 @@ require (
 	github.com/knadh/koanf/providers/confmap v0.1.0 // indirect
 )
 
+{{- range .Providers}}
+{{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
+{{- end}}
+{{- range .Converters}}
+{{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
+{{- end}}
 {{- range .Connectors}}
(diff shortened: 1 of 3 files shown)


jackgopack4:
for some reason ad8ab82e4f450037bf590aafe4eb1a8f15d2f97a is still pending for removal even though I force pushed.

jackgopack4:
if the commit doesn't disappear today I'll close this PR and open a new one with a fresh branch. opened a ticket with GitHub because this is weird behavior

songy23 (approved):
LGTM

cmd/builder/internal/builder/templates/go.mod.tmpl
```diff
@@ -34,6 +34,12 @@ require (
 	github.com/knadh/koanf/providers/confmap v0.1.0 // indirect
 )
 
+{{- range .Providers}}
+{{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
+{{- end}}
+{{- range .Converters}}
+{{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
+{{- end}}
```
bogdandrutu, line 42: Can you do a separate PR and rename .Providers -> .ConfmapProviders same for converter?

bogdandrutu, line 42:
```suggestion
{{- range .Converters}}
{{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
{{- end}}
{{- range .Providers}}
{{if ne .Path ""}}replace {{.GoMod}} => {{.Path}}{{end}}
{{- end}}
```

jackgopack4: yeah that makes sense to me

jackgopack4: do you want to wait to merge this one in until that PR is open? seems like it would make sense to merge this and then I can create a new branch from main with this included

Issue #11238, opened 2024-09-20
Scraper feedback, and what to do next?


bogdandrutu:
**Current state:**

While https://github.com/open-telemetry/opentelemetry-collector/issues/9473 was a right call in terms of strangeness that from a string we construct an ID, we should not pass a "type" because the idea was that a "receiver" can have multiple scrapers, but if that is true, all of them will be the same receiver/component type correct?

We have made a mistake probably (big part of the problem is the author of this PR) with the design of the scrapers. Here are a set of questions we need to answer, and based on that we can change the design of the scraper package:
1. Should the "scraperhelper" be a receiver that supports multiple "scrapers" or a receiver that supports one Scraper and we can have a "multi source" scraper in case we really need to support a receiver to scrape multiple sources?
2. Should we have a standalone "scraper" type instead of making this a receiver? Or maybe we have only one "scraperreceiver" that allows users to configure multiple scrapers?

During the decision and discussion it should be considered that currently we have a big problem, which is that for each scraper/receiver we create its one timer and we lose the opportunity of batching at the source. Another feedback we got in the past with the current a scraper as a receiver is that user has to configure the scrape interval individually for each receiver.

A proposed idea is to start from how user should configure this, and see how we can achieve this making sure we "batch" as much as possible all these scrapings:

```yaml
receivers:
  scraper/30sec:
    collection_interval: 30s
    initial_delay: 10s
    timeout: 10s
    sources:
      sql:
        ... // things specific to sql including enable/disable metrics
      nginx:
        ... // things specific to nginx including enable/disable metrics.
```

Later we can add capability to do "source discovery" or build a different receiver like the current "receivercreator" which can be adopted to follow the same pattern/structure.

dmitryax:
I like the idea. We can have just one `scraper` receiver eventually. I see a few opportunities with this approach:

1. It'll remove the confusion between "push" and "pull" receivers. All the receivers will be "push" except for the `scraper`, and we don't need `kafkametrics` in addition to the `kafka` receiver.

2. The batching, as you mentioned, with consistent timestamps. However, it can lead to huge payloads that potentially have to be split on the exporter side. So, batching can probably be a configurable behavior. Also, I think we should still initiate separate workers for each source scraping the data concurrently + one main worker batching them together if it's enabled.

[...]

PR #11657, opened 2024-11-12
Add a new scraper package to replace scraperhelper

bogdandrutu:
Update https://github.com/open-telemetry/opentelemetry-collector/issues/11238

No changelog since this is still in progress.


scraper/metadata.yaml
```diff
@@ -0,0 +1,7 @@
+type: service
+github_project: open-telemetry/opentelemetry-collector
+
+status:
+  class: pkg
+  stability:
+    development: [metrics]
```
songy23, line 7: Are traces & logs in scope too?

versions.yaml
```diff
@@ -77,6 +77,7 @@ module-sets:
       - go.opentelemetry.io/collector/receiver/otlpreceiver
       - go.opentelemetry.io/collector/receiver/receiverprofiles
       - go.opentelemetry.io/collector/receiver/receivertest
+      - go.opentelemetry.io/collector/scraper
```
songy23, line 80: Please update https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/builder/test/core.builder.yaml as well

bogdandrutu: No need for the moment, since noboday depends on it.

scraper/metadata.yaml
bogdandrutu: I think we will extend to logs at one point, see request https://github.com/open-telemetry/opentelemetry-collector/issues/11238#issuecomment-2457770329.

For me the first part is to move what we have (only metrics) to this new framework proposed in the issue, then we can extend to logs.

songy23 (approved):

sincejune:
@bogdandrutu May I ask what would be the next step of this PR? Are we planning to remove scraperhelper/scrapercontroller as well?

PR #11676, opened 2024-11-14
[chore] Run `make genotelcorecol` to fix "build-and-test / checks"

jade-guiton-dd:
#### Description

The "build-and-test / checks" CI job specifies `go-version: ~1.22.8`. Today, this started resolving to `1.22.9`, causing [errors in CI](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11839445874/job/32990922879?pr=11670) because of the `toolchain 1.22.8` directive in `otelcorecol/go.mod`. This PR just updates the directive to fix the error. A more long-term solution may be preferred, such as using exact versions instead of `~` versions in CI.

### DIFF
--- a/.github/workflows/api-compatibility.yml
+++ b/.github/workflows/api-compatibility.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
-          go-version: ~1.22.8
+          go-version: ~1.22.9
 
       # Generate apidiff states of Main
       - name: Generate-States
--- a/.github/workflows/build-and-test-windows.yaml
+++ b/.github/workflows/build-and-test-windows.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
-          go-version: ~1.22.8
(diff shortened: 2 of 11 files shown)


bogdandrutu (approved):

bogdandrutu:
You need to update https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/build-and-test.yml#L48 (and everywhere)

jade-guiton-dd:
Interesting, it seems some runs of `check` resolve the version to 1.22.8, and others to 1.22.9...

bogdandrutu:
>  Interesting, at the moment it seems some runs of check resolve ~1.22.8 to 1.22.8, and others to 1.22.9...

Can you clarify ?

jade-guiton-dd:
> Can you clarify ?

The failing run I linked earlier specified version ~1.22.8, which [resolved to 1.22.9](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11839445874/job/32990922879?pr=11670#step:3:12), causing a conflict with the `toolchain 1.22.8` directive.

But the first run in *this* PR also failed, because it it [resolved to 1.22.8](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11841264595/job/32997107261#step:3:12) from the same specification, causing a conflict with the update `toolchain 1.22.9` directive. I suspect some kind of cache effect causing different runs to find different "latest" versions.

I just updated the CI to use ~1.22.9, so for the time being it should be unambiguous, but I think that's an additional reason to want to switch to exact versions.

PR #11683, opened 2024-11-15
[chore] Generate `replace` directives automatically in `check-contrib`

jade-guiton-dd:
#### Description

`make check-contrib` use `go mod edit` to allow running the `opentelemetry-collector-contrib` unit tests against the local version of the core libraries. It also removes the `replace` statements after tests have been run. At the moment, this is done with two long handwritten `go mod edit` commands. But this very easily gets out-of-date: for reference, we currently only `-replace` 70 modules out of the 73 in versions.yaml, and only `-dropreplace` 63 out of _those_.

This PR changes the `Makefile` to programatically generate the commands, to avoid needing to keep them up to date.

This was split off from PR #11670.

You can check the output in the `contrib-tests` CI job output.


mx-psi (approved):

mx-psi:
You can double check e.g. here: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11858185454/job/33048223148?pr=11683 that the generated replaces look reasonable

PR #11688, opened 2024-11-18
Fix sample count iteration.

constanca-m:
#### Description

Sample count iteration for profiles is incrementing the wrong variable inside the loop, which causes the result to be wrong.

#### Testing

The refactored unit test covers this case.


dmathieu:
Nice catch, thank you! Could you add a changelog entry?

constanca-m:
Yes @dmathieu, I have updated it with a new changelog entry.

dmathieu (approved):

Issue #11606, opened 2024-11-05
Use Renovate for markdown-link-check updates


andrzej-stencel:
The version of `markdown-link-check` took has been pinned to a specific version in https://github.com/open-telemetry/opentelemetry-collector/pull/11605 due to a bug in the most recent release https://github.com/tcort/markdown-link-check/issues/368.

Ideally the version of the tool should be updated automatically by Renovate bot by creating a pull request on every new release of the tool. This would allow us to control the version used and only ever use a version that actually works.

andrzej-stencel:
Looks like nothing needs to be changed, Renovate is already doing its job: https://github.com/open-telemetry/opentelemetry-collector/pull/11608.

andrzej-stencel:
Not exactly true. It's true that Renovate correctly creates a PR to update markdown-link-check, but the problem is, the check-links job is not run for this PR, so there is no way to know if the update does not break the tool.

We need to change the `check-links` job to also be run when the tool is updated - e.g. when `package.json` file is changed.

PR #11725, opened 2024-11-21
[chore] Run markdown link check if package.json changes

niwoerner:
#### Description
This PR will run the "markdown-link-check" tool if the package.json changes even if no markdown files were updated. This will help to confirm that "markdown-link-check" is still working as expected after a version update/pin. 

If 1+ *.md files gets changed AND the package.json gets updated, it will only check the updated files to keep the same behaviour as before. 

If only the package.json gets updated it will test all existing *.md files. 

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/11606 

#### Testing

I tested/reproduced this manually: 

1. *.md file changes, but package.json not: https://github.com/niwoerner/opentelemetry-collector/actions/runs/11959750938
2. *.md file and package.json changes: https://github.com/niwoerner/opentelemetry-collector/actions/runs/11959777300
3. *.md file doesn't change but package.json: https://github.com/niwoerner/opentelemetry-collector/actions/runs/11959797335

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -18,7 +18,7 @@ jobs:
     env:
       PR_HEAD: ${{ github.event.pull_request.head.sha }}
     outputs:
-      md: ${{ steps.changes.outputs.md }}
+      files: ${{ steps.changes.outputs.files }}
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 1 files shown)


codeboten (approved):
thanks @niwoerner

PR #11731, opened 2024-11-22
[chore] Export the confMap Unmarshal function

dehaansa:
#### Description
In Alloy, we have older components using a fork of the opentelemetry collector to access the confmap unmarshal function. In order to reduce our dependencies/maintenance requirement we'd like to get rid of that fork, and exporting this function will help with that goal.

### DIFF
--- a/otelcol/configprovider.go
+++ b/otelcol/configprovider.go
@@ -80,7 +80,7 @@ func (cm *configProvider) Get(ctx context.Context, factories Factories) (*Config
 	}
 
 	var cfg *configSettings
-	if cfg, err = unmarshal(conf, factories); err != nil {
+	if cfg, err = Unmarshal(conf, factories); err != nil {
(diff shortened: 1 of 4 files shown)


bogdandrutu (changes requested):
Hi @dehaansa 

Thank you for your issue. I do understand your problem, but I would prefer not to expose an internal detail unnecessary.

Also, this is not going to work for you because it returns a private member that you cannot use.

dehaansa:
> Hi @dehaansa
> 
> Thank you for your issue. I do understand your problem, but I would prefer not to expose an internal detail unnecessary.
> 
> Also, this is not going to work for you because it returns a private member that you cannot use.

While the configSettings type is not exported, as its fields are all exported I believe it should work as it is simply used immediately to populate an otelcol.Config object. That said, I'll see if I can either work with one of the existing providers or propose another, thanks.

Issue #10266, opened 2024-05-26
How to declare optional parts of the config?


yurishkuro:
Slightly related to the discussion in open-telemetry/opentelemetry-collector-contrib#9478 (cc @mx-psi), but more about the conflict between a desire to have optional configs which nonetheless have defaults. Most obvious example is OTLP Receiver, which is often configured like this:
```yaml
receivers:
  otlp:
    protocols:
      grpc:
      http:
```
Here the grpc/http subsections do not have any fields, in particular port numbers, because they are provided by `createDefaultConfig()`. However, it makes it look like one could simple remove e.g. `http` to disable creating an HTTP server, which is not actually true - because the default config still populates `http` even if it's absent in the config.

Now if there were no defaults, this would work fine - both `http` and `grpc` fields are declared as pointers and thus could be nil, indicating an absence. But this is completely circumvented by the `createDefaultConfig()` mechanism.

## Motivation

1. In case of OTLP exporter I might not want to support HTTP endpoint. Right now there's no way to disable it afaiu.
2. In Jaeger components we often have a need to define alternate sub-sections of the config, e.g. remote sampling extension can be backed by a file or by an adaptive calculator. Each of the options requires its own configuration with defaults (e.g. file refresh interval).

## Solution?

I don't have a clear one, but it would be nice to have more granular `createDefault` functions that would only be called once the config is read and a given section is seen as non-empty in the config. As I understand the first pass of the config parsing is by reading YAML into nested maps, which are then decoded into the objects.

mx-psi:
> The issue with (2) is that it only fires when there is a corresponding entry in the input config, so cannot be used to unset the default values already in the struct.

One option is to put it on the configuration consumer to check if the value is none. I think this is what we currently do on the fields where we use 'pointer to T'. 

Another option is to make a type that stores the default value (so, the same as your `Optional` but with a default value if unset). I don't have a good name for that (`Default[T]`? `WithFallback[T]`?), but it would be simple to implement.

yurishkuro:
@mx-psi `WithDefault[T]()` was a good idea, but it didn't work either. When an item is defined in YAML but without any value, i.e. `item:` with no content, then `mapstructure` treats the value as nil, and does not perform any processing. As a result, the custom `Unmarshal()` is still not being called.

[...]

PR #11734, opened 2024-11-23
[WIP][confmap] Enable decoding nil values in confmap

mahadzaryab1:
#### Description
- This PR is a prelude to https://github.com/open-telemetry/opentelemetry-collector/pull/10260 and towards #10266. We made an upstream patch in https://github.com/mahadzaryab1/mapstructure to allow decoding nil values to enable the use of the `Optional` unmarshaller in https://github.com/open-telemetry/opentelemetry-collector/pull/10260. This PR enables the decoding of nil values in `confmap` to unblock the enablement of optional configs. 

#### Link to tracking issue
Towards #https://github.com/open-telemetry/opentelemetry-collector/issues/10266

### DIFF
--- a/cmd/mdatagen/internal/loader_test.go
+++ b/cmd/mdatagen/internal/loader_test.go
@@ -33,7 +33,7 @@ func TestLoadMetadata(t *testing.T) {
 						component.StabilityLevelBeta:        {"traces"},
 						component.StabilityLevelStable:      {"metrics"},
 					},
-					Distributions: []string{},
+					Distributions: nil,
 					Codeowners: &Codeowners{
 						Active: []string{"dmitryax"},
 					},
--- a/cmd/mdatagen/internal/loader_test.go
+++ b/cmd/mdatagen/internal/loader_test.go
@@ -33,7 +33,7 @@ func TestLoadMetadata(t *testing.T) {
 						component.StabilityLevelBeta:        {"traces"},
 						component.StabilityLevelStable:      {"metrics"},
 					},
-					Distributions: nil,
+					Distributions: []string{},
 					Codeowners: &Codeowners{
 						Active: []string{"dmitryax"},
 					},
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -133,7 +133,7 @@ func sanitizeExpanded(a any, useOriginal bool) any {
 		return c
 	case []any:
 		var newSlice []any
-		if reflect.ValueOf(m).IsNil() {
+		if m == nil {
 			return newSlice
 		}
 		newSlice = []any{}
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -330,6 +330,9 @@ func useExpandValue() mapstructure.DecodeHookFuncType {
 // Config{Thing: &SomeStruct{}} instead of Config{Thing: nil}
 func expandNilStructPointersHookFunc() mapstructure.DecodeHookFuncValue {
 	return func(from reflect.Value, to reflect.Value) (any, error) {
+		if !from.IsValid() {
+			return from, nil
+		}
 		// ensure we are dealing with map to map comparison
 		if from.Kind() == reflect.Map && to.Kind() == reflect.Map {
 			toElem := to.Type().Elem()
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -133,6 +133,10 @@ func sanitizeExpanded(a any, useOriginal bool) any {
 		return c
 	case []any:
 		var newSlice []any
+		if reflect.ValueOf(m).IsNil() {
+			return newSlice
+		}
+		newSlice = []any{}
 		for _, e := range m {
 			newSlice = append(newSlice, sanitizeExpanded(e, useOriginal))
 		}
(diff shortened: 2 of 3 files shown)


cmd/mdatagen/internal/loader_test.go
```diff
@@ -33,7 +33,7 @@ func TestLoadMetadata(t *testing.T) {
 						component.StabilityLevelBeta:        {"traces"},
 						component.StabilityLevelStable:      {"metrics"},
 					},
-					Distributions: []string{},
+					Distributions: nil,
```
mahadzaryab1, line 36: @yurishkuro I saw this change was made on https://github.com/open-telemetry/opentelemetry-collector/pull/10260. Do you know why?

yurishkuro: don't recall, probably was failing somewhere

yurishkuro: if you don't need it don't include it

confmap/confmap.go
```diff
@@ -537,7 +541,9 @@ type Marshaler interface {
 func zeroSliceHookFunc() mapstructure.DecodeHookFuncValue {
 	return func(from reflect.Value, to reflect.Value) (interface{}, error) {
 		if to.CanSet() && to.Kind() == reflect.Slice && from.Kind() == reflect.Slice {
-			to.Set(reflect.MakeSlice(to.Type(), from.Len(), from.Cap()))
+			if !from.IsNil() {
```
yurishkuro, line 544: is this check needed? I could see having this at the beginning:
```
		if !from.IsValid() {
			return from, nil
		}
```
But if the Value is valid then calling Len/Cap on nil slice should still be ok.

confmap/confmap.go
```diff
@@ -537,7 +541,9 @@ type Marshaler interface {
...
+				to.Set(reflect.MakeSlice(to.Type(), from.Len(), from.Cap()))
+			}
 		}
 
 		return from.Interface(), nil
```
yurishkuro, line 549: why are you not changing this to `from`? That's the main change that needs to be applied across all hooks, otherwise you never know under which condition they might return untyped nil and break the chain.

confmap/confmap.go
mahadzaryab1: @yurishkuro I agree with you that adding that block in the beginning is the way to go. It also fixes some other failures we were seeing. However, it causes one failure in `confmap_test.go`.
```
    --- FAIL: TestZeroSliceHookFunc/not_overridden_by_nil (0.00s)
        confmap_test.go:688: 
            	Error Trace:	D:/a/opentelemetry-collector/opentelemetry-collector/confmap/confmap_test.go:688
            	Error:      	Not equal: 
            	            	expected: &confmap.structWithSlices{Strings:[]string{"xxx", "yyyy"}}
            	            	actual  : &confmap.structWithSlices{Strings:[]string{}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,5 +1,3 @@
            	            	 (*confmap.structWithSlices)({
            	            	- Strings: ([]string) (len=2) {
            	            	-  (string) (len=3) "xxx",
            	            	-  (string) (len=4) "yyyy"
            	            	+ Strings: ([]string) {
            	            	  }
            	Test:       	TestZeroSliceHookFunc/not_overridden_by_nil
```
Should this test expectation be changed now that we have enabled `DecodeNil`?

yurishkuro: No, there has to be a very high bar for changing test expectations, definitely not because we changed some implementation detail. Does the test make sense? Isn't it validating exactly what the hook promises to do?

confmap/confmaptest/configtest_test.go
```diff
@@ -33,8 +33,7 @@ func TestLoadConf(t *testing.T) {
 func TestToStringMapSanitizeEmptySlice(t *testing.T) {
 	cfg, err := LoadConf(filepath.Join("testdata", "empty-slice.yaml"))
 	require.NoError(t, err)
-	var nilSlice []interface{}
-	assert.Equal(t, map[string]any{"slice": nilSlice}, cfg.ToStringMap())
+	assert.Equal(t, map[string]any{"slice": []interface{}{}}, cfg.ToStringMap())
```
mahadzaryab1, line 36: @yurishkuro What do you think of the change to this expectation here? It looks like it was conflicting with rest of the decoding flow. This is what `empty-slice.yaml` looks like:
```yaml
slice: [] # empty slices are sanitized to nil in ToStringMap
```
Given that `slice` is explicitly identified, shouldn't `slice` be empty instead of nil?

confmap/confmap.go
```diff
@@ -133,6 +133,10 @@ func sanitizeExpanded(a any, useOriginal bool) any {
 		return c
 	case []any:
 		var newSlice []any
+		if reflect.ValueOf(m).IsNil() {
```
yurishkuro, line 136: why do you need to check via reflect package, and not just `m == nil`?

confmap/confmaptest/configtest_test.go
yurishkuro: if there is indeed a contradiction, I would suggest opening a ticket with description, to let the maintainers opine what behavior they expect.

confmap/confmap.go
mahadzaryab1: @yurishkuro 
- We create a nil slice here: https://github.com/go-viper/mapstructure/blob/main/mapstructure.go#L496
- That nil slice gets turned into a concrete value in https://github.com/go-viper/mapstructure/blob/main/decode_hooks.go#L95 which gets passed in https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/confmap.go#L290
As a result of this, we need to dig into the type using reflection to see if it is nil. Let me know if you see an alternative here.

mahadzaryab1: done

PR #11738, opened 2024-11-25
[chore][docs/rfc] Add text regarding conversations

### LINKED ISSUE #11623: Document module split in coding guidelines
On #11216 and others we have taken the approach of splitting existing modules into the parts that we intend to stabilize. This is not documented on our coding guidelines, but it should be!

We want to document the following:
- What are the rough rules of thumb we use for deciding when to split into different modules vs different packages in the same module
- What are some situations where we will always split modules (experimental stuff, testing helpers, see #11216)
- How to go about splitting a new module
  - Roughly, what changes are necessary
  - How to make sure there are no issues during the release because of the new modules

### LINKED ISSUE #11737: Add requirement to solve conversations on RFCs
From private conversation:

> I think it might be reasonable to require that all conversations are solved. However, we could maybe articulate a reasonable way to proceed if a conversation doesn't appear to be going anywhere. e.g. "the author of the RFC may resolve conversations at their discretion but they must explain in the thread why they believe it is appropriate to do so". This pings the people involved and gives them a chance to unresolve/request changes.

mx-psi:
#### Description

Based on the experience of this process going through once, I am suggesting we make the following changes:

1. Require announcing the final comment period via a comment and on the #otel-collector-dev channel.
2. Require that all conversations are resolved. Allow RFC author to resolve any conversation, but require them to leave a comment explaining why

#### Link to tracking issue
Fixes #11737


andrzej-stencel (approved):

codeboten (approved):

evan-bradley (approved):

TylerHelmuth (approved):

docs/rfcs/README.md
```diff
@@ -46,15 +46,19 @@ We use a [Lazy Consensus](https://www.apache.org/foundation/glossary.html#LazyCo
 1. *Quorum*: For an RFC to be mergeable, it needs to have at least **two approvals** from the
    approvers set as well as approvals from any additional stakeholders.
 2. *Waiting period*: Maintainers need to announce their intent to merge the RFC with a GitHub
-   comment. They will need to add a `rfc:final-comment-period` label to the PR and wait for at least
-   **4 business days** after making the announcement to merge the RFC.
+   comment. They will need to add a `rfc:final-comment-period` label to the PR, comment on the PR
+   and note the final comment period in the #otel-collector-dev CNCF Slack channel, and wait for at
+   least **4 business days** after making the announcement to merge the RFC.
```
MovieStoreGuy, line 51: I worry the idea of business days might be fuzzy considering the amount of approvers/maintainers in different geos.

Would it make more sense to say 5 days (this would account for plus or minus a day depending on which part of the world you're in)?

dmitryax (approved):

docs/rfcs/README.md
mx-psi: > I worry the idea of business days might be fuzzy considering the amount of approvers/maintainers in different geos.
> Would it make more sense to say 5 days (this would account for plus or minus a day depending on which part of the world you're in)?

Since this is already part of the text and was already approved at #11416, I think we can discuss this independently from this PR, see #11763

mx-psi:
Given the approvals, this is entering the final comment period. Since we have Thanksgiving I am going to exclude Thursday and Friday and merge on or after 2024-12-04

Aneurysm9 (approved):

bogdandrutu (approved):

Issue #11707, opened 2024-11-19
Automate halting merging in core with a CI check


mx-psi:
We want to automate halting merges with either a bot or some sort of lockfile that is checked by CI

PR #11744, opened 2024-11-25
[chore] Add CI check to enforce merge freezes

jade-guiton-dd:
#### Description

As stated in the [Release Procedure document](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#releasing-opentelemetry-collector), all merging in Core should be halted while the "Prepare release" PR is open. This PR adds a CI job checking the existence of such a PR, allowing us to enforce these freezes automatically.

To make things simple, I decided to distinguish the "Prepare release" PR from others by using a new `prepare:merge-freeze` label. I modified the "Prepare release" action to include that label on the automatically created PR, but in rare cases where maintainers might create the PR themselves, the label can be added manually. Triagers will need to be careful not to accidentally add this label to any random PR to avoid artificial freezes.

Because, in the current state, this CI check is run on each commit of a PR, and at no other time, there is a chance for:
- False positives: If a commit is pushed during a freeze, and the freeze is later rescinded, the PR will remain unmergeable until a new commit is pushed. A simple solution is to re-run the check manually.
- False negatives: If a commit is pushed outside of a freeze, then a freeze is enacted, the PR will remain mergeable despite the freeze. This is more problematic.

To solve this second problem and make this check truly useful, two changes in the repository configuration are necessary:
- the merge queue should be enabled, allowing us to run checks again just before merging;
- this new CI check should be added to the list of required checks for the main branch. 

This will allow us to automatically cancel the merging of PRs during a freeze. A later PR on the community repository should take care of these configuration changes.

Note that another solution to handle outdated CI checks would be to use a tool like [MergeFreeze](https://www.mergefreeze.com), but reading [this document](https://github.com/open-telemetry/community/blob/main/docs/using-github-extensions.md), it sounds like it would be a harder sell with the community than enabling the merge queue.

#### Link to tracking issue
Updates #11707

#### Testing

Simple tests were made on a separate repository to check that enabling the merge queue fixes the false negatives.

### DIFF
--- a/.github/workflows/check-merge-freeze.yml
+++ b/.github/workflows/check-merge-freeze.yml
@@ -10,7 +10,8 @@ on:
 jobs:
   check-merge-freeze:
     name: Check
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:prepare') }}
+    # This condition is to avoid blocking the PR causing the freeze in the first place.
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:merge-freeze') }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/check-merge-freeze.yml
+++ b/.github/workflows/check-merge-freeze.yml
@@ -14,7 +14,7 @@ jobs:
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:merge-freeze') }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           sparse-checkout: .github/workflows/scripts
       - run: ./.github/workflows/scripts/check-merge-freeze.sh
--- a/.github/workflows/check-merge-freeze.yml
+++ b/.github/workflows/check-merge-freeze.yml
@@ -1,0 +1,22 @@
+name: Merge freeze
+
+on:
+  pull_request:
+    types: [opened, ready_for_review, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+  merge_group:
+    types: [checks_requested]
+
+jobs:
+  check-merge-freeze:
+    name: Check
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:prepare') }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/workflows/scripts
+      - run: ./.github/workflows/scripts/check-merge-freeze.sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: open-telemetry/opentelemetry-collector
--- a/.github/workflows/scripts/check-merge-freeze.sh
+++ b/.github/workflows/scripts/check-merge-freeze.sh
@@ -6,5 +6,6 @@
 BLOCKERS=$( gh pr list --search "label:release:merge-freeze" --json url --jq '.[].url' --repo "${REPO}" )
 if [ "${BLOCKERS}" != "" ]; then
     echo "Merging in main is frozen, as there are open PRs labeled 'release:merge-freeze': ${BLOCKERS}"
+    echo "If you believe this is no longer true, re-run this job to unblock your PR."
     exit 1
 fi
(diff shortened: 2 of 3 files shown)


.github/workflows/check-merge-freeze.yml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    types: [opened, ready_for_review, synchronize, reopened, labeled, unlabeled]
+    branches: [main]
+  merge_group:
+    types: [checks_requested]
+
+jobs:
+  check-merge-freeze:
+    name: Check
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:prepare') }}
```
mx-psi, line 13: I wonder if we should call the label something that more explicitly relates to the freeze. Something like `release:halt-merges` or `release:freeze-main`?

mx-psi, line 13: I feel like this warrants a comment explaining how it works, at a first glance one could think that this is checking that the _current_ PR has this label, but it is actually checking IIUC if _any_ open PR has this label

.github/workflows/scripts/release-prepare-release.sh
```diff
@@ -41,7 +41,7 @@ if [ "${CANDIDATE_BETA}" != "" ]; then
 fi
 git push origin "${BRANCH}"
 
-gh pr create --title "[chore] Prepare release ${RELEASE_VERSION}" --body "
+gh pr create --title "[chore] Prepare release ${RELEASE_VERSION}" --label release:prepare --body "
```
mx-psi, line 44: Can you also add a comment here stating why we add the label and referencing the workflow?

.github/workflows/scripts/check-merge-freeze.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+#
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+BLOCKERS=$( gh pr list --search "label:release:prepare" --json url --jq '.[].url' --repo "${REPO}" )
+if [ "${BLOCKERS}" != "" ]; then
+    echo "Merging in main is frozen while release PR is open: ${BLOCKERS}"
```
mx-psi, line 8: Should this more explicitly talk about the label. Something like:
```suggestion
    echo "Merging in main is frozen since there are one or more PRs with the 'release:prepare' label: ${BLOCKERS}"
```

.github/workflows/check-merge-freeze.yml
jade-guiton-dd: No, it is in fact checking the labels of the current PR. We wouldn't want the release PR to be frozen by itself.

jade-guiton-dd: That's a good idea. Maybe `release:freeze` is clear enough?

jade-guiton-dd:
> As I understand it this problem remains even after we make the repository changes. Should we explicitly encourage re-running the check in the error message so that people know how to deal with false positives?

That's probably a good idea.

By the way, do you know who has the rights to re-run jobs?

jade-guiton-dd:
I went with `release:merge-freeze` for the triggering label in the end.

.github/workflows/check-merge-freeze.yml
```diff
@@ -0,0 +1,23 @@
...
+    types: [checks_requested]
+
+jobs:
+  check-merge-freeze:
+    name: Check
+    # This condition is to avoid blocking the PR causing the freeze in the first place.
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:merge-freeze') }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
mx-psi, line 17: Can we use the version we use elsewhere? https://github.com/open-telemetry/opentelemetry-collector/blob/e27c4aeae8da140e286534c69fb9f8231b4b79a2/.github/workflows/changelog.yml#L29

To follow [best practices](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) we refer to the version by its full length commit SHA with a comment specifying the version (tooling like dependabot [support this natively](https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/)):

```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```

mx-psi (approved):

mx-psi:
@open-telemetry/collector-approvers This adds a check to prevent merging other PRs during merge freezes (step 3 of [the release process](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md)), which will be more useful after we have a merge queue on this repository. I will merge this on Thursday EU morning unless somebody objects before then.

PR #11755, opened 2024-11-26
[confmap][fix] Correctly differentiate between nil and empty slices in `ToStringMap`

### LINKED ISSUE #11749: Behaviour of `ToStringMap` Not Matching Confmap Decoding Workflow
This question is related to https://github.com/open-telemetry/opentelemetry-collector/pull/11734 which is working towards https://github.com/open-telemetry/opentelemetry-collector/pull/10260 and #10266. The PR I have open enables the decoding of `nil` values in `mapstructure` so that we can use an Optional type with a default value (see https://github.com/open-telemetry/opentelemetry-collector/pull/10260#issuecomment-2241363814 for more context). While enabling the decoding of nil values, I came across an issue where we were unable to differentiate between nil and empty slices. The reason for that is because we always initialize a nil slice and do not populate it if it is empty (see https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/confmap.go#L135). In order to get around this, I added this patch in the PR above. 
```go
		if m == nil {
			return newSlice
		}
		newSlice = []any{}
```
This allows us to differentiate between a nil slice and an empty slice which we need in the `zeroSliceHookFunc`. However, that patch causes [this](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/confmaptest/configtest_test.go#L33-L38) unit test to fail because there's an expectation here that an empty slice gets rendered as nil when calling `ToStringMap`. I wanted to ask if it was valid to change this expectation so that it matches the decoding workflow we have in which an empty slice is preserved. cc @yurishkuro

mahadzaryab1:
#### Description
- Background
  - The confmap decoder hook has a `zeroSliceHookFunc` which treats empty slices provided in the config as empty slices rather than nil
  - However in `confmap.ToStringMap`, empty slices were getting rendered as nil which is different from the decoder hook as described above. 
- This PR fixes `confmap.ToStringMap` by adding a check in the sanitizer function that returns an empty slice when one is provided. 
- This PR is a prelude to https://github.com/open-telemetry/opentelemetry-collector/pull/11734 to allow for the decoding of nil values in `confmap` to enable optional configs.

#### Link to tracking issue
Fixes #11749

#### Testing
- Adjusted the existing unit test and added a new unit test


confmap/confmap.go
```diff
@@ -133,6 +133,10 @@ func sanitizeExpanded(a any, useOriginal bool) any {
 		return c
 	case []any:
 		var newSlice []any
+		if m == nil {
+			return newSlice
+		}
```
mahadzaryab1, line 138: @mx-psi This block is currently not covered by tests because `DecodeNil` because we're not decoding nil values PR (that will happen in https://github.com/open-telemetry/opentelemetry-collector/pull/11734). We could remove this if block here and add it in the following PR or leave it in here but merge with missing code coverage or add a unit test for this function directly. Let me know what you think.

mx-psi:
Let's wait to discuss this with @songy23

yurishkuro:
The description is confusing to me

> which differs from the behaviour of confmap's decoding workflow

What does this tell me?

What I would rather see is something like this:
  * the hook's description says 
    * empty slice `[]` in yaml should override the default value assigned to config
    * but a nil slice in yaml should fall back to default value
  * but in practice the code was doing this: ...

yurishkuro (approved):

mahadzaryab1:
> The description is confusing to me
> 
> > which differs from the behaviour of confmap's decoding workflow
> 
> What does this tell me?
> 
> What I would rather see is something like this:
> 
> * the hook's description says
>   
>   * empty slice `[]` in yaml should override the default value assigned to config
>   * but a nil slice in yaml should fall back to default value
> * but in practice the code was doing this: ...

@yurishkuro Thanks for the feedback! I updated the description. One thing to note is that the discrepancy is not within the hook but rather `ToStringMap` which is more of a helper function to create a `map[string]any` from a `Conf`. This PR matches the behaviour of `ToStringMap` with the behaviour of the hook and more specifically with that of `zeroSliceHookFunc`.

confmap/confmaptest/configtest_test.go
```diff
@@ -30,11 +30,16 @@ func TestLoadConf(t *testing.T) {
 	assert.Equal(t, map[string]any{"floating": 3.14}, cfg.ToStringMap())
 }
 
+func TestToStringMapSanitizeNil(t *testing.T) {
+	cfg, err := LoadConf(filepath.Join("testdata", "nil.yaml"))
+	require.NoError(t, err)
+	assert.Equal(t, map[string]any{"slice": nil}, cfg.ToStringMap())
+}
+
 func TestToStringMapSanitizeEmptySlice(t *testing.T) {
```
songy23, line 39:
```suggestion
func TestToStringMapEmptySlice(t *testing.T) {
```

The test name should change

.chloggen/nil-empty-slice.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
```
songy23, line 4:
```suggestion
change_type: 'breaking'
```

am not sure if the previous behavior is intentional, but this PR is a breaking change to confmap users

songy23 (approved):
OK with this change

.chloggen/nil-empty-slice.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'breaking'
```
mx-psi, line 4: I was okay merging this with the expectation that we do not consider this a breaking change, but rather a change in an unspecified part of `confmap`. If we think this is a breaking change, then I am against merging this, since confmap is 1.0

songy23: I'm a bit on the fence here. 
- On one hand this does change the current output data type of confmap so it will break users who rely on the type or format. 
- On the other hand,  the current behavior is subtle  - it is not documented or specified anywhere. And from the PR description it sounds like the current behavior should not have been there in the first place. I'm OK with changing this back to bug_fix if treating it as a breaking change will block it.

confmap/confmaptest/configtest_test.go
```diff
@@ -30,11 +30,16 @@ func TestLoadConf(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.Equal(t, map[string]any{"slice": nil}, cfg.ToStringMap())
+}
+
+func TestToStringMapEmptySlice(t *testing.T) {
 	cfg, err := LoadConf(filepath.Join("testdata", "empty-slice.yaml"))
 	require.NoError(t, err)
-	var nilSlice []interface{}
-	assert.Equal(t, map[string]any{"slice": nilSlice}, cfg.ToStringMap())
+	assert.Equal(t, map[string]any{"slice": []interface{}{}}, cfg.ToStringMap())
```
evan-bradley, line 42:
```suggestion
	assert.Equal(t, map[string]any{"slice": []any{}}, cfg.ToStringMap())
```
Minor, but this is more stylistically conventional for the repo now, and easier to read in my opinion.

.chloggen/nil-empty-slice.yaml
evan-bradley: I've finally spent some time digging into this, and I think we should consider merging this after we define what we want to be the correct behavior. In my opinion, the current behavior is undefined and that's a bug in confmap, even if changing the existing behavior will break users. Personally, I'd like to keep a 1:1 mapping between YAML <-> `map[string]any` <-> `confmap.Conf` <-> Config struct as much as is reasonable.

I'm also not sure if we should keep `zeroSliceHookFunc`: removing it from our list of decoder hooks only breaks a single test in cmd/mdatagen, and that test passes with the hook removed if the changes in this PR are applied. In general, we should likely try to keep a distinction between `[]any(nil)` and `any{}`, whereas the hook currently maps `[]any(nil)` to `[]any{}` (and `ToStringMap`, without the changes in this PR, maps `[]any{}` to `[]any(nil)`). Additionally, if I understand the changes in https://github.com/open-telemetry/opentelemetry-collector/pull/11734, I'm not 100% sure the hook does much anymore.

evan-bradley: I created a PR to illustrate a little of what I have in mind here: https://github.com/open-telemetry/opentelemetry-collector/pull/11882.

I ran the k8s attributes processor tests with this change and didn't see any failing tests from removing `zeroSliceHookFunc`.

mahadzaryab1: @evan-bradley thanks for the great explanation and the illustration in the PR. Let me know how you think we should proceed here. Do we want to replace my PR with the one you have open?

mahadzaryab1:
Hi @mx-psi @evan-bradley - just wanted to check in on this PR. Is this still something we can push forward?

mx-psi:
Apologies for dropping this, we are re-doing it over at #11882

Issue #11562, opened 2024-10-29
Add codeowners tooling to this repository


mx-psi:
We have some tooling on the opentelemetry-collector-contrib that allows us to automate some actions related to codeowners:
- [ping-codeowners-prs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/ping-codeowners-prs.yml)
- [ping-codeowners-on-new-issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/ping-codeowners-on-new-issue.yml)
- [ping-codeowners-issues.yml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/ping-codeowners-issues.yml)
- [add-codeowners-to-pr](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/add-codeowners-to-pr.yml)

We can add the tooling one by one. Some of these may need additional tooling (more issue templates?):

- [ ] Ping code owners on PRs for a component
- [ ] Ping code owners on new issues
- [ ] Ping code owners on existing issues
- [ ] Add code owners as reviewers in a PR

mowies:
i'd like to work on this :)

mowies:
@mx-psi so after a quick review of the 4 links above, I noticed that probably only the issue related things will really work on this repo. That's because the [codeowners file](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/CODEOWNERS) is not set up as granular here compared to the contrib repo so the PR related workflows, that would ping codeowners for specific folders would just generate a ton of notifications for the collector maintainers.

mx-psi:
@mowies That's fair. There will be a long transition period in which a lot of things do not have a codeowner. I think we should ignore the `*` line and only trigger automation if we are on a more specific line.

I am working on a PR that should be ready on Wednesday adding more specific codeowners, maybe after that is ready it's a good time to see how this would work in practice

mowies:
The automations only look for specific labels as components anyways, so they already ignore the `*` case.

mowies:
For some of the automations to work, I would also need to adjust the existing issue templates and probably mostly exchange them for the ones from the contrib repo. Should I add that to this PR or rather create a separate one?

mx-psi:
> The automations only look for specific labels as components anyways, so they already ignore the * case.

[...]

PR #11756, opened 2024-11-26
[chore] add githubgen to enhance codeowners and issue templates

mowies:
### This PR
- adds the githubgen tool as a dependency in internal/tools
- uses githubgen to generate codeowners and issue template files
- updates lots of metadata files by
  - taking the existing codeowners file and feeding the info from there back into the component metadata.yaml files or creating new metadata.yaml files where none existed yet
- adds distributions.yaml as a basis the mostly already existing `distributions:` keys in metadata.yaml files (needed for githubgen to work correctly)
- adds relevant make commands to make the githubgen tool usage mostly transparent to users

This change is a prerequisite to be able to ping codeowners reliably with automated tooling as a next step.

Part of #11562


cmd/mdatagen/internal/sampleprocessor/metadata.yaml
```diff
@@ -15,7 +15,7 @@ status:
   distributions: []
   unsupported_platforms: [freebsd, illumos]
   codeowners:
-    active: [some]
+    active: []
```
mowies, line 18: removed since `some` is an actual user and not affiliated or related to OTel

.github/CODEOWNERS
```diff
@@ -1,19 +1,30 @@
...
+# https://help.github.com/en/articles/about-code-owners
 #
 
 * @open-telemetry/collector-approvers
 
 # List all maintainers individually so all of them are notified
 .github/CODEOWNERS @codeboten @BogdanDrutu @dmitryax @mx-psi
-docs/rfc           @codeboten @BogdanDrutu @dmitryax @mx-psi
+
+cmd/githubgen/                         @open-telemetry/collector-approvers @atoulme
```
mowies, line 21: hope this fine since @atoulme is also maintainer for the tool in contrib :)

docs/rfcs/metadata.yaml
```diff
@@ -0,0 +1,10 @@
+type: rfcs
```
mowies, line 1: i added this for completeness sake and to make githubgen work correctly. owners list taken from current codeowners file

evan-bradley:
>Follow ups: move the githubgen tool to a central place so that we don't need to duplicate it in contrib and core repos

https://github.com/open-telemetry/opentelemetry-go-build-tools/ is probably the right place. Could we just pull the module from contrib (like we do with chloggen) and if we move it update the reference? Referencing it in the `internal/tools` package should be fairly straightforward.

mowies:
@evan-bradley that would be possible but I made a few minor changes to the tool to make it work with this repo (e.g. changing the file templates), so that would need to be parameterized in the original tool first to be able to do that. I wanted to do exactly that in a follow up issue afterwards to be able to clearly see what we need to change in the tool to make it general enough to serve both or even more repos in the future.

mowies:
I filed https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/36691 to take care of moving githubgen to the build-tools repo

.github/CODEOWNERS
atoulme: feel free to put your name here too.

atoulme:
yes, go-build-tools seems like a better home.

mowies:
this is still active

atoulme:
Looking good to me, how about getting it out of draft and fixing conflicts, and then we can get it in?

mowies:
yes! but ideally, i'd like to merge and maybe also release https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/655 first

receiver/xreceiver/metadata.yaml
```diff
@@ -0,0 +1,9 @@
+type: profile
```
mx-psi, line 1: While this type is correct today, `xreceiver` will eventually also have the entities implementation. Maybe `type: experimental`?

service/internal/graph/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+type: profile
```
mx-psi, line 1: This is type does not make a lot of sense to me, but I am not sure which one to suggest

mowies: this is probably a copy/paste error. i'll try to find something else to put there

receiver/xreceiver/metadata.yaml
mowies: is this something to change for all of the `x*` components?

mx-psi: Yes!

service/internal/graph/metadata.yaml
mowies: i adapted this to class: service, type: graph
maybe that works

receiver/xreceiver/metadata.yaml
mowies: i updated this again. all the x* components now have type: <the component name> and class: receiver|processor|...
I think that is more appropriate and more in line with the other components

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,123 @@
...
+      # NOTE: The list below is autogenerated using `make generate-gh-issue-templates`
+      # Do not manually edit it.
+      # Start Collector components list
+      - cmd/mdatagen
+      - cmd/mdatagen/internal/sampleprocessor
+      - cmd/mdatagen/internal/samplereceiver
+      - cmd/mdatagen/internal/samplescraper
+      - confmap
+      - connector/forward
+      - connector/x
```
andrzej-stencel, line 27: This should be `connector/xconnector`, right?
```suggestion
      - connector/xconnector
```

andrzej-stencel (approved):

.github/ISSUE_TEMPLATE/bug_report.yaml
mowies: this is kind of a limitation of the githubgen tool currently. I didn't get around to fix it yet, but I think for now it should be fine. I don't think those components are selected often if at all in the issue templates.

mx-psi (approved):

mx-psi:
I will merge this in one week if there are no further comments cc @open-telemetry/collector-approvers

songy23 (approved):

atoulme (approved):

PR #11769, opened 2024-11-28
Add confighttp.WithOtelHTTPOptions

### LINKED ISSUE #11770: Allow confighttp.ToServer() to accept options for otelhttp instrumentation
When a component creates an HTTP server using ToServer, it may need the ability to customize otelhttp instrumentation that is automatically applied.

Related issue: https://github.com/jaegertracing/jaeger/issues/6265

A new option to allow passing otelhttp options to otelhttp instrumentation, such as a filter.

yurishkuro:
#### Description
Add ability to provide OTel HTTP intrumentation options to `ToServer()` function.

#### Link to tracking issue
Fixes #11770
Related to https://github.com/jaegertracing/jaeger/issues/6265

#### Testing

Added unit test


mx-psi (commented):
`otelhttp` is not 1.0 and has no plans to be 1.0 in the foreseeable future. Introducing this as part of the public API of `confighttp` would prevent us from marking it as 1.0 until `otelhttp` is marked as 1.0.

Is there a way to introduce this so that `WithOtelHTTPOptions` either does not depend on a symbol in `otelhttp` or is not in the `confighttp` package?

mx-psi:
> > Introducing this as part of the public API of confighttp would prevent us from marking it as 1.0 until otelhttp is marked as 1.0.
> 
> * The chance that otelhttp would change the "options" concept is near zero.

Agreed

> * If they do, we can always deprecate the new option here or make it no-op.

Worst case scenario is that the concept is removed altogether. I think this is extremely unlikely, but we wouldn't have a non-breaking way of dealing with it.

> * This module depends on plenty of other 0.x modules, both internal and external e.g. github.com/golang/snappy v0.0.4

The particular rule we are breaking is this one [from the VERSIONING.md doc](https://github.com/open-telemetry/opentelemetry-collector/blob/main/VERSIONING.md#general-go-api-considerations):

> **Dependency updates**. Updating dependencies is not considered a breaking change except when their types are part of the public API or the update may change the behavior of applications in an incompatible way.

I am fine accepting this change if we first amend VERSIONING.md to reflect the new policy we are following.

mx-psi:
One way of doing this without changing `VERSIONING.md` is to move this option to a separate module. It feels like an overkill to have a separate module for just one option but I guess modules are free :shrug: This is basically the pattern we have here: https://pkg.go.dev/go.opentelemetry.io/collector/pipeline/pipelineprofiles

yurishkuro:
> to move this option to a separate module

don't think it will work, the options operate on a private struct in `confighttp`

mx-psi:
> > to move this option to a separate module
> 
> don't think it will work, the options operate on a private struct in `confighttp`

The way to do this is to move the struct to an internal package, define all options in that internal package and re-export them in the appropriate module (the `otelhttp` one in a separate module, the others in `confighttp`)

yurishkuro:
I don't follow that. Internal package would have to be in _some_ module, so another module would not be able to reference it.

But if we do it in the same module with an experimental sub-package then it could work, e.g `confighttp.confighttpx.WithOtelHTTPOptions`

mx-psi:
@yurishkuro Here is a more explicit explanation of the pattern: https://dmathieu.com/en/development/go-architecture-semver/stable-experimental-modules/ I think we are talking about the same thing

yurishkuro:
@mx-psi I move it to new experimental module, lmk if any concerns

mx-psi (approved):
This looks good to me (modulo solving the EasyCLA thing), but as I commented on #11779 I think we should settle in a convention for experimental module names by opening a PR [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md). I personally like the `exp` suffix better

yurishkuro:
@mx-psi https://github.com/open-telemetry/opentelemetry-collector/issues/11783

mx-psi (commented):
Can you update the changelog to reference that this is in `xcongihttp`? LGTM otherwise :)

PR #11772, opened 2024-11-28
Define observability requirements for stable components

### LINKED ISSUE #11581: Document observability requirements for stable components
We want to document observability requirements for stable components. We probably want to do it under [the 'stable' section](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#stable) in the component stability document.

jade-guiton-dd:
## Description

This PR defines observability requirements for components at the "Stable" stability levels. The goal is to ensure that Collector pipelines are properly observable, to help in debugging configuration issues.

#### Approach

- The requirements are deliberately not too specific, in order to be adaptable to each specific component, and so as to not over-burden component authors.
- After discussing it with @mx-psi, this list of requirements explicitly includes things that may end up being emitted automatically as part of the Pipeline Instrumentation RFC (#11406), with only a note at the beginning explaining that not everything may need to be implemented manually.

Feel free to share if you don't think this is the right approach for these requirements.

#### Link to tracking issue
Resolves #11581

## Important note regarding the Pipeline Instrumentation RFC

I included this paragraph in the part about error count metrics:
> The goal is to be able to easily pinpoint the source of data loss in the Collector pipeline, so this should either:
>   - only include errors internal to the component, or;
>   - allow distinguishing said errors from ones originating in an external service, or propagated from downstream Collector components.

The [Pipeline Instrumentation RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md) (hereafter abbreviated "PI"), once implemented, should allow monitoring component errors via the `outcome` attribute, which is either `success` or `failure`, depending on whether the `Consumer` API call returned an error.

Note that this does not work for receivers, or allow differentiating between different types of errors; for that reason, I believe additional component-specific error metrics will often still be required, but it would be nice to cover as many cases as possible automatically.

However, at the moment, errors are (usually) propagated upstream through the chain of `Consume` calls, so in case of error the `failure` state will end up applied to all components upstream of the actual source of the error. This means the PI metrics do not fit the first bullet point.

Moreover, I would argue that even post-processing the PI metrics does not reliably allow distinguishing the ultimate source of errors (the second bullet point). One simple idea is to compute `consumed.items{outcome:failure} - produced.items{outcome:failure}` to get the number of errors originating in a component. But this only works if output items map one-to-one to input items: if a processor or connector outputs fewer items than it consumes (because it aggregates them, or translates to a different signal type), this formula will return false positives. If these false positives are mixed with real errors from the component and/or from downstream, the situation becomes impossible to analyze by just looking at the metrics.

For these reasons, I believe we should do one of four things:
1. Change the way we use the `Consumer` API to no longer propagate errors, making the PI metric outcomes more precise.
  We could catch errors in whatever wrapper we already use to emit the PI metrics, log them for posterity, and simply not propagate them.
  Note that some components already more or less do this, such as the `batchprocessor`, but this option may in principle break components which rely on downstream errors (for retry purposes for example).
3. Keep propagating errors, but modify or extend the RFC to require distinguishing between internal and propagated errors (maybe add a third `outcome` value, or add another attribute).
  This could be implemented by somehow propagating additional state from one `Consume` call to another, allowing us to establish the first appearance of a given error value in the pipeline.
5. Loosen this requirement so that the PI metrics suffice in their current state.
6. Leave everything as-is and make component authors implement their own somewhat redundant error count metrics.


docs/component-stability.md
```diff
@@ -66,6 +66,74 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+#### Observability requirements
+
+Stable components should emit enough internal telemetry to let users detect errors, as well as data
+loss and performance issues inside the component, and to help diagnose them if possible.
+
+The internal telemetry of a stable component should allow observing the following:
+
+1. How much data the component receives.
+
+  For receivers, this could be a metric counting requests, received bytes, scraping attempts, etc.
```
mx-psi, line 78: What would this metric measure for the [hostmetrics receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver)? Would the measurement be different depending on the OS?

docs/component-stability.md
```diff
@@ -66,6 +66,74 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+
+  - How much data is created by the component.
+
+  - How much data is currently held by the component (eg. an UpDownCounter keeping track of the
+    size of an internal queue).
+
+6. Processing performance.
+
+  This could be a histogram of end-to-end component latency, measured as the time between external
+  requests or `Consumer` API calls.
```
mx-psi, line 121: I feel like this is something that would be capture by the RED metrics from a component like the [spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector#overview). Should we instead require that the main loop of the component is be traced?

docs/component-stability.md
```diff
@@ -66,6 +66,74 @@ Stable components MUST be compatible between minor versions unless critical secu
 component owner MUST provide a migration path and a reasonable time frame for users to upgrade. The same rules from beta
 components apply to stable when it comes to configuration changes.
 
+#### Observability requirements
```
mx-psi, line 69: You mention a couple of times the concept of items. It may make sense to define this first at the beginning of this section so you don't have to explain it multiple times

docs/component-stability.md
```diff
@@ -66,6 +66,74 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+
+6. Processing performance.
+
+  This could be a histogram of end-to-end component latency, measured as the time between external
+  requests or `Consumer` API calls.
+
+When measuring amounts of data, counting data items (spans, log records, metric points) is
+recommended. Where this can't easily be done, any relevant unit may be used, as long as zero is a
+reliable indicator of the absence of data. In any case, the type of all metrics should be properly
+documented (not "1").
```
mx-psi, line 126: I think you are talking here about the metric unit, not its type, right?
```suggestion
reliable indicator of the absence of data. In any case, all metrics should have a defined unit (not "1").
```

docs/component-stability.md
mx-psi, line 1: The clarifications for each item in the list are not rendered as part of the item, could you fix this?

docs/component-stability.md
```diff
@@ -66,6 +66,74 @@ Stable components MUST be compatible between minor versions unless critical secu
 component owner MUST provide a migration path and a reasonable time frame for users to upgrade. The same rules from beta
 components apply to stable when it comes to configuration changes.
 
+#### Observability requirements
+
+Stable components should emit enough internal telemetry to let users detect errors, as well as data
+loss and performance issues inside the component, and to help diagnose them if possible.
+
+The internal telemetry of a stable component should allow observing the following:
```
mx-psi, line 74: The text seems to exclude extensions, I would clarify it like this:
```suggestion
The internal telemetry of a stable pipeline component should allow observing the following:
```

mx-psi (commented):
Thanks! Left a few comments. I think we also want to clarify this is not an exhaustive list: components may want to add other telemetry if it makes sense

docs/component-stability.md
jade-guiton-dd: I think "scraping attempts", or maybe "number of metric points" scraped could make sense for the `hostmetrics` receiver, and those would not need to differ based on the OS.

I don't think we need to be much more specific than that, as the "data received" metrics for receiver only really seem useful if you also have a comparable "data lost" metric.

docs/component-stability.md
jade-guiton-dd: If by "main loop" you mean everything from after the component entry point (HTTP/gRPC/whatever server for receivers, `Consume` for others) to before the exit point (HTTP/gRPC/whatever request for exporters, `Consume` for others), I think it would probably be doable. The requirement cites histograms as an example, so this would not technically be excluded.

I avoided requiring spans because the PI RFC was pretty frilly about them and preferred logs when metrics are insufficient. If we're considering using spans again, we may also want to reconsider outputting errors as logs and encourage emitting them as span events instead.

docs/component-stability.md
```diff
@@ -66,6 +66,74 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+
+  For other components, this would typically be the number of items forwarded through the `Consumer`
+  API.
+
+3. How much data is dropped because of errors.
+
+  For receivers, this could include a metric counting payloads that could not be parsed in.
+  
+  For receivers and exporters, this could include a metric counting requests that failed because
+  of network errors.
```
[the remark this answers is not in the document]
jade-guiton-dd: That's why this is a "could" statement, it only applies for receivers and exporters which use the network. Do you think I should add a note about how these example implementations are non-normative?

docs/component-stability.md
mx-psi: Makes sense to me, we can look into adding builtin telemetry for the scraperhelper

docs/component-stability.md
mx-psi: Discussed offline, we can probably add some wording around being able to observe performance and errors but it will be quite generic

docs/component-stability.md
```diff
@@ -66,6 +66,99 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+information can be derived from the emitted telemetry; everything after the basic category
+description is a recommendation, and is not normative.
+
+- Of course, a component may define additional internal telemetry which is not in this list.
+
+- Some of this internal telemetry may already be provided by pipeline auto-instrumentation or
+helper modules (such as `receiverhelper`, `scraperhelper`, `processorhelper`, or `exporterhelper`).
+Please check the documentation to verify which parts, if any, need to be implemented manually.
+
+**Definition:** In the following, an "item" refers generically to a single log record, metric event,
```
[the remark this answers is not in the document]
jade-guiton-dd: Yes. I tried searching the docs for the "official" terminology, but I think I got confused. I'll change it back to "metric point" if that sounds better.

docs/component-stability.md
```diff
@@ -66,6 +66,99 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+
+4. Details for error conditions.
+
+    This could be in the form of logs or spans detailing the reason for an error. As much detail as
+    necessary should be provided to ease debugging. Processed signal data should not be included for
+    security and privacy reasons.
+
+5. Other possible discrepancies between input and output, if any. This may include:
+
+    - How much data is dropped as part of normal operation (eg. filtered out).
```
jaronoff97, line 136: In the past we've discussed data dropped being its own metric so you could understand the reason for the discrepancy (i.e. maybe a filter processor rule match vs admission control), with this today `reason` wouldn't necessarily be clear. Does that matter? If not, why not?

jaronoff97: it's mentioned below that some attributes could be added, but i'm wondering if there should be more guidance around it.

docs/component-stability.md
```diff
@@ -66,6 +66,99 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+    This could include spans for each operation of the component, or a histogram of end-to-end
+    component latency.
+    
+    The goal is to be able to easily pinpoint the source of latency in the Collector pipeline, so
+    this should either:
+    - only include time spent processing inside the component, or;
+    - allow distinguishing this latency from that caused by an external service, or from time spent
+        in downstream Collector components.
+
+When measuring amounts of data, counting items is recommended. Where this can't easily be done, any
```
jaronoff97, line 154: are there any metrics where we should explicitly recommend an updowncounter/gauge?

docs/component-stability.md
```diff
@@ -66,6 +66,99 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+    security and privacy reasons.
+
+5. Other possible discrepancies between input and output, if any. This may include:
+
+    - How much data is dropped as part of normal operation (eg. filtered out).
+
+    - How much data is created by the component.
+
+    - How much data is currently held by the component (eg. an UpDownCounter keeping track of the
+        size of an internal queue).
```
jaronoff97, line 141: is it worth defining a set of internal telemetry for queueing / batching? i.e. things that hold on to items?

docs/component-stability.md
```diff
@@ -66,6 +66,99 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+    this should either:
+    - only include time spent processing inside the component, or;
+    - allow distinguishing this latency from that caused by an external service, or from time spent
+        in downstream Collector components.
+
+When measuring amounts of data, counting items is recommended. Where this can't easily be done, any
+relevant unit may be used, as long as zero is a reliable indicator of the absence of data. In any
+case, all metrics should have a defined unit (not "1").
+
+If data can be dropped/created/held at multiple distinct points in a component's pipeline (eg.
```
jaronoff97, line 158: is there a recommendation about when it's okay to break a trace? Right now traces can be broken in components that do batching/queueing, and I think it would be useful to recommend that breaking the context should be avoided or a spanlink should be added.

docs/component-stability.md
jade-guiton-dd: For components that can intentionally drop data for multiple reasons, adding attributes or defining multiple metrics to distinguish between them could potentially be useful. But I'm not sure we want to explicitly require it; I would assume that in most cases "this data was dropped intentionally, no errors to see here" is enough for the end user, and I'd like to keep the burden on component authors as light as possible.

docs/component-stability.md
jade-guiton-dd: "Counting" in this sentence was not meant to recommend using a Counter, it was only meant to specify that we want to measure in units of "items" rather than bytes or something else. Since it doesn't seem to be clear, I'll reword this sentence.

Notably, "how much data is *currently* being held by the component" could very naturally be implemented with an UpDownCounter. I'll add it explicitly as an example implementation.

docs/component-stability.md
jade-guiton-dd: Do you mean we should define *more specific* requirements? It sounds like many components may have an internal queue, so I don't want to increase the component author's burden too much. But we could perhaps suggest a gauge for the queue's capacity on top of its current size?

Or do you mean that maybe this category is not worth it? For the record, we already emit similar telemetry for the internal queues of exporters relying on `exporterhelper`.

docs/component-stability.md
jade-guiton-dd: That sounds good. Following the recommendation that spans should "only include time spent processing inside the component", I think we should *recommend* having multiple spans when data is queued internally, but that the two spans must be either in the same trace (no breaking the context, which may be complex to implement), or have a link between them. I'll add that if it sounds good to you.

docs/component-stability.md
jaronoff97: yeah more specific requirements, essentially codifying how exporterhelper does this today to make it clear to end-users where component authors are using internal queues. This is really vital for understanding backpressure from an end-user perspective IMO.

docs/component-stability.md
jaronoff97: yep, sounds reasonable. thank you!

docs/component-stability.md
jade-guiton-dd: I added a requirement to emit the queue's capacity (if it is fixed), and added "size as an updowncounter and capacity as a gauge" (the current `exporterhelper` implementation) as the example implementation.

mx-psi (approved):

docs/component-stability.md
```diff
@@ -66,6 +66,110 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+
+- However, for each category, many reasonable implementations are possible, as long as the relevant
+information can be derived from the emitted telemetry; everything after the basic category
+description is a recommendation, and is not normative.
+
+- Of course, a component may define additional internal telemetry which is not in this list.
+
+- Some of this internal telemetry may already be provided by pipeline auto-instrumentation or
+helper modules (such as `receiverhelper`, `scraperhelper`, `processorhelper`, or `exporterhelper`).
+Please check the documentation to verify which parts, if any, need to be implemented manually.
```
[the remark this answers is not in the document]
jade-guiton-dd: Right, I discovered those yesterday haha. I'll switch to that, it'll look better.

docs/component-stability.md
```diff
@@ -66,6 +66,108 @@ Stable components MUST be compatible between minor versions unless critical secu
...
+    `Consumer` API.
+
+2. How much data the component outputs.
+
+    For exporters, this could be a metric counting requests, sent bytes, etc.
+
+    For other components, this would typically be the number of items forwarded to the next
+    component through the `Consumer` API.
+
+3. How much data is dropped because of errors.
```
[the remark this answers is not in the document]
jade-guiton-dd: We probably want to require the use of the RFC's conventions for component-identifying attributes, and I will definitely include using an `outcome` attribute on the input metric instead of a separate error metric as a recommended implementation.

However, if we want to incentivize contributing external components, I don't think we want to require strict adherence to all of the RFC's choices, so divergences are somewhat inevitable. What do you think?

Relatedly, have you read the "Important note" about the RFC in the PR description? I'm interested in feedback on the best course of action.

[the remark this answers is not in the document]
mx-psi: > Per @djaglowski 's RFC [here](https://github.com/open-telemetry/opentelemetry-collector/pull/11406/files) I think this would just be an attribute on the output metric called outcome?

That would be the most natural way to go about this. I feel like this document should not be too prescriptive as to how to accomplish the requirements listed, but making a recommendation like this would make sense to me to ensure consistency across components.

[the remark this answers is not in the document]
jaronoff97: @jade-guiton-dd I hadn't seen that yet, thanks for bringing my attention to it, i think i had only seen the initial description.

I would strongly advise against option 1 as error back propagation is key if you are running a collector in gateway mode and you want to propagate backpressure to an agent. I think options 2 or 3 are sufficient, option 4 feels not prescriptive enough IMO.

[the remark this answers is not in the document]
jaronoff97: Given this is the recommendation for a component, it makes sense to have the component author use a custom error metric that they can decide to either include or exclude any downstream errors as part of it. (this is what you have written, and i agree with it 😄)

jaronoff97 (approved):

docs/component-stability.md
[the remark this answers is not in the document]
jade-guiton-dd: > they can decide to either include or exclude any downstream errors as part of it. (this is what you have written, and i agree with it 😄)

To be clear, the current requirements allow including downstream errors in a custom error metric, but there has to be some way to distinguish them from internal errors.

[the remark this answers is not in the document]
jaronoff97: yep, this makes sense to me. Thank you! :bow:

djaglowski (approved):

docs/component-stability.md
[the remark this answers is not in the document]
jade-guiton-dd: Just to be sure @djaglowski, do you support option 2 I detailed in the PR description, ie. amending the Pipeline Instrumentation RFC to require the implementation to distinguish errors coming directly from the next pipeline component from errors propagated from components further downstream, in order to fit the last paragraph of point 3?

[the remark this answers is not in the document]
djaglowski: I think it makes sense in principal, as long as there is a clear mechanism for communicating this information so that instrumentation that is automatically wrapped around components can unambiguously know the correct outcome.

Issue #11531, opened 2024-10-23
confmap env provider fails when trying to set a default value


dloucasfx:
**Describe the bug**
When attempting to define an env variable with default value, ex: `'${env:app:-""}'`, the confmap env provider fails the validation https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider/envprovider/provider.go#L51

**Steps to reproduce**
Use the below config and make sure that `app` env variable is NOT set

**What did you expect to see?**
The collector should start and a resource attribute `app` with empty string should be added

**What did you see instead?**
The collector exited and this gets logged

````
Error: failed to get config: cannot resolve the configuration: configsource provider failed retrieving: environment variable "app:-\"\"" has invalid name: must match regex ^[a-zA-Z_][a-zA-Z0-9_]*$
2024/10/23 18:20:36 main.go:92: application run finished with error: failed to get config: cannot resolve the configuration: configsource provider failed retrieving: environment variable "app:-\"\"" has invalid name: must match regex ^[a-zA-Z_][a-zA-Z0-9_]*$

````

**What version did you use?**
main

**What config did you use?**
```
receivers:
  prometheus/internal:
    config:
      scrape_configs:
      - job_name: 'otel-collector'
        scrape_interval: 10s
        static_configs:
        - targets: ['0.0.0.0:8888']
        metric_relabel_configs:
          - source_labels: [ __name__ ]
            regex: '.*grpc_io.*'
            action: drop

processors:
  resource:
    attributes:
      - key: app
        value: '${env:app:-""}'
        action: insert

exporters:
  logging:
    loglevel: debug

service:
  pipelines:
    metrics:
      receivers: [prometheus/internal]
      processors: [resource]
      exporters: [logging]
```

**Environment**

**Additional context**

**Comment by VihasMakwana** [comment]
@dloucasfx can you confirm the latest commit you're on? I tried to reproduce this using v0.112.0 of `opentelemetry-collector-contrib` and it works as expected.

**Comment by pebo** [comment]
Tested fallback to default with `v0.112.0` and it works if the env var is not set, however `:-` does not trigger the fallback if the env var is set with an empty string as the value.

Example:
```yaml
            - key: foo
              # Empty values are not allowed so use "novalue" as default
              value: ${env:FOO:-novalue}
```

If `FOO` is not set the value will be set to `novalue`. If `FOO` is set to the empty string we'll get something like:

`
[...]

PR #11776, opened 2024-11-29
[envprovider] - If env is set to empty string, fallback to default

VihasMakwana:
See https://github.com/open-telemetry/opentelemetry-collector/issues/11531#issuecomment-2454200937

In my opinion, we should fallback to default when either:
1. env variable is unset
2. env variable is set, but is an empty string


dmitryax (changes requested):
It needs discussion before changing the behaviour. See https://github.com/open-telemetry/opentelemetry-collector/issues/11531#issuecomment-2513127544

VihasMakwana:
After some thought over Dmitry's comment,

> There must be a way to pass an empty string even if there is a default.

I agree. I'll close this issue. I think it's better to consider the empty string because the user might have explicitly set it to "".

Issue #7443, opened 2023-03-28
[CI] 'Inform Incompatible PRs / Check-Compatibility' doesn't work


mx-psi:
The `Inform Incompatible PRs / Check-Compatibility` CI check passes on PRs that make breaking changes. An example can be seen on #7440 and in #7442.

One problem with the check right now is that it only accounts for packages on the main module: the variable `ALL_PKGS` is built from `go list ./...` that only lists packages on the top-level module: https://github.com/open-telemetry/opentelemetry-collector/blob/92e542357151fb3ae237035322b0143954655389/Makefile.Common#L1-L2

This does not explain the whole problem, since in #7442 a breaking change was made on one of the main modules but the check passes.

I am able to reproduce this locally: if I run `make apidiff-build` on main, switch to the PR's branch and  run `make apidiff-compare` I don't get any changes listed at all.

mx-psi:
cc @Aneurysm9 maybe you can help here?

PR #11797, opened 2024-12-04
Fix merge queue CI

### LINKED ISSUE #11790: Make CI workflows compatible with merge queues
After [enabling the merge queue](https://github.com/open-telemetry/community/issues/2469) on this repository, a PR [was rejected](https://github.com/open-telemetry/opentelemetry-collector/pull/11781#event-15496733797) from the merge queue because of two failing checks, `changelog` and `api-compatibility`, which were passing in the PR tests. Looking into it, it seems the issue is that these checks rely on the context variables `github.event.pull_request.head.sha`, `github.base_ref`, and `github.head_ref`, which don't exist in merge queue-triggered workflows ([docs for `github` context](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context), [docs for `github.merge_group`](https://docs.github.com/en/webhooks/webhook-events-and-payloads#merge_group)). We need to figure out if these checks can be modified to work in merge queues, or if we need to disable them in merge queues entirely.

jade-guiton-dd:
#### Description

This PR removes the `merge_group` trigger from the two CI checks that failed in merge queues (but not in PRs). See tracking issue for details.

The reasoning for simply disabling them instead of modifying them to work in merge queues is three-fold:
- For `changelog` to be meaningful, it needs to only run for PRs with no `[chore]` title tag or `Skip changelog` label, but it seems there is no simple way to access this data in merge queue workflows unfortunately. (A possible complicated way may be to parse the ref (PR branch) name and query the title and labels in a script with the `gh` command.)
- `api-compatibility` is never meaningful because it seems to have been broken for a long time (see #7443)
- Both `changelog` and `api-compatibility` are checks on the _set of changes_ made by a PR, not on the resulting state of the codebase. As such, their result should not meaningfully change when merged against an old main vs. a newer main, making running them in the merge queue pointless.

#### Link to tracking issue
Fixes #11790

This should let us try [enabling the merge queue](https://github.com/open-telemetry/community/issues/2469) again.

#### Testing
Testing merge queue checks is unfortunately quite difficult, but since we're only disabling checks I don't foresee any issues.

### DIFF
--- a/.github/workflows/api-compatibility.yml
+++ b/.github/workflows/api-compatibility.yml
@@ -9,8 +9,6 @@ on:
   pull_request:
     branches:
       - main
-  merge_group:
-    types: [checks_requested]
 
 permissions:
   contents: read
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -9,8 +9,6 @@ on:
     types: [opened, ready_for_review, synchronize, reopened, labeled, unlabeled]
     branches:
       - main
-  merge_group:
-    types: [checks_requested]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref }}


evan-bradley (approved):
Makes sense to me, thanks for the detailed reasoning.

mowies (approved):

PR #11805, opened 2024-12-04
[confighttp] Add support for configuring compression levels

rnishtala-sumo:
#### Description
Add the option of configuring compression levels. This is not intended to be a breaking change for the user.

#### Link to tracking issue
Issue - https://github.com/open-telemetry/opentelemetry-collector/issues/10467

#### Testing
Unit tests were added/changed


config/configcompression/compressiontype.go
```diff
@@ -3,30 +3,43 @@
...
+	TypeDeflate Type  = "deflate"
+	TypeSnappy  Type  = "snappy"
+	TypeZstd    Type  = "zstd"
+	TypeLz4     Type  = "lz4"
+	typeNone    Type  = "none"
+	typeEmpty   Type  = ""
+	LevelNone   Level = 0
 )
 
+var typ Type
```
dmitryax, line 33: We cannot use global variable for this. It's not guaranteed that `UnmarshalText` is called before `Validate` and is part of the same parent object

dmitryax: We need to have the `Validate` func at a parent struct. I see two options:

1. We can wrap Type and CompressionConfig in one struct that will be squashed in the `confighttp`. Then we can have `Validate` with access to both fields. It'll be a breaking change for Go API but not for users.

2. Do the validation on `confighttp.ClientConfig`. The problem with that is that we couple `confighttp` and `configcompression` even more. It may be wort just removing `configcompression` in that case...

config/configcompression/compressiontype.go
```diff
@@ -40,3 +53,28 @@ func (ct *Type) UnmarshalText(in []byte) error {
...
+		typ == typeNone ||
+		typ == typeEmpty {
+		return nil
+	}
+
+	return fmt.Errorf("unsupported compression type and level %s - %d", typ, cc.Level)
+}
+
+// Checks the validity of zlib/gzip/flate compression levels
+func isValidLevel(level int) bool {
```
dmitryax, line 73: Why don't we accept `Level` instead of casting it back and forth?
```suggestion
func isValidLevel(level Level) bool {
```

config/configcompression/compressiontype.go
rnishtala-sumo: yeah, that global variable was hacky. If we want to stay away from any breaking change, option 2 might be the way.

rnishtala-sumo: I chose option2 and moved the validation to ClientConfig. If we want to avoid API breaking changes, we could keep the `configcompression` package.

dmitryax: > If we want to avoid API breaking changes, we could keep the configcompression package.
Right. We don't need to do anything with that right now. I was thinking long term

.chloggen/compression-config-option2.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'breaking'
```
dmitryax, line 4: Not breaking anymore

.chloggen/compression-config-option2.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added support for configuring compression levels.
+
+# One or more tracking issues or pull requests related to the change
+issues: [10467]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
dmitryax, line 18: Please add more details here describing a new field

config/configcompression/compressiontype.go
```diff
@@ -3,26 +3,35 @@
 
 package configcompression // import "go.opentelemetry.io/collector/config/configcompression"
 
-import "fmt"
+import (
+	"fmt"
+)
```
dmitryax, line 8: Nit please avoid unrelated changes

config/configgrpc/go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/klauspost/compress v1.17.9 // indirect
+	github.com/klauspost/compress v1.17.11 // indirect
```
dmitryax, line 39: Why is this change? It doesn't require new version right?

config/confighttp/compression.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 type compressRoundTripper struct {
-	rt              http.RoundTripper
-	compressionType configcompression.Type
-	compressor      *compressor
+	rt                http.RoundTripper
+	compressionType   configcompression.Type
+	CompressionParams configcompression.CompressionParams
```
dmitryax, line 26: to be consistent
```suggestion
	compressionParams configcompression.CompressionParams
```

config/confighttp/compression_test.go
```diff
@@ -35,9 +36,12 @@ func TestHTTPClientCompression(t *testing.T) {
 	compressedZstdBody := compressZstd(t, testBody)
 	compressedLz4Body := compressLz4(t, testBody)
 
+	const invalidGzipLevel configcompression.Level = 100
+
 	tests := []struct {
 		name        string
 		encoding    configcompression.Type
+		enclevel    configcompression.Level
```
dmitryax, line 44: shoud be `level` or `encLevel`

config/confighttp/compression_test.go
```diff
@@ -104,16 +119,26 @@ func TestHTTPClientCompression(t *testing.T) {
...
+				message := fmt.Sprintf("unsupported compression type and level %s - %d", tt.encoding, tt.enclevel)
+				assert.Equal(t, message, err.Error())
+				return
 			}
 			client, err := clientSettings.ToClient(context.Background(), componenttest.NewNopHost(), componenttest.NewNopTelemetrySettings())
 			require.NoError(t, err)
 			res, err := client.Do(req)
 			if tt.shouldError {
-				assert.Error(t, err)
+				require.Error(t, err)
```
dmitryax, line 141: should not be changed

config/confighttp/compression_test.go
```diff
@@ -104,16 +119,26 @@ func TestHTTPClientCompression(t *testing.T) {
 
 			req, err := http.NewRequest(http.MethodGet, srv.URL, reqBody)
 			require.NoError(t, err, "failed to create request to test handler")
-
+			compressionType := tt.encoding
+			err = compressionType.UnmarshalText([]byte(tt.encoding))
```
dmitryax, line 123: Why do we need UnmarshalText?

config/confighttp/README.md
```diff
@@ -26,6 +26,31 @@ README](../configtls/README.md).
...
+    - `deflate`
+      - NoCompression: `0`
+      - BestSpeed: `1`
+      - BestCompression: `9`
+      - DefaultCompression: `-1`
+    - `zstd`
+      - SpeedFastest: `1`
+      - SpeedDefault: `3`
+      - SpeedBetterCompression: `6`
+      - SpeedBestCompression: `11`
```
dmitryax, line 51: Does it mean that `zstd` has more levels? I see we don't validate that

config/confighttp/confighttp.go
```diff
@@ -134,6 +138,33 @@ func NewDefaultClientConfig() ClientConfig {
 	}
 }
 
+// Checks the validity of zlib/gzip/flate compression levels
+func isValidLevel(level configcompression.Level) bool {
+	return level == zlib.DefaultCompression ||
+		level == configcompression.LevelNone ||
```
dmitryax, line 144: Why not `zlib.NoCompression`?

config/confighttp/confighttp.go
```diff
@@ -134,6 +138,33 @@ func NewDefaultClientConfig() ClientConfig {
...
+		level == zlib.HuffmanOnly ||
+		level == zlib.NoCompression ||
+		level == zlib.BestSpeed ||
```
dmitryax, line 147: This seems redundant as we check it in the next line

config/confighttp/README.md
rnishtala-sumo: zstd doesn't seem to need validation because it picks the nearest level to whatever number is entered.

config/configgrpc/go.mod
rnishtala-sumo: that's right, will revert

rnishtala-sumo:
@dmitryax this latest [commit](https://github.com/open-telemetry/opentelemetry-collector/pull/11805/commits/d7109c1c59b49aa29747c38b96f036c3e3b0a6b7) has all the changes from the most recent review.

.chloggen/compression-config-option2.yaml
rnishtala-sumo: Added some description, PTAL, thanks!

rnishtala-sumo:
@dmitryax please let me know if there are any other changes that you'd like to see!

config/confighttp/compressor.go
```diff
@@ -23,45 +24,72 @@ type writeCloserReset interface {
...
+	_                writeCloserReset = (*snappy.Writer)(nil)
+	_                writeCloserReset = (*zstd.Encoder)(nil)
+	_                writeCloserReset = (*zlib.Writer)(nil)
+	_                writeCloserReset = (*lz4.Writer)(nil)
+)
+
 // writerFactory defines writer field in CompressRoundTripper.
 // The validity of input is already checked when NewCompressRoundTripper was called in confighttp,
-func newCompressor(compressionType configcompression.Type) (*compressor, error) {
+func newCompressor(compressionType configcompression.Type, compressionParams configcompression.CompressionParams) (*compressor, error) {
```
axw, line 48: This function is no longer concurrent-safe, and needs to be. Accessing/updating `compressorPools` should be guarded by a mutex.

config/confighttp/compressor.go
```diff
@@ -23,45 +24,72 @@ type writeCloserReset interface {
...
-	lz4Pool                   = &compressor{pool: sync.Pool{New: func() any {
-		lz := lz4.NewWriter(nil)
-		// Setting concurrency to 1 to disable async decoding by goroutines. This will reduce the overall memory footprint and pool
-		_ = lz.Apply(lz4.ConcurrencyOption(1))
-		return lz
-	}}}
-)
+type compressorMap struct {
+	pools map[string]*compressor
+}
```
axw, line 29: I think this type may as well be a map, i.e. `type compressorMap map[string]*compressor`. (See my other comment about using a non-string key, though.)

config/confighttp/compressor.go
```diff
@@ -23,45 +24,72 @@ type writeCloserReset interface {
...
+	_                writeCloserReset = (*zstd.Encoder)(nil)
+	_                writeCloserReset = (*zlib.Writer)(nil)
+	_                writeCloserReset = (*lz4.Writer)(nil)
+)
+
 // writerFactory defines writer field in CompressRoundTripper.
 // The validity of input is already checked when NewCompressRoundTripper was called in confighttp,
-func newCompressor(compressionType configcompression.Type) (*compressor, error) {
+func newCompressor(compressionType configcompression.Type, compressionParams configcompression.CompressionParams) (*compressor, error) {
+	mapKey := fmt.Sprintf("%s/%d", compressionType, compressionParams.Level)
```
axw, line 49: Using only the Level field will work for now, but if another field is added then it may lead to confusion if the parameter is not considered in the pool key. I suggest changing the `compressorMap` type like so:

```go
type compressorMap map[compressionMapKey]*compressor
 
type compressionMapKey struct {
        compressionType   configcompression.Type
        compressionParams configcompression.CompressionParams
}
```

config/confighttp/compressor.go
```diff
@@ -23,45 +24,72 @@ type writeCloserReset interface {
...
 	case configcompression.TypeGzip:
-		return gZipPool, nil
+		gZipCompressor, gzipExists := compressorPools.pools[mapKey]
+		if gzipExists {
+			return gZipCompressor, nil
+		}
+		gZipCompressor = &compressor{}
+		gZipCompressor.pool = sync.Pool{New: func() any { w, _ := gzip.NewWriterLevel(nil, int(compressionParams.Level)); return w }}
+		compressorPools.pools[mapKey] = gZipCompressor
+		return gZipCompressor, nil
```
axw, line 59: The map lookup/update is repeated a lot. I suggest splitting this into two functions:

```go
func newCompressor(compressionType configcompression.Type, compressionParams configcompression.CompressionParams) (*compressor, error) {
        compressorPoolsMu.Lock()
        defer compressorPoolsMu.Unlock()

        mapKey := compressionMapKey{compressionType, compressionParams}
        c, ok := compressorPools[mapKey]
        if ok {
                return c, nil
        }

        f, err := newWriteCloserResetFunc(compressionType, compressionParams)
        if err != nil {
                return nil, err
        }
        c = &compressor{pool: sync.Pool{New: func() any { return f() }}}
        compressorPools[mapKey] = c
        return c, nil
}

func newWriteCloserResetFunc(compressionType configcompression.Type, compressionParams configcompression.CompressionParams) (func() writeCloserReset, error) {
        switch compressionType {
        case configcompression.TypeGzip:
                return func() writeCloserReset {
                        w, _ := gzip.NewWriterLevel(nil, int(compressionParams.Level))
                        return w
                }, nil
        case configcompression.TypeSnappy:
                return func() writeCloserReset {
                        return snappy.NewBufferedWriter(nil)
                }, nil
        case configcompression.TypeZstd:
                level := zstd.WithEncoderLevel(zstd.EncoderLevelFromZstd(int(compressionParams.Level)))
                return func() writeCloserReset {
                        zw, _ := zstd.NewWriter(nil, zstd.WithEncoderConcurrency(1), level)
                        return zw
                }, nil
        case configcompression.TypeZlib, configcompression.TypeDeflate:
                return func() writeCloserReset {
                        w, _ := zlib.NewWriterLevel(nil, int(compressionParams.Level))
                        return w
                }, nil
        case configcompression.TypeLz4:
                return func() writeCloserReset {
                        lz := lz4.NewWriter(nil)
                        _ = lz.Apply(lz4.ConcurrencyOption(1))
                        return lz
                }, nil
        }
        return nil, errors.New("unsupported compression type")
}
```

config/confighttp/compressor.go
```diff
@@ -23,45 +24,72 @@ type writeCloserReset interface {
...
 
+var (
+	compressorPools                   = &compressorMap{pools: make(map[string]*compressor)}
+	snappyCompressor                  = &compressor{}
+	lz4Compressor                     = &compressor{}
+	_                writeCloserReset = (*gzip.Writer)(nil)
+	_                writeCloserReset = (*snappy.Writer)(nil)
+	_                writeCloserReset = (*zstd.Encoder)(nil)
+	_                writeCloserReset = (*zlib.Writer)(nil)
+	_                writeCloserReset = (*lz4.Writer)(nil)
```
axw, line 43:
```suggestion
```

These are unnecessary if you apply my other suggestions: snappy & lz4 would also go in the map, and the separate newWriteCloserResetFunc function would guarantee that the encoders/writers implement the required interface.

rnishtala-sumo:
@axw I appreciate the review. If possible, I'd also like @dmitryax or @open-telemetry/collector-approvers to chime in before I  implement these changes.

config/confighttp/confighttp.go
```diff
@@ -137,6 +141,31 @@ func NewDefaultClientConfig() ClientConfig {
...
+func (hcs *ClientConfig) Validate() error {
+	if hcs.Compression.IsCompressed() {
+		if (hcs.Compression == configcompression.TypeGzip && isValidLevel(hcs.CompressionParams.Level)) ||
+			(hcs.Compression == configcompression.TypeZlib && isValidLevel(hcs.CompressionParams.Level)) ||
+			(hcs.Compression == configcompression.TypeDeflate && isValidLevel(hcs.CompressionParams.Level)) ||
+			hcs.Compression == configcompression.TypeSnappy ||
+			hcs.Compression == configcompression.TypeLz4 ||
+			hcs.Compression == configcompression.TypeZstd ||
+			hcs.Compression == configcompression.TypeNone ||
+			hcs.Compression == configcompression.TypeEmpty {
```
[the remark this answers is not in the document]
Aneurysm9: This also seems to be the only place those symbols are referenced outside of the `configcompression` package.  If they're removed here they no longer need to be exported.  More generally, should `configcompression` gain a `Validate(configcompression.CompressionParams)` method so that this whole validation routine can be encapsulated where knowledge of validity lies?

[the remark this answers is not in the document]
rnishtala-sumo: The reason this is here, is because both `CompressionType` and `CompressionParams` are part of the client config in `confighttp`.  The Validation needs to be done on a combination of compression type(`Compression`) and level (part of `CompressionParams`). This cannot be done in the `configcompression` package.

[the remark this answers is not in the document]
rnishtala-sumo: > Is this correct? Should we be checking that level is unspecified for lz4 and snappy?
The level if any is specified is ignored for lz4 and snappy

config/confighttp/compressor.go
rnishtala-sumo: Implemented this change! PTAL

config/confighttp/compressor.go
rnishtala-sumo: Yes makes sense!

config/confighttp/confighttp.go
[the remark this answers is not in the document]
axw: > The level if any is specified is ignored for lz4 and snappy

The collector will normally error out if you specify undefined config parameters, rather than ignore them. e.g. if you specified the following:

```yaml
compression_params:
  foo: bar
```

then that will result in the error `'compression_params' has invalid keys: foo`

Why should that fail, but not when you specify level for compression algorithms that don't support it?

config/confighttp/compression_test.go
```diff
@@ -104,10 +126,17 @@ func TestHTTPClientCompression(t *testing.T) {
...
+				Compression:       tt.encoding,
+				CompressionParams: newCompressionParams(tt.level),
+			}
+			err = clientSettings.Validate()
+			if tt.shouldError {
+				require.Error(t, err)
+				message := fmt.Sprintf("unsupported compression type and level %s - %d", tt.encoding, tt.level)
+				assert.Equal(t, message, err.Error())
+				return
 			}
```
axw, line 140: I can't seem to unresolve comments (https://github.com/open-telemetry/opentelemetry-collector/pull/11805/files#r1904797587), so creating a new one: should there be a `require.NoError(t, err)` in the `tt.shouldError == false` branch?

config/confighttp/confighttp.go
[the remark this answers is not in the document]
axw: confighttp would still need to invoke the method, but you can move the majority of the logic to configcompression: https://github.com/axw/opentelemetry-collector/commit/3defa89f23d09ad5369082a714cf265f22670209

axw (commented):
Thanks for the updates, mostly looks good - just a few minor issues remaining.

config/confighttp/confighttp.go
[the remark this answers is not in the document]
rnishtala-sumo: cool makes sense, I added tests for it in the latest commit.

config/confighttp/compression_test.go
rnishtala-sumo: yes added that after the `if` condition.

config/configcompression/compressiontype.go
```diff
@@ -21,22 +30,37 @@ const (
...
+			p.Level == zlib.NoCompression ||
+			(p.Level >= zlib.BestSpeed && p.Level <= zlib.BestCompression) {
+			return nil
+		}
+	case TypeZstd:
+		// Supports arbitrary levels: zstd will map any given
+		// level to the nearest internally supported level.
+		return nil
+	case TypeSnappy, TypeLz4, typeNone, typeEmpty:
+		if p.Level != 0 && (t == TypeSnappy || t == TypeLz4) {
```
axw, line 60:
```suggestion
		if p.Level != 0 {
```

Level isn't relevant for any of these cases

config/configcompression/compressiontype_test.go
axw, line 1: We should probably add a test for Type.ValidateParams in this file

axw (approved):
Thanks for iterating, LGTM

config/configcompression/compressiontype_test.go
rnishtala-sumo: Added some tests for this, thanks!

config/configcompression/compressiontype.go
```diff
@@ -3,40 +3,65 @@
...
-	if typ == TypeGzip ||
-		typ == TypeZlib ||
-		typ == TypeDeflate ||
-		typ == TypeSnappy ||
-		typ == TypeZstd ||
-		typ == TypeLz4 ||
-		typ == typeNone ||
-		typ == typeEmpty {
-		*ct = typ
+func (t Type) Validate() error {
```
dmitryax, line 38: Why `UnmarshalText` replaced with `Validate`? This is a breaking change. I don't think it's necessary.

config/configcompression/compressiontype.go
```diff
@@ -40,3 +50,25 @@ func (ct *Type) UnmarshalText(in []byte) error {
...
+		// Supports arbitrary levels: zstd will map any given
+		// level to the nearest internally supported level.
+		return nil
+	case TypeSnappy, TypeLz4, typeNone, typeEmpty:
+		if p.Level != 0 {
+			return fmt.Errorf("unsupported parameters %+v for compression type %q", p, t)
+		}
+		return nil
+	}
+	return fmt.Errorf("unsupported parameters %+v for compression type %q", p, t)
```
dmitryax: Do we need to list all the types explicitly here? It'll be hard to keep it in sync when new types are added. What about this instead?
```suggestion
	if p.Level != 0 {
		return fmt.Errorf("unsupported parameters %+v for compression type %q", p, t)
	}
	return nil
```

rnishtala-sumo: Maybe this is fine? We'll need to revisit this function when new types are added anyway...

rnishtala-sumo: on second thought, it seems unnecessary, made the update, thanks

dmitryax (approved):

jmacd:
Recommend changing the PR title to emphasize that this is being addressed for HTTP configurations, not gRPC.

FYI the OTel-Arrow project also addressed this problem for itself internally on the gRPC path, and I think it would be worth planning for HTTP and gRPC configurations to do the same. 
 
Please see https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/otelarrow/compression, which has support for both exporter- and receiver- side configuration of the Zstd compressor.

PR #11816, opened 2024-12-06
[chore] move explicitly to use ubuntu 24.04 runners

atoulme:
#### Description
Rather than use ubuntu-latest, which will move over a period of time to ubuntu 24.04, directly move to ubuntu 24.04 to avoid having flaky or unexpected failures.

See https://github.com/actions/runner-images/issues/10636 for context.

### DIFF
--- a/.github/workflows/build-and-test-arm.yml
+++ b/.github/workflows/build-and-test-arm.yml
@@ -53,7 +53,7 @@ jobs:
         run: make -j4 gotest
   arm-unittest:
     if: ${{ github.actor != 'dependabot[bot]' && (contains(github.event.pull_request.labels.*.name, 'Run ARM') || github.event_name == 'push' || github.event_name == 'merge_group') }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     needs: [arm-unittest-matrix]
     steps:
       - name: Print result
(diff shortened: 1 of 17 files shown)


codeboten (commented):
If tests are passing with ubuntu-latest and ubuntu-24.04, is there a reason for changing this? i'd expect this to be more needed if we saw failures that we would need to address, but maybe i'm missing something

atoulme:
No, you understood right. We can let go and keep using ubuntu-latest as this proves we should be ok either way. I was just making sure we would be ok.

dmathieu:
If failures were expected (which doesn't seem to be the case), should we run on both 22.04 and 24.04 (and possibly even 20.04 until it's EOL next april)?
Folks will probably keep running their collector on those versions for some time.

jpkrohling:
> should we run on both 22.04 and 24.04

I don't think it brings much value, and would increase our CI time significantly.

andrzej-stencel (approved):

andrzej-stencel:
>> If tests are passing with ubuntu-latest and ubuntu-24.04, is there a reason for changing this? i'd expect this to be more needed if we saw failures that we would need to address, but maybe i'm missing something
> No, you understood right. We can let go and keep using ubuntu-latest as this proves we should be ok either way. I was just making sure we would be ok.

I like the Ubuntu version being stated explicitly, but It's probably not a big deal, can have it either way.

Issue #11818, opened 2024-12-07
Unnecessary high memory in sharedcomponent


bogdandrutu:
The https://github.com/open-telemetry/opentelemetry-collector/pull/10777 added the "capability" to keep all previously reported events in shared components and the list grows unbounded causing memory issues.

Proposed solution: only preserve the last reported event.

TylerHelmuth:
For an status aggregator I believe it is important that the previous events be played back in order. Since `SharedComponent` is internal and we control how it is used, and since we only ever add sources during startup, I suggest a circular buffer with a small size.  This allows us to replay all the events during startup, but then during normal operation, once we are done adding sources, we won't have to remember every event we've ever sent.

PR #11826, opened 2024-12-09
[interna/sharedcomponent] Use a ring buffer to restrict remembered status count

TylerHelmuth:
#### Description
Use a ring buffer to only remember the last 5 events. This ensures we remember a reasonable number of events during startup, so that a status aggregator gets the events for all instances. Then during normal operation, when we're done adding sources and no longer need to replay events, we don't have to remember every single event.

#### Link to tracking issue
Closes https://github.com/open-telemetry/opentelemetry-collector/issues/11818

#### Testing

`go test status_test.go -count 1000 -failfast` still passes with many tries.


internal/sharedcomponent/sharedcomponent.go
```diff
@@ -108,19 +108,30 @@ var _ componentstatus.Reporter = (*hostWrapper)(nil)
...
 }
 
 func (h *hostWrapper) GetExtensions() map[component.ID]component.Component {
 	return h.host.GetExtensions()
 }
 
+func (h *hostWrapper) hasEvent(e *componentstatus.Event) bool {
+	hasEvent := false
+	h.previousEvents.Do(func(a any) {
+		if previousEvent, ok := a.(*componentstatus.Event); ok && previousEvent == e {
```
bogdandrutu, line 122: How can this happen?

TylerHelmuth: Added a test.  It can happen if a component were to ever create an event and then emit it twice.  Components should really never do that, but this defensive code makes sure that if they do we don't fill up our buffer with duplicate events.

bogdandrutu: Another thing is that events have a "timestamp" right now witch makes them to be unique every time you create a new event, but there is a discussion to remove that, which will cause lots of "duplicate" events to be detected and that is wrong most likely.

So overall I think this is a "premature defensive code" (optimization) that we should remove.

TylerHelmuth: ok

bogdandrutu (approved):

mx-psi:
@TylerHelmuth Ready to merge once you fix the merge conflicts

codeboten (approved):

PR #11829, opened 2024-12-09
add changelog entry for Dockerfile USER_GID as non root

rogercoll:
#### Description

This PR just adds a new changelog entry for the changes we are doing in the releases repository: https://github.com/open-telemetry/opentelemetry-collector-releases/pull/738

The main reason is that the `releases` repository does not have any changelog mechanism, but the corresponding change is important enough to make the users aware of. 

- Is this the appropriate approach? Have there been any similar situations in the past?

**My understanding is that this PR should be merged after https://github.com/open-telemetry/opentelemetry-collector-releases/pull/738 but before its next release**

#### Link to tracking issue
Fixes 
https://github.com/open-telemetry/opentelemetry-collector-releases/issues/662

### DIFF
--- a/.chloggen/dockerfile_remove_root_group.yaml
+++ b/.chloggen/dockerfile_remove_root_group.yaml
@@ -7,10 +7,10 @@ change_type: 'enhancement'
 component: container
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Set Dockerfile's USER_GID as non root
+note: "Set Dockerfile's USER_GID as non root. OpenTelemetry Collector releases issue: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/662"
 
 # One or more tracking issues or pull requests related to the change
-issues: [https://github.com/open-telemetry/opentelemetry-collector-releases/issues/662]
+issues: []
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/dockerfile_remove_root_group.yaml
+++ b/.chloggen/dockerfile_remove_root_group.yaml
@@ -10,7 +10,7 @@ component: container
 note: "Set Dockerfile's USER_GID as non root. OpenTelemetry Collector releases issue: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/662"
 
 # One or more tracking issues or pull requests related to the change
-issues: []
+issues: [11829]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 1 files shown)


.chloggen/dockerfile_remove_root_group.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: container
```
songy23, line 7: There is no `container` component in this repo, maybe use `release` or `image` instead

rogercoll: Thanks! Replaced in https://github.com/open-telemetry/opentelemetry-collector/pull/11829/commits/9593eba27c71e3f134da0fd2f615633b9f6fbe02

songy23 (approved):
LGTM

.chloggen/dockerfile_remove_root_group.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: image
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Set Dockerfile's USER_GID as non root. OpenTelemetry Collector releases issue: https://github.com/open-telemetry/opentelemetry-collector-releases/issues/662"
```
mowies, line 10: [nitpick] can you not add the link to the `issues:` field below?

mowies (approved):

.chloggen/dockerfile_remove_root_group.yaml
songy23: no, you can only link issues / PRs in this repo in `issues: `

mowies: ok 🤷🏼

atoulme:
You can add to the changelog of the release on opentelemetry-collector-releases, I don't think the opentelemetry-collector is the right place for this change entry.

andrzej-stencel (approved):

mx-psi:
I agree with Antoine, I think this change should be noted in the opentelemetry-collector-releases repository instead of here

songy23:
agree if we can changelog to the -releases repo, that would be more ideal

atoulme:
Changelog is in -releases now!

rogercoll:
Sounds good to me as well. The primary reason for adding the changelog entry in the collector's repository was to ensure visibility, particularly for environments that might be impacted.

PR #11836, opened 2024-12-10
[docs] Add guidelines for experimental modules naming

dmitryax:
Adding guidelines for experimental module naming as a result of [the vote](https://github.com/open-telemetry/opentelemetry-collector/issues/11783).

Resolves #11783

PR #11860, opened 2024-12-12
[chore] Document module split in more detail

### LINKED ISSUE #11436: Document best practice regarding experimental modules
Some modules are experimental and will continue to be experimental even after we declare the parent module 1.0:
- [`pdata/pprofile`](https://pkg.go.dev/go.opentelemetry.io/collector/pdata/pprofile)
- [`component/componentstatus`](https://pkg.go.dev/go.opentelemetry.io/collector/component/componentstatus)
- (possibly) `consumer/consumererror`

We may want to merge these modules back into the parent module when the versions match, so that these modules become packages instead of separate modules. We should document what the best practice regarding this is.

### LINKED ISSUE #11623: Document module split in coding guidelines

mx-psi:
#### Description

- Documents when to split into separate modules, including general rules as well as specific conventions we are currently using
- Rephrases the wording on #11836 to add it into a general list.
- Documents how to split into separate modules.

#### Link to tracking issue
Follows #11836, Fixes #11436, Fixes #11623


docs/coding-guidelines.md
```diff
@@ -47,11 +47,41 @@ When naming configuration structs, use the following guidelines:
...
+   `config/configgrpc` and `config/confighttp` are separate modules.
+1. Testing helpers should be in a separate submodule with the suffix `test`. For example, if you
+   have a module `component`, the helpers should be in `component/componenttest`.
+1. Experimental packages or groups that will later be merged into the another module should be in
+   their own module, named as it will be after integration. For example, if adding profile support
+   to `pdata`, you should add a separate module `pdata/pprofile` for the experimental code.
+1. Experimental code that will be added to an existing stable module can be a submodule with the
+   same name, but prefixed with x. They MUST have the same name as the stable modules prefixed with
+   `x`. For example, `config/confighttp` module can have an experimental module named
+   `config/confighttp/xconfighttp` that contains experimental APIs.
```
jade-guiton-dd, line 72: I find the distinction between points 5 and 6 confusing; if I add experimental profiling code to `pdata`, should I name my experimental module `pdata/pprofile`, or `pdata/xpdata`?

(Also, I assume "prefixed with x" being repeated twice is a mistake)

docs/coding-guidelines.md
```diff
@@ -47,11 +47,41 @@ When naming configuration structs, use the following guidelines:
...
+1. Each top-level directory should be a separate module.
+1. Each component referenceable by the OpenTelemetry Collector Builder should be in a separate
+   module. For example, the OTLP receiver is in its own module, different from that of other
+   receivers.
+1. Consider splitting into separate modules if the API may evolve independently in separate groups
+   of packages. For example, the configuration related to HTTP and gRPC evolve independently, so
+   `config/configgrpc` and `config/confighttp` are separate modules.
+1. Testing helpers should be in a separate submodule with the suffix `test`. For example, if you
+   have a module `component`, the helpers should be in `component/componenttest`.
+1. Experimental packages or groups that will later be merged into the another module should be in
```
jade-guiton-dd, line 66: Small typo

```suggestion
1. Experimental packages or groups that will later be merged into another module should be in
```

docs/coding-guidelines.md
```diff
@@ -47,11 +47,41 @@ When naming configuration structs, use the following guidelines:
...
+1. Experimental code that will be added to an existing stable module can be a submodule with the
+   same name, but prefixed with x. They MUST have the same name as the stable modules prefixed with
+   `x`. For example, `config/confighttp` module can have an experimental module named
+   `config/confighttp/xconfighttp` that contains experimental APIs.
+
+When adding a new module remember to update the following:
+1. Add a changelog note for the new module.
+1. Add the module in `versions.yaml`.
+1. Use `make crosslink` to make sure the module replaces are added correctly throughout the
+   codebase.
```
jade-guiton-dd, line 78: Maybe add a note that `crosslink` may fail if there was a package with the same path as the new module. I had to semi-manually add replaces in all the module split PRs I did because of that.

jade-guiton-dd (approved):

docs/coding-guidelines.md
mx-psi: > I find the distinction between points 5 and 6 confusing; if I add experimental profiling code to pdata, should I name my experimental module pdata/pprofile, or pdata/xpdata?

If you are adding a new package/package group you don't add the `x`. If you are adding functionality that you want to eventually add to an existing package, you add the `x`. So for example:
- Profiles are in a separate package, so they don't get the `x`
- Something like #11769 is something we don't want to have in a separate package: the only reason we put it in a separate package is because we can't have experimental and stable API on the same package.

Does that make more sense? If so, how could we rephrase this to make the distinction more clear?

jade-guiton-dd: That makes sense. I think one way to make this clearer would be:

> 1. Experimental packages or groups ***of packages*** that will later be ***added to a*** module should be in their own module, named as ***they*** will be after integration. For example, if adding ***a `pprofile` package*** to `pdata`, you should add a separate module `pdata/pprofile` for the experimental code.
> 1. Experimental code that will be added to an existing ***package in a*** stable module can be a submodule with the same name, but prefixed with an `x`. For example, `config/confighttp` module can have an experimental module named `config/confighttp/xconfighttp` that contains experimental APIs.

mx-psi: Addressed, thanks!

codeboten (approved):

Issue #11553, opened 2024-10-28
Document requirements for components moving stability levels and reaching 1.0


mx-psi:
We have a [list of component stability levels](https://github.com/open-telemetry/opentelemetry-collector?tab=readme-ov-file#stability-levels), as well as guidelines for [configuration changes](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md#configuration-changes). We also have the concept of a component being 1.0. 

Based on some recent discussions on the Collector SIG (2024-10-23) and the Collector stability working group (2024-10-28), it would beneficial to have a consolidated document with a list of requirements for a component to move through the stability stages and for reaching 1.0

- [ ] Move existing stability level definitions to a document in the `docs/` folder
- [ ] Move configuration changes guidelines to the new document
- [ ] Establish codeowners minimum criteria for moving up through the stability ladder
- [ ] Document relation between signal stability levels and component versioning
- [ ] Document observability requirements for 1.0 components

mx-psi:
We may also want to have requirements about keeping the size of internal state in check for stateful components

YohanSciubukgian:
Could you also consider major releases with LTS? (E.g. 3 years)

mx-psi:
I am going to keep this one open but remove this from Collector v1 since we have almost completed it and the last item refers to semconv which doesn't really affect the OTLP receiver/exporters.

PR #11864, opened 2024-12-12
[docs/component-stability.md] Add criteria for graduating between stability levels

### LINKED ISSUE #11850: Establish codeowners minimum criteria for moving up through the stability ladder

mx-psi:
#### Description

Code ownership and maintenance of components continues to be an issue, with varying levels of support across contrib. As we approach 1.0 and the ability to mark components as stable, we want to make sure that components that we deem as 'stable' have a healthy community around them. We have three datapoints that we can leverage here: how many codeowners a component has, how diverse these are in terms of employers and how actively the codeowners have been responding to issues/PRs in the recent past.

We need criteria that
1. Are reasonable predictors of the component health over the short/medium term
2. Are not too onerous on the code owners

Some notes:
1. Some beta components do not meet the criteria listed on the PR. This will be the case even after the transition for some components. This PR makes no claim as to what should happen to these components stability (so, de facto, they will stay as is). 
2. The OTLP receiver and exporters do not meet this criteria today because they don't have listed code owners. We can solve this either by carving out an exception or by listing code owners.
3. We need automation and templates to enforce this.

#### Link to tracking issue
Fixes #11850


docs/component-stability.md
```diff
@@ -78,3 +78,52 @@ Components that were accepted based on being vendor-specific components will be
...
+
+## Alpha to beta
+
+To graduate any signal from alpha to beta on a component:
+1. The component MUST have at least two active code owners.
+2. The code owners for non-vendor-specific components SHOULD have at least two different employers.
+3. Within the 30 days prior to the graduation request, the code owners MUST have reviewed and
+   replied to at least 80% of the issues and pull requests opened against the component. This
+   excludes general PRs or issues that are not specific to the component itself (e.g. repo-wide API
+   updates). It is not necessary that the issues and PRs are closed or merged, but the thay have
```
ChrsMark, line 117: Looks like a typo?:
```suggestion
   updates). It is not necessary that the issues and PRs are closed or merged, but that they have
```

docs/component-stability.md
```diff
@@ -78,3 +78,52 @@ Components that were accepted based on being vendor-specific components will be
...
+
+## Beta to stable
+
+To graduate any signal from beta to stable on a component:
+1. The component MUST have at least three active code owners.
+2. The code owners for non-vendor-specific components SHOULD have at least two different employers.
+3. Within the 60 days prior to the graduation request, the code owners MUST have reviewed and
+   replied to at least 80% of the issues and pull requests opened against the component. This
+   excludes general PRs or issues that are not specific to the component itself (e.g. repo-wide API
+   updates). It is not necessary that the issues and PRs are closed or merged, but the thay have
```
ChrsMark, line 128:
```suggestion
   updates). It is not necessary that the issues and PRs are closed or merged, but that they have
```

docs/component-stability.md
```diff
@@ -78,3 +78,52 @@ Components that were accepted based on being vendor-specific components will be
...
+usage and scrutiny that comes with a higher stability level, and that the community around it is
+sufficiently healthy.
+
+If the graduation criteria are not met, the approver should provide feedback on what is missing and
+how to address it. The component owners can then address the feedback and re-request graduation on
+the same issue.
+
+## In development to alpha
+
+No additional criteria are required to graduate from development to alpha.
```
atoulme, line 107: Maybe we can request that the codeowners understand that moving to alpha sends the signal that the component is ready for limited use, and therefore must have enough documentation, and should be expected to have a consistent design that will not change completely moving forward.

docs/component-stability.md
```diff
@@ -78,3 +78,52 @@ Components that were accepted based on being vendor-specific components will be
...
+## Alpha to beta
+
+To graduate any signal from alpha to beta on a component:
+1. The component MUST have at least two active code owners.
+2. The code owners for non-vendor-specific components SHOULD have at least two different employers.
+3. Within the 30 days prior to the graduation request, the code owners MUST have reviewed and
+   replied to at least 80% of the issues and pull requests opened against the component. This
+   excludes general PRs or issues that are not specific to the component itself (e.g. repo-wide API
+   updates). It is not necessary that the issues and PRs are closed or merged, but that they have
+   been reviewed and replied to appropriately.
```
atoulme, line 118: this is formalizing the existing practice and I agree with it.

docs/component-stability.md
```diff
@@ -78,3 +78,52 @@ Components that were accepted based on being vendor-specific components will be
...
+
+## Beta to stable
```
ChrsMark, line 120: I know this PR only focuses on the code-owners part, however I wonder what would be the plan/requirement for making components SemConv compliant.

For example `hostmetrics` receiver will go through this process at some point: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35325

I foresee K8s components will follow this path.

To my mind, for a component to graduate as stable it should only include attributes that are defined in SemConv (preferably in `ga` stability ???).

What would be the best place to consider and include this as stability requirement?

docs/component-stability.md
```diff
@@ -78,3 +78,52 @@ Components that were accepted based on being vendor-specific components will be
...
+
+To graduate any signal from beta to stable on a component:
+1. The component MUST have at least three active code owners.
```
atoulme, line 123: I think there should be a commitment from codeowners that there is a SLA for first response on bug issues.
The commitment should be measured in days.

docs/component-stability.md
mx-psi: I think everything you mention is covered above or will be covered soon (see #11871 for the documentation requirements, feel free to review that one as well :wink: ). I added an extra sentence to make it super clear that you need to fulfill the other criteria

docs/component-stability.md
mx-psi: > I know this PR only focuses on the code-owners part, however I wonder what would be the plan/requirement for making components SemConv compliant.

I agree that this is important to document, could you create an issue for this? I can add it to #11553. 

> To my mind, for a component to graduate as stable it should only include attributes that are defined in SemConv (preferably in ga stability ???).

I think it is a bit more subtle than this, since there are some things that will not be in Semantic Conventions. The Semantic Conventions SIG is working on a document to define what conventions should be in their repo vs outside of it, once we have this we can say "stable components need to follow the semantic conventions in [insert link here]".

docs/component-stability.md
mx-psi: This raises some questions for me including:
- What happens when people go on vacation/have a kid/[insert activity here that leads to a prolonged period of absence]?
- What happens if people don't follow this SLA? Typically an SLA means that you pay if you don't meet a certain standard, how do you "pay" here?

docs/component-stability.md
ChrsMark: Issue created https://github.com/open-telemetry/opentelemetry-collector/issues/11878.

julianocosta89 (commented):
I think we shouldn't make it too hard to have community components.
I think vendor components and widely used components will not have any issue to follow the guidelines.

When we think about components that add value to the overall project, but may not be interesting/priority to vendors, they may struggle to get folks involved on it, and that would disqualify them from being moved to stable.

I do understand that we need to provide a way to ensure maintainability of stable components, but maybe we could draw something over the ideas of:
- Being active
- Replying to issues related to its components timely
- Fixing reported bug timely
- ...

If we have a component that is not vendor related, but maintained by 2 folks from a single employer, that wouldn't allow them to move on.

Also, let's imagine the following scenario:
- 3 folks are codeowners of a component, 2 from one company and another one from another company. 
- They graduate to stable.
- A couple of months later the person that was from the other company moves to the same company of the other 2. Would the component be demoted?

I don't think it should be, if they are active and responsive in issues related to that component.
I know it is a corner case, and may never happen, but it still can.

My main point here, is actually that we shouldn't make it too hard to have community components.
I know a couple of companies that develop internal components to solve their customers' issues. It would be awesome to have a couple of those contributed back to upstream, and let the community grow together.

mx-psi:
@julianocosta89 

> When we think about components that add value to the overall project, but may not be interesting/priority to vendors, they may struggle to get folks involved on it, and that would disqualify them from being moved to stable.

There is a trade-off between having more components and having fewer components that are more actively maintained. We need to be mindful of where we draw the line, but my feeling is that right now we have too many components that are not well maintained.

> I do understand that we need to provide a way to ensure maintainability of stable components, but maybe we could draw something over the ideas of:
> 
> * Being active
> * Replying to issues related to its components timely
> * Fixing reported bug timely
> * ...

There's two questions to consider here: 
1. When do we make a decision to move a component to stable?
2. When do we make a decision to move a component to unmaintained?

On this PR I am focusing on (1). For doing (1), we need to focus on things that we can measure/check at the time of marking as stable. Some of the things you mention are, I feel like, important criteria for deciding if a component should be moved to unmaintained, but not to move to stable.

> If we have a component that is not vendor related, but maintained by 2 folks from a single employer, that wouldn't allow them to move on.

The point with the 'vendor diversity' is that I think it is a good predictor of component quality (more than one vendor means more focus on a wide number of use cases) and maintainability (we don't depend on a single company). Maybe it would help to do an analysis of existing components to see how difficult this is to achieve?

> Also, let's imagine the following scenario:
> 
> * 3 folks are codeowners of a component, 2 from one company and another one from another company.
> * They graduate to stable.
> * A couple of months later the person that was from the other company moves to the same company of the other 2. Would the component be demoted?

This PR makes no claims about when a component should be 'demoted'. Currently the only way to be demoted is to be moved to unmaintained, we have [some rules](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#unmaintained) about when that can happen. I personally don't think we should move from  stable->beta, I think that would be confusing for end users.

The way I see this it is a bit like the CNCF project status: there is no moving from graduated to incubating, only from graduated to deprecated.

mx-psi:
I split off part of this PR in #11937. PTAL at that one first

julianocosta89:
> There is a trade-off between having more components and having fewer components that are more actively maintained. We need to be mindful of where we draw the line, but my feeling is that right now we have too many components that are not well maintained.

Agree.

> For doing (1), we need to focus on things that we can measure/check at the time of marking as stable. Some of the things you mention are, I feel like, important criteria for deciding if a component should be moved to unmaintained, but not to move to stable.

Makes sense.

> Maybe it would help to do an analysis of existing components to see how difficult this is to achieve?

Let's take the Connector first, from all components none of them have 3 ACTIVE codeowners. `sumconnector` is the only component with 3 codeowners, but the 3 of them do not seem much active

- Connectors:
  - `countconnector`: 2 codeowners from different vendors
  - `exceptionsconnector`: 1 codeowner
  - `failoverconnector`: 2 codeowners (from different vendors?)
  - `otlpjsonconnector`: 2 codeowners from different vendors
  - `roundrobinconnector`: 1 codeowner
  - `routingconnector`: 2 codeowners from different vendors
  - `servicegraphconnector`: 2 codeowners from different vendors
  - `signaltometricsconnector`: 2 codeowners (from different vendors?)
  - `sumconnector`: 3 codeowners (from different vendors?)

On the other hand, I think this "rule" would bring more awareness to the components, and maybe it would also bring more codeowners to "important" components.

Still not sure about the the amount of codeowners.
Most of the connectors (and had a quick look at processors) have just 1 or 2.
I agree 1 is not enough, but wouldn't 2 (from different companies) be enough?

julianocosta89:
> Still not sure about the the amount of codeowners. Most of the connectors (and had a quick look at processors) have just 1 or 2. I agree 1 is not enough, but wouldn't 2 (from different companies) be enough?

After thinking further and discussing with @mx-psi, I believe this criteria is going to be the foundation for users to further engage and assume codeowner's responsibility in the components they would like to move to stable.
I'd say that at least 3 codeowners is a good number too keep the collector components maintainable are not just pilling up responsibilities to Collector maintainers.

atoulme (approved):
I agree with the addition of those criteria.

dmitryax (approved):
LGTM

ArthurSens (commented):
The criteria look reasonable; my only concern is whether our current automation can support approvers when reviewing Graduation requests.

For the most used components, 60 days is enough time for dozens of PRs/Issues to be opened, and code owners may have changed during this period. Asking approvers to look at all of them and consider code owner changes might be too difficult if done manually. If a process is too difficult to follow, people tend to bypass it 😅

mx-psi:
@ArthurSens I think those concerns are reasonable, we definitely need more automation to be able to do this at scale and over time. I think the focus at this stage should be whether these requirements are automatable, we will automate things once we have the rules defined

douglascamata:
Are there any plans regarding "technical stability/maturity"? For instance, a threshold of test coverage and/or linting issues. 

I'm thinking here from a "user" point of view. What does stability mean to users? Number and activity of maintainers is a point, but I also believe there are some technical aspects to it that could be considered (or not).

docs/component-stability.md
```diff
@@ -260,6 +260,44 @@ To move within the 'Maintained' ladder ("graduate"), the process for doing so is
...
+## In development to alpha
+
+No additional criteria are required to graduate from development to alpha.
+The component still needs to meet the general requirements for alpha components.
+
+## Alpha to beta
+
+To graduate any signal from alpha to beta on a component:
+1. The component MUST have at least two active code owners.
+2. The code owners for non-vendor-specific components SHOULD have at least two different employers.
```
djaglowski, line 283: I believe we previously removed all language pertaining to the notion of "vendor-specific" components. Is this a meaningful distinction at this point?

mx-psi:
> Are there any plans regarding "technical stability/maturity"? For instance, a threshold of test coverage and/or linting issues.

@douglascamata Yes! see #11553 for the full list of things I want to cover in this document. If you think the list is incomplete or you would like to help me put this into wording feel free to comment on the issues/file a PR :)

docs/component-stability.md
mx-psi: Would your suggestion be to remove this requirement entirely? Or to make it apply to all components? (The latter feels difficult to me)

djaglowski: I think it could be removed entirely but this isn't something I feel strongly about. My main point is that it is an ambiguous requirement since we don't have a definition for vendor-specific component.

mx-psi: Removed, we can always add this later if we find a good definition of vendor-specific

mx-psi:
@djaglowski I will wait a bit more in case you have further comments and merge this on Monday morning

Issue #11778, opened 2024-12-02
Rename "profiles" modules to "experimental"


dmitryax:
In order to add support for the new profiling data to any of the components, the component has to use the factories defined in experimental modules with `profiles` suffix, e.g. [processorprofiles](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/processorprofiles) and [exporterprofiles](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterprofiles). The naming of the modules assumes that there is only one experimental module. If we want to add another experimental data signal (e.g. [entities](https://github.com/open-telemetry/opentelemetry-collector/pull/11282)), it would be required to add it in the same modules so the components can support them together. However, the current naming schema is reserved for the profiling data type.

I propose we rename the experimental factory modules from `processorprofiles` to `processorexp` (or `processorexperimental`) etc. This can be done as a 2-step change without breaking existing users.

PR #11889, opened 2024-12-13
Deprecate processorhelperprofiles module in favor of xprocessorhelper

dmitryax:
to allow adding more experimental data types

Updates https://github.com/open-telemetry/opentelemetry-collector/issues/11778


dmitryax:
I'm putting the `release:blocker` label with an intention to deprecate all the modules before the next release. Technically, it's not a blocker

songy23 (approved):

PR #11894, opened 2024-12-14
[chore] Update version of all new experimental modules to latest commit

dmitryax:
This unblocks updating core dependencies in contrib


songy23 (approved):
thanks!

Issue #11570, opened 2024-10-30
The `components` command should output providers


andrzej-stencel:
Now that the [Builder](https://opentelemetry.io/docs/collector/custom-collector/) allows to specify providers in its config, I would expect the providers to be listed in the output of the `otelcol components` command beside connectors, exporters, extensions, processors and receivers.

**Describe alternatives you've considered**

I don't know of any way to find what providers my build uses other than trying them out with `otelcol validate --config=provider:value`, which is cumbersome.

ChrsMark:
I would like to try my hand at it. Can I get assigned?

ChrsMark:
Now that [#11900](https://github.com/open-telemetry/opentelemetry-collector/pull/11900) is merged I think we can close this issue. (I don't remember why I didn't link them originally)

@andrzej-stencel @mx-psi @dmathieu feel free to close it if you agree.

PR #11900, opened 2024-12-16
Add support for providers in components command

ChrsMark:
#### Description
This PR adds support for printing config providers with the `components` command.

#### Link to tracking issue

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/11570

### DIFF
--- a/cmd/builder/internal/builder/templates/main.go.tmpl
+++ b/cmd/builder/internal/builder/templates/main.go.tmpl
@@ -33,7 +33,11 @@ func main() {
 					{{- range .ConfmapProviders}}
 					{{.Name}}.NewFactory(),
 					{{- end}}
-				},
+				}, ProviderModules: map[string]string{
+                    {{- range .ConfmapProviders}}
+                    "{{.Name}}": "{{.GoMod}}",
+                    {{- end}}
+                },
 				{{- if .ConfmapConverters }}
 				ConverterFactories: []confmap.ConverterFactory{
 					{{- range .ConfmapConverters}}
--- a/confmap/resolver.go
+++ b/confmap/resolver.go
@@ -39,6 +39,10 @@ type ResolverSettings struct {
 	// It is required to have at least one factory.
 	ProviderFactories []ProviderFactory
 
+	// ProviderFactories is a slice of Provider factories.
+	// It is required to have at least one factory.
+	ProviderModules map[string]string
+
 	// DefaultScheme is the scheme that is used if ${} syntax is used but no schema is provided.
 	// If no DefaultScheme is set, ${} with no schema will not be expanded.
 	// It is strongly recommended to set "env" as the default scheme to align with the
--- a/otelcol/command_components.go
+++ b/otelcol/command_components.go
@@ -31,6 +31,7 @@ type componentsOutput struct {
 	Exporters  []componentWithStability
 	Connectors []componentWithStability
 	Extensions []componentWithStability
+	Providers  []string
 }
 
 // newComponentsCommand constructs a new components command using the given CollectorSettings.
@@ -109,6 +110,14 @@ func newComponentsCommand(set CollectorSettings) *cobra.Command {
 				})
 			}
 			components.BuildInfo = set.BuildInfo
+
+			confmapProviderFactories := set.ConfigProviderSettings.ResolverSettings.ProviderFactories
+			for _, confmapProvider := range confmapProviderFactories {
+				provider := confmapProvider.Create(set.ConfigProviderSettings.ResolverSettings.ProviderSettings)
+				scheme := provider.Scheme()
+				components.Providers = append(components.Providers, scheme)
+			}
+
 			yamlData, err := yaml.Marshal(components)
 			if err != nil {
 				return err
--- a/otelcol/command_components.go
+++ b/otelcol/command_components.go
@@ -24,6 +24,11 @@ type componentWithStability struct {
 	Stability map[string]string
 }
 
+type componentWithOutStability struct {
+	Name   string
+	Module string
+}
+
 type componentsOutput struct {
 	BuildInfo  component.BuildInfo
 	Receivers  []componentWithStability
@@ -31,7 +36,7 @@ type componentsOutput struct {
 	Exporters  []componentWithStability
 	Connectors []componentWithStability
 	Extensions []componentWithStability
-	Providers  []string
+	Providers  []componentWithOutStability
 }
 
 // newComponentsCommand constructs a new components command using the given CollectorSettings.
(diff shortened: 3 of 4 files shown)


ChrsMark:
### Sample output

```console
./bin/otelcorecol_linux_amd64 components
buildinfo:
    command: otelcorecol
    description: Local OpenTelemetry Collector binary, testing only.
    version: 0.115.0-dev
receivers:
    - name: nop
      module: go.opentelemetry.io/collector/receiver/nopreceiver v0.115.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
    - name: otlp
      module: go.opentelemetry.io/collector/receiver/otlpreceiver v0.115.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
processors:
    - name: batch
      module: go.opentelemetry.io/collector/processor/batchprocessor v0.115.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
    - name: memory_limiter
      module: go.opentelemetry.io/collector/processor/memorylimiterprocessor v0.115.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
exporters:
    - name: debug
      module: go.opentelemetry.io/collector/exporter/debugexporter v0.115.0
      stability:
        logs: Development
        metrics: Development
        traces: Development
    - name: nop
      module: go.opentelemetry.io/collector/exporter/nopexporter v0.115.0
      stability:
        logs: Beta
        metrics: Beta
        traces: Beta
    - name: otlp
      module: go.opentelemetry.io/collector/exporter/otlpexporter v0.115.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
    - name: otlphttp
      module: go.opentelemetry.io/collector/exporter/otlphttpexporter v0.115.0
      stability:
        logs: Beta
        metrics: Stable
        traces: Stable
connectors:
    - name: forward
      module: go.opentelemetry.io/collector/connector/forwardconnector v0.115.0
      stability:
        logs-to-logs: Beta
        logs-to-metrics: Undefined
        logs-to-traces: Undefined
        metrics-to-logs: Undefined
        metrics-to-metrics: Beta
        metrics-to-traces: Undefined
        traces-to-logs: Undefined
        traces-to-metrics: Undefined
        traces-to-traces: Beta
extensions:
    - name: memory_limiter
      module: go.opentelemetry.io/collector/extension/memorylimiterextension v0.115.0
      stability:
        extension: Development
    - name: zpages
      module: go.opentelemetry.io/collector/extension/zpagesextension v0.115.0
      stability:
        extension: Beta
providers:
    - name: env
      module: go.opentelemetry.io/collector/confmap/provider/envprovider v1.21.0
    - name: file
      module: go.opentelemetry.io/collector/confmap/provider/fileprovider v1.21.0
    - name: http
      module: go.opentelemetry.io/collector/confmap/provider/httpprovider v1.21.0
    - name: https
      module: go.opentelemetry.io/collector/confmap/provider/httpsprovider v1.21.0
    - name: yaml
      module: go.opentelemetry.io/collector/confmap/provider/yamlprovider v1.21.0
```

cmd/builder/internal/builder/templates/main.go.tmpl
```diff
@@ -33,7 +33,11 @@ func main() {
 					{{- range .ConfmapProviders}}
 					{{.Name}}.NewFactory(),
 					{{- end}}
-				},
+				}, ProviderModules: map[string]string{
+                    {{- range .ConfmapProviders}}
+                    "{{.Name}}": "{{.GoMod}}",
+                    {{- end}}
+                },
```
evan-bradley, line 40:
```suggestion
					{{- range .ConfmapProviders}}
					"{{.Name}}": "{{.GoMod}}",
					{{- end}}
				},
```

otelcol/command_components.go
```diff
@@ -24,13 +24,19 @@ type componentWithStability struct {
 	Stability map[string]string
 }
 
+type componentWithOutStability struct {
```
andrzej-stencel, line 27: "without" is one word.
```suggestion
type componentWithoutStability struct {
```

andrzej-stencel (approved):
@ChrsMark sorry for the delay. I would love to see this merged. Can you please merge main  and run `make genotelcorecol` to update the provider versions in the generated `main.go` file?

ChrsMark:
Thank's for reviewing this @andrzej-stencel! The patch is up to date now.

/cc @mx-psi

confmap/resolver.go
```diff
@@ -39,6 +39,9 @@ type ResolverSettings struct {
 	// It is required to have at least one factory.
 	ProviderFactories []ProviderFactory
 
+	// ProviderModules maps provider types to their respective go modules.
+	ProviderModules map[string]string
```
mx-psi, line 43: The PR in general LGTM but I don't feel like this is the best place for this field. The resolver does not really use this field. Should this be in `otelcol.CollectorSettings` maybe?

otelcol/command_components.go
```diff
@@ -109,6 +115,18 @@ func newComponentsCommand(set CollectorSettings) *cobra.Command {
 				})
 			}
 			components.BuildInfo = set.BuildInfo
+
+			confmapProviderFactories := set.ConfigProviderSettings.ResolverSettings.ProviderFactories
+			for _, confmapProvider := range confmapProviderFactories {
+				provider := confmapProvider.Create(set.ConfigProviderSettings.ResolverSettings.ProviderSettings)
+				scheme := provider.Scheme()
+				module := set.ProviderModules[scheme+"provider"]
```
evan-bradley, line 123: Using the scheme as the name for the provider feels like it's not guaranteed to always work, I'm a bit concerned some providers may have a difference in the future even if all upstream providers have matching names and schemes.

Could we do the following instead:
1. In the builder, set up all the metadata we need given the Go module metadata. We should probably just do all the transformations we need here.
2. Loop through `set.providerModules` and retrieve this data directly.

What do you think?

evan-bradley: Thinking on it more, we should call out the scheme instead of or in addition to the module name since the scheme is the part users actually care about. So we'll need to still read from `confmapProviderFactories` to get the scheme.

However I still don't think we can guarantee that future providers won't have names like `acmecorpsecretsprovider` that read URIs with schemes like `acs://corp.my/secret`. We could potentially use reflection to find the module for a given provider and match them up, but I'm not sure about the best way to do this given the provider interfaces don't give a module name.

ChrsMark: I see! Would that make sense to extend the provider interface so as to be able to provide the package or module name?

evan-bradley: Unfortunately we likely can't expand the interface any further now that confmap is stable. However it looks like we can do something like this in the map inside main.go.tmpl:

```
{{.Name}}.NewFactory().Create(confmap.ProviderSettings{}).Scheme()
```

If we key on the scheme like this and have the go module as the value, I think we only need a few changes inside the main.go template.

ChrsMark: That works, indeed! Thank's for the suggestion!

otelcol/testdata/components-output.yaml
```diff
@@ -41,3 +41,6 @@ extensions:
       module: go.opentelemetry.io/collector/extension/extensiontest v1.2.3
       stability:
         extension: Stable
+providers:
+    - name: nop
```
evan-bradley, line 45: One last thing, should we call this `scheme` instead of `name`?

ChrsMark: Sure! I changed that.

ChrsMark:
Hey @evan-bradley, I think your comments should be addressed now. Feel free to take a final look.

evan-bradley (approved):

PR #11932, opened 2024-12-17
Upgrade pdata to v1.5.0

dmathieu:
#### Description

Proto 1.5.0 has been released. This upgrades pdata to use that version.
See https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.5.0


pdata/pprofile/profile.go
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package pprofile // import "go.opentelemetry.io/collector/pdata/pprofile"
+
+import "go.opentelemetry.io/collector/pdata/pcommon"
+
+// Attributes returns the Attributes associated with this Profile.
+//
+// Deprecated: [v0.117.0] This field has been removed, and replaced with AttributeIndices
+func (ms Profile) Attributes() pcommon.Map {
```
dmathieu, line 11: This method prevents panics on contrib. It's not a great solution either though, as it makes attributes readable, but not writable (and we can't make them writable, since we can't look for changes in `pcommon.Map`).

dmathieu:
Unfortunately, due to a change in the way profile attributes are handled, this brings a breaking change to contrib, which isn't really fixable (we can read the data from the old method, but not write).
The contrib fix is here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36868

I'd suggest a coordinated merge of the two PRs.

exporter/debugexporter/internal/normal/profiles.go
```diff
@@ -35,10 +36,15 @@ func (normalProfilesMarshaler) MarshalProfiles(pd pprofile.Profiles) ([]byte, er
 				buffer.WriteString(" samples=")
 				buffer.WriteString(strconv.Itoa(profile.Sample().Len()))
 
-				if profile.Attributes().Len() > 0 {
-					profileAttributes := writeAttributes(profile.Attributes())
+				if profile.AttributeIndices().Len() > 0 {
+					attrs := []string{}
+					for _, i := range profile.AttributeIndices().AsRaw() {
+						a := profile.AttributeTable().At(int(i))
+						attrs = append(attrs, fmt.Sprintf("%s=%s", a.Key(), a.Value().AsString()))
```
rockdaboot, line 43: Can you write directly into the buffer here?

rockdaboot, line 43: Adding the space in front of the format string (` %s=%s`) avoids having a space+linefeed at the end.

rockdaboot, line 43: What about values containing spaces? Better escape + quote it, maybe using `%q`.

dmathieu: I'd rather not refactor this, to keep what exactly was changed readable.

dmathieu: To keep things readable, this should happen in another PR.

rockdaboot: Hm, looks to me that this has been introduced by your changes. Another PR is easily forgotten without a tracking issue.

dmathieu: No, we previously used the common `writeAttributes` method, which does the same thing.
https://github.com/open-telemetry/opentelemetry-collector/blob/4fc50a85ed71025010b18ceae315accba2bfa0ab/exporter/debugexporter/internal/normal/common.go#L15

Changing it there impacts all signals. Hence the new PR.

pdata/pprofile/profile.go
mx-psi: I wonder if it makes more sense to just remove this. It's a bigger breakage, but it's more noticeable that writes silently failing

mx-psi (approved):

PR #11830, opened 2024-12-09
Added Zap as the default Error log #11820

tarun-kavipurapu:
#### Description
This Issue Fixes the Issue #11820 
Fixes 
- Made the Zap as the default ErrorLog Supplied to the Http server

PR #11935, opened 2024-12-17
Add Zap as the default error log in confighttp

### LINKED ISSUE #11820: confighttp.ToServer should be setting ErrorLog

Standard `http.Server` has a field `ErrorLog`:
```
	// ErrorLog specifies an optional logger for errors accepting
	// connections, unexpected behavior from handlers, and
	// underlying FileSystem errors.
	// If nil, logging is done via the log package's standard logger.
	ErrorLog *log.Logger
```

However, `confighttp.ToServer` leaves it empty, which will lead to error logs printed using a different logger.

In Jaeger we initialize it with a logger backed by Zap:
```
	errorLog, _ := zap.NewStdLogAt(params.Logger, zapcore.ErrorLevel)
	server.ErrorLog = errorLog
```

I think `confighttp.ToServer()` should do the same.

tarun-kavipurapu:
#### Description
This Issue Fixes the Issue https://github.com/open-telemetry/opentelemetry-collector/issues/11820
This is a continuation pull request for #11830 

Fixes #11820 
 -Made the Zap as the default ErrorLog Supplied to the Http server


.chloggen/zap-default-errorlog.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confighttp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Added Zap as the default error log in confighttp "
```
dmathieu, line 10:
```suggestion
note: "Added Zap as the default error log in confighttp"
```

config/confighttp/confighttp.go
```diff
@@ -478,16 +480,18 @@ func (hss *ServerConfig) ToServer(_ context.Context, host component.Host, settin
 		next:            handler,
 		includeMetadata: hss.IncludeMetadata,
 	}
+	errorLog, err := zap.NewStdLogAt(settings.Logger, zapcore.ErrorLevel)
```
dmathieu, line 483: We should return errors as soon as they happen, not later.

```suggestion
	errorLog, err := zap.NewStdLogAt(settings.Logger, zapcore.ErrorLevel)
	if err != nil {
		 return nil, err
	}
```

config/confighttp/confighttp.go
```diff
@@ -478,16 +480,18 @@ func (hss *ServerConfig) ToServer(_ context.Context, host component.Host, settin
...
 		Handler:           handler,
 		ReadTimeout:       hss.ReadTimeout,
 		ReadHeaderTimeout: hss.ReadHeaderTimeout,
 		WriteTimeout:      hss.WriteTimeout,
 		IdleTimeout:       hss.IdleTimeout,
+		ErrorLog:          errorLog,
 	}
 
-	return server, nil
+	return server, err
```
dmathieu, line 494:
```suggestion
	return server, nil
```

config/confighttp/confighttp.go
tarun-kavipurapu: the suggested change was the  initial code too
 but in 
 https://github.com/open-telemetry/opentelemetry-collector/pull/11830/files/71c16e0b293b4d6ed25146c4dd295093e672afd7#r1880527497
 it was suggested to return server,err so that tests can pass

dmathieu: Then, how about something like:

```golang
server := &http.Server{
	Handler:           handler,
	ReadTimeout:       hss.ReadTimeout,
	ReadHeaderTimeout: hss.ReadHeaderTimeout,
	WriteTimeout:      hss.WriteTimeout,
	IdleTimeout:       hss.IdleTimeout,
}

errorLog, err := zap.NewStdLogAt(settings.Logger, zapcore.ErrorLevel)
if err != nil {
	return server, err
}

server.ErrorLog = errorLog
return server, nil
```

Then, we can handle multiple things that could return errors, and we also keep error return alongside its definition.

tarun-kavipurapu: I will check the report locally and may be make the change according to it 
the command for it is `make gotest-with-cover` right?

tarun-kavipurapu:
@dmathieu @dmitryax Hey any reason this is waiting to be merged i mean am i expected to make any changes?

PR #11849, opened 2024-12-11
[chore] Fix "Prepare Release" action

jade-guiton-dd:
#### Description

The "Prepare Release" action failed during the latest release process, as the OpenTelemetry Bot, which opens the release PR, does not have the permissions to add the new `release:merge-freeze` label. This PR fixes this by creating the PR with the bot token, then adding the label separately using the regular CI token.

Since we already use default CI tokens on -contrib to add labels, I assume they have the permission to do that on this repo as well. I deemed this simpler than the alternative solutions mentioned in the tracking issue, though I'm not certain it will work.

Note that the tracking issue is a release blocker.

#### Link to tracking issue
Fixes #11808

#### Testing
I'm not sure how I could test this without getting it merged first unfortunately. Once merged, we could try running the action, then closing the resulting PR and issue.

PR #11936, opened 2024-12-17
[chore] Merge freeze check uses title to find Release PRs

### LINKED ISSUE #11808: "Prepare Release" action fails to add label
I recently merged a PR (#11744) to add a CI check which prevents PRs from being merged in `main` while a "Prepare Release" PR is open, enforcing the merge freeze described in the [release procedure](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md). To detect the Prepare Release PR, I decided to rely on a new "release:merge-freeze" label, to be added automatically to said PR.

However, during the latest release, the ["Prepare Release" action failed](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/12130228415/job/33840863531), with the message _"opentelemetrybot does not have permission to update the pull request"_. The PR was still created, but the label was not added, presumably because the `OPENTELEMETRYBOT_GITHUB_TOKEN` used by the action does not have the permissions to add labels to PRs.

I see two possible solutions:
1. Generate a new token with the appropriate additional permission;
2. Have the bot comment on the PR / add something in the description that triggers another workflow which adds the label.

### LINKED ISSUE #11906: Label release:merge-freeze should not block the release PR
See an example in https://github.com/open-telemetry/opentelemetry-collector/pull/11904

jade-guiton-dd:
#### Description

The last attempt at making the Merge freeze check work in release PRs failed (#11906). This PR tries a different approach: it changes the criteria of the Merge freeze check, so that a freeze is enacted iff there is an open PR authored by @opentelemetrybot whose title contains "[chore] Prepare release" (note that if it weren't for the author, this PR would qualify).

This PR additionally reverts #11849, so no label is added to the release PR. I also added the `pull_request.enqueued` trigger, taking inspiration [from Merge Freeze](https://docs.mergefreeze.com/github-merge-queue#how-it-works), to see if it could help reject PRs earlier.

I tried to make sure the freeze check would be properly skipped for the release PR itself, both in PR checks and in the merge queue, but given the state of Github's documentation, I'm not very confident about this.

Notably, these is an edge case where I'm not sure what would happen: what if another PR gets added to the merge queue at the same time as the release PR? How many times would the "merge_group" check run, and with what values for "github.event.merge_group.head_commit"? Would both PR be booted out of the queue (not great)? Would both be accepted (way worse)? Does it depend on the order?

#### Link to tracking issue
Fixes #11906 and fixed #11808

#### Testing
As always with this, it's pretty much impossible to test before merging. Once merged, I strongly recommend we do the following test to make sure that this issue does not block the real release process again:
- Create two dummy PRs that change nothing of consequence: the freeze check should pass
- Run the "Prepare release" action
- Rerun the freeze check on one of the dummy PRs: it should now fail
- Approve the second PR and try to merge it: it should be booted out of the merge queue
- Close all test PRs

This unfortunately does not test whether the release PR gets merged properly, but I don't see how to test until the next release process, unfortunately.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -101,8 +101,7 @@ jobs:
       #   - Run make prepare-release PREVIOUS_VERSION=0.52.0 RELEASE_CANDIDATE=0.53.0 MODSET=beta
       - name: Prepare release for core
         env:
-          BOT_GITHUB_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
           REPO: open-telemetry/opentelemetry-collector
           CANDIDATE_BETA: ${{ inputs.candidate-beta }}
           CANDIDATE_STABLE: ${{ inputs.candidate-stable }}
--- a/.github/workflows/check-merge-freeze.yml
+++ b/.github/workflows/check-merge-freeze.yml
@@ -11,7 +11,9 @@ jobs:
   check-merge-freeze:
     name: Check
     # This condition is to avoid blocking the PR causing the freeze in the first place.
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'release:merge-freeze') }}
+    if: |
+      (!startsWith(github.event.pull_request.title || github.event.merge_group.head_commit.message, '[chore] Prepare release')) ||
+      (github.event.pull_request.user.login || github.event.merge_group.head_commit.author.name) != 'opentelemetrybot'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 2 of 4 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -101,8 +101,7 @@ jobs:
       #   - Run make prepare-release PREVIOUS_VERSION=0.52.0 RELEASE_CANDIDATE=0.53.0 MODSET=beta
       - name: Prepare release for core
         env:
-          BOT_GITHUB_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.OPENTELEMETRYBOT_GITHUB_TOKEN }}
```
mx-psi, line 104: I don't understand this change

mx-psi (commented):
This looks good overall, I can take care of the testing

.github/workflows/prepare-release.yml
jade-guiton-dd: Originally, the release script used the OTelBot token to create the PR (we can't use a normal CI token because workflows won't run). I previously changed it to use OTelBot token to create the PR, and the CI token to add the label. Since we're no longer adding a label, I'm just reverting it to how it used to be.

codeboten (approved):

PR #11940, opened 2024-12-17
[chore] Unrevert automatic `replace` generation in builder-integration-test

### LINKED ISSUE #11607: [release] builder integration test succeeds on main but fails on release prep
**Describe the bug**

When a new module was added since the last release, the `Builder - Integration tests` github action succeeds on main but fails on the [release prep PR](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/11687864326/job/32546947029?pr=11602#step:4:24).

CI should be able to catch this issue immediately and appropriate replace statements should be required in the codebase.

**What did you see instead?**

The release process fails.

---

Is it possible to detect and require a solution to this problem at the time the module was added? What new tests or changes to existing tests are necessary to ensure releases are not blocked by this in the future?

jade-guiton-dd:
#### Description

PR #11793, which added code to automatically generate `replace` clauses in builder-integration-test in order to avoid issues in release PRs, caused [issues in a release PR](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/12358329683/job/34488569900?pr=11904), and was reverted in #11909.

The issue ended up being that the `replace` clause for the root collector module (`go.opentelemetry.io/collector`) was missing, because ([cf. previous source code](https://github.com/open-telemetry/opentelemetry-collector/blob/6a47030ba2bf247ec2c533322d1beaf5eada8981/cmd/builder/test/test.sh#L118)) I was generating a list of module paths relative to the root, with an empty string for the root module, which bash helpfully ignored during the for loop below.

This PR unreverts #11793, with a fix for the above issue. (We make sure to only strip the initial dot of the folder path _inside_ the for loop.)

#### Link to tracking issue
Fixes #11607, hopefully correctly this time.

#### Testing
I replicated the release PR in which the issues arose with a .1 patch version number, replicated the failure that was observed, and checked that the new script has the expected behavior.


songy23:
In case it helps: the failed CI during the release is https://github.com/open-telemetry/opentelemetry-collector/actions/runs/12358329683/job/34488569900?pr=11904

songy23 (approved):
please rebase

cc @dmitryax release manager for the next release

PR #11946, opened 2024-12-17
[chore] Use gotestsum wrapper for go tests

ankitpatel96:
**Description:**

Use [gotestyourself/gotestsum](https://github.com/gotestyourself/gotestsum?rgh-link-date=2024-02-09T12%3A20%3A03Z) for running tests. 
This is following contrib's lead - they implemented this in this [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31163).  

Gotestsum is a lot more readable and user friendly than go test - the output format is nicer and the test output is more parseable.  We could also use [rerun on failed](https://github.com/gotestyourself/gotestsum#re-running-failed-tests) or some of their other features!


mx-psi (approved):
LGTM since we already use this in contrib

codeboten (approved):

PR #11947, opened 2024-12-18
RFC: Error transmission through batching components

### LINKED ISSUE #11308: Batch processor concurrency and error transmission

jmacd:
#### Description

Draft RFC addresses #11308.

#### Link to tracking issue

Fixes #11308.

#### Testing

n/a

#### Documentation

This is a new RFC. As these changes are accepted and made, user-facing
documentation will be added.


jmacd:
I will present this RFC in the next two Collector SIG meetings 1/7/2025 (APAC/PT) and 1/15/2025 (NA).

docs/rfcs/batching-process-design.md
```diff
@@ -0,0 +1,282 @@
...
+A central aspect of this presentation focuses on the queue sender,
+which along with one of the batching processors defines the behavior
+of most OpenTelemetry pipelines. Therefore, the default queue sender
+behavior is critical.
+
+This proposal argues in favor of the user, who does not want default
+behavior that leads to data loss. This requires one or more changes
+in the exporterhelper:
+
+1. Disable the queue sender by default. In this configuration, requests
```
dmitryax, line 180: This is a significant change in the behavior. I don't remember any other user requests to make the in-memory queue disabled by default. I believe a significant portion of users would favor keeping in-memory queue enabled by default, even if it's prone to data loss in case of Collector's crash. If we want to change that, there should be a more extended discussion and evaluation. AFAIR, in-memory buffering is a common default setting in other data processing agents.

docs/rfcs/batching-process-design.md
```diff
@@ -0,0 +1,282 @@
...
+
+The batch sender MUST be modified to achieve the following
+outcomes:
+
+- Allow concurrent splitting. When multiple full-size requests
+  are produced from an input, they are exported concurrently and
+  independently.
+- Recognize partial splitting. When a request is split, leaving
+  part of a request that is not full, it remains in the active
+  request (i.e., still pending).
```
dmitryax, line 212: I'm afraid this might introduce a significant delay when several incoming requests that are partially split block each other until the timeout is reached.

jmacd:
This work is out of date since #12245, #12233, #12118, #12154, #12246, #12242, and so on.

I believe all of the above work deserves an RFC.

PR #11951, opened 2024-12-18
[exporterqueue] Implement blocking queue

sfc-gh-sili:
#### Description

This PR

1. Implements blocking memory queue. The queue blocks until the incoming request is successgfully sent out.
2. Uses blocking queue + queue batcher in place of `batch_sender` if the feature gate `UsePullingBasedExporterQueueBatcher` is on.

#### Link to tracking issue


dmitryax:
@sfc-gh-sili please rebase and look at the failing CI

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -101,8 +101,23 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
 		be.QueueSender = NewQueueSender(q, be.Set, be.queueCfg.NumConsumers, be.ExportFailureMessage, be.Obsrep, be.BatcherCfg)
 	}
 
-	if !usePullingBasedExporterQueueBatcher.IsEnabled() && be.BatcherCfg.Enabled ||
-		usePullingBasedExporterQueueBatcher.IsEnabled() && be.BatcherCfg.Enabled && !be.queueCfg.Enabled {
+	if usePullingBasedExporterQueueBatcher.IsEnabled() && be.BatcherCfg.Enabled && !be.queueCfg.Enabled {
+		be.queueFactory = exporterqueue.NewBlockingMemoryQueue[internal.Request]()
+		be.queueCfg.QueueSize = 20
```
dmitryax, line 106: Why do we need to set it to 20?

sfc-gh-sili: Thanks, Dmitrii. 
The blocking queue is introduced as a hidden logical abstraction to handle the case when batch is enabled and queue is not enabled - it means a component that blocks until the incoming request is successfully sent out, so **logically the queue size is infinite.** I'd think setting this number to MaxInt would make sense but that broke a bunch of tests, and 20 is just some random number I picked as a work around.

Let me experiment with some alternative solutions to "infinite queue" and get back to this.

sfc-gh-sili: Hi Dmitrii, I made a temporary hack that treats `queueSize = -1` as

* A queue that blocks until the items are read and "processed"
* and it has unlimited capacity (it never returns "queue is full" error)

but I think the proper way is to use a slimmer version of `sizedChannel` (I'll implement in the next PR)

bogdandrutu:
https://github.com/open-telemetry/opentelemetry-collector/pull/12090

PR #11956, opened 2024-12-18
Amend Pipeline Component Telemetry RFC to add a "rejected" outcome

jade-guiton-dd:
### Context

The [Pipeline Component Telemetry RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md) was recently accepted (#11406). The document states the following regarding error monitoring:
> For both [consumed and produced] metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to whether or not the corresponding function call returned an error. Specifically, consumed measurements will be recorded with `outcome` as `failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, produced measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and `success` otherwise.

[Observability requirements for stable pipeline components](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#observability-requirements) were also recently merged (#11772). The document states the following regarding error monitoring:
> The goal is to be able to easily pinpoint the source of data loss in the Collector pipeline, so this should either:
> - only include errors internal to the component, or;
> - allow distinguishing said errors from ones originating in an external service, or propagated from downstream Collector components.

Because errors are typically propagated across `ConsumeX` calls in a pipeline (except for components with an internal queue like `processor/batch`), the error observability mechanism proposed by the RFC implies that Pipeline Telemetry will record failures for every component interface upstream of the component that actually emitted the error, which does not match the goals set out in the observability requirements, and makes it much harder to tell which component errors are coming from from the emitted telemetry.

### Description

This PR amends the Pipeline Component Telemetry RFC with the following:
- restrict the `outcome=failure` value to cases where the error comes from the very next component (the component on which `ConsumeX` was called);
- add a third possible value for the `outcome` attribute: `rejected`, for cases where an error observed at an interface comes from further downstream (the component did not "fail", but its output was "rejected");
- propose a mechanism to determine which of the two values should be used.

The current proposal for the mechanism is for the pipeline instrumentation layer to wrap errors in an unexported `downstream` struct, which upstream layers could check for with `errors.As` to check whether the error has already been "attributed" to a component. This is the same mechanism currently used for tracking permanent vs. retryable errors. Please check the diff for details.

### Possible alternatives

There are a few alternatives to this amendment, which were discussed as part of the observability requirements PR:
- loosen the observability requirements for stable components to not require distinguishing internal errors from downstream ones → makes it harder to identify the source of an error;
- modify the way we use the `Consumer` API to no longer propagate errors upstream → prevents proper propagation of backpressure through the pipeline (although this is likely already a problem with the `batch` prcessor);
- let component authors make their own custom telemetry to solve the problem → higher barrier to entry, especially for people wanting to opensource existing components.


jade-guiton-dd:
Happy new year to you too, stale Github action.

djaglowski (approved):

docs/rfcs/component-universal-telemetry.md
```diff
@@ -90,11 +90,20 @@ The location of these measurements can be described in terms of whether the data
...
+Specifically, a call to `ConsumeX` is recorded with:
+- `outcome = success` if the call returns `nil`;
+- `outcome = failure` if the call returns a regular error;
+- `outcome = rejected` if the call returns an error tagged as coming from downstream.
+After inspecting the error, the instrumentation layer should tag it as coming from downstream before returning it to the parent component.
+
+The upstream component which called `ConsumeX` will have this `outcome` attribute applied to its produced measurements, and the downstream
+component that `ConsumeX` was called on will have the attribute applied to its consumed measurements.
+
+Errors should be "tagged as coming from downstream" the same way permanent errors are currently handled: they can be wrapped in a `type downstreamError struct { err error }` wrapper error type, then checked with `errors.As`. Note that care may need to be taken when dealing with the `multiError`s returned by the `fanoutconsumer`. (If PR #11085 introducing a single generic `Error` type is merged, an additional `downstream bool` field can be added to it to serve the same purpose.)
```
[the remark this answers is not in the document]
jade-guiton-dd: You're absolutely right. I think those components would already be broken anyway, because of the `permanentError` and `multiError` wrappers we already use.

mx-psi (approved):

docs/rfcs/component-universal-telemetry.md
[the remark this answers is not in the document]
jade-guiton-dd: If #11085 gets merged before this PR, I'll update this paragraph to only include the parenthetical. If this PR gets merged first, I think presenting the two alternatives is probably good enough?

docs/rfcs/component-universal-telemetry.md
```diff
@@ -90,11 +90,20 @@ The location of these measurements can be described in terms of whether the data
...
 while metrics which contain the term "consumed" describe data which is received by the component.
 
-For both metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to
-whether or not the corresponding function call returned an error. Specifically, consumed measurements will be recorded with `outcome` as
-`failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, produced
-measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and
-`success` otherwise.
+For both metrics, an `outcome` attribute with possible values `success`, `failure`, and `rejected` should be automatically recorded,
+corresponding to whether or not the corresponding function call returned an error, and whether the error originates from the next
+component(s) in the pipeline, or from one further downstream.
```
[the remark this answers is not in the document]
jade-guiton-dd: I find this wording is a bit confusing. What would you think of the following?

> based on whether the corresponding function call returned successfully, returned an internal error, or propagated an error from a component further downstream.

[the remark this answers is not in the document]
jaronoff97: perfect :) thank you!

mx-psi:
Given the two approvals and the announcements on #otel-collector-dev and the Collector SIG meeting from last week, this is entering final comment period. cc @open-telemetry/collector-approvers

docs/rfcs/component-universal-telemetry.md
```diff
@@ -90,11 +90,20 @@ The location of these measurements can be described in terms of whether the data
...
 while metrics which contain the term "consumed" describe data which is received by the component.
 
-For both metrics, an `outcome` attribute with possible values `success` and `failure` should be automatically recorded, corresponding to
-whether or not the corresponding function call returned an error. Specifically, consumed measurements will be recorded with `outcome` as
-`failure` when a call from the previous component the `ConsumeX` function returns an error, and `success` otherwise. Likewise, produced
-measurements will be recorded with `outcome` as `failure` when a call to the next consumer's `ConsumeX` function returns an error, and
-`success` otherwise.
+For both metrics, an `outcome` attribute with possible values `success`, `failure`, and `rejected` should be automatically recorded,
+based on whether the corresponding function call returned successfully, returned an internal error, or propagated an error from a
+component further downstream.
```
jmacd, line 95: I would like to ensure there is detail about the OTLP PartialSuccess fields `rejected_spans`, `rejected_log_records`, `rejected_data_points`. Here's what I propose:

An OTLP exporter that receives success with one of the rejected counts set will:

- return `nil` indicating success
- count `N-j` success points where N is the item count and j is the number rejected
- count `j` rejected points

jade-guiton-dd: In this RFC, the exporter wouldn't be the one outputting the metrics; this would be done by an external mechanism — a wrapper of the `Consumer` API placed in between each component.

This means that adding information about partial successes to pipeline instrumentation metrics would require a whole new mechanism for propagating partial successes through the Collector pipeline, and probably other considerations such as how retry exporters should handle them, which sounds like a much broader discussion...

Considering this mechanism would likely only be used by the OTLP exporter, I would suggest emitting a custom metric in said exporter instead of including it as part of general pipeline telemetry. Even in the latter case, I think this may warrant a separate PR, or even an RFC of its own.

mx-psi: From the conversation above this seems like something that can be dealt with independently from this PR.

@jmacd I will merge this on Tuesday unless you consider this a blocker for this PR

djaglowski: > The best way to avoid my concerns, in the short term, is not to overload the term "rejected" in favor of the term we've used in the past, "refused".

This seems reasonable to me. I don't think we're intending to modify any definitions outside of what is defined in this RFC. 

The larger problem, if I understand correctly, boils down to reconciling the "all or nothing" interface which stands in between components vs the notion of partial success between the exporter and destination. I suggest this ought to be proposed in a separate PR.

jmacd: I don't wish to block this effort, and I agree we should disambiguate requests that fail inside a component vs downstream. The term "reject" was chosen for its dictionary definition, "dismiss as inadequate", "failure to meet standards", because it describes a returned judgement about the data. If there's a downstream failure because of timeout, unavailable destination, and so on, the term "reject" feels less applicable to me. For "refuse" the dictionary has a more applicable definition ("not willing to perform an action").

I think "rejected" is a OK term for downstream failures, but receivers have been using "refused" for this. Mostly, hope we can eventually count the OTLP partial success rejections. This is discussed in #9243.

jade-guiton-dd: I see, I wasn't at all aware of the ongoing discussions about partial successes, and this distinction between "internal error / invalid data" and "data is valid but intentionally rejected for backend-specific reasons". Given this and the current use of the word "refused" in receiver metrics, I absolutely agree we should use that instead. To be honest, I hadn't thought very hard about the exact word used, so I'll update the PR with that in mind.

mx-psi:
I'll give this a few more days since there was a minor change and merge this on Tuesday next week

PR #3126, opened 2021-05-07
Validate that batch config max size is greater than send size

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/3119

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

Issue #3119, opened 2021-05-06
Batch Processor send_batch_max_size 


kisieland:
**Describe the bug**
The main issues I would like to summarise here is connected to how the `send_batch_max_size` and `send_batch_size` flags interact.
Currently `processItem` function does something like:
```
if bp.sendBatchMaxSize > 0 {
  itemCount := bp.batch.itemCount()
  if itemCount+uint32(td.MetricCount()) > bp.sendBatchMaxSize {
  tdRemainSize := splitMetrics(int(bp.sendBatchSize-itemCount), td)
  item = tdRemainSize
  // put td back to the queue
}

bp.batch.add(item)
if bp.batch.itemCount() >= bp.sendBatchSize {
  // send data and clear batch
}
```
Note that this checks if batch after adding new data is bigger than `sendBatchMaxSize`, but when splitting uses `sendBatchSize` to compute remaining space. 
In effect (when `sendBatchSize > sendBatchMaxSize`) batch sent can be bigger than `sendBatchMaxSize`.
If you look at the latter part of the code it shows that the batch is sent only when it is bigger than `sendBatchSize`.
(For the simplicity I did not cover timeouts).

I would expect to batch be send as soon as size of `sendBatchMaxSize` is reached. Possible fix is to set those fields to the same values. I believe this should be at least mentioned in the Readme.

The original issue that brought me here is the fact that when configured in a way that `sendBatchSize` is much bigger than `sendBatchMaxSize`. Then the batch processor is using a lot of CPU (about 50x more than normally). 
I made some profiling and it spends majority of the time on scheduling and creating gorutines in `processItem`.
I will update the bug once I find some more information about the source of this CPU usage.

**What version did you use?**
Version: `v0.23.0`

**What config did you use?**
Config:
```
processors:
  batch:
    send_batch_max_size: 200
    timeout: 10s
```

bogdandrutu:
There were some fixes related to this in the latest release, but we still have the problem if the values are reversed. Maybe that should be a validation in the config to not allow that

bogdandrutu:
@kisieland I closed this since the main issue is not resolved. If you want to still improve performance please open an issue for that

PR #3029, opened 2021-04-26
Fix batchprocessor to avoid reordering and send max size

bogdandrutu:
* The reordering could happen if during processing of an item other items were added to the newItems channel. This is fixed by avoiding to re-add the left items to the channel.
* The logic of spliting was wrong by forcing sendBatchSize instead of sendBatchMaxSize when split was called.
* Fix logic if a very large message is received > 2x sendBatchMaxSize to call export multiple times instead of once and reset the timer.

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

PR #11962, opened 2024-12-19
[processor/batch] Allow multiple batches for each batch size trigger

lopes-felipe:
#### Description
There is a config validation which enforces that `send_batch_max_size` must be greater than `send_batch_size`. This makes it impossible to have a setup of a batch that is triggered when a number of items are processed (`SendBatchSize`), but we want to break that set into multiple smaller batches (`SendBatchMaxSize` < `SendBatchSize`).

This logic was introduced [here](https://github.com/open-telemetry/opentelemetry-collector/pull/3126), due to [this](https://github.com/open-telemetry/opentelemetry-collector/issues/3119) issue, but the underlying bug was fixed [here](https://github.com/open-telemetry/opentelemetry-collector/pull/3029), so this validation shouldn't exist anymore.


lopes-felipe:
Hey @codeboten. Could you please take a look whenever you can? Thanks! :)

jpkrohling:
cc @bogdandrutu , as you worked on similar PRs in the past.

PR #11994, opened 2024-12-28
[chore] Remove impossible infinite workers use-case for Batcher

bogdandrutu:
It is impossible to configure infinite workers today, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterqueue/config.go#L41

### DIFF
--- a/exporter/internal/queue/batcher.go
+++ b/exporter/internal/queue/batcher.go
@@ -25,10 +25,8 @@ type Batcher interface {
 }
 
 type BaseBatcher struct {
-	batchCfg exporterbatcher.Config
-	queue    exporterqueue.Queue[internal.Request]
-	// TODO: Remove when the -1 hack for testing is removed.
-	maxWorkers int
+	batchCfg   exporterbatcher.Config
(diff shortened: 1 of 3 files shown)


dmitryax:
Ok this is dead code, but we still want to support that eventually, right?

bogdandrutu:
@dmitryax do we?

dmitryax:
We need that behavior if the queue is disabled, but batching is enabled. So, the validation is correct. I'm not sure if @sfc-gh-sili needs this code for https://github.com/open-telemetry/opentelemetry-collector/pull/11951

jmacd:
> We need that behavior if the queue is disabled, but batching is enabled. So, the validation is correct. I'm not sure if @sfc-gh-sili needs this code for https://github.com/open-telemetry/opentelemetry-collector/pull/11951

@sfc-gh-sili could you help clarify?  I'm trying to understand how, when the queue is disabled, concurrency will be limited (or not limited) by callers, as well I am trying to understand whether cancellation is supported. Related to #11947.

dmitryax (approved):

PR #11963, opened 2024-12-19
[chore] Create gotest-with-junit Makefile target

ankitpatel96:
Creates a target to output junit files from testing.  This PR also starts to use it in CI - we don't consume the Junits yet but it would be easy to add on.  This will be useful for a number of CI/Devx initiatives in the future.

My company, Datadog, plans to use this information to track the build stability of our code in contrib, as well as the stability of code we depend on here in collector.

PR #12001, opened 2024-12-31
[chore] tweak junit.xml upload

ankitpatel96:
#### Description
This PR tweaks the output junit test files introduced in https://github.com/open-telemetry/opentelemetry-collector/pull/11963. 
 It introduces underscores as a delimiter, adds the architecture to name,  and add test file outputs to arm and windows builds.

#### Testing
Hopefully just these unit tests.  For some reason, the arm tests aren't running, and I'm not sure why.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -141,9 +141,9 @@ jobs:
   unittest-matrix:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        runner: [ubuntu-latest]
         go-version: ["~1.23", "~1.22"] # 1.20 needs quotes otherwise it's interpreted as 1.2
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.runner }}
     needs: [setup-environment]
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -141,15 +141,16 @@ jobs:
   unittest-matrix:
     strategy:
       matrix:
-        runner: [ubuntu-latest]
```
songy23, line 144: is this relevant to junit?

songy23 (approved):

songy23:
> For some reason, the arm tests aren't running, and I'm not sure why.

It needs the `Run Arm` label (added)

.github/workflows/build-and-test.yml
```diff
@@ -160,18 +161,18 @@ jobs:
...
           path: ~/.cache/go-build
-          key: unittest-${{ runner.os }}-${{ matrix.runner }}-go-build-${{ matrix.go-version }}-${{ hashFiles('**/go.sum') }}
+          key: unittest-${{ runner.os }}-${{ matrix.os }}-go-build-${{ matrix.go-version }}-${{ hashFiles('**/go.sum') }}
       - name: Run Unit Tests
         run: |
           make -j4 gotest-with-junit
       - uses: actions/upload-artifact@v4
         with:
-          name: test-results-${{ runner.os }}-${{ matrix.runner }}-${{ matrix.go-version }}
+          name: test-results_${{ runner.os }}_${{ matrix.os }}_${{ runner.arch }}_${{ steps.setup-go.outputs.go-version }}
```
codeboten, line 175: wont this result in duplicated os name in the `name` string?

ankitpatel96: Sort of - not quite.  Here's what the artifacts look like - 
 `test-results_Linux_ubuntu-latest_X64_1.22.10 `
so the runner.os is Linux, Macos, etc, and the matrix.os is what we use to request a runner - so ubuntu-latest in this particular example.  I captured both since technically one day we may test against different Linux distros

.github/workflows/build-and-test.yml
ankitpatel96: i use matrix.os later on - and we use matrix.os in other files.  I chose to align the different files.

.github/workflows/build-and-test.yml
ankitpatel96: I changed my mind.  If we ever want to do this we can - for now I eliminated the matrix.os variable for simplicity.  Plus the information is pretty duplicated and unspecific

.github/workflows/build-and-test.yml
ankitpatel96: Reversed this

songy23 (approved):

PR #12012, opened 2025-01-04
[chore] Deprecate the interface exporter.Request.Merge()

sfc-gh-sili:
#### Description

This PR removes the interface `exporter.Request.Merge()` because `Merge()` is just a special case of `MergeSplit()`.
With just one interface, we no longer need an if-else loop in the batcher based on MaxConfigSize. `MergeSplit()` should be able to handle both the case where MaxConfigSize is specified or the case where it is not

#### Link to tracking issue
Fixes #

#### Testing

All the existing cases should still pass.

### DIFF
--- a/exporter/exporterhelper/internal/batch_sender.go
+++ b/exporter/exporterhelper/internal/batch_sender.go
@@ -202,12 +202,12 @@ func (bs *BatchSender) sendMergeBatch(ctx context.Context, req internal.Request)
 	bs.mu.Lock()
 
 	if bs.activeBatch.request != nil {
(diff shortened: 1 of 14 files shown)


bogdandrutu (approved):

bogdandrutu:
I think we need to mark this as a "breaking" change even if it is for experimental interface we still need to add a changelog. Ideally if not that hard, is it possible to just mark the method as deprecated and then remove next release?

bogdandrutu:
Created a different PR for the changelog.

PR #12017, opened 2025-01-05
[POC][DEPRECATED] exporter batcher - byte size based batching

### LINKED ISSUE #3262: [processors/batch] Size calculations inconsistent, causing unbounded batch size in bytes

sfc-gh-sili:
DEPRECATED
See for https://github.com/open-telemetry/opentelemetry-collector/pull/12091 for the updated POC

## Description

This is an POC of serialized size based batching.

Configuration is supported via an additional field to `MaxSizeConfig`.
```
type MaxSizeConfig struct {
	MaxSizeBytes int `mapstructure:"max_size_bytes"`
}
```
We will validate that at most one of the above fields are specified (TODO) and switch between item count-based batching vs. byte size-based batching accordingly.

To get the byte size of otlp protos, this PR updates `pdata/internal/cmd/pdatagen/internal/templates/message.go.tmpl` to expose an interface `Size()`. This change will apply to all `pdatagen`-generated files.

-------------------------------------------------------------------------------------------------------------------------

### Conclusions

* Byte size based batching is more expensive.
* Both merging and splitting are expensive
  * Cost of splitting is dependent on number of logs that are "extracted". The most unfortunate case is where "the incoming request has to be split, but the remaining capacity is large enough to hold most of the logs"
* As a best practice, users should make sure `maxSizeLimit` is large enough for most requests. (**Assumption 1**)
* With the **Assumption 1** , **Optimization 1** would save us from the above unfortunate case.
* "Optimization 2" can significantly improvement the performance at the cost of accuracy. It helps with the case where many requests are merged into a single batch.

### Benchmark 1
This benchmark merge splits 1000 requests x 10 logs / request. None of the incoming request would result in a split. The accumulated batch is ~1MB
```
BenchmarkSplittingBasedOnItemCountManySmallLogs-10    	     338	   3,551,946 ns/op
BenchmarkSplittingBasedOnByteSizeManySmallLogs-10     	       9	 121,470,181 ns/op
```

### Benchmark 2
The following benchmarks test the case where every request would result in a split.

**Case 1**: Merging 10 request, where each contains 10001 logs / ~ 1 MB and is slightly above the limit
```
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyAboveLimit-10    	      49	  24,918,513 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyAboveLimit-10     	       1	3,388,820,625 ns/op
```

**Case 2**: Merging 10 request, where each contains 9999 logs / ~ 1 MB and is slightly below the limit
```
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyBelowLimit-10    	      54	  22,659,051 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyBelowLimit-10     	      37	  31,614,578 ns/op
```

### Benchmark 3
This benchmark merge splits a request with 100000 logs / 9.6MB in to 10 batches. With the above mentioned assumption, this should rarely occur in practice.
```
BenchmarkSplittingBasedOnItemCountHugeLog-10    	      40	  30,575,706 ns/op
BenchmarkSplittingBasedOnByteSizeHugeLog-10     	       1	9,909,146,541 ns/op
```

### Optimization 1

Instead of splitting precisely, simply put the new request into a new batch if it goes beyond capacity.
Running Benchmark 2 again with optimization:

**Case 1**: Merging 10 request, where each contains 10001 logs / ~ 1 MB and is slightly above the limit
```
// Before optimization
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyAboveLimit-10    	      49	  24,918,513 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyAboveLimit-10     	       1	3,388,820,625 ns/op

// After optimization
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyAboveLimit-10    	      48	  24,296,155 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyAboveLimit-10     	       1	11,746,665,625 ns/op
```

Takeaway: if size limit is not large enough, this optimization might hurt the performance.

**Case 2**: Merging 10 request, where each contains 9999 logs / ~ 1 MB and is slightly below the limit
```
// Before optimization
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyBelowLimit-10    	      54	  22,659,051 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyBelowLimit-10     	      37	  31,614,578 ns/op

// After optimization
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyBelowLimit-10    	      52	  22,252,627 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyBelowLimit-10     	      43	  25,942,651 ns/op
```

### Optimization 2
Optimization 2 is done at the cost of being inaccurate. For example, with Benchmark 1, the actual byte size is 1010000 while the estimated byte size is 1011010. Delta is expected to be larger in real situation as we will have more metadata with logs.

(The original cost is not high either, so optimization 2 is not essential)

```
// Before optimization
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyAboveLimit-10    	      49	  24,918,513 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyAboveLimit-10     	       1	3,388,820,625 ns/op

// After optimization
BenchmarkSplittingBasedOnItemCountManySmallLogs-10    	     340	   3470374 ns/op
BenchmarkSplittingBasedOnByteSizeManySmallLogs-10     	     493	   2784610 ns/op
```

#### Link to tracking issue
Fixes #3262

#### Documentation

TODO: ByteSize() should return int64 instead of int

### DIFF
--- a/exporter/exporterbatcher/config.go
+++ b/exporter/exporterbatcher/config.go
@@ -31,6 +31,7 @@ type MinSizeConfig struct {
 	// sent regardless of the timeout. There is no guarantee that the batch size always greater than this value.
 	// This option requires the Request to implement RequestItemsCounter interface. Otherwise, it will be ignored.
 	MinSizeItems int `mapstructure:"min_size_items"`
+	MinSizeBytes int `mapstructure:"min_size_bytes"`
 }
 
 // MaxSizeConfig defines the configuration for the maximum number of items in a batch.
@@ -41,17 +42,31 @@ type MaxSizeConfig struct {
 	// If the batch size exceeds this value, it will be broken up into smaller batches if possible.
 	// Setting this value to zero disables the maximum size limit.
 	MaxSizeItems int `mapstructure:"max_size_items"`
+	MaxSizeBytes int `mapstructure:"max_size_bytes"`
 }
 
 func (c Config) Validate() error {
+	if c.MinSizeBytes != 0 && c.MinSizeItems != 0 || c.MinSizeBytes != 0 && c.MaxSizeItems != 0 || c.MinSizeItems != 0 && c.MaxSizeBytes != 0 {
+		return errors.New("size limit and bytes limit cannot be specified at the same time")
+	}
+
 	if c.MinSizeItems < 0 {
 		return errors.New("min_size_items must be greater than or equal to zero")
+	}
+	if c.MinSizeBytes < 0 {
+		return errors.New("min_size_bytes must be greater than or equal to zero")
 	}
 	if c.MaxSizeItems < 0 {
 		return errors.New("max_size_items must be greater than or equal to zero")
 	}
+	if c.MaxSizeBytes < 0 {
+		return errors.New("max_size_bytes must be greater than or equal to zero")
+	}
 	if c.MaxSizeItems != 0 && c.MaxSizeItems < c.MinSizeItems {
 		return errors.New("max_size_items must be greater than or equal to min_size_items")
+	}
+	if c.MaxSizeBytes != 0 && c.MaxSizeBytes < c.MinSizeBytes {
+		return errors.New("max_size_bytes must be greater than or equal to min_size_bytes")
 	}
 	if c.FlushTimeout <= 0 {
 		return errors.New("timeout must be greater than zero")
--- a/exporter/exporterhelper/internal/request.go
+++ b/exporter/exporterhelper/internal/request.go
@@ -56,6 +56,10 @@ func (r *fakeRequest) ItemsCount() int {
 	return r.items
 }
 
+func (r *fakeRequest) ByteSize() int {
+	return r.items
+}
+
 func (r *fakeRequest) MergeSplit(_ context.Context, cfg exporterbatcher.MaxSizeConfig, r2 internal.Request) ([]internal.Request, error) {
 	if r.mergeErr != nil {
 		return nil, r.mergeErr
--- a/exporter/exporterhelper/internal/retry_sender_test.go
+++ b/exporter/exporterhelper/internal/retry_sender_test.go
@@ -418,6 +418,10 @@ func (mer *mockErrorRequest) ItemsCount() int {
 	return 7
 }
 
+func (mer *mockErrorRequest) ByteSize() int {
+	return 7
+}
+
 func (mer *mockErrorRequest) MergeSplit(context.Context, exporterbatcher.MaxSizeConfig, internal.Request) ([]internal.Request, error) {
 	return nil, nil
 }
@@ -461,6 +465,10 @@ func (m *mockRequest) checkNumRequests(t *testing.T, want int) {
 }
 
 func (m *mockRequest) ItemsCount() int {
+	return m.cnt
+}
+
+func (m *mockRequest) ByteSize() int {
 	return m.cnt
 }
 
--- a/exporter/exporterhelper/logs.go
+++ b/exporter/exporterhelper/logs.go
@@ -25,8 +25,9 @@ var (
 )
 
 type logsRequest struct {
-	ld     plog.Logs
-	pusher consumer.ConsumeLogsFunc
+	ld       plog.Logs
+	pusher   consumer.ConsumeLogsFunc
+	byteSize int
 }
 
 func newLogsRequest(ld plog.Logs, pusher consumer.ConsumeLogsFunc) Request {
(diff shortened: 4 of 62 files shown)


sfc-gh-sili:
Feedback from Bogdan:

Issue #11524, opened 2024-10-23
Should ConfigValidator be in confmap?


bogdandrutu:
**Is your feature request related to a problem? Please describe.**

Currently the `ConfigValidator` is part of the `component` package, and for example in `mdatagen` where we use confmap it is a bit strange to depend on the `component.ConfigValidator` for config validation.

**Describe the solution you'd like**

Provide the configuration `Validator` as part of the `confmap` repository, and probably make it part of `Conf.Unmarshal` (configurable) to also do validation.

**Describe alternatives you've considered**

Leave where it is right now.

TylerHelmuth:
Is it correct to say from a component developer perspective it doesn't matter where it lives?

sfc-gh-bdrutu:
> Is it correct to say from a component developer perspective it doesn't matter where it lives?

Yes, but if we want to use confmap in other places like I said in `mdatagen` it is confusing.

dmitryax:
I agree `confmap` seems like a better place for it

mx-psi:
I don't have a strong opinion on this (which makes me lean slightly into "leave it as it is")

TylerHelmuth:
> probably make it part of Conf.Unmarshal (configurable) to also do validation.

Today I believe that `otelcol` is our collector's use of `ConfigValidator`. What problem would we be solving by adding something configurable to `Conf.Unmarshal`?

mx-psi:
@bogdandrutu If we were to move this, what would we do with `component.ValidateConfig`?

dmitryax:
Do we even need `component.ValidateConfig`? I can't see any value in it...

mx-psi:
@dmitryax Here are the places where this is used: https://github.com/search?q=org%3Aopen-telemetry+language%3AGo+component.ValidateConfig&type=code

There are a lot of usages that could be avoided/converted, but I am not confident that we would be able to remove all of them.

atoulme:
We discussed this issue during the stability SIG meeting of 1/6 and came up with the following suggestions:
1. Move the validation code out of component to an experimental package (prefixed with x)
2. Look at whether this validation code should be moved to stable, or use a new approach, using for example https://github.com/go-playground/validator so we don't implement our own validation framework. A new issue for this will be cut to continue the work.

TylerHelmuth:
I'll submit a PR to move validation stuff to `xcomponent`.

TylerHelmuth:
[...]

PR #12031, opened 2025-01-06
[confmap] Automatically invoke `Validate()` during `Conf.Unmarshal`

TylerHelmuth:
#### Description

Duplicate `component.ConfigValidator` into `confmap`.  Add capability for `ConfigValidator.Validate` to be invoked as part of unmarshal. Add a new `UnmarshalOption` to allow controlling if validation is invoked.

If we like this feature, we'll need to do some refactoring to how `otelcol` unmarshals if we want to take advantage of it. See https://github.com/open-telemetry/opentelemetry-collector/issues/11524#issuecomment-2574045326. After implementing it I'm still not convinced this solution is better than what we already have.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector/issues/11524

#### Testing

Added new unit tests.  Tested manually using local build.

#### Documentation

Added godoc comments.

### DIFF
--- a/confmap/confmap.go
+++ b/confmap/confmap.go
@@ -15,6 +15,7 @@ import (
 	"github.com/knadh/koanf/maps"
 	"github.com/knadh/koanf/providers/confmap"
 	"github.com/knadh/koanf/v2"
+	"go.uber.org/multierr"
 
 	encoder "go.opentelemetry.io/collector/confmap/internal/mapstructure"
 )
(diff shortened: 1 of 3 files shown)


confmap/confmap.go
```diff
@@ -71,20 +73,109 @@ func WithIgnoreUnused() UnmarshalOption {
...
+}
+
 type unmarshalOptionFunc func(*unmarshalOption)
 
 func (fn unmarshalOptionFunc) apply(set *unmarshalOption) {
 	fn(set)
 }
 
+// ConfigValidator defines an optional interface for configurations to implement to do validation.
+type ConfigValidator interface {
```
bogdandrutu, line 93: Just `Validator`.

bogdandrutu (changes requested):

TylerHelmuth:
@bogdandrutu I was able to test the custom unmarshal scenario with the otlp receiver and confirmed it works since the custom unmarshal implementation invokes `conf.Unmarshal`. 

If the custom unmarshaller does not invoked `conf.Unmarshal` it will still work since `validate` is ultimately called on the top level object that is being unmarshalled into.

I've added a tests to demonstrate these 2 scenarios.

PR #12033, opened 2025-01-07
[chore] Move exporterhelper queue code to exporterqueue

bogdandrutu:


mx-psi (commented):
Could you add a link/explanation on why we want to do this? This would increase the public API surface so I feel like we need to do some due diligence

bogdandrutu:
> This would increase the public API surface so I feel like we need to do some due diligence

Why? Everything is private types/funcs unless I missed something?

> Could you add a link/explanation on why we want to do this? 

Not sure we need to explain that is better to have the implementation of the interfaces and the interface definition close to each other.

mx-psi:
> Why? Everything is private types/funcs unless I missed something?

Ah, looks like I got confused, sorry! LGTM then

mx-psi (approved):

bogdandrutu:
@mx-psi thank you

Issue #11814, opened 2024-12-06
Loggers for shared components should not report signal type.


djaglowski:
**Describe the bug**

For each instantiated component, the collector [assigns a set of attributes to a logger](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/components/loggers.go) which are sufficient to identify the individual instance of the component.

When instantiating "shared" components (those which consolidate instances using e.g. a singleton pattern), this logger may contain a "signal" attribute that indicates what would have been a separate instance of the component, but is in fact representative of the shared instance. The effect is that internally generated logs are inaccurately attributed to a "metrics instance", or a "logs instance", or a "traces instance" of the component, when in reality logs were generated by an instance shared by all signals.

**Steps to reproduce**

Configure a single otlp receiver for metrics, logs, and traces. The collector will _think_ that it is instantiating three separate instances. Supposing it instantiates the metrics instance first, this logger passed in will have the metrics signal as an attribute. Then, when it tries to instantiate the traces and logs instances, the otlp receiver factory will just return the same instance, which still only has the "metrics logger". For the lifecycle of the instance, it will always attribute all logs to the metrics instance. 

Additionally, the order of instantiation is essentially non-deterministic, meaning that minor changes to the configuration may result in attribution to a different signal.

**What did you expect to see?**

The otlp receiver (and all other "shared" components) should use a logger which does not contain a signal attribute.

**Potential solutions**

While this issue is specifically focused on loggers, we can more broadly look at this as a question of how to ensure component instances are always using the correct set of attributes for all internally generated telemetry. In my opinion, the best solution is one that confidently hands the correct attribute set to the instance, rather than relying on component implementations to correctly follow documented guidelines.

Therefore, I suggest that we should have a simple mechanism for indicating whether or not component implementations are using a singleton pattern. This would act much like `consumer.Capabilities.MutatesData`, where component authors must set the flag correctly, but otherwise do not need to worry about implementation details.

Note: In theory, there could be other patterns for "shared" components, (e.g. combine logs and metrics instances, but not traces) but I'm unaware of any examples or reasons to do this. If we can assume the only valid patterns are "singleton" and "default", then a single flag is sufficient. If this doesn't seem like a safe assumption, then an enum with only two possible values would be appropriate.

[...]

PR #12057, opened 2025-01-08
Singleton flags and attributes

djaglowski:
This PR is an extension of #11344 that addresses #11814.

Major changes:
- Add  a `Metadata()` method to `receiver.Factory`, `exporter.Factory` and `connector.Factory` (`processor.Factory` excluded intentionally as described in #11814). These `Metadata()` methods return `Metadata` structs, which act similarly to `consumer.Capabilities` in that they are intended as a set of information which describes the components which the factory produces. 
- Remodels each component's as a proper `go.opentelemetry.io/otel/attribute.Set`. This set is used as the ID in the graph, but also as the set of attributes that creates the logger. In the future, the set will be automatically used to describe other aspects of the collector's own telemetry.
- The logger fields are changed to match those described in the [component telemetry RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md#attributes).

Minor changes:
- Unknown component types are detected earlier in the startup process. (When the graph is being assembled, rather than when the components of the graph are being instantiated.) This is necessary so that we can obtain the `Metadata` from the appropriate factory. There is a minor user facing change in that the error message is not wrapped in a "could not build" prefix. I moved the processors to this pattern as well in order to match the other components.


djaglowski:
If desired, I believe I can split this PR into several smaller PRs. However, I wanted to put this version up in order to show the overall picture.

djaglowski:
I discovered when adding tests that singleton connectors will introduce broader changes into the way we build the graph, since we're wrapping each instances as a specific type of consumer. I've not been able to get it working yet but either way, the additional complexity seems not worth for now. I believe there are some reasons why singleton connectors would be useful but I'm going to propose for now that singleton instances are supported only for receivers and exporters. In the future we can try to add singleton connectors.

exporter/exporter.go
```diff
@@ -72,9 +72,17 @@ type Factory interface {
 	// LogsStability gets the stability level of the Logs exporter.
 	LogsStability() component.StabilityLevel
 
+	// Metadata returns the metadata describing the receiver.
+	Metadata() Metadata
```
mx-psi, line 76: I wonder if this should have a different name. The metadata name suggests to me that the properties listed are properties of the component itself, but this is not the case: the service is the one that uses this information to decide how to build the graph. I see this more as a 'hint' than 'metadata'?

mx-psi (commented):
The approach makes sense to me, just one question about naming

exporter/exporter.go
djaglowski: I was thinking that since it's a method on the factory, it describes the type of component that the factory produces. Although the immediate usage is in the service, I assume we might use it elsewhere, e.g. documentation or cli output.

That said, I don't feel strongly about the name. Hints makes sense to me too, just seems a little more obscure.

mx-psi: Re-reading the PR what I said does not make much sense: the service only handles logging, which it should, and building the component as a singleton is left to the component internals itself.

I don't feel strongly about the name, if you feel like `Metadata` is okay I am okay with that.

mx-psi (approved):
Looks good to me. 

One thing I don't like about `Capabilities` is that a component developer can accidentally 'lie' about their component, which can also happen here. I don't see a great way to do this here without introducing a lot more API, so this seems reasonable enough

djaglowski:
> Do you have ideas for making the new attributes available to components for use in spans and metrics? (Adding the `attribute.Set` in the `TelemetrySettings` for example?)

It is the next thing I intend to look into. Either as you suggested, or maybe it's possible to save them into the Meter/Tracer providers in some way?

.chloggen/singleton-flags-and-attributes-exporter.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
mx-psi, line 4: I don't think this is a breaking change since the interface is sealed, right?

mx-psi: > Open to other opinions but I think the error message does not warrant a breaking change. If it were in a stable package, maybe it would warrant it, but since it's not, I'll remove the breaking notice

This is covered in the  [VERSIONING.md document](https://github.com/open-telemetry/opentelemetry-collector/blob/main/VERSIONING.md#general-go-api-considerations) (emphasis mine):

> Unless otherwise specified in the documentation, the following may change in any way between minor versions:
> - String representation. **The** String or **Error method** of any struct is intended to be human-readable and **may change its output in any way**.

mx-psi:
I will merge this on Monday unless somebody comments, cc @open-telemetry/collector-approvers

songy23 (approved):
LGTM

mx-psi:
@dmitryax will take a look at this before we merge it

sfc-gh-bdrutu:
I have a feedback we discuss in the otel collector stability meeting (which may be already address, but want to make sure is documented): We should be able to have internal telemetry (logs/traces/metrics) annotated with the signal type that produces them, so if I fail to parse a message I should know that is the "metrics" pipeline that failed.

exporter/exporter.go
```diff
@@ -176,6 +189,13 @@ func WithLogs(createLogs CreateLogsFunc, sl component.StabilityLevel) FactoryOpt
 	})
 }
 
+// AsSingletonInstance indicates that the factory always returns the same instance of the component.
+func AsSingletonInstance() FactoryOption {
```
bogdandrutu, line 193: For option we use "With" prefix, please be consistent.

djaglowski: I was on the fence about it. `As` seemed better grammatically but I don't really care. I will change to `With` for consistency.

djaglowski:
> I think this will come with the downside that the signal specific part of the component is not handled correctly (or has to be handled by manually annotating the signal tag to the telemetry).

I agree this is a downside, but your proposal leaves the same problem in place. It reduces the scope of the problem slightly at the cost of a lot of refactoring, but fundamentally the same problem remains. Any shared sub-component may have good reason to report telemetry that pertains to one specific signal, so how does the shared sub-component manage this? This still leaves us with manually annotating the signal tag to the telemetry.

exporter/exporter.go
djaglowski: It occurred to me that "singleton" is a poor choice of word here too, since we're not actually intending for the component to create just one instance globally. It is shared by all components with the same config. Therefore, I think `WithSharedInstance` is better.

dmitryax:
What if we just another field to `component.TelemetrySettings`, let's call it `SharedLogger` (same as `Logger` but without the signal attribute), which can be used by the shared parts of the components while `Logger` will be kept to signal specific code parts? `memory_limiter` processor needs that as well.

It's hard to be convinced that this new public API (Metadata) is required. We just pass a boolean flag back to the factory. It's supposed to be connected to the fact that `sharedComponent` is used, but there is nothing to support that connection.

djaglowski:
I'm closing this as I no longer believe that adding notions of shared/singleton parts is a realistic solution to all the novel cases. New proposal here: https://github.com/open-telemetry/opentelemetry-collector/issues/12217

PR #12059, opened 2025-01-09
Support unix socket for OTLP with gRPC export of internal telemetry

### LINKED ISSUE #11941: Using Unix socket for internal telemetry metrics
**Describe the bug**
When OTel Collector is run We use Unix socket

**Steps to reproduce**
In internal telemetry config, try to use `unix:otlp.sock` instead of URI in `config.service.telemetry.metrics.readers[periodic].exporters.otlp.endpoint`

Unix socket recognized properly as internal telemetry endpoint, internal telemetry metrics are being sent to Unix socket in the same way app container in the same pod does.

**What did you see instead?**
Pod fails to start, sidecar OTel Collector container logs error:
```
2024/12/17 09:29:01 collector server run finished with error: failed to create metric provider: parse "http://unix:otlp.sock": invalid port ":otlp.sock" after host
```

**What version did you use?**
We run OTel Collectors in our k8s clusters, installed as Helm charts.
Chart version: `opentelemetry-operator:0.74.3`
OTel image version override: `0.115.1` (tried on `0.114.0` too)

**What config did you use?**
```
  config:
    service:
      telemetry:
        metrics:
          level: detailed
          readers:
            - periodic:
                interval: 10000
                exporter:
                  otlp:
                    protocol: grpc/protobuf
                    endpoint: unix:otlp.sock # <- only network URI is being accepted due to address:port validation
```

**Environment**
 All workloads are run in Azure k8s/AKS v1.30

**Additional context**
It looks like the error hints on what's happening:
```
2024/12/17 09:29:01 collector server run finished with error: failed to create metric provider: parse "http://unix:otlp.sock": invalid port ":otlp.sock" after host
```
Instead of accepting the connection string the way it is, the validation adds `http://` to the beginning of the string, and considers everything after colon as the port, like `:<port_expected>`.

OTel Collector [OTLP exporter documentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlpexporter/README.md#getting-started) refers to the [gRPC name resolution specification](https://github.com/grpc/grpc/blob/master/doc/naming.md) which should allow unix sockets in `unix:socket` syntax, so this extra connection string validation looks excessive and actually breaks the usage of unix sockets.

dmathieu:
#### Description

This allows specifying a unix socket for the internal telemetry OTLP endpoint, when using gRPC.

#### Link to tracking issue
Fixes #11941

### DIFF
--- a/.chloggen/intenal-telemetry-unix.yaml
+++ b/.chloggen/intenal-telemetry-unix.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: support unix socket as internal telemetry endpoint for the gRPC OTLP exporter
+
+# One or more tracking issues or pull requests related to the change
+issues: [12059]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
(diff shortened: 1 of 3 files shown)


bogdandrutu (changes requested):
Thanks for this PR @dmathieu 

I will let @codeboten jump onto this since we are planning to delete this otelinit and use the contrib config initialization. 

@codeboten feel free to dismiss my change request if you are ok with this, I added here only to ensure you see this.

codeboten:
Thanks @dmathieu, @bogdandrutu is correct, the code in otelinit will be removed completely in favour of the config package once https://github.com/open-telemetry/opentelemetry-collector/pull/11611 lands. It's currently waiting on the next release of otel-go and otel-go-contrib to avoid bumping the minimum go version.

It would be great to add a test to the config package to validate that unix sockets are supported

dmathieu:
Closing in favor of https://github.com/open-telemetry/opentelemetry-go-contrib/pull/6636

PR #12074, opened 2025-01-11
Add capability for memory and persistent queue to block when add items

bogdandrutu:
This functionality is not yet exposed to the user because there is no config available. One option to consider is to add a "blocking" bool in the queue config.

#### Link to tracking issue

### DIFF
--- a/exporter/exporterqueue/persistent_queue.go
+++ b/exporter/exporterqueue/persistent_queue.go
@@ -11,7 +11,6 @@ import (
 	"strconv"
 	"sync"
 
-	"go.uber.org/multierr"
 	"go.uber.org/zap"
 
 	"go.opentelemetry.io/collector/component"
@@ -42,6 +41,7 @@ var (
 type persistentQueueSettings[T any] struct {
 	sizer       sizer[T]
 	capacity    int64
+	blocking    bool
 	signal      pipeline.Signal
 	storageID   component.ID
 	marshaler   Marshaler[T]
@@ -81,7 +81,8 @@ type persistentQueue[T any] struct {
 
 	// mu guards everything declared below.
 	mu                       sync.Mutex
-	hasElements              *sync.Cond
+	hasMoreElements          *sync.Cond
+	hasMoreSpace             *cond
 	readIndex                uint64
 	writeIndex               uint64
 	currentlyDispatchedItems []uint64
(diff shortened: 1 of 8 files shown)


exporter/exporterqueue/persistent_queue.go
```diff
@@ -269,32 +276,38 @@ func (pq *persistentQueue[T]) putInternal(ctx context.Context, req T) error {
...
-			// because we don't flush the current queue size on the disk on every read/write.
-			// In that case we need to make sure it doesn't go below 0.
-			if pq.queueSize < 0 {
+		// Read until either a successful retrieved element or no more elements in the storage.
+		for pq.readIndex != pq.writeIndex {
+			index, req, consumed := pq.getNextItem(ctx)
+			// Ensure the used size and the channel size are in sync.
+			if pq.readIndex == pq.writeIndex {
 				pq.queueSize = 0
+				pq.hasMoreSpace.Signal()
```
dmitryax, line 292: Why do we need this line here as well? It'll be called twice when the last element consumed from the queue

dmitryax: Nevermind. I see why. And calling twice it's not a problem

exporter/exporterqueue/sized_queue.go
```diff
@@ -104,17 +114,20 @@ func (sq *sizedQueue[T]) pop() (context.Context, T, bool) {
...
 
 		if sq.stopped {
 			var el T
 			return context.Background(), el, false
 		}
 
-		sq.hasElements.Wait()
+		// TODO: Change the Queue interface to return an error to allow distinguish between shutdown and context canceled.
+		// Ok to ignore the error, since the context.Background() will never be done.
+		_ = sq.hasMoreElements.Wait(context.Background())
```
dmitryax, line 130: Do we need to switch `hasMoreElements` to the custom cond to resolve the TODO later?

bogdandrutu: We need only to have "Read" return an error instead of bool, and have options like "ErrQueueShutdown" and "context.ErrDeadlineExceed" .

bogdandrutu: If your suggestion was to still use the `sync.Cond` and change that when also solving the TODO, I think it is better. I did that.

dmitryax (approved):

PR #12091, opened 2025-01-15
[POC][exporter] exporter batcher - byte size based batching

### LINKED ISSUE #3262: [processors/batch] Size calculations inconsistent, causing unbounded batch size in bytes

sfc-gh-sili:
## Description

This is an POC of serialized size based batching (version 2)

Configuration is supported via an additional field to `MaxSizeConfig`.
```
type MaxSizeConfig struct {
	MaxSizeBytes int `mapstructure:"max_size_bytes"`
}
```
We will validate that at most one of the above fields are specified (TODO) and switch between item count-based batching vs. byte size-based batching accordingly.

To get the byte size of otlp protos, this PR updates `pdata/internal/cmd/pdatagen/internal/templates/message.go.tmpl` to expose an interface `Size()`. This change will apply to all `pdatagen`-generated files.

-------------------------------------------------------------------------------------------------------------------------

## Getting serialized byte size

Getting serialized byte size is an expensive operation that gets repeatedly invoked during bytes based merge-splitting. The POC shows two optimizations that improves by 1. reducing size calculation times 2. reducing size calculation cost

### Caching size in `Request`

We repeated call `request.ByteSize()` while merg-splitting. By caching previous result in `request.ByteSize()`, we were able to improve the performance especially for the case where many requests are merged into a single batch.

```
BenchmarkSplittingBasedOnItemCountManySmallLogs-10                 	     324	   3729469 ns/op
/*Before*/ BenchmarkSplittingBasedOnByteSizeManySmallLogs-10         	       9	 120696384 ns/op
/*After*/ BenchmarkSplittingBasedOnByteSizeManySmallLogs-10                  358	   2906934 ns/op
```

### Proto size caculation

While filling the batch, we repeated evaluate size of new items and the remaining space, and both calculations are expensive. This PR shows an optimization that calculates remaining space iteratively using new item size:

```
remaning -= (1 + sizeof(newItem) + sov(sizeof(newItem)))
```

where

```
func sov(x uint64) (n int) {
	return (math_bits.Len64(x|1) + 6) / 7
}
```

This helped significantly improve the performance. For benchmark result without this optimization, see https://github.com/open-telemetry/opentelemetry-collector/pull/12017. 

### Benchmark 1 - 1000 small requests merge into one batch
This benchmark merges 1000 requests x 10 logs / request. The accumulated batch is ~1MB
```
BenchmarkSplittingBasedOnItemCountManySmallLogs-10                 	     328	   3556645 ns/op
BenchmarkSplittingBasedOnByteSizeManySmallLogs-10                  	     447	   2689311 ns/op
```

### Benchmark 2 - Every incoming request causes a split

**Case 1**: Merging 10 request, where each contains 10001 logs / ~ 1 MB and is slightly above the limit
```
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyAboveLimit-10    	      45	  24271094 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyAboveLimit-10     	      31	  35001137 ns/op
```

**Case 2**: Merging 10 request, where each contains 9999 logs / ~ 1 MB and is slightly below the limit
```
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyBelowLimit-10    	      52	  21750268 ns/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyBelowLimit-10     	      38	  28904701 ns/op
```

### Benchmark 3 - A huge request splits into 10 batches
This benchmark merge splits a request with 100000 logs / ~10MB in to 10 batches.
```
BenchmarkSplittingBasedOnItemCountHugeLog-10                       	      43	  29004844 ns/op
BenchmarkSplittingBasedOnByteSizeHugeLog-10                        	      15	  66705669 ns/op
```

#### Link to tracking issue
Fixes #3262

#### Documentation

TODO: ByteSize() should return int64 instead of int

### DIFF
--- a/exporter/exporterbatcher/config.go
+++ b/exporter/exporterbatcher/config.go
@@ -31,6 +31,7 @@ type MinSizeConfig struct {
 	// sent regardless of the timeout. There is no guarantee that the batch size always greater than this value.
 	// This option requires the Request to implement RequestItemsCounter interface. Otherwise, it will be ignored.
 	MinSizeItems int `mapstructure:"min_size_items"`
+	MinSizeBytes int `mapstructure:"min_size_bytes"`
 }
 
 // MaxSizeConfig defines the configuration for the maximum number of items in a batch.
@@ -41,17 +42,31 @@ type MaxSizeConfig struct {
 	// If the batch size exceeds this value, it will be broken up into smaller batches if possible.
 	// Setting this value to zero disables the maximum size limit.
 	MaxSizeItems int `mapstructure:"max_size_items"`
+	MaxSizeBytes int `mapstructure:"max_size_bytes"`
 }
 
 func (c Config) Validate() error {
+	if c.MinSizeBytes != 0 && c.MinSizeItems != 0 || c.MinSizeBytes != 0 && c.MaxSizeItems != 0 || c.MinSizeItems != 0 && c.MaxSizeBytes != 0 {
+		return errors.New("size limit and bytes limit cannot be specified at the same time")
+	}
+
 	if c.MinSizeItems < 0 {
 		return errors.New("min_size_items must be greater than or equal to zero")
+	}
+	if c.MinSizeBytes < 0 {
+		return errors.New("min_size_bytes must be greater than or equal to zero")
 	}
 	if c.MaxSizeItems < 0 {
 		return errors.New("max_size_items must be greater than or equal to zero")
 	}
+	if c.MaxSizeBytes < 0 {
+		return errors.New("max_size_bytes must be greater than or equal to zero")
+	}
 	if c.MaxSizeItems != 0 && c.MaxSizeItems < c.MinSizeItems {
 		return errors.New("max_size_items must be greater than or equal to min_size_items")
+	}
+	if c.MaxSizeBytes != 0 && c.MaxSizeBytes < c.MinSizeBytes {
+		return errors.New("max_size_bytes must be greater than or equal to min_size_bytes")
 	}
 	if c.FlushTimeout <= 0 {
 		return errors.New("timeout must be greater than zero")
--- a/exporter/exporterhelper/internal/retry_sender_test.go
+++ b/exporter/exporterhelper/internal/retry_sender_test.go
@@ -418,6 +418,10 @@ func (mer *mockErrorRequest) ItemsCount() int {
 	return 7
 }
 
+func (mer *mockErrorRequest) ByteSize() int {
+	return 7
+}
+
 func (mer *mockErrorRequest) MergeSplit(context.Context, exporterbatcher.MaxSizeConfig, internal.Request) ([]internal.Request, error) {
 	return nil, nil
 }
@@ -461,6 +465,10 @@ func (m *mockRequest) checkNumRequests(t *testing.T, want int) {
 }
 
 func (m *mockRequest) ItemsCount() int {
+	return m.cnt
+}
+
+func (m *mockRequest) ByteSize() int {
 	return m.cnt
 }
 
--- a/exporter/exporterhelper/logs.go
+++ b/exporter/exporterhelper/logs.go
@@ -25,8 +25,9 @@ var (
 )
 
 type logsRequest struct {
-	ld     plog.Logs
-	pusher consumer.ConsumeLogsFunc
+	ld             plog.Logs
+	pusher         consumer.ConsumeLogsFunc
+	cachedByteSize int
 }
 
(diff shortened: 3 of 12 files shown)


sfc-gh-sili:
TODO: Try if we can get away with not exposing that many size functions. Ideally we can do everything from one API.

krin-wang:
@sfc-gh-sili  It looks pretty good. Any further plan to merge it into main branch?

PR #12097, opened 2025-01-15
[confmap] - new feature flag for append merging strategy

VihasMakwana:
Koanf's default merging strategy currently overrides static values such as slices, numbers, and strings. However, lists of components should be treated as a special case. This pull request introduces a new command line option to allow for merging lists instead of discarding the existing ones.

With this new merging strategy:
- All the lists are merged rather than replaced.
- The merging logic is name-aware, meaning that if components with the same name appear in both lists, they will only appear once in the final merged list.

#### Link to tracking issue
Related issues:
- https://github.com/open-telemetry/opentelemetry-collector/issues/8754
- https://github.com/open-telemetry/opentelemetry-collector/issues/8394
- https://github.com/open-telemetry/opentelemetry-collector/issues/10370

#### Testing
- Added

#### Documentation
- Added under readme.

##### Example
Consider the following configs,

```yaml
# main.yaml
receivers:
  otlp/in:
processors:
  batch:
exporters:
  otlp/out:
extensions:
  file_storage:

service:
  pipelines:
    traces:
      receivers: [ otlp/in ]
      processors: [ batch ]
      exporters: [ otlp/out ]
  extensions: [ file_storage ]
```

```yaml
# extra_extension.yaml
extensions:
  healthcheckv2:

service:
  extensions: [ healthcheckv2 ]
```

If you run the collector with following command,
```
otelcol --config=main.yaml --config=extra_extension.yaml --feature-gates=-confmap.enableMergeAppendOption
```
then the final configuration after config resolution will look like following:

```yaml
# main.yaml
receivers:
  otlp/in:
processors:
  batch:
exporters:
  otlp/out:
extensions:
  file_storage:
  healthcheckv2:

service:
  pipelines:
    traces:
      receivers: [ otlp/in ]
      processors: [ batch ]
      exporters: [ otlp/out ]
  extensions: [ file_storage, healthcheckv2 ]
```

For backward compatibly, the default behaviour is **not** to merge lists. Users who want to explicitly merge lists can enable the command line option.

Note: I’d appreciate your feedback on this 🙏


confmap/confmap.go
```diff
@@ -17,8 +17,16 @@ import (
 	"github.com/knadh/koanf/v2"
 
 	encoder "go.opentelemetry.io/collector/confmap/internal/mapstructure"
+	"go.opentelemetry.io/collector/featuregate"
 )
 
+const MergeComponentsAppendID = "confmap.MergeComponentsAppend"
+
+var MergeComponentsAppend = featuregate.GlobalRegistry().MustRegister(MergeComponentsAppendID,
```
Aneurysm9, line 25: I'm not sure that a feature gate is appropriate for controlling this behavior.  Feature gates are [intended](https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate#collector-feature-gates) for experimental or transitional behavior changes.  This seems to be neither as we would need to allow the user to choose which behavior they prefer in perpetuity.  Changing this gate to the beta stage would be a staggeringly impactful breaking change and leaving it at alpha forever indicates it should be an option elsewhere and not a feature gate.

Aneurysm9 (changes requested):

confmap/confmap.go
evan-bradley: If possible, I'd like a command line flag that looks something like this:

```
--merge-lists=service::* --merge-lists=processor::transform::log_statements
```

Ideally we would allow the user to specify paths they want merged using some kind of glob-like syntax. We could then also allow users to merge everything pretty easily:

```
--merge-lists=*
```

Something like `--merge-lists=all` or a similar alias may be helpful for this case as well.

VihasMakwana: I see. Instead of glob-like syntax, would it make sense to assume that user wants to merge all lists under given config path? For `--merge-lists=service::extensions --merge-lists=processor::transform::log_statements` would only merge `extensions` and `log_statements` lists. 
But `--merge-lists=service --merge-lists=processor::transform::log_statements` would merge all lists under `servive` and only `log_statements`. 

----

Ofcourse, cases such as `--merge-lists=processor::**::log_statements`, `--merge-lists=service::pipeline::**::receivers` etc. are missing in above assumption, but I would like to keep things simple in this PR and work on glob patter in a follow up because we'll need to do some extra computation to flatten/unflatten the config and that can complicate the PR.
Let me know what do you think?

evan-bradley: That works for me. I would just like a way for this to be configurable while also allowing users to specify that everything should be merged.

VihasMakwana:
@Aneurysm9 @evan-bradley @mx-psi can you take another look?

i've removed feature flag and added command line argument to control the paths user wants merged.

I'll work on the documentation once we've established a common ground.

.chloggen/new-merge-strategy.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confmap
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce a new feature gate to allow for merging lists instead of discarding the existing ones.
```
dehaansa, line 10: Note needs to be corrected after change from feature gate

confmap/merge.go
```diff
@@ -0,0 +1,142 @@
...
+
+		// Note: path is a delimited by KeyDelimiter. We will call "Unflatten" later to get the final config
+		src[path] = mergeSlice(srcVal, destVal)
+	case reflect.Map:
+		// both of them are maps. Recursively call the mergeMaps
+		mergeMaps(sVal.(map[string]any), dVal.(map[string]any))
+	}
+}
+
+func mergeMaps(src, dest map[string]any) {
```
dehaansa, line 95: If I'm reading this correctly the "src" value is the one that is overwritten and updated, and the "dest" one is used to merge/used to provide value if the key does not exist in "src".

To me these names feel backwards, is there an existing place where we use the pattern of the names source & destination in this way?

VihasMakwana: This was actually confusing to me as well. It took a while to digest this.

> If I'm reading this correctly the "src" value is the one that is overwritten and updated, and the "dest" one is used to merge/used to provide value if the key does not exist in "src".

It's the other way. "src" is the new config and "dest" is old config. In other words, "dest" is the one that is overwritten and "src" is new one and it will override "dest" keys. 
`dest`: destination config i.e. where existing config is stored
`src`: one is used to merge/used to provide value if the key does not exist in "dest"

> To me these names feel backwards, is there an existing place where we use the pattern of the names source & destination in this way?

I feel it's backwards too but I named it this way to be consistent with [koanf's mergeFunc signature](https://github.com/knadh/koanf/blob/4a21f9b42785bb75e40cd9eced14dd5518cf3d96/options.go#L25-L33).

dehaansa: Thanks for answering, if it's consistent with koanf's implementation that makes sense. 👍

mx-psi (commented):
Thanks for taking this Vihas! I have two comments:
1. I think this is the kind of thing that is hard to get right on the first try, so I would want to do this under some sort of experimental flag.
2. I would like to do this if possible with little to no modification of stable confmap APIs. confmap is 1.0, it is a *very* important piece of the Collector, and if we get it wrong it would be quite difficult to reverse course.

VihasMakwana:
@mx-psi Thanks for your comments!
1. Makes sesne. I will move it under experimental flag.
2. I understand. I will again take a fresh look and make sure we tick all the boxes.

VihasMakwana:
@mx-psi I have moved this under a feature flag and little modification to existing API (only `Resolver.Resolve(...)` function). Can you please take another look?

otelcol/flags.go
```diff
@@ -49,6 +63,10 @@ func flags(reg *featuregate.Registry) *flag.FlagSet {
 			return nil
 		})
 
+	mergePaths := new(mergePathsValue)
+
+	flagSet.Var(mergePaths, mergePathsAppendFlag, "Specify config paths you want merged")
```
evan-bradley, line 68:
```suggestion
	flagSet.Var(mergePaths, mergePathsAppendFlag, "Specify paths in the config where lists should be merged instead of overwritten. `*` specifies that all paths with lists should be merged.")
```

confmap/confmap.go
```diff
@@ -169,6 +169,13 @@ func (l *Conf) Merge(in *Conf) error {
 	return l.k.Merge(in.k)
 }
 
+// MergeAppend merges the input given configuration into the existing config.
+// Note that the given map may be modified.
+// Note: MergeAppend merges the lists in a name-aware fashion.
```
evan-bradley, line 174: Could you expand a bit on "name-aware" here just to make it clearer?

confmap/merge.go
```diff
@@ -0,0 +1,122 @@
...
+func mergeSlice(src, dest reflect.Value) any {
+	if src.Type() != dest.Type() {
+		return src
+	}
+	slice := reflect.MakeSlice(src.Type(), 0, src.Cap()+dest.Cap())
+	for i := 0; i < dest.Len(); i++ {
+		slice = reflect.Append(slice, dest.Index(i))
+	}
+
+OUTER2:
```
evan-bradley, line 112: I just want to check since using go-to style statements isn't present anywhere else in this repo and is relatively uncommon in contrib, how much more complex does this code become if we avoid this?

confmap/resolver.go
```diff
@@ -55,6 +65,11 @@ type ResolverSettings struct {
 	// ConverterSettings contains settings that will be passed to Converter
 	// factories when instantiating Converters.
 	ConverterSettings ConverterSettings
+
+	// MergePaths contains the paths specified by the user.
+	// This paths will be used while merging the configs and all the lists under the
```
evan-bradley, line 70:
```suggestion
	// These paths will be used while merging the configs and all the lists under the
```

confmap/confmap.go
```diff
@@ -169,6 +169,13 @@ func (l *Conf) Merge(in *Conf) error {
...
+func (l *Conf) MergeAppend(in *Conf, paths []string) error {
```
evan-bradley, line 175:
```suggestion
func (l *Conf) mergeAppend(in *Conf, paths []string) error {
```
Could we unexport this for now? Since we only call this method within this package, it would be nice not to offer it externally until we're positive we want this to be part of the public API. confmap is a stable module, so once something is made public, we can't break it if we find a better solution later.

confmap/resolver.go
```diff
@@ -55,6 +65,11 @@ type ResolverSettings struct {
...
+	// specified paths will be merged rather than overridden.
+	MergePaths []string
```
evan-bradley, line 72: I suspect there is not, but if there is any way to remove this from the public API, that would be preferable for now. As @mx-psi called out, since confmap is stable, the ideal situation is that we don't add anything public-facing outside the feature gate.

otelcol/flags.go
```diff
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	configFlag = "config"
+	configFlag           = "config"
+	mergePathsAppendFlag = "merge-paths-append"
```
evan-bradley, line 16: Do we think we will want a prepending mode or other merging behavior in the future? It would be nice to just call this `merge-paths` and simply state that the behavior is to append new list items to existing lists, if we feel confident this will cover the vast majority of use-cases.

We could also potentially have the API accept custom merging functions when we're ready to make it public, which could allow things like prepending or other custom merge logic.

evan-bradley (commented):
Thanks for your continued efforts @VihasMakwana! The functionality mostly looks good to me, my comments are largely around how we expose it.

confmap/confmap.go
VihasMakwana: Sure. That makes sense.

otelcol/flags.go
VihasMakwana: > It would be nice to just call this merge-paths and simply state that the behavior is to append new list items to existing lists, if we feel confident this will cover the vast majority of use-cases.

For now, I think it makes sense to call it just `merge-paths`. We don't plan to have any "prepend" option.

confmap/resolver.go
```diff
@@ -10,10 +10,19 @@ import (
...
 	"strings"
 
+	"go.opentelemetry.io/collector/featuregate"
+
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
 
+var EnableMergeAppendOption = featuregate.GlobalRegistry().MustRegister(
+	"exporter.enableMergeAppendOption",
```
mx-psi, line 20: I would namespace it under confmap
```suggestion
	"confmap.enableMergeAppendOption",
```

confmap/resolver.go
```diff
@@ -10,10 +10,19 @@ import (
 	"regexp"
 	"strings"
 
+	"go.opentelemetry.io/collector/featuregate"
+
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
 
+var EnableMergeAppendOption = featuregate.GlobalRegistry().MustRegister(
```
mx-psi, line 19: We can make this symbol internal, no need to put it here

confmap/resolver.go
mx-psi: I wonder if we can make this work in a simpler way. Do we expect users to want to do this in a fine-grained fashion, specifying individual paths? If we make it all-or-none, at least to start, we don't need to have an option here

confmap/confmap.go
mx-psi, line 174: We can make this method private for now, no uses outside of confmap

mx-psi (changes requested):

confmap/resolver.go
evan-bradley: The idea for allowing users to specify individual paths is that we could allow users to do things like append extensions in the `service::extensions` section of the config while allowing other sections to be overridden. That said, I think we could take an iterative approach and have it be all-or-nothing for now.

mx-psi: If this is usable/useful with an all-or-nothing approach, I would rather focus on testing this first like that before adding any public API. What do you think @VihasMakwana ?

VihasMakwana: I think all-or-nothing makes sense for now. We can followup on an issue for more fine grained control.
Just to be on same page, when you say all-or-none, you mean all the lists in a given config, right? Or just the lists under `service::*`?

VihasMakwana: I'm confirming because the original issue was only concerned about the components lists (under `service::`)

mx-psi: All the lists in a given config, yes. We may even have a way to configure this in the YAML itself using YAML tags (but that's something to do in a different PR)

VihasMakwana: Thanks for confirming! I'll revert the fine grained control for now and store it safely in another branch (we might need it 😆). 
For now, I'll update the function to merge all the lists in a config.

VihasMakwana:
@mx-psi @evan-bradley I've removed fine-grained control for now. I'll open up a follow-up issue for further improvements. Can you please re-review this? Thanks you!

.chloggen/new-merge-strategy.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce a new command line option to allow for merging lists instead of discarding the existing ones.
+
+# One or more tracking issues or pull requests related to the change
+issues: [8394, 8754, 10370]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
mx-psi, line 18: Can we add details here for how to use this?

mx-psi (commented):
Implementation looks good to me, I think we just need some more docs. Can you do the following:
1. Add information on the changelog note on how to use it.
2. Add a section to https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap#resolver that explains how this works with an example CLI invocation

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
...
+extensions:
+  healthcheckv2:
+
+service:
+  extensions: [ healthcheckv2 ]
+```
+
+If you run the collector with following command,
+```
+otelcol --config=main.yaml --config=extra_extension.yaml --feature-gates=-confmap.enableMergeAppendOption
```
mx-psi, line 126:
```suggestion
otelcol --config=main.yaml --config=extra_extension.yaml --feature-gates=confmap.enableMergeAppendOption
```

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
 4. For each "Converter", call "Convert" for the "result".
 5. Return the "result", aka effective, configuration.
 
+#### New Merging strategy
+
+When merging configs from different sources, we currently override static values such as slices, numbers, and strings.
```
mx-psi, line 88:
```suggestion
When merging configs from different sources, lists are overridden instead of being merged.
```

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
 4. For each "Converter", call "Convert" for the "result".
 5. Return the "result", aka effective, configuration.
 
+#### New Merging strategy
```
mx-psi, line 86:
```suggestion
#### (Experimental) Append merging strategy for lists
```

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
...
+
+When merging configs from different sources, we currently override static values such as slices, numbers, and strings.
+However, you can combine lists instead of discarding the existing ones by enabling `confmap.enableMergeAppendOption` feature flag.
```
mx-psi, line 89:
```suggestion
You can opt-in to experimentally combine slices instead of discarding the existing ones by enabling `confmap.enableMergeAppendOption` feature flag. Lists are appended in the order in which they appear in their configuration sources.

This will **not** become the default in the future, we are still deciding how this should be configured and want your feedback on [this issue](https://github.com/open-telemetry/opentelemetry-collector/issues/8754).
```

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
...
+```yaml
+# extra_extension.yaml
+extensions:
+  healthcheckv2:
+
+service:
+  extensions: [ healthcheckv2 ]
+```
+
+If you run the collector with following command,
```
mx-psi, line 124:
```suggestion
If you run the Collector with following command,
```

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
...
+service:
+  pipelines:
+    traces:
+      receivers: [ otlp/in ]
+      processors: [ batch ]
+      exporters: [ otlp/out ]
+  extensions: [ file_storage, healthcheckv2 ]
+```
+
+Notice that the `service::extensions` list is a combination of both configurations. By default, it is `service::extensions: [healthcheckv2]`.
```
mx-psi, line 151:
```suggestion
Notice that the `service::extensions` list is a combination of both configurations. By default, the value of the last configuration source passed, `extra_extension`, would be used, so the extensions list would be: `service::extensions: [healthcheckv2]`.
```

confmap/README.md
```diff
@@ -83,6 +83,74 @@ The `Resolve` method proceeds in the following steps:
 4. For each "Converter", call "Convert" for the "result".
 5. Return the "result", aka effective, configuration.
 
+#### New Merging strategy
+
+When merging configs from different sources, we currently override static values such as slices, numbers, and strings.
+However, you can combine lists instead of discarding the existing ones by enabling `confmap.enableMergeAppendOption` feature flag. 
+
+##### Example
```
mx-psi, line 91: I wonder if we should use an example in which the order matters (e.g. `processors`). I think this is an important detail that may not be obvious from this example (since the order in which you declare extensions should not matter)

confmap/resolver.go
```diff
@@ -10,10 +10,19 @@ import (
...
+
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 )
 
+var enableMergeAppendOption = featuregate.GlobalRegistry().MustRegister(
+	"confmap.enableMergeAppendOption",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterFromVersion("v0.120.0"),
+	featuregate.WithRegisterDescription("if set to true, combines lists when resolving configs from different sources"),
```
mx-psi, line 23:
```suggestion
	featuregate.WithRegisterDescription("Combines lists when resolving configs from different sources. This feature gate will not be stabilized 'as is'; the current behavior will remain the default."),
```

confmap/README.md
```diff
@@ -131,6 +199,8 @@ Calling the `onChange` func from a provider triggers the collector to re-resolve
 
 An example of a `Provider` with an `onChange` func that periodically gets notified can be found in provider_test.go as UpdatingProvider
 
+> Note: By enabling this feature gate, all the lists in the given configuration will be merged.
```
mx-psi, line 202: This note should be in the section above, right? Also, could you use the syntax described [here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)?

confmap/resolver.go
```diff
@@ -10,10 +10,19 @@ import (
...
+)
```
mx-psi, line 24:
```suggestion
	featuregate.WithRegisterReferenceURL("[if set to true, combines lists when resolving configs from different sources]("https://github.com/open-telemetry/opentelemetry-collector/issues/8754"),
)
```

confmap/README.md
VihasMakwana: You're right. I put it here by mistake

confmap/README.md
VihasMakwana: I'll amend this example to include processors as well.

mx-psi (approved):

confmap/README.md
```diff
@@ -93,6 +93,91 @@ The `Resolve` method proceeds in the following steps:
 4. For each "Converter", call "Convert" for the "result".
 5. Return the "result", aka effective, configuration.
 
+#### (Experimental) Append merging strategy for lists
+
+You can opt-in to experimentally combine slices instead of discarding the existing ones by enabling `confmap.enableMergeAppendOption` feature flag. Lists are appended in the order in which they appear in their configuration sources.
```
evan-bradley, line 98:
```suggestion
You can opt-in to experimentally combine slices instead of discarding the existing ones by enabling the `confmap.enableMergeAppendOption` feature flag. Lists are appended in the order in which they appear in their configuration sources.
```

evan-bradley (approved):
Thanks for being patient through so many review cycles @VihasMakwana. I'm really happy to finally see this land, I've been wanting it for a while now.

evan-bradley:
@Aneurysm9 Do you have any concerns with the approach we've arrived at? We're still toggling it with a feature gate, but we're framing it such that the feature gate enables users to test a preview of merging lists in configs instead of being intended as the sole way to toggle it.

Aneurysm9:
> We're still toggling it with a feature gate, but we're framing it such that the feature gate enables users to test a preview of merging lists in configs instead of being intended as the sole way to toggle it.

That still doesn't feel right.  If the feature gate will never move beyond the alpha stage then it isn't a feature gate.  This feature gate as is could never move beyond the alpha stage as it would break an untold number of configurations.  What would the feature enablement look like if the test is successful and we want to make this generally available?  Would there be a new CLI flag a user would need to use to indicate this merging should take place?  If so, maybe the feature gate should control the operability of that CLI flag.  By introducing it behind a feature gate we keep the signal to the user that it is an experimental feature that may be removed in the future but also have a path forward for the feature gate that allows it to become enabled by default without breaking everyone who doesn't intend to use the feature.

mx-psi:
> If so, maybe the feature gate should control the operability of that CLI flag

I am okay with this, @VihasMakwana  could we reintroduce a flag (all-or-nothing, something like, `--merge-strategy=append`)?

VihasMakwana:
@mx-psi Yes. That should be quite simple to do.

So, the final use case would look something like:

`otelcol --config=main.yaml --config=extra_extension.yaml --feature-gates=-confmap.enableMergeAppendOption --merge-strategy=append`

please correct me if I'm wrong.

mx-psi:
@VihasMakwana you got it, that's what I had in mind :)

dmitryax:
I'm not sure I agree with the proposed solution. It's intended to solve one particular problem but doesn't provide enough controls to limit it specifically to that problem. 
1. I think, typically, users would like to apply this to the pipelines only. In other parts of the collector, there are lists that are intended to be overridden ([one example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/system/config.go#L17)). I think we need to provide an option to control either (maybe even both): 
  - which config file is going to be merged with a particular policy
  - which part of the configs to be merged with a particular policy
2. This PR adds support for only one way to merge lists by removing duplicates. In cases where the order is essential (processors), it's unclear how to control it. In case of a duplicate, do we keep the original item and drop the newly added one, or otherwise? I think we should provide more control options like:
  - append newer entries to the older ones, allow duplicates
  - append newer entries to the older ones, remove duplicates from the original list
  - append newer entries to the older ones, drop duplicates in the new list
  - prepend newer entries to the older ones, allow duplicates
  - prepend newer entries to the older ones, remove duplicates from the original list
  - prepend newer entries to the older ones, drop duplicates in the new list
  - keep the old list if present, discard the new one
  - overwrite the old list with a new one (default)

For now, I just want to raise the concern and discuss other options. So far, I think about something like this:
```
otelcol --config=main.yaml --config=config_override.yaml --config=config_append.yaml?list_merge_policy=append_override_duplicates
```
But I'm open to other suggestions that keep it flexible.

If we just add an option to merge all config, it'll be hard to make it more targeted later.

mx-psi:
>  it's unclear how to control it.

The order of the config files defines this, just like the order defines which file overrides which right now. We can document this if it is unclear.

> In case of a duplicate, do we keep the original item and drop the newly added one, or otherwise?

We can also document this.

> I think we should provide more control options like: [...]

This is very complex, I would like to postpone the decision of which of these to add until we have user feedback. This first version is simple enough for users to test out and provide feedback to us about which of these are really important.

> If we just add an option to merge all config, it'll be hard to make it more targeted later.

I would argue just the opposite. The solution was more configurable but this meant committing to specific changes in the confmap Go API. This more generic solution does not suffer from that problem: all changes to confmap are internal and completely reversible.

VihasMakwana:
@dmitryax Thanks for sharing your POV!

Let me give you some context about this PR's history, as it went through many review cycles. 
Initially, I introduced option to control which part of the configs will be merged. As that change was major one and it involved lots of changes to public APIs, we were concerned about introducing it before v1.0. Refer [this discussion](https://github.com/open-telemetry/opentelemetry-collector/pull/12097/#discussion_r1949329112).
Hence, we decided to implement a all-or-nothing merge strategy for now to keeps things simple and iterate from it in a follow-up. 

Coming to your concerns:
> 1. I think, typically, users would like to apply this to the pipelines only. In other parts of the collector, there are lists that are intended to be overridden ([one example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/system/config.go#L17)). I think we need to provide an option to control either (maybe even both):
> 
> * which config file is going to be merged with a particular policy
> * which part of the configs to be merged with a particular policy
> 

As I mentioned, this was implemented in the first iteration. 

> 2. This PR adds support for only one way to merge lists by removing duplicates. In cases where the order is essential (processors), it's unclear how to control it. In case of a duplicate, do we keep the original item and drop the newly added one, or otherwise? I think we should provide more control options like:
> 
In case of duplicates, we keep the original item and drop newly added one.

> * append newer entries to the older ones, allow duplicates
> * append newer entries to the older ones, remove duplicates from the original list
> * append newer entries to the older ones, drop duplicates in the new list
> * prepend newer entries to the older ones, allow duplicates
> * prepend newer entries to the older ones, remove duplicates from the original list
> * prepend newer entries to the older ones, drop duplicates in the new list
> * keep the old list if present, discard the new one
> * overwrite the old list with a new one (default)
> 
This is on the cards and I'll open a follow-up issue to discuss the options.

> For now, I just want to raise the concern and discuss other options. So far, I think about something like this:
> 
> ```
> otelcol --config=main.yaml --config=config_override.yaml --config=config_append.yaml?list_merge_policy=append_override_duplicates
> ```
> 
> But I'm open to other suggestions that keep it flexible.
> 
> If we just add an option to merge all config, it'll be hard to make it more targeted later.

I believe as I initially implemented the targeted version, it should be fairly simple to re-introduce it.

@mx-psi @evan-bradley feel free to add anything I might have missed.

dmitryax:
> The order of the config files defines this, just like the order defines which file overrides which right now. We can document this if it is unclear.

My concern is not only about the ordering but about controlling that it's applied to configuration parts where it's expected. For example:

I want to define a component in another config file and add it to a "standard" pipeline defined in a "base" config, but the "base" config may already have it as well (I don't have control over it). In the case of a conflict, I'd like to be able to choose whether to override the original component's config or keep it as is. Overriding maps is currently impossible, given that we have things like `receivers.otlp.protocols.grpc` (something we can eventually address). But at least we can control overriding lists in the component's config. This all-or-nothing approach doesn't allow that.

@VihasMakwana @mx-psi I'm ok with this as an initial version. I'm just afraid that it'll be stuck in this state for a long time, and users will be relying on the behavior of this feature gate going forward. Can we have another issue where we track the progress? @VihasMakwana, will you be able to keep working on this until it's stabilized?

confmap/resolver.go
```diff
@@ -55,6 +68,10 @@ type ResolverSettings struct {
 	// ConverterSettings contains settings that will be passed to Converter
 	// factories when instantiating Converters.
 	ConverterSettings ConverterSettings
+
+	// Following option is used to specify the merging strategy to be used while
+	// merging configurations from multiple sources.
+	MergeStrategy string
```
VihasMakwana, line 74: @mx-psi I couldn't find a better way to keep this option away.

We need to pass the command line's option to Resolver and this is our best option.

We don't change anything and the core implementation still remains same. Can you take a look?

VihasMakwana:
> > The order of the config files defines this, just like the order defines which file overrides which right now. We can document this if it is unclear.
> 
> My concern is not only about the ordering but about controlling that it's applied to configuration parts where it's expected. For example:
> 
> I want to define a component in another config file and add it to a "standard" pipeline defined in a "base" config, but the "base" config may already have it as well (I don't have control over it). In the case of a conflict, I'd like to be able to choose whether to override the original component's config or keep it as is. Overriding maps is currently impossible, given that we have parts with meaningful `null` values like `receivers.otlp.protocols.grpc` (something we can eventually address). But at least we can control overriding lists in the component's config. This all-or-nothing approach doesn't allow that.
> 
> @VihasMakwana @mx-psi I'm ok with this as an initial version. I'm just afraid that it'll be stuck in this state for a long time, and users will be relying on the behavior of this feature gate going forward. Can we have another issue where we track the progress? @VihasMakwana, will you be able to keep working on this until it's stabilized?

yes. This is one of my top priorities and I'll open up an follow up issue to track progress.

confmap/resolver.go
evan-bradley: My understanding of what @dmitryax wrote is that we are okay with the initial version only including the feature gate. @dmitryax can you confirm?

Especially after the most recent discussion on this PR, I think we likely need to have an RFC to explore how we want this to ultimately work. I still think the feature gate will be useful for a meaningful number of users, and will hopefully drive feedback for additional functionality we should consider, so I would like to add it as an initial step.

VihasMakwana: I agree. Let's wait for Dmitry's comments.

I just want to confirm on thing. For this PR, we agree that we should include the feature gate and command line as an initial version.
Then, discuss it in detail on an RFC to finalize the implementation.

evan-bradley: >I just want to confirm on thing. For this PR, we agree that we should include the feature gate and command line as an initial version.

My understanding is that we can go with only the feature gate on this PR. I think even the command line argument will need some initial design and we will need to be careful, since we won't get a chance to revise it later.

VihasMakwana: Alright. So, we will go with the version before I introduced the command line i.e. just feature gate.
Then, discuss potential options over an RFC.

VihasMakwana:
@mx-psi @evan-bradley @dmitryax I think we can go forward this PR as an initial version. 

I'll open up an RFC as a follow-up and discuss how to address concerns related to ordering and controlling of specific parts of configuration.

confmap/README.md
```diff
@@ -93,6 +93,91 @@ The `Resolve` method proceeds in the following steps:
 4. For each "Converter", call "Convert" for the "result".
 5. Return the "result", aka effective, configuration.
 
+#### (Experimental) Append merging strategy for lists
+
+You can opt-in to experimentally combine slices instead of discarding the existing ones by enabling the `confmap.enableMergeAppendOption` feature flag. Lists are appended in the order in which they appear in their configuration sources.
+This will **not** become the default in the future, we are still deciding how this should be configured and want your feedback on [this issue](https://github.com/open-telemetry/opentelemetry-collector/issues/8754).
```
dmitryax, line 99: > This will **not** become the default in the future

Do we want to replace it with another feature gate in future? I think we can keep the feature gate and just keep changing its behavior. So, I would change the wording to something like "The behavior of this feature gate will change over time before it can be considered for beta stage".

However there have been a lot of discussion around that. So I maybe missing something

confmap/merge.go
```diff
@@ -0,0 +1,79 @@
...
+			continue
+		}
+
+		switch srcVal.Kind() {
+		case reflect.Array, reflect.Slice:
+			// both of them are array. Merge them
+			src[dKey] = mergeSlice(srcVal, destVal)
+		case reflect.Map:
+			// both of them are maps. Recursively call the mergeMaps
+			mergeMaps(sVal.(map[string]any), dVal.(map[string]any))
```
dmitryax, line 52: Should we call `koanf.maps.IntfaceKeysToStrings` before instead of casting like this?

confmap/README.md
dmitryax: SGTM 👍

confmap/merge.go
VihasMakwana: I surfed web and it turns out that YAML keys can only be strings, so we should be okay. wdyt?

confmap/merge.go
```diff
@@ -0,0 +1,79 @@
...
+	maps.Merge(src, dest)
+	return nil
+}
+
+func mergeMaps(src, dest map[string]any) {
+	for dKey, dVal := range dest {
+		sVal, sOk := src[dKey]
+		if !sOk {
+			// old key is not present in new config. Hence, add it to src
+			src[dKey] = dVal
```
dmitryax, line 34: Why do we modify `src` not `dest`?

dmitryax: Oh I see why... We call `maps.Merge(src, dest)` after that to override `dest`. This is. a bit confusing. How much manual code do we save by relying on calling `maps.Merge` after?

VihasMakwana: because `src` will override `dest`. It's counter-intuitive. Refer this [discussion](https://github.com/open-telemetry/opentelemetry-collector/pull/12097#discussion_r1937421659) around the "werid" naming convention.

confmap/merge.go
dmitryax: I checked the code, looks like koanf actually calls if before calling a custom `merge` function. So we are good 👍

confmap/resolver.go
dmitryax: Sounds good to me 👍

dmitryax (approved):
LGTM

dmitryax:
Please update the nolint comment to fix CI

evan-bradley (approved):

VihasMakwana:
A few of test cases are failing but they seem unrelated to my PR.

cmd/otelcorecol/go.mod
```diff
@@ -104,7 +104,7 @@ require (
 	go.opentelemetry.io/collector/exporter/exporterhelper/xexporterhelper v0.120.0 // indirect
 	go.opentelemetry.io/collector/exporter/exportertest v0.120.0 // indirect
 	go.opentelemetry.io/collector/exporter/xexporter v0.120.0 // indirect
-	go.opentelemetry.io/collector/extension/extensionauth v0.0.0-20250227134758-6a98ee8b14a2 // indirect
+	go.opentelemetry.io/collector/extension/extensionauth v0.120.0 // indirect
```
mx-psi, line 107: Please don't revert this!

VihasMakwana: Woops

VihasMakwana: @mx-psi it's good now.

VihasMakwana:
@mx-psi I have fixed the the go.mod and I'm not reverting `extensionauth`. can you please re-review?

PR #12102, opened 2025-01-15
[component] Allow calling `component.ValidateConfig` at the top level

evan-bradley:
#### Description

Updates `component.ValidateConfig` to recurse through the entire config object by checking for `ConfigValidator` implementations of structs that are under interfaces. Right now `ValidateConfig` stops at things like `component.Config` which themselves can't implement `component.ValidateConfig` because `callValidateIfPossible` can only see the interface and not the concrete type before this change.

I can break out some changes in this PR so we only implement this for component here and actually call `component.ValidateConfig` inside otelcol in a subsequent PR, but I want to show what the effect of this will be.

#### Link to tracking issue
Replaces https://github.com/open-telemetry/opentelemetry-collector/pull/12058.

Works toward https://github.com/open-telemetry/opentelemetry-collector/issues/11524.


service/config.go
```diff
@@ -22,15 +20,3 @@ type Config struct {
...
 }
-
-func (cfg *Config) Validate() error {
-	if err := cfg.Pipelines.Validate(); err != nil {
-		return fmt.Errorf("service::pipelines config validation failed: %w", err)
-	}
-
-	if err := cfg.Telemetry.Validate(); err != nil {
-		fmt.Printf("service::telemetry config validation failed: %v\n", err)
-	}
```
TylerHelmuth, line 33: Are you sure we want to lose these error messages?

evan-bradley: In my opinion we don't, they're essential for users knowing where an error occurred. I think this is motivation for moving `ConfigValidator` to confmap: we can use the mapstructure tags to recreate these paths automatically.

TylerHelmuth: If this PR is merged and ConfigValidator is not moved what is the resulting error message when `otelcol` does the validation?

evan-bradley: It will just be the direct message from the `Validate` function. So for not configuring any protocols for the OTLP receiver, the Collector will print:

```
collector server run finished with error: invalid configuration: must specify at least one protocol when using the OTLP receiver
```

We could still make sure this message is printed inside component, but I think it's undesirable and adds to the reasons why we should move `ConfigValidator`:

1. Simply use the struct field names to create paths and ignore any mapstructure tags. The paths would still be "correct" but may have names that don't match what's in the YAML file.
2. Use mapstructure anyway either implicitly or explicitly to do this. This won't have any negative impacts for most users since mapstructure used by confmap and confmap is pretty much universal in the Collector. However I think it's not ideal.

evan-bradley: If we like this, I want to issue the following PRs:

1. This change with just the `ValidateConfig` changes. otelcol and service will still manually call `Validate` and keep the config paths in the error messages.
2. Introduce printing the path to `ValidateConfig`.
3. Remove direct calls to `Validate` in the codebase.
4. Move `ConfigValidator` to confmap.

mx-psi: I support this plan. I think it is important to print the full path to configuration options (like we are doing today or with a new mechanism)

TylerHelmuth: Ya that sounds good to me.

evan-bradley:
I'm skipping the contrib tests for now, I'll fix any errors with an `update-otel` PR in contrib.

service/config.go
evan-bradley: @mx-psi @TylerHelmuth this is ready to review.

Follow-up is here: https://github.com/open-telemetry/opentelemetry-collector/pull/12108.

mx-psi (approved):

jade-guiton-dd (approved):

TylerHelmuth (approved):

PR #12104, opened 2025-01-16
[chore][exporter] Consolidate merge splitting for the case where maxLimit is set and the case it's not

sfc-gh-sili:
This PR consolidates merge splitting for the case where maxLimit is set and the case it's not. This should not have any effect from users' perspective.

#### Link to tracking issue
Fixes #

### DIFF
--- a/exporter/internal/queue/default_batcher.go
+++ b/exporter/internal/queue/default_batcher.go
@@ -52,7 +52,7 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
 				reqList, mergeSplitErr = qb.currentBatch.req.MergeSplit(ctx, qb.batchCfg.MaxSizeConfig, req)
 			}
 
-			if mergeSplitErr != nil || reqList == nil {
+			if mergeSplitErr != nil || len(reqList) == 0 {
 				qb.queue.OnProcessingFinished(idx, mergeSplitErr)
 				qb.currentBatchMu.Unlock()
 				continue
--- a/exporter/internal/queue/default_batcher.go
+++ b/exporter/internal/queue/default_batcher.go
@@ -43,77 +43,41 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
 
 			qb.currentBatchMu.Lock()
 
-			if qb.batchCfg.MaxSizeItems > 0 {
-				var reqList []internal.Request
-				var mergeSplitErr error
-				if qb.currentBatch == nil || qb.currentBatch.req == nil {
-					qb.resetTimer()
-					reqList, mergeSplitErr = req.MergeSplit(ctx, qb.batchCfg.MaxSizeConfig, nil)
-				} else {
-					reqList, mergeSplitErr = qb.currentBatch.req.MergeSplit(ctx, qb.batchCfg.MaxSizeConfig, req)
-				}
-
-				if mergeSplitErr != nil || reqList == nil {
-					qb.queue.OnProcessingFinished(idx, mergeSplitErr)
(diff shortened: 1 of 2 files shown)


exporter/internal/queue/default_batcher.go
```diff
@@ -43,77 +43,41 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
...
+			} else {
+				reqList, mergeSplitErr = qb.currentBatch.req.MergeSplit(ctx, qb.batchCfg.MaxSizeConfig, req)
+			}
 
-				if mergeSplitErr != nil || reqList == nil {
-					qb.queue.OnProcessingFinished(idx, mergeSplitErr)
-					qb.currentBatchMu.Unlock()
-					continue
-				}
+			if mergeSplitErr != nil || reqList == nil {
```
bogdandrutu, line 55: Not sure you want to do anything if error, unless I am missing something.

bogdandrutu, line 55: A slice never compare to nil, you must most likely compare len to 0

exporter/internal/queue/default_batcher.go
```diff
@@ -43,77 +43,41 @@ func (qb *DefaultBatcher) startReadingFlushingGoroutine() {
...
-							idxList: []uint64{idx},
-						})
-						// TODO: handle partial failure
-					}
-					qb.resetTimer()
-				} else {
-					qb.currentBatch = &batch{
-						req:     reqList[0],
+			// If there was a split, we flush everything immediately.
+			if reqList[0].ItemsCount() >= qb.batchCfg.MinSizeItems || len(reqList) > 1 {
```
bogdandrutu, line 62: What if len is 0, get a segfault here?

exporter/internal/queue/default_batcher.go
jmacd: Hmmm. In Go a slice can be nil and the comparison is valid. Note however that `nil != []type{}`.
https://go.dev/tour/moretypes/12

jmacd:
I tried to understand this PR, but I found myself with background-level questions as in #12118, would love to see package-level documentation on the new relationship between the queue sender and batch sender, so that I can keep track of timeout and cancellation-related topics. Thank you @sfc-gh-sili for leading this effort.

exporter/internal/queue/default_batcher.go
sfc-gh-sili: According to the `exporter.request` specification, the result is guaranteed to be non-empty if there is no error.

https://github.com/open-telemetry/opentelemetry-collector/blob/d977da4d32de422140049f99492c270355a21525/exporter/internal/request.go#L28

But to be safe, we can add a check and return early.

exporter/internal/queue/default_batcher.go
sfc-gh-sili: merge error is reserved for things like failed type check, and means the request should not be retried. We want to remove the request (by calling `onProcessingFinished`) from the queue in that case.

sfc-gh-sili: Makes sense! I updated `reqList == nil` check to `len(reqList) == 0`.

bogdandrutu: > Hmmm. In Go a slice can be nil and the comparison is valid. Note however that nil != []type{}.
https://go.dev/tour/moretypes/12

I know that :) It is just that in practice we should always compare len with 0.

sfc-gh-sili:
Closing this PR as the two paths are consolidated in https://github.com/open-telemetry/opentelemetry-collector/pull/12260.

PR #12108, opened 2025-01-16
[component] Print path to error in `ValidateConfig`

evan-bradley:
#### Description

Follow up to https://github.com/open-telemetry/opentelemetry-collector/pull/12102.

This prints the full path to where the error occurred in a config object using available reflection metadata. Currently some paths will be duplicated until we shift otelcol to call `component.ValidateConfig` on `otelcol.Config` and remove manual calls to `Validate`, but no information will be missing. If this is a concern we can do both steps at once.

Example error messages:

```
collector server run finished with error: invalid configuration: receivers::otlp: must specify at least one protocol when using the OTLP receiver
service::telemetry: collector telemetry metrics reader should exist when metric level is not None, level is Normal
```

I'll clean this PR up once https://github.com/open-telemetry/opentelemetry-collector/pull/12102 is merged, but you can see the latest commit for the changes relevant to only this PR.


mx-psi:
I think this can be rebased and marked as ready for review

component/config.go
```diff
@@ -31,46 +33,131 @@ type ConfigValidator interface {
...
+			keyErrs := validate(iter.Key())
+			valueErrs := validate(iter.Value())
+			var key string
+
+			if str, ok := iter.Key().Interface().(string); ok {
+				key = str
+			} else if stringer, ok := iter.Key().Interface().(fmt.Stringer); ok {
+				key = stringer.String()
+			} else {
+				key = fmt.Sprintf("[%T key]", iter.Key().Interface())
```
evan-bradley, line 135: I'm not sure what we want to put here, but for keys that can't be made into a reasonable string representation, I've just chosen to print the type so we have something displayed. There are tests that show how this looks.

evan-bradley:
As expected, some contrib tests are failing despite no outward API changes. I feel like this is more of an indication that the tests should be rewritten than that we've made a breaking change: most of the test failures seem to be happening because they assume a specific structure of the error object instead of working with the `error` interface.

component/config.go
```diff
@@ -31,46 +33,131 @@ type ConfigValidator interface {
...
+	var err error
+	errs := validate(reflect.ValueOf(cfg))
+
+	for _, suberr := range errs {
+		if suberr.err != nil {
+			if suberr.path != "" {
+				err = errors.Join(err, fmt.Errorf("%s: %w", suberr.path, suberr.err))
+			} else {
+				err = errors.Join(err, suberr.err)
+			}
```
mx-psi, line 45: nit: We could move this logic inside a `Error` method for `pathError`

component/config.go
```diff
@@ -31,46 +33,131 @@ type ConfigValidator interface {
 // ValidateConfig validates a config, by doing this:
 //   - Call Validate on the config itself if the config implements ConfigValidator.
 func ValidateConfig(cfg Config) error {
-	return validate(reflect.ValueOf(cfg))
+	var err error
+	errs := validate(reflect.ValueOf(cfg))
+
+	for _, suberr := range errs {
+		if suberr.err != nil {
```
mx-psi, line 40: When does this happen?

component/config.go
```diff
@@ -31,46 +33,131 @@ type ConfigValidator interface {
...
+					tags := strings.Split(tag, ",")
+					if len(tags) > 0 {
+						fieldName = tags[0]
+					}
+				}
+				// Even if the mapstructure tag exists, the field name may not
+				// be available, so set it if it is still blank.
+				if len(fieldName) == 0 {
+					fieldName = strings.ToLower(field.Name)
+				}
```
mx-psi, line 86: nit: move to helper method `fieldName`?

mx-psi, line 86: Couldn't this be done outside of this loop?

component/config.go
```diff
@@ -31,46 +33,131 @@ type ConfigValidator interface {
...
+			}
+		}
+	}
+
+	return err
+}
+
+type pathError struct {
+	err  error
+	path string
```
mx-psi, line 54: I would store this as a slice here
```suggestion
	path []string
```

and then do the joining in the `Error` method, I think that will give you simpler code (you avoid all the ifs).

You can probably store in reverse order as well so that you just append to the list, and then reverse when printing

component/config.go
mx-psi: We could also use `%v` or `%#v`

component/config.go
evan-bradley: I like that. Done.

component/config.go
evan-bradley: I previously allowed `nil` errors so we didn't have to do a nil check every time we called `callValidateIfPossible` on the value. I'm okay either way, so I've updated it so we don't include `pathError` objects with nil `err` fields in the slice anymore.

component/config.go
evan-bradley: Good thinking, done.

evan-bradley: Great catch, I've moved it outside the loop.

component/config.go
evan-bradley: I opted for a different name, but I agree we should just join at the end. This also cleans up the code a good amount, thanks for calling this out.

component/config.go
evan-bradley: I avoided printing the value because it produces odd-looking error messages like:

```
mapkeychild::[component.configChildStruct{Child:component.errConfig{err:(*errors.errorString)(0xc000039c80)}, ChildPtr:(*component.errConfig)(nil)}]::child: child key error
```

We can probably be smarter about this by doing additional reflection to determine whether we're looking at a primitive or complex type, my initial goal was just to keep it simple for this iteration. I'll play with this.

evan-bradley: I arrived at something I think is simple enough, take a look in the latest commit.

component/config.go
mx-psi: > Edit: To clarify, I went with toError instead of Error so as not to conflict with the error interface, which specifies that Error returns a string.

Oops, I missed this and I left a suggestion below to change it to implement the `Error` interface :smile:

component/config.go
mx-psi: Looks reasonable!

mx-psi:
Can we fix (some or all of) the contrib tests before merging this?

evan-bradley:
It's definitely possible, let me see what I can do.

component/config.go
```diff
@@ -33,22 +33,15 @@ type ConfigValidator interface {
...
 func ValidateConfig(cfg Config) error {
-	var err error
-	errs := validate(reflect.ValueOf(cfg))
-
-	for _, suberr := range errs {
-		err = errors.Join(err, suberr.toError())
-	}
-
-	return err
+	return errors.Join(validate(reflect.ValueOf(cfg)))
```
mx-psi, line 36: Ah, damn you would need to change the return type of `validate` or cast items one by one :cry:

mx-psi: Sorry, I hadn't thought about this. I still like using the standard Go interfaces more, but I see less value in my suggestion now

evan-bradley: I can't find a way to pull this off. The spread operator doesn't work since the type system still considers it a slice and not separate arguments. Changing the return type of `validate` will require a ton of casting since we use the `pathError` type directly in recursive calls to `validate`.

I liked consolidating this onto one line, any other ideas for how to solve this?

evan-bradley: We could reapply the loop with what's here, though part of the reason I had the `toError` method create new errors instead of having `pathError` implement `error` was to keep the `pathError` type internal and remove it from the error chain (not that it really hurts anything).

evan-bradley: I'm fine sticking with the standard Go interfaces even if the `pathError` type ends up in the chain. I've reintroduced the loop in `ValidateConfig`.

mx-psi (approved):

component/config.go
mx-psi: I have not been able to find a way, let's keep it like this if you are okay with it

evan-bradley:
I opened https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37579 to address hopefully most of the failing contrib tests. Unfortunately it's proven bit challenging to run everything locally.

PR #12112, opened 2025-01-17
[chore] Remove unnecessary usage of nolint:gosec 

### LINKED ISSUE #11425: Remove usage of nolint:gosec added by #10935
There are lots of `// nolint:gosec` annotations to move forward with https://github.com/open-telemetry/opentelemetry-collector/pull/10935 since the problems are not new. We should fix the wrong conversions where possible.

zzzk1:
#### Description
- Remove unnecessary usage of `// nolint:gosec` introduced by PR #10935
- Add exclude-rules G115

#### Link to tracking issue
Fixes #11425


bogdandrutu:
> Remove incorrect usage of // nolint:gosec introduced by PR https://github.com/open-telemetry/opentelemetry-collector/pull/10935

I don't think they were incorrect, but the rules were relaxed.

bogdandrutu (approved):

bogdandrutu (approved):
No need for chlog here.

zzzk1:
> > Remove incorrect usage of // nolint:gosec introduced by PR #10935
> 
> I don't think they were incorrect, but the rules were relaxed.

So, all we need to do is add exclude rules and then ignore them, right?

.golangci.yml
```diff
@@ -225,6 +225,9 @@ issues:
     - text: "G402:"
       linters:
         - gosec
+    - text: "G115:"
+      linters:
+        - gosec
```
bogdandrutu, line 230: I am confused, I did not see that you excluded it.

bogdandrutu (commented):
I missed that you excluded. We should not exclude the rule, but fix the usage.

bogdandrutu (changes requested):
I missed that you excluded. We should not exclude the rule, but fix the usage.

zzzk1:
@bogdandrutu `G115` has not been fixed yet: https://golangci-lint.run/product/changelog/. I tried using a condition like `if a <= math.MaxInt64()` (where `a` is of type `uint64`), but it did not work.

PR #12118, opened 2025-01-18
[exporter][queuebatching] Implemented "DisabledQueue" which is used when batch is enabled but queue is disabled

sfc-gh-sili:
#### Description

This PR implements a component called "DisabledQueue".

| Component    | Behavior |
| -------- | ------- |
| DisabledQueue  | Offer(req) blocks until the request is sent out |
| Blocking BoundedMemoryQueue | Offer(req) blocks until there is space in the queue  |
| Non Blocking BoundedMemoryQueue | Offer(req) returns an error if the queue is full  |

DisabledQueue is used in place of PersistentStorageQueue or BoundedMemoryQueue when batching is enabled but queueing is disabled. Having this component enables us to have a unified implementation for the case when queueing is enabled or not.

#### Link to tracking issue


jmacd:
@sfc-gh-sili I read this PR trying to understand why when the queue sender is disabled, we have to create a "real" queue in some sense. For example, this PR you allocate a `sizedQueue[disabledMemQueueEl[T]]` as the stand-in for the batcher to use when there is no queueing. I would like a little documentation explaining how this works--or maybe put another way, why would we use the pull-based batcher if there is not a real queue? It would seem to make sense to continue with push-based batching if there is no queue, and I'm looking for a document that explains the current architecture and where we are headed so I can update #11948.

sfc-gh-sili:
@jmacd
Hi Joshua!
> I read this PR trying to understand why when the queue sender is disabled, we have to create a "real" queue in some sense.
The disabled queue is introduced so we can entirely get rid of `batch_sender`. That way `batcher` (owned by `queue_sender`) would be the one and only component that does batching.

For example, this PR you allocate a sizedQueue[disabledMemQueueEl[T]] as the stand-in for the batcher to use when there is no queueing. I would like a little documentation explaining how this works--or maybe put another way, why would we use the pull-based batcher if there is not a real queue? It would seem to make sense to continue with push-based batching if there is no queue, and I'm looking for a document that explains the current architecture and where we are headed so I can update https://github.com/open-telemetry/opentelemetry-collector/pull/11948.

jmacd:
Thank you @sfc-gh-sili. I see also #12245 addresses this topic, I'll continue paying attention. I've closed both of my RFCs #11947 #11948 (they're out of date) and will re-open after the dust settles.

sfc-gh-sili:
Closing this PR as https://github.com/open-telemetry/opentelemetry-collector/pull/12242 resolves the issue.

PR #12150, opened 2025-01-22
doc: add detail to confmap watcher

### LINKED ISSUE #12115: Add documentation and example for confmap/Provider interface which actually has a callback mechanism
The comment in code says this
```
	// `watcher` callback is called when the config changes. watcher may be called from
	// a different go routine. After watcher is called Retrieved.Get should be called
	// to get the new config. See description of Retrieved for more details.
	// watcher may be nil, which indicates that the caller is not interested in
	// knowing about the changes.
```
On writing my implementation, I want to fetch the config from a remote source every 30 seconds. How to provide the updated config back to the caller after invoking matcher func? I dont find any way to do that nor did I find any examples. 

Here's my code:
```
package testprovider

import (
	"context"
	"go.opentelemetry.io/collector/confmap"
	"io"
	"net/http"
	"time"
)

var _ confmap.Provider = (*TestProvider)(nil)

type TestProvider struct{}

func (t TestProvider) Retrieve(_ context.Context, uri string, watcher confmap.WatcherFunc) (*confmap.Retrieved, error) {
	client := http.Client{
		Timeout: 5 * time.Second,
	}
	tickerChan := time.Tick(30 * time.Second)
	for {
		select {
		case <-tickerChan:
			res, err := client.Get(uri)
			if err != nil {
				watcher(&confmap.ChangeEvent{Error: err})
				continue
			}
			defer res.Body.Close()
			resBody, err := io.ReadAll(res.Body)
			if err != nil {
				watcher(&confmap.ChangeEvent{Error: err})
				continue
			}
			// but this will exit out of my loop after one invocation, I want the config to be reloaded every 30 seconds
			return confmap.NewRetrievedFromYAML(resBody)
		}
	}
}

func (t TestProvider) Scheme() string {
	return "http"
}

func (t TestProvider) Shutdown(_ context.Context) error {
	return nil
}

```

mattsains:
#### Description
I tried to improve the docs on watchers based on the issue below

#### Link to tracking issue
Fixes #12115

#### Documentation
Clarified some code comments that were incorrect (there's no Retrieved.Get method) and included a new code sample for a Provider that uses the watcher func

### DIFF
--- a/confmap/README.md
+++ b/confmap/README.md
@@ -129,44 +129,7 @@ Calling the `onChange` func from a provider triggers the collector to re-resolve
 ◄───────────┤                     │
 ```
 
-Example of `Provider` with `onChange` func that periodically gets notified:
-
-```golang
-type UpdatingProvider struct { }
-
-func (p UpdatingProvider) Retrieve(ctx context.Context, uri string, watcher confmap.WatcherFunc) (*confmap.Retrieved, error) {
-    ticker := time.Ticker(30 * time.Second)
-
-    retrieved := provider.NewRetrieved(getCurrentConfig(), provider.WithRetrievedClose(func (ctx context) error {
-        // the provider should call this function when it no longer wants config updates
-        close(ticker)
-    }))
-
-    go func() {
-        for {
-            select {
-            case <-ctx.Done():
-                // if the context is closed, then we should stop sending updates
-                close(ticker)
-                return
-            case _, ticking <- ticker:
-                if !ticking {
-                    // if ticking stopped, then closeFunc was called
-                    return
-                }
-                // otherwise, notify the watcher that a new config is available
-                // the watcher is expected to call Provider.Retrieve again to get the update
-                // note that the collector calls closeFunc before calling Retrieve for the second time,
-                // so these go functions don't accumulate indefinitely. (see otelcol/collector.go, Collector.reloadConfiguration)
-                watcher(&confmap.ChangeEvent{})
-            }
-        }
-    }()
-
-    return retrieved, nil
-}
-```
-
+An example of a `Provider` with an `onChange` func that periodically gets notified can be found in provider_test.go as UpdatingProvider
 
 ## Troubleshooting
 
--- a/confmap/README.md
+++ b/confmap/README.md
@@ -37,7 +37,7 @@ that can be used by code that is oblivious to the usage of `Providers` and `Conv
 or an individual value (partial configuration) when the `configURI` is embedded into the `Conf` as a values using
 the syntax `${configURI}`.
 
-**Limitation:** 
+**Limitation:**
 - When embedding a `${configURI}` the uri cannot contain dollar sign ("$") character unless it embeds another uri.
 - The number of URIs is limited to 100.
 
@@ -87,7 +87,7 @@ The `Resolve` method proceeds in the following steps:
 After the configuration was processed, the `Resolver` can be used as a single point to watch for updates in the
 configuration retrieved via the `Provider` used to retrieve the “initial” configuration and to generate the “effective” one.
 
-```terminal      
+```terminal
          Resolver              Provider
             │                     │
    Watch    │                     │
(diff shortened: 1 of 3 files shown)


mattsains:
I can write a CHANGELOG entry for this, but I don't think it's helpful. Let me know if I should anyway to get the CI to pass

confmap/README.md
```diff
@@ -99,16 +118,62 @@ configuration retrieved via the `Provider` used to retrieve the “initial” co
...
+───────────►│                     │
+            │                     │
+            │      Retrieve       │
+            ├────────────────────►│
+            │        Conf         │
+            │◄────────────────────┤
+◄───────────┤                     │
+```
+
+Example of `Provider` with `onChange` func that periodically gets notified:
```
mx-psi, line 132: I wonder if we could turn this into [a testable example](https://go.dev/blog/examples). That way we are able to lint and test this code over time and ensure it doesn't get outdated (which is harder in a Markdown file)

mx-psi: Thank you!

mx-psi (approved):

confmap/provider_test.go
```diff
@@ -6,13 +6,87 @@ package confmap
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				// if the context is closed, then we should stop sending updates
+				ticker.Stop()
+				return
+			case <-stop:
+				// closeFunc was called, so stop updating the watcher
+				fmt.Println("unsubscribing watcher")
```
evan-bradley, line 48:
```suggestion
				t.Println("unsubscribing watcher")
```
Could we put a `*testing.T` on the `UpdatingProvider` type and use that to print here?

confmap/provider_test.go
```diff
@@ -6,13 +6,87 @@ package confmap
...
+				// so these go functions don't accumulate indefinitely. (see otelcol/collector.go, Collector.reloadConfiguration)
+				watcher(&ChangeEvent{})
+			}
+		}
+	}()
+
+	return retrieved, nil
+}
+
+func ExampleProvider() {
```
evan-bradley, line 63:
```suggestion
// Tests the example scenario in README.md
func TestProviderWithWatcher(t *testing.T) {
```
Could we make this into a test and verify the output specified in the "Output" comment section of this function?

mattsains: @mx-psi  wanted this to be a testable example

mx-psi: @evan-bradley we don't have to change the name (we can keep it as an example) to use the "Output comment", right?

evan-bradley: Maybe I misunderstood the intent of this, I guess what has me confused is that this is in the test file despite not being a test. Should we move it to the `doc.go` file or an internal `examples` directory instead?

mx-psi: > I guess what has me confused is that this is in the test file despite not being a test

Go allows you to write down [testable examples](https://go.dev/blog/examples) if you start the function name with `Example`. This is the thing that powers https://pkg.go.dev/net/http#pkg-examples, it shows up nicely in the Godoc

evan-bradley: Oh wow, I didn't know about that, that's a cool feature. Sorry, just now seeing your earlier comment where you linked to this earlier. I think what's here is perfect then.

evan-bradley (approved):

mx-psi:
@bogdandrutu Could you double check this since you designed this in the first place?

confmap/provider_test.go
```diff
@@ -6,13 +6,87 @@ package confmap
...
+
+	retrieved, err := provider.Retrieve(context.Background(), "example", watcherFunc)
+	if err != nil {
+		fmt.Println("received an error")
+	} else {
+		fmt.Printf("received: %s\n", retrieved.rawConf)
+	}
+
+	// sleep for 1.5 seconds (1s for a new configuration, and 0.5 more to prevent races. Too close to 2s and we risk getting notified twice)
+	time.Sleep(1500 * time.Millisecond)
```
mx-psi, line 78: Thinking about this, I think this introduces the risk of flakes on CI, what about 
1. defining a channel `ack := make(chan struct{})`
2. On `watcherFunc`, sending a `struct{}{}` to `ack`
3. Replacing this sleep with a wait on `ack`

We can probably replace the second sleep with a similar pattern. 

What do you think?

mattsains: I'll work on this

mattsains: I replaced the first sleep with a channel (the one waiting for a notification of a new config. However, keeping in mind that this is destined to be documentation and not test case, I couldn't find a natural way to notify when the retriever has unsubscribed the watchFunc. Any way I tried to do it compromised the structure of the Provider in a way that I think would detract from the documentational value. For example:

* I considered returning a channel as the rawConf return, but that doesn't make sense because the config is never going to be a channel in normal uses. 
* I considered cancelling the context of the Retrieve() method call, but that doesn't make sense because you usually wouldn't do that in a Provider. 
* I also considered creating a channel as a public instance member of the Provider, but again that is not didactically sound because a single Retrieve call has a different duration/lifetime than the entire Provider instance. 

I do agree that the sleeps introduce a chance of a race in the tests. I've removed the test for the unsubscribe of the watcher. If you'd still like that tested, I'm just not sure how to do that without making it more like a confusing test and less like documentation

mx-psi: Thank you for taking a look at this, and sorry that this slipped on the first review. I think this looks good to merge now, if we get feedback that this is confusing or the 'unsubscribe watcher' would be helpful, we can add it.

Issue #9462, opened 2024-02-05
[exporterhelper] User interface for controling queue measurement


dmitryax:
Following the discussion https://github.com/open-telemetry/opentelemetry-collector/pull/8853#discussion_r1435671202

As part of https://github.com/open-telemetry/opentelemetry-collector/issues/8122, the batching capability is moved to the exporter side. The batch sizes are configured in terms of items by default. The current queue can be configured only in terms of requests. The different measures can be confusing to the end user. At least, we need to provide a way to configure the queue in terms of items that would be consistent with the batching.

This issue is intended to answer the following questions:

1. How to define the user interface for the end user to configure different queue measurements.
2. Do we want to migrate the default measure for the queue from requests to items to be consistent with the batching configuration? 

Currently, the `sending_queue` config group in the exporter configuration has the only field to control the queue capacity called `queue_size`, which means the maximum number of requests the queue can hold. The `queue` prefix can potentially be dropped as redundant during the migration to the new interface.

The following measures can be supported by the new exporter queue additionally. The options 1 and 2 are already implemented, just not exposed to the user until this issue is resolved. Option 3 can be added in the future.

1. Number of basic items: log records, spans, data points - this should be the default measure going forward to be consistent with config for batching
2. Number of incoming requests - this is the current measure used by the `queue_size` field. We should keep it as an option.
3. Size in MiB - will be nice to have in the future. We have similar options in the memory limiter: `limit_mib` and `spike_limit_mib`.

We have the following options how the different queue measurements can be configured:

A. Provide different fields for each option and validate that only one is set. The following options can be added under `sending_queue` config group in the exporter configuration

- `max_size_items`, `max_size_requests` and `max_size_mib` (vote with 😄 )
- `size_limit_items`, `size_limit_requests` and `size_limit_mib` (vote with 🎉 )
- `items_limit`, `requests_limit` and `mib_limit` (vote with 😕 )

B. Keep `queue_size` (or chose another name like `size`) and introduce another field for the `queue_measure` (or `measure`) enum which can be `items`, `requests` or `MiB`. (vote with ❤️ )

C. Keep `queue_size` (or chose another name like `size`) and support suffixes like `10requests`, `10items`, `10[B|KB|MB..]` and keep the queue_size field. (vote with 🚀 )

[...]

PR #12154, opened 2025-01-22
[exporter][batching] configuration and config validation for bytes based batching

sfc-gh-sili:
#### Description

This PR adds config API that will be used for serialized bytes based batching. 

We will deprecate `MinSizeConfig` and `MaxSizeConfig` in favor of:

```
type SizeConfig struct {
	Sizer string `mapstructure:"sizer"`
	MinSize int `mapstructure:"mix_size"`
	MaxSize int `mapstructure:"max_size"`
}
```

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/issues/12303


exporter/exporterbatcher/config.go
```diff
@@ -31,6 +31,7 @@ type MinSizeConfig struct {
 	// sent regardless of the timeout. There is no guarantee that the batch size always greater than this value.
 	// This option requires the Request to implement RequestItemsCounter interface. Otherwise, it will be ignored.
 	MinSizeItems int `mapstructure:"min_size_items"`
+	MinSizeBytes int `mapstructure:"min_size_bytes"`
```
bogdandrutu, line 34: What if we have a "Sizer"("SizerType") as enum with 3 values (request, items, bytes) and the "size" value instead?

sfc-gh-sili: Good idea. How about something like this:
```
type BatcherConfig struct {
	...
	minSize SizeConfig `mapstructure:"min_size"`
	maxSize SizeConfig `mapstructure:"max_size"`
}
type SizeConfig struct {
	sizer string `mapstructure:"sizer"`
	size int `mapstructure:"size"`
}
func (c SizerType) validate() {
	...
}
```

Does the option `request` mean that the request will implement its own sizing method?

bogdandrutu: Not sure if we need "sizer" in both min/max, since I don't think we need to accept different sizer for min and max.

sfc-gh-sili: > I don't think we need to accept different sizer for min and max.
Good point. How about
```
type BatcherConfig struct {
	...
	sizer SizeType `mapstructure:",squash"`
	minSize int `mapstructure:"max_size"`
	maxSize int `mapstructure:"max_size"`
}
type SizerType struct {
	sizer string `mapstructure:"sizer"`
}
func (c SizerType) validate() {
}
```

jmacd: Why is "Sizer" a configuration string? What does it name, and how does the user configure it?

sfc-gh-sili: How about

bogdandrutu: We are considering the options for the user `Sizer -> "requests"|"items"|"bytes"`.

bogdandrutu: @sfc-gh-sili still need an enum with possible values for `Sizer` because usages of the `SizeConfig` need to do different things based on different types of sizers.

exporter/exporterbatcher/config.go
```diff
@@ -44,6 +77,11 @@ type MaxSizeConfig struct {
 }
 
 func (c Config) Validate() error {
+	err := c.SizeConfig.Validate()
+	if err != nil {
+		return err
+	}
```
bogdandrutu, line 83: You don't need to call Validate, it is called automatically.

exporter/exporterbatcher/config_test.go
```diff
@@ -29,3 +30,49 @@ func TestConfig_Validate(t *testing.T) {
...
+	cfg := Config{
+		Enabled:      true,
+		FlushTimeout: 200 * time.Millisecond,
+		SizeConfig: SizeConfig{
+			Sizer:   "invalidsizer",
+			MaxSize: 100,
+			MinSize: 100,
+		},
+	}
+	require.EqualError(t, cfg.Validate(), "sizer should either be bytes or items")
```
bogdandrutu, line 44: To test that it is called automatically, use `component.ValidateConfig()`

exporter/exporterbatcher/config.go
sfc-gh-sili: @bogdandrutu Sounds good! I updated the config to have BatchConfig -> SizeConfig -> SizerType. Would you mind taking another look?

exporter/exporterbatcher/config.go
sfc-gh-sili: Makes sense. Thanks!

exporter/exporterbatcher/config.go
```diff
@@ -19,10 +19,29 @@ type Config struct {
...
+	SizeConfig `mapstructure:",squash"`
+
+	// Deprecated. Ignored if SizeConfig is set.
 	MinSizeConfig `mapstructure:",squash"`
+	// Deprecated. Ignored if SizeConfig is set.
 	MaxSizeConfig `mapstructure:",squash"`
 }
 
+type SizeConfig struct {
+	SizerType SizerType `mapstructure:",squash"`
```
bogdandrutu, line 31:
```suggestion
	SizerType SizerType `mapstructure:"sizer"`
```

exporter/exporterbatcher/config.go
```diff
@@ -19,10 +19,29 @@ type Config struct {
...
+
+	MinSize int `mapstructure:"mix_size"`
+	MaxSize int `mapstructure:"max_size"`
+}
+
+type SizerType struct {
+	// Sizer should either be bytes or items.
+	Sizer string `mapstructure:"sizer"`
```
bogdandrutu, line 39: make this private and create an `UnmarshalText` (implement this interface https://pkg.go.dev/encoding#TextUnmarshaler). See https://github.com/open-telemetry/opentelemetry-collector/blob/main/pipeline/internal/globalsignal/signal.go

That way, we can control to only allow specific values.

exporter/exporterbatcher/config.go
```diff
@@ -19,10 +19,29 @@ type Config struct {
...
+}
+
+const SizerTypeItems = "items"
+const SizerTypeBytes = "bytes"
```
bogdandrutu, line 43: they should be `SizerType` type.

exporter/exporterbatcher/config.go
```diff
@@ -59,6 +78,30 @@ func (c Config) Validate() error {
...
+	if c.MaxSize != 0 && c.MaxSize < c.MinSize {
+		return errors.New("max_size must be greater than or equal to mix_size")
+	}
+	return nil
+}
+
+func (c SizerType) Validate() error {
+	if c.Sizer != SizerTypeItems && c.Sizer != SizerTypeBytes {
+		return errors.New("sizer should either be bytes or items")
+	}
```
bogdandrutu, line 101: Don't need this if you do what I suggest.

exporter/exporterbatcher/config.go
sfc-gh-sili: Makes sense.

sfc-gh-sili:
I noticed an existing issue that discusses the config API for queue https://github.com/open-telemetry/opentelemetry-collector/issues/9462 and I think we should align batch config API and queue config API as possible.

Opened an issue https://github.com/open-telemetry/opentelemetry-collector/issues/12303 for further discussion.

jmacd (approved):

exporter/exporterbatcher/config.go
```diff
@@ -59,6 +70,23 @@ func (c Config) Validate() error {
 	return nil
 }
 
+func (c SizeConfig) Validate() error {
+	if c.Sizer.string == "" && c.MinSize == 0 && c.MaxSize == 0 {
+		return nil
+	}
```
dmitryax, line 76: This seems redundant

exporter/exporterbatcher/sizer_type.go
```diff
@@ -0,0 +1,61 @@
...
+
+package exporterbatcher // import "go.opentelemetry.io/collector/exporter/exporterbatcher"
+
+import (
+	"fmt"
+)
+
+type SizerType struct {
+	string
+}
```
dmitryax, line 12: No need to make it struct
```suggestion
type SizerType string
```

exporter/exporterbatcher/sizer_type.go
```diff
@@ -0,0 +1,61 @@
...
+
+const (
+	itemsSizerString = "items"
```
dmitryax, line 15: Then we can expose these instead on the vars:
```suggestion
	SizerTypeItemst SizerType = "items"
```

exporter/exporterbatcher/sizer_type.go
```diff
@@ -0,0 +1,61 @@
...
+	BytesSizer = SizerType{bytesSizerString}
+)
+
+// String converts SizerType to a string
+func (s *SizerType) String() string {
+	return s.string
+}
+
+// NewSizer creates a new sizer. It returns an error if the input is invalid.
+func NewSizer(text string) (SizerType, error) {
```
dmitryax, line 30: No need to expose these constructors along with `MustNewSizer`. Users should use the constants instead. Please take  a look at the pattern we follow in other places of this codebase

exporter/exporterbatcher/sizer_type.go
sfc-gh-sili: That's neat. Thanks!

exporter/exporterbatcher/sizer_type.go
sfc-gh-sili: Thanks! With `SizerType` defined as `string` instead of `struct`, it's much simpler now.
I was referring to `component.Type` that checks the input against a regular expression instead of a finite list - that one provides a lot more APIs than what is needed for this use case.

dmitryax (approved):

Issue #11061, opened 2024-09-05
remove `MetricsLevel` from `TelemetrySettings`


codeboten:
A discussion on 4-Sept-2024 lead to the agreement that the struct field `MetricsLevel` will be removed from `TelemetrySettings`. This field is there to allow component authors to check a level before deciding whether to incur the cost of calculating an expensive operation. 

The alternative for component authors will be to check whether an instrument is enabled or not prior to calculating the metric. This is dependent on the [PR](https://github.com/open-telemetry/opentelemetry-go/pull/5768) in otel go to implement this and on the spec to stabilize this feature.

The decision was that if the change would take too long to implement upstream, then we would remove this field and move any component to configure their own metric level instead. This will allow the unblocking of the component module stabilization work.

mx-psi:
@bogdandrutu Are you going to submit any changes to the spec related to this? (Maybe related to open-telemetry/opentelemetry-specification/issues/4200?)

mx-psi:
Given that stabilization of `Enabled` is not on September's release (see open-telemetry/opentelemetry-specification/pull/4204), I think we should go with the alternative plan, since mid-October would delay us significantly

mx-psi:
Summary of current state: open-telemetry/opentelemetry-specification/issues/4256 and open-telemetry/opentelemetry-specification/issues/4207 are blocking stabilization in the spec. There is a PR for 4256: open-telemetry/opentelemetry-specification/pull/4262 that is blocked by adding a prototype. Nobody is working on a prototype. For 4207, things seem to be moving forward.

This is one of the last issues blocking component 1.0. I have been focusing on #11499 and related issues in case it changes the roadmap and this becomes less important, but given the state above, we may want to go with the approach of "remove MetricsLevel and use feature gates for current use cases"/

mx-psi:
There is also now an OTEP related to one of the issues above: open-telemetry/opentelemetry-specification/pull/4290

mx-psi:
We discussed open-telemetry/opentelemetry-specification/issues/4256 at the 2024-11-19 spec meeting. We will close it on 2024-12-03 unless there are use case brought up based on https://github.com/open-telemetry/opentelemetry-specification/issues/4256#issuecomment-2486242726.

This also would mean that https://github.com/open-telemetry/opentelemetry-specification/issues/4207 is not blocking stabilization, so we may (fingers crossed) be able to merge the stabilization PR.

mx-psi:
[...]

PR #12159, opened 2025-01-22
[component] Deprecate TelemetrySettings.MetricsLevel

mx-psi:
#### Description

- Deprecates `component.TelemetrySettings.MetricsLevel`
- Sets the value passed to components to always be `configtelemetry.LevelDetailed` 

#### Link to tracking issue

Updates #11061


service/service.go
```diff
@@ -173,7 +173,8 @@ func New(ctx context.Context, set Settings, cfg Config) (*Service, error) {
 		Logger:         logger,
 		MeterProvider:  mp,
 		TracerProvider: tracerProvider,
-		MetricsLevel:   cfg.Telemetry.Metrics.Level,
+		// Set MetricsLevel to Detailed during the transition period.
+		MetricsLevel: configtelemetry.LevelDetailed,
```
bogdandrutu, line 177: Why? This may increase metrics for other unintentionally.

mx-psi: It was an attempt to force people to use the new mechanism, but since it can be missed I reverted it

bogdandrutu (approved):

PR #12164, opened 2025-01-22
[chore] Move exporter id attribute to the span

bogdandrutu:
Even though we change the Span attributes, this is ok since tracing is still not stable.

### DIFF
--- a/exporter/exporterhelper/internal/obsexporter.go
+++ b/exporter/exporterhelper/internal/obsexporter.go
@@ -11,7 +11,6 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 
-	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter"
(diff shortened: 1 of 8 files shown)


dmitryax (approved):

exporter/exporterhelper/internal/obsexporter.go
```diff
@@ -111,7 +114,7 @@ func (or *ObsReport) EndProfilesOp(ctx context.Context, numSpans int, err error)
 // the updated context and the created span.
 func (or *ObsReport) startOp(ctx context.Context, operationSuffix string) context.Context {
 	spanName := or.spanNamePrefix + operationSuffix
-	ctx, _ = or.tracer.Start(ctx, spanName)
+	ctx, _ = or.tracer.Start(ctx, spanName, or.spanAttrs)
```
dmitryax, line 117: Shouldn't we have a changelog since it changes the internal span attributes?

bogdandrutu: Done.

Issue #11140, opened 2024-09-11
[exporter/otlp] Review telemetry


mx-psi:
As part of the stabilization of the OTLP exporter, we want to ensure that it has the core self telemetry needed to understand the component and debug common issues. This issue tracks reviewing existing telemetry and defining any new telemetry needed to make the component observable. It should be rare that we add new telemetry as per the 1.0 tenet that "The Collector is already used in production at scale and has been tested in a variety of environments".

## Blockers

- [ ] #12212

jade-guiton-dd:
# `exporter/otlp` and `exporter/otlphttp` telemetry review

Because most of the internal telemetry in these two components is emitted by `exporterhelper`, I will make a joint review of both of them here.

## Review

This review is based on the [observability requirements for stable components](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#observability-requirements).

⚠️ Like the receiver, a lot of this telemetry is not emitted for profiles, only traces/metrics/logs. These components should NOT be considered stable for profiles.

> 1. How much data the component receives.

✅ This should be covered by [pipeline instrumentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md), once implemented.

> 2. How much data the component outputs.

✅ The `otelcol_exporter_sent_(log_records / metric_points / spans)` metric should suffice.

> 3. How much data is dropped because of errors.

✅ When queueing and batching are disabled, this should be covered by pipeline instrumentation, although this is probably a rare case.

✅ Otherwise, the `otelcol_exporter_enqueue_failed / _send_failed / _sent` metrics do a good job of counting errors of different types.

⚠️ The only caveat is that `_send_failed` includes connection errors and errors caused by the Collector shutting down. The former can reasonably be considered downstream errors, and the latter should be rare and easy to distinguish, so I don't think this necessarily needs modification.

> 4. Details for error conditions.

✅ Error messages are variously available as part of emitted spans, span events, and logs.

> 5. Other possible discrepancies between input and output, if any.
>     - How much data is currently held by the component, and how much can be held if there is a fixed capacity.

✅ This is handled by the `otelcol_exporter_queue_size` and `otelcol_exporter_queue_capacity` metrics (both gauges).

> 6. Processing performance.
>     As an application of this, components which hold items in a queue should allow differentiating between time spent processing a batch of data and time where the batch is simply waiting in the queue.
[...]

PR #12165, opened 2025-01-22
Merge observability report sender for all signals, remove duplicate code

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/12164


jade-guiton-dd:
The only behavior change here is renaming span attributes from (taking spans as example) `sent_spans` / `send_failed_span` to a generic `items.sent` / `items.failed`, right? (Just confirming, since I'm in the process (#11140) of reviewing the telemetry we have on the OTLP exporters)

jmacd:
> The only behavior change here is renaming span attributes from (taking spans as example) sent_spans / send_failed_span to a generic items.sent / items.failed, right?

Is that intentional? We can't skip the changelog for a change like that.

bogdandrutu:
@jmacd added changelog

PR #12167, opened 2025-01-23
Generate helper metrics equal functions to simplify asserts

bogdandrutu:

PR #12168, opened 2025-01-23
[mdatagen] Fix bug where Histograms were marked as not supporting temporaly aggregation

bogdandrutu:
See https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L254


codeboten (approved):
surprised this change doesn't modify any generated code, i guess no components in core are using histograms?

bogdandrutu:
Yes, very rarely used. I found it because I am trying to see if we can simplify the metrics testing, see https://github.com/open-telemetry/opentelemetry-collector/pull/12167. But for telemetry so far we don't test temporality.

PR #12170, opened 2025-01-23
[chore] Fix MergeSplit logic, update tests to use Request

bogdandrutu:
No need for a changelog since the cache items was added this release.


exporter/exporterhelper/logs_batch.go
```diff
@@ -24,6 +24,8 @@ func (req *logsRequest) MergeSplit(_ context.Context, cfg exporterbatcher.MaxSiz
 	}
 
 	if cfg.MaxSizeItems == 0 {
+		req.setCachedItemsCount(req.ItemsCount() + req2.ItemsCount())
+		req2.setCachedItemsCount(0)
```
jmacd, line 28: This makes me wonder what the real benefit of having a cached items count is? How many times do we expect that value to be read, aside from once by the observability sender?

jmacd (commented):
Can we eliminate cached item counts? I wonder if it's a premature optimization.

bogdandrutu:
> Can we eliminate cached item counts? I wonder if it's a premature optimization.

I would prefer not to https://github.com/open-telemetry/opentelemetry-collector/pull/12136

dmitryax (approved):

exporter/exporterhelper/logs_batch.go
sfc-gh-sili: Hi Joshua! This optimization comes from the observation that `ItemsCount()` is repeatedly called during merge-splitting and is a fairly expensive operation. The optimization has proved to save 30 - 75% cpu time for merging based on the benchmark (see more details in https://github.com/open-telemetry/opentelemetry-collector/pull/12136)

I understand your concern around having an additional variable to book-keep could be error-prone, but I would argue that no one, except for the batcher, should be messing around with the request once its created. We just need to make sure that the cached items count is updated when the wrapped proto is updated.

PR #12176, opened 2025-01-23
Add FromAttributeIndices helper to pprofile

dmathieu:
#### Description

Profiles attributes are stored in a single `AttributeTable` attribute, and then referenced in each substruct as `AttributeIndices`.

This means to read them as a `pcommon.Map`, a bit of pre-processing is required.
This adds an helper method so each component manipulating profiles doesn't have to reimplement it.

Benchmark:

```
go test -bench=BenchmarkFromAttributeIndices ./...
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/collector/pdata/pprofile
cpu: Apple M1 Max
BenchmarkFromAttributeIndices-10         9789714               123.2 ns/op           140 B/op          4 allocs/op
PASS
ok      go.opentelemetry.io/collector/pdata/pprofile    1.952s
PASS
ok      go.opentelemetry.io/collector/pdata/pprofile/pprofileotlp       0.407s
```


pdata/pprofile/attributes.go
```diff
@@ -0,0 +1,27 @@
...
+type attributable interface {
+	AttributeIndices() pcommon.Int32Slice
+}
+
+// BuildAttributes builds a [pcommon.Map] containing the attributes of a record.
+// The record can by any struct that implements an `AttributeIndices` method.
+func BuildAttributes(profile Profile, record attributable) (pcommon.Map, error) {
+	attrs := map[string]any{}
+
+	for i := range record.AttributeIndices().AsRaw() {
```
bogdandrutu, line 19: Unnecessary slice allocation, avoid everywhere AsRaw and FromRaw and add a benchamark.

bogdandrutu (changes requested):

pdata/pprofile/attributes.go
dmathieu: Thank you. I've added the benchmark and switched to manual type setting. There's still 9 allocations though. Happy to look into any improvement proposal.

pdata/pprofile/attributes.go
```diff
@@ -0,0 +1,50 @@
...
+
+type attributable interface {
+	AttributeIndices() pcommon.Int32Slice
+}
+
+// BuildAttributes builds a [pcommon.Map] containing the attributes of a record.
+// The record can by any struct that implements an `AttributeIndices` method.
+func BuildAttributes(profile Profile, record attributable) pcommon.Map {
+	m := pcommon.NewMap()
+
```
bogdandrutu, line 18: m.EnsureCapacity(record.AttributeIndices().Len())

pdata/pprofile/attributes.go
```diff
@@ -0,0 +1,50 @@
...
+	for i := 0; i < record.AttributeIndices().Len(); i++ {
+		id := int(record.AttributeIndices().At(i))
+		a := profile.AttributeTable().At(id)
+
+		key := a.Key()
+		val := a.Value()
+
+		switch val.Type() {
```
bogdandrutu, line 26: Simply replace whole switch statement with: `val.CopyTo(m.PutEmpty(key))`

pdata/pprofile/attributes.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// BuildAttributes builds a [pcommon.Map] containing the attributes of a record.
+// The record can by any struct that implements an `AttributeIndices` method.
+func BuildAttributes(profile Profile, record attributable) pcommon.Map {
+	m := pcommon.NewMap()
+
+	for i := 0; i < record.AttributeIndices().Len(); i++ {
+		id := int(record.AttributeIndices().At(i))
+		a := profile.AttributeTable().At(id)
```
bogdandrutu, line 21: kv := profile.AttributeTable().At(int(record.AttributeIndices().At(i)))

pdata/pprofile/attributes.go
```diff
@@ -0,0 +1,50 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+type attributable interface {
+	AttributeIndices() pcommon.Int32Slice
+}
+
+// BuildAttributes builds a [pcommon.Map] containing the attributes of a record.
+// The record can by any struct that implements an `AttributeIndices` method.
+func BuildAttributes(profile Profile, record attributable) pcommon.Map {
```
bogdandrutu, line 16: I would probably use something like `FromAttributeIndices`?

bogdandrutu, line 16: Should this be a func on `Profile` then?

dmathieu: I don't mind moving this into Profile, but mixing generated code within non-generated usually tends to harder to read code.
I've switched to passing only the `AttributeTable`.

bogdandrutu (commented):
This looks good to me in general, would like other to specify where this function should live.

Couple things:
1. Did you update benchmarks after latest changes?
2. We should document that changes to the return map are not reflected to the profile. Would like to understand how you think we should design the API that can do changes.

dmathieu:
> Did you update benchmarks after latest changes?

Sorry, they're updated now.

> We should document that changes to the return map are not reflected to the profile. Would like to understand how you think we should design the API that can do changes.

I've added a mention that updates on the map will not change the record.
Making changes is a bit tricky, as we have to check the AttributeTable to find existing entries, possibly add new ones.
And we won't be able to remove data that's not used anymore, as we can't know it's not used by other records (that shouldn't matter much though).

We could have something like:

```golang
ToAttributeIndices(data pcommon.Map, table AttributeTable, record attributable)
```

That would properly update the AttributeTable with new entries, as well as the attributable indices.
That would be a full replace. So we could also have a single attribute add:

```golang
Addattribute(key, value string, table AttributeTable, record attributable)
```

bogdandrutu (approved):
PR looks good. But I want you to add an issue and mark it as blocker for stabilizing profiles to find a solution for updates and try it before stabilizing this package.

bogdandrutu:
> Sorry, they're updated now.

Numbers look much better than at the beginning 👯

.chloggen/pprofile-build-attributes.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: pdata/pprofile
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add new helper method `BuildAttributes` to build a `pcommon.Map` out of `AttributeIndices`.
```
dmathieu, line 10:
```suggestion
note: Add new helper method `FromAttributeIndices` to build a `pcommon.Map` out of `AttributeIndices`.
```

Issue #12000, opened 2024-12-31
Regression: Error message for duplicate keys in configuration


tsloughter:
**Describe the bug**

A duplicate key in the configuration like this duplicate use of `http`:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: "0.0.0.0:4317"
      http:
        endpoint: "0.0.0.0:4318"
      http:
        cors:
          allowed_origins:
          - "http://localhost:8000"
        endpoint: "0.0.0.0:4318"
```

when run with `main` results in the error message:

```shell
$ bin/otelcorecol_linux_amd64 --config ../roll_dice_gleam/otel-collector-config.yaml
Error: failed to get config: cannot resolve the configuration: retrieved value (type=string) cannot be used as a Conf
2024/12/31 10:58:11 collector server run finished with error: failed to get config: cannot resolve the configuration: retrieved value (type=string) cannot be used as a Conf
```

This makes no mention of where the issue in the file was found or that it was a duplicate key.

Compare to what the error message was as of commit `334d95aebcffc88fa19ac4233dc166dd51aaa55f`:

```shell
Error: failed to resolve config: cannot resolve the configuration: cannot retrieve the configuration: yaml: unmarshal errors:
  line 9: mapping key "http" already defined at line 7
2024/12/31 10:53:08 collector server run finished with error: failed to resolve config: cannot resolve the configuration: cannot retrieve the configuration: yaml: unmarshal errors:
  line 9: mapping key "http" already defined at line 7
```

The line number and the key `http` was given and that it was a duplicate issue.

**Steps to reproduce**

Run a recent collector with the configuration like the above.

**What did you expect to see?**

The previous error message as shown above.

**What did you see instead?**

An error message that does not give details on the actual issue.

**What version did you use?**

does't work: ffcef93b05175c21f967d43af95141e95bdb2f3d
works: 334d95aebcffc88fa19ac4233dc166dd51aaa55f

**What config did you use?**

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: "0.0.0.0:4317"
      http:
        endpoint: "0.0.0.0:4318"
      http:
        cors:
          allowed_origins:
          - "http://localhost:8000"
        endpoint: "0.0.0.0:4318"

exporters:
  debug:
    verbosity: detailed
    sampling_initial: 1
    sampling_thereafter: 1

  pipelines:
    traces:
      receivers: [otlp]
      processors: [h]
[...]

PR #12180, opened 2025-01-24
[confmap] Add error hint when invalid YAML was passed

mx-psi:
#### Description

Add error hint field to surface YAML parsing errors. This only surfaces the error for top-level URIs but it's a step in the right direction. 

Error when passing a file with contents `[invalid:,` as the config:

Before:
```
retrieved value (type=string) cannot be used as a Conf
```

After: 
```
retrieved value (type=string) cannot be used as a Conf: assuming string type since contents are not valid YAML: yaml: line 1: did not find expected node content
```

#### Link to tracking issue
Updates #12000


mattsains:
One thing I noticed a lot while researching this issue is that a lot of the test cases only test that some kind of error was produced, and not the content of the error. Perhaps you could consider fixing that as part of this PR. Examples:

* https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/collector_test.go#L315C1-L316C1
* https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider_test.go#L64
* https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/provider_test.go#L80

Improving these tests would prevent a regression like this in the future, they're easy to find by searching for `require.Error` and easy to fix by changing them to `require.ErrorEqual(t, err, "expected message")`. You can even just run the tests with the expected message set to blank and figure out what the expected error is.

mx-psi:
@mattsains That makes sense. I am not going to work on this further at least until next week, if you want to file a PR adding the `EqualError`s or continue on top of this branch feel free to do so!

ms-jcorley (approved):

mx-psi:
@mattsains Can you take another look?

mattsains (approved):
looks good to me! Although I'm not a maintainer on this repo

mattsains:
Not stale, needs approver

jade-guiton-dd (approved):

PR #12182, opened 2025-01-25
[mdatagen] Generate wrappers for recording telemetry metrics

crobert-1:
#### Description
This is essentially reviving #10911, the first step towards #10801. The goal is to be able to use `mdatagen` to handle more use cases around telemetry metrics, similar to regular metrics, especially to be able to set attributes on the telemetry metrics.

#### Link to tracking issue
Part of #10801

#### Testing
Code has been generated for different telemetry metric types, existing tests use this behavior pretty heavily.


processor/batchprocessor/internal/metadata/generated_telemetry.go
```diff
@@ -98,3 +98,19 @@ func NewTelemetryBuilder(settings component.TelemetrySettings, options ...Teleme
...
+	builder.ProcessorBatchBatchSendSizeBytes.Record(ctx, val, opts...)
+}
+
+func (builder *TelemetryBuilder) RecordProcessorBatchBatchSizeTriggerSend(ctx context.Context, val int64, opts ...metric.AddOption) {
+	builder.ProcessorBatchBatchSizeTriggerSend.Add(ctx, val, opts...)
+}
+
+func (builder *TelemetryBuilder) RecordProcessorBatchTimeoutTriggerSend(ctx context.Context, val int64, opts ...metric.AddOption) {
+	builder.ProcessorBatchTimeoutTriggerSend.Add(ctx, val, opts...)
+}
```
bogdandrutu, line 116: Few things here:
1. We will confuse the user when using Add/Record because we don't respect the same semantics as otel.
2. What is the next level for this PR? Can you just provide a sketch code that you want to end with, please don't make the code changes, just provide a draft code for me to understand.

crobert-1: > 1. We will confuse the user when using Add/Record because we don't respect the same semantics as otel.

Good call out. My thought process was to make it as simple and consistent for the user as possible. This also matches `mdatagen`'s functionality for metrics, which always have `Record` as the prefix irrespective of metric type.

> 2. What is the next level for this PR? Can you just provide a sketch code that you want to end with, please don't make the code changes, just provide a draft code for me to understand.

The goal is to bring up internal telemetry functionality to the same level as receiver metrics using `mdatagen`. My intention would be to make the metric itself private to the `internal/metadata` package within a component, then only expose the `Record` method.

Something like:
```
type AttributeFoo string

type metricProcessorBatchBatchSendSize struct {
}

type TelemetryBuilder {
  metricProcessorBatchBatchSendSendSize metricProcessorBatchBatchSendSendSize
}

func (builder *TelemetryBuilder) RecordProcessorBatchBatchSendSendSize(ctx context.Context, val int64, foo string) {
}
```
Going this direction allows us to implement support for the [documented schema](https://github.com/open-telemetry/opentelemetry-collector/blob/c4f1c3f58ef8807501227284e3fc33ea5717826a/cmd/mdatagen/metadata-schema.yaml#L135) of mdatagen, especially telemetry metric attributes, and enforcing the set value for `enabled` in the schema.

crobert-1:
Apologies for letting this go stale, I've updated.

@bogdandrutu I've replied to your question, can you review again when you get a chance?

processor/batchprocessor/internal/metadata/generated_telemetry.go
dmitryax: 1. @bogdandrutu, do you want to use `Add<MetricName>` for sum type? That sounds reasonable. We don't have `DataPoint` suffix here as we have in the normal builder, and that makes sense. So we can probably provide an interface closer to the go SDK instead.

bogdandrutu: >The goal is to bring up internal telemetry functionality to the same level as receiver metrics using mdatagen. My intention would be to make the metric itself private to the internal/metadata package within a component, then only expose the Record method.

I would go with that. Why do we need a different API though? For the other metrics we have a helper builder because otherwise is hard to construct the whole metrics, and because we want support aggregations and other stuff, but for OTEL that is already supported?

crobert-1: I've changed `Record` to `Add` for `sum` telemetry metrics.

PR #12189, opened 2025-01-27
[component] Remove TelemetrySettings.MetricsLevel

### LINKED ISSUE #11061: remove `MetricsLevel` from `TelemetrySettings`

mx-psi:
#### Description

Fixes #11061.

Depends on releasing the deprecation.


mx-psi:
Needs open-telemetry/opentelemetry-collector-contrib/pull/37648

bogdandrutu (approved):

PR #12203, opened 2025-01-28
Change receiver shared component part

bogdandrutu:


receiver/otlpreceiver/grpc.go
```diff
@@ -0,0 +1,85 @@
...
+	tel        component.TelemetrySettings
+	shutdownWG sync.WaitGroup
+	servers    []registerServerFunc
+}
+
+func newSharedGRPC(cfg *configgrpc.ServerConfig, tel component.TelemetrySettings) (*sharedcomponent.Component[*grpcComponent], error) {
+	return grpcs.LoadOrStore(
+		cfg,
+		func() (*grpcComponent, error) {
+			return &grpcComponent{cfg: cfg, tel: tel}, nil
```
djaglowski, line 35: When you pass in `tel` here, I assume you want it to _not_ have an `otel.signal` attribute on the logger, etc?

This still leaves us in a situation where it needs to be present sometimes and not others. What's the mechanism for managing this?

bogdandrutu: My hope is that, now that we know a component has to create different "types" of sub-component, the mechanism we design will allow that. Very important is that the "main" component still has all the "attributes".

The easiest way I can think of is via the TelemetrySettings we provide, we can have there a mechanism of retrieving the Logger via some functions instead of direct access to 1 variable.

PR #12204, opened 2025-01-29
Make registration of callback for async metric always optional

bogdandrutu:
This PR also solves one more issues with the new interface, which is the ability to record multiple observations for a single async metric using one callback.


dmitryax (approved):
LGTM

Issue #11443, opened 2024-10-14
Remove `GetExtensions` from `component.Host`


bogdandrutu:
Remove `GetExtensions` from `component.Host`. This should probably be an "optional" interface in the Extension package, so that it can return `GetExtensions() map[component.ID]extension.Extension`.

mx-psi:
@bogdandrutu I feel like this should stay in `component.Host`: this feels like *the* essential functionality a `component.Host` must have in order to support basic use cases like authentication, encoding or storage configuration.

jpkrohling:
Isn't this what we use in confighttp and configgrpc in order to get the available auth extensions?

mx-psi:
@jpkrohling This would still be available, just as an optional interface.

Right now:

```go

var host component.Host = getComponentHostSomehow()

doSomethingWithExtensions(host.GetExtensions())
```

After the change proposed by @bogdandrutu we would have an interface like this:

```go
type GetExtensionsInterface interface {
  GetExtensions() map[[component.ID](component.Component)   
}
```

and you would use it like this:

```go

var host component.Host = getComponentHostSomehow()

getExtHost, ok := host.(GetExtensionsInterface)

if !ok {
  panic("host does not support GetExtensions")
}

doSomethingWithExtensions(host.GetExtensions())
```

This would be similar to [`componentstatus.Reporter`](https://pkg.go.dev/go.opentelemetry.io/collector/component/componentstatus#Reporter)

jpkrohling:
Got it -- I'm still leaning towards having it as part of the main interface, unless there's a concrete reason to not have it there. I agree with your previous statement:

> this feels like the essential functionality

jade-guiton-dd:
One possible argument for making this interface optional is that it gives us more flexibility to change the method signature later, even after stabilization, for instance if returning a `map` no longer seems like the right choice. (Considering we removed `GetExporters` from `Host` way back it doesn't seem implausible.) Instead of supporting the old signature forever, it could be removed without completely breaking components relying on it, though at the cost that these components will likely assume there are no extensions set up. On the other hand, if we're fine with that last part, keeping the old signature around with a dummy implementation would not be a difficult alternative either... for that reason, I don't have strong feelings one way or another.

TylerHelmuth:
This functionality is the is only concrete function on the host interface. I also agree is it the core feature that host is providing to components. If we removed it what would we do with the host interface?

What problem is being solved by removing it?

[...]

PR #12208, opened 2025-01-29
[extension] Change from interface alias to embedding alias

mx-psi:
#### Description

Changes `extension.Extension` to a proper interface that embeds `component.Component` instead of being an alias.

#### Link to tracking issue
Fixes the second point mentioned in https://github.com/open-telemetry/opentelemetry-collector/issues/11443#issuecomment-2616396687


mx-psi:
@bogdandrutu I think we can do this without much of an issue, I had to file open-telemetry/opentelemetry-collector-contrib/pull/37568 but AFAICT that should be all

bogdandrutu (approved):

PR #12214, opened 2025-01-30
[component] Make component.Kind a struct

### LINKED ISSUE #11443: Remove `GetExtensions` from `component.Host`

mx-psi:
#### Description

Changes the underlying type of `component.Kind` to be closer to [`pipeline.Signal`](https://pkg.go.dev/go.opentelemetry.io/collector/pipeline#Signal). Right now the type is defined within `component`, but it could be moved to an internal module so that we could have a different module exposing other value on this enum.

This is already doable today, the only thing this PR gives us is 
1. slightly more flexibility on things like making the concept of kind more complex (e.g. adding an adjective to a kind).
2. restricting external consumers from implementing their own component kinds without our explicit approval (with some kind of API we design)

I am not convinced this is _necessary_ to do, but we may as well do it.

This is technically a breaking change since `component.Kind(42)` was a valid expression and it no longer is. I think this is extremely rare, so I suggest if we go ahead we do so in one go.

#### Link to tracking issue
Fixes #11443

### DIFF
--- a/.chloggen/mx-psi_component-kind-structify.yaml
+++ b/.chloggen/mx-psi_component-kind-structify.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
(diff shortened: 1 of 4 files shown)


mx-psi:
cc @bogdandrutu do you think we should do this?

bogdandrutu (approved):

PR #12220, opened 2025-01-31
Make MakeFactoryMap generic and move to otelcol

### LINKED ISSUE #12222: [extension/receiver/exporter/connector/processor] Move `MakeFactoryMap` to `otelcol`
The `MakeFactoryMap` functions are (i) redundant after the introduction of generics and (ii) a concept only useful if you have an `otelcol`. We should move them to `otelcol`.

From the 2025-01-27 Collector stability working group meeting.

jade-guiton-dd:
#### Description

At the moment, `receiver`, `scraper`, `processor`, `exporter`, and `extension` each have their own version of the `MakeFactoryMap` function, which takes a list of Factories and returns it as a map, with the component name as the key (and an error if there is overlap). Because all versions are near-identical except for the Factory type used, and this function is only used in tests and in the code generated by OCB (`components.go`), it was suggested to make it generic and move it to the `otelcol` package.

This PR does exactly that. The old `MakeFactoryMap` functions are deprecated (and may be removed alongside their tests in a future release). I also had to make a few other odd changes to fit the new dependency graph.

#### Link to tracking issue
Resolves #12222


jade-guiton-dd:
Okay I think we have an infinite loop of `make crosslink` adds a `replace`, `make genotelcorecol` removes it, will think about this harder next week

mx-psi (approved):

mx-psi:
@bogdandrutu will merge this on Friday, this seems fine to me but since otelcol is far from being stabilized we can iterate on the function if we need to

@jade-guiton-dd can you solve the merge conflicts?

service/internal/builders/builders.go
```diff
@@ -13,7 +13,7 @@ import (
 
 var (
 	errNilNextConsumer = errors.New("nil next Consumer")
-	nopType            = component.MustNewType("nop")
+	NopType            = component.MustNewType("nop")
```
bogdandrutu, line 16: Why this change?

bogdandrutu (commented):
The pattern to avoid circular deps that I've seen is to move tests in a package "builder_test" in the same directory.

service/internal/builders/builders.go
jade-guiton-dd: I moved the tests into the separate `builderstest` package to avoid circular dependencies, and this variable is used in the tests, so I needed to make it exported to keep it accessible. Note that all of this is still under `internal`, so I don't think this expands the public API.

jade-guiton-dd: I based the `builderstest` naming convention on `componenttest`, `connectortest`, `consumertest`, etc. Do you think I should make it a child of `builders` instead of a sibling?

mx-psi:
I want us to be able to mark extension as 1.0 as soon as we mark component 1.0, so I am marking this as release blocker for the next release. I believe the changes discussed are all internal, so this is more of a reminder.

service/internal/builders/builders.go
bogdandrutu: builderstest -> is the test files, not testing helpers

connector/connector.go
```diff
@@ -388,6 +388,8 @@ func NewFactory(cfgType component.Type, createDefaultConfig component.CreateDefa
 
 // MakeFactoryMap takes a list of connector factories and returns a map with factory type as keys.
 // It returns a non-nil error when there are factories with duplicate type.
+//
+// Deprecated: [v0.119.0] Use otelcol.MakeFactoryMap[connector.Factory] instead
```
codeboten, line 392:
```suggestion
// Deprecated: [v0.120.0] Use otelcol.MakeFactoryMap[connector.Factory] instead
```

exporter/exporter.go
```diff
@@ -190,6 +190,8 @@ func NewFactory(cfgType component.Type, createDefaultConfig component.CreateDefa
 
 // MakeFactoryMap takes a list of factories and returns a map with Factory type as keys.
 // It returns a non-nil error when there are factories with duplicate type.
+//
+// Deprecated: [v0.119.0] Use otelcol.MakeFactoryMap[exporter.Factory] instead
```
codeboten, line 194:
```suggestion
// Deprecated: [v0.120.0] Use otelcol.MakeFactoryMap[exporter.Factory] instead
```

extension/extension.go
```diff
@@ -82,6 +82,8 @@ func NewFactory(
 
 // MakeFactoryMap takes a list of factories and returns a map with Factory type as keys.
 // It returns a non-nil error when there are factories with duplicate type.
+//
+// Deprecated: [v0.119.0] Use otelcol.MakeFactoryMap[extension.Factory] instead
```
codeboten, line 86:
```suggestion
// Deprecated: [v0.120.0] Use otelcol.MakeFactoryMap[extension.Factory] instead
```

codeboten (commented):
Change looks good, just need to update the deprecation version

service/internal/builders/builders.go
jade-guiton-dd: Ah I see, so `xtest` usually contains test helpers rather than tests of `x`. Do you have a suggestion for what the test package should be named then?

jade-guiton-dd: I renamed it that way. Since this test package isn't referenced anyway we can easily rename it again later if this wasn't what you had in mind.

mx-psi (approved):
Re-approving after the move to `builders_test`

PR #12224, opened 2025-01-31
[chore] Remove manual calls to `Validate`

evan-bradley:
#### Description

Follow-up to https://github.com/open-telemetry/opentelemetry-collector/pull/12108.

This calls `component.ValidateConfig` on `otelcol.Config`, which causes `Validate` to be recursively called on all eligible structs in the config.

This has the following benefits:
1. `Validate` will now automatically be called instead of config structs needing to call it on member structs.
2. Paths to the exact key that caused a validation error in the config are now consistently printed.
3. If there are multiple errors in the config, they are now all returned to the user instead of just the first error encountered.

### DIFF
--- a/otelcol/collector.go
+++ b/otelcol/collector.go
@@ -171,7 +171,7 @@ func (col *Collector) setupConfigurationComponents(ctx context.Context) error {
 		return fmt.Errorf("failed to get config: %w", err)
 	}
 
-	if err = cfg.Validate(); err != nil {
+	if err = component.ValidateConfig(cfg); err != nil {
 		return fmt.Errorf("invalid configuration: %w", err)
 	}
 
@@ -261,7 +261,7 @@ func (col *Collector) DryRun(ctx context.Context) error {
 		return fmt.Errorf("failed to get config: %w", err)
 	}
 
-	return cfg.Validate()
+	return component.ValidateConfig(cfg)
 }
 
 func newFallbackLogger(options []zap.Option) (*zap.Logger, error) {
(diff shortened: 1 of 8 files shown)


evan-bradley:
Here's what this looks like with the following config:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: localhost:28973
  otlp/2:
    protocols:
      # http:
      #   endpoint: localhost:26726

processors:
  memory_limiter:

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    logs:
      receivers: [otlp]
      processors: [memory_limiter, memory_limiter]
      exporters: [debug]
    logs/2:
      receivers: [otlp/2]
      exporters: [debug]
```

**Old output:**

```
Error: invalid configuration: receivers::otlp/2: must specify at least one protocol when using the OTLP receiver
2025/01/31 10:30:41 collector server run finished with error: invalid configuration: receivers::otlp/2: must specify at least one protocol when using the OTLP receiver
```

**New output:**

```
Error: invalid configuration: receivers::otlp/2: must specify at least one protocol when using the OTLP receiver
processors::memory_limiter: 'check_interval' must be greater than zero
service::pipelines::logs: references processor "memory_limiter" multiple times
2025/01/31 10:44:22 collector server run finished with error: invalid configuration: receivers::otlp/2: must specify at least one protocol when using the OTLP receiver
processors::memory_limiter: 'check_interval' must be greater than zero
service::pipelines::logs: references processor "memory_limiter" multiple times
```

Errors created with `errors.Join` are printed separated by newlines, but I can try to manually replace them with semicolons like `multierr` does or use some other way of delimiting them, if we would like.

mx-psi (approved):

PR #12226, opened 2025-01-31
[xconfmap] Create module and add validation facilities

evan-bradley:
#### Description

Builds on https://github.com/open-telemetry/opentelemetry-collector/pull/12224 and starts the move of config validation from component to confmap. We can keep this in `xconfmap` while we determine whether to add the ability to validate using struct field tags.

I think this has the following advantages:
1. Everything configuration-related is now in confmap instead of split between confmap and component.
2. We can share things like the mapstructure tag and config key separator as constants between unmarshaling and validation without creating dependencies between confmap and component.

~The one uncertainty this creates is what to do with `component.Config`, which would now be used as a thin alias for `any` without any meaningful usage in component.~

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/11524


evan-bradley:
The relevant changes are in this commit: [`aa1b14c` (#12226)](https://github.com/open-telemetry/opentelemetry-collector/pull/12226/commits/aa1b14c4d7f350a96c7f2068e6eb43d2a0706a56)

confmap/config.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// Config represents an interface for a configuration struct.
+//
+// Implementations and/or any sub-configs (other types embedded or included in the Config implementation)
+// MUST implement the Validator if any validation is required for that part of the configuration
+// (e.g. check if a required field is present).
+//
+// A valid implementation MUST pass the check componenttest.CheckConfigStruct (return nil error).
+type Config any
```
evan-bradley, line 21: This is the one part of this PR I'm not as clear on. Do we want this type? If we do, should it replace `component.Config`?

Just to demonstrate what this would look like, I used this type in `Validate`, `Unmarshal`, and `Marshal` within confmap. It's just an alias to `any`, so there's no real functional difference, but it could help annotate which functions expect a config struct.

evan-bradley: We discussed this point during the 2025-02-03 Collector stability meeting and determined that we agree with the above and will keep `component.Config` and simply use `any` in the `confmap` validation functions.

confmap/confmap.go
```diff
@@ -24,6 +24,12 @@ const (
 	KeyDelimiter = "::"
 )
 
+const (
+	// MapstructureTag is the struct field tag used to record marshaling/unmarshaling settings.
+	// See https://pkg.go.dev/github.com/go-viper/mapstructure/v2 for supported values.
+	MapstructureTag = "mapstructure"
```
evan-bradley, line 30: We could just hardcode the tag name in `xconfmap` for now, but I figure it's a good time to make this an exported constant.

mx-psi: We could also move this to an `confmap/internal` package I guess, if we want to avoid exposing this

mx-psi (approved):

bogdandrutu (approved):

PR #12231, opened 2025-02-01
Fix context propagation for DisabledBatcher

bogdandrutu:
Finishes the implementation started in https://github.com/open-telemetry/opentelemetry-collector/pull/12225

### DIFF
--- a/.chloggen/disabled-batcher-context.yaml
+++ b/.chloggen/disabled-batcher-context.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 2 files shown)


sfc-gh-sili:
Thank you, Bogdan!
I agree that using the callback function is a cleaner approach compared to using an index. However, I'm not sure if we can handle partial error handling with a callback.

For example, consider this scenario: we start with a batch that contains request `x`, then we received another request `y` and this resulted in a split. Then, we have
* Batch A that contains request x and part of request y
* Batch B that contains the other part of request y

If batch A was sent out successfully but batch B failed, we still propagate the failure back to the queue?

dmitryax:
@bogdandrutu please rebase

bogdandrutu:
@dmitryax done

dmitryax (approved):

PR #12233, opened 2025-02-01
Remove old push based batcher

bogdandrutu:
See https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate#feature-lifecycle


exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -21,17 +21,9 @@ import (
 	"go.opentelemetry.io/collector/exporter/exporterbatcher"
 	"go.opentelemetry.io/collector/exporter/exporterqueue" // BaseExporter contains common fields between different exporter types.
 	"go.opentelemetry.io/collector/exporter/internal"
-	"go.opentelemetry.io/collector/featuregate"
 	"go.opentelemetry.io/collector/pipeline"
 )
 
-var usePullingBasedExporterQueueBatcher = featuregate.GlobalRegistry().MustRegister(
-	"exporter.UsePullingBasedExporterQueueBatcher",
-	featuregate.StageBeta,
```
bogdandrutu, line 30: In the previous release was alpha, turned to beta in this release. Decided to remove this sooner, since we want to complete the exportehelper this cycle.

dmitryax:
Do we have a replacement for batching-only behavior? I don't think we initialize a 0-sized blocking queue if the queue is disabled and only batching is enabled... At least the queue code path isn't executed as far as I can tell.

exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -101,23 +100,8 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, osf ObsrepSe
 		be.QueueSender = NewQueueSender(q, be.Set, be.queueCfg.NumConsumers, be.ExportFailureMessage, be.Obsrep, be.BatcherCfg)
 	}
 
-	if !usePullingBasedExporterQueueBatcher.IsEnabled() && be.BatcherCfg.Enabled ||
-		usePullingBasedExporterQueueBatcher.IsEnabled() && be.BatcherCfg.Enabled && !be.queueCfg.Enabled {
-		bs := NewBatchSender(be.BatcherCfg, be.Set)
```
dmitryax, line 106: Iven with the feature gate set to Stable, we still have this condition `be.BatcherCfg.Enabled && !be.queueCfg.Enabled` because we haven't replaced the batching-only use case

exporter/exporterhelper/internal/base_exporter.go
bogdandrutu: Because it was alpha:

> Features that prove unworkable in the alpha stage may be discontinued without proceeding to the beta stage. Instead, they will proceed to the deprecated stage, which will feature is permanently disabled. A feature gate will be removed once it has been deprecated for at least 2 releases of the collector.

dmitryax: But this feature hasn't proved unworkable. We still want to proceed with this, right?

jmacd (approved):

bogdandrutu:
Replaced by https://github.com/open-telemetry/opentelemetry-collector/pull/12425

PR #12122, opened 2025-01-19
[exporter][queuebatcher] set feature gate UsePullingBasedExporterQueueBatcher to beta

sfc-gh-sili:
#### Description

This PR sets feature gate UsePullingBasedExporterQueueBatcher to beta.

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/issues/8122

PR #12242, opened 2025-02-02
Change queue to embed the consumers

bogdandrutu:
This PR will facilitate:
1. The implementation of the disabled queue, can be a simpler call into the "batcher" with the done callback setup and block until that is called.
2. The implementation of the disabled batcher becomes trivial, and does not suffer from the limitation that only one reader of the queue is setup. Even though the request was sent async by multiple go-routines, still one reader which may have been a bottleneck.

In a separate PR will try to get rid of readableQueue. Initial version of the PR had "read" as a private func on Queue, but the "unused" linter complain that `boundedMemoryQueue.read` was unused.

Technically this is a breaking change, but Queue is an experimental API that we allow changes.


exporter/internal/queue/default_batcher_test.go
```diff
@@ -288,18 +304,22 @@ func TestDefaultBatcher_Shutdown(t *testing.T) {
 	batchCfg.MinSizeItems = 10
 	batchCfg.FlushTimeout = 100 * time.Second
 
+	ba, err := NewBatcher(batchCfg,
+		func(ctx context.Context, req internal.Request) error { return req.Export(ctx) },
+		2)
+	require.NoError(t, err)
+
+	// TODO: https://github.com/open-telemetry/opentelemetry-collector/issues/12244
```
dmitryax, line 312: Should we revert https://github.com/open-telemetry/opentelemetry-collector/pull/12122 until this is fixed?

dmitryax (approved):
Looks good to me

exporter/internal/queue/default_batcher_test.go
bogdandrutu: I was thinking the same.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/pull/12246

PR #12245, opened 2025-02-02
Implement sync disabled queue

bogdandrutu:
This PR implements a version of a queue that can be used when users want to not have an async behavior and want to block until the request is send by the exporter.

This implementation is only used when batching is enabled, otherwise if both batching and queueing are disabled no queue or batch is configured.

**Benchmark results:**
```
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/collector/exporter/exporterqueue
cpu: Apple M2 Max
BenchmarkDisabledQueueOffer
BenchmarkDisabledQueueOffer-12    	 5839711	       187.5 ns/op	      24 B/op	       1 allocs/op
```

### DIFF
--- a/exporter/exporterqueue/disabled_queue.go
+++ b/exporter/exporterqueue/disabled_queue.go
@@ -1,0 +1,67 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package exporterqueue // import "go.opentelemetry.io/collector/exporter/exporterqueue"
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+
+	"go.opentelemetry.io/collector/component"
+)
+
+var donePool = sync.Pool{
+	New: func() any {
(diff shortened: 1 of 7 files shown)


dmitryax:
Please rebasePlease rebase

exporter/exporterqueue/disabled_queue.go
```diff
@@ -0,0 +1,55 @@
...
+	ch := chanPool.Get().(chan struct{})
+	defer chanPool.Put(ch)
+	done := func(err error) {
+		ch <- struct{}{}
+	}
+	d.size.Add(1)
+	d.consumeFunc(ctx, req, done)
+	<-ch
+	d.size.Add(-1)
+	return nil
```
jmacd, line 44: Shouldn't there be an error returned here, somehow?
(see also #11947)

bogdandrutu: Fixed this, thanks for catching.

dmitryax (approved):
LGTM

PR #12248, opened 2025-02-02
[chore] remove test only request impl for profile, use requesttest

bogdandrutu:


exporter/exporterhelper/xexporterhelper/profiles_batch_test.go
```diff
@@ -125,13 +125,6 @@ func TestMergeSplitProfiles(t *testing.T) {
 	}
 }
 
-func TestMergeSplitProfilesInvalidInput(t *testing.T) {
```
bogdandrutu, line 128: Removed because it is exactly same are test on line 28

dmitryax (approved):

PR #12253, opened 2025-02-03
Add pprofile aggregation temporality constants

dmathieu:
This adds constants to more easily use the aggregation temporality enum in pprofile.

This behavior is very similar to how pmetrics handles those constants.
See https://pkg.go.dev/go.opentelemetry.io/collector/pdata/pmetric#pkg-constants

cc @rockdaboot


mx-psi (commented):
This deviates from the protobuf, right? Is the intent to add it to the protobuf later?

dmathieu:
Hmm how does this deviate from the protobuf? It's adding public definitions for protobuf enums that are otherwise private.

mx-psi:
Ah, I was not aware this is part of protobuf. Maybe "deviate" is not the right word; what I meant is that this was not generated code as it is the case in [the metrics constants](https://pkg.go.dev/go.opentelemetry.io/collector/pdata/pmetric#pkg-constants) and I didn't understand why. Seems like it's all good now :)

mx-psi (approved):

rockdaboot (approved):

mx-psi:
Will merge this on Friday if there are no further comments

Issue #12254, opened 2025-02-03
regression in service telemetry metrics configuration


codeboten:
The following configuration causes errors in the current version in main:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: localhost:5317
      http:
        endpoint: localhost:5318

exporters:
  debug:
    verbosity: detailed

service:
  telemetry:
    metrics:
      level: detailed
      readers:
        - periodic:
            interval: 10000
            exporter:
              otlp:
                protocol: http/protobuf
                endpoint: localhost:4318
  pipelines:
    metrics:
      receivers: [otlp]
      exporters: [debug]
```

With this config, if I wait a few seconds, I get this error message (no need to send any telemetry):

```
2025/02/03 10:13:45 failed to upload metrics: Post "https:///v1/metrics": http: no Host in request URL
```

_Originally posted by @mx-psi in https://github.com/open-telemetry/opentelemetry-collector/issues/12252#issuecomment-2630401591_

codeboten:
The issue is that the following code is no longer present in the configuration code:


```
func normalizeEndpoint(endpoint string) string {
	if !strings.HasPrefix(endpoint, "https://") && !strings.HasPrefix(endpoint, "http://") {
		return "http://" + endpoint
	}
	return endpoint
}
```

This is causing the URL (`localhost:4318`) to be parsed incorrectly.

PR #12258, opened 2025-02-03
[service] ensure endpoint is prefixed w/ scheme

codeboten:
#### Description

Users can enter the OTLP endpoint w/o a scheme in the prefix. This causes issues with the URL parsing code in the config package.

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/12254


codeboten:
Note this code could be removed in the future if we decided to make parsing of the endpoint more strict, but for now it allows users not to make changes to their config

service/telemetry/internal/migration/normalize.go
```diff
@@ -0,0 +1,10 @@
+package migration
+
+import "strings"
+
+func normalizeEndpoint(endpoint string) *string {
+	if !strings.HasPrefix(endpoint, "https://") && !strings.HasPrefix(endpoint, "http://") {
+		endpoint = "http://" + endpoint
```
TylerHelmuth, line 7: If someone uses grpc instead of http/proto will this break them?

codeboten: no, the same logic was applied to both grpc + http in the old code path, i tested it locally to ensure it works

TylerHelmuth (approved):

bogdandrutu (approved):

Issue #12217, opened 2025-01-30
System for managing own telemetry attributes within pipeline components


djaglowski:
Some types of components are fully or partially shared across multiple instances. (e.g. otlp receiver / memory limiter)

I have previously pursued some optimistic models in which the collector would define certain ways in which sharing is "supported", but the reality is that sharing is an unbounded problem. e.g. Someone could write a processor that shares a data structure between logs and traces pipelines only if it's instantiated on a Wednesday and the component name is "foo". I am convinced that there are too many novel cases to model ahead of time, so here is a proposal to normalize attributes as much as possible, while still allowing component authors to _both add or subtract_ attributes from their telemetry when appropriate.

I will use Loggers as the example here, but I believe the same general ideas can be applied to TracerProviders and MeterProviders.

---

General Idea:

Create a "Logger" struct specifically for use within components. This logger satisfies the same interface and behaves essentially the same as what we have today. However, it has the following important differences:
1. The attributes are preset for the component as if it uses no sharing of parts with other instances of the component. (As defined in [this RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md#attributes))
2. A `Without(key ...string)` method is provided, which allows component authors to _remove_ attributes as appropriate for portions of the code that are shared. This is opposite the typical pattern used by loggers, but necessary if we want our default set of attributes to be correct.

For example:
- OTLP receiver: Since the entire component struct is shared, the factory will use `Without("otelcol.signal")` during construction to create a logger appropriate for the shared instance.
- Memory limiter processor: For any code outside of the shared part, it can use the logger which it was given. For the shared part, is can create a new logger by calling `Without("otelcol.signal", "otelcol.component.id", "otelcol.pipeline.id")`.
- Custom component mentioned above: Typically uses the given logger, but if creating a logs or traces instance on a Wednesday and the component name is "foo", then it calls `Without("otelcol.signal", "otelcol.component.id")` to adjust the scope.

--- 

This implementation looks something like this:

```go
import (
	"go.opentelemetry.io/otel/attribute"
	"go.uber.org/zap"
)

type ComponentLogger struct {
	*zap.Logger
	root *zap.Logger
	diff []attribute.KeyValue
}

func NewComponentLogger(root *zap.Logger, defaultAttrs ...attribute.KeyValue) *ComponentLogger {
	defaultLogger := root
	for _, attr := range defaultAttrs {
[...]

PR #12259, opened 2025-02-03
Introduce component logger with appropriate attributes

### LINKED ISSUE #11814: Loggers for shared components should not report signal type.

djaglowski:
Implements the logger described in https://github.com/open-telemetry/opentelemetry-collector/issues/12217

Alternative to #12057

Resolves #11814

`component/componentattribute`:
- Initializes new module
- Defines constants for component telemetry attribute keys
- Defines a `zapcore.Core` which can remove attributes from the root logger

`service`:
- Rebases component instantiation on attribute sets
- Internal constructors for attribute sets for each component type
- Constructs loggers from `componentattribute`

`otlpreceiver`:
- Uses `componentattribute` to remove `otelcol.signal` attribute from logger

`memorylimiter`:
- Uses `componentattribute` to remove `otelcol.signal`, `otelcol.pipeline.id` and `otelcol.component.id` attributes from logger


component/componentattribute/logger.go
```diff
@@ -0,0 +1,52 @@
...
+	"go.uber.org/zap/zapcore"
+)
+
+var _ zapcore.Core = (*Core)(nil)
+
+type Core struct {
+	zapcore.Core
+	from  *zap.Logger
+	attrs attribute.Set
+}
```
bogdandrutu, line 18: Looking at usages (outside the service) the only think needed is `Without(keys ...string) *zap.Logger`

Can we do some golang magic, and actually public expose only an interface (rest of the implementation is in otelcol/service)?

bogdandrutu: Here is where I got https://github.com/open-telemetry/opentelemetry-collector/pull/12289

I don't have to expose public any interface. We will still need to expose the fields keys not sure where though.

component/componentattribute/logger.go
```diff
@@ -0,0 +1,52 @@
...
+			newAttrs = append(newAttrs, kv)
+			withAttributes = withAttributes.With(zap.String(string(kv.Key), kv.Value.AsString()))
+		}
+	}
+
+	return zap.New(&Core{
+		Core:  withAttributes.Core(),
+		from:  l.from,
+		attrs: attribute.NewSet(newAttrs...),
+	})
```
bogdandrutu, line 51: With this people lose other settings (options of the logger). You need to use `logger.WithOptions(zap.WrapCore(...){})`

component/componentattribute/logger.go
djaglowski: > I don't have to expose public any interface

I'm trying to incorporate your solution to simplify mine, but one consideration overlooked is that in practice we pass different kinds of loggers into tests, so we must make a type assertion to check for the `Without` method. If we're doing this, I think we need to at least give a name the interface so that components do not need to assert against an anonymous interface. The alternative would be to enforce somehow that tests use the same logger, but this is impractical IMO.

djaglowski: Similarly, for tests that do want to validate the functionality of the shared logger, they need to be able to instantiate a logger, so at least then you need to expose a testing logger which is very similar to the actual logger.

component/componentattribute/logger.go
```diff
@@ -0,0 +1,54 @@
...
+
+var _ zapcore.Core = (*coreWithout)(nil)
+
+type coreWithout struct {
+	zapcore.Core
+	from  *zap.Logger
+	attrs attribute.Set
+}
+
+func NewLogger(from *zap.Logger, attrs *attribute.Set) *zap.Logger {
```
djaglowski, line 20: I'm not sure how to avoid exposing this. It's used in component tests, and also in the graph when instantiating the loggers for components.

component/componentattribute/logger.go
djaglowski: I was able to reduce the surface area of `componentattribute` but still think we need a constructor for the logger

component/componentattribute/attribute.go
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package componentattribute // import "go.opentelemetry.io/collector/component/componentattribute"
+
+const (
+	ComponentKindKey = "otelcol.component.kind"
+	ComponentIDKey   = "otelcol.component.id"
+	PipelineIDKey    = "otelcol.pipeline.id"
+	SignalKey        = "otelcol.signal"
+	SignalOutputKey  = "otelcol.signal.output"
```
bogdandrutu, line 11: I would not couple this change with breaking the keys.
```
	zapKindKey            = "kind"
	zapNameKey            = "name"
	zapDataTypeKey        = "data_type"
	zapStabilityKey       = "stability"
	zapPipelineKey        = "pipeline"
	zapExporterInPipeline = "exporter_in_pipeline"
	zapReceiverInPipeline = "receiver_in_pipeline"
```

bogdandrutu (commented):
LGTM, few comments:

1. Need good comments when to use `LoggerWithout`, no comments at all now.
2. I think the attributes keys should be "semconv", maybe someone can start that effort before we change them. Also, I would not change them in this PR.
3. I think we can have the "NewLogger" in otelcol as private, for testing we can use a mock/fake core.

component/componentattribute/attribute.go
djaglowski: I'm not sure what you're suggesting

djaglowski:
> I think the attributes keys should be "semconv", maybe someone can start that effort before we change them. Also, I would not change them in this PR.

We went through a months long RFC process just to decide on these keys. Why wouldn't we switch to them immediately? That's the entire point of this PR. I can see the value of moving them to a `xsemconv` package instead but beyond that needs an explanation why your feeling about it overrides the RFC.

> I think we can have the "NewLogger" in otelcol as private, for testing we can use a mock/fake core.

If it's going to be private and not available in tests, I think it can be in `service/internal`.

For testing, we should make it easy to validate certain things happen expected number of times. e.g. Logger is created once instead of once-per-signal, or log is logged once with correct attributes. Setting aside that a fake/mock core would implement mostly the same functionality as the actual, which package would expose this functionality so that each component does not need to write their own? `componenttest`?

mx-psi:
> 2. I think the attributes keys should be "semconv", maybe someone can start that effort before we change them. Also, I would not change them in this PR.

This was discussed during the RFC https://github.com/open-telemetry/opentelemetry-collector/pull/11406#discussion_r1830355455 resulting in open-telemetry/semantic-conventions/issues/1555. Quoting from the thread, this is the plan we are following:

> Decide which prefix you want to use: "otelcol." or "otel.collector." appear to be your top choices.
> Submit a request to get this added to semconv. We should be able to convince semconv group to accept it and I do not anticipate difficulties here.
> You don't need to be blocked or wait until acceptance by semconv group, there is a ton of work that needs to happen anyway and replacing metric names should be trivial anytime before you declare them stable.

Therefore, I agree with @djaglowski  that we should change them here and **not** wait for any semconv work.

bogdandrutu:
> Therefore, I agree with @djaglowski that we should change them here and not wait for any semconv work.

Sorry, maybe I was not clear. I am not suggesting to wait for that. Just that I was hoping to have a simple PR after this that only does the breaking changes/bug fixes: rename and use of the new API in otelreceiver and memorylimiter.

The main reason for that is to have clean changelog for that, and a simple enough change so that user can go to that PR and easily see what changed.

Happy to move forward with this way as well.

bogdandrutu (approved):
Since my feedback is mostly esthetic will let @mx-psi merge when he feels comfortable with this.

I can play later with  componenattribute structure if you folks want to move faster with this PR.

mx-psi:
Since `componentattributes` is a separate, experimental module, and the only addition to `component` is the `LoggerWithout` method, I think we can go ahead with merging this yes, and iterate on it.

Issue #12247, opened 2025-02-02
Pull batcher does not correctly count the done/index to signal on terminate


bogdandrutu:
The bug was before the conversion to done callback, but here I will show the links after done callback change. 

* If an existing batch is current with data, and a new request comes that makes the batch to exceed the minimum limit or getting split the previous done callback is not called.

See https://github.com/open-telemetry/opentelemetry-collector/blob/eb4f6400d7ca9f36dd4bca9d9d642d0168893766/exporter/internal/queue/default_batcher.go#L68 even if the reqList is created https://github.com/open-telemetry/opentelemetry-collector/blob/eb4f6400d7ca9f36dd4bca9d9d642d0168893766/exporter/internal/queue/default_batcher.go#L54 we don't signal existing "dones".

PR #12260, opened 2025-02-03
Fix default batcher to correctly call all done callbacks exactly once

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/12247

### DIFF
--- a/exporter/internal/queue/default_batcher.go
+++ b/exporter/internal/queue/default_batcher.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/multierr"
+
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter/exporterbatcher"
 	"go.opentelemetry.io/collector/exporter/exporterqueue"
@@ -15,9 +17,9 @@ import (
 )
 
 type batch struct {
-	ctx   context.Context
-	req   internal.Request
-	dones []exporterqueue.DoneCallback
+	ctx  context.Context
+	req  internal.Request
+	done multiDone
 }
 
 // defaultBatcher continuously batch incoming requests and flushes asynchronously if minimum size limit is met or on timeout.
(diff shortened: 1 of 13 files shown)


exporter/internal/queue/default_batcher.go
```diff
@@ -55,76 +57,102 @@ func (qb *defaultBatcher) resetTimer() {
...
-				done(mergeErr)
-				qb.currentBatchMu.Unlock()
-				return
-			}
+
+		// We have at least one result in the reqList. Last in the list may not have enough data to be flushed.
+		// Find if it has at least MinSizeItems, and if it does then move that as the current batch.
+		lastReq := reqList[len(reqList)-1]
+		if lastReq.ItemsCount() < qb.batchCfg.MinSizeItems {
+			// Do not flush the last item and add it to the current batch.
```
dmitryax, line 82: I'm not sure if this is the desired behavior for cases when the queue is disabled (eventually)... If the queue isn't enabled, it can block the caller with a big request for the time of flush timeout. Let's say we have the max batch size set to 200 and the min batch size set to 100. If I send a request with 1001 spans, the first 5 requests will be sent quickly, but the caller will still be blocked until the 6th batch with 1 span can be sent, which is potentially can be done only at the next flush timeout.

It's a tradeoff between making sure the collector always sends at least min batch size requests or ovoid being unnecessary blocked. For me it seems like the second is more important. Let me know what do you think.

exporter/internal/queue/default_batcher.go
```diff
@@ -55,76 +57,102 @@ func (qb *defaultBatcher) resetTimer() {
...
 
-			// flush() blocks until successfully started a goroutine for flushing.
-			qb.flush(batchToFlush.ctx, batchToFlush.req, batchToFlush.dones)
+		return
+	}
+
+	reqList, mergeSplitErr = qb.currentBatch.req.MergeSplit(ctx, qb.batchCfg.MaxSizeConfig, req)
+	// If failed to merge signal all Done callbacks from current batch as well as the current request and reset the current batch.
+	if mergeSplitErr != nil || len(reqList) == 0 {
+		// TODO: investigate why is ok to keep old data here, this is the current behavior in bunch of tests.
```
dmitryax, line 103: I believe this error should be associated with the new incoming request only  We can just send the batch accumulated in `qb. current batch` instead. It must be valid by itself. I'm not sure why we would need to drop it.

bogdandrutu: > I believe mergeSplitErr should be associated with the new incoming request only. 

I could not find anywhere in the documentation that MergeSplit should not "touch" the first request passed if fails. Since we do pass `currentBatch.req` to that function, unless we have a contract that it is "valid" not sure we can continue to use it.

I left this unchanged, but we need to document the expected behavior, and if we say that should not affect passed request then we can remove this.

exporter/internal/queue/default_batcher.go
bogdandrutu: Not sure about this. I do see the "downside" of waiting for 1 cycle but that is not always true anyway since other requests may come sooner.

Here are the cons about your proposal:
* Assuming minSize = 100 and maxSize = 200 and we send only requests with 99 then 102 items, we will always send 2 requests one with 150 the other with 1. This defeats the purpose of batching since we want to batch request because the communication is over network (across data centers) and sending a request with 2 items is very bad from performance and cost perspective.
* This was the previous behavior for the batch processor and nobody complained about it.
* If user want to optimize for low latency then there are options: 1. don't batch since that adds latency anyway; 2. set a low flush interval to ensure low latency.

I see no real benefit, and also not sure I understand the "avoiding being unnecessarily blocked.".

exporter/internal/queue/default_batcher.go
bogdandrutu: Filed https://github.com/open-telemetry/opentelemetry-collector/issues/12268, and I remove this TODO.

exporter/internal/queue/default_batcher.go
dmitryax: Ok. Sounds reasonable. Let's change it this way

PR #12281, opened 2025-02-04
Fix concurrent issues caused by accessing request after calling Send

bogdandrutu:
This can happen when items are added to the memory queue and processed by the batcher in the same time with reading the number of items from this code.

Though our queue does not return error and still add the item unless the Context is cancelled, but likely because of a bug that does not work as expected, see https://github.com/open-telemetry/opentelemetry-collector/issues/12282


dmitryax (approved):
I was actually looking into this particular problem yesterday. Thank you!

PR #12285, opened 2025-02-04
Fix async queue to propagate cancellation all they way to the queue

bogdandrutu:
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/12282

### DIFF
--- a/exporter/exporterqueue/async_queue.go
+++ b/exporter/exporterqueue/async_queue.go
@@ -55,9 +55,7 @@ func (qc *asyncQueue[T]) Start(ctx context.Context, host component.Host) error {
 
 func (qc *asyncQueue[T]) Offer(ctx context.Context, req T) error {
 	span := trace.SpanFromContext(ctx)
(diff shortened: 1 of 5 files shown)


exporter/exporterqueue/async_queue.go
```diff
@@ -57,7 +57,7 @@ func (qc *asyncQueue[T]) Offer(ctx context.Context, req T) error {
 	span := trace.SpanFromContext(ctx)
 	// Prevent cancellation and deadline to propagate to the context stored in the queue.
 	// The grpc/http based receivers will cancel the request context after this function returns.
```
dmitryax, line 59: The comment needs to be moved as well

dmitryax (approved):
Otherwise LGTM

exporter/exporterqueue/async_queue.go
bogdandrutu: Doing now, thanks

PR #12287, opened 2025-02-04
[chore][service/telemetry] Fill out valid level values and match with definition

crobert-1:
#### Description
This updates the trace and metric configs for service telemetry to have [all config options commented](https://github.com/open-telemetry/opentelemetry-collector/blob/677b87e3ab5c615bc3f93b8f99bb1fa5be951751/config/configtelemetry/configtelemetry.go#L13-L20), along with the exact case of the defined string constants of each level.

Note: The case of `Level` doesn't matter in practice as they're all compared in `strings.ToLower` calls, but I figure exact matching is best here.

### DIFF
--- a/service/telemetry/internal/migration/v0.3.0.go
+++ b/service/telemetry/internal/migration/v0.3.0.go
@@ -15,8 +15,10 @@ import (
 
 type TracesConfigV030 struct {
 	// Level configures whether spans are emitted or not, the possible values are:
-	//  - "none" indicates that no tracing data should be collected;
-	//  - "basic" is the recommended and covers the basics of the service telemetry.
+	//  - "None" indicates that no tracing data should be collected.
+	//  - "Basic" is the recommended and covers the basics of the service telemetry.
+	//  - "Normal" adds some other indicators on top of basic.
+	//  - "Detailed" adds dimensions and views to the previous levels.
 	Level configtelemetry.Level `mapstructure:"level"`
 	// Propagators is a list of TextMapPropagators from the supported propagators list. Currently,
 	// tracecontext and  b3 are supported. By default, the value is set to empty list and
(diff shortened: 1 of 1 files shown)


songy23 (approved):

service/telemetry/internal/migration/v0.3.0.go
```diff
@@ -15,8 +15,10 @@ import (
 
 type TracesConfigV030 struct {
 	// Level configures whether spans are emitted or not, the possible values are:
-	//  - "none" indicates that no tracing data should be collected;
-	//  - "basic" is the recommended and covers the basics of the service telemetry.
+	//  - "None" indicates that no tracing data should be collected.
+	//  - "Basic" is the recommended and covers the basics of the service telemetry.
+	//  - "Normal" adds some other indicators on top of basic.
```
codeboten, line 20: Basic and Normal aren't documented here because their behaviour is no different from basic for traces.

service/telemetry/internal/migration/v0.3.0.go
```diff
@@ -57,10 +59,10 @@ func (c *TracesConfigV030) Unmarshal(conf *confmap.Conf) error {
...
 type MetricsConfigV030 struct {
 	// Level is the level of telemetry metrics, the possible values are:
-	//  - "none" indicates that no telemetry data should be collected;
-	//  - "basic" is the recommended and covers the basics of the service telemetry.
-	//  - "normal" adds some other indicators on top of basic.
-	//  - "detailed" adds dimensions and views to the previous levels.
+	//  - "None" indicates that no telemetry data should be collected.
+	//  - "Basic" is the recommended and covers the basics of the service telemetry.
+	//  - "Normal" adds some other indicators on top of basic.
+	//  - "Detailed" adds dimensions and views to the previous levels.
```
codeboten, line 65: Don't know if this means any documentation/examples should also be updated or not.

service/telemetry/internal/migration/v0.3.0.go
crobert-1: Is there anything currently enforcing the fact that `Basic`, `Normal`, and `Detailed` are all the same at this point?

Out of ignorance, my assumption is that at any point a change could be introduced that makes them different, so we may as well include valid values. Maybe we can add a note that they're currently the same thing? I just lean towards documenting valid values.

service/telemetry/internal/migration/v0.3.0.go
crobert-1: I can update documentation and examples. It technically doesn't make a difference, but I figure the closer we are to the defined constant, the less likely we are to run into problems.

crobert-1: I looked into it a bit more, I think updating documentation examples may be a bit more disconcerting to users than helpful. With the widespread use of these constants in other components (such as the debug exporter), I think it's not worth it. There are also [tests to ensure case doesn't impact results](https://github.com/open-telemetry/opentelemetry-collector/blob/ea38dc4a251a1a0409277601f52b4465e43b2eee/config/configtelemetry/configtelemetry_test.go#L19), so it should be safe to keep as-is for now.

I can revert this change or leave as-is, let me know if you have a preference.

Issue #12283, opened 2025-02-04
[extension] [extensioncapabilities] report list of installed collector components and build status in an interface like ConfigWatcher


jackgopack4:
**Is your feature request related to a problem? Please describe.**
ConfigWatcher/NotifyConfig will pass the confmap to an extension to report the configuration of enabled components in an OpenTelemetry Collector.

However, it cannot and does not report a list of included components in the collector. If a user wants to start using a new processor or other component, they will have to refer back to their `builder-config.yaml` in order to see which components are included, and if it is not in their list, rebuild a new collector.

Extension receives [BuildInfo](https://github.com/open-telemetry/opentelemetry-collector/blob/68ff067415c1d467315beba7c31a384a4b532a4f/extension/extension.go#L37) and [ModuleInfo](https://github.com/open-telemetry/opentelemetry-collector/blob/68ff067415c1d467315beba7c31a384a4b532a4f/extension/extension.go#L20-L27) when created, but in my testing it is not clear that the list is complete/full at time of extension creation.

**Describe the solution you'd like**
I would like to propose an addition to `extensioncapabilities.ConfigWatcher` to report complete ModuleInfo and BuildInfo after collector is already up and running, not just at extension creation.

**Describe alternatives you've considered**
A separate interface could be added if this doesn't fit under `ConfigWatcher` domain.

**Additional context**
This would be something that could help vendors troubleshoot their customer's configurations and custom collector distributions without having to consult the original builder-config.yaml or have access to the actual compiled binary to run something like `go tool nm` or `ldd`

**Example code and result**
```
type ConfigWatcher interface {
  ...
  // BuildStatus reports list of installed collector components (e.g. processors, receivers, exporters, etc.) and 
  // information about the tools used to build the binary

  BuildStatus(ctx context.Context, modules extension.ModuleInfo, buildInfo component.BuildInfo) error
}
```

jackgopack4:
It appears that BuildInfo and ModuleInfo do actually work correctly and serve this purpose.

PR #12296, opened 2025-02-05
[extension] Remove Settings.ModuleInfo and move to service.Host hidden method

mx-psi:
#### Description

Moves `extension.Settings.ModuleInfo` to a `service.Host.GetModuleInfo` method.

This field probably fits better on `extension.Settings` or even in `component.BuildInfo`, but since its contents are a bit in flux per discussions like #12283, I would like to move it here to not stabilize it alongside the rest of `extension`.

The field is not used in any public code on Github, so I think it won't have a huge impact to remove it in one go, happy to do this in two steps if preferred.

One relevant difference is that this information is no longer available at extension build time, but rather after the `Start` method is called. Another relevant difference is that this is now available for all component kinds, not just extensions. This could be worked around (we could pass a different host depending on the component kind) but I don't see the use right now.

#### Link to tracking issue
Updates #12283

### DIFF
--- a/service/internal/moduleinfo/moduleinfo.go
+++ b/service/internal/moduleinfo/moduleinfo.go
@@ -1,0 +1,15 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package moduleinfo // import "go.opentelemetry.io/collector/service/internal/moduleinfo"
+
+import "go.opentelemetry.io/collector/component"
+
+// ModuleInfo describes the go module for each component.
+type ModuleInfo struct {
+	Receiver  map[component.Type]string
+	Processor map[component.Type]string
+	Exporter  map[component.Type]string
+	Extension map[component.Type]string
+	Connector map[component.Type]string
+}
--- a/service/internal/moduleinfo/moduleinfo.go
+++ b/service/internal/moduleinfo/moduleinfo.go
@@ -5,11 +5,16 @@ package moduleinfo // import "go.opentelemetry.io/collector/service/internal/mod
 
 import "go.opentelemetry.io/collector/component"
 
-// ModuleInfo describes the go module for each component.
 type ModuleInfo struct {
-	Receiver  map[component.Type]string
-	Processor map[component.Type]string
-	Exporter  map[component.Type]string
-	Extension map[component.Type]string
-	Connector map[component.Type]string
+	// BuilderRef is the raw string passed in the builder configuration used to build this service.
+	BuilderRef string
+}
+
+// ModuleInfos describes the go module for each component.
+type ModuleInfos struct {
+	Receiver  map[component.Type]ModuleInfo
+	Processor map[component.Type]ModuleInfo
+	Exporter  map[component.Type]ModuleInfo
+	Extension map[component.Type]ModuleInfo
+	Connector map[component.Type]ModuleInfo
 }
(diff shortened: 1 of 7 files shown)


mx-psi:
@djaglowski @dpaasman00 would this be an acceptable solution for you? We may revert back to the current state once we have decided on the `ModuleInfo` struct contents and structure, but I want to make sure we get it right first.

service/internal/moduleinfo/moduleinfo.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "go.opentelemetry.io/collector/component"
+
+// ModuleInfo describes the go module for each component.
+type ModuleInfo struct {
+	Receiver  map[component.Type]string
+	Processor map[component.Type]string
+	Exporter  map[component.Type]string
+	Extension map[component.Type]string
+	Connector map[component.Type]string
```
bogdandrutu, line 14: Thinking if we should use this moment at least to re-design a bit this:

```
type ModuleInfo struct {
	ModuleName string
}

type ModuleInfos struct {
	Receiver  map[component.Type]ModuleInfo
	Processor map[component.Type]ModuleInfo
	Exporter  map[component.Type]ModuleInfo
	Extension map[component.Type]ModuleInfo
	Connector map[component.Type]ModuleInfo
}
```

This will allow to extend in the future if needed.

mrsillydog:
@mx-psi - responding on behalf of @dpaasman00 and @djaglowski. This is an acceptable change for us - work is currently [in flight](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37249) that depends on this information, but it's not problematic if that work needs to be adjusted based on this change. Simple enough alteration to modify where we are interacting with the information from component initialization to the Start() method.

mx-psi:
> @mx-psi - responding on behalf of @/dpaasman00 and @/djaglowski. This is an acceptable change for us - work is currently [in flight](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37249) that depends on this information, but it's not problematic if that work needs to be adjusted based on this change. Simple enough alteration to modify where we are interacting with the information from component initialization to the Start() method.

Nice, glad to hear this :)

service/internal/moduleinfo/moduleinfo.go
mx-psi: Done, PTAL

service/internal/moduleinfo/moduleinfo.go
```diff
@@ -0,0 +1,20 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package moduleinfo // import "go.opentelemetry.io/collector/service/internal/moduleinfo"
+
+import "go.opentelemetry.io/collector/component"
+
+type ModuleInfo struct {
+	// BuilderRef is the raw string passed in the builder configuration used to build this service.
+	BuilderRef string
```
bogdandrutu, line 10: Should we have version as well here in the future?

bogdandrutu (approved):

service/internal/moduleinfo/moduleinfo.go
mx-psi: Yes, I think it would make sense to move this to have more structure, but I would rather do this in a separate PR (I would have to start parsing the builder ref and I am worried that this is not trivial)

PR #12297, opened 2025-02-05
[otelcol] Make the `ConfigProvider` interface a struct

evan-bradley:
#### Description

The `ConfigProvider` interface is not very useful since it isn't used by `otelcol.Collector` anymore since we moved to creating it through `ConfigProviderSettings` inside `otelcol.Collector`. Therefore, we should just make `configProvider` the concrete implementation and remove the interface. We could consider entirely removing `ConfigProvider` from the API, but it is used by some downstream distributions, and there is no other way to marshal YAML into `otelcol.Config`, so we would need to make additional changes to provide an adequate replacement.

We could also follow our deprecation process for this change, but in my opinion that will only cause more churn for downstream consumers since they will need to update their code multiple times throughout the process instead of just once.

### DIFF
--- a/otelcol/configprovider.go
+++ b/otelcol/configprovider.go
@@ -20,35 +20,9 @@ import (
 //	cfgProvider.Watch() // wait for an event.
 //	// repeat Get/Watch cycle until it is time to shut down the Collector process.
 //	cfgProvider.Shutdown()
-type ConfigProvider interface {
-	// Get returns the service configuration, or error otherwise.
-	//
-	// Should never be called concurrently with itself, Watch or Shutdown.
-	Get(ctx context.Context, factories Factories) (*Config, error)
-
-	// Watch blocks until any configuration change was detected or an unrecoverable error
-	// happened during monitoring the configuration changes.
-	//
-	// Error is nil if the configuration is changed and needs to be re-fetched. Any non-nil
-	// error indicates that there was a problem with watching the config changes.
-	//
-	// Should never be called concurrently with itself or Get.
-	Watch() <-chan error
-
-	// Shutdown signals that the provider is no longer in use and the that should close
-	// and release any resources that it may have created.
-	//
-	// This function must terminate the Watch channel.
-	//
-	// Should never be called concurrently with itself or Get.
-	Shutdown(ctx context.Context) error
-}
-
-type configProvider struct {
+type ConfigProvider struct {
 	mapResolver *confmap.Resolver
 }
-
-var _ ConfigProvider = (*configProvider)(nil)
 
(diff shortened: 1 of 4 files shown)


otelcol/configprovider.go
```diff
@@ -62,18 +36,21 @@ type ConfigProviderSettings struct {
 //   - Then applies all the confmap.Converter in the given order.
 //
 // * Then unmarshalls the confmap.Conf into the service Config.
-func NewConfigProvider(set ConfigProviderSettings) (ConfigProvider, error) {
+func NewConfigProvider(set ConfigProviderSettings) (*ConfigProvider, error) {
```
evan-bradley, line 39: This is probably the biggest breaking change in this PR. If we return `ConfigProvider` instead of `ConfigProvider`, it's likely that the change would go unnoticed by downstream users of this function. However, we typically return pointers for structs like this one, so this would break that pattern.

The only field in this struct is a pointer, so from an efficiency standpoint I don't think it matters much whether this is a pointer or not.

bogdandrutu: I will ask a different question. Does this need to be public at all?

evan-bradley: My initial preference was to make everything in this file internal. However, if we do that, I see two issues:

1. There's no meaningful replacement for distro authors without forking this code, and I see quite a few usages. For example:

* https://github.com/observIQ/bindplane-otel-collector/blob/69053af3a2a2cbc23b994faf40865aaabff0b1c5/receiver/pluginreceiver/supplied_plugins_test.go#L109
* https://github.com/grafana/tempo/blob/60245c6c2e84d798a01d5af27b7325bdf16c1593/modules/distributor/receiver/shim.go#L214
* https://github.com/grafana/alloy/blob/60f43c5ad23f11e01c7e95829a4bd71baec29486/internal/converter/internal/otelcolconvert/otelcolconvert.go#L87
* https://github.com/SigNoz/signoz-otel-collector/blob/f10f88640de8cf52ce888b8c229949b2f111b9c2/pkg/collectorsimulator/collectorsimulator.go#L114
* https://github.com/xobserve/xobserve/blob/35c2745b913bae4a78dfa7ba212e24150190fc3b/otel-collector/pkg/config/factory.go#L39
2. There would be no way to unmarshal an `otelcol.Config` struct from the Collector's YAML, since all of the unmarshaling facilities are private. We should solve this separately.

My preference would be to make this smaller change for now, and reevaluate unexporting this at least after (2) is resolved.

ms-jcorley (approved):

bogdandrutu (approved):

codeboten (approved):

PR #12299, opened 2025-02-05
[exporter][batching] Serialized bytes based batching for logs

sfc-gh-sili:
#### Description

This PR implements serialized bytes based batching.

#### Link to tracking issue

#### Testing

```
BenchmarkSplittingBasedOnItemCountManySmallLogs-10                    	     409	   2822478 ns/op	 3148718 B/op	   71082 allocs/op
BenchmarkSplittingBasedOnByteSizeManySmallLogs-10                     	     412	   2890594 ns/op	 3156103 B/op	   71242 allocs/op
BenchmarkSplittingBasedOnItemCountManyLogsSlightlyAboveLimit-10       	      38	  27579702 ns/op	43531582 B/op	  700471 allocs/op
BenchmarkSplittingBasedOnByteSizeManyLogsSlightlyAboveLimit-10        	      24	  43067219 ns/op	43639889 B/op	  702927 allocs/op
BenchmarkSplittingBasedOnItemCountHugeLogs-10                         	      40	  29722710 ns/op	41922925 B/op	  690300 allocs/op
BenchmarkSplittingBasedOnByteSizeHugeLogs-10                          	      16	  64321198 ns/op	42107144 B/op	  694144 allocs/op
```

#### Caveat
We book keep both item count and byte size regardless of batching option, which adds an overhead in both cases.

### DIFF
--- a/exporter/exporterhelper/internal/sizer/logs_sizer.go
+++ b/exporter/exporterhelper/internal/sizer/logs_sizer.go
@@ -1,0 +1,68 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sizer // import "go.opentelemetry.io/collector/exporter/exporterhelper/internal/sizer"
+
+import (
+	math_bits "math/bits"
+
+	"go.opentelemetry.io/collector/pdata/plog"
+)
+
+type LogsSizer interface {
+	LogsSize(ld plog.Logs) int
+	ResourceLogsSize(rl plog.ResourceLogs) int
+	ScopeLogsSize(sl plog.ScopeLogs) int
+	LogRecordSize(lr plog.LogRecord) int
+
+	// DeltaSize() returns the delta size when a ResourceLog, ScopeLog or LogRecord is added.
+	DeltaSize(newItemSize int) int
+}
+
+// LogsByteSizer returns the byte size of serialized protos.
+type LogsBytesSizer struct {
+	plog.ProtoMarshaler
+}
+
+// DeltaSize() returns the delta size of a proto slice when a new item is added.
+// Example:
+//
+//	prevSize := proto1.Size()
+//	proto1.RepeatedField().AppendEmpty() = proto2
+//
+// Then currSize of proto1 can be calculated as
+//
+//	currSize := (prevSize + sizer.DeltaSize(proto2.Size()))
+//
+// This is derived from opentelemetry-collector/pdata/internal/data/protogen/logs/v1/logs.pb.go
+// which is generated with gogo/protobuf.
+func (s *LogsBytesSizer) DeltaSize(newItemSize int) int {
+	return 1 + newItemSize + math_bits.Len64(uint64(newItemSize|1)+6)/7 //nolint:gosec // disable G115
+}
+
+// LogsCountSizer returns the nunmber of logs entries.
+type LogsCountSizer struct{}
+
+func (s *LogsCountSizer) LogsSize(ld plog.Logs) int {
+	return ld.LogRecordCount()
+}
+
+func (s *LogsCountSizer) ResourceLogsSize(rl plog.ResourceLogs) int {
+	count := 0
+	for k := 0; k < rl.ScopeLogs().Len(); k++ {
+		count += rl.ScopeLogs().At(k).LogRecords().Len()
+	}
+	return count
+}
+
+func (s *LogsCountSizer) ScopeLogsSize(sl plog.ScopeLogs) int {
+	return sl.LogRecords().Len()
+}
+
+func (s *LogsCountSizer) LogRecordSize(_ plog.LogRecord) int {
+	return 1
+}
+
+func (s *LogsCountSizer) DeltaSize(newItemSize int) int {
+	return newItemSize
+}
--- a/exporter/exporterhelper/logs_batch.go
+++ b/exporter/exporterhelper/logs_batch.go
@@ -8,57 +8,78 @@ import (
 	"errors"
 
 	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+	"go.opentelemetry.io/collector/exporter/exporterhelper/internal/sizer"
 	"go.opentelemetry.io/collector/pdata/plog"
 )
+
+func sizerFromConfig(cfg exporterbatcher.SizeConfig) sizer.LogsSizer {
+	switch cfg.Sizer {
+	case exporterbatcher.SizerTypeItems:
+		return &sizer.LogsCountSizer{}
+	case exporterbatcher.SizerTypeBytes:
+		return &sizer.LogsBytesSizer{}
+	default:
+		return &sizer.LogsCountSizer{}
+	}
+}
 
 // MergeSplit splits and/or merges the provided logs request and the current request into one or more requests
 // conforming with the MaxSizeConfig.
 func (req *logsRequest) MergeSplit(_ context.Context, cfg exporterbatcher.SizeConfig, r2 Request) ([]Request, error) {
+	sizer := sizerFromConfig(cfg)
 	if r2 != nil {
 		req2, ok := r2.(*logsRequest)
 		if !ok {
 			return nil, errors.New("invalid input type")
 		}
-		req2.mergeTo(req)
+		req2.mergeTo(req, sizer)
 	}
 
 	// If no limit we can simply merge the new request into the current and return.
 	if cfg.MaxSize == 0 {
 		return []Request{req}, nil
 	}
-	return req.split(cfg)
+
+	return req.split(cfg.MaxSize, sizer), nil
 }
 
-func (req *logsRequest) mergeTo(dst *logsRequest) {
-	dst.setCachedItemsCount(dst.ItemsCount() + req.ItemsCount())
-	req.setCachedItemsCount(0)
+func (req *logsRequest) mergeTo(dst *logsRequest, sizer sizer.LogsSizer) {
+	if sizer != nil {
+		dst.setCachedSize(dst.Size(sizer) + req.Size(sizer))
+		req.setCachedSize(0)
+	}
 	req.ld.ResourceLogs().MoveAndAppendTo(dst.ld.ResourceLogs())
 }
 
-func (req *logsRequest) split(cfg exporterbatcher.SizeConfig) ([]Request, error) {
+func (req *logsRequest) split(maxSize int, sizer sizer.LogsSizer) []Request {
 	var res []Request
-	for req.ItemsCount() > cfg.MaxSize {
-		ld := extractLogs(req.ld, cfg.MaxSize)
-		size := ld.LogRecordCount()
-		req.setCachedItemsCount(req.ItemsCount() - size)
-		res = append(res, &logsRequest{ld: ld, pusher: req.pusher, cachedItemsCount: size})
+	for req.Size(sizer) > maxSize {
+		ld := extractLogs(req.ld, maxSize, sizer)
+		size := sizer.LogsSize(ld)
+		req.setCachedSize(req.Size(sizer) - size)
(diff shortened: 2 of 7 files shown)


exporter/exporterbatcher/config.go
```diff
@@ -41,6 +41,9 @@ type MaxSizeConfig struct {
 	// If the batch size exceeds this value, it will be broken up into smaller batches if possible.
 	// Setting this value to zero disables the maximum size limit.
 	MaxSizeItems int `mapstructure:"max_size_items"`
+
+	// MaxSizeItems is the maximum number of the batch serialized size.
```
ms-jcorley, line 45: Comment has a copy/paste error:  MaxSizeBytes vs MaxSizeItems.  
Also I'd say "bytes" in the comment, i.e. something like:
//  MaxSizeBytes is the maximum serialized size of the batch in bytes

exporter/exporterhelper/logs_batch.go
```diff
@@ -32,16 +33,27 @@ func (req *logsRequest) MergeSplit(_ context.Context, cfg exporterbatcher.MaxSiz
...
 }
 
 func (req *logsRequest) split(cfg exporterbatcher.MaxSizeConfig) ([]Request, error) {
 	var res []Request
-	for req.ItemsCount() > cfg.MaxSizeItems {
-		ld := extractLogs(req.ld, cfg.MaxSizeItems)
-		size := ld.LogRecordCount()
-		req.setCachedItemsCount(req.ItemsCount() - size)
-		res = append(res, &logsRequest{ld: ld, pusher: req.pusher, cachedItemsCount: size})
+	if cfg.MaxSizeItems != 0 {
```
ms-jcorley, line 43: If both MaxSizeItems and MaxSizeBytes are set, doesn't this logic only handle MaxSizeItems due to the "else" here?

exporter/exporterhelper/logs_batch.go
```diff
@@ -109,3 +121,99 @@ func resourceLogsCount(rl plog.ResourceLogs) int {
 	}
 	return count
 }
+
+func (req *logsRequest) splitBasedOnByteSize(maxByteSize int) ([]Request, error) {
```
ms-jcorley, line 125: Is this function unused?

exporter/exporterhelper/logs_batch_test.go
```diff
@@ -123,7 +123,106 @@ func TestMergeSplitLogs(t *testing.T) {
...
 			assert.Equal(t, len(tt.expected), len(res))
 			for i := range res {
-				assert.Equal(t, tt.expected[i], res[i])
+				assert.Equal(t, tt.expected[i].(*logsRequest).ld, res[i].(*logsRequest).ld)
+			}
+		})
+	}
+}
+
+func TestMergeSplitLogsBasedOnByteSize(t *testing.T) {
```
ms-jcorley, line 132: You may want a test or two where both MaxSizeBytes and MaxSizeItems is set.

sfc-gh-sili: Thank you, Jeremy. I needed some placeholder config API to unblock the implementation, but this config API will be quite different from how it ends up being.

I just opened up https://github.com/open-telemetry/opentelemetry-collector/issues/12303 following some discussion in thishttps://github.com/open-telemetry/opentelemetry-collector/pull/12154. Mainly, we want to
1. Enforces the same sizer for max limit and min limit.
2. Ensure a unified config API for both queuing and batching.

I am happy to hear some feedback and continue discussion there. 

For this PR per se, I wanted the minimal change as possible to unblock implementation, because whichever option we settle on, we will need some refactoring before we can actually read the configuration.

exporter/exporterhelper/logs_batch.go
sfc-gh-sili: Thanks for catching!

exporter/exporterhelper/logs_batch.go
sfc-gh-sili: See [the other comment](https://github.com/open-telemetry/opentelemetry-collector/pull/12299#discussion_r1943708980)

exporter/exporterbatcher/config.go
sfc-gh-sili: Thanks! Updated the comment.

exporter/exporterhelper/logs_batch_test.go
ms-jcorley: I may not be following... why are we adding code for MaxSizeBytes if users can't use it yet?
Looking that the linked issue, wouldn't it be better to wait for the input config format to be determined and do this together as a feature rather than add dead code that will likely need to be changed later?

sfc-gh-sili: @ms-jcorley Once the batch config API settles, we need to update `MergeSplit` API to support passing in bytes configuration. Right now `MergeSplit` is referenced from 13 files, and 44 lines of code, and all those will need to be updated accordingly.

The implementation is separate though. Whichever API we converge to, the underlying implementation will be same.  I just need the bare minimum so I can invoke the feature from unit test - I picked `MaxSizeBytes` because it requires minimum change.

ms-jcorley: Ah, that makes more sense to me now.

exporter/exporterhelper/logs_batch.go
ms-jcorley: Call me paranoid, but since we only expect one to be set, should we fail if we find both?

dmitryax: We currently fail if `MaxSizeItems` is [set at all](https://github.com/open-telemetry/opentelemetry-collector/pull/12299/files#diff-2569611ab8d6f62dce2cace7cba426b2452366d2314736ad8fba9231461f5698R62). So we can resolve this comment

.chloggen/exporter-batcher-bytes-based-batching.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterhelper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Support bytes based baching for logs.
```
dmitryax, line 10: I don't think we need a changelog here since it's unavailable for the end user

exporter/exporterhelper/logs_batch.go
```diff
@@ -109,3 +121,87 @@ func resourceLogsCount(rl plog.ResourceLogs) int {
 	}
 	return count
 }
+
+// extractLogs extracts logs from the input logs and returns a new logs with the specified number of log records.
+func extractLogsBasedOnByteSize(srcLogs plog.Logs, capacity int) plog.Logs {
```
dmitryax, line 126: Have you looked into ways to reuse the existing extract functions by sending a sizer function as an additional argument? It'd be great if we could avoid the code duplication, given that we will need to add other ways to size/split the data later.

.chloggen/exporter-batcher-bytes-based-batching.yaml
sfc-gh-sili: Makes sense. Thanks!

exporter/exporterhelper/logs_batch.go
sfc-gh-sili: You are right. I abstracted the sizing into an interface called `logSizer` that provides a list of sizing functions.  That way `logSizer` would corresponds to a sizer defined in https://github.com/open-telemetry/opentelemetry-collector/pull/12154. Let me know how it looks.

exporter/exporterhelper/logs_batch.go
```diff
@@ -8,104 +8,127 @@ import (
...
 }
 
 // extractResourceLogs extracts resource logs and returns a new resource logs with the specified number of log records.
-func extractResourceLogs(srcRL plog.ResourceLogs, count int) plog.ResourceLogs {
+func extractResourceLogs(srcRL plog.ResourceLogs, capacity int, sizer sizer.LogsSizer) (plog.ResourceLogs, bool) {
+	capacityReached := false
 	destRL := plog.NewResourceLogs()
 	destRL.SetSchemaUrl(srcRL.SchemaUrl())
 	srcRL.Resource().CopyTo(destRL.Resource())
+	capacityLeft := capacity - sizer.ResourceLogsSize(destRL)
```
dmitryax, line 98: Is it expected that we don't use `sizer.DeltaSize` here?

exporter/exporterhelper/logs_batch.go
dmitryax: It looks good to me. Thanks  you!

exporter/exporterhelper/logs_batch.go
sfc-gh-sili: This is intended because this is the first time we take proto size off from capacity. Alternatively this could be:

```
destRL := plog.NewResourceLogs()
capacityLeft = capacity - sizer.Size(destRL)
destRL.SetSchemaUrl(schemaUrl)
capacityLeft -= deltaCapacity(len(schemaUrl)
```

dmitryax: Got it. Thanks for clarifying 👍

dmitryax (approved):

exporter/exporterhelper/internal/sizer/logs_sizer.go
```diff
@@ -0,0 +1,67 @@
...
+
+// DeltaCapacity() returns the delta size of a proto slice when a new item is added.
+// Example:
+//
+//	prevSize := proto1.Size()
+//	proto1.RepeatedField().AppendEmpty() = proto2
+//
+// Then currSize of proto1 can be calculated as
+//
+//	currSize := (prevSize + deltaCapacity(proto2.Size()))
```
jmacd, line 35:
```suggestion
//	currSize := (prevSize + sizer.DeltaSize(proto2.Size()))
```

exporter/exporterhelper/internal/sizer/logs_sizer.go
```diff
@@ -0,0 +1,67 @@
...
+	"go.opentelemetry.io/collector/pdata/plog"
+)
+
+type LogsSizer interface {
+	LogsSize(ld plog.Logs) int
+	ResourceLogsSize(rl plog.ResourceLogs) int
+	ScopeLogsSize(sl plog.ScopeLogs) int
+	LogRecordSize(lr plog.LogRecord) int
+
+	// DeltaCapacity() returns the delta size when a ResourceLog, ScopeLog or LogRecord is added.
```
jmacd, line 18:
```suggestions
	// DeltaSize() returns the delta size when a ResourceLog, ScopeLog or LogRecord is added.
```

exporter/exporterhelper/internal/sizer/logs_sizer.go
```diff
@@ -0,0 +1,67 @@
...
+// Example:
+//
+//	prevSize := proto1.Size()
+//	proto1.RepeatedField().AppendEmpty() = proto2
+//
+// Then currSize of proto1 can be calculated as
+//
+//	currSize := (prevSize + deltaCapacity(proto2.Size()))
+//
+// This is derived from gogo/protobuf 's Size() function
```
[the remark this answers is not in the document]
jmacd: Yes, and you might add that the constant `1` below is specialized for OTLP; it relies on protobuf tags < 16 (which are expressed w/ a single byte of overhead).

jmacd (approved):

dmitryax:
@sfc-gh-sili please fix failures in `lint` and `checks` CI jobs

Issue #12221, opened 2025-01-31
[extension/receiver/exporter/processor/connector] Should we remove `component.Type` from `Settings.ID`?


mx-psi:
The `component.Type` bit from `Settings.ID` is redundant for components: a component knows its own `component.Type`.

Should we remove `component.Type` from `Settings.ID`? 

I think not, but I want to do a prototype, and see how it looks.

From the 2025-01-27 stability working group meeting.

mx-psi:
@open-telemetry/collector-approvers I intend to close this as wontfix on Friday unless somebody objects.

bogdandrutu:
>While there is repetition, it seems justified based on the fact that the component.ID is used as-is much more frequently than the component name, which is used very rarely.

There is a simple script to replace that: `component.NewID(metadata.Type, set.ID)` will be exactly that. Or we can even create a function `ID(component.Type)` and will use `set.ID(metadata.Type)`

mx-psi:
> > While there is repetition, it seems justified based on the fact that the component.ID is used as-is much more frequently than the component name, which is used very rarely.
> 
> There is a simple script to replace that: `component.NewID(metadata.Type, set.Name)` will be exactly that. Or we can even create a function `ID(component.Type)` and will use `set.ID(metadata.Type)`

I am not saying it can't be done, I am just saying that that would mean users having to figure out how to build an `ID`. If 90% of the time users want to just use the ID, why not just give that to them?

bogdandrutu:
> If 90% of the time users want to just use the ID, why not just give that to them?

Not saying to not give the ID, or a function like `ID()` in setting to the user. The fundamental question I have is which component sets the `Type` part.

mx-psi:
@bogdandrutu My point is that it's okay if we have this duplication since it leads to a better user experience.

I understand the concern of mismatch, one possible solution is to explicitly check for it. I filed #12305 to do that for extension; if that makes sense we can extend it to other component types and close this issue.

PR #12305, opened 2025-02-06
[extension] Error out if passed extension.Settings has incorrect type

mx-psi:
#### Description

Explicitly error out if the passed `component.ID` does not have a matching `component.Type` 

#### Link to tracking issue
Updates #12221

### DIFF
--- a/.chloggen/mx-psi_type-mismatch.yaml
+++ b/.chloggen/mx-psi_type-mismatch.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
 component: extension
--- a/.chloggen/mx-psi_type-mismatch.yaml
+++ b/.chloggen/mx-psi_type-mismatch.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: extension
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Explicitly error out at extension creation time if there is a type mismatch.
+
+# One or more tracking issues or pull requests related to the change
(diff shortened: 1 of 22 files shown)


.chloggen/mx-psi_type-mismatch.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
mx-psi, line 4: Given the breakage on contrib I am not sure if this should be marked as a 'breaking' API change. I don't think it is (I feel like it's a misuse of the API to not pass a matching `component.Type`), and the breakage should only happen on tests but still it may be a bit confusing

songy23: I would call it a bug_fix in that case

songy23 (approved):

.chloggen/mx-psi_type-mismatch.yaml
mx-psi, line 4:
```suggestion
change_type: bug_fix
```

mx-psi: Alright, fixed!

extension/extension.go
```diff
@@ -65,6 +65,14 @@ func (f *factory) Stability() component.StabilityLevel {
 	return f.extensionStability
 }
 
+func (f *factory) Create(ctx context.Context, set Settings, cfg component.Config) (Extension, error) {
```
bogdandrutu, line 68: I was thinking this morning about exactly same thing. I would deprecate func at line 35 since no longer useful and don't want to maintain it.

extension/memorylimiterextension/factory_test.go
```diff
@@ -37,7 +38,10 @@ func TestCreate(t *testing.T) {
 	pCfg.MemorySpikeLimitMiB = 1907
 	pCfg.CheckInterval = 100 * time.Millisecond
 
-	tp, err := factory.Create(context.Background(), extensiontest.NewNopSettings(), cfg)
+	set := extensiontest.NewNopSettings()
+	set.ID = component.NewID(factory.Type())
```
bogdandrutu, line 42: Should we switch `extensiontest.NewNopSettings()` to accept a type? Also we can generate a helper like `metadatatest.NewNopSettings` for components that calls the extensiontest but passes the type

extension/extension.go
mx-psi: Done

bogdandrutu (approved):

PR #12308, opened 2025-02-06
[configgrpc] Add ClientConfig.EnhanceContextWithHeaders

### LINKED ISSUE #12307: [configgrpc] Handle gRPC metadata generically
**Problem:**

`configgrpc` and `confighttp` both allow setting a `headers` section to set headers on client requests. However, while `confighttp` [handles this internally](https://github.com/open-telemetry/opentelemetry-collector/blob/43087bdd92ed0a7d0ebc58c1e6652248d58e06b9/config/confighttp/confighttp.go#L226), `configgrpc` does not, as there doesn't seem to be a way to set headers on the `grpc.ClientConn` struct it returns. The OTLP/gRPC exporter fills the gap by extracting the `headers` config, [wrapping it in a `metadata.MD`](https://github.com/open-telemetry/opentelemetry-collector/blob/43087bdd92ed0a7d0ebc58c1e6652248d58e06b9/exporter/otlpexporter/otlp.go#L73), then later [adding it to the `Context` passed to the gRPC call](https://github.com/open-telemetry/opentelemetry-collector/blob/43087bdd92ed0a7d0ebc58c1e6652248d58e06b9/exporter/otlpexporter/otlp.go#L156), which is then taken into account by the gRPC library.

This is not ideal, as it means the `configgrpc.ClientConfig.Headers` is a no-op by default, and components using a gRPC client must manually do this abstraction-leaking processing to take it into account.

**Proposed solution:**

~~I considered wrapping `grpc.ClientConn` to automatically add the metadata in `Invoke` calls, but because the code calling this method is generated by protogen and expects a `grpc.ClientConn` specifically, this does not seem possible.~~

~~My proposed solution is a new method on `ClientConfig` which adds the headers to a provided `Context`, and documenting that the `Headers` field requires calling this function and passing the resulting `Context` into the gRPC call.~~

**Update:** We can actually do it pretty simply with `grpc.WithUnaryInterceptor`.

jade-guiton-dd:
#### Description

See [related issue](https://github.com/open-telemetry/opentelemetry-collector/issues/12307#issue-2835601950) for context

#### Link to tracking issue
Resolves #12307

### DIFF
--- a/config/configgrpc/configgrpc.go
+++ b/config/configgrpc/configgrpc.go
@@ -91,6 +91,9 @@ type ClientConfig struct {
 	WaitForReady bool `mapstructure:"wait_for_ready"`
 
 	// The headers associated with gRPC requests.
+	//
+	// The connection returned from ClientConfig.ToClientConn will not automatically set these headers.
+	// Make sure to use ClientConfig.EnhanceContextWithHeaders in your gRPC call.
 	Headers map[string]configopaque.String `mapstructure:"headers"`
 
 	// Sets the balancer in grpclb_policy to discover the servers. Default is pick_first.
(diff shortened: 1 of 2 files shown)


jade-guiton-dd:
Turns out we can do this without an additional method with `grpc.WithUnaryInterceptor`, closing this

PR #12317, opened 2025-02-07
[cmd/mdatagen] add field for configuring individual metric interval

cuichenli:
#### Description
Add an `Interval` field to the individual metric configuration, allowing users to specify custom intervals for each metric. This enables flexibility in cases where different metrics require distinct collection intervals.

Note: Additional changes are needed in the receiver layer to fully support this functionality.


cmd/mdatagen/internal/templates/config.go.tmpl
```diff
@@ -13,7 +13,7 @@ import (
 // MetricConfig provides common config for a particular metric.
 type MetricConfig struct {
 	Enabled bool `mapstructure:"enabled"`
-
+	Interval uint64 `mapstructure:"interval"`
```
bogdandrutu, line 16: What does an interval mean for a metric?

cuichenli: Apologies for the confusion in the description. This is primarily intended for database scrapers. Since these metrics are queried directly from the database and some of them are resource-intensive, it would be beneficial to allow certain metrics to be scraped less frequently.

bogdandrutu: We should make the generation of this config optional then.

So, do you mean that in the SQL world, every metric is a sql request?

cuichenli: I cannot speak for all databases, but let's take MSSQL and PostgreSQL as examples. In the contrib repository, these metrics are collected by sending SQL queries.

For reference:

PostgreSQL: [client.go#L293](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/postgresqlreceiver/client.go#L293)

MSSQL: [queries.go#L14](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlserverreceiver/queries.go#L14)

I will update the code to make this behavior optional

cuichenli: updated, please take a look, thanks!

dmitryax:
@cuichenli please add a changelog entry in https://github.com/open-telemetry/opentelemetry-collector/tree/main/.chloggen

cuichenli:
> @cuichenli please add a changelog entry in [`main`/.chloggen](https://github.com/open-telemetry/opentelemetry-collector/tree/main/.chloggen)

thanks, updated!

Issue #12212, opened 2025-01-30
[exporterhelper] Preserve trace context through queue


jade-guiton-dd:
**Is your feature request related to a problem? Please describe.**

The [observability requirements for stable components](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#observability-requirements) requires the processing performance of a component to be observable.

At the moment, this is not quite true for the OTLP exporter (cf. [telemetry review](https://github.com/open-telemetry/opentelemetry-collector/issues/11140#issuecomment-2610436390)), and prevents its stabilization. Using the `exporterhelper`, it emits a span for the exporting operation, (including retries), but there is no link with the initial trace that enqueued the data, which prevents analysis of latency and queueing times.

**Describe the solution you'd like**

When the queue is enabled in `exporterhelper`, I think a span should be emitted, its `SpanContext` should be saved in the queue (potentially serialized if using the persistent queue), and the span for the export operation should have a span link linking it to the queueing operation. Note that the links may be N to N, because the batching exporter may fuse or split batches. Adding an attribute on each link describing how many items were processed from an initial batch may be useful.

**Describe alternatives you've considered**

It would also be possible to measure the latency using a new metric, but this would provide less information (especially with batching), and would still require some context propagation through the queue.

**Additional context**

Related issues: #11740, #8804

bogdandrutu:
This is only a problem for the persistent queue implementation. The in-memory queue propagates everything but the deadline. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queue_sender.go#L167C15-L167C28

jade-guiton-dd:
It looks like my observations are because of a bug with the batching code: when batching is not enabled, data goes through the `DisabledBatcher`, which [discards the provided Context and creates a new one with context.Background for some reason](https://github.com/open-telemetry/opentelemetry-collector/blob/349b84b059e52bd152829ee90422c5f897763a76/exporter/internal/queue/disabled_batcher.go#L32). I can file a separate bug issue for that.

However, even with we fix that bug, there is still an issue with the in-memory queue: it currently creates a parent-child relationship between the previous component's span and the export operation span, which will not always make sense because of the asynchronous nature of the queue, and the potentially N-to-N relation between enqueue operations and export operations caused by the batcher, as explained above.

bogdandrutu:
[...]

PR #12318, opened 2025-02-07
[exporter][batcher] MergedContext implemented with SpanLink

sfc-gh-sili:
#### Description

This PR implements a component called mergedContext that is linked to all incoming context via a spanLink.

For example: let's say req1, req2, req3 are batched in the same request, then

* `batch = { requests: [req1, req2, req3], mergedContext: ctx1}`
* `ctx1.span` is linked to `ctx2.span` and `ctx3.span` via a `SpanLink`

#### Link to tracking issue
https://github.com/open-telemetry/opentelemetry-collector/issues/12212

### DIFF
--- a/exporter/exporterhelper/internal/batcher/batch_context.go
+++ b/exporter/exporterhelper/internal/batcher/batch_context.go
@@ -1,0 +1,31 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package batcher // import "go.opentelemetry.io/collector/exporter/exporterhelper/internal/batcher"
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+type traceContextKeyType int
+
+const batchSpanLinksKey traceContextKeyType = iota
+
+// LinksFromContext returns a list of trace links registered in the context.
+func LinksFromContext(ctx context.Context) []trace.Link {
+	if ctx == nil {
+		return []trace.Link{}
+	}
+	if links, ok := ctx.Value(batchSpanLinksKey).([]trace.Link); ok {
+		return links
+	}
+	return []trace.Link{trace.LinkFromContext(ctx)}
+}
+
+func contextWithMergedLinks(ctx1 context.Context, ctx2 context.Context) context.Context {
+	return context.WithValue(
+		context.Background(),
+		batchSpanLinksKey,
+		append(LinksFromContext(ctx1), LinksFromContext(ctx2)...))
+}
--- a/.chloggen/merged_context.yaml
+++ b/.chloggen/merged_context.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterhelper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Link batcher context to all batched request's span contexts.
+
+# One or more tracking issues or pull requests related to the change
+issues: [12212, 8122]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/exporter/exporterhelper/internal/batcher/batch_context_test.go
+++ b/exporter/exporterhelper/internal/batcher/batch_context_test.go
@@ -1,0 +1,39 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package batcher
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.opentelemetry.io/collector/component/componenttest"
+)
+
+func TestBatchContextLink(t *testing.T) {
+	tracerProvider := componenttest.NewTelemetry().NewTelemetrySettings().TracerProvider
+	tracer := tracerProvider.Tracer("go.opentelemetry.io/collector/exporter/exporterhelper")
+
+	ctx1 := context.Background()
+
+	ctx2, span2 := tracer.Start(ctx1, "span2")
+	defer span2.End()
+
+	ctx3, span3 := tracer.Start(ctx1, "span3")
+	defer span3.End()
+
+	ctx4, span4 := tracer.Start(ctx1, "span4")
+	defer span4.End()
+
+	batchContext := contextWithMergedLinks(ctx2, ctx3)
+	batchContext = contextWithMergedLinks(batchContext, ctx4)
+
+	actualLinks := LinksFromContext(batchContext)
+	require.Len(t, actualLinks, 3)
+	require.Equal(t, trace.SpanContextFromContext(ctx2), actualLinks[0].SpanContext)
+	require.Equal(t, trace.SpanContextFromContext(ctx3), actualLinks[1].SpanContext)
+	require.Equal(t, trace.SpanContextFromContext(ctx4), actualLinks[2].SpanContext)
+}
--- a/exporter/exporterhelper/internal/batcher/default_batcher.go
+++ b/exporter/exporterhelper/internal/batcher/default_batcher.go
@@ -120,9 +120,10 @@ func (qb *defaultBatcher) Consume(ctx context.Context, req request.Request, done
 	// - Last result may not have enough data to be flushed.
 
 	// Logic on how to deal with the current batch:
-	// TODO: Deal with merging Context.
 	qb.currentBatch.req = reqList[0]
 	qb.currentBatch.done = append(qb.currentBatch.done, done)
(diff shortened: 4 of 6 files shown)


exporter/internal/queue/merged_context.go
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		mc.ctxArr = append(mc.ctxArr, other)
+		context.AfterFunc(other, func() {
+			mc.cancel()
+		})
+	}
+	return mc
+}
+
+func (mc mergedContext) Deadline() (time.Time, bool) {
```
bogdandrutu, line 35: This may be called multiple times, so I think we should pre-calculate this.

exporter/internal/queue/merged_context.go
```diff
@@ -0,0 +1,73 @@
...
+				mergedDeadline = deadline
+			}
+		}
+	}
+	return mergedDeadline, mergedOk
+}
+
+func (mc mergedContext) Done() <-chan struct{} {
+	return mc.ctx.Done()
+}
```
bogdandrutu, line 54: We don't need to propagate this.

exporter/internal/queue/merged_context.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (mc mergedContext) Err() error {
+	var mergedErr error
+	for _, c := range mc.ctxArr {
+		if err := c.Err(); err != nil {
+			mergedErr = multierr.Append(mergedErr, c.Err())
+		}
+	}
+	return mergedErr
+}
```
bogdandrutu, line 64: Ditto.

exporter/internal/queue/merged_context.go
```diff
@@ -0,0 +1,73 @@
...
+
+package queue // import "go.opentelemetry.io/collector/exporter/internal/queue"
+import (
+	"context"
+	"time"
+
+	"go.uber.org/multierr"
+)
+
+type mergedContext struct {
```
bogdandrutu, line 12: The scope is:
1. Deadline is max of all.
2. Keep a list of all SpanContext.

We just need these 2 things, not the whole contexts.

jade-guiton-dd (changes requested):
I don't think the logic in this PR will work as intended.

- The pipeline in the exporterhelper is queue → batcher → obsReportSender, so exporterhelper only creates its own span AFTER queuing and batching. This means the spans you are creating links between are the spans of the parent component (for example the span emitted by a batchprocessor), which will be hard to interpret for users.
- Let's say that four dequeue operations D1, D2, D3, and D4 are performed, with associated parent spans S1, S2, S3, and S4. Let's say that the data from D1, D2, and part of D3 are put into a batch B1, and the rest of D3 + D4 are put into a second batch B2. Your code will attempt to add links from S1 to S2 and S3, then from S3 to S4. This will be also be difficult to interpret.
- Most importantly, by the time S2 and S3 are added to B1, span S1 will likely already have ended, and will likely already have been exported, without any span links. I tried your code by enabling the batcher in the OTLP exporter, and I suspect this is the reason why I don't see any span links being exported in my tests.

I think the proper solution to adding span links across the queue and batcher would involve:
- Creating a span for the enqueue operation in the queueSender, to avoid relying on or adding links to spans from parent components
- Adding our span links to the span created by the obsReportSender at span creation time, to ensure they are all exported. This would involve not adding the links in the batcher, but only recording them and passing them to the obsReportSender, presumably through a Context key.

sfc-gh-sili:
@jade-guiton-dd Thanks for the feedback! You convinced me on the "connecting to span links in obsReportSender" part - it does make the trace structure clearer. I updated the implementation accordingly.

Regarding spans created from the queue, you proposed to create a span for the enqueue operation, but the enqueue and dequeue uses different contexts if the queue is persistent storage queue. Did you mean we want to create a new span for every dequeue operation?

jade-guiton-dd:
No, I did mean "enqueue". The ultimate goal of the span links is to link the "export" trace with the "input" trace that pushed the data into the exporter. Since only the enqueue operation is a synchronous part of the input trace, I think we will need to create a span for it. To make the link work across the persistent queue, we will need to persist the `SpanContext` across it, but that should hopefully be doable.

However for now, it's probably simpler to add a span for the dequeue operation and link to that across the batching operations. I think we can leave linking the enqueue and dequeue spans across the queues as future work.

exporter/exporterhelper/internal/batcher/batch_context.go
```diff
@@ -0,0 +1,93 @@
...
+	if ctx == nil {
+		return []trace.Link{}
+	}
+
+	if bctx, ok := ctx.(batchContext); ok {
+		if links, ok := bctx.Value(batchSpanLinksKey).([]trace.Link); ok {
+			return links
+		}
+	} else {
+		panic("lalal")
```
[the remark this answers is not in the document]
sfc-gh-sili: Sorry! That was me debugging the unit test. Let me clean this up.

sfc-gh-sili:
@jade-guiton-dd Thanks! Just updated this PR with a new iteration that propagates span context with a thinner layer. Now trace links are stored in a dummy background context for `obs_report_sender` to read from.

Regarding timeout: it's not hard to implement. One way is to keep track of deadline time through the batching process and create context with deadline when the batch flushes. I removed it from this PR because I am not sure how useful the timeout would be. It feels awkward to have an additional timeout when we have timeout in batching and `timeout_sender` already.

Regarding start of span: This PR now starts a new span whenever a item is read from the queue, but I think the proper way is to
1. Like you mentioned, move it to enqueue
2. and delegate the operation to `obs_report_sender` (or something similar)
```
obs_report_sender -> queue_sender -> [ queue ] -> batch_sender -> obs_report_sender -> retry_sender -> timeout_sender
```

Hopefully this PR is enough for the purpose of linking span context in the batcher. Let me know what you think.

exporter/exporterhelper/internal/batcher/batch_context.go
```diff
@@ -0,0 +1,31 @@
...
+
+// LinksFromContext returns a list of trace links registered in the context.
+func LinksFromContext(ctx context.Context) []trace.Link {
+	if ctx == nil {
+		return []trace.Link{}
+	}
+	if links, ok := ctx.Value(batchSpanLinksKey).([]trace.Link); ok {
+		return links
+	}
+	return []trace.Link{trace.LinkFromContext(ctx)}
```
jade-guiton-dd, line 23: When the batcher is disabled (or when it is enabled but only one request ends up in the batch), `contextWithMergedLinks` is never called, and we pass the parent context through directly to the obsreportsender. I think this is fine, but because the latter calls `LinksFromContext`, the parent span ends up as both the parent AND as a link.

It's not a big deal, but I think it would be better to create links only when we cut the trace, which is to say, when calling `contextWithMergedLinks`.

jade-guiton-dd (changes requested):
Thank you for all your work! Just one more fix needed, and a nitpick. (Although I'm not an approver, so a second review will be necessary anyway)

exporter/exporterhelper/internal/queue_sender.go
```diff
@@ -129,7 +125,13 @@ func NewQueueSender(
 		// Have to read the number of items before sending the request since the request can
 		// be modified by the downstream components like the batcher.
 		itemsCount := req.ItemsCount()
+
+		// TODO: move start of span to enqueue instead to dequeue.
+		// Figure out how to preserve span context across persistent storage.
+		ctx, _ = metadata.Tracer(qSet.ExporterSettings.TelemetrySettings).Start(ctx, "exporter/enqueue")
 		err := next.Send(ctx, req)
+		trace.SpanFromContext(ctx).End()
+
```
jade-guiton-dd, line 134: I believe that's still not the right place for to start the trace; this `exportFunc` is only called AFTER the batcher.

I would recommend doing some manual tests to check whether your code behaves as intended. When testing your code, I usually start a local Collector with self-telemetry and the batcher enabled, send a sample trace to it, then check the output in [otel-tui](https://github.com/ymtdzzz/otel-tui/). (Writing automated tests would also be good.)

exporter/exporterhelper/internal/batcher/batch_context.go
jmacd: Yeah, I wondered about created a struct like:

```
type linkContext struct {
  Link trace.SpanLink
  More context.Context
}
```

at each contextWithMergedLinks 

```
	return context.WithValue(
		context.Background(),
		batchSpanLinksKey,
        linkContext{
            Link: trace.SpanFromContext(ctx2),
            More: ctx1,
		)
```

I believe this will reduce overall allocation count. The LinksFromContext() method would repeatedly get the batchSpanLinksKey value and append a link. To improve on this, the caller of LinksFromContext will (or can) know how many requests were put into the batch (I think?), so it can allocate a slice of the correct capacity, then ideally the call `LinksFromContext(into []trace.SpanLink) error` to populate the result, expecting to find `len(into)` many elements.

jade-guiton-dd: I think that's a premature optimization. I think that creating a linked list of new Contexts, which is then resolved into a slice of SpanLinks, would require just as many (if not more) allocations than just `append`ing the SpanLinks to a single slice, even if you allocate the slice with the right length. 

My suggestion was very different:
- in `LinksFromContext`, retrieve the slice associated with the `batchSpanLinksKey`, or return an *empty* slice if the key is absent
- in `contextWithMergedLinks`, do something like this:
```go
links := append(LinksFromContext(ctx1), LinksFromContext(ctx2)...)
for _, ctx := range []Context{ctx1, ctx2} {
  spanCtx := trace.SpanContextFromContext(ctx)
  if spanCtx.IsValid() { // ctx has a parent span
    links = append(links, Link{ SpanContext: spanCtx }) // turn into a link
  }
}
return context.WithValue(context.Background(), batchSpanLinksKey, links)
```

jade-guiton-dd:
@sfc-gh-sili Do you think you'll have the time to finish this PR? I believe the remaining steps would be signing the CLA, fixing the failing tests, and addressing the remaining comment threads.

jade-guiton-dd:
I continued the work in this PR: #12768

PR #12320, opened 2025-02-07
[configgrpc] Insert headers automatically

### LINKED ISSUE #12307: [configgrpc] Handle gRPC metadata generically

jade-guiton-dd:
#### Description

This PR uses the `grpc.WithUnaryInterceptor` option to add the metadata configured in `ClientConfig.Headers` automatically to requests made with the created `grpc.ClientConn`. Currently, users of `configgrpc` must extract the headers from the config and insert them manually in each request.

#### Link to tracking issue
Fixes #12307

#### Testing
The tests in `otlpexporter` check that headers are properly inserted. Removing the OTLP-specific headers code fails the check. Enabling the new code in configgrpc passes it again.

I'm not sure how to easily test this for streaming RPC calls, but since [the docs](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) imply that adding metadata to streaming RPC calls is the same as unary calls, it should hopefully work just as well.


mx-psi (approved):

.chloggen/configgrpc-metadata-interceptor.yaml
```diff
@@ -0,0 +1,26 @@
...
+note: Apply configured Headers automatically
+
+# One or more tracking issues or pull requests related to the change
+issues: [12307]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  configgrpc now calls metadata.AppendToOutgoingContext automatically in an interceptor.
```
mx-psi, line 19: Can we add here that exporters that do this manually will need to remove the manual code doing this?

.chloggen/configgrpc-metadata-interceptor.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
mx-psi, line 4: I am torn about whether this should be called a bug fix instead, if only because people may be a bit more likely to notice this change and adapt their components

jade-guiton-dd: Not sure how much people pay attention to anything non-breaking, but I'll set it to `bug_fix`

PR #12323, opened 2025-02-07
[confighttp] Remove pointers from integer fields

evan-bradley:
#### Description

I was looking through our defaults for a few `confighttp.ClientConfig` fields and found that `0` disables these fields in the underlying `http.Transport` object, which means we don't need them to be pointers.

Docs for the four fields we set in `http.Transport` here: https://pkg.go.dev/net/http#Transport.MaxIdleConns

I also stopped assigning defaults to any fields we take from `http.DefaultTransport` where setting them is a no-op since they aren't set in `http.DefaultTransport`.

#### Link to tracking issue

Helps unblock confighttp 1.0 from depending on https://github.com/open-telemetry/opentelemetry-collector/issues/9478

### DIFF
--- a/config/confighttp/confighttp.go
+++ b/config/confighttp/confighttp.go
@@ -77,21 +77,20 @@ type ClientConfig struct {
 	CompressionParams configcompression.CompressionParams `mapstructure:"compression_params"`
 
 	// MaxIdleConns is used to set a limit to the maximum idle HTTP connections the client can keep open.
-	// By default, it is set to 100.
-	MaxIdleConns *int `mapstructure:"max_idle_conns"`
+	// By default, it is set to 100. Zero means no limit.
+	MaxIdleConns int `mapstructure:"max_idle_conns"`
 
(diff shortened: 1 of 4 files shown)


config/confighttp/confighttp.go
```diff
@@ -77,21 +77,20 @@ type ClientConfig struct {
 	CompressionParams configcompression.CompressionParams `mapstructure:"compression_params"`
 
 	// MaxIdleConns is used to set a limit to the maximum idle HTTP connections the client can keep open.
-	// By default, it is set to 100.
-	MaxIdleConns *int `mapstructure:"max_idle_conns"`
+	// By default, it is set to 100. Zero means no limit.
```
codeboten, line 80: Based on the change at line 163, it looks like it's not possible to set this to 0

codeboten (commented):
I'm ok with removing the pointers here, just one question

jmacd (approved):

atoulme (approved):

mx-psi (approved):

iblancasa (approved):

PR #12343, opened 2025-02-11
[chore] Improve usage/setup of metadata.yaml

bogdandrutu:
Few miss-usages:
1. Using a type like "receiver" when a simple "pkg" can be used, see "receiverhelper";
2. Defining metadata.yaml without a generate statement, see "pdata";
3. Legacy files that no longer need to be generated, see "exporterhelpe" the generated_status;
4. Missing entry into the README.md to generate the documentation, see "pdata"


pdata/metadata.yaml
```diff
@@ -1,8 +1,11 @@
...
+github_project: open-telemetry/opentelemetry-collector
 
 status:
   class: pkg
   codeowners:
     active:
       - BogdanDrutu
       - dmitryax
+  stability:
+    stable: [traces, metrics, logs]
```
songy23, line 11:
```suggestion
    stable: [traces, metrics, logs]
    development: [profiles]
```

not sure if profiles is in development or alpha, but would be good to include

songy23 (approved):

pdata/metadata.yaml
bogdandrutu: Pressed the merge button to fast. Profile is not in this package but in the xprofile.

songy23:
needs `make gotidy`

Issue #12117, opened 2025-01-17
[receiverhelper] Implement new component-identifying attributes


jade-guiton-dd:
The new [Pipeline Component Telemetry](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md) RFC defines a set of attributes that should be used to describe which component instance a telemetry item comes from.

At the moment, the `receiverhelper` emits the receiver's component ID as a `receiver` attribute on metrics and as part of the span name for traces. To bring this in line with the RFC, we would need to add the following attributes to metrics and spans:
- `otelcol.component.kind`: always `receiver`
- `otelcol.component.id`: the component ID
- `otelcol.signal`: `logs`, `metrics`, `traces`, `profiles`

(As mentioned in the RFC, `otelcol.signal` is not mandatory for singleton receivers, but adding it in all cases should be fine.)

All this information should already be available in the helper, but instead of implementing the attributes manually, I would recommend waiting to see if the pipeline telemetry implementation introduces a general mechanism for getting these attributes.

This would be a step towards the stabilization of the otlpreceiver: cf. #11139.

PR #12351, opened 2025-02-11
[otlpreceiver] Add component ID as HTTP span name prefix

jade-guiton-dd:
#### Description

In the spirit of #9382, this PR adds the component ID as a prefix to the name of spans generated by otelhttp instrumentation on the OTLP endpoint. Instead of, for instance, `/v1/traces`, that span will now be called `otlp/foo:/v1/traces`, to easily differentiate between spans from unrelated HTTP endpoints.

I believe this is not a breaking change, as traces are not stable in self-telemetry.

#### Testing
I manually checked that the span name changed. Not sure if a test would be necessary?

### DIFF
--- a/receiver/otlpreceiver/otlp.go
+++ b/receiver/otlpreceiver/otlp.go
@@ -17,6 +17,7 @@ import (
 	"go.opentelemetry.io/collector/component/componentattribute"
 	"go.opentelemetry.io/collector/component/componentstatus"
 	"go.opentelemetry.io/collector/config/confighttp"
+	"go.opentelemetry.io/collector/config/confighttp/xconfighttp"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/consumer/xconsumer"
 	"go.opentelemetry.io/collector/pdata/plog/plogotlp"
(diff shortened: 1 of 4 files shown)


songy23 (approved):
LGTM

receiver/otlpreceiver/otlp.go
```diff
@@ -164,8 +166,15 @@ func (r *otlpReceiver) startHTTPServer(ctx context.Context, host component.Host)
 		})
 	}
 
+	opts := []confighttp.ToServerOption{
+		confighttp.WithErrorHandler(errorHandler),
+		xconfighttp.WithOtelHTTPOptions(otelhttp.WithSpanNameFormatter(func(_ string, req *http.Request) string {
```
bogdandrutu, line 171: I think, now that we have more options, I would rather add the ID as attribute https://pkg.go.dev/go.opentelemetry.io/otel/trace#WithAttributes

jade-guiton-dd: I see, that would make sense, especially with the new component identifying attributes from the RFC. In that case, I will close this issue, and I can add them as part of #12117.

PR #12366, opened 2025-02-12
Fix otlphttpexporter parsing logic for Retry-After

bogdandrutu:
Small improvement to not create a throttled error when delay is 0, same behavior.

### DIFF
--- a/exporter/otlphttpexporter/otlp.go
+++ b/exporter/otlphttpexporter/otlp.go
@@ -222,24 +222,34 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
 	}
 	formattedErr = httphelper.NewStatusFromMsgAndHTTPCode(errString, resp.StatusCode).Err()
 
-	if isRetryableStatusCode(resp.StatusCode) {
-		// A retry duration of 0 seconds will trigger the default backoff policy
-		// of our caller (retry handler).
-		retryAfter := 0
-
-		// Check if the server is overwhelmed.
-		// See spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-throttling
-		isThrottleError := resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusServiceUnavailable
-		if val := resp.Header.Get(headerRetryAfter); isThrottleError && val != "" {
-			if seconds, err2 := strconv.Atoi(val); err2 == nil {
-				retryAfter = seconds
-			}
-		}
-
-		return exporterhelper.NewThrottleRetry(formattedErr, time.Duration(retryAfter)*time.Second)
+	if !isRetryableStatusCode(resp.StatusCode) {
+		return consumererror.NewPermanent(formattedErr)
 	}
 
-	return consumererror.NewPermanent(formattedErr)
+	// Check if the server is overwhelmed.
+	// See spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-throttling
+	isThrottleError := resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusServiceUnavailable
+	if isThrottleError {
+		// Use Values to check if the header is present, and if present even if it is empty return ThrottleRetry.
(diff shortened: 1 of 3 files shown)


exporter/otlphttpexporter/otlp.go
```diff
@@ -223,25 +223,46 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
...
 		// See spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-throttling
 		isThrottleError := resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusServiceUnavailable
-		if val := resp.Header.Get(headerRetryAfter); isThrottleError && val != "" {
-			if seconds, err2 := strconv.Atoi(val); err2 == nil {
-				retryAfter = seconds
+		if isThrottleError {
+			// A retry duration of 0 seconds will trigger the default backoff policy of our caller (retry handler).
+			throttleDuration := getRetryAfterSeconds(resp)
+			if throttleDuration != 0 {
+				return exporterhelper.NewThrottleRetry(formattedErr, getRetryAfterSeconds(resp))
```
dmitryax, line 233: Why do we need to parse it again? 
```suggestion
				return exporterhelper.NewThrottleRetry(formattedErr, throttleDuration)
```

exporter/otlphttpexporter/otlp.go
```diff
@@ -223,25 +223,46 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
...
-		if val := resp.Header.Get(headerRetryAfter); isThrottleError && val != "" {
-			if seconds, err2 := strconv.Atoi(val); err2 == nil {
-				retryAfter = seconds
+		if isThrottleError {
+			// A retry duration of 0 seconds will trigger the default backoff policy of our caller (retry handler).
+			throttleDuration := getRetryAfterSeconds(resp)
+			if throttleDuration != 0 {
+				return exporterhelper.NewThrottleRetry(formattedErr, throttleDuration)
 			}
+			return formattedErr
```
dmitryax, line 235: We still need to return `exporterhelper.NewThrottleRetry` when the header isn't set and falls back to the default intervals

bogdandrutu: Why?

jmacd: I think I agree -- if the server returned a RetryAfter, even a 0 or an already-past timestamp, maintaining the value makes sense.

bogdandrutu: > even a 0 or an already-past timestamp, maintaining the value makes sense.

What if is invalid?

bogdandrutu: Changed to preserve if the header is present and has a valid value (based on the RFC).

dmitryax: Currently, we always return `throttleRetry` if `isRetryableStatusCode(resp.StatusCode)` (true for 4 codes). This PR changes the behavior to return `throttleRetry` only for `http.StatusTooManyRequests` and `http.StatusServiceUnavailable`. Is that an expected change?

dmitryax: Nevermind. I see that `throttleRetry` isn't needed for the retry sender to retry

exporter/otlphttpexporter/otlp.go
```diff
@@ -222,24 +222,34 @@ func (e *baseExporter) export(ctx context.Context, url string, request []byte, p
...
+		// Retry-After = HTTP-date / delay-seconds
+		//
+		// First try to parse delay-seconds, since that is what the receiver will send.
+		if seconds, err := strconv.Atoi(values[0]); err == nil {
+			return exporterhelper.NewThrottleRetry(formattedErr, time.Duration(seconds)*time.Second)
+		}
+		if date, err := time.Parse(time.RFC1123, values[0]); err == nil {
+			return exporterhelper.NewThrottleRetry(formattedErr, time.Until(date))
+		}
+		return formattedErr
```
dmitryax, line 250: This line must be redundant then

dmitryax (approved):

PR #12367, opened 2025-02-12
Fix OTLP http receiver to correctly set Retry-After

bogdandrutu:

PR #12368, opened 2025-02-12
Change memorylimiter and queue full errors to carry time-after information

bogdandrutu:
Depends on https://github.com/open-telemetry/opentelemetry-collector/pull/12367

### DIFF
--- a/exporter/exporterqueue/queue.go
+++ b/exporter/exporterqueue/queue.go
@@ -5,7 +5,12 @@ package exporterqueue // import "go.opentelemetry.io/collector/exporter/exporter
 
 import (
 	"context"
-	"errors"
+	"time"
+
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/durationpb"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter"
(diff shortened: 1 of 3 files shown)


exporter/exporterqueue/queue.go
```diff
@@ -16,7 +21,16 @@ import (
 // not block.
 // Experimental: This API is at the early stage of development and may change without backward compatibility
 // until https://github.com/open-telemetry/opentelemetry-collector/issues/8122 is resolved.
-var ErrQueueIsFull = errors.New("sending queue is full")
+var ErrQueueIsFull = func() error {
+	st := status.New(codes.ResourceExhausted, "sending queue is full")
+	dt, err := st.WithDetails(&errdetails.RetryInfo{
+		RetryDelay: durationpb.New(1 * time.Second),
```
jmacd, line 27: Suggestion: 

Make `ErrQueueIsFull` a struct type supporting the Golang `errors` package's `Is()` and `As()` methods, and let the OTLP receiver translate this into a `status.New(codes.ResourceExhausted, err.Error())` and the `WithDetails` parts itself. 

```
type QueueIsFullError struct {
  Duration time.Duration
} 
func (e QueueIsFullError) Error() string {
  return fmt.Sprintf("sending queue is full: retry after %v", e.Duration)
}
var ErrQueueIsFull = QueueIsFullError{time.Second}
```

Then, allow the component to configure a specific retry-after constant via the `Config` struct. At runtime, inside the component, use the configured retry-after and store a `queueFullErr` field in the component, then the user can configure this value (e.g., `self.queueFullErr = QueueIsFullError{cfg.RetryAfter}`. The OTLP receiver will use e.g.,

```
  err := pipe.Consume(ctx, request)
  candidate := exporterqueue.QueueIsFullError{}
  if errors.As(err, &candidate) {
    gst := status.New(codes.ResourceExhausted, ...)
    if dt, err := gst.withDetails(...)
       gst = dt
    }
    return gst.Err()
  }
  // ...
```

exporter/exporterqueue/queue.go
```diff
@@ -16,7 +21,16 @@ import (
...
+	})
+	if err != nil {
+		panic(err)
+	}
```
jmacd, line 31:
```suggestion
```

exporter/exporterqueue/queue.go
sfc-gh-bdrutu: I feel that this is not a scalable solution that you are proposing unless we say there is only one error like this that we support.

sfc-gh-bdrutu: The reason is because if we start adding checks for 10s of types of errors in every receiver this will become a big problem to maintain.

PR #12370, opened 2025-02-12
bump minimum supported go version to 1.23

codeboten:
With the release of 1.24, we need to bump the tested versions.

### DIFF
--- a/.chloggen/codeboten_1.24.yaml
+++ b/.chloggen/codeboten_1.24.yaml
@@ -7,7 +7,7 @@ change_type: breaking
 component: all
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Added support for go.14, bumped minimum version to 1.23
+note: Added support for go1.24, bumped minimum version to 1.23
 
 # One or more tracking issues or pull requests related to the change
 issues: [12370]
--- a/.chloggen/codeboten_1.24.yaml
+++ b/.chloggen/codeboten_1.24.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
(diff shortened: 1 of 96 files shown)


.chloggen/codeboten_1.24.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: all
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Added support for go.14, bumped minimum version to 1.23
```
mx-psi, line 10:
```suggestion
note: Added support for go 1.24, bumped minimum version to 1.23
```

evan-bradley, line 10:
```suggestion
note: Added support for go 1.24, bumped minimum version to 1.23
```

mx-psi (approved):
LGTM, we could try to do `1.23` instead of `1.23.0`, but I think we'll have to do `1.23.0` eventually, so I am not sure it is worth it

evan-bradley (approved):
There are also notes in the builder readme and in CONTRIBUTING.md that need to be updated.

.chloggen/codeboten_1.24.yaml
codeboten:
```suggestion
note: Added support for go1.24, bumped minimum version to 1.23
```

songy23 (approved):

codeboten:
>  There are also notes in the builder readme and in CONTRIBUTING.md that need to be updated.

@evan-bradley i updated the docs and removed the bits that only seemed relevant before 1.23 was the minimum, ptal

evan-bradley (approved):
Doc changes LGTM.

dmathieu (approved):

Issue #11795, opened 2024-12-03
cicd: check-contrib fails during the release process


codeboten:
The `check-contrib` step failed at release process during the v0.115.0 release with the following message:

```
running go generate ./... in ./extension/encoding/avrologencodingextension
running go generate ./... in ./extension/encoding/jaegerencodingextension
go: updates to go.mod needed; to update it:
	go mod tidy
make[3]: *** [Makefile:179: for-all] Error 1
make[3]: Leaving directory '/tmp/opentelemetry-collector-contrib'
make[2]: *** [Makefile:310: generate] Error 2
make[2]: Leaving directory '/tmp/opentelemetry-collector-contrib'
make[1]: *** [Makefile:281: generate-contrib] Error 2
make[1]: Leaving directory '/home/runner/work/opentelemetry-collector/opentelemetry-collector'
make: *** [Makefile:266: check-contrib] Error 2
Error: Process completed with exit code 2.
```

See logs here: https://github.com/open-telemetry/opentelemetry-collector/actions/runs/12137529261/job/33840992749

codeboten:
To unblock the release, this was reverted in https://github.com/open-telemetry/opentelemetry-collector/pull/11794/files

codeboten:
The reason a tidy needs to be run is that there are modules that pull in dependencies from `collector-contrib` (for example `jaegerencodingextension`) that need to be updated once their dependency (ie. `internal/core`) is update in the contrib repo. it's unclear why `internal/core` updates its versions since it's using a replace statement.

jade-guiton-dd:
@codeboten I'm not sure I understand your explanation. `go mod tidy` is run just before `make generate`, so are you saying to `go mod tidy`s are required? Or that the generate step for one of the other modules changes `go.mod` for `jaegerencodingextension`? Unfortunately, using core from before the revert and the latest contrib main, I can't reproduce the error locally, so I'm not sure how to investigate this.

jade-guiton-dd:
Update on this issue:

I [added a subcommand to `crosslink`](https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/642) which generates a list of modules to tidy, in the appropriate order and with the appropriate repetitions to ensure(?) no modules are left in the "updates to go.mod needed" state at the end.

[...]

PR #12371, opened 2025-02-12
[chore] Enable `generate-contrib` step in `check-contrib`

### LINKED ISSUE #11167: [CI] Contrib tests should also test updated mdatagen

jade-guiton-dd:
#### Description

This PR enables the `make generate-contrib` step in `make check-contrib` originally introduced in #11670, which allows testing the latest version of the mdatagen against contrib. It had been disabled as initial remediation against #11795, which has since been resolved by making `make gotidy` on contrib tidy modules in topological order.

#### Link to tracking issue
Fixes #11167

#### Testing
Modifying `mdatagen` locally to crash on startup causes `make check-contrib` to fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,7 +284,7 @@ generate-contrib:
 	@echo -e "\nGenerating files in contrib"
 	$(MAKE) -C $(CONTRIB_PATH) -B install-tools
 	$(MAKE) -C $(CONTRIB_PATH) generate
-	$(MAKE) -C $(CONTRIB_PATH) gofmt
+	$(MAKE) -C $(CONTRIB_PATH) gofmt GROUP=all
 
 # Restores contrib to its original state after running check-contrib.
 .PHONY: restore-contrib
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -281,7 +283,8 @@ check-contrib:
 generate-contrib:
 	@echo -e "\nGenerating files in contrib"
 	$(MAKE) -C $(CONTRIB_PATH) -B install-tools
-	$(MAKE) -C $(CONTRIB_PATH) generate GROUP=all
+	$(MAKE) -C $(CONTRIB_PATH) generate
+	$(MAKE) -C $(CONTRIB_PATH) gofmt GROUP=all
```
bogdandrutu, line 287: generate on contrib does gofmt correct?

jade-guiton-dd: Ah, you're right! So *that's* why past me wrote "generate GROUP=all"...

bogdandrutu (approved):

PR #12377, opened 2025-02-12
Add a featuregatetest package and module to share featuregate testing support

bogdandrutu:


dmitryax (approved):

featuregate/featuregatetest/test_util.go
```diff
@@ -11,7 +11,9 @@ import (
 	"go.opentelemetry.io/collector/featuregate"
 )
 
-func setFeatureGateForTest(tb testing.TB, gate *featuregate.Gate, enabled bool) {
+// SetGate sets the value to the given gate. Also, it installs a cleanup function to restore
+// the gate to the initial value when the test is done.
+func SetGate(tb testing.TB, gate *featuregate.Gate, enabled bool) {
```
mx-psi, line 16: Why do this instead of using a local registry for testing? I am a bit worried about encouraging this pattern, it modifies global state and it can be hard to reason about

bogdandrutu: Any example where I can see what you envision?

andrzej-stencel (approved):

mx-psi (approved):
I don't see a better way of doing this and don't want to block this indefinitely, approving!

mx-psi:
@bogdandrutu we can merge this once the conflicts are resolved

Issue #11982, opened 2024-12-25
[pdata] Implement Go 1.23 style iterators


yurishkuro:
**Is your feature request related to a problem? Please describe.**

Iterating through `pdata` structures today is pretty convoluted, e.g.

```go
for i := 0; i < traces.ResourceSpans().Len(); i++ {
	resource := traces.ResourceSpans().At(i)
	for j := 0; j < resource.ScopeSpans().Len(); j++ {
		scope := resource.ScopeSpans().At(j)
		for k := 0; k < scope.Spans().Len(); k++ {
			span := scope.Spans().At(k)
```

**Describe the solution you'd like**
Go 1.23 introduced the ability to use iterators for looping through custom structs. The above code could look like this, if we introduce iterator functions to pdata, e.g. `Seq()` and `Seq2()`:

```go
for resource := range traces.ResourceSpans().Seq() {
	for scope := range resource.ScopeSpans().Seq() {
		for span := range scope.Spans().Seq() {
```

```go
for i, resource := range traces.ResourceSpans().Seq2() {
	for j, scope := range resource.ScopeSpans().Seq2() {
		for k, span := range scope.Spans().Seq2() {
```

bogdandrutu:
@yurishkuro good call, will do when we move to minimum supported version to be 1.23

mauri870:
I think this is perfect use case for iterators, the way this is implemented today is quite verbose to write and almost every single use case requires looping over the entries in some way.

I suggest we go with an `All` method, following the api signature for the already existing [maps](https://github.com/golang/go/issues/61900) and [slices](https://github.com/golang/go/issues/61899) counterparts.

Since this involves changes in [autogenerated code](https://github.com/open-telemetry/opentelemetry-collector/tree/ba4ab80/pdata/internal/cmd/pdatagen/internal/templates), there might be oportunities for reusing existing iterator routines from the standard library, for example `ResourceSpans` uses an slice internally.

mauri870:
I started working on this https://github.com/open-telemetry/opentelemetry-collector/compare/main...mauri870:opentelemetry-collector:pdata-iterator-all. It requires Go 1.23 so we cannot accept it until all the other collector packages are bumped to 1.23.

PR #12380, opened 2025-02-13
pdata: add iterator All method to slice and map types 

mauri870:
#### Description

This PR adds support for iterators to Slice and Map types and their autogenerated counterparts. Iterators were stabilized in Go 1.23. 

The All method is analogous to [maps.All](https://pkg.go.dev/maps#All) and [slices.All ](https://pkg.go.dev/slices#All) and is a more idiomatic alternative to the more verbose `for i := 0; i < es.Len(); i++ { z := es.At(i) }` way of looping.

Code that is written like this today:

```go
for i := 0; i < ld.ResourceLogs().Len(); i++ {
        rl := ld.ResourceLogs().At(i)
        for j := 0; j < rl.ScopeLogs().Len(); j++ {
                sl := rl.ScopeLogs().At(j)
                for k := 0; k < sl.LogRecords().Len(); k++ {
                        lr := sl.LogRecords().At(k)
                        // use lr
                }
        }
}
```

Can now be written like this:

```go
for _, rl := range ld.ResourceLogs().All() {
        for _, sl := range rl.ScopeLogs().All() {
                for _, lr := range sl.LogRecords().All() {
                        // use lr
                }
        }
}
```

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/11982


mx-psi:
@djaglowski Do you think this would solve #8927 ?

pdata/pcommon/map.go
```diff
@@ -216,6 +218,8 @@ func (m Map) Len() int {
 //	sm.Range(func(k string, v Value) bool {
 //	    ...
 //	})
+//
+// Deprecated: Use the iterator returned by All instead.
```
mx-psi, line 222: Deprecation messages should have the version that the symbol was deprecated on

pdata/internal/cmd/pdatagen/internal/templates/primitive_slice.go.tmpl
```diff
@@ -57,6 +57,17 @@ func (ms {{ .structName }}) At(i int) {{ .itemType }} {
 	return (*ms.getOrig())[i]
 }
 
+// All returns an iterator over index-value pairs in the slice.
```
mx-psi, line 60: I am curious if the name is the standard used by other libraries? Do we have any examples of the name from widely used libraries (ideally the standard library)?

mauri870: All is a name that that is being used currently in slices and maps https://pkg.go.dev/maps#All and https://pkg.go.dev/slices#All, but that is standard functions.

I'm looking for examples of methods that return iterators, and I found go/types in go 1.24

```txt
go/types

All go/types data structures that expose sequences using a pair of methods such as Len() int and At(int) T now also have methods that return iterators, allowing you to simplify code such as this:

params := fn.Type.(*types.Signature).Params()
for i := 0; i < params.Len(); i++ {
   use(params.At(i))
}

to this:

for param := range fn.Signature().Params().Variables() {
   use(param)
}
```

So another approach is to use more idiomatic names that reflect the library usage, for example `resourceLogs.ScopeLogs()` for example, that would make the code even shorter:

```go
for _, lr := ld.ResourceLogs.ScopeLogs().LogRecords() {
         // use lr
}
```

I think both seem quite useful.

mx-psi: Personally I like the idiomatic names approach more, I think it is clearer

djaglowski: > So another approach is to use more idiomatic names that reflect the library usage, for example `resourceLogs.ScopeLogs()`, that would make the code even shorter:
> 
> ```go
> for _, lr := ld.ResourceLogs().ScopeLogs().LogRecords() {
>          // use lr
> }
> 
> # or
> assert.Len(t, 10, ld.ResourceLogs().ScopeLogs().LogRecords())
> ```

I like the idea of being able to chain like this but could it actually work?

`ld.ResourceLogs()` returns a `ResourceLogsSlice`, which is not an iterator, so you couldn't do `for _, sl := range ld.ResourceLogs()`.

You could add a method `func (rls ResourceLogsSlice) ScopeLogs() iter.Seq2[int, ScopeLogs]` but this is confusing to me for several reasons:
1. It's strange that there's no equivalent way to get an iterator for `ResourceLogs` (because where would you add an equivalent?)
2. `ResourceLogs` already has a `ScopeLogs` method, but it returns `ScopeLogsSlice`. So if you have a `ResourceLogsSlice` level, you can directly get an iterator, but if you have a `ResourceLogs` then you get a `ScopeLogsSlice`. 
3. If you use `ResourceLogsSlice.ScopeLogs()`, this presumably iterates over all scopes _in all resources_, but if you use `ResourceLogs.ScopeLogs()` then the result is only the scopes within a single resource. Sure, it makes sense but having exactly the same name makes this nuance much less obvious IMO.

Personally I find that working with e.g. `ResourceLogs` vs `ResourceLogsSlice` vs `Resource` is already confusing enough. I would hope that whatever iterators we introduce make it easier to _not_ have to think about these nuances every time I implement a loop. If we add more methods with similar names and behave differently, then I think it only compounds the problem.

djaglowski: All that is to say that I prefer `All` or something similar which explicitly indicates that we're asking for all scopes within all resources, or all records within all scopes within all resources.

mauri870: Ah, yeah it would be nice if it worked like I shown. There is no way to make a type implement an iterator , without having an explicit method for it. We could have variants such as ResourceLogsIter() for example for every variant that we need, but it gets out of hand pretty quickly.

mx-psi: Sounds like a convincing case for `All`, I am sold :)

pdata/pcommon/map.go
mauri870: Updated, thanks.

mauri870:
Any ideas on what could be causing the failures https://github.com/open-telemetry/opentelemetry-collector/actions/runs/13333923272/job/37244695282?pr=12380#step:4:3611? It looks like a check for the new All method is being automatically added in some tests. Since Range isn't included there, there must be a way to bypass this—I just don't know what it is.

pdata/pcommon/generated_byteslice.go
```diff
@@ -55,6 +57,17 @@ func (ms ByteSlice) At(i int) byte {
 	return (*ms.getOrig())[i]
 }
 
+// All returns an iterator over index-value pairs in the slice.
+func (ms ByteSlice) All() iter.Seq2[int, byte] {
```
bogdandrutu, line 61: Is "All" a canonical name for this? What do other places use?

djaglowski: Some discussion about this [here](https://github.com/open-telemetry/opentelemetry-collector/pull/12380#discussion_r1954380804)

jmacd (approved):

mauri870:
I see that the code coverage stage failed. Most of the missing coverage is from the generated code around the yield check. I don't think we gain anything functional by pursuing full coverage. Let me know if there is anything I should do before submitting this.

andrzej-stencel (approved):

mx-psi:
@bogdandrutu @dmitryax Could you review as codeowners of the pdata module? I will merge this on or after next Thursday if there are no further blocking comments

yurishkuro (approved):

mx-psi:
This has a bunch of conflicts now :( I can merge once those conflicts are resolved

mauri870:
> This has a bunch of conflicts now :( I can merge once those conflicts are resolved

Thanks for the ping, I just fixed the conflicts.

PR #12381, opened 2025-02-13
[connector,exporter,processor,receiver] Error out on mismatched type

### LINKED ISSUE #12221: [extension/receiver/exporter/processor/connector] Should we remove `component.Type` from `Settings.ID`?

mx-psi:
#### Description

Error out on mismatched type for all component kinds. Same as #12305 but for all component kinds.

#### Link to tracking issue

Requires #12357
Fixes #12221


bogdandrutu (approved):
Please make sure we have tests for all the main use-cases.

jade-guiton-dd:
- Regarding telemetry:

    The component-related attributes defined in the RFC only really work within the context of a standard Collector pipeline (because of `otelcol.pipeline.id`) and configuration (because of `otelcol.component.id`), so I think components that are created manually and used as part of an opaque "sub-pipeline" like that should just inherit the attributes of their containing component. This should happen by default with #12384 as long as the parent component passes the `TelemetrySettings` they received.

    However, It may be useful to add an API to let (for instance) the JMX receiver manually add additional attributes to the `TelemetrySettings` passed to the inner OTLP receiver to identify the source more accurately.

- Regarding the API:

    I'm not sure if this kind of "sub-pipeline" should be considered a supported use case of components, so I think the onus should be on component authors to manually adjust the settings passed into the child component to make it work, in this case manually adjusting the component ID.

    If we DO want to support this use case, I think the best way would be some kind of API to create a "derived" `receiver/exporter/etc.Settings` with the appropriate values for the child component, instead of parents passing their own as-is.

codeboten (commented):
Is this unblocked now that #12357 is merged?

mx-psi:
@codeboten I want to fix these first https://github.com/open-telemetry/opentelemetry-collector/pull/12381#issuecomment-2668331333

PR #12201, opened 2025-01-28
Ensure unique aliases

mackjmr:
#### Description
This PR ensures that unique aliases are used for modules with same suffix.

#### Link to tracking issue
Fixes #12162

PR #12385, opened 2025-02-13
add providers converters to moduleinfo and converters to component co…

jackgopack4:
…mmand

#### Link to tracking issue
Fixes #


mx-psi (commented):
Can we split the two changes (adding converters and adding this to moduleInfo) into two separate PRs?

jackgopack4:
> Can we split the two changes (adding converters and adding this to moduleInfo) into two separate PRs?

yes, no problem. I will make this into the "add converters to ModuleInfo" and split out adding the both of them to the ModuleInfos struct for a future PR after this one is merged.

jackgopack4:
I have removed the "add providers and converters to ModuleInfos struct" portions of the PR and limited this PR to just adding Converters to components command

confmap/converter.go
```diff
@@ -35,4 +35,6 @@ func NewConverterFactory(f CreateConverterFunc) ConverterFactory {
 type Converter interface {
 	// Convert applies the conversion logic to the given "conf".
 	Convert(ctx context.Context, conf *Conf) error
+	// Type returns the type of the converter.
+	Type() string
```
mx-psi, line 39: Adding a method to a non-sealed interface is a breaking change, we can't do it this way, since `confmap` is 1.0. As it stands, converters don't have a type

otelcol/collector.go
```diff
@@ -70,7 +70,10 @@ type CollectorSettings struct {
 	ConfigProviderSettings ConfigProviderSettings
 
 	// ProviderModules maps provider schemes to their respective go modules.
-	ProviderModules map[string]string
+	ProviderModules map[component.Type]string
```
mx-psi, line 73: This is a breaking change (which is fine since `otelcol` is 1.0), but I don't understand why we are doing it

jackgopack4: It will make this compatible with `buildModuleInfos` used to build the ModuleInfos struct, allowing these components to be reported in ModuleInfos to extensions, which I plan to propose in the next PR after this
https://github.com/jackgopack4/opentelemetry-collector/blob/5631f36480d21aa1d2850680ec28ffc5a5f08a00/otelcol/collector.go#L212

.chloggen/jackgopack4-add-converters-component-command.yaml
```diff
@@ -0,0 +1,25 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Converters are now available in the `components` command. 
+
+# One or more tracking issues or pull requests related to the change
+issues: [11900, 12385]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: buildinfo in components command is now reported as "build_info" for legibility.
```
mx-psi, line 18: Let's keep the PRs focused on a single purpose. AFAICT this is independent from the other changes, so I would like to drop it or move it to a separate PR

otelcol/collector.go
```diff
@@ -72,6 +72,9 @@ type CollectorSettings struct {
 	// ProviderModules maps provider schemes to their respective go modules.
 	ProviderModules map[string]string
 
+	// ConverterModules maps converter names to their respective go modules.
+	ConverterModules map[string]string
```
mx-psi, line 76: I don't see the point on adding a map here, there is no converter name so we could just have a list of Go modules, right?

otelcol/command_components.go
```diff
@@ -19,24 +19,26 @@ import (
 )
 
 type componentWithStability struct {
-	Name      component.Type
-	Module    string
-	Stability map[string]string
+	Name      component.Type    `yaml:"name"`
+	Module    string            `yaml:"module"`
+	Stability map[string]string `yaml:"stability"`
```
mx-psi, line 24: Is this change necessary? If not, let's keep it out of this PR

otelcol/collector.go
jackgopack4: it is possible builder has a different name as reference such as changes made in https://github.com/open-telemetry/opentelemetry-collector/pull/12201
I think it could still be worthwhile to pass in the `Name` field from the Converters

mx-psi (approved):

mx-psi:
Will merge this next Wednesday

evan-bradley (approved):

Issue #7455, opened 2023-03-29
Core component to compute attribute sets needed


jmacd:
**Is your feature request related to a problem? Please describe.**

It is common to aggregate telemetry signals by attribute sets. It requires custom code to do this with `pdata` objects, which leads to inefficiency and repetition in our code.

As a concrete example, note that the current core component batchprocessor makes no effort to combine similar containers in the OTLP hierarchy.  For example, when two scopes are identical it should be possible to combine the entries and avoid repeating the scope details, but for this we need a helper library. Compression performance would improve with better batching, but it takes a lot of code to achieve this and we should standardize this code as part of the collector core.

**Describe the solution you'd like**
This is a complicated question because there are heavy performance costs associated with the decisions we make here. In early prototypes for OTel-Go metrics, the https://github.com/open-telemetry/opentelemetry-go/blob/main/attribute/set.go package was created. This package is one ideal solution to the problem, except that it takes OTel-Go attribute lists as input instead of `pcommon.Map` objects. We need a solution that works for `pcommon.Map` without requiring error-prone conversion to `[]attribute.KeyValue` from the OTel-Go library.

The cost of attribute set construction is expensive--no matter what. Frequently, we see code written that attempts to avoid a memory allocation by first hashing the set of attributes. This may require sorting the attributes, which also can trigger memory allocation. There is a question of whether the user of a hashing approach will also check for hash collisions, because checking hash collisions implies the use of memory to maintain the exact set of attributes.

In my own code, when I want to optimize the cost of attribute processing I arrange to use OTel-Go's `attribute.NewSetWithSortable()` because it avoids all but one obligatory allocation associated with constructing a flat set of attributes. However, even that one allocation can be too much in a fast code path, so for the [Lightstep metrics SDK](https://github.com/lightstep/otel-launcher-go/blob/main/lightstep/sdk/metric/README.md) I've written an optimization to use a hash function. As discussed above, a hash function alone does not give as much optimization as some users want _if you also implement collision detection_. To be very clear, the cost benefit of hashing is limited by the impact of collision detection, so we should survey existing services and libraries to see how they treat this. For the Lightstep metrics SDK, I gave the user an [option](https://github.com/lightstep/otel-launcher-go/blob/main/lightstep/sdk/metric/README.md#ignorecollisions) to ignore collisions. (Lightstep chooses to run with this option.)

[...]

PR #12390, opened 2025-02-14
introduce pprofile.AddAttribute helper

### LINKED ISSUE #12206: [pdata/pprofile] Add helper method to allow modifying a struct's attributes
In profiles, attributes are set with an `AttributeTable` at the profile level. Then, every internal object has an `AttributeIndices` attribute which holds references to the main table.

This means components need an additional step to modify the attributes (add, remove them). So we should provide an helper method to facilitate the action and make it efficient all the time.

We could have something like:

```golang
ToAttributeIndices(data pcommon.Map, table AttributeTable, record attributable)
```

Which would do a full replace of the attributes with a new map.
However, folks would have to first get the attributes and then update them to make a single update.

We could also have a single attribute add:

```golang
Addattribute(key, value string, table AttributeTable, record attributable)
```

Related: https://github.com/open-telemetry/opentelemetry-collector/pull/12176

cc @open-telemetry/profiling-approvers @open-telemetry/profiling-maintainers

dmathieu:
#### Description

This introduces the `pprofile.AddAttribute` helper method so profile extensions can modify attributes.
Closes #12206

```
BenchmarkAddAttribute
BenchmarkAddAttribute/with_a_new_string_attribute
BenchmarkAddAttribute/with_a_new_string_attribute-10            30461164                39.55 ns/op           32 B/op          2 allocs/op
BenchmarkAddAttribute/with_an_existing_attribute
BenchmarkAddAttribute/with_an_existing_attribute-10             30714298                39.44 ns/op           32 B/op          2 allocs/op
BenchmarkAddAttribute/with_a_duplicate_attribute
BenchmarkAddAttribute/with_a_duplicate_attribute-10             29132547                39.47 ns/op           32 B/op          2 allocs/op
BenchmarkAddAttribute/with_a_hundred_attributes_to_loop_through
BenchmarkAddAttribute/with_a_hundred_attributes_to_loop_through-10               6762793               176.8 ns/op            32 B/op          2 allocs/op
```

Those benchmarks kind of have an N+1 issue, where the method will take longer the more attributes there are.
However, since the attribute table is a map, we can't go around with looping through it.

I also had to use `FromRaw` and `AsRaw` in there, to properly handle `any` values.

This currently doesn't support removing entries. But doing so would also have a non-trivial performance impact, since we'd have to loop through the `AttributeIndices` slice.


dmathieu:
@bogdandrutu I'm happy to hear your input on improvements here, especially with the loop through all attributes issue, and the use of `FromRaw` and `AsRaw` to be able to have an `any` value.

pdata/pprofile/attributes.go
```diff
@@ -26,3 +30,35 @@ func FromAttributeIndices(table AttributeTableSlice, record attributable) pcommo
 
 	return m
 }
+
+// AddAttribute updates an AttributeTable and a record's AttributeIndices to
+// add a new attribute.
+// The record can by any struct that implements an `AttributeIndices` method.
+func AddAttribute(table AttributeTableSlice, record attributable, key string, value any) error {
+	for i := 0; i < table.Len(); i++ {
```
rockdaboot, line 38:
```suggestion
	for i := range table.Len() {
```

pdata/pprofile/attributes.go
```diff
@@ -26,3 +30,35 @@ func FromAttributeIndices(table AttributeTableSlice, record attributable) pcommo
...
+		a := table.At(i)
+
+		if a.Key() == key && value == a.Value().AsRaw() {
+			if i >= math.MaxInt32 {
+				return fmt.Errorf("Attribute %s=%#v has too high an indice to be added to AttributeIndices", key, value)
+			}
+
+			record.AttributeIndices().Append(int32(i)) //nolint:gosec // overflow checked
```
rockdaboot, line 46: Can we avoid identical entries by looping over the AttributeIndices and checking whether `i` already exists?

pdata/pprofile/attributes.go
```diff
@@ -26,3 +30,35 @@ func FromAttributeIndices(table AttributeTableSlice, record attributable) pcommo
...
+			return nil
+		}
+	}
+
+	if table.Len() >= math.MaxInt32 {
```
rockdaboot, line 51: Before this check ever triggers, many GB of RAM are used. Does it make sense to have a smaller limit, like 10k or 100k?

pdata/pprofile/attributes.go
```diff
@@ -26,3 +30,35 @@ func FromAttributeIndices(table AttributeTableSlice, record attributable) pcommo
...
+		return errors.New("AttributeTable can't take more attributes")
+	}
+	table.EnsureCapacity(table.Len() + 1)
+	entry := table.AppendEmpty()
+	entry.SetKey(key)
+	err := entry.Value().FromRaw(value)
+	if err != nil {
+		return err
+	}
```
rockdaboot, line 60:
```suggestion
	if err := entry.Value().FromRaw(value); err != nil {
		return err
	}
```

pdata/pprofile/attributes.go
dmathieu: This check is there only for the gosec linter when converting int into an int32.
While actually limiting the number of attributes may make sense, this is not the proper location to do it.

pdata/pprofile/attributes.go
dmathieu: Every other loop through slices in pdata uses the old for loop. I'd rather not deviate from the convention.

pdata/pprofile/attributes.go
dmathieu: That kind of goes with the concern regarding loops and deletion of attributes I brought up in the PR description.

pdata/pprofile/attributes.go
rockdaboot: Take it as a nit, not a blocker.
Is there really a "convention" or is it more a copy&paste thing everywhere? 🤷🏻

pdata/pprofile/attributes.go
rockdaboot: Where else if not at the point where we append data?

dmathieu: If we already have 100k of attributes here, it should have been limited by the receiver.

pdata/pprofile/attributes.go
dmathieu: I'd like to hear @bogdandrutu's feedback first.

pdata/pprofile/attributes.go
rockdaboot: More likely that processors are using this helper to add attributes.
Not a blocker, though.

pdata/pprofile/attributes.go
rockdaboot: Just to understand the background correctly: when processors add attributes, are these explicitly requested by the user via configuration? If the user is in control and aware, I'd agree that a check for duplication here is not needed.

In any case, we should document the behavior.

rockdaboot (approved):

pdata/pprofile/attributes.go
```diff
@@ -26,3 +30,41 @@ func FromAttributeIndices(table AttributeTableSlice, record attributable) pcommo
 
 	return m
 }
+
+// AddAttribute updates an AttributeTable and a record's AttributeIndices to
+// add a new attribute.
+// The record can by any struct that implements an `AttributeIndices` method.
+func AddAttribute(table AttributeTableSlice, record attributable, key string, value any) error {
```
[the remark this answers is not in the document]
dmathieu: attributable is there to allow the behavior on any of the generated types that have attribute indices.
Location, Mapping, Sample.
Folks will only ever need to manipulate the generated struct. They don't need the interface.

pdata/pprofile/attributes.go
```diff
@@ -26,3 +30,41 @@ func FromAttributeIndices(table AttributeTableSlice, record attributable) pcommo
...
+	for i := range table.Len() {
+		a := table.At(i)
+
+		if a.Key() == key && value == a.Value().AsRaw() {
```
jmacd, line 41: nit: I know that `AsRaw()` could make a copy of the data. This is a reminder that it would be nice if you could write `a.Value().Equal(value)` and avoid the allocation.

pdata/pprofile/attributes.go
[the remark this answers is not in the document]
jmacd: Thanks. I have a feeling there is a more general solution out there, but this is pragmatic. 

Part of https://github.com/open-telemetry/opentelemetry-collector/issues/7455.

jmacd (approved):

pdata/pprofile/attributes.go
mx-psi: @dmathieu Happy to merge after you reply to this :)

dmathieu: I have opened https://github.com/open-telemetry/opentelemetry-collector/issues/12561 to address this in another PR, since it requires changed in `pcommon`.

pdata/pprofile/attributes.go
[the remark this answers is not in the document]
bogdandrutu: why do you not accept a value Value?

PR #12406, opened 2025-02-17
Preallocate a map of the right size in pcommon Map.AsRaw()

kamstrup:
#### Description
We are seeing pcommon/Map.AsRaw() show up in the memory profiles.

This PR ensure we pre-allocate a map[string]any of the expected size, instead of just an empty map.


bogdandrutu:
This PR is ok, but I am curious why do you need to call this function and in which component?

bogdandrutu:
Please add a changelpg

kamstrup:
> Please add a changelog entry

Done :-)

bogdandrutu:
@kamstrup ping on 

> but I am curious why do you need to call this function and in which component?

kamstrup:
> @kamstrup ping on
> 
> > but I am curious why do you need to call this function and in which component?

Good question :-) We have custom storage engine and we do some massaging of otel payloads to store them most efficiently. We pull the attributes out as a map and rearrange the data.

When we have huge nested maps AsRaw starts to show up in the profiles. Not a deal breaker by any means, of course, but just thought it was a nice small improvement

jmacd (approved):

PR #12407, opened 2025-02-17
Add clarification about attribute types to Telemetry RFC

djaglowski:
As suggested in https://github.com/open-telemetry/opentelemetry-collector/issues/12217#issuecomment-2648752916

### DIFF
--- a/docs/rfcs/component-universal-telemetry.md
+++ b/docs/rfcs/component-universal-telemetry.md
@@ -27,7 +27,7 @@ be automatically captured for each kind of pipeline component.
 
 ## Attributes
 
-All signals should use the following attributes:
+Spans, data points, and log records should carry the following attributes:
 
 ### Receivers
 


docs/rfcs/component-universal-telemetry.md
```diff
@@ -27,7 +27,7 @@ be automatically captured for each kind of pipeline component.
 
 ## Attributes
 
-All signals should use the following attributes:
+Spans, data points, and log records should carry the following attributes:
```
djaglowski, line 30: What is the equivalent for profiles?

mx-psi: There's multiple things that can have attributes in the case of profiles: samples, mappings and locations. I guess we can put them all

songy23 (approved):

dmitryax (approved):

jade-guiton-dd:
Since we ended up pivoting to instrumentation scope attributes, maybe we should update the wording of the RFC?

PR #12433, opened 2025-02-19
[service] Add `service::telemetry::metrics::views` config key

jade-guiton-dd:
#### Description

This PR adds a `service::telemetry::metrics::views` config key, which explicitly sets the list of metric views used for internal telemetry, mirroring `meter_provider::views` in the SDK config. This can be used to disable specific internal metrics, among other uses.

This key will cause an error if used alongside other features which would normally implicitly create views, such as:
- not setting `service::telemetry::metrics::level` to `detailed`;
- enabling the `telemetry.disableHighCardinalityMetrics` feature gate.

#### Further discussion needed

- A comment notes that the `telemetry.disableHighCardinalityMetrics` alpha gate *"will be removed when the collector allows for view configuration"*. I think setting the gate as deprecated first would be the correct thing to do, but it means that users relying on it will see their Collectors crash on update. Is that okay?

- In the context of being able to enable/disable specific metrics, this key is only useful to disable metrics from an "all enabled" baseline. It cannot easily be used to customize the set of metrics emitted at `level: normal`, `level: basic`. Discussion is ongoing in #10769 on how to handle that, but the solution will probably involve a new key, which should hopefully be backward-compatible with the user-visible changes in this PR.

#### Link to tracking issue
Updates #10769

#### Testing
None yet

#### Documentation
None yet, except the changelog


service/service.go
```diff
@@ -404,9 +412,19 @@ func dropViewOption(selector *config.ViewSelector) config.View {
...
+		// Drop all otelhttp and otelgrpc metrics if the level is not detailed.
+		views = append(views,
+			dropViewOption(&config.ViewSelector{
+				MeterName: ptr("go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"),
+			}),
+			dropViewOption(&config.ViewSelector{
+				MeterName: ptr("go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"),
+			}),
+		)
+	} else if disableHighCardinalityMetricsFeatureGate.IsEnabled() {
```
mx-psi, line 425: Can you add a comment explaining the Go SDK behavior when there are multiple views that affect the same set of metrics?

service/telemetry/config.go
```diff
@@ -111,5 +111,9 @@ func (c *Config) Validate() error {
 		return errors.New("collector telemetry metrics reader should exist when metric level is not none")
 	}
 
+	if c.Metrics.Views != nil && c.Metrics.Level != configtelemetry.LevelDetailed {
+		return errors.New("metric views can only be set when level is detailed")
```
mx-psi, line 115: nit: Use explicit names for the configuration
```suggestion
		return errors.New("service::telemetry::metrics::views can only be set when`service::telemetry::metrics::level is set to detailed")
```

mx-psi (commented):
Can you increase the test coverage? Could you also add information about this (in a draft PR) to opentelemetry.io docs?

mx-psi:
> Regarding the lack of test coverage: it's almost entirely from the existing code for the feature gate that I just moved. I can try to add a new test for it; is there a way to set feature gates in tests?

You can check the tests that are modified in #12377

service/telemetry/config.go
```diff
@@ -111,5 +111,9 @@ func (c *Config) Validate() error {
 		return errors.New("collector telemetry metrics reader should exist when metric level is not none")
 	}
 
+	if c.Metrics.Views != nil && c.Metrics.Level != configtelemetry.LevelDetailed {
+		return errors.New("service::telemetry::metrics::views can only be set when service::telemetry::metrics::level is detailed")
```
bogdandrutu, line 115: Why this? I think we should allow views only if the level is not set.

jade-guiton-dd: The way we implement levels is by adding views that filter out unwanted metrics. But setting views explicitly completely replaces the default views. There is discussion of this in the tracking issue, but I thought it would be confusing to add an empty `views` section and suddenly see more metrics than before. For that reason, I decided to only allows views with `level: detailed`, which enables all metrics and thus sets no views by default.

jade-guiton-dd: There was also the idea of adding a new `level: custom` value, so users explicitly declare their intent to bypass the level-based system and use more advanced view-based configuration, not sure what you think of it.

jade-guiton-dd: @bogdandrutu Do you think changes are necessary here?

mx-psi: > Why this? I think we should allow views only if the level is not set.

Level detailed is the "please ignore level" option, since with detailed the level has no effect.

mx-psi (approved):

service/telemetry/internal/migration/v0.3.0.go
```diff
@@ -73,6 +73,11 @@ type MetricsConfigV030 struct {
 	// Readers allow configuration of metric readers to emit metrics to
 	// any number of supported backends.
 	Readers []config.MetricReader `mapstructure:"readers"`
+
+	// Views allows advanced configuration of emitted metrics by setting
+	// the list of views passed to the SDK.
+	// This is incompatible with `level` values other than `detailed`.
+	Views []config.View `mapstructure:"views"`
```
bogdandrutu, line 80: instead, use: 
```
MeterProvider MeterProvider `mapstructure:",squash"`
```

This will show that we use everything from the config.

bogdandrutu (approved):

jmacd (approved):

service/telemetry/internal/migration/v0.3.0.go
jade-guiton-dd: Done. I modified the SDK instantiation code to pass the modified original struct to `otelconf` instead of creating a new one, in case new fields get added later upstream.

PR #12435, opened 2025-02-19
[docs] add guidance around what a collector/distro is

### LINKED ISSUE #8555: What is an OpenTelemetry Collector, what is a distribution?
We had a discussion recently around what is an OpenTelemetry Collector and what is a distribution of the Collector. I would like to gather your opinions.

@dyladan proposed that only what the SIG Collector produces can be called an "OpenTelemetry Collector" and that a distribution has to fulfill the following requirements:

* uses the collector framework (upstream not a fork)
* includes only plugins/components which are compatible with the collector framework. they don't need to be in the otel repos, but you should be able to point the upstream collector builder at them

I tend to agree with him, but I'm eager to hear your opinions. The GC might have the right to make the final decision if we can't get an agreement, but I think we can indeed reach a consensus, at least between the GC and the Collector maintainers (core and contrib).

---

Update - 2024-07-17: based on the state of the discussion so far, here are the issues we identified:

- [ ] #10638
- [ ] #10639
- [ ] #10640
- [ ] #10641
- [ ] #10642

codeboten:
Fixes #8555


mx-psi (approved):

austinlparker (approved):

jpkrohling (approved):
Bringing my approval from the original PR.

songy23 (approved):
LGTM

jaronoff97 (approved):

tedsuo (approved):

jmacd (approved):
🎉

evan-bradley (approved):

docs/README.md
```diff
@@ -0,0 +1,60 @@
+# OpenTelemetry Collector
+
```
tigrannajaryan, line 2: Can we mark this document with a [stability level](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0232-maturity-of-otel.md#maturity-levels) to give us a chance to refine it before it is considered final?

tigrannajaryan: Not a blocker. Can be merged without it, but if we anticipate changes to the doc it would be good to have stability level listed.

tedsuo (approved):

docs/README.md
codeboten, line 2:
```suggestion

**Status**: [Beta](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0232-maturity-of-otel.md#beta)
```

docs/README.md
```diff
@@ -0,0 +1,61 @@
...
+**Status**: [Beta](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0232-maturity-of-otel.md#beta)
```
codeboten, line 3:
```suggestion
**Status**: [Beta](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0232-maturity-of-otel.md#beta)

```

PR #12421, opened 2025-02-18
Remove a lot of deprecated symbols

mx-psi:
#### Description

Remove deprecated symbols

#### Link to tracking issue
Updates #12222, #12305, #11524

PR #12444, opened 2025-02-20
Remove deprecated NewNopSettings functions

sincejune:
#### Description
This PR removes deprecated `NewNopSettings` functions which are missing in #12421 

cc @mx-psi 

#### Link to tracking issue
Relevant to #12305 

#### Testing
n/a

#### Documentation
n/a

### DIFF
--- a/.chloggen/removing-deprecated-newnopsettings.yaml
+++ b/.chloggen/removing-deprecated-newnopsettings.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: connector, exporter, extension, processor, receiver, scraper
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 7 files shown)


mx-psi:
@sincejune Thanks for the PR! What do you think is the best end state? Having a `NewNopSettingsWithType(typ)` function? Or having a `NewNopSettings(typ)` function? I was thinking about the latter, and so I was following [Example 3](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md#example-3---changing-the-arguments-of-a-function) from the coding guidelines. If you agree, could you, instead of removing `NewNopSettings`, change its signature to pass a the `component.Type` and deprecate `NewNopSettingsWithType`?

sincejune:
> @sincejune Thanks for the PR! What do you think is the best end state? Having a `NewNopSettingsWithType(typ)` function? Or having a `NewNopSettings(typ)` function? I was thinking about the latter, and so I was following [Example 3](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md#example-3---changing-the-arguments-of-a-function) from the coding guidelines. If you agree, could you, instead of removing `NewNopSettings`, change its signature to pass a the `component.Type` and deprecate `NewNopSettingsWithType`?

@mx-psi Sorry, didn't notice that we're still in progress of refactoring. I raised #12452 for the change instead.

PR #12450, opened 2025-02-21
Add support to configure min GC intervals for soft and hard limits

bogdandrutu:
When servers are in a high memory usage mode, the amount of times GC is called creates a high CPU usage which combined with a high ingest rate limits capability to offload existing queued data.

Configuring the minimum interval even for hard limit, allows the system to spend some CPU cycles between GCs to offload old data from queues/batch processor. The most amount of data I've seen accumulated are blocked in the batch processor queue (incoming channel) not in the exporter queue.


internal/memorylimiter/memorylimiter.go
```diff
@@ -50,11 +46,14 @@ type MemoryLimiter struct {
 
 	ticker *time.Ticker
 
-	lastGCDone time.Time
+	minGCIntervalWhenSoftLimited time.Duration
+	minGCIntervalWhenHardLimited time.Duration
+	lastGCDone                   time.Time
 
-	// The function to read the mem values is set as a reference to help with
+	// The functions to read the mem values and run GS are set as a reference to help with
```
jmacd, line 53:
```suggestion
	// The functions to read the mem values and run GC are set as a reference to help with
```

internal/memorylimiter/memorylimiter.go
```diff
@@ -180,38 +177,42 @@ func (ml *MemoryLimiter) CheckMemLimits() {
...
-	if !wasRefusing && mustRefuse {
+	if ml.usageChecker.aboveHardLimit(ms) {
+		// We are above hard limit, do a GC if it wasn't done recently and see if
+		// it brings memory usage below the soft limit.
+		if time.Since(ml.lastGCDone) > ml.minGCIntervalWhenHardLimited {
+			ml.logger.Warn("Memory usage is above hard limit. Forcing a GC.", memstatToZapField(ms))
+			ms = ml.doGCandReadMemStats()
+			// Check the limit again to see if GC helped.
+			aboveSoftLimit = ml.usageChecker.aboveSoftLimit(ms)
+		}
```
jmacd, line 199: nit: would be nice to see this factored into a function or closure that is called twice inside section of code.

bogdandrutu: 2 out of 4 lines of code have different arguments (the if statement and the log message). I though it will not help that much. If you feel strong, I will do it.

jmacd (approved):

PR #12466, opened 2025-02-24
POC for how to avoid duplicating WithStart and WithShutdown

bogdandrutu:
This way, we allow in the future to add more options in the component that will automatically be picked by every component implementation, as well as we remove duplicate definition.

### DIFF
--- a/processor/processorhelper/logs.go
+++ b/processor/processorhelper/logs.go
@@ -21,8 +21,7 @@ import (
 type ProcessLogsFunc func(context.Context, plog.Logs) (plog.Logs, error)
 
 type logs struct {
-	component.StartFunc
-	component.ShutdownFunc
+	component.Component
 	consumer.Logs
 }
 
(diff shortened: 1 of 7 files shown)


processor/processorhelper/logs.go
```diff
@@ -70,8 +69,7 @@ func NewLogs(
 	}
 
 	return &logs{
-		StartFunc:    bs.StartFunc,
-		ShutdownFunc: bs.ShutdownFunc,
-		Logs:         logsConsumer,
+		Component: component.NewComponent(bs.componentOptions...),
```
ms-jcorley, line 72: tiny nit:  The abbreviated name of this "bs" variable isn't clear entirely clear here (metrics.go and traces.go also use it).  The other files use names that are clearer from context.

ms-jcorley (approved):

mx-psi:
This looks alright to me, I think we can apply it everywhere

codeboten (commented):
i think this approach makes sense 👍🏻

PR #12467, opened 2025-02-24
[mdatagen] Add Converter and Provider support

evan-bradley:
#### Description

Adds support for confmap Converters and Providers to mdatagen, which allows us to declare stability levels for these module types.

Helps unblock https://github.com/open-telemetry/opentelemetry-collector-releases/pull/546.


bogdandrutu (commented):
What is special for them vs the generic "pkg"?

evan-bradley:
As a class of module: currently only the prefix (`pkg/x` vs. `converter/x` or `provider/x`) which are used in the labels in the generated readme. As a target for a certain level of stability: They're like extensions, so stability doesn't apply to a particular signal.

Considering the bulk of this change is just adding two new items to a few slices, I'm more inclined to explicitly add these as opposed to come up with exceptions for existing classes or make them work for more general cases. None of the existing options are really a great fit.

atoulme (approved):

MovieStoreGuy (approved):

jmacd (approved):

dmitryax (approved):

PR #12480, opened 2025-02-25
[extension/extensionauth] Deprecate methods on Func types

mx-psi:
#### Description

Depends on #12478, deprecates methods that are unnecessary

### DIFF
--- a/.chloggen/mx-psi_deprecate-func-methods.yaml
+++ b/.chloggen/mx-psi_deprecate-func-methods.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: extension/extensionauth
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deprecate methods on `*Func` types.
+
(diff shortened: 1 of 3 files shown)


jade-guiton-dd:
Do you think [this panic in contrib-tests](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/13544013530/job/37851308585?pr=12480) is relevant? It looks to me like that test should have panicked even with the previous code however ([creating an `auth.NewClient()` with no options](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/04d3b439bb4706d54a114e23115c735c3a13c721/receiver/splunkenterprisereceiver/client_test.go#L50) should result in a segfault when trying to call the function, whether it's embedded or not)

bogdandrutu (approved):

mx-psi:
> Do you think [this panic in contrib-tests](https://github.com/open-telemetry/opentelemetry-collector/actions/runs/13544013530/job/37851308585?pr=12480) is relevant? It looks to me like that test should have panicked even with the previous code however ([creating an `auth.NewClient()` with no options](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/04d3b439bb4706d54a114e23115c735c3a13c721/receiver/splunkenterprisereceiver/client_test.go#L50) should result in a segfault when trying to call the function, whether it's embedded or not)

I added the nil check present in the Func types: 38d366c358e1ecb0c71bf9c3fe0d047bfc69b03e. I think that should solve the panic, thanks for pointing this out

PR #12464, opened 2025-02-24
[mdatagen] Add deprecation date and migration note fields for deprecated components

ChrsMark:
#### Description

This PR adds deprecation date and migration note fields for deprecated components as described at https://github.com/open-telemetry/opentelemetry-collector/issues/12359.

Example metadata file:

```yaml
status:
  class: receiver
  stability:
    development: [logs]
    beta: [traces]
    stable: [metrics]
    deprecated: [profiles]
  deprecation:
    profiles:
      migration: "no migration needed"
      date: "2025-02-05"
```

Example README.md:

| Status        |           |
| ------------- |-----------|
| Stability     | [deprecated]: profiles   |
|               | [development]: logs   |
|               | [beta]: traces   |
|               | [stable]: metrics   |
| Deprecation of profiles | [Date]: 2025-02-05   |
|                      | [Migration Note]: no migration needed   |

[deprecated]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#deprecated
[development]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#development
[beta]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#beta
[stable]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#stable
[Date]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#deprecation-information
[Migration Note]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#deprecation-information

I'd appreciate any suggestions if there is a better way to represent this information in the markdown table. 

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/12359

#### Testing

Added

#### Documentation

Added

/cc @atoulme who filed the feature request issue originally.

PR #12484, opened 2025-02-25
[chore] ignore extra metadata fields

atoulme:
#### Description
As we need to be able to move forward with mdatagen changes, we want to relax how it reads metadata yaml to ignore extra fields so we can roll out changes without breaking existing parsing.


ChrsMark (approved):
I think it's not a big deal ignoring extra fields unless I miss something important here.

LGTM

cmd/mdatagen/internal/loader.go
```diff
@@ -38,7 +39,7 @@ func LoadMetadata(filePath string) (Metadata, error) {
 	}
 
 	md := Metadata{ShortFolderName: shortFolderName(filePath), Tests: Tests{Host: "componenttest.NewNopHost()"}}
-	if err = conf.Unmarshal(&md); err != nil {
+	if err = conf.Unmarshal(&md, confmap.WithIgnoreUnused()); err != nil {
```
ms-jcorley, line 42: It is slightly sad that the tooling doesn't offer an easy way to "warn" for unused fields.  My only fear with this change is that slight typos will get easily missed.  But I don't see an easy way around it.

ms-jcorley (approved):

cmd/mdatagen/internal/loader.go
atoulme: You are welcome to file requests for enhancements to offer additional support to warn instead of ignoring fields.

dmitryax (commented):
Please consider moving without this https://github.com/open-telemetry/opentelemetry-collector/pull/12464#issuecomment-2683937917

dmitryax (changes requested):
I don't think we should allow random fields going forward unless we have a good reason for that. It'll be prone to typos

atoulme:
Your approach is better than what I was putting together. I will go with it. Thank you!

PR #12490, opened 2025-02-25
[chore] make contrib tests more efficient

atoulme:
Instead of having each test group edit modules, tidy and generate, do it once and share the output in each subtest to save 5 minutes per branch.


bogdandrutu (approved):

songy23 (approved):
👍

codeboten (approved):

.github/workflows/contrib-tests.yml
```diff
@@ -17,8 +17,26 @@ concurrency:
...
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'Skip Contrib Tests') }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Prepare Contrib Tests
+        run: |
+          contrib_path=/tmp/opentelemetry-collector-contrib
+          git clone --depth=1 https://github.com/open-telemetry/opentelemetry-collector-contrib.git $contrib_path
+          make CONTRIB_PATH=$contrib_path prepare-contrib
+      - uses: actions/upload-artifact@v4
```
dmitryax, line 31: Doesn't it mean that we will be testing against one contrib version for 90 days?

dmitryax: I'm looking at https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period. Should we add git sha in the `contrib_path`?

atoulme: ouch. I'll open a new PR to add that

dmitryax: Why do we need to upload it then? Should we use https://github.com/actions/cache instead?

atoulme: I followed the recipe I have seen used most of the time. A cache is a step better I didn't look into because I was afraid of having cache issues. That would shave a few dozen more seconds off the build.

PR #12496, opened 2025-02-26
[chore] Do not introduce pointers in the deprecated batcher fields

dmitryax:
Do not introduce an unnecessary Go API breaking change.

Fixes contrib tests.

A follow-up to https://github.com/open-telemetry/opentelemetry-collector/pull/12486.

### DIFF
--- a/exporter/exporterbatcher/config.go
+++ b/exporter/exporterbatcher/config.go
@@ -42,13 +42,13 @@ type SizeConfig struct {
 // Deprecated: [v0.121.0] use SizeConfig.
 type MinSizeConfig struct {
 	// Deprecated: [v0.121.0] use SizeConfig.MinSize.
-	MinSizeItems *int `mapstructure:"min_size_items"`
+	MinSizeItems int `mapstructure:"min_size_items"`
 }
 
 // Deprecated: [v0.121.0] use SizeConfig.
 type MaxSizeConfig struct {
 	// Deprecated: [v0.121.0] use SizeConfig.MaxSize.
-	MaxSizeItems *int `mapstructure:"max_size_items"`
+	MaxSizeItems int `mapstructure:"max_size_items"`
 }
 
 func (c *Config) Unmarshal(conf *confmap.Conf) error {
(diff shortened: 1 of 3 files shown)


exporter/exporterhelper/internal/base_exporter.go
```diff
@@ -70,7 +70,7 @@ func NewBaseExporter(set exporter.Settings, signal pipeline.Signal, options ...O
 	}
 
 	//nolint: staticcheck
-	if be.batcherCfg.MinSizeItems != nil || be.batcherCfg.MaxSizeItems != nil {
+	if be.batcherCfg.MinSizeItems != 0 || be.batcherCfg.MaxSizeItems != 0 {
```
dmitryax, line 73: Potentially, we could update the `Unmarshal` func to set a private field indicating whether `min/max_size_items` were set, but that would require moving/aliasing the `exporterbatcher.Config` in some internal module along with a function showing this warning. I could do that separately.

exporter/exporterbatcher/config.go
```diff
@@ -42,26 +42,26 @@ type SizeConfig struct {
...
 
 func (c *Config) Unmarshal(conf *confmap.Conf) error {
 	if err := conf.Unmarshal(c); err != nil {
 		return err
 	}
 
-	if c.MinSizeItems != nil {
-		c.SizeConfig.MinSize = *c.MinSizeItems
+	if conf.IsSet("min_size_items") {
+		c.SizeConfig.MinSize = c.MinSizeItems
```
jade-guiton-dd, line 60: Not directly related to the PR, but the deprecation comment says that min_size_items is "ignored if SizeConfig is set". Doesn't this code do the opposite?

dmitryax:
Some contrib tests still needs to be updated. So maybe it's ok to switch to pointers to keep it simple. I'll proceed with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38214

exporter/exporterbatcher/config.go
dmitryax: Submitted https://github.com/open-telemetry/opentelemetry-collector/pull/12502

PR #12500, opened 2025-02-26
document breaking change in metric names for prometheus exporter

### LINKED ISSUE #12458: [service] Internal counter metrics now have a `_total` suffix in Prometheus
### Component(s)

service

### What happened?

**Describe the bug**

Between v0.118.0 and v0.120.0, the name of internal counter metrics exposed through the Prometheus endpoint gained a `_total` suffix, breaking the stability of internal metrics.

**Steps to reproduce**

Exercise the OTLP receiver (sending a span for example) with an explicitly configured Prometheus reader.

Old output on the Prometheus endpoint:
```
# HELP otelcol_receiver_accepted_spans Number of spans successfully pushed into the pipeline. [alpha]
# TYPE otelcol_receiver_accepted_spans counter
otelcol_receiver_accepted_spans{receiver="otlp/1",service_instance_id="dfcfad0f-6d19-4340-9d76-1e0df4883526",service_name="otelcol",service_version="0.118.0",transport="grpc"} 123
```

**What did you see instead?**

New output on the Prometheus endpoint:
```
# HELP otelcol_receiver_accepted_spans_total Number of spans successfully pushed into the pipeline. [alpha]
# TYPE otelcol_receiver_accepted_spans_total counter
otelcol_receiver_accepted_spans_total{otel_scope_name="go.opentelemetry.io/collector/receiver/receiverhelper",otel_scope_version="",receiver="otlp/1",transport="grpc"} 123
```

### Collector version

v0.120.0

### OpenTelemetry Collector configuration

```yaml
service:
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: '0.0.0.0'
                port: 8888
```

### Additional context

This was reported by Jon Bates on the CNCF Slack.

I believe this was introduced by #11611. The previous code for instantiating the Prometheus exporter explicitly added the `WithoutCounterSuffixes` option, but the new code now relies on the go-contrib config module to instantiate it based on the user's config.

While the default `telemetry` config has a Prometheus reader with the corresponding `without_type_suffix: true` option set, I believe that setting your own list of readers without specifying `without_type_suffix` will make it default to false, adding the `_total` suffix to counter metrics.

A workaround is to add `without_type_suffix: true` to the `prometheus:` section.

codeboten:
Fixes #12458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@ If you are looking for developer-facing changes, check out [CHANGELOG-API.md](./
 - `service`: Value for telemetry exporter `otlp.protocol` updated from `grpc/protobuf` to `grpc`. (#12337)
 - `service`: internal metrics exported over Prometheus may differ from previous versions. (#11611)
 
-  The change to update the internal telemetry to use [otel-go config](https://pkg.go.dev/go.opentelemetry.io/contrib/config) can cause unexpected behaviour
+  Users who do not customize the Prometheus reader should not be impacted. The change to update the internal telemetry to use [otel-go config](https://pkg.go.dev/go.opentelemetry.io/contrib/config) can cause unexpected behaviour
   for end users. This change is caused by the default values in `config` being different from what the Collector has used in previous versions. The
   following changes can occur when users configure their `service::telemetry::metrics::readers`:
   - the metric name will append a `_total` suffix if `without_type_suffix` is not configured. Set `without_type_suffix` to `true` to disable this.


CHANGELOG.md
```diff
@@ -51,6 +51,11 @@ If you are looking for developer-facing changes, check out [CHANGELOG-API.md](./
 - `cmd/mdatagen`: Remove `level` field from metrics definition (#12145)
   This mechanism will be added back once a new views mechanism is implemented.
 - `service`: Value for telemetry exporter `otlp.protocol` updated from `grpc/protobuf` to `grpc`. (#12337)
+- `service`: internal metrics exported over Prometheus will append a `_total` suffix to the metric name if `without_type_suffix` is not configured. (#11611)
+  The change to update the internal telemetry to use [otel-go config](https://pkg.go.dev/go.opentelemetry.io/contrib/config) can cause unexpected behaviour
+  for end users. The default value in `config` for `without_type_suffix` is `false` as per the specification which causes the `_total` suffix to be appended. Users
+  that configure their `service::telemetry::metrics::readers` can set `without_type_suffix` to `true` to keep their metrics name consistent with previous versions
+  of the collector.
```
bogdandrutu, line 58: Extra problem that was not clearly documented in the issue: same applies with unit. Also an extra metric the target_info.

codeboten: Updated to include those options as well

songy23 (approved):

jmacd (approved):

CHANGELOG.md
```diff
@@ -51,6 +51,14 @@ If you are looking for developer-facing changes, check out [CHANGELOG-API.md](./
 - `cmd/mdatagen`: Remove `level` field from metrics definition (#12145)
   This mechanism will be added back once a new views mechanism is implemented.
 - `service`: Value for telemetry exporter `otlp.protocol` updated from `grpc/protobuf` to `grpc`. (#12337)
+- `service`: internal metrics exported over Prometheus may differ from previous versions. (#11611)
+
+  The change to update the internal telemetry to use [otel-go config](https://pkg.go.dev/go.opentelemetry.io/contrib/config) can cause unexpected behaviour
+  for end users. This change is caused by the default values in `config` being different from what the Collector has used in previous versions. The
+  following changes can occur when users configure their `service::telemetry::metrics::readers`:
```
songy23, line 58: QQ: does it mean as long as you don't explicitly configure `readers` you won't run into this? e.g. if I just have
```
service:
  telemetry:
    metrics:
      level: detailed
```

the behavior of this should be the same in all versions, right?

bogdandrutu (approved):

CHANGELOG.md
dmitryax: Yes, I think it worth mentioning that the breaking change only affects user how customized the prometheus reader in any way. The default behavior is the same

CHANGELOG.md
```diff
@@ -51,6 +51,14 @@ If you are looking for developer-facing changes, check out [CHANGELOG-API.md](./
 - `cmd/mdatagen`: Remove `level` field from metrics definition (#12145)
   This mechanism will be added back once a new views mechanism is implemented.
 - `service`: Value for telemetry exporter `otlp.protocol` updated from `grpc/protobuf` to `grpc`. (#12337)
+- `service`: internal metrics exported over Prometheus may differ from previous versions. (#11611)
+
+  The change to update the internal telemetry to use [otel-go config](https://pkg.go.dev/go.opentelemetry.io/contrib/config) can cause unexpected behaviour
```
codeboten, line 56:
```suggestion
  Users who do not customize the Prometheus reader should not be impacted. The change to update the internal telemetry to use [otel-go config](https://pkg.go.dev/go.opentelemetry.io/contrib/config) can cause unexpected behaviour
```

CHANGELOG.md
codeboten: Tried to add a note at the beginning to capture this, ptal @dmitryax @songy23

songy23 (approved):
LGTM

Issue #12191, opened 2025-01-27
Add `omitempty` to all appropriate config struct fields


evan-bradley:
**Is your feature request related to a problem? Please describe.**

Right now when marshaling our config structs into YAML, they produce a large number of keys that are unset. We should get as close to only emitting fields that have been actually set as possible.

**Describe the solution you'd like**

Use mapstructure's [`omitempty`](https://pkg.go.dev/github.com/go-viper/mapstructure/v2#hdr-Omit_Empty_Values) suffix to omit these values from serialization.

**Describe alternatives you've considered**

We can add an option to `confmap.Marshal` that forces this regardless of whether it has been set on the struct. We may need this anyway, but it would be nice if our config structs handled this out-of-the-box.

PR #12501, opened 2025-02-26
[service] Add `omitempty` tag to config fields

evan-bradley:
#### Description

Update config structs in the service module to include `omitempty` tags.

#### Link to tracking issue
Works toward https://github.com/open-telemetry/opentelemetry-collector/issues/12191

### DIFF
--- a/service/telemetry/config.go
+++ b/service/telemetry/config.go
@@ -30,13 +30,13 @@ var disableAddressFieldForInternalTelemetryFeatureGate = featuregate.GlobalRegis
 type Config struct {
 	Logs    LogsConfig    `mapstructure:"logs"`
 	Metrics MetricsConfig `mapstructure:"metrics"`
-	Traces  TracesConfig  `mapstructure:"traces"`
+	Traces  TracesConfig  `mapstructure:"traces,omitempty"`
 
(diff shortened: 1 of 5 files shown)


service/telemetry/config.go
```diff
@@ -30,13 +30,13 @@ var disableAddressFieldForInternalTelemetryFeatureGate = featuregate.GlobalRegis
 type Config struct {
 	Logs    LogsConfig    `mapstructure:"logs"`
 	Metrics MetricsConfig `mapstructure:"metrics"`
```
songy23, line 32: Do these need `omitempty` too?

songy23 (approved):

service/telemetry/config.go
evan-bradley: I probably should have included this in the PR description, my bad. I only added `omitempty` to fields where the zero value is also the default. Otherwise the zero value won't be emitted during serialization and it would be impossible to generate configs that set the zero value.

For these fields, we have default configurations for both the `Logs` and `Metrics` fields, so I didn't add `omitempty` so they can be unset if need be.

codeboten (approved):

PR #12508, opened 2025-02-26
[component] Add namespace parameter to component.BuildInfo

### LINKED ISSUE #12505: [component] Add `service.namespace` semantic convention to component.BuildInfo
### Component(s)

_No response_

**Problem:**
In OpAMP we use [Identifying Attributes](https://opentelemetry.io/docs/specs/opamp/#agentdescriptionidentifying_attributes) to help differentiate collectors. These attributes are based on a set of [Resource Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/resource/#service). Currently all of these conventions except `service.namespace` are reported via the OpAMP extension. The addition of `service.namespace` would enable OpAMP servers to better differentiate collector types rather than relying on just `service.name` and `service.version`. `service.name` and `service.version` are insufficient on their own because `service.name` is set to the collector binary name which has a high chance of being the same across distributions. 

**Solution:**
Add a new field to `component.BuildInfo` called `Namespace`. All components would have access to this new field, specifically the OpAMP extension. By default this value would be set to "opentelemetry". The OCB should also be updated with a new configuration parameter, `distribution.namespace`, to facilitate setting this value.

**Context:**
There was a previous [issue](https://github.com/open-telemetry/opentelemetry-collector/issues/6599) that aimed to solve this, but `service.namespace` wasn't added. Looking at the issue and [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/7531) there doesn't appear to be any reason why not.

dpaasman00:
#### Description

Adds a new field to `component.BuildInfo` to facilitate support for reporting the `service.namespace` semantic convention in OpAMP. Includes corresponding updates to resource attributes reported by the collector. Also adds a new parameter to the builder configuration to facilitate setting this via the OCB.

#### Link to tracking issue
Fixes #12505 

Need follow up PR to OpAMP extension making use of this value.

#### Testing
Relevant unit tests updated. 

Running the updated otelcorecol with logging enabled in `service.telemetry` and the new resource attribute for `service.namespace` shows up. This was done sending to another collector running the otlp receiver and logging via the debug exporter.

[02-26-25_monitoring-output.txt](https://github.com/user-attachments/files/18993755/02-26-25_monitoring-output.txt)

### DIFF
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -67,6 +67,7 @@ type ConfResolver struct {
 type Distribution struct {
 	Module      string `mapstructure:"module"`
 	Name        string `mapstructure:"name"`
+	Namespace   string `mapstructure:"namespace"`
 	Go          string `mapstructure:"go"`
 	Description string `mapstructure:"description"`
 	// Deprecated: [v0.113.0] only here to return a detailed error and not failing during unmarshalling.
--- a/service/internal/resource/config.go
+++ b/service/internal/resource/config.go
@@ -28,6 +28,11 @@ func New(buildInfo component.BuildInfo, resourceCfg map[string]*string) *resourc
 		telAttrs = append(telAttrs, attribute.String(semconv.AttributeServiceName, buildInfo.Command))
 	}
 
+	if _, ok := resourceCfg[semconv.AttributeServiceNamespace]; !ok {
+		// AttributeServiceNamespace is not specified in the config. Use the default namespace.
+		telAttrs = append(telAttrs, attribute.String(semconv.AttributeServiceNamespace, buildInfo.Namespace))
+	}
+
 	if _, ok := resourceCfg[semconv.AttributeServiceInstanceID]; !ok {
 		// AttributeServiceInstanceID is not specified in the config. Auto-generate one.
 		instanceUUID, _ := uuid.NewRandom()
--- a/service/internal/resource/config.go
+++ b/service/internal/resource/config.go
@@ -29,7 +29,7 @@ func New(buildInfo component.BuildInfo, resourceCfg map[string]*string) *resourc
 	}
 
 	if _, ok := resourceCfg[semconv.AttributeServiceNamespace]; !ok {
-		// AttributeServiceNamespace is not specified in the config. Use the default namespace.
+		// AttributeServiceNamespace is not specified in the config. Use the default namespace built into the binary.
 		telAttrs = append(telAttrs, attribute.String(semconv.AttributeServiceNamespace, buildInfo.Namespace))
 	}
 
--- a/.chloggen/dakotapaasman_bpop-2116-configurable-servicenamespace-value-in-ocb.yaml
+++ b/.chloggen/dakotapaasman_bpop-2116-configurable-servicenamespace-value-in-ocb.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: component
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add namespace field to component.BuildInfo to support service.namespace semantic convention
+
+# One or more tracking issues or pull requests related to the change
+issues: [12505]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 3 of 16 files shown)


iblancasa (approved):

djaglowski (approved):

cmd/builder/internal/builder/config.go
```diff
@@ -67,6 +67,7 @@ type ConfResolver struct {
 type Distribution struct {
 	Module      string `mapstructure:"module"`
 	Name        string `mapstructure:"name"`
+	Namespace   string `mapstructure:"namespace"`
```
evan-bradley, line 70: Could you add this field to the config example in the builder readme?

cmd/otelcorecol/otelcorecol
evan-bradley, line 1: I think this file was accidentally committed.

service/internal/resource/config.go
```diff
@@ -28,6 +28,11 @@ func New(buildInfo component.BuildInfo, resourceCfg map[string]*string) *resourc
 		telAttrs = append(telAttrs, attribute.String(semconv.AttributeServiceName, buildInfo.Command))
 	}
 
+	if _, ok := resourceCfg[semconv.AttributeServiceNamespace]; !ok {
+		// AttributeServiceNamespace is not specified in the config. Use the default namespace.
```
evan-bradley, line 32:
```suggestion
		// AttributeServiceNamespace is not specified in the config. Use the namespace built into the binary.
```
If possible I think we should specify where the default comes from.

evan-bradley (commented):
Overall looks good to me.

evan-bradley (approved):

dpaasman00:
Contrib tests failing

bogdandrutu:
I am not sure I understand, how this. Why do you not configure the "service.namespace" in the `service::telemetry::resource`?

bogdandrutu (changes requested):
The `service.namespace` is a runtime identification configuration and not a build time identification configuration, so I am confused why we need a change to the build info.

evan-bradley:
@dpaasman00 I've discussed this offline with other core leads and received some feedback that we may need to make some changes. Could you make the following updates to this PR?

1. Remove `Namespace` from `component.BuildInfo`. `component` is a stable module, so we need to be absolutely sure of anything we put in there. We will likely need to pass information from the builder down to the service through a new parameter for now, and we can consider moving it to `component.BuildInfo` in the future. Additionally, is there a term other than `namespace` that we could use to uniquely identify distributions in a way that doesn't include the binary name?
2. Clarify whether `service.namespace` is the correct attribute for something set at build time. My initial impression was the same as Bogdan's, that service.namespace is primarily intended to describe the runtime context of a service than describe its functionality, but reading the docs and seeing how we set `service.name`, I'm not so clear on this. Could you determine whether we should set this information in `service.namespace`, or whether `service.name` may be more appropriate? If we don't set `service.namespace`, maybe we add a `distro` field and make the `service.name` attribute something like `opentelemetry.otelcol`.

evan-bradley:
A relatively minor aside: I read through [the docs](https://opentelemetry.io/docs/specs/semconv/resource/#service) again and saw this:

>Note: `service.namespace` and `service.name` are not intended to be concatenated for the purpose of forming a single globally unique name for the service.

I don't think this strictly impacts anything here since we can define the semantics if we concatenate the distro and binary inside `service.name`, but we do need to think about what it means if we put the distro into `service.namespace`.

dpaasman00:
After spending some time thinking about this, I agree that `service.namespace` is not the correct attribute for what we're trying to accomplish. The goal is to better differentiate collectors now that the OCB enables users to create and manage their own distributions and versions of the collector. To achieve this, I think we should add 2 new attributes to the `service` resource semantic convention, `service.description` and `service.distribution`. 

- `service.description` would be a long, human friendly description of what the collector is intended for. There is already a `distribution.description` config parameter for the builder that gets added to the BuildInfo, this would just need to get reported by the OpAMP extension once the semantic convention is established.
- `service.distribution` would represent who created and manages the manifest the collector was built with. Implementing this attribute would require adding a new parameter to the builder, adding a new field to BuildInfo, and updating the OpAMP extension to report this value once the semantic convention is established.

I understand this approach would take a while since we're changing the semantic conventions and BuildInfo, but I do think it is a better approach to the original problem this PR was trying to solve. 

@evan-bradley @bogdandrutu What do you think of this approach?

evan-bradley:
Thanks for your thoughts, I'm largely in agreement with everything you wrote. There is an existing proposal for something similar, but under the `agent` attribute namespace: https://github.com/open-telemetry/semantic-conventions/pull/950. I think it would be worth checking on that PR or at the next OpAMP SIG meeting how we should proceed there.

PR #12509, opened 2025-02-27
Remove exporterqueue.Factory and simplify exportehelper.WithQueueRequest

bogdandrutu:
#### Link to tracking issue
Fixes #

### DIFF
--- a/.chloggen/break-exporterqueue-factory.yaml
+++ b/.chloggen/break-exporterqueue-factory.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterqueue
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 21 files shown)


.chloggen/break-exporterqueue-factory.yaml
```diff
@@ -0,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: exporterqueue
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove exporterqueue.Factory in favor of the NewQueue function, and merge configs for memory and persistent.
```
dmitryax, line 10: So it means that custom queue implementations are not allowed anymore? I guess that's ok. But we should probably explicitly call this out here. Also, the `exporterqueue.Queue` interface and all related API can be hidden now as well, right?

bogdandrutu: Correct we can hide all of that. I think if customer wants custom queue they can have their own exporter pipeline, since queue is such a core of the pipeline now.

bogdandrutu: Updated the subtext.

dmitryax: Sounds good to me

dmitryax (approved):

PR #12511, opened 2025-02-27
[chore][extensionauth] Return error on NewClient and NewServer constructors

mx-psi:
#### Description

Adds errors in return type to `extensionauth.New*` functions. This is not a breaking change since `extensionauth` is an unreleased module.

### DIFF
--- a/config/configauth/configauth_test.go
+++ b/config/configauth/configauth_test.go
@@ -17,6 +17,13 @@ import (
 
 var mockID = component.MustNewID("mock")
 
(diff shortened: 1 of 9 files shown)


jade-guiton-dd (approved):
LGTM

jade-guiton-dd:
Are there plans to do something with these errors or is this "just in case"?

mx-psi:
> Are there plans to do something with these errors or is this "just in case"?

This is a 'just in case', so that we don't have to define a new function in case we need to return errors in the future. The main way I would see errors happening would be if there are incompatible options

Issue #11754, opened 2024-11-26
[component] Add support for components specifying what instruments to enable depending on metric level


mx-psi:
After doing #11061, we still need a way for components that make use of `service::telemetry::metrics::level` to specify what instruments are enabled at which level (or figure out how to use this exactly).

We need this at SDK creation time so this would probably be factory option.

If we want to sidestep this issue, we could hardcode things for the components that use this  ([batch processor](https://github.com/open-telemetry/opentelemetry-collector/blob/e27c4aeae8da140e286534c69fb9f8231b4b79a2/processor/batchprocessor/metrics.go#L48), [configgrpc](https://github.com/open-telemetry/opentelemetry-collector/blob/e27c4aeae8da140e286534c69fb9f8231b4b79a2/config/configgrpc/configgrpc.go#L486), [confighttp](https://github.com/open-telemetry/opentelemetry-collector/blob/e27c4aeae8da140e286534c69fb9f8231b4b79a2/config/confighttp/confighttp.go#L560), [otelarrow](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/28f23aaa8bb29b9af004e939d8d657dbd4e9950e/receiver/otelarrowreceiver/otelarrow.go#L145)) instead. 

cc @bogdandrutu

mx-psi:
This can be done in a backwards compatible way (a method on the Factory interface) and we have been able to implement a workaround, so, while we need to do this at some point, I am going to remove it from Collector 1.0

PR #12525, opened 2025-02-28
[service] Move batchprocessor metrics to normal level and update level guidelines

### LINKED ISSUE #7890: Revisit how collector internal metric are distributed across telemetry levels

jade-guiton-dd:
#### Description

This PR:
- requires "level: normal" before outputting batch processor metrics (in addition to one specific metric which was already restricted to "level: detailed")
- clarifies wording in the telemetry level guidelines and documentation, and adds said guidelines to the requirements for stable components.

Some rationale for these changes can be found in the tracking issue and [this comment](https://github.com/open-telemetry/opentelemetry-collector/issues/7890#issuecomment-2684652956).

#### Link to tracking issue
Resolves #7890

#### To be discussed

Should we add a feature gate for this, in case a user relies on "level: basic" outputting batch processor metrics? This feels like a niche use case, so considering the "alpha" stability level of these metrics, I don't think it's really necessary.

Considering batch processor metrics had already been switched to "normal" once (#9767), but were turned back to basic at some later point (not sure when), we might also want to add tests to avoid further regressions (especially as the handling of telemetry levels is bound to change further with #11754).


dmitryax (approved):
LGTM. Thanks, @jade-guiton-dd!

docs/component-stability.md
```diff
@@ -214,6 +214,9 @@ If data can be dropped/created/held at multiple distinct points in a component's
 scraping, validation, processing, etc.), it is recommended to define additional attributes to help
 diagnose the specific source of the discrepancy, or to define different signals for each.
 
+The breakdown of emitted telemetry per telemetry level (basic / normal / detailed) should follow
+the guidelines in [the Go package documentation for `configtelemetry`](config/configtelemetry/doc.go).
```
dmitryax, line 218: I hope this will fix the failing links check
```suggestion
the guidelines in [the Go package documentation for `configtelemetry`](/config/configtelemetry/doc.go).
```

PR #12528, opened 2025-02-28
[builder] add doc generation

atoulme:
#### Description
Adds a go script that you can run manually as part of the distribution to output documentation.

The script takes an optional argument pointing to a go text template file. If no argument is given, the script will use the default `md.tmpl` file to generate a set of markdown tables for each component categories.


mx-psi (commented):
Should this be a separate subcommand? It feels independent to me from the main command

PR #12531, opened 2025-03-02
[mdatagen] Add variable for metric name

### LINKED ISSUE #12459: Access names from `mdatagen` generated metrics
### Component(s)

cmd/mdatagen

I have a use case where I'm trying to produce a configuration validation error for a receiver that intends to reference a particular metric name. In particular the scenario is if a particular technology is disabled but a particular metric that requires that subsystem is enabled, I want to produce an error like:

`Metric metric.name requires subsystem to be enabled`

What would be great would be to reference the name directly from the `mdatagen` generated package, as putting the metric name directly in the error isn't future-proof against the metric name changing.

I have no particular preference for what this API looks like, provided it allows me to access any metric or resource attribute names. If there already is a way to do this and I've missed it, I will close this issue.

sincejune:
#### Description
Add variable for metric name in mdatagen, accessing via `metadata.MetricsInfo.<metric-variable>.Name`

#### Link to tracking issue
Fixes #12459 

#### Testing
n/a

#### Documentation
Added


sincejune:
I will address the contrib test issue by updating otel PR once this is merged

cmd/mdatagen/internal/templates/config.go.tmpl
```diff
@@ -12,7 +12,8 @@ import (
 
 // MetricConfig provides common config for a particular metric.
 type MetricConfig struct {
-	Enabled bool `mapstructure:"enabled"`
+	Enabled bool   `mapstructure:"enabled"`
+	Name    string `mapstructure:"name"`
```
dmitryax, line 16: Why are we adding this to the end-user config while it doesn't have any effect if user changes it? AFAIU goal of https://github.com/open-telemetry/opentelemetry-collector/issues/12459 is to have the metric name generated and available for a component developer, not end user.

sincejune: fixed in https://github.com/open-telemetry/opentelemetry-collector/pull/12531/commits/030ce9ec3cc9ba9d8e295444215562723c856e0f

sincejune: fixed in https://github.com/open-telemetry/opentelemetry-collector/pull/12531/commits/030ce9ec3cc9ba9d8e295444215562723c856e0f

sincejune: fixed in https://github.com/open-telemetry/opentelemetry-collector/pull/12531/commits/030ce9ec3cc9ba9d8e295444215562723c856e0f

cmd/mdatagen/internal/sampleconnector/internal/metadata/generated_config.go
```diff
@@ -9,7 +9,8 @@ import (
 
 // MetricConfig provides common config for a particular metric.
 type MetricConfig struct {
-	Enabled bool `mapstructure:"enabled"`
+	Enabled bool   `mapstructure:"enabled"`
+	Name    string `mapstructure:"-"`
```
dmitryax, line 13: Even if this is not available to the end user, it's still part of `Config` Go public API which is confusing. Why don't generate a separately, as part of `MetricsBuilder` maybe, e.g. `MetricsBuilder.Info.MyMetric.Name`?

sincejune: updated

dmitryax (approved):

Issue #7444, opened 2023-03-28
[CI] Migrate 'Inform Incompatible PRs / Check-Compatibility' check to Ubuntu runner


mx-psi:
The 'Inform Incompatible PRs / Check-Compatibility' check runs on a macOS machine: https://github.com/open-telemetry/opentelemetry-collector/blob/92e542357151fb3ae237035322b0143954655389/.github/workflows/api-compatibility.yml#L16-L17 This is inconsistent with the rest of our test suite, that (generally) runs on `ubuntu-latest`. It also eats into the org's CI minutes: [1 minute in a macOS runner is equivalent to 10 minutes on a Linux runner](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers).

The main (only?) thing to change is this call to `mktemp`: https://github.com/open-telemetry/opentelemetry-collector/blob/a25d4385afca78db00dea994a784aed44dd740db/internal/buildscripts/gen-apidiff.sh#L45 which uses macOS-specific options. Changing this I am able to run this without errors on an Ubuntu machine, but we may want to tackle #7443 first to ensure the targets do not silently fail on Linux because of something else.

mowies:
@mx-psi PR should be ready for review 🚀

PR #12549, opened 2025-03-04
[chore] move api-diff pipeline to ubuntu runner

mowies:
#### Description
Move the api-diff workflow to the standard `ubuntu-latest` runner and adjust the OS dependent call to mktemp so that it works for both linux and macos.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/7444

#### Testing

Tested on my fork:
https://github.com/mowies/opentelemetry-collector/actions/runs/13941800348/job/39019991437

Also tested this locally on MacOS.


mowies:
The workflow is failing in the PR because it still uses the version from main branch I believe. It works on my fork and locally on macOS

mx-psi:
> The workflow is failing in the PR because it still uses the version from main branch I believe. It works on my fork and locally on macOS

Any way to fix that before merging this PR?

mowies:
I don't think so...

mx-psi (approved):

PR #12563, opened 2025-03-05
[chore] Simplify release doc

mx-psi:
#### Description
Simplifies description of automated release steps. While there is some value in having the description of the automated steps somewhere, I think this runs the risk of getting outdated and us having to look at the code directly, so I would rather just remove it from here and improve the comments/code of the automation over time.


songy23 (approved):
cc @jackgopack4

atoulme (approved):

jackgopack4:
thanks, looks great

PR #12567, opened 2025-03-05
[extensionauthtest] Remove extensionauthtest.MockClient

mx-psi:
#### Description

Removes `extensionauthtest.MockClient`. Users can use either `extensionauth.NewClient` or use `extensionauthtest.NewErrorClient` as a convenience method to replicate the `MustError` field behavior.

Doing a Github search, I am unable to find usages of this outside of contrib, which I dealt with on PR open-telemetry/opentelemetry-collector-contrib/pull/38401, thus I believe this should be fine to remove in one go.


songy23 (approved):
need to fix lint http://github.com/open-telemetry/opentelemetry-collector/actions/runs/13703523758/job/38323128465?pr=12567

PR #12574, opened 2025-03-06
[extensionauth] Split extensionauth.Client by protocol type

mx-psi:
#### Description

An attempt at splitting `extensionauth.Client` by protocol type. open-telemetry/opentelemetry-collector-contrib/pull/38451 removes usages of the `NewClient` and `NewServer` constructor


extension/extensionauth/client.go
```diff
@@ -12,27 +12,34 @@ import (
...
 
 	// RoundTripper returns a RoundTripper that can be used to authenticate HTTP requests.
 	RoundTripper(base http.RoundTripper) (http.RoundTripper, error)
+}
+
+// GRPCClient is an Extension that can be used as an HTTP authenticator for the configauth.Authentication option.
+// Authenticators are then included as part of OpenTelemetry Collector builds and can be referenced by their
+// names from the Authentication configuration.
+type GRPCClient interface {
+	extension.Extension
```
bogdandrutu, line 29: I don't think we should follow this model, but instead the optional interface model, which means that we will only define the  `PerRPCCredentials()` func.

If we do this, implementations can pick to implement both http and gRPC or just one of them. Also, allows implementations to implement also other types of extensions.

extension/extensionauth/client.go
```diff
@@ -42,25 +49,29 @@ type ClientRoundTripperFunc func(base http.RoundTripper) (http.RoundTripper, err
...
 
-var _ Client = (*defaultClient)(nil)
+var (
+	_ HTTPClient = (*Client)(nil)
+	_ GRPCClient = (*Client)(nil)
+)
 
-type defaultClient struct {
+// Client is an Extension that can be used as an HTTP and gRPC authenticator for the configauth.Authentication option.
+type Client struct {
```
bogdandrutu, line 58: I don't think we need this anymore if we go with the optional interface as suggested.

extension/extensionauth/extensionauthtest/error_client.go
```diff
@@ -4,25 +4,48 @@
...
+	return nil
+}
+
+// Start implements extension.Extension.
+func (e *errorClient) Start(context.Context, component.Host) error {
+	return nil
+}
+
+// NewErrorClient returns a new [extension.Extension] that implements the [extensionauth.HTTPClient] and [extensionauth.GRPCClient] and always returns an error on both methods.
+func NewErrorClient() extension.Extension {
```
bogdandrutu, line 49: I would rather expect an error as an argument and return that.

extension/extensionauth/extensionauthtest/error_client.go
```diff
@@ -4,25 +4,48 @@
...
-		}),
-	}
-	return extensionauth.NewClient(append(errorOpts, opts...)...)
+var (
+	_ extension.Extension      = (*errorClient)(nil)
+	_ extensionauth.HTTPClient = (*errorClient)(nil)
+	_ extensionauth.GRPCClient = (*errorClient)(nil)
+)
+
+type errorClient struct{}
```
bogdandrutu, line 26:
```suggestion
type errorClient struct{
  component.StartFunc
  component.ShutdownFunc
}
```

bogdandrutu: For me the rule when a Func like "StartFunc" should implement the interface function "Start" is that, when we have an unsealed interface like Component, users can do what I suggested and not have to write the signature themselves. And when is sealed like in the Factory interface we don't need them since we control the only implementation.

bogdandrutu: So in this case we should provide the functions for the http/grpc clients and they should implement the signature, so that you can write this as:


```
type errorClient struct{
  component.StartFunc
  component.ShutdownFunc
  extensionauth.ClientRoundTripperFunc
}

func NewErrorClient(err error) extension.Extension {
  return &errorClient{ClientRoundTripperFunc: func() (credentials.PerRPCCredentials, error) { return err; }}
}
```

bogdandrutu (approved):
You will need to "undeprecate" the interface function implementations like https://github.com/open-telemetry/opentelemetry-collector/pull/12574/files#diff-23c4f67c72c9521b926d80da1f9233f3375e624222432f9983ae0aab05e8218aR64

extension/extensionauth/client.go
```diff
@@ -12,20 +12,34 @@ import (
...
 
+// HTTPClient is an optional Extension interface that can be used as an HTTP authenticator for the configauth.Authentication option.
+// Authenticators are then included as part of OpenTelemetry Collector builds and can be referenced by their
+// names from the Authentication configuration.
+type HTTPClient interface {
 	// RoundTripper returns a RoundTripper that can be used to authenticate HTTP requests.
 	RoundTripper(base http.RoundTripper) (http.RoundTripper, error)
+}
 
+// GRPCClient is an optional Extension interface that can be used as an HTTP authenticator for the configauth.Authentication option.
```
jade-guiton-dd, line 33:
```suggestion
// GRPCClient is an optional Extension interface that can be used as a gRPC authenticator for the configauth.Authentication option.
```

extension/extensionauth/client.go
```diff
@@ -93,13 +128,16 @@ func WithClientRoundTripper(roundTripperFunc ClientRoundTripperFunc) ClientOptio
...
+// Deprecated: [v0.122.0] This type is deprecated and will be removed in the next release.
 func WithClientPerRPCCredentials(perRPCCredentialsFunc ClientPerRPCCredentialsFunc) ClientOption {
 	return clientOptionFunc(func(o *defaultClient) {
 		o.clientPerRPCCredentialsFunc = perRPCCredentialsFunc
 	})
 }
 
 // NewClient returns a Client configured with the provided options.
+// Deprecated: [v0.122.0] This type is deprecated and will be removed in the next release.
+// Manually implement the [HTTPClient] and/or [GRPCClient] extensions instead.
```
jade-guiton-dd, line 140:
```suggestion
// Manually implement the [HTTPClient] and/or [GRPCClient] interfaces instead.
```

jade-guiton-dd (approved):
LGTM, just a few nitpicks on the doc comments

mx-psi:
I addressed all comments, however I do not want to merge this until the discussion over at #12601 is finalized

mx-psi:
Given https://github.com/open-telemetry/opentelemetry-collector/pull/12601#discussion_r1990310809 I am going to merge this

PR #12580, opened 2025-03-07
[componenttest] Move `CheckConfigStruct` to confmaptest

evan-bradley:
#### Description

This function is mostly concerned with config unmarshaling, so it relates much more to `confmap` than it does to `component`.

### DIFF
--- a/component/componenttest/configtest.go
+++ b/component/componenttest/configtest.go
@@ -20,6 +20,7 @@ var configFieldTagRegExp = regexp.MustCompile("^[a-z0-9][a-z0-9_]*$")
 // of components and extensions. It is recommended for implementers of components
 // to call this function on their tests passing the default configuration of the
 // component factory.
+// Deprecated: [v0.122.0] Use confmaptest.CheckConfigStruct instead.
 func CheckConfigStruct(config any) error {
 	t := reflect.TypeOf(config)
 	if t.Kind() == reflect.Ptr {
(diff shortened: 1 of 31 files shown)


evan-bradley:
Contrib tests are broken due to the changes in mdatagen; will need a `make update-otel` after this is merged. I'm skipping the contrib tests for now.

mx-psi (approved):
One small hiccup with this: we split all the test modules except for confmaptest, so this will be moved from a 0.x module to a 1.x module. That's okay for me but wanted to highlight it

component/componenttest/configtest.go
```diff
@@ -20,6 +20,7 @@ var configFieldTagRegExp = regexp.MustCompile("^[a-z0-9][a-z0-9_]*$")
 // of components and extensions. It is recommended for implementers of components
 // to call this function on their tests passing the default configuration of the
 // component factory.
+// Deprecated: [v0.122.0] Use confmaptest.CheckConfigStruct instead.
 func CheckConfigStruct(config any) error {
```
bogdandrutu, line 24:
```suggestion
// Deprecated: [v0.122.0] Use confmaptest.CheckConfigStruct instead.
var CheckConfigStruct = confmaptest.CheckConfigStruct
```

evan-bradley: That's a good idea, I'll do that.

andrzej-stencel (approved):

PR #12583, opened 2025-03-07
Propagate telemetry.resource to the internal logs

### LINKED ISSUE #12582: Resource attributes from telemetry.resource are missing for internal logs
When the `telemetry.resource` field is populated like this:
```yaml
service:
  telemetry:
    resource:
      service.name: "your-service-name"
      service.version: "your-version"
    logs:
      level: "debug"
```

Those resource attributes appear in the metrics and traces but not in the logs. This issue is to fix this and add those to the logs.

*Update 27Mar2025:*
this happens only on the logs printed via console

iblancasa:
#### Description
Propagate `telemetry.resource` to the internal logs.

#### Link to tracking issue
Fixes #12582

### DIFF
--- a/service/telemetry/logger.go
+++ b/service/telemetry/logger.go
@@ -39,6 +39,22 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
 		return nil, nil, err
 	}
 
+	// The attributes in cfg.Resource are added as resource attributes for logs exported through the LoggerProvider instantiated below.
+	// To make sure they are also exposed in logs written to stdout, we add them as fields to the Zap core created above using WrapCore.
+	// We do NOT add them to the logger using With, because that would apply to all logs, even ones exported through the core that wraps
+	// the LoggerProvider, meaning that the attributes would be exported twice.
+	logger = logger.WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+		fields := []zap.Field{}
+		for k, v := range cfg.Resource {
+			if v != nil {
+				f := zap.Stringp(k, v)
+				fields = append(fields, f)
+			}
+		}
+		r := zap.Dict("resource", fields...)
+		return c.With([]zapcore.Field{r})
+	}))
+
 	var lp log.LoggerProvider
 
 	logger = logger.WithOptions(zap.WrapCore(func(core zapcore.Core) zapcore.Core {
--- a/service/telemetry/logger_test.go
+++ b/service/telemetry/logger_test.go
@@ -6,14 +6,27 @@ package telemetry // import "go.opentelemetry.io/collector/service/telemetry"
 import (
 	"context"
 	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"reflect"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	config "go.opentelemetry.io/contrib/otelconf/v0.3.0"
+	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+
+	"go.opentelemetry.io/collector/pdata/plog/plogotlp"
+	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
 )
+
+type shutdownable interface {
+	Shutdown(context.Context) error
+}
 
 func TestNewLogger(t *testing.T) {
 	tests := []struct {
@@ -102,9 +115,6 @@ func TestNewLogger(t *testing.T) {
 				require.NoError(t, err)
 				gotType := reflect.TypeOf(l.Core()).String()
 				require.Equal(t, wantCoreType, gotType)
-				type shutdownable interface {
-					Shutdown(context.Context) error
-				}
 				if prov, ok := lp.(shutdownable); ok {
 					require.NoError(t, prov.Shutdown(context.Background()))
 				}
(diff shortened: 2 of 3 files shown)


bogdandrutu (changes requested):
Please add a test when used in combination with the otlp exporter for logs

iblancasa:
> Please add a test when used in combination with the otlp exporter for logs

Could you point to where or how should I add that test? I'm checking the source code and I'm a bit lost. Do we test somehow that the bridge is working?

iblancasa:
> Please add a test when used in combination with the otlp exporter for logs

Test added.

iblancasa:
Can you check again @bogdandrutu? Thanks!

iblancasa:
@open-telemetry/collector-approvers can somebody take a look to this PR? Thank you

service/telemetry/logger.go
```diff
@@ -37,23 +37,34 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
...
+					otelzap.NewCore("go.opentelemetry.io/collector/service/telemetry",
+						otelzap.WithLoggerProvider(lp),
+					),
+				), zap.NewAtomicLevelAt(cfg.Logs.Level))
+				if err != nil {
+					panic(err)
+				}
+				return core
+			}),
+			zap.Fields(fields...),
```
jade-guiton-dd, line 66: Adding `zap.Fields(...)` to the Logger here will add resource attributes as log attributes on all emitted logs, including those emitted using the OTel SDK, and which already have resource attributes set. This will create confusion and increate data volume.

Moreover, this code path is only taken when internal log exporting using processors is enabled, which is probably not what you want.

I think you should add those fields specifically to the logger core which handles stdout output; see next comment for suggestion.

service/telemetry/logger.go
```diff
@@ -37,23 +37,34 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
...
 	}
 
+	fields := []zap.Field{}
+	for k, v := range cfg.Resource {
+		if v != nil {
+			f := zap.Field{Key: k, Type: zapcore.StringType, String: *v}
+			fields = append(fields, f)
+		}
+	}
+
```
jade-guiton-dd, line 47:
```suggestion
	logger = logger.WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
		return c.With(fields)
	}))
```

See previous comment for rationale

service/telemetry/logger_test.go
```diff
@@ -79,13 +86,114 @@ func TestNewLogger(t *testing.T) {
...
+
+		logs := req.Logs()
+		rl := logs.ResourceLogs().At(0)
+		sl := rl.ScopeLogs().At(0)
+		logRecord := sl.LogRecords().At(0)
+		attrs := logRecord.Attributes().AsRaw()
+
+		assert.Contains(t, attrs, semconv.AttributeServiceName)
+		assert.Contains(t, attrs, semconv.AttributeServiceVersion)
+		assert.Contains(t, attrs, testAttribute)
```
jade-guiton-dd, line 124: I know @bogdandrutu asked to add this test with the OTLP exporter, but I don't think it's a good idea.

I believe the resource attributes should already be present in the OTLP-exported logs before your change, in `rl.Resource().Attributes()`. If anything, we should check that they are present there, and *not* duplicated in `rl.ScopeLogs().At(*).LogRecords().At(*).Attributes()`.

Moreover, this doesn't test that the attributes are present in the console output, which (to me at least) is the main contribution in this PR. (And unfortunately, the structure of the code in `service/telemetry/logger.go` would make that a bit difficult to test without refactoring.)

.chloggen/iblancasa_fix-resource-attributes-internal-logs.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+# One or more tracking issues or pull requests related to the change
+issues: [12582]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Resource attributes from telemetry.resource were not added to the internal
+  logs but to metrics and traces. This fix ensures they are added to the logs.
```
jade-guiton-dd, line 20: I think we should clarify that the resource attributes were only missing from the console output.

jade-guiton-dd (changes requested):
Thank you for the PR. I have some concerns related to [my comment on the tracking issue](https://github.com/open-telemetry/opentelemetry-collector/issues/12582#issuecomment-2758055221) however.

service/telemetry/logger_test.go
iblancasa: Yes, they are. Sorry, I was wrong about a manual test I ran locally and that make me implement something that is not needed.

jade-guiton-dd (commented):
This looks good to me now, but I'll leave it to @bogdandrutu to evaluate whether the test he requested is no longer necessary.

iblancasa:
> This looks good to me now, but I'll leave it to @bogdandrutu to evaluate whether the test he requested is no longer necessary.

Thanks a lot for your time to review and your help.

jade-guiton-dd (approved):
Looking at this again, since this PR no longer affects the export of logs through OTLP, I don't think the test requested by Bogdan makes sense anymore. Approving.

iblancasa:
@open-telemetry/collector-maintainers can somebody take a look there?

service/telemetry/logger_test.go
sfc-gh-bdrutu: Again, for me is to confirm that resource attributes are not duplicate, and this is not tested now

service/telemetry/logger.go
```diff
@@ -41,6 +41,18 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
...
 	}
 
+	// Add the resource fields to the logger
+	fields := []zap.Field{}
+	for k, v := range cfg.Resource {
+		if v != nil {
+			f := zap.Field{Key: k, Type: zapcore.StringType, String: *v}
+			fields = append(fields, f)
+		}
+	}
```
sfc-gh-bdrutu, line 51: Should we also add these fields in at "resource: map[string]string"?

service/telemetry/logger_test.go
jade-guiton-dd: Ah, I understand now. It seems we already have a test for resource attributes in internal metrics in `service/service_test.go`, so I guess we could add a similar one for internal logs.

service/telemetry/logger.go
jade-guiton-dd: Sorry, I don't understand what you mean. Where would this map be?

iblancasa: I don't get it neither.

iblancasa:
Hey, could you take a look to this PR? @bogdandrutu @sfc-gh-bdrutu

iblancasa:
ping @bogdandrutu @sfc-gh-bdrutu

iblancasa:
Ping @open-telemetry/collector-approvers

service/telemetry/logger.go
```diff
@@ -39,6 +39,18 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
 		return nil, nil, err
 	}
 
+	// Add the resource fields to the logger
+	fields := []zap.Field{}
+	for k, v := range cfg.Resource {
+		if v != nil {
+			f := zap.Field{Key: k, Type: zapcore.StringType, String: *v}
```
bogdandrutu, line 46: Please use zap.String()

bogdandrutu: Actually may consider `Stringp`

service/telemetry/logger.go
```diff
@@ -39,6 +39,18 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
...
+			fields = append(fields, f)
+		}
+	}
+	logger = logger.WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+		return c.With(fields)
+	}))
```
bogdandrutu, line 52: Should you call `logger.With` directly? If not, please document why.

jade-guiton-dd: This was [a suggestion of mine](https://github.com/open-telemetry/opentelemetry-collector/pull/12583#discussion_r2016767115), here is the rationale:

> Adding zap.Fields(...) to the Logger here will add resource attributes as log attributes on all emitted logs, including those emitted using the OTel SDK, and which already have resource attributes set. This will create confusion and increase data volume. I think you should add those fields specifically to the logger core which handles stdout output; see next comment for suggestion.

I agree that we should add a comment to explain why.

service/telemetry/logger.go
sfc-gh-bdrutu: The clarification this suggestion is to change line 51 `c.With(fields)` with `c.With(zap.Dict("resource", fields...))` if we think this is important.

jade-guiton-dd: The upside is that it distinguishes resource attributes from regular attributes by namespacing them, which could for example be useful in the `filelog` receiver if we want to turn them back into resource attributes.

On the other hand, this wouldn't be consistent with most other exporters, which tend to merge the different attribute sets together. This includes the new component attributes, which are turned into regular Zap fields in the console output.

I think either way is fine personally, especially since we have currently [no guidance from the spec](https://github.com/open-telemetry/opentelemetry-specification/issues/2535) about how to handle this.

Is this a blocker for merging the PR, or do we have leeway to change our minds later?

iblancasa: > The upside is that it distinguishes resource attributes from regular attributes by namespacing them, which could for example be useful in the `filelog` receiver if we want to turn them back into resource attributes.

This is exactly our use case. The current solution makes it work as desired.

iblancasa:
@bogdandrutu @sfc-gh-bdrutu ready for review

service/telemetry/logger.go
jade-guiton-dd: Are you sure? I would assume they would be turned into regular log record attributes rather than resource attributes.

iblancasa: Let me double check...

service/telemetry/logger.go
```diff
@@ -39,6 +39,20 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
 		return nil, nil, err
 	}
 
+	// The resouce attributes are added to the logger sending the logs to the OTLP endpoint
```
bogdandrutu, line 42: I don't think this comment explains the use of WrapCore vs directly With

jade-guiton-dd: I agree it's not very clear, I'll make a suggestion for clearer wording

service/telemetry/logger.go
```diff
@@ -39,6 +39,20 @@ func newLogger(set Settings, cfg Config) (*zap.Logger, log.LoggerProvider, error
...
+	// but not to the one handling the STDOUT from the collector. Here, we add them
+	// to the logger handling the STDOUT from the collector.
```
jade-guiton-dd, line 44: Suggestion on how to explain the use of WrapCore more clearly:
```suggestion
	// The attributes in cfg.Resource are added as resource attributes for logs exported through the LoggerProvider instantiated below.
	// To make sure they are also exposed in logs written to stdout, we add them as fields to the Zap core created above using WrapCore.
	// We do NOT add them to the logger using With, because that would apply to all logs, even ones exported through the OTel core, meaning that the attributes would be exported twice.
```

service/telemetry/logger.go
iblancasa: I added them under `resource`.

iblancasa:
I think everything should be ok now, @bogdandrutu @sfc-gh-bdrutu

jmacd (approved):

bogdandrutu (approved):

iblancasa:
@open-telemetry/collector-approvers could somebody merge this? Thanks!

codeboten:
Thanks for this fix @iblancasa, this makes me wonder if at some point we'd be able to switch to using a console exporter to ensure consistency in the future

PR #11576, opened 2024-10-30
Remove reprecated flags from Builder

bogdandrutu:

Issue #11453, opened 2024-10-15
`builder`: Please retain functionality to supply Go binary at command time


braydonk:
**Is your feature request related to a problem? Please describe.**
My intended setup for a collector binary built by OCB is the following:
* Write a `manifest.yaml`
* Download a local copy of a Go binary at a specific version
* Use that Go binary to install a local copy of `builder`
* Provide the locally installed Go binary as the Go binary for `builder`, today using the `--go` flag

This flag is being deprecated in favour of the `distribution::go` config value. However in my desired setup, I have to separately track within the config the eventual location of the Go binary that I'm going to install. It is logically much cleaner to simply provide the Go binary I want at command-time rather than being forced to track this in configuration.

**Describe the solution you'd like**
Either keep the `--go` flag, or provide something other mechanism to provide a Go binary at command-time like an env var.

**Describe alternatives you've considered**
As stated above, I could just track the location that I know the local Go binary will be installed when it is, but this is less clean than simply providing it at command-time.

mx-psi:
Relates to #7976

braydonk:
## Additional Context

I have a repo where I am managing multiple separate builds of the collector. The way each build works is by downloading a local copy of the Go version required for that particular build. What I used to be able to do was in the command to initiate the builder call it with `$(PWD)/path/to/downloaded/go`. With that functionality removed, I can put the directory I know it will be installed to in the `dist::go` field, but it can't be a relative path; if the path is relative, it is unable to find it. So I can't actually find a way to make this work with the removal of the `--go` flag.

seankhliao:
can't you just set PATH before calling the builder?

PR #12589, opened 2025-03-09
[builder] Add --go-build-flags cli flag

### LINKED ISSUE #12588: [builder] Allow passing additional Go build args to `ocb`
### Component(s)

_No response_

NOTE: I wanted to make the builder the component for this issue, but it's missing from the list.

# Description

There are some additional flags I would like to be able to pass to the `go build` subcommand when compiling distributions. The main ones I have in mind are `-buildvcs=false`, `-p`, and build tags in case I want to change what is set in the distribution config.

# Proposed Solution

Add a new command arg that accepts additional `go build` arguments which will be appended to what already gets added to go build command. The way the build command arguments work is that if a duplicate is passed, the latter will take precedence (i.e. `go build -o one -o two .` will build a binary called "two"), so a user can pass in any go build args they want to override from the defaults that `ocb` adds and there shouldn't be a need to worry about `ocb` checking and keeping track of that.

# Alternative

Today the user can already `--skip-compilation` and use their own `go build`, which is what I'm doing right now. However it is nice to have the defaults that `ocb` already sets.

### LINKED ISSUE #12591: [builder] Produces binary name with no `.exe` extension on Windows
### Component(s)

_No response_

### What happened?

**Describe the bug**
If you use the builder to build a Windows binary, that binary will not have a `.exe` extension. That's because the `-o` flag is passed to `go build` as the distribution name only, with no platform consideration.

**Steps to reproduce**
Run the builder on Windows, or with `GOOS=windows`.

`<distribution name>.exe`

**What did you see instead?**

`<distribution name>`

### Collector version

v0.121.0

## Environment
OS: Windows
Compiler(if manually compiled): go 1.24.0

### OpenTelemetry Collector configuration

```yaml

```

### Log output

```shell
2025-03-09T22:12:26.248-0400    INFO    internal/command.go:99  OpenTelemetry Collector Builder  {"version": "(devel)"}
2025-03-09T22:12:26.249-0400    INFO    internal/command.go:104 Using config file{"path": "internal/config/default.yaml"}
2025-03-09T22:12:26.251-0400    INFO    builder/config.go:165   Using go        {"go-executable": "C:\\Users\\Braydon\\sdk\\go1.24.0\\bin\\go.exe"}
2025-03-09T22:12:26.264-0400    INFO    builder/main.go:99      Sources created {"path": "C:\\Users\\Braydon\\AppData\\Local\\Temp\\otelcol-distribution148321280"}
2025-03-09T22:13:05.790-0400    INFO    builder/main.go:201     Getting go modules
2025-03-09T22:13:08.976-0400    INFO    builder/main.go:110     Compiling
2025-03-09T22:13:32.737-0400    INFO    builder/main.go:140     Compiled        {"binary": "C:\\Users\\Braydon\\AppData\\Local\\Temp\\otelcol-distribution148321280/otelcorecol"}
```

### Additional context

_No response_

braydonk:
#### Description
This PR adds the `--go-build-flags` cli flag which will allow the user to pass additional flags for the `go build` command used to compile the collector.

While doing the refactor, I also made it so that the default binary name now adds the `.exe` extension when building for Windows.

#### Link to tracking issue
Fixes #12588 
Fixes #12591 

#### Testing
New unit tests were added to verify the build command args based on the config. A new CompileAndGenerate test case was added as well.

#### Documentation
A section has been added to the README.

### DIFF
--- a/cmd/builder/README.md
+++ b/cmd/builder/README.md
@@ -225,3 +225,22 @@ file for the following things after `go get`ing all components and calling
 The `--skip-strict-versioning` flag disables these versioning checks.
 This flag is available temporarily and
 **will be removed in a future minor version**.
+
+## Customize Go Build
+
+By default, the builder will set a number of default arguments for the `go build` command it runs to build the collector. This includes: 
+
+* `-o` matching `dist::name`
+* `-trimpath` 
+* Default or provided `-ldflags`
+* Provided `-gcflags`
+* `-tags` configured from `dist::build_tags`
+
+You can provide additional flags to `go build` with the builder flag `--go-build-flags`:
+```console
+ocb --config=config.yaml --go-build-flags="-p 32 -buildvcs=false"
+```
+If you provide any flags in `--go-build-flags` that are already configured by the builder automatically, **they will be overridden**:
+```console
+ocb --config=config.yaml --go-build-flags="-o alt-bin-name" # The binary will be called "alt-bin-name" instead of dist::name
+```
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -36,9 +36,9 @@ type Config struct {
 	SkipGetModules       bool   `mapstructure:"-"`
 	SkipStrictVersioning bool   `mapstructure:"-"`
 	LDFlags              string `mapstructure:"-"`
-	LDSet                bool   `mapstructure:"-"`
+	LDSet                bool   `mapstructure:"-"` // true if ldflags is set from cli flag
 	GCFlags              string `mapstructure:"-"`
-	GCSet                bool   `mapstructure:"-"`
+	GCSet                bool   `mapstructure:"-"` // true if gcflags is set from cli flag
 	GoBuildFlags         string `mapstructure:"-"`
 	Verbose              bool   `mapstructure:"-"`
 
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -242,7 +242,8 @@ func (c *Config) getGoBuildArgs() []string {
 	// Typically the -ldflags values are quoted i.e. -ldflags="<flags>".
 	// However since these args are used as exec.Cmd args, adding quotes
 	// will break the exec.Cmd argument parser. So they are instead provided
-	// as -ldflags=<flags>.
+	// as -ldflags=<flags>, and when it is applied to the go build command
+	// it will effectively be `go build "-ldflags=<flags>"`.
 	args = append(args, "-ldflags="+ldflags)
 	args = append(args, "-gcflags="+gcflags)
 
--- a/cmd/builder/internal/builder/config.go
+++ b/cmd/builder/internal/builder/config.go
@@ -36,7 +36,9 @@ type Config struct {
 	SkipGetModules       bool   `mapstructure:"-"`
 	SkipStrictVersioning bool   `mapstructure:"-"`
 	LDFlags              string `mapstructure:"-"`
+	LDSet                bool   `mapstructure:"-"`
 	GCFlags              string `mapstructure:"-"`
+	GCSet                bool   `mapstructure:"-"`
 	GoBuildFlags         string `mapstructure:"-"`
 	Verbose              bool   `mapstructure:"-"`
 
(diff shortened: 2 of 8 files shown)


cmd/builder/internal/builder/main_test.go
```diff
@@ -113,6 +113,7 @@ var replaceModules = []string{
 func newTestConfig(tb testing.TB) *Config {
 	cfg, err := NewDefaultConfig()
 	require.NoError(tb, err)
+	cfg.Distribution.Name = "test_distribution"
```
braydonk, line 116: Currently, the integration tests are running with a blank name field, which means the args end up being `build -o  'ldflags=-s -w' ...`. So it's building a binary with the name 'ldflags=-s -w' instead of applying the ldflags. It didn't break any tests but is probably not intentional.

cmd/builder/internal/builder/config.go
```diff
@@ -212,6 +214,51 @@ func (c *Config) allComponents() []Module {
 	return slices.Concat[[]Module](c.Exporters, c.Receivers, c.Processors, c.Extensions, c.Connectors, c.ConfmapProviders, c.ConfmapConverters)
 }
 
+func (c *Config) getGoBuildArgs() []string {
+	ldflags := "-s -w" // we strip the symbols by default for smaller binaries
+	gcflags := ""
+	executableName := c.Distribution.Name
+	if runtime.GOOS == "windows" {
+		executableName += ".exe"
```
braydonk, line 222: Codecov seems to think these aren't covered, but I think they are covered by `TestGetGoBuildArgs`.

mattsains: the reason you're getting this warning is because the body of the if statement isn't being run in codecov because it's not running in windows. I think this finding should probably be ignored because the test TestGetGoBuildArgs explicitly tests correctly on windows systems

mattsains (approved):

mattsains:
@open-telemetry/collector-approvers

cmd/builder/README.md
```diff
@@ -225,3 +225,22 @@ file for the following things after `go get`ing all components and calling
...
+* `-trimpath` 
+* Default or provided `-ldflags`
+* Provided `-gcflags`
+* `-tags` configured from `dist::build_tags`
+
+You can provide additional flags to `go build` with the builder flag `--go-build-flags`:
+```console
+ocb --config=config.yaml --go-build-flags="-p 32 -buildvcs=false"
+```
+If you provide any flags in `--go-build-flags` that are already configured by the builder automatically, **they will be overridden**:
```
mx-psi, line 243: Is this flexible enough to remove any option passed by default? For example, what if I don't want to pass `-trimpath`? Could I do this by overriding it with `--go-build-flags`?

cmd/builder/README.md
```diff
@@ -225,3 +225,22 @@ file for the following things after `go get`ing all components and calling
...
+
+* `-o` matching `dist::name`
+* `-trimpath` 
+* Default or provided `-ldflags`
+* Provided `-gcflags`
+* `-tags` configured from `dist::build_tags`
+
+You can provide additional flags to `go build` with the builder flag `--go-build-flags`:
+```console
+ocb --config=config.yaml --go-build-flags="-p 32 -buildvcs=false"
```
mx-psi, line 241: Any advantage of this being a CLI flag vs being in `config.yaml`?

jackgopack4 (commented):
I am not convinced that we should be adding more CLI arguments to ocb, especially when we have recently removed several recently (https://github.com/open-telemetry/opentelemetry-collector/pull/11576)
I think there is an argument to be made that this [and the ld/gc flags](https://github.com/open-telemetry/opentelemetry-collector/pull/11996) should be moved to a config.yaml section

braydonk:
I'm just going to close this I think. The reason I wanted to add this is similar to the reason I opened #11453, which is that there are ways I want to be able to build the Collector from `ocb` that can't be known at config time. However, the reason that issue was dismissed probably applies here too. I am generally opposed to this unilateral move to config over CLI flags, but not willing to fight for it. So at this point I just `--skip-compilation` and use Go directly.

#12591 which this PR addressed still should be fixed.

DesAWSume:
Looks like this has been merged, could we have an update README or a instruction to build a exe/msi file will work on Windows?

braydonk:
@DesAWSume This wasn't merged. The instructions that exist on the README work the same on Windows, the only difference is you will have to rename the resulting binary to have a `.exe` extension.

PR #12590, opened 2025-03-09
[componenttest] Improve config struct checks

evan-bradley:
#### Description

Improves our config struct checks by allowing for a few more cases that are possible with mapstructure.

### DIFF
--- a/component/componenttest/configtest.go
+++ b/component/componenttest/configtest.go
@@ -65,8 +65,8 @@ func validateConfigDataType(t reflect.Type) error {
 
 // checkStructFieldTags inspects the tags of a struct field.
 func checkStructFieldTags(f reflect.StructField) error {
-	tagValue := f.Tag.Get("mapstructure")
-	if tagValue == "" {
+	tagValue, ok := f.Tag.Lookup("mapstructure")
+	if !ok {
 		// Ignore special types.
 		switch f.Type.Kind() {
 		case reflect.Interface, reflect.Chan, reflect.Func, reflect.Uintptr, reflect.UnsafePointer:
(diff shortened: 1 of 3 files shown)


component/componenttest/configtest.go
```diff
@@ -93,20 +97,17 @@ func checkStructFieldTags(f reflect.StructField) error {
...
+		switch tag {
+		case "squash":
+			if (f.Type.Kind() != reflect.Struct) && (f.Type.Kind() != reflect.Ptr || f.Type.Elem().Kind() != reflect.Struct) {
+				return fmt.Errorf(
+					"attempt to squash non-struct type on field %q", f.Name)
+			}
+		case "remain":
+			if f.Type.Kind() != reflect.Map && f.Type.Kind() != reflect.Interface {
+				return fmt.Errorf(`attempt to use "remain" on non-map or interface type field %q`, f.Name)
+			}
```
mx-psi, line 110: Do we have any uses of remain currently? Should we just forbid its usage entirely?

evan-bradley: We don't use it ourselves, but we do use it through the OTel SDK config package: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/v0.3.0/generated_config.go#L422.

In general I think it would make sense to advise against `remain` for Collector config structs, but I'm not sure there's a good way to do it. The best I can think of would be to update this function to take an option that enables support for it.

mx-psi (approved):

bogdandrutu (approved):

atoulme (approved):

Issue #12468, opened 2025-02-25
[confighttp] otelhttp-generated span names may have high cardinality


axw:
### Component(s)

_No response_

### What happened?

**Describe the bug**

The `otelhttp` instrumentation in `confighttp` uses the URL path for span names: https://github.com/open-telemetry/opentelemetry-collector/blob/2b5fa0ed51fabd128d30ad07d50535df95444e8f/config/confighttp/confighttp.go#L468-L470

According to https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name:

> HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available. If there is no (low-cardinality) {target} available, HTTP span names SHOULD be {method}.

I realise it's a SHOULD and not a MUST, but I would expect OpenTelemetry Collector to be setting an example and following documented recommendations. Producing high cardinality span names can be problematic for certain pipelines, e.g. when aggregating spans by name.

**Steps to reproduce**

1. Add trace telemetry to `examples/local/otel-config.yaml`:

```diff
$ git diff
diff --git a/examples/local/otel-config.yaml b/examples/local/otel-config.yaml
index 168a1631a..3e6266e70 100644
--- a/examples/local/otel-config.yaml
+++ b/examples/local/otel-config.yaml
@@ -39,3 +39,10 @@ service:
       exporters: [debug]
 
   extensions: [zpages]
+  telemetry:
+    traces:
+      level: basic
+      processors:
+        - batch:
+            exporter:
+              console: {}
```

2. `make run`
3. `curl http://localhost:4318/foo/bar`

**What did you expect to see?**

I would expect to see a span logged with the name `GET`, or something low-cardinality but more descriptive like `GET unknown route`.

**What did you see instead?**

A span is logged with the name `/foo/bar`:

```
{                                                                                                                                                                                                                                                       
        "Name": "/foo/bar",                                                                                                                                                                                                                             
        "SpanContext": {                                                                                                                                                                                                                                
                "TraceID": "9a023a975a32440d01b47c6f8f7a60c7",                                                                                                                                                                                          
                "SpanID": "aa890cf63d51904a",                                                                                                                                                                                                           
                "TraceFlags": "01",                                                                                                                                                                                                                     
                "TraceState": "",                                                                                                                                                                                                                       
                "Remote": false                                                                                                                                                                                                                         
        },
...
```

### Collector version

2b5fa0ed51fabd128d30ad07d50535df95444e8f

[...]

PR #12593, opened 2025-03-10
confighttp: route-based span naming for ServeMux

axw:
#### Description

Update confighttp.ServerConfig.ToServer to handle ServeMux specially: if the handler supplied to ToServer is an http.ServeMux, we use its [Handler method](https://pkg.go.dev/net/http#ServeMux.Handler) to determine the matching pattern and use that to name the span as described at https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name. If there is no matching pattern then we fall back to "unknown route".

This approach means that everything will magically just work for components that supply a ServeMux, but there are a couple of downsides:
 - this is a little bit magical, and won't help for components that _don't_ happen to use a ServeMux
 - route pattern matching needs to be done twice: once for the instrumentation, and once for routing to the correct handler

The alternative that I started working on before I came to this solution was to introduce a new method, ServerConfig.ToServeMux, which would not accept an `http.Handler`, and would instead return a type that wraps `*http.ServeMux`, and would instrument handlers as they are registered. With that approach the instrumentation would come after the routing, and we would install a default `/` NotFoundHandler that is also instrumented. This approach is a bit less magical, and a bit more efficient, but would require all the components to be updated.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/12468 (for any receiver that uses a ServeMux, like otlpreceiver)

#### Testing

- Added a unit test to confighttp for the instrumentation added by ServerConfig.ToServer.
- Tested with a otlpreceiver:
   - ran `make run` with trace telemetry enabled, exporting to console
   - used `telemetrygen logs --otlp-http`, confirmed that a span named "POST /v1/logs" is produced
   - used `curl http://localhost:4318`, confirmed that a span named "GET unknown route" is produced

#### Documentation

Added doc comment to confighttp.ServerConfig.ToServer explaining the new behaviour.

### DIFF
--- a/.chloggen/confighttp-server-spanname.yaml
+++ b/.chloggen/confighttp-server-spanname.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: confighttp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use low-cardinality route pattern for confighttp server span names
+
+# One or more tracking issues or pull requests related to the change
+issues: [12468]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  For components that route HTTP requests using a net/http.ServeMux, such as otlpreceiver,
+  server spans will now be given low-cardinality span names based on the route pattern.
+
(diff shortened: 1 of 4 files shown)


.chloggen/confighttp-server-spanname.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext: |
+  For components that route HTTP requests using a net/http.ServeMux, such as otlpreceiver,
+  server spans will now be given low-cardinality span names based on the route pattern.
+
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api, user]
```
axw, line 27:
```suggestion
change_logs: [user]
```

No API change (there was, I reverted it)

PR #12568, opened 2025-03-05
Introduce `Equal` comparison in pcommon types

dmathieu:
#### Description

This introduces `.Equal()` methods to most pcommon types, so equality comparison can be performed with no allocations.

The types `Equal` is added to are:

* `Value`
* `ByteSlice`
* `Float64Slice`
* `Int32Slice`
* `Int64Slice`
* `StringSlice`
* `Uint64Slice`
* `Map`
* `Slice`

The original intent was to add it to `Value`. However, to be able to handle every type in there, I had to also add it to the other types, some of which are coming too because they are auto-generated.

#### Link to tracking issue
Fixes #12561

#### Testing

Here are the results from the new benchmarks.

```
go test -bench=. ./...
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/collector/pdata/pcommon
cpu: Apple M1 Max
BenchmarkByteSliceEqual-10              518354892                2.216 ns/op           0 B/op          0 allocs/op
BenchmarkFloat64SliceEqual-10           536303960                2.213 ns/op           0 B/op          0 allocs/op
BenchmarkInt32SliceEqual-10             544727418                2.208 ns/op           0 B/op          0 allocs/op
BenchmarkInt64SliceEqual-10             544786770                2.218 ns/op           0 B/op          0 allocs/op
BenchmarkStringSliceEqual-10            161984316                7.491 ns/op           0 B/op          0 allocs/op
BenchmarkUInt64SliceEqual-10            543405762                2.211 ns/op           0 B/op          0 allocs/op
BenchmarkMapEqual-10                    95199063                12.37 ns/op            0 B/op          0 allocs/op
BenchmarkSliceEqual-10                  180702288                6.640 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/nil-10              583397124                2.065 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/strings-10          252571316                4.735 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/booleans-10         421377843                2.865 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/ints-10             420362413                2.861 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/ints8-10            420972375                2.867 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/ints16-10           420432676                2.865 ns/op           0 B/op          0 allocs/op
[...]

PR #12594, opened 2025-03-10
Introduce `Equal` comparison in pcommon types

### LINKED ISSUE #12561: Add `Equal()` helper method to `pcommon.Value`
nit: I know that `AsRaw()` could make a copy of the data. This is a reminder that it would be nice if you could write `a.Value().Equal(value)` and avoid the allocation.

_Originally posted by @jmacd in https://github.com/open-telemetry/opentelemetry-collector/pull/12390#discussion_r1979808316_

So we don't need to use `AsRaw` to do value comparison.

dmathieu:
#### Description

This introduces `.Equal()` methods to most pcommon types, so equality comparison can be performed with no allocations.

The types `Equal` is added to are:

* `Value`
* `ByteSlice`
* `Float64Slice`
* `Int32Slice`
* `Int64Slice`
* `StringSlice`
* `Uint64Slice`
* `Map`
* `Slice`

The original intent was to add it to `Value`. However, to be able to handle every type in there, I had to also add it to the other types, some of which are coming too because they are auto-generated.

#### Link to tracking issue
Fixes #12561

This is the same thing as #12568, but with comparison of pcommon types rather than raw ones.

#### Testing

Here are the results from the new benchmarks.

```
goos: darwin
goarch: arm64
pkg: go.opentelemetry.io/collector/pdata/pcommon
cpu: Apple M1 Max
BenchmarkByteSliceEqual-10              513754251                2.181 ns/op           0 B/op          0 allocs/op
BenchmarkFloat64SliceEqual-10           550760200                2.189 ns/op           0 B/op          0 allocs/op
BenchmarkInt32SliceEqual-10             550836993                2.186 ns/op           0 B/op          0 allocs/op
BenchmarkInt64SliceEqual-10             536615931                2.178 ns/op           0 B/op          0 allocs/op
BenchmarkStringSliceEqual-10            166586130                7.055 ns/op           0 B/op          0 allocs/op
BenchmarkUInt64SliceEqual-10            550221884                2.191 ns/op           0 B/op          0 allocs/op
BenchmarkMapEqual-10                    100000000               10.29 ns/op            0 B/op          0 allocs/op
BenchmarkSliceEqual-10                  165937951                7.185 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/nil-10              425391500                2.817 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/strings-10          226853427                5.441 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/booleans-10         296628630                4.080 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/ints-10             296841543                4.070 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/doubles-10          296242837                4.050 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/byte_slices-10              173515455                6.855 ns/op           0 B/op          0 allocs/op
BenchmarkValueEqual/slices-10                   83512392                14.10 ns/op            0 B/op          0 allocs/op
BenchmarkValueEqual/maps-10                     63556654                18.41 ns/op            0 B/op          0 allocs/op
PASS
ok      go.opentelemetry.io/collector/pdata/pcommon     24.650s
```


pdata/pcommon/value.go
```diff
@@ -452,6 +452,33 @@ func (v Value) AsRaw() any {
...
 }
 
+func (v Value) Equal(c Value) bool {
+	if v.Type() != c.Type() {
+		return false
+	}
+
+	switch v.Type() {
+	case ValueTypeEmpty:
+		return v.Type() == c.Type()
```
bogdandrutu, line 462: Already checked, just return true.

bogdandrutu (approved):

pdata/pcommon/value.go
dmathieu: Hmm no, the type check at the beginning of the method checks that the types are not the same (and therefore returns false).
Here, we check that if a is empty and b is empty, they are equal.

pdata/pcommon/map_test.go
```diff
@@ -564,3 +564,29 @@ func TestInvalidMap(t *testing.T) {
...
+func TestMapEqual(t *testing.T) {
+	m := NewMap()
+	m2 := NewMap()
+	assert.True(t, m.Equal(m2))
+
+	m.PutStr("hello", "world")
+	assert.False(t, m.Equal(m2))
+
+	m2.PutStr("hello", "world")
+	assert.True(t, m.Equal(m2))
```
bogdandrutu, line 577: Add a test for when both have "hello" but different values. Also a test when both have same len but different keys.

pdata/pcommon/value.go
```diff
@@ -452,6 +452,33 @@ func (v Value) AsRaw() any {
 	return fmt.Sprintf("<Unknown OpenTelemetry value type %q>", v.Type())
 }
 
+func (v Value) Equal(c Value) bool {
+	if v.Type() != c.Type() {
```
bogdandrutu, line 456: Please add test.

bogdandrutu (approved):

PR #10694, opened 2024-07-22
Generate config Go code from schema

ptodev:
#### Description

This is an attempt to generate the `config.go` files from  Json schema. The goal of the PR is to delete all the `config.go` files for each component and replace them with some sort of Yaml or Json schema, which will most likely be sourced from the `metadata.yaml` file for each component. This PR uses an [earlier PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27003) as a starting point. 

There are many unknowns related to this project. I will start threads on various problems via file comments and line comments, so that we can keep the discussion organised.

#### Link to tracking issue
Fixes #9769

PR #12596, opened 2025-03-10
[chore] Create RFC for Optional config types

evan-bradley:
#### Description

This RFC will help us explore the use cases Optional types solve and move us toward a resolution for whether to adopt these in our config.

Related to #10266


evan-bradley:
@yurishkuro could you take a look at this? We're taking a closer look at whether we want this type for 1.0 and want to make sure if we add it that it will cover all config use cases and won't have any major downsides. I would appreciate your input here.

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+facilities and while not common, could be a rough edge for users looking to
+implement similar config structures.
+
+## Disadvantages of an Optional type
+
+There are two noteworthy disadvantages of introducing an Optional type:
+
+1. Since the type isn't standard, external packages working with config may
+   require additional adaptations to work with our config structs.
+2. It will still be possible to use pointer types for optional config fields,
```
yurishkuro, line 182: I don't think this is a disadvantage of the Optional type specifically. There can be a feature gate added that will run a validator that prohibits (or warns about) pointer fields, to ensure consistency.

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+This situation is something of an edge case with our current unmarshaling
+facilities and while not common, could be a rough edge for users looking to
+implement similar config structures.
+
+## Disadvantages of an Optional type
+
+There are two noteworthy disadvantages of introducing an Optional type:
+
+1. Since the type isn't standard, external packages working with config may
+   require additional adaptations to work with our config structs.
```
yurishkuro, line 181: Can we give examples of external packages working with config? E.g. validation libs come to mind, but they might still work, for them Optional is just an extra nesting container.

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+
+func Some[T any](value T) Optional[T] {
+	return Optional[T]{value: value, hasValue: true}
+}
+
+func None[T any](value T) Optional[T] {
+	return Optional[T]{}
+}
+
+func WithDefault[T any](defaultVal T) Optional[T] {
```
yurishkuro, line 63: I think `WithDefault` should accept a lambda, `func() T`, not an actual value. The defaults may be runtime sensitive, an example would be the `localhost` vs. `0.0.0.0` that could be controlled by an external mechanism like a feature gate.

yurishkuro (commented):
+1

docs/rfcs/optional-config-type.md
mx-psi: Agreed, I think we have the same issue with, say, `configopaque`, and it's so far fine (Evan can keep me honest here wrt e.g. serializing)

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+
+1. Using a separate boolean field to indicate whether the field is enabled or
+   disabled.
+2. Making the type a pointer, which makes a `nil` pointer represent that a
+   value is not present, and a valid pointer represent the desired value.
+
+Each of these approaches has deficiencies: Using a separate boolean field
+requires the user to set the boolean field to `true` in addition to setting the
+actual config option, leading to suboptimal UX. Using a pointer value has a few
+drawbacks:
```
mx-psi, line 24: Mutability is another concern that you mentioned over at https://github.com/open-telemetry/opentelemetry-collector/issues/9478#issuecomment-2414487668 Should we add it here?

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+		cfg.HTTP = nil
+	}
+
+	return nil
+}
+```
+
+With an Optional type, the checks in `Unmarshal` become unnecessary, and it's
+possible the entire `Unmarshal` function may no longer be needed. Instead, when
+the the config is unmarshaled, no value would be put into the default Optional
```
mx-psi, line 169:
```suggestion
the config is unmarshaled, no value would be put into the default Optional
```

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+func (o Optional[T]) Default() T {
+    return o.defaultVal
+}
+```
+
+## Use cases
+
+Optional types can fulfill the following use cases we have when decoding config.
+
+### Representing optional config fields
```
mx-psi, line 84: Could we list other use cases? e.g. TLSConfig

docs/rfcs/optional-config-type.md
yurishkuro: I had another motivation for function in my prototype https://github.com/open-telemetry/opentelemetry-collector/pull/10260/files#diff-4825c80a1cbc3775a3f3cd1aa7e6d01b6226bc31efaa9ca7b6a1a466b31512f2R36
```
// The reason we pass a function instead of T directly
// is to allow this to work with T being a pointer,
// because we wouldn't want to copy the value of a pointer
// since it might reuse (and override) some shared state.
```

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,184 @@
...
+     // HTTP will be enabled if this line is not run
+	if !conf.IsSet(protoHTTP) {
+		cfg.HTTP = nil
+	}
+
+	return nil
+}
+```
+
+With an Optional type, the checks in `Unmarshal` become unnecessary, and it's
```
yurishkuro, line 167: having `Unmarshal` by itself is not a problem, the problem is that it cannot be made type-safe, since `conf.IsSet(protoGRPC)` checks against a string with field name, disconnected from the actual field name.

evan-bradley: Fully agreed, that's why I call out the checks in Unmarshal here as opposed to the function itself. Granted, in this case specifically, this implementation of `Unmarshal` could be removed since it only handles unsetting default values.

docs/rfcs/optional-config-type.md
evan-bradley: Sorry for the confusion here, the reason I didn't have `WithDefault` accept a function was to simplify the implementation to focus on high-level concepts instead of the design of the API. This is not at all supposed to be a carefully-designed or production-ready implementation, which I try to call out above the code block.

If there's a consideration in the design of `WithDefault` that we should be considering for whether the type is worth adding at all, or if using a value instead of a function makes things less clear, I can adjust the example implementation in this document.

yurishkuro: fair enough, but the "simplified" implementation is different from the one with lambda in about 10 characters.

docs/rfcs/optional-config-type.md
evan-bradley: That's a good callout, I'll remove this point.

docs/rfcs/optional-config-type.md
yurishkuro: ok, but how do you want to move the discussion forward? Just take a vote? Or do more cost/benefit analysis, and potentially block this on unresolved #10694?

yurishkuro (approved):

docs/rfcs/optional-config-type.md
evan-bradley: We'll essentially need a vote to get this merged anyway per the Collector's [RFC process](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/README.md#merge-rules), so I think that's what it will come down to.

I think this is far more straightforward than https://github.com/open-telemetry/opentelemetry-collector/pull/10694 anyway, so I would prefer to not block it on that. I'm just calling it out here as a risk.

docs/rfcs/optional-config-type.md
evan-bradley: I missed that when reviewing the issues related to this PR, but that's a meaningful concern. I've added it.

docs/rfcs/optional-config-type.md
evan-bradley: Sure thing, I added another example for how this would impact `confighttp.ServerConfig`. It doesn't look substantially different from how it's used in the OTLP receiver, but it should help illustrate how this might look.

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,276 @@
...
+the config is unmarshaled, no value would be put into the default Optional
+values and `HasValue` would return false when using this config object.
+
+This situation is something of an edge case with our current unmarshaling
+facilities and while not common, could be a rough edge for users looking to
+implement similar config structures.
+
+## Disadvantages of an Optional type
+
+There are two noteworthy disadvantages of introducing an Optional type:
```
mx-psi, line 268:
```suggestion
There is one noteworthy disadvantage of introducing an Optional type:
```

mx-psi (approved):

bogdandrutu (approved):

bogdandrutu (commented):
I think we can then remove https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/confmap.go#L351

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,276 @@
...
+
+Optional types can fulfill the following use cases we have when decoding config.
+
+### Representing optional config fields
+
+To use the optional type to mark a config field as optional, the type can simply
+be used as a type parameter to `Optional[T]`. The following config struct shows
+how this may look, both in definition and in usage:
+
+```golang
```
bogdandrutu, line 93: More than golang code, would be good to show how users in yaml will use it.

yurishkuro: there's no expected impact on YAML

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,276 @@
...
+   a field is optional.
+2. The distinction between values that are conventionally pointers (e.g. gRPC
+   configs) and optional values is lost.
+3. Setting a default value for a pointer field when decoding will set the field
+   on the resulting config struct, and additional logic must be done to unset
+   the default if the user has not specified a value.
+4. The potential for null pointer exceptions is created.
+5. Config structs are generally intended to be immutable and may be passed
+   around a lot, which makes the mutability property of pointer fields
+   an undesirable property.
```
jmacd, line 36: Drawback numbers 2 and 5 appear contradictory. Is there a reason why pointers are "conventionally" pointers, aside from mutability?
```suggestion
2. While they can be copied cheaply, pointers convey a mutable reference making
   them unsafe to share.
3. Setting a default value for a pointer field when decoding will set the field
   on the resulting config struct, and additional logic must be done to unset
   the default if the user has not specified a value.
4. The potential for null pointer exceptions is created.
```

jmacd (approved):

docs/rfcs/optional-config-type.md
evan-bradley: This may not really apply to config structs, but outside mutability, another reason to use a pointer type for a field in Go is so you don't have to pass around a large struct by value.

The specific point I was making with item 2 is that some types in popular libraries are pointers, e.g. `*zap.Logger`, `*grpc.Server`, etc. If pointers are used to indicate optional values, if you have a field that is a `*zap.Logger` type, it's difficult to tell if it's intentionally optional, or if that type is almost always used as a pointer type (`zap.Logger` as a non-pointer is rare from what I've seen). Obviously `*zap.Logger` isn't a perfect example because nobody is setting a logger in YAML, but hopefully my point still makes sense.

mx-psi:
cc @open-telemetry/collector-approvers This has entered final comment period, I intend to merge this on Friday if there are no further blocking comments

docs/rfcs/optional-config-type.md
```diff
@@ -0,0 +1,276 @@
...
+
+This situation is something of an edge case with our current unmarshaling
+facilities and while not common, could be a rough edge for users looking to
+implement similar config structures.
+
+## Disadvantages of an Optional type
+
+There is one noteworthy disadvantage of introducing an Optional type:
+
+1. Since the type isn't standard, external packages working with config may
```
kristinapathak, line 270: What is the intended implementation plan for Optional config types? In the above example, it says:
>This situation is something of an edge case

So is the intention to only modify fields who meet some required use case? Or is it to modify all fields that are currently pointers?

Changing the field type is a breaking API change. Is there a different approach depending on the stability level of a component? Or is the expectation that anyone importing and using a config struct from an otel component will have to make any required changes given the new struct field types?

mx-psi: @kristinapathak Do you have examples of components/modules that are 1.x and need to use this optional type? We are blocking `confighttp`/`configgrpc` on this so that we can apply this pattern on these modules.

The intent is to apply this new type everywhere, go through an (API) breaking change, and ensure this type is used everywhere going forward.

docs/rfcs/optional-config-type.md
mx-psi: I think there is no specific concern on this thread, and the JSON schema libraries support custom types so I am resolving this comment

docs/rfcs/optional-config-type.md
mx-psi: @bogdandrutu does Yuri's comment solve your question?

docs/rfcs/optional-config-type.md
mx-psi: @jmacd does Evan's comment solve your question?

docs/rfcs/optional-config-type.md
mx-psi: @kristinapathak does this solve your question?

mx-psi:
I believe none of the pending conversations are blocking, but I pinged the authors just in case.

docs/rfcs/optional-config-type.md
kristinapathak: yes, thank you

jmacd (approved):

docs/rfcs/optional-config-type.md
mx-psi: Clarified inline on 5d5f6527f1d4444c657d30c19f3055e781830913

PR #12600, opened 2025-03-11
Limiter extension configgrpc integration draft

jmacd:
#### Description

Demonstrates how a `Limiters []limter.Limiter` will be threaded through the gRPC ServerConfig struct, allowing limiters to integrate with gRPC servers easily. 

#### Link to tracking issue
Part of #9591.

#### Testing

TODO

#### Documentation

TODO

PR #12601, opened 2025-03-11
Limiter extension interface for memorylimiterextension draft

jmacd:
#### Description

This demonstrates how the existing memorylimiter extension will be upgraded to support the proposed limiter extension interface.

#### Link to tracking issue

Part of #9591.

#### Testing

TODO

#### Documentation

TODO


extension/xextension/limit/limit.go
```diff
@@ -0,0 +1,70 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package limit // import "go.opentelemetry.io/collector/extension/xextension/limit"
```
bogdandrutu, line 4: Based on auth case, I think this should be consistent and be named `extensionlimit`

extension/xextension/limit/limit.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Extension is the interface that storage extensions must implement
+type Extension interface {
+	extension.Extension
+
+	// GetClient will create a client for use by the specified
+	// component.  The component can use the client to limit
+	// admission to the pipeline.
+	GetClient(ctx context.Context, kind component.Kind, id component.ID) (Client, error)
+}
```
bogdandrutu, line 21: Some questions/asks here:
1. Need to be consistent with decision in https://github.com/open-telemetry/opentelemetry-collector/pull/12574, and not include the `extension.Extension` into this.
2. Do we need the "extra" client interface? Asking because we don't follow this pattern in `extensionauth` but we do this in `storage` extension. If we need that (there may be good reasons to have it), I would like to make the auth follow the same pattern. cc @mx-psi

extension/xextension/limit/limit.go
```diff
@@ -0,0 +1,70 @@
...
+	// approximate method to determine weight in bytes.
+	//
+	// The limiter is permitted to block the request. After making
+	// its decision, the return value is exclusively a function
+	// value or an error. Admit will return when one of the
+	// following events occurs:
+	//
+	//   (1) admission is allowed, or
+	//   (2) the provided ctx becomes canceled, or
+	//   (3) the limiter determines that the request should fail.
```
bogdandrutu, line 52: To clarify, in this case returned err is not nil.

extension/xextension/limit/nop_client.go
```diff
@@ -0,0 +1,20 @@
...
+package limit // import "go.opentelemetry.io/collector/extension/xextension/limit"
+
+import "context"
+
+type nopClient struct{}
+
+var nopClientInstance Client = &nopClient{}
+
+// NewNopClient returns a nop client
+func NewNopClient() Client {
```
bogdandrutu, line 13: We usually have this in a `limittest` package/module. Please follow that model.

bogdandrutu (commented):
Very nice. Thank you

extension/xextension/limit/limit.go
mx-psi: It makes more sense to me, without further context, to use the approach on my PR: if multiple instances of the client are required then the user can define multiple extensions (`limiter/1`, `limiter/2`...)

jmacd: I didn't mean to be inconsistent with #12574, sorry. That approach looks fine to me. 

As noted, the "extra" interface appears valuable to me in the sense that static context (e.g., component metadata) can be used to specialize the limiting conditions at start time similar to `x/storage`. See my comment in https://github.com/open-telemetry/opentelemetry-collector/issues/12603#issuecomment-2714189817 on this topic. Today's `configgrpc` and `confighttp` helpers do not include the component metadata, so to be realistic I removed the extra parameters. If we remove the extra parameters and do not arrange to pass component metadata (e.g., signal kind, component ID) via context, then there is no reason for the extra interface.

See I opened several drafts--the copy you are commenting on, here, is an older iteration of mine, where I tried to follow the `x/storage` model before I realized it would not work. In #12600 the interface became simply `GetLimiter(ctx)`, see https://github.com/open-telemetry/opentelemetry-collector/pull/12600/files#diff-8be8b8c994bb03af15a8c1a7cc8953d44053c7b40726d0ec99f365ca29eee608R19.

jmacd:
See https://github.com/open-telemetry/opentelemetry-collector/issues/12603#issuecomment-2715417816.

PR #12613, opened 2025-03-11
[service] Allow running the Collector without any pipelines

evan-bradley:
#### Description

Adds a feature gate that allows starting the Collector without running any pipelines. This will mainly be used for OpAMP, where the Collector may need to be started with only extensions


evan-bradley:
I decided to go with only a feature gate instead of a command line option because adding the command line option meaningfully complicates the code for what is in my opinion not sufficient gain. I think we could either:

1. Abuse the feature gate system a bit and leave this as an alpha-stage feature gate for the indefinite future.
2. Move to print warnings or similar when no pipelines are enabled. I don't think this would likely be a big deal.

otelcol/config.go
```diff
@@ -50,13 +51,13 @@ func (cfg *Config) Validate() error {
...
+	if !pipelines.AllowNoPipelines.IsEnabled() && len(cfg.Receivers) == 0 {
 		return errMissingReceivers
 	}
 
 	// Currently, there is no default exporter enabled.
 	// The configuration must specify at least one exporter to be valid.
-	if len(cfg.Exporters) == 0 {
+	if !pipelines.AllowNoPipelines.IsEnabled() && len(cfg.Exporters) == 0 {
 		return errMissingExporters
 	}
```
evan-bradley, line 62: I wasn't looking to make this change in this PR, but I think we should consider removing these checks, since validating the pipelines configuration will catch the same error.

jmacd (approved):

service/pipelines/config.go
```diff
@@ -25,14 +25,20 @@ var (
 		featuregate.WithRegisterFromVersion("v0.112.0"),
 		featuregate.WithRegisterDescription("Controls whether profiles support can be enabled"),
 	)
+	AllowNoPipelines = featuregate.GlobalRegistry().MustRegister(
+		"service.AllowNoPipelines",
+		featuregate.StageAlpha,
+		featuregate.WithRegisterFromVersion("v0.122.0"),
+		featuregate.WithRegisterDescription("Allow starting the Collector without starting any pipelines."),
+	)
```
bogdandrutu, line 33: Is the plan to eventually stabilize this, or keep it as an option forever?

evan-bradley: I'm not sure yet. Ideally we eventually stabilize it, but I'm not sure how we want to enable this going forward. I figure for now it's okay to just have the feature gate and we can adjust course later.

mx-psi: I think there is a reasonable chance that we would stabilize this. The main downside of stabilizing this is people getting confused about how to configure the Collector, so I guess we could think about some further validation (say, allowing this kind of configuration only when you have configured extensions that use OpAMP?), but that's something that we can figure out over time

mx-psi (approved):

service/pipelines/config.go
evan-bradley: @bogdandrutu Is this okay with you? I'd like to get this in this week (before Monday's release) so I can have the OpAMP Supervisor start using it.

bogdandrutu: It is ok...

bogdandrutu (approved):

yurishkuro:
why not a flag in the config?
```
service:
  extensions: [...]
  pipelines:
    disabled: true
```
This would be useful for Jaeger where `jaeger-query` also needs only extensions, not the pipeline. Currently we're using `nop` components to disable the pipelines.

mx-psi:
> why not a flag in the config?

I think this is something we can discuss before we mark this feature gate as beta. I filed #12646 to continue the discussion

PR #12616, opened 2025-03-11
[chore] Reduce port contention in service tests

evan-bradley:
#### Description

I noticed the service tests were a little flaky due to using the default Prometheus port. This should at least reduce the frequency where this happens if not eliminate it.

I did two things to help elimination contention for the port:
1. Disable the Prometheus server in tests where it isn't needed.
2. Ensure the port is randomized in tests where it is needed.


codeboten (approved):
👍🏻

Issue #12316, opened 2025-02-06
[telemetry] extend collector telemetry with additional attributes


vigneshshanmugam:
### Is your feature request related to a problem? Please describe.

The current [telemetry configuration](https://github.com/open-telemetry/opentelemetry-collector/blob/50b76b95bf3578ce960bab664bf7809612da044c/service/telemetry/config.go#L30) in the collector https://github.com/open-telemetry/opentelemetry-collector/issues/4970 doesnt allow the use of custom providers (e.g., Meter, Trace) or the addition of custom dimensions that can be dynamically injected to the components collecting the internal telemetry data. This limitation makes it challenging to use the data in a multi-tenant environment or use it impact analysis etc. 

### Describe the solution you'd like

Given there is already an open issue to extend the Telemetry factory to accept various providers, I would like to describe the solution only for extending the collected metrics/trace/logs data with additional dimensions via attributes. 

- Extend the [telemetry configuration](https://github.com/open-telemetry/opentelemetry-collector/blob/50b76b95bf3578ce960bab664bf7809612da044c/service/telemetry/tracer.go#L71) to support `baggage propagator`
- Add the Contents from the baggage to the produced spans/metrics/logs as attributes. Common place to start adding them would be to the [ObsReport](https://github.com/open-telemetry/opentelemetry-collector/blob/6cae4115b392d90b0fbaee5f6538a96c25ffddca/receiver/receiverhelper/obsreport.go#L167) package that is the base for capturing telemetry for Receiver, Processors and Exporters. 
```go
	for _, member := range baggage.FromContext(ctx).Members() {
		if filter(member) {
			span.SetAttributes(attribute.String(member.Key(), member.Value()))
		}
	}
// similarly for metrics to be recorded
```
### Alternatives considered

I am not sure if there is any easy way to propagate them across all components, For custom components it was possible via setting `TelemetrySettings`. 

Happy to send a PR if there is a consensus.

vigneshshanmugam:
@mx-psi Since you wrote the original proposal for the Telemetry extensions, Adding you incase if you have any thoughts on this approach. Thanks.

mx-psi:
@vigneshshanmugam What about setting the resource https://github.com/open-telemetry/opentelemetry-collector/blob/3d5171244b4bcebe2e3b903745eb5de684659d1f/service/telemetry/config.go#L35-L39 ?

vigneshshanmugam:
[...]

PR #12617, opened 2025-03-12
Inject component-identifying scope attributes

### LINKED ISSUE #12117: [receiverhelper] Implement new component-identifying attributes

### LINKED ISSUE #12213: [exporterhelper] Implement new component-identifying attributes

### LINKED ISSUE #12217: System for managing own telemetry attributes within pipeline components

jade-guiton-dd:
#### Description

Fork of #12384 to showcase how component attributes can be injected into scope attributes instead of log/metric/span attributes. See that PR for more context.

To see the diff from the previous PR, filter changes starting from the "Prototype using scope attributes" commit.

#### Link to tracking issue
Resolves #12217 
Also incidentally resolves #12213 and resolves #12117

#### Testing
I updated the existing tests to check for scope attributes, and did some manual testing with a debug exporter to check that the scope attributes are added/removed properly.


.chloggen/metric-span-component-attributes.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - `otelcol.component.id`
+  - `otelcol.pipeline.id`
+  - `otelcol.signal`
+  - `otelcol.signal.output`
+  These attributes are defined in the [Pipeline Component Telemetry RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md#attributes),
+  and identify the component instance from which the telemetry originates.
+  These attributes are added automatically without changes to component code by wrapping the
+  `MeterProvider`, `TracerProvider`, and `Logger` provided to them.
+  Components will be able to opt out of one or more of these attributes, but that API is not yet
+  publicly available.
```
mx-psi, line 30: I would remove the implementation details and forward-looking statements
```suggestion
  These attributes are added automatically without changes to component code.
```

internal/telemetry/componentattribute/logger_zap.go
```diff
@@ -0,0 +1,50 @@
...
+		scopeName: scopeName,
+		wrapCore:  wrapCore,
+	}
+}
+
+func ZapLoggerWithAttributes(logger *zap.Logger, attrs attribute.Set) *zap.Logger {
+	return logger.WithOptions(zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+		sc, ok := c.(serviceCore)
+		if !ok {
+			// Logger does not use OTel under the hood, cannot change Logger parameters
```
mx-psi, line 45: Does it make sense to add a log about this?

internal/telemetry/telemetry.go
```diff
@@ -4,22 +4,43 @@
...
-	}
-	if _, ok := ts.Logger.Core().(coreWithout); !ok {
-		return ts.Logger
-	}
-	return ts.Logger.WithOptions(
-		zap.WrapCore(func(from zapcore.Core) zapcore.Core {
-			return from.(coreWithout).Without(fields...)
-		}),
-	)
+type TelemetrySettings struct {
```
mx-psi, line 16: Add a comment that says "THIS IS PART OF THE PUBLIC API PLZ DO NOT CHANGE IN ANY BREAKING WAY"?

internal/telemetry/componentattribute/logger_zap.go
jade-guiton-dd: I don't think so. This is not an error condition, as logs are only exported with OTel [if some processors are set in the config](https://github.com/open-telemetry/opentelemetry-collector/blob/cfb5488c6edbd0454e90ab0017a83bb4d96ba825/service/telemetry/logger.go#L44). When that isn't the case, we get a pure Zap-based console logger.

internal/telemetry/telemetry.go
jade-guiton-dd: Done, with slightly different wording 😆

internal/telemetry/componentattribute/logger_zap.go
jade-guiton-dd: As discussed in the Collector stability meeting, I'm working on adding the attributes to the console logger output as well, which I think will require a core wrapper in all cases. I'll add a warn-level log if the wrapper isn't found.

jade-guiton-dd: In the end it was a bit tricky so I didn't do it, but I'm thinking relying on our tests might be enough? Since we have tests in `componentattribute` checking that attributes are set properly after multiple calls to `ZapLoggerWithAttributes`, and a test in `service/telemetry` checking that the instantiated logger is the expected type

jmacd (approved):

internal/telemetry/componentattribute/logger_zap.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return c
+}
+
+type consoleCoreWithAttributes struct {
+	zapcore.Core
+	from zapcore.Core
+}
+
+var _ coreWithAttributes = consoleCoreWithAttributes{}
```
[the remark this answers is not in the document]
jade-guiton-dd: I didn't think about it too hard, but considering we embed zapcore.Core, which is already a pointer, I wanted to avoid an unnecessary layer of indirection. Are there benefits to using a pointer here?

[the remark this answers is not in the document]
jade-guiton-dd: Thinking about it again, since these structs will be used behind the zapcore.Core interface and are too big to be inlined, it wouldn't actually matter wrt indirection.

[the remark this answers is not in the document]
mx-psi: If there is no good reason to not use pointers I would prefer using pointers. It can be a bit confusing (e.g. if we were to modify the code to do something that modifies member fields like it happens on [the example on the Go tour](https://go.dev/tour/methods/4))

internal/telemetry/componentattribute/logger_zap.go
```diff
@@ -0,0 +1,110 @@
...
+
+func NewConsoleCoreWithAttributes(c zapcore.Core, attrs attribute.Set) zapcore.Core {
```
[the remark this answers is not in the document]
mx-psi: It seems to me that the function is agnostic to where the core is actually writing and console is just one of the use cases. I would have understood it better if console was not in the name and we clarified the intended use cases for each constructor in godoc comments instead

internal/telemetry/componentattribute/logger_zap.go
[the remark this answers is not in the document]
jade-guiton-dd: Fair enough, I'll change it.

internal/telemetry/componentattribute/logger_zap.go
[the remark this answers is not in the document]
jade-guiton-dd: That makes sense. I'm not sure what to call it if not based on the use case though... Any ideas? `zapCoreWithAttributes`? `attributesAsFieldsWrapper` vs. `attributesInScopeWrapper`? (although the latter also has the Tee feature, so that might be confusing)

internal/telemetry/componentattribute/logger_zap.go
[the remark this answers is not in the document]
jade-guiton-dd: I think using a pointer actually introduces some confusion, because even if we only implement `withAttributeSet` on `*core`, `core` will still implement `zapcore.Core`, so if we accidentally return the raw struct at some point, we get no type errors, but `tryWithAttributeSet` will silently do nothing. That's what the tests are for, but still.

internal/telemetry/componentattribute/logger_zap.go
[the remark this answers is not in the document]
jade-guiton-dd: I added some comments to the different core wrappers to clarify what they do and the intended use, hopefully that should alleviate the questionable naming.

mx-psi:
We discussed offline adding a feature gate for this and all other internal telemetry related changes, I intend to merge this once the comments are addressed and the feature gate has been added

internal/telemetry/componentattribute/logger_zap.go
```diff
@@ -0,0 +1,146 @@
...
+	level       zapcore.Level
+}
+
+var _ coreWithAttributes = (*otelTeeCoreWithAttributes)(nil)
+
+// Wraps a Zap core in order to copy logs to a [log.LoggerProvider] using [otelzap]. For the copied
+// logs, component attributes are injected as instrumentation scope attributes.
+//
+// This is used when service::telemetry::logs::processors is configured.
+func NewOTelTeeCoreWithAttributes(consoleCore zapcore.Core, lp log.LoggerProvider, scopeName string, level zapcore.Level, attrs attribute.Set) zapcore.Core {
```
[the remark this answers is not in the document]
jade-guiton-dd: It comes from `cfg.Logs.Level`. The `panic` is part of the original code from `service/telemetry/logger.go`. It should never be triggered in normal operation. (It would require both the console core and the otelzap core to be restricted to a less verbose level than requested; the first is created with level `cfg.Logs.Level` as well, and the second I don't think has any level restrictions.)

.chloggen/metric-span-component-attributes.yaml
```diff
@@ -0,0 +1,34 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. otlpreceiver)
+component: service
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Internal telemetry uses scope attributes to identify the source component
```
mx-psi, line 10: Needs an update to mention the feature gate

mx-psi (approved):

.chloggen/metric-span-component-attributes.yaml
jade-guiton-dd: Done. I also added a note about the change to logs, and about this relying on instrumentation scope attribute support.

receiver/otlpreceiver/otlp.go
```diff
@@ -54,7 +54,7 @@ type otlpReceiver struct {
 // responsibility to invoke the respective Start*Reception methods as well
 // as the various Stop*Reception methods to end it.
 func newOtlpReceiver(cfg *Config, set *receiver.Settings) (*otlpReceiver, error) {
-	set.Logger = telemetry.LoggerWithout(set.TelemetrySettings, componentattribute.SignalKey)
+	set.TelemetrySettings = telemetry.WithoutAttributes(set.TelemetrySettings, componentattribute.SignalKey)
```
vigneshshanmugam, line 57: @jade-guiton-dd Apologies for asking on a old PR, Was there any specific reason for not including the signal info on the exposed metrics from OTLP receiver? This would be really useful to see the amount of data ingested based on the signal type. 

We did already have this info exposed in the older metrics via `otelcol_receiver_accepted_log_records`, `otelcol_receiver_accepted_metric_points` and `otelcol_receiver_accepted_spans`. Trying to understand if we have a plan to add them down the line? Thanks.

jade-guiton-dd: The OTLP receiver is internally a single object, even when configured in multiple pipelines for multiple signals. For that reason, the telemetry it emits can't easily be associated with a single signal, so it removes the "otelcol.signal" attribute from its set of attributes on startup. If we didn't do that, all telemetry from the component would be associated with whichever signal pipeline happened to be created first, which would not be helpful.

However, the OTLP receiver _could_ manually add back a signal attribute on specific metric points which *are* associated with a specific signal. But I don't believe this is currently needed:
- The older `otelcol_receiver_X` metrics (which aren't going anywhere for the foreseeable future) already differentiate between signals in their name
- The new metrics emitted by [pipeline auto-instrumentation](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md#auto-instrumented-metrics) (implemented in a later PR) use the original attribute set of the component before startup, which includes "otelcol.signal".

Do you have any examples of internal metrics emitted by the OTLP receiver which are lacking association with a specific signal (and which _could_ be associated with one despite the singleton architecture)?

vigneshshanmugam: Thanks for the detailed answer, I was going off based on this PR and upon testing the different metrics/logs, I do see the `otelcol.signal` being present in all of the emitted telemetry - even the custom ones generated via mdatagen which is super cool. Thanks for making that happen 👍🏽 

One thing I noticed is since we are currently treating [Middlewares](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configmiddleware/configmiddleware.go) as part of the Extension interface, some of the pipeline attributes like `signal and outcome` are missing from the reported metrics. Should we treat them similar to Receivers as they are running as part of the Receiver end?

jade-guiton-dd: I'm not extremely familiar with the middleware interface considering how new it is (are there even any implementations of it yet?), but I think this would be of questionable use and difficult to accomplish:
- Because middlewares act at the HTTP/gRPC request level, there's no generic reliable way to know which signal (if any) they're processing. This is determined later by the receiver, after the request has been processed by the middleware. The only case where I think this would be doable is if the receiver only handles a single type of signal, in which case `otelcol.signal` is much less useful anyway.
- We only use the `outcome` attribute on auto-instrumented pipeline metrics, not arbitrary receiver telemetry, because the auto-instrumentation layer is at the right place to know whether the next component succeeded or not. I think adding a similar instrumentation layer inside middlewares would be difficult, but I'm not familiar enough with the middleware API to tell for sure.

However, I think we could make a stronger case for adding an attribute to middleware telemetry to know which receiver instance it's used in. This would be doable, though we would need to redesign the middleware API to pass in a new `TelemetrySettings` with the appropriate attributes for each call to `GetHTTPHandler` / `GetGRPCServerOptions`.

vigneshshanmugam: Thanks for the details again. Just to give some context, I am working with the Middleware based extension to add additional attributes to the custom telemetry records without relying on some of the proposed alternatives like Baggage propagation, Processors etc - https://github.com/open-telemetry/opentelemetry-collector/issues/12316 

> This is determined later by the receiver, after the request has been processed by the middleware. 

Makes sense 👍🏽. 

> However, I think we could make a stronger case for adding an attribute to middleware telemetry to know which receiver instance it's used in

That would be useful since middleware extension can basically be run on any HTTP/gRPC receivers. 

Do you think its worth creating a issue for this so we can move the discussion?

jade-guiton-dd: Yes, I think that would make sense. It would be good to make the contributors working on the middleware interface aware that this need exists.

Issue #12603, opened 2025-03-11
Outline steps to add "limiter" extension component


jmacd:
### Component(s)

extension/memorylimiter

### Describe the issue you're reporting

Following a proposal roughly sketched in https://github.com/open-telemetry/opentelemetry-collector/issues/9591#issuecomment-2675430567, this is a concrete set of steps to upgrade the Collector's capabilities with regards to rate-limiting and admission-limiting requests.

This proposal has the following major steps:

1. Add `extension/xextension/limiter` package, where `limiter.Extension` and `limiter.Limiter` types are modeled on the `x/storage` package
2. Add `config/configlimiter` package, similar to the `configauth` package in exporting a type for naming extension components, performing a map lookup and type check over host extensions
3. Modify `configgrpc` adding `Limiters []configlimiter.Limitation`
- (a) w/ unittests in configgrpc
- (b) end-to-end test in `extension/admissionlimiterextension`
4. Modify `confighttp` adding `Limiters []configlimiter.Limitation`
- (a) w/ unittests in confighttp
- (b) end-to-end test in `extension/admissionlimiterextension`
5. Update `memorylimiterextension` to support the new `limiter.Extension` interface
6. Add `admissionlimiterextension` supporting the new `limiter.Extension` interface modeled on the OTel-Arrow admission controller
- (a) Initial skeleton
- (b) Copy implementation from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/otelarrow/admission2/README.md
7. Remove `internal/otelarrow/admission2` from collector-contrib, update `otelarrowreceiver` to use limiter extensions.

Here are the major steps outlined as rough drafts:

Includes step (1) and (2). These can be easily separated into two PRs, first extension then config.
https://github.com/open-telemetry/opentelemetry-collector/pull/12599

Includes (3) w/o tests plus the content above, look mainly at `configgrpc.go`
https://github.com/open-telemetry/opentelemetry-collector/pull/12600

Includes (5) w/o tests plus the first, https://github.com/open-telemetry/opentelemetry-collector/pull/12601

Includes (6a) w/o tests plus the first, https://github.com/open-telemetry/opentelemetry-collector/pull/12602

I am looking for maintainers and approvers to sign off on the general approach. I would send part (1) as a stand-alone change and proceed in roughly the order presented above.

mx-psi:
Is there a possibility to design some parts in a more generic way to cover the use cases discussed in #7441?

[...]

Issue #7441, opened 2023-03-28
Add support for "middleware extensions"


jpkrohling:
**Is your feature request related to a problem? Please describe.**

When developing a custom auth extension, @rlankfo mentioned he needs access to the request URI to determine the proper auth parameters (tenant, for instance). Instead of just adding the possibility of specifying the interceptors to use as part of the auth package and using the experience gathered over the past couple of years since implementing the auth capabilities, I think it might be helpful to add support for middleware extensions instead: middleware extensions would have the possibility of intercepting the incoming requests (HTTP and gRPC) and potentially doing things that are beyond auth (and possibly advanced auth scenarios, like @rlankfo's). Another example of a middleware extension would be @MovieStoreGuy's https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/6908. 

Depending on how we progress with this idea, server auth extensions could be implemented as a specialization of a middleware extension.

**Describe the solution you'd like**
I had a proposal for "authorizers" as part of @MovieStoreGuy's rate limiting extension proposal, and I would change that a bit, so that it reads middleware instead:

```yaml
extensions:
  oidc:

  ratelimiter:
    qps: 1000
  blocklist:
    source: https://example.com/blocklist.txt

receivers:
  otlp:
    protocols:
      grpc:
        auth:
          authenticator: oidc
        middlewares:
        - blocklist
        - ratelimiter


processors:

exporters:
  logging:

service:
  extensions: [oidc, ratelimiter]
  pipelines:
    traces:
      receivers: [otlp]
      processors: []
      exporters: [logging]
```

**Describe alternatives you've considered**
Solving @rlankfo's problem could also be done by adding the notion of Interceptors to the server auth extension interface, with the current interptors (http.Handler, gRPC's Unary and Stream interceptors) as default implementations, allowing auth extensions to override when needed via options (WithInterceptor in addition to the current WithAuthenticate). Another possible solution to the original problem would be to add request metadata (URL, for instance) to the client's context, which is made available later on to auth extensions.

codeboten:
Capturing some thoughts from the discussion in the SIG call on March 29:
- middleware/interceptors should not have access to the telemetry data

mx-psi:
[...]

PR #12633, opened 2025-03-13
Second draft: configmiddleware and extensionlimiter

jmacd:
#### Description

Second draft for Limiter extension w/ Middleware config as discussed in #12603 incorporating feedback from @axw and @bogdandrutu and following @dmitryax's example in #12574.

Relative to the first draft, this adds:
- Settings field passed to Acquire() for signal kind, can be extended w/ caller's component ID
- Weight field passed to Acquire() supports limiting by request count, item count, and bytes count
- Middleware have multiple types as an indicator of what they do.

As an example (and to emphasize that previous examples were incorrect), note that to follow the `configauth` pattern here means each entry in a middleware list has a nested field.

```
receivers:
  otlp:
    protocols:
      grpc:
        // ...
        middleware:
        - interceptor: blockinterceptor
        - limiter: ratelimiter
        - limiter: admissionlimiter
        - interceptor: decoratorinterceptor
```

If we don't follow this example, each of the "limiter" and "interecptor" fields above will become "middleware" which feels not right.

#### Link to tracking issue

Part of #7441 
Part of #9591 
Part of #12603 

#### Testing

TODO

#### Documentation

👍


extension/extensionlimiter/limiter.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Potential additions:
+	// - component.ID of the calling component
+	// - protocol/method of the request
+}
+
+// Weight is the argument to a limiter. Any of three fields may be
+// set; if a field is zero, no limit will apply. Receivers are meant
+// to call a limiter one or more times as this information becomes
+// available.
```
sfc-gh-bdrutu, line 36: A bit confused about the "one or more times". How do a limiter deal with that? Let's say we want a limiter to do N concurrent requests, how do I know that is the same request? Do you expect to call Done before calling second time?

extension/extensionlimiter/limiter.go
```diff
@@ -0,0 +1,138 @@
...
+//	  Requests: 1,
+//	  Records: 0,
+//	  Bytes: len(compressedData),
+//	})
+//
+// Later, when the data is uncompressed and the actual record count
+// is known, the receiver should call the limiter again:
+//
+//	rel, err := Acquire(ctx, Settings{...}, Weight{
+//	  Requests: 0,
```
sfc-gh-bdrutu, line 53: Why request is 0?

extension/extensionlimiter/limiter.go
```diff
@@ -0,0 +1,138 @@
...
+//	  Bytes: len(uncompressedData),
+//	})
+//
+// or, if the code has dropped all references to the compressed data
+// it could call instead:
+//
+//	rel, err := Acquire(ctx, Settings{...}, Weight{
+//	  Requests: 0,
+//	  Records: recordNum,
+//	  Bytes: len(uncompressedData) - len(compressedData),
```
sfc-gh-bdrutu, line 64: Seems to be a sort of "sum" involved here, which is not clear from the comments.

extension/extensionlimiter/limiter.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// Limiters should ignore zero values in the Weight struct and they
+// should ignore values they are not concerned with limiting.
+// Therefore, in the examples above, if a limiter is based on request
+// count or record count, they are guaranteed one call to the limiter
+// will be a no-op because either Requests or Records is zero in one
+// of the two calls. On the other hand, for this example, if the
+// limiter is based on bytes there will be two real calls to the
+// limiter, to account for the two non-zero Bytes fields.
+type Weight struct {
```
axw, line 75: I'm finding it hard to reason about this. Are they all part of the same limit? It seems to me we're looking at multiple distinct limits here, which would be applied in order:

 1. HTTP/gRPC requests
 2. HTTP/gRPC request body size, i.e. potentially-compressed bytes
 3. Signal-specific record count
 4. Estimated size of decoded signals

Would it be simpler to have a single numeric weight, and distinct limiters for each unit? e.g.

```go
type Limiter interface {
	// Acquire requests admission of a predefined unit, such as "requests", "records", or "bytes".
	Acquire(ctx context.Context, settings Settings, n int64) (ReleaseFunc, error)
}
```

Then you would configure a separate limiter for each unit.

jmacd: @axw Thank you. Between your comment and the three above @sfc-gh-bdrutu (cc @bogdandrutu 🤣), I realize that more work is needed. 

@axw as you pointed out here, there are 4 ways we can limit a pipeline at the receiver, including compressed body size, request number, item number, and "resident memory size". I am excluding "(uncompressed) body size" for reasons discussed below: mainly that "resident memory size" is better measurement. Instead of having a `Weight` with multiple fields, following your suggestion, we can have a single-valued weight with an enum indicating which kind of weight it is. What I would like to preserve, is that callers can call the limiter fewer than 4 times.

@bogdandrutu About your questions "why 0?", "why one or more times?". I think in most cases a limiter will be called once or twice. In some cases, we can derive all four values from a single interceptor (especially OTLP signals); in other cases, the interceptor may report compressed bytes and request count, but a later stage computes the item-count and resident-memory-size.

In pseudo-code, I see this looking like:

```
type Variety int

const (
    CompressedBytesCount Variety = iota
    RequestCount
    ItemCount
    ResidentBytesCount
)

type Weight struct {
    Variety Variety
    Value uint64
}

type Limiter interface {
    Acquire(ctx context.Context, settings Settings, weights []Weight) (ReleaseFunc, error)
}
```

Some details, as I understand them:

An interceptor middleware can be used to measure and limit compressed network bytes or request count.

A limiter extension can be used to measure and limit item count and resident memory size.

Note: Uncompressed network bytes, as calculated by stream interceptors in particular, is not always useful, especially not in the `otelarrow` case I've mentioned in the past. For a streaming protocol, the "uncompressed" size as seen by gRPC is not a good estimate for the uncompressed size in an OTLP representation.

As an aside, OTel-Arrow exporter and receiver have integrated middleware used for monitoring compression rate, which requires computing a ratio of compressed to uncompressed bytes. For the reason stated above, we use different middleware configuration for the unary and stream cases (i.e., OTLP vs. OTAP); in some cases, uncompressed bytes is reported from gRPC, in other cases it is reported explicitly using resident memory size. Roughly speaking, this demonstrates how we can derive a compression-rate signal from the same list of weights presented to the limiter: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/2042f437c45db1f53694bd574c2bfa33ec9dae94/internal/otelarrow/netstats/netstats.go#L65. I see a way to address https://github.com/open-telemetry/opentelemetry-collector/issues/6638 using the `limitermiddlewareextension`, consequently: the limiter sees both compressed bytes and resident memory size signals, i.e. the necessary inputs.

jmacd (commented):
Thank you reviewers. I will close this PR in 24 hours. I will open another with the idea described first by @axw below, which is (1) to separate limiters from middleware, with a dedicated `limitermiddleware` extension to apply limiters as middleware, (2) to use a slice of weights in the Acquire function(), giving receivers flexibility to handle some compressed-bytes and request-coutn limits in interceptors while counting items and resident memory size at the receiver level.

PR #12638, opened 2025-03-15
[builder] Rewrite replace paths to be absolute in generated go.mod

milas:
#### Description
Convert all relative paths to be absolute for `replace` directives the same way that is done for modules/components.

This ensures that the `output_path` isn't tied to the location of the builder YAML config.

For example:
```
/
├── work
│   ├── mycollector
│   │   └── builder.yaml
│   ├── mylib
│   │   └── go.mod
└── dist
    └── my-collector
```

**OCB YAML - `/work/mycollector/builder.yaml`**
```yaml
replaces:
 - registry.test/somemodule => ../mylib
```

**Build**
```shell
cd /work/mycollector
DIST_OUTPUT_PATH=/dist ocb --config ./builder.yaml
```

**Before these changes**
* Generated `go.mod`'s replace would end up resolving to `/mylib`

**After these changes**
* Generated `go.mod`'s replace resolves to `/work/mylib`

#### Link to tracking issue
n/a

#### Testing
Added unit test

#### Documentation
Copious comments in code, nothing user-facing


milas:
RE: `stale` label - Happy to make any changes / rebase on `main` based on review if of interest to maintainers. Note that I'm not sure what's up with codecov, as there are unit test cases included. Otherwise, will let stalebot close.

atoulme:
It's unclear if we would like to adopt this change. Please take a look at the checks CI failure - this is generating a path that is absolute and therefore no longer portable across machines. When would this approach make sense?

You also need to add a changelog entry for this PR. `make chlog-new` to generate the yaml template.

milas:
Thanks for the reply!

> this is generating a path that is absolute and therefore no longer portable across machines

Correct. I agree that behavior is not ideal, but fwiw it's the status quo:

https://github.com/open-telemetry/opentelemetry-collector/blob/e367eb17ff21f90b99dc15e686087e35626870ec/cmd/builder/internal/builder/config.go#L246-L249

Today, if you do:
```yaml
providers:
  - gomod: example.com/module v1.0.0
    path: ./mymodule
```

The resulting `go.mod` produced by `ocb` _will_ have an absolute/non-portable path in the generated `replace` directive.

However, if you do:
```yaml
replaces:
  - example.com/other => ./myvendor
```

The resulting `go.mod` produced by `ocb` will NOT have an absolute path for _that_ `replace` directive.

In practice, this will break things:
```
/home/milas/mycollector/
  mymodule/
    go.mod
  myvendor/
    go.mod
  builder.yaml
```

When you run `ocb`, you'll end up with `/home/milas/mycollector/dist/go.mod`:
```
replace example.com/module v1.0.0 => /home/milas/mycollector/mymodule
replace example.com/other => ./myvendor
```

However, the correct _relative_ path would be `../myvendor` here, so it does not work as expected currently IMO, even non-portably.

Arguably, OCB could instead try to generate relative paths for the "dist" path based on the `builder.yaml` path: I did not attempt that for this since the module/component replacement was already using absolute/non-portable paths for the local replacement use case.

As a workaround, if you _know_ the dist path will be a child, you can manually make the paths `../` instead of `./`, though this inconsistent with the behavior of other paths. 

In particular, that breaks my CI build flow (in a Docker container):
 * Run `DIST_OUTPUT_PATH=/dist ocb --config builder.yaml /src/builder-config.yaml`
 * From `/dist`, run `go build`

Currently, I end up with a `go.mod` where my components have `replace ... => /src/mymodule` (works) but the generic "replace" has `replace ... => ./myvendor` (does not work).

PR #12576, opened 2025-03-06
[chore] replace deprecated tenv with usetesting

mattsains:
#### Description
There are a bunch of warnings in the lint output about tenv being deprecated, and moving to usetesting. Reference: https://golangci-lint.run/usage/linters/#tenv

#### Link to tracking issue
none

#### Testing
Ran `make golint` and `make all`

PR #12639, opened 2025-03-16
[chore] Fix ARM unit test

sincejune:
#### Description
This PR fixes a test in ARM by using a shorter file name in socket (rolling back the behavior introduced in #12576 ). Seems that we ran into an issue similar to https://github.com/golang/go/issues/6895

#### Link to tracking issue
n/a

#### Testing
n/a

#### Documentation
n/a


config/configgrpc/configgrpc_test.go
```diff
@@ -1168,7 +1168,7 @@ func sendTestRequest(t *testing.T, gcs ClientConfig) (ptraceotlp.ExportResponse,
 
 // tempSocketName provides a temporary Unix socket name for testing.
 func tempSocketName(t *testing.T) string {
-	tmpfile, err := os.CreateTemp(t.TempDir(), "sock")
```
bogdandrutu, line 1171: Why? Please add a comment why `t.TempDir()` is not good.

bogdandrutu (approved):

config/configgrpc/configgrpc_test.go
sincejune: @bogdandrutu @codeboten I believe the issue stems from the path length limit for sockets on Unix systems. The `os.TempDir()` function works because it generates a shorter file name compared to `t.TempDir()`. This link might be helpful: https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

atoulme (approved):

config/configgrpc/configgrpc_test.go
sincejune: updated

sincejune: updated

PR #12635, opened 2025-03-14
[exporter][batching] Enable batching based on size

sfc-gh-sili:
#### Description

This PR enable serialized batching based on size rather than item count.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector/issues/3262

PR #12641, opened 2025-03-16
[exporterbatcher] Update batcher to support serialized bytes based batching.

sfc-gh-sili:
#### Description

This PR
1. Updates batcher to use Size(sizerType) instead of itemsCount()
2. For logs, traces, and metrics, it also makes sure that the `cachedSize` is associated with the `sizerType`.

This will be followed by https://github.com/open-telemetry/opentelemetry-collector/pull/12635/files which enables turning on this feature from configuration.

#### Link to tracking issue

### DIFF
--- a/exporter/exporterhelper/internal/request/request.go
+++ b/exporter/exporterhelper/internal/request/request.go
@@ -15,6 +15,8 @@ type Request interface {
 	// sent. For example, for OTLP exporter, this value represents the number of spans,
 	// metric data points or log records.
 	ItemsCount() int
+	// ByteSize returns the size of the request in bytes.
+	ByteSize() int
 	// MergeSplit is a function that merge and/or splits this request with another one into multiple requests based on the
 	// configured limit provided in MaxSizeConfig.
 	// MergeSplit does not split if all fields in MaxSizeConfig are not initialized (zero).
--- a/exporter/exporterhelper/logs.go
+++ b/exporter/exporterhelper/logs.go
@@ -44,14 +44,16 @@ func NewLogsQueueBatchSettings() QueueBatchSettings {
 }
 
 type logsRequest struct {
-	ld         plog.Logs
-	cachedSize int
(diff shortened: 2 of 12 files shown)


exporter/exporterhelper/internal/request/request.go
```diff
@@ -31,6 +33,10 @@ type Request interface {
 	MergeSplit(context.Context, exporterbatcher.SizeConfig, Request) ([]Request, error)
 }
 
+type Sizer interface {
```
bogdandrutu, line 36: Where is this used?

exporter/exporterhelper/logs.go
```diff
@@ -27,14 +28,14 @@ var (
 type logsRequest struct {
 	ld         plog.Logs
 	pusher     consumer.ConsumeLogsFunc
-	cachedSize int
+	cachedSize map[exporterbatcher.SizerType]int
```
bogdandrutu, line 31: Better to have 2 values than this map which needs heap allocation and lookup is more expensive.

exporter/exporterhelper/logs.go
```diff
@@ -70,15 +71,26 @@ func (req *logsRequest) ItemsCount() int {
...
-	return req.cachedSize
+
+	var sz sizer.LogsSizer
+	switch sizerType {
+	case exporterbatcher.SizerTypeItems:
+		sz = &sizer.LogsCountSizer{}
+	case exporterbatcher.SizerTypeBytes:
+		sz = &sizer.LogsBytesSizer{}
+	default:
+		return -1
```
bogdandrutu, line 86: this design seems a bit off to return -1 here. At least document this behavior.

exporter/exporterhelper/internal/request/request.go
sfc-gh-sili: Removed.

PR #12652, opened 2025-03-17
[chore] Rename disabled queue to waiting queue, implement size limit

bogdandrutu:
No need yet for changelog entry because this is not enabled, pending the new queue-batch config.


exporter/exporterqueue/waiting_queue.go
```diff
@@ -0,0 +1,121 @@
...
+
+type waitingQueue[T any] struct {
+	component.StartFunc
+	component.ShutdownFunc
+	consumeFunc ConsumeFunc[T]
+	sizer       sizer[T]
+	capacity    int64
+
+	mu           sync.Mutex
+	hasMoreSpace *cond
```
jmacd, line 38: is this sync.Cond? I'm missing a `type cond` somewhere, or is there a missing file?

bogdandrutu: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterqueue/cond.go

It is a fork of the sync.Cond that supports context cancellation.

bogdandrutu:
> If this were only the renaming, I'd agree it's a chore not needing a changelog. But the new condition variable... I would add a changelog about the size limit. If a user is using queue-disabled today, does the condition variable mean slightly more synchronization cost? I'm trying to imagine a user who might like to read that changelog.

Because the "size limit" is max int64, it will never get to the Wait() path, and then it means the Signal path if a if check, see from https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterqueue/cond.go:
```
// Signal wakes one goroutine waiting on c, if there is any.
// It requires for the caller to hold c.L during the call.
func (c *cond) Signal() {
	if c.waiting == 0 { // Nobody waits ever here.
		return
	}
	c.waiting--
	c.ch <- struct{}{}
}
```

PR #12660, opened 2025-03-18
Define the final stable config for QueueBatch component

bogdandrutu:
Add this as a separate config than exporterqueue.Config until the queue and batcher are changed to support this new config. Once approved, will merge and in sub-sequent PRs will change the components to accept this new config and expose it to the users.


exporter/exporterhelper/internal/queuebatch/config.go
```diff
@@ -0,0 +1,100 @@
...
+	// If blocking is enabled, the component will wait for space; otherwise, operations will immediately return.
+	Blocking bool `mapstructure:"blocking"`
+
+	// StorageID if not empty, enables the persistent storage and uses the component specified
+	// as a storage extension for the persistent queue.
+	// TODO: This will be changed to Optional when available.
+	StorageID *component.ID `mapstructure:"storage"`
+
+	// NumConsumers is the maximum number of concurrent consumers from the queue.
+	// TODO: This will also control the maximum number of shards, when supported:
```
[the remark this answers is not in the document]
sfc-gh-bdrutu: Correct. Do you have a suggestion on how to clarify that?

exporter/exporterhelper/internal/queuebatch/config.go
```diff
@@ -0,0 +1,100 @@
...
+	// TODO: This will be removed when Optional is available.
+	Enabled bool `mapstructure:"enabled"`
+
+	// WaitForResult determines if incoming requests are blocked until the request is processed or not.
+	// Currently, this option is not available when persistent queue is configured using the storage configuration.
+	WaitForResult bool `mapstructure:"wait_for_result"`
+
+	// Sizer configuration determines the type of size measurement used by this component.
+	// It accepts "requests", "items", or "bytes".
+	Sizer exporterbatcher.SizerType `mapstructure:"sizer"`
```
dmitryax, line 26: Do you want to use one field for `queue_size` and batch sizes going forward? This might be confusing given that we have `requests` for the queue and `items` for batcher by default. I think we shouldn't change that default and split this field instead to make it clear. Probably, just expose only batch sizer until the queue sizer is available for now?

exporter/exporterhelper/internal/queuebatch/config.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// QueueSize represents the maximum data size allowed for concurrent storage and processing.
+	QueueSize int `mapstructure:"queue_size"`
+
+	// Blocking configuration dictates the behavior when the component's TotalSize limit is reached.
+	// If blocking is enabled, the component will wait for space; otherwise, operations will immediately return.
```
dmitryax, line 32: ?
```suggestion
	// If blocking is enabled, the component will wait for space; otherwise, operations will immediately return an error.
```

exporter/exporterhelper/internal/queuebatch/config.go
sfc-gh-bdrutu: >Do you want to use one field for queue_size and batch sizes going forward? 

I would definitely prefer this. 

> This might be confusing given that we have requests for the queue and items for batcher by default. 

Batching is experimental and not enabled by default, so we can tune that.

> I think we shouldn't change that default and split this field instead to make it clear. 

Will make everything harder to resonate in my opinion that in the same config you have "sizes" in different parts that are using different sizers.

> Even without the default, we might want to give the flexibility to use different sizers, especially given that we won't always have all the options consistently available for both.

For queue we will have all for sure because we will implement them. For batch, yes we may have different supported (requests is always though).

> For now, we can probably just expose only the batch sizer until the queue sizer is available. 

As mentioned in the comment, this is the final config I want to get to, and will implement what is missing before then.

dmitryax: > Will make everything harder to resonate in my opinion that in the same config you have "sizes" in different parts that are using different sizers.

The goal sounds good to me. It's just unclear to me how to achieve that.

Do you want to keep the existing default of `items` for batcher and `requests` for the queue or you want to have one particular sizer by default for both?

> For queue we will have all for sure because we will implement them. For batch, yes we may have different supported (requests is always though).

What would be the behavior if user picks a sizer that is only available for the queue and enables batcher with some custom numbers there?

bogdandrutu: >What would be the behavior if user picks a sizer that is only available for the queue and enables batcher with some custom numbers there?

I think we should try to fail during config validation (or at most start time) if the sizer picked is not valid. I have some ideas on how to do that (for me that is implementation detail), but want to focus on how the config should look like for the user and then figure out how can I fail fast.

bogdandrutu: >The goal sounds good to me. It's just unclear to me how to achieve that. Do you want to keep the existing default of items for batcher and requests for the queue or you want to have one particular sizer by default for both?

I think in the end the default sizer for this component (for all sizes) should be "items", which has the best tradeoff between performance (cpu cost) and granularity.

dmitryax: > I think we should try to fail during config validation (or at most start time) if the sizer picked is not valid. I have some ideas on how to do that (for me that is implementation detail), but want to focus on how the config should look like for the user and then figure out how can I fail fast.

Ok. It means that some sizer options won't be available if users want to use both queue and batching. That sounds reasonable. We should probably implement all the sizers before migrating to this interface.

> I think in the end the default sizer for this component (for all sizes) should be "items", which has the best tradeoff between performance (cpu cost) and granularity.

Sounds reasonable to me. We need to think carefully about a migration approach here I guess.

exporter/exporterhelper/internal/queuebatch/config.go
```diff
@@ -0,0 +1,100 @@
...
+	"time"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/exporter/exporterbatcher"
+)
+
+// Config defines configuration for queueing and batching incoming requests.
+type Config struct {
+	// Enabled indicates whether to not enqueue and batch before exporting.
+	// TODO: This will be removed when Optional is available.
```
dmitryax, line 17: How `Optional` can replace this if we keep `Enabled: true` by default?

exporter/exporterhelper/internal/queuebatch/config.go
bogdandrutu: Done.

exporter/exporterhelper/internal/queuebatch/config.go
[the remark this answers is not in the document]
jmacd:
```suggestion
	// This applies across all configurations of the optional members above.
	// TODO: This will also control the maximum number of shards, when supported:
```

exporter/exporterhelper/internal/queuebatch/config.go
bogdandrutu: I was thinking that users will embed this in the config as optional. For example in otlpexporter we will have `Optional[queuebatch.Config] sending_queue`, the problem with that is there is no way to disable if otlpexporter enables it by default.

jmacd: I was wondering the same. In a parallel PR #12652 you are replacing the disabled behavior -- does that PR reflect how you will change the "disabled" case after Optional is available, in other words? Or, does disabling the queue/batch mean "synchronously call the exporter"?

jmacd (approved):

exporter/exporterhelper/internal/queuebatch/config.go
jmacd: Question: I wasn't sure which exporters would not support certain sizer options. Isn't it possible, for exporters that don't implement a given sizer option, to use a default implementation of that sizer? For requests and items, there is a natural definition. For bytes, when the exporter doesn't implement a specific sizer, couldn't we silently fall back on the OTLP sizer?

exporter/exporterhelper/internal/queuebatch/config.go
dmitryax: > I want to get to "disabling" means equivalent of the whole component not being present at all.

How disabling would look from user's perspective? By setting `sending_queue` to `nil`?

bogdandrutu: Not sure yet how the design of Optional will be, and if it supports opt-out or only opt-in. So, just to avoid more comments/questions will remove the TODO, but will keep this in my mind if there is an opportunity to use Optional for this.

exporter/exporterhelper/internal/queuebatch/config.go
bogdandrutu: > For bytes, when the exporter doesn't implement a specific sizer, couldn't we silently fall back on the OTLP sizer?

Not really, since splitting the data may not be a valid operation after the conversion to the exporter specific format.

exporter/exporterhelper/internal/queuebatch/config.go
```diff
@@ -0,0 +1,104 @@
...
+	// Sizer configuration determines the type of size measurement used by this component.
+	// It accepts "requests", "items", or "bytes".
+	Sizer exporterbatcher.SizerType `mapstructure:"sizer"`
+
+	// QueueSize represents the maximum data size allowed for concurrent storage and processing.
+	QueueSize int `mapstructure:"queue_size"`
+
+	// Blocking configuration dictates the behavior when the component's TotalSize limit is reached.
+	// If blocking is enabled, the component will wait for space; otherwise, operations will immediately return a retryable error.
+	Blocking bool `mapstructure:"blocking"`
```
dmitryax, line 32: If we have `wait_for_result`, should this be called something like `block_on_overflow`? I think `blocking` might be confused the the former one.

bogdandrutu: I like `block_on_overflow`, will do that.

dmitryax (approved):

PR #12665, opened 2025-03-18
[chore] Prepare release v1.28.1/v0.122.1

evan-bradley:
The following commands were run to prepare this release:

* make chlog-update VERSION=v1.28.1/v0.122.1
* make prepare-release PREVIOUS_VERSION=1[.]28[.]0 RELEASE_CANDIDATE=1.28.1 MODSET=stable
* make prepare-release PREVIOUS_VERSION=0[.]122[.]0 RELEASE_CANDIDATE=0.122.1 MODSET=beta


codeboten:
some failures in contrib....

evan-bradley:
If I had to guess, it's because the contrib tests are running on main and there was a recent update to a Prometheus dependency that appears to be used in the modules that are failing: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38746.

Are we comfortable skipping contrib tests here? Is there another way we should validate this against contrib?

songy23 (approved):
would be good to test and validate in the v0.122.x branch in contrib (if there is an easy way)

Issue #12668, opened 2025-03-18
aix builds failing


codeboten:
### Describe the issue you're reporting

See https://github.com/open-telemetry/opentelemetry-collector/actions/runs/13934523135/job/38999449578

codeboten:
Reopening this issue to keep it around in case anyone can address the problems in aix builds

PR #12669, opened 2025-03-18
[chore] disable aix builds as they're failing

codeboten:
This build can be re-enabled when #12668 is addressed.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -234,8 +234,9 @@ jobs:
           # on macOS: https://go.dev/doc/go1.15
           #- goos: darwin
           #  goarch: 386
-          - goos: aix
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -234,8 +234,9 @@ jobs:
           # on macOS: https://go.dev/doc/go1.15
           #- goos: darwin
           #  goarch: 386
-          - goos: aix
-            goarch: ppc64
+          # disabling this test until https://github.com/open-telemetry/opentelemetry-collector/issues/12668 is addressed
```
songy23, line 237: Do we still need this? Looks like http://github.com/open-telemetry/opentelemetry-collector/pull/12674 fixed it

PR #12686, opened 2025-03-20
Remove deprecated func types from exporterhelper

bogdandrutu:


bogdandrutu:
>LGTM modulo the changelog

If only I would have pushed the changes from my branch :)

